PR #701, opened 2017-02-27
Fix negative refcount in PyCapsule destructor

dean0x7d:
When running the tests with a debug build of Python, the following error occurs:
```
Fatal Python error: /pybind11/include/pybind11/pytypes.h:159 object at 0x7f4672c01100 has negative ref count -1
```

This is due to a `reinterpret_steal` in a PyCapsule destructor. Note that `reinterpret_borrow` also isn't appropriate because the destructor is called just before the capsule is to be deallocated. All reference counting should be avoided at that point. The fix in this PR just uses the raw C API.

Since this was only picked up by a debug build of Python itself, I figure it would be nice to have one of the CI configs cover it. The existing `DEBUG=1` config is modified to also use `python3-dbg` in addition to the debug build of `pybind11_tests`. This Travis config now reports the refcount error above. [See the log](https://travis-ci.org/dean0x7d/pybind11/jobs/205982087#L1031).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,9 +147,13 @@ install:
 - |
   # Install dependencies
   if [ -n "$DOCKER" ]; then
+    if [ -n "$DEBUG" ]; then
+      PY_DEBUG="python$PY-dbg python$PY-scipy-dbg"
+      export CMAKE_EXTRA_ARGS="${CMAKE_EXTRA_ARGS} -DPYTHON_EXECUTABLE=/usr/bin/python${PYTHON}dm"
+    fi
     docker exec --tty "$containerid" sh -c "for s in 0 15; do sleep \$s; \
       apt-get -qy --no-install-recommends $APT_GET_EXTRA install \
-        python$PY-dev python$PY-pytest python$PY-scipy \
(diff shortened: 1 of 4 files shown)


dean0x7d:
The inclusion of the debug packages in the `.travis.yml` is a bit crude. Let me know if there's a better solution for it.

I was sidetracked with a bizarre cmake build test failure -- with the debug build of Python and the imported pybind11 target, `NO_SYSTEM_FROM_IMPORTED` is required for CMake to ignore the default Python release headers and use debug instead.

wjakob:
This is great -- both the bugfix and using a debug Python on the CI bots (I didn't realize there was a package for that).

jagerman:
It might be worth also installing numpy and scipy debug packages, and adding Eigen debug flags to that build.

jagerman:
(though I'm not sure what needs to be done to actually use the debug packages for numpy/scipy)

dean0x7d:
Numpy and scipy are covered (`python$PY-scipy-dbg` is included) and Eigen turns on additional checks for `!defined(NDEBUG)`.

jagerman:
Aha, good (sorry for not looking at the commit; I'm stuck on my phone for the moment).

PR #702, opened 2017-02-28
Minor pytest maintenance

dean0x7d:
* Add `pytest.ini` config file and set default options there instead of in `CMakeLists.txt` (command line arguments).

* Change all output capture from `capfd` (filedescriptors) to `capsys` (Python's `sys.stdout` and `sys.stderr`). This avoids capturing low-level C errors, e.g. from the debug build of Python. (PR #701 changed only the global capture, but not the context managers which resulted in some "capture leaks". This PR sets both to `capsys` to avoid any issues.)

* Set pytest minimum version to 3.0 to make it easier to use new features. Remove conditional use of `excinfo.match()`.

* Clean up some leftover function-level `@pytest.requires_numpy`.

### DIFF
--- a/tests/pytest.ini
+++ b/tests/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+minversion = 3.0
+addopts =
+    # show summary of skipped tests
+    -rs
+    # capture only Python print and C++ py::print, but not C output (low-level Python errors)
+    --capture=sys
--- a/tests/test_numpy_array.py
+++ b/tests/test_numpy_array.py
@@ -264,7 +264,6 @@ def test_constructors():
     assert results["array_t<double>"].dtype == np.float64
 
 
-@pytest.requires_numpy
 def test_greedy_string_overload():  # issue 685
     from pybind11_tests.array import issue685
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -150,16 +150,20 @@ endif()
 
 # Make sure pytest is found or produce a fatal error
 if(NOT PYBIND11_PYTEST_FOUND)
-  execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import pytest" OUTPUT_QUIET ERROR_QUIET
-                  RESULT_VARIABLE PYBIND11_EXEC_PYTHON_ERR)
-  if(PYBIND11_EXEC_PYTHON_ERR)
-    message(FATAL_ERROR "Running the tests requires pytest.  Please install it manually (try: ${PYTHON_EXECUTABLE} -m pip install pytest)")
+  execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import pytest; print(pytest.__version__)"
+                  RESULT_VARIABLE pytest_not_found OUTPUT_VARIABLE pytest_version ERROR_QUIET)
+  if(pytest_not_found)
+    message(FATAL_ERROR "Running the tests requires pytest. Please install it manually"
+                        " (try: ${PYTHON_EXECUTABLE} -m pip install pytest)")
+  elseif(pytest_version VERSION_LESS 3.0)
+    message(FATAL_ERROR "Running the tests requires pytest >= 3.0. Found: ${pytest_version}"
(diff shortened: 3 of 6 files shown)


wjakob:
Very fancy. TIL about ``addopts``. I also never heard about the ``pytestmark`` variable -- is this an official pytest feature? (I only find ``@pytest.mark`` in searches)

dean0x7d:
`pytestmark` -- it's the last paragraph [here](http://doc.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules). @jagerman originally made the switch to global `pytestmark` in the big Eigen PR (https://github.com/pybind/pybind11/commit/2a75784420059a3d3661febf8f39b385a9b08abf) which looks much cleaner.

BTW [`pytest.importorskip`](http://doc.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency) is also an interesting feature which can be used as `suppress(ImportError)` + `pytestmark` in one line, but it messes up static analysis tools (obscures `import`) so I think it's better to avoid it.

wjakob:
Great. One more suggestion would be to add put pytest>=3.0.0 in the cmake message that asks the user to install pytest. If a smaller version is installed (e.g. on Anaconda, which ships an older version by default), it's quite surprising because entering the command does nothing.

dean0x7d:
The CMake script will now also check `pytest.__version__` and complain accordingly.

jagerman:
I've noticed some C++ output making its way into the test output recently, e.g.:

```
test_operator_overloading.py .
test_pickling.py ..
test_python_types.py ..### ExamplePythonTypes @ 0x55c604005760 destroyed
........s......
test_sequences_and_iterators.py ....
test_smart_ptr.py ### MyObject3 @ 0x55c60447b2d0 destroyed
### MyObject3 @ 0x55c604452f00 destroyed
### MyObject3 @ 0x55c603f222f0 destroyed
......
test_stl_binders.py .........
test_virtual_functions.py ### ExampleVirt @ 0x55c604478f00 destroyed
### ExampleVirt @ 0x55c6043d9610 destroyed
### ExampleVirt @ 0x55c6043dbd80 destroyed
...
```

Is this the "capture leaks" you referred to?

jagerman:
Could we also add some way to invoke pytest in verbose mode (i.e. with `-v`)?  Some of the numpy/eigen tests need it in order to get enough detail to see the test failure.

Ideally it would be nice to have it turned on with `make pytest VERBOSE=1` rather than a separate target, but I don't see a simple way to do that.

aldanor:
Or maybe, more generally, just allow to forward any args to pytest (e.g. selecting specific test, passing `-v`, or anything else).

dean0x7d:
> Is this the "capture leaks" you referred to?

Indeed. That's fixed here by switching everything to `capsys`.

> Or maybe, more generally, just allow to forward any args to pytest

It's already covered by pytest itself, see [`PYTEST_ADDOPTS`](http://doc.pytest.org/en/latest/customize.html#adding-default-options):

```bash
 make pytest PYTEST_ADDOPTS="-v -s"  # output all the things
```

tests/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+minversion = 3.0
+addopts =
+    # show summary of skipped tests
+    -rs
```
jagerman: This removed `w` from what was in the CMake version.  I assume that was intentional (I can't even find `w` in the pytest manpage; was it a deprecated option?), but just checking.

tests/test_numpy_array.py
```diff
@@ -264,7 +264,6 @@ def test_constructors():
     assert results["array_t<double>"].dtype == np.float64
 
 
-@pytest.requires_numpy
```
jagerman: Good catch; I think that was the result of a merge conflict.

jagerman:
This looks good to me.

tests/pytest.ini
dean0x7d: Yeah, deprecated. It's enabled by default in pytest >= 3.0.

wjakob:
Merged!

PR #703, opened 2017-02-28
Fix compilation of Eigen casters with complex scalars

dean0x7d:
Just a missing `typename`. The new test function produces the compilation error (the `std::complex` variant of `npy_format_descriptor` wasn't being compiled before).


wjakob:
Whoops, nice catch!

PR #704, opened 2017-02-28
fix python version check

eirrgang:
Commit 11a337f1 added major and minor python version
checking to cast.h but does not use the macros defined
via the Python.h inclusion. This may be due to an
intention to use the variables defined by the cmake
module FindPythonInterpreter, but nothing in the
pybind11 repo does anything to convert the cmake
variables to preprocessor defines.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -646,14 +646,14 @@ struct type_caster<std::basic_string<CharT, Traits, Allocator>, enable_if_t<is_s
     using StringType = std::basic_string<CharT, Traits, Allocator>;
 
     bool load(handle src, bool) {
-#if PY_VERSION_MAJOR < 3
+#if PY_MAJOR_VERSION < 3
         object temp;
 #endif
         handle load_src = src;
         if (!src) {
             return false;
         } else if (!PyUnicode_Check(load_src.ptr())) {
-#if PY_VERSION_MAJOR >= 3
+#if PY_MINOR_VERSION >= 3
             return false;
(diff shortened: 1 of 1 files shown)


dean0x7d:
Good catch! The name was most likely just accidentally inverted. Looks like the Python 2 code was used in both cases, but it's always `false` (dead code) in Python 3 so it went undetected.

GitHub web hooks are a being [a bit sluggish](https://status.github.com/messages), Travis hasn't started the test run yet.

include/pybind11/cast.h
```diff
@@ -646,14 +646,14 @@ struct type_caster<std::basic_string<CharT, Traits, Allocator>, enable_if_t<is_s
...
-#if PY_VERSION_MAJOR < 3
+#if PY_MAJOR_VERSION < 3
         object temp;
 #endif
         handle load_src = src;
         if (!src) {
             return false;
         } else if (!PyUnicode_Check(load_src.ptr())) {
-#if PY_VERSION_MAJOR >= 3
+#if PY_MINOR_VERSION >= 3
```
dean0x7d: Should be `MAJOR` instead of `MINOR`.

wjakob:
I wonder if it's related to the S3 outage.

eirrgang:
Whoah. How did I mess that one up?! New PR...

jagerman:
I seem to recall github allows you to undo a branch delete, at least for a little while.

jagerman:
(The PR was closed because the branch was deleted)

Issue #711, opened 2017-03-08
Errors with clang-4.0


jcelerier:
Clang 4.0 seems to introduce a new warning, treated as an error, that breaks pybind11 for me: `-Wunusable-partial-specialization`.

The error can be seen here : https://travis-ci.org/OSSIA/libossia/jobs/209039296#L1914

My whole pybind11 code is here : https://github.com/OSSIA/libossia/blob/master/OSSIA/ossia-python/ossia_python.cpp

jagerman:
This is fixed in current master (PR #677 fixed it).

I'll close this, but of course reopen if you're still getting it on current master.

jcelerier:
nice, thanks, pulling asap!

PR #677, opened 2017-02-16
Revert noexcept deduction in favour of SFINAE lambda function matching

jagerman:
`noexcept` deduction, added in PR #555, doesn't work with clang's `-std=c++1z`; and while it works with g++, it isn't entirely clear to me that it is required to work in C++17.

What should work, however, is C++17's umplicit conversion of a `noexcept(true)` function pointer to a `noexcept(false)` one (i.e.  default, `noexcept`-specifier-omitted).  That was breaking in C++17 mode in pybind11 before #555 because the `cpp_function` template used for lambdas provided a better match (i.e. without requiring an implicit conversion), but it then failed when trying to treat the matched function pointer as a lambda object.

This commit takes a different approach of using SFINAE on the lambda function to prevent it from matching a non-lambda object, which then gets implicit conversion from a `noexcept` function pointer to a `noexcept(false)` function pointer.  This much nicer solution also gets rid of the C++17 NOEXCEPT macros, and works in both clang and g++.

With this, pybind11 builds and tests successfully with clang 4.0 and libc++ 4.0 on linux under `-std=c++1z`.

PR #712, opened 2017-03-08
Add failure-allowed clang-4.0/libc++-4.0 build

jagerman:
Bug #711 found the same issues that I did (which led me to #677), so I decided to convert my clang build code into a proper travis-ci build.

This adds a clang 4.0 with libc++ 4.0 docker-based build, using `-std=c++1z` for the build.

It's marked failure-allowed (like the gcc 7 one) since this is still an unreleased compiler with an experimental c++ standard.

The one annoyance here is that I couldn't find a pre-built libc++ 4.0 package, so had to build it from llvm-source in the build.  This is reasonably quick (it adds around a minute to the build time), but is needed for libc++'s c++17 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,9 @@ matrix:
   - sudo: true
     services: docker
     env: PYTHON=3.5 CPP=17 GCC=7
+  - sudo: true
+    services: docker
+    env: PYTHON=3.5 CPP=17 CLANG=4.0
   - os: osx
     osx_image: xcode7.3
     env: PYTHON=2.7 CPP=14 CLANG
@@ -77,6 +80,7 @@ matrix:
     - flake8
   allow_failures:
     - env: PYTHON=3.5 CPP=17 GCC=7
+    - env: PYTHON=3.5 CPP=17 CLANG=4.0
 cache:
   directories:
   - $HOME/.cache/pip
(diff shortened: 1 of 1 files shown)


wjakob:
Starting to test C++1z is a good idea. One somewhat related thought: our build matrix on Travis CI is getting fairly big. I assume we'll eventually run into some kind of quota and our jobs will be slowed down (potentially this is happening already). Shall we remove some of the mostly redundant targets?

jagerman:
One other thing we could do is combine some of them.  For example, by having just one c++17 build that does both a clang 4 and gcc 7 build.

jagerman:
The travis limit for build matrix size is 200, so we're nowhere near that.  Throttling is applied long before that, of course, by limiting the number of concurrent jobs (across all branches/PRs), with the specific number depending on the current travis-ci load.

As far as delayed builds, during low load times there isn't much of an issue: everything builds more or less at once; we're using idle travis-ci machines.  During high load, it is *always* the OS X builds that are the bottleneck.

So while I agree we should be good citizens by not wasting builds that aren't helpful, I don't think we're hitting limits in any practical way at the moment.

wjakob:
Ok, sounds good. We can remove configurations from the build matrix once this becomes an actual problem. Feel free to merge this.

PR #714, opened 2017-03-09
Don't use static constexpr char * for value passed to python

### LINKED ISSUE #713: type_caster<std::basic_string<...>>::cast crashes
#### Issue description
* python 2.7.11
* MSVC 2015, gcc 4.8.5
When setting dictionary item value to const char *, assignment uses type_caster<std::string>::cast(), which crashes. 

Crash happens inside PyUnicode_Decode method when called with with encoding="utf8", it looks like python bug, however if PyUnicode_Decode invoked with encoding utf-8 it uses 'fast path' and call PyUnicode_DecodeUTF8 instead of dealing with codecs, also PyUnicode_FromStringAndSize uses PyUnicode_DecodeUTF8  as well, so fix maybe just change encoding to "utf-8" and it will be faster, consistent with PyUnicode and will not crash

```c++
namespace py = pybind11;

int main()
{
	py::dict dict;
	dict["a"] = "b"; //crash here
        return 0;
}
```

jagerman:
Attempt to fix #713.  I *think* the `static constexpr char *` here is okay by the standard, but MSVC and gcc 4.8 reportedly crash when the encoding value is passed to Python, so just hard-code the value in the two places we use it instead of using a static member.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -641,7 +641,6 @@ struct type_caster<std::basic_string<CharT, Traits, Allocator>, enable_if_t<is_s
     static_assert(!std::is_same<CharT, wchar_t>::value || sizeof(CharT) == 2 || sizeof(CharT) == 4,
             "Unsupported wchar_t size != 2/4");
     static constexpr size_t UTF_N = 8 * sizeof(CharT);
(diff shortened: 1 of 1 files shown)


jagerman:
Apparently not the issue.

PR #715, opened 2017-03-09
Call PyUnicode_DecodeUTF* directly

jagerman:
Some versions of Python 2.7 reportedly (#713) have issues with
PyUnicode_Decode being passed the encoding string, so just skip it
entirely by calling the PyUnicode_DecodeUTF* function directly.


wjakob:
This looks good to me. Something really strange is going on with the Travis/PyPy configuration.

jagerman:
It's just #716 - the real error is here: https://travis-ci.org/pybind/pybind11/jobs/209433805#L400; a rebase will fix it (just restarting an existing build doesn't merge).

jagerman:
PyPy really didn't like the change to PyDecode_UTFx: the UTF8 version segfaults, and the UTF16/32 ones only take a non-const `char *` (and may similarly segfault--I didn't get past the utf-8 segfault).  Instead I just reverted the PyPy version to just a `PyUnicode_Decode` call essentially as we had before, passing the encoding as a string, which seems to work reliably.  I think PyPy also doesn't have the same fast-path advantages as CPython for these encodings, nor the issue from #713, so this seems a pretty reasonable compromise.

PR #716, opened 2017-03-10
Use trusty for travis-ci pypy build

jagerman:
Nightlies for pypy no longer run on Ubuntu 12.04 as of a couple days ago; change the pypy build distribution to the travis-ci trusty (i.e. 14.04) beta container to get them working again.

The pypy build was also [installing numpy and scipy for the *system* python version](https://travis-ci.org/pybind/pybind11/jobs/209056475#L441), which was pointless.  We could fix it by calling `pypy -m pip`, but that ends up building all of numpy and scipy from source, which doesn't seem desirable, so this commit also adds a guard around the eigen/numpy/scipy install code with a not-PYPY check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,10 @@ matrix:
     env: PYTHON=3.6 CPP=14 CLANG
   # Test a PyPy 2.7 nightly build
   - os: linux
+    dist: trusty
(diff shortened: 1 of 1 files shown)


wjakob:
Looks good, thank you!

PR #692, opened 2017-02-23
Minor docs build maintenance

dean0x7d:
* Switch breathe to stable releases. It was previously pulling directly from master because a required bugfix was not in a stable release yet.

* Force update sphinx and RTD theme. When using conda, readthedocs pins `sphinx==1.3.5` and `sphinx_rtd_theme==0.1.7`, which is a bit older than the ones used in the RTD regular (non-`conda`) build. The newer theme has nicer sidebar navigation (4-level depth vs. only 2-level on the older version*). Note that the `python==3.5` requirement must stay because RTD still installs the older sphinx at one point which isn't available with Python 3.6. This just updated the theme at a later stage.

  (*) Compare the sidebar navigation depth on [stable (pre-doxygen, non-conda build)](http://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#buffer-protocol) vs. [master (doxygen, conda build)](http://pybind11.readthedocs.io/en/master/advanced/pycpp/numpy.html#buffer-protocol). The latter is currently missing third level navigation (regression after moving to the conda build for doxygen). This PR restores the nicer RTD sidebar.

**Note**: Skipped CI tests since this is a pure RTD config change, i.e. completely transparent to CI.

PR #721, opened 2017-03-12
Fix readthedocs build

dean0x7d:
As noted in https://github.com/pybind/pybind11/pull/692#issuecomment-285863682, the RTD build with doxygen is broken. This was due to a recent build environment update on the RTD side. The 1.8.14.dev build of doxygen that we were using broke as a result. The update also broke the conda-forge build of 1.8.13 (but that version has other issues).

Luckily, the RTD update did bring their doxygen version up to 1.8.11 which is enough to parse the C++11 code we need (ref qualifiers) and it also avoids the segfault found in 1.8.13.

Since we're using the native doxygen, conda isn't required anymore and we can simplify the RTD configuration. Sample output from new config: http://pybind11-rtdtest.readthedocs.io/en/latest/reference.html

**[skip ci]** Skipping Travis and AppVeyor since they're not affected by RTD config changes anyway (it would just waste their time).


jagerman:
Looks good, and the simplification is nice.

wjakob:
Awesome, thanks!

PR #722, opened 2017-03-12
VS 2017 AppVeyor build

jagerman:
This adds VS 2017 to the build matrix, updates the appveyor Python version we build against to 3.6, and adds a workaround for one ICE that is triggered under VS 2017.

The appveyor build matrix support is a bit annoying: unlike travis, there is no ability to exclude some builds other than a dirty hack of making them fail early, and the 1.5 year old [appveyor bug](https://github.com/appveyor/ci/issues/386) has nothing from AppVeyor itself other than a suggestion to use this dirty hack.  I've done that here by marking the "skipped" tests as allow_failures.

The VS 2017 builds are still a bit slow to queue (they take a couple minutes to start), even for the "skipped" ones, but that should improve sometime next week once Appveyor migrates the 2017 builds to more efficient build infrastructure.

I'm not sure whether we should do it this way (i.e. with fake failures for skipped tests), or else just go ahead with enlarging the build matrix to 8 jobs (2 archs × 2 compilers × 2 python versions) since upstream apparently doesn't care to make it practical to incur a lighter build load.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,16 +1,22 @@
 version: 1.0.{build}
-os: Visual Studio 2015
+image:
+- Visual Studio 2017
+- Visual Studio 2015
 test: off
 platform:
-- x86
 - x64
+- x86
 environment:
   matrix:
+  - CONDA: 36
   - CONDA: 27
-  - CONDA: 35
+matrix:
+  fast_finish: true  # Stop remaining jobs after a job failure
 install:
 - ps: |
     if ($env:PLATFORM -eq "x64") { $env:CMAKE_ARCH = "x64" }
+    if ($env:APPVEYOR_JOB_NAME -like "*Visual Studio 2017*") { $env:CMAKE_GENERATOR = "Visual Studio 15 2017" }
+    else { $env:CMAKE_GENERATOR = "Visual Studio 14 2015" }
     if ($env:PYTHON) {
       if ($env:PLATFORM -eq "x64") { $env:PYTHON = "$env:PYTHON-x64" }
       $env:PATH = "C:\Python$env:PYTHON\;C:\Python$env:PYTHON\Scripts\;$env:PATH"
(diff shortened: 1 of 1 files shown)


dean0x7d:
Nice! Good timing, I really needed this. I was just also looking into some VS issues, wanted to add the test config and encountered the same AppVeyor limitations. I'd say let all the configurations run If AppVeyor doesn't offer any nice way to exclude builds.

Also, I was surprised that CMake 3.7 can't find VS2017 without being given an explicit generator. Nice solution with `APPVEYOR_JOB_NAME`. Hopefully, CMake 3.8 will be able to find it automatically.

include/pybind11/pybind11.h
```diff
@@ -50,13 +50,25 @@ class cpp_function : public function {
         initialize(f, f, extra...);
     }
 
+#if !defined(_MSC_VER)
     /// Construct a cpp_function from a lambda function (possibly with internal state)
     template <typename Func, typename... Extra,
              typename FuncType = typename detail::remove_class<decltype(&std::remove_reference<Func>::type::operator())>::type>
     cpp_function(Func &&f, const Extra&... extra) {
```
dean0x7d: If you'd like to avoid the preprocessor, an alternative solution would be to do SFINAE without `decltype` and move `FuncType` inside the constructor. I've tried the following purely `<type_traits>`-base SFINAE condition and it works well with VS2017:
```
detail::enable_if_t<!std::is_pointer<F>::value && !std::is_member_pointer<F>::value>>
```

jagerman: GCC and clang 4 (or possibly just clang 4 in C++1z mode) also needed a `remove_reference` and `!std::is_function` test to avoid compilation errors, but I think that approach is nicer (and should keep working as MSVC gains C++17 support).

jagerman:
These builds are pretty slow right now (about 3 minutes to start, and 5+ to build and test, so a total addition of about 24 minutes), so I think we should hold off merging the build changes until [AppVeyor migrates the VS 2017 images to the faster hardware](https://github.com/appveyor/ci/issues/1179#issuecomment-285534695).

I'll merge the 2017 compilation fix separately, as other people are hitting it (someone just reported the failure in the gitter room).

jagerman:
VS 2017 jobs are fast now.  Anything else before I merge this?

wjakob:
no, it looks good to me!

dean0x7d:
Looks great!

PR #723, opened 2017-03-12
Add static_assert failure for non-static def_static

### LINKED ISSUE #697: def_static with a member function
If I accidentally use `def_static` instead of `def` for a member function, it compiles but leads to a runtime argument error when called. It's not hard to fix once it pops up, but could this typo be caught at compile-time?

jagerman:
Fixes #697 

Anyone see any problem with this?  I can't think of any reason you would give a member function to `def_static` that isn't an error.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -924,7 +924,9 @@ public:
     }
 
     template <typename Func, typename... Extra> class_ &
-    def_static(const char *name_, Func f, const Extra&... extra) {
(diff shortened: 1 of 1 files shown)


jagerman:
This also fixes a non-rvalue forwarding (though I'm not sure it caused any practical problems).

wjakob:
Looks good to me. Was this motivated by a concrete issue?

wjakob:
ah, never mind -- I saw it now.

dean0x7d:
Looks good to me.

Issue #631, opened 2017-01-31
vectorize dispatch std::complex issue


nbecker:
```c++
#include "pybind11/numpy.h"     // vectorize
#include "pybind11/pybind11.h"
#include "pybind11/operators.h"
#include "numpy/arrayobject.h"
#include <complex>

namespace py = pybind11;


PYBIND11_PLUGIN (bug1) {
  if (_import_array() < 0) {
    PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
    return nullptr;
  }

  py::module m("bug1", "pybind11 example plugin");

  m.def("mag_sqr_ufunc", py::vectorize([](double x) { return x * x; }));
  m.def("mag_sqr_ufunc", py::vectorize([](std::complex<double> x) { return real(x) * real(x) + imag(x) * imag(x); }));

  return m.ptr();
}
```
```python
import numpy as np
from bug1 import mag_sqr_ufunc
u = np.ones (1000000, dtype=complex)
mag_sqr_ufunc(u)
```
produces:

```
In [3]: /home/nbecker/.local/bin/ipython3:4: ComplexWarning: Casting complex values to real discards the imaginary part
  import re
```

This leads me to believe that the wrong ufunc was dispatched, when given complex data type, the scalar overload was erroneously chosen.
This is confirmed by commenting out the scalar overload, in which case there is no message and a correct result is returned.

jagerman:
Pybind's default overload dispatch tries in FIFO order, and stops when one works.  That means the `float`/`double` overload is never going to work (it's always going to call the first one registered).  As to the `std::complex` one, what happens if you swap the order of the definitions (that'll certainly fix the problem for actual complex, but what I'm unsure of is whether regular python numeric value arguments will also end up there).

nbecker:
I followed this example:
https://github.com/pybind/pybind11/issues/563

to do my own dispatch:
```
  m.def("mag_sqr_ufunc", [](py::array a) {
      if (py::isinstance<py::array_t<complex_t>>(a))
        return py::vectorize([](complex_t x) { return real(x) * real(x) + imag(x) * imag(x); })(a);
      if (py::isinstance<py::array_t<double_t>>(a))
        return py::vectorize([](double_t x) { return x * x; })(a);
      if (py::isinstance<py::array_t<complex64_t>>(a))
        return py::vectorize([](complex64_t x) { return real(x) * real(x) + imag(x) * imag(x); })(a);
[...]

PR #724, opened 2017-03-13
Improve py::array_t scalar type information

### LINKED ISSUE #632: add typedef to get element type of array_t
I believe this makes it a lot easier to write generic code using array_t

```diff
diff --git a/include/pybind11/numpy.h b/include/pybind11/numpy.h
index c224bd4..5795347 100644
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -563,6 +563,7 @@ protected:
 
 template <typename T, int ExtraFlags = array::forcecast> class array_t : public array {
 public:
+    typedef T el_t;
     array_t() : array(0, static_cast<const T *>(nullptr)) {}
     array_t(handle h, borrowed_t) : array(h, borrowed) { }
     array_t(handle h, stolen_t) : array(h, stolen) { }
```

dean0x7d:
Resolves #632 and the comment in https://github.com/pybind/pybind11/issues/631#issuecomment-286100470.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1142,7 +1142,9 @@ struct vectorize_helper {
 };
 
 template <typename T, int Flags> struct handle_type_name<array_t<T, Flags>> {
-    static PYBIND11_DESCR name() { return _("numpy.ndarray[") + make_caster<T>::name() + _("]"); }
+    static PYBIND11_DESCR name() {
+        return _("numpy.ndarray[") + npy_format_descriptor<T>::name() + _("]");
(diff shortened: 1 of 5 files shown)


dean0x7d:
Updated to also improve the signatures of `py::array_t` functions. See https://github.com/pybind/pybind11/issues/631#issuecomment-286100470. Having both `float` and `double` reported as `float` is confusing. Using `npy_format_descriptor` makes those `float32` and `float64`. This is already the case for Eigen types so I don't think it should be an issue.

Issue #726, opened 2017-03-13
Make a v2.1 release?


dean0x7d:
Seems like there is a decent amount of new features and bug fixes for a v2.1 release. This should also be good timing with the recent release of Visual Studio 2017 and the related fixes in pybind11.

**Bugs:** Are there any critical bugs that should be fixed? Looking at the issues, there don't seem to be any showstoppers.

**Features:** Of the feature PRs, #488 and #693 are close to ready. I'd say #693 would be best to merge right at the beginning of v2.2.dev so it gets some time in master due to the low-level core changes, rather than going straight to release. #488 doesn't affect the core library in that way, but perhaps a general feature freeze for v2.1 would be best.

Thoughts?

jagerman:
Sounds good to me.  I think current master is in pretty good shape, and I agree that #693 (or anything of a similar scope) should be in master for a while before a release.  For non-core things like #488, I think we can accept them if they are ready to go in the next couple days.

jagerman:
#488 is merged now; I also created v2.1 and v2.2 milestones to help categorize issues/PRs.

jagerman:
I think #671 will be an easy fix, as will adding compiler version checks to close #706; I've tagged them both for v2.1, and will submit PRs tomorrow.

wjakob:
Creating a v2.1 with the current master sounds good to me. I'll wait for @jagerman's two commits before pushing it out (and perhaps add 1-2 days of delay to be on the safe side?).

wjakob:
https://travis-ci.org/pybind/pybind11/jobs/210795174 :( -- I sense nondeterminism.

MathMagique:
A 2.1 release would be awesome, since I would love to use the (seemingly new) support for `u16string`. Thanks for the great work!

jagerman:
@wjakob that looks like a pypy bug: they just merged a fix for the tp_bases-is-incomplete issue, but when I tried removing the workaround from the MI PR I got a segfault in the same place.  I'm going to investigate and report upstream (assuming it's the same issue).

jagerman:
The segfault issue is [fixed in PyPy](https://bitbucket.org/pypy/pypy/issues/2480/cpyext-tp_bases-doesnt-properly-expose).

dean0x7d:
@jagerman What do you think about bumping #671 to v2.2? It's not a bug fix and it seems like there's still some discussion on performance going on there.

In that case, there would be just #727 left to merge and we can call it a release.

[...]

PR #727, opened 2017-03-14
Added minimum compiler version assertions

### LINKED ISSUE #706: std::negation requires Visual Studio 2015 Update 3
As reported [here](https://www.bountysource.com/issues/41112911-error-building-the-python-plugin-with-visual-studio-2015).

jagerman:
We now require (and enforce at compile time):
- GCC 4.8+
- clang 3.3+ (5.0+ for Apple's renumbered clang)
- MSVC 2015u3+
- ICC 15+

This also updates the versions listed in the README, and removes a now-redundant MSVC version macro check.

Fixes #706 (by producing a more useful error message citing the required MSVC version).

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -28,6 +28,33 @@
 #  endif
 #endif
 
+// Compiler version assertions
+#if defined(__INTEL_COMPILER)
+#  if __INTEL_COMPILER < 1500
+#    error pybind11 requires Intel C++ compiler v15 or newer
+#  endif
+#elif defined(__clang__) && !defined(__apple_build_version__)
+#  if __clang_major__ < 3 || (__clang_major__ == 3 && __clang_minor__ < 3)
+#    error pybind11 requires clang 3.3 or newer
+#  endif
+#elif defined(__clang__)
+// Apple changes clang version macros to its Xcode version; the first Xcode release based on
+// (upstream) clang 3.3 was Xcode 5:
+#  if __clang_major__ < 5
+#    error pybind11 requires Xcode/clang 5.0 or newer
+#  endif
+#elif defined(__GNUG__)
+#  if __GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 8)
+#    error pybind11 requires gcc 4.8 or newer
+#  endif
+#elif defined(_MSC_VER)
+// Pybind hits various compiler bugs in 2015u2 and earlier, and also makes use of some stl features
+// (e.g. std::negation) added in 2015u3:
+#  if _MSC_FULL_VER < 190024213
+#    error pybind11 requires MSVC 2015 update 3 or newer
+#  endif
+#endif
+
 #if !defined(PYBIND11_EXPORT)
 #  if defined(WIN32) || defined(_WIN32)
 #    define PYBIND11_EXPORT __declspec(dllexport)
@@ -651,8 +678,8 @@ struct error_scope {
 /// Dummy destructor wrapper that can be used to expose classes with a private destructor
 struct nodelete { template <typename T> void operator()(T*) { } };
 
-// overload_cast requires variable templates: C++14 or MSVC 2015 Update 2
-#if defined(PYBIND11_CPP14) || _MSC_FULL_VER >= 190023918
+// overload_cast requires variable templates: C++14 or MSVC
+#if defined(PYBIND11_CPP14) || defined(_MSC_VER)
 #define PYBIND11_OVERLOAD_CAST 1
 
 NAMESPACE_BEGIN(detail)
--- a/README.md
+++ b/README.md
@@ -96,9 +96,9 @@ In addition to the core functionality, pybind11 provides some extra goodies:
 
 ## Supported compilers
 
-1. Clang/LLVM (any non-ancient version with C++11 support)
+1. Clang/LLVM 3.3 or newer (for Apple Xcode's clang, this is 5.0.0 or newer)
 2. GCC 4.8 or newer
-3. Microsoft Visual Studio 2015 or newer
+3. Microsoft Visual Studio 2015 Update 3 or newer
 4. Intel C++ compiler 16 or newer (15 with a [workaround](https://github.com/pybind/pybind11/issues/276))
 5. Cygwin/GCC (tested on 2.5.1)
 


include/pybind11/common.h
```diff
@@ -28,6 +28,27 @@
 #  endif
 #endif
 
+// Compiler version assertions
+#if defined(__INTEL_COMPILER)
+// ICC v15 or newer required
+static_assert(__INTEL_COMPILER >= 1500, "pybind11 requires Intel C++ compiler v15 or newer");
```
dean0x7d: The offending compiler may not actually have `static_assert` -- might be better to use `#error`.

Minor: The comments are a bit redundant as they just say the same thing as the error message and/or assertion statement.

jagerman: Is `#error` universally supported?

jagerman: Oh, it is.  I thought it was a non-standard extension.  I'll update.

include/pybind11/common.h
```diff
@@ -28,6 +28,33 @@
...
+#if defined(__INTEL_COMPILER)
+#  if __INTEL_COMPILER < 1500
+#    error pybind11 requires Intel C++ compiler v15 or newer
+#  endif
+#elif defined(__clang__) && !defined(__apple_build_version__)
+#  if __clang_major__ < 3 || (__clang_major__ == 3 && __clang_minor__ < 3)
+#    error pybind11 requires clang 3.3 or newer
+#  endif
+#elif defined(__clang__)
+// Apple changes clang version macros to its Xcode version for some reason ("Think different"?)
```
wjakob: I sympathize with the rant, but perhaps we don't need it permanently in ``common.h``? ;)

jagerman: Heh, fair enough.

SylvainCorlay:
The changes that would be required to support Update 1 of MSVC seem pretty minor.

It would be nice to still support this version...

dean0x7d:
Like I said in #706:
> VS 2015 Update 2 should be able to compile pybind11 with just some minor workarounds, but I think the bigger issue is that it's difficult to test. AFAIK AppVeyor only has a "Visual Studio 2015" option and it's always kept up to date, so there's only Update 3 to test against.

Without continuous testing, the support could break at any point and we wouldn't know about it. It's not just the standard library. The different MSVC compiler updates handle some template code differently which requires specific workarounds. It looks bad to pledge support for Update 1 and 2, when it would most likely break often. The VS updates themselves are free so I don't think it's too much to ask for Update 3.

SylvainCorlay:
Indeed, it seems that this breakage was not intentional in the first place. However, updating a compiler version in large projects / production environments is not a trivial decision...

SylvainCorlay:
It appears that appveyor has images for older versions of visual studio.

wjakob:
In the partial order of compiler standards compliance, the service packs are comparable and clearly getting better over time. Hence, if we require a service pack, then I don't think it's unreasonable to ask for SP3. While it would be nice to support the non-patched MSVC2015 release, that seems impossible given the many issues this release has. Thus I'm leaning against supporting earlier MSVC 2015 service packs, particularly if this means adding ugly workarounds somewhere.

dean0x7d:
> PS: it appears that appveyor has images for older versions of visual studio.

Not the individual updates as far as I can see. There are 2013, 2015 and 2017 images but they all have the latest update of their respective Visual Studio. There are also the "Previous Visual Studio xxxx" images, but those are only the "last known good build" as they update all software continuously (so not necessarily the previous VS update).

SylvainCorlay:
This comes from the following thread: http://help.appveyor.com/discussions/questions/3167-visual-studio-2015-update-2

dean0x7d:
> This comes from the following thread: http://help.appveyor.com/discussions/questions/3167-visual-studio-2015-update-2

Looks like it's a workaround for specific accounts, rather than a regular feature: [https://github.com/appveyor/ci/issues/891](https://github.com/appveyor/ci/issues/891)

jagerman:
Aside from the testing issues, update 3 versus update 2 gives us two concrete things:

1. The `all_of_t` and related templates become much simpler than the required workaround in update 2.
2. We have at least one feature, `overload_cast`, that appears to be impossible with update 2.

If 1 was the only issue we could address it by reintroducing the (ugly) workaround instead of using `std::conjunction`/`std::disjunction`, but 2 is a bigger deal: documented features of pybind simply wouldn't be available under certain compilers.  I think this is a very bad thing, and I'd honestly prefer that we remove features that aren't supported in every environment we support.  But that means we can't have nice things like `overload_cast` because of *fixed* bugs in an older compiler that can't be worked around.

SylvainCorlay:
It seems that on other platforms, you already disable `overload_cast` when C++14 is not available.

jagerman:
It comes down to this: is there a strong, pressing need for continuing to support update 2 that outweighs the maintenance burden?  Who strongly needs this and cannot upgrade, and are they using, or likely to start using, pybind11 but only if they don't have to install Microsoft's software update for their compiler?

On the flip side, I know that, personally, I've spent considerably longer dealing with MSVC bugs (often ICEs) than I have for bugs in all other compilers and compiler versions combined.  It's not a well-loved compiler, to put it mildly, and dealing with its bugs is a pain in the ass; I'm going to need something more than "there might be hypothetical users out there that want to continue using older, buggy software without installing the released, free, bug-fix updates" to be convinced here.

I don't restrict this point of view to Microsoft: I'd have little sympathy for dealing with, say, a gcc error in Ubuntu that is fixed by a released update to that specific version of Ubuntu.

jagerman:
I think @dean0x7d, @wjakob and I are in agreement that, at present, the maintenance burden of supporting MSVC 2015 update 2 or earlier is too high relative to the fairly small gains.

I think these two quotes (from #726), with which I'm in complete agreement, essentially sum it up:
@dean0x7d:
> Adding support for Update 2 now would require someone to install it locally (can't be installed side-by-side with Update 3), make the necessary workarounds and then hope it doesn't break in the future as there is no CI support.

@wjakob:
> Part of the motivation of this project is that it's okay to assume a reasonably correct compiler to avoid having to build complicated abstractions over semi-broken metaprogramming facilities.

If someone feels that support for earlier versions of compilers is sufficiently worthwhile and wants to commit to not only contribute the necessary workarounds for the current code **but also** to regularly dealing with future issues (because, quite frankly, I think most regular pybind11 contributors hate dealing with MSVC issues, and the older the release the more issues), I'd happily accept such a contribution.  But I don't want to deal with a semi-broken compiler, and I don't think it's fair to ask contributors to pybind11 to deal with potential issues; so until/unless such a contribution commitment is made, the list in this PR is going to be the minimum supported.

PR #693, opened 2017-02-23
Python multiple inheritance

jagerman:
This builds on @dean0x7d's #679 to support multiple inheritance of C++ classes from Python.  This required no dispatch changing at all, but does require some changes to be able to store multiple C++ base objects in a single instance.

This is done by changing the `instance` layout to contain a single `void**` which is allocated to the correct size and contains value/holder pointer pairs.  The value pointers are themselves each set to newly-allocated memory of the size required for each value, and the holders are initialized to nullptr.

Thus the instance changes from this:

- ... python stuff ...
- pointer to VALUE data
- weakrefs *
- bools
- HOLDER

to:

- ... python stuff ...
- pointer to value/holder data: [[VALUE1 *][HOLDER1 *][VALUE2 *][HOLDER2 *]...]
- weakrefs *
- bools

where each VALUEn pointer is to newly allocated memory of the size required by the value type, and each HOLDERn starts out set to nulptr.

This has three effects: First, all instances have the same size (more on this below).  Second, `instance` can store multiple base values and holders for those base values.  Third, we don't need the `instance_essentials`/`instance<HolderType>` distinction anymore: there is now just one `instance` which can contain any number of values and holders of whatever size.

Structure aside, a new `get_all_type_info()` function is added which takes a type and returns a vector of all the pybind types that make up the input type.  For types that are already a pybind type, or that inheriting just one pybind-registered type, this is just that the `typeinfo` for that type.  For multiple inheritance, this will be a list.

A companion function, `get_value_and_holder()`, contains the logic of using the `get_all_type_info()` to extract the correct references to the value and holder pointers for a given type from the `instance`.

As for method dispatching: we don't have to do anything.  Python takes care of the MRO.  (Cross-class overloads won't work, but that's true in Python in general).  The generic type caster now knows how to extract the correct value from a multi-valued instance, and an invoked constructor can get the correct holder type/pointer by simply using the parent of the constructor itself (rather than the instance itself, which may have multiple holders).

.so size increases slightly (~1.2%, or about 20KB, on the full test suite compared to #679 under g++6).  I think it's worthwhile (and I expect a good chunk of that increase is constant).

Currently this fails under PyPy: I haven't investigated in detail, but will do so.

Some other comments:

[...]

PR #729, opened 2017-03-14
Fix pydoc failure because of missing __module__

### LINKED ISSUE #728: help()/pydoc fails when a module has classes
This must be a regression as http://pybind11.readthedocs.io/en/master/classes.html#overloaded-methods has an example of `help()` and this example doesn't work with the current master.

I can get a a traceback if I call an internal function directly:
```
>>> print pydoc.text.docmodule(example, 'example')
  File "/usr/lib/python2.7/pydoc.py", line 1140, in docmodule
    inspect.getclasstree(classlist, 1), name)]
  File "/usr/lib/python2.7/inspect.py", line 739, in getclasstree
    return walktree(roots, children, None)
  File "/usr/lib/python2.7/inspect.py", line 708, in walktree
    classes.sort(key=attrgetter('__module__', '__name__'))
AttributeError: __module__
```

I started pdb and it seems that it's `pybind11_object_48` that is missing `__module__`:

```
(Pdb) 
> /usr/lib/python2.7/inspect.py(739)getclasstree()
-> return walktree(roots, children, None)
(Pdb) p roots
[<class 'pybind11_object_48'>]
(Pdb) p children
{<class 'pybind11_object_48'>: [<class 'example.Pet'>]}
(Pdb) s
--Call--
> /usr/lib/python2.7/inspect.py(705)walktree()
-> def walktree(classes, children, parent):
(Pdb) n
> /usr/lib/python2.7/inspect.py(707)walktree()
-> results = []
(Pdb) n
> /usr/lib/python2.7/inspect.py(708)walktree()
-> classes.sort(key=attrgetter('__module__', '__name__'))
(Pdb) p classes
[<class 'pybind11_object_48'>]
(Pdb) p classes[0].__class__
<class 'pybind11_type'>
(Pdb) p classes[0].__module__
*** AttributeError: AttributeError('__module__',)
(Pdb) p classes[0].__name__
'pybind11_object_48'
```

I don't understand how all this works, so that's all I can tell.

dean0x7d:
Fixes #728.

The `__module__` for all pybind11 types is set to `pybind11_builtins`, analogous to Python's own `builtins`.

### DIFF
--- a/include/pybind11/class_support.h
+++ b/include/pybind11/class_support.h
@@ -57,6 +57,8 @@ inline PyTypeObject *make_static_property_type() {
     if (PyType_Ready(type) < 0)
         pybind11_fail("make_static_property_type(): failure in PyType_Ready()!");
 
+    setattr((PyObject *) type, "__module__", str("pybind11_builtins"));
+
     return type;
 }
 
@@ -170,6 +172,8 @@ inline PyTypeObject* make_default_metaclass() {
     if (PyType_Ready(type) < 0)
         pybind11_fail("make_default_metaclass(): failure in PyType_Ready()!");
 
+    setattr((PyObject *) type, "__module__", str("pybind11_builtins"));
+
     return type;
 }
 
@@ -269,6 +273,8 @@ inline PyObject *make_object_base_type(size_t instance_size) {
 
     if (PyType_Ready(type) < 0)
         pybind11_fail("PyType_Ready failed in make_object_base_type():" + error_string());
+
+    setattr((PyObject *) type, "__module__", str("pybind11_builtins"));
 
     assert(!PyType_HasFeature(type, Py_TPFLAGS_HAVE_GC));
     return (PyObject *) heap_type;
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -52,3 +52,11 @@ def test_importing():
 
     assert OD is OrderedDict
     assert str(OD([(1, 'a'), (2, 'b')])) == "OrderedDict([(1, 'a'), (2, 'b')])"
+
+
+def test_pydoc():
+    """Pydoc needs to be able to provide help() for everything inside a pybind11 module"""
+    import pybind11_tests
+    import pydoc
+
+    assert pydoc.text.docmodule(pybind11_tests)


wjakob:
Looks good to me!

wojdyr:
Maybe a shorter name for pybind11_builtins.pybind11_object_*?
Pydoc output feels a bit cluttered. For example the example from http://pybind11.readthedocs.io/en/master/classes.html#overloaded-methods is now:
```
class Pet(pybind11_builtins.pybind11_object_48)
 |  Method resolution order:
 |      Pet
 |      pybind11_builtins.pybind11_object_48
 |      __builtin__.object
 |  
 |  Methods defined here:
 |  
 |  __init__(...)
 |      __init__(self: example.Pet, arg0: unicode, arg1: int) -> None
 |  
 |  set(...)
 |      set(*args, **kwargs)
 |      Overloaded function.
 |      
 |      1. set(self: example.Pet, arg0: int) -> None
 |      
 |      Set the pet's age
 |      
 |      2. set(self: example.Pet, arg0: unicode) -> None
 |      
 |      Set the pet's name
 |  
 |  ----------------------------------------------------------------------
 |  Data and other attributes inherited from pybind11_builtins.pybind11_object_48:
 |  
 |  __new__ = <built-in method __new__ of pybind11_type object>
 |      T.__new__(S, ...) -> a new object with type S, a subtype of T
```

jagerman:
I suppose it could be just `pybind11_builtins.object_48`.

It might also be nice to drop the "_48" for the default size (i.e. for the size used for a typical `unique_ptr` holder).  So then we'd have `pybind11_builtins.object` typically, but the size suffix (e.g. `pybind11_builtins.object_56`) for other types (e.g. a type with a `shared_ptr` holder).

In 2.2 (with PR #693) the size suffix should be gone regardless of the holder.

dean0x7d:
I think it would be nice to avoid shadowing builtin names like `object` and `type`. While they are formally in the `__builtin__` (2.7) / `builtins` (3.x) modules, the objects are always available unqualified globally. I don't think that shadowing them would be too much of a technical issue, but perhaps confusing upon user inspection, especially as Python 2.7 uses unqualified names. E.g. looking up `type(A).__name__` for a pybind11 class and seeing `type` just like a regular Python class may lead someone to the wrong conclusion while debugging. Similarly for bases and `object`. 

That said, I'm not attached to the `pybind11_`-prefix-everywhere naming scheme and don't mind a bit of bikeshedding on this, just without shadowing the builtins. Note that `__module__ == ""` is a possibility, I just  didn't go for it here because `pydoc` leaves the dot prefix anyway.

PR #730, opened 2017-03-15
Stop forcing c-contiguous in py::vectorize

### LINKED ISSUE #671: optimization of strided array vectorize
I'm comparing performance of py::vectorize with performance of nd::Array for different test conditions.  I find that py::vectorize is a factor of 2 slower on this test than ndarray for a 1-d array with stride 2.

In the following test code, when the comments are removed, a function is overloaded for the 1D non-contiguous case to use nd::Array, while without it the py::vectorize is used.  In the ndarray case, the profile looks like:
```
     340  67.3%  67.3%      421  83.4% pybind11::cpp_function::initialize::{lambda#3}::_FUN [clone .lto_priv.280]
      81  16.0%  83.4%       81  16.0% __ieee754_atan2_avx
      36   7.1%  90.5%       36   7.1% __brk
```

While for py::vectorize:
```
     608  50.5%  50.5%      608  50.5% _aligned_strided_to_contig_size16
     167  13.9%  64.4%      170  14.1% pybind11::detail::vectorize_helper::run [clone .isra.536] [clone .lto_priv.257]
     123  10.2%  74.6%      123  10.2% __ieee754_atan2_avx
     116   9.6%  84.3%      279  23.2% pybind11::detail::vectorize_helper::run [clone .isra.533] [clone .lto_priv.258]
     104   8.6%  92.9%      104   8.6% __brk
```

```
  // m.def("rect_to_polar", [](nd::Array<complex_t,1,0> const& a) {
  //     nd::Vector<size_t,1> shape = a.getShape();
  //     nd::Array<double,1,1> r (shape[0]);
  //     nd::Array<double,1,1> theta (shape[0]);
  //     for (int i = 0; i < shape[0]; ++i) {
  //       r[i] = std::abs(a[i]);
  //       theta[i] = std::arg(a[i]);
  //     }
  //     return py::make_tuple (r, theta);
  //   });

  m.def("rect_to_polar", [](py::array a) {
      if (py::isinstance<py::array_t<complex_t>>(a))
        return py::make_tuple (py::vectorize([](complex_t x) { return std::abs(x); })(a),
                               py::vectorize([](complex_t x) { return std::arg(x); })(a)
                               );
      else
        throw py::type_error("rect_to_polar unhandled type");
    });
```

The python test code:
```
from timeit import timeit
w = np.ones(1000000, dtype=complex)
#print (timeit('rect_to_polar(w)', 'from __main__ import rect_to_polar, w', number=1000))
print (timeit('rect_to_polar(w[::2])', 'from numpy_fncs import rect_to_polar; from __main__ import w', number=1000))
#print (timeit('rect_to_polar(w.reshape(w.shape[0]//2, 2))', 'from numpy_fncs import rect_to_polar; from __main__ import w', number=1000))
```

run as:
```
[nbecker@nbecker2 pybind11_test]$ python3 -m yep -- test_numpy_fncs.py 
11.713223681988893
PROFILE: interrupts/evictions/bytes = 1203/109/41896
[nbecker@nbecker2 pybind11_test]$ pprof --text /usr/bin/python3 test_numpy_fncs.py.prof > test_numpy_fncs.test2.pyarray.profile
Using local file /usr/bin/python3.
Using local file test_numpy_fncs.py.prof.
```

Thanks for your help,
Neal

jagerman:
The only part of the vectorize code that actually needs c-contiguous storage is the "trivial" broadcast; for non-trivial arguments, the code already uses strides properly (and so handles C-style, F-style, neither, slices, etc.)

This commit rewrites `broadcast` to additionally check for C-contiguous storage, then takes off the `c_style` flag for the arguments, which will keep the functionality more or less the same, except for no longer requiring an array copy for non-c-contiguous input arrays.

Fixes #671 

Edit: this also enables trivial broadcasting on full-size F-contiguous inputs.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1052,28 +1052,47 @@ private:
     std::array<common_iter, N> m_common_iterator;
 };
 
+// Populates the shape and number of dimensions for the set of buffers.  Returns true if the
+// broadcast is "trivial"--that is, has each buffer being either a singleton or a full-size,
+// C-contiguous storage buffer.
 template <size_t N>
-bool broadcast(const std::array<buffer_info, N>& buffers, size_t& ndim, std::vector<size_t>& shape) {
+bool broadcast(const std::array<buffer_info, N> &buffers, size_t &ndim, std::vector<size_t> &shape) {
     ndim = std::accumulate(buffers.begin(), buffers.end(), size_t(0), [](size_t res, const buffer_info& buf) {
         return std::max(res, buf.ndim);
     });
 
-    shape = std::vector<size_t>(ndim, 1);
+    shape.clear();
+    shape.resize(ndim, 1);
+
     bool trivial_broadcast = true;
     for (size_t i = 0; i < N; ++i) {
+        trivial_broadcast = trivial_broadcast && (buffers[i].size == 1 || buffers[i].ndim == ndim);
+        size_t expect_stride = buffers[i].itemsize;
         auto res_iter = shape.rbegin();
-        bool i_trivial_broadcast = (buffers[i].size == 1) || (buffers[i].ndim == ndim);
-        for (auto shape_iter = buffers[i].shape.rbegin();
-             shape_iter != buffers[i].shape.rend(); ++shape_iter, ++res_iter) {
+        auto stride_iter = buffers[i].strides.rbegin();
+        auto shape_iter = buffers[i].shape.rbegin();
+        while (shape_iter != buffers[i].shape.rend()) {
+            const auto &dim_size_in = *shape_iter;
+            auto &dim_size_out = *res_iter;
 
-            if (*res_iter == 1)
-                *res_iter = *shape_iter;
-            else if ((*shape_iter != 1) && (*res_iter != *shape_iter))
+            // Each input dimension can either be 1 or `n`, but `n` values must match across buffers
+            if (dim_size_out == 1)
+                dim_size_out = dim_size_in;
+            else if (dim_size_in != 1 && dim_size_in != dim_size_out)
                 pybind11_fail("pybind11::vectorize: incompatible size/dimension of inputs!");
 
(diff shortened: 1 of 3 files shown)


wjakob:
+1 for the improved readability and comments. The copies by converting to c_style are indeed completely unnecessary.

Regarding the specific changes: I'm wondering if the full-size C storage requirement is too strict? A trivial broadcast should only require that the strides of non-singleton arguments are identical. In practice, NumPy arrays come in all sorts of flavors (especially F-style when using some of the builtin matrix decompositions implemented in Fortran). This would lead to minor changes to the logic in line 1084.

jagerman:
Ah, good catch (coincidentally, that's the same issue I just fixed in #745).

jagerman:
Our meaning of "trivial" right now is whether we can apply a pointer-for-pointer translation, which requires either singleton or full size.  (There was a minor bug for a singleton slice with non-unit strides that I just pushed, but that's a pretty obscure case.)  Partial sizes (e.g. passing 2x2 and 2x1 arguments) aren't going to be trivial because we can't do the trivial 2*2-pass for loop--that's what we need the non-trivial iterators to handle.

It should, however, be possible to support Fortran-contiguous storage as long as all inputs are F-contiguous.

wjakob:
The way that ``expect_stride`` is computed (starting with ``itemsize`` and then multiplying dimensions) looks fairly specific to C-contiguous arrays. It seems to me that ``broadcast()`` will return ``false`` for F-contigous arrays. Or am I missing something?

wjakob:
Currently, there are some tests involving F-contiguous arrays, but they just check that ``broadcast()==false``.

jagerman:
Right; I'm working on the F-contiguous support right now.

jagerman:
Trivial Fortran-contiguous support added.

jagerman:
@wjakob - do the additions look good?  If so, I'll squash and merge.

wjakob:
Yes, this looks all good now!

PR #732, opened 2017-03-15
Fix nullptr to None conversion for builtin type casters

### LINKED ISSUE #731: returning nullptr segfaults for some types
When I have a function returning a pointer, for some types (std::string*, int*) returning a nullptr segfaults, for others (char*, Class*) it returns None.

If it's a known limitation then take this issue as a feature request to handle nullptr in all cases.

Example:
rtest.cc
```cpp
#include <string>

namespace py = pybind11;

struct Custom { std::string s; };

int* r_int(bool nil) { static int v=12; return nil ? nullptr : &v; }
Custom* r_cust(bool nil) { static Custom v={"a"}; return nil ? nullptr : &v; }
std::string* r_str(bool nil) { static std::string v="b"; return nil ? nullptr : &v; }
char* r_char(bool nil) { static char v[]="c"; return nil ? nullptr : v; }

PYBIND11_PLUGIN(rtest) {
    py::module m("rtest", "");
    py::class_<Custom>(m, "Custom").def(py::init<>()).def_readwrite("s", &Custom::s);
    m.def("r_int", &r_int, py::return_value_policy::reference);
    m.def("r_cust", &r_cust, py::return_value_policy::reference);
    m.def("r_str", &r_str, py::return_value_policy::reference);
    m.def("r_char", &r_char, py::return_value_policy::reference);
    return m.ptr();
}
```
rtest.py
```py
from rtest import *

print r_int(False)
print r_cust(False)
print r_str(False)
print r_char(False)

print r_int(True)  # segfault
print r_cust(True)
print r_str(True)  # segfault
print r_char(True)
```

dean0x7d:
Fixes #731.

Generally, this applies to any caster made with `PYBIND11_TYPE_CASTER()`.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -469,6 +469,7 @@ public:
     public: \
         static PYBIND11_DESCR name() { return type_descr(py_name); } \
         static handle cast(const type *src, return_value_policy policy, handle parent) { \
+            if (!src) return none().release(); \
             return cast(*src, policy, parent); \
         } \
         operator type*() { return &value; } \
--- a/tests/test_python_types.cpp
+++ b/tests/test_python_types.cpp
@@ -464,6 +464,12 @@ test_initializer python_types([](py::module &m) {
     m.def("ord_char16", [](char16_t c) -> uint16_t { return c; });
     m.def("ord_char32", [](char32_t c) -> uint32_t { return c; });
     m.def("ord_wchar", [](wchar_t c) -> int { return c; });
+
+    m.def("return_none_string", []() -> std::string * { return nullptr; });
+    m.def("return_none_char",   []() -> const char *  { return nullptr; });
+    m.def("return_none_bool",   []() -> bool *        { return nullptr; });
+    m.def("return_none_int",    []() -> int *         { return nullptr; });
+    m.def("return_none_float",  []() -> float *       { return nullptr; });
 });
 
 #if defined(_MSC_VER)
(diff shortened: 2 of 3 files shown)


include/pybind11/cast.h
```diff
@@ -469,6 +469,7 @@ template <typename type> class type_caster<std::reference_wrapper<type>> : publi
     public: \
         static PYBIND11_DESCR name() { return type_descr(py_name); } \
         static handle cast(const type *src, return_value_policy policy, handle parent) { \
+            if (!src) return none().release(); \
```
wojdyr: I'm not familiar with the pybind code, but in other places in this file None is returned with `return none().inc_ref();` rather than `.release()`?

dean0x7d: They're essentially the same, but it turns out that `.release()` gives the compiler an opportunity for code size optimization (eliminating the destructor call) in Release mode (as opposed to MinSizeRel, where it doesn't make a difference).

Might need to look into changing the other call sites as well.

wjakob:
This looks good to me!

dean0x7d:
One thing to note: With this change `T* -> None` is allowed for both builtin and user-defined types (defined with `py::class_`). However, the other direction `None -> T*` is possible for user-defined types and `char*`, but no other builtins. `None -> T*` will never segfault like the issue here. It will present a proper `TypeError` with a useful message. 

So there's a bit of asymmetry there, which I'm personally fine with, but I'd just like to note it in case this might be an issue. Supporting `None -> T*` for builtins would also have non-trivial binary size and runtime costs, so I'd be agains supporting it.

jagerman:
I think the asymmetry here is perfectly reasonably, especially because on a returned `nullptr` it's unambiguous to convert to `None`, while the other way is not: e.g. `None` could be both a null `int *` or `double *`.  Better, I think, to force someone to explicitly accept `py::none` and make the appropriate conversion manually.

dean0x7d:
@jagerman That's a good point. OK, merging.

PR #733, opened 2017-03-16
Add note about module destructors

### LINKED ISSUE #638: Documentation: module destructor
Python supports a module destructor via `tp_dealloc`... I can't seem to figure out how to hook that in pybind11. I tried the most obvious thing I could think of (subclass module), and that doesn't work. I'm sure it's supported, right?

virtuald:
- Technique provided by @jagerman
- Fixes #638

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -171,6 +171,26 @@ would be then able to access the data behind the same pointer.
 .. [#f6] https://docs.python.org/3/extending/extending.html#using-capsules
 
 
+Module Destructors
+==================
+
+pybind11 does not explicitly provide a mechanism that can be called at module
+destruction time (and most users will not need this functionality). However,
+if you do happen to need the ability to do something when a module is
+destroyed, you can use a Python capsule to accomplish this:
+
+.. code-block:: cpp
+
+    // the capsule needs something to reference
+    static int unused;
+    
+    py::capsule cleanup(&unused, [](PyObject * capsule) {
+        // do cleanup here -- this function is called with the GIL held
+    });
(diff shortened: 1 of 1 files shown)


wjakob:
This looks good. Quick question: can't the *unused* value simply be ``nullptr``?

jagerman:
If I recall correctly, Python won't allow a nullptr for it.  Of course, if you have some meaningful value, you should use it instead.

dean0x7d:
Would `Py_None` work?

virtuald:
Looking at the API docs for capsule:

* The pointer argument may not be NULL
* It's a void*, so it can be anything you want it to be

Unless someone tries to get the pointer, I don't believe the API will ever try to dereference it, so in theory you could do `Py_None` if you wanted.

jagerman:
Instead of a static local variable, how about amending the example to:

```C++
// the capsule needs a pointer to reference; if you don't have anything better, create a simple one:
int *data = new int(123);

py::capsule cleanup(&unused, [](PyObject * capsule) {
    // do cleanup here -- this function is called with the GIL held
    int *data = PyCapsule_GetPointer(capsule, nullptr);
    delete data;
});
```

which also suggests an obvious way of how to pass data into the capsule.

wjakob:
I was  surprised at how inelegant our way of dealing with ``py::capsule`` destructors is ;)

I submitted a separate PR to deal with this before we document the API and then can't change it anymore #752.

wjakob:
Can you update the PR to take the future changes into account?

Thanks!

wjakob:
closing in favor of #752

PR #735, opened 2017-03-16
Put std::negation usage behind a feature test

jcelerier:
In current AppleClang, -std=c++1z triggers the PYBIND11_CPP17 macro, but std::negation is not implemented. This prevents compile errors in this case.

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -396,7 +396,7 @@ template<size_t ...S> struct make_index_sequence_impl <0, S...> { typedef index_
 template<size_t N> using make_index_sequence = typename make_index_sequence_impl<N>::type;
 #endif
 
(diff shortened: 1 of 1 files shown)


wjakob:
It would be good to only do this extra test if __clang__ is defined.

dean0x7d:
Honestly, the best thing would be to completely remove the special C++17 condition around `bool_constant` and `negation` and just use the simple one-line aliases in the `#else` branch. There isn't really any advantage to using the C++17 `std` library implementations there (in contract to `index_sequence` where the `std` implementation can leverage compiler intrinsics for improved efficiency).

jcelerier:
The problem seems to be the same for g++ < 6 : https://godbolt.org/g/C58PzX

jcelerier:
@dean0x7d Did the modification in #737

Issue #625, opened 2017-01-30
Provide overload similar to return_value_policy to unlock the GIL


virtuald:
It would be ideal to not have to create a lambda each time one wants to unlock the gil:

```
.def("putFrame", [](CvSource &__inst, cv::Mat &image) {
        py::gil_scoped_release release;
        __inst.PutFrame(image);
      })
```

Instead, it seems like an ideal API would be something like:

```
.def("putFrame", &CvSource::PutFrame, py::release_gil)
```

jagerman:
One of the big issues with this is that it would have to be in the core function call dispatch code, but adding it would increase the object size of *all* function calls.  I just pushed a change to #611 that might aid with this, however, by allowing this to happen through the precall mechanism without impacting other functions.  See the commit message for how I think it could easily work without impacting functions that don't use it if you want to take a stab at implementing it.

virtuald:
It does seem that the precall/postcall mechanism would be ideal for implementing such a thing. Assuming that having an extra `unique_ptr` in the function object isn't a huge deal, that could work. I might look at this at the end of the week.

virtuald:
I looked a little bit at your commit, and I don't believe that mechanism would be quite what one would need. It would release the gil too early, as it would presumably be needed for any arg casting that might happen?. Same thing with the return value.

jagerman:
Hmm, I think you're right: the precall would be fine for initializing it, as long as it was the last precall, but it needs to expire between the function call and the casting of the return value.

(There's also another problem added since that commit: with #643, the function_call might not expire right away: there are overload cases where it is saved to try again a second time).


Another idea that might work: a wrapper generator so that you could write something like:

```C++
m.def("f", gil_released(&f))
```

`gil_release()` would return a lambda with the same signature as `f`, and simply take out a `gil_scoped_release` then forward all arguments to the real `f`.  Basically what you're doing manually, but with some clever templates to do it all for you.

virtuald:
Alright, so I wrote a wrapper function that seems to do the job (inspired by http://stackoverflow.com/questions/30679445/python-like-c-decorators and the cpp_function implementation) for my use cases:

``` c++
namespace pybind11 {

    // normal functions
    template <typename Return, typename... Args>
    std::function<Return(Args...)> release_gil(Return (*f)(Args ...) PYBIND11_NOEXCEPT_SPECIFIER) {
        return [f](Args... args) -> Return {
            gil_scoped_release r;
            return f(args...);
        };
    }
    
[...]

PR #740, opened 2017-03-16
Add a scope guard call policy

dean0x7d:
This is a proposed solution to #625 + a bit of generalization to allow any scope guard. The basic idea is that the following:
```c++
m.def("foo", foo, py::call_guard<T>());
```
is equivalent to:
```c++
m.def("foo", [](Args&&... args) {
    T guard{};
    return foo(std::forward<Args>(args)...);
});
```

To release the GIL, one would pass `py::call_guard<py::gil_scoped_release>()` (could also be aliased to something shorter). In general, any default-constructible RAII class could be used.

### Implementation

It compiles to nothing when the guard isn't used so there's no binary size or runtime overhead. It mainly uses the type system and integrates with the existing wrapper created for all functions.

An alternative would be to create a `cpp_function`-like wrapper as proposed in #625 and there are definitely pros and cons to both approaches. The main reason for submitting this PR is that the implementation uses minimal code compared to actual function wrapping and argument forwarding.


jagerman:
Nice approach, I like it!

wjakob:
Cool, I like it as well!

dean0x7d:
OK, the documentation is in place. Feel free to review.

The only thing I couldn't find out is how to check the GIL state with PyPy (it doesn't seem to be supported via the C API) so that test is skipped there, but the generic guard test covers the basic functionallity.

wjakob:
This looks great, the documentation is very clear. 

I think it's only a matter of time until people will try to mix multiple call guards for a single bound function (just like one can specify multiple ``keep_alive`` policies). I'm wondering is there is a more general implementation of the ``extract_call_guard_t`` interface that would allow this to work?

wjakob:
(This could be something like a filter over ``Extras`` that assembles a ``std::tuple`` with all of the call guards)

dean0x7d:
That's a good point about multiple guards. I forgot to document how that would work with the current implementation.

Unfortunately, I don't think that a `std::tuple` or any parameter pack based implementation is viable. I had considered it originally, but the issue is that the construction and destruction order of tuple elements is not specified. Same goes for argument initialization order in a function call (if multiple guards were expanded where there is currently one). This could yield some surprises if users assume that `call_guard<T1>, call_guard<T2>` will follow that order but it's actually random depending on the platform (with pretty nasty results, e.g. if `T2` releases the GIL, but `T1` uses the CPython API).

A solution with well-defined order would be for users to create a simple composite guard:
```c++
struct CompositeGuard {
    T1 guard1;
    T2 guard2;
};
```

So perhaps just add that to the documentation + a `static_assert` to ensure only a simple `call_guard` is allowed.

dean0x7d:
Thinking about it a bit more, (ab)using `std::initializer_list` would work with well-defined guard order. But I think the implementation is on the ugly side:
```c++
Return call_impl(Func &&f, index_sequence<Is...>, std::initializer_list<int>);
```
```c++
call_impl<Return>(std::forward<Func>(f), indices{}, {((void) Guard{}, 0)...});
```

jagerman:
I have some trouble seeing why that would guarantee the guards aren't destroyed until the call is finished; wouldn't it be valid for a compiler to destroy them after evaluating the initializer list?

jagerman:
Something like:

```
template <typename Guard, typename...> struct call_guardian { Guard guard; };
template <typename Guard1, typename Guard2, typename... Guards>
struct call_guardian<Guard1, Guard2, Guards...> {
    Guard1 guard;
    call_guardian<Guard2, Guards...> next;
};
```
ought to be enough to guarantee FILO guard construction/destruction order.

include/pybind11/common.h
```diff
@@ -536,9 +536,13 @@ using is_template_base_of = decltype(is_template_base_of_impl<Base>::check((remo
 struct is_template_base_of : decltype(is_template_base_of_impl<Base>::check((remove_cv_t<T>*)nullptr)) { };
 #endif
 
+/// Check if T is an instantiation of the template `Class`. For example:
+/// `is_<std::shared_ptr, T>` is true if `T == std::shared_ptr<U>` where U can be anything.
+template <template<typename...> class Class, typename T> struct is_ : std::false_type { };
+template <template<typename...> class C, typename... Us> struct is_<C, C<Us...>> : std::true_type { };
```
jagerman: Perhaps rename this to `is_specialization`?  `is_` isn't immediately obvious.

dean0x7d:
> I have some trouble seeing why that would guarantee the guards aren't destroyed until the call is finished; wouldn't it be valid for a compiler to destroy them after evaluating the initializer list?

Temporary objects are cleaned up at the end of the full expression in which they appear -- essentially at `;`. This is what makes nested and chained calls work in C++ (reference lifetime extension doesn't propagate through function returns).

That being said, the init list is still ugly and I don't like. Your `call_guardian` idea is much better!

The only question I have now: `call_guard<T1, T2>` or `call_guard<T1>, call_guard<T2>`? The former is trivial to implement and it would make a terser syntax. The latter is more versatile since additional guards can be added at a later stage, but it does require passing `Extra...` through a filter metafunction. I assume the more versatile one is more desirable?

wjakob:
Both are fine with me. It sounds like the first variant is easier to implement and faster to compile, so I'd lean towards that one.

jagerman:
I thought I commented on that, but apparently not: I like the first variant more as well, unless there's some compelling alternative use you have in mind for extra parameters.

dean0x7d:
OK, I've implemented the `call_guard<T1, T2, ...>` version and changed `first_of_t` with `exactly_one_t` (implementation in terms of the existing `constexpr` functions + `pack_element` which may be useful for other things as well).

include/pybind11/common.h
dean0x7d: True, renamed. I kind of liked the naming as a prefix of some kind, e.g. `is_call_guard = is_<call_guard, T>`, but a proper name is more useful.

jagerman:
I can definitely use `pack_element` in #762; right now it [uses](https://github.com/pybind/pybind11/pull/762/commits/e2acf16d31f12724028bb2cddd722b702026495e#diff-7174f0e1ad0a58b4622057815d7f8199R1330) `std::tuple_element<N, std::tuple<...>>`, but since I don't actually need the tuple, that seemed rather heavy.

include/pybind11/pybind11.h
```diff
@@ -143,8 +143,11 @@ class cpp_function : public function {
             /* Override policy for rvalues -- usually to enforce rvp::move on an rvalue */
             const auto policy = detail::return_value_policy_override<Return>::policy(call.func.policy);
 
+            /* Function scope guard -- defaults to the compile-to-nothing `void_type` */
```
jagerman: I think `` `void_type` `` should be `` `call_guard<>` ``

tests/test_call_policies.cpp
```diff
@@ -0,0 +1,92 @@
+/*
+    tests/test_keep_alive.cpp -- keep_alive modifier (pybind11's version
+    of Boost.Python's with_custodian_and_ward / with_custodian_and_ward_postcall)
```
jagerman: The comment filename and description are obsolete now.  Perhaps:

    tests/test_call_policies.cpp -- keep_alive (pybind11's version of Boost.Python's
    with_custodian_and_ward / with_custodian_and_ward_postcall) and call_guard
    examples and tests.

include/pybind11/pybind11.h
jagerman: Scratch that, I see now: it's going to be `call_guard<>::type`, which is the `void_type`.

tests/test_call_policies.cpp
dean0x7d: Fixed, thanks.

include/pybind11/pybind11.h
dean0x7d: Yeah, the 0 and 1 argument versions just return the type directly since those are the most common cases. The additional wrappers start with 2 arguments and above.

dean0x7d:
Fixed the comment and squashed the commits. If there's nothing else, I'll merge this after the CI tests are done.

wjakob:
This looks great, please go ahead!

PR #742, opened 2017-03-16
Fail to compile if attempting to specifying MI bases via `class_` ctor parameters

### LINKED ISSUE #741: Cannot call parent method when use multiple inheritance with pybind11::class_
I am using Python 3.5 and VS 2015. The pybind11 is in the master branch.
When I use py::class_ to define multiple inheritance, like this.
```C++
	// namespace py = pybind11;
	
	auto PropertyGrid = py::class_<PyPropertyGrid>(m, "PropertyGrid")
		.def(py::init<py::object>(), "data"_a=None)
		.def("addCategory", &PyPropertyGrid::addCategory, "title"_a);

	auto PropertyGridPage = py::class_<PyPropertyGridPage>(m, "PropertyGridPage", PropertyGrid)
		.def(py::init<py::object>(), "data"_a = py::none());

	auto ListBox = py::class_<PyListBox>(m, "ListBox")
		.def(py::init<>())
		.def(py::init<py::capsule>())
		.def("insertItems", &PyListBox::insertItems)
		.def("setListener", &PyListBox::setListener);

// if delete the line bellow, typeinfo->simple_type become true and work well
	py::class_<PyPropertyGridListPage>(m, "PropertyGridListPage", ListBox, PropertyGridPage)
		.def(py::init<py::object>(), "data"_a = py::none());
```
And I call addCategory in PropertyGridPage instance, I got this:
TypeError: addCategory(): incompatible function arguments. The following argument types are supported:
    1. (self: fefactory_api.PropertyGrid, title: str) -> None

Invoked with: <fefactory_api.PropertyGridPage object at 0x000001EE51163DB0>, 'name'

fefactory_api is my module name, and issubclass(fefactory_api.PropertyGridPage, fefactory_api.PropertyGridPage) is true;

in cpp:
```C++
class PyPropertyGridListPage: public PyListBox, public PyPropertyGridPage
{
public:
	PyPropertyGridListPage(py::object data) : PyPropertyGridPage(data)
	{
		
	}
};
```

I debug and finally find this in cast.h
```C++
/* Case 2: multiple inheritance */
            /* Check if we can safely perform a reinterpret-style cast */
            if (tobj == typeinfo->type) {
                value = reinterpret_cast<instance<void> *>(src.ptr())->value;
                return true;
            }

            /* If this is a python class, also check the parents recursively */
            auto const &type_dict = get_internals().registered_types_py;
            bool new_style_class = PyType_Check((PyObject *) tobj);
            if (type_dict.find(tobj) == type_dict.end() && new_style_class && tobj->tp_bases) {
                auto parents = reinterpret_borrow<tuple>(tobj->tp_bases);
                for (handle parent : parents) {
                    bool result = load(src, convert, (PyTypeObject *) parent.ptr());
                    if (result)
                        return true;
                }
            }
```
After i modifying "type_dict.find(tobj) == type_dict.end()" into "type_dict.find(tobj) != type_dict.end()", it work on.

Is my usage wrong or is this a bug?

jagerman:
~~Currently we only allow multiple inheritance via `class_` template parameters; this extends the support to `class_` instances passed via the derived `class_` constructor arguments as well.~~

Edit: see discussion below.  We *can't* support MI via parameters in [the most useful case](http://pybind11.readthedocs.io/en/latest/advanced/misc.html#partitioning-code-over-multiple-extension-modules), so produce a compile time error instead.

Fixes #741

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -885,11 +885,21 @@ public:
 
     template <typename... Extra>
     class_(handle scope, const char *name, const Extra &... extra) {
-        detail::type_record record;
+        using namespace detail;
+
+        // MI can only be specified via class_ template options, not constructor parameters
+        static_assert(
+            none_of<is_pyobject<Extra>...>::value || // no base class arguments, or:
+            (   constexpr_sum(is_pyobject<Extra>::value...) == 1 && // Exactly one base
+                constexpr_sum(is_base<options>::value...)   == 0 && // no template option bases
+                none_of<std::is_same<multiple_inheritance, Extra>...>::value), // no multiple_inheritance attr
+            "Error: multiple inheritance bases must be specified via class_ template options");
+
+        type_record record;
         record.scope = scope;
         record.name = name;
         record.type = &typeid(type);
-        record.type_size = sizeof(detail::conditional_t<has_alias, type_alias, type>);
+        record.type_size = sizeof(conditional_t<has_alias, type_alias, type>);
         record.instance_size = sizeof(instance_type);
         record.init_holder = init_holder;
         record.dealloc = dealloc;
@@ -900,12 +910,12 @@ public:
         (void) unused;
 
         /* Process optional arguments, if any */
-        detail::process_attributes<Extra...>::init(extra..., &record);
+        process_attributes<Extra...>::init(extra..., &record);
 
-        detail::generic_type::initialize(record);
+        generic_type::initialize(record);
 
         if (has_alias) {
-            auto &instances = pybind11::detail::get_internals().registered_types_cpp;
+            auto &instances = get_internals().registered_types_cpp;
             instances[std::type_index(typeid(type_alias))] = instances[std::type_index(typeid(type))];
         }
     }
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -333,7 +333,7 @@ template <> struct process_attribute<arg_v> : process_attribute_default<arg_v> {
     }
 };
 
-/// Process a parent class attribute
+/// Process a parent class attribute.  Single inheritance only (class_ itself already guarantees that)
 template <typename T>
 struct process_attribute<T, enable_if_t<is_pyobject<T>::value>> : process_attribute_default<handle> {
     static void init(const handle &h, type_record *r) { r->bases.append(h); }
--- a/tests/test_multiple_inheritance.cpp
+++ b/tests/test_multiple_inheritance.cpp
@@ -31,18 +31,27 @@ struct MIType : Base12 {
 };
 
 test_initializer multiple_inheritance([](py::module &m) {
-    py::class_<Base1>(m, "Base1")
-        .def(py::init<int>())
-        .def("foo", &Base1::foo);
+    py::class_<Base1> b1(m, "Base1");
+    b1.def(py::init<int>())
+      .def("foo", &Base1::foo);
 
-    py::class_<Base2>(m, "Base2")
-        .def(py::init<int>())
-        .def("bar", &Base2::bar);
+    py::class_<Base2> b2(m, "Base2");
+    b2.def(py::init<int>())
+      .def("bar", &Base2::bar);
 
     py::class_<Base12, Base1, Base2>(m, "Base12");
 
     py::class_<MIType, Base12>(m, "MIType")
         .def(py::init<int, int>());
+
+    // Uncommenting this should result in a compile time failure (MI can only be specified via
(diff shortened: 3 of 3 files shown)


wjakob:
neat! That looks great (including your implementation)

wjakob:
Looks like MSVC doesn't like your template forward declaration.

dean0x7d:
As far as I have gathered, the main motivation for the alternative constructor-argument base specification is this: http://pybind11.readthedocs.io/en/latest/advanced/misc.html#partitioning-code-over-multiple-extension-modules

If that's the case, the `class_<...>` compile-time information is lost and it reverts to adding bases via `handle` and the original problem comes back. It would be useful to add an error message in that case.

jagerman:
Ah, right, that's going to be a problem.

dean0x7d:
If the compile-time type information is always required, perhaps it would be enough to require that MI must use the template parameter specification, limit the ctor arguments to single-inheritance-only and just report an error if > 1 `handle`s passed via ctor.

jagerman:
Yeah, that sounds reasonable.

jagerman:
Even better: it should be possible to make it a compile-time error.

wjakob:
@dean0x7d: I'm not sure I follow -- why is compile-time information needed for MI? The documentation you referenced talks about cyclic dependencies, but that's not the case here, no?

jagerman:
@wjakob - because MI requires a `static_cast<BaseN>(reinterpret_cast<Derived>(voidptr))`: unlike single inheritance, we can't `reinterpret_cast` directly to the `BaseN` because `BaseN` might not be at the pointer itself.

wjakob:
ah right.. I forgot about that. We need to create all the converters that let us jump through the hierarchy.

jagerman:
So while we *could* support it for actual `class_<Base, ...>` instances, it's probably not worth bothering with since we can't support it in the most useful base-class-as-argument case.

include/pybind11/pybind11.h
```diff
@@ -885,11 +885,21 @@ class class_ : public detail::generic_type {
...
     class_(handle scope, const char *name, const Extra &... extra) {
-        detail::type_record record;
+        using namespace detail;
+
+        // MI can only be specified via class_ template options, not constructor parameters
+        static_assert(
+            none_of<is_pyobject<Extra>...>::value || // no base class arguments, or:
+            (   constexpr_sum(is_pyobject<Extra>::value...) == 1 && // Exactly one base
+                constexpr_sum(is_base<options>::value...)   == 0 && // no template option bases
+                none_of<std::is_same<multiple_inheritance, Extra>...>::value), // no multiple_inheritance attr
```
dean0x7d: It's pretty awesome that this can be a compile-time error!

One minor point, and feel free to ignore this since it's more about code style. The first `none_of` and the first `constexpr_sum` are repeating a bit of work. Consider assigning the sum to a variable to reuse the result. In general, assigning to variables would make the code more self-documenting, avoiding the need for any extra comments. E.g.:

```c++
constexpr auto compiletime_bases = constexpr_sum(is_pyobject<options>::value...);
constexpr auto runtime_bases     = constexpr_sum(is_pyobject<Extra>::value...);
constexpr auto needs_multiple_inheritance = compiletime_bases + runtime_bases > 1 ||
                                            any_of<std::is_same<multiple_inheritance, Extra>...>();
static_assert(!(needs_multiple_inheritance && runtime_bases > 0),
              "Error: multiple inheritance bases must be specified via class_ template options");
```

(In addition to the variable names, the error message itself shows the overall intent.)

jagerman: There's one annoying issue (MSVC related, as usual) with doing it that way: MSVC considers all the constexpr variables unused if only used in the `static_assert`, and so issues a warning which breaks the build, and so separate variables would need a `(void) (compiletime_bases, runtime_bases, needs_mi);` to silence it, which is a bit ugly.

I thought it might also end up being potentially less work for the compiler this way: typically the number of base classes is 0, so short-circuiting avoids needing to compute everything but the first one.

jagerman:
Rebased to make sure this works with MSVC 2017 (MSVC 2015u3 was very fragile around the static_assert: changing the `constexpr_sum(..) == 0` bit to a `none_of<..>` made it sometimes fail to compile with a "not a constexpr value" error, and other times ICE at link time, depending on how exactly I wrote the expression.)

include/pybind11/pybind11.h
jagerman: Given the MSVC warning bug, I think I'll merge as is.

dean0x7d: Sure, better than adding workarounds.

PR #752, opened 2017-03-20
Nicer API to pass py::capsule destructor

wjakob:
The current way of specifying a destructor for ``py::capsule`` instances strikes me as quite inelegant (the Python C API leaks through, requiring the user to deal with  ``PyCapsule_GetPointer``). This PR changes the convention.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1032,6 +1032,16 @@ public:
             pybind11_fail("Could not set capsule context!");
     }
 
+    capsule(void (*destructor)()) {
+        m_ptr = PyCapsule_New(reinterpret_cast<void *>(destructor), nullptr, [](PyObject *o) {
+            auto destructor = reinterpret_cast<void (*)()>(PyCapsule_GetPointer(o, nullptr));
+            destructor();
+        });
+
+        if (!m_ptr)
+            pybind11_fail("Could not allocate capsule object!");
+    }
+
     template <typename T> operator T *() const {
         T * result = static_cast<T *>(PyCapsule_GetPointer(m_ptr, nullptr));
         if (!result) pybind11_fail("Unable to extract capsule contents!");
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1004,10 +1004,27 @@ public:
     PYBIND11_OBJECT_DEFAULT(capsule, object, PyCapsule_CheckExact)
     PYBIND11_DEPRECATED("Use reinterpret_borrow<capsule>() or reinterpret_steal<capsule>()")
     capsule(PyObject *ptr, bool is_borrowed) : object(is_borrowed ? object(ptr, borrowed) : object(ptr, stolen)) { }
-    explicit capsule(const void *value, void (*destruct)(PyObject *) = nullptr)
-        : object(PyCapsule_New(const_cast<void*>(value), nullptr, destruct), stolen) {
-        if (!m_ptr) pybind11_fail("Could not allocate capsule object!");
+
+    explicit capsule(const void *value)
+        : object(PyCapsule_New(const_cast<void *>(value), nullptr, nullptr), stolen) {
+        if (!m_ptr)
+            pybind11_fail("Could not allocate capsule object!");
     }
+
+    explicit capsule(const void *value, void (*destructor)(void *)) {
+        m_ptr = PyCapsule_New(const_cast<void *>(value), nullptr, [](PyObject *o) {
+            auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
+            void *ptr = PyCapsule_GetPointer(o, nullptr);
+            destructor(ptr);
+        });
+
+        if (!m_ptr)
+            pybind11_fail("Could not allocate capsule object!");
+
+        if (PyCapsule_SetContext(m_ptr, (void *) destructor) != 0)
+            pybind11_fail("Could not set capsule context!");
+    }
+
     template <typename T> operator T *() const {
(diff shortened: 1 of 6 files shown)


virtuald:
I'm already using it in a [module deployed to pypi](https://github.com/robotpy/robotpy-cscore/blob/master/src/_cscore.cpp#L25).. any chance you can add a macro to detect the API change? :(

wjakob:
If it becomes part of 2.1 or 2.2, you can of course #ifdef based on ``PYBIND11_VERSION_MAJOR`` and ``PYBIND11_VERSION_MINOR``

include/pybind11/pytypes.h
```diff
@@ -1004,10 +1004,27 @@ class capsule : public object {
...
-        : object(PyCapsule_New(const_cast<void*>(value), nullptr, destruct), stolen) {
-        if (!m_ptr) pybind11_fail("Could not allocate capsule object!");
+
+    explicit capsule(const void *value)
+        : object(PyCapsule_New(const_cast<void *>(value), nullptr, nullptr), stolen) {
+        if (!m_ptr)
+            pybind11_fail("Could not allocate capsule object!");
+    }
+
+    explicit capsule(const void *value, void (*destructor)(void *)) {
```
dean0x7d: `explicit` isn't needed here.

include/pybind11/pytypes.h
```diff
@@ -1004,10 +1004,27 @@ class capsule : public object {
     PYBIND11_OBJECT_DEFAULT(capsule, object, PyCapsule_CheckExact)
     PYBIND11_DEPRECATED("Use reinterpret_borrow<capsule>() or reinterpret_steal<capsule>()")
     capsule(PyObject *ptr, bool is_borrowed) : object(is_borrowed ? object(ptr, borrowed) : object(ptr, stolen)) { }
-    explicit capsule(const void *value, void (*destruct)(PyObject *) = nullptr)
```
dean0x7d: Maybe just deprecate this constructor for backward compatibility? (Just without the `= nullptr` default and `explicit`).

wjakob: good idea!

include/pybind11/pytypes.h
jagerman: I was thinking it would be nice to make this take a function with a (deduced) `T *` argument, with a reinterpret cast of the capsule pointer to the deduced type when invoking the destructor.  Unfortunately I think that would require some special handling to accept a lambda, so might not be worthwhile.

wjakob: That's what I had implemented at first (with an extra closure object stored in the capsule context), but it struck me as a bit overkill.

wjakob: (In any case, that sort of thing can still be added later on if there a good case for this functionality can be made. This PR is about preventing fallout from a problematic API that is about to be documented.)

wjakob:
Thanks @dean0x7d, I added a deprecated constructor now. I'll merge this into 2.1 unless there are objections.

jagerman:
No objections.

dean0x7d:
I'm not sure how common the use case of #733 is -- would it warrant adding a dedicated constructor? E.g. `explicit capsule(void (*destructor)())` where "dummy" pointer would be handled internally.

jagerman:
I think it would be nice to have.  I don't understand why Python imposes `!= NULL` on the pointer in the first place, but I don't see why we should be subject to the same limitation.

virtuald:
Well, the only reason why #733 uses a capsule is because module destructors aren't exposed. If they were, then it would just be added to wherever the module docs are and wouldn't need to worry about capsules. 

Of course, only Python 3 supports module destructors IIRC, so you'd need this hack anyways -- but maybe it could be done internally?

jagerman:
Module destructors, using this as a workaround for Python 2, seems like a useful addition, but can be added later (i.e. for 2.2).

jagerman:
This is the last 2.1 PR; I think it's good as is, but the dummy-pointer version might also be useful.  (Though I think it can wait as well, if you prefer to merge without it).

wjakob:
I had the same thought and just added it.

wjakob:
Added docs & the destructor-only capsules.

wjakob:
That's a pretty weird PyPy failure -- it looks like the capture object is not working.

jagerman:
Rebase against current master: I just pushed a change to use pypy 5.7 instead of the 5.8 nightly in the travis-ci test, which fixes the PyPy build.

wjakob:
thanks -- done.

jagerman:
I just turned on this cool new travis-ci setting for pybind11 PR builds: https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation

wjakob:
Cool -- that's a useful feature indeed!

jagerman:
Hmm, that didn't fix the PyPy test failure.  I wonder if you could work around it by not using a capture, perhaps by setting/checking a static variable?

dean0x7d:
Doesn't PyPy need a couple of GC calls after a `del`?

wjakob:
that's right -- I forgot about that.

tests/test_python_types.py
```diff
@@ -512,3 +512,25 @@ def test_builtins_cast_return_none():
...
+    assert capture.unordered == """
+        creating capsule
+        destructing capsule
+    """
+
+    with capture:
+        a = m.return_capsule_with_destructor_2()
+        del a
+        pytest.gc_collect()
+    print(capture)
```
dean0x7d: Leftover from debugging.

wjakob:
that cancel feature is already starting to pay off :)

wjakob:
Ok, that did it!

PR #588, opened 2017-01-04
Add check for matching holder_type when inheriting

pschella:
One frequently encountered source of errors are mismatching holder types between bases and derived classes. If either the base has `unique_ptr` holder type and the derived class `shared_ptr` (or vice versa) a segfault results.

Tracking holder types for all classes in a large codebase can be cumbersome and error prone.

This pull request is an example of how import-time checks for matching holder types can be added. I welcome your comments.

PR #753, opened 2017-03-20
Throw an exception when attempting to load an incompatible holder

### LINKED ISSUE #751: Different holder type unchecked in argument loading
Though I know it might be considered a case of "stupid code, don't write that", I just found the following way of segfaulting a pybind11 extension module:

```C++
#include <iostream>

struct Test {
	int x = 42;
};

namespace py = pybind11;

PYBIND11_PLUGIN(test) {
	py::module m("test");
	
	py::class_<Test /* implicitly held by std::unique_ptr */>(m, "Test")
		.def(py::init<>())
	;
	
	m.def("f", [] (std::shared_ptr<Test> test) {
		std::cout << "Here" << std::endl;
		std::cout << test->x << std::endl;
		std::cout << "There" << std::endl;
	});
	
	return m.ptr();
}
```

Calling function `f` with an argument of type `Test` does not result in a type exception, but segfaults Python:

```Python
>>> import test
>>> test.f(test.Test())
Here
42
There
Segmentation fault (core dumped)
```
(Note that you might need to repeat this, as it seems that sometimes the 'undefined behaviour' causes nothing to happen. I am guess this is the case because the layouts of `std::unique_ptr` and `std::shared_ptr` overlap and so the pointer itself is not invalid.)

As far as I see, the problem lies in the fact that 'copyable_holder_caster' does not check (and has no way of checking?) the type of the holder once the held type matches (i.e., [this line in cast.h](https://github.com/pybind/pybind11/blob/0f5ec0a87e3d3879cd70740422adf3e3728a9d84/include/pybind11/cast.h#L959). This is actually how I found this way of segfaulting, by trying to understand the casting mechanism, rather than writing such stupid code.

I am aware you have to make a rather severe programming mistake to trigger this, but it does not seem not to match the philosophy of pybind11 regarding type safety, and since it could not find a previous bug report on this, I reckoned this might be interesting.

dean0x7d:
Instead of a segfault. Fixes #751.

This covers the case of loading a custom holder from a default-holder instance. Attempting to load one custom holder from a different custom holder (i.e. not `std::unique_ptr`) yields undefined behavior, just as #588 established for inheritance. That should be enough to catch the most common errors.

I tried to keep the error message concise. May have overdone it. Is it informative enough?

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -919,6 +919,9 @@ public:
             return true;
         }
 
+        if (typeinfo->default_holder)
+            throw cast_error("Unable to load a custom holder type from a default-holder instance");
+
         if (typeinfo->simple_type) { /* Case 1: no multiple inheritance etc. involved */
             /* Check if we can safely perform a reinterpret-style cast */
             if (PyType_IsSubtype(tobj, typeinfo->type))
--- a/tests/test_smart_ptr.cpp
+++ b/tests/test_smart_ptr.cpp
@@ -265,4 +265,10 @@ test_initializer smart_ptr_and_references([](py::module &pm) {
 
     py::class_<C, CustomUniquePtr<C>>(m, "TypeWithMoveOnlyHolder")
         .def_static("make", []() { return CustomUniquePtr<C>(new C); });
+
+    struct HeldByDefaultHolder { };
+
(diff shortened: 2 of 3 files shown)


wjakob:
This looks good to me -- the error message is fine.

PR #755, opened 2017-03-21
Support class-specific operator new and delete

### LINKED ISSUE #754: pybind11 doesn't seem to honor `operator new()` (e.g. for Eigen aligned allocators)
#### Issue description

I would like to expose classes in Python which have Eigen matrices as fields. Such classes need to be [allocated aligned to 16 bytes for vectorized operations](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html). A canonical way of ensuring this is to use Eigen's "operator new" implementation via the macro `EIGEN_MAKE_ALIGNED_OPERATOR_NEW`.

However it seems that pybind11 doesn't honor the alignment-aware `operator new()` when exposing such a class to Python, causing segfaults. Crashes seem to appear only when compiling with vectorized operations enabled.

This may be a feature request if pybind11 currently does not support a custom `operator new()`. In this case, could this be mentioned in the docs and/or a compile-time error be generated? It would be very helpful instead of having to trace the reason of a segfault.

Thanks a lot for your time and for the great tool!

See attachment [pybind11_eigenaligned.tar.gz](https://github.com/pybind/pybind11/files/856416/pybind11_eigenaligned.tar.gz). I'm assuming the host system has vectorized operations available via `-march=native`. Then, in python:

    >>> import pybind11_eigenaligned
    >>> pybind11_eigenaligned.Data().showB()
    Assertion failed: ((internal::UIntPtr(array) & (31)) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"), function plain_array, file /usr/local/include/eigen3/Eigen/src/Core/DenseStorage.h, line 128.
    Abort trap: 6

I'm running Mac OS Sierra with Apple's clang and Eigen 3.3.3. The problem occurs both with pybind11 2.0.1 as well as current pybind11 git master.

dean0x7d:
Fixes #754.

`void_t` was needed to work around a GCC 4.8 issue, but it's pretty useful to have anyway.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -174,6 +174,9 @@ struct type_record {
     /// How large is pybind11::instance<type>?
     size_t instance_size = 0;
 
+    /// The global operator new can be overridden with a class-specific variant
+    void *(*operator_new)(size_t) = ::operator new;
+
     /// Function pointer to class_<..>::init_holder
     void (*init_holder)(PyObject *, const void *) = nullptr;
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -25,6 +25,7 @@ inline PyTypeObject *make_default_metaclass();
 struct type_info {
     PyTypeObject *type;
     size_t type_size;
+    void *(*operator_new)(size_t);
     void (*init_holder)(PyObject *, const void *);
     void (*dealloc)(PyObject *);
     std::vector<PyObject *(*)(PyObject *, PyTypeObject *)> implicit_conversions;
--- a/include/pybind11/class_support.h
+++ b/include/pybind11/class_support.h
@@ -183,7 +183,7 @@ extern "C" inline PyObject *pybind11_object_new(PyTypeObject *type, PyObject *, 
     PyObject *self = type->tp_alloc(type, 0);
     auto instance = (instance_essentials<void> *) self;
     auto tinfo = get_type_info(type);
-    instance->value = ::operator new(tinfo->type_size);
+    instance->value = tinfo->operator_new(tinfo->type_size);
     instance->owned = true;
     instance->holder_constructed = false;
     get_internals().registered_instances.emplace(instance->value, self);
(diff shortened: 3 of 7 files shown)


jagerman:
Looks good to me.

dean0x7d:
I'd recommend this waits for v2.2 since it does affect a pretty key part of class support. Also might need to see about future-proofing for C++17 alignment-aware `new`.

jagerman:
C++17 support can definitely wait for 2.2, but I think the current PR fixes an important issue that is likely going to come up again in 2.1's stable lifetime (given the improved Eigen support in 2.1).

wjakob:
I've merged this now so that it can be part of 2.1

dean0x7d:
Note: Clang 4.0 doesn't compile in C++17mode due to `#ifdef __cpp_lib_void_t`. I'll push a minor fix shortly.

jagerman:
Whoops, didn't notice that.

Should we promote the gcc7/clang4 builds to non-optional?  (clang 4.0 is building with the stable released version now; gcc7 is still a snapshot, so perhaps just clang 4?)

dean0x7d:
I just pushed a fix for clang 4.0: f0e58a69d382724697f9d0154e94a59529a9d512

wjakob:
Just out of curiosity: what's the issue with __cpp_lib_void_t? Does Clang define it but then not offer ``void_t``?

jagerman:
No, it isn't working as its supposed to.  http://bugs.llvm.org/show_bug.cgi?id=26086 is the issue, I believe.

jagerman:
... which basically means LLVM libc++'s `void_t` is broken, since doing exactly this is the entire point of `std::void_t` in the first place.

dean0x7d:
Exactly that. The new-style implementation of `void_t` is `template<class...> using void_t = void;`, i.e. no need for a helper `struct`. The new libc++ implements it that way, but clang 4.0 doesn't implement [CWG1558](http://open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#1558) so it just ignores unused alias template arguments. The interesting thing is that this did work before, sometime in clang 3.x.

Anyway, I just removed `__cpp_lib_void_t` rather than adding extra compiler version checks since both gcc and clang are in `c++1z` mode for now.

PR #762, opened 2017-03-23
Make py::vectorize() ignore non-vectorizable arguments & support member functions

### LINKED ISSUE #761: vectorize() for class member functions?
Hi,

I'm currently trying to make some improvements to the healpy package (https://github.com/healpy/healpy), and pybind11 seems to be an almost perfect tool for the interfacing between Python and C++ that this package needs.
My first experiments went well, but then I hit the problem that the vectorize() functionality of pybind11 does not appear to support class member functions.

Assuming I have in C++

```
class foo {
  public:
    double bar(double arg);
};
```
I'd like to provide a Python function that takes either a scalar or a numpy array, calls foo::bar for all individual values, and then returns either a scalar or a numpy array. (The class object itself will always be a scalar.)
Is there a way to achieve this with current pybind11 versions? Naive application of vectorize() doesn't seem to work, like so:

```
  py::class_<foo>(m, "foo")
     .def("bar",py::vectorize(&foo::bar))
```

I'd be grateful for any tips!

jagerman:
Edit: PR amended from this initial description, see below.

This commit adds py::vectorize() support for class methods, so that you can write:

```C++
    class_<MyClass>(m, "MyClass")
        .def("method", py::vectorize(&MyClass::method));
```

Currently the only way to do this is via a lambda method wrapper with the exclusion argument technique described in the documentation.  That way still works—this PR also adds a test for it—and is still needed if you want to wrap a lambda method wrapper.

Fixes #761

### DIFF
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -239,27 +239,13 @@ by the compiler. The result is returned as a NumPy array of type
 The scalar argument ``z`` is transparently replicated 4 times.  The input
 arrays ``x`` and ``y`` are automatically converted into the right types (they
 are of type  ``numpy.dtype.int64`` but need to be ``numpy.dtype.int32`` and
-``numpy.dtype.float32``, respectively)
+``numpy.dtype.float32``, respectively).
 
-Sometimes we might want to explicitly exclude an argument from the vectorization
-because it makes little sense to wrap it in a NumPy array. For instance,
-suppose the function signature was
+.. note::
 
-.. code-block:: cpp
-
-    double my_func(int x, float y, my_custom_type *z);
-
-This can be done with a stateful Lambda closure:
-
-.. code-block:: cpp
-
-    // Vectorize a lambda function with a capture object (e.g. to exclude some arguments from the vectorization)
-    m.def("vectorized_func",
-        [](py::array_t<int> x, py::array_t<float> y, my_custom_type *z) {
-            auto stateful_closure = [z](int x, float y) { return my_func(x, y, z); };
-            return py::vectorize(stateful_closure)(x, y);
-        }
-    );
+    Only arithmetic, complex, and POD types passed by value or by ``const &``
+    reference are vectorized; all other arguments are passed through as-is.
+    Functions taking rvalue reference arguments cannot be vectorized.
 
 In cases where the computation is too complicated to be reduced to
 ``vectorize``, it will be necessary to create and access the buffer contents
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -504,6 +504,14 @@ struct is_input_iterator<T, void_t<decltype(*std::declval<T &>()), decltype(++st
 /// Ignore that a variable is unused in compiler warnings
 inline void ignore_unused(const int *) { }
 
+/// Apply a function over each element of a parameter pack
+#ifdef __cpp_fold_expressions
+#define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) (((PATTERN), void()), ...)
+#else
+using expand_side_effects = bool[];
+#define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
+#endif
+
 NAMESPACE_END(detail)
 
 /// Returns a named pointer that is shared among all extension modules (using the same
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -348,10 +348,12 @@ inline internals &get_internals();
 using std::enable_if_t;
 using std::conditional_t;
 using std::remove_cv_t;
+using std::remove_reference_t;
 #else
 template <bool B, typename T = void> using enable_if_t = typename std::enable_if<B, T>::type;
 template <bool B, typename T, typename F> using conditional_t = typename std::conditional<B, T, F>::type;
 template <typename T> using remove_cv_t = typename std::remove_cv<T>::type;
+template <typename T> using remove_reference_t = typename std::remove_reference<T>::type;
 #endif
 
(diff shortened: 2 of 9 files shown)


include/pybind11/numpy.h
```diff
@@ -1368,22 +1401,33 @@ template <typename T, int Flags> struct handle_type_name<array_t<T, Flags>> {
 
 NAMESPACE_END(detail)
 
-template <typename Func, typename Return, typename... Args>
-detail::vectorize_helper<Func, Return, Args...>
-vectorize(const Func &f, Return (*) (Args ...)) {
-    return detail::vectorize_helper<Func, Return, Args...>(f);
-}
```
jagerman: Moved this into `detail` as it didn't seem to be an intentionally exposed part of the API.

wjakob:
I'm wondering if this is the nicest way of solving the issue with #761. I think the cleanest solution would be for ``vectorize()`` to simply to act as a pass-through for non array arguments. This allows for a very general and much shorter implementation. If you want, I can invite you to another project that does something similar with pybind11 in the context of SIMD vectorization. (currently still in private mode, so I can't post a public link)

jagerman:
Yeah, I'm not entirely happy with how much needed to be added to make this work.  Making it skip anything non-array-convertible would indeed be nicer.

jagerman:
(Though it would still need some of this code--the current `vectorize` code won't accept a member function pointer).

jagerman:
I'm not really seeing how this could work in practice.

Say `vectorize(f)` is called with `double f(T1 a, T2 b)`; it will then generate a lambda-like callable object with `operator()(array_t<T1>, array_t<T2>)`, so the templated implementation has the needed array types.  So given arbitrary `T1` and `T2` types, what we would consider a non-array arguments?

wjakob:
Something like ``!std::is_arithmetic`` or more general: ``!std::is_pod``?

jagerman:
I've updated the PR; copying and pasting the main commit message here:

This extends py::vectorize to automatically pass through non-vectorizable arguments.  This removes the need for the documented "explicitly exclude an argument" workaround.

Vectorization now applies to arithmetic, std::complex, and POD types, passed as plain value or by const lvalue reference (previously only pass-by-value types were supported).  Non-const lvalue references and any other types are passed through as-is.

Functions with rvalue reference arguments (whether vectorizable or not) are explicitly prohibited: an rvalue reference is inherently not something that can be passed multiple times and is thus unsuitable to being in a vectorized function.

The vectorize returned value is also now more sensitive to inputs: previously it would return by value when all inputs are of size 1; this is now amended to having all inputs of size 1 *and* 0 dimensions.  Thus if you pass in, for example, [[1]], you get back a 1x1, 2D array, while previously you got back just the resulting single value.

Vectorization of member function specializations is now also supported via `py::vectorize(&Class::method)`; this required passthrough support for the initial object pointer on the wrapping function pointer.

This commit also removes the documentation about manually dealing with buffers; that feature is much better covered by the direct access implementation (which is the subsequent section).

docs/advanced/pycpp/numpy.rst
```diff
@@ -239,70 +239,13 @@ by the compiler. The result is returned as a NumPy array of type
...
-
-    // Vectorize a lambda function with a capture object (e.g. to exclude some arguments from the vectorization)
-    m.def("vectorized_func",
-        [](py::array_t<int> x, py::array_t<float> y, my_custom_type *z) {
-            auto stateful_closure = [z](int x, float y) { return my_func(x, y, z); };
-            return py::vectorize(stateful_closure)(x, y);
-        }
-    );
-
-In cases where the computation is too complicated to be reduced to
```
wjakob: It might still be good to keep the piece of code below as an complete self-contained example how how to deal with arrays when a function being vectorized is not a universal function.

include/pybind11/common.h
```diff
@@ -504,6 +512,14 @@ struct is_input_iterator<T, void_t<decltype(*std::declval<T &>()), decltype(++st
 /// Ignore that a variable is unused in compiler warnings
 inline void ignore_unused(const int *) { }
 
+/// Apply a function over each element of a parameter pack
+#ifdef __cpp_fold_expressions
+#define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) (((PATTERN), void()), ...)
```
wjakob: good idea!

include/pybind11/numpy.h
```diff
@@ -1390,83 +1403,143 @@ broadcast_trivial broadcast(const std::array<buffer_info, N> &buffers, ssize_t &
...
+    template <size_t Index> using param_n_t = typename pack_element<Index, typename vectorize_arg<Args>::call_type...>::type;
+
+    // Runs a vectorized function given arguments tuple and three index sequences:
+    //     - Index is the full set of 0 ... (N-1) argument indices;
+    //     - VIndex is the subset of argument indices with vectorized parameters, letting us access
+    //       vectorized arguments (anything not in this sequence is passed through)
+    //     - BIndex is a incremental sequence (beginning at 0) of the same size as VIndex, so that
+    //       we can store vectorized buffer_infos in an array (argument VIndex has its buffer at
+    //       index BIndex in the array).
+    template <size_t... Index, size_t... VIndex, size_t... BIndex> object run(
```
wjakob: cool, that's a nice trick (with ``select_indices``).

docs/advanced/pycpp/numpy.rst
jagerman: Yeah.  Looking at it now, I don't know why I removed it; I'm guessing it was a mistake of just removing too much along with the above.

jagerman: No, scratch that.  I took it out deliberately, but on second thought I see the merits in keeping it (it doesn't only apply to `py::array` arguments); I've added it back in.

wjakob:
LGTM!

PR #764, opened 2017-03-24
Arch-indep CMake packaging

ghisvail:
Since pybind11 is a header-only library, the CMake packaging does not have to carry any architecture specific checks. Without this patch, the detection of pybind11 will fail on 32-bit architectures if the project was built on a 64-bit machine and vice-versa. This fix is similar to what is applied to `Eigen` and other header-only C++ libraries.


dean0x7d:
Looks good to me. The only thing I'd suggest is adding a quick comment above those lines in the CMake file.

wjakob:
Looks good to me as well, and all agreed about adding a comment.

ghisvail:
There you go

PR #766, opened 2017-03-24
Support raw string literals as input for py::eval

dean0x7d:
A small convenience feature for `py::eval`.

**Before:**
```c++
py::eval<py::eval_statements>(
    "x = call_test();\n"
    "if x == 42:\n"
    "    print('Hello World!');\n"
    "else:\n"
    "    print('Bye!')",
    global, local
);
```

**After:**
```c++
py::eval<py::eval_statements>(R"(
    x = call_test()
    if x == 42:
        print('Hello World!')
    else:
        print('Bye!')
    )", global, local
);
```

### DIFF
--- a/docs/advanced/pycpp/utilities.rst
+++ b/docs/advanced/pycpp/utilities.rst
@@ -56,7 +56,9 @@ is always ``none``).
     // Evaluate the statements in an separate Python file on disk
     py::eval_file("script.py", scope);
 
-C++11 raw string literals are also supported and quite handy for this purpose:
+C++11 raw string literals are also supported and quite handy for this purpose.
+The only requirement is that the first statement must be on a new line following
+the raw string delimiter ``R"(``, ensuring all lines have common leading indent:
 
 .. code-block:: cpp
 
--- a/docs/advanced/pycpp/utilities.rst
+++ b/docs/advanced/pycpp/utilities.rst
@@ -55,3 +55,16 @@ is always ``none``).
 
     // Evaluate the statements in an separate Python file on disk
     py::eval_file("script.py", scope);
+
+C++11 raw string literals are also supported and quite handy for this purpose:
+
+.. code-block:: cpp
+
+    py::eval<py::eval_statements>(R"(
+        x = get_answer()
+        if x == 42:
+            print('Hello World!')
+        else:
+            print('Bye!')
+        )", scope
+    );
--- a/include/pybind11/eval.h
+++ b/include/pybind11/eval.h
@@ -11,8 +11,6 @@
 
 #pragma once
 
-#pragma once
-
 #include "pybind11.h"
 
 NAMESPACE_BEGIN(pybind11)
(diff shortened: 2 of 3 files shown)


wjakob:
Not so sure about this one! Calling ``textwrap`` from Python seems like a very heavy solution. None of this is necessary if the string is not indented.

dean0x7d:
Not indenting the raw string literal looks really ugly in my opinion:
```c++
// ... a couple of levels of indent in the preceding code
            // ...
            // C++ code
            // ...
            py::eval<py::eval_statements>(R"(
x = call_test()
if x == 42:
    print('Hello World!')
else:
    print('Bye!')
)", global, local
            );

            // ...
            // more C++
            // ...
```

An alternative could be to only do the dedent if the string starts with `\n` (check with `.attr("startswith")`).

aldanor:
I doubt `py::eval` would ever be used in performance parts of the code, its more of a once-off thing or a debug tool anyway? You could probably cache the `textwrap` function object with a static though to avoid import + getattr...

dean0x7d:
Caching with a static variable would cause problems with interpreter shutdown (the static object outlives the interpreter). I've also benchmarked a bit and it looks like the `import` + `attr` part is only about 10% of the total cost of the `dedent` so it doesn't seem worth caching. The entire `dedent()` call adds about 40% runtime overhead to `eval()` for the simplest one- or two-liners, but this overhead goes down for longer-running evals.

The new commit changes the behavior so that only raw string literals will go through `dedent()` thus avoiding the overhead for the simple stuff. What do you think?

In general, I agree with @aldanor that this isn't a performance sensitive area. A bit of syntax comfort feels more important.

wjakob:
This looks good to me -- restricting to raw strings and checking for "\n" reduces the likelihood of unexpected dedent calls.

jagerman:
> The new commit changes the behavior so that only raw string literals will go through ...

Isn't that going to match all string literals, not just raw ones (or perhaps that's what you meant by raw string literals)?  The check for "\n" on string literals seems the pertinent part.  As for that part--will it work correctly on a source file with CRLF linefeeds?

jagerman:
To answer the last bit myself: it seems that pretty much all compilers will treat a CRLF as a newline, and thus end up with a `\n`.  To do anything else would be pretty unreasonable.

wjakob:
Merged!

dean0x7d:
> @jagerman said:
> Isn't that going to match all string literals, not just raw ones (or perhaps that's what you meant by raw string literals)? The check for "\n" on string literals seems the pertinent part. As for that part--will it work correctly on a source file with CRLF linefeeds?

I'm a bit late with a reply, but it doesn't affect the PR in any case. Yeah, it matches all string literals, the "\n" check selects the `dedent()` path as that is the most common way to write the statements with raw literals (and no harm is done if regular literals also start with "\n"). All compilers use "\n" as the new line, independent of source line endings. Otherwise, the same source would generate different binaries.

Issue #749, opened 2017-03-20
Interpreter shutdown support


thorink:
In our project we have to perform some shutdown tasks. At the moment we use the Python API function `Py_AtExit` directly (Actually a pybind11 wrapper for this would be nice. Should I work on a MR for this?). I can attach a function that gets called when the python interpreter exits.

Now, when I tidy up remaining `py::object`s in the callback attached to `Py_AtExit`, the `handle::dec_ref` calls the Python API in the destructor of `py::object`. This is not allowed as stated in the CPython docs (https://docs.python.org/3.5/c-api/sys.html#c.Py_AtExit):

> Since Python’s internal finalization will have completed before the cleanup function, no Python APIs should be called by func.

Does anyone has a nice solution for this?

thorink:
Reading the release notes of v2.1 I found out about this module destructor trick ([link](http://pybind11.readthedocs.io/en/latest/advanced/misc.html?highlight=capsule#module-destructors)):

```c++
auto cleanup_callback = []() {
    // perform cleanup here -- this function is called with the GIL held
};

m.add_object("_cleanup", py::capsule(cleanup_callback));
```

Am I right, that this solves my issue? Is `cleanup_callback` called before the interpreter goes down but after all my python objects are destructed?

wjakob:
Yes, this should solve your issue. Can you give it a try and report back?

thorink:
I'm on it.

thorink:
It seems to work, though I needed to modify the example a bit because clang gave me a `error: non-const lvalue reference to type 'pybind11::object' cannot bind to a temporary of type 'py::capsule'`:

```c++
auto cleanup_callback = []() {
    // perform cleanup here -- this function is called with the GIL held
};

// does not compile:
// m.add_object("_cleanup", py::capsule(cleanup_callback));

// works:
py::capsule cap(cleanup_callback);
m.add_object("_cleanup", cap);
```

I'll set up an documentation fix MR.

wjakob:
@thorink: can you check if the code works with the latest master version that includes patch aa1b316?

thorink:
I've tested it with a patched v2.1 and it worked for me.

PR #768, opened 2017-03-27
module destructor example code clang fix

thorink:
clang does not like the example code, see #749

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -183,7 +183,8 @@ is possible to emulate it using Python capsules with a destruction callback.
         // perform cleanup here -- this function is called with the GIL held
     };
 
-    m.add_object("_cleanup", py::capsule(cleanup_callback));
+    py::capsule cleanup(cleanup_callback));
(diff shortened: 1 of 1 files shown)


wjakob:
IMO it would be nicer to fix add_object so that it can deal with an rvalue.

thorink:
Yes, you're right. I will work on it..

jagerman:
It's probably as simple as just changing the argument from `object &obj` to `object obj`.  You can even save the extra reference increment/decrement that would result if you simplify `add_object`'s:
```C++
        obj.inc_ref(); // PyModule_AddObject() steals a reference
        PyModule_AddObject(ptr(), name, obj.ptr());
```
to:
```C++
        PyModule_AddObject(ptr(), name, obj.release().ptr() /*steals a reference*/);
```

wjakob:
Taking a ``py::object`` increased the test suite by 32792 bytes (this pushes an ``inc_ref`` to all call sites). I switched to ``py::handle``.

See @aa1b316

thorink:
Okay, this MR is obsolete after @aa1b316.

PR #652, opened 2017-02-07
CMake installation path consistency

SylvainCorlay:
**Problem:**

1) In a given installation PREFIX, `pip install pybind11` or `conda install pybind11` will install pybind11 headers into a new `pybind11` subdirectory of the python include directory, while `cmake + make install` results in pybind11 headers being put into `PREFIX/include/pybind11`.

    For example, with a conda installation, pip will give

    `PREFIX/include/pythonX.Y/pybind11` while cmake yields `PREFIX/include/pybind11`.

2) It would be nice to have a consistent installation scheme between the two methods, plus the benefit of having the cmake project config files being made available, in the context of conda.

**Solution:**

This changes the destination of the headers of pybind into `PYTHON_INCLUDE_DIRS/pybind11`.

The generated project config cmake files are installed in the same location as before, but reflect the new  location of the headers.

PR #769, opened 2017-03-28
Missing CMakelist in source tarball

SylvainCorlay:
Another possibility is to have conda-forge fetch the tarball from GH.

@wjakob thoughts?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
 include include/pybind11/*.h
-include LICENSE README.md CONTRIBUTING.md
+include LICENSE README.md CONTRIBUTING.md CMakeLists.txt


wjakob:
Can you explain the context of these two PRs in more detail? I looked at both and quite frankly don't understand. (For instance, "just" the top level CMakeLists.txt is not enough to do anything, since it includes ``pybind11Tools``, etc.)

wjakob:
Why are there two PRs, and how are they related? What is this trying to accomplish?

SylvainCorlay:
The goal of the other PR is to implement what was described in https://github.com/pybind/pybind11/pull/652 i.e. enable the inclusion of the `pybind11Config.cmake` files  in the conda package.

SylvainCorlay:
But

SylvainCorlay:
I have updated the other PR for it to fetch the tarball from GH archives, and unlike the pypi tarball, it has everything.

wjakob:
ok, sounds good!

PR #771, opened 2017-03-28
Deprecate `object::borrowed`/`object::stolen`

### LINKED ISSUE #770: Unedfined symbol pybind11::object::borrowed
Hi devs!

Faced topic issue when building modules with Python bindings. Seems that this is related to my custom array class that is inherited from `pybind11::array_t`. It's a templated type that imports (inherits) all constructors from `array_t`, but couple of methods are only declared. Module1 contains explicit specializations of these methods (from my custom array) and have no problems.
Then, I use array in Module2 and it fails to load because of the issue above. I think the same would apply for `pybind11::object::stolen`, but I just didn't use it -- my array code contains `reinterpret_borrow<>`.

Right now I have a situation when release build *does not* have undefined symbol, but debug build *does*. In some other Module3 both release and debug builds have undefined symbol. Tested with gcc 6.3.1 and clang 3.9.1 -- results are the same.

Any ideas how to fix this?

jagerman:
~~Provides (via a dummy template parameter hack) inline definitions of object::borrowed/object::stolen.~~

The constexpr static instances can cause linking failures if the compiler doesn't optimize away the reference, as reported in #770.

There's no particularly nice way of fixing this in C++11/14: we can't inline definitions to match the declaration aren't permitted for non-templated static variables (C++17 *does* allows "inline" on variables, but that obviously doesn't help us.)

One solution that could work around it is to add an extra inherited subclass to `object`'s hierarchy, but that's a bit of a messy solution and was decided against in #771 in favour of just deprecating (and eventually dropping) the constexpr statics.

Fixes #770.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -714,16 +714,16 @@ public:
     using value_type = T;
 
     array_t() : array(0, static_cast<const T *>(nullptr)) {}
-    array_t(handle h, borrowed_t) : array(h, borrowed) { }
-    array_t(handle h, stolen_t) : array(h, stolen) { }
+    array_t(handle h, borrowed_t) : array(h, borrowed_t{}) { }
+    array_t(handle h, stolen_t) : array(h, stolen_t{}) { }
 
     PYBIND11_DEPRECATED("Use array_t<T>::ensure() instead")
-    array_t(handle h, bool is_borrowed) : array(raw_array_t(h.ptr()), stolen) {
+    array_t(handle h, bool is_borrowed) : array(raw_array_t(h.ptr()), stolen_t{}) {
         if (!m_ptr) PyErr_Clear();
         if (!is_borrowed) Py_XDECREF(h.ptr());
     }
 
-    array_t(const object &o) : array(raw_array_t(o.ptr()), stolen) {
+    array_t(const object &o) : array(raw_array_t(o.ptr()), stolen_t{}) {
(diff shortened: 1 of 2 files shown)


jagerman:
(Awaiting confirmation from #770 that this actually fixes it).

jagerman:
One comment: In C++17 this could be fixed by adding just:

```C++
inline constexpr object::borrowed_t object::borrowed;
inline constexpr object::stolen_t object::stolen;
```
which is, of course, much nicer than the template hack, but seeing as we need the hack for 14/11 already, I don't see much advantage in using `#ifdef`s to provide the alternative in C++17 mode.

jagerman:
I'm fine with that.  Do you mean heavyweight in terms of added compiler complexity, or something else?  (I did check that this doesn't change `sizeof(object)`).

dean0x7d:
I mean complex both for the compiler and for humans. The only advantage is avoiding `_t{}` in the syntax, but the cost is multiple + private inheritance, template tricks, and a separate C++17 version. Definitely not worth it when the goal is simple-as-can-be tag dispatch.

jagerman:
Okay; updated the PR to deprecate `borrowed`/`stolen` and replace all internal use with `borrowed_t{}`/`stolen_t{}`.

dean0x7d:
Looks good to me.

wjakob:
Looks good -- merged!

Issue #700, opened 2017-02-27
Catching specific exception types via error_already_set


llchan:
If I want to perform the C++ equivalent of this python code:
```python
try:
    val = d['key']
except KeyError:
    return
```

I can do it like this, but it's rather verbose:
```c++
py::object val;
try {
  val = d["key"];
} catch (py::error_already_set& exc) {
  exc.restore();
  if (PyErr_ExceptionMatches(PyExc_KeyError)) {
    PyErr_Clear();
    return;
  } else {
    throw py::error_already_set();
  }
}
```

Is there a more idiomatic way to do this? If not, could we modify the type hierarchy and/or the `error_already_set` mechanism to allow for something closer to this?
```c++
py::object val;
try {
  val = d["key"];
} catch (const py::key_error& exc) {
  return;
}
```

llchan:
@romanvm @dean0x7d thanks for adding the helper func. Can we also add a mention of this in the docs?

dean0x7d:
@llchan It's mentioned in the [API reference for error_already_set](http://pybind11.readthedocs.io/en/master/reference.html#_CPPv217error_already_set).

llchan:
I see, looks good, but I was thinking we could also add something in the tutorial-ish section here:
http://pybind11.readthedocs.io/en/master/advanced/exceptions.html

dean0x7d:
Ah, right, it does seem like that section is overwhelmingly about C++ -> Python exception translation, but not the other way around. If you have something in mind for it, would you be willing to submit a PR?

PR #772, opened 2017-03-31
Add a method to check Python exception types

romanvm:
This commit adds `error_already_set::matches()` convenience method to
check if the exception trapped by `error_already_set` matches a given
Python exception type. This will address #700 by providing a less
verbose way to check exceptions:
```c++
py::dict val;
try {
  val = d["key"];
} catch (py::error_already_set& exc) {
  if (exc.matches(PyExc_KeyError)) {
    exc.clear();
    return;
  } else {
    throw;
  }
}
```

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -597,6 +597,9 @@ public:
     /// Clear the held Python error state (the C++ `what()` message remains intact)
     void clear() { restore(); PyErr_Clear(); }
 
+    /// Check if the trapped exception matches a given Python exception class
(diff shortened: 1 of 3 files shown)


dean0x7d:
Looks pretty convenient! The only suggestion I have is to turn the example code into a test and add it to `tests/test_exceptions.cpp`.

romanvm:
Done.

wjakob:
LGTM

dean0x7d:
Merged. Thanks @romanvm!

PR #773, opened 2017-03-31
Wrong msvc version

SylvainCorlay:
190024210 is for update 3
190024215 is for update 3 with Cumulative Servicing Release

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -50,7 +50,7 @@
 #elif defined(_MSC_VER)
 // Pybind hits various compiler bugs in 2015u2 and earlier, and also makes use of some stl features
 // (e.g. std::negation) added in 2015u3:
-#  if _MSC_FULL_VER < 190024213
+#  if _MSC_FULL_VER < 190024210
 #    error pybind11 requires MSVC 2015 update 3 or newer
 #  endif
 #endif


SylvainCorlay:
cc @wjakob you probably want to get this one in shortly and push out a `2.1.1`...

jagerman:
Do you have easy access to a machine with stock 2015u3 installed?  If so, can you confirm that the full test suite builds and runs successfully in 190024210?  (In other words, do we depend on anything fixed between u3 and u3+CSR?)  AppVeyor checks obviously can't tell that, since AppVeyor runs with the update installed.

JohanMabille:
vs_tb1

jagerman:
@wjakob - thoughts re: 2.1.1 release for this?

SylvainCorlay:
@wjakob would you guys consider making a release with that fix?

The current version number check prevent the compilation of pybind11 for a lot of people on windows even with MSVC update 3.

wjakob:
@SylvainCorlay: ok, done -- feel free to merge the feedstock PR once the CI tests have passed.

SylvainCorlay:
The issue is that this comes after the backward incompatible removal of `borrow` and `stolen`, so 2.1.1 would not be a good number, unless you branch out 2.1.0, include the patch and release.

SylvainCorlay:
Actually creating a 2.x branch sounds reasonable, so that you can make bug-fix releases for 2.x, while working on 3.0 in the master branch.

wjakob:
@SylvainCorlay 
Actually I just cherry-picked the MSVC version check -- otherwise it is completely identical to 2.1.0. However, there is some weirdness with SHA256 checksums that's preventing the feedstock from completing.

dean0x7d:
@wjakob On a semi-related note, the stable version of the docs is outdated: http://pybind11.readthedocs.io/en/stable/changelog.html. Updating the stable branch to v2.1 requires a force push so I wanted to check in, rather than just do it. As an alternative, RTD considers the newest tagged release to be stable so the branch isn't strictly needed and can be simply removed -- RTD would automatically update when new tags are pushed.

SylvainCorlay:
Btw, thanks a lot for making this release.

wjakob:
@dean0x7d: I've pushed v2.1.1 to stable.

PR #779, opened 2017-04-05
Fix -Wmissing-braces warning

aldanor:
@jagerman ^


jagerman:
Good catch.  Not sure why I didn't get a warning for it, though.

PR #666, opened 2017-02-13
Expose enum_ entries as new "__members__" attribute

mdcb:
Brings convenience of being able to list the entries making an Enum. Side effect simplifies the .export_values() implementation.

PR #781, opened 2017-04-05
WIP: Python 3 enums

### LINKED ISSUE #530: enum_ QoL improvements and enum.IntEnum
I've been using Python 3 `enum.IntEnum` lately instead of pybind11 `enum_<>` enums (although they're less straightforward to bind), since the former are (a) faster and (b) more ergonomic/Pythonic. 

I wonder if we could either improve `enum_` a little (a lot?), and/or provide a shortcut to bind standard library enums on Python 3?

I'll try to prove my point using this mini-example:

```cpp

#include <stdexcept>

enum class Side1 : char { Right = 1, Left = -1 };
enum class Side2 : char { Right = 1, Left = -1 };

template<> struct type_caster<::Side1> {
    static handle cast(const ::Side1& src, return_value_policy, handle) {
        // this is the only flaky-ish part here; a better way to share `test_enums.Side1`?
        static object cls = module::import("test_enums").attr("Side1");
        if (!cls) throw std::runtime_error("unable to import test_enums.Side1");

        static object v_right = cls.attr("Right"), v_left = cls.attr("Left");

        switch (src) {
        case ::Side1::Right: return v_right.inc_ref();
        case ::Side1::Left: return v_left.inc_ref();
        default: return {};
        }
    }

    static PYBIND11_DESCR name() { return _("Side1"); }
};
} }

PYBIND11_PLUGIN(test_enums) {
    namespace py = pybind11;
    py::module m("test_enums");
    m.attr("Side1") = py::module::import("enum").attr("IntEnum")
        ("Side1", py::dict("Right"_a=1, "Left"_a=-1));
    py::enum_<Side2>(m, "Side2")
        .value("Right", Side2::Right).value("Left", Side2::Left);
    m.def("f1", []() { return Side1::Left; });
    m.def("f2", []() { return Side2::Left; });
    return m.ptr();
}
```

1) Casting is 3x-4x faster:

```python
>>> from test_enums import *
>>> %timeit f1()
10000000 loops, best of 3: 84.7 ns per loop
>>> %timeit f2()
1000000 loops, best of 3: 305 ns per loop
```

2) Comparison is 4x-5x faster:

```python
>>> e1, e2 = f1(), f2()
>>> %timeit e1 == e1
10000000 loops, best of 3: 40 ns per loop
>>> %timeit e2 == e2
1000000 loops, best of 3: 212 ns per loop
```

3) singleton values versus new instances:

```python
>>> f1() is f1()
True
>>> f2() is f2()
False
```

4) `enum.IntEnum` type:

```python
>>> list(Side1)
[<Side1.Left: -1>, <Side1.Right: 1>]  # ease of enumeration over values
>>> dict(Side1.__members__)
{'Left': <Side1.Left: -1>, 'Right': <Side1.Right: 1>}
```

5) `enum_` values:

```python
>>> Side2.Left.Left.Left.Left.Left.Left.Left  # stop showing up in my tab completion!
Side2.Left
>>> int(Side2.Left)
TypeError: __int__ returned non-int (type str)  # because `char` caster, I guess...
>>> isinstance(Side2.Left, int)
False
```

6) `enum.IntEnum` values:

```python
>>> isinstance(Side1.left, int)  # no need to reinvent the wheel, IntEnum is a subclass of int
True
>>> Side1.Left.name
'Left'
>>> Side1.Left.value
-1
```

aldanor:
This adds support for wrapping C/C++ enums as Python 3's `enum.IntEnum` (with `@unique` decorator applied). It would also work with `enum34` backport. If neither is available, `ImportError` will be thrown. The existing `enum_` wrappers are untouched and should work as before.

- [ ] Don't derive `enum_` from `class_`
- [x] Allow `value().value().def().def()` API, like the existing `enum_<>`
- [ ] Add full tests for `class_interface<>` API
- [ ] Add support for `py::arithmetic`
- [x] Add support for `.export_values()`
- [ ] Rename to `enum34`
- [ ] Implement automatic compile-time selection between `enum_` and `enum34`
- [ ] Update docs

(Closes #530)

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -443,3 +443,46 @@ The entries defined by the enumeration type are exposed in the ``__members__`` p
            ...
 
     By default, these are omitted to conserve space.
+
+Python 3 enumerations
+=====================
+
+Python 3 provides support for rich enumeration types in ``enum`` module [#enum]_ of the
+standard library; it can be also used in Python 2 by installing ``enum34`` backport package.
+
+In order to expose a C++ enumeration type as a subclass of ``enum.IntEnum`` in Python,
+one can use the :class:`py3_enum` wrapper whose interface is similar to that of :class:`enum_`:
+
+.. code-block:: cpp
+
+    enum class Captain {
+        Kirk = 0,
+        Picard
+    };
+
+The binding code will look like this:
+
+.. code-block:: cpp
+
+    py::py3_enum<Captain>(m, "Captain")
+        .value("Kirk", Captain::Kirk)
+        .value("Picard", Captain::Picard);
+
+The corresponding Python type is a subclass of ``enum.IntEnum`` and, as a result, is also
+a subclass of ``int``:
+
+.. code-block:: pycon
+
+    >>> Captain.mro()
+    [<enum 'Captain'>, <enum 'IntEnum'>, int, <enum 'Enum'>, object]
+    >>> list(Captain)
+    [<Captain.Kirk: 0>, <Captain.Picard: 1>]
+    >>> int(Captain.Picard)
+    1
+    >>> Captain.Kirk.name, Captain.Kirk.value
+    ('Kirk', 0)
+
+For more details on ``IntEnum`` functionality, see the official docs [#intenum]_.
+
+.. [#enum] https://docs.python.org/3/library/enum.html
+.. [#intenum] https://docs.python.org/3/library/enum.html#intenum
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -950,6 +950,77 @@ protected:
     std::tuple<make_caster<Tuple>...> value;
 };
 
+struct py3_enum_info {
+    handle type = {};
+    std::unordered_map<long long, handle> values = {};
+
+    py3_enum_info() = default;
+
+    py3_enum_info(handle type, const dict& values) : type(type) {
+        for (auto item : values)
+            this->values[static_cast<long long>(item.second.cast<int>())] = type.attr(item.first);
+    }
+
+    static std::unordered_map<std::type_index, py3_enum_info>& registry() {
+        static std::unordered_map<std::type_index, py3_enum_info> map = {};
+        return map;
+    }
+
+    template<typename T>
+    static void bind(handle type, const dict& values) {
+        registry()[typeid(T)] = py3_enum_info(type, values);
+    }
+
+    template<typename T>
+    static const py3_enum_info* get() {
+        auto it = registry().find(typeid(T));
+        return it == registry().end() ? nullptr : &it->second;
+    }
+};
+
+template<typename T>
+struct type_caster<T, enable_if_t<std::is_enum<T>::value>> {
+private:
+    using base_caster = type_caster_base<T>;
+    base_caster caster;
+    bool py3 = false;
+    T value;
+
+public:
+    template<typename U> using cast_op_type = pybind11::detail::cast_op_type<U>;
+
+    operator T*() { return py3 ? &value : static_cast<T*>(caster); }
+    operator T&() { return py3 ? value : static_cast<T&>(caster); }
+
+    static handle cast(const T& src, return_value_policy rvp, handle parent) {
+        if (auto info = py3_enum_info::get<T>()) {
+            auto it = info->values.find(static_cast<long long>(src));
+            if (it == info->values.end())
+                return {};
+            return it->second.inc_ref();
+        }
+        return base_caster::cast(src, rvp, parent);
+    }
+
+    bool load(handle src, bool convert) {
+        if (!src)
+            return false;
+        if (auto info = py3_enum_info::get<T>()) {
+            py3 = true;
+            if (!isinstance(src, info->type))
+                return false;
+            value = static_cast<T>(src.cast<long long>());
+            return true;
+        }
+        py3 = false;
+        return caster.load(src, convert);
+    }
+
+    static PYBIND11_DESCR name() {
+        return base_caster::name();
+    }
+};
+
 /// Helper class which abstracts away certain actions. Users can provide specializations for
 /// custom holders, but it's only necessary if the type has a non-standard interface.
 template <typename T>
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -950,52 +950,45 @@ protected:
     std::tuple<make_caster<Tuple>...> value;
 };
 
+template<typename U>
 struct py3_enum_info {
     handle type = {};
-    std::unordered_map<long long, handle> values = {};
+    std::unordered_map<U, handle> values = {};
 
     py3_enum_info() = default;
 
     py3_enum_info(handle type, const dict& values) : type(type) {
-        for (auto item : values)
-            this->values[static_cast<long long>(item.second.cast<int>())] = type.attr(item.first);
-    }
-
-    static std::unordered_map<std::type_index, py3_enum_info>& registry() {
-        static std::unordered_map<std::type_index, py3_enum_info> map = {};
-        return map;
-    }
-
-    template<typename T>
-    static void bind(handle type, const dict& values) {
-        registry()[typeid(T)] = py3_enum_info(type, values);
-    }
-
-    template<typename T>
-    static const py3_enum_info* get() {
-        auto it = registry().find(typeid(T));
-        return it == registry().end() ? nullptr : &it->second;
+        for (auto item : values) {
+            this->values[item.second.cast<U>()] = type.attr(item.first);
+        }
     }
 };
 
 template<typename T>
 struct type_caster<T, enable_if_t<std::is_enum<T>::value>> {
+    using underlying_type = typename std::underlying_type<T>::type;
+
 private:
     using base_caster = type_caster_base<T>;
+
+    static std::unique_ptr<py3_enum_info<underlying_type>>& py3_info() {
+        static std::unique_ptr<py3_enum_info<underlying_type>> info;
+        return info;
+    }
+
     base_caster caster;
-    bool py3 = false;
     T value;
 
 public:
     template<typename U> using cast_op_type = pybind11::detail::cast_op_type<U>;
 
-    operator T*() { return py3 ? &value : static_cast<T*>(caster); }
-    operator T&() { return py3 ? value : static_cast<T&>(caster); }
+    operator T*() { return py3_info() ? &value : static_cast<T*>(caster); }
+    operator T&() { return py3_info() ? value : static_cast<T&>(caster); }
 
     static handle cast(const T& src, return_value_policy rvp, handle parent) {
-        if (auto info = py3_enum_info::get<T>()) {
-            auto it = info->values.find(static_cast<long long>(src));
-            if (it == info->values.end())
+        if (py3_info()) {
+            auto it = py3_info()->values.find(static_cast<underlying_type>(src));
+            if (it == py3_info()->values.end())
                 return {};
             return it->second.inc_ref();
         }
@@ -1005,19 +998,21 @@ public:
     bool load(handle src, bool convert) {
         if (!src)
             return false;
-        if (auto info = py3_enum_info::get<T>()) {
-            py3 = true;
-            if (!isinstance(src, info->type))
+        if (py3_info()) {
+            if (!isinstance(src, py3_info()->type))
                 return false;
-            value = static_cast<T>(src.cast<long long>());
+            value = static_cast<T>(src.cast<underlying_type>());
             return true;
         }
-        py3 = false;
         return caster.load(src, convert);
     }
 
     static PYBIND11_DESCR name() {
         return base_caster::name();
+    }
+
+    static void bind(handle type, const dict& values) {
+        py3_info().reset(new py3_enum_info<underlying_type>(type, values));
     }
 };
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1293,7 +1293,7 @@ private:
 };
 
 template<typename T>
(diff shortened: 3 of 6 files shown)


aldanor:
Upon interpreter shutdown something gets decrefed too many times, I think :/

```
    frame #3: 0x00000001002ab072 Python`Py_FatalError(msg="PyThreadState_Get: no current thread") + 130 at pythonrun.c:2712
    frame #4: 0x00000001002a8f9a Python`PyThreadState_Get + 202 at pystate.c:477
    frame #5: 0x0000000100285d39 Python`PyErr_Fetch(p_type=0x00007fff5fbfecd0, p_value=0x00007fff5fbfecc8, p_traceback=0x00007fff5fbfecc0) + 25 at errors.c:337
    frame #6: 0x00000001002b0ec2 Python`_Py_PrintFatalError(fd=2) + 34 at pythonrun.c:2599
    frame #7: 0x00000001002ab063 Python`Py_FatalError(msg="PyThreadState_Get: no current thread") + 115 at pythonrun.c:2684
    frame #8: 0x00000001002a8f9a Python`PyThreadState_Get + 202 at pystate.c:477
    frame #9: 0x000000010014e216 Python`tupledealloc(op=0x000000010319a630) + 38 at tupleobject.c:232
    frame #10: 0x0000000100130eca Python`_Py_Dealloc(op=0x000000010319a630) + 58 at object.c:1749
    frame #11: 0x00000001000ce305 Python`code_dealloc(co=0x00000001031bd100) + 293 at codeobject.c:365
    frame #12: 0x0000000100130eca Python`_Py_Dealloc(op=0x00000001031bd100) + 58 at object.c:1749
    frame #13: 0x00000001000f46ee Python`func_dealloc(op=0x00000001031c16d0) + 318 at funcobject.c:555
    frame #14: 0x0000000100130eca Python`_Py_Dealloc(op=0x00000001031c16d0) + 58 at object.c:1749
    frame #15: 0x0000000104006864 pybind11_tests.so`pybind11::handle::dec_ref(this=0x0000000104572018) const & + 148 at pytypes.h:159
    frame #16: 0x0000000104008df1 pybind11_tests.so`pybind11::object::~object(this=0x0000000104572018) + 17 at pytypes.h:200
    frame #17: 0x0000000104004685 pybind11_tests.so`pybind11::object::~object(this=0x0000000104572018) + 21 at pytypes.h:200
```

jagerman:
Thoughts on having this replace the current `py::enum_` implementation (with the current code preserved as a `detail`-level fallback implementation when `enum.IntEnum` isn't available) instead of having dual enum implementations?  2.1 had some additions (in #666) to make `enum_` behave more like enum.IntEnum already; this seems like a logical next step.

It would break some potential backwards compatibility in that you can't use `enum_` as a `class_` anymore, but I don't think that was documented to begin with, nor does it seem particularly useful for an enum implementation.

I definitely like the idea of storing a the values in a static variable in the `type_caster<EnumType>`.  It would also let you store the keys as a `std::underlying_type_t<EnumType>` (rather than the `long long`--which might have a problem if someone stashed a (size_t) -1 in an enum).

jagerman:
> Upon interpreter shutdown something gets decrefed too many times...

It's caused by the `static object ctor` and `unique`.  Changing them to private, non-static local members with the import+attr as default values fixes it.  (I don't see much point in having them static: it seems pretty unlikely that a `py3_enum<T>` instance for a given T is going to be instantiated a second time for the same T, and if it is, an extra import+attr is not a big deal).

jagerman:
I just had a thought about recreating the object each time you call `value()`: you could, perhaps, make `value()` return a temporary object which itself has a `value()` method that records new values, but doesn't call `update()` until destruction.  Something like this (untested):

```C++
template <typename T>
class py3_enum {
...
    class value_loader {
    private:
        py3_enum &owner;
        friend class py3_enum;
        value_loader(py3_enum &e) : owner{e} {}
    public:
        value_loader &value(const char *name, T value) {
            owner.entries[name] = cast(static_cast...);
            return *this;
        }
        ~value_loader() { owner.update(); }
    };

    value_loader value(const char *name, T value) {
        value_loader vl(*this);
        vl.value(name, value);
        return vl;
    }
...
};
```

Then the common approach:
```C++
py_enum3<MyEnum>(m, "MyEnum")
    .value("v1", MyEnum.v1)
    .value("v2", MyEnum.v2)
    .value("v3", MyEnum.v3)
    ;
```
will only do the `IntEnum` construction once.  (I suppose it *could* do it twice if it doesn't apply RVO, but that shouldn't be a big deal).

aldanor:
> It's caused by the static object ctor and unique. Changing them to private, non-static local members with the import+attr as default values fixes it.

Thanks -- yes, that fixes it.

> I just had a thought about recreating the object each time you call value(): you could, perhaps, make value() return a temporary object which itself has a value() method that records new values, but doesn't call update() until destruction.

This could work if you still call `update()` once in `py3_enum`'s ctor, so that it works for empty enums. You could also potentially move `update()` into `~py3_enum()`: assuming you don't assign it to a variable, the dtor will be called before the next line of code; elsewise, will be called at the end of scope.

aldanor:
The only disadvantage of the current implementation, as I see it, is that you can't define methods / properties on the newly created enum class, because it's not a `class_<>`. In fact, `py3_enum` is not even an `object` (should it be, somehow?).

Hypothetically, here's also an alternative implementation: define py3 enums through `class_`, somewhat similar to the current `enum_`. This, however, will require two features of `class_` that we don't currently have:
- Accept a class dict to be used at the type instantiation site -- because many metaclasses (like those of enum types) use that. This could be also a useful thing to have in general. I tried it out in a local branch and it's a trivial addition. However, it's not very useful without having the next point implemented.
- A non-trivial part is actually calling metaclass's `__new__()` if it's provided, similar to what `builtins.__build__class__()` does -- we currently don't, so metaclass support is somewhat half-baked and is limited to post-class-creation use.

Also, in this case you won't be able to do `enum_<>().value().value().def()` anyway since the full dict of entries is needed at type instantiation point, so it would either have to be `enum_<>().value().value().finalize().def()`, or `enum_<>(...=..., ...=...).def()`.

aldanor:
@jagerman In a4381aa I've collapsed most of py3 enum impl into the `type_caster<>` which gets rid of the unnecessary map lookup on cast/load. Also, `py3_enum_info` is templated on the underlying type which gets rid of all `long long`.

aldanor:
I've fixed a few things like setting `__module__` / `__qualname__`, so the new enums now behave similar to existing types.

@jagerman I've also tried out a very cheesy way of allowing to define properties and methods on Python 3 enums  -- surprisingly it seemed to work, at least for simple cases I've tested -- so I pushed it as WIP in the last commit.

Basically, this seems to work:

```cpp
py3_enum<Foo>(m, "Foo")
    .value("A", Foo::A)
    .value("B", Foo::B)
    .extend() // requires rvalue *this
    .def_property_readonly("is_a", [](Foo x) { return x == Foo::A; });
```

aldanor:
Darn AppVeyor complaining all the time...

jagerman:
I wonder if we should just turn off C4456 through C4459 warnings within pybind code.  I don't really subscribe to the every-variable-must-be-unique idea of MS coding mentality that leads to silly things like prefixing all class members with some random character.

aldanor:
I'd be +1 for sure, those warnings are very opinionated.

wjakob:
> I wonder if we should just turn off C4456 through C4459 warnings within pybind code. 
Sound good to me.

wjakob:
Actually, on second thought: let's try to keep pybind11 clean with respect to these warnings. Other projects may choose to activate a similarly strict set of warnings, and it would be unfortunate if pybind11 causes issues in that context.

jagerman:
It won't cause external issues: the warnings are pushed at the beginning of pybind11.h and popped off at the end.

wjakob:
Right.. I was thinking of ``/wd...``-style parameters, which would be problematic. Pushing via a pragma sounds all good to me.

dean0x7d:
I'd keep the warnings. C++ core guidelines / `clang-tidy` flag the same shadowing issues: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es12-do-not-reuse-names-in-nested-scopes

jagerman:
Fair enough.

aldanor:
Guys, any ideas on this? I'd be willing to finish it (if anything) but there's a few unclear points.

jagerman:
Assuming you mean your original post checklist, I vote for calling it `enum34` (matching the Python backport name) if it's going to be distinct from `enum_` and public.

But I vote for it being neither distinct nor public.  Rather I think that `enum_` should continue to be the single public interface, with `py3_enum` and `old_enum` (or whatever) as implementation details: `enum_` would then map to a `py3_enum` when available, falling back transparently to `old_enum` if not available.

I'm not sure it's really worth it to worry about supporting additional definitions and attributes on the bound enum.  (Or in other words, I'm perfectly happy for this to be a documented limitation, possibly to be revisited later if someone has a strong use case for it).

dean0x7d:
I agree with @jagerman: `enum34` seems like an appropriate name for the public interface, but it would be even better if it was rolled into `enum_` with a Python >= 3.4 `ifdef` if this can be done in a fully backward-compatible way. Note that, `py::arithmetic()` would seem to which the binding behavior between `enum.Enum` and `enum.IntEnum`.

aldanor:
If we roll it into one `enum_`, should it be compile-time or runtime? The first option being presumably easier.

> @dean0x7d: if this can be done in a fully backward-compatible way

The biggest question re: backwards-compatibility here is what to do with `.def()` on enum types?

Basically, declaration semantics are slightly different as enum34 needs to be "finalised", so while this would work with the current `enum_`:

```cpp
py::enum_<Foo>(m, "Foo").value("A", Foo::A).def("f", ...).value("B", Foo::B).def("g", ...);
```

-- I can't see how to make it work with enum34 implementation.

dean0x7d:
> so while this should work with the current enum_:
> `py::enum_<Foo>(m, "Foo").value("A", Foo::A).def("f", ...).value("B", Foo::B);`

That doesn't actually compile with the current `enum_`. `.def` returns `class_&` which doesn't have `.value()`. So mixing in `class_` functions is already semi-broken and probably isn't there intentionally but just as a consequence of the implementation via `class_` base.

I think the more important things for maintaining compatibility are: `.export_values()`, `py::arithmetic` and the behavior of the resulting types in Python. If the new implementation can appear simply as an optimization selected at compile-time, that would be great.

aldanor:
Hmm, seems like I've managed to make this work:

```cpp
py3_enum<Foo>(m, "Foo").value(...).value(...).def(...).def(...)
```

-- however it required slight changes to `class_` (namely, all public methods have been moved out to a separate CRTP). All `class_::def_*()` methods should work, which brings it closer to the existing `enum_<>` behaviour.

If this is acceptable, I'll take a look at `export_values()`, `py::arithmetic` and other combat stuff.

Re: `py::arithmetic`, I'd personally like it to be a default option because it really makes most sense in Python as `IntEnum` is specifically geared for ints whereas `Enum` is a more generic option, however this differs from the `enum_<>` default, so I guess we'll have to go with `enum.Enum` as default?

jagerman:
Is there a significant value in being able to define custom methods on an enum?  It feels a bit like chasing backwards compatibility with a feature which was never really intended or advertised, which I think we'd be reasonably justified to just stop supporting given the complexity that it adds.

aldanor:
Well, idk, we actually have a few in our pybind11 codebase, it's quite handy -- mostly `def_property_readonly`, indicating some feature of an enum value, or whether it belong to a certain group/category ("is_this, is_that"). `__str__` / `__repr__` are also quite reasonable to overload.

dean0x7d:
If adding methods to `enum_` is a desirable feature, I'd suggest making that official and documented first (preferably in a separate PR for clarity). I feel like the current undocumented way of adding methods is unfortunate because of the broken chaining. Making it explicit and documented would be nice. Something like:

```c++
py::enum_<Foo>(m, "Foo")
    .value("First", Foo::First)
    .value("Second", Foo::Second)
    .classify() // `class_` functions permitted after this point; no more `enum_`
    .def("__str__", [](Foo const&) { ... });
```

The `enum_` implementation in this case would go from:
```c++
class enum_ : public class_ { ... };
```
to:
```c++
class enum_ {
public:
    class_& clasify() { return cls; }

private:
    class_ cls;
};
```

Using composition here instead of inheritance would simplify things overall since there would be no need for CRTP.

aldanor:
@dean0x7d That's definitely reasonable and involves less hacks. In fact, that's exactly what I did initially if you look at the commit history for this branch (the only difference was that the method was called `.extend()`). If we make `enum_<>` behave this way though, we'll break backwards-compat a bit so there's that.

Note that there are various edge cases with your suggested implementation like:

```cpp
auto e =py::enum_<Foo>(m, "Foo")
    .value("First", Foo::First)
    .value("Second", Foo::Second);
auto c = e.classify()
    .def("__str__", [](Foo const&) { ... });
e.value("Third", Foo::Third);  // ???
```

To avoid this, I tried marking `.extend()` with `&&`, but not sure that's the best solution. Any thoughts?

P.S. `.classify()` sounds a bit like a machine-learning term; maybe `into_class()` (in Rust's lingo), or `to_class()`?

aldanor:
@dean0x7d @jagerman I've posted suggested changes to the legacy `enum_<>` in #837. If that's all good, I'll backtrack this branch and remove CRTP hacks so it has the same API.

wjakob:
I'll close this PR.: While wrapping Python 3 enums would be nice, it was never quite clear how to do this in a backwards-compatible way that preserves the semantics of the current enums. pybind11 now uses a less bloaty way of creating `enums` that addresses some of the flaws of how this worked in the past.

PR #782, opened 2017-04-05
Python buffer objects can have negative strides.

crisluengo:
I have made changes such that strides are stored as `ssize_t`. I have added only one test case, I'd like advice on where else this feature needs to be tested.

Eigen does not allow for negative strides, and will throw if you pass a numpy array `[::-1]` to it. I don't think the interface needs to test for this anywhere, since Eigen will test. Unless there is a feature that can copy the array if the data doesn't fit?

### DIFF
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -41,8 +41,8 @@ completely avoid copy operations with Python expressions like
                 py::format_descriptor<float>::format(), /* Python struct-style format descriptor */
                 2,                                      /* Number of dimensions */
                 { m.rows(), m.cols() },                 /* Buffer dimensions */
-                { (ssize_t)( sizeof(float) * m.rows() ),/* Strides (in bytes) for each index */
-                  (ssize_t)( sizeof(float) ) }
+                { sizeof(float) * m.rows(),             /* Strides (in bytes) for each index */
+                  sizeof(float) }
             );
         });
 
@@ -118,11 +118,10 @@ as follows:
             /* Number of dimensions */
             2,
             /* Buffer dimensions */
-            { (size_t) m.rows(),
-              (size_t) m.cols() },
+            { m.rows(), m.cols() },
             /* Strides (in bytes) for each index */
-            { (ssize_t)( sizeof(Scalar) * (rowMajor ? m.cols() : 1) ),
-              (ssize_t)( sizeof(Scalar) * (rowMajor ? 1 : m.rows()) ) }
+            { sizeof(Scalar) * (rowMajor ? m.cols() : 1),
+              sizeof(Scalar) * (rowMajor ? 1 : m.rows()) }
         );
      })
 
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -41,8 +41,8 @@ completely avoid copy operations with Python expressions like
                 py::format_descriptor<float>::format(), /* Python struct-style format descriptor */
                 2,                                      /* Number of dimensions */
                 { m.rows(), m.cols() },                 /* Buffer dimensions */
-                { sizeof(float) * m.rows(),             /* Strides (in bytes) for each index */
-                  sizeof(float) }
+                { (ssize_t)( sizeof(float) * m.rows() ),/* Strides (in bytes) for each index */
+                  (ssize_t)( sizeof(float) ) }
             );
         });
 
@@ -61,7 +61,7 @@ specification.
         std::string format;
         int ndim;
         std::vector<size_t> shape;
-        std::vector<size_t> strides;
+        std::vector<ssize_t> strides;
     };
 
 To create a C++ function that can take a Python buffer object as an argument,
@@ -121,8 +121,8 @@ as follows:
             { (size_t) m.rows(),
               (size_t) m.cols() },
             /* Strides (in bytes) for each index */
-            { sizeof(Scalar) * (rowMajor ? m.cols() : 1),
-              sizeof(Scalar) * (rowMajor ? 1 : m.rows()) }
+            { (ssize_t)( sizeof(Scalar) * (rowMajor ? m.cols() : 1) ),
+              (ssize_t)( sizeof(Scalar) * (rowMajor ? 1 : m.rows()) ) }
         );
      })
 
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -57,10 +57,10 @@ specification.
 
     struct buffer_info {
         void *ptr;
-        size_t itemsize;
+        ssize_t itemsize;
         std::string format;
-        int ndim;
-        std::vector<size_t> shape;
+        ssize_t ndim;
+        std::vector<ssize_t> shape;
         std::vector<ssize_t> strides;
     };
 
@@ -95,8 +95,8 @@ buffer objects (e.g. a NumPy matrix).
                 throw std::runtime_error("Incompatible buffer dimension!");
 
             auto strides = Strides(
-                info.strides[rowMajor ? 0 : 1] / sizeof(Scalar),
-                info.strides[rowMajor ? 1 : 0] / sizeof(Scalar));
+                info.strides[rowMajor ? 0 : 1] / (py::ssize_t)sizeof(Scalar),
+                info.strides[rowMajor ? 1 : 0] / (py::ssize_t)sizeof(Scalar));
 
             auto map = Eigen::Map<Matrix, 0, Strides>(
                 static_cat<Scalar *>(info.ptr), info.shape[0], info.shape[1], strides);
@@ -111,17 +111,14 @@ as follows:
 
     .def_buffer([](Matrix &m) -> py::buffer_info {
         return py::buffer_info(
-            m.data(),                /* Pointer to buffer */
-            sizeof(Scalar),          /* Size of one scalar */
-            /* Python struct-style format descriptor */
-            py::format_descriptor<Scalar>::format(),
-            /* Number of dimensions */
-            2,
-            /* Buffer dimensions */
-            { m.rows(), m.cols() },
-            /* Strides (in bytes) for each index */
+            m.data(),                                /* Pointer to buffer */
+            sizeof(Scalar),                          /* Size of one scalar */
+            py::format_descriptor<Scalar>::format(), /* Python struct-style format descriptor */
+            2,                                       /* Number of dimensions */
+            { m.rows(), m.cols() },                  /* Buffer dimensions */
             { sizeof(Scalar) * (rowMajor ? m.cols() : 1),
               sizeof(Scalar) * (rowMajor ? 1 : m.rows()) }
+                                                     /* Strides (in bytes) for each index */
         );
      })
 
@@ -321,17 +318,17 @@ where ``N`` gives the required dimensionality of the array:
     m.def("sum_3d", [](py::array_t<double> x) {
         auto r = x.unchecked<3>(); // x must have ndim = 3; can be non-writeable
         double sum = 0;
-        for (size_t i = 0; i < r.shape(0); i++)
-            for (size_t j = 0; j < r.shape(1); j++)
-                for (size_t k = 0; k < r.shape(2); k++)
+        for (ssize_t i = 0; i < r.shape(0); i++)
+            for (ssize_t j = 0; j < r.shape(1); j++)
+                for (ssize_t k = 0; k < r.shape(2); k++)
                     sum += r(i, j, k);
         return sum;
     });
     m.def("increment_3d", [](py::array_t<double> x) {
         auto r = x.mutable_unchecked<3>(); // Will throw if ndim != 3 or flags.writeable is false
-        for (size_t i = 0; i < r.shape(0); i++)
-            for (size_t j = 0; j < r.shape(1); j++)
-                for (size_t k = 0; k < r.shape(2); k++)
+        for (ssize_t i = 0; i < r.shape(0); i++)
+            for (ssize_t j = 0; j < r.shape(1); j++)
+                for (ssize_t k = 0; k < r.shape(2); k++)
                     r(i, j, k) += 1.0;
     }, py::arg().noconvert());
 
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -15,31 +15,31 @@ NAMESPACE_BEGIN(pybind11)
 
 /// Information record describing a Python buffer object
 struct buffer_info {
-    void *ptr = nullptr;         // Pointer to the underlying storage
-    size_t itemsize = 0;         // Size of individual items in bytes
-    size_t size = 0;             // Total number of entries
-    std::string format;          // For homogeneous buffers, this should be set to format_descriptor<T>::format()
-    size_t ndim = 0;             // Number of dimensions
-    std::vector<size_t> shape;   // Shape of the tensor (1 entry per dimension)
+    void *ptr = nullptr;          // Pointer to the underlying storage
+    ssize_t itemsize = 0;         // Size of individual items in bytes
+    ssize_t size = 0;             // Total number of entries
+    std::string format;           // For homogeneous buffers, this should be set to format_descriptor<T>::format()
+    ssize_t ndim = 0;             // Number of dimensions
+    std::vector<ssize_t> shape;   // Shape of the tensor (1 entry per dimension)
     std::vector<ssize_t> strides; // Number of entries between adjacent entries (for each per dimension)
 
     buffer_info() { }
 
-    buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t ndim,
-                detail::any_container<size_t> shape_in, detail::any_container<ssize_t> strides_in)
+    buffer_info(void *ptr, ssize_t itemsize, const std::string &format, ssize_t ndim,
+                detail::any_container<ssize_t> shape_in, detail::any_container<ssize_t> strides_in)
     : ptr(ptr), itemsize(itemsize), size(1), format(format), ndim(ndim),
       shape(std::move(shape_in)), strides(std::move(strides_in)) {
-        if (ndim != shape.size() || ndim != strides.size())
+        if (ndim != (ssize_t) shape.size() || ndim != (ssize_t) strides.size())
             pybind11_fail("buffer_info: ndim doesn't match shape and/or strides length");
-        for (size_t i = 0; i < ndim; ++i)
+        for (size_t i = 0; i < (size_t) ndim; ++i)
             size *= shape[i];
     }
 
-    buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t size)
+    buffer_info(void *ptr, ssize_t itemsize, const std::string &format, ssize_t size)
     : buffer_info(ptr, itemsize, format, 1, {size}, {itemsize}) { }
 
     explicit buffer_info(Py_buffer *view, bool ownview = true)
-    : buffer_info(view->buf, (size_t) view->itemsize, view->format, (size_t) view->ndim,
+    : buffer_info(view->buf, view->itemsize, view->format, view->ndim,
             {view->shape, view->shape + view->ndim}, {view->strides, view->strides + view->ndim}) {
         this->view = view;
         this->ownview = ownview;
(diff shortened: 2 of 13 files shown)


jagerman:
Indeed, [Eigen doesn't](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747) (but might someday).  For Eigen, it should make a copy when the reference type is non mutable (e.g. `Eigen::Ref<const MatrixXd>`).  For mutable types (`Eigen::Ref<MatrixXd>`) obviously it needs to fail.

Making it work should be pretty easy: you'd just need to add a couple of "return false;" in `static EigenConformable<row_major> conformable(const array &a)`: one in the `if (dims == 2) { ... }` case if either stride is negative, and another after that if the 1D stride is negative.

jagerman:
Also take a look at the travis-ci failures: clang wants an explicit (ssize_t) cast on the shape in `strides[i] = strides[i - 1] * shape[i - 1];` in [numpy.h](https://travis-ci.org/pybind/pybind11/jobs/219072374#L246), and the python test code needs a space after the , to make flake8 happy.  (I've been bitten by that same style warning more than once).

crisluengo:
@jagerman, thanks so much for your pointers. This change was not as trivial as that, but it's working correctly now. I did not change `conformable` because that meant that no copies were made. Instead, I adjusted the `EigenConformable::stride_compatible` function. I also had to change one of the `type_caster` to force a copy on negative strides. That copy seems a little ugly to me, as it could mean that two copies get made, but I don't know how to solve this otherwise.

jagerman:
Needing to do *two* copies is definitely undesirable.  I think in that case we are better off doing an element-by-element copy, bypassing the Eigen::Map.  I think this will work:

```C++
        if (fits.negativestrides) {

            // Eigen does not support negative strides, so we need to make a copy here with normal strides.
            // TODO: when Eigen bug #747 is fixed, remove this if case, always execute the else part.
            // http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747
            value = Type(fits.rows, fits.cols);
            if (dims == 1) {
                auto input = buf.unchecked_reference<1>();
                ssize_t i = 0;
                for (EigenIndex c = 0; c < fits.cols; c++) for (EigenIndex r = 0; r < fits.rows; r++)
                    value(r, c) = input(i++);
            }
            else {
                auto input = buf.unchecked_reference<2>();
                for (EigenIndex c = 0; c < fits.cols; c++) for (EigenIndex r = 0; r < fits.rows; r++)
                    value(r, c) = input(r, c);
            }
        } else {

           value = Eigen::Map<const Type, 0, EigenDStride>(buf.data(), fits.rows, fits.cols, fits.stride);

        }
```

crisluengo:
Sorry, I don't understand. Then you're still doing a second copy. You are replacing copy #2 by a different copy. What we'd want to do is avoid the first call to `array_t::ensure` (COPY 1) below if the strides are negative, but I don't see how to do that.

Actually, coming to think of it, if COPY 1 happens, then the strides will not be negative, no? If so, COPY 2 will not happen. COPY 2 can only happen if the first `ensure` didn't need to copy. Is that right?

```cpp
    bool load(handle src, bool) {
        auto buf = array_t<Scalar>::ensure(src);  // *** COPY 1 ***
        if (!buf)
            return false;

        auto dims = buf.ndim();
        if (dims < 1 || dims > 2)
            return false;

        auto fits = props::conformable(buf);
        if (!fits)
            return false; // Non-comformable vector/matrix types

        if (fits.negativestrides) {

            // Eigen does not support negative strides, so we need to make a copy here with normal strides.
            // TODO: when Eigen bug #747 is fixed, remove this if case, always execute the else part.
            // http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747
            auto buf2 = array_t<Scalar,array::forcecast || array::f_style>::ensure(src); // *** COPY 2 ***
            if (!buf2)
                return false;
            // not checking sizes, we already did that
            fits = props::conformable(buf2);
            value = Eigen::Map<const Type, 0, EigenDStride>(buf2.data(), fits.rows, fits.cols, fits.stride);

        } else {

           value = Eigen::Map<const Type, 0, EigenDStride>(buf.data(), fits.rows, fits.cols, fits.stride);

        }

        return true;
    }
```

crisluengo:
I see the compile bots don't understand `np.testing.assert_raises`. I'll look for a different way to test exceptions.

jagerman:
COPY 1 generally only copies if type conversion is needed (e.g. you pass a numpy array of ints to an Eigen matrix of floats), or when what you pass in isn't an array at all (e.g. a list).  But if it happens, I believe you'll get something with default (row major) strides, which we can copy via the Eigen map.  (There are still two copies, but the first is also doing a conversion).

jagerman:
Re assert testing: lines 11-13 of test_buffers.py has an example (there are many others through the test suite).

jagerman:
Re: copying, instead of using Eigen to copy out of numpy, we could add something to tell numpy to copy into a destination, which would avoid a copy in some cases, and avoid the need for the special case for negative strides.

jagerman:
(I'm not asking you to do that, just thinking out loud)

crisluengo:
`Eigen::Map` doesn't copy, it just takes a pointer to the data and builds a `Matrix` object around that. Unless pybind is doing something more behind the scenes, there should only be one copy made if type conversion or stride normalization is needed.

jagerman:
Right, I was a bit imprecise: it's the assignment of the Eigen::Map to `value` that does the copy.

crisluengo:
I see what you mean now. It's a matrix that is passed to the called function, not the map, so a copy needs to be made.

So you are suggesting to create the matrix, give Python a buffer view of it, and ask python to copy the data into it? That sounds very reasonable. I don't think I'll have time to implement that, sorry.

jagerman:
I just filed a PR against your branch that makes numpy do the conversion.  (I've never filed a PR against a PR before--I'm hoping that if you merge it, it'll automatically show up here).

crisluengo:
Thank you so much, that looks pretty simple, but it would have taken me a few days to figure out how to do it. :)

docs/advanced/pycpp/numpy.rst
```diff
@@ -41,8 +41,8 @@ completely avoid copy operations with Python expressions like
                 py::format_descriptor<float>::format(), /* Python struct-style format descriptor */
                 2,                                      /* Number of dimensions */
                 { m.rows(), m.cols() },                 /* Buffer dimensions */
-                { sizeof(float) * m.rows(),             /* Strides (in bytes) for each index */
-                  sizeof(float) }
+                { (ssize_t)( sizeof(float) * m.rows() ),/* Strides (in bytes) for each index */
```
wjakob: strange indentation: the convention used elsewhere is ``(ssize_t) (sizeof(float) * m.rows())``

docs/advanced/pycpp/numpy.rst
```diff
@@ -121,8 +121,8 @@ as follows:
             { (size_t) m.rows(),
               (size_t) m.cols() },
             /* Strides (in bytes) for each index */
-            { sizeof(Scalar) * (rowMajor ? m.cols() : 1),
-              sizeof(Scalar) * (rowMajor ? 1 : m.rows()) }
+            { (ssize_t)( sizeof(Scalar) * (rowMajor ? m.cols() : 1) ),
```
wjakob: ditto

include/pybind11/eigen.h
```diff
@@ -246,8 +254,14 @@ struct type_caster<Type, enable_if_t<is_eigen_dense_plain<Type>::value>> {
...
     using props = EigenProps<Type>;
 
-    bool load(handle src, bool) {
-        auto buf = array_t<Scalar>::ensure(src);
+    bool load(handle src, bool convert) {
+        // If we're in no-convert mode, only load if given an array of the correct type
+        if (!convert && !isinstance<array_t<Scalar>>(src))
+            return false;
+
+        // Coerce into an array, but don't do type conversion yet; the copy below handle it.
```
wjakob: handle*s* it

crisluengo:
Thanks, I'll fix these things tonight.

jagerman:
There's another issue here that needs addressing: the current PR is breaking backwards compatibility for `buffer_info`/`array`/`array_t` constructors that take the strides as a `std::vector<size_t>`: you're changing them all to `std::vector<ssize_t>`, but that's going to break existing pybind-using code that builds a `std::vector<size_t>` for strides.

I've submitted PR #788 which should address this.  It's going to conflict with this PR (because it replaces all the `std::vector<size_t>` constructors with templated ones), unfortunately, but hopefully the conflicts aren't too hard to sort out.

jagerman:
(I don't mind fixing the conflicts in #788 if we want to merge this one first--but I think we need both commits to go more or less together).

crisluengo:
That's a really good point. The other public interface changes are the return type of `array::strides()`, `array::strides(size_t)`, `array::offset_at`, `array::index_at`. The first one will break existing code, the other three will have implicit casts (though hopefully with a compiler warning). I don't see any ways around this, except maybe adding a preprocessor macro that causes these functions to test for the stride sign and return unsigned values.

I don't think it will be a huge amount of work to sort out the conflicts, so I wouldn't mind having to fix this PR if you merge yours first. Whatever you prefer.

jagerman:
The return type of `strides` is less of a concern: if you have code that relies on strides being a `size_t`, that's broken code (essentially the point of this PR).  Cases where it breaks code are cases that are likely already broken and need fixing, so it's not *that* big a problem.

The input is more of an issue because forcing people to change the `size_t` in code such as
```C++
std::vector<size_t> str{{ 1, 2, 3 }};
```
is both likely to be more common, and isn't a change that causes any harm, so it's harder to justify breaking the API unnecessarily.

Besides all that, it felt a bit restrictive to me to require a specific vector type for input: if my strides were already in a C-array or `std::list` or something, I have to mess around copying it into a vector, which is just annoying.

dean0x7d:
Any thoughts on fully committing to signed types for all buffer/array values? (itemsize, ndim, shape)

Having a mix of signed/unsigned can be accident-prone. E.g. if a user commits a cast while multiplying a negative stride with an unsigned `itemsize`, they're going to have a bad time. Having all signed values would also be nicely consistent, both internally and with Python/Numpy APIs.

(This may, however, raise additional backward compatibility issues.)

jagerman:
That's a good point.  I think, given that this is already going to unavoidably break backwards compatibility, we might as well go all the way.  Promotion to unsigned values probably won't break anything for addition and subtraction, but will definitely break multiplication or division, such as:

```C++
auto itemstride = a.strides()[0] / a.itemsize(); // With -16/8 I get itemstride = 536870910
```

It is also going to *reduce* compatibility issues in at least some cases by avoiding compiler warnings in calling code for mixed arithmetic.

crisluengo:
It does require adding some tests for `size < 0`, etc.  But on the positive side, it'll reduce the number of explicit casts in the code, and will avoid unpleasant surprises with mixed arithmetic.

Both Eigen and Python use signed integers all the way through, I believe.

jagerman:
Eigen's is configurable at compile time, but yes, the default is signed.

dean0x7d:
> Eigen's is configurable at compile time, but yes, the default is signed.

Funny thing about that option in Eigen: Much like the Ford Model T, Eigen lets you pick any index type you want, as long as it's a signed integer :)

jagerman:
> as long as it's a signed integer :)

Oh, you're right.  I remember a long and interesting thread on Eigen's mailing list a couple months ago about this, but had forgotten that "Eigen doesn't currently support unsigned indices" was the starting point of the thread.

jagerman:
#782 is merged (sorry for the conflicts!)

crisluengo:
Nice! It was pretty easy to fix the conflicts. Do you need me to rebase and/or squish commits?

docs/advanced/pycpp/numpy.rst
```diff
@@ -41,8 +41,8 @@ completely avoid copy operations with Python expressions like
                 py::format_descriptor<float>::format(), /* Python struct-style format descriptor */
                 2,                                      /* Number of dimensions */
                 { m.rows(), m.cols() },                 /* Buffer dimensions */
-                { sizeof(float) * m.rows(),             /* Strides (in bytes) for each index */
-                  sizeof(float) }
+                { (ssize_t)(sizeof(float) * m.rows()),  /* Strides (in bytes) for each index */
+                  (ssize_t)(sizeof(float)) }
```
jagerman: The `(ssize_t)` cast shouldn't be needed here anymore.

docs/advanced/pycpp/numpy.rst
```diff
@@ -121,8 +121,8 @@ as follows:
             { (size_t) m.rows(),
               (size_t) m.cols() },
             /* Strides (in bytes) for each index */
-            { sizeof(Scalar) * (rowMajor ? m.cols() : 1),
-              sizeof(Scalar) * (rowMajor ? 1 : m.rows()) }
+            { (ssize_t)(sizeof(Scalar) * (rowMajor ? m.cols() : 1)),
+              (ssize_t)(sizeof(Scalar) * (rowMajor ? 1 : m.rows())) }
```
jagerman: ... or here.

include/pybind11/buffer_info.h
```diff
@@ -36,11 +36,11 @@ struct buffer_info {
     }
 
     buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t size)
-    : buffer_info(ptr, itemsize, format, 1, size, itemsize) { }
+    : buffer_info(ptr, itemsize, format, 1, size, (ssize_t) itemsize) { }
```
jagerman: ... or here (and so on).

jagerman:
A rebase + squash into fewer commits will be needed eventually, but it can probably wait until it's ready to merge.

I think this looks pretty good now: the one outstanding bit is whether we should apply the same treatment to `shape` and `itemsize`.  I think we should, to avoid forcing dealing with mixed signed/unsigned integer types on the user any time they use a shape and stride in the same equation.  It does break the numpy API a little more, but we're already breaking it a little here (primarily by changing the stride return type), and I'm worried that not doing it introduces a potential source of obscure bugs if callers aren't very careful with shape/stride types.

crisluengo:
It is probably better to make it all the same type. I'm not sure how much time that will take, I'll try to do it tomorrow or Monday. I'm afraid it'll require adding tests for negativity in multiple places, and it'll be easy to miss those places...

crisluengo:
I have made `shape`, `size`, `ndim` and `itemsize` into signed integers. The one that bugs me most is `itemsize`, since `sizeof` returns an unsigned integer, which causes all sorts of warnings if not explicitly cast. On the other hand, since both Python and Euler use signed integers everywhere, there's a lot fewer casts in the pybind11 code now.

jagerman:
This looks great!  (I haven't yet gone through everything, but I think this is getting close).  @wjakob @dean0x7d @aldanor thoughts?

I also built this against the 2.1.1 test suite, just to see what would happen, and it (almost*) passed, albeit with a slew of comparison between signed and unsigned warnings from where the tests are using `size_t`s in loops.  So backwards compatibility looks decent, at least as far as the test suite goes.

(* - almost in that there were two acceptable failures: the vectorize test that directly calls into py::detail::broadcast didn't compile (it needed types changed to `ssize_t`), which is fine, it's in `detail`; and there was an unrelated pytest failure (from the #791 fix)).

docs/advanced/pycpp/numpy.rst
```diff
@@ -37,7 +37,7 @@ completely avoid copy operations with Python expressions like
        .def_buffer([](Matrix &m) -> py::buffer_info {
             return py::buffer_info(
                 m.data(),                               /* Pointer to buffer */
-                sizeof(float),                          /* Size of one scalar */
+                (py::ssize_t)sizeof(float),             /* Size of one scalar */
```
dean0x7d: Are you sure an explicit cast is necessary here? As far as I know, compilers don't issue warnings for compile-time constants which are guaranteed to fit into the destination type:

```c++
#include <cstddef>

void foo(int) {}

int main() {
    foo(sizeof(float)); // clean (no warnings)
    size_t s = 4;
    foo(s);             // <-- conversion warning here
}
```

GCC, clang and MSVC for example: https://godbolt.org/g/zgibas

This might help avoid some of the casts on the user side.

crisluengo:
@dean0x7d , thanks so much for the code review! Will fix these things and solve conflicts with main branch today.

include/pybind11/numpy.h
```diff
@@ -1252,10 +1257,11 @@ broadcast_trivial broadcast(const std::array<buffer_info, N> &buffers, size_t &n
...
         // Check for C contiguity (but only if previous inputs were also C contiguous)
         if (trivial_broadcast_c) {
-            size_t expect_stride = buffers[i].itemsize;
+            ssize_t expect_stride = buffers[i].itemsize;
             auto end = buffers[i].shape.crend();
-            for (auto shape_iter = buffers[i].shape.crbegin(), stride_iter = buffers[i].strides.crbegin();
-                    trivial_broadcast_c && shape_iter != end; ++shape_iter, ++stride_iter) {
+            auto shape_iter = buffers[i].shape.crbegin();
+            auto stride_iter = buffers[i].strides.crbegin();
+            for (; trivial_broadcast_c && shape_iter != end; ++shape_iter, ++stride_iter) {
```
crisluengo: We changed the strides to signed first, which meant that shape and stride were different types. It's no longer necessary.

include/pybind11/class_support.h
```diff
@@ -349,7 +349,7 @@ inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type) {
 #endif
     type->tp_flags |= Py_TPFLAGS_HAVE_GC;
     type->tp_dictoffset = type->tp_basicsize; // place dict at the end
-    type->tp_basicsize += sizeof(PyObject *); // and allocate enough space for it
+    type->tp_basicsize += (Py_ssize_t)sizeof(PyObject *); // and allocate enough space for it
```
crisluengo: GCC with -Wsign-conversion does give a warning there without the cast. Assuming 2's complement, the operation yields the same result regardless, but I think it's better to do the operation as a signed addition.

include/pybind11/numpy.h
```diff
@@ -767,6 +769,9 @@ template <typename T, int ExtraFlags = array::forcecast> class array_t : public
     explicit array_t(ShapeContainer shape, const T *ptr = nullptr, handle base = handle())
         : array(std::move(shape), ptr, base) { }
 
+    explicit array_t(size_t count, const T *ptr = nullptr, handle base = handle())
```
dean0x7d: Shouldn't this be `ssize_t`?

crisluengo: I was thinking of that. This is a new constructor from the merge with the master branch. I wasn't sure what to do with it. If `count` is `ssize_t`, do we need a test for non-negativity?

jagerman: It's not exactly new ([here's the same constructor in 2.1.1](https://github.com/pybind/pybind11/blob/v2.1/include/pybind11/numpy.h#L513)), it just disappeared between #788 and #822.  I tried to handle it in #788 by having any_container accept single values, but that broke 2-element initializer_list construction (due to ambiguity; see #822), so #822 reintroduced the older singleton value constructors.

Yes, I think throwing an exception here for a negative value is warranted; `std::invalid_argument`?

dean0x7d: Numpy should already be doing this check for us in `PyArray_NewFromDescr`. Perhaps `pybind11_fail` should just be replaced with `error_already_set` to relay the original error message.

jagerman: +1 for changing the `pybind11_fail` to `error_already_set` in the main constructor.

crisluengo: Is this what you mean?https://github.com/pybind/pybind11/pull/782/commits/a06f2f782d55f6dc2881305ee145c336ce0fec37

jagerman: Yup.  I suggest also adding a one-line test that tries to create with -1, just to verify that the exception is thrown and comes through properly, but that should work.

dean0x7d:
Look good to me!

There's a [flake8 style error on Travis](https://travis-ci.org/pybind/pybind11/jobs/228147305#L301-L303). Just fix that and squash the PR to remove the "merge from master" commits and I think this is good to merge.

crisluengo:
@jagerman @dean0x7d I have rebased onto master and squashed all but 3 commits. I cannot guarantee that the first two commits are as they were originally, it is possible things got a little mixed up because of the move of `buffer_info` to its own file in between my first changes and now. But the head of the branch looks good to me.

dean0x7d:
Merged, thanks @crisluengo!

jagerman:
> Indeed, Eigen doesn't (but might someday)

It's a bit earlier to say for sure, but the [Eigen bug tracking this](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747) got marked today as a blocking-3.4 feature, so, assuming it doesn't run into major problems on the Eigen side, it looks like negative strides will work there soon too.

PR #784, opened 2017-04-06
Fix overriding static properties in derived classes

### LINKED ISSUE #775: Static properties override
Hello!

The question: do static properties override allowed by pybind11?

I have a base class, say `Foo` defined with its Python bindings in `module1`. Also I have a derived class `Bar: public Foo` defined with its Python bindings in `module2` (`pybind11::class_<Bar, Foo>`, i.e. base class `Foo` is passed as template argument to `pybind11::class_`). Both have static method `int test()` which I want to export to Python as static readonly property. I accomplish this with `.def_property_readonly_static` following the documentation.

But when `module2` is being loaded in Python interpreter I'm getting an exception: `AttributeError: can't set attribute`. I can fix this if I export `test()` as static function, i.e. with `.def_static` -- then everything seems to work properly. But overriding static properties doesn't work for me.

dean0x7d:
Fixes #775.

Assignments of the form `Type.static_prop = value` should be translated to `Type.static_prop.__set__(value)` except when `isinstance(value, static_prop)`.

### DIFF
--- a/include/pybind11/class_support.h
+++ b/include/pybind11/class_support.h
@@ -124,8 +124,15 @@ extern "C" inline int pybind11_meta_setattro(PyObject* obj, PyObject* name, PyOb
     // descriptor (`property`) instead of calling `tp_descr_get` (`property.__get__()`).
     PyObject *descr = _PyType_Lookup((PyTypeObject *) obj, name);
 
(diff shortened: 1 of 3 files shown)


jagerman:
Looks good to me.

wjakob:
Same here, please feel free to merge.

PR #786, opened 2017-04-07
Add static_assert to holder casters

### LINKED ISSUE #785: Keyword arguments mismatch with C++ function arguments
I have a C++ function with the following signature:
```C++
spv_float well_path_ident(
	ulong nx, ulong ny, spv_float coord, spv_float zcorn,
	spv_float well_info, bool include_well_nodes, const char* strat_traits = "online_tops",
	const ulong min_split_threshold = 0, spv_ulong hit_idx = nullptr
);
```
Here `ulong` is `std::size_t`, and `spv_float, spv_ulong` are `std::shared_ptr` pointers to my custom array type which is inherited from `pybind11::array_t`. I've written custom converters for it that are very similar to `Eigen` binding code. Converters are working as expected in my tests.

Next, I export the function above with the following snippet:
```C++
m.def("well_path_ident", &well_path_ident,
	"nx"_a, "ny"_a, "coord"_a, "zcorn"_a, "well_info"_a, "include_well_nodes"_a = true,
	"mesh_traits"_a = "online_tops", "min_split_threshold"_a = 0, "hit_idx"_a = nullptr
);
```
Everything compiles just fine! `pybind11` correctly counts arguments of `well_path_ident` -- if I make a mistake (for ex. forget one argument) it will notify me with an error at compile time.

But after module is loaded in Python here's what I get with `help(my_module)` command:
```
well_path_ident(...)
        well_path_ident(*args, **kwargs)
        Overloaded function.

        1. well_path_ident(nx: int, ny: int, numpy.ndarray[float64], numpy.ndarray[float64], numpy.ndarray[float64], coord: bool, zcorn: unicode, well_info: int, numpy.ndarray[uint64]) -> numpy.ndarray[float64]
```
(I have one more overload of this function, but it has the same issue).

As you can see, keyword arguments "skip" `numpy.ndarray[float64]` (spv_float) for some reason, and starting from "coord" they bind to wrong C++ function arguments.

I feel that this is related in some way to custom type conversion code, but what really can cause such issue?
Btw, when I try to call this function from Python, I get an exception:
```
RuntimeError: Unknown internal error occurred
```
And actual C++ function isn't called, I can't get into it in debugger.

jagerman:
The holder casters assume but don't check that a `holder<type>`'s `type` is really a `type_caster_base<type>`; this adds a `static_assert` to make sure this is really the case, to turn things like `std::shared_ptr<array>` into a compilation failure.

Fixes #785

There is some merit to making the holder casters use a proper `type_caster<type>` rather than a `type_caster_base<type>` so that you could, for example, have a `std::shared_ptr<std::string>` or `std::unique_ptr<Eigen::MatrixXd>`.  With this commit, holder arguments around non-generic type casters fails at compile time; before this commit, those failures show up in strange runtime failures (due to invoking the generic type caster rather than the specialized type caster).

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -902,6 +902,8 @@ template <typename type, typename holder_type>
 struct copyable_holder_caster : public type_caster_base<type> {
 public:
     using base = type_caster_base<type>;
+    static_assert(std::is_base_of<base, type_caster<type>>::value,
+            "Holder classes are only supported for custom types");
     using base::base;
     using base::cast;
     using base::typeinfo;
@@ -1018,6 +1020,9 @@ class type_caster<std::shared_ptr<T>> : public copyable_holder_caster<T, std::sh
 
 template <typename type, typename holder_type>
 struct move_only_holder_caster {
+    static_assert(std::is_base_of<type_caster_base<type>, type_caster<type>>::value,
+            "Holder classes are only supported for custom types");
+
     static handle cast(holder_type &&src, return_value_policy, handle) {
         auto *ptr = holder_helper<holder_type>::get(src);
         return type_caster_base<type>::cast_holder(ptr, &src);


jagerman:
Just to be clear: I did not change this to use proper type casters because it would require a fairly substantial rewrite of the holder casters, which are currently quite closely tied to `type_caster_base<type>` internals.

wjakob:
+1 That seems like a really nice way to catch this early and avoid confusion.

jagerman:
Actually I don't think this really solves #785; this just changes the error to a compile-time failure.  We're still unable to handle a holder around a type with a custom converter, which I think is going to happen in #785 when he switches to composition rather than inheritance.

jagerman:
I'm going to merge this to take care of the current internal error issue; I opened #787 for the second part of this.

PR #788, opened 2017-04-07
Generic containers/iterators for shape/strides

jagerman:
This adds support for constructing `buffer_info` and `array`s using arbitrary containers or iterators instead of requiring a vector.

On its own, this adds quite a few templated constructors which isn't all that desirable, but it's needed by PR #782 to preserve backwards compatibility for previous pybind11 versions that accept a `std::vector<size_t>` for strides.  Given a choice between duplicating all the stride-taking constructors (accepting both a `ssize_t` and `size_t` vector) and making the whole interface more general, I think this is a better approach.

I also seem to recall some discussion a few months ago about wanting something like this, but I can't find the issue/PR where it was mentioned.

### DIFF
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -7,7 +7,7 @@
     BSD-style license that can be found in the LICENSE file.
 */
 
-#pragma once 
+#pragma once
 
 #include "common.h"
 
@@ -26,25 +26,22 @@ struct buffer_info {
     buffer_info() { }
 
     buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t ndim,
-                const std::vector<size_t> &shape, const std::vector<size_t> &strides)
-        : ptr(ptr), itemsize(itemsize), size(1), format(format),
-          ndim(ndim), shape(shape), strides(strides) {
+                detail::any_container<size_t> shape_in, detail::any_container<size_t> strides_in)
+    : ptr(ptr), itemsize(itemsize), size(1), format(format), ndim(ndim),
+      shape(std::move(shape_in)), strides(std::move(strides_in)) {
+        if (ndim != shape.size() || ndim != strides.size())
+            pybind11_fail("buffer_info: ndim doesn't match shape and/or strides length");
         for (size_t i = 0; i < ndim; ++i)
             size *= shape[i];
     }
 
     buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t size)
-    : buffer_info(ptr, itemsize, format, 1, std::vector<size_t> { size },
-                  std::vector<size_t> { itemsize }) { }
+    : buffer_info(ptr, itemsize, format, 1, { size }, { itemsize }) { }
 
-    explicit buffer_info(Py_buffer *view, bool ownview = true)
-        : ptr(view->buf), itemsize((size_t) view->itemsize), size(1), format(view->format),
-          ndim((size_t) view->ndim), shape((size_t) view->ndim), strides((size_t) view->ndim), view(view), ownview(ownview) {
-        for (size_t i = 0; i < (size_t) view->ndim; ++i) {
-            shape[i] = (size_t) view->shape[i];
-            strides[i] = (size_t) view->strides[i];
-            size *= shape[i];
-        }
+    explicit buffer_info(Py_buffer *view, bool ownview_in = true)
+    : buffer_info(view->buf, (size_t) view->itemsize, view->format, (size_t) view->ndim,
+            {view->shape, view->shape + view->ndim}, {view->strides, view->strides + view->ndim}) {
+        ownview = ownview_in;
     }
 
     buffer_info(const buffer_info &) = delete;
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -1,0 +1,96 @@
+/*
+    pybind11/buffer_info.h: Python buffer object interface
+
+    Copyright (c) 2016 Wenzel Jakob <wenzel.jakob@epfl.ch>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once 
+
+#include "common.h"
+
+NAMESPACE_BEGIN(pybind11)
+
+/// Information record describing a Python buffer object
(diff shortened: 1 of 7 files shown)


jagerman:
@SylvainCorlay / @JohanMabille - I have a vague recollection that it was one (or both) of you who wanted something like this.  (Apologies if I'm wrong about that--I couldn't find the discussion).

dean0x7d:
I wonder if this could be simplified by shifting most of the logic from the constructors into a kind of vector adaptor class. E.g., instead of:

```c++
template <typename Shape, typename Strides,
          typename = decltype(std::begin(std::declval<const Shape &>())),
          typename = decltype(std::begin(std::declval<const Strides &>()))>
 array(const Shape &shape, const Strides &strides);
```

use:

```c++
array(const Adaptor& shape, const Adaptor& strides);
```
where:
```c++
class Adaptor {
public:
    template <typename Container, typename /*= ... SFINAE*/>
    Adaptor(const Container& c);

private:
    std::vector<T> data; // or whatever
};
```

So instead of templating every constructor, the adaptor takes care of everything.

jagerman:
That's a good idea.  The same thing could work for the iterator versions.

SylvainCorlay:
@jagerman @dean0x7d we now use the Numpy API directly and only inherit from py::object in our pybind11 bindings.

SylvainCorlay:
@jagerman @dean0x7d do you guys want to hop on gitter for a sec

jagerman:
Pushed update with @dean0x7d 's adaptor suggestion.

include/pybind11/common.h
```diff
@@ -657,4 +657,32 @@ static constexpr auto const_ = std::true_type{};
 
 #endif // overload_cast
 
+NAMESPACE_BEGIN(detail)
+// Aids in converting arbitrary containers into a vector; implicitly convertible from any standard
+// container (or C-style array) supporting std::begin/std::end.
+template <typename T>
+struct any_container {
+    std::vector<T> v; // Intended to be std::move'd
```
aldanor: Nitpicking: if you don't want to make `v` public, you could `operator std::vector<int>&&() && { return std::move(v); }` and then move the vector directly out of the container.

include/pybind11/numpy.h
```diff
@@ -478,7 +480,7 @@ class array : public buffer {
         }
 
         auto tmp = reinterpret_steal<object>(api.PyArray_NewFromDescr_(
-            api.PyArray_Type_, descr.release().ptr(), (int) ndim, shape.data(), strides.data(),
+            api.PyArray_Type_, descr.release().ptr(), (int) ndim, shape.v.data(), strides.v.data(),
```
aldanor: Maybe just provide a const ref `.data()` on the container?

include/pybind11/common.h
jagerman: Good point.  There is something a bit icky about moving a struct's member.

include/pybind11/numpy.h
jagerman: I added an lvalue reference conversion operator (in addition to the rvalue operator), so I can just grab a vector reference before that and call `.data()` on it.

jagerman: Actually, that won't work; clang considers it ambiguous.

aldanor: Weird, it's not obvious how it could be ambiguous.

So maybe just `.data()` then? Would we need anything else from it? (data ptr / size / move out into a vec)

aldanor: Btw you only use the lvalue (const) ref in passing it to `default_strides()` though? So maybe if you remove the mutable `&` overload and leave just `&&` and `const &` there would be no ambiguity?

include/pybind11/buffer_info.h
```diff
@@ -0,0 +1,100 @@
...
+            pybind11_fail("buffer_info: ndim doesn't match shape and/or strides length");
+        for (size_t i = 0; i < ndim; ++i)
+            size *= shape[i];
+    }
+
+
+    template <typename ShapeIt, typename StridesIt,
+             typename = detail::enable_if_t<detail::is_input_iterator<ShapeIt>::value && detail::is_input_iterator<StridesIt>::value>>
+    buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t ndim,
+            ShapeIt shape_first, ShapeIt shape_last, StridesIt strides_first, StridesIt strides_last)
```
dean0x7d: Isn't this already covered by the `any_container` constructor?

```C++
// ShapeIt, StridesIt
buffer_info(..., shape_first, shape_last, strides_first, strides_last);
// any_container
buffer_info(..., {shape_first, shape_last}, {strides_first, strides_last});
```

jagerman: Yeah, you're right.  I was intending to provide a more stl-like interface, but cutting down the number of constructors seems like a better idea.

jagerman:
Any other comments on this before I merge it?  (This is needed PR #782 and would be nice for #795 as well).

aldanor:
Looks good to me.

wjakob:
I just noticed a regression in some of my code regarding this patch: initializer lists seem to no longer work, which is a bummer. In particular,
```
size_t n = ..., m = ...;
py::array_t<float> array({n, m});
```
is now ambiguous. Same issue with this code:
```
size_t n = ..., m = ...;
py::array_t<float> array({(ssize_t) n, (ssize_t) m});
```
I had to change all occurrences of this pattern to
```
py::array_t<float> array(std::vector<ssize_t>{(ssize_t) n, (ssize_t) m});
```
for the compiler to accept it. I would be in favor of continuing to allow initializer lists for convenience -- what do you think?

jagerman:
Definitely a bug (and only occurs with exactly 2 arguments in the initializer list; with 1 or 3 it's fine).

It's caused by the arithmetic constructor in `any_container`: so {1, 2} could either be a implicit construction of an `array_t({shape={1}, strides={2} })`, or an implicit construction of an `any_container` with the initializer list.  You can fix it with `py::array_t<float> array({{n, m}})`, but that's not a nice solution.

I think taking out the arithmetic constructor makes the most sense here, so `any_container` needs a container (or initializer_list), but won't accept a single scalar value.  That was convenient, but not at the cost of introducing ambiguity.

PR #791, opened 2017-04-09
Fix Eigen argument doc strings

jagerman:
Many of the Eigen type casters' name() methods weren't wrapping the type description in a `type_descr` object, which thus wasn't adding the "{...}" annotation used to identify an argument which broke the help output by skipping Eigen arguments.

The test code I had added even had some (unnoticed) broken output (with the "arg0: " showing up in the return value).

This commit also adds test code to ensure that named eigen arguments actually work properly, despite the invalid help output.  (The added tests pass without the rest of this commit).


jagerman:
Whoops, that was only partially committed.  Fixed.

PR #793, opened 2017-04-10
array: always enable compile-time optimized 1D unchecked proxy for contiguous arrays

uentity:
Rationale: every C or Fortran contiguous array can be viewed as
1D-vector. Therefore if 1D unchecked proxy is requested for such
an array, skip dimensions checking and always return compile-time optimized
proxy object (strides and sizes stored in std::array).

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -620,11 +620,8 @@ public:
      * care: the array must not be destroyed or reshaped for the duration of the returned object,
      * and the caller must take care not to access invalid dimensions or dimension indices.
      */
-    template <typename T, ssize_t Dims = -1> detail::unchecked_mutable_reference<T, Dims> mutable_unchecked() {
-        if (Dims >= 0 && ndim() != (size_t) Dims)
-            throw std::domain_error("array has incorrect number of dimensions: " + std::to_string(ndim()) +
-                    "; expected " + std::to_string(Dims));
-        return detail::unchecked_mutable_reference<T, Dims>(mutable_data(), shape(), strides(), ndim());
+    template <typename T, ssize_t Dims = -1> auto mutable_unchecked() {
+        return mutable_unchecked_impl<T, Dims>(std::integral_constant<bool, Dims == 1>());
     }
 
     /** Returns a proxy object that provides const access to the array's data without bounds or
@@ -633,11 +630,8 @@ public:
      * reshaped for the duration of the returned object, and the caller must take care not to access
      * invalid dimensions or dimension indices.
      */
-    template <typename T, ssize_t Dims = -1> detail::unchecked_reference<T, Dims> unchecked() const {
-        if (Dims >= 0 && ndim() != (size_t) Dims)
-            throw std::domain_error("array has incorrect number of dimensions: " + std::to_string(ndim()) +
-                    "; expected " + std::to_string(Dims));
-        return detail::unchecked_reference<T, Dims>(data(), shape(), strides(), ndim());
+    template <typename T, ssize_t Dims = -1> auto unchecked() const {
+        return unchecked_impl<T, Dims>(std::integral_constant<bool, Dims == 1>());
     }
 
     /// Return a new view with all of the dimensions of length 1 removed
@@ -706,6 +700,41 @@ protected:
             return nullptr;
         return detail::npy_api::get().PyArray_FromAny_(
             ptr, nullptr, 0, 0, detail::npy_api::NPY_ARRAY_ENSUREARRAY_ | ExtraFlags, nullptr);
+    }
+
+    /// Unchecked proxy access implementation is selected by Dims == 1 condition
+    template <typename T, ssize_t Dims> detail::unchecked_reference<T, Dims>
+    unchecked_impl(std::false_type) const {
+        if (Dims >= 0 && ndim() != (size_t) Dims)
+            throw std::domain_error("array has incorrect number of dimensions: " + std::to_string(ndim()) +
+                    "; expected " + std::to_string(Dims));
+        return detail::unchecked_reference<T, Dims>(data(), shape(), strides(), ndim());
+    }
+    template <typename T, ssize_t Dims> detail::unchecked_reference<T, Dims>
+    unchecked_impl(std::true_type) const {
+        if (ndim() != 0 && (flags() & (c_style | f_style))) {
+            std::array<const size_t, 1> shape1d{{size()}};
+            std::array<const size_t, 1> strides1d{{itemsize()}};
+            return detail::unchecked_reference<T, 1>(data(), shape1d.data(), strides1d.data(), 1);
+        }
+        return unchecked_impl<T, Dims>(std::false_type());
+    }
+
+    template <typename T, ssize_t Dims> detail::unchecked_mutable_reference<T, Dims>
+    mutable_unchecked_impl(std::false_type) {
+        if (Dims >= 0 && ndim() != (size_t) Dims)
+            throw std::domain_error("array has incorrect number of dimensions: " + std::to_string(ndim()) +
+                    "; expected " + std::to_string(Dims));
+        return detail::unchecked_mutable_reference<T, Dims>(mutable_data(), shape(), strides(), ndim());
+    }
+    template <typename T, ssize_t Dims> detail::unchecked_mutable_reference<T, Dims>
+    mutable_unchecked_impl(std::true_type) {
+        if (ndim() != 0 && (flags() & (c_style | f_style))) {
+            std::array<const size_t, 1> shape1d{{size()}};
+            std::array<const size_t, 1> strides1d{{itemsize()}};
+            return detail::unchecked_mutable_reference<T, 1>(mutable_data(), shape1d.data(), strides1d.data(), 1);
+        }
+        return mutable_unchecked_impl<T, Dims>(std::false_type());
     }
 };
 
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -264,4 +264,13 @@ test_initializer numpy_array([](py::module &m) {
     sm.def("array_auxiliaries2", [](py::array_t<double> a) {
         return auxiliaries(a, a);
     });
+
+    sm.def("proxy_init3_ct", [](double start) {
+        py::array_t<double, py::array::c_style> a({ 3, 3, 3 });
+        auto r = a.mutable_unchecked<1>();
+        if (r.ndim() != 1) throw std::domain_error("error: ndim != 1, no compile-time opt");
+        for (size_t i = 0; i < r.shape(0); i++)
+            r(i) = start++;
+        return a;
+    });
 });
--- a/tests/test_numpy_array.py
+++ b/tests/test_numpy_array.py
@@ -367,13 +367,14 @@ def test_array_unchecked_fixed_dims(msg):
 
 def test_array_unchecked_dyn_dims(msg):
     from pybind11_tests.array import (proxy_add2_dyn, proxy_init3_dyn, proxy_auxiliaries2_dyn,
-                                      array_auxiliaries2)
+                                      array_auxiliaries2, proxy_init3_ct)
     z1 = np.array([[1, 2], [3, 4]], dtype='float64')
     proxy_add2_dyn(z1, 10)
     assert np.all(z1 == [[11, 12], [13, 14]])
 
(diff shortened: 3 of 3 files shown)


wjakob:
A nice part of the dimension check is that it prevents mistakes where a function expecting an 1D array is e.g. accidentally called with a matrix. If you want to treat everything as 1D arrays, you could just call ``mutable_data()`` and ``size()`` on the (checked) array and work with that.

jagerman:
Conceptually, a x-D and a y-D array are fundamentally different objects; that their storage is the same is more of an implementation detail.  There would also be some unexpected characteristics of this, for example `a.unchecked<1>()[2]` on a 2x2 matrix would give you either `a[1,0]` or `a[0,1]` depending on the storage order, and would have to fail for non-contiguous storage.

Something along similar lines that might be useful, however, we be to provide a wrapper for numpy's `reshape` method: if possible (i.e. if strides and order are compatible) it gives you a view into the original array, otherwise it makes a copy.

jagerman:
Another useful addition, perhaps more along the lines of what you're looking for here, would be to add an iterator to `unchecked_reference`, so that you could write:

```C++
for (auto &v : arr.unchecked<2>()) {
    // do something with v
}
```
where this iterator would iterator in row-major order, by default, but would have some mechanism to also allow column-major iteration.

uentity:
Thanks for your replies! Great idea with iterator on `unchecked_reference` -- I'll try to implement that. And `reshape` also would be quite useful.

My thinking behind this PR was the following.
1) If array is 1D then row/column major doesn't matter
2) if array is > 1D and somebody forced query of 1D `unchecked_reference` - then he probably knows what's happening and wants direct vector-like view of multidimensional array.

Always remembering dimensions seems inconvenient  to me, when just quick per-element processing is needed. And such kind of operations (in my practice) on numpy arrays in C++ are dominating for ex. linear algebra (which would be better done with `Eigen`) where row/column major order makes a lot of sense. To be correct, order always makes sense but in real life application you just *have* to postulate row/major common ordering of processed arrays, otherwise at least severe performance issues will arise.
And yes, we can address different elements of matrix depending of ordering, but this is an expected effect.

On the other side `pybind11` is generic library that should provide best possible implementation for every possible case. Here @wjakob note makes sense -- dimension-aware code can just directly use `data()` and `size()` for processing. So added complexity may be of no worth really.

jagerman:
Here's some (uncommitted) code I used to test unchecked performance when implementing unchecked references.  I think you could easily add an iteration version (for comparison of the iterator approach performance vs loops):

### unsafe.cpp:
```C++
#include <pybind11/numpy.h>

namespace py = pybind11;

double squared(double x) {
    return x*x;
}

PYBIND11_PLUGIN(unsafe) {
    py::module m("unsafe");

    m.def("incr_dbl_1D", [](py::array a) {
        auto r = a.mutable_unchecked<double, 1>();
        for (size_t i = 0; i < r.shape(0); i++)
            r[i]++;
    });
    m.def("incr_dbl_2D", [](py::array_t<double> a) {
        auto r = a.mutable_unchecked<2>();
        for (size_t i = 0; i < r.shape(0); i++)
        for (size_t j = 0; j < r.shape(1); j++)
            r(i, j)++;
    });
    m.def("incr_dbl_5D", [](py::array_t<double> a) {
        auto r = a.mutable_unchecked<5>();
        for (size_t i0 = 0; i0 < r.shape(0); i0++)
        for (size_t i1 = 0; i1 < r.shape(1); i1++)
        for (size_t i2 = 0; i2 < r.shape(2); i2++)
        for (size_t i3 = 0; i3 < r.shape(3); i3++)
        for (size_t i4 = 0; i4 < r.shape(4); i4++)
            r(i0, i1, i2, i3, i4)++;
    });

    m.def("sum_dbl_1D", [](const py::array_t<double> &a) {
        auto r = a.unchecked<1>();
        double sum = 0;
        for (size_t i = 0; i < r.shape(0); i++)
            sum += r[i];
        return sum;
    });
    m.def("sum_dbl_2D", [](const py::array &a) {
        auto r = a.unchecked<double, 2>();
        double sum = 0;
        for (size_t i = 0; i < r.shape(0); i++)
        for (size_t j = 0; j < r.shape(1); j++)
            sum += r(i, j);
        return sum;
    });
    m.def("sum_dbl_5D", [](const py::array &a) {
        auto r = a.unchecked<double, 5>();
        double sum = 0;
        for (size_t i0 = 0; i0 < r.shape(0); i0++)
        for (size_t i1 = 0; i1 < r.shape(1); i1++)
        for (size_t i2 = 0; i2 < r.shape(2); i2++)
        for (size_t i3 = 0; i3 < r.shape(3); i3++)
        for (size_t i4 = 0; i4 < r.shape(4); i4++)
            sum += r(i0, i1, i2, i3, i4);
        return sum;
    });


    // Dynamic-dimensions versions:
    m.def("dincr_dbl_1D", [](py::array a) {
        auto r = a.mutable_unchecked<double>();
        for (size_t i = 0; i < r.shape(0); i++)
            r[i]++;
    });
    m.def("dincr_dbl_2D", [](py::array_t<double> a) {
        auto r = a.mutable_unchecked();
        for (size_t i = 0; i < r.shape(0); i++)
        for (size_t j = 0; j < r.shape(1); j++)
            r(i, j)++;
    });
    m.def("dincr_dbl_5D", [](py::array_t<double> a) {
        auto r = a.mutable_unchecked();
        for (size_t i0 = 0; i0 < r.shape(0); i0++)
        for (size_t i1 = 0; i1 < r.shape(1); i1++)
        for (size_t i2 = 0; i2 < r.shape(2); i2++)
        for (size_t i3 = 0; i3 < r.shape(3); i3++)
        for (size_t i4 = 0; i4 < r.shape(4); i4++)
            r(i0, i1, i2, i3, i4)++;
    });

    m.def("dsum_dbl_1D", [](const py::array_t<double> &a) {
        auto r = a.unchecked();
        double sum = 0;
        for (size_t i = 0; i < r.shape(0); i++)
            sum += r[i];
        return sum;
    });
    m.def("dsum_dbl_2D", [](const py::array &a) {
        auto r = a.unchecked<double>();
        double sum = 0;
        for (size_t i = 0; i < r.shape(0); i++)
        for (size_t j = 0; j < r.shape(1); j++)
            sum += r(i, j);
        return sum;
    });
    m.def("dsum_dbl_5D", [](const py::array &a) {
        auto r = a.unchecked<double>();
        double sum = 0;
        for (size_t i0 = 0; i0 < r.shape(0); i0++)
        for (size_t i1 = 0; i1 < r.shape(1); i1++)
        for (size_t i2 = 0; i2 < r.shape(2); i2++)
        for (size_t i3 = 0; i3 < r.shape(3); i3++)
        for (size_t i4 = 0; i4 < r.shape(4); i4++)
            sum += r(i0, i1, i2, i3, i4);
        return sum;
    });

    // Non-proxy versions:
    m.def("aincr_dbl_1D", [](py::array_t<double> a) {
        for (size_t i = 0; i < a.shape(0); i++)
            a.mutable_at(i)++;
    });
    m.def("aincr_dbl_2D", [](py::array_t<double> a) {
        for (size_t i = 0; i < a.shape(0); i++)
        for (size_t j = 0; j < a.shape(1); j++)
            a.mutable_at(i, j)++;
    });
    m.def("aincr_dbl_5D", [](py::array_t<double> a) {
        for (size_t i0 = 0; i0 < a.shape(0); i0++)
        for (size_t i1 = 0; i1 < a.shape(1); i1++)
        for (size_t i2 = 0; i2 < a.shape(2); i2++)
        for (size_t i3 = 0; i3 < a.shape(3); i3++)
        for (size_t i4 = 0; i4 < a.shape(4); i4++)
            a.mutable_at(i0, i1, i2, i3, i4)++;
    });

    m.def("asum_dbl_1D", [](const py::array_t<double> &a) {
        double sum = 0;
        for (size_t i = 0; i < a.shape(0); i++)
            sum += a.at(i);
        return sum;
    });
    m.def("asum_dbl_2D", [](const py::array_t<double> &a) {
        double sum = 0;
        for (size_t i = 0; i < a.shape(0); i++)
        for (size_t j = 0; j < a.shape(1); j++)
            sum += a.at(i, j);
        return sum;
    });
    m.def("asum_dbl_5D", [](const py::array_t<double> &a) {
        double sum = 0;
        for (size_t i0 = 0; i0 < a.shape(0); i0++)
        for (size_t i1 = 0; i1 < a.shape(1); i1++)
        for (size_t i2 = 0; i2 < a.shape(2); i2++)
        for (size_t i3 = 0; i3 < a.shape(3); i3++)
        for (size_t i4 = 0; i4 < a.shape(4); i4++)
            sum += a.at(i0, i1, i2, i3, i4);
        return sum;
    });

    return m.ptr();
}
```

### test_unsafe.py
```Python
from __future__ import print_function
from timeit import timeit
print("fixed, 1D, 10M,  incr", timeit("incr_dbl_1D(m)", "from unsafe import incr_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("fixed, 1D, 10M,  sum ", timeit("sum_dbl_1D(m)", "from unsafe import sum_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("fixed, 10K x 1K, incr", timeit("incr_dbl_2D(m)", "from unsafe import incr_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("fixed, 10K x 1K, sum ", timeit("sum_dbl_2D(m)", "from unsafe import sum_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("fixed, 100x100x10x10x10, incr", timeit("incr_dbl_5D(m)", "from unsafe import incr_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
print("fixed, 100x100x10x10x10, sum ", timeit("sum_dbl_5D(m)", "from unsafe import sum_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
print("dynamic, 1D, 10M,  incr", timeit("dincr_dbl_1D(m)", "from unsafe import dincr_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("dynamic, 1D, 10M,  sum ", timeit("dsum_dbl_1D(m)", "from unsafe import dsum_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("dynamic, 10K x 1K, incr", timeit("dincr_dbl_2D(m)", "from unsafe import dincr_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("dynamic, 10K x 1K, sum ", timeit("dsum_dbl_2D(m)", "from unsafe import dsum_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("dynamic, 100x100x10x10x10, incr", timeit("dincr_dbl_5D(m)", "from unsafe import dincr_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
print("dynamic, 100x100x10x10x10, sum ", timeit("dsum_dbl_5D(m)", "from unsafe import dsum_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
print("checked, 1D, 10M,  incr", timeit("aincr_dbl_1D(m)", "from unsafe import aincr_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("checked, 1D, 10M,  sum ", timeit("asum_dbl_1D(m)", "from unsafe import asum_dbl_1D; import numpy as np; m = np.ones((10000000,))", number=500))
print("checked, 10K x 1K, incr", timeit("aincr_dbl_2D(m)", "from unsafe import aincr_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("checked, 10K x 1K, sum ", timeit("asum_dbl_2D(m)", "from unsafe import asum_dbl_2D; import numpy as np; m = np.ones((10000,1000))", number=500))
print("checked, 100x100x10x10x10, incr", timeit("aincr_dbl_5D(m)", "from unsafe import aincr_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
print("checked, 100x100x10x10x10, sum ", timeit("asum_dbl_5D(m)", "from unsafe import asum_dbl_5D; import numpy as np; m = np.ones((100,100,10,10,10))", number=500))
```

dean0x7d:
Looks like the general recommendation here was a change of direction to iterators. Since there hasn't been any activity here for a long time, I'll close this PR for now, but feel free to reopen if there is any news.

PR #794, opened 2017-04-10
array: set exception message on failure

### LINKED ISSUE #785: Keyword arguments mismatch with C++ function arguments

jagerman:
When attempting to get a raw array pointer we return nullptr if given a nullptr, which triggers an `error_already_set()`, but we haven't set an exception message, which results in `Unknown internal error`.

Callers that want explicit allowing of a `nullptr` here already handle it (by clearing the exception after the call).

Fixes #785.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -702,8 +702,10 @@ protected:
 
     /// Create array from any object -- always returns a new reference
     static PyObject *raw_array(PyObject *ptr, int ExtraFlags = 0) {
-        if (ptr == nullptr)
+        if (ptr == nullptr) {
+            PyErr_SetString(PyExc_ValueError, "cannot create a pybind11::array from a nullptr");
             return nullptr;
+        }
         return detail::npy_api::get().PyArray_FromAny_(
             ptr, nullptr, 0, 0, detail::npy_api::NPY_ARRAY_ENSUREARRAY_ | ExtraFlags, nullptr);
     }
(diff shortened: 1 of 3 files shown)


jagerman:
No one has commented, and this seems fairly harmless (and fixes a bug, if a somewhat obscure one), so I'll plan to merge this later today.

aldanor:
This was definitely overlooked, looks good to me

Issue #792, opened 2017-04-09
Random appveyor x86 build failures


jagerman:
I'm seeing occassional random build failures on the appveyor x86 builds, always during the linking stage.  I have a feeling we're running into memory limitations, perhaps combined with the addition of the /m flag for parallel building and the new multi-CPU abilities.  (I haven't yet seen a failure for exactly the same builds on https://ci.appveyor.com/project/jagerman/pybind11).  If this persists, perhaps we should turn off the /m flag? (I'm assuming--though may be wrong--that it's default to 2 cores for the main project builds and 1 for my open-source-free account)

jagerman:
Another option: we could disable eigen tests on the 2 appveyor x86 builds; in practice it seems to be the most resource heavy test script.

jagerman:
Example: [pre-merge build](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1645) and [post-merge build](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1646), with no changes to master between the PR and the merge.

wjakob:
Removing the ``/m`` flag sounds good to me, that doesn't really seem to have helped in any way.

jagerman:
I removed it; we'll see if it helps.

jagerman:
Another failure without /m (this time on `jagerman` not `wjakob`): https://ci.appveyor.com/project/jagerman/pybind11/build/1.0.529

And this strange one: https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1648/job/jrflasp8e5tsr95l

I have a feeling that there are some appveyor x86 issues.

jagerman:
And [another](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1653/job/yfe7e4a2dcds4lej), which [worked here](https://ci.appveyor.com/project/jagerman/pybind11/build/1.0.530).

jagerman:
Now *every* PR is triggering it, even on x64.  Um...

wjakob:
Huh? I don't see that here.

jagerman:
Okay, not every; but it did show up on [the first](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1657) and [second](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.1657) x64 builds, so I guess this isn't purely x86-specific.

[...]

PR #795, opened 2017-04-11
array: implement array resize

uentity:
Resize won't work in every conditions, but at least it should work
1) if new array was created from scratch in C++ and resized before being exported to Python
2) if array passed from Python, owns it's data and doesn't have dependant arrays and refcheck is 0.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -129,6 +129,11 @@ struct npy_api {
         NPY_STRING_, NPY_UNICODE_, NPY_VOID_
     };
 
+    typedef struct {
+        Py_intptr_t *ptr;
+        int len;
+    } PyArray_Dims;
+
     static npy_api& get() {
         static npy_api api = lookup();
         return api;
@@ -159,6 +164,7 @@ struct npy_api {
                                              Py_ssize_t *, PyObject **, PyObject *);
     PyObject *(*PyArray_Squeeze_)(PyObject *);
     int (*PyArray_SetBaseObject_)(PyObject *, PyObject *);
+    PyObject* (*PyArray_Resize_)(PyObject*, PyArray_Dims*, int, int);
 private:
     enum functions {
         API_PyArray_GetNDArrayCFeatureVersion = 211,
@@ -168,6 +174,7 @@ private:
         API_PyArray_DescrFromType = 45,
         API_PyArray_DescrFromScalar = 57,
         API_PyArray_FromAny = 69,
+        API_PyArray_Resize = 80,
         API_PyArray_NewCopy = 85,
         API_PyArray_NewFromDescr = 94,
         API_PyArray_DescrNewFromType = 9,
@@ -197,6 +204,7 @@ private:
         DECL_NPY_API(PyArray_DescrFromType);
         DECL_NPY_API(PyArray_DescrFromScalar);
         DECL_NPY_API(PyArray_FromAny);
+        DECL_NPY_API(PyArray_Resize);
         DECL_NPY_API(PyArray_NewCopy);
         DECL_NPY_API(PyArray_NewFromDescr);
         DECL_NPY_API(PyArray_DescrNewFromType);
@@ -650,6 +658,21 @@ public:
     array squeeze() {
         auto& api = detail::npy_api::get();
         return reinterpret_steal<array>(api.PyArray_Squeeze_(m_ptr));
+    }
+
+    /// Resize array to given shape
+    /// If refcheck is true and more that one reference exist to this array
+    /// then resize will succeed only if it makes a reshape, i.e. original size doesn't change
+    void resize(ShapeContainer new_shape, bool refcheck = true) {
+        detail::npy_api::PyArray_Dims d = {
+            new_shape->data(), int(new_shape->size())
+        };
+        // try to resize, set ordering param to -1 cause it's not used anyway
+        object new_array = reinterpret_steal<object>(
+            detail::npy_api::get().PyArray_Resize_(m_ptr, &d, int(refcheck), -1)
+        );
+        if (!new_array) throw error_already_set();
+        if (isinstance<array>(new_array)) { *this = std::move(new_array); }
     }
 
     /// Ensure that the argument is a NumPy array
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -273,4 +273,25 @@ test_initializer numpy_array([](py::module &m) {
     sm.def("array_initializer_list", []() { return py::array_t<float>({ 1, 2 }); });
     sm.def("array_initializer_list", []() { return py::array_t<float>({ 1, 2, 3 }); });
     sm.def("array_initializer_list", []() { return py::array_t<float>({ 1, 2, 3, 4 }); });
+
+    // reshape array to 2D without changing size
+    sm.def("array_reshape2", [](py::array_t<double> a) {
+        const size_t dim_sz = (size_t)std::sqrt(a.size());
+        if (dim_sz * dim_sz != a.size())
+            throw std::domain_error("array_reshape2: input array total size is not a squared integer");
+        a.resize({dim_sz, dim_sz});
+    });
+
+    // resize to 3D array with each dimension = N
+    sm.def("array_resize3", [](py::array_t<double> a, size_t N, bool refcheck) {
+        a.resize({N, N, N}, refcheck);
+    });
+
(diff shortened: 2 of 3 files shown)


tests/test_numpy_array.py
```diff
@@ -377,3 +377,18 @@ def test_array_unchecked_dyn_dims(msg):
 
     assert proxy_auxiliaries2_dyn(z1) == [11, 11, True, 2, 8, 2, 2, 4, 32]
     assert proxy_auxiliaries2_dyn(z1) == array_auxiliaries2(z1)
+
+def test_array_resize(msg):
```
jagerman: You should also test some failure conditions here, i.e. that you get the expected exception and exception message if attempting to resize a referenced array.

tests/test_numpy_array.cpp
```diff
@@ -264,4 +264,24 @@ test_initializer numpy_array([](py::module &m) {
     sm.def("array_auxiliaries2", [](py::array_t<double> a) {
         return auxiliaries(a, a);
     });
+
+    // resize array to 3D without changing size
```
jagerman: `3D` -> `2D`

tests/test_numpy_array.cpp
```diff
@@ -264,4 +264,24 @@ test_initializer numpy_array([](py::module &m) {
...
+    // should probably succeed
+    sm.def("array_reshape2", [](py::array_t<double> a) {
+        const size_t dim_sz(std::sqrt(a.size()));
```
jagerman: A self-commenting input validation test:
```C++
if (dim_sz * dim_sz != a.size())
    throw std::domain_error("array_reshape2: input array total size is not a squared integer");
```

jagerman:
A few things to fix, but nothing major: this looks pretty good to me.

include/pybind11/numpy.h
```diff
@@ -646,6 +654,26 @@ class array : public buffer {
         return reinterpret_steal<array>(api.PyArray_Squeeze_(m_ptr));
     }
 
+    /// Resize array to given shape
+    /// If refcheck is true, then resize will fail if > 1 references exists to this array
+    void resize(const std::vector<size_t> &new_shape, bool refcheck = false) {
```
dean0x7d: +1 for `refcheck=true` by default.

dean0x7d:
There is also the possibility of simplifying the implementation so it doesn't need the Numpy C API:

```c++
void resize(const std::vector<size_t> &new_shape, bool refcheck = true) {
    attr("resize")(new_shape, refcheck);
}
```

Which would be a bit slower because of the dynamic method lookup, but at the same time I don't think `resize` is going into any hot loops anyway. (I don't mind the C API either, I'm just mentioning this as a possibility.)

include/pybind11/numpy.h
jagerman: Also note that in the [Python API](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.resize.html), refcheck defaults to True.

include/pybind11/numpy.h
```diff
@@ -655,23 +655,18 @@ class array : public buffer {
     }
 
     /// Resize array to given shape
-    /// If refcheck is true, then resize will fail if > 1 references exists to this array
-    void resize(const std::vector<size_t> &new_shape, bool refcheck = false) {
+    /// If refcheck is true and more that one reference exist to this array
+    /// then resize will succeed only if it makes a reshape, i.e. original size doesn't change
+    void resize(const std::vector<size_t> &new_shape, bool refcheck = true) {
```
jagerman: I'm not a huge fan of taking concrete containers (`std::vector<size_t>`) like this--it's restrictive and something of a nuisance if you happen to have your shape in some other container.  Just to be clear: this isn't your fault at all; what you wrote here is perfectly consistent with the current API.  But I'm aiming to solve it with PR #788, which would allow you to change the signature to `void resize(ShapeContainer new_shape, bool refcheck = true)`, where the `ShapeContainer` is implicitly convertible from any sort of container, copy the contents into a `std::vector<Py_intptr_t>`.  That would also allow you to drop both the `reinterpret_cast` and `const_cast` from:
`reinterpret_cast<Py_intptr_t*>(const_cast<size_t*>(new_shape.data())), int(new_shape.size())`
replacing it with just:
`new_shape->data(), int(new_shape.size())`

jagerman: I've merged #788, so you should be able to switch the input to a `ShapeContainer new_shape` now.  (Unfortunately I also caused some conflicts--both had tests added in the same place--but I added a commit to hopefully resolve them).

aldanor:
(Minor: it's best to not have merge commits from master in feature branches, ideally this PR would get rebased or squashed)

jagerman:
Yeah, it'll get squashed and rebased before being merged.

include/pybind11/numpy.h
```diff
@@ -129,6 +129,11 @@ struct npy_api {
         NPY_STRING_, NPY_UNICODE_, NPY_VOID_
     };
 
+    typedef struct {
+        const Py_intptr_t *ptr;
```
uentity: I think it's safe to make ptr `const` here instead of having `const_cast<Py_intptr_t*>(new_shape->data()` later in resize()

tests/test_numpy_array.cpp
```diff
@@ -264,7 +264,28 @@ test_initializer numpy_array([](py::module &m) {
         return auxiliaries(a, a);
     });
 
-    // Issue #785: Uninformative "Unknown internal error" exception when constructing array from empty object:
+      // Issue #785: Uninformative "Unknown internal error" exception when constructing array from empty object:
```
uentity: oops

include/pybind11/numpy.h
jagerman: Just change the `const ShapeContainer &new_shape` to `ShapeContainer new_shape`; it's essentially always going to be passed by value (via implicit conversion), so the const reference doesn't gain anything.

uentity: Done.

jagerman:
You can ignore the MSVC failure; that's something broken in appveyor (#792).

uentity:
Travis also didn't complete for some reason...

jagerman:
I just restarted the failed job.

jagerman:
Can you rebase this against current master?  I have a feeling the travis-ci commit I merged today is going to cause some failures in the PyPy test here from some things numpy doesn't support when under PyPy (and so need to be skipped for PyPy).

uentity:
Rebased.
I failed to build tests with latest PyPy (5.7.1) as well as with bundled with Fedora 25 (5.4.0), so didn't check if tests succeeds on it.

jagerman:
Don't worry about 5.4.0; we don't support anything before 5.7 as it had several fixes we reported to get it working with pybind.  The [one test failure](https://travis-ci.org/pybind/pybind11/jobs/223699895#L493) was exactly what I expected to fail; just skip that test under PyPy: move the failing assertions (basically anything with `refcheck=True`) into a separate test and mark that test with `@pytest.unsupported_on_pypy`

jagerman:
> I failed to build tests

Wait, failed to *build*?  That part should work (with 5.7 at least).

uentity:
Year, I see what you have done with disabling tests on PyPy in prev commits. I just wanted to check it myself before pushing.

But as I can see now, you build tests on *PyPy 5.7 with Python 2.7* and I tried to build against *PyPy 5.7 with Python 3.5*. That combination gives compile error:
```
class_support.h:169:13: error: «struct _heaptypeobject» has no member named «ht_qualname»; did you mean «ht_name»?
  heap_type->ht_qualname = name_obj.inc_ref().ptr();
             ^~~~~~~~~~~
```

wjakob:
Aha, looks like some Python 3.x compatibility issues in the Python 3 version of PyPy. There is no need to worry about it here.

jagerman:
Oh, right, I forgot about the PyPy3 version.  We currently only support PyPy2 5.7.

uentity:
Yes, I noticed that support is only for PyPy2 5.7.
Are there any stoppers preventing this from being merged upstream?

dean0x7d:
This looks good to me. There's a merge conflict right now. Just resolve that and I think it's good to merge.

jagerman:
Can you do a rebase against master rather than a merge?  It makes it easier to commit without getting merge commits in the log.

uentity:
Sorry, I happened to use this slick github "resolve merge conflicts" button.

jagerman:
Merged!  Thanks!

PR #798, opened 2017-04-14
Travis: switch to trusty, pip and libc++ caching, pypy numpy

jagerman:
This applies several changes to the non-docker (and the C++17 clang docker) travis-ci builds:

- Make all non-docker linux builds use *trusty* rather than *precise*.  pybind can't actually build in *precise* anyway (we install essentially the entire toolchain backported from *trusty* on every build), and so this saves needing to install all the backported packages during the build setup.
- Updated the 3.5 build to 3.6 (via deadsnakes, which didn't backport 3.6 to ubuntu releases earlier than trusty).
- As a result of the switch to trusty, the BAREBONES build now picks up the (default-installed) python 3.5 installation.
- `pip` is now invoked everywhere via `$PY_CMD -m pip` (where `PY_CMD` is the python binary name) rather than the `pip` executable, which saves us having to figure out what the pip executable is (e.g. `pip` vs `pip3`), and ensures that we are always using the correct `pip`.
- Add the local user python package archive to the travis-ci cache (rather than the pip cache).  This saves needing to install packages during installation (unless there are updates, in which case the package and the cache are updated), and saves a huge amount of time for the pypy and clang 4.0 builds (see below).
- Install packages with `pip --user` rather than in a virtualenv.  This makes the caching a bit easier: we can just cache `~/.local/{lib,bin}`.  (We could alternatively cache a `virtualenv` directory, but caching things under .local/ make it easier to also cache the local libc++ install, below).
- Install numpy and scipy on the pypy build.  This has to build from source (which adds about 8 minutes to the job time), but due to the above caching, this only happens *once* per numpy/scipy release.  I think this testing is quite valuable: numpy has various behaviour differences under pypy (I'm anticipating that this is going to break #795, which will need to skip some tests on pypy).
- Enable numpy tests for pypy exposed a segfault in the stl_binder test: it was previously skipped because it was marked as requiring numpy, but I'm fairly sure this is the same issue as on the other test here, which didn't require numpy and had an explicit pypy skip.
- Added `set -e`/`+e` around the before_install/install blocks so that a failure here (e.g. a pip install failure or dependency download failure) triggers a build failure.  Previously such failures during dependency installation were just silently ignored.
- Update eigen version to latest (3.3.3), mainly to be consistent with the appveyor build.
- Switch the c++17 clang build from docker to trusty (for which clang builds are available from LLVM); the container setup is faster than docker to begin with, but it also makes caching the local libc++ installation possible, which saves a good chunk of time on every build.
- Various small cleanups; the most notable is getting rid of the all `export`s on script variables (e.g. `PY`, `GCC`) which are just local script variables that don't need to be exported.  The remaining `export`s are only for things like `CXX` which we *do* need to be exported to cmake.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,40 +94,40 @@ before_install:
       export CXX=clang++-$CLANG CC=clang-$CLANG
       COMPILER_PACKAGES="clang-$CLANG llvm-$CLANG-dev"
     else
-      if [ -z "$GCC" ]; then export GCC=4.8
-      else export COMPILER_PACKAGES=g++-$GCC
+      if [ -z "$GCC" ]; then GCC=4.8
+      else COMPILER_PACKAGES=g++-$GCC
       fi
       export CXX=g++-$GCC CC=gcc-$GCC
     fi
-    if [ "$GCC" = "6" ] || [ "$CLANG" = "3.9" ]; then export DOCKER=${ARCH:+$ARCH/}debian:testing
-    elif [ "$GCC" = "7" ]; then export DOCKER=debian:experimental APT_GET_EXTRA="-t experimental"
+    if [ "$GCC" = "6" ] || [ "$CLANG" = "3.9" ]; then DOCKER=${ARCH:+$ARCH/}debian:testing
+    elif [ "$GCC" = "7" ]; then DOCKER=debian:experimental APT_GET_EXTRA="-t experimental"
     fi
   elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
     export CXX=clang++ CC=clang;
   fi
-  if [ -n "$CPP" ]; then export CPP=-std=c++$CPP; fi
-  if [ "${PYTHON:0:1}" = "3" ]; then export PY=3; fi
-  if [ -n "$DEBUG" ]; then export CMAKE_EXTRA_ARGS="-DCMAKE_BUILD_TYPE=Debug"; fi
+  if [ -n "$CPP" ]; then CPP=-std=c++$CPP; fi
+  if [ "${PYTHON:0:1}" = "3" ]; then PY=3; fi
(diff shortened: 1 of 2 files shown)


jagerman:
And another spurious appveyor failure. :(

wjakob:
This looks great! (+1 for testing NumPy with PyPy). I restarted the AppVeyor build.

dean0x7d:
Awesome overhaul!

Just one small issue: pybind11's minimum supported CMake version 2.8.12 is not longer being tested anywhere. I guess this can be easily fixed by reverting `BAREBONES` to `precise`.

.travis.yml
```diff
@@ -58,126 +57,141 @@ matrix:
...
-  elif [ -z "$PYPY" ]; then
-    pip install numpy scipy pytest
 
-    wget -q -O eigen.tar.gz https://bitbucket.org/eigen/eigen/get/3.3.0.tar.gz
+    export NPY_NUM_BUILD_JOBS=2
+    echo "Installing numpy, pytest..."
+    $PY_CMD -m pip install --user --upgrade --quiet numpy pytest
+    echo -e "done.\nInstalling scipy..."
+    $PY_CMD -m pip install --user --upgrade --quiet scipy
+    echo "done."
```
dean0x7d: Why are pytest/numpy/scipy installed in two stages?

jagerman: I had a failure at one point when installing both at the same time, which looked like pip had tried to build scipy without numpy being installed yet.  It didn't actually complete fail: it seemed as though pip *then* installed numpy, and ran the scipy build again in the same command.

I can't reproduce this locally now, however, so I'll go ahead and recombine them.

jagerman:
Good point about cmake: I hadn't intended that.  Trusty itself has cmake 2.8.12, but the travis trusty environment has a ppa-upgraded version.  I've updated the first two builds to downgrade to the system version during initialization.

.travis.yml
jagerman: ... it also apparently prevented a travis no-output-for-10-minute failure, but I've added a travis-specific fix for that.

PR #802, opened 2017-04-17
Don't let `PyInstanceMethod` hide itself

### LINKED ISSUE #443: Bizarre Python 3.x failure when creating method aliases
Consider the following program, which creates an alias of a method. My use case for this kind of design pattern were two different precision implementations of some floating point code (i.e. `computeSingle` and `computeDouble`) as well as an alias for the default compile time precision of the application (i.e. `computeFloat`).

``` cpp

struct Test { void foo() { printf("foo\n"); } };

namespace py = pybind11;

PYBIND11_PLUGIN(test) {
    py::module m("test");
    py::class_<Test> tclass(m, "Test");

    tclass
        .def(py::init<>())
        .def("foo", &Test::foo);

    tclass.attr("bar") = tclass.attr("foo");

    return m.ptr();
}
```

Calling this from Python 3.x produces the following issue (Python 2.7 works without problems):

``` pycon
Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 23 2015, 02:52:03)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import test
>>> t = test.Test()
>>> t.foo()
foo
>>> t.bar()
TypeError: foo(): incompatible function arguments. The following argument types are supported:
    1. (self: test.Test) -> None

Invoked with:
```

What specifically happens here is that the implicit `self` argument is dropped, which causes method resolution to fail. The root cause is `PyInstanceMethod_New`, which seems to be really picky about the name of the attribute that does the function call.

So when creating an alias, this means that we would have to un-box the method from it's method wrapper, and re-box it with a different name. It wasn't clear to me how to do that in a transparent way, and whether that's a generally reasonable workaround. What's also weird is that this kind of behavior does not occur in native Python classes, where method aliases can be created without such problems.

jagerman:
Python 3's `PyInstanceMethod_Type` hides itself via its `tp_descr_get`, which prevents aliasing methods via `cls.attr("m2") = cls.attr("m1")`: instead the `tp_descr_get` returns a plain function, when called on a
class, or a `PyMethod`, when called on an instance.  Override that behaviour for pybind11 types with a special bypass for `PyInstanceMethod_Types`.

Fixes #443.

### DIFF
--- a/include/pybind11/class_support.h
+++ b/include/pybind11/class_support.h
@@ -149,6 +149,24 @@ extern "C" inline int pybind11_meta_setattro(PyObject* obj, PyObject* name, PyOb
     }
 }
 
+#if PY_MAJOR_VERSION >= 3
+/** Python 3's PyInstanceMethod_Type hides itself via its tp_descr_get, which prevents aliasing
+ * methods via cls.attr("m2") = cls.attr("m1"): instead the tp_descr_get returns a plain function,
+ * when called on a class, or a PyMethod, when called on an instance.  Override that behaviour here
+ * to do a special case bypass for PyInstanceMethod_Types.
+ */
+extern "C" inline PyObject *pybind11_meta_getattro(PyObject *obj, PyObject *name) {
+    PyObject *descr = _PyType_Lookup((PyTypeObject *) obj, name);
+    if (descr && PyInstanceMethod_Check(descr)) {
+        Py_INCREF(descr);
+        return descr;
+    }
+    else {
+        return PyType_Type.tp_getattro(obj, name);
+    }
+}
+#endif
+
 /** This metaclass is assigned by default to all pybind11 types and is required in order
     for static properties to function correctly. Users may override this using `py::metaclass`.
     Return value: New reference. */
@@ -176,6 +194,9 @@ inline PyTypeObject* make_default_metaclass() {
 
     type->tp_new = pybind11_meta_new;
     type->tp_setattro = pybind11_meta_setattro;
+#if PY_MAJOR_VERSION >= 3
+    type->tp_getattro = pybind11_meta_getattro;
+#endif
 
     if (PyType_Ready(type) < 0)
         pybind11_fail("make_default_metaclass(): failure in PyType_Ready()!");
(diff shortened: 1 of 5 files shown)


wjakob:
That's really cool -- I'm glad that #443 can be fixed at all, and that the new metaclass approach allows doing this so cleanly. LGTM to commit.

dean0x7d:
Great fix. Looks good to me too.

Minor style comment: It might be good to agree on a docstring style for consistency. In C++:
```c++
/** documentation
 * string
 */
```
vs.
```c++
/** documentation
    string */
```

In Python:
```python
# Documentation string
def foo():
    pass
```
vs.
```python
def foo():
    """Documentation string"""
```

In Python, I'd strongly recommend using the standard triple-quote docstrings. In C++, I don't think it matters as long as it's consistent (although the asterisk-less version has better sphinx compatibility).

jagerman:
I'm fine with either style for C++; I'm using the first version just because that's vim's default for comment rewrapping, but I can change that easily enough.

jagerman:
@wjakob - any preference for the C++ documenting comment format?

wjakob:
I prefer
```cpp
/** documentation
 * string
 */
```
but it's really a minor preference. Is Sphinx not able to strip out the asterisks?

wjakob:
I think a lot of the comment in question are really internal notes to developers. For those, the format matters less and could even be in the body of the function. For comments that document public API, the format should be more rigid.

dean0x7d:
Breathe can handle the leading asterisks with additional markup: http://breathe.readthedocs.io/en/latest/markups.html Of course, for internal docs, sphinx doesn't matter. I'm mainly asking for the purpose of maintaining a consistent internal style. In the introduction of sphinx/breathe, I assumed asterisks-less was the preferred pybind11 style based on the existing docstrings (e.g. [pre-sphinx/breathe docs here](https://github.com/pybind/pybind11/blob/master/include/pybind11/common.h#L238)) and so the breathe defaults were set according to that. Like I said, I don't think the style itself matters, just as long as it's consistent.

jagerman:
So what's the consensus here?  Should I update the comment style, or leave as-is?

wjakob:
Up to you -- there are probably other inconsistent comments. I propose that if comment style bothers anyone, we fix it globally via an independent PR.

jagerman:
Sounds good to me.  Just for the record, I've no opposition to a global comment style policy, but agree with @wjakob that it should belong as an independent PR/discussion.

PR #804, opened 2017-04-17
Don't allow mixed static/non-static overloads

jagerman:
We currently fail at runtime when trying to call a method that is overloaded with both static and non-static methods.  This is something python won't allow: the object is either a function or an instance, and can't be both.


wjakob:
LGTM!

dean0x7d:
👍

PR #805, opened 2017-04-17
Allow binding factory functions as constructors

jagerman:
Edit:

Redid this (a few times).  The interface is now:

```C++
    cls.def(py::init(&factory_function));
```

where `factory_function` is some pointer- or holder-generating factory function or lambda of the type that `cls` binds, such as:

```C++
    cls.def(py::init([](int a) { return new MyClass(a); }));
```

Internally, this still results in a (non-static) `__init__` method, but we handle the C++-static-factory <-> python method translation by calling the factory function, then stealing the resulting object internal pointer and holder.

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -322,6 +322,8 @@ can now create a python class that inherits from ``Dog``:
     See the file :file:`tests/test_virtual_functions.cpp` for complete examples
     using both the duplication and templated trampoline approaches.
 
+.. _extended_aliases:
+
 Extended trampoline class functionality
 =======================================
 
@@ -358,16 +360,124 @@ Custom constructors
 ===================
 
 The syntax for binding constructors was previously introduced, but it only
-works when a constructor with the given parameters actually exists on the C++
-side. To extend this to more general cases, let's take a look at what actually
-happens under the hood: the following statement
+works when a constructor of the appropriate arguments actually exists on the
+C++ side.  To extend this to more general cases, pybind11 offers two different
+approaches: binding factory functions, and placement-new creation.
+
+Factory function constructors
+-----------------------------
+
+It is possible to expose a Python-side constructor from a C++ function that
+returns a new object by value or pointer.  For example, suppose you have a
+class like this:
+
+.. code-block:: cpp
+
+    class Example {
+    private:
+        Example(int); // private constructor
+    public:
+        // Factory function:
+        static Example create(int a) { return Example(a); }
+    };
+
+While it is possible to expose the ``create`` method to Python, it is often
+preferrable to expose it on the Python side as a constructor rather than a
+named static method.  You can do this by calling ``.def(py::init(...))`` with
+the function reference returning the new instance passed as an argument.  It is
+also possible to use this approach to bind a function returning a new instance
+by raw pointer or by the holder (e.g. ``std::unique_ptr``).
+
+The following example shows the different approaches:
+
+.. code-block:: cpp
+
+    class Example {
+    private:
+        Example(int); // private constructor
+    public:
+        // Factory function - returned by value:
+        static Example create(int a) { return Example(a); }
+
+        // These constructors are publicly callable:
+        Example(double);
+        Example(int, int);
+        Example(std::string);
+    };
+
+    py::class_<Example>(m, "Example")
+        // Bind the factory function as a constructor:
+        .def(py::init(&Example::create))
+        // Bind a lambda function returning a pointer wrapped in a holder:
+        .def(py::init([](std::string arg) {
+            return std::unique_ptr<Example>(new Example(arg));
+        }))
+        // Return a raw pointer:
+        .def(py::init([](int a, int b) { return new Example(a, b); }))
+        // You can mix the above with regular C++ constructor bindings as well:
+        .def(py::init<double>())
+        ;
+
+When the constructor is invoked from Python, pybind11 will call the factory
+function and store the resulting C++ instance in the Python instance.
+
+When combining factory functions constructors with :ref:`overriding_virtuals`
+there are two approaches.  The first is to add a constructor to the alias class
+that takes a base value by rvalue-reference.  If such a constructor is
+available, it will be used to construct an alias instance from the value
+returned by the factory function.  The second option is to provide two factory
+functions to ``py::init()``: the first will be invoked when no alias class is
+required (i.e. when the class is being used but not inherited from in Python),
+and the second will be invoked when an alias is required.
+
+You can also specify a single factory function that always returns an alias
+instance: this will result in behaviour similar to ``py::init_alias<...>()``,
+as described in :ref:`extended_aliases`.
+
+The following example shows the different factory approaches for a class with
+an alias:
+
+.. code-block:: cpp
+
+    #include <pybind11/factory.h>
+    class Example {
+    public:
+        // ...
+        virtual ~Example() = default;
+    };
+    class PyExample : public Example {
+    public:
+        using Example::Example;
+        PyExample(Example &&base) : Example(std::move(base)) {}
+    };
+    py::class_<Example, PyExample>(m, "Example")
+        // Returns an Example pointer.  If a PyExample is needed, the Example
+        // instance will be moved via the extra constructor in PyExample, above.
+        .def(py::init([]() { return new Example(); }))
+        // Two callbacks:
+        .def(py::init([]() { return new Example(); } /* no alias needed */,
+                      []() { return new PyExample(); } /* alias needed */))
+        // *Always* returns an alias instance (like py::init_alias<>())
+        .def(py::init([]() { return new PyExample(); }))
+        ;
+
+Low-level placement-new construction
+------------------------------------
+
+A second approach for creating new instances use C++ placement new to construct
+an object in-place in preallocated memory.  To do this, you simply bind a
+method name ``__init__`` that takes the class instance as the first argument by
+pointer or reference, then uses a placement-new constructor to construct the
+object in the pre-allocated (but uninitialized) memory.
+
+For example, instead of:
 
 .. code-block:: cpp
 
     py::class_<Example>(m, "Example")
         .def(py::init<int>());
 
-is short hand notation for
+you could equivalently write:
 
 .. code-block:: cpp
 
@@ -378,9 +488,7 @@ is short hand notation for
             }
         );
 
-In other words, :func:`init` creates an anonymous function that invokes an
-in-place constructor. Memory allocation etc. is already take care of beforehand
-within pybind11.
+which will invoke the constructor in-place at the pre-allocated memory.
 
 .. _classes_with_non_public_destructors:
 
--- a/include/pybind11/detail/init.h
+++ b/include/pybind11/detail/init.h
@@ -40,7 +40,7 @@ PYBIND11_NOINLINE inline value_and_holder load_v_h(handle self_, type_info *tinf
 }
 
 
-// Implementing functions for py::init(...)
+// Implementing functions for all forms of py::init<...> and py::init(...)
 template <typename Class> using Cpp = typename Class::type;
 template <typename Class> using Alias = typename Class::type_alias;
 template <typename Class> using Holder = typename Class::holder_type;
@@ -160,6 +160,62 @@ void construct(value_and_holder &v_h, Alias<Class> &&result, bool) {
         "pybind11::init() return-by-alias-value factory function requires a movable alias class");
     deallocate(v_h) = new Alias<Class>(std::move(result));
 }
+
+// Implementing class for py::init<...>()
+template <typename... Args> struct constructor {
+    template <typename Class, typename... Extra, enable_if_t<!Class::has_alias, int> = 0>
+    static void execute(Class &cl, const Extra&... extra) {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        cl.def("__init__", [cl_type](handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // Ignore duplicate __init__ calls (see above)
+
+            construct<Class>(v_h, new Cpp<Class>(std::forward<Args>(args)...), false);
+        }, extra...);
+    }
+
+    template <typename Class, typename... Extra,
+              enable_if_t<Class::has_alias &&
+                          std::is_constructible<Cpp<Class>, Args...>::value, int> = 0>
+    static void execute(Class &cl, const Extra&... extra) {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        cl.def("__init__", [cl_type](handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // Ignore duplicate __init__ calls (see above)
+
+            if (Py_TYPE(v_h.inst) == cl_type->type)
+                construct<Class>(v_h, new Cpp<Class>(std::forward<Args>(args)...), false);
+            else
+                construct<Class>(v_h, new Alias<Class>(std::forward<Args>(args)...), true);
+        }, extra...);
+    }
+
+    template <typename Class, typename... Extra,
+              enable_if_t<Class::has_alias &&
+                          !std::is_constructible<Cpp<Class>, Args...>::value, int> = 0>
+    static void execute(Class &cl, const Extra&... extra) {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        cl.def("__init__", [cl_type](handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // Ignore duplicate __init__ calls (see above)
+            construct<Class>(v_h, new Alias<Class>(std::forward<Args>(args)...), true);
+        }, extra...);
+    }
+};
+
+// Implementing class for py::init_alias<...>()
+template <typename... Args> struct alias_constructor {
+    template <typename Class, typename... Extra,
+              enable_if_t<Class::has_alias && std::is_constructible<Alias<Class>, Args...>::value, int> = 0>
+    static void execute(Class &cl, const Extra&... extra) {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        cl.def("__init__", [cl_type](handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // Ignore duplicate __init__ calls (see above)
+            construct<Class>(v_h, new Alias<Class>(std::forward<Args>(args)...), true);
+        }, extra...);
+    }
+};
 
 // Implementation class for py::init(Func) and py::init(Func, AliasFunc)
 template <typename CFunc, typename AFuncIn, typename... Args> struct factory {
--- a/include/pybind11/detail/init.h
+++ b/include/pybind11/detail/init.h
@@ -1,0 +1,274 @@
+/*
+    pybind11/detail/init.h: init factory function implementation and support code.
+
+    Copyright (c) 2017 Jason Rhinelander <jason@imaginary.ca>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include "class.h"
+
+NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+NAMESPACE_BEGIN(initimpl)
+
+inline void no_nullptr(void *ptr) {
+    if (!ptr) throw type_error("pybind11::init(): factory function returned nullptr");
+}
+
+// Makes sure the `value` for the given value_and_holder is not preallocated (e.g. by a previous
+// old-style placement new `__init__` that requires a preallocated, uninitialized value).  If
+// preallocated, deallocate.  Returns the (null) value pointer reference ready for allocation.
+inline void *&deallocate(value_and_holder &v_h) {
+    if (v_h) v_h.type->dealloc(v_h);
+    return v_h.value_ptr();
+}
+
+PYBIND11_NOINLINE inline value_and_holder load_v_h(handle self_, type_info *tinfo) {
+    if (!self_ || !tinfo)
+        throw type_error("__init__(self, ...) called with invalid `self` argument");
+
+    auto *inst = reinterpret_cast<instance *>(self_.ptr());
+    auto result = inst->get_value_and_holder(tinfo, false);
+    if (!result.inst)
+        throw type_error("__init__(self, ...) called with invalid `self` argument");
+
+    return result;
+}
+
+
+// Implementing functions for py::init(...)
+template <typename Class> using Cpp = typename Class::type;
+template <typename Class> using Alias = typename Class::type_alias;
+template <typename Class> using Holder = typename Class::holder_type;
+
+template <typename Class> using is_alias_constructible = std::is_constructible<Alias<Class>, Cpp<Class> &&>;
+
+// Takes a Cpp pointer and returns true if it actually is a polymorphic Alias instance.
+template <typename Class, enable_if_t<Class::has_alias, int> = 0>
+bool is_alias(Cpp<Class> *ptr) {
+    return dynamic_cast<Alias<Class> *>(ptr) != nullptr;
+}
+// Failing fallback version of the above for a no-alias class (always returns false)
+template <typename /*Class*/>
+constexpr bool is_alias(void *) { return false; }
+
+// Attempts to constructs an alias using a `Alias(Cpp &&)` constructor.  This allows types with
+// an alias to provide only a single Cpp factory function as long as the Alias can be
+// constructed from an rvalue reference of the base Cpp type.  This means that Alias classes
+// can, when appropriate, simply define a `Alias(Cpp &&)` constructor rather than needing to
+// inherit all the base class constructors.
+template <typename Class>
+void construct_alias_from_cpp(std::true_type /*is_alias_constructible*/,
+                              value_and_holder &v_h, Cpp<Class> &&base) {
+    deallocate(v_h) = new Alias<Class>(std::move(base));
+}
+template <typename Class>
+[[noreturn]] void construct_alias_from_cpp(std::false_type /*!is_alias_constructible*/,
+                                           value_and_holder &, Cpp<Class> &&) {
+    throw type_error("pybind11::init(): unable to convert returned instance to required "
+                     "alias class: no `Alias<Class>(Class &&)` constructor available");
+}
+
+// Error-generating fallback for factories that don't match one of the below construction
+// mechanisms.
+template <typename Class>
+void construct(...) {
+    static_assert(!std::is_same<Class, Class>::value /* always false */,
+            "pybind11::init(): init function must return a compatible pointer, "
+            "holder, or value");
+}
+
+// Pointer return v1: the factory function returns a class pointer for a registered class.
+// If we don't need an alias (because this class doesn't have one, or because the final type is
+// inherited on the Python side) we can simply take over ownership.  Otherwise we need to try to
+// construct an Alias from the returned base instance.
+template <typename Class>
+void construct(value_and_holder &v_h, Cpp<Class> *ptr, bool need_alias) {
+    no_nullptr(ptr);
+    if (Class::has_alias && need_alias && !is_alias<Class>(ptr)) {
+        // We're going to try to construct an alias by moving the cpp type.  Whether or not
+        // that succeeds, we still need to destroy the original cpp pointer (either the
+        // moved away leftover, if the alias construction works, or the value itself if we
+        // throw an error), but we can't just call `delete ptr`: it might have a special
+        // deleter, or might be shared_from_this.  So we construct a holder around it as if
+        // it was a normal instance, then steal the holder away into a local variable; thus
+        // the holder and destruction happens when we leave the C++ scope, and the holder
+        // class gets to handle the destruction however it likes.
+        deallocate(v_h) = ptr;
+        v_h.set_instance_registered(true); // To prevent init_instance from registering it
+        v_h.type->init_instance(v_h.inst, nullptr); // Set up the holder
+        Holder<Class> temp_holder(std::move(v_h.holder<Holder<Class>>())); // Steal the holder
+        v_h.type->dealloc(v_h); // Destroys the moved-out holder remains, resets value ptr to null
+        v_h.set_instance_registered(false);
+
+        construct_alias_from_cpp<Class>(is_alias_constructible<Class>{}, v_h, std::move(*ptr));
+    }
+    else {
+        // Otherwise the type isn't inherited, so we don't need an Alias and can just store the Cpp
+        // pointer directory:
+        deallocate(v_h) = ptr;
+    }
+}
+
+// Pointer return v2: a factory that always returns an alias instance ptr.  We simply take over
+// ownership of the pointer.
+template <typename Class, enable_if_t<Class::has_alias, int> = 0>
+void construct(value_and_holder &v_h, Alias<Class> *alias_ptr, bool) {
+    no_nullptr(alias_ptr);
+    deallocate(v_h) = static_cast<Cpp<Class> *>(alias_ptr);
+}
+
+// Holder return: copy its pointer, and move or copy the returned holder into the new instance's
+// holder.  This also handles types like std::shared_ptr<T> and std::unique_ptr<T> where T is a
+// derived type (through those holder's implicit conversion from derived class holder constructors).
+template <typename Class>
+void construct(value_and_holder &v_h, Holder<Class> holder, bool need_alias) {
+    auto *ptr = holder_helper<Holder<Class>>::get(holder);
+    // If we need an alias, check that the held pointer is actually an alias instance
+    if (Class::has_alias && need_alias && !is_alias<Class>(ptr))
+        throw type_error("pybind11::init(): construction failed: returned holder-wrapped instance "
+                         "is not an alias instance");
+
+    deallocate(v_h) = ptr;
+    v_h.type->init_instance(v_h.inst, &holder);
+}
+
+// return-by-value version 1: returning a cpp class by value.  If the class has an alias and an
+// alias is required the alias must have an `Alias(Cpp &&)` constructor so that we can construct
+// the alias from the base when needed (i.e. because of Python-side inheritance).  When we don't
+// need it, we simply move-construct the cpp value into a new instance.
+template <typename Class>
+void construct(value_and_holder &v_h, Cpp<Class> &&result, bool need_alias) {
+    static_assert(std::is_move_constructible<Cpp<Class>>::value,
+        "pybind11::init() return-by-value factory function requires a movable class");
+    if (Class::has_alias && need_alias)
+        construct_alias_from_cpp<Class>(is_alias_constructible<Class>{}, v_h, std::move(result));
+    else
+        deallocate(v_h) = new Cpp<Class>(std::move(result));
+}
+
+// return-by-value version 2: returning a value of the alias type itself.  We move-construct an
+// Alias instance (even if no the python-side inheritance is involved).  The is intended for
+// cases where Alias initialization is always desired.
+template <typename Class>
+void construct(value_and_holder &v_h, Alias<Class> &&result, bool) {
+    static_assert(std::is_move_constructible<Alias<Class>>::value,
+        "pybind11::init() return-by-alias-value factory function requires a movable alias class");
+    deallocate(v_h) = new Alias<Class>(std::move(result));
+}
+
+// Implementation class for py::init(Func) and py::init(Func, AliasFunc)
+template <typename CFunc, typename AFuncIn, typename... Args> struct factory {
+private:
+    using CFuncType = typename std::remove_reference<CFunc>::type;
+    using AFunc = conditional_t<std::is_void<AFuncIn>::value, void_type, AFuncIn>;
+    using AFuncType = typename std::remove_reference<AFunc>::type;
+
+    CFuncType class_factory;
+    AFuncType alias_factory;
+
+public:
+    // Constructor with a single function/lambda to call; for classes without aliases or with
+    // aliases that can be move constructed from the base.
+    factory(CFunc &&f) : class_factory(std::forward<CFunc>(f)) {}
+
+    // Constructor with two functions/lambdas, for a class with distinct class/alias factories: the
+    // first is called when an alias is not needed, the second when the alias is needed.  Requires
+    // non-void AFunc.
+    factory(CFunc &&c, AFunc &&a) :
+        class_factory(std::forward<CFunc>(c)), alias_factory(std::forward<AFunc>(a)) {}
+
+    // Add __init__ definition for a class that either has no alias or has no separate alias
+    // factory; this always constructs the class itself.  If the class is registered with an alias
+    // type and an alias instance is needed (i.e. because the final type is a Python class
+    // inheriting from the C++ type) the returned value needs to either already be an alias
+    // instance, or the alias needs to be constructible from a `Class &&` argument.
+    template <typename Class, typename... Extra,
+              enable_if_t<!Class::has_alias || std::is_void<AFuncIn>::value, int> = 0>
+    void execute(Class &cl, const Extra&... extra) && {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        #if defined(PYBIND11_CPP14)
+        cl.def("__init__", [cl_type, func = std::move(class_factory)]
+        #else
+        CFuncType &func = class_factory;
+        cl.def("__init__", [cl_type, func]
+        #endif
+        (handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            // If this value is already registered it must mean __init__ is invoked multiple times;
+            // we really can't support that in C++, so just ignore the second __init__.
+            if (v_h.instance_registered()) return;
+
+            construct<Class>(v_h, func(std::forward<Args>(args)...), Py_TYPE(v_h.inst) != cl_type->type);
+        }, extra...);
+    }
+
+    // Add __init__ definition for a class with an alias *and* distinct alias factory; the former is
+    // called when the `self` type passed to `__init__` is the direct class (i.e. not inherited), the latter
+    // when `self` is a Python-side subtype.
+    template <typename Class, typename... Extra,
+              enable_if_t<Class::has_alias && !std::is_void<AFuncIn>::value, int> = 0>
+    void execute(Class &cl, const Extra&... extra) && {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+
+        #if defined(PYBIND11_CPP14)
+        cl.def("__init__", [cl_type, class_func = std::move(class_factory), alias_func = std::move(alias_factory)]
+        #else
+        CFuncType &class_func = class_factory;
+        AFuncType &alias_func = alias_factory;
+        cl.def("__init__", [cl_type, class_func, alias_func]
+        #endif
+        (handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // (see comment above)
+
+            if (Py_TYPE(v_h.inst) == cl_type->type)
+                // If the instance type equals the registered type we don't have inheritance, so
+                // don't need the alias and can construct using the class function:
+                construct<Class>(v_h, class_func(std::forward<Args>(args)...), false);
+            else
+                construct<Class>(v_h, alias_func(std::forward<Args>(args)...), true);
+        }, extra...);
+    }
+};
+
+template <typename Func> using functype =
+    conditional_t<std::is_function<remove_reference_t<Func>>::value, remove_reference_t<Func> *,
+    conditional_t<is_function_pointer<remove_reference_t<Func>>::value, remove_reference_t<Func>,
+    Func>>;
+
+// Helper definition to infer the detail::initimpl::factory template types from a callable object
+template <typename Func, typename Return, typename... Args>
+factory<functype<Func>, void, Args...> func_decltype(Return (*)(Args...));
+
+// metatemplate that ensures the Class and Alias factories take identical arguments: we need to be
+// able to call either one with the given arguments (depending on the final instance type).
+template <typename Return1, typename Return2, typename... Args1, typename... Args2>
+inline constexpr bool require_matching_arguments(Return1 (*)(Args1...), Return2 (*)(Args2...)) {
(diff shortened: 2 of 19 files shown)


jagerman:
(Credit to @YannickJadoul and @bmerry for the idea).

jagerman:
Also worth noting: this isn't affected by #804, because the `def_static("__init__", ...)` support here doesn't actually bind a *python*-side static method, so `def("__init__")` and `def_static("__init__")` can be mixed.

wjakob:
This strikes me a fairly intrusive change to support a relatively narrow set of situations. (In particular, the ``instance_swap()`` function, and that we need to stick something into the dispatch code regardless of whether this feature is used or not.)

As an alternative, could we have something like:

```cpp
m.def("__init__", py::factory_adapter(&Example::create));
```

where the ``factory_adapter`` either uses an in-place move constructor, or it takes a ``py::object`` as ``self`` argument and then does something similar to ``instance_swap``.

jagerman:
> As an alternative, could we have something like:
>
> ```C++
> m.def("__init__", py::factory_adapter(&Example::create));
> ```
> where the factory_adapter either uses an in-place move constructor, or it takes a py::object as self argument and then does something similar to instance_swap.

Yeah, that should be doable.  I'll take another shot at this in the next day or two.

jagerman:
And that MSVC job failure is more bizarre than ever.

YannickJadoul:
>it takes a py::object as self argument and then does something similar to instance_swap.

I don't want to come across whiny, but could I bring up that gist I've mentioned in the gitter chat, again?
https://gist.github.com/YannickJadoul/38dc4ef8f0faa56fa10824626886edab

I'm sure it could be improved a lot with better knowledge of pybind11, but what it more or less does is:
- Define a class `unconstructed_holder` which takes care of the deallocation of the allocated memory for the instance, then constructs the holder immediately.
- Define `type_caster<unconstructed_holder>` that is only able to load a `py::instance`, not just extracting the pointer to the instance but also the memory reserved for the holder

jagerman:
Redesigned this.

The redesign is far less intrusive, and avoids some unnecessary Python object creation in several cases.

I still want to add some tests to verify all the various edge cases (e.g. for things like returning an alias pointer or value, or a subclass pointer).

docs/advanced/classes.rst
```diff
@@ -366,6 +366,35 @@ In other words, :func:`init` creates an anonymous function that invokes an
...
+    class Example {
+        // ...
+        static Example *create(int a) { return new Example(a); }
+    };
+
+    py::class_<Example>(m, "Example")
+        // Bind an existing pointer-returning factory function:
+        .def(py::init_factory(&Example::create))
+        // Similar, but returns the pointer wrapped in a holder:
+        .def("__init__", []() {
```
wjakob: ``init_factory`` missing.

wjakob:
Nifty -- +1 for covering all the different kinds of things that could be returned. Since the ``detail::factory_init`` got quite large, I'm wondering if it might make sense to move it into a separate header file (e.g. ``pybind11/factory.h``) since it addresses a fairly specific requirement.

jagerman:
As far as organization, moving it to another header is a good idea.  It can't *quite* be all moved, though: the `class_::init(py::init_factory &&)` member function is still needed (because it needs to pass the `class_<...>` type to `execute()`).

wjakob:
Yep, that is to be expected. With a forward declaration, it should be possible though, no?.

jagerman:
I assume you meant making it an optional (i.e. requires an extra include before using) header.

YannickJadoul:
I don't know about all the subtleties of handling e.g. `py::handle` or `py::object`, but this seems a really nice solution to the problem to me, especially the way it hardly requires an internal change :-)
Trying this out on my code and test case I'd written works like a charm.

I've got two possibly useful things:
- As far as I know, `init_holder` is still called afterwards, which might mess things up by overwriting the holder once again? I.e., since the `owned` field of the instance is still true, `init_holder` will think it needs to wrap the pointer again in a holder? (But maybe I've overlooked something.)
- Just a detail, but nevertheless: I believe your `alias_constructible` does not need SFINAE but can do with just a `using alias_constructible = std::integral_constant<bool, that_long_boolean_expression>` ?

But for the rest, just "wow", again!

jagerman:
Re: `init_holder`: I added a check to the if-constructor-then-init-holder call in `dispatch`, but apparently it got lost when I was cleaning up the commit: good catch!

The SFINAE was necessary because the `bool_constant` version fails when the type has no alias because `Alias<Class>` is `void` and so `Alias<Class> &&` is taking a reference to void, which is invalid.

YannickJadoul:
>The SFINAE was necessary because the bool_constant version fails when the type has no alias because Alias<Class> is void and so Alias<Class> && is taking a reference to void, which is invalid.

Aha, my bad, sorry

But thanks again for the quick and elegant solution!

jagerman:
I've added various tests and fixes related to value and alias initialization, and moved the tests out into a new `test_factory_constructors` rather than bulking up `method_and_attributes` with things that are only sort-of methods.

wjakob:
This probably needs to be rebased on the latest master, it looks like the PR has some conflicts and contains MI-related changes.

jagerman:
Fixed.  (This was already rebased against #812--I just had to delete all the duplicate commits).

jagerman:
(There's still an outstanding MSVC issue I need to track down here).

jagerman:
Oh, it was just an annoying warning (because `factory.h` is outside `pybind11.h`'s ignore warnings stack).

include/pybind11/factory.h
```diff
@@ -0,0 +1,269 @@
...
+    using ForwardReturn = typename std::add_rvalue_reference<Return>::type;
+
+    template <typename Class> using Cpp = typename Class::type;
+    template <typename Class> using Alias = typename Class::type_alias;
+    template <typename Class> using Inst = typename Class::instance_type;
+    template <typename Class> using Holder = typename Class::holder_type;
+
+    template <typename Class, typename SFINAE = void> struct alias_constructible : std::false_type {};
+    template <typename Class> struct alias_constructible<Class, enable_if_t<Class::has_alias &&
+            std::is_convertible<Return, Alias<Class>>::value && std::is_constructible<Alias<Class>, Alias<Class> &&>::value>>
```
dean0x7d: `std::is_constructible<Alias<Class>, Alias<Class> &&>` == `std::is_move_constructible<Alias<Class>>` which might better reveal the intent.

include/pybind11/factory.h
```diff
@@ -0,0 +1,269 @@
...
+        : std::true_type {};
+    template <typename Class> using cpp_constructible = bool_constant<
+        !(Class::has_alias && std::is_base_of<Alias<Class>, Return>::value) &&
+        std::is_convertible<Return, Cpp<Class>>::value && std::is_constructible<Cpp<Class>, Cpp<Class> &&>::value>;
```
dean0x7d: Same here.

include/pybind11/factory.h
jagerman: Good point.  I had thought `is_move_constructible` was more specifically "has move constructor", but I see that's not the case.

jagerman:
> Did I overlook something?

The idea I had in mind for that was making it easier to return values that come from Python.  So, for example, a particular constructor could defer to a Python factory function to create the instance without having to do the unboxing in a C++ factory function, while wrapping the whole thing up by exposing it as a constructor.  I'll add a couple of example of this to the tests.

I do agree that making a pure C++ factory function wrap the return in a `py::object` seems fairly pointless.

jagerman:
One last question: should `factory.h` get included by default by `pybind11.h`?  It doesn't actually add any dependencies, and is all templated code, so shouldn't really affect anything if it isn't actually being used.

jagerman:
> > Did I overlook something?

Looking at the pypy failure, I think you're right that this isn't useful; there's no guarantee a python factory function will be uniquely referenced (though seems to be in cpython).  I'll go ahead and remove it.

dean0x7d:
> @jagerman
> One last question: should factory.h get included by default by pybind11.h? It doesn't actually add any dependencies, and is all templated code, so shouldn't really affect anything if it isn't actually being used.

I think @wjakob's suggestion for moving out the header was for compilation performance (even just parsing C++ is difficult because so much is context sensitive). I think that if some of the metafunctions could be simplified a bit (more below) it might be worthwhile to include it in `pybind11.h` just to better organize the features (make them easier to teach). Specifically, with the separate header, there are 3 ways of creating constructors which look like this:
```c++
.def(py::init<Args...>())
.def("__init__", [](Class&, Args...) { ... })
.def(py::init_factory([](Args...) { ... }))
```
where the last one is a superset of the second one, but requires an extra header. The 3 methods also use wildly different syntax. If the factory is made into a first-class feature, this could become more uniform:
```c++
.def(py::init<Args...>())
.def(py::init([](Args...) { ... }))
```
Note that `init_factory` is renamed to `init` (just an overload of the existing `init` function). The factory `init` is a type-safe alternative to placement `new`, so it would be preferable to teach this in the docs (leave the raw `__init__` as an expert-level feature).

To make the factory implementation a bit more light-weight, would it be possible to rely more on simple overload resolution than the metafunctions + tags? To me, it looks like `construct` overloads are already unique, so the `factory_type` metafunction may not be completely necessary.

YannickJadoul:
> The factory `init` is a type-safe alternative to placement `new`

Just as a side note: this type-safety only relates to the C++ side of things. Because of the `py::handle` as first argument and the implicit assumption that this will always be of the right type, you cán actually break this at 'Python runtime' by doing something silly like `X.__init__(Y.__new__(Y))`.
(I know, very far fetched, but pybind11 won't stop you from segfaulting your Python instance, while - I think - the standard `init` would.)

dean0x7d:
> Because of the py::handle as first argument and the implicit assumption that this will always be of the right type, you cán actually break this at 'Python runtime' by doing something silly like `X.__init__(Y.__new__(Y))`.

Interesting point. I guess a quick type check should be able to prevent that.

jagerman:
I took another stab at this in light of @dean0x7d 's comments and some other issues to simplify the interface and implementation.  Most of the issues related to alias classes.  The implementation is a bit lighter, but the code logic grew a bit to handle the special cases aliases bring up.

Summary of changes:
- renamed `py::init_factory` to `py::init`.  This is easy to revert, but this doesn't conflict with the templated no-argument `py::init<...>()`, while keeping the naming nicer (especially if we decide to roll this into the core, documented approach replacing raw placement-new `__init__`s).
- factories can no longer return types *convertible* to the base/alias class: they can now return only a {base,alias} {value,pointer}, a holder, or a shared_ptr<T> to some base or derived `T` (only applies when the class has a `shared_ptr` holder to begin with).
- cutting out the convertible support allowed the implementation to be a bit lighter by avoiding the need for the tag classes (which were needed to prioritize ambiguous calls).
- If returning a base class (whether by value or pointer) when an alias is required (i.e. python-side inheritance and a declared alias), a dynamic_cast to the alias is attempted (for the pointer version); if it fails, or if returned by value, an Alias(Class &&) constructor needs to be defined (otherwise a type_error is thrown).
- for holder returns when an alias is required, we try a dynamic_cast of the pointer to the alias; if it fails, we raise an error.
- a new `py::init(class_factory, alias_factory)` function is available that takes two factories: the first is called when an alias is not needed, the second when it is.  (This, in particular, is how holder-returning factories can still be used, but puts the onus on the caller to sort out the base/alias creation logic).

(I'm holding off on the writing the documentation until the implementation looks good).

bmerry:
I'm not clear why one would want to accept a factory function that returns a `shared_ptr<T>` where T is a base class of the class being defined (as opposed to a derived class). That sounds like it discards type safety, and if the user really wants to do that, they can write it in a lambda that does the pointer cast themselves.

However, I have barely used pybind11 and I don't really understand the aliases concept, so I might be missing something.

jagerman:
@bmerry - The story is the same for both base class pointers and `std::shared_ptr<Base>`: it doesn't discard type safety as long as the type is polymorphic.  It *does* need a `dynamic_cast` (via `std::dynamic_pointer_cast` for a `shared_ptr`) to do a runtime RTTI check as to whether the pointer actually is the derived type, and the get the derived pointer (which doesn't have to be equal to the base pointer, particularly when multiple inheritance is involved).

bmerry:
I was thinking more of static than dynamic type safety. If I saw code that used a factory function returning `shared_ptr<Base>` being used as a constructor for the Python wrapper of Derived, my first guess would be that either
(a) The factory only ever returns `shared_ptr<Derived>`, and ought to have been declared as such; or
(b) The factory doesn't always return `shared_ptr<Derived>`, and the user either ought to be wrapping Base rather than Derived, or they meant to use a different factory function.

Maybe there are legitimate use cases where one wishes to do this, and get an exception thrown if the Python caller passes a combination of arguments that causes the factory not to spit out a `shared_ptr<Derived>`, but my (prejudiced, non-scientific) feeling is that this would be much rarer than the cases above. Removing the support for it from pybind11 would help catch errors at compile time. It wouldn't prevent users implementing this use case, because the user can write a wrapper around his/her factory function that does the dynamic cast, and perhaps raises a more application-aware error whose message explains why the constructor is failing.

docs/advanced/classes.rst
```diff
@@ -366,6 +366,51 @@ In other words, :func:`init` creates an anonymous function that invokes an
...
+    class Example {
+        // ...
+        static Example *create(int a) { return new Example(a); }
+    };
+    py::class_<Example>(m, "Example")
+        // Bind an existing pointer-returning factory function:
+        .def(py::init_factory(&Example::create))
+        // Similar, but returns the pointer wrapped in a holder:
+        .def(py::init_factory([](std::string arg) {
+            return std::unique_ptr<Example>(new Example(arg, "another arg"));
```
bmerry: Should this be changed init_factory -> init now?

include/pybind11/factory.h
```diff
@@ -0,0 +1,316 @@
...
+    return true;
+}
+template <typename CFunc, typename AFunc,
+          typename CReturn, typename... CArgs, typename AReturn, typename... AArgs,
+          bool = init_factory_require_matching_arguments((CReturn (*)(CArgs...)) nullptr, (AReturn (*)(AArgs...)) nullptr)>
+init_factory<CFunc, CReturn, AFunc, AReturn, CArgs...> init_factory_decltype(
+    CReturn (*)(CArgs...), AReturn (*)(AArgs...));
+
+template <typename... Func> using init_factory_t = decltype(init_factory_decltype<Func...>(
+    (typename detail::remove_class<decltype(&std::remove_reference<Func>::type::operator())>::type *) nullptr...));
```
bmerry: When I tried this out with a plain function it didn't work. I suspect it's because a plain function doesn't have a `::operator()` when it tries to expand this. I can't think of a simple way to get the return type of either a function or a function object (`result_of` requires you to know the calling argument types). I guess it'll just need another traits helper class to handle the various cases.

docs/advanced/classes.rst
YannickJadoul: Yes, but @jagerman said he'd wait before updating that
> (I'm holding off on the writing the documentation until the implementation looks good).

include/pybind11/factory.h
bmerry: Your fix for this is working for me - thanks.

include/pybind11/factory.h
```diff
@@ -0,0 +1,322 @@
...
+template <typename type, typename... options>
+template <typename... Args, typename... Extra>
+class_<type, options...> &class_<type, options...>::def(detail::init_factory<Args...> &&init, const Extra&... extra) {
+    std::move(init).execute(*this, extra...);
+    return *this;
+}
+
+NAMESPACE_BEGIN(detail)
+
+inline static void init_factory_no_nullptr(void *ptr) {
```
dean0x7d: Shouldn't be `static`.

dean0x7d:
> @bmerry said:
> I'm not clear why one would want to accept a factory function that returns a shared_ptr<T> where T is a base class of the class being defined (as opposed to a derived class).
>
> If I saw code that used a factory function returning shared_ptr<Base> being used as a constructor for the Python wrapper of Derived, my first guess would be that either
(a) The factory only ever returns shared_ptr<Derived>, and ought to have been declared as such; or
(b) The factory doesn't always return shared_ptr<Derived>, and the user either ought to be wrapping Base rather than Derived, or they meant to use a different factory function.

I think this is a good point. `init_factory` is checking everything at runtime, but really, the user should already know this when binding a constructor. A function that can return an invalid type depending on the given arguments isn't a good choice for a constructor anyway.

A pretty large portion of the code is dedicated to supporting downcasting (+ double that because of the alias constructor). Removing it would be major simplification.

jagerman:
I've updated this PR to slim this down some more by removing the downcasting support (except for the `Base *` -> `Alias *` cast attempt when we need an Alias), which simplifies it quite a bit.

Thoughts, now, on including this by default (and updating to docs to prefer this over placement-new in simpler cases)?

Any other feedback?

jagerman:
Oops, I missed some of @dean0x7d's comments, I'll update with them shortly.

include/pybind11/common.h
```diff
@@ -509,6 +509,31 @@ template <typename T>
 struct is_input_iterator<T, void_t<decltype(*std::declval<T &>()), decltype(++std::declval<T &>())>>
     : std::true_type {};
 
+template <typename T> using is_function_pointer = bool_constant<
+    std::is_pointer<T>::value && std::is_function<typename std::remove_pointer<T>::type>::value>;
+
+// Extracts the function signature from a function, function pointer or lambda.
+template <typename F, typename SFINAE = void> struct function_signature {};
```
jagerman: This fails to compile because of the `&F::operator()` when `F` is a plain function pointer; any idea for a workaround?

dean0x7d: Untested, but try:
```c++
typename std::conditional_t<
    std::is_pointer<F>::value || std::is_member_pointer<F>::value,
    remove_pointer<F>, // Note: no `_t`
    strip_function_object<F>
>::type // deferred `::type` access
```
where:
```c++
template <typename F>
struct strip_function_object {
    using type = typename remove_class<decltype(&F::operator())>::type;
}
```

jagerman: That worked.  I actually dropped the whole `struct function_signature` and just declared `using function_signature_t = conditional_t<...>`.

jagerman:
I've rebased and updated this PR with the new testing style.

As for whether or not to include it by default, I ran some benchmarks building the test suite, testing both compiler CPU time and peak memory usage (RSS).  The only change between the two runs below is whether `#include <pybind11/factory.h>` is added to `pybind11_tests.h` (i.e. included in every compilation unit).

Here are the gcc 7.1.0 results:
```
                                   gcc-wo-factory.txt [n=50]                     gcc-w-factory.txt [n=50]              
                                ================================   ====================================================
                                 CPU   (s.d. )   MaxRSS ( s.d. )    CPU   (s.d. ) (   ±   )   MaxRSS ( s.d. ) (   ±   )
                                --------------   ---------------   ------------------------   -------------------------
                   (linking):   27.865 (0.119)   288764 (  82.8)   27.866 (0.126) ( +0.01%)   288786 (  67.6) ( +0.01%)
              pybind11_tests:    2.201 (0.029)   294996 ( 132.3)    2.206 (0.029) ( +0.22%)   295576 (  51.9) ( +0.20%)
                test_buffers:    2.557 (0.025)   310068 (  79.9)    2.567 (0.027) ( +0.36%)   310264 ( 118.4) ( +0.06%)
        test_builtin_casters:    2.969 (0.038)   340145 (  98.6)    2.967 (0.034) ( -0.07%)   339710 (  78.9) ( -0.13%)
          test_call_policies:    1.882 (0.031)   254697 (  43.4)    1.882 (0.026) ( -0.01%)   255205 (  77.1) ( +0.20%)
              test_callbacks:    2.735 (0.033)   310796 (  50.5)    2.740 (0.030) ( +0.18%)   310648 ( 131.2) ( -0.05%)
                 test_chrono:    1.762 (0.027)   252543 ( 125.8)    1.771 (0.027) ( +0.53%)   252720 (  70.0) ( +0.07%)
                  test_class:    3.057 (0.035)   356680 (  54.0)    3.069 (0.039) ( +0.38%)   356962 (  54.4) ( +0.08%)
test_constants_and_functions:    2.136 (0.028)   291964 ( 100.5)    2.133 (0.029) ( -0.18%)   292425 ( 107.2) ( +0.16%)
              test_copy_move:    3.064 (0.035)   342694 ( 134.9)    3.061 (0.034) ( -0.10%)   342977 (  60.1) ( +0.08%)
      test_docstring_options:    1.761 (0.030)   242729 ( 134.8)    1.755 (0.030) ( -0.32%)   243192 ( 140.1) ( +0.19%)
                  test_eigen:    9.299 (0.091)   728918 (  57.2)    9.283 (0.075) ( -0.17%)   729366 (  58.9) ( +0.06%)
                   test_enum:    3.316 (0.035)   354021 (  92.6)    3.312 (0.037) ( -0.10%)   353723 (  75.8) ( -0.08%)
                   test_eval:    1.524 (0.021)   215105 (  60.4)    1.527 (0.023) ( +0.20%)   215526 (  54.1) ( +0.20%)
             test_exceptions:    1.459 (0.026)   208468 (  67.0)    1.463 (0.019) ( +0.23%)   208796 (  78.5) ( +0.16%)
   test_factory_constructors:    4.581 (0.050)   418475 (  53.9)    4.564 (0.033) ( -0.38%)   418496 (  60.1) ( +0.01%)
    test_kwargs_and_defaults:    2.121 (0.023)   287457 (  52.0)    2.121 (0.027) ( +0.03%)   287698 (  58.3) ( +0.08%)
 test_methods_and_attributes:    4.541 (0.043)   390432 (  55.8)    4.545 (0.051) ( +0.09%)   390622 (  52.6) ( +0.05%)
                test_modules:    2.200 (0.024)   296003 (  51.6)    2.205 (0.026) ( +0.25%)   296336 (  93.7) ( +0.11%)
   test_multiple_inheritance:    4.896 (0.051)   454484 (  44.4)    4.896 (0.051) ( +0.02%)   454642 (  73.2) ( +0.03%)
            test_numpy_array:    4.222 (0.044)   370925 (  56.9)    4.218 (0.039) ( -0.09%)   371269 (  98.1) ( +0.09%)
           test_numpy_dtypes:    3.072 (0.031)   333738 ( 132.7)    3.067 (0.033) ( -0.14%)   333883 ( 109.4) ( +0.04%)
        test_numpy_vectorize:    3.168 (0.038)   342012 (  82.3)    3.172 (0.037) ( +0.15%)   341343 (  71.4) ( -0.20%)
           test_opaque_types:    2.180 (0.028)   295067 (  84.8)    2.189 (0.033) ( +0.41%)   295338 ( 122.7) ( +0.09%)
   test_operator_overloading:    2.877 (0.030)   334324 ( 135.7)    2.881 (0.043) ( +0.14%)   334708 (  98.7) ( +0.11%)
               test_pickling:    1.853 (0.030)   258136 ( 108.8)    1.855 (0.024) ( +0.12%)   258567 (  76.1) ( +0.17%)
                test_pytypes:    2.088 (0.029)   270000 (  46.6)    2.091 (0.024) ( +0.13%)   270404 ( 105.9) ( +0.15%)
test_sequences_and_iterators:    4.152 (0.037)   377033 (  62.2)    4.147 (0.041) ( -0.12%)   378544 (  79.7) ( +0.40%)
              test_smart_ptr:    4.205 (0.036)   378643 (  69.6)    4.212 (0.037) ( +0.19%)   378940 (  53.0) ( +0.08%)
                    test_stl:    3.546 (0.032)   366140 (  89.6)    3.549 (0.038) ( +0.08%)   366496 ( 111.9) ( +0.10%)
            test_stl_binders:   11.039 (0.090)   806419 ( 145.5)   11.037 (0.088) ( -0.02%)   806921 ( 186.6) ( +0.06%)
      test_virtual_functions:    4.029 (0.032)   377739 (  74.1)    4.039 (0.035) ( +0.25%)   377863 (  63.5) ( +0.03%)
```

The takeaway is that the CPU usage increase is basically indistinguishable from noise (19 increases, 11 decreases, not counting the added factory test); the compiler memory usage increases look slightly less like noise (26 increased, 4 decreased), but the impact is still quite small (only one exceeds 0.2% increase).

For clang 4.0.1 the results are similarly miniscule.  (CPU times increase more consistently positive than gcc, but it is still a very small increase, with pretty much everything under half a percent).
```
                                  clang-wo-factory.txt [n=50]                   clang-w-factory.txt [n=50]             
                                ================================   ====================================================
                                 CPU   (s.d. )   MaxRSS ( s.d. )    CPU   (s.d. ) (  +/-  )   MaxRSS ( s.d. ) (  +/-  )
                                --------------   ---------------   ------------------------   -------------------------
                   (linking):   37.108 (0.616)   250701 (6422.2)   36.970 (0.585) ( -0.37%)   247911 (5011.3) ( -1.11%)
              pybind11_tests:    2.492 (0.032)   149898 ( 373.8)    2.504 (0.028) ( +0.50%)   150165 ( 330.1) ( +0.18%)
                test_buffers:    2.704 (0.034)   159831 ( 247.3)    2.707 (0.031) ( +0.10%)   160006 ( 267.3) ( +0.11%)
        test_builtin_casters:    3.256 (0.033)   182829 ( 217.3)    3.259 (0.034) ( +0.10%)   182961 ( 202.6) ( +0.07%)
          test_call_policies:    2.161 (0.026)   136706 ( 203.1)    2.174 (0.032) ( +0.57%)   136876 ( 192.6) ( +0.12%)
              test_callbacks:    3.058 (0.031)   166637 ( 317.2)    3.068 (0.035) ( +0.31%)   166780 ( 331.1) ( +0.09%)
                 test_chrono:    1.910 (0.021)   138702 ( 221.2)    1.916 (0.025) ( +0.30%)   138812 ( 196.0) ( +0.08%)
                  test_class:    3.400 (0.029)   189369 ( 215.2)    3.397 (0.034) ( -0.10%)   189575 ( 161.1) ( +0.11%)
test_constants_and_functions:    2.433 (0.021)   149718 ( 254.9)    2.441 (0.030) ( +0.34%)   149864 ( 251.1) ( +0.10%)
              test_copy_move:    3.289 (0.032)   189856 ( 200.6)    3.290 (0.036) ( +0.05%)   190092 ( 170.7) ( +0.12%)
      test_docstring_options:    2.018 (0.025)   133543 ( 238.5)    2.022 (0.025) ( +0.16%)   133715 ( 189.7) ( +0.13%)
                  test_eigen:    9.780 (0.073)   412217 ( 248.5)    9.794 (0.085) ( +0.15%)   412457 ( 304.5) ( +0.06%)
                   test_enum:    3.646 (0.024)   192517 ( 198.4)    3.666 (0.046) ( +0.55%)   192678 ( 188.9) ( +0.08%)
                   test_eval:    1.758 (0.018)   124507 ( 214.8)    1.764 (0.023) ( +0.36%)   124641 ( 174.6) ( +0.11%)
             test_exceptions:    1.648 (0.020)   122190 ( 166.4)    1.653 (0.018) ( +0.32%)   122422 ( 195.0) ( +0.19%)
   test_factory_constructors:    5.396 (0.052)   235282 ( 228.5)    5.393 (0.052) ( -0.06%)   235291 ( 214.6) ( +0.00%)
    test_kwargs_and_defaults:    2.387 (0.024)   150717 ( 383.2)    2.397 (0.024) ( +0.39%)   150876 ( 405.2) ( +0.11%)
 test_methods_and_attributes:    4.864 (0.045)   228137 ( 241.3)    4.861 (0.050) ( -0.05%)   228315 ( 249.5) ( +0.08%)
                test_modules:    2.561 (0.025)   150216 ( 357.0)    2.565 (0.027) ( +0.16%)   150489 ( 398.9) ( +0.18%)
   test_multiple_inheritance:    4.973 (0.051)   259231 ( 359.4)    4.972 (0.039) ( -0.03%)   259470 ( 416.0) ( +0.09%)
            test_numpy_array:    4.740 (0.039)   220536 ( 277.9)    4.731 (0.053) ( -0.19%)   220680 ( 217.6) ( +0.07%)
           test_numpy_dtypes:    3.595 (0.039)   181304 ( 324.2)    3.604 (0.038) ( +0.26%)   181479 ( 250.1) ( +0.10%)
        test_numpy_vectorize:    3.561 (0.034)   188478 ( 183.4)    3.567 (0.040) ( +0.18%)   188645 ( 207.6) ( +0.09%)
           test_opaque_types:    2.493 (0.029)   153162 ( 477.6)    2.501 (0.031) ( +0.32%)   153496 ( 455.0) ( +0.22%)
   test_operator_overloading:    3.062 (0.044)   172689 ( 234.3)    3.068 (0.031) ( +0.18%)   172895 ( 192.2) ( +0.12%)
               test_pickling:    2.116 (0.030)   137702 ( 245.7)    2.123 (0.029) ( +0.30%)   137854 ( 206.4) ( +0.11%)
                test_pytypes:    2.725 (0.029)   143312 ( 259.7)    2.728 (0.028) ( +0.09%)   143504 ( 308.4) ( +0.13%)
test_sequences_and_iterators:    4.408 (0.052)   216225 ( 211.2)    4.406 (0.046) ( -0.04%)   216568 ( 191.3) ( +0.16%)
              test_smart_ptr:    4.387 (0.042)   209329 ( 205.3)    4.390 (0.044) ( +0.06%)   209564 ( 205.9) ( +0.11%)
                    test_stl:    3.873 (0.033)   202854 ( 204.3)    3.880 (0.035) ( +0.17%)   202967 ( 214.6) ( +0.06%)
            test_stl_binders:   11.174 (0.082)   439104 ( 215.3)   11.185 (0.093) ( +0.10%)   439334 ( 239.1) ( +0.05%)
      test_virtual_functions:    4.214 (0.045)   211784 ( 263.2)    4.220 (0.040) ( +0.15%)   211997 ( 301.1) ( +0.10%)
```

So a slight CPU impact (well under half a percent for almost everything), and a smaller memory impact than gcc.

---

If this seems acceptable to everyone, I'll go ahead with adding this as a core include from `pybind11.h`, updating the documentation to prefer this over direct `__init__` initializers (relegating `__init__` to an advanced feature), and update the test suite accordingly.


(I've also committed the tools that generated these with quick instructions in https://github.com/pybind/pybind11_compiler_bench)

--- 

For the curious, here's the comparison between gcc/clang:

```
                                   gcc-wo-factory.txt [n=50]                   clang-wo-factory.txt [n=50]             
                                ================================   ====================================================
                                 CPU   (s.d. )   MaxRSS ( s.d. )    CPU   (s.d. ) (  +/-  )   MaxRSS ( s.d. ) (  +/-  )
                                --------------   ---------------   ------------------------   -------------------------
                   (linking):   27.865 (0.119)   288764 (  82.8)   37.108 (0.616) (+33.17%)   250701 (6422.2) (-13.18%)
              pybind11_tests:    2.201 (0.029)   294996 ( 132.3)    2.492 (0.032) (+13.21%)   149898 ( 373.8) (-49.19%)
                test_buffers:    2.557 (0.025)   310068 (  79.9)    2.704 (0.034) ( +5.74%)   159831 ( 247.3) (-48.45%)
        test_builtin_casters:    2.969 (0.038)   340145 (  98.6)    3.256 (0.033) ( +9.67%)   182829 ( 217.3) (-46.25%)
          test_call_policies:    1.882 (0.031)   254697 (  43.4)    2.161 (0.026) (+14.81%)   136706 ( 203.1) (-46.33%)
              test_callbacks:    2.735 (0.033)   310796 (  50.5)    3.058 (0.031) (+11.83%)   166637 ( 317.2) (-46.38%)
                 test_chrono:    1.762 (0.027)   252543 ( 125.8)    1.910 (0.021) ( +8.41%)   138702 ( 221.2) (-45.08%)
                  test_class:    3.057 (0.035)   356680 (  54.0)    3.400 (0.029) (+11.21%)   189369 ( 215.2) (-46.91%)
test_constants_and_functions:    2.136 (0.028)   291964 ( 100.5)    2.433 (0.021) (+13.89%)   149718 ( 254.9) (-48.72%)
              test_copy_move:    3.064 (0.035)   342694 ( 134.9)    3.289 (0.032) ( +7.32%)   189856 ( 200.6) (-44.60%)
      test_docstring_options:    1.761 (0.030)   242729 ( 134.8)    2.018 (0.025) (+14.64%)   133543 ( 238.5) (-44.98%)
                  test_eigen:    9.299 (0.091)   728918 (  57.2)    9.780 (0.073) ( +5.16%)   412217 ( 248.5) (-43.45%)
                   test_enum:    3.316 (0.035)   354021 (  92.6)    3.646 (0.024) ( +9.95%)   192517 ( 198.4) (-45.62%)
                   test_eval:    1.524 (0.021)   215105 (  60.4)    1.758 (0.018) (+15.38%)   124507 ( 214.8) (-42.12%)
             test_exceptions:    1.459 (0.026)   208468 (  67.0)    1.648 (0.020) (+12.92%)   122190 ( 166.4) (-41.39%)
   test_factory_constructors:    4.581 (0.050)   418475 (  53.9)    5.396 (0.052) (+17.79%)   235282 ( 228.5) (-43.78%)
    test_kwargs_and_defaults:    2.121 (0.023)   287457 (  52.0)    2.387 (0.024) (+12.57%)   150717 ( 383.2) (-47.57%)
 test_methods_and_attributes:    4.541 (0.043)   390432 (  55.8)    4.864 (0.045) ( +7.10%)   228137 ( 241.3) (-41.57%)
                test_modules:    2.200 (0.024)   296003 (  51.6)    2.561 (0.025) (+16.41%)   150216 ( 357.0) (-49.25%)
   test_multiple_inheritance:    4.896 (0.051)   454484 (  44.4)    4.973 (0.051) ( +1.58%)   259231 ( 359.4) (-42.96%)
            test_numpy_array:    4.222 (0.044)   370925 (  56.9)    4.740 (0.039) (+12.26%)   220536 ( 277.9) (-40.54%)
           test_numpy_dtypes:    3.072 (0.031)   333738 ( 132.7)    3.595 (0.039) (+17.03%)   181304 ( 324.2) (-45.67%)
        test_numpy_vectorize:    3.168 (0.038)   342012 (  82.3)    3.561 (0.034) (+12.41%)   188478 ( 183.4) (-44.89%)
           test_opaque_types:    2.180 (0.028)   295067 (  84.8)    2.493 (0.029) (+14.40%)   153162 ( 477.6) (-48.09%)
   test_operator_overloading:    2.877 (0.030)   334324 ( 135.7)    3.062 (0.044) ( +6.43%)   172689 ( 234.3) (-48.35%)
               test_pickling:    1.853 (0.030)   258136 ( 108.8)    2.116 (0.030) (+14.20%)   137702 ( 245.7) (-46.66%)
                test_pytypes:    2.088 (0.029)   270000 (  46.6)    2.725 (0.029) (+30.54%)   143312 ( 259.7) (-46.92%)
test_sequences_and_iterators:    4.152 (0.037)   377033 (  62.2)    4.408 (0.052) ( +6.15%)   216225 ( 211.2) (-42.65%)
              test_smart_ptr:    4.205 (0.036)   378643 (  69.6)    4.387 (0.042) ( +4.33%)   209329 ( 205.3) (-44.72%)
                    test_stl:    3.546 (0.032)   366140 (  89.6)    3.873 (0.033) ( +9.24%)   202854 ( 204.3) (-44.60%)
            test_stl_binders:   11.039 (0.090)   806419 ( 145.5)   11.174 (0.082) ( +1.22%)   439104 ( 215.3) (-45.55%)
      test_virtual_functions:    4.029 (0.032)   377739 (  74.1)    4.214 (0.045) ( +4.58%)   211784 ( 263.2) (-43.93%)
```

So gcc compiles everything faster, but uses almost twice as much memory when doing so.  (Linking is actually considerably faster in *wallclock* seconds under clang's lld as it runs in parallel, while regular `ld` does not).

jagerman:
This would likely also provide a way to deal with #948 (binding a class that doesn't support placement new).

include/pybind11/factory_support.h
```diff
@@ -0,0 +1,280 @@
...
+// Makes sure the `value` for the given value_and_holder is preallocated; if it isn't, preallocate
+// it.  Returns the value pointer reference.
+inline void *&preallocate(value_and_holder &v_h) {
+    if (!v_h) v_h.value_ptr() = v_h.type->operator_new(v_h.type->type_size);
+    return v_h.value_ptr();
+}
+
+// Makes sure the `value` for the given value_and_holder is not allocated; if it is, deallocate it.
+// Returns the (null) value pointer reference.
+inline void *&deallocate(value_and_holder &v_h) {
```
jagerman: One possible case is when you have overloaded constructors when you have mixed placement new and init factory constructors, such as:

```
cl.def("__init__", [](Class &self, double i) { new (&self) Class(i); });
cl.def("__init__", py::init([](int j) { return new Class(j); });
```
when called from Python as `Class(3)` to go to the second overload.  Because the lazy allocation happens during argument loading, the `&self` will get preallocated even though the overload fails, and the other init factories need to deal with it.

I suppose we could deprecated placement new constructors and get rid of this eventually (i.e. in an eventual 3.0), in which case we wouldn't need to worry about this case.

Re: comments, that's a fair criticism; I'll edit them to be more "why" than "what".

include/pybind11/factory_support.h
```diff
@@ -0,0 +1,280 @@
...
+    // Pointer return v3: returning a cpp class for a registered class with an alias.  If we don't
+    // need an alias, we simply use the cpp pointer.  Otherwise, we try a dynamic_cast to see if the
+    // cpp pointer is actually a polymorphic alias instance.  If it isn't, but we can construct the
+    // alias via an `Alias(Cpp &&)` constructor, we do.  Otherwise we throw a type error.
+    template <typename Class, enable_if_t<Class::has_alias, int> = 0>
+    static void construct(value_and_holder &v_h, Cpp<Class> *ptr, PyTypeObject *cl_type) {
+        no_nullptr(ptr);
+        if (Py_TYPE(v_h.inst) != cl_type) {
+            // Inherited from on the Python side: an alias instance is required
+            if (!is_alias<Class>(ptr)) {
```
jagerman: It should have been, but may have been dropped accidentally somewhere along the line.  I'll investigate.

include/pybind11/factory_support.h
```diff
@@ -0,0 +1,280 @@
...
+        deallocate(v_h) = ptr;
+        v_h.type->init_instance(v_h.inst, &holder);
+    }
+
+    // return-by-value version 1: returning a cpp class by value when there is no alias
+    template <typename Class, enable_if_t<!Class::has_alias, int> = 0>
+    static void construct(value_and_holder &v_h, Cpp<Class> &&result, PyTypeObject *) {
+        static_assert(std::is_move_constructible<Cpp<Class>>::value,
+            "pybind11::init() return-by-value factory function requires a movable class");
+        new (preallocate(v_h)) Cpp<Class>(std::move(result));
```
jagerman: `preallocate` doesn't do anything if the value is already preallocated, so if you mix old-style placement-new `__init__`s with factory inits you get fewer allocation/deallocations.  If placement-new constructors are rare/deprecated, it seems a reasonable approach though.

(And on a side note, non-factory inits have a pretty big issue with the constructor being invoked multiple time if being used under multiple inheritance, so discouraging them is probably a good thing).

include/pybind11/pybind11.h
```diff
@@ -203,6 +199,8 @@ class cpp_function : public function {
                 a.descr = strdup(a.value.attr("__repr__")().cast<std::string>().c_str());
         }
 
+        rec->is_constructor = !strcmp(rec->name, "__init__") || !strcmp(rec->name, "__setstate__");
```
jagerman: It was part of an earlier commit that was since dropped to fix the constructor signature when the first argument was a `handle`, but that got replaced with `v_h_proxy<Cpp<Class>>`.  I think you want me to go back to that, though.

include/pybind11/factory_support.h
dean0x7d: Ah, that makes sense with the mixed constructors. Deprecating the placement new constructors sounds good to me.

jagerman:
I've been a bit swamped with other things for the last few days, but I'll try to get this PR updated tomorrow.

jagerman:
The last commit just pushed should address the outstanding comments.

There is still some growth in .so size (~19KB) from the `py::init`-via-factory implementation.  This is buying us a fairly important fix for Python-side multiple inheritance (where multiple `__init__`s will invoke multiple C++ constructions on the same object) that can't be easily avoided with the previous placement-new implementation, so I think it's worthwhile.  I tried to extract some other bits into common code, but couldn't get anything that made a noticeable difference in `.so` size.

If all looks good, there's one last cleanup I think we should do which is to rename `factory_support.h` to `init_support.h` and also move the `struct init` and `struct init_alias` into it from `pybind11.h`.  Neither is very large, but they seem logically well matched with the factory code (and takes a bit of `detail` code out of the main header).

include/pybind11/pybind11.h
jagerman: The constructor signature rewriting got reintroduced with removal of the proxy class, so I kept this moved.

jagerman:
I've renamed this to `detail/init.h` to match #1001.  I also moved the various static functions in the factory class into free templated functions in the namespace, then used them to implement a more direct path for the common standard `py::init<...>()` constructors, which further cuts down the .so size increase (and should also make regular constructors slightly faster--they save the extra function pointer dereference and call that using factory functions would incur).

jagerman:
(The change looks bigger than it actually is; most of the commit diff is just from moving the static functions in the factory class out, and moving the `detail::init`/`detail::init_alias` structs into `detail/init.h`).

include/pybind11/detail/init.h
```diff
@@ -0,0 +1,333 @@
...
+// inherit all the base class constructors.
+template <typename Class, enable_if_t<std::is_constructible<Alias<Class>, Cpp<Class> &&>::value, int> = 0>
+static bool construct_alias_from_cpp(value_and_holder &v_h, Cpp<Class> &&base) {
+    deallocate(v_h) = new Alias<Class>(std::move(base));
+    return true;
+}
+// Fallback of the above for aliases without an `Alias(Cpp &&) constructor; does nothing and
+// returns false to signal the non-availability of the special constructor.
+template <typename Class, enable_if_t<!std::is_constructible<Alias<Class>, Cpp<Class> &&>::value, int> = 0>
+static constexpr bool construct_alias_from_cpp(value_and_holder &, Cpp<Class> &&) { return false; }
```
dean0x7d: The control flow built around this feels a bit more complicated than it should be. This `return false` results in two different `throw type_error` messages below, but they have the same cause. Consider switching to `void` functions and place the `throw` statement in the `false` fallback. (I'm also using simple tag dispatch in the example below because it's a nice lightweight option, especially for binary overloads, and it also nicely maps to the future `if constexpr`.)

```c++
template <typename Class>
using is_alias_constructible = std::is_constructible<Alias<Class>, Cpp<Class> &&>;

template <typename Class>
void construct_alias_from_cpp(std::true_type /*is_alias_constructible*/,
                              value_and_holder &v_h, Cpp<Class> &&base) {
    deallocate(v_h) = new Alias<Class>(std::move(base));
}

template <typename Class>
[[noreturn]] void construct_alias_from_cpp(std::false_type /*is_alias_constructible*/,
                                           value_and_holder &, Cpp<Class> &&) {
    throw type_error("pybind11::init(): unable to convert returned instance to required "
                     "alias class: no `Alias<Class>(Class &&)` constructor available");
}
```

```c++
if (Class::has_alias && need_alias && !is_alias<Class>(ptr)) {
    ...
    construct_alias_from_cpp<Class>(is_alias_constructible<Class>{}, v_h, std::move(*ptr));
} else {
    deallocate(v_h) = ptr;
}
```

tests/test_factory_constructors.py
```diff
@@ -0,0 +1,451 @@
...
+        pass
+
+    # Attempts to initialize with an invalid type passed as `self`:
+    class BrokenTF1(m.TestFactory1):
+        def __init__(self, bad):
+            if bad == 1:
+                a = m.TestFactory2(tag.pointer, 1)
+                m.TestFactory1.__init__(a, tag.pointer)
+            elif bad == 2:
+                m.TestFactory1.__init__(NotPybindDerived.__new__(NotPybindDerived), tag.pointer)
```
jagerman: I was looking for something allocated but uninitialized.  Will it hurt anything?

tests/test_factory_constructors.py
```diff
@@ -0,0 +1,451 @@
...
+        NoisyAlloc(int 5)    # construction
+        ---
+        ~NoisyAlloc()  # Destructor
+        noisy delete   # operator delete
+    """)
+
+
+@pytest.unsupported_on_py2
+@pytest.mark.parametrize('which, bad', [(1, 1), (1, 2), (6, 1), (6, 2), (6, 3), (6, 4)])
+def test_invalid_self(which, bad, capture):
```
jagerman: That was to suppress the capture leak (which it does on my system, but apparently doesn't on one of the travis-ci builds).

I don't quite understand why that capture leak is happening at all, though.  Could the `mark.parametrized` be involved somehow?

jagerman: Deparameterizing it worked around the issue.

tests/test_factory_constructors.py
jagerman: Updated to just pass an ordinary non-pybind instance (essentially matching to previous case, except for the non-pybind aspect).

jagerman:
Fixed up everything and squashed the commits down; I think this is ready to go (once the builds finish).

tests/test_factory_constructors.py
dean0x7d: It doesn't hurt anything I just thought it looked like unusual syntax for the functionally equivalent `NotPybindDerived()`. In general, for a native Python `class A`, the constructor `A() = A.__new__() + A.__init__()`. If `A` doesn't have `__init__()` then `A() == A.__new__()`, which is the case for `NotPybindDerived` in the test.

jagerman:
Merged!

include/pybind11/detail/init.h
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// Implementing class for py::init<...>()
+template <typename... Args> struct constructor {
+    template <typename Class, typename... Extra, enable_if_t<!Class::has_alias, int> = 0>
+    static void execute(Class &cl, const Extra&... extra) {
+        auto *cl_type = get_type_info(typeid(Cpp<Class>));
+        cl.def("__init__", [cl_type](handle self_, Args... args) {
+            auto v_h = load_v_h(self_, cl_type);
+            if (v_h.instance_registered()) return; // Ignore duplicate __init__ calls (see above)
```
wjakob: I stumbled across this comment (duplicate __init__ calls) while looking at ``init.h`` and was a bit puzzled. I wasn't able to find what the "see above" comment refers to. Would you mind clarifying this?

wjakob: ^ @jagerman

jagerman: Ah, the code got shuffled around; the comment it's referring to is at lines 258/259 now.  I'll fix it up on `master`.

jagerman: (They used to be right next to each other).

jagerman: Fixed in 9f6a636e.

dean0x7d: Ah, just got a conflict with those changes :) Those checks are gone now in #1014.

jagerman: That was pretty incredibly bad timing :)

PR #806, opened 2017-04-17
WIP: cast: fix holder construction for std::enable_shared_from_this derived types

### LINKED ISSUE #801: Incorrect objects construction with multiple inheritance
I've decided to open a new issue, following already closed #797, because I was able to narrow down the error case and I think it's related to multiple inheritance and/or incorrect usage of trampolines, which I cannot detect.

Here's modified test case from #797:
```C++
#include <iostream>

namespace py = pybind11;

// classes
class objbase : public std::enable_shared_from_this< objbase > {
	int prop_ = 42;
public:
	virtual std::string name() const {
		return "objbase";
	}

	int prop() const { return prop_; }
};

class iface {
public:
	int n = 42;
	virtual int f() const = 0;
};

class mytype : public iface, public objbase {
public:
	std::string name() const override {
		return "mytype";
	}
	int f() const override {
		return 42;
	}
};
using sp_my = std::shared_ptr<mytype>;
using sp_cmy = std::shared_ptr<const mytype>;

// trampoline
template<class T = objbase>
class py_objbase : public T {
public:
	using T::T;

	std::string name() const override {
		PYBIND11_OVERLOAD(std::string, T, name);
	}
};

class py_objbase_ : public py_objbase<> {
	double pyprop_ = 42.84;

public:
	using py_objbase<>::py_objbase;

	double pyprop() const { return pyprop_; }
};

class py_iface : public iface {
public:
	using iface::iface;

	int f() const override {
		PYBIND11_OVERLOAD_PURE(int, iface, f);
	}
};

// bind mytype
class py_mytype : public py_iface, public py_objbase<mytype> {
public:
	using py_objbase<mytype>::py_objbase;

	int f() const override {
		PYBIND11_OVERLOAD(int, mytype, f);
	}
};

PYBIND11_PLUGIN(example) {
	py::module m("example");

	// bind objbase
	py::class_<objbase, py_objbase_, std::shared_ptr<objbase>>(m, "objbase", py::multiple_inheritance())
		.def(py::init_alias<>())
		.def("name", &objbase::name)
		.def_property_readonly("prop", &objbase::prop)
		.def_property_readonly("refs", [](const objbase& src) { return src.shared_from_this().use_count() - 1; })
		.def_property_readonly("pyprop", [](const objbase& src) {
			return static_cast<const py_objbase_*>(&src)->pyprop();
		})
	;

	// iface
	py::class_<iface, py_iface, std::shared_ptr<iface>>(m, "iface", py::multiple_inheritance())
		.def("f", &iface::f)
	;

	// bind mytype
	py::class_<mytype, iface, objbase, py_mytype, sp_my>(m, "mytype", py::multiple_inheritance())
		.def(py::init<>())
		.def("name", &mytype::name)
		.def("f", &mytype::f)
		.def_property_readonly("n", [](mytype& src){ return src.n; })
	;

	return m.ptr();
}
```
The main difference from #797 is that `objbase` now has `py_objbase_` trampoline instead of `py_objbase<>`. And `py_objbase_`, in it's turn, contains specific variable `pyprop`.

Now, if `objbase` is constructed directly, everything is fine:
```Python
In [1]: import example as e

In [2]: a = e.objbase()

In [3]: a.pyprop
Out[3]: 42.84
```
But `mytype` is constructed incorrectly:
```Python
In [1]: import example as e

In [2]: a = e.mytype()

In [3]: a.pyprop
Out[3]: 4.65557599738463e-310
```
More that that, in this concrete example if I change `mytype`'s constructor from `.def(py::init<>())` to `.def(py::init_alias<>())`, then Python crashes with segfault on `a = e.mytype()` line.

As you can see, I've added `multiple_inheritance()` tag to all exported types, but it makes no difference at all, because it makes sense only for `mytype` which already has multiple bases specified as `class_` template arguments.

Testes on clang 3.9.1, Python 2.7.13.

uentity:
Properly cast `instance_type::value` before calling
`shared_from_this()` inside corresponding `class_::init_holder_helper`.

Previous implementation throwed `bad_weak_ptr` when initializing holder
in `type_caster_generic::cast` and source object being casted is
held by `std::shared_ptr`. In that case, if policy is set to
`take_ownership`, `init_holder_helper` creates new shared ptr and
blindly takes ownership of passed object already managed by source shared
ptr. This in turn leads to premature object destruction and segfault.

This partially fixes #801.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1100,8 +1100,26 @@ private:
     template <typename T>
     static void init_holder_helper(instance_type *inst, const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
         try {
-            new (&inst->holder) holder_type(std::static_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this()));
+            new (&inst->holder) holder_type(std::dynamic_pointer_cast<typename holder_type::element_type>(
+                dynamic_cast<std::enable_shared_from_this<T>*>(
+                    static_cast<T*>((void*)inst->value))->shared_from_this()
+            ));
             inst->holder_constructed = true;
+            if(!inst->holder.get()) {
+                // create dumb holder for the fail case
+                // otherwise, if holder_constructed is false, cleanup code will try to `delete inst->value`
+                // that would probably lead to double free and segfault
+                // (because instance is already managed by shared_ptr)
+                new (&inst->holder) holder_type();
+                auto &internals = detail::get_internals();
+                internals.registered_instances.emplace((void*)inst->value, (void*)inst);
+                // ... and throw exception
+                std::string tname = typeid(*inst->value).name();
+                detail::clean_type_id(tname);
+                throw type_error(std::string("Unable to create holder for type : ") + tname);
+            }
+            // correct pointer to actual type instance
+            inst->value = dynamic_cast<type*>(inst->holder.get());
         } catch (const std::bad_weak_ptr &) {
             if (inst->owned) {
                 new (&inst->holder) holder_type(inst->value);
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -963,7 +963,7 @@ public:
 
     bool load_value_and_holder(handle src) {
         auto inst = (instance<type, holder_type> *) src.ptr();
-        value = (void *) inst->value;
+        value = const_cast<void *>((const void*) inst->value);
         if (inst->holder_constructed) {
             holder = inst->holder;
             return true;
--- a/tests/pybind11_tests.cpp
+++ b/tests/pybind11_tests.cpp
@@ -41,7 +41,15 @@ void bind_ConstructorStats(py::module &m) {
         .def_readwrite("move_assignments", &ConstructorStats::move_assignments)
         .def_readwrite("copy_constructions", &ConstructorStats::copy_constructions)
         .def_readwrite("move_constructions", &ConstructorStats::move_constructions)
-        .def_static("get", (ConstructorStats &(*)(py::object)) &ConstructorStats::get, py::return_value_policy::reference_internal);
+        .def_static("get", (ConstructorStats &(*)(py::object)) &ConstructorStats::get, py::return_value_policy::reference_internal)
+
+        // Not exactly ConstructorStats, but related: expose the internal pybind number of registered instances
+        // to allow instance cleanup checks (invokes a GC first)
+        .def_static("detail_reg_inst", []() {
+            ConstructorStats::gc();
+            return py::detail::get_internals().registered_instances.size();
+        })
+        ;
 }
 
 PYBIND11_PLUGIN(pybind11_tests) {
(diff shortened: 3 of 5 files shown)


jagerman:
Can you add a simple-as-possible test case to the test suite for this (i.e. which currently throws a `bad_weak_ptr` and is fixed by this PR)?

I assume the basic issue is that the `enable_shared_from_this<T>` comes via inheritance from `T`, where `T` is some base class of `type`.

jagerman:
(`test_inheritance.cpp`/`.py` seems like a good place for an added test)

include/pybind11/pybind11.h
```diff
@@ -1100,7 +1100,9 @@ class class_ : public detail::generic_type {
     template <typename T>
     static void init_holder_helper(instance_type *inst, const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
         try {
-            new (&inst->holder) holder_type(std::static_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this()));
+            new (&inst->holder) holder_type(std::static_pointer_cast<typename holder_type::element_type>(
+                static_cast<T*>((void*)inst->value)->shared_from_this()
```
dean0x7d: If `T` is a base, a single `static_cast` should be enough -- no need to go through `void*` (which is essentially reinterpret casting).

Alternatively, the casting can be avoided altogether by using the currently commented out parameter:
```c++
// existing unnamed parameter
const std::enable_shared_from_this<T> * /* dummy */

// -> name it
const std::enable_shared_from_this<T> *esft

// -> use it
esft->shared_from_this()
```

And as @jagerman has already mentioned, a minimal test case would be very good here.

uentity: Nice point. Will do shortly.

uentity:
While developing tests for casting problem, I shed more light on it. But I've got stuck and need some help, please take a look at the code.

1) The real trigger for this issue is *multiple inheritance*. If you remove `mytype` inheritance from `iface` then execution goes on another path -- registered instance of `mytype` is found for slot argument and `type_caster_generic::cast` just adds one reference to it instead of constructing new type holder.
2) If you invoke `break_mytype_cast()` before calling slot, then "old" `init_holder_helper` code is activated for `mytype` and it will throw `std::bad_weak_ptr`. But what is the correct way to process this exception? I wasn't able to somehow just abort calling function and print error message in Python. All I do it leads to segfault (my last attempt is in `invalid_mytype_cast::init_holder_helper`).
3) Even if `std::shared_ptr<objbase>` is constructed correctly (see fix from first commit) calling slot would  segfault. I don't know why.
4) You may argue that *both* `objbase` and `iface` should inherit virtually from `std::enable_shared_from_this`. But I did that and behaviour is unchanged -- segfault in all cases.
5) As we deal with multiple inheritance -- should we use *dynamic* casts instead of static in `init_holder_helper`? I tried this, shared pointer is constructed correctly (no `bad_weak_ptr` exception), but calling slot = segfault.

So now all code paths calling `pyslot.act` in this test lead to segfault.

jagerman:
I have not had time to look at this in detail, but if `T` is a derived class (i.e. we're doing a base to derived downcast) then we really should be using `std::dynamic_pointer_cast` (with failure detection) there instead of `static_pointer_cast`.  Try changing it (as you did) plus adding a `throw type_error` if, after the call, the resulting shared_ptr is empty.

jagerman:
Is this PR still necessary with #812, or does #812 address the issues this PR was solving?

uentity:
This PR should be killed. Your PR #812 resolves everything.

PR #808, opened 2017-04-19
Make slice::compute return signed integers

TallJimbo:
Signed results are necessary for at least the step, which may be negative.

If unsigned results are needed for the others, this should be done with a `static_cast`; the result of a `reintepret_cast` between signed and unsigned is compiler-defined.  I think it's better to just pass the ssize_t values directly to the user and let them handle a conversion to unsigned if needed.


jagerman:
This seems a good fit with #782.

dean0x7d:
I agree that this would work well alongside #782. 

@TallJimbo Looks like there are addition conversion issues that would need to be cleaned up here.

wjakob:
Superseded by #1394

Peterisfar:
How to express this "a[0:3, 0:3]" in python as c ++?

PR #811, opened 2017-04-21
Add type caster for std::variant and other variant-like classes

### LINKED ISSUE #810: variant type
Can you give me advice how handle variant types(std::variant from c++17, mpark/variant for c++14, boost::variant and etc ) with pybind11?

dean0x7d:
C++17 `std::variant` is supported automatically, but custom variants can also be added easily. See the explanation in the docs.

For conforming C++11/14 `std::variant` implementations (like `mpark::variant`) a `variant_caster` specialization is enough. For implementations with a different `visit()` function (like `boost::variant`) the `visit_helper` is also required.

Resolves #810.

### DIFF
--- a/docs/advanced/cast/overview.rst
+++ b/docs/advanced/cast/overview.rst
@@ -144,6 +144,8 @@ as arguments and return values, refer to the section on binding :ref:`classes`.
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::experimental::optional<T>`` | STL optional type (exp.)  | :file:`pybind11/stl.h`        |
 +------------------------------------+---------------------------+-------------------------------+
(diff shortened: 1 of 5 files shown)


jagerman:
This looks good to me.

I suggest we remove the clang C++17 build from failure-allowed status, given that clang 4 is released.  Moving gcc-7 should probably wait for a while (i.e. until it's released rather than built from gcc snapshots in debian experimental).

wjakob:
Nice and short -- this looks great! I look forward to using this in my projects once compiler support is a bit more widespread :)

@jagerman: sounds good to me.

jagerman:
Done (via 2761f78).

PR #812, opened 2017-04-21
Multiple inheritance fixes

### LINKED ISSUE #801: Incorrect objects construction with multiple inheritance

jagerman:
This fixes two issues with multiple inheritance.

The first is that when we cast a base class pointer, we correctly detect (via a `typeid`) the derived class, but then do (in essense) a reinterpret_cast on it, which is wrong when the base class is offset (due to multiple inheritance).  The fix is to dynamic_cast to the most derived type before passing the pointer on.

The second fix is that we can often end up with multiple Python instances around the same C++ instance when we are given a base class pointer in the same situation as above: the pointer doesn't match anything in `registered_instances`, so we treat it as a new type, even though, internally, it is really wrapping the same type.  The commit to fix this adds into `registered_instances` all the base type pointers (for base types that we know about) that aren't equal to the derived pointer.  This is a bit of extra work as we have to do it recursively, and both during registering the instance and deregistering, but this is only done for types that actually have multiple bases somewhere above them in their inheritance tree.

Fixes #801.

@uentity - I think this should fix at least some of your multiple inheritance issues; your feedback here would be very valuable.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -266,7 +266,7 @@ struct type_record {
             dynamic_attr = true;
 
         if (caster)
-            base_info->implicit_casts.push_back(std::make_pair(type, caster));
+            base_info->implicit_casts.emplace_back(type, caster);
     }
 };
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -24,6 +24,7 @@ inline PyTypeObject *make_default_metaclass();
 /// Additional type information which does not fit into the PyTypeObject
 struct type_info {
     PyTypeObject *type;
+    const std::type_info *cpptype;
     size_t type_size;
     void *(*operator_new)(size_t);
     void (*init_holder)(PyObject *, const void *);
@@ -35,9 +36,11 @@ struct type_info {
     void *get_buffer_data = nullptr;
     /** A simple type never occurs as a (direct or indirect) parent
      * of a class that makes use of multiple inheritance */
-    bool simple_type = true;
+    bool simple_type : 1;
+    /* True if there is no multiple inheritance in this type's inheritance tree */
+    bool simple_ancestors : 1;
     /* for base vs derived holder_type checks */
-    bool default_holder = true;
+    bool default_holder : 1;
 };
 
 PYBIND11_NOINLINE inline internals &get_internals() {
@@ -197,7 +200,7 @@ inline PyThreadState *get_thread_state_unchecked() {
 
 // Forward declarations
 inline void keep_alive_impl(handle nurse, handle patient);
-inline void register_instance(void *self);
+inline void register_instance(void *self, const type_info *tinfo);
 inline PyObject *make_new_instance(PyTypeObject *type, bool allocate_value = true);
 
 class type_caster_generic {
@@ -338,7 +341,7 @@ public:
                 throw cast_error("unhandled return_value_policy: should not happen!");
         }
 
-        register_instance(wrapper);
+        register_instance(wrapper, tinfo);
         tinfo->init_holder(inst.ptr(), existing_holder);
 
         return inst.release();
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -195,8 +195,10 @@ inline PyThreadState *get_thread_state_unchecked() {
 #endif
 }
 
-// Forward declaration
+// Forward declarations
 inline void keep_alive_impl(handle nurse, handle patient);
+inline void register_instance(void *self);
+inline PyObject *make_new_instance(PyTypeObject *type, bool allocate_value = true);
 
 class type_caster_generic {
 public:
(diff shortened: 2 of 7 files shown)


uentity:
@jagerman I thought I have found a solution by playing with casts inside `init_holder_helper()` (you can check it in my updated commit  for PR #806). But then this PR appeared and I tried to run your tests and they failed.

I think your proposal is versatile and address the problem in its toots. Tomorrow I'll check it in my real-world project, but it has to work.

jagerman:
I found another problem here, when returning an derived instance of an unregistered type as an offset base class pointer to a registered type.  We would cast to most-derived, pass that to `cast(...)`, which would then recognize that the type is unregistered and so cast it to the backup type.  But that isn't valid: by the time we get to `cast()` there's no way to get back to the base class pointer.

The update moves the type decision logic out of cast entirely: we should dynamic cast if the type is polymorphic *and* a registered type; otherwise we don't dynamic_cast at all, keeping the `itype` pointer as-is, and use `itype` as the type.  (The updated commit also adds another a test that triggers this case).

uentity:
All my code works even without your latest update, and update didn't break it either. My issue #801 is solved.

jagerman:
The latest update is a bit of an obscure case: you'd have to have something like:
```
 B   C
  \ /
   A
```
where A is *not* pybind11-registered, but `C` is, and you return a `C` pointer to an `A` instance.  (Whether or not `B` is pybind11-registered doesn't matter; if it is, nothing would have broken when returning a `B *` because it happens to be valid in that case to reinterpret_cast an `A *` back to `B *`).

uentity:
It's great that pybind11 finally correctly manages even such a corner cases with multiple inheritance. Thumbs up!

PR #815, opened 2017-04-24
functional: support bound methods

### LINKED ISSUE #814: Bound methods not working correctly as callbacks
#### Issue description

When using bound methods of instantiated python objects as std::function callbacks, the "self" parameter is incorrectly set to the first parameter instead of the bound object. Causing exceptions and incorrect behavior when using callbacks with member bound objects.

I'm using:
python 3.4.2
pybind11 git hash deb85f46efcba5fea624ee89daea8144abd77c54
gcc: arm-linux-gnueabihf-gcc ( 4.9.2-10) 4.9.2

It seems like bound functions are getting the same handling as normal functions, and thus do not properly capture "self". In pure python code this happens automatically, I do not know enough about the python C API to guess where this could be going wrong.

C++ code from example:
```

int func_arg(const std::function<int(int)> &f) {
    return f(10);
}

PYBIND11_PLUGIN(example) {
    py::module m("example", "pybind11 example plugin");
    m.def("func_arg", &func_arg);
    return m.ptr();
}
```

Python code:
```
import example
class Test:
  def boundMemberFunction(self, arg=-1): #default is needed, as pybind11 incorrectly does not fill the argument
    print(self, arg)
    return 1

t = Test()
example.func_arg(t.boundMemberFunction)
```

The boundMemberFunction incorrectly prints "10 -1", while I would expect it to print "<__main__.Test object at 0xa6c98070> 10"

jagerman:
If a bound std::function is invoked with a bound method, the implicit bound self is lost; the method is called with the first argument as self.  This commits adds bound method detection and re-adds the
implicit `self` into the call arguments for bound methods.

Fixes #814.

### DIFF
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -22,13 +22,14 @@ struct type_caster<std::function<Return(Args...)>> {
     using function_type = Return (*) (Args...);
 
 public:
-    bool load(handle src_, bool) {
-        if (src_.is_none())
+    bool load(handle src, bool) {
+        if (src.is_none())
             return true;
 
-        src_ = detail::get_function(src_);
-        if (!src_ || !PyCallable_Check(src_.ptr()))
+        if (!isinstance<function>(src))
             return false;
+
+        auto func = reinterpret_borrow<function>(src);
 
         /*
            When passing a C++ function as an argument to another C++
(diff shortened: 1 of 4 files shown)


wjakob:
Interesting -- this case had never occurred to me. Would it be easier to simply not unbox the method in ``detail::get_function``?

jagerman:
It looks like other code expects unboxing: removing it causes various other tests to fail with maximum recursion errors.

jagerman:
I cleaned it up a bit to go through `py::function`, which doesn't unbox, but supports explicit unboxing (for the special really-a-C++-function bypass case).

wjakob:
LGTM!

dean0x7d:
Looks good to me!

This also reminded me that `else if` is just `else` + `if`. I had a mental compiler error for a second :)

PR #624, opened 2017-01-30
Unicode fixes and docs

jagerman:
As discussed in #591, we currently don't fail gracefully when encountering invalid unicode.  This PR propagates the unicode error, while also adding support for `std::u16string`, `std::u32string` and the associated `char16_t`/`char32_t` types.

PR #817, opened 2017-04-26
Fix Python 3 `bytes` conversion to std::string/char*

### LINKED ISSUE #816: Conversion from bytes() to std::string
#### Issue description

Conversion from bytes() from python to std::string should be automatic according to the documentation at:
http://pybind11.readthedocs.io/en/master/advanced/cast/strings.html#passing-bytes-to-c

However, when I actually have a function that accepts std::string, passing a bytes object causes a TypeError.

python 3.4.2
pybind11 git hash deb85f4
gcc: arm-linux-gnueabihf-gcc ( 4.9.2-10) 4.9.2

With py::bytes also being available, I'm not sure if this is a bug in the code or an error in the documentation.

```

namespace py = pybind11;

void test(const std::string& s) {
}

PYBIND11_PLUGIN(example) {
    py::module m("example", "pybind11 example plugin");
    m.def("test", &test, "");
    return m.ptr();
}
```

```
import example
example.test('x') # works
example.test(b'x') # TypeError
```

jagerman:
The Unicode support added in 2.1 (PR #624) inadvertently broke accepting `bytes` as std::string/char* arguments.  This restores it with a separate path that does a plain conversion (i.e. completely bypassing all the encoding/decoding code), but only for single-byte string types.

Fixes #816.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -655,9 +655,9 @@ struct type_caster<std::basic_string<CharT, Traits, Allocator>, enable_if_t<is_s
             return false;
         } else if (!PyUnicode_Check(load_src.ptr())) {
 #if PY_MAJOR_VERSION >= 3
-            return false;
-            // The below is a guaranteed failure in Python 3 when PyUnicode_Check returns false
+            return load_bytes(load_src);
 #else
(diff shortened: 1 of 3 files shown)


wjakob:
LGTM

PR #819, opened 2017-04-27
Add numpy version check

### LINKED ISSUE #818: Regression in v2.1.1 returning Eigen Map Type
#### Issue description

A simple test case that returns an Eigen::Map works fine with v2.0.0 but now segfaults with v2.1.1.  The issue appears to have been introduced with a change on 16 Jan 2017: https://github.com/pybind/pybind11/commit/f86dddf7ba45810c97dc2cf8e8cb6ee1e8477307

This is with python 2.7.2 and gcc 4.9.2

```
// The following test case should run fine on v2.0.0 but segfault on v2.1.1
#include "Eigen/Dense"
#include "pybind11/eigen.h"

namespace py = pybind11;

static Eigen::Map<Eigen::Matrix3i> return_eigen_map() {
  static int array[9];
  for (int ii = 0; ii < 9; ++ii) {
    array[ii] = ii;
  }
  return Eigen::Map<Eigen::Matrix3i>(array);
}

PYBIND11_PLUGIN(test_eigen_map) {
  py::module mm("test_eigen_map", "Eigen Map Tester");
  mm.def("return_eigen_map", &return_eigen_map)
  return mm.ptr();
}
```

jagerman:
The numpy API constants can check past the end of the API array if the numpy version is too old thus causing a segfault.  The current list of functions requires numpy >= 1.7.0, so this adds a check and exception if numpy is too old.

The added feature version API element was added in numpy 1.4.0, so this could still segfault if loaded in 1.3.0 or earlier, but given that 1.4.0 was released at the end of 2009, it seems reasonable enough to not worry about that case.  (1.7.0 was released in early 2013).

Fixes #818

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -141,6 +141,7 @@ struct npy_api {
         return (bool) PyObject_TypeCheck(obj, PyArrayDescr_Type_);
     }
 
+    unsigned int (*PyArray_GetNDArrayCFeatureVersion_)();
     PyObject *(*PyArray_DescrFromType_)(int);
     PyObject *(*PyArray_NewFromDescr_)
         (PyTypeObject *, PyObject *, int, Py_intptr_t *,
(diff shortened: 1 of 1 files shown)


wjakob:
Looks good!

PR #821, opened 2017-04-28
enum_: fix implicit conversion on Python 2.7

wjakob:
Enumerations on Python 2.7 were not always implicitly converted to
integers (depending on the target size). This patch adds a `__long__`
conversion function (only enabled on 2.7) which fixes this issue.

The attached test case fails without this patch.


wjakob:
Interesting -- looks like PyPy does not pass the new test even with the fix.

wjakob:
The solution to the puzzle is that PyPy generates a type error (rather than a 2.7.x-style system error) when a conversion to an arithmetic type fails. I added a patch that switches PyPy over to the 3.x convention. With that, the test suite now passes on PyPy as well.

Any thoughts on this patch?

dean0x7d:
LGTM

PR #822, opened 2017-04-28
Fix ambiguous initialize_list arguments

jagerman:
This removes the convert-from-arithemtic-scalar constructor of `any_container` added in #788 as it can result in ambiguous calls, as in:

```C++
    py::array_t<float>({ 1, 2 })
```

which could be intepreted as either of:

```C++
    py::array_t<float>(py::array_t<float>(1, 2))
    py::array_t<float>(py::detail::any_container({ 1, 2 }))
```

Removing the convert-from-arithmetic constructor reduces the number of implicit conversions, avoiding the ambiguity for array and array_t.  This also re-adds the array/array_t constructors taking a scalar argument for backwards compatibility.

### DIFF
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -36,7 +36,7 @@ struct buffer_info {
     }
 
     buffer_info(void *ptr, size_t itemsize, const std::string &format, size_t size)
-    : buffer_info(ptr, itemsize, format, 1, size, itemsize) { }
+    : buffer_info(ptr, itemsize, format, 1, {size}, {itemsize}) { }
 
     explicit buffer_info(Py_buffer *view, bool ownview = true)
     : buffer_info(view->buf, (size_t) view->itemsize, view->format, (size_t) view->ndim,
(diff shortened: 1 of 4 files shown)


jagerman:
@wjakob - I think this should fix the ambiguity issue; can you give it a try with your various (unmodified) projects to confirm?

(A related idea: it might be interesting to have some failure-allowed travis-ci builds that install pybind (without compiling the test suite since that's already done in the other builds), then clones a few open source pybind-using projects and tries to build and/or test them.  It could help catching some backwards compatibility issues like this.)

wjakob:
Thanks! I confirm that this PR fixes the issue.

Regarding your suggestion: I think that our test suite is actually pretty good; it's gotten increasingly rare that something slips through. Downsides of testing further projects are that this would likely consume quite a few extra cycles -- not so nice given that we're on the free tier of Travis CI. My impression is that the previous approach of leaving enough stabilization time to catch rare regressions before releases suffices, though we can certainly discuss this further.

jagerman:
That's a good point, and given that we almost never remove things from the test suite, the coverage is probably pretty good.

PR #825, opened 2017-04-29
Deprecate handle::operator== in favor of object_api::is

dean0x7d:
Working with `py::object`s, it really seems confusing that `obj1 == obj2` does pointer comparison instead of actual object equality. ~~Explicitly writing `obj1.ptr() == obj2.ptr()` is easy enough and much clearer.~~ Edit: Switched to `obj1.is(obj2)` as suggested.

I propose deprecating `handle::opeator==` now so that the syntax can be freed up for Python's rich comparison methods at some point in the future. The changes to pybind11 are minimal since the operator is actually used very sparsely.

This PR also adds messages for deprecation warnings to all compilers (the `= default` to `{}` constructor change is due to a VS2015 bug).

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -58,7 +58,7 @@ struct metaclass {
     handle value;
 
     PYBIND11_DEPRECATED("py::metaclass() is no longer required. It's turned on by default now.")
-    metaclass() = default;
+    metaclass() {}
 
     /// Override pybind11's default metaclass
     explicit metaclass(handle value) : value(value) { }
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -69,14 +69,10 @@
 #  define PYBIND11_NOINLINE __attribute__ ((noinline))
 #endif
 
-#if defined(PYBIND11_CPP14)
+#if defined(PYBIND11_CPP14) || defined(_MSC_VER)
 #  define PYBIND11_DEPRECATED(reason) [[deprecated(reason)]]
-#elif defined(__clang__)
+#else
 #  define PYBIND11_DEPRECATED(reason) __attribute__((deprecated(reason)))
-#elif defined(__GNUG__)
-#  define PYBIND11_DEPRECATED(reason) __attribute__((deprecated))
-#elif defined(_MSC_VER)
-#  define PYBIND11_DEPRECATED(reason) __declspec(deprecated)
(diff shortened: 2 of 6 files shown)


aldanor:
Good idea. In Python, there's an `is` keyword that does pointer comparison, would `py::is()` be too much?

jagerman:
Deprecating it seems fine to me, because the confusion seems genuine, but I think that repurposing it in the future to map to Python equality would be similarly confusing.  `py::object` is, in many respects, a holder and C++ holders like `std::shared_ptr` typically do map `==` to pointer equality, not dereferenced value equality.  I think to avoid the confusion going the other way, it might be better to use a named method (e.g. `obj.equals(other)`) for that.  And then perhaps `obj.is(other)` for pointer equality.

jagerman:
Actually, looking at the diff, replacing the various `if (obj1.ptr() != obj2.ptr())` with `if (!obj1.is(obj2))` seems like a nice syntax clarity cleanup.

dean0x7d:
Appveyor is failing randomly again: #792

Regarding `is`: the `obj1.is(obj2)` syntax looks pretty good to me. Resembles Python a bit closer than `is(obj1, obj2)`.

However `obj.equals(other)` just screams Java to me. I think the syntax parallel with `std::shared_ptr` went out the window the moment `operator*` became unpacking instead of dereferencing (there's also `operator()` and `operator[]` which also break typical smart pointer syntax). Users don't see the underlying implementation of `PyObject` at all. The syntax is already so close to Python that proper object comparison is definitely the expected behavior of `operator==`.

jagerman:
I suppose so.  Also with `.is` available and used (rather than direct `.ptr()` comparisons) the distinction becomes a bit more obvious.

aldanor:
Agreed, `.is()` looks nicer.

jagerman:
Re: `PYBIND11_DEPRECATED`, does the Intel compiler support `[[deprecated(msg)]]`?  Aside from that, this PR looks good to me.

dean0x7d:
> does the Intel compiler support [[deprecated(msg)]] ?

Looks like it does since [v15 on Linux/macOS and v16 on Windows](https://software.intel.com/en-us/articles/c14-features-supported-by-intel-c-compiler). v15 doesn't matter for pybind11 on Windows since only v16+ support VS2015.

PR #836, opened 2017-05-07
buffer_info.h is missing from setup.py

yungyuc:
It seems everyone is using cmake, but unfortunately I am still using distutils ... :)  Without this line the pybind11 package installed by `pip install -U https://github.com/pybind/pybind11/archive/master.zip` doesn't work with distutils.


jagerman:
Thanks for this.  I merged it outside the PR (2b941b3) along with a comment in CMakeLists.txt to hopefully reduce the likelihood of doing it again.

PR #837, opened 2017-05-07
Do not derive enum_ from class_

aldanor:
(This is related to work on Python 3 enums in #781)

- Using `class_` API now requires calling `.into_class()` or doing a rvalue cast
- `.value()` and `.export_values()` now return rvalue `*this` instead of lvalue
- As a result of changing this, no existing tests were broken
- Added a few additional tests; updated the docs
 
Note: this is technically a breaking change, but the consensus seems to be that this functionality was more of a side-effect (of deriving from `class_`) rather than an intended feature.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1173,6 +1173,9 @@ template <typename Type> class enum_ {
 public:
     using Scalar = typename std::underlying_type<Type>::type;
 
+    enum_(const enum_&) = delete;
+    enum_(enum_&&) = default;
+
     template <typename... Extra>
     enum_(const handle &scope, const char *name, const Extra&... extra)
       : cls(scope, name, extra...), m_entries(), m_parent(scope) {
@@ -1242,18 +1245,28 @@ public:
     }
 
     /// Export enumeration entries into the parent scope
-    enum_& export_values() {
+    enum_& export_values() & {
         for (const auto &kv : m_entries)
             m_parent.attr(kv.first) = kv.second;
         return *this;
     }
 
+    enum_ export_values() && {
+        this->export_values();
+        return std::move(*this);
+    }
+
     /// Add an enumeration entry
-    enum_& value(char const* name, Type value) {
+    enum_& value(char const* name, Type value) & {
         auto v = pybind11::cast(value, return_value_policy::copy);
         cls.attr(name) = v;
         m_entries[pybind11::str(name)] = v;
         return *this;
+    }
+
+    enum_ value(char const* name, Type value) && {
+        this->value(name, value);
+        return std::move(*this);
     }
 
     /// Get the associated class object.
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1169,72 +1169,76 @@ private:
 };
 
 /// Binds C++ enumerations and enumeration classes to Python
-template <typename Type> class enum_ : public class_<Type> {
+template <typename Type> class enum_ {
 public:
-    using class_<Type>::def;
-    using class_<Type>::def_property_readonly_static;
     using Scalar = typename std::underlying_type<Type>::type;
 
     template <typename... Extra>
     enum_(const handle &scope, const char *name, const Extra&... extra)
-      : class_<Type>(scope, name, extra...), m_entries(), m_parent(scope) {
+      : cls(scope, name, extra...), m_entries(), m_parent(scope) {
 
         constexpr bool is_arithmetic = detail::any_of<std::is_same<arithmetic, Extra>...>::value;
 
         auto m_entries_ptr = m_entries.inc_ref().ptr();
-        def("__repr__", [name, m_entries_ptr](Type value) -> pybind11::str {
+
+        cls
+        .def("__repr__", [name, m_entries_ptr](Type value) -> pybind11::str {
             for (const auto &kv : reinterpret_borrow<dict>(m_entries_ptr)) {
                 if (pybind11::cast<Type>(kv.second) == value)
                     return pybind11::str("{}.{}").format(name, kv.first);
             }
             return pybind11::str("{}.???").format(name);
-        });
-        def_property_readonly_static("__members__", [m_entries_ptr](object /* self */) {
+        })
+        .def_property_readonly_static("__members__", [m_entries_ptr](object /* self */) {
             dict m;
             for (const auto &kv : reinterpret_borrow<dict>(m_entries_ptr))
                 m[kv.first] = kv.second;
             return m;
-        }, return_value_policy::copy);
-        def("__init__", [](Type& value, Scalar i) { value = (Type)i; });
-        def("__int__", [](Type value) { return (Scalar) value; });
+        }, return_value_policy::copy)
+        .def("__init__", [](Type& value, Scalar i) { value = (Type)i; })
+        .def("__int__", [](Type value) { return (Scalar) value; })
         #if PY_MAJOR_VERSION < 3
-            def("__long__", [](Type value) { return (Scalar) value; });
+            .def("__long__", [](Type value) { return (Scalar) value; })
         #endif
-        def("__eq__", [](const Type &value, Type *value2) { return value2 && value == *value2; });
-        def("__ne__", [](const Type &value, Type *value2) { return !value2 || value != *value2; });
+        .def("__eq__", [](const Type &value, Type *value2) { return value2 && value == *value2; })
+        .def("__ne__", [](const Type &value, Type *value2) { return !value2 || value != *value2; });
         if (is_arithmetic) {
-            def("__lt__", [](const Type &value, Type *value2) { return value2 && value < *value2; });
-            def("__gt__", [](const Type &value, Type *value2) { return value2 && value > *value2; });
-            def("__le__", [](const Type &value, Type *value2) { return value2 && value <= *value2; });
-            def("__ge__", [](const Type &value, Type *value2) { return value2 && value >= *value2; });
+            cls
+            .def("__lt__", [](const Type &value, Type *value2) { return value2 && value < *value2; })
+            .def("__gt__", [](const Type &value, Type *value2) { return value2 && value > *value2; })
+            .def("__le__", [](const Type &value, Type *value2) { return value2 && value <= *value2; })
+            .def("__ge__", [](const Type &value, Type *value2) { return value2 && value >= *value2; });
         }
         if (std::is_convertible<Type, Scalar>::value) {
             // Don't provide comparison with the underlying type if the enum isn't convertible,
             // i.e. if Type is a scoped enum, mirroring the C++ behaviour.  (NB: we explicitly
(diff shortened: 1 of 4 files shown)


include/pybind11/pybind11.h
```diff
@@ -1169,90 +1169,107 @@ class class_ : public detail::generic_type {
...
+    enum_ export_values() && {
         for (const auto &kv : m_entries)
             m_parent.attr(kv.first) = kv.second;
-        return *this;
+        return std::move(*this);
     }
 
     /// Add an enumeration entry
-    enum_& value(char const* name, Type value) {
+    enum_ value(char const* name, Type value) && {
```
dean0x7d: Why only rvalues? I'm a bit more worried about this breaking backward compatibility than the inherited `class_` functions.

include/pybind11/pybind11.h
```diff
@@ -1169,90 +1169,107 @@ class class_ : public detail::generic_type {
...
-        return *this;
+        return std::move(*this);
+    }
+
+    /// Get the associated class object.
+    class_<Type> into_class() && {
+        return cls;
+    }
+
+    operator class_<Type>() && {
```
dean0x7d: Why make it implicitly convertible?

dean0x7d:
> Note: this is technically a breaking change, but the consensus seems to be that this functionality was more of a side-effect (of deriving from class_) rather than an intended feature.

Just to make sure we don't break anything important, I'll just cc this for others how have also contributed to `enum_`: @pschella, @mdcb .

The gist is that `enum_` supported defining methods like `class_` but this was undocumented and would also break chaining, i.e. `enum_().value().def().value()` would not compile. The first `class_` API call would implicitly turn the `enum_` into a `class_`. This PR makes that explicit, documented and tested, but this will break the old undocumented behavior.

include/pybind11/pybind11.h
aldanor: That's how I had it initially (in the other PR) -- with hindsight, this could probably be removed.

include/pybind11/pybind11.h
aldanor: Well, if you return a reference, you won't be able to call `.into_class()` which needs an rvalue (which is an important point, in order to avoid all kinds of edge cases). Taking an rvalue `*this` here, disable copy ctor and returning by value basically helps ensure we never copy it (accidentally or not) and always move. The usual chaining-style enum definitions still all work as expected.

dean0x7d: Ah, right, it would be required for the new `enum34`. I was only looking at it for the `enum_` here, which does not need to return an rvalue.

aldanor: I've added back the lvalue overloads which seems to work and won't break backwards compatibility if someone chose to assign enum object to a variable and not chain the calls:

```cpp
// ok
auto e = enum_<>();
e.value().value();

// ok
enum_<>().value().into_class().def();

// not ok
auto e = enum_<>().value();
e.into_class().def(); // <-- compile fail
```

include/pybind11/pybind11.h
aldanor: This is now removed.

wjakob:
One minor suggestion: can we call it ``as_class`` instead of ``into_class``?

wjakob:
Actually, thinking a bit more about this, I find problematic that ``enum_`` is no longer a ``py::object``. This will definitely break a few of my projects that stick aliases of enumerations into other (sub-) modules to mirror ``using`` namespace statements on the C++ side. 

Why are we changing this?

aldanor:
@wjakob I've initially chosen `into_` as opposed to `as_` here to indicate the && requirement (same as in Rust conventions, `as_` returns a reference/view on the object, where as `into_` moves out of it and returns something else). Could definitely change it though, of course.

We are changing it because it was a bit broken initially (and undocumented) since `class_<>` is not a CRTP and its methods return `class_`. So, code like this won't compile: `.value().def().value()`. This wasn't the initial reason though -- in order to make the new py3 enums be API-compatible with the existing `enum_` there are two options (see https://github.com/pybind/pybind11/pull/781) -- adjust `enum_` so it's no longer a `class_`, or split out the public part of `class_` into a CRTP interface (the few latest commits in #781). The latter was deemed too intrusive, hence this PR.

wjakob:
It would be straightforward to fix the ``.def`` issue by adding a variadic template implementation that forwards its arguments to the ``.def`` implementation of the base class and then returns ``this&``. Does this address the issue you mentioned?

wjakob:
(i.e. no CRTP needed)

aldanor:
Sort of, yea. It's a bit more than def though... `.def`, `.def_property_readonly`, `.def_readonly_static`, and all of that. This would basically be a copy/paste-driven implementation, so to speak :)

aldanor:
(For the CRTP example, see cf1ef95a311547d55c755d990b9e725beb4df6a4)

wjakob:
This seems fine to me -- ~9-ish lines of code that will all be optimized away, no?

wjakob:
That's it, no?
```
template <typename... Args> def(Args&&... args) { return (enum_&) Base::def(std::forward<Args>(args)...); }
template <typename... Args> def_property_readonly(Args&&... args) { return (enum_&) Base:: def_property_readonly(std::forward<Args>(args)...); }
template <typename... Args> def_readonly_static(Args&&... args) { return (enum_&) Base:: def_readonly_static(std::forward<Args>(args)...); }
```
Or am I missing something?

aldanor:
For the new enum34 implementation - yea, something like this (x3) would work; hacky but should suffice.

If we're going this way... this is exactly what a separate CRTP would do, less the copy/paste. With a bit of effort I think it could benefit both enum implementations -- for `enum_` which derives from `class_` (which derives from CRTP) it could allow `.value().def().value()`; for `enum34` (which derives directly from CRTP) it could enable `.def_*()` methods.

wjakob:
I don't think it's hacky :). CRTP brings its own set of issues -- it definitely doesn't make the source easier to read ;). In one of my current projects, MSVC 2017 needs to be spoon-fed some of the CRTP-related type aliases to avoid compiler segfaults (accessing members of the derived class in template default parameters still seems somewhat fragile).

wjakob:
If the duplication bothers you, we could add a macro:
```cpp
#define PYBIND11_FORWARD(name) template <typename... Args> name(Args&&... args) { return static_cast<decltype(*this)>(Base::name(std::forward<Args>(args)...)); }
```
I'd really prefer this to making something that should be a ``py::object`` into a non-object.

aldanor:
A macro would be reasonable, it will certainly deal with `enum_`, but it won't work with enum34 as is :/ (first, because enum34 can't be derived from `class_`, and second, it cannot give you a `class_` lvalue, only a by-value when the instance is rvalue -- to prevent defining methods on a type that may get recreated). It's possible to handle all this lvalue/rvalue mess with crtp and proper template methods, but not with macro.

aldanor:
Btw if `enum_` remains a `py::object` and there's code actively relying on it then enum34 cannot be made an automatic drop-in replacement (which is what @dean0x7d and @jagerman seemed to suggest), because enum34 cannot have the same API as `enum_` and be a `py::object` at the same time.

jagerman:
Could `enum34` be derived directly from `py::object` (rather than indirectly through `py::class_`)?

aldanor:
Btw @wjakob could you post examples of code you've have that relies on `enum_` being an object (eg storing aliases to it)?

aldanor:
@jagerman I was thinking about it too, will take a look tomorrow.

wjakob:
One simple example are cases where an enumeration exists in separate classes. Something like this:

```cpp
class A {
   enum E {Value  = 0};
};

class B {
  using E = A::E;
};

class_<A> cls_a(m, "A");
class_<B> cls_b(m, "B");

enum_<A::E> e(cls_a, "E");
e.value("Value", A::E::Value);

cls_b.attr("E") = e;
```

aldanor:
@wjakob Here's your example working :) 9c75fd060981298e447153c95047af488b536948

wjakob:
Hm -- what about @jagerman's suggestion (``enum_`` being an object)?

wjakob:
Hm -- what about @jagerman's suggestion (``enum_`` being an object)?

aldanor:
I thought a little bit about it and don't think `enum34` can be an `object` (and the whole purpose of all this is to try and unify the two as much as possible), because it has a "finalize" step and has to recreate itself multiple times, so you have a `py::object` whose `m_ptr` may change outside of ctor, that doesn't sound too good. At least with cast ops you can control all this business by & and && method qualifiers.

jagerman:
I'm not sure I see the problem with the `m_ptr` changing.  If I do something like `m.attr("foo") = myenum;` and then add additional enum values to `myenum`, it's going to break in exactly the same way with the `handle` operator as it would with direct `object` inheritance.

aldanor:
@jagerman Hmm yea you're right, in the current state `handle` operator is a bit leaky. However... if we remove `enum_& value() &` (which I didn't have initially in this PR but then added, see comments above) and leave just `enum_ value() &&` then your example won't compile -- specifically you won't be able to "add an additional value".

jagerman:
> then your example won't compile

... nor will @wjakob's if the lvalue version is removed, which doesn't seem desirable.  Requiring you to write `std::move(myenum).into_class()` when you want both a local `myenum` variable and then a `class_` out of it seems pretty reasonable to me, since it means not forcing existing, simple code into a particular style (i.e. the all-at-once value definitions).  (And that's even ignoring the backwards compatibility issue).

aldanor:
Yep, you're right. Either way something will be broken it seems :) Unless there's some other, smarter way (deriving from object but not class_? idk).

If there's no middle ground, I'm ok with retracting this PR altogether since it was opened out of a suggestion voiced in enum34 PR. if everyone thinks the `enum_<>` is fine as is (being derived from `class_` in a half-broken way, that is), then ok; as I see it, the only way enum34 could be compatible with this then would be either allowing knowingly broken constructs to compile and leave it to users discretion (e.g. .value().def().value()), or pulling class_ apart into a CRTP which was deemed too intrusive iiuc.

dean0x7d:
I guess the entire point here was to make `enum_` and `enum34` API compatible and interchangeable to the point that users don't need to be aware of it and `enum34` is just an invisible optimization of `enum_` on compatible Python versions. But if that's not possible without breaking stuff and/or a lot of pain, I would recommend simplifying `enum34` to the essentials and making it an optional header `enum34.h` for those that need it/target the required Python versions.

aldanor:
Yea; far as I see now, without big changes it's only possible to make them API compatible if we accept the possibility of runtime errors thrown in enum34 API at module initialization time due to incorrect use (which gets compiled ok) -- I really wouldn't like this to be the case.

So (soon as I get back to IE...) I'll probably close this and revert all changes to `pybind11.h` in #781.

aldanor:
In fact I'll close this one now so it doesn't annoy anyone anymore :)

(I still think `enum_<>` is a bit broken in the current state due to how `.def()` works; I'm not sure if we care though).

PR #841, opened 2017-05-08
Set PYBIND11_CPP1{4,7} macros for MSVC

### LINKED ISSUE #833: MSVC 2017 and define PYBIND11_CPP17
PYBIND11_CPP17 not defined for msvc 2017 but msvc 2017 has support for extended constexpr and std::variant with compiler option `/std:c++latest`
I try explicit define PYBIND11_CPP17 on msvc and run test succesufull. (as far as I understand)

[https://github.com/pybind/pybind11/blob/master/include/pybind11/common.h#L19](url)
```
// Neither MSVC nor Intel support enough of C++14 yet (in particular, as of MSVC 2015 and ICC 17
// beta, neither support extended constexpr, which we rely on in descr.h), so don't enable pybind
// CPP14 features for them.
#if !defined(_MSC_VER) && !defined(__INTEL_COMPILER)
#  if __cplusplus >= 201402L
#    define PYBIND11_CPP14
#    if __cplusplus > 201402L /* Temporary: should be updated to >= the final C++17 value once known */
#      define PYBIND11_CPP17
#    endif
#  endif
#endif
```
Can you add PYBIND11_CPP17  support for msvc 2017  (with and without  `/std:c++latest` )?

jagerman:
We currently don't set `PYBIND11_CPP14` or `_CPP17` for MSVC because MSVC doesn't support the code in `descr.h`.  Originally, that's all the `_CPP14` flag guarded, but it now enables several other things as well (most of which are `#ifdef CPP14 || MSVC`).  This flips that, setting `CPP14` and `CPP17` (the latter only when compiling under MSVC 2017 with `/std:c++latest`), then changing the `descr.h` code to require `CPP14 && !MSVC`.

This PR also enables the `PYBIND11_CPP_STANDARD` cmake variable under MSVC, defaulting it to `/std:c++14`, then uses it to add a `/std:c++latest` appveyor job which runs the optional and variant tests.

Fixes #833.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,10 +11,19 @@ platform:
 environment:
   matrix:
   - CONDA: 36
+    CPP: 14
   - CONDA: 27
+    CPP: 14
+  - CONDA: 36
+    CPP: latest
 matrix:
   exclude:
     - image: Visual Studio 2015
+      platform: x86
+    - image: Visual Studio 2015
+      CPP: latest
+    - image: Visual Studio 2017
+      CPP: latest
       platform: x86
 install:
 - ps: |
@@ -37,7 +46,7 @@ install:
     7z x 3.3.3.zip -y > $null
     $env:CMAKE_INCLUDE_PATH = "eigen-eigen-67e894c6cd8f"
 build_script:
-- cmake -G "%CMAKE_GENERATOR%" -A "%CMAKE_ARCH%" -DPYBIND11_WERROR=ON -DCMAKE_SUPPRESS_REGENERATION=1
+- cmake -G "%CMAKE_GENERATOR%" -A "%CMAKE_ARCH%" -DPYBIND11_CPP_STANDARD=/std:c++%CPP% -DPYBIND11_WERROR=ON -DCMAKE_SUPPRESS_REGENERATION=1
 - set MSBuildLogger="C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 - cmake --build . --config Release --target pytest -- /v:m /logger:%MSBuildLogger%
 - cmake --build . --config Release --target test_cmake_build -- /v:m /logger:%MSBuildLogger%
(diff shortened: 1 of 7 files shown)


wjakob:
I'm curious -- what happens when using the constexpr ``descr`` with MSVC2017? When I tried this with MSVC2015 at some point, everything could be made to compile with minor tweaks -- however, when inspecting the resulting binary, it was impossible to find the function signature strings, suggesting that they were not precomputed at compile time.

jagerman:
It fails with: `cast.h(1121): error C2131: expression did not evaluate to a constant` and `descr.h(91): note: failure was caused by non-constant arguments or reference to a non-constant symbol`.

I didn't try investigating any further to try to fix it.

RedSkotina:
ok. give me time for investigating then.

jagerman:
Another MSVC annoyance: it looks like
```C++
using std::enable_if_t;
```
doesn't work with MSVC 2015.

jagerman:
A #792 failure, but otherwise passes now ([same code, no failure](https://ci.appveyor.com/project/jagerman/pybind11/build/1.0.690)).

RedSkotina:
pass all tests with and without -DPYBIND11_CPP_STANDARD=/std:c++latest on msvc 2017

dean0x7d:
LGTM

RedSkotina:
can you then update documentation about -DPYBIND11_CPP_STANDARD=/std:c++latest ?

jagerman:
It's already documented, though it didn't mention the doesn't-do-anything-on-MSVC limitation (which this PR removes).  I did reword things a bit, and added MSVC `/std:c++...` examples to the documentation.

jagerman:
It also changes the wording a bit to say that we explicitly target C++14 by default, rather than "the latest available," but mentions that it is subject to change to the latest available, which I think captures a little bit more clearly what we are doing.

jagerman:
Squashed and merged.

Re: `descr.h` C++14 path and MSVC 2017, it would be cool if someone figured that out, but this change is worthwhile whether or not we do (and makes enabling it for MSVC simple: just change the macro condition in `descr.h`).

PR #842, opened 2017-05-09
Add tests for pointer defaults.

abergmeier-dsfishlabs:
Not sure whether these are intended to work or not.

In 2.1.1 currently throws exceptions.

### DIFF
--- a/tests/test_issues.cpp
+++ b/tests/test_issues.cpp
@@ -394,6 +394,9 @@ void init_issues(py::module &m) {
 #elif defined(PYBIND11_HAS_EXP_OPTIONAL)
     m2.def("tpl_constr_optional", [](std::experimental::optional<TplConstrClass> &) {});
 #endif
+
+    m2.def("test_default_set_null", [](std::unordered_set<std::uint64_t>*) {}, "ids"_a = nullptr);
+    m2.def("test_default_set_none", [](std::unordered_set<std::uint64_t>*) {}, "ids"_a = py::none());
 }
 
 // MSVC workaround: trying to use a lambda here crashes MSVC
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -249,3 +249,9 @@ def test_inheritance_override_def_static():
     assert isinstance(b, MyBase)
     assert isinstance(d1, MyDerived)
     assert isinstance(d2, MyDerived)
+
+def test_default_set():
+    from pybind11_tests.issues import test_default_set_null, test_default_set_none
+
+    test_default_set_null()
+    test_default_set_none()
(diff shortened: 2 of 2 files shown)


wjakob:
This is not expected to work, unless you treat these maps as opaque types (see the documentation).

abergmeier-dsfishlabs:
Ok, but this is straight forward what I would expect to work.

wjakob:
The default STL converters are all pass-by-value, which is an elementary design decision. Did you read the documentation on this?

abergmeier-dsfishlabs:
> The default STL converters are all pass-by-value, which is an elementary design decision.

AFAIU this is no reason to not have additional logic, which enables defaults for pointers.
My problem is less the behavior but the unhelpful error message. `TypeError` is really not sufficient here.

abergmeier-dsfishlabs:
So now it is at least tested, that it throws a `TypeError`.

jagerman:
The implementation detail underlying this helps explain the behaviour a bit.  The main issue is that defaults get cast to Python objects before the method definition happens.  In this case, a nullptr is going to be cast to `None` when defining the method.  Later, when calling it from Python, we see the argument missing but that it has a default of `None`, so we copy `None` into the argument list before passing it on to the function.  That fails because there's no conversion from `None` to `set` (or the various other `stl` types).

Ideally we'd fail when you define the method, but that's not really feasible: the only way to know for sure whether `None` can be accepted is to pass it to the relevant type caster and see whether it accepts it or not.  (The `stl` ones don't).

To really solve this properly would need a sort of "do-you-accept-X?" method to type casters: then we could catch the error at definition time, rather than when calling the method.  (I'm not advocating for adding it: while it would help here, it would also make type casters more complex, which weighs against adding it).  Until/unless it supports such things, it's possible to pass defaults that can't actually be passed to the function.

(This also isn't restricted to `stl.h`: you could get the same error by specifying any default that can't be converted to the required type).

abergmeier-dsfishlabs:
@jagerman Thanks for the great explanation. There does seem to be room for improvement, though. What about, before raising the "TypeError", stepping back and detecting whether the error was due to a None/nullptr being passed and in that case raising "NoneValueError"?

jagerman:
It's possible, but it would be a bit messy: we're have to count the argument that failed to load in the argument loader, return that to the dispatcher, which would then, if there are no other overloads, inspect the arguments to see if the one that failed happened to be None.  But then you could have other complications: maybe None is allowed sometimes (e.g. if I set some object state earlier), so None isn't necessarily an error with a failed argument load.

I think a better approach would be to allow type casters to propagate a load error back to the function dispatcher, which could display the failures next to each attempted overload.  It would probably be feasible in a backwards-compatible way by repurposing (and renaming) the `reference_cast_error` exception to support nested another exception so that a type caster's `load()` could do something like `throw load_failure(std::invalid_argument_error("can't load a map from None"))` to try the next overload, but register the failure with the current overload in case no overloads succeed.

dean0x7d:
Since `test_issues.cpp` is going away as part of a test suite reorganization, I've cherry picked the changes from here into PR #898, so they will be merged along with that.

Issue #839, opened 2017-05-08
variant_caster cant handle c++ type nullptr_t


RedSkotina:
I have compile errors while compile this code (gcc 7.1 and vs 2017 checked)
```
    py::module m("xpytest", "pybind testing plugin");
    m.def("load_variant", [](std::variant<int, std::nullptr_t> v) { 
        return std::visit(visitor(), v);
    });
```
gcc output:
```
F:/devel/mingw-w64/i686-7.1.0-posix-dwarf-rt_v5-rev0/mingw32/bin/g++.exe  -c  "E:/code/zog/ptest3/xpytest.cpp" -O0 -g -std=c++17 -D_hypot=hypot -D_DEBUG  -o ./Debug/up_xpytest.cpp.o -I. -I. -If:\lang\Python35-32\include
In file included from f:\lang\Python35-32\include/pybind11/attr.h:13:0,
                 from f:\lang\Python35-32\include/pybind11/pybind11.h:36,
                 from E:/code/zog/ptest3/xpytest.cpp:3:
f:\lang\Python35-32\include/pybind11/cast.h: In instantiation of 'typename pybind11::detail::make_caster<T>::cast_op_type<T> pybind11::detail::cast_op(pybind11::detail::make_caster<T>&) [with T = std::nullptr_t; typename pybind11::detail::make_caster<T>::cast_op_type<T> = std::nullptr_t&; pybind11::detail::make_caster<T> = pybind11::detail::type_caster<std::nullptr_t>; typename pybind11::detail::intrinsic_type<T>::type = std::nullptr_t]':
f:\lang\Python35-32\include/pybind11/stl.h:305:31:   required from 'bool pybind11::detail::variant_caster<V<Ts ...> >::load_alternative(pybind11::handle, bool, pybind11::detail::type_list<U, Us ...>) [with U = std::nullptr_t; Us = {}; V = std::variant; Ts = {int, std::nullptr_t}]'
f:\lang\Python35-32\include/pybind11/stl.h:308:32:   required from 'bool pybind11::detail::variant_caster<V<Ts ...> >::load_alternative(pybind11::handle, bool, pybind11::detail::type_list<U, Us ...>) [with U = int; Us = {std::nullptr_t}; V = std::variant; Ts = {int, std::nullptr_t}]'
[...]

PR #843, opened 2017-05-09
Defer None loading to second pass

jagerman:
Following up the [discussion in #839](https://github.com/pybind/pybind11/issues/839#issuecomment-299932843), this changes most of the `is_none()` checks to defer to the `convert = true` pass.  This makes them consistent with the `char` caster (which already deferred), and seems more correct.  In particular, this allows a `py::none` overload to be used even if earlier overloads would accept a `None`.

This keeps `None` in the no-convert pass for the two type casters where it makes sense: the void caster, and the `std::optional` caster.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -218,6 +218,8 @@ public:
         if (!src || !typeinfo)
             return false;
         if (src.is_none()) {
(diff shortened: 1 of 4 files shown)


wjakob:
Makes sense -- LGTM!

dean0x7d:
Looks good to me.

PR #845, opened 2017-05-10
Improve constructor resolution in variant_caster

dean0x7d:
Currently, `py::int_(1).cast<variant<double, int>>()` fills the `double` slot of the variant. This commit switches the loader to a 2-pass scheme in order to correctly fill the `int` slot.

Together with #843, this will result in correct resolution of pointers in a variant as discussed in #839.


wjakob:
Looks great!

jagerman:
That's a nice, simple fix.  LGTM.

PR #846, opened 2017-05-11
Fix Eigen shape assertion error in dense matrix caster

dean0x7d:
Missing conformability check was causing Eigen to create a 0x0 matrix with an error in debug mode and silent corruption in release mode.


jagerman:
Good catch.  LGTM.

PR #849, opened 2017-05-12
std::reference_wrapper: (some) non-generic type support; failure on None

### LINKED ISSUE #848: std::reference_wrapper accepts nullptr / None arguments
Seems I'm able to pass None to a function that accepts an std::reference_wrapper. (While for an ordinary reference, I get the appropriate cast error.)

So I'm assuming this example is not supposed to give this output:

```C++
#include <iostream>

class A {};

void f(std::reference_wrapper<A> a)
{
	std::cout << "called with &a.get() == " << (void *) &a.get() << std::endl;
}

namespace py = pybind11;

PYBIND11_PLUGIN(test) {
    py::module m("test");

    py::class_<A>(m, "A")
        .def(py::init<>())
    ;

    m.def("f", &f);

    return m.ptr();
}
```

Python output:
```Python
>>> import test
>>> test.f(test.A())
called with &a.get() == 0x28dd7e0
>>> test.f(None)
called with &a.get() == 0
```

jagerman:
This reimplements the `std::reference_wrapper<T>` caster to be a shell around the underlying `T` caster (rather than assuming `T` is a generic type), which lets it work for things like `std::reference_wrapper<int>` or anything else custom type caster with a lvalue cast operator.

Fixes #848: this also makes it properly fail when `None` is provided, just as an ordinary lvalue reference argument would similarly fail.

This also adds a static assert to test that `T` has an appropriate type caster.  It triggers for casters like `std::pair`, which have return-by-value cast operators.  (In theory this could be supported by storing a local temporary for such types, but that's beyond the scope of this PR).  This shouldn't break anything in terms of backwards compatibility: currently `reference_wrapper` only actually works for generic types, which always have an lvalue cast operator.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -527,13 +527,24 @@ cast_op(make_caster<T> &&caster) {
         typename make_caster<T>::template cast_op_type<typename std::add_rvalue_reference<T>::type>();
 }
 
-template <typename type> class type_caster<std::reference_wrapper<type>> : public type_caster_base<type> {
+template <typename type> class type_caster<std::reference_wrapper<type>> {
+private:
+    using caster_t = make_caster<type>;
+    caster_t subcaster;
+    using subcaster_cast_op_type = typename caster_t::template cast_op_type<type>;
+    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value,
+            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
 public:
+    bool load(handle src, bool convert) { return subcaster.load(src, convert); }
+    static PYBIND11_DESCR name() { return caster_t::name(); }
     static handle cast(const std::reference_wrapper<type> &src, return_value_policy policy, handle parent) {
-        return type_caster_base<type>::cast(&src.get(), policy, parent);
+        // It is definitely wrong to take ownership of this pointer, so mask that rvp
+        if (policy == return_value_policy::take_ownership || policy == return_value_policy::automatic)
+            policy = return_value_policy::automatic_reference;
+        return caster_t::cast(&src.get(), policy, parent);
     }
     template <typename T> using cast_op_type = std::reference_wrapper<type>;
-    operator std::reference_wrapper<type>() { return std::ref(*((type *) this->value)); }
+    operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type&(); }
 };
 
 #define PYBIND11_TYPE_CASTER(type, py_name) \
--- a/tests/test_issues.cpp
+++ b/tests/test_issues.cpp
@@ -117,19 +117,6 @@ void init_issues(py::module &m) {
         .def(py::init<int>())
         .def("__repr__", [](const Placeholder &p) { return "Placeholder[" + std::to_string(p.i) + "]"; });
 
-    // #171: Can't return reference wrappers (or STL datastructures containing them)
-    m2.def("return_vec_of_reference_wrapper", [](std::reference_wrapper<Placeholder> p4) {
-        Placeholder *p1 = new Placeholder{1};
-        Placeholder *p2 = new Placeholder{2};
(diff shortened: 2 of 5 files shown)


jagerman:
Struck by random CI failures on both appveyor and travis on this one.  (Restarted the travis, and the appveyor is just a single #792 failure).

include/pybind11/cast.h
```diff
@@ -495,13 +495,20 @@ template <typename T> typename make_caster<T>::template cast_op_type<T> cast_op(
...
+template <typename type> class type_caster<std::reference_wrapper<type>> {
+private:
+    using caster_t = make_caster<type>;
+    caster_t subcaster;
 public:
+    bool load(handle src, bool convert) { return subcaster.load(src, convert); }
+    static PYBIND11_DESCR name() { return caster_t::name(); }
     static handle cast(const std::reference_wrapper<type> &src, return_value_policy policy, handle parent) {
-        return type_caster_base<type>::cast(&src.get(), policy, parent);
+        return caster_t::cast(&src.get(), policy, parent);
```
dean0x7d: This existed before this PR, but I'm just noticing: casting as a pointer here means that `rvp::automatic` will behave like `rvp::take_ownership` which goes completely against `std::reference_wrapper`. Would there be anything wrong with just hard coding `rvp::reference` here? I don't think any other policy could possibly be valid.

jagerman: Sounds reasonable to me.

jagerman: So I agree that `rvp::automatic` and `take_ownership` seem wrong and should be masked, but `copy` could be valid.  Perhaps it would be better to just set it to `rvp::reference` if it's set to `automatic` or `take_ownership`, and otherwise leave it alone?

dean0x7d: When is `copy` valid?

jagerman: `copy` is the default currently if you `py::cast(x)` a `reference_wrapper<T> x`.  Essentially both `copy` and `reference` currently work the way I'd expect:

```C++

namespace py = pybind11;

struct Foo { int value = 2; };

std::vector<std::reference_wrapper<Foo>> foos() {
    static Foo foo1, foo2;
    std::vector<std::reference_wrapper<Foo>> r;
    r.emplace_back(foo1);
    r.emplace_back(foo2);
    return r;
}

PYBIND11_PLUGIN(copy) {
    py::module m("copy");

    py::class_<Foo>(m, "Foo").def(py::init<>()).def_readwrite("value", &Foo::value);

    m.def("z1", []() {
        py::list l;
        for (auto &f : foos()) l.append(py::cast(f, py::return_value_policy::copy));
        return l;
    });
    m.def("z2", []() {
        py::list l;
        for (auto &f : foos()) l.append(py::cast(f, py::return_value_policy::reference));
        return l;
    });

    return m.ptr();
}
```

```pycon
>>> from copy import z1, z2, Foo
>>> a = z1()
>>> a2 = z1()
>>> b = z2()
>>> b2 = z2()
>>> a[0] is a2[0]
False
>>> b[0] is b2[0]
True
```

(The `::copy` there is redundant: it'll be the default for an lvalue reference wrapper already.)

jagerman: Though to be honest, I'm not sure why we pass a pointer here at all, instead of just passing an lvalue reference (i.e. changing `&src.get()` -> `src.get()`).

dean0x7d: Doesn't `&src.get()` -> `src.get()` make the following a copy?

```c++
T x = ...; // T is a user-defined type
py::function f = ...;
f(std::ref(x)); // <-- rvp::automatic_reference: the `cast(const&)` overload does a copy 
```

jagerman: That doesn't actually copy (because of the rvalue), but if you change the reference wrapper argument to an lvalue it is indeed a problem.

I'll change back to the pointer cast, with a `take_ownership` mask.

dean0x7d:
Looks like there's a conflict now. But otherwise, looks good to me.

PR #851, opened 2017-05-15
Add movable cast support to type casters

jagerman:
This commit allows type casters to allow their local values to be moved away, rather than copied, when the type caster instance itself is an rvalue.

This only applies (automatically) to type casters using `PYBIND11_TYPE_CASTER`; the generic type type casters don't own their own pointer, and various value casters (e.g. `std::string`, `std::pair`, arithmetic types) already cast to an rvalue (i.e. their cast op is a return-by-value).

This updates various calling code to attempt to get a movable value whenever the value is itself coming from a type caster about to be destroyed: for example, when constructing a `std::pair` or various `stl.h` containers.  For types that don't support value moving, these casts fall back to an lvalue cast.

There wasn't an obvious place to add the tests, so I added them to test_copy_move_policies, but also renamed it to drop the _policies as it now tests more than just policies.

.so size (before adding the tests) is slightly reduced, though the reduction is small enough (just under 4K) that it could just be noise; suffice it to say, there is no significant increase.


dean0x7d:
Looks good to me!

include/pybind11/cast.h
```diff
@@ -374,11 +374,31 @@ class type_caster_generic {
     object temp;
 };
 
-/* Determine suitable casting operator */
+/** Determine suitable casting operator for pointer-or-lvalue-casting type casters.  The type caster
+ * needs to provide `operator T*()` and `operator T&()` operators.
```
dean0x7d: Style nitpick: could the first and second text line be aligned? I.e.:
```c++
/** 
 * Determine suitable casting operator for pointer-or-lvalue-casting type casters.  The type caster
 * needs to provide `operator T*()` and `operator T&()` operators.
 */
```
(The misalignment probably bothers me more than it should.)

jagerman: Yeah, that does look better.  37b2383 does this on master, and I'll update this PR.

wjakob:
This looks great! The 4KB change is definitely above the noise floor :)

PR #855, opened 2017-05-16
Override deduced Base class when defining Derived methods

### LINKED ISSUE #854: Problems wrapping member function defined in base class
#### Issue description

If a member function being wrapped by `py::class_::def` is actually implemented in a base class, then calling the member fails.

In the example code, it gives
```
  File "./testit.py", line 6, in <module>
    b.do_nothing()
TypeError: do_nothing(): incompatible function arguments. The following argument types are supported:
    1. (self: A) -> None

Invoked with: <pbtest._pbtest.B object at 0x7f1a691c3810>
```

This is presumably because the type of &B::do_nothing is actually `void (A::*)()` rather than `void (B::*)()`. While it can be fixed by explicitly casting it, it seems like it should be possible for pybind11 to fix this itself, since it knows at compile time which class is being wrapped.

C++:
```c++

namespace py = pybind11;

class A
{
public:
    void do_nothing() {}
};

class B : public A
{
public:
    using A::A;
};

PYBIND11_PLUGIN(_pbtest)
{
    py::module m("_pbtest");

    py::class_<B>(m, "B")
        .def(py::init<>())
        .def("do_nothing", &B::do_nothing);

    return m.ptr();
}
```

Python:
```python
from pbtest._pbtest import B

b = B()
b.do_nothing()
```

jagerman:
When defining method from a member function pointer (e.g. `.def("f", &Derived::f)`) we run into a problem if `&Derived::f` is actually implemented in some base class `Base` when `Base` isn't pybind-registered.

This happens because the class type is deduced, which then becomes a lambda with first argument this deduced type.  For a base class implementation, the deduced type is `Base`, not `Derived`, and so we generate and registered an overload which takes a `Base *` as first argument.  Trying to call this fails if `Base` isn't registered (e.g. because it's an implementation detail class that isn't intended to be exposed to Python) because the type caster for an unregistered type always fails.

This commit extends the pybind11::is_method annotation into a templated annotation containing the class being registered, which we can then extract to override the first argument to the derived type when attempting to register a base class method for a derived class.

Fixes #854 , #910

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -620,6 +620,11 @@ using exactly_one_t = typename exactly_one<Predicate, Default, Ts...>::type;
 template <typename T, typename... /*Us*/> struct deferred_type { using type = T; };
 template <typename T, typename... Us> using deferred_t = typename deferred_type<T, Us...>::type;
 
+/// Like is_base_of, but requires a strict base (i.e. `is_strict_base_of<T, T>::value == false`,
+/// unlike `std::is_base_of`)
+template <typename Base, typename Derived> using is_strict_base_of = bool_constant<
+    std::is_base_of<Base, Derived>::value && !std::is_same<Base, Derived>::value>;
+
 template <template<typename...> class Base>
 struct is_template_base_of_impl {
     template <typename... Us> static std::true_type check(Base<Us...> *);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -910,11 +910,22 @@ inline void call_operator_delete(void *p) { ::operator delete(p); }
 
 NAMESPACE_END(detail)
 
+/// Given a pointer to a member function, cast it to its `Derived` version.
+/// Forward everything else unchanged.
+template <typename /*Derived*/, typename F>
+auto method_adaptor(F &&f) -> decltype(std::forward<F>(f)) { return std::forward<F>(f); }
+
+template <typename Derived, typename Return, typename Class, typename... Args>
+auto method_adaptor(Return (Class::*pmf)(Args...)) -> Return (Derived::*)(Args...) { return pmf; }
+
+template <typename Derived, typename Return, typename Class, typename... Args>
+auto method_adaptor(Return (Class::*pmf)(Args...) const) -> Return (Derived::*)(Args...) const { return pmf; }
+
 template <typename type_, typename... options>
 class class_ : public detail::generic_type {
     template <typename T> using is_holder = detail::is_holder_type<type_, T>;
-    template <typename T> using is_subtype = detail::bool_constant<std::is_base_of<type_, T>::value && !std::is_same<T, type_>::value>;
-    template <typename T> using is_base = detail::bool_constant<std::is_base_of<T, type_>::value && !std::is_same<T, type_>::value>;
+    template <typename T> using is_subtype = detail::is_strict_base_of<type_, T>;
+    template <typename T> using is_base = detail::is_strict_base_of<T, type_>;
     // struct instead of using here to help MSVC:
     template <typename T> struct is_valid_class_option :
         detail::any_of<is_holder<T>, is_subtype<T>, is_base<T>> {};
@@ -980,7 +991,7 @@ public:
 
     template <typename Func, typename... Extra>
     class_ &def(const char *name_, Func&& f, const Extra&... extra) {
-        cpp_function cf(std::forward<Func>(f), name(name_), is_method(*this),
+        cpp_function cf(method_adaptor<type>(std::forward<Func>(f)), name(name_), is_method(*this),
                         sibling(getattr(*this, name_, none())), extra...);
         attr(cf.name()) = cf;
         return *this;
@@ -1044,15 +1055,17 @@ public:
 
     template <typename C, typename D, typename... Extra>
     class_ &def_readwrite(const char *name, D C::*pm, const Extra&... extra) {
-        cpp_function fget([pm](const C &c) -> const D &{ return c.*pm; }, is_method(*this)),
-                     fset([pm](C &c, const D &value) { c.*pm = value; }, is_method(*this));
+        static_assert(std::is_base_of<C, type>::value, "def_readwrite() requires a class member (or base class member)");
+        cpp_function fget([pm](const type &c) -> const D &{ return c.*pm; }, is_method(*this)),
+                     fset([pm](type &c, const D &value) { c.*pm = value; }, is_method(*this));
         def_property(name, fget, fset, return_value_policy::reference_internal, extra...);
         return *this;
     }
 
     template <typename C, typename D, typename... Extra>
     class_ &def_readonly(const char *name, const D C::*pm, const Extra& ...extra) {
-        cpp_function fget([pm](const C &c) -> const D &{ return c.*pm; }, is_method(*this));
+        static_assert(std::is_base_of<C, type>::value, "def_readonly() requires a class member (or base class member)");
+        cpp_function fget([pm](const type &c) -> const D &{ return c.*pm; }, is_method(*this));
         def_property_readonly(name, fget, return_value_policy::reference_internal, extra...);
         return *this;
     }
@@ -1075,7 +1088,8 @@ public:
     /// Uses return_value_policy::reference_internal by default
     template <typename Getter, typename... Extra>
     class_ &def_property_readonly(const char *name, const Getter &fget, const Extra& ...extra) {
-        return def_property_readonly(name, cpp_function(fget), return_value_policy::reference_internal, extra...);
+        return def_property_readonly(name, cpp_function(method_adaptor<type>(fget)),
+                                     return_value_policy::reference_internal, extra...);
     }
 
     /// Uses cpp_function's return_value_policy by default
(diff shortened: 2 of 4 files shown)


bmerry:
This doesn't appear to work for `def_property`, I'm guessing because the `is_method` attribute is only applied later by `detail::process_attributes` rather than at the time the `cpp_function` is constructed.

jagerman:
Should work for `def_property` now.

include/pybind11/pybind11.h
```diff
@@ -1075,15 +1083,27 @@ class class_ : public detail::generic_type {
     }
 
     /// Uses return_value_policy::reference_internal by default
+    template <typename Getter, typename Setter, typename... Extra>
+    class_ &def_property(const char *name, const Getter &fget, const Setter &fset, const Extra& ...extra) {
+        return def_property(name,
+                cpp_function(fget, is_method<type>(*this)), cpp_function(fset, is_method<type>(*this)),
+                return_value_policy::reference_internal, extra...);
+    }
```
bmerry: You've ended up with 4 versions of def_property, for the four combination of {cpp_function, generic} for fget and for fset. Possibly the code would be simpler given a helper template function that takes a getter/setter and a (list of) default call policy, and wraps it into a cpp_function with `is_method` if it isn't already a cpp_function, or returns the provided cpp_function as-is if it is. I think that would allow for only a single version of `def_property`, and probably also simplifies `def_property_readonly`.

By the way, is there any particular reason that def and def_static take their function via perfect forwarding while def_property takes it by const reference?

tests/test_methods_and_attributes.cpp
```diff
@@ -322,4 +338,19 @@ test_initializer methods_and_attributes([](py::module &m) {
...
+    // The methods and member pointers below actually resolve to members/pointers in
+    // UnregisteredBase; before this test/fix they would be registered via lambda with a first
+    // argument of an unregistered type, and thus uncallable.
+    py::class_<RegisteredDerived>(m, "RegisteredDerived")
+        .def(py::init<>())
+        .def("do_nothing", &RegisteredDerived::do_nothing)
+        .def("increase_value", &RegisteredDerived::increase_value)
+        .def_readwrite("rw_value", &RegisteredDerived::rw_value)
+        .def_readonly("ro_value", &RegisteredDerived::ro_value)
+        .def_property("rw_value_prop", &RegisteredDerived::get_int, &RegisteredDerived::set_int)
```
bmerry: Missing a test for def_property_readonly (I don't think the other tests provide full coverage: def_readonly calls def_property_readonly, but passes it a cpp_function).

include/pybind11/pybind11.h
jagerman: I know; I don't like it that much.  The alternative is adding a move constructor that takes and ignores extra arguments (i.e. `template <typename... Extra> cpp_function(cpp_function &&f, const Extra &...)`), but I like that approach even less (mainly because it's too late to apply any of the Extra arguments).

bmerry: I don't see the extra arguments as that problematic, but I concede that it's a matter of taste.

However, I think you might be able to delete the `cpp_function &fget, const Setter &fset` overload that you added (L1098) if you change the fully generic overload above to only wrap the setter and forward the getter as-is to be wrapped (if necessary) at the next layer.

jagerman: I'm not seeing how: if either of the `cpp_function, Setter` or `Getter, cpp_function` versions aren't there, then a call with one `cpp_function` and one lambda is going to end up at the fully-generic `Getter, Setter` template, which doesn't handle `cpp_function` arguments.

bmerry: That's why you also need to change the fully-generic template to only wrap the setter. Let's say that c, d are instances of cpp_function, C=cpp_function, G=Getter (template parameter), S=setter (template parameter) and p, q are non-cpp_function getter/setter. Then the chain of calls for each of the cases is
1. p, q: <G, S> -> <G, C> -> <C, C>
2. c, q: <G, S> -> <C, C>
3. p, c: <G, C> -> <C, C>
4. c, c: <C, C>
I don't know if that notation makes sense. If I've left you just as unclear, let me know and I can send you a PR against your branch.

wjakob:
I'm on the fence here. I think it would be reasonable to expect the user to create a (tiny) binding for the base class rather than making the pybind11 implementation more complex.

bmerry:
My worry is that which class in a hierarchy defines a member function is not normally considered part of an API. The author of the C++ class might push some members into a base class (which may be an implementation detail class) without believing that it will break API compatibility, but it will break the pybind11 bindings (and in a way that is only apparent at runtime with a cryptic error message).

It's not entirely related to this PR, but as a workaround for this issue I've written a macro that takes the pain out of wrapping the PTMF: see [here](https://github.com/ska-sa/spead2/blob/90b59840be54b131e3c849fb6d2e9c013b6ed787/include/spead2/py_common.h#L241-L287). If something like that was added to pybind11 it would at least be less onerous to wrap up every PTMF for robustness.

jagerman:
A simpler approach would be to fix the error message with an error about `Base` being an unregistered class.  That, however, would require letting the argument loader communicate failures back to the dispatcher, as I roughly outlined in https://github.com/pybind/pybind11/issues/864#issuecomment-302407868

Right now the biggest issue in call failures is that we have no way to communicate *why* a call failed.  Sometimes you can look at the "invoked with" and see that it doesn't match, but having each overload communicate a failure reason would be pretty nice.

bmerry:
As an example, consider that GCC derives most of its STL classes from base classes with names like `_Vector_base`. From a quick look I haven't found an example where the public member functions are actually defined in the base class, but that doesn't mean that some other compiler won't do so - and if it did then code in `stl_bind.h` that binds pointers to members would break.

More helpful error messages would definitely be better than not having them, but it doesn't address the fragility/portability concerns.

YannickJadoul:
What would actually be the use case of nót having the class itself as first argument of a method? Is there a situation where that's not a mistake?
I seem to remember that Boost.Python enforces this by default when registering a method to a class, not just when it's an unregistered base, even. (Just wondering if that might approach might simplify the proposed solution to a point where the complexity is acceptable.)

jagerman:
I've simplified this implementation now by adding a `cpp_function::method<D>` that works just like the constructors except in the special case where it's called with a Base member function pointer: in that case, it generates a lambda taking the derived type as `self`.

include/pybind11/common.h
```diff
@@ -582,6 +582,11 @@ using exactly_one_t = typename exactly_one<Predicate, Default, Ts...>::type;
 template <typename T, typename... /*Us*/> struct deferred_type { using type = T; };
 template <typename T, typename... Us> using deferred_t = typename deferred_type<T, Us...>::type;
 
+/// Like is_base_of, but requires a strict base (i.e. `is_strict_base_of<T, T>::value == false`,
+/// unlike `std::is_base_of`)
+template <typename Base, typename Derived> using is_strict_base_of = bool_constant<
+    std::is_base_of<Base, Derived>::value && !std::is_base_of<Derived, Base>::value>;
```
bmerry: Might be clearer/more self-documenting to say `... && !std::is_same<Derived, Base>::value`.

include/pybind11/pybind11.h
```diff
@@ -906,13 +906,40 @@ void call_operator_delete(T *p) { T::operator delete(p); }
...
+// Wrapper around cpp_function that binds a method pointer of a base class with a derived class
+// argument, as opposed to calling cpp_function directly which always uses the inferred class.  For
+// any other type of argument, this is identical to cpp_function.
+template <typename Derived, typename Func, typename SFINAE = void> struct method_adaptor_impl {
+    template <typename... Args> static cpp_function adapt(Args &&... args) { return {std::forward<Args>(args)...}; }
+};
+template <typename Derived, typename Return, typename Class, typename... Arg>
+struct method_adaptor_impl<Derived, Return (Class::*)(Arg...), enable_if_t<is_strict_base_of<Class, Derived>::value>> {
+    template <typename... Extra> static cpp_function adapt(Return (Class::*f)(Arg...), const Extra&... extra) {
+        return {[f](Derived *c, Arg... args) -> Return { return (c->*f)(args...); }, extra...};
```
bmerry: Any particular reason that the specialisations use `const &` for the extras while the generic implementation uses perfect forwarding?

jagerman: There's no advantage to forwarding extras given the way they are used.  The generic one uses perfect forwarding because there could be an potential advantage in forwarding a stateful lambda (the first argument), and it's simpler to just forward everything.

dean0x7d:
Hope you don't mind me pushing directly here, but your simplification gave me an idea to take it further still. I wasn't sure about it, had to write it out completely and test it. Looks like it works. 

Now,  `method_adaptor` just does a simple cast from `Return (Source::*)(Args...)` to `Return (Target::*)(Args...)`. Because it's an implicit cast, the compiler will only allow compatible `Source` -> `Target` member function conversions, like `Base` -> `Derived`, but `Class` -> `UnrelatedClass` will result in a compile-time error.

jagerman:
Very nice.  :+1:

jagerman:
Rebased and squashed.  Any other comments before committing?

include/pybind11/pybind11.h
```diff
@@ -908,11 +908,24 @@ inline void call_operator_delete(void *p) { ::operator delete(p); }
...
 NAMESPACE_END(detail)
 
+/// Given a pointer to a member function, cast it to its `Derived` version.
+/// Forward everything else unchanged.
+template <typename /*Derived*/, typename F>
+auto method_adaptor(F &&f) -> decltype(std::forward<F>(f)) { return std::forward<F>(f); }
+
+template <typename Derived, typename Return, typename Class, typename... Args,
+          typename Adapted = Return (Derived::*)(Args...)>
+Adapted method_adaptor(Return (Class::*pmf)(Args...)) { return pmf; }
```
bmerry: I'm not clear why Adapted is a template parameter. It can't be explicitly specified (because Args will swallow all trailing arguments) nor inferred (because it is only used in the return type), so it seems to exist solely to assign a name to the return type. While the C++03 syntax for a function returning a function pointer is horrific, would
```c++
template <typename Derived, typename Return, typename Class, typename... Args>
auto method_adaptor(Return (Class::*pmf)(Args...)) -> Return (Derived::*)(Args...) { return pmf; }
```
be clearer? Seems to pass on my system (GCC 5.4), haven't tried running it through Travis and Appveyor or applying it to the const version.

bmerry: (LGTM apart from this one comment)

dean0x7d: `Adapted` is only there to avoid the horror of the regular function pointer return type syntax. But the trailing return type solution looks much nicer!

PR #859, opened 2017-05-17
Add py::arg().none(false) class attribute

jagerman:
Edit: revamped; see below.

Old description:

---

This new class attribute lets you specify that a pointer or copyable holder type should not be constructed with a `nullptr` when given `None`, but rather that `None` should result in an incompatible function call error.

Enabling this by default would break backwards compatibility, so this is an opt-in feature, to be used as:

```C++
    py::class_<C>(m, "C", !py::accepts_none())
```

but I think that eventually (i.e. when we break backwards compatibility, 3.0) we should consider making this the default, requiring:

```C++
    py::class_<C>(m, "C", py::accepts_none())
```

in cases where `None` -> `nullptr` conversion is explicitly wanted.  This PR accepts that to explicitly request None conversion in a way that will continue to work if the default changes.

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -406,6 +406,53 @@ name, i.e. by specifying ``py::arg().noconvert()``.
     need to specify a ``py::arg()`` annotation for each argument with the
     no-convert argument modified to ``py::arg().noconvert()``.
 
+Allow/Prohibiting None arguments
+================================
+
+When a C++ type registered with :class:`py::class_` is passed as an argument to
+a function taking the instance as pointer or shared holder (e.g. ``shared_ptr``
+or a custom, copyable holder as described in :ref:`smart_pointers`), pybind
+allows ``None`` to be passed from Python which results in calling the C++
+function with ``nullptr`` (or an empty holder) for the argument.
+
+To explicitly enable or disable this behaviour, using the
+``.none`` method of the :class:`py::arg` object:
+
+.. code-block:: cpp
+
+    py::class_<Dog>(m, "Dog").def(py::init<>());
+    py::class_<Cat>(m, "Cat").def(py::init<>());
+    m.def("bark", [](Dog *dog) -> std::string {
+        if (dog) return "woof!"; /* Called with a Dog instance */
+        else return "(no dog)"; /* Called with None, d == nullptr */
+    }, py::arg("dog").none(true));
+    m.def("meow", [](Cat *cat) -> std::string {
+        // Can't be called with None argument
+        return "meow";
+    }, py::arg("cat").none(false));
+
+With the above, the Python call ``bark(None)`` will return the string ``"(no
+dog)"``, while attempting to call ``meow(None)`` will throw a :exc:`TypeError`:
+
+.. code-block:: pycon
+
+    >>> from animals import Dog, Cat, bark, meow
+    >>> bark(Dog())
+    'woof!'
+    >>> meow(Cat())
+    'meow'
+    >>> bark(None)
+    '(no dog)'
+    >>> meow(None)
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+    TypeError: meow(): incompatible function arguments. The following argument types are supported:
+        1. (cat: animals.Cat) -> str
+
+    Invoked with: None
+
+The default behaviour when the tag is unspecified is to allow ``None``.
+
 Overload resolution order
 =========================
 
--- a/tests/test_methods_and_attributes.cpp
+++ b/tests/test_methods_and_attributes.cpp
@@ -162,6 +162,14 @@ public:
 /// Issue/PR #648: bad arg default debugging output
 class NotRegistered {};
 
+// Test None-allowed py::arg argument policy
+class NoneTester { public: int answer = 42; };
+int none1(const NoneTester &obj) { return obj.answer; }
+int none2(NoneTester *obj) { return obj ? obj->answer : -1; }
+int none3(std::shared_ptr<NoneTester> &obj) { return obj ? obj->answer : -1; }
+int none4(std::shared_ptr<NoneTester> *obj) { return obj && *obj ? (*obj)->answer : -1; }
+int none5(std::shared_ptr<NoneTester> obj) { return obj ? obj->answer : -1; }
+
 test_initializer methods_and_attributes([](py::module &m) {
     py::class_<ExampleMandA> emna(m, "ExampleMandA");
     emna.def(py::init<>())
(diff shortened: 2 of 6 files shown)


bmerry:
I haven't looked at the implementation, but the interface feels wrong to me. Whether a null pointer is acceptable is in general part of the contract of the function it is being passed to, rather than the type. So I think it ought to be a modifier on `py::arg`, similar to `noconvert`. I don't know if that's possible to do generically.

jagerman:
Hmm, that's a good point.

jagerman:
Rewrote it to move the disable-none handling to the function/method level:

```C++
    m.def("f", [](C *c) { return c->whatever(); }, py::arg().none(false));
```
gives you an `f()` that can't be called with `None`.

dean0x7d:
What's the motivation for this feature? I don't really see when it would be useful. If I have a `void foo(T* p)` function in C++, I can write `foo(nullptr)` in C++. At that point, this is either a valid argument for `foo` or `foo` should handle it in some way (e.g. `throw`). The validity of `nullptr` as an argument is already determined by the C++ API and the same should apply to `foo(None)` in Python.

jagerman:
Also a good point.  @bmerry?

jagerman:
I think the main motivation is that there is code that takes a pointer that really isn't `nullptr`-safe (our test suite is full of them).  I think also that the expectations regarding segfaulting for a C++ programmer passing `nullptr` are rather different than for a Python programmer passing `None`.  This essentially lets you bind known-nullptr-unsafe code with extra information that it isn't nullptr safe (thus allowing an exception rather than a segfault).

bmerry:
What @jagerman said. And even if C++ (or C) code checks for a null pointer, it doesn't necessarily handle it in a Pythonic way e.g. it could assert/abort, or throw an exception but not one that becomes `TypeError` in Python, or print a warning then return (ala GLib assertions).

docs/advanced/functions.rst
```diff
@@ -406,6 +406,38 @@ name, i.e. by specifying ``py::arg().noconvert()``.
...
+    py::class_<Cat>(m, "Cat")
+        .def(py::init<>());
+    m.def("bark", [](Dog *d) -> std::string {
+        if (d) return "woof!"; /* Called with a Dog instance */
+        else return "(no dog)"; /* Called with None, d == nullptr */
+    });
+    m.def("meow", [](Cat *c) -> std::string {
+        // Can't be called with None argument
+        return "meow";
+    });
```
bmerry: You neglected to actually demonstrate the `none` call.

docs/advanced/functions.rst
```diff
@@ -406,6 +406,38 @@ name, i.e. by specifying ``py::arg().noconvert()``.
...
+
+With the above, the Python call ``bark(None)`` will return the string ``"(no
+dog)"``, while attempting to call ``meow(None)`` will result in a type error.
```
bmerry: "type error" -> ":exc:`TypeError`"?

tests/test_methods_and_attributes.cpp
```diff
@@ -322,4 +330,18 @@ test_initializer methods_and_attributes([](py::module &m) {
...
+        .def(py::init<>());
+    m.def("no_none1", &none1, py::arg().none(false));
+    m.def("no_none2", &none2, py::arg().none(false));
+    m.def("no_none3", &none3, py::arg().none(false));
+    m.def("no_none4", &none4, py::arg().none(false));
+    m.def("no_none5", &none5, py::arg().none(false));
+    m.def("ok_none1", &none1);
+    m.def("ok_none2", &none2, py::arg().none(true));
+    m.def("ok_none3", &none4);
+    m.def("ok_none4", &none3, py::arg().none(true));
```
bmerry: none3 and none4 seem to be swapped here.

tests/test_methods_and_attributes.py
```diff
@@ -369,3 +369,47 @@ def test_bad_arg_default(msg):
...
+    assert "incompatible function arguments" in str(excinfo.value)
+    with pytest.raises(TypeError) as excinfo:
+        no_none3(None)
+    assert "incompatible function arguments" in str(excinfo.value)
+    with pytest.raises(TypeError) as excinfo:
+        no_none4(None)
+    assert "incompatible function arguments" in str(excinfo.value)
+    with pytest.raises(TypeError) as excinfo:
+        no_none5(None)
+    assert "incompatible function arguments" in str(excinfo.value)
```
bmerry: This whole test probably be made more DRY (although I guess at the expense of being able to immediately tell from the line number which test failed) using loops over the functions to test.

docs/advanced/functions.rst
jagerman: Whoops, fixed.

docs/advanced/functions.rst
jagerman: 👍

tests/test_methods_and_attributes.cpp
jagerman: 👍

tests/test_methods_and_attributes.py
jagerman: While I normally agree, I rather prefer D(o)RY for tests for exactly that reason: when something goes wrong, I want to know exactly which one failed.

YannickJadoul:
Just my two - possibly useless, sorry in that case - cents: you can actually already achieve this by wrapping your function in a lambda that takes a reference (or std::reference_wrapper), though. I believe that's the C++11 semantics of a pointer vs. a reference?
I also ran into the need of not accepting a nullptr, but solved it that way. (I was even planning on maybe writing some function template to do it automatically.)

bmerry:
@YannickJadoul that works okay for plain pointers (although it's mildly inconvenient), but it doesn't work so well for copyable holder types like std::shared_ptr. There are probably workarounds (and if the class being wrapped derives from `enable_shared_from_this` it becomes a lot easier), but I don't know if there is a nice general solution.

jagerman:
@YannickJadoul - yes, it can be solved that way, even (I think) for a `shared_ptr` holder (i.e. make the lambda take a `shared_ptr` and see if its empty).  This is mainly about making that process a bit more convenient.

All of that said, I'm more or less indifferent w.r.t. this PR at the moment; that is, if @wjakob and @dean0x7d are opposed, I'm perfectly fine with dropping it.

YannickJadoul:
@bmerry @jagerman Hmmm, yes, I see; hadn't considered that, sorry.

Just as yet another possible alternative, have you considered making it somehow part of the holder type or having a tag for the type_casters? Somehow, it feels as if the `py::arg` arguments are handling the Python interface, while this is part of the C++ typing.
But then again, I guess you could argue that this is an extra restriction on the Python inferface, and it ís a rather elegant solution.

jagerman:
> Just as yet another possible alternative, have you considered making it somehow part of the holder type or having a tag for the type_casters?

That's how I approached this initially (see the description in the original PR post).  As @bmerry pointed out, though, whether `T *arg` can be given a nullptr is really up to the function rather than `T`.

YannickJadoul:
> That's how I approached this initially (see the description in the original PR post). As @bmerry pointed out, though, whether `T *arg` can be given a nullptr is really up to the function rather than `T`.

Yes, I'd read that, but I meant having 2 variants of the holder (i.e. `std::shared_ptr<T>` and `not_null<std::shared_ptr<T>>` or something like that, just like you have `T *` and `T &`). Then it would be part of the C++ type system and not of `py::arg`.
But I guess you'd constantly be wrapping your functions in new lambdas, so it's probably still a bad idea.

dean0x7d:
Well, I can't say I find segfaults in high-level C++ interfaces any more acceptable than in Python (and I'm not sure anyone will be using `None` as control flow in Python so that the exception must be `TypeError`), but if this saves people from writing lots of lambda wrappers then I guess I don't have any strong objection. 

@YannickJadoul A `gsl::not_null`-like class can be supported by adding a `type_caster` specialization which would just check for `None` and raise `TypeError`, or delagate to the appropriate subcaster (raw `T` or `Holder<T>`, doesn't matter).

YannickJadoul:
@dean0x7d Ok, yes, that's more or less what I meant, but I can now see that that's not the goal of this PR (but rather just a more convenient way of doing this), sorry.

wjakob:
I would be on the fence if it added a lot of complexity, but this is a really nice and simple patch. LGTM.

PR #861, opened 2017-05-17
Add comment in test_issues deprecating it

jagerman:
I think we decided a while ago (it was in a PR, but I forget which) to stop putting new things in `test_issues.cpp`; either because of that, or perhaps by chance, we haven't added anything there for a while.  I think deprecating it makes sense: there's always some better than a catch-all test file.

### DIFF
--- a/tests/test_issues.cpp
+++ b/tests/test_issues.cpp
@@ -1,11 +1,17 @@
 /*
-    tests/test_issues.cpp -- collection of testcases for miscellaneous issues
+    tests/test_issues.cpp -- collection of testcases for miscellaneous issues (DEPRECATED!)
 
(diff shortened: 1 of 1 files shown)


dean0x7d:
It would be nice to move the tests into their appropriate files and outright delete `test_issues.cpp`. I know that's a chore -- I can volunteer if needed.

jagerman:
+1 from me if you want to undertake it.  This was mainly a stop gap to save other people work from dropping new tests there then being told to move them.

jagerman:
If/when you undertake this, feel free to close this PR.

wjakob:
+1 from me as well to move the tests to their respective files (and I agree that it's a chore ;))

PR #862, opened 2017-05-17
Make static internals ptr pybind version specific

### LINKED ISSUE #796: Improve mixed-version handling
I upgraded pybind11 and rebuilt an extension, but there was another extension in an unrelated package/repo that was not rebuilt. When I imported both of them, the process segfaulted.

Can we add some runtime version checking to ensure that an newly imported module is compatible with any existing pybind11 version(s) that have already been loaded? It would be a better user experience to get an informative exception about mismatched versions rather than a segfault + gdb on a core file to figure out it's pybind11-related.

I'm not sure how strict we want to make the version checking, but this particular instance was 2.1.x vs 2.0.x, so minor versions arent compatible. Should we add an internal ABI version that evolves independent of the user-facing API version?

jagerman:
Under gcc, the `static internals *internals_ptr` is shared across .so's, which breaks for obvious reasons.

This commit fixes it by moving the static pointer declaration into a pybind-version-templated function.

Fixes #796.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -43,8 +43,15 @@ struct type_info {
     bool default_holder : 1;
 };
 
+// Store the static internals pointer in a version-specific function so that we're guaranteed it
+// will be distinct for modules compiled for different pybind11 versions.  Without this, some
+// compilers (i.e. gcc) can use the same static pointer storage location across different .so's,
+// even though the `get_internals()` function itself is local to each shared object.
+template <int = PYBIND11_VERSION_MAJOR, int = PYBIND11_VERSION_MINOR>
+internals *&get_internals_ptr() { static internals *internals_ptr = nullptr; return internals_ptr; }
+
 PYBIND11_NOINLINE inline internals &get_internals() {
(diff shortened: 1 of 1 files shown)


jagerman:
(This doesn't affect test suite .so size at all.)

dean0x7d:
It's strange that gcc does that. The fix looks good to me. Might be nice to add a comment above the new function template since its purpose is a bit mysterious out of context.

(As a side-note, placing `get_internals` inside an `inline` namespace with a version-macro-derived name could also be a solution, but more code than the template parameter version.)

jagerman:
Added a comment, and removed the unneeded "inline".

jagerman:
A bit of further investigation: gcc only seems to behave this way when *not* compiling with `-flto`.  Static locals then get exported as `STB_GNU_UNIQUE`, which makes them unique, even across the `RTLD_LOCAL` boundary.

With `-flto` this doesn't appear to happen.

jagerman:
Another way to fix this is to compile with `-fno-gnu-unique`.

PR #866, opened 2017-05-19
Use dynamic cast for shared_from_this holder init

### LINKED ISSUE #865: Compile error: std::enable_shared_from_this and virtual inheritance
#### Issue description

Compilation fails of the code below (gcc 6.3.1). The problem is that pybind attempts a static cast from Base* to Derived*, which is not possible due to virtual inheritance.

```bash
In file included from /usr/include/c++/6.3.1/memory:82:0,
                 from /home/matthias/src/pybind11/include/pybind11/common.h:133,
                 from /home/matthias/src/pybind11/include/pybind11/pytypes.h:12,
                 from /home/matthias/src/pybind11/include/pybind11/cast.h:13,
                 from /home/matthias/src/pybind11/include/pybind11/attr.h:13,
                 from /home/matthias/src/pybind11/include/pybind11/pybind11.h:39,
                 from /home/matthias/src/pybind11/tests/test_virtual_inheritance.cpp:1:
/usr/include/c++/6.3.1/bits/shared_ptr.h: In instantiation of ‘std::shared_ptr<_Tp1> std::static_pointer_cast(const std::shared_ptr<_Tp2>&) [with _Tp = Derived; _Tp1 = Base]’:
/home/matthias/src/pybind11/include/pybind11/pybind11.h:1122:105:   required from ‘static void pybind11::class_<type_, options>::init_holder_helper(pybind11::class_<type_, options>::instance_type*, const holder_type*, const std::enable_shared_from_this<_Tp1>*) [with T = Base; type_ = Derived; options = {std::shared_ptr<Derived>}; pybind11::class_<type_, options>::instance_type = pybind11::detail::instance<Derived, std::shared_ptr<Derived> >; typename pybind11::detail::exactly_one<pybind11::class_<type_, options>::is_holder, std::unique_ptr<_Tp>, options ...>::type = std::shared_ptr<Derived>; pybind11::class_<type_, options>::holder_type = std::shared_ptr<Derived>]’
/home/matthias/src/pybind11/include/pybind11/pybind11.h:1156:27:   required from ‘static void pybind11::class_<type_, options>::init_holder(PyObject*, const void*) [with type_ = Derived; options = {std::shared_ptr<Derived>}; PyObject = _object]’
/home/matthias/src/pybind11/include/pybind11/pybind11.h:938:28:   required from ‘pybind11::class_<type_, options>::class_(pybind11::handle, const char*, const Extra& ...) [with Extra = {}; type_ = Derived; options = {std::shared_ptr<Derived>}]’
/home/matthias/src/pybind11/tests/test_virtual_inheritance.cpp:13:70:   required from here
/usr/include/c++/6.3.1/bits/shared_ptr.h:447:35: error: cannot convert from pointer to base class ‘Base’ to pointer to derived class ‘Derived’ because the base is virtual
     { return shared_ptr<_Tp>(__r, static_cast<_Tp*>(__r.get())); }

```

```c++
#include <memory>

struct Base : public std::enable_shared_from_this<Base> {
    virtual ~Base() {}
};
struct Derived : public virtual Base{
    virtual ~Derived() {}
};

PYBIND11_PLUGIN(foo) {
    pybind11::module m("foo", "pybind foo");
    pybind11::class_<Derived, std::shared_ptr<Derived>> (m, "Derived");
    return m.ptr();
}

```
 #### Suggested fix
Use dynamic_cast instead of static_cast in init_holder_helper().

```diff
diff --git a/include/pybind11/pybind11.h b/include/pybind11/pybind11.h
index 098d7e0..560ac07 100644
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1119,7 +1119,7 @@ private:
     template <typename T>
     static void init_holder_helper(instance_type *inst, const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
         try {
-            new (&inst->holder) holder_type(std::static_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this()));
+            new (&inst->holder) holder_type(std::dynamic_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this()));
             inst->holder_constructed = true;
         } catch (const std::bad_weak_ptr &) {
             if (inst->owned) {

```

jagerman:
Using a `dynamic_cast` instead of a `static_cast` is needed to safely cast from a base to a derived type.  The previous `static_pointer_cast` isn't safe, however, when downcasting (and fails to compile when downcasting with virtual inheritance).

Switching this to always use a `dynamic_pointer_cast` shouldn't incur any additional overhead when a `static_pointer_cast` is safe (i.e. when upcasting, or self-casting): compilers don't need RTTI checks in those cases.

Fixes #865

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1119,13 +1119,15 @@ private:
     template <typename T>
     static void init_holder_helper(instance_type *inst, const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
         try {
-            new (&inst->holder) holder_type(std::static_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this()));
-            inst->holder_constructed = true;
-        } catch (const std::bad_weak_ptr &) {
-            if (inst->owned) {
-                new (&inst->holder) holder_type(inst->value);
+            auto sh = std::dynamic_pointer_cast<typename holder_type::element_type>(inst->value->shared_from_this());
+            if (sh) {
+                new (&inst->holder) holder_type(std::move(sh));
                 inst->holder_constructed = true;
             }
+        } catch (const std::bad_weak_ptr &) {}
+        if (!inst->holder_constructed && inst->owned) {
+            new (&inst->holder) holder_type(inst->value);
+            inst->holder_constructed = true;
         }
     }
 
--- a/tests/test_smart_ptr.cpp
+++ b/tests/test_smart_ptr.cpp
@@ -214,6 +214,12 @@ struct SharedFromThisRef {
     std::shared_ptr<B> shared = std::make_shared<B>();
 };
 
+// Issue #865: shared_from_this doesn't work with virtual inheritance
+struct SharedFromThisVBase : std::enable_shared_from_this<SharedFromThisVBase> {
+    virtual ~SharedFromThisVBase() = default;
+};
+struct SharedFromThisVirt : virtual SharedFromThisVBase {};
+
 template <typename T>
 class CustomUniquePtr {
     std::unique_ptr<T> impl;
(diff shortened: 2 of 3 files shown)


dean0x7d:
Looks good to me.

PR #867, opened 2017-05-21
Operator fixes

jagerman:
This fixes a couple issues in `operators.h`.

The first is that `/=` isn't being defined properly under Python 3 (`__idiv__` is set rather than `__itruediv__`) so it doesn't get called.  Instead Python converts `x /= 3` to `x = x / 3` and ends up create a new object via `__truediv__`, which probably works in some cases, but isn't correct.

The second issue (third commit) is that the `__truediv__` (and variants) are needed under Python 2 with `from __future__ import division`, otherwise division will fail (it does *not* fall back to calling an available `__div__`).  The fix here adds a second binding under the alternate names under Python 2.


dean0x7d:
LGTM!

PR #875, opened 2017-05-26
Correct error in numpy.rst

chenzeyuczy:
When follow the instruction in [http://pybind11.readthedocs.io/en/master/advanced/pycpp/numpy.html] , error occurs as following. 

> error:  ‘static_cat’ was not declared in this scope

This PR correct the mistake in documentation.


jagerman:
Good catch, thanks!

PR #879, opened 2017-05-28
Replace PYBIND11_PLUGIN with PYBIND11_MODULE

dean0x7d:
As discussed in #774, this PR changes the module initialization macro from:

```c++
PYBIND11_PLUGIN(example) {
    pybind11::module m("example", "pybind11 example plugin");
    m.def("add", [](int a, int b) { return a + b; });
    return m.ptr();
}
```
to:
```c++
PYBIND11_MODULE(example, m) {
    m.def("add", [](int a, int b) { return a + b; });
}
```

Notes:

* Using the old macro results in a deprecation warning. The warning actually points to the `pybind11_init` function (since attributes don't bind to macros), but the message should be quite clear: "PYBIND11_PLUGIN is deprecated, use PYBIND11_MODULE".
* The docs have been updated to reflect the change. This makes one less entry in the FAQ.
* `doc()` is implemented as a simple shortcut for the `attr("__doc__")` accessor.

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -194,6 +194,8 @@ extern "C" {
     PYBIND11_TOSTRING(PYBIND11_VERSION_MAJOR) "_" PYBIND11_TOSTRING(PYBIND11_VERSION_MINOR) "__"
 
 /** \rst
+    ***Deprecated in favor of PYBIND11_MODULE***
+
     This macro creates the entry point that will be invoked when the Python interpreter
     imports a plugin library. Please create a `module` in the function body and return
     the pointer to its underlying Python object at the end.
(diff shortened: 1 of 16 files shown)


wjakob:
LGTM!

include/pybind11/common.h
```diff
@@ -234,6 +237,54 @@ extern "C" {
...
+            m.def("foo", []() {
+                return "Hello, World!";
+            });
+        }
+\endrst */
+#define PYBIND11_MODULE(name, variable)                                        \
+    static void pybind11_init_##name(pybind11::module &);                      \
+    PYBIND11_PLUGIN_IMPL(name) {                                               \
+        int major, minor;                                                      \
+        if (sscanf(Py_GetVersion(), "%i.%i", &major, &minor) != 2) {           \
```
jagerman: This is really minor nitpicking, but this should probably be `std::sscanf`.  (Maybe not worth bothering with, since we know the C header is going to have been included via Python).

dean0x7d: Looks like `Python.h` is guaranteed to include `<stdio.h>` (along with a few other standard headers) so we get `sscanf` anyway. Given that `std::sscanf` needs `<cstdio>`, I'd rather keep the plain version just to avoid cluttering the includes.

PR #885, opened 2017-05-30
Allowed multiple types in brace-enclosed initializer list constructor of any_container

Lunderberg:
With the previous templated initializer_list<TIn>, TIn could only be deduced if all values had the same type.  By accepting initializer_list<T> instead, arguments passed in can be of different types, so long as each type can be converted to type T.

As tested on my machine, this passes all unit tests, along with a new one added to test the new behavior.  Without the new change, the new test fails to compile.  The change also causes a large number of `-Wnarrowing` warnings, calling out the conversions from `unsigned long int` to `long int`.  I'm not sure how best to fix this, other than changing all of the call sites.  I have currently left them uncorrected, in order to keep the pull request small.

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -755,8 +755,7 @@ public:
 
     // initializer_list's aren't deducible, so don't get matched by the above template; we need this
     // to explicitly allow implicit conversion from one:
-    template <typename TIn, typename = enable_if_t<std::is_convertible<TIn, T>::value>>
-    any_container(const std::initializer_list<TIn> &c) : any_container(c.begin(), c.end()) { }
+    any_container(const std::initializer_list<T> &c) : any_container(c.begin(), c.end()) { }
 
     // Avoid copying if given an rvalue vector of the correct type.
     any_container(std::vector<T> &&v) : v(std::move(v)) { }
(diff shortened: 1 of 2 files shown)


dean0x7d:
Those aren't narrowing warnings, they're outright errors because of the C++11 brace initializer rules. Removing the template constructor causes the narrowing errors to also pop up for existing same-type initializers. I'm not sure that there is a good solution to handle these narrowing issues with mixed types. The existing implementation managed to get around them, at least for same-type initializers, so I'd be in favor of keeping that.

It seems to me that the mixed-type initializer from the test could be handled easily on the user side:
```c++
{ points.size(), size_t{3} }
```

Lunderberg:
Ah, whoops.  Looks like gcc is in the minority with narrowing in C++11 being only a warning, rather than an error, so this is not as promising of a solution as I had thought.  Use of mixed types with brace initialization in `buffer_info` would compile up until 5f3838629.  I have since updated a package that this broke, but I was hoping to prevent others from having a similar issue.  It looks like it is more complicated than I had thought.

Attempts to rewrite the change so that `buffer_info` can be made from a variadic template instead of an initializer_list cause ambiguous function overloads elsewhere, so I don't think there is anything to do for it at the moment.  Thank you.

PR #888, opened 2017-06-05
Explicitly define move constructor for accessor

eirrgang:
I suppose it depends on when template instantiation occurs or something, but in my case,
I want to avoid the following compiler warning.

```
pybind11/include/pybind11/pytypes.h:385:10: warning: definition of implicit copy constructor for 'accessor<pybind11::detail::accessor_policies::str_attr>' is deprecated because it has a
      user-declared copy assignment operator [-Wdeprecated]
    void operator=(const accessor &a) && { std::move(*this).operator=(handle(a)); }
         ^
pybind11/include/pybind11/pybind11.h:1480:18: note: implicit copy constructor for 'accessor<pybind11::detail::accessor_policies::str_attr>' first required here
    auto write = file.attr("write");
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -393,6 +393,7 @@ public:
     // to replace default compiler-generated assignments).
     void operator=(const accessor &a) && { std::move(*this).operator=(handle(a)); }
     void operator=(const accessor &a) & { operator=(handle(a)); }
+    accessor(accessor &&a) = default;
 
     template <typename T> void operator=(T &&value) && {
         Policy::set(obj, key, object_or_cast(std::forward<T>(value)));
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -393,6 +393,7 @@ public:
     // to replace default compiler-generated assignments).
     void operator=(const accessor &a) && { std::move(*this).operator=(handle(a)); }
     void operator=(const accessor &a) & { operator=(handle(a)); }
+    accessor(accessor &a) = default;
     accessor(accessor &&a) = default;
 
     template <typename T> void operator=(T &&value) && {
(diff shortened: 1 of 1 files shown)


jagerman:
Looks like that caused many other test suite build failures.

Do you have a compact bit of sample code that reproduces the warning?

include/pybind11/pytypes.h
```diff
@@ -393,6 +393,7 @@ class accessor : public object_api<accessor<Policy>> {
     // to replace default compiler-generated assignments).
     void operator=(const accessor &a) && { std::move(*this).operator=(handle(a)); }
     void operator=(const accessor &a) & { operator=(handle(a)); }
+    accessor(accessor &a) = default;
```
jagerman: Missing `const`

eirrgang:
Unfortunately, it wasn't clear to me why the warning didn't always appear. But did all of the tests fail for the same reason? I think the problem for the CI tests was that declaring the move constructor caused the copy constructor to be implicitly deleted. Is default copy and move construction appropriate? Or does handle(a) have different different semantics w.r.t. data members?

include/pybind11/pytypes.h
eirrgang: oops. Thanks.

jagerman:
Default move and copy ctors should be fine.  The class doesn't currently provide the move constructor (since there's a copy constructor), but I don't see a problem with adding one, and relying on a default copy constructor is indeed C++11-deprecated.  Strange that the warning hasn't come up before—perhaps copy elision results in the compiler not hitting the warning?

Anyway, I think this looks good, assuming all the tests pass.

eirrgang:
I just realized that some code I compile with c++11 and some with c++14. There are some differences in copy elision between the two standards, so maybe that's it.

jagerman:
Merged.  Thanks for the PR/report!

PR #891, opened 2017-06-06
Fix passing in utf8 encoded strings with python 2

benfred:
Passing utf8 encoded strings from python to a C++ function taking a std::string
was broken.  The previous version was trying to call 'PyUnicode_FromObject'
on this data, which failed to convert the string to unicode with the default
ascii codec. Also this incurs an unnecessary conversion to unicode for data
this is immediately converted back to utf8.

Fix by treating python 2 strings the same python 3 bytes objects, and just
copying over the data if possible.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -745,15 +745,14 @@ struct type_caster<std::basic_string<CharT, Traits, Allocator>, enable_if_t<is_s
 #if PY_MAJOR_VERSION >= 3
             return load_bytes(load_src);
 #else
+            if (sizeof(CharT) == 1) {
+                return load_bytes(load_src);
+            }
+
             // The below is a guaranteed failure in Python 3 when PyUnicode_Check returns false
             if (!PYBIND11_BYTES_CHECK(load_src.ptr()))
                 return false;
 
-            // try just loading directly from object
-            if (load_bytes(load_src))
-                return true;
-
-            // otherwise convert to unicode (if sizeof(CharT) > 1)
             temp = reinterpret_steal<object>(PyUnicode_FromObject(load_src.ptr()));
             if (!temp) { PyErr_Clear(); return false; }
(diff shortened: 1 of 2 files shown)


jagerman:
This looks reasonable to me.  A couple comment:

- `load_bytes` is obviously pointless when `sizeof(CharT) > 1`.  Is `PyUnicode_FromObject` conversely pointless when `sizeof(CharT) == 1`?  I'm not 100% sure, but I *think* it is, in which case it might be clearer to write to code as: `if (sizeof(CharT) == 1) { if (load_bytes(...)) return true; } else { PyUnicode_FromObject(...) }` so that the compiler can optimize away the dead code path.  (It also makes the code self-documenting, rather than putting the "else" logic inside a comment).
- the current code calls `PYBIND11_BYTES_CHECK` (i.e. `PyBytes_Check`) twice—once immediately before then again inside `load_bytes()`.  Since `load_bytes()` is going to succeed in the common case, I think you could clean up the logic slightly more to have something like: `if (sizeof(CharT) == 1 && load_bytes(...)) return true; else if (sizeof(CharT) > 1 && PYBIND11_BYTES_CHECK(...)) { ... } else { return false }`
- 💩 seems a totally apt unicode character choice for code dealing with unicode encoding issues under Python 2 :)

benfred:
Awesome! Thanks for the feedback.

For your first point - reading the documentation for ```PyUnicode_FromObject``` it does seem pointless when ```sizeof(CharT)``` is 1.  I've updated to just return the results of load_bytes in that case.

I've also updated so that it doesn't call ```PYBIND11_BYTES_CHECK``` twice when ```sizeof(CharT)``` is 1.

jagerman:
Merged this (in 74b501c, outside as github as I needed to squash and rebase).

jagerman:
Thanks for the work!

PR #892, opened 2017-06-07
Destroy internals if created during Py_Finalize()

jagerman:
`Py_Finalize` could potentially invoke code that calls `get_internals()`,
which could create a new internals object if one didn't exist.
`finalize_interpreter()` didn't catch this because it only used the
pre-finalize interpreter pointer status; if this happens, it results in
the internals pointer not being properly destroyed with the interpreter,
which leaks, and also causes a `get_internals()` under a future
interpreter to return an internals object that is wrong in various ways.

The two lines added to `embed.h` are the fix; the rest is just adding a test case that triggers and detects the issue.

### DIFF
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -143,7 +143,11 @@ inline void finalize_interpreter() {
     handle builtins(PyEval_GetBuiltins());
     const char *id = PYBIND11_INTERNALS_ID;
 
-    detail::internals **internals_ptr_ptr = nullptr;
+    // Get the internals pointer (without creating it if it doesn't exist).  It's possible for the
+    // internals to be created during Py_Finalize() (e.g. if a py::capsule calls `get_internals()`
+    // during destruction), so we get the pointer-pointer here and check it after Py_Finalize().
+    detail::internals **internals_ptr_ptr = &detail::get_internals_ptr();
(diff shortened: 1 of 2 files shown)


jagerman:
Cc @dean0x7d for confirmation that the fix looks good.

include/pybind11/embed.h
```diff
@@ -149,6 +149,12 @@ inline void finalize_interpreter() {
 
     Py_Finalize();
 
+    // If something in Py_Finalize() calls get_internals() the pointer could be created, even if it
+    // didn't exist above.  (But we still use the above because if nothing calls get_internals() we
+    // might not have the static pointer set in get_internals_ptr())
+    if (!internals_ptr_ptr)
+        internals_ptr_ptr = &detail::get_internals_ptr();
```
dean0x7d: Good catch! I'd suggest moving up `get_internals_ptr()` to mirror the code from `detail:: get_internals()`. This is invariant with respect to `Py_Finalize()`:

```c++
auto *&internals_ptr = detail::get_internals_ptr();
if (builtins.contains(id) && isinstance<capsule>(builtins[id]))
    internals_ptr = *static_cast<internals **>(capsule(builtins[id]));

Py_Finalize();

delete internals_ptr;
internals_ptr = nullptr;
```

PR #893, opened 2017-06-08
Enable detection of private operator new on MSVC

dean0x7d:
MSVC 2015 Update 3 and 2017 can handle enough expression SFINAE to make this work now.


jagerman:
Looks good, and :+1: for moving another test of out `issues`.

dean0x7d:
I'm cleaning out everything in `test_issues` now (hopefully done soon). This was the only feature change that came out of it. It's pretty nice not to have to worry about MSVC older than 2015u3.

PR #897, opened 2017-06-09
Fix py::make_iterator's __next__() for past-the-end calls

### LINKED ISSUE #896: Attempting to iterate over python iterable, in c++, iterates past the end with bound c++ types.
### Issue
Attempting to iterate over python iterable, in c++, iterates past the end with bound c++ types. It seems that comparing the end iterator to the sentinel attempts to advance the end iterator past the end. When iterating a python type calling PyIter_Next past the end simply returns NULL however for a bound STL container this invokes undefined behaviour.

Occurs on:
Windows 10 (VS 2017), Ubuntu 17.04 (gcc 7.0)
pybind11 2.1.1 and master ( 0365d49 )

### Example
cpp:
```cpp
#include <pybind11/stl_bind.h>

namespace py = pybind11;

PYBIND11_MAKE_OPAQUE(std::vector<int>)

PYBIND11_MODULE(IterableTest, m)
{

    m.def("iter_test", [] (py::iterable iterable)
    {
        auto iter = py::iter(iterable);
        while (iter != py::iterator::sentinel())
        {
            py::print("got value: ", *iter);
            ++iter;
        }
    });
}
```
python:
```python
from IterableTest import *

ivec = VectorInt()
ivec.append(3)
ivec.append(2)
ivec.append(1)

iter_test(ivec)
```
example output:
```
got value:  3
got value:  2
got value:  1
*debug assert iterator is not incrementable*
```

dean0x7d:
Fixes #896.

From Python docs: Once an iterator’s `__next__()` method raises `StopIteration`, it must continue to do so on subsequent calls. Implementations that do not obey this property are deemed broken.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1353,7 +1353,7 @@ template <typename Iterator, typename Sentinel, bool KeyIterator, return_value_p
 struct iterator_state {
     Iterator it;
     Sentinel end;
-    bool first;
+    bool first_or_done;
 };
 
 NAMESPACE_END(detail)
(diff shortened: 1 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -1378,8 +1378,10 @@ iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
                     ++s.it;
                 else
                     s.first = false;
-                if (s.it == s.end)
+                if (s.it == s.end) {
+                    s.first = true;
```
jagerman: It was a bit non-obvious to me at first glance why this works, which I think may have been due to the `first` name; a rename of `first` to something like `first_or_done` would be more self-documenting.

jagerman:
LGTM, with one minor comment about the variable name being slightly confusing.

include/pybind11/pybind11.h
dean0x7d: Good point. Renamed.

PR #900, opened 2017-06-12
Upgrade pypy to 5.8, use pre-built numpy/scipy wheels

jagerman:
numpy 1.13.0 fails to install properly with pypy 5.7.1, so this upgrades pypy to the latest 5.8.0, which can build numpy successfully.

I've also uploaded pre-built pypy numpy and scipy wheels to imaginary.ca (checked every 2 hours and rebuilt if needed), and list that repository as an extra pypi location under the pypy pip install to avoid the long travis pypy build times for a new numpy/scipy release, new branch, or cache-cleared build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
     env: PYTHON=3.6 CPP=14 CLANG
   # Test a PyPy 2.7 build
   - os: linux
-    env: PYPY=5.7 PYTHON=2.7 CPP=11 GCC=4.8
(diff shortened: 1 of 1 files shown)


dean0x7d:
The pre-built wheels are a great idea! LGTM.

jagerman:
I've stashed the build scripts [here](https://github.com/jagerman/pypy-numpy).  A cron job runs `./check-for-updates.py` every 2 hours.

PR #902, opened 2017-06-13
add the capsule name to the py::capsule constructor

phaustin:
Here's a first attempt at adding a name to py::capsule

I've only changed the first constructor, because trying to add an optional name to the other two constructors produces an error message:

    ValueError: PyCapsule_GetPointer called with incorrect name

### DIFF
--- a/tests/test_python_types.cpp
+++ b/tests/test_python_types.cpp
@@ -578,7 +578,7 @@ test_initializer python_types([](py::module &m) {
             py::print("creating capsule");
             auto capsule=py::capsule((void *) 1234, "pointer type description",
                 [](PyObject *ptr) {
-                 if (ptr){
+                 if (ptr) {
                     py::print("destructing capsule");
                  }
             });
--- a/tests/test_python_types.cpp
+++ b/tests/test_python_types.cpp
@@ -578,7 +578,7 @@ test_initializer python_types([](py::module &m) {
             py::print("creating capsule");
             auto capsule=py::capsule((void *) 1234, "pointer type description",
                 [](PyObject *ptr) {
-                 if(ptr){
+                 if (ptr){
                     py::print("destructing capsule");
                  }
             });
@@ -588,7 +588,7 @@ test_initializer python_types([](py::module &m) {
             return capsule;
         }
     );
-    
+
     m.def("load_nullptr_t", [](std::nullptr_t) {}); // not useful, but it should still compile
     m.def("cast_nullptr_t", []() { return std::nullptr_t{}; });
 
--- a/tests/test_python_types.cpp
+++ b/tests/test_python_types.cpp
@@ -575,6 +575,7 @@ test_initializer python_types([](py::module &m) {
 
     m.def("return_capsule_with_name_and_destructor_3",
         []() {
+            py::print("creating capsule");
             auto capsule=py::capsule((void *) 1234, "pointer type description",
                 [](PyObject *ptr) {
                  if(ptr){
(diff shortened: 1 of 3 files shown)


tests/test_python_types.cpp
```diff
@@ -575,15 +575,18 @@ test_initializer python_types([](py::module &m) {
...
         []() {
-            py::print("creating capsule");
             auto capsule=py::object(py::capsule((void *) 1234, "pointer type description",
-                [](PyObject *ptr) {py::print("destructing capsule");}));
+                [](PyObject *ptr) {
+                 if(ptr){
+                    py::print("destructing capsule");
+                 }
+            }));
             auto name=PyCapsule_GetName(capsule.ptr());
```
jagerman: We should provide enough of the interface here that we can avoid hitting the internals, particularly because the test suite also serves as a source of examples.  Let's add a simple `const char *name() const { return PyCapsule_GetName(m_ptr); }` method to get it.

tests/test_python_types.cpp
```diff
@@ -575,15 +575,18 @@ test_initializer python_types([](py::module &m) {
...
-            py::print(name);
+            auto contents=PyCapsule_GetPointer(capsule.ptr(),name);
```
jagerman: Here likewise you should be able to use the pybind API: `void *contents = capsule;`

tests/test_python_types.cpp
```diff
@@ -575,15 +575,18 @@ test_initializer python_types([](py::module &m) {
 
     m.def("return_capsule_with_name_and_destructor_3",
         []() {
-            py::print("creating capsule");
             auto capsule=py::object(py::capsule((void *) 1234, "pointer type description",
```
jagerman: Is there a reason to wrap the `capsule` inside an `object` here?  `capsule` is already an `object` subclass.

tests/test_python_types.cpp
phaustin: Fixed hopefully  -- but the pytest is failing for some reason:

    >       assert capture.unordered == """
                created capsule with name --pointer type description-- and contents 1234
                creating capsule
                destructing capsule
            """
    E           created capsule with name --pointer type description-- and contents 1234
    E         + creating capsule
    E           destructing capsule

not sure why it's unhappy.

jagerman: It looks like you removed the `py::print("creating capsule");` but forgot to take that line out of the expected output in the test.

phaustin: ok, I misunderstood the pytest output, added the line back in.

jagerman:
This looks good to me; any other comments?

phaustin:
Not directly related, but as long as I'm here, a note that my /bin/bash on OSX Sierra won't run tools/check-style.sh

   $ ./tools/check-style.sh 
    ./tools/check-style.sh: line 22: 3: ambiguous redirect
    ./tools/check-style.sh: line 23: read: 3: invalid file descriptor: Bad file descriptor
   ...

works fine on Ubuntu though.

best, Phil

phaustin:
I'll be putting together a Jupyter notebook to demonstrate how to use py::capsule and xtensor-python to pass functions to an ode integrator, will ping you when I post it to my class.

jagerman:
> /bin/bash on OSX Sierra won't run tools/check-style.sh

Thanks for the report; that should be fixed by d080f833beb7e18944961d523d408c524b70001a

jagerman:
Merged; thanks for the PR!

PR #911, opened 2017-06-20
def_readwrite/readonly: bind using class type, not inferred type

### LINKED ISSUE #910: Exposing base member from derived, when base is not wrapped
If I have a base class `B` with a data member `x` (of type `float`, but that is irrelevant) and a derived class `D`. And if I now expose only `D` (because `B` for some other reason should not be exposed) as: `py::class_<D>(m, "D").def(py::init<>()).def_readonly("x", &D::x);` Accessing `x` from Python fails with:

```
E TypeError: (): incompatible function arguments. The following argument types are supported:
E 1. (self: B) -> float
E
E Invoked with: <pybind11_tests.D object at 0x108e2e538>

```

This can be fixed with an explicit cast as in `.def_readonly("x", (float D::*) &D::x);`, but that should not really be required (as pointed out my @jagerman on Gitter: Pybind doesn't know anything about the `D : B` relationship here, which is why it happens. It ought to be possible to make it work properly (and `static_assert` fail when the member isn't a superclass member)).

jagerman:
Current `def_readwrite`/`def_readonly` create getter/setter lambdas that
take the inferred type `C` as the first argument, but this is a problem
if the member is actually a base class member (e.g.

    derived.def_readonly("x", &Base::x)

because this ends up creating a setter/getter that aren't callable with
a derived instance when `Base` isn't registered: it invariably produces
an invalid argument failure when attempting to access the property due
to the `D` instance/`B` argument mismatch.

This commit changes the generated lambdas to use the `type` (as given in
`class_<type>`) rather than the inferred `C` for the lambdas, and adds a
static_assert failure for an attempt to call it with a member that isn't
part of `type` (or a base class).

Fixes #910

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1042,15 +1042,17 @@ public:
 
     template <typename C, typename D, typename... Extra>
     class_ &def_readwrite(const char *name, D C::*pm, const Extra&... extra) {
-        cpp_function fget([pm](const C &c) -> const D &{ return c.*pm; }, is_method(*this)),
-                     fset([pm](C &c, const D &value) { c.*pm = value; }, is_method(*this));
+        static_assert(std::is_base_of<C, type>::value, "def_readwrite() requires a class member (or base class member)");
+        cpp_function fget([pm](const type &c) -> const D &{ return c.*pm; }, is_method(*this)),
+                     fset([pm](type &c, const D &value) { c.*pm = value; }, is_method(*this));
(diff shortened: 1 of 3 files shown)


bmerry:
Isn't this already fixed by #855?

jagerman:
Yes, if #855 is merged it fixes this as well.  There is some concern than #855 is too complicated, however, so if we decide to drop it, this is a partial alternative.

jagerman:
I've simplified #855; closing this in favour of it.

Issue #912, opened 2017-06-21
Typeid information not synchronized across shared library boundaries


wjakob:
Dear all,

I've been running into strange inconsistencies where type information is not synchronized across shared library boundaries. This happens when using Clang & libc++ on Linux/x86_64 machines.

Before escalating to Clang/LLVM, I wanted to gauge if anyone here has previously run into this. To reproduce:

issue.py
```
import mod1, mod2

mod2.test(mod1.Test())

# Expected output
# ================
# $ python issue.py
# $


# Actual output:
# ================
# $ python issue.py
# Traceback (most recent call last):
#   File "issue.py", line 3, in <module>
#     mod2.test(mod1.Test())
# TypeError: test(): incompatible function arguments. The following argument types are supported:
#     1. (arg0: Test) -> None
#
# Invoked with: <mod1.Test object at 0x7f12861f5df8>
```

CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 2.8.12)
project(example)

# Issue is triggered in a release build
set(CMAKE_BUILD_TYPE Release CACHE STRING "" FORCE)

# Issue is triggered when using libc++
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")

add_subdirectory(pybind11)

add_library(shared SHARED shared.cpp)

pybind11_add_module(mod1 mod1.cpp)
pybind11_add_module(mod2 mod2.cpp)

target_link_libraries(mod1 shared)
target_link_libraries(mod2 shared)
```

shared.h
```cpp
#pragma once

#define EXPORT __attribute__ ((visibility("default")))

class EXPORT Test {
public:
    Test();
};
```

shared.cpp
```cpp
#include "shared.h"

Test::Test() { }
```

mod1.cpp
```cpp
#include <pybind11/pybind11.h>
#include "shared.h"

namespace py = pybind11;

PYBIND11_MODULE(mod1, m) {
    py::class_<Test>(m, "Test")
        .def(py::init<>());
}
```

mod2.cpp
```cpp
#include <pybind11/pybind11.h>
#include "shared.h"

namespace py = pybind11;

void test(Test b) { }

PYBIND11_MODULE(mod2, m) {
    py::module::import("mod1");
    m.def("test", &test);
}
```

Best,
Wenzel

aldanor:
Hm, is `typeid(Test)` not the same in mod1/mod2, can you check?

wjakob:
[...]

PR #914, opened 2017-06-21
Avoid std::type_info pointer comparison

jagerman:
Type equality needs to be done with `operator==` on the
`std::type_info` object; we do this in some places, but in a few places
use pointer equality (i.e. the equivalent of &typeid(A) == &typeid(B)).
This replaces all such cases with std::type_info equality comparisons.

I noticed this in investigation of #912, but it does *not* fix that issue.


jagerman:
Closed in favour of #915.

PR #915, opened 2017-06-21
Use std::type_info::name() for type lookups outside stdlibc++

### LINKED ISSUE #912: Typeid information not synchronized across shared library boundaries

jagerman:
Using `std::type_info::operator==` fails under libc++ because the .so
is loaded with RTLD_LOCAL.  libc++ considers types under such .sos
distinct, and so comparing typeid() values directly isn't going to work.

This adds a custom hasher and equality class for the type lookup maps
when not under stdlibc++, and adds a `detail::same_type` function to
perform the equality test.  It also converts a few pointer arguments to
const lvalue references, particularly since doing the pointer
comparison wasn't technically valid to begin with (though in practice,
appeared to work everywhere).

This fixes #912, and closes #914 (it supersedes that one).

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -243,17 +243,17 @@ struct type_record {
     /// Is the default (unique_ptr) holder type used?
     bool default_holder : 1;
 
-    PYBIND11_NOINLINE void add_base(const std::type_info *base, void *(*caster)(void *)) {
-        auto base_info = detail::get_type_info(*base, false);
+    PYBIND11_NOINLINE void add_base(const std::type_info &base, void *(*caster)(void *)) {
+        auto base_info = detail::get_type_info(base, false);
         if (!base_info) {
-            std::string tname(base->name());
+            std::string tname(base.name());
             detail::clean_type_id(tname);
             pybind11_fail("generic_type: type \"" + std::string(name) +
                           "\" referenced unknown base type \"" + tname + "\"");
         }
 
         if (default_holder != base_info->default_holder) {
-            std::string tname(base->name());
+            std::string tname(base.name());
             detail::clean_type_id(tname);
             pybind11_fail("generic_type: type \"" + std::string(name) + "\" " +
                     (default_holder ? "does not have" : "has") +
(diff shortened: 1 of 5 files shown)


jagerman:
(I kept this separate from #914 because it we *don't* want to merge this, we should at least merge the #914 changes).

wjakob:
You beat me to it! :) -- This looks great. The PR resolves all typeid issues on my Linux/x86_64 box.

wjakob:
Any objections to merging this? @dean0x7d?

dean0x7d:
No objections from me. Looks good!

PR #917, opened 2017-06-22
Allow type_casters to expose temporary dependencies

### LINKED ISSUE #916: Matrix conversion and variant issue
Hi!

I came across the following issue working with variants of Eigen matrices (which is an amazing feature, by the way):

## Issue description

When calling the function, parameterized with a variant over Eigen const matrix references with an Eigen matrix over a certain size that needs to be converted, I get a segmentation fault. This works flawlessly with smaller matrices. I do not suspect the variant implementation is the problem, because if I shift the conversion out of pybind11 and do it myself in a c++ function, it works flawlessly, also it works if the matrix doesn't have to be converted.

## Reproducible example code

example.py:

```
#!/usr/bin/env python2
import example
import numpy as np

# Works.
example.test_v(np.ones((30, 10))[:, (7, 3, 9, 5)])

# Segmentation fault (conversion because of stride).
#example.test_v(np.ones((500000, 10))[:, (7, 3, 9, 5)])
# Segmentation fault (conversion because of dtype).
for i in range(71475):
    print i
    # Fails for me at i = 71474, i.e., 71475 is the first with too much data.
    example.test_v(np.ones((i+1,), dtype=np.uint8))

# Raises correctly.
# example.test_v_noconv(np.ones((500000, 10))[:, (7, 3, 9, 5)])
```

CMakeLists.txt:

```
cmake_minimum_required(VERSION 3.2)
project(example)

add_subdirectory(pybind11-master)
include_directories (eigen3-v3.3)
include_directories (variant-v1.1.4/include)

pybind11_add_module(example MODULE example.cpp)

add_custom_command (TARGET example POST_BUILD
  COMMAND "${CMAKE_COMMAND}" -E copy
     "$<TARGET_FILE:example>"
     "${PROJECT_SOURCE_DIR}/$<TARGET_FILE_NAME:example>"
  COMMENT "Copying to root directory")
```

example.cpp
```
#include <pybind11/eigen.h>
#include <pybind11/stl.h>
#include <Eigen/Dense>
#include <mapbox/variant.hpp>

namespace py = pybind11;
namespace mu = mapbox::util;

// Register the variant.

    template <typename... Ts>
    struct type_caster<mapbox::util::variant<Ts...>> :
    variant_caster<mapbox::util::variant<Ts...>> {};

    template <>
    struct visit_helper<mapbox::util::variant> {
      template <typename... Args>
        -> decltype(mapbox::util::apply_visitor(std::forward<Args>(args)...)) {
        return mapbox::util::apply_visitor(std::forward<Args>(args)...);
      }
    };

} }  // namespace detail  namespace pybind11

template <typename DT>
using Mat = Eigen::Matrix<DT,
                          Eigen::Dynamic,
                          Eigen::Dynamic,
                          Eigen::RowMajor>;

template <typename DT>
using MatCRef = Eigen::Ref<const Mat<DT>>;

struct Empty{};

PYBIND11_MODULE(example, m) {
  m.doc() = "Example module";

  m.def("test_v", [](const mu::variant<Empty,
                                       MatCRef<float>,
                                       MatCRef<double>> &val_v) {
          val_v.match([&](const auto &val) {
              std::cout << val(0, 0) << std::endl;
            },
            [](const Empty &) {});
        });

  m.def("test_v_noconv", [](const mu::variant<Empty,
                                              MatCRef<float>,
                                              MatCRef<double>> &val_v) {
          val_v.match([&](const auto &val) {
              std::cout << val(0, 0) << std::endl;
            },
            [](const Empty &) {});
        },
        py::arg().noconvert());

}
```

This assumes that `pybind11-master`, `eigen3-v3.3` and `variant-v1.1.4` are subfolders, variant can be downloaded from [here](https://github.com/mapbox/variant), but probably `std::variant` can be used in the same way. If `.noconvert()` is used, the function works just as expected, which I use as a workaround.

Another probably only slightly related issue I came across in this setting is the `keep_alive` policy, which also doesn't work for these kind of converted arguments (I assume the original object that is passed to the variant is kept alive, but the converted object that the Eigen::Ref points to in the end is not). I'm not 100% sure on this one, though, and may open a dedicated issue once I have a proper test case.

jagerman:
The `Eigen::Ref` caster, in particular, requires that a temporary array be
kept alive in some cases.  This allows a type caster to expose a
keep-alive dependency by providing a `handle subcaster_keepalive()`
method so that stl.h casters (which employ temporary subcasters during
loading) can hold onto any required extra dependencies.

Fixes #916

### DIFF
--- a/include/pybind11/eigen.h
+++ b/include/pybind11/eigen.h
@@ -425,6 +425,9 @@ private:
     // storage order conversion.  (Note that we refuse to use this temporary copy when loading an
     // argument for a Ref<M> with M non-const, i.e. a read-write reference).
     Array copy_or_ref;
+    // Set when copy_or_ref is a copy, in which case subcaster_keepalive() returns it to keep the
+    // temporary copy alive when this caster is invoked by a subcaster.
+    bool is_copy;
 public:
     bool load(handle src, bool convert) {
         // First check whether what we have is already an array of the right type.  If not, we can't
@@ -462,6 +465,7 @@ public:
             if (!fits || !fits.template stride_compatible<props>())
                 return false;
             copy_or_ref = std::move(copy);
+            is_copy = true;
         }
 
         ref.reset();
@@ -470,6 +474,8 @@ public:
 
         return true;
     }
+
+    handle subcaster_keepalive() { return is_copy ? (handle) copy_or_ref : handle(); }
 
     operator Type*() { return ref.get(); }
     operator Type&() { return *ref; }
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -49,6 +49,27 @@
 NAMESPACE_BEGIN(pybind11)
 NAMESPACE_BEGIN(detail)
 
+template <typename caster_t, typename = void>
+struct subcaster_keepalive_impl { static handle get(caster_t &) { return {}; } };
+template <typename caster_t>
+struct subcaster_keepalive_impl<caster_t, void_t<decltype(std::declval<caster_t>().subcaster_keepalive())>> {
+    static handle get(caster_t &caster) { return caster.subcaster_keepalive(); } };
+template <typename caster_t>
+
+// Calls caster.subcaster_keepalive() if the caster has it, otherwise returns an empty handle
+inline handle subcaster_keepalive(caster_t &caster) { return subcaster_keepalive_impl<caster_t>::get(caster); }
+
+
+
+template <typename caster_t>
+void append_subcaster_keepalive(object &deps, caster_t &caster) {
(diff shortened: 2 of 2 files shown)


dean0x7d:
Perhaps this might work as an alternative: make the temporary object a patient of the original `handle src` which is being loaded (the simpler `weakref` variant of the `keep_alive` mechanism should be enough here). That should ensure correct lifetime with the upside that supercasters would not need to be aware of the extra dependencies. (This should also work for the `string_view` caster and remove the extra `object` from the struct.)

jagerman:
Closing in favour of #924.

PR #918, opened 2017-06-23
Fix linker issue with move constructors on MSVC

dean0x7d:
Fixes the issue as described in the comments of commit e27ea47. 

This just adds `enable_if_t<std::is_move_constructible<T>::value>` to `make_move_constructor`. The change fixes MSVC and is harmless with other compilers. A few line breaks are also added to make the functions easier to read.

@r-englund Can you verify that this fixes the issue for you?


dean0x7d:
Thanks for the confirmation @r-englund.

PR #920, opened 2017-06-23
Fix GIL release and acquire when embedding the interpreter

dean0x7d:
This fixes a race condition when multiple threads try to acquire the GIL before `detail::internals` have been initialized. `gil_scoped_release` is now tasked with initializing `internals` (guaranteed single-threaded) to ensure the safety of subsequent `acquire` calls from multiple threads.

An alternative would be to have `py::initialize_interpreter()` do this, but strictly single-threaded programs don't need to initialize the GIL at all (which slows down the interpreter in general). Another way would be to ensure thread safety by locking in `detail::get_internals()`, but that's overkill.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1661,9 +1661,13 @@ private:
 class gil_scoped_release {
 public:
     explicit gil_scoped_release(bool disassoc = false) : disassoc(disassoc) {
+        // `get_internals()` must be called here unconditionally in order to initialize
+        // `internals.tstate` for subsequent `gil_scoped_acquire` calls. Otherwise, an
+        // initialization race could occur as multiple threads try `gil_scoped_acquire`.
+        const auto &internals = detail::get_internals();
         tstate = PyEval_SaveThread();
         if (disassoc) {
-            auto key = detail::get_internals().tstate;
+            auto key = internals.tstate;
             #if PY_MAJOR_VERSION < 3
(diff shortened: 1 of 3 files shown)


wjakob:
LGTM!

jagerman:
It looks like this is causing intermittent failures under MSVC 2015/Python 2.7 (not the dreaded -1, thankfully, but a proper test failure):
https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.2019/job/kig85y403w39abx5
https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.2021/job/f6n9chvi940plt2x
but this one, which has the merged change, passed:
https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.2020/job/jsms662esm4qm052

Possible race condition in the test?

dean0x7d:
I've been able to reproduce this consistently by increasing the number of threads, but it only seems to show up with MSVC2015 + Python 2.7 + exec. I've tried using the raw C API (`Py_BEGIN_ALLOW_THREADS`/`PyGILState_Ensure`/`PyRun_String`) and it still happens for 2015/2.7, so I'm assuming the issue is something specific to that compiler/Python combination and not any underlying issue with pybind11's GIL release/acquire. I've switched to a more direct test in cd2d3ad5df688e7df208896ace31364350488b7e which seems to make 2015/2.7 happy.

PR #921, opened 2017-06-23
`python -m pybind11 --includes` prints include paths

aldanor:
A very minor/cosmetic addition, makes building modules manually slightly easier:

```sh
$ c++ -I `python -m pybind11`
```

instead of

```sh
$ c++ -I `python -c 'import pybind11; print(pybind11.get_include())'`
```

The docs in #907 could also get updated if this gets accepted.

### DIFF
--- a/pybind11/__main__.py
+++ b/pybind11/__main__.py
@@ -1,0 +1,31 @@
+from __future__ import print_function
+
+import argparse
+import sys
+import sysconfig
+
+from . import get_include
+
+
+def print_includes():
+    dirs = [sysconfig.get_path('include')]
+    if sysconfig.get_path('platinclude') not in dirs:
+        dirs.append(sysconfig.get_path('platinclude'))
+    if get_include() not in dirs:
+        dirs.append(get_include())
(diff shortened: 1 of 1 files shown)


jagerman:
I realize there's no `__main__` at all at the moment, but I'd be a bit more receptive to this if it put the print statement behind an argument, so that you need to invoke something like `python -m pybind11 --includes`, which leads the door wide open for making it do other nifty things in the future.

aldanor:
Agreed, good point re: `--includes`. Would you expect it to behave the same as `python-config`? (i.e., prefix the include paths with `-I` on Linux/macOS).

jagerman:
That's probably the most convenient.  Perhaps throw in the Python includes, too?

aldanor:
Alright.

```sh
python -m pybind11 --includes
```

now prints include flags for both Python (borrowed from python3-config code) and pybind11.

aldanor:
Does this look better now; anything else to do here?

dean0x7d:
LGTM

PR #923, opened 2017-06-26
travis-ci: switch gcc-7 build to buster, remove from allow_failures

jagerman:
gcc 7 is now in debian testing ("buster"), with a proper stable upstream
release; this updates the associated travis-ci to use "buster" (rather
than "sid"), and removes the build from `allow_failures`.


dean0x7d:
Nice, no more `allow_failures` 👍

PR #924, opened 2017-06-26
Add a life support system for type_caster temporaries (+ binary size reduction)

### LINKED ISSUE #916: Matrix conversion and variant issue

dean0x7d:
The first commit here changes the handling of temporary objects created by `type_caster::load()`. It uses an RAII `loader_life_support` class, placed in `dispatcher`, which holds a list of patients only for the lifetime of the function call. This has two nice advantages:

1. Supercasters don't need to care about maintaining the life of their subcasters.
2. Switching `type_caster_generic` to this system reduced the binary size of the test module by 3-4% (!), depending on the compiler. The main reason seems to be the removal of `object temp` from `type_caster_generic`, which makes the struct trivially copyable.

The second commit fixes #916 (alternative to #917) by using the new system in the `Eigen::Ref` caster.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -111,6 +111,53 @@ PYBIND11_NOINLINE inline internals &get_internals() {
     return *internals_ptr;
 }
 
+/// A life support system for temporary objects created by `type_caster::load()`.
+/// Adding a patient will keep it alive up until the enclosing function returns.
+class loader_life_support {
+public:
+    /// A new patient frame is created when a function is entered
+    loader_life_support() {
+        get_internals().loader_patient_stack.push_back(nullptr);
+    }
+
+    /// ... and destroyed after it returns
+    ~loader_life_support() {
+        auto &stack = get_internals().loader_patient_stack;
+        if (stack.empty())
+            pybind11_fail("loader_life_support: internal error");
+
+        auto ptr = stack.back();
+        stack.pop_back();
+        Py_CLEAR(ptr);
+
+        // A heuristic to reduce the stack's capacity (e.g. after long recursive calls)
+        if (stack.capacity() > 16 && stack.size() != 0 && stack.capacity() / stack.size() > 2)
+            stack.shrink_to_fit();
+    }
+
+    /// This can only be used inside a pybind11-bound function, either by `argument_loader`
+    /// at argument preparation time or by `py::cast()` at execution time.
+    PYBIND11_NOINLINE static void add_patient(handle h) {
+        auto &stack = get_internals().loader_patient_stack;
+        if (stack.empty())
+            throw cast_error("When called outside a bound function, py::cast() cannot "
+                             "do Python -> C++ conversions which require the creation "
+                             "of temporary values");
+
+        auto &list_ptr = stack.back();
+        if (list_ptr == nullptr) {
+            list_ptr = PyList_New(1);
+            if (!list_ptr)
+                pybind11_fail("loader_life_support: error allocating list");
+            PyList_SET_ITEM(list_ptr, 0, h.inc_ref().ptr());
+        } else {
+            auto result = PyList_Append(list_ptr, h.ptr());
+            if (result == -1)
+                pybind11_fail("loader_life_support: error adding patient");
+        }
+    }
+};
+
 // Gets the cache entry for the given type, creating it if necessary.  The return value is the pair
 // returned by emplace, i.e. an iterator for the entry and a bool set to `true` if the entry was
 // just created.
@@ -643,9 +690,11 @@ protected:
         // Perform an implicit conversion
         if (convert) {
             for (auto &converter : typeinfo->implicit_conversions) {
-                temp = reinterpret_steal<object>(converter(src.ptr(), typeinfo->type));
-                if (load_impl<ThisT>(temp, false))
+                auto temp = reinterpret_steal<object>(converter(src.ptr(), typeinfo->type));
+                if (load_impl<ThisT>(temp, false)) {
+                    loader_life_support::add_patient(temp);
                     return true;
+                }
             }
             if (this_.try_direct_conversions(src))
                 return true;
@@ -674,7 +723,6 @@ protected:
 
     const type_info *typeinfo = nullptr;
     void *value = nullptr;
-    object temp;
 };
 
 /**
@@ -1064,7 +1112,7 @@ template <typename StringType, bool IsView = false> struct string_caster {
 
         // If we're loading a string_view we need to keep the encoded Python object alive:
         if (IsView)
-            view_into = std::move(utfNbytes);
+            loader_life_support::add_patient(utfNbytes);
 
         return true;
     }
@@ -1080,8 +1128,6 @@ template <typename StringType, bool IsView = false> struct string_caster {
     PYBIND11_TYPE_CASTER(StringType, _(PYBIND11_STRING_NAME));
 
 private:
-    object view_into;
-
     static handle decode_utfN(const char *buffer, ssize_t nbytes) {
 #if !defined(PYPY_VERSION)
         return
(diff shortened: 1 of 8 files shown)


wjakob:
This generally looks great. However, one issue that should be considered here is that local static variables like ``static PyObject *ptr = nullptr;`` aren't going to be mutually visible in larger projects that may involve many separate modules. This is really the purpose of the ``internals`` data structure, thus I would suggest to keep it there.

jagerman:
I was thinking of something like this; 👍 for implementing it.  Also good for catching the `py::cast` issue; I was pretty sure the following was unsafe (and not addressed by #917):
```C++
using namespace Eigen;
void some_func(Ref<MatrixXd> x) { /* do something with x */ }
some_func(py::cast<Ref<MatrixXd>>(myarray));
```
With this PR that now raises an exception, which is good (it would be nicer still to make that work, but that doesn't really seem feasible).

The one issue I see is that with the life support instance being outside the `dispatch` loop, which means there's a single lifeboat for *all* overloads across both passes, rather than just the one that finally makes it (e.g. because the overload fails on some later argument, but the earlier argument already added a lifeboat entry).  I think it would suffice to make the guard localized to the function calls, i.e. to put the `guard` localized to the `try { ... }` that calls the function (the first pass just after `// 6. Call the function` and the second in the loop over `second_pass`).

As for lifetime, I'm a bit concerned that this approach has the potential to vastly extend the lifetime of temporaries: it seems as though no pybind11 function call is ever going to result in freeing the temporaries when there is some pybind11 function call somewhere up the stack.  If all my pybind11 functions are short, that's not an issue, but if a pybind function call sits high up the stack (e.g. perhaps it calls a python function that does significant work) this is going to build up temporaries without freeing them for potentially a very long time.

An alternative approach would be to get rid of the nesting and counter, and have `loader_life_support` contain a static `forward_list<PyObject *> stack`: construction pops a new (nullptr) value on the stack, destruction pops it off, and adding a new item appends, vivifying a new list at the beginning of the stack when needed.  (It might be simpler and perhaps more efficient to just use a `forward_list<forward_list<py::object>>`: construction does an `stack.emplace_front()`, destruction does `stack.pop_front()`, and adding does a `stack.front().emplace_front(h)`.

dean0x7d:
Very good points. I've made the appropriate changes. The patients are now stashed in `internals` and they should have the minimal required lifetime.

I've now added tests for `py::cast` with temporary values and this does actually work without error in the majority of cases because the temp objects are supported by the same life support frame as the enclosing function. So things like this work now:

```c++
// when inside a bound function...
class_<A>(m, "A");
class_<B>(m, "B").def(py::init<A>());
py::implicitly_convertible<A, B>();

m.def("foo", []() {
    // implicit conversion A -> B
    auto obj = py::cast(A{});
    auto b = obj.cast<const B&>(); // reference to temporary
    // ... use b -- previously dangling reference, now works

    auto s = str_obj.cast<std::string_view>();
    // ... use s -- also works now
});
```

Is there any reason not to allow this to work? I suppose there could be a danger of creating too many temp objects (which are only destroyed on `return`). This could be explicitly disabled with an additional guard in `py::cast`, but it might actually be nice to keep it.

include/pybind11/cast.h
```diff
@@ -114,44 +114,38 @@ PYBIND11_NOINLINE inline internals &get_internals() {
...
-        return value;
-    }
-
 public:
+    /// A list of patiences is created when a function is entered
     loader_life_support() {
-        if (level() == 0)
-            patients() = PyList_New(0);
-        ++level();
+        get_internals().loader_patient_stack.push_back(PyList_New(0));
```
jagerman: This seems like it's going to be creating a lot of lists here, the vast majority of which are never used (i.e. life support is the exception rather than the rule).  Is there a reason to prefer always allocating here to allocating the list on demand when the first `add_patient` is called?

jagerman:
Ah, I hadn't thought about the life support already being present; that's a nice feature.

include/pybind11/cast.h
dean0x7d: Indeed, there's no need to allocate the list in the majority of cases. Fixed.

jagerman:
This looks good to me!

PR #931, opened 2017-07-01
Fix unsigned error value casting

### LINKED ISSUE #929: No TypeError when converting negative int to unsigned 64bit on Windows
### Issue
Function argument type conversion from negative python integer to unsigned 64 bit c++ type does not raise TypeError. For a 32 bit unsigned type a TypeError is thrown as expected.

Note: It doesn't happen on Linux (with gcc 7.0.1) perhaps because `uint64_t` is an alias for `unsigned long` where as on Windows with VS 2017 `uint64_t` is an `unsigned long long`.

Occurs on:
Windows 10 (VS 2017)
pybind11 2.1.1 and master ( 30f6c3b )

### Example
cpp:
```cpp

namespace py = pybind11;

PYBIND11_MODULE(IntTest, m)
{
    m.def("take_uint64", [](uint64_t val)
    {
        py::print("taken uint64: ", val);
    });

    m.def("take_uint32", [](uint32_t val)
    {
        py::print("taken uint32: ", val);
    });
}
```
python:
```python
from IntTest import *

try:
    take_uint32(-1)
except TypeError as e:
    assert "take_uint32(): incompatible function arguments." in str(e)

take_uint64(-1)

```
example output:
```
taken uint64:  4294967295
```

jagerman:
When casting to an unsigned type from a python 2 `int`, we currently cast using `(unsigned long long) PyLong_AsUnsignedLong(src.ptr())`.  If the Python cast fails, it returns `(unsigned long) -1`, but then we
cast this to `unsigned long long`, which means we get 4294967295, but because that isn't equal to `(unsigned long long) -1`, we don't detect the failure.

This commit moves the unsigned casting into a `detail::as_unsigned` function which, upon error, casts -1 to the final type, and otherwise casts the return value to the final type to avoid the problematic double cast when an error occurs.

The error most commonly shows up under Python 2 wherever `long` is 32-bits (e.g. under both 32- and 64-bit Windows, and under 32-bit linux) when passing a negative value to a bound function taking an `unsigned long`.  The same issue would show up, however, for other size-increasing casts regardless of the Python version (for example, when binding a function taking gcc's `__uint128_t` type).

Fixes #929.

### DIFF
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -72,6 +72,12 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("string_view32_return", []() { return std::u32string_view(U"utf32 secret \U0001f382"); });
 #endif
 
+    // test_integer_casting
+    m.def("i32_str", [](std::int32_t v) { return std::to_string(v); });
+    m.def("u32_str", [](std::uint32_t v) { return std::to_string(v); });
+    m.def("i64_str", [](std::int64_t v) { return std::to_string(v); });
+    m.def("u64_str", [](std::uint64_t v) { return std::to_string(v); });
+
     // test_tuple
     m.def("pair_passthrough", [](std::pair<bool, std::string> input) {
         return std::make_pair(input.second, input.first);
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -929,31 +929,27 @@ public:
                 py_value = (py_type) PyFloat_AsDouble(src.ptr());
             else
                 return false;
-        } else if (sizeof(T) <= sizeof(long)) {
-            if (PyFloat_Check(src.ptr()))
-                return false;
-            if (std::is_signed<T>::value)
-                py_value = (py_type) PyLong_AsLong(src.ptr());
-            else
-                py_value = (py_type) PyLong_AsUnsignedLong(src.ptr());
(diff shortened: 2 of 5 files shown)


jagerman:
The added tests segfaulted under PyPy; it turns out to be a latent issue (i.e. occurs in current master) when attempting to call:
```C++
m.def("f", [](int32_t) {});
```
with a value that fits in a `long` but not a `int32_t` (e.g. 3333333333 on a 64-bit-`long` architecture).

I've squashed the fix into this commit: we aren't allowed to call `PyErr_ExceptionMatches()` if we didn't actually get an exception from the conversion.  (We were calling it unconditionally if either an exception happened, *or* we detect an integer overflow case).

tests/test_builtin_casters.cpp
```diff
@@ -72,6 +72,11 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("string_view32_return", []() { return std::u32string_view(U"utf32 secret \U0001f382"); });
 #endif
 
+    m.def("i32_str", [](std::int32_t v) { return std::to_string(v); });
```
dean0x7d: Can you add a comment above this line to associate it with the Python test function:
```c++
// test_integer_casting
```

jagerman: Done.

PR #936, opened 2017-07-03
tuple/pair/stl.h casters: use rvalue subcasting when casting an rvalue tuple/pair/container

### LINKED ISSUE #853: Returning std::vector of move-only holders
The `const holder &` entries in an `std::vector` (and supposedly also other STL containers) cannot be passed to `move_only_holder_caster::cast(holder_type &&src, return_value_policy, handle)`.

(Apparently, for custom non-holder move-only types, this is ok, since the `return_value_policy` is forced to be `move` and the overload of `type_caster::cast(const type &, return_value_policy, handle)` creates a move_constructor that casts away the constness.)

Example code:
```C++
py::class_<A> classA(m, "A");
m.def("test", [] () { return std::vector<std::unique_ptr<A>>(); });
```

### LINKED ISSUE #935: scipy.optimize.fmin_l_bfgs_b thinks that my array isn't fortran contiguous
I have C++ code that calculates a function and its gradient (toy example):

```
#include <Eigen/Dense>
#include <pybind11/stl.h>
#include <pybind11/eigen.h>

std::tuple<double, Eigen::VectorXd> quadratic (const Eigen::Ref<const Eigen::VectorXd>& x) {
    Eigen::VectorXd grad (x);
    return {.5 * x.array().square().sum(), grad};
}

    m.def("quadratic", quadratic);
}
```

And I would like to minize it using scipy.

```
> scipy.optimize.fmin_l_bfgs_b(func=mymodule.quadratic, x0=np.array([1., 2, 3]), fprime=None)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
...
lib/python2.7/site-packages/scipy/optimize/lbfgsb.pyc in fmin_l_bfgs_b(func, x0, fprime, args, approx_grad, bounds, m, factr, pgtol, epsilon, iprint, maxfun, maxiter, disp, callback, maxls)
    191 
    192     res = _minimize_lbfgsb(fun, x0, args=args, jac=jac, bounds=bounds,
--> 193                            **opts)
    194     d = {'grad': res['jac'],
    195          'task': res['message'],

lib/python2.7/site-packages/scipy/optimize/lbfgsb.pyc in _minimize_lbfgsb(fun, x0, args, jac, bounds, disp, maxcor, ftol, gtol, eps, maxfun, maxiter, iprint, callback, maxls, **unknown_options)
    319         _lbfgsb.setulb(m, x, low_bnd, upper_bnd, nbd, f, g, factr,
    320                        pgtol, wa, iwa, task, iprint, csave, lsave,
--> 321                        isave, dsave, maxls)
    322         task_str = task.tostring()
    323         if task_str.startswith(b'FG'):

ValueError: failed to initialize intent(inout) array -- input not fortran contiguous
```
I can fix the issue by writing a wrapper (in python) for `mymodule.quadratic` that copies the gradient.

Also, the issue does not occur if I only return the gradient from C++:

```
Eigen::VectorXd copy(const Eigen::Ref<const Eigen::VectorXd>& x) {
    Eigen::VectorXd grad (x);
    return grad;
}
```

```
def quadratic (x):
    value = (x*x).sum()/2.0;
    grad = mymodule.copy(x);
    return value, grad
```

so it's related to returning a tuple.

And a possibly relevant scipy issue: https://github.com/scipy/scipy/issues/5832

Thank you for your help and apologies if this is already answered in the documentation.

jagerman:
The tuple caster had only a const-lvalue-ref cast method, which meant
that subcaster casts couldn't distinguish between an lvalue and rvalue
cast; this caused Eigen values in a tuple, for instance, to end up with
a readonly flag.

This fixes it by adding an rvalue versions of `cast()` and `cast_impl()`
that move the tuple values when subcasting them.

Fixes #935 and #853 

Edit: I've updated this PR to also consolidate the tuple/pair casters into a single `tuple_caster` implementation.  It is essentially just the `std::tuple` implementation, which works perfectly well with `std::pair`s.  Thus the fix here also gets applied to `std::pair`.

Edit 2: Also `stl.h` casters get the same treatment, which (incidentally) also fixes #853.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1255,50 +1255,13 @@ public:
     template <typename _T> using cast_op_type = remove_reference_t<pybind11::detail::cast_op_type<_T>>;
 };
 
-template <typename T1, typename T2> class type_caster<std::pair<T1, T2>> {
-    typedef std::pair<T1, T2> type;
-public:
-    bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src))
-            return false;
-        const auto seq = reinterpret_borrow<sequence>(src);
-        if (seq.size() != 2)
-            return false;
-        return first.load(seq[0], convert) && second.load(seq[1], convert);
-    }
-
-    static handle cast(const type &src, return_value_policy policy, handle parent) {
-        auto o1 = reinterpret_steal<object>(make_caster<T1>::cast(src.first, policy, parent));
-        auto o2 = reinterpret_steal<object>(make_caster<T2>::cast(src.second, policy, parent));
-        if (!o1 || !o2)
-            return handle();
-        tuple result(2);
-        PyTuple_SET_ITEM(result.ptr(), 0, o1.release().ptr());
-        PyTuple_SET_ITEM(result.ptr(), 1, o2.release().ptr());
-        return result.release();
-    }
-
-    static PYBIND11_DESCR name() {
-        return type_descr(
-            _("Tuple[") + make_caster<T1>::name() + _(", ") + make_caster<T2>::name() + _("]")
-        );
-    }
-
-    template <typename T> using cast_op_type = type;
-
-    operator type() & { return type(cast_op<T1>(first), cast_op<T2>(second)); }
-    operator type() && { return type(cast_op<T1>(std::move(first)), cast_op<T2>(std::move(second))); }
-protected:
-    make_caster<T1> first;
-    make_caster<T2> second;
-};
-
-template <typename... Tuple> class type_caster<std::tuple<Tuple...>> {
-    using type = std::tuple<Tuple...>;
-    using indices = make_index_sequence<sizeof...(Tuple)>;
+// Base implementation for std::tuple and std::pair
+template <template<typename...> class TupleType, typename... Tuple> class tuple_caster {
+    using type = TupleType<Tuple...>;
     static constexpr auto size = sizeof...(Tuple);
-
+    using indices = make_index_sequence<size>;
 public:
+
     bool load(handle src, bool convert) {
         if (!isinstance<sequence>(src))
             return false;
@@ -1327,7 +1290,6 @@ protected:
     template <size_t... Is>
     type implicit_cast(index_sequence<Is...>) && { return type(cast_op<Tuple>(std::move(std::get<Is>(subcasters)))...); }
 
-
     static constexpr bool load_impl(const sequence &, bool, index_sequence<>) { return true; }
 
     template <size_t... Is>
@@ -1337,9 +1299,6 @@ protected:
                 return false;
         return true;
     }
-
-    static handle cast_impl(const type &, return_value_policy, handle,
-                            index_sequence<>) { return tuple().release(); }
 
     /* Implementation: Convert a C++ tuple into a Python tuple */
     template <size_t... Is>
@@ -1357,8 +1316,14 @@ protected:
         return result.release();
     }
 
-    std::tuple<make_caster<Tuple>...> subcasters;
+    TupleType<make_caster<Tuple>...> subcasters;
 };
+
+template <typename T1, typename T2> class type_caster<std::pair<T1, T2>>
+    : public tuple_caster<std::pair, T1, T2> {};
+
+template <typename... Tuple> class type_caster<std::tuple<Tuple...>>
+    : public tuple_caster<std::tuple, Tuple...> {};
 
 /// Helper class which abstracts away certain actions. Users can provide specializations for
 /// custom holders, but it's only necessary if the type has a non-standard interface.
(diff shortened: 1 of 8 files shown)


jagerman:
I've updated this PR to also consolidate the tuple/pair casters into a single `tuple_caster` implementation.  It is essentially just the `std::tuple` implementation, which works perfectly well with `std::pair`s.  Thus the fix here also gets applied to `std::pair`.

I've also added tests.

wjakob:
This looks great! Combining the pair and tuple casters is a nice simplification.

YannickJadoul:
I'm curious: this reminds me of/sounds like it is very similar to the problem of returning an vector of move-only types (cfr. #853). Could this kind of solution also help there, for `list_caster` ?

jagerman:
I'm working on doing the same for `stl.h` right now; will update this PR with it shortly.

YannickJadoul:
Awesome, thanks! :-)

jagerman:
@YannickJadoul - This also fixes #853 now (I used an adapted version your test case in the tests, which now passes).

include/pybind11/stl.h
```diff
@@ -49,6 +49,13 @@
 NAMESPACE_BEGIN(pybind11)
 NAMESPACE_BEGIN(detail)
 
+// Extracts an const lvalue reference or rvalue reference for a container element based on the
+// lvalue/rvalue status of the container
+template <typename ContainerArg, typename Value>
+using forward_item = conditional_t<
+    std::is_lvalue_reference<ContainerArg>::value, const Value &,
+    typename std::add_rvalue_reference<Value>::type>;
```
dean0x7d: Explicitly using `const Value &` is a bit less generic than it could be: `remove_reference_t<Value>&` and `remove_reference_t<Value>&&`.

Consider wrapping it in a `forward_like` function like here: https://stackoverflow.com/a/29780197/7189436
`forward_like<Container>(element)` is pretty nice and expressive.

include/pybind11/cast.h
```diff
@@ -1255,50 +1255,13 @@ template <typename CharT> struct type_caster<CharT, enable_if_t<is_std_char_type
...
-    make_caster<T1> first;
-    make_caster<T2> second;
-};
-
-template <typename... Tuple> class type_caster<std::tuple<Tuple...>> {
-    using type = std::tuple<Tuple...>;
-    using indices = make_index_sequence<sizeof...(Tuple)>;
+// Base implementation for std::tuple and std::pair
+template <template<typename...> class TupleType, typename... Tuple> class tuple_caster {
+    using type = TupleType<Tuple...>;
```
dean0x7d: Consider renaming `Tuple...` into `Ts...` or similar. At first glance `TupleType<Tuple...>` looks like a tuple of tuples.

tests/test_stl.cpp
```diff
@@ -61,6 +61,50 @@ TEST_SUBMODULE(stl, m) {
         return set.count("key1") && set.count("key2") && set.count("key3");
     });
 
+    #if defined(_MSC_VER) && _MSC_VER < 1910
+    // We get some really long type names here which causes MSVC 2015 to emit warnings
+    #  pragma warning(disable: 4503) // warning C4503: decorated name length exceeded, name was truncated
```
dean0x7d: I'd suggest adding this globally to `pybind11_tests.h` since it's already required in two files and likely more in the future.

include/pybind11/cast.h
```diff
@@ -1338,14 +1304,15 @@ template <typename... Tuple> class type_caster<std::tuple<Tuple...>> {
         return true;
     }
 
-    static handle cast_impl(const type &, return_value_policy, handle,
+    template <typename T>
+    static handle cast_impl(T &&, return_value_policy, handle,
```
jagerman: With just the `const type &` an rvalue argument will end up at the full template version.  I suppose I could put in dual empty methods (one lvalue, one rvalue) if you like that better (though there are still things like non-const lvalue and const rvalue that would slip by).

dean0x7d: Right, I missed that. Never mind, it's better to leave it with `T &&` then.

jagerman: I'll try just taking it out completely (i.e. so it lands in the base case) and see if that causes any problems.

jagerman:
> adding a `forward_like<Container>()` function

That's a nice suggestion, and it shortens up the code a bit too (by eliminating the `using Forward = ...`).  I kept it in `stl.h` for now; we can move it to the block of utility templates/functions in `common.h` if/when something else wants it.

include/pybind11/cast.h
jagerman: Do you prefer keeping it in (and templated), or taking it out entirely, as in the last commit?

dean0x7d: 👍 for removing it completely -- perfect!

PR #937, opened 2017-07-05
Fix compilation with Intel's compiler

dean0x7d:
ICC failed to compile and was reporting that `try_direct_conversions()` cannot be `constexpr` because `handle` is not a literal type (it is, ICC just doesn't know it). The fix removes `constexpr` from the function since it isn't strictly needed.

This commit also suppresses new false positive warnings which mostly appear in `constexpr` contexts (where the compiler knows conversions are safe). Suppressing all of these may seem heavy-handed, but these really are false positives -- other compilers have the same warnings enabled without any of the issues for compile-time constants.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -846,7 +846,7 @@ protected:
     }
 
     template <typename T, typename = enable_if_t<std::is_move_constructible<T>::value>>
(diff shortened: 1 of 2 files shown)


jagerman:
LGTM.  We've pretty much always taken the path of disabling false positive warnings rather than changing code to work around them, so it seems pretty reasonable to keep doing so for ICC.

PR #938, opened 2017-07-07
Custom type caster pointer ownership support (and container regression fix)

jagerman:
PR #936 broke the ability to return a pointer to a stl container (and, likewise, to a tuple) because the added deduced type matched a non-const pointer argument: the pointer-accepting `cast` in PYBIND11_TYPE_CASTER had a `const type *`, which is a worse match for a non-const pointer than the universal reference template #936 added.

This changes the provided TYPE_CASTER cast(ptr) to take the pointer by template arg (so that it will accept either const or non-const pointer). It has two other effects: it slightly reduces .so size (because many type casters never actually need the pointer cast at all), and it allows type casters to provide their untemplated pointer `cast()` that will take precedence over the templated version provided in the macro.

Edit:
I've also added a second commit here that makes `take_ownership` work for custom type casters.  Previously it wouldn't do anything (it was essentially ignored and worked like one would expect `reference` to work), which meant it was fairly easy to leak by returning an unmanaged pointer while expecting pybind11 to take over ownership.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -892,14 +892,15 @@ public:
         type value; \
     public: \
         static PYBIND11_DESCR name() { return type_descr(py_name); } \
-        static handle cast(const type *src, return_value_policy policy, handle parent) { \
+        template <typename T_, enable_if_t<std::is_same<type, remove_cv_t<T_>>::value, int> = 0> \
+        static handle cast(T_ *src, return_value_policy policy, handle parent) { \
             if (!src) return none().release(); \
             return cast(*src, policy, parent); \
         } \
         operator type*() { return &value; } \
         operator type&() { return value; } \
         operator type&&() && { return std::move(value); } \
-        template <typename _T> using cast_op_type = pybind11::detail::movable_cast_op_type<_T>
+        template <typename T_> using cast_op_type = pybind11::detail::movable_cast_op_type<T_>
 
 
 template <typename CharT> using is_std_char_type = any_of<
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -895,7 +895,11 @@ public:
         template <typename T_, enable_if_t<std::is_same<type, remove_cv_t<T_>>::value, int> = 0> \
         static handle cast(T_ *src, return_value_policy policy, handle parent) { \
             if (!src) return none().release(); \
-            return cast(*src, policy, parent); \
+            if (policy == return_value_policy::take_ownership) { \
+                auto h = cast(std::move(*src), policy, parent); delete src; return h; \
+            } else { \
+                return cast(*src, policy, parent); \
+            } \
         } \
         operator type*() { return &value; } \
         operator type&() { return value; } \
--- a/tests/test_methods_and_attributes.cpp
+++ b/tests/test_methods_and_attributes.cpp
@@ -154,7 +154,28 @@ public:
     }
 
     static handle cast(const ArgAlwaysConverts &, return_value_policy, handle) {
-        return py::none();
+        return py::none().release();
+    }
+};
+}}
+
+// test_custom_caster_destruction
+class DestructionTester {
+public:
+    DestructionTester() { print_default_created(this); }
+    ~DestructionTester() { print_destroyed(this); }
+    DestructionTester(const DestructionTester &) { print_copy_created(this); }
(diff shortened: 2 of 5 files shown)


jagerman:
Something else that I thought about here is that `take_ownership` on a function returning a container pointer doesn't do what I'd expect (it acts like `reference`).

I tried extending the TYPE_CASTER macro pointer `cast()` to:
```C++
if (policy == return_value_policy::take_ownership) {
    handle h = cast(std::move(*src), policy, parent);
    delete src;
    return h;
} else {
}
```
i.e. to destroy if a pointer-returning method is called with `take_ownership`, but it seems it's too late: `automatic` gets also converted into `take_ownership` before it ends up in the cast(), but we have tests expecting that a pointer return for a non-registered class doesn't get destroyed.  One possible approach is to remove the code that changes `automatic` to `take_ownership` for a pointer, so that type casters can interpret `automatic` in different ways--here, for example, it would make more sense for the default (i.e. `automatic`) to work more like `automatic_reference`.  I'm not sure how much *that* approach would break, however.

jagerman:
(The appveyor failure looks like a false positive from an error downloading catch).

include/pybind11/cast.h
```diff
@@ -889,14 +889,25 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
     operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type&(); }
 };
 
+/// Call class-specific delete if it exists or global otherwise. Can also be an overload set.
+template <typename T, typename = void_t<decltype(static_cast<void (*)(void *)>(T::operator delete))>>
+void call_operator_delete(T *p) { T::operator delete(p); }
+
+inline void call_operator_delete(void *p) { ::operator delete(p); }
```
dean0x7d: I don't think this needed to be moved.

include/pybind11/cast.h
```diff
@@ -889,14 +889,25 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
...
+
 #define PYBIND11_TYPE_CASTER(type, py_name) \
     protected: \
         type value; \
     public: \
         static PYBIND11_DESCR name() { return type_descr(py_name); } \
-        static handle cast(const type *src, return_value_policy policy, handle parent) { \
+        template <typename _T, enable_if_t<std::is_same<type, remove_cv_t<_T>>::value, int> = 0> \
+        static handle cast(_T *src, return_value_policy policy, handle parent) { \
```
dean0x7d: Identifiers that begin with an underscore followed by an uppercase letter (`_Capital`) are reserved by the standard. Change to `T_`?

tests/test_methods_and_attributes.cpp
```diff
@@ -399,4 +420,15 @@ test_initializer methods_and_attributes([](py::module &m) {
 
     using Adapted = decltype(py::method_adaptor<RegisteredDerived>(&RegisteredDerived::do_nothing));
     static_assert(std::is_same<Adapted, void (RegisteredDerived::*)() const>::value, "");
+
+    // test_custom_caster_destruction
+    // Test that `take_ownership` works on types with a custom type caster when given a pointer
+    auto dt = new DestructionTester();
```
dean0x7d: Why not a function scope `static`? (Inside the `custom_caster_no_destroy` lambda.)

include/pybind11/cast.h
jagerman: Whoops, that was for part of the change that I reverted, but forgot to move it back.  I'll put it back.

include/pybind11/cast.h
jagerman: Good point; I was copying the `typename _T` for the `cast_op_type` just below; I'll change both to `T_`.

tests/test_methods_and_attributes.cpp
jagerman: I want to be sure of the number of live instances of the object in the associated python tests so as to be sure there is no extra copy or extra destruction; `custom_caster_no_destroy` takes a reference (without creating anything new) while `destruction_tester_finish` is the counterpart meant to destroy that object.

jagerman: ... although I suppose that would still work with a function static instance.

jagerman: I've updated this, and improved the comments around it.

PR #941, opened 2017-07-11
Added *.class to .gitignore

kFYatek:
Some IDEs such as NetBeans use Jython as its internal Python interpreter/parser, which causes `*.class` files to appear seemingly out of nowhere. This diff adds it to `.gitignore` so that they don't creep in `git status` all the time, which is especially annoying when `pybind11` is used as a Git submodule.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ MANIFEST
 /docs/.build
 *.py[co]
 *.egg-info
+*.class
 *~
 .DS_Store
 /dist


wjakob:
Wouldn't it be better to fix this upstream rather than pushing the idiosyncrasies of an IDE to other projects.

kFYatek:
Hm, I don't know. Jython's `something$py.class` does not seem substantially different from CPython's `something.pyc`, and the latter is included in `.gitignore` here.

jagerman:
It is different in one key way: the `.pyc`s are generated when running the test suite, part of pybind itself.

I'm a bit curious, though: where are the `.class` files appearing?

kFYatek:
Well, my project has Pybind11 configured as a Git submodule, and upon loading the main project in NetBeans 8.2 Patch 2 (Build 201609300101) with nbPython plugin version 0.151118, two files appear under the Pybind11 submodule directory:
* `pybind11/__init__$py.class`
* `pybind11/_version$py.class`
Curiously, `__main__$py.class` is not generated even though that's the third `.py` file in that same directory.

kFYatek:
Well, it turns out I can just use `ignore = untracked` in my `.gitmodules` (https://stackoverflow.com/a/5127213), so I guess this can be closed if you're not willing to merge it.

jagerman:
It's a pretty small addition, but is there no way to configure NetBeans/Jython to stick its class files in a temporary directory rather than `.`?  Leaving junk behind in the directories of files you edit seems like a pretty major editor defect.

kFYatek:
I don't know, and frankly, I stopped caring. I just proposed adding this `.gitignore` in the spirit of ignoring any possible temporary files just because you can - but if it's against your philosophy, I'll just stick to `ignore = untracked`, which is good enough for me. Sorry for the trouble.

jagerman:
I don't have any strong objections to this.  I do agree with @wjakob that it seems like a bad editor feature, but on the other hand this wouldn't be the first thing added to `.gitignore` for junk that other IDEs drop in there (`*.vcxproj`, `*.sln`, `*~`).

I expect that I would be pretty strongly opposed to actually adding `.class` files into the pybind11 git repository in the first place, but in the highly unlikely event that someone convinces me that that's a good idea we could always remove the `.class` filter.

But I think we should either merge this or close it.

aldanor:
Well... might then as well add a bunch of other stuff like `compile_commands.json`, `.idea/`, `.python-version` (cmake, clion/pycharm, pyenv).

YannickJadoul:
@kFYatek There's also the `.git/info/exclude` file for non-versioned, local ignores. I recently started using that for IDE-specific things that I don't want 'polluting' my repository/history. Then you don't have to do this `ignore = untracked` trick (ignoring changes that should possibly nót be ignored!).
And if there's something you want to ignore álways, you can set a user-wide ignore (by default in `~/.config/git/ignore`, but configurable by `git config` for `core.excludesFile`).

(See [git documentation](https://git-scm.com/docs/gitignore) for details)

dean0x7d:
> And if there's something you want to ignore álways, you can set a user-wide ignore (by default in ~/.config/git/ignore, but configurable by git config for core.excludesFile).

I would agree with this. Git's global ignore seems like the best place for system or IDE-specific files.

Since no one has been overly enthusiastic about adding this to `.gitignore`, it might be best to close it.

PR #944, opened 2017-07-12
Detect c++ standard unconditionally (needed for embed target)

### LINKED ISSUE #943: Embedding build doesn't include compilation flags when following a regular module
If I build both a regular module and an embedded target the c++ standard doesn't get added to the compilation flags:

```CMake
cmake_minimum_required(VERSION 3.0)
project(example)
add_subdirectory(pybind11)
pybind11_add_module(example example_pybind11.cpp)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE pybind11::embed)
```

Compilation lines (extracted from `make VERBOSE=1`):
```
/usr/bin/c++   -Dexample_EXPORTS -I/home/jagerman/src/pybind11/scratch/tasty/pybind11/include -I/usr/include/python3.6m  -fPIC   -std=c++14 -flto -fno-fat-lto-objects -o CMakeFiles/example.dir/example_pybind11.cpp.o -c /home/jagerman/src/pybind11/scratch/tasty/example_pybind11.cpp
```
```
/usr/bin/c++    -I/home/jagerman/src/pybind11/scratch/tasty/pybind11/include -I/usr/include/python3.6m   -o CMakeFiles/main.dir/main.cpp.o -c /home/jagerman/src/pybind11/scratch/tasty/main.cpp
```

If I comment out the `pybind11_add_module` the main.cpp compilation line becomes:
```
/usr/bin/c++    -I/home/jagerman/src/pybind11/scratch/tasty/pybind11/include -I/usr/include/python3.6m  -std=c++14 -o CMakeFiles/main.dir/main.cpp.o -c /home/jagerman/src/pybind11/scratch/tasty/main.cpp
```

jagerman:
Currently `select_cxx_standard()`, which sets `PYBIND11_CPP_STANDARD` when not externally set, is only called from `pybind11_add_module()`, but the `embed` target setup (which runs unconditionally) makes use of `${PYBIND11_CPP_STANDARD}`, which isn't set yet.  This commit removes the `select_cxx_standard` function completely and just always runs the standard detection code.

This also tweaks the detection code to not bothering checking for the `-std=c++11` flag when the `-std=c++14` detection succeeded.

Fixes #943.

One odd thing about #943 is that when `pybind11_add_module` was never called at all (and hence `select_cxx_standard()` was never called at all), an embedded target *would* get the `std=c++14` flag, but it's a mystery to me how that was happening.

### DIFF
--- a/tools/pybind11Config.cmake.in
+++ b/tools/pybind11Config.cmake.in
@@ -90,7 +90,6 @@ if(NOT TARGET ${PN}::pybind11)
       set_property(TARGET ${PN}::module APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${PYTHON_LIBRARIES})
     endif()
 
-    select_cxx_standard()
     set_property(TARGET ${PN}::pybind11 APPEND PROPERTY INTERFACE_COMPILE_OPTIONS "${PYBIND11_CPP_STANDARD}")
 
     get_property(_iid TARGET ${PN}::pybind11 PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -18,27 +18,27 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
 include(CheckCXXCompilerFlag)
 include(CMakeParseArguments)
 
-function(select_cxx_standard)
-  if(NOT PYBIND11_CPP_STANDARD)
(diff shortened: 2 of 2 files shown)


jagerman:
The commit removes `select_cxx_standard()`; does anyone see a reason to keep it as a cmake function?  (i.e. and just always call `select_cxx_standard()` from pybind's main `CMakeLists.txt`)

dean0x7d:
Looks good to me. I don't see any reason to keep the function.

Issue #439, opened 2016-10-12
generc_type xxxx is already registered


kanaero:
Suppose in test_stl_binders.cpp we have two types mapping back to the same underlying c++ type in this case "unsigned int".

    py::bind_vector<std::vector<unsigned int>>(global, "VectorInt");
    py::bind_vector<std::vector<unsigned int>>(global, "AnotherVectorInt");

When importing the module, the error generic_type: type "AnotherVectorInt" is already registered! arises.

Are we disallowing type alias within python. Can we not refer to the same vector type using multiple python types ?

If we choose to keep the current behaviour of disallowing multiple aliases for the same underlying type we should improve the error message to say "generic_type: type "AnotherVectorInt" has underlying type "unsigned int" that is already registered.

wjakob:
You can use ``global.attr("VectorInt") = global.attr("AnotherVectorInt")`` to create an alias on the Python side. However, you can't register the type two times independently since it would cause various inconsistencies in Python-internal data structures.

kliegeois:
@SanPen I am facing the same issue; how have you resolved this?

SanPen:
Hi! yes I solved it by adding `py::module_local()` to the class definition 

example:

```
py::class_<Zone, Device, std::shared_ptr<Zone>>(m, "Zone", py::module_local())
.def(py::init<std::string,
                           std::string,
                           std::string,
                           uword>(),
                  py::arg("uuid"),
                  py::arg("secondary_id"),
                  py::arg("name"),
                  py::arg("time_steps"))
            .def_property("area", &Zone::getArea, &Zone::setArea, "Area object to which this zone belongs to. It may be null.")
            ;
```

THis is somewhere in the docs and should be in every example!

kliegeois:
Thanks a lot @SanPen !

PR #949, opened 2017-07-19
Module-local bindings

### LINKED ISSUE #919: Eventual collisions between opaque STL containers
After some chatting on gitter, @jagerman suggested that I raise an issue here, mostly so that it can be kept track of.

This is related to #439, but has more to do with what happens in the long run when more and more packages are developed using pybind11.

Currently,  opaque STL containers are most often generated with a combo of the `PYBIND11_MAKE_OPAQUE` macro and a call to `py::bind_vector`.

If two different modules by different developers generate the same opaque containers, an import error occurs if both modules are used together.  One simple solution is to wrap the binding in try/catch blocks as shown here: https://github.com/molpopgen/opaque_test.  (That example uses ccpimport to compile the code.)

However, the try/catch solution is imperfect: the name seen on the Python side depends on the import order.  Further, there is the deeper problem that `py::bind_vector` may optionally be invoked with `py::buffer_protocol()` and that the class type returned may be further augmented via `.defs` in the usual way.  The possible result then is that both the name and features/behavior depend on import order, which seems less than ideal.

The gitter discussion ended with the agreement that a solution is needed and that it is not clear what that solution should look like.

jagerman:
This PR adds a `py::local` attribute that lets one confine a type registration to the module (more technically, the shared object) in which it is defined, by registering it with:
```C++
    py::class_<C>(m, "C", py::local())
```
This will allow the C++ class `C` to be registered in different modules with independent sets of class definitions.  On the Python side, two such types will be completely distinct; on the C++ side, the C++ resolves to a different Python types in each module.

This also fixes #919 / #439 by making the `stl_bind.h` binding code apply `py::local()` by default (but passing in `py::local(false)` can be used in cases where you want to use the stl-bound type across module boundaries).

Tests are included showing this working; I haven't yet written up the documentation as I wanted feedback/input on the idea and approach first.


wjakob:
I think this is a great idea, and your implementation looks good to me.

jagerman:
Note: the tests added here rely on the fix from #951; I've included the #951 fix here, but filed the separate PR to allow separate discussion of that fix.

include/pybind11/cast.h
```diff
@@ -242,13 +267,18 @@ PYBIND11_NOINLINE inline detail::type_info* get_type_info(PyTypeObject *type) {
...
+PYBIND11_NOINLINE inline detail::type_info *get_type_info(const std::type_index &tp,
                                                           bool throw_if_missing = false) {
+    std::type_index type_idx(tp);
     auto &types = get_internals().registered_types_cpp;
-
-    auto it = types.find(std::type_index(tp));
+    auto it = types.find(type_idx);
     if (it != types.end())
         return (detail::type_info *) it->second;
+    auto &locals = registered_local_types_cpp();
```
dean0x7d: I'm reasonably sure this is an ODR violation. `registered_local_types_cpp()` is located inside an anonymous namespace inside a header which means there's going to be one unique function per translation unit, i.e.:
```
a.cpp -> pybind11::detail::<unique_name_a>::registered_local_types_cpp()
b.cpp -> pybind11::detail::<unique_name_b>::registered_local_types_cpp()
```
On the other hand, `get_type_info()` is not in an anonymous namespace, so per ODR we are promising that there is just one of those for all transaltion units. But because of the call highlighted above, the body of `get_type_info()` will be defined differently in different translalation units, i.e. one will call `<unique_name_a>::registered_local_types_cpp()`, another `<unique_name_b>::`, etc.

A possible solution could be to move the definition of `registered_local_types_cpp` into the `PYBIND11_MODULE` macro (without the anon namespace) which would ensure a single definition per module. Although, I'm not sure if GCC will consider the function-local `static` variable as unique or shared among different modules.

jagerman:
> I believe there is an ODR violation here.

My thinking on it was that we might have wiggled out of it by virtue of python modules being loaded with `RTLD_LOCAL`, so that whichever `registered_local_types_cpp` get used will be the "right" one.  That's probably assuming too much, though, and furthermore it's not going to work if the module is loaded with `RTLD_GLOBAL`, so indeed a better solution is needed.

Putting it into the `PYBIND11_MODULE` macro is a nice idea; I'll try that and see how it goes.

jagerman:
`PYBIND11_MODULE` works just fine (and gcc does indeed consider the variable unique, which is nice).

There is one thorny issue though—what do do about embedding that doesn't have a macro at all?

dean0x7d:
> There is one thorny issue though—what do do about embedding that doesn't have a macro at all?

Yeah, that does throw a wrench in the works. Technically, there is the `PYBIND11_EMBEDDED_MODULE` macro, but it doesn't help much.

This also made me realize that the bindings are also made local to a `.so` file not a `py::module` which adds some nuance: all modules created with `def_submodule` would share the same local bindings, as would all embedded modules.

Do you think it would be possible to place the local type map inside a capsule and attach it to a `py::module`? (Instead of the function-scope static type map.) This would also require `get_type_info` to take an extra `PyObject *module` argument -- the context of the type lookup. This might not be the most efficient solution, but I can't think of anything leaner right now. However, it would have the advantage of making the local bindings have `py::module` scope.

jagerman:
I can't see how storing a capsule in the module would work.  Storing the capsule is fine, but I don't see an obvious way to actually get the module (to pass it in to `get_type_info()`) in various contexts.

How about if we simply drop the anonymous namespace, and go back to the local-to-the-current-so approach? (and perhaps name it `py::local_type` instead of `py::local`--but it's not quite module-local, as you pointed out).  GCC (at least gcc 7; I haven't tested earlier versions yet) doesn't actually appear to have any problems with it, and I think it solves the ODR issue.

dean0x7d:
That definitely solves the ODR issue, but I was under the impression that GCC shared function-scope statics across modules -- the issues with `get_internals_ptr()`.

jagerman:
A bit of further investigation: gcc only seems to behave this way when not compiling with `-flto`. Static locals then get exported with `STB_GNU_UNIQUE`, which makes them unique, even across the `RTLD_LOCAL` boundary.

With `-flto` this doesn't appear to happen.  There's also a flag dedicated to this, `-fno-gnu-unique`, which I'm very tempted to force on (and perhaps require for the feature to work).

jagerman:
66d836c should fix the gcc visibility issue (I also simplified `get_internals()` back to what it used to be using the same trick).  There are a few other static local variable functions here and there, but they are for things like caching the numpy api object--I don't think any of them are going to be an issue if shared across modules (even for different pybind versions).

dean0x7d:
Nice investigative work! The solution with the function attribute looks ideal (as opposed to the compiler flag, which we have no way of enforcing). Right now, you've included the attribute for gcc but not clang. I'd suggest including it for any compiler which defines `__GNUG__` (gcc, clang and intel). It's required anyway, so best to enable it explicitly than rely on whatever the defaults are for the different gcc-compatible compilers.

jagerman:
> Right now, you've included the attribute for gcc but not clang. I'd suggest including it for any compiler which defines `__GNUG__` (gcc, clang and intel). It's required anyway, so best to enable it explicitly than rely on whatever the defaults are for the different gcc-compatible compilers.

It won't hurt on other compilers, but as far as I can tell this only affects g++.  It can't hurt to turn it on for the others, though.

jagerman:
I've rewritten the git history here, and added a test for the cross-module exception handler (suggested by @dean0x7d in #951; I'll close that one and just merge the fix via this one).

dean0x7d:
The revised implementation here looks excellent. Sorry, I've been busy as of late and fallen behind on the reviews. Just one quick question: what happens if a type is registered as both local and global? I didn't find any tests for that case, but I may have missed it.

jagerman:
> Just one quick question: what happens if a type is registered as both local and global? I didn't find any tests for that case, but I may have missed it.

If a type is already registered globally, you'll get a failure if you try to register it locally.  If you go the other way, there will be no failure, but the global type will take precedence (local types are only checked if the global type isn't found).

jagerman:
Now that I think about it, I wonder if that's the right behaviour.  Perhaps it would make more sense to make the local type take precedence.

dean0x7d:
Appleclang seems to be broken in [debug mode](https://travis-ci.org/dean0x7d/pybind11/builds/261067255). Strangely, the older appleclang 7.3.0 on Travis and the latest 8.1.0 fail with different errors:

7.3.0
```
E    TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
E        1. pybind11_cross_module_tests.LocalVec()
E        2. pybind11_cross_module_tests.LocalVec(arg0: pybind11_cross_module_tests.LocalVec)
E        3. pybind11_cross_module_tests.LocalVec(arg0: iterable)
E       
E    Invoked with:
```

8.1.0
```
test_exceptions.py:4: in <module>
    import pybind11_cross_module_tests as cm
E   ImportError: generic_type: type "LocalType" is already registered!
```

The key difference in debug mode is that `-fvisibility=hidden` isn't being given to the compiler, but it shouldn't really matter since the relevant functions are explicitly marked with `__attribute__ ((visibility("hidden")))`.

jagerman:
I think the issue is coming about because the functions calling the functions with the static local are not becoming hidden visibility, and so we're back into the ODR issue.  It seems to me that the same issue is going to apply under appleclang for `get_internals()` when compiled not under `-fvisibility=hidden`.

What if we just force hidden visibility for everything under the `pybind11` namespace?  I tested with changing the `NAMESPACE_BEGIN` definition to `namespace name __attribute__((visibility("hidden"))) {`, which seems to fix clang on OS X.

Obviously we'd need to rename the macro (e.g. `PYBIND11_NAMESPACE`), and condition on `__GNUG__`, etc.  But would this cause any other problems?

dean0x7d:
Adding the visibility attribute to the namespace seems good to me. The only other options are visibility pragmas or function-level attributes, but neither are very convenient.

jagerman:
There is one potential downside to this in that it leads to a warning under gcc for code like this (from `test_exceptions.cpp`) if compiling without `-fvisibility=hidden`:

```C++
struct PythonCallInDestructor {
    // ...
    py::dict d;
};
```

```
/home/jagerman/src/pybind11/tests/test_exceptions.cpp:61:8: warning: ‘PythonCallInDestructor’ declared with greater visibility than the type of its field ‘PythonCallInDestructor::d’ [-Wattributes]
 struct PythonCallInDestructor {
        ^~~~~~~~~~~~~~~~~~~~~~
```

That's probably minor enough to ignore, but we can at least mention it in the release notes.

dean0x7d:
What about limiting the visibility attribute to the `detail` namespace?

jagerman:
I'm not confident that that's going to fix the issue under macOS: e.g. some non-hidden, non-`detail` function resolves to a 2.1 definition module calls something in `detail` that in turn calls `get_internals()` and gets the internals pointer for the wrong version.

I'm not particular familiar with the OS X `ld`, but I wonder if there is some optional flags that might address this in a nicer way?  Otherwise it seems that there is all sorts of potential for conflict for mixed-version modules when `-fvisibility=hidden` isn't turned on.

jagerman:
Indeed I just verified--it's not enough to just force it on `detail`.

dean0x7d:
The same issue can be reproduced on Linux by changing Python's `dlopenflags` to `RTLD_GLOBAL`, which makes perfect sense. For pybind11's test suite, this can be done in `conftest.py`:
```python
if sys.platform.startswith("linux"):
    sys.setdlopenflags(os.RTLD_NOW | os.RTLD_GLOBAL)
```
This reproduces the issue when compiled without `-fvisibility=hidden` (i.e. in debug mode when building with CMake).

On the macOS side, this can't be configured at run time, but there's a [compile-time option](https://github.com/python/cpython/blob/c82b7f332aff606af6c9c163da75f1e86514125e/Python/dynload_next.c#L14-L21) which looks to be enabled by default (at least with the homebrew build; haven't looked into others yet).

So it seems like there are two possible solutions here:

1. Always compile pybind11 modules with `-fvisibility=hidden`. This is pretty simple in CMake, but it would need to be documented for other build systems. The downside here is that the functionality of `py::module_local` is tied to a specific (off-by-default) compiler flag.

2. Add the visibility attribute to pybind11's namespaces. The nice part is that it's independent of the build system, but it has the downside of the attribute warning mentioned above (which may be annoying to track down and work around on the user side).

Right now I'm leaning toward solution 1. It seems a bit cleaner and it would only require user intervention for custom build systems in conjunction with `py::module_local`. On the other hand, the warnings produced by the namespace attribute would be a lot more common.

jagerman:
I don't see why it would only affect `py::module_local`, though; if, say, `make_new_instance()` could randomly go to either the 2.1 or 2.2 version there are going to be many potential segfaults.

The warning that comes up under 2. *only* comes up if you're *not* using `-fvisibility=hidden`; if we're going to enforce 1. anyway, the downside of 2. disappears.

dean0x7d:
> I don't see why it would only affect py::module_local, though; if, say, make_new_instance() could randomly go to either the 2.1 or 2.2 version there are going to be many potential segfaults.

Good point. Although, the versioning issues have a possible alternative solution of using an inline namespace (which could similarly be added to the namespace macro), but this isn't an option for `module_local`.

One other thing to consider with the namespace attribute is that it would need to be applied consistently, but there are cases where the namespace macro isn't used, like used-defined casters and also a couple of places internally. That could also result in mixed visibility issues. My main worry is that it could be fragile.

> what I mean is, if we require 1., it's harmless to do 2., but doing 2. allows things to still work if the user ignores 1. anyway 

So you're suggesting to do both the compiler flag and namespace attribute?

jagerman:
> So you're suggesting to do both the compiler flag and namespace attribute?

Yeah.  Then you'll only actually run into warnings if you forget the `-fvisibility=hidden`, but the warnings might actually be helpful in that case in notifying you that you forgot the flag.

jagerman:
Ideally we would add a warning ourselves if not compiling under `-fvisibility=hidden`, but I don't know if there's a way to do that.

dean0x7d:
> Yeah. Then you'll only actually run into warnings if you forget the -fvisibility=hidden, but the warnings might actually be helpful in that case in notifying you that you forgot the flag.

OK, that sounds good. As far as I've gathered, you've already been experimenting with the namespace attributes. Would you be willing to turn that into a PR along with the debug visibility flag change?

jagerman:
> Would you be willing to turn that into a PR along with the debug visibility flag change?

Done in #995.  We can carry on discussion there.

PR #951, opened 2017-07-19
Fix builtin exception handlers to work across modules

jagerman:
The builtin exception handler currently doesn't work across modules under clang/libc++ for builtin pybind exceptions like `pybind11::error_already_set` or `pybind11::stop_iteration`: under RTLD_LOCAL module loading clang considers each module's exception classes distinct types (this is essentially the same as issue #912).  This then means that the base exception translator fails to catch the exceptions and falls through to the generic `std::exception` handler, which completely breaks things like `stop_iteration`: only the `stop_iteration` of the first module loaded actually works properly.  Later modules raise a RuntimeError with no message when their iterators try to signal the end.

For example, two modules defined like this exhibit the behaviour under clang++/libc++:

z1.cpp:
```C++
#include <pybind11/stl_bind.h>
namespace py = pybind11;
PYBIND11_MODULE(z1, m) {
    py::bind_vector<std::vector<long>>(m, "IntVector");
}
```

z2.cpp:
```C++
#include <pybind11/stl_bind.h>
namespace py = pybind11;
PYBIND11_MODULE(z2, m) {
    py::bind_vector<std::vector<double>>(m, "FloatVector");
}
```

Python:
```Python
import z1, z2
for i in z2.FloatVector():
    pass
```

results in:
```
  File "zs.py", line 2, in <module>
    for i in z2.FloatVector():
RuntimeError
```
while if you remove `z1` from the import the code runs.  (You could also change the order of the import to `z2, z1` which lets it run, but would break `z1.IntVector`'s iterator).

This commit fixes the issue by adding a new exception translator each time the internals pointer is initialized from python builtins: this generally means the internals data was initialized by some other module.  (The extra translator(s) are skipped under libstdc++).

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -75,6 +75,23 @@ PYBIND11_NOINLINE inline internals &get_internals() {
     const char *id = PYBIND11_INTERNALS_ID;
     if (builtins.contains(id) && isinstance<capsule>(builtins[id])) {
         internals_ptr = *static_cast<internals **>(capsule(builtins[id]));
+
+        // We loaded builtins through python's builtins, which means that our error_already_set and
+        // builtin_exception may be different local classes than the ones set up in the initial
+        // exception translator, below, so add another for our local exception classes.
+        //
+        // stdlibc++ doesn't require this (types there are identified only by name)
+        #if !defined(__GLIBCXX__)
+        internals_ptr->registered_exception_translators.push_front(
+            [](std::exception_ptr p) -> void {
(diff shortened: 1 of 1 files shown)


dean0x7d:
Looks good to me, but adding a test would be nice. See my comment in #949 about using the new tests module for cross-module testing.

jagerman:
Closing (this will be merged via #949, which depends on this fix, but also adds the cross module testing apparatus this PR needs).

PR #952, opened 2017-07-20
Add support custom sized operator deletes

jagerman:
If a class doesn't provide a `T::operator delete(void *)` but does have
a `T::operator delete(void *, size_t)` the latter is invoked by a
`delete someT`.  Pybind currently only looks for and calls the former;
this commit adds detection and calling of the latter when the former
doesn't exist.

### DIFF
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -127,3 +127,55 @@ def test_implicit_conversion_life_support():
     assert m.implicitly_convert_variable(UserType(5)) == 5
 
     assert "outside a bound function" in m.implicitly_convert_variable_fail(UserType(5))
+
+
+def test_operator_new_delete(capture):
+    """Tests that class-specific operator new/delete functions are invoked"""
+
+    class SubAliased(m.AliasedHasOpNewDelSize):
+        pass
+
+    with(capture):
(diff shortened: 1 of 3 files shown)


tests/test_class.py
```diff
@@ -127,3 +127,55 @@ def test_implicit_conversion_life_support():
...
     assert "outside a bound function" in m.implicitly_convert_variable_fail(UserType(5))
+
+
+def test_operator_new_delete(capture):
+    """Tests that class-specific operator new/delete functions are invoked"""
+
+    class SubAliased(m.AliasedHasOpNewDelSize):
+        pass
+
+    with(capture):
```
dean0x7d: Unnecessary parentheses here and below.

tests/test_class.py
```diff
@@ -127,3 +127,55 @@ def test_implicit_conversion_life_support():
...
+        a = m.HasOpNewDel()
+        b = m.HasOpNewDelSize()
+        d = m.HasOpNewDelBoth()
+    assert capture == """
+    A new 8
```
dean0x7d: For style consistency: capture triple quote strings are usually indented.

PR #954, opened 2017-07-21
Simplify error_already_set

### LINKED ISSUE #927: Link error when using LLVM or Intel Toolset under Visual Studio
When using the last LLVM and Intel C++ Compiler versions (4.0 and 17.0 respectively) under Visual Studio, the following error occurs (both with 2.1.1 and the current master):

``
error LNK2001: unresolved external symbol "public: __cdecl pybind11::error_already_set::error_already_set(class pybind11::error_already_set const &)" (??0error_already_set@pybind11@@QEAA@AEBV01@@Z)
1>C:\Dev\PythonLibs\Tests\Pybind11\build\bin\Release\pybind11test.cp36-win_amd64.pyd : fatal error LNK1120: 1 unresolved externals
``

The following code was used:

```c++

namespace py = pybind11;

struct S
{
	int value;

	int get() const { return value; }
	void set(int v) { value = v; }
};


PYBIND11_MODULE(pybind11test, m)
{
	py::class_<S>(m, "S")
		.def(py::init<>())
		.def("get", &S::get)
		.def("set", &S::set)
		.def_readwrite("spectrumType", &S::value);
}
```

I was able to fix this problem by changing line 704 in common.h (from the master branch):
```c++
error_already_set(const error_already_set &) = delete;
```
to
```c++
error_already_set(const error_already_set &) = default;
```

But I'm not sure this is the right way to fix the problem though.

jagerman:
`error_already_set` is more complicated than it needs to be, partly because it manages reference counts itself rather than using `py::object`, and partly because it tries to do more exception clearing than is needed.  This commit greatly simplifies it, and (hopefully) fixes #927.

I hope I'm not overlooking something, but I think the simplifications I'm making here make sense.

First, using `py::object` instead of `PyObject *` means we can rely on implicit copy/move constructors.  (This necessitated a move into `pytypes.h`, but that doesn't seem a terrible fit given the purpose of `error_already_set`).

The current logic did both a `PyErr_Clear` on deletion *and* a `PyErr_Fetch` on creation.  I can't see how the `PyErr_Clear` on deletion is ever useful: the `Fetch` on creation itself clears the error, so the only way doing a `PyErr_Clear` on deletion could do anything if is something else set the current python exception while the `error_already_set` object was alive--but in that case, clearing some potentially-unrelated other exception seems wrong.  (Code that is worried about an exception handler raising another exception would already catch a second `error_already_set` from exception code).

The destructor itself called `clear()`, but `clear()` was a little bit convoluted: it called `restore()` to restore the currently captured error, but then immediately cleared it with `PyErr_Clear();`.  The only point seems to be because `PyErr_Restore` decrements the reference counts on the held objects, but we can simply do that manually and avoid restoring the error just to clear it again.  Or, actually, we can just *not* decrement them at all and let them be unreferenced (via the `py::object` destructor) when the `error_already_set` is itself destroyed.

`clear()` also had the side effect of clearing the current error, even if the current `error_already_set` didn't have a current error (e.g. because of a previous `restore()` or `clear()` call).  I don't really see how clearing the error here is ever the right thing to do: since the error is cleared on construction, the only way there could be a current error is if you called `restore()` (in which case the current stored error-related members have already been released), or if some *other* code raised the error, in which case `clear()` on *this* object is clearing an error for which it shouldn't be responsible.

Neither of those seem like intentional or desirable features, and manually requesting deletion of the stored references similarly seems pointless, so I've just made `clear()` an empty method and marked it
deprecated in case any code out there is calling it.

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -231,7 +231,6 @@ extern "C" {
         try {                                                                  \
             return pybind11_init();                                            \
         } catch (pybind11::error_already_set &e) {                             \
-            e.clear();                                                         \
             PyErr_SetString(PyExc_ImportError, e.what());                      \
             return nullptr;                                                    \
         } catch (const std::exception &e) {                                    \
@@ -278,7 +277,6 @@ extern "C" {
             pybind11_init_##name(m);                                           \
             return m.ptr();                                                    \
         } catch (pybind11::error_already_set &e) {                             \
-            e.clear();                                                         \
             PyErr_SetString(PyExc_ImportError, e.what());                      \
             return nullptr;                                                    \
         } catch (const std::exception &e) {                                    \
@@ -352,8 +350,6 @@ inline static constexpr int log2(size_t n, int k = 0) { return (n <= 1) ? k : lo
 
 // Returns the size as a multiple of sizeof(void *), rounded up.
 inline static constexpr size_t size_in_ptrs(size_t s) { return 1 + ((s - 1) >> log2(sizeof(void *))); }
-
-inline std::string error_string();
 
 /**
  * The space to allocate for simple layout instance holders (see below) in multiple of the size of
@@ -702,36 +698,6 @@ template<typename T> T& get_or_create_shared_data(const std::string& name) {
     }
     return *ptr;
 }
-
-/// Fetch and hold an error which was already set in Python
-class error_already_set : public std::runtime_error {
-public:
-    error_already_set() : std::runtime_error(detail::error_string()) {
-        PyErr_Fetch(&type, &value, &trace);
-    }
-
-    error_already_set(const error_already_set &) = delete;
-
-    error_already_set(error_already_set &&e)
-        : std::runtime_error(e.what()), type(e.type), value(e.value),
-          trace(e.trace) { e.type = e.value = e.trace = nullptr; }
-
-    inline ~error_already_set(); // implementation in pybind11.h
-
-    error_already_set& operator=(const error_already_set &) = delete;
-
-    /// Give the error back to Python
-    void restore() { PyErr_Restore(type, value, trace); type = value = trace = nullptr; }
-
-    /// Clear the held Python error state (the C++ `what()` message remains intact)
-    void clear() { restore(); PyErr_Clear(); }
-
-    /// Check if the trapped exception matches a given Python exception class
-    bool matches(PyObject *ex) const { return PyErr_GivenExceptionMatches(ex, type); }
-
-private:
-    PyObject *type, *value, *trace;
-};
 
 /// C++ bindings of builtin Python exceptions
 class builtin_exception : public std::runtime_error {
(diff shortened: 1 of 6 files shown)


dean0x7d:
The extra work for exception clearing originates and is explained in 099d6e9c48e5a2a76c4bad375e33d58e0834e587. The first part about acquiring the GIL in `~error_already_set()` is definitely important and should be added back to this PR. (Would be nice to add a regression test for that if it's not too elaborate.)

The same commit also makes mention of always using `PyErr_Fetch` and `PyErr_Restore` in pairs instead of simply decrementing the refcount. But I don't think this is necessary. Python's docs include examples where `type, value, trace` are simply decremented (e.g. [here](https://docs.python.org/3/faq/extending.html#how-do-i-tell-incomplete-input-from-invalid-input) toward the end of the second code block). Avoiding the `PyErr_Restore`/`PyErr_Clear` combo definitely simplifies `error_already_set`, and it looks like that's fine with the CPython API.

So I'd say the `py::object`-base approach in this PR is pretty nice as long as the GIL destructor is added back. `PyErr_Restore`/`PyErr_Clear` is not necessary, unless I'm overlooking something.

jagerman:
Ah, right, I suppose the `error_already_set` could be thrown from within a `gil_scoped_acquire` and caught outside it.

jagerman:
@wjakob - do you recall what the motivation was to always match `Fetch`/`Restore`?

wjakob:
IIRC the restore & clear on destruction was just paranoia regarding reference counting, I wasn't sure if decreasing the refcounts of the three objects representing the exception would be enough.

Another part that took a few iterations to work correctly is the propagation of stack traces through nested calls with C++ -> Python -> C++ -> .. transitions. As long as that still works with the proposed change, I'm fine with it.

jagerman:
I've added some tests for the exception round-tripping, which seems to work as expected.

dean0x7d:
I would very much like to recommend against the `object.reset()` function. It signals that it's expected that a `py::object` can be in a `null` state (in addition to `None`). It adds an additional level of uncertainty for users. If a function returns a `py::object` is it guaranteed to be in a valid state or must the user do an explicit check? So far, the answer has pretty much always been that it is a valid object and there's no *convenient* way to invalidate it on the user side (there is currently a way, just not a convenient one).

I think we should be working in the opposite direction and make sure that `py::object`s are never `nullptr` except after `std::move`.

jagerman:
The point about not exposing it publically makes sense to me.  In this particular case, though, I think I'd need either a `o.release().dec_ref()` or a `o = object();` to clear the object before destruction.  Any preference on either of those?

Another alternative is to make `~object()` take out the `gil_scoped_acquire` before decrementing the reference.  I'm not sure how much overhead this would add to the gil-already-head common case, though.

dean0x7d:
`o.release().dec_ref()` seem good to me. Acquiring the GIL for every `~object()` would add overhead even if the GIL is already held.

jagerman:
> Acquiring the GIL for every `~object()` would add overhead even if the GIL is already held.

More than you would think: I just tried it to see if it works and it segfaults: the `gil_scoped_acquire` constructor apparently triggers a `py::object` destruction, resulting in a segfault while attempting to obtain an infinite number of `gil_scoped_acquire`s.

PR #962, opened 2017-07-25
Only allow unchecked()/mutable_unchecked() on an lvalue

### LINKED ISSUE #961: Could unchecked views prevent garbage collection of the underlying arrays?
## Issue description

Unchecked views appear to not prevent garbage collection of the underlying array (object).  Unless I am mistaken, it should be possible for them to incref the underlying object upon construction of the unchecked view and decref it upon destruction, for a minimal cost; I do not see a case where the opposite behavior is desirable.

## Reproducible example code

```
#include <pybind11/numpy.h>

namespace py = pybind11;

int foo() {
    auto a =
        py::module::import("numpy").attr("zeros")(1)
        .cast<py::array_t<double>>().mutable_unchecked<1>();
    auto b =
        py::module::import("numpy").attr("zeros")(1)
        .cast<py::array_t<double>>().mutable_unchecked<1>();
    a(0) = 1;
    // Rely on the fact that after a and b are (usually) allocated at the same position as the
    // first one has been immediately garbage collected.
    py::print(a(0), b(0));
}

PYBIND11_PLUGIN(python_example) {
    py::module m("python_example");
    m.def("foo", &foo);
    return m.ptr();
}
```

```
$ python -c 'from python_example import *; foo()'
1.0 1.0  # Should be 1.0 0.0 if a did not get GC before b gets allocated.
```

jagerman:
This should mitigate accidental invocation on a temporary array.

Fixes #961.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -689,7 +689,7 @@ public:
      * care: the array must not be destroyed or reshaped for the duration of the returned object,
      * and the caller must take care not to access invalid dimensions or dimension indices.
      */
-    template <typename T, ssize_t Dims = -1> detail::unchecked_mutable_reference<T, Dims> mutable_unchecked() {
+    template <typename T, ssize_t Dims = -1> detail::unchecked_mutable_reference<T, Dims> mutable_unchecked() & {
         if (Dims >= 0 && ndim() != Dims)
             throw std::domain_error("array has incorrect number of dimensions: " + std::to_string(ndim()) +
                     "; expected " + std::to_string(Dims));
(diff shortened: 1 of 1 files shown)


dean0x7d:
Seems fine to me -- preferable to keeping the underlying object alive via refcount. References/views in C++ never do that anyway. Even more so in our case, since the type itself (`detail::unchecked_reference`) is an implementation detail, the reference isn't expected to be kept around and the type isn't meant to be named.

As far as I can see, the only use case that might be prevented by the lvalue qualifier is:
```c++
py::array_t<double> get_array();
void foo(detail::unchecked_reference ref);

foo(get_array().unchecked<1>()); // valid temporary lifetime
```
But this isn't very likely, especially because of the `detail::` part.

PR #963, opened 2017-07-26
Update tests to new style

jagerman:
This updates the test suite tests to the new style proposed and started in #898.  This is still in progress.  (I'm adding commits one-by-one for now, but intend to merge them all into one before merging onto `master`).

(Also note that `test_exceptions` and `test_multiple_inheritance` won't be included here—they are covered by PRs #954 and #960, respectively.)

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -799,6 +799,8 @@ struct nodelete { template <typename T> void operator()(T*) { } };
 NAMESPACE_BEGIN(detail)
 template <typename... Args>
 struct overload_cast_impl {
+    constexpr overload_cast_impl() {} // MSVC 2015 needs this
+
     template <typename Return>
     constexpr auto operator()(Return (*pf)(Args...)) const noexcept
                               -> decltype(pf) { return pf; }
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -196,7 +196,7 @@ def pytest_namespace():
     except ImportError:
         scipy = None
     try:
-        from pybind11_tests import have_eigen
+        from pybind11_tests.eigen import have_eigen
     except ImportError:
         have_eigen = False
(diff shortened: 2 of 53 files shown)


jagerman:
This PR broke the current `py::overload_cast` with ICEs under both MSVC 2015 *and* under clang 3.8 on debian stretch.

I spent some time fighting with MSVC to get it to work, and finally settled on e4fb05a, which manages to simplify it, fix the ICEs on MSVC and clang, *and* makes it C++11-compatible.  I came on the really weird trick to making it work accidentally: dropping the defaulted `std::false_type` argument makes it work.  For some bizarre reason it made MSVC choke on the deductions.  But having *two* functions here--one with the `std::false_type` argument and one without--makes MSVC perfectly happy.

@dean0x7d - do we need the `std::false_type` argument at all?  We could just drop the second function that takes it, but I'm assuming you had a reason for adding the argument in the first place.

dean0x7d:
The solution in e4fb05ab494d531331423111f94e5bbb430b67c5 will not work for overload sets which have a variable number of arguments, e.g. `void foo()` and `void foo(int)`. That's the main reason why the two stage template parameters are needed (class template parameters + static member template parameters). See the `test.cpp` file in https://github.com/dean0x7d/overload_cast for the relevant tests.

The `std::false_type` argument is there to allow programmatic selection of `const` vs non-`const` overloads, based on some type trait. E.g. `overload_cast<int>(&W::foo, std::is_same<T, U>{})`. It's not needed for manual casting, but it's pretty nice for metaprogramming.

jagerman:
I think I've found a fix (for the original `overload_cast` implementation) for MSVC 2015: it seems adding an explicit `constexpr cast_overload_impl() {}` makes it happy.  (But changing the `{}` to `= default;` makes the ICE happen again).

PR #967, opened 2017-07-28
Modify build logic to enable exporting pybind11 when it is a subdirec…

tjcorona:
…tory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,13 @@ project(pybind11)
 set(PYBIND11_MASTER_PROJECT OFF)
 if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   set(PYBIND11_MASTER_PROJECT ON)
+  set(PYBIND11_EXPORT_NAME "${PROJECT_NAME}Targets")
+  set(PYBIND11_INSTALL_EXPORT ON)
+else ()
+  if (NOT PYBIND11_EXPORT_NAME)
+    set(PYBIND11_EXPORT_NAME "${PROJECT_NAME}Targets")
+  endif ()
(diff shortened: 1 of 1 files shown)


tjcorona:
@mathstuf

mathstuf:
The failure seems to be due to the setup of Travis more than this change. Is PyPy really inside of the build directory for the test build?

jagerman:
Can we get some more context on this?  What's the motivating issue, why is this needed, why is the solution doing it *this* way instead of some other way?  (I'm not suggesting that there is necessarily a problem with any of it, but I think those are the basic questions needed to help understand/evaluate a PR).

> Is PyPy really inside of the build directory for the test build?

Yes, but there's no particular reason it has to be.  (It gets downloaded as part of the build, and I think it was slightly easier to just unpack it in `.` (i.e. the pybind11 dir) rather than put it somewhere else.  I've no objection to unpacking it elsewhere, though.)

mathstuf:
There are actually two commits here:

  - One to remove the `BUILD_INTERFACE` on the linking and including of Python. The install of these libraries also need these to happen, so `BUILD_INTERFACE` is not right.
  - The other allows Pybind11 to be used in another project via `add_subdirectory` and have it add its targets to the parent project's `EXPORT` set. When this happens, it is up to the parent project to install that `EXPORT` set. This arises because when Pybind11 is linked to by a library which is installed, CMake notes that a dependency of that library (and is made in this project) is missing from that `EXPORT` set.

@tjcorona Please split the commits and add this info to the commit messages. A third commit (before the fixes) should change the `.travis.ci.yml` file to extract PyPy somewhere not inside of the build directory.

tjcorona:
@mathstuf I think it would be better if you were to submit this PR.

PR #971, opened 2017-08-01
WIP Add a consume call policy

torokati44:
I tried to interface with a C++ library that has a couple of functions which take ownership of some of their pointer parameters.
I could not find a way to call them from Python using pybind11, because they kept being either double-freed or leaked.
An included test demonstrates this use case with a minimal, and not very elegant example.

This is my attempt to add a call policy named consume. It means that the ownership of any parameters marked by it is taken by the called C++ function, and must not be accessed or deleted by Python code or pybind11 after the call.
It is incomplete, because I could not figure out how to make it work in the general case, with custom holder types. I have marked the code with ??? where I tried to do something that did not work out.

Any help is much appreciated.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -44,6 +44,9 @@ template <typename T> struct base {
 /// Keep patient alive while nurse lives
 template <size_t Nurse, size_t Patient> struct keep_alive { };
 
+/// Give ownership of parameter to the called function
+template <size_t Consumed> struct consume { };
+
 /// Annotation indicating that a class is involved in a multiple inheritance relationship
 struct multiple_inheritance { };
 
@@ -117,6 +120,7 @@ enum op_type : int;
 struct undefined_t;
 template <op_id id, op_type ot, typename L = undefined_t, typename R = undefined_t> struct op_;
 inline void keep_alive_impl(size_t Nurse, size_t Patient, function_call &call, handle ret);
+inline void consume_impl(size_t Consumed, function_call &call);
 
 /// Internal data structure which holds metadata about a keyword argument
 struct argument_record {
@@ -448,6 +452,13 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
     static void precall(function_call &) { }
     template <size_t N = Nurse, size_t P = Patient, enable_if_t<N == 0 || P == 0, int> = 0>
     static void postcall(function_call &call, handle ret) { keep_alive_impl(Nurse, Patient, call, ret); }
+};
+
+template <size_t Consumed> struct process_attribute<consume<Consumed>> : public process_attribute_default<consume<Consumed>> {
+    template <size_t C = Consumed, enable_if_t<C != 0, int> = 0>
+    static void precall(function_call &call) {
+        consume_impl(Consumed, call);
+    }
 };
 
 /// Recursively iterate over variadic template arguments
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1371,6 +1371,7 @@ template <typename... Ts> class type_caster<std::tuple<Ts...>>
 template <typename T>
 struct holder_helper {
     static auto get(const T &p) -> decltype(p.get()) { return p.get(); }
+    static void release(T &p) { p.release(); }
 };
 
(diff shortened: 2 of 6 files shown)


jagerman:
If I understand correctly, essentially what you're trying to do here is to convert an owned, held reference into an unowned, unheld reference—essentially to turn a regular object into one that would have been created had it been returned with `return_value_policy::reference`.  There are, of course, some dangers involved in doing that (e.g. if a Python instance stays around past the point where the pointer becomes invalid), but that's just the same danger you'd get with `rvp::reference` in the fist place.

Some comments for fixing this up:

- In addition to releasing the holder, you should invoke its destructor, and call `set_holder_constructed(false)` on the `value_and_holder`.  (That call isn't in `master` yet, but will be soon as part of #805).
- You need to set `owned = false` on the instance so that the pointed-to memory doesn't get deallocated when the Python variable gets deallocated.
- Related to that, we're now going to be in a situation where `owned` applies at the pointer level rather than the `instance` level.  (Currently that can't happen because the only way to get `owned = false` is through a cast, but a cast can only ever give you a single-pointer instance so this isn't an issue).  With Python-side multiple inheritance, this would result in the consumed pointer being unowned while other, unconsumed pointers in the instance (for the other C++ types) are still owned.  #805 adds status bits for non-simple layouts, so moving this into `value_and_holder` (and renaming `owned` to `simple_owned` for the simple holder layout) would probably be fairly easy.
- I think it makes more sense to propagate this feature through the `py::arg` mechanism rather than needing to specify the argument number.  Thus you would do something like:
    ```C++
    m.def("consume", [](int a, Type *b, Type2 *c) { ... },
          py::arg(), py::arg(), py::arg().disown());
    ```
    to indicate that the argument should be converted to unowned before calling the function (if not already disowned).  (And of course, you could add keyword names, e.g. `py::arg("c").disown()`).
- You'll have to deal with the `unique_ptr` assumption, and also the assumption that `.release()` will work.  Adding a static method to `holder_helper` would probably work (it could also have a `static_assert` fallback for holder types without a `release()` so that attempting to use the feature on a type without a `.release()` wouldn't allow the feature to be specified).

Another idea worth thinking about:
- it might be nice for symmetry to be able to go the other way, i.e. to reclaim ownership after the call (so, essentially, to convert a `reference` into a `take_ownership`), via something like `py::arg().take_ownership()`.

torokati44:
> If I understand correctly, essentially what you're trying to do here is to convert an owned, held reference into an unowned, unheld reference

Yes, either that, or just completely getting rid of the reference to the object from the Python side, if that's possible at all. However, the solution you described would be just fine too, if not better.

Otherwise, thank you for your valuable insight, I will consider your comments thoroughly, and hopefully come up with a more acceptable solution based on them in the future.

jagerman:
> just completely getting rid of the reference to the object from the Python side, if that's possible at all.

Unfortunately, it isn't: we can't do anything about references to the instance held elsewhere.  At best you could replace the pointer with something else, but then the question is what "something else" should be--and since it has to be some sort of object that you're going to have to duplicate, you might as well just consume the duplicate.  (I suppose you could add some sort of error detection for trying to use an instance that is already cleared, but that's going a bit far into esoteric territory).

torokati44:
I meant to removing that single reference that was passed to the function. Something like an implicit del statement to that name. If other references are held to it... Well, I don't think we can/should do much about them. But never mind, deleting is not that important.

And if I make the reference into a non-holding weak one, is that going to apply to all Python references? Is there only a single holder for a given instance that is used by all Python references, or all references have their own holders?

jagerman:
I think the key thing to realize is that every variable in Python is essentially a shared reference; the instance that gets passed into the function is the same instance held by other copies of that variable.  So yes, any changes to one variable are going to affect every other reference to that variable.  The holder itself is unique, by default, but the holder is inside the *object*; each python variable is simply a reference to that one object.

torokati44:
Awesome, thanks! I knew how Python references work, just wasn't sure if there isn't any black magic to somehow keep track of all Python references to an object, and assign a different holder to all of them. However, in hindsight, I'm not sure why this even came up in my mind, it would be kind of silly and pointless.

torokati44:
Superseded by #1226. See details for closing there.

PR #978, opened 2017-08-03
Free created types' tp_name in default metaclass

### LINKED ISSUE #977: Possible leak in make_new_python_type()?
Sounds like something fails to dealloc properly?

```cpp

struct A { int x };

PYBIND11_PLUGIN(test1) {
    pybind11::module m("test1");
    pybind11::class_<A>(m, "A", "A");
    return m.ptr();
}
```

```sh
$ valgrind --tool=memcheck --leak-check=full \
    --show-possibly-lost=no python -c 'import test1' 
```

Yields:

```valgrind
==138390== 8 bytes in 1 blocks are definitely lost in loss record 4 of 305
==138390==    at 0x4C29C3D: malloc (in /opt/rh/devtoolset-4/root/usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==138390==    by 0x5EDCAB9: strdup (in /usr/lib64/libc-2.17.so)
==138390==    by 0xCBB5864: pybind11::detail::make_new_python_type(pybind11::detail::type_record const&) (class_support.h:559)
==138390==    by 0xCBB9008: pybind11::detail::generic_type::initialize(pybind11::detail::type_record const&) (pybind11.h:829)
==138390==    by 0xCBBCCD2: pybind11::class_<A<> >::class_<char [2]>(pybind11::handle, char const*, char const (&) [2]) (pybind11.h:986)
==138390==    by 0xCBAD108: pybind11_init() (test1.cpp:9)
==138390==    by 0xCBACFE6: PyInit_test1 (test1.cpp:7)
==138390==    by 0x4F952DA: _PyImport_LoadDynamicModuleWithSpec (importdl.c:154)
==138390==    by 0x4F9206E: _imp_create_dynamic_impl (import.c:2027)
==138390==    by 0x4F9206E: _imp_create_dynamic (import.c.h:282)
==138390==    by 0x4EF2D98: PyCFunction_Call (methodobject.c:109)
==138390==    by 0x4F79CE2: ext_do_call (ceval.c:5046)
==138390==    by 0x4F79CE2: PyEval_EvalFrameEx (ceval.c:3290)
==138390==    by 0x4F7BFC8: _PyEval_EvalCodeWithName (ceval.c:4033)
==138390==

```

I've checked this on:

- Python 3.5 / Python 3.6 (conda, Linux)
- pybind11: both v2.1 and latest master

with the same results.

jagerman:
Fixes #977, via a weakref added to the type.

This feels like a bit of overkill, though; the leak in #977 seems fairly minor.

### DIFF
--- a/include/pybind11/common.h
+++ b/include/pybind11/common.h
@@ -484,6 +484,7 @@ struct internals {
     std::forward_list<void (*) (std::exception_ptr)> registered_exception_translators;
     std::unordered_map<std::string, void *> shared_data; // Custom data to be shared across extensions
     std::vector<PyObject *> loader_patient_stack; // Used by `loader_life_support`
+    std::forward_list<std::string> static_strings; // Stores the std::strings backing detail::c_str()
     PyTypeObject *static_property_type;
     PyTypeObject *default_metaclass;
     PyObject *instance_base;
@@ -677,6 +678,16 @@ inline void ignore_unused(const int *) { }
 using expand_side_effects = bool[];
 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
 #endif
+
+/// Constructs a std::string with the given arguments, stores it in `internals`, and returns its
+/// `c_str()`.  Such strings objects have a long storage duration -- the internal strings are only
+/// cleared when the program exits or after interpreter shutdown (when embedding), and so are
+/// suitable for c-style strings needed by Python internals (such as PyTypeObject's tp_name).
+template <typename... Args> const char *c_str(Args &&...args) {
+    auto &strings = get_internals().static_strings;
+    strings.emplace_front(std::forward<Args>(args)...);
+    return strings.front().c_str();
+}
 
 NAMESPACE_END(detail)
 
--- a/include/pybind11/class_support.h
+++ b/include/pybind11/class_support.h
@@ -518,12 +518,11 @@ inline PyObject* make_new_python_type(const type_record &rec) {
             module = rec.scope.attr("__name__");
     }
 
+    auto full_name = c_str(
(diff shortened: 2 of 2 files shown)


aldanor:
:+1:   Don't think it's an overkill, looks quite straightforward. Keeping the the codebase valgrind-clean is nice.

What about `strdup()` of docstrings, wonder if it should be done the same? (in `initialize_generic()`, IIRC)

jagerman:
I tried that, but at least with a simple approach it results in invalid frees because the metaclass's `tp_dealloc` gets called for Python-inherited types as well, where `free()`ing is definitely wrong.

jagerman:
Another way we could go about this is to store the type names in a `std::forward_list<std::string>` in `internals`.  Then at least it'll be destroyed when the interpreter is finalized (which AFAICT is the only way a pybind-created type is going to be destroyed anyway).

include/pybind11/common.h
```diff
@@ -484,6 +484,7 @@ struct internals {
     std::forward_list<void (*) (std::exception_ptr)> registered_exception_translators;
     std::unordered_map<std::string, void *> shared_data; // Custom data to be shared across extensions
     std::vector<PyObject *> loader_patient_stack; // Used by `loader_life_support`
+    std::forward_list<std::string> type_names; // Stores the std::strings holding created types' tp_names
```
aldanor: Minor: might as well name it `static_strings` or the like? (unlike all other fields of `internals`, this one you would never actually read, it's just there for storage with static lifetime and doesn't really have to do much with types)

Also, in case we ever need it again?

```cpp
const char* detail::strdup(const std::string& what) {
    auto& internals = get_internals();
    internals.static_strings.emplace_front(what);
    return internals.static_strings.front().c_str();
}

aldanor:
Looks great. Merge?

PR #980, opened 2017-08-04
Added catkin package definition

VladimirIvan:
Adds native support for the Catkin build system (ROS compatibility)

Additions:
 - ```catkin/package.xml``` Package description.
 - ```catkin/CMakeLists.txt``` Replicates some of the top level CMakeLists but uses Catkin specific paths.
 - ```catkin/cmake/pybind.cmake``` Splits out the ```pybind_add_module``` macro and several other commands that have to be included in every package that depends on this code.

This will turn the code into a catkin package. I have found only [one](https://github.com/ethz-asl/voxblox) or two instance of pybind11 being used in ROS so far. Native support will make integration into ROS a lot easier.

Additional notes:
The list of header files has to be updated manually (same as in the top level ```CMakeLists.txt``` and in ```setup.py```). Version, maintainer info and other bit should also be updated in ```catkin/package.xml``` if they change in the future.

Catkin/ROS only support python2.7. To get around this, it is possible to compile the pybind11 package by setting the ```PYBIND_PYTHON_EXECUTABLE``` cmake variable to the path to desired python executable. This will only change the python executable/library paths for compilation of pybind11 modules while the rest of the build system will keep using python2.7. The result will be of limited use within ROS but the code will compile nonetheless.

There is no particular test that can be added and maintained cross platform since Catkin/ROS is compatible with only a handful OSs (mainly Ubuntu). To manually test the compatibility:
 1. Follow [these](http://catkin-tools.readthedocs.io/en/latest/quick_start.html) or [these](http://wiki.ros.org/catkin/Tutorials) instructions to create a catkin workspace.
 1. Clone pybind11 into the source directory.
 1. Create a second test package that will wrapp some c++ code.
     1. Make the test package depend on pybind11.
     1. Add a target using the ```pybind_add_module``` macro (instead of the native ```pybind11_add_module``` one).
 1. Compile and source the workspace (Catkin sets up the sys.path when the workspace is sourced).
 1. The package can now be imported into python and used.

### DIFF
--- a/catkin/CMakeLists.txt
+++ b/catkin/CMakeLists.txt
@@ -1,0 +1,74 @@
+cmake_minimum_required(VERSION 2.8.12)
+project(pybind11)
+
+find_package(catkin REQUIRED)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+include(pybind)
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+
+file(MAKE_DIRECTORY include)
+catkin_package(
+  INCLUDE_DIRS include ${EIGEN3_INCLUDE_DIR}
+  LIBRARIES ${PYTHON_LIBRARIES}
+  CATKIN_DEPENDS
+  CFG_EXTRAS pybind.cmake
+)
+
+set(PYBIND11_HEADERS
+  include/pybind11/attr.h
+  include/pybind11/buffer_info.h
+  include/pybind11/cast.h
+  include/pybind11/chrono.h
+  include/pybind11/class_support.h
+  include/pybind11/common.h
+  include/pybind11/complex.h
+  include/pybind11/descr.h
+  include/pybind11/eigen.h
+  include/pybind11/embed.h
+  include/pybind11/eval.h
+  include/pybind11/functional.h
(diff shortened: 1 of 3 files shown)


catkin/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+install(FILES ${PYBIND11_HEADERS}
+        DESTINATION include/pybind11)
+
+set(PYBIND11_CMAKECONFIG_INSTALL_DIR "${CATKIN_PACKAGE_SHARE_DESTINATION}/cmake/" CACHE STRING "install path for pybind11Config.cmake")
+
+configure_package_config_file(../tools/${PROJECT_NAME}Config.cmake.in
+                              "${CMAKE_CURRENT_BINARY_DIR}/pybindConfig.cmake"
```
wxmerkt: This is a typo I've just discovered in our local copy - it should be ``pybind11Config`` consistently, otherwise catkin with an install step will fail

wjakob:
Hi Vladimir,

unfortunately we have to be fairly restrictive in the number of wrappers/build system generators that we can directly include here in the project, or this project would quickly get rather cluttered (there are simply too many). It's probably best if you provide your patch in a separate repository.

Thanks,
Wenzel

PR #984, opened 2017-08-06
reshape for numpy arrays

ncullen93:
implemented reshape, since `resize` is rarely useful and often fails.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -166,6 +166,8 @@ struct npy_api {
     PyObject *(*PyArray_Squeeze_)(PyObject *);
     int (*PyArray_SetBaseObject_)(PyObject *, PyObject *);
     PyObject* (*PyArray_Resize_)(PyObject*, PyArray_Dims*, int, int);
+    PyObject* (*PyArray_Newshape_)(PyObject*, PyArray_Dims*, int);
+
 private:
     enum functions {
         API_PyArray_GetNDArrayCFeatureVersion = 211,
@@ -176,6 +178,7 @@ private:
         API_PyArray_DescrFromScalar = 57,
         API_PyArray_FromAny = 69,
         API_PyArray_Resize = 80,
+        API_PyArray_Newshape = 135,
         API_PyArray_CopyInto = 82,
         API_PyArray_NewCopy = 85,
         API_PyArray_NewFromDescr = 94,
@@ -207,6 +210,7 @@ private:
         DECL_NPY_API(PyArray_DescrFromScalar);
         DECL_NPY_API(PyArray_FromAny);
         DECL_NPY_API(PyArray_Resize);
+        DECL_NPY_API(PyArray_Newshape);
         DECL_NPY_API(PyArray_CopyInto);
         DECL_NPY_API(PyArray_NewCopy);
         DECL_NPY_API(PyArray_NewFromDescr);
@@ -216,6 +220,7 @@ private:
         DECL_NPY_API(PyArray_GetArrayParamsFromObject);
         DECL_NPY_API(PyArray_Squeeze);
         DECL_NPY_API(PyArray_SetBaseObject);
+
 #undef DECL_NPY_API
         return api;
     }
@@ -730,6 +735,19 @@ public:
         if (!new_array) throw error_already_set();
         if (isinstance<array>(new_array)) { *this = std::move(new_array); }
     }
+
+    void reshape(ShapeContainer new_shape) {
+        detail::npy_api::PyArray_Dims d = {
+            new_shape->data(), int(new_shape->size())
+        };
+        // try to reshape, set ordering param to -1 cause it's not used anyway
+        object new_array = reinterpret_steal<object>(
+              detail::npy_api::get().PyArray_Newshape_(m_ptr, &d, 0)
+        );
+        if (!new_array) throw error_already_set();
+        if (isinstance<array>(new_array)) { *this = std::move(new_array); }
+    }
+    
 
     /// Ensure that the argument is a NumPy array
     /// In case of an error, nullptr is returned and the Python error is cleared.
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -292,4 +292,17 @@ TEST_SUBMODULE(numpy_array, sm) {
         std::fill(a.mutable_data(), a.mutable_data() + a.size(), 42.);
         return a;
     });
+
+    m.def("array_reshape1", [](py::array_t<double> a, size_t N) {
+        a.reshape({N, N, N});
+        return a;
+    });
+
+    m.def("create_and_reshape", [](size_t N, size_t M, size_t O) {
+        py::array_t<double> a;
+        a.resize({N*M*O});
+        std::fill(a.mutable_data(), a.mutable_data() + a.size(), 42.);
+        a.reshape({N, M, O});
+        return a;
+    });
 }
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -293,16 +293,20 @@ TEST_SUBMODULE(numpy_array, sm) {
         return a;
     });
 
-    m.def("array_reshape1", [](py::array_t<double> a, size_t N) {
+    sm.def("array_reshape1", [](py::array_t<double> a, size_t N) {
         a.reshape({N, N, N});
         return a;
     });
 
-    m.def("create_and_reshape", [](size_t N, size_t M, size_t O) {
+    sm.def("create_and_reshape", [](size_t N, size_t M, size_t O) {
         py::array_t<double> a;
         a.resize({N*M*O});
         std::fill(a.mutable_data(), a.mutable_data() + a.size(), 42.);
         a.reshape({N, M, O});
+        return a;
+    });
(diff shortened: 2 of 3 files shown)


rwgk:
First impression: looks useful, filling in a blank, not doing anything substantially new or different, therefore unlikely to upset anything.

@Skylion007 if you want to modernize this one (CI green here), I could run all our sanitizers, if that's clean, merge.

I don't think this needs global testing. My testing turn-around could therefore be really quick.

tests/test_numpy_array.py
```diff
@@ -477,6 +477,25 @@ def test_array_create_and_resize(msg):
...
+def test_create_and_reshape(msg):
+    x = m.create_and_reshape(10, 20, 30)
+    assert x.shape == (10, 20, 30)
+
+
+def test_reshape_tuple(msg):
+    a = np.random.randn(10 * 10 * 10).astype("float64")
+    x = m.reshape_tuple(a, (10, 10, 10))
+    assert x.shape == (10, 10, 10)
+    assert a.size == 4
```
Skylion007, line 495: Hmm, I don't see how these test cases are suppose to be right at all.

tests/test_numpy_array.py
```diff
@@ -477,6 +477,25 @@ def test_array_create_and_resize(msg):
...
 
 
+def test_array_reshape(msg):
+    a = np.random.randn(10 * 10 * 10).astype("float64")
+    x = m.array_reshape1(a, 10)
+    assert x.shape == (10, 10, 10)
+
+
+def test_create_and_reshape(msg):
+    x = m.create_and_reshape(10, 20, 30)
```
Skylion007, line 487: @henryiii Thoughts on how to deal with this PyPy specific error?

tests/test_numpy_array.py
Skylion007: This was my bad, a merge conflict.

Skylion007:
@rwgk @henryiii , I fixed this PR.

Skylion007 (approved):

include/pybind11/numpy.h
```diff
@@ -790,6 +795,15 @@ class array : public buffer {
         if (isinstance<array>(new_array)) { *this = std::move(new_array); }
     }
 
+    array reshape(ShapeContainer new_shape) {
```
rwgk, line 798: while we're at it, I'd also plumb through the `order`.
even if it's not actually used now (idk), it may be in future numpy versions.
easier to just do it than to explain why not.

Skylion007: I think the opposite and it may have been used in older versions and its kept on for backwards compat.

Skylion007: @rwgk We would need to add support for it as a function arg, which we currently don't. That can be added in a followup PR.

rwgk: > That can be added in a followup PR.
OK

include/pybind11/numpy.h
```diff
@@ -213,6 +215,7 @@ struct npy_api {
         API_PyArray_DescrFromScalar = 57,
         API_PyArray_FromAny = 69,
         API_PyArray_Resize = 80,
+        API_PyArray_Newshape = 135,
```
henryiii, line 218: Aren't the rest of these in numerical order?

Skylion007: Look closely,, they aren't. ;-;

henryiii: Ahh, true. They are in https://github.com/numpy/numpy/blob/623bc1fae1d47df24e7f1e29321d0c0ba2771ce0/numpy/core/code_generators/numpy_api.py

tests/test_numpy_array.cpp
```diff
@@ -405,6 +405,19 @@ TEST_SUBMODULE(numpy_array, sm) {
         return a;
     });
 
+    sm.def("array_reshape1", [](py::array_t<double> a, size_t N) {
```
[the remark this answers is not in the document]
Skylion007: This did uncover a minor bug. Giving it an empty tuple gives PyBind11 a conversion error when it should be a TypeError to match NumPy's behavior.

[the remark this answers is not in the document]
Skylion007: @rwgk I found it, but i am not sure how to expose the proper ENUM to PyBind11: https://docs.scipy.org/doc/numpy-1.6.0/reference/c-api.array.html#NPY_ORDER

include/pybind11/numpy.h
Skylion007: @henryiii I fixed them.

include/pybind11/numpy.h
Skylion007: There is an NORDER Enum that it takes, but I am not sure how to expose that properly.

include/pybind11/numpy.h
```diff
@@ -203,6 +203,8 @@ struct npy_api {
     // Unused. Not removed because that affects ABI of the class.
     int (*PyArray_SetBaseObject_)(PyObject *, PyObject *);
```
Skylion007, line 204: Since we are altering the ABI, should this function be removed?

include/pybind11/numpy.h
rwgk: sounds good, thanks for looking into it!

tests/test_numpy_array.py
```diff
@@ -477,6 +477,37 @@ def test_array_create_and_resize(msg):
...
+
+def test_reshape_tuple(msg):
+    a = np.random.randn(10, 20, 30).astype("float64")
+    x = m.reshape_tuple(a, (30, 20, 10))
+    assert x.shape == (30, 20, 10)
+
+
+def test_reshape_tuple_empty(msg):
+    a = np.random.randn(10 * 20 * 30).astype("float64")
+    # FIXME: This should be a ValueError for maximum numpy compat
```
rwgk, line 500: Yep, I just spent a couple minutes playing myself, we definitely need to make this compatible in this PR. Otherwise we'd be introducing a wrinkle of the really silly kind that can lead to surprisingly huge stumbles.

If you feel out of your comfort zone, I could help out working on it here. Please let me know.

Skylion007: Yeah, just not sure how to fix the caster code here to allow empty iterable to be coercible into a std::vector<int> if it's empty. Could definitely use some help here, that seems like it might even warrant it's own PR, right?

Skylion007: I get the following error back: 
```
E           TypeError: Unable to convert function return value to a Python type! The signature was
E           	(arg0: numpy.ndarray[numpy.float64], arg1: List[int]) -> numpy.ndarray
```

Skylion007: @rwgk Aha, figured it out. Behavior matches Numpy now.

henryiii (approved):

include/pybind11/numpy.h
```diff
@@ -203,6 +203,8 @@ struct npy_api {
...
     PyObject* (*PyArray_Resize_)(PyObject*, PyArray_Dims*, int, int);
+    PyObject* (*PyArray_Newshape_)(PyObject*, PyArray_Dims*, int);
```
henryiii, line 206: This causes an ABI break?

Skylion007: @henryiii *shrugs* @rwgk thoughts?

rwgk: I don't see any `virtual` here. If there is no vtable, how does adding or removing member functions change the ABI? I'm having doubts about the correctness of the comment in line 203. I could imagine maybe getting into trouble removing a function, but adding? If a newer extension knows about it, the machine code for it will be in that extension for sure. No?

include/pybind11/numpy.h
```diff
@@ -790,6 +795,18 @@ class array : public buffer {
         if (isinstance<array>(new_array)) { *this = std::move(new_array); }
     }
 
+    array reshape(ShapeContainer new_shape) {
+        detail::npy_api::PyArray_Dims d
+            = {reinterpret_cast<Py_intptr_t *>(new_shape->data()), int(new_shape->size())};
+        // try to reshape, set ordering param to 0 cause it's not used anyway
```
rwgk, line 801: I think it's best to remove this comment, and add a comment just above the `array reshape(ShapeContainer new_shape) {` line, e.g.
```
// Optional `order` parameter omitted here, to be added as needed.
```

tests/test_numpy_array.cpp
```diff
@@ -405,6 +405,19 @@ TEST_SUBMODULE(numpy_array, sm) {
...
+        return a.reshape({N, N, N});
+    });
```
rwgk, line 410: To keep the tests lean, I'd remove this function completely (and `test_array_reshape` with it). The exact same interface is covered by `test_array_reshape`. (I realize it takes away from PyPy coverage because of the `xfail` for `test_array_reshape`, but that's not enough reason in my mind to keep this one.)
If you feel differently, I'd at least change this to something like `{N, N+5, N+7}` to get a little bit more value.

rwgk (approved):

rwgk:
Hi @Skylion007, I streamlined the tests. Could you please take a look?
(PyPy doesn't need an xfail anymore.)

henryiii:
These need to go into the docs, too.

PR #986, opened 2017-08-06
Add quick blurb about __init__ with inherited types.

EricCousineau-TRI:
Per discussion on the [`pybind11` Lobby chatroom](https://gitter.im/pybind/Lobby?at=59866327ee5c9a4c5f947878), this adds a quick note about explicitly calling C++ constructors from Python.

Thanks!

\cc @jagerman

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -123,7 +123,7 @@ Bindings should be made against the actual class, not the trampoline helper clas
             .def("go", &PyAnimal::go); /* <--- THIS IS WRONG, use &Animal::go */
 
 Note, however, that the above is sufficient for allowing python classes to
-extend ``Animal``, but not ``Dog``: see ref:`virtual_and_inheritance` for the
+extend ``Animal``, but not ``Dog``: see :ref:`virtual_and_inheritance` for the
 necessary steps required to providing proper overload support for inherited
 classes.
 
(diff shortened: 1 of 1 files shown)


docs/advanced/classes.rst
```diff
@@ -144,6 +144,29 @@ a virtual method call.
     >>> call_go(c)
     u'meow! meow! meow! '
 
+If you are defining a custom constructor in a derived Python class, you *must*
+ensure that you explicitly call the bound C++ constructor using ``__init__``,
+*regardless* of whether it is a default constructor or not. Otherwise, the
+memory for the C++ portion of the instance will be left uninitialized, which
+could do nasty things, such as give you a corrupted ``vtable``.
```
jagerman: One minor wording change: `could do nasty things` could be a bit more technical.  Perhaps: `will generally leave the C++ instance in an invalid state and cause undefined behavior if the C++ instance is subsequently used.`.  Undefined behavior is the C++ technical term for "could do nasty things", of course.

It doesn't automatically do nasty things: you would actually be okay if you never do anything with the object that touches the C++ instance.  You could also call the parent `__init__` *later*, as long as you do it before invoking anything needing the C++ instance.  (I don't know why you'd do this, but it's technically possible).  Also I don't think there's any reason to go into the corrupted vtable description specifically (there are many other ways this could be a problem even if the object doesn't need a vtable; also it isn't so much "corrupted" as never created in the first place).

EricCousineau-TRI: Done. Much more concise, thank you for the explanation!

dean0x7d:
LGTM. Thanks for contributing to the docs!

PR #989, opened 2017-08-08
Add support for boost::variant in C++11 mode

### LINKED ISSUE #988: boost::variant snippet in docs fails to compile
The snippet for using `boost::variable` at  http://pybind11.readthedocs.io/en/latest/advanced/cast/stl.html?highlight=variant? fails to compile. The code
```cpp 
#include <pybind11/stl.h>
#include <boost/variant.hpp>

// `boost::variant` as an example -- can be any `std::variant`-like container
    template <typename... Ts>
    struct type_caster<boost::variant<Ts...>> : variant_caster<boost::variant<Ts...>> {};

    // Specifies the function used to visit the variant -- `apply_visitor` instead of `visit`
    template <>
    struct visit_helper<boost::variant> {
        template <typename... Args>
            -> decltype(boost::apply_visitor(std::forward<Args>(args)...)) {
            return boost::apply_visitor(std::forward<Args>(args)...);
        }
    };
}} // namespace pybind11::detail

PYBIND11_MODULE(cpp, m)
{ m.def("value", []() { boost::variant<int, double> v = 1.1; return v; }); }
```
gives the error
```
/Users/garth/tmp/pybind11-test/pybind11/include/pybind11/stl.h:344:16: error: no matching function for call to 'call'
        return visit_helper<V>::call(variant_caster_visitor{policy, parent},
               ^~~~~~~~~~~~~~~~~~~~~
/Users/garth/tmp/pybind11-test/pybind11/include/pybind11/pybind11.h:157:39: note: in instantiation of function template specialization 'pybind11::detail::variant_caster<boost::variant<int, double> >::cast<boost::variant<int, double> >' requested here
            handle result = cast_out::cast(
                                      ^
/Users/garth/tmp/pybind11-test/pybind11/include/pybind11/pybind11.h:69:9: note: in instantiation of function template specialization 'pybind11::cpp_function::initialize<(lambda at /Users/garth/tmp/pybind11-test/test.cpp:22:18), boost::variant<int, double>, pybind11::name, pybind11::scope, pybind11::sibling>' requested here
        initialize(std::forward<Func>(f),
        ^
/Users/garth/tmp/pybind11-test/pybind11/include/pybind11/pybind11.h:755:22: note: in instantiation of function template specialization 'pybind11::cpp_function::cpp_function<(lambda at /Users/garth/tmp/pybind11-test/test.cpp:22:18), pybind11::name, pybind11::scope, pybind11::sibling, void>' requested here
        cpp_function func(std::forward<Func>(f), name(name_), scope(*this),
                     ^
/Users/garth/tmp/pybind11-test/test.cpp:22:5: note: in instantiation of function template specialization 'pybind11::module::def<(lambda at /Users/garth/tmp/pybind11-test/test.cpp:22:18)>' requested here
{ m.def("value", []() { boost::variant<int, double> v = 1.1; return v; }); }
    ^
/Users/garth/tmp/pybind11-test/test.cpp:14:21: note: candidate template ignored: substitution failure [with Args = <pybind11::detail::variant_caster_visitor, boost::variant<int, double>>]: no matching function for call to 'apply_visitor'
                    ^
1 error generated.
```
with clang. It's a lot longer with gcc! Tested against Boost 1.63 and 1.64 with pybind11 master.

The same code does work against mapbox/variant.

`boost::variable` is super useful and it will be a while before many projects can use `std::variant`, so it might be and idea to add a Boost version to the tests.

dean0x7d:
Resolves #988. 

It also turns out that we can only support `boost::variant` from Boost 1.56 or newer since that's when it started using C++11 variadic templates. In older versions `variant` has a fixed number of template arguments with the last few being placeholders. We'd need to add workarounds for this, but it's not worth it, so I've just noted the requirement in the docs.

The non-docker trusty builds on Travis only have access to Boost 1.54, which unfortunately means CI doesn't cover C++11 mode. Any ideas how to get a newer boost on Travis without docker?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,8 +28,12 @@ matrix:
 install:
 - ps: |
     if ($env:PLATFORM -eq "x64") { $env:CMAKE_ARCH = "x64" }
-    if ($env:APPVEYOR_JOB_NAME -like "*Visual Studio 2017*") { $env:CMAKE_GENERATOR = "Visual Studio 15 2017" }
-    else { $env:CMAKE_GENERATOR = "Visual Studio 14 2015" }
+    if ($env:APPVEYOR_JOB_NAME -like "*Visual Studio 2017*") {
+      $env:CMAKE_GENERATOR = "Visual Studio 15 2017"
+      $env:CMAKE_INCLUDE_PATH = "C:\Libraries\boost_1_64_0"
+    } else {
+      $env:CMAKE_GENERATOR = "Visual Studio 14 2015"
+    }
     if ($env:PYTHON) {
       if ($env:PLATFORM -eq "x64") { $env:PYTHON = "$env:PYTHON-x64" }
       $env:PATH = "C:\Python$env:PYTHON\;C:\Python$env:PYTHON\Scripts\;$env:PATH"
@@ -45,7 +49,7 @@ install:
 - ps: |
     Start-FileDownload 'http://bitbucket.org/eigen/eigen/get/3.3.3.zip'
     7z x 3.3.3.zip -y > $null
-    $env:CMAKE_INCLUDE_PATH = "eigen-eigen-67e894c6cd8f"
+    $env:CMAKE_INCLUDE_PATH = "eigen-eigen-67e894c6cd8f;$env:CMAKE_INCLUDE_PATH"
 build_script:
 - cmake -G "%CMAKE_GENERATOR%" -A "%CMAKE_ARCH%"
     -DPYBIND11_CPP_STANDARD=/std:c++%CPP%
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ install:
     $SCRIPT_RUN_PREFIX sh -c "for s in 0 15; do sleep \$s; \
       apt-get -qy --no-install-recommends install \
         $PY_DEBUG python$PYTHON-dev python$PY-pytest python$PY-scipy \
-        libeigen3-dev cmake make ${COMPILER_PACKAGES} && break; done"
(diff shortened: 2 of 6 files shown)


jagerman:
Try adding `boost-latest` to `addons->apt->sources`, and add `libboost-dev` to `addons->apt->packages`.

jagerman:
Scratch that; it looks like the `boost-latest` ppa isn't maintained (the latest version appears to be 1.55).

jagerman:
I suppose you could turn on `sudo: required` for one of the builds, then add a:

```
sudo apt-get -qy remove libboost1.54-dev &&
wget http://us.archive.ubuntu.com/ubuntu/pool/main/b/boost1.58/libboost1.58-dev_1.58.0+dfsg-5ubuntu3_amd64.deb && \
sudo dpkg -i libboost1.58-dev_1.58.0+dfsg-5ubuntu3_amd64.deb
```

If you want more recent than 1.58, the Debian `stretch` deb should also work (and like the `xenial` URL above should stay valid for a long time): `http://ftp.us.debian.org/debian/pool/main/b/boost1.62/libboost1.62-dev_1.62.0+dfsg-4_amd64.deb`.

dean0x7d:
It turns out that `boost::variant` can't use variadic templates on MSVC 2015 so the variant caster can't support it. I've added an additional note about MSVC 2017 being required.

Regarding the C++11 tests, I've added a `static_assert` to make sure that `::result_type` is present, which should be enough to cover that. Probably best not to complicate the Travis config for this. The main thing is that the variant caster is being tested with at least one third-party `variant` (with the C++14 docker builds).

PR #995, opened 2017-08-10
Visibility hidden

jagerman:
This is a few commits to force hidden visibility on the `pybind11` namespace and to unconditionally add it into the compilation flags (also this sets the flag through the cmake property rather than setting the actual compilation flags).

This needs one more thing before merging--the FAQ entry talking about hidden visibility seems largely obsolete with these changes.  Should we shorten it? Amend it? Drop it?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,7 @@ if(NOT (CMAKE_VERSION VERSION_LESS 3.0))  # CMake >= 3.0
                                                 $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
   target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:${PYBIND11_CPP_STANDARD}>)
   if(NOT MSVC)
-    target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:-fvisibility=hidden>)
+    target_compile_options(pybind11 INTERFACE -fvisibility=hidden)
   endif()
 
   add_library(module INTERFACE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,9 @@ if(NOT (CMAKE_VERSION VERSION_LESS 3.0))  # CMake >= 3.0
                                                 $<BUILD_INTERFACE:${PYTHON_INCLUDE_DIRS}>
                                                 $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
   target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:${PYBIND11_CPP_STANDARD}>)
+  if(NOT MSVC)
+    target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:-fvisibility=hidden>)
+  endif()
 
   add_library(module INTERFACE)
   add_library(pybind11::module ALIAS module)
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -137,10 +137,10 @@ Naturally, both methods will fail when there are cyclic dependencies.
 
 Note that pybind11 code compiled with hidden-by-default symbol visibility (e.g.
 via the command line flag ``-fvisibility=hidden`` on GCC/Clang), which is
-strongly recommended for pybind11 modules, can interfere with the ability to
+required proper pybind11 functionality, can interfere with the ability to
 access types defined in another extension module.  Working around this requires
-manually export types that are accessed by multiple extension modules; pybind11
-provides a macro to do just this:
+manually exporting types that are accessed by multiple extension modules;
+pybind11 provides a macro to do just this:
 
 .. code-block:: cpp
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -135,22 +135,16 @@ has been executed:
 
 Naturally, both methods will fail when there are cyclic dependencies.
 
-Note that compiling code which has its default symbol visibility set to
-*hidden* (e.g. via the command line flag ``-fvisibility=hidden`` on GCC/Clang) can interfere with the
-ability to access types defined in another extension module. Workarounds
-include changing the global symbol visibility (not recommended, because it will
-lead unnecessarily large binaries) or manually exporting types that are
-accessed by multiple extension modules:
+Note that pybind11 code compiled with hidden-by-default symbol visibility (e.g.
+via the command line flag ``-fvisibility=hidden`` on GCC/Clang), which is
+strongly recommended for pybind11 modules, can interfere with the ability to
+access types defined in another extension module.  Working around this requires
+manually export types that are accessed by multiple extension modules; pybind11
+provides a macro to do just this:
 
 .. code-block:: cpp
 
-    #ifdef _WIN32
-    #  define EXPORT_TYPE __declspec(dllexport)
-    #else
-    #  define EXPORT_TYPE __attribute__ ((visibility("default")))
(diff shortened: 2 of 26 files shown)


include/pybind11/attr.h
```diff
@@ -12,7 +12,7 @@
 
 #include "cast.h"
 
-NAMESPACE_BEGIN(pybind11)
+NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
```
dean0x7d: Wouldn't it be simpler to add the visibility attribute to the `NAMESPACE_BEGIN` macro instead of introducing a new one?

tools/pybind11Tools.cmake
```diff
@@ -139,6 +139,13 @@ function(pybind11_add_module target_name)
   set_target_properties(${target_name} PROPERTIES PREFIX "${PYTHON_MODULE_PREFIX}")
   set_target_properties(${target_name} PROPERTIES SUFFIX "${PYTHON_MODULE_EXTENSION}")
 
+  # -fvisibility=hidden is required to allow multiple modules compiled against
+  # different pybind versions to work properly, and for some features (e.g.
+  # py::module_local).  We force it on everything inside the `pybind11`
+  # namespace; also turning it on for a pybind module compilation here avoids
+  # potential warnings or issues from having mixed hidden/non-hidden types.
+  set_target_properties(${target_name} PROPERTIES CXX_VISIBILITY_PRESET "hidden")
```
dean0x7d: 👍

docs/advanced/misc.rst
```diff
@@ -135,22 +135,16 @@ has been executed:
...
 
-Note that compiling code which has its default symbol visibility set to
-*hidden* (e.g. via the command line flag ``-fvisibility=hidden`` on GCC/Clang) can interfere with the
-ability to access types defined in another extension module. Workarounds
-include changing the global symbol visibility (not recommended, because it will
-lead unnecessarily large binaries) or manually exporting types that are
-accessed by multiple extension modules:
+Note that pybind11 code compiled with hidden-by-default symbol visibility (e.g.
+via the command line flag ``-fvisibility=hidden`` on GCC/Clang), which is
+strongly recommended for pybind11 modules, can interfere with the ability to
```
dean0x7d: I guess it's required, not just recommended anymore. Maybe simplify to something along the lines of: "By default,  pybind11 modules are compiled with hidden symbol visibility (see :ref:`...`), which can interfere..."

docs/advanced/misc.rst
```diff
@@ -135,22 +135,16 @@ has been executed:
...
+access types defined in another extension module.  Working around this requires
+manually export types that are accessed by multiple extension modules; pybind11
```
dean0x7d: exporting

docs/compiling.rst
```diff
@@ -74,13 +74,15 @@ removes this target from the default build (see CMake docs for details).
...
-additional flags enable LTO (Link Time Optimization), set default visibility
-to *hidden* and strip unneeded symbols. See the :ref:`FAQ entry <faq:symhidden>`
-for a more detailed explanation. These optimizations are never applied in
-``Debug`` mode. If ``NO_EXTRAS`` is given, they will always be disabled, even
-in ``Release`` mode. However, this will result in code bloat and is generally
-not recommended.
+symbol names and duplication of code in different translation units. It
+sets default visibility to *hidden*, which is required for some pybind11
+features and functionality when attempting to load multiple pybind11 modules
+compiled under different pybind11 versions.  It also adds additional flags
```
dean0x7d: Weird question, but: Why two spaces after a period?

docs/compiling.rst
```diff
@@ -181,9 +183,8 @@ to an independently constructed (through ``add_library``, not
     flags (i.e. this is up to you).
 
     These include Link Time Optimization (``-flto`` on GCC/Clang/ICPC, ``/GL``
-    and ``/LTCG`` on Visual Studio). Default-hidden symbols on GCC/Clang/ICPC
-    (``-fvisibility=hidden``) and .OBJ files with many sections on Visual Studio
-    (``/bigobj``). The :ref:`FAQ <faq:symhidden>` contains an
+    and ``/LTCG`` on Visual Studio) and .OBJ files with many sections on Visual
+    Studio (``/bigobj``). The :ref:`FAQ <faq:symhidden>` contains an
```
dean0x7d: Unlike `pybind11_add_module()`, the CMake targets haven't been including any visibility flags. It would be good to point this out here.

Alternatively, the visibility property could be added to the targets, but that's somewhat of a breaking change.

include/pybind11/attr.h
wjakob: I'm against that, for selfish reasons :). I have several other projects that use the ``NAMESPACE_BEGIN`` macro, where that behavior may not be intended.

dean0x7d: Ah, I guess that's why `NAMESPACE_BEGIN/END` are the only macros not prefixed with `PYBIND11` :)

docs/compiling.rst
jagerman: In monospace fonts, it is visually closer to, e.g., a LaTeX inter-sentence space.  It's also consistent with what you get from tools like `man` when used in a terminal.  And while I do think it looks a bit better, it's mainly habit: I do it without even thinking about it.

docs/compiling.rst
jagerman: Oh, actually I meant to add that as well but forgot.  Why would it be breaking?

dean0x7d: If the library was relying on the default visibility to export symbols, the change would break cross-module types. It would require a one line change in CMake to restore default visibility, so it might not be a big deal.

jagerman: It seems it's not so easy to add it; there isn't any sort of `INTERFACE_CXX_VISIBILITY_PRESET`, which means we'd be back to figuring out when to add it based on compilers ourselves.  Better, I think, to just document that target users should set the property themselves.

dean0x7d: The condition should be as simple as `if(NOT MSVC)`, shouldn't it?

After some thinking, I feel that the hidden visibility should be enforced as much as possible. With an increasing number of pybind11 modules on PyPI, it would be quite easy to run into `internals` versioning issues and/or bindings of common STL containers.

jagerman: Okay, I'll add it in.

docs/compiling.rst
jagerman: Also (I just noticed) vim does this automatically when rewrapping (e.g. if a line ends with a `.`).  I'd be quite surprised if there isn't an option to control that, but it isn't something I turned on specifically.

CMakeLists.txt
```diff
@@ -91,6 +91,9 @@ if(NOT (CMAKE_VERSION VERSION_LESS 3.0))  # CMake >= 3.0
                                                 $<BUILD_INTERFACE:${PYTHON_INCLUDE_DIRS}>
                                                 $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
   target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:${PYBIND11_CPP_STANDARD}>)
+  if(NOT MSVC)
+    target_compile_options(pybind11 INTERFACE $<BUILD_INTERFACE:-fvisibility=hidden>)
```
dean0x7d: No need for build/install interface separation here because the same flag is applied to both.
If `target_compile_options(pybind11 INTERFACE -fvisibility=hidden)` is set here then nothing needs to be added to `pybind11Config.cmake`.

dean0x7d:
LGTM!

PR #997, opened 2017-08-13
Add test for mixing STL casters and local binders across modules

dean0x7d:
This is an attempt to address one of the lingering questions from #919. Even if it doesn't directly answer that, it's still a useful test to add. 

One module uses a generic vector caster from `<pybind11/stl.h>` while the other exports `std::vector<int>` with a local `py:bind_vector`. The vector caster from `<pybind11/stl.h>` accepts anything that looks like a sequence, so it will happily load the `py::module_local` opaque vector from a different module. Even if it can't see it's C++ type data, it still behaves like a sequence and can be converted generically.

Note: Type aliases and `PYBIND11_MAKE_OPAQUE` needed to be added for several types because of the inclusion of `<pybind11/stl.h>`.

### DIFF
--- a/tests/local_bindings.h
+++ b/tests/local_bindings.h
@@ -21,6 +21,14 @@ using MixedLocalGlobal = LocalBase<4>;
 /// Mixed: global first, then local (which fails)
 using MixedGlobalLocal = LocalBase<5>;
 
+using LocalVec = std::vector<LocalType>;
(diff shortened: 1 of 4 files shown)


jagerman:
Nice!  I didn't actually realize that would work, but it's nifty that it does.

PR #999, opened 2017-08-13
Avoid duplicate C++ standard flags if CMAKE_CXX_STANDARD is set

### LINKED ISSUE #998: PYBIND11_CPP_STANDARD detects and adds the /std:c++14 flag when CMAKE_CXX_STANDARD 17 resulting in both flags
Hi!

I'm using the latest VS 2017 Preview (15.3.0 Preview 7.1) with the latest CMake nightly (cmake-3.9.20170812-g71ee6).

CMake has recently added support for `set(CMAKE_CXX_STANDARD 17)` on MSVC, which will set `/std:c++17` (in fact it sets `-std:c++17` which still works but I guess they should use the slash instead so that the VS IDE displays it correctly) (https://gitlab.kitware.com/cmake/cmake/issues/16482).

Now pybind11 detects `PYBIND11_CPP_STANDARD` to be `/std:c++14`, and adds the `/std:c++14` flag in addition to `/std:c++17` which is added by CMake, resulting in a (benign) compiler warning:

> cl : Command line warning D9025: overriding '/std:c++14' with '/std:c++17'

I realise I'm a bit cutting-edge here but both of these things will be released officially soon so we may well improve them now or at least have it on the radar. 
I think the logic here (https://github.com/pybind/pybind11/blob/master/tools/pybind11Tools.cmake#L35) will need some adjustments. I'm happy to try to change it if you give me some guidance, or test a fix.

Cheers!

dean0x7d:
Fixes #998.

`CMAKE_CXX_STANDARD` is only available on CMake >= 3.1. If the flag is set, we avoid initializing `PYBIND11_CPP_STANDARD`.

I didn't add any tests because all the CI builds currently rely on `PYBIND11_CPP_STANDARD` (for older CMake support) and there's no need to complicate things there.

### DIFF
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -18,7 +18,7 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
 include(CheckCXXCompilerFlag)
 include(CMakeParseArguments)
 
-if(NOT PYBIND11_CPP_STANDARD)
+if(NOT PYBIND11_CPP_STANDARD AND NOT CMAKE_CXX_STANDARD)
   if(NOT MSVC)
     check_cxx_compiler_flag("-std=c++14" HAS_CPP14_FLAG)
 


jagerman:
This is going to take care of the global, but what if the `CXX_STANDARD` property is set on a target?

dean0x7d:
> This is going to take care of the global, but what if the CXX_STANDARD property is set on a target?

We can't do anything about it as far as I know. It would be set by users after `pybind11_add_module()` has finished or the `pybind11::` targets have been constructed. 

We could conditionally, on CMake >= 3.1, set the per-target `CXX_STANDARD` and let CMake sort it out. But this starts to get really messy if we need to support both the old and new standard styles. I'm also not completely happy with the `CXX_STANDARD` implementation because it requires CMake to actually know about the target compiler and C++ standard. For example, older CMake versions may not know about c++17 or c++1z flags and refuse to set them or even the basic 11/14 flags on some not so common compilers (I've had issues with ICC on early CMake 3.x versions). Also, the way the `CXX_STANDARD_REQUIRED` property works (on/off) isn't really ideal (it would be nice if it was the minimum *required* standard, while `CXX_STANDARD` was the *preferred if available* standard).

All in all, I think it would be best to stick with our existing standard variable and just make this provision for the global `CMAKE_CXX_STANDARD` without too much messing around with the properties.

Issue #953, opened 2017-07-21
Make a 2.2 release


jagerman:
We have a decent amount of new features and bug fixes for a new stable v2.2 release.  Some big features (embedding support seems the biggest), lots of minor ones, and various bug fixes have landed into `master` over the last four months.  I'm not suggesting we release it tomorrow, but I think we should start thinking about where to draw the line for 2.2.

PRs I think we should aim to merge:

- [x] #951, #952 are bug fixes (and ready to merge).
- [x] #805 (init factories) - the implementation is done; we need to decide whether to make this core (i.e. included by `pybind11.h` or not), and then I'll update the documentation accordingly before merging.
- [x] #949 (`py::local` support for keeping a type's registration local to the module).  This was a sort of experiment to think about fixing #919, but I think it worked out into a nice feature -- and I was actually surprised just how little code it actually needed to implement.  This still needs me to write up the documentation, but I'll do that in the next couple of days.
- [x] #928 (FindPythonLibs{,New} cmake compatibility) - this is ready to merge AFAICT.
- [x] #925 (bool and numpy bool conversions)
- [x] #907 would be nice to have it for the clearer command line instructions.
- ~~#901 (cutting off the maximum length in generated signatures) is a nice cosmetic feature.  It needs some more tweaking, but if that happens it seems worth including.~~ --> for v2.3
- ~~#934 (constexpr type signatures for C++11 and MSVC).  Currently has a minor merge conflict, but aside from that I think it's ready to merge.~~ --> for v2.3

Other features in the queue that I think might not make it:

- #864 (Revising type casters) - I think this is going to be a pretty significant change, and likely to go through a fair amount of discussion/suggestions/changes from start to finish.  I'm not opposed to including it if its done soon, but of course I also think it's important to get it right.
- #781 (python 3.4 enums) - this is still marked WIP and, I guess, still in progress.  I think the last discussion was that it was going to be separate from `enum_<>` due to backwards-compatibility issues with how `enum_` worked; beyond that I'm not entirely sure of the status.

Thoughts?

dean0x7d:
Yeah, seems like a good time for a new release. There's a good number of bug fixes and some nice new features.

I'd suggest postponing #934 for the next release, so it gets some time in master just in case there are any backward compatibility concerns. Also, it's related to #864 so if we find any need to adjust the casters it would be good to keep everything in one release and change things only once.

For the rest of the PRs, I very much agree with your assessment.

garth-wells:
[...]

PR #1000, opened 2017-08-13
Proposed changelog for v2.2.0

dean0x7d:
Here is a proposed changelog for v2.2. I've tried to keep it in the same style as the previous entries. There's a pretty large number of changes, so please let me know if I've missed anything, if something needs to be reworded, or if the order of the items should change. A lot of PRs ended up in the "various improvements" item, but perhaps some of those need a full entry (or vice versa).

Unless I've missed any, the relevant PRs should all be in the [v2.2 milestone](https://github.com/pybind/pybind11/milestone/3) -- there's around a 100 PRs. I didn't check any of the out-of-PR commits.

### DIFF
--- a/docs/advanced/cast/stl.rst
+++ b/docs/advanced/cast/stl.rst
@@ -26,6 +26,8 @@ next sections for more details and alternative approaches that avoid this.
     The file :file:`tests/test_stl.cpp` contains a complete
     example that demonstrates how to pass STL data types in more detail.
 
+.. _cpp17_container_casters:
+
 C++17 library containers
 ========================
 
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -406,6 +406,8 @@ name, i.e. by specifying ``py::arg().noconvert()``.
     need to specify a ``py::arg()`` annotation for each argument with the
     no-convert argument modified to ``py::arg().noconvert()``.
 
+.. _none_arguments:
+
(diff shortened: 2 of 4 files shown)


jagerman:
I think the hidden visibility change should be mentioned fairly prominently as it may potentially affect anyone not using pybind's cmake tools.

Also a suggestion for post-2.2: lets start putting changelog entries into the changelog as we merge PRs rather than waiting to the end so that all we need to do at release time is some light editing and reordering rather than writing them all up (and potentially missing things).

dean0x7d:
> I think the hidden visibility change should be mentioned fairly prominently as it may potentially affect anyone not using pybind's cmake tools.

Good point. I'll add it.

> Also a suggestion for post-2.2: lets start putting changelog entries into the changelog as we merge PRs

Agreed.

jagerman:
Another release-related issue: we should also update the example repositories for 2.2 (to use the `PYBIND11_MODULE` macro, but possible other things).

wjakob:
This looks like an epic release -- sorry for not helping out more (I've been busy with other projects). In any case, I'm happy to push the button to send out the release, just let me know.

dean0x7d:
I think the release is shaping up nicely. Maybe just one or two things to wrap up. I'll merge this changelog now so that it's accessible to PRs. Issue #953 is still open for release v2.2 in general.

Issue #864, opened 2017-05-18
Revising the design of type_caster<T>::load()


dean0x7d:
Currently, type casters defined using `PYBIND11_TYPE_CASTER` require that the C++ type is default-constructible and copy/move-assignable. Apart from the fact that not all types fulfill these requirements, doing default init + assign is less efficient than directly constructing an object (both binary size and runtime overhead). There are also possible complications as in the case of #850 where an additional helper struct and specialization are required but could be avoided with a different `load()` design.

### The proposal

The idea is to make `type_caster<T>::load()` a static function which returns an optional-like container. Looking only at the load-specific parts, a type caster currently looks like this:

```c++
struct caster {
    T value;

    operator type*();
    operator type&();
    template <typename U> using cast_op_type;

    bool load(handle src, bool convert);
}
```

This proposal changes it to:

```c++
struct caster {
    static maybe<T> load(handle src, bool convert);
}
```

Where `maybe<T, ExtractionPolicy>` is an optional-like container with some simplifications and specializations for the purpose of pybind11's casters (named `maybe` to avoid confusion with the standard `optional`). Here, the `ExtractionPolicy` template parameter takes on the job of `cast_op_type` and the cast operators. The default policy is the same as defined by `PYBIND11_TYPE_CASTER` which makes `maybe<T>` the only thing needed for user-defined casters. This also drastically reduces the footprint of the `PYBIND11_TYPE_CASTER` macro.

This would also simplify usage:
```c++
// currently
U value;

bool load(handle src, bool convert) {
    make_caster<T> subcaster;
    if (subcaster.load(src, convert)) {
        value = cast_op<T>(subcaster);
        return true;
    }
    return false;
}
```

```c++
// proposed
static maybe<U> load(handle src, bool convert) {
    if (auto maybe_value = make_caster<T>::load(src, convert))
        return maybe_value.extract<T>();
    return {};
}
```

The behavior of `maybe::extract<T>()` is determined by the `ExtractionPolicy`. Given that these containers have a very special purpose and are only used once, the default policy could move when `T` is a value or rvalue ref (to simplify caster code).

### Backward compatibility

Existing user-defined casters would still function either by inserting a compatibility layer into `PYBIND11_TYPE_CASTER` or by using SFINAE in pybind11 internals (only if compatibility is required for user-defined casters which do not use the `PYBIND11_TYPE_CASTER` macro).

[...]

Issue #708, opened 2017-03-01
Header reorganization


jagerman:
#679 initially added a `detail/` directory to the header includes.  It was withdrawn to be discussed later in a separate issue/PR; this is meant to be the start of discussion.  Also read the following not as a done deal, but as a request for comments where I'm mostly thinking out loud to start the discussion.

I think we should make the move for most (or even all?) of our `detail` namespace code.  Basically, all the things in `class_support.h` are an obvious first start to move to `detail/class.h`; but there are also many other potential additions:

Detail headers
===
- the various template meta-code in common.h (for things like satisfied_all_of, would fit nicely into a `detail/meta.h`.  It's relatively centralized now (in common.h), but that hasn't always been the case and has led to me, at least, occassionally duplicating meta template functionality that was already implemented (but implemented elsewhere, i.e. immediately before it was used/needed).

- The same sort of moving of `detail` namespace bits into `detail`-namespace companions would be a nice cleanup to the non-detail code.  E.g. pybind11.h's `generic_type` and `init` classes, keep alive implementation, etc. are all already in the `detail` namespace; moving them into a detail directory header would be appropriate.

Type casters
===

Along a similar line, we could move type caster implementations into their own namespace, `pybind11::cast`, and directory, `pybind11/cast`.  For now, for backwards compatibility, `cast` would be a namespace alias of `detail`, but the plan would be to eventually drop the alias so that all type casters live as `pybind11::cast::type_caster<T>`, thus having external type casters put themselves in the `detail` namespace.

Thus you would include `pybind11/cast/stl.h` for stl casters, `pybdin11/cast/eigen.h` for eigen casters, etc.  (The current `pybind11/stl.h`, `pybind11/eigen.h` would just one-line compatibility headers that just include the new location).

Many of the built-in casters would be usefully moved as well:
- the arithmetic caster in `cast/arithmetic.h`
- std::string and char* casters to `cast/string.h`
- std::pair/std::tuple into `cast/tuple.h`
- pointer casts, null casters, generic casters I think could remain in `cast.h`.
- the argument loader seems a bit out of place in `cast.h`—I think it's there mainly for historic reasons, since it was originally part of the `std::tuple` caster.  Perhaps `detail/argloader.h`?

dean0x7d:
[...]

PR #1001, opened 2017-08-13
Move internal headers into `detail` subdirectory

dean0x7d:
The header reorganization discussed in #708 can be broken up into 3 main parts:

1. Moving internal headers into a `detail` subdirectory.
2. Breaking up large detail-heavy headers into smaller ones (e.g. `detail/meta.h` for the metaprogramming bits).
3. Separate `cast` subdirectory for type casters.

Instead of tackling all this at the same time, it would be best to take it in different stages. This PR takes care of number 1 by simply moving the non-user-facing headers (everything fully intact and no user-side changes). This would allow new features like #805 to already make use of the `detail` subdirectory.

Number 2 can be done separately in smaller pieces. Number 3 would have some user-facing changes and it would be best to do in conjunction with the caster changes from #864.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,14 +36,15 @@ set(PYTHON_MODULE_EXTENSION ${PYTHON_MODULE_EXTENSION} CACHE INTERNAL "")
 
 # NB: when adding a header don't forget to also add it to setup.py
 set(PYBIND11_HEADERS
+  include/pybind11/detail/class.h
+  include/pybind11/detail/common.h
+  include/pybind11/detail/descr.h
+  include/pybind11/detail/typeid.h
   include/pybind11/attr.h
   include/pybind11/buffer_info.h
   include/pybind11/cast.h
   include/pybind11/chrono.h
-  include/pybind11/class_support.h
-  include/pybind11/common.h
   include/pybind11/complex.h
-  include/pybind11/descr.h
   include/pybind11/options.h
   include/pybind11/eigen.h
   include/pybind11/embed.h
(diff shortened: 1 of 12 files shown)


jagerman:
Are you proposing this first step for 2.2?  I can update #805 to match.

dean0x7d:
> Are you proposing this first step for 2.2? I can update #805 to match.

Yes. Forgot to mention that. Step 1 for v2.2 and steps 2/3 for v2.3.

jagerman:
Sounds good to me.

CMakeLists.txt
```diff
@@ -55,7 +56,6 @@ set(PYBIND11_HEADERS
   include/pybind11/pytypes.h
   include/pybind11/stl.h
   include/pybind11/stl_bind.h
-  include/pybind11/typeid.h
```
jagerman: `typeid.h` does provide the non-`detail` namespace `py::type_id<T>()`, which is a fairly useful function and *might* be in use.  Maybe this one should be left for later?

include/pybind11/pytypes.h
```diff
@@ -1,5 +1,5 @@
 /*
-    pybind11/typeid.h: Convenience wrapper classes for basic Python types
+    pybind11/pytypes.h: Convenience wrapper classes for basic Python types
```
jagerman: Good catch 👍

CMakeLists.txt
dean0x7d: My thinking here was that `typeid.h` is included anyway by `cast.h` and in turn by `pybind11.h`, so there's no way to lose out on `py::type_id<T>()` unless someone is using it without anything else from pybind11. But I wouldn't mind reverting it if I'm overlooking something.

jagerman: That's true; it's fine with me to include it in the move.  (Maybe `type_id()` should be moved out into `cast.h` or somewhere else, but that's not really a big deal—and as an undocumented mostly internal function it was arguably misplaced by not being in `detail` in the first place).

PR #1003, opened 2017-08-14
Don't force hidden visibility on the embed target, just the module target

jagerman:
Embedding may well be used in places where hidden visibility isn't
desired.  It should be relatively safe to allow it there; any potential
conflict would come in if modules are loaded into that embedded
interpreter, but as long as the modules are compiled with hidden
visibility they shouldn't conflict.

There could still be warnings if the embedded code attempts to export
classes with internal (hidden) pybind members, but that seems a
legitimate warning (and already has a FAQ entry).


dean0x7d:
LGTM

PR #1007, opened 2017-08-17
Allow module-local classes to be loaded by external modules

### LINKED ISSUE #919: Eventual collisions between opaque STL containers

jagerman:
The main point of `py::module_local` is to make the C++ -> Python cast unique so that returning/casting a C++ instance is well-defined.  Unfortunately it also makes *loading* unique, but this isn't particularly desirable or necessary: when an instance contains `Type` instance there's no reason it shouldn't be possible to pass that instance to a bound function taking a `Type` parameter, even if that function is in another module.

This commit solves the issue by storing the module-local `&type_caster_generic::load` in a global multimap so that we can call into local modules' `load()` methods for `module_local` types.  For module-local types that override a globally-registered types a global type conversion is also attempted.

Fixes #919.

It increases `.so` size slightly (a little over 4KB, not counting the added tests), but as the additions are entirely in non-templated code, that number shouldn't change with the number of bound types.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -51,6 +51,7 @@ struct type_info {
     std::vector<bool (*)(PyObject *, void *&)> *direct_conversions;
     buffer_info *(*get_buffer)(PyObject *, void *) = nullptr;
     void *get_buffer_data = nullptr;
+    void *(*module_local_load)(PyObject *, const type_info *) = nullptr;
     /* A simple type never occurs as a (direct or indirect) parent
      * of a class that makes use of multiple inheritance */
     bool simple_type : 1;
@@ -265,23 +266,30 @@ PYBIND11_NOINLINE inline detail::type_info* get_type_info(PyTypeObject *type) {
     return bases.front();
 }
 
-/// Return the type info for a given C++ type; on lookup failure can either throw or return nullptr.
-/// `check_global_types` can be specified as `false` to only check types registered locally to the
-/// current module.
-PYBIND11_NOINLINE inline detail::type_info *get_type_info(const std::type_index &tp,
-                                                          bool throw_if_missing = false,
-                                                          bool check_global_types = true) {
-    std::type_index type_idx(tp);
+inline detail::type_info *get_local_type_info(const std::type_index &tp) {
     auto &locals = registered_local_types_cpp();
-    auto it = locals.find(type_idx);
+    auto it = locals.find(tp);
     if (it != locals.end())
         return (detail::type_info *) it->second;
-    if (check_global_types) {
-        auto &types = get_internals().registered_types_cpp;
-        it = types.find(type_idx);
-        if (it != types.end())
-            return (detail::type_info *) it->second;
-    }
+    return nullptr;
+}
+
+inline detail::type_info *get_global_type_info(const std::type_index &tp) {
+    auto &types = get_internals().registered_types_cpp;
+    auto it = types.find(tp);
+    if (it != types.end())
+        return (detail::type_info *) it->second;
+    return nullptr;
+}
+
+/// Return the type info for a given C++ type; on lookup failure can either throw or return nullptr.
+PYBIND11_NOINLINE inline detail::type_info *get_type_info(const std::type_index &tp,
+                                                          bool throw_if_missing = false) {
+    if (auto ltype = get_local_type_info(tp))
+        return ltype;
+    if (auto gtype = get_global_type_info(tp))
+        return gtype;
+
     if (throw_if_missing) {
         std::string tname = tp.name();
         detail::clean_type_id(tname);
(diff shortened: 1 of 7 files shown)


dean0x7d:
Just a quick comment for now. The search for an external loader looks really complicated. Instead of the multimap in `internals`, have you considered attaching the external load as a capsule to the Python type object in `generic_type::initialize()`?

jagerman:
That simplifies it; I've pushed a couple of commits to stash things in an attribute of the python type.  The second commit is simpler, but increases the `type_info` size by the size of a member function pointer.

dean0x7d:
I had a go at simplifying this just a bit more: reduced the reliance on caster state (recursion flag, typeinfo state change) because these will need to go anyway with the planned new-style `maybe` loaders. Also switched to a regular function pointer for portability (pointer to member function can be weird). The changes reduce binary size (at least on clang). Everything should still work as before. Let me know if you agree with these changes.

jagerman:
Looks great!

jagerman:
I think we should include this in 2.2 -- I noticed you also put it in the upgrade doc, so I'm guessing you agree.

dean0x7d:
Yes, definitely. It makes `module_local` significantly more convenient (and easier to explain because there's one less caveat). If everything else looks good, feel free to squash and merge this.

jagerman:
The current build is just for the comment change, right?  (i.e. safe to not wait for the ci tests to finish?)

dean0x7d:
Yeah.

include/pybind11/cast.h
```diff
@@ -686,13 +694,32 @@ class type_caster_generic {
...
+    /// native typeinfo, or when the native one wasn't able to produce a value.
+    PYBIND11_NOINLINE bool try_load_foreign_module_local(handle src) {
+        constexpr auto *local_key = "_pybind11_module_local_typeinfo";
+        const auto pytype = src.get_type();
+        if (!hasattr(pytype, local_key))
+            return false;
+
+        type_info *foreign_typeinfo = reinterpret_borrow<capsule>(getattr(pytype, local_key));
+        if (foreign_typeinfo == typeinfo)
+            return false; // Only consider this foreign loader if actually foreign
```
jagerman: I think there's a slight problem here: if the type has a masked global registration, `typeinfo` will have been changed to the global one, in which case the `foreign_typeinfo == typeinfo` isn't going to be true, so we could end recursing.

jagerman:
There's another problem here that I'm looking into: We aren't checking that the foreign loader is actually a loader of the right type.

include/pybind11/cast.h
```diff
@@ -655,8 +665,6 @@ class type_caster_generic {
         return inst.release();
     }
 
-protected:
```
jagerman: Was this removal intentional?

dean0x7d: Yes. Direct access to `value` is needed in the `module_local_load` lambda. And seeing as `detail::type_caster_generic` is already an implementation detail, there isn't really much of a need to protect access.

dean0x7d:
> There's another problem here that I'm looking into: We aren't checking that the foreign loader is actually a loader of the right type.

Good catch.

include/pybind11/cast.h
dean0x7d: Do you mean a local binding masking a local one? That should work and is covered by the tests. (Not the `cpptype` mismatch mentioned below.)

jagerman: I'm thinking:
- the local type masks the global type, is the correct type_info for the given argument, but the load fails anyway.
- the global type gets checked, which also fails (the global type is not right).
- foreign loaders get checked, but because `typeinfo` has been changed to the global one, we don't short-circuit in `try_load_foreign` and so end up recursing.

It's not immediately obvious to me that the first step is impossible; though it's not immediately obvious to me how I can trigger it in a test, either.

jagerman: I think 225caf3 ought to solve this: when using a proper function (rather than a lambda) we can just test whether that function is foreign rather than relying on `typeinfo`.

dean0x7d: I see, seems reasonable.

PR #1009, opened 2017-08-18
Stdout redirection

### LINKED ISSUE #1005: Redirect std::cout
I'm interested in redirecting `std::cout` for my functions (some of which are long running library functions that report on progress). Is there an officially sanctioned way to redirect `std::cout` to python's stdout? (other than the custom `pybind11::print` function)? Running inside of a Jupyter notebook, for example, requires the output go to Python's stdout to be visible. 

After lots of googling and some playing around, I've come up with the following solution, which works well. I can make into a PR if this is useful and not already available. It's at https://github.com/GooFit/GooFit/blob/573044ea20eedc8444f38cc6a7ce66bef27d2524/python/Print.cpp .

henryiii:
This is the `std::cout` redirection discussed in #1005. The new docs:

## Capturing standard output from ostream

Often, a library will use the streams `std::cout` and `std::cerr` to
print, but this does not play well with Python's standard `sys.stdout`
and `sys.stderr` redirection. Replacing a library's printing with
`py::print` may not be feasible. This can be fixed using a guard
around the library function that redirects output to the corresponding
python streams:

```cpp
// At beginning of file
#include <pybind11/iostream.h>

...

// Add a scoped redirect for your noisy code
m.def("noisy_func", []() {
    py::scoped_ostream_redirect stream(
        std::cout,                               // std::ostream&
        py::module::import("sys").attr("stdout") // Python output
    );
    call_noisy_func();
});
```

This method respects flushes on the output streams, and will flush if
needed when the scoped guard is destroyed. This allows the output to be
redirected in real time, such as to a Jupyter notebook. The two
arguments, the C++ stream and the Python output, are optional, and
default to standard output if not given. An extra type,
`py::scoped_estream_redirect`, is identical execpt for defaulting to the
error stream; this can be useful with `py::call_guard`, which allows
multiple items, but uses the default constuctor:

```py
// Alternative: Call single function using call guard
m.def("noisy_func", &call_noisy_function,
      py::call_guard<py::scoped_ostream_redirect,
                     py::scoped_estream_redirect>());
```

The redirection can also be done in python with the addition of a
context manager, using the `py::add_ostream_redirect()` function:

```cpp
py::add_ostream_redirect(m, "ostream_redirect");
```

The name in Python defaults to `ostream_redirect` if no name is passed.
This creates the following context manager in Python:

```py
with ostream_redirect(stdout=True, stderr=True):
    noisy_function()
```

The added context manager defaults to redirecting both streams, though
you can use the keyword arguments to disable one of the streams if
needed.

Closes #1005.


henryiii:
Odd, you don't seem to be able to `del` a `py::container` in Pypy, it calls the destructor twice (segfault).

aldanor:
My subjective 2c: this is somewhat half-baked since it doesn't work with C-level output. 

In fact, this is possible to implement completely outside of `pybind11` -- e.g. I've done [something similar](https://github.com/aldanor/ipybind/blob/master/ipybind/stream.py) in `ipybind` that works via `os.dup2` and redirects all kinds of streams.

henryiii:
As long as you are working with C++, this works, and it does the redirection directly in the C++ code. Since libraries usually don't change IO, this either works if `cout`/`cerr` is used, or it doesn't if `printf` and raw file descriptors are used for a given library. This is much simpler and completely portable, and can be added in C++ instead of wrapping in Python.

This could be extended eventually to provide a FD version; it would need a series of platform checks, but that code (even on Python) has a series of checks as well. Like the Python version, it would not be fully standard, relying on non-standard [extensions like `__stdoutp`](https://github.com/aldanor/ipybind/blob/143bda6d85ad8aab47b6ec0216eee1813a999220/ipybind/ext/wurlitzer.py#L38-L39). Libraries that do this, like [PyROOT](https://root.cern.ch), tend to behave weirdly.

henryiii:
Just my 2c as well, and I'm biased. 😉

tests/test_globaliostream.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+
+#include <pybind11/iostream.h>
+#include "pybind11_tests.h"
+#include <iostream>
+
+TEST_SUBMODULE(globaliostream, m) {
+    // test_evals
+    m.def("redirect_output", []() {
+            return py::capsule(new py::scoped_output_redirect(
```
dean0x7d: I see that this is going for an RAII-style guard, but those should really be implemented using context managers in Python. Something like the following (not actually tested):

```c++
struct RedirectOutput {
    std::unique_ptr<py:: scoped_output_redirect> redirect;
}

py::class_<RedirectOutput>(m, "redirect_output")
    .def("__enter__", [](RedirectOutput &self) {
        self.redirect.reset(new py::scoped_output_redirect(...))
    })
    .def("__exit__", [](RedirectOutput &self, py::args) {
        self.redirect.reset();
    });
```

Note: it's convention for Python context managers to have lowercase names.

```python
with m.redirect_output():
    # redirected here
    with capture:
        # can be nested
```

henryiii:
Fixed the easier comments, will work on the rest of it tomorrow. Thanks for the review!

I could put it in a detail folder then include in `stl.h`.

henryiii:
Remaining tasks:

- [ ] Try removing multiple inheritance
- [ ] Add note about C output to docs
- [ ] Update docs for new redirect
- [ ] Try idea about removing non-captured output
- [ ] Move the code into `stl.h` (last, since it's easier to develop by itself)

jagerman:
> I'm not sure if this should be a new header or just added to pybind11/stl.h. Does anyone else have any opinions on this?

I vote, at least weakly, in favour of a new header.  `stl.h`'s current self-professed (i.e. in the file header) purpose is: `Transparent conversion for STL data types`.  This feature, on the other hand, doesn't really fit with that.  While it lets you manipulate the stl output interface, it isn't exactly "conversion" nor is it transparent.  It seems different enough to me to warrant its own header.

dean0x7d:
> stl.h's current self-professed (i.e. in the file header) purpose is: Transparent conversion for STL data types

Ah, indeed. And with the proposed header reorg from #708 it would become `pybind11/cast/stl.h` -- the `cast` part definitely doesn't fit here. Then it would be best to keep it `iostream.h` as is.

include/pybind11/iostream.h
```diff
@@ -0,0 +1,96 @@
...
+        return 0;
+    }
+public:
+    pythonbuf(object pyostream)
+        : pyostream(pyostream) {
+        this->setp(this->d_buffer, this->d_buffer + sizeof(this->d_buffer) - 1);
+    }
+};
+
+class opythonstream : private virtual pythonbuf, public std::ostream {
```
henryiii: Not trivially, since it will be constructed in the wrong order, as the final object is required to construct the class it would be a member of. Let me know if you have any ideas other than manually building objects first and then passing around smart pointers or similar.

henryiii:
That's most of the changes, plus a bit of renaming/redesign based on the comments. The remaining questions: multiple inheritance doesn't seem clean to avoid, due to construction order. The class is otherwise very simple, doing almost nothing past the multiple inheritance part.

The second question is if the test files should be combined. I like keeping the python files separate, but I can combine the C++ files (I have a local change stashed that I can push up to do that if that's still agreed upon). The code in the files is mostly orthogonal.

tests/test_globaliostream.cpp
henryiii: Redesigned, and made available in the iostream.h header. I've tried to tweak the names a bit too to be clearer.

jagerman:
> multiple inheritance doesn't seem clean to avoid, due to construction order.

Would something like the following work?

```C++
class opythonstream : public std::ostream {
public:
    opythonstream(object pyostream) : std::ostream(new pythonbuf(pyostream)) { }
    virtual ~opythonstream() { delete rdbuf(nullptr); }
};
```

include/pybind11/iostream.h
```diff
@@ -0,0 +1,188 @@
+#pragma once
```
dean0x7d: Add a license notice like in `test_iostream.cpp` and the other headers.

include/pybind11/iostream.h
```diff
@@ -0,0 +1,188 @@
...
+    std::streambuf * old {nullptr};
+    std::ostream& costream;
+    detail::opythonstream buffer;
+
+public:
+    scoped_ostream_redirect(
+            std::ostream& costream = std::cout,
+            object pyostream = module::import("sys").attr("stdout") )
+        : costream(costream), buffer(pyostream) {
+        old = costream.rdbuf(buffer.rdbuf());
```
dean0x7d: Seem like the `opythonstream` class isn't even needed. The buffer could just be:
```c++
detail::pythonbuf buffer;
```
And then:
```c++
old = costream.rdbuf(&buffer);
```

include/pybind11/iostream.h
henryiii: Copied from there.

include/pybind11/iostream.h
henryiii: That makes it simpler and cleaner! Done.

dean0x7d:
Oh, cool, I just realized that this can be combined with call guard:
```c++
m.def("noisy_func", &noisy_func, py::call_guard<py::scoped_ostream_redirect>());
```

henryiii:
You are right, I'm adding that to the tests and docs 👍

henryiii:
Added tests to verify flushing worked, and auto-flush if pythonbuf is destroyed (a problem revealed by the added tests). I dropped a bit of unneeded verbosity, as well.

henryiii:
I think this is ready, finished polishing it off.

Primarily due to the usefulness of `call_guard`, a `scoped_estream_redirect` has been added with `cerr` defaults. This allows both `cout` and `cerr` to be redirected easily:

```cpp
m.def("noisy_function", &noisy_function,
      py::call_guard<py::scoped_ostream_redirect,
                     py::scoped_estream_redirect>());
```

This also simplified some other code without losing generality. Other changes are mostly cosmetic; Docs have been nicely formatted and expanded a bit, etc.

jagerman:
> This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

Can you squash the commits down to one or two commits with appropriate merged/updated commit messages, and rebase against current master (rather than merging the current master branch)?  (That will keep the `master` commit history much cleaner when this gets merged).

henryiii:
Rebased, squashed down to 5 commits that each had a valid, working version.

henryiii:
Finally the Mac build finished.

henryiii:
Missed an inline (`py::class_` is a function), re-pushed.

dean0x7d:
@henryiii Our sphinx setup is a bit peculiar so I went over and fixed up some reference links and added a section/links to the API docs. Minor (whitespace) style changes in the header.

Only significant bits:

* `add_ostream_redirect` now adds the class as `module_local` to avoid module conflicts.
* Dropped `scoped_estream_redirect` constructors in favor of inherited ones.

Let me know if anything looks off here. If not, I think this PR can be merged.

dean0x7d:
There was also a missing header in `setup.py`. I rebased the PR to have the new installation tests cover it. The commit order is fine offline, but for some reason, GitHub decided to randomize them...

henryiii:
The `module_local` was new in 2.2, so I didn't know about it. I checked the `scoped_estream_redirect` constructors, the new version seems to work correctly. (I was overly paranoid, but `delete`d constructors and assignments stay deleted).

Looks good, thanks!

dean0x7d:
Merged! Thanks, @henryiii!

> The module_local was new in 2.2, so I didn't know about it.

It was added very recently, so I also didn't remember to add it initially, but it looks like a good fit for `add_ostream_redirect`.

henryiii:
Great, thanks for the help polishing! Looking forward quite a few of the features in 2.2!

Issue #909, opened 2017-06-20
Upgrading to 2 leads to: Unknown/invalid class_ template parameters provided


abergmeier-dsfishlabs:
I am trying to upgrade to 2.x but new template magic leads to incomprehensible template error stacks: `Unknown/invalid class_ template parameters provided`.
 This seems to have been introduced in 5fffe200e337f1a1bb13034fac29a63ad6eae304.

I have 2 issues with this:

1. I am puzzled whether this really is the user experience that pybind11 wants to provide to users. The error message conveys near zero information. I am wondering whether this code has been reviewed?

2. Also the new `class_` API is not well documented.
  There seems to be no description what the upgrade path from the old API is or what traits the new API awaits.

As a result, can someone explain what the upgrade path for the following code from:

    void bindMyResource(py::module& m) {
        py::class_<MyResource, IntrusivePtr<MyResource>>(m, "Model", py::base<Resource>);
    }

to:

    void bindMyResource(py::module& m) {
        py::class_<MyResource, IntrusivePtr<MyResource>>(m, "Model", /*probably some type that gets passed in here*/);
    }

or:

    void bindMyResource(py::module& m) {
        py::class_<MyResource, Resource, IntrusivePtr<MyResource>>(m, "Model");
    }

is? Thanks.

jagerman:
Both approaches are valid.  The rational for allowing the template arguments to be specified in any order is that it allows other things (such as the base class) to be specified as a template argument, and makes the template arguments position-independent.

It seems more likely that are are [missing a `PYBIND11_DECLARE_HOLDER_TYPE`](http://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#custom-smart-pointers) for your holder type; in 2.0 that is now detected as a compilation error; before 2.0 it wasn't caught but could fail in unpredictable ways.

jagerman:
Closing this due to lack of response.  (Feel free to reopen if this is still an issue.)

PR #1010, opened 2017-08-18
Add upgrade guide to the documentation

dean0x7d:
The basic idea here is to have a separate documentation page which sits right below the changelog, but goes into more detail about changes which specifically affect version upgrades. It can spell out what to watch out for and what the optimal upgrade path is. This is only a small number of items per version, but ones that would feel weird to describe in detail in the changelog. This would hopefully avoid some confusion going forward (e.g. #909).

The sections about hidden visibility and local STL containers feel a bit verbose. Let me know your thoughts about those and the whole thing.

### DIFF
--- a/docs/upgrade.rst
+++ b/docs/upgrade.rst
@@ -1,0 +1,329 @@
+Upgrade guide
+#############
+
+This is a companion guide to the :doc:`changelog`. While the changelog briefly
+lists all of the new features, improvements and bug fixes, this upgrade guide
+focuses only the subset which directly impacts your experience when upgrading
+to a new version. But it goes into more detail. This includes things like
+deprecated APIs and their replacements, build system changes, general code
+modernization and other useful information.
+
+
+v2.2
+====
+
+Deprecation of the ``PYBIND11_PLUGIN`` macro
+--------------------------------------------
+
+``PYBIND11_MODULE`` is now the preferred way to create module entry points.
+The old macro emits a compile-time deprecation warning.
+
+.. code-block:: cpp
+
+    // old
+    PYBIND11_PLUGIN(example) {
+        py::module m("example", "documentation string");
+
+        m.def("add", [](int a, int b) { return a + b; });
+
+        return m.ptr();
+    }
+
+    // new
+    PYBIND11_MODULE(example, m) {
+        m.doc() = "documentation string"; // optional
+
+        m.def("add", [](int a, int b) { return a + b; });
+    }
+
+
+Stricter enforcement of hidden symbol visibility for pybind11 modules
+---------------------------------------------------------------------
+
+pybind11 now tries to actively enforce hidden symbol visibility for modules.
+If you're using either one of pybind11's :doc:`CMake or Python build systems
+<compiling>` (the two example repositories) and you haven't been exporting any
+symbols, there's nothing to be concerned about. All the changes have been done
+transparently in the background. If you were building manually or relied on
+specific default visibility, read on.
+
+Setting default symbol visibility to *hidden* has always been recommended for
+pybind11 (see :ref:`faq:symhidden`). On Linux and macOS, hidden symbol
+visibility (in conjunction with the ``strip`` utility) yields much smaller
+module binaries. `CPython's extension docs`_ also recommend hiding symbols
+by default, with the goal of avoiding symbol name clashes between modules.
+Starting with v2.2, pybind11 enforces this more strictly: (1) by declaring
+all symbols inside the ``pybind11`` namespace as hidden and (2) by including
+the ``-fvisibility=hidden`` flag on Linux and macOS (only for extension
+modules, not for embedding the interpreter).
+
+.. _CPython's extension docs: https://docs.python.org/3/extending/extending.html#providing-a-c-api-for-an-extension-module
+
+The namespace-scope hidden visibility is done automatically in pybind11's
+headers and it's generally transparent to users. It ensures that:
+
+* Modules compiled with different pybind11 versions don't clash with each other.
+
+* Some new features, like ``py::module_local`` bindings, can work as intended.
+
+The ``-fvisibility=hidden`` flag applies the same visibility to user bindings
+outside of the ``pybind11`` namespace. It's now set automatic by pybind11's
+CMake and Python build systems, but this needs to be done manually by users
+of other build systems. Adding this flag:
+
+* Minimizes the chances of symbol conflicts between modules. E.g. if two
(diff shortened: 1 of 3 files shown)


docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+localize all common type bindings in order to avoid conflicts with
+third-party modules.
+
+
+New syntax for custom constructors
+----------------------------------
+
+The old placement-new custom constructors are still valid, but the new approach
+greatly improves type safety (placement-new can be called accidentally with an
+incompatible type, without any compiler errors or warnings). See
```
jagerman: ... with an incompatible type, or can be called multiple times on the same object if not careful with the Python-side `__init__` calls.

docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+
+    // old
+    py::class<Foo>(m, "Foo")
+        .def("__init__", [](Foo &self, ...) {
+            new (&self) Foo(...); // uses placement-new
+        });
+
+    // new
+    py::class<Foo>(m, "Foo")
+        .def(py::init([](...) {
```
jagerman: Suggest: ``// Note: no `self` argument`` to draw attention to that distinction.

docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+
+Deprecation of some ``py::object`` APIs
+---------------------------------------
+
+The equality comparison operator for ``py::object`` only compares the
+underlying pointers, therefore ``obj1 == obj2`` in C++ is equivalent to the
+expression ``obj1 is obj2`` in Python. To avoid confusion, the equality
+operator now emits a deprecation warning (in the future, it can be replaced
+with proper rich object comparison). To compare objects by pointer, use
+``obj1.is(obj2)``.
```
jagerman: I think this should be rewritten to mention `is` first (otherwise it seems to be suggesting that `==` is the right way to do it if I don't read the paragraph in full).

```
To compare ``py::object``s by pointer, you should now use ``obj1.is(obj2)``,
which is equivalent to ``obj1 is obj2`` in Python.  Previously pybind11 used
the `==` operator for this (`obj1 == obj2`), but that could be confusing and
is now deprecated (so that it can eventually be replaced with proper rich
object comparison in a future release).
```

docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+v2.0
+====
+
+Breaking changes in ``py::class_``
+----------------------------------
+
+These changes were necessary to make type definitions in pybind11
+future-proof, to support PyPy via its ``cpyext`` mechanism (`#527
+<https://github.com/pybind/pybind11/pull/527>`_), and to improve efficiency
+(`rev. 86d825 <https://github.com/pybind/pybind11/commit/86d825>`_).
```
jagerman: The referenced change (86d825 - moving trampolines declaration into the `class_<...>` template types) looks like it was already in 1.8.  Should that be a different commit?

jagerman: Oops, never mind, I was misreading the commit status.

docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+           .def(py::init<...>())
+           .def_buffer(...);
+
+2. Classes which include static properties (e.g. ``def_readwrite_static()``)
+   must now include the ``py::metaclass()`` attribute. Note: this requirement
+   has since been removed in v2.1. If you're upgrading from 1.x, it's
+   recommended to skip directly to v2.1 or newer.
+
+3. This version of pybind11 uses a redesigned mechnism for instantiating
+   trempoline classes that are used to override virtual methods from within
```
jagerman: trempoline -> trampoline

docs/upgrade.rst
```diff
@@ -0,0 +1,317 @@
...
+       py::class_<TrampolineClass>("MyClass")
+           .alias<MyClass>()
+           ...
+
+       // new v2.x syntax
+       py::class_<MyClass, TrampolineClass>("MyClass")
+           ...
+
+   Importantly, both the original and the trampoline class are now specified
+   as an arguments (in arbitrary order) to the ``py::class_`` template, and
```
jagerman: The "in arbitrary order" here is a bit confusing as it seems to be applying to "the original and the trampoline", but of course the original class still has to be first.  Perhaps: "(mixed in arbitrary order with other arguments such as a base class or holder type)".

jagerman:
I left a few minor wording comments, but overall I think this is a great idea.

docs/upgrade.rst
dean0x7d: Good point. The original was awkwardly worded.

docs/upgrade.rst
```diff
@@ -0,0 +1,322 @@
...
+The namespace-scope hidden visibility is done automatically in pybind11's
+headers and it's generally transparent to users. It ensures that:
+
+* Modules compiled with different pybind11 versions don't clash with each other.
+
+* Some new features, like ``py::module_local`` bindings, can work as intended.
+
+The ``-fvisibility=hidden`` flag applies the same visibility to user bindings
+outside of the ``pybind11`` namespace. It's now set automatic by pybind11's
+CMake and Python build systems, but this needs to be done manually by users
```
wjakob: I'm confused by this paragraph -- didn't the CMake build system always set -fvisibility=hidden on OSX & Linux?

wjakob: Aha, that point clarified itself a few lines below :)

docs/upgrade.rst
```diff
@@ -0,0 +1,322 @@
...
+* GCC >= 4.8
+* clang >= 3.3 (appleclang >= 5.0)
+* MSVC >= 2015u3
+* Intel C++ >= 15.0
+
+
+Deprecation of the ``py::metaclass()`` attribute
+------------------------------------------------
+
+It's no longer required for binding classes with static properties.
```
wjakob: Hmm, I think this is too strong. We didn't deprecate the ``py::metaclass(handle)`` constructors, which permits creating pybind11-bound classes with custom metaclasses (fairly powerful IMO).

How about the following instead? The ``py::metaclass`` attribute is no longer required for binding classes with static properties. It is still needed in rare cases, e.g.  to set a custom metaclasses for pybind11-bound data structures.

wjakob:
This is a great idea -- very useful!

docs/upgrade.rst
dean0x7d: It did read a bit like `py::metaclass` was completely deprecated instead of just the zero-parameter version. I've rewritten it now.

PR #1011, opened 2017-08-19
Fix typos in Eigen documentation

patrikhuber:
Fixes one small variable name typo, and two instances where `py::arg().nocopy()` is used, where I think it should be `py::arg().noconvert()` instead. Probably `nocopy()` was the old/original name for it and then it was changed.


jagerman:
Good catches, thanks!

PR #1012, opened 2017-08-20
Change internals ID and versioning scheme to avoid module conflicts

dean0x7d:
The current `PYBIND11_INTERNALS_ID` depends on the version of the library in order to isolate binary incompatible `internals` capsules. However, this does not preclude conflicts between modules built from different (binary incompatible) commits with the same version number. 

For example, if one module was built with an early v2.2.dev and submitted to PyPI, it could not be loaded alongside a v2.2.x release module -- it would segfault because of incompatible internals with the same ID. (Well, whichever module was loaded first would work, but the second one would segfault on initialization.) I'm sure that a lot of projects include pybind11 as a submodule and check out a particular commit from the master branch.

This PR changes the ID to depend on `PYBIND11_INTERNALS_VERSION` which is independent of the main library version. It's an integer which should be incremented whenever a binary incompatible change is made to `internals`.

`PYBIND11_INTERNALS_KIND` is also introduced for a similar reason.

Note: This would require some manual effort to make sure `PYBIND11_INTERNALS_VERSION` was correctly incremented for incompatible changes to `internals` or the related functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ set(PYBIND11_HEADERS
   include/pybind11/detail/common.h
   include/pybind11/detail/descr.h
   include/pybind11/detail/init.h
+  include/pybind11/detail/internals.h
   include/pybind11/detail/typeid.h
   include/pybind11/attr.h
   include/pybind11/buffer_info.h
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -169,7 +169,7 @@ inline detail::type_info *get_local_type_info(const std::type_index &tp) {
     auto &locals = registered_local_types_cpp();
     auto it = locals.find(tp);
     if (it != locals.end())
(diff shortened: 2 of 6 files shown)


jagerman:
Could we move the definition down a bit, right before `struct internals`, so that it's a very visible reminder whenever some touches `internals`?  I'm worried that it might be easy to forget and having it right there is a nice reminder.

dean0x7d:
I consolidated everything related to `internals` into a new header: `detail/internals.h` (~200 LOC). This code was previously split between `common.h` and `cast.h`.

I also made a couple of tweaks as stated in the commit messages. I'm don't think there was any reason for using `void*` instead of `type_info*`, or `noinline` on those functions.

dean0x7d:
`type_info` and `module_local` are also affected by incompatible changes. I modified the `module_local` attribute lookup to use the same versioning scheme as internals.

PR #1020, opened 2017-08-22
Fix "is_template_base_of" compilation.

### LINKED ISSUE #1018: Compilation fails when using Eigen under Visual Studio + LLVM or Intel Toolchain
When using a class containing an Eigen object (a VectorXd or MatrixD for example), it is not possible to compile under Visual Studio with LLVM or Intel Toolchain, we get the following error:

```
C:\Dev\libs\pybind11\include\pybind11/detail/common.h(664,92): error : 'type name' declared as a pointer to a reference of type 'remove_cv_t<const Eigen::Matrix<double, -1, 1, 0, -1, 1> &>' (aka 'const Eigen::Matrix<double, -1, 1, 0, -1, 1> &')
```

The following code illustrates the bug:

```c++
#include <pybind11/eigen.h>

namespace py = pybind11;

struct Vec
{
	Eigen::VectorXd v_;
};

PYBIND11_MODULE(test_eigen, m)
{
	py::class_<Vec>(m, "Vec")
		.def(py::init<>())
		.def_readwrite("v", &Vec::v_);

}
```
The problem comes from the fact that even if another toolchain is used, Visual Studio still defines `_MSC_VER`.

To correct it problem, I modified line 661 from `common.h` to this:

```c++
#if !defined(_MSC_VER) || defined(__clang__) || defined(__INTEL_COMPILER)
```

If there is no better way, I'll create a PR.

Baljak:
This corrects the compilation of Eigen types under Visual Studio when using another toolchain like LLVM or Intel C++ Compiler.

```is_template_base_of``` now uses ```intrinsic_t``` instead of ```remove_cv_t``` as suggested by @dean0x7d in #1018.

Fixes #1018.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -659,9 +659,9 @@ struct is_template_base_of_impl {
 /// `is_template_base_of<Base, T>` is true if `struct T : Base<U> {}` where U can be anything
 template <template<typename...> class Base, typename T>
 #if !defined(_MSC_VER)
-using is_template_base_of = decltype(is_template_base_of_impl<Base>::check((remove_cv_t<T>*)nullptr));
+using is_template_base_of = decltype(is_template_base_of_impl<Base>::check((intrinsic_t<T>*)nullptr));
 #else // MSVC2015 has trouble with decltype in template aliases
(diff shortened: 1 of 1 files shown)


dean0x7d:
Merged, thanks!

PR #1021, opened 2017-08-22
Speed up AppVeyor build

dean0x7d:
The AppVeyor builds have gotten ridiculously slow: 9-10 minutes per config and over an hour in total. This PR aims to bring them back to a more reasonable level: 3-4 minutes per config.

The `latest` build remains as is, but all others are modified to:

* Use regular Python instead of conda. `pip install` is much faster than conda, but scipy isn't available. Numpy is still tested.

* Compile in debug mode instead of release.

* Skip CMake build tests. For some reason, CMake configuration is very slow on AppVeyor and these tests are almost entirely CMake.

The `latest` config still covers scipy, release mode and the CMake build tests, so the others don't need to.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,12 +10,15 @@ platform:
 - x86
 environment:
   matrix:
-  - CONDA: 36
+  - PYTHON: 36
     CPP: 14
-  - CONDA: 27
+    CONFIG: Debug
+  - PYTHON: 27
     CPP: 14
+    CONFIG: Debug
   - CONDA: 36
     CPP: latest
+    CONFIG: Release
 matrix:
   exclude:
     - image: Visual Studio 2015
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -38,7 +41,7 @@ install:
       if ($env:PLATFORM -eq "x64") { $env:PYTHON = "$env:PYTHON-x64" }
       $env:PATH = "C:\Python$env:PYTHON\;C:\Python$env:PYTHON\Scripts\;$env:PATH"
       pip install --disable-pip-version-check --user --upgrade pip wheel
-      pip install pytest numpy scipy
+      pip install pytest numpy
```
jagerman: Maybe I'm missing something, but doesn't this end up taking scipy out of the `latest` build as well?

dean0x7d: `latest` is configured with `CONDA` while the others are `PYTHON`.

jagerman: Oh, right.  I forgot we had that whole (previously inactive) `PYTHON` build code already there.

dean0x7d: It was there waiting for a `scipy` Windows wheel, but that never came to be.

jagerman:
LGTM, and that's a very nice reduction.

YannickJadoul:
How much of this time is actually being spent on compilation? I haven't found the time to complete it yet, but I've been looking at integrating [clcache](https://github.com/frerich/clcache) into AppVeyor to speed up the compilation of my project. (I don't know what the gain would be in this case though, since pybind11 is header-only, so a single change could cause a lot of preprocessed compilation units to change?)

dean0x7d:
@wjakob AppVeyor now has an option for [rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds) similar to what was enabled on Travis a while ago. It might be a nice to enable it on AppVeyor as well, in order to ease some of the congestion.

@YannickJadoul Caching isn't really worth it because it's header-only and most of the headers are included in all the tests files. There's not much reuse and saving/loading the cache ends up being more time-consuming than the compilation.

PR #1023, opened 2017-08-23
Add missing comma in setup.py.

garth-wells:
Test framework must be missing a test for `setup.py`.


dean0x7d:
Fixed in #1026 along with subdirectory header fixes and new automated distribution/installation testing. Thanks for reporting!

PR #1027, opened 2017-08-24
Always specified CXX as language.

apollo13:
there should be no need to test for a C compiler since pybind only uses CPP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,12 @@ if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif()
 
-project(pybind11)
+# CMake versions < 3.4.0 do not support try_compile/pthread checks without C as active language.
+if(CMAKE_VERSION VERSION_LESS 3.4.0)
+  project(pybind11)
+else()
+  project(pybind11 CXX)
+endif()
 
 # Check if pybind11 is being used directly or via add_subdirectory
 set(PYBIND11_MASTER_PROJECT OFF)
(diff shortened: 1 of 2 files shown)


jagerman:
It's correct, in theory, but as you can see from the travis-ci builds, it causes some issues with older (but supported) versions of cmake.

apollo13:
This new patch should fix that, not sure if it is worth it. (I'd like to see it because I do not really need a C compiler and the checks sometimes cause more problems…)

dean0x7d:
Well, this does shave off a bit of time of the CI builds (especially AppVeyor) which is nice. But I haven't really been using the language settings in CMake, so I have a couple of questions:

1. What's actually the issue with CMake < 3.4?
2. Would this cause any issues for mixed C and C++ parent projects? (I would guess not, but just want to make sure we don't break anyone's build.)

apollo13:
> What's actually the issue with CMake < 3.4?

The issue is that try_compile on cmake < 3.4 does not work if you have no C compiler, later version happily use the C++ compiler which works just fine to test features (especially if you target CPP anyways).

> Would this cause any issues for mixed C and C++ parent projects? (I would guess not, but just want to make sure we don't break anyone's build.)

This should not cause issues for parent or child projects. The final supported languages basically are an union of all defined ones. There might be the weird situation where a parent project specified only CXX but relied on a C compiler to be present and usable… I think this case should not be a concern for pybind itself though (also by default cmake defaults to C & CXX, so that project would have to actively set CXX only…)

henryiii:
I believe @apollo13 is correct; the same thing happens if you are specifying CUDA as a language (CMake 3.7+); the final language is a union of the defined languages.

Older versions of CMake didn't take into account that someone might be only compiling CXX, so modules (notoriously the pthread module, for example) didn't work without C specified. That's been fixed in newer CMake versions.

So I don't think there is any downside to avoiding the C specification in PyBind11 for newer CMake versions.

dean0x7d:
OK, that seems good to me. @apollo13 Could you just add a very brief comment to `CMakeLists.txt` to explain the `CMake < 3.4` check?

dean0x7d:
Thank you, @apollo13.

Issue #579, opened 2016-12-29
Default named arguments for init functions


rmjarvis:
I'm working on converting a [project](https://github.com/GalSim-developers/GalSim) from Boost Python to PyBind11, mostly to try to make it pip installable, since the Boost dependency is kind of a stopper for that.

So far, one thing that boost.python does nicely that I can't figure out how to do easily in pybind11 is to give default values to the arguments of an `__init__` method.  

For instance, here is the code to wrap one of our classes using boost::python:
```
namespace bp = boost::python;
bp::class_<GaussianDeviate, bp::bases<BaseDeviate> >
        pyGaussianDeviate("GaussianDeviate", "", bp::no_init);
pyGaussianDeviate
    .def(bp::init<long, double, double>(
            (bp::arg("seed")=0, bp::arg("mean")=0., bp::arg("sigma")=1.)
    ))
```
(There's more than this of course, but this is the relevant bit.)

So far, my attempt to convert this to pybind11 API involves:
* change namespace to pybind11 (of course)
* remove `no_init`
* remove `bases` wrapper for `BaseDeviate`
* remove the parentheses around the args

which leads to
```
namespace bp = pybind11;
bp::class_<GaussianDeviate, BaseDeviate>
        pyGaussianDeviate("GaussianDeviate");
pyGaussianDeviate
    .def(bp::init<long, double, double>(
            bp::arg("seed")=0, bp::arg("mean")=0., bp::arg("sigma")=1.
    ))
```
However, this is still insufficient.  I get the following error from the compiler (using pybind11 2.0.0rc1)
```
pysrc/Random.cpp:143:40: error: use of overloaded operator '=' is ambiguous (with operand types
      'bp::arg' and 'int')
                        bp::arg("seed")=0, bp::arg("mean")=0., bp::arg("sigma")=1.
                        ~~~~~~~~~~~~~~~^~
/Users/Mike/Library/Python/2.7/include/python2.7/pybind11/attr.h:20:8: note: candidate is the
      implicit move assignment operator
struct arg {
       ^
/Users/Mike/Library/Python/2.7/include/python2.7/pybind11/attr.h:20:8: note: candidate is the
      implicit copy assignment operator
/Users/Mike/Library/Python/2.7/include/python2.7/pybind11/attr.h:24:24: note: candidate function
      [with T = int]
    constexpr arg_t<T> operator=(const T &value) const { return {name, value}; }
                       ^
```
[...]

PR #1028, opened 2017-08-24
[setup.cfg] set zip_safe to False

marscher:
This avoids a bug, if pybind11 is dragged in with setuptools.setup_requires,
as without the flag a zip file (.egg) would be created, which can not be used
as include directory.

[ci skip] until a proper test

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,6 @@
+[global]
+zip_safe=False
+
 [bdist_wheel]
 universal=1
 


marscher:
Eg. this fails for pip invocations:

```
(pyemma_dev) marscher@wallaby:/tmp$  pip install git+https://github.com/markovmodel/PyEMMA.git -v
Collecting git+https://github.com/markovmodel/PyEMMA.git
  Cloning https://github.com/markovmodel/PyEMMA.git to ./pip-lk2h0s8n-build
  Running command git clone -q https://github.com/markovmodel/PyEMMA.git /tmp/pip-lk2h0s8n-build
  Running command git submodule update --init --recursive -q
  Running setup.py (path:/tmp/pip-lk2h0s8n-build/setup.py) egg_info for package from git+https://github.com/markovmodel/PyEMMA.git
    Running command python setup.py egg_info
    /tmp/pip-lk2h0s8n-build/setup.py:272: UserWarning: using git, require cython
      warnings.warn('using git, require cython')
    zip_safe flag not set; analyzing archive contents...

    Installed /tmp/pip-lk2h0s8n-build/.eggs/pybind11-2.1.1-py3.6.egg
    running egg_info
    creating pip-egg-info/pyEMMA.egg-info
```

Note that the temporary install of pybind11 is an egg file (zip file), so it can not be used as an include directory in the parent setup.py

marscher:
https://github.com/pybind/pybind11/issues/579#issuecomment-269715352 I think this is causing by the fact that install_headers is not called, when setup.py install is called.

marscher:
This PR does not solve the underlying problem. Actually the Python package should include the headers it self and should not just be misused as an installer to the system headers location. The pybind11.get_include() is not of any use at the moment, because the default install location should always be contained in the compilers default include list.

Issue #1029, opened 2017-08-24
Reference leaks


wjakob:
pybind11 has a fairly obnoxious reference leak related to instances (it appears to me that it has *always* had this issue):

To reproduce:
```cpp
#include <pybind11/pybind11.h>
PYBIND11_MODULE(example, m) {
    struct A { };
    auto a = pybind11::class_<A>(m, "A").def(pybind11::init<>());
}
```

```pycon
Python 3.6.1 |Anaconda 4.4.0 (x86_64)| (default, May 11 2017, 13:04:09)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import example
>>> import sys
>>> sys.getrefcount(example.A)
4
>>> l = [example.A() for i in range(1000)]
>>> sys.getrefcount(example.A)
1004
>>> del l
>>> sys.getrefcount(example.A)
1004
>>>
```

Let's try to get this fixed for v2.2!

PR #1030, opened 2017-08-24
Fix for reference leak issue (#1029)

wjakob:
See issue #1029

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -334,9 +334,17 @@ inline void clear_instance(PyObject *self) {
 /// to destroy the C++ object itself, while the rest is Python bookkeeping.
 extern "C" inline void pybind11_object_dealloc(PyObject *self) {
     clear_instance(self);
+
     auto type = Py_TYPE(self);
     type->tp_free(self);
-    Py_DECREF(type);
+
+    // `type->tp_dealloc != pybind11_object_dealloc` means that we're being called
+    // as part of a derived type's dealloc, in which case we're not allowed to decref
+    // the type here. For cross-module compatibility, we shouldn't compare directly
+    // with `pybind11_object_dealloc`, but with the common one stashed in internals.
+    auto pybind11_object_type = (PyTypeObject *) get_internals().instance_base;
+    if (type->tp_dealloc == pybind11_object_type->tp_dealloc)
+        Py_DECREF(type);
 }
 
 /** Create the type which can be used as a common base for all classes.  This is
@@ -585,6 +593,8 @@ inline PyObject* make_new_python_type(const type_record &rec) {
     /* Register type with the parent scope */
     if (rec.scope)
         setattr(rec.scope, rec.name, (PyObject *) type);
+    else
+        Py_INCREF(type); // Keep it alive forever (reference leak)
 
     if (module) // Needed by pydoc
         setattr((PyObject *) type, "__module__", module);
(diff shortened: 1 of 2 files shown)


wjakob:
@dean0x7d, @jagerman: This patch fixes the reference leak for "normal" ``py::init<>`` construction, but it seems to cause with the new factory initialization (some kind of double free?). Would you mind taking a look?

dean0x7d:
As far as I can see, the new negative refcount issue has to do with types derived in Python:
```python
class PyDog(m.Dog):
    pass

instance = PyDog("Molly")
del instance  # `m.Dog` is decremented twice
```

Looks like the issue is [here in `subtype_dealloc()`](https://github.com/python/cpython/blob/8621bb5d93239316f97281826461b85072ff6db7/Objects/typeobject.c#L1248-L1254) which is the `dealloc` used by types created in Python. It calls `basedealloc(self)` and then `Py_DECREF(type)`. But since `basedealloc == pybind11_object_dealloc`, it calls another `Py_DECREF(type)` and we end up with a negative refcount.

The solution could be to replace the `Py_DECREF(type)` in `pybind11_object_dealloc` with:
```c++
if (type->tp_dealloc == pybind11_object_dealloc)
    Py_DECREF(type);
```

Unfortunately, while this does fix everything in `test_class`, I get additional errors in `test_sequences_and_iterators` and `test_virtual_functions`, so there's more to this.

dean0x7d:
That should do it. One issue was types derived in Python as mentioned above. The other issue had to do with types which are defined without a scope, e.g. like [`py:: make_iterator`](https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1451). The patch above fixes unscoped types by leaking the reference so they end up living forever. The only alternative I can think of is to ban unscoped types.

wjakob:
I think that is a very reasonable fix.

wjakob:
By the way: I noticed this issue in one of my projects where I had been using a ``py::capule``-based module shutdown handler similar to what we currently describe [in the documentation](http://pybind11.readthedocs.io/en/stable/advanced/misc.html#module-destructors). The trouble with that approach is that instances of classes defined within the module may still be alive when the shutdown handler is called. 

I thus switched to something like the following
```cpp
py::cpp_function cleanup_callback(
    [](py::handle weakref) {
        /* ... shutdown logic .. */
        weakref.dec_ref();
    }
);
(void) py::weakref(m.attr("BaseClass"), cleanup_callback).release();
```
and was surprised that it stopped getting called as soon as any ``BaseClass``-derived instance was created.

In any case, now that this is fixed, it might be worth mentioning this alternative destruction logic in the docs.

wjakob:
Fantastic, all tests check out 👍

dean0x7d:
> In any case, now that this is fixed, it might be worth mentioning this alternative destruction logic in the docs. What do you think?

Seems like it would go along nicely with module destructors. As long as the refcounts are working, I think a simple `setattr` should work as well:
```c++
m.attr("BaseClass").attr("_cleanup") = py::capsule(cleanup_callback);
```

> PS: Looks like we're on hacker news front page (https://news.ycombinator.com/item?id=15095757), cool!

Neat. It's certainly an interesting spike in traffic :)

visitors

wjakob:
> Seems like it would go along nicely with module destructors. As long as the refcounts are working, I think a simple setattr should work as well.

Indeed, putting the capsule into the ``BaseClass`` is actually much simpler, never mind the weak reference suggestion. I'll make a PR for the precise wording.

wjakob:
I went ahead and squash-merged this.

wjakob:
Upon further thought, I think it's helpful to document the ``weakref``-based approach as well. The big downside of the capsule approach is that a ``_cleanup`` attribute is exposed on the Python side. Calling that may lead to any number of unintended & undefined behaviors/crashes. The weakref is invisible and does not suffer from that problem.

jagerman:
I know I'm a little late (I'm in the middle of a move, and will probably be mostly inactive for the next 2-3 days as well), but the fix looks good; 👍 to @dean0x7d for tracking that down!

PR #1037, opened 2017-08-28
Make implicit conversions non-reentrant (fixes #1035)

### LINKED ISSUE #1035: Runaway recursion in implicit conversion may trigger undefined behavior
The following program with a bogus implicit conversion exhibits curious behavior:

```cpp

namespace py = pybind11;

PYBIND11_MODULE(example, m) {
    struct A { A() { } };

    py::class_<A>(m, "A")
        .def(py::init<const A &>());

    py::implicitly_convertible<int, A>();
}
```

Test driver:
```python
import example
example.A(1)
```

The implicit conversion gets triggered again and again (due to the copy constructor) until the stack overflows.  On some platforms (Linux, OSX), Python seems to catch the stack overflow internally and throws an exception which then causes the implicit conversion to fail. After that, pybind11 just prints a normal parameter error, hiding the undesired program behavior that just transpired:

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. example.A(arg0: example.A)
```

On Windows + Python 2.7 (MSVC2017 debug mode), this seems to crash fairly reliably. (I stumbled across this issue while debugging an AppVeyor CI failure of a piece of software I develop. The implicit conversion was a lot more subtle in my case and additionally involved a custom type converter.)

Although one could argue that the infeasible implicit conversion is to blame for the problem here, I don't think that it should be allowed to trigger undefined behavior. 

One potential fix would be to disable implicit conversion by default when executing a copy constructor. I'm not sure about the best way to add that and wanted to bring it up for discussion:

One way would be to detect it at compile time -- for instance, the following seems to do the job:
```cpp
/// in common.h
template <typename T, typename... Args> struct is_copy_constructor : std::false_type { };
template <typename T> struct is_copy_constructor<T, T> : std::true_type { };

/// in init.h, constructor::Execute
constexpr bool copy_constructor = detail::is_copy_constructor<Cpp<Class>, intrinsic_t<Args>...>::value;
// now what?
```
This would need to be turned int some kind of attribute that gets passed into ``class_::def``, potentially using a construction like ``std::conditional_t<noconvert, nop>()``. The downside of this is adding further attributes and flags to ``function_record``.

Another alternative would be to it in ``cpp_function::initialize_generic`` by analyzing the function signature. I tried this as well, but it seemed brittle.

Thoughts?

Wenzel

wjakob:
See issue #1035

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -585,6 +585,10 @@ Python side:
     Implicit conversions from ``A`` to ``B`` only work when ``B`` is a custom
     data type that is exposed to Python via pybind11.
 
+    To prevent runaway recursion, implicit conversions are non-reentrant: an
+    implicit conversion invoked as part of another implicit conversion of the
+    same type (i.e. from ``A`` to ``B``) will fail.
+
 .. _static_properties:
 
 Static properties
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1510,7 +1510,16 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
 }
 
 template <typename InputType, typename OutputType> void implicitly_convertible() {
+    struct set_flag {
+        bool &flag;
+        set_flag(bool &flag) : flag(flag) { flag = true; }
+        ~set_flag() { flag = false; }
+    };
     auto implicit_caster = [](PyObject *obj, PyTypeObject *type) -> PyObject * {
+        static bool currently_used = false;
+        if (currently_used) // implicit conversions are non-reentrant
+            return nullptr;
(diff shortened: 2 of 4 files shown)


wjakob:
Weird.. not sure I understand these CI failures.

dean0x7d:
Try an empty copy constructor instead of an explicitly defaulted one:
```c++
BogusImplicitConversion(const BogusImplicitConversion &) {}
```

Early C++11 compilers had a bug where braces would prefer aggregate initialization over copy constructors. A struct with an explicitly defaulted copy constructor is still considered an aggregate.

wjakob:
Thank you for the suggestion

PR #1039, opened 2017-08-28
Clarify what is meant by debug mode.

anntzer:
This should clarify the error message for those relying on setuptools
(which does not have the concept of debug mode, but can be configured to
add arbitrary flags to the compiler call) rather than cmake.

See discussion starting at https://gitter.im/pybind/Lobby?at=59a3c8db614889d4759441b8

Feel free to modify the error message as you see fit by force-pushing to my branch.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -382,7 +382,7 @@ template <> struct process_attribute<arg_v> : process_attribute_default<arg_v> {
 #else
             pybind11_fail("arg(): could not convert default argument "
                           "into a Python object (type not registered yet?). "
-                          "Compile in debug mode for more information.");
+                          "Compile in debug mode (#undef NDEBUG) for more information.");
 #endif
         }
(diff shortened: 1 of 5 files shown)


wjakob:
I'd rather mention this in the part of the documentation that references ``setuptools``. This modified comment pushes people to manually undefine NDEBUG rather than using the intended mechanism for that (e.g. on CMake/Visual Studio/..)

dean0x7d:
This is more of a `setuptools` issue since other build systems handle `NDEBUG` automatically. `setuptools` does have the concept of a debug mode (`python setup.py build --debug`) but it always defines `NDEBUG`. It's pretty easy to add extra code to `setup.py` to handle this. E.g. [`cmake_example/setup.py` already does this](https://github.com/pybind/cmake_example/blob/master/setup.py#L39-L40). Something similar with `undef NDEBUG` could be done for `python_example`'s `setup.py`.

wjakob:
Closing this issue (it's fairly old, and there were a few objections)

Issue #1041, opened 2017-08-29
function<Type*(void)> will attempt to copy Type on return


gratzdhg:
Firstly, thank you for this wonderful library. As I was using it I found some unexpected/undesired behaviour. Wrapped function objects from std::functional will copy the result even when the result is of type pointer.
Here is my sample code to reproduce the issue.
```c++
#include <functional>
#include <pybind11/pybind11.h>
#include <pybind11/functional.h>

namespace py = pybind11;

struct Test {
    Test() {};
    Test(const Test &obj) = delete;
    int i;
};
Test* fullFunction()
{
    return new Test();
}
std::function<Test*(void)> func = &fullFunction;

PYBIND11_MODULE(example, m) {
    // optional module docstring
    m.attr("full") = py::cpp_function(func);
    m.attr("func") = func;
    py::class_<Test>(m,"Test")
        .def_readwrite("i",&Test::i);
}
```
```python
import example
example.full()
example.func()
```
example.func() will result in an error: return_value_policy = copy, but the object is non-copyable!

It's my understanding that both full & func should return a Test* and that the underlying Test will not be copied however when the pybind11/functional wrapper handles it, it will attempt to copy Test*. I believe this occurs because a function object is an object and not a pointer which causes copy to be the default behaviour that is detected. Sorry if I have misunderstood anything.

henryiii:
The problem seems to lie in 

https://github.com/pybind/pybind11/blob/37de2da9dd72d22e4d57f18ce61bac0738053371/include/pybind11/functional.h#L72-L75

where it inherits the parent call policy, which is `copy` since the function object is not a pointer. This is the correct parent call policy, but inheriting it is probably incorrect. Removing the explicit call policies here fixes this problem.

gratzdhg:
thanks, my solution was to replace those lines with
```c++
        auto res_policy = policy;
        res_policy = std::is_pointer<Return>::value ? return_value_policy::take_ownership : return_value_policy::copy;

        auto result = f_.template target<function_type>();
        if (result)
            return cpp_function(*result, res_policy).release();
        else
            return cpp_function(std::forward<Func>(f_), res_policy).release();
```
but I'm not confidant that this would always be correct

henryiii:
Simply removing the explicit policies seems to also pass all tests, so I'm assuming this is either also correct or not tested. Waiting for someone with a bit more experience with PyBind11 to weigh in. I think something like you've written above probably is a reasonable solution?

[...]

PR #1043, opened 2017-08-30
Functional returning a pointer policy

henryiii:
This is a proposed fix for #1041. This auto-selects a return policy when a functional gets auto-wrapped, instead of inheriting the copy policy that is left over from the auto policy selection.  Added tests that fail without the patch, and succeed with the patch.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -891,7 +891,7 @@ T cast(const handle &handle) { return T(reinterpret_borrow<object>(handle)); }
 template <typename T, detail::enable_if_t<!detail::is_pyobject<T>::value, int> = 0>
 object cast(T &&value, return_value_policy policy = return_value_policy::automatic_reference,
             handle parent = handle()) {
-    return reinterpret_steal<object>(detail::make_caster<T>::cast(value, policy, parent));
+    return reinterpret_steal<object>(detail::make_caster<T>::cast(std::forward<T>(value), policy, parent));
 }
 
 template <typename T> T handle::cast() const { return pybind11::cast<T>(*this); }
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -891,14 +891,7 @@ T cast(const handle &handle) { return T(reinterpret_borrow<object>(handle)); }
 template <typename T, detail::enable_if_t<!detail::is_pyobject<T>::value, int> = 0>
 object cast(T &&value, return_value_policy policy = return_value_policy::automatic_reference,
             handle parent = handle()) {
-    using no_ref_T = typename std::remove_reference<T>::type;
-    if (policy == return_value_policy::automatic)
-        policy = std::is_pointer<no_ref_T>::value ? return_value_policy::take_ownership :
-                 std::is_lvalue_reference<T>::value ? return_value_policy::copy : return_value_policy::move;
-    else if (policy == return_value_policy::automatic_reference)
-        policy = std::is_pointer<no_ref_T>::value ? return_value_policy::reference :
-                 std::is_lvalue_reference<T>::value ? return_value_policy::copy : return_value_policy::move;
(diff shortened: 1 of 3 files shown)


henryiii:
Feel free to close this (or I will) if the decision is that forwarding on function creation is correct (it very well may be).

wjakob:
I don't think this is the best way to address the underlying issue, since it imposes a fixed policy -- instead, we should fix the logic that forwards the return value policy to the std::function caster (see my comment on #1041)

henryiii:
Sounds good. The tests in the PR are possibly still useful as a basis to see if the forwarding behavior is fixed.

henryiii:
I tracked down the policy calls and it seems to be making a decision to copy at the end if you keep automatic, as well. So if you simply comment out these four lines:

https://github.com/pybind/pybind11/blob/689867927033d3de22241b4e8e95d2bf25a350a7/include/pybind11/cast.h#L1607-L1610

Automatic gets passed down the chain and the test I added passes.

The place where automatic goes to copy is here:

https://github.com/pybind/pybind11/blob/689867927033d3de22241b4e8e95d2bf25a350a7/include/pybind11/cast.h#L786-L787

Is there a reason for those initial lines?

dean0x7d:
Removing those policy translations from `py::cast()` does seem reasonable to me and it looks like it doesn't affect the tests in any way. AFAIK all the casters which need policies already know how to translate policies internally because other to-Python conversions (which don't do through `py::cast()`) don't have these extra policy checks.

The only questions is if there are any casters or situations that do require the policy translations made in `py::cast()`, but that are not currently covered by the test suite. @wjakob Is there a special reason for those 4 lines in `py::cast()`? If they can be removed, it would propagate the automatic policy to the casters which seems desirable.

(Although, I wouldn't recommend to make this change for v2.2.1. Better for v2.3 so it can sit in `master` for  a while and we can see if there are any issues with it.)

henryiii:
Does this need to be merged to master so it can sit in master for a while? I've rebased on current master.

include/pybind11/cast.h
```diff
@@ -891,14 +891,7 @@ T cast(const handle &handle) { return T(reinterpret_borrow<object>(handle)); }
...
             handle parent = handle()) {
-    using no_ref_T = typename std::remove_reference<T>::type;
-    if (policy == return_value_policy::automatic)
-        policy = std::is_pointer<no_ref_T>::value ? return_value_policy::take_ownership :
-                 std::is_lvalue_reference<T>::value ? return_value_policy::copy : return_value_policy::move;
-    else if (policy == return_value_policy::automatic_reference)
-        policy = std::is_pointer<no_ref_T>::value ? return_value_policy::reference :
-                 std::is_lvalue_reference<T>::value ? return_value_policy::copy : return_value_policy::move;
-    return reinterpret_steal<object>(detail::make_caster<T>::cast(std::forward<T>(value), policy, parent));
+    return reinterpret_steal<object>(detail::make_caster<T>::cast(value, policy, parent));
```
henryiii, line 894:
```suggestion
    return reinterpret_steal<object>(detail::make_caster<T>::cast(std::forward<T>(value), policy, parent));
```

tests/test_callbacks.cpp
```diff
@@ -12,6 +12,16 @@
 #include <pybind11/functional.h>
 #include <thread>
 
+struct TestNotCopyable{
+    TestNotCopyable() {};
```
henryiii, line 16:
```suggestion
    TestNotCopyable() = default;
```

tests/test_callbacks.py
```diff
@@ -193,3 +193,12 @@ def test_callback_num_times():
                 min(rates), sum(rates) / len(rates), max(rates)
             )
         )
+
+
+def test_full_func_return_ptr():
+    m.full_func_return_ptr()
+
+
+@pytest.mark.xfail(reason="Currently auto policy is not inherited")
```
henryiii, line 202:
```suggestion
```

PR #1045, opened 2017-08-30
Fix style script and add comment to failing blocks

### LINKED ISSUE #1044: `check_style.sh` script issue
The `check_style.sh` script is reporting an error in the `iostream` example in the docs. There are several problems:

* The "error" is incorrect; the brackets is is picking up with the simple grep are not supposed to be attached to a statement.
* The "error" is in a doc-string
* The error is not causing the script to return a non-zero exit code for some reason, causing the test to pass (and only showing up in the log or if you run it manually)

~~Possible solution: Make sure it is looking for `)\w*{`~~, and find out why it is returning 0.

Edit: Since it looks line-by-line, the fix above might be a bit tricky to implement. Another solution would be to replace the single brackets on the two lines in the iostream examples with:
```cpp
{ // block
```

henryiii:
This fixes #1044. A bash while loop cannot modify the surrounding environment (runs in a sub-shell), so this was rewritten to be simpler and give the correct error code so travis will report errors correctly.

This also fixes the failing lines in iostream's `rst` file by adding `{ // block`.

### DIFF
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -75,7 +75,7 @@ NAMESPACE_END(detail)
 
         ...
 
-        {   // Block
+        {
             py::scoped_ostream_redirect output;
             std::cout << "Hello, World!"; // Python stdout
         } // <-- return std::cout to normal
@@ -85,7 +85,7 @@ NAMESPACE_END(detail)
 
     .. code-block:: cpp
 
-        {   // Block
+        {
             py::scoped_ostream_redirect output{std::cerr, py::module::import("sys").attr("stderr")};
             std::cerr << "Hello, World!";
         }
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -75,7 +75,7 @@ NAMESPACE_END(detail)
 
         ...
 
-        {
+        /* block */ {
             py::scoped_ostream_redirect output;
             std::cout << "Hello, World!"; // Python stdout
         } // <-- return std::cout to normal
@@ -85,7 +85,7 @@ NAMESPACE_END(detail)
 
     .. code-block:: cpp
 
-        {
+        /* block */ {
             py::scoped_ostream_redirect output{std::cerr, py::module::import("sys").attr("stderr")};
             std::cerr << "Hello, World!";
         }
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -75,7 +75,7 @@ NAMESPACE_END(detail)
 
         ...
 
-        /* block */ {
+        {   // Block
             py::scoped_ostream_redirect output;
             std::cout << "Hello, World!"; // Python stdout
         } // <-- return std::cout to normal
@@ -85,7 +85,7 @@ NAMESPACE_END(detail)
 
     .. code-block:: cpp
 
-        /* block */ {
+        {   // Block
             py::scoped_ostream_redirect output{std::cerr, py::module::import("sys").attr("stderr")};
             std::cerr << "Hello, World!";
         }
--- a/tools/check-style.sh
+++ b/tools/check-style.sh
@@ -37,7 +37,7 @@ fi
 found="$(GREP_COLORS='mt=41' GREP_COLOR='41' grep '[[:blank:]]\+$' include tests/*.{cpp,py,h} docs/*.rst -rn --color=always )"
 if [ -n "$found" ]; then
     # The mt=41 sets a red background for matched trailing spaces
-    echo -e '\033[31,01mError: found trailing spaces in the following files:\033[0m'
+    echo -e '\033[31;01mError: found trailing spaces in the following files:\033[0m'
     check_style_errors=1
     echo "$found" | sed -e 's/^/    /'
 fi
--- a/tools/check-style.sh
+++ b/tools/check-style.sh
@@ -64,7 +64,7 @@ last && /^\s*{/ {
 if [ -n "$found" ]; then
     check_style_errors=1
     echo -e '\033[31;01mError: braces should occur on the same line as the if/while/.. statement. Found issues in the following files:\033[0m'
-    echo "$found" | sed -e 's/^/    /'
+    echo "$found"
 fi
 
 exit $check_style_errors
--- a/tools/check-style.sh
+++ b/tools/check-style.sh
@@ -21,50 +21,50 @@ IFS=$'\n'
 found="$( GREP_COLORS='mt=41' GREP_COLOR='41' grep $'\t' include tests/*.{cpp,py,h} docs/*.rst -rn --color=always )"
 if [ -n "$found" ]; then
     # The mt=41 sets a red background for matched tabs:
-    echo -e '\033[31m\033[01mError: found CRLF characters in the following files:\033[0m'
+    echo -e '\033[31;01mError: found tab characters in the following files:\033[0m'
     check_style_errors=1
-    echo "$found" | while read line; do
-        echo -e '\033[31m\033[01mError: found tabs instead of spaces in the following files:\033[0m'
-    done
+    echo "$found" | sed -e 's/^/    /'
 fi
 
 
 found="$( grep -IUlr $'\r' include tests/*.{cpp,py,h} docs/*.rst --color=always )"
 if [ -n "$found" ]; then
-    echo -e '\033[31m\033[01mError: found CRLF characters in the following files:\033[0m'
+    echo -e '\033[31;01mError: found CRLF characters in the following files:\033[0m'
     check_style_errors=1
-    echo "$found" | while read line; do
-        echo "    $line"
-    done
+    echo "$found" | sed -e 's/^/    /'
 fi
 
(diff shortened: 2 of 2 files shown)


henryiii:
Oops, that was a mistake on my part. Can't put a block comment inside a comment. I should just fix the script that looks for `{`.

henryiii:
By the way, a related fix that I could include in this PR:

The current Travis order has the "check" style environment last, which causes it to wait until most of the previous environments in the matrix are finished. Since this fails often and is a bit harder to test locally, it makes sense to make it first or second, so developers can get faster feedback if they are watching the build.

Would that be a good idea?

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
-    if [ -z "$found" ]; then
-        echo -e '\033[31m\033[01mError: found CRLF characters in the following files:\033[0m'
-        found=1
-        errors=1
-    fi
-
-    echo "    $f"
-done
+found="$( GREP_COLORS='mt=41' GREP_COLOR='41' grep $'\t' include tests/*.{cpp,py,h} docs/*.rst -rn --color=always )"
+if [ ! -z "$found" ]; then
```
jagerman: Simpler: `if [ -n "$found" ]; then`

jagerman: Simpler: `if [ -n "$found "]; then`.  (The same goes for the later checks).

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
+    # The mt=41 sets a red background for matched tabs:
```
jagerman: This comment got moved out of place--it should be moved up (to just before the `found=$(...`).

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
+    echo -e '\033[31m\033[01mError: found CRLF characters in the following files:\033[0m'
+    check_style_errors=1
+    echo "$found" | while read line; do
+        echo -e '\033[31m\033[01mError: found tabs instead of spaces in the following files:\033[0m'
```
jagerman: This looks like a copy-and-paste error: the echo inside the read loop should just be `echo "    $line"`.  However I think it would be easier to just replace the entire
```bash
    echo "$found" | while read line; do
        echo "    $line"
    done
```
with just:
```bash
    echo "$found" | sed -e 's/^/    '
```

The following blocks print the right thing, but would still benefit from the same one-line `sed` simplification.

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
-        echo -e '\033[31m\033[01mError: found trailing spaces in the following files:\033[0m'
-        found=1
-        errors=1
-    fi
 
-    echo "    $f"
-done
+found="$( grep -IUlr $'\r' include tests/*.{cpp,py,h} docs/*.rst --color=always )"
+if [ ! -z "$found" ]; then
+    echo -e '\033[31m\033[01mError: found CRLF characters in the following files:\033[0m'
```
jagerman: This existed before your PR, but we can simplify `\033[31m\033[01m` to `\033[31;01m`.

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
-grep '\<\(if\|for\|while\|catch\)(\|){' include tests/*.{cpp,py,h} -rn --color=always |
-        while read line; do
-    if [ -z "$found" ]; then
-        echo -e '\033[31m\033[01mError: found the following coding style problems:\033[0m'
-        found=1
-        errors=1
-    fi
+found="$(GREP_COLORS='mt=41' GREP_COLOR='41' grep '[[:blank:]]\+$' include tests/*.{cpp,py,h} docs/*.rst -rn --color=always )"
+if [ ! -z "$found" ]; then
+    # The mt=41 sets a red background for matched trailing spaces
```
jagerman: Misplaced comment (as above).

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
+    echo -e '\033[31m\033[01mError: found trailing spaces in the following files:\033[0m'
+    check_style_errors=1
+    echo "$found" | while read line; do
+        echo "    $line"
+    done
+fi
 
-    echo "    $line"
-done
+found="$(grep '\<\(if\|for\|while\|catch\)(\|){' include tests/*.{cpp,py,h} -rn --color=always)"
```
jagerman: I think it's a mistake (in the existing code, copied over here) to run this test again `tests/*.py` as far as I can see; let's remove the `,py` from the grep argument input files.

tools/check-style.sh
```diff
@@ -11,72 +11,60 @@
...
-    echo "    $f"
-done
+found="$(GREP_COLORS='mt=41' GREP_COLOR='41' grep '^\s*{\s*$' include docs/*.rst -rn --color=always)"
+if [ ! -z "$found" ] ; then
+    check_style_errors=1
+    echo -e '\033[31m\033[01mError: braces should occur on the same line as the if/while/.. statement. Found issues in the following files: \033[0m'
+    echo "$found" | while read line; do
+        echo "    $line"
+    done
+fi
```
jagerman: This whole block is hitting false positives (both before and after the PR).  We aren't running it on `tests/*.{cpp,h}` basically because of this, and now, as you pointed out, we hit the same error in the documentation.

Adding a comment such as `{ // block` is really an ugly hack--it isn't a helpful comment, but is just serving to work around a defect in the style checker.

I think we're going to have to move beyond `grep` to fix this.  The following uses `awk` (with emulation of `grep`'s colour output) to do the check with a little more sophistication, but should get the behaviour we're after and lets us include the `cpp`/`h` files in `tests`:

```bash
found="$($(awk 'function prefix(filename, lineno) {
    return "    \033[35m" filename "\033[36m:\033[32m" lineno "\033[36m:\033[0m"
}
function mark(pattern, string) { sub(pattern, "\033[01;31m&\033[0m", string); return string }
last && /^\s*{/ {
    print prefix(FILENAME, FNR-1) mark("\\)\\s*$", last)
    print prefix(FILENAME, FNR)   mark("^\\s*{", $0)
    last=""
}
{ last = /(if|for|while|catch|switch)\s*\(.*\)\s*$/ ? $0 : "" }
' `find include -type f` tests/*.{cpp,h} docs/*.rst)"
if [ -n "$found" ]; then
    check_style_errors=1
    echo -e '\033[31;01mError: braces should occur on the same line as the if/while/.. statement. Found issues in the following files:\033[0m'
    echo "$found"
fi
```

include/pybind11/iostream.h
```diff
@@ -75,7 +75,7 @@ NAMESPACE_END(detail)
 
         ...
 
-        {
+        /* block */ {
```
jagerman: These shouldn't be necessary anymore (with the `awk` replacement I give below).

include/pybind11/iostream.h
```diff
@@ -85,7 +85,7 @@ NAMESPACE_END(detail)
 
     .. code-block:: cpp
 
-        {
+        /* block */ {
```
jagerman: Likewise.

include/pybind11/iostream.h
```diff
@@ -75,7 +75,7 @@ NAMESPACE_END(detail)
 
         ...
 
-        {
+        {   // Block
```
jagerman: These shouldn't be necessary anymore (with the `awk` replacement).

jagerman:
Sorry for that--I meant to submit all the review at once rather than just the first comment about the `if`.  It looks like I also applied my comments to just the first commit, and some now appear outdated (but are still things I think should be addressed).

tools/check-style.sh
henryiii: Added, though needed: 
 ```bash
echo "$found" | sed -e 's/^/    /'
```

tools/check-style.sh
henryiii: Fixed some typos, added.

tools/check-style.sh
henryiii: Done (too soon, as you know)

tools/check-style.sh
henryiii: Didn't really look at the colors, yes that's obvious. Fixed.

tools/check-style.sh
henryiii: Done.

include/pybind11/iostream.h
henryiii: Removed.

tools/check-style.sh
jagerman: Whoops, yeah, accidentally left out the last `/`.

jagerman:
> The current Travis order has the "check" style environment last ...

It looks like current master already does that as of a few hours ago (in cdf38dc6ba346f2aba52240d906b3ef0eba544ec).

henryiii:
Yep, @dean0x7d told me on Gitter that he added it to another PR.

jagerman:
I created some [test files](https://jagerman.com/bad-style.tar.gz) to trigger the various failures, and pushed a couple minor fixes to your branch.  Once the tests pass, this should be good to merge (unless anyone else has anything to add/change).

PR #1056, opened 2017-09-02
Speed up Travis CI build

dean0x7d:
Similarly to the recent AppVeyor changes, this removes the `test_cmake_build` target from all configurations except for two: one on Linux and one on macOS. This is pretty much all that's needed to cover those (they only depend on the platform, not on the compiler) and it saves ~30 seconds build time on the other configs.

The Python 3 on macOS build gets an updated image to speed up the homebrew Python installation (the Python 2.7 build is still using Travis's default osx image).

This also removes two configurations completely (down to 11 from 13):

- Style/docs/pip is merged with barebones.
- The 32-bit build is merged with CMake install/standalone tests (this removes clang 3.9 from testing, but there are already 3 other clang versions being tested on Travis and the new xcode8.3 image should be close to clang 3.9).

Each change is committed separately for easier review -- to be squashed before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
         packages: [python3.6-dev python3.6-venv, cmake=2.\*, cmake-data=2.\*]
   - sudo: true
     services: docker
-    env: PYTHON=2.7 CPP=14 GCC=6
+    env: PYTHON=2.7 CPP=14 GCC=6 CMAKE=1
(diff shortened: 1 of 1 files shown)


dean0x7d:
If there are no objections, I'll go ahead and merge this. There's a huge backlog of macOS builds on Travis again. Looks like it will need all the help it can get.

PR #1058, opened 2017-09-02
Fix nullptr dereference when loading an external-only module_local type

dean0x7d:
I ran into a cross-module segfault. The test case reproduces it and the `nullptr` check fixes it.


dean0x7d:
Note: this fix is incomplete, it's missing a type check when `typeinfo == nullptr`. I'll push a more complete solution in a bit.

dean0x7d:
OK, this fix is complete now and the test cases should cover it. Note that this did require adding an extra `std::type_info *` member to `type_caster_generic`.

PR #1062, opened 2017-09-04
Fix compilation with Clang on host GCC < 5 (old libstdc++)

dean0x7d:
This issue comes from Gitter: https://gitter.im/pybind/Lobby?at=59ad8678162adb6d2e68311c

The libstdc++ which comes with GCC 4.x does not implement `std::is_trivially_copyable`, so there's a workaround in place. But it does not take into account that Clang can also be installed on a host GCC 4 with the same incomplete libstdc++.

The fix here uses `_GLIBCXX_USE_CXX11_ABI` which is defined in the libstdc++ from GCC 5 and not defined in older versions.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,9 @@ v2.3.0 (Not yet released)
 v2.2.1 (Not yet released)
 -----------------------------------------------------
 
+* Fixed compilation with Clang on host GCC < 5 (old libstdc++ which isn't fully
+  C++11 compliant). `#1062 <https://github.com/pybind/pybind11/pull/1062>`_.
+
 * Fixed a regression where the ``py::keep_alive`` policy could not be applied
   to constructors. `#1065 <https://github.com/pybind/pybind11/pull/1065>`_.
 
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
This looks good to me. Thanks for the quick fix.

SylvainCorlay:
For reference: log from gitter

> Also, this seems to break the build with clang + linux, where the STL of GCC is used, if the host GCC is <5.
> Indeed, the following is broken when clang uses the STL of GCC (which is always on linux), and the host GCC is <5.
> 
> ```cpp
> #if !defined(__GNUG__) || defined(__clang__) || __GNUC__ >= 5
>     std::is_trivially_copyable<T>,
> #else
>     // GCC 4 doesn't implement is_trivially_copyable, so approximate it
>     std::is_trivially_destructible<T>,
>     satisfies_any_of<T, std::has_trivial_copy_constructor, std::has_trivial_copy_assign>,
> #endif
> ```
> 
> This is really not a simple problem to fix, because clang overrides the `__GNUC__ `macros, and the underlying version of the GCC STL cannot be detected with macros.
> 
> There could be a temptation to use the `__GLIBCXX__`  macro instead , but it is a date, and old versions of gcc may have very recent releases for security patches. Also, sometimes, this can be equal for two different versions of gcc.
> So there is no reliable means to detect the version of the STL from gcc used by clang with macros.
> You could force the user to use `libc++` instead of `libstdc++` on linux, but it would break a lot of downstream code.
> A reason why libstdc++ is the default on linux is that libc++ still has issues on this platform with IO operations.

PR #1063, opened 2017-09-04
Add a dummy common.h header with a deprecation warning

dean0x7d:
`common.h` was never a public header (undocumented and it never did anything when included alongside `pybind11.h` or any other header), but it does look like it was being used: https://gitter.im/pybind/Lobby?at=59ad8483614889d475bd4d3a

This PR adds a dummy file with a deprecation warning. To be removed in v3.0.


SylvainCorlay:
:+1:

PR #1065, opened 2017-09-04
Fix application of keep_alive policy to constructors (regression)

dean0x7d:
The regression was introduced in #1014 where the constructor's `self` argument became `value_and_holder` instead of the instance itself, so the `keep_alive` policy stoped binding to the created instance. This PR fixes it by stashing the original `self` instance in `function_call` (only for constructor calls).

This also adds a quick paragraph to the docs to clear up argument indexing of constructors.


jagerman:
I was just about to ask about creating a v2.2.1 milestone, but I see that's already done.

The fix looks good to me.

PR #1078, opened 2017-09-10
Fix a reference leak in the number converter

### LINKED ISSUE #1075: Passing a numpy.array as a vector<unsigned> (but not as vector<int> or py::array_t) leaks memory
This `numpyleak` module leaks memory when passing a `numpy.array` to the `leak` function, observe with e.g. `watch -n .2 "ps aux | sed -n '1p;/[n]umpyleak.py/p'"`. Tested on various 64bit Linux systems with python 3.5 and 3.6.

```
#include <pybind11/numpy.h>
#include <pybind11/stl.h>

namespace py = pybind11;

using leaking_uint = const std::vector<unsigned>&;
using ok_signed_int = const std::vector<int>&;

PYBIND11_MODULE(numpyleak, m) {
    m.def("leak", [](leaking_uint i) {
       return i.back() + i.front();
    });
    m.def("ok", [](const py::array_t<unsigned>& i) {
       return i.at(0) + i.at(i.size()-1);
    });
    m.def("noleak", [](ok_signed_int i) {
       return i.back() + i.front();
    });
}
```

```
#!/usr/bin/env python3

import numpy as np
from numpyleak import leak, noleak, ok

def loop(n, fn):
    # dtype.uint makes no difference
    X = np.random.randint(1000, size=40, dtype=np.int)

    i = 0
    while True:
        a = fn(X) # !
        if i % 500 == 0:
            print(i//5000, a)
        if i > n:
            break
        else:
            i += 1

loop(40100, leak)
input('memory not freed ')
loop(40100, leak)
input('done leaking ')

loop(202100, ok)
loop(202100, ok)
input('array_t done ')

loop(80200, noleak)
input('done ')
print('no double free on exit')
```

dean0x7d:
Fixes #1075.

`PyNumber_Float()` and `PyNumber_Long()` return new references.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,9 @@ v2.2.1 (Not yet released)
 * Fixed compilation with Clang on host GCC < 5 (old libstdc++ which isn't fully
   C++11 compliant). `#1062 <https://github.com/pybind/pybind11/pull/1062>`_.
 
+* Fixed a reference leak in the number converter.
+  `#1078 <https://github.com/pybind/pybind11/pull/1078>`_.
+
 * Fixed a regression where the automatic ``std::vector<bool>`` caster would
   fail to compile. The same fix also applies to any container which returns
   element proxies instead of references.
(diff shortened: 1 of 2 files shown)


wjakob:
Oops, that should definitely go into the next patch release..

PR #1079, opened 2017-09-11
docs: Describe importing Python modules and Python methods

jbarlow83:
Expand documentation to include explicit example of py::module::import where one would expect it. 

Describe how to use unbound and bound methods to class Python classes.

### DIFF
--- a/docs/advanced/pycpp/object.rst
+++ b/docs/advanced/pycpp/object.rst
@@ -33,18 +33,90 @@ The reverse direction uses the following syntax:
 
 When conversion fails, both directions throw the exception :class:`cast_error`.
 
+.. _python_libs:
+
+Accessing Python libraries from C++
+===================================
+
+It is also possible to import types and classes defined in the Python standard
+library or available in the current Python environment (``sys.path``) and work
+with these in C++.
+
+This example obtains a reference to the Python ``Decimal`` class.
+
+.. code-block:: cpp
+
+    // Equivalent to "from decimal import Decimal"
+    py::object Decimal = py::module::import("decimal").attr("Decimal");
+
+.. code-block:: cpp
+
+    // Try to import scipy
+    py::object scipy = py::module::import("scipy");
+    return scipy.attr("__version__");
+
 .. _calling_python_functions:
 
 Calling Python functions
 ========================
 
-It is also possible to call python functions via ``operator()``.
+It is also possible to call Python, classes, functions and methods 
+via ``operator()``.
+
+.. code-block:: cpp
+
+    // Construct a Python object of class Decimal
+    py::object pi = Decimal("3.14159");
+
+.. code-block:: cpp
+
+    // Use Python to make our directories
+    py::object os = py::module::import("os");
+    py::object makedirs = os.attr("makedirs");
+    makedirs("/tmp/path/to/somewhere");
+
+One can convert the result obtained from Python to a pure C++ version 
+if a ``py::class_`` or type conversion is defined.
 
 .. code-block:: cpp
 
     py::function f = <...>;
     py::object result_py = f(1234, "hello", some_instance);
     MyClass &result = result_py.cast<MyClass>();
+
+.. _calling_python_methods:
+
+Calling Python methods
+========================
+
+To call a object's method, one can again use ``.attr`` to obtain access to the
+Python method.
+
+.. code-block:: cpp
+
+    // Calculate e^π in decimal
+    py::object pi_squared = pi.attr("exp")();
+    std::cout << py::str(pi_squared);
+
+In the example above ``pi.attr("exp")`` is a *bound method*: it will always call
+for the method for that same instance of the class. (The first argument of the
+method, by convention ``self``, is bound to ``pi``, even if called for other
+arguments.)
+
+Alternately one can create an *unbound method* via the Python class (instead of
+instance) and pass the ``self`` object explicitly, followed by other arguments.
+
+.. code-block:: cpp
+
+    py::object decimal_exp = Decimal.attr("exp");
+
+    // Compute the e^n for n=0..4
+    for (int n = 0; n < 5; n++) {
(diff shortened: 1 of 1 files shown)


docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
...
     py::function f = <...>;
     py::object result_py = f(1234, "hello", some_instance);
     MyClass &result = result_py.cast<MyClass>();
 
+.. _calling_python_methods:
+
+Calling Python methods
+========================
+
+To call a object's method, one can again use ``.attr`` to obtain access to the
```
dean0x7d: **an** object's

docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
...
+Python method.
+
+.. code-block:: cpp
+
+    // Calculate e^π in decimal
+    py::object pi_squared = pi.attr("exp")();
+    std::cout << py::str(pi_squared);
```
dean0x7d: This only works with the optional `pybind11/stl.h` header. Consider using `py::print` instead.

docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
...
+
+In the example above ``pi.attr("exp")`` is a *bound method*: it will always call
+for the method for that same instance of the class. (The first argument of the
```
dean0x7d: ~~for~~ the method

docs/advanced/pycpp/object.rst
```diff
@@ -62,6 +134,9 @@ In C++, the same call can be made using:
     using namespace pybind11::literals; // to bring in the `_a` literal
     f(1234, "say"_a="hello", "to"_a=some_instance); // keyword call in C++
 
+Return value unpacking
```
dean0x7d: Return value? These are arguments.

docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
 
 When conversion fails, both directions throw the exception :class:`cast_error`.
 
+.. _python_libs:
+
+Accessing Python libraries from C++
+===================================
+
+It is also possible to import types and classes defined in the Python standard
```
dean0x7d: A type and a class are the same thing in Python. Maybe replace "types and classes" with "types and functions".

docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
...
+    py::object scipy = py::module::import("scipy");
+    return scipy.attr("__version__");
+
 .. _calling_python_functions:
 
 Calling Python functions
 ========================
 
-It is also possible to call python functions via ``operator()``.
+It is also possible to call Python, classes, functions and methods
```
dean0x7d: Extra comma between Python and classes.

docs/advanced/pycpp/object.rst
```diff
@@ -33,19 +33,91 @@ The reverse direction uses the following syntax:
...
+
+.. code-block:: cpp
+
+    // Calculate e^π in decimal
+    py::object pi_squared = pi.attr("exp")();
+    std::cout << py::str(pi_squared);
+
+In the example above ``pi.attr("exp")`` is a *bound method*: it will always call
+for the method for that same instance of the class. (The first argument of the
+method, by convention ``self``, is bound to ``pi``, even if called for other
```
dean0x7d: To me, the wording here with "even if" sounds like it could be called with a different `self` but it would be ignored. Consider rephrasing to something like "it's implicit and doesn't appear alongside other call arguments". (Although, readers should already be familiar with this from pure Python.)

docs/advanced/pycpp/object.rst
jbarlow83: I went with objects since that should cover everything.

docs/advanced/pycpp/object.rst
jbarlow83: fixed

docs/advanced/pycpp/object.rst
jbarlow83: fixed

docs/advanced/pycpp/object.rst
jbarlow83: Changed to `py::print` and changed variable name.

docs/advanced/pycpp/object.rst
jbarlow83: Removed sentence. I agree anyone using pybind11 is probably familiar with this but it's not that commonly used in Python; mentioning bound/unbound methods should be enough of a clue.

dean0x7d:
Thanks!

TD22057:
Thanks for these - they're very helpful.  It would also be nice if the examples showed how to handle errors that will get thrown from unknown modules, unknown symbols, invalid calling sequences, etc.  At a minimum a table showing which exception types get generated for every possible error would be really useful.  I'm going to use this in a production environment where I need to catch and handle any error that occurs and having it in the docs will avoid the need for me (and anyone else) to find out what's going to happen by experimentation.

jbarlow83:
There already is a table of exception mapping from C++ to Python
https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

All other Python exceptions not in that table are raised on the C++ as `std::exception`. So `ImportError` on failing to load a module will be raised as `std::exception` unless you register a specific handler. A C++ exception can be trapped in Python or C++, and a Python exception can be trapped in Python or C++.

I agree that this page should have some commentary on exceptions and will add that when I have some time.

TD22057:
Thanks - I missed this line:

> When a Python function invoked from C++ throws an exception, it is converted into a C++ exception of  type error_already_set whose string payload contains a textual summary.

on that page when I was looking the first time.

PR #1084, opened 2017-09-13
Fix casting of classes with mixed polymorphic inheritance

### LINKED ISSUE #1080: Incorrect binding when the derived class is polymorphic and the base is not
The issue from [Gitter]( https://gitter.im/pybind/Lobby?at=59b64c40319100804e1e054f)
## Reproducible example code
```c++
namespace py = pybind11;

struct BaseNonPolymorphic {
	unsigned int a;
	unsigned int b;
	unsigned int c;
	unsigned int d;
};
struct Derived : BaseNonPolymorphic {
	Derived() {
		a = 1;
		b = 2;
		c = 3;
		d = 4;
	};
	virtual ~Derived() {};
};
PYBIND11_MODULE(test_inheritance, m) {
	py::class_<BaseNonPolymorphic > base_class(m, "Base");
	base_class
		.def_readwrite("d", &BaseNonPolymorphic::d)
	;

	py::class_<Derived> derived_class(m, "Derived", base_class);
	derived_class
		.def(py::init<>())
	;
}
```

```python
import test_inheritance
derived_instance = test_inheritance.Derived()
assert derived_instance.d == 4
```

dean0x7d:
Fixes #1080.

When a polymorphic class inherits from a non-polymorphic base, casting between the two adds an offset to account for the vtable pointer. Exposing such a base and derived type to Python currently results in undefined behavior (in practice, a wrong variable value or segfault due to the pointer offset). This PR fixes this by detecting when this kind of mixed inheritance takes place and marks the types as non-simple just like multiple inheritance. Everything else is taken care of by the existing non-simple type caster path which accounts for offset bases.

One problem is that this fix requires a modification of the `internals` ABI (bumped `PYBIND11_INTERNALS_VERSION`), so it's not a candidate for v2.2.1. There are currently ways to work around this (I'll update the issue) so that could work for now and the fix could be merged for v2.3.


dean0x7d:
Fixed the MSVC linker error.

PR #1092, opened 2017-09-16
Fix segfault when reloading interpreter with external modules

jagerman:
When embedding the interpreter and loading external modules in that
embedded interpreter, the external module correctly shares its
internals_ptr with the one in the embedded interpreter.  When the
interpreter is shut down, however, only the `internals_ptr` local to
the embedded code is actually reset to nullptr: the external module
remains set.

The result is that loading an external pybind11 module, letting the
interpreter go through a finalize/initialize, then attempting to use
something in the external module fails because this external module is
still trying to use the old (destroyed) internals.  This causes
undefined behaviour (typically a segfault).

This commit fixes it by adding a level of indirection in the internals
path, converting the local internals variable to `internals **` instead
of `internals *`.  With this change, we can detect a stale internals
pointer and reload the internals pointer (either from a capsule or by
creating a new internals instance).

(No issue number: this was reported on gitter by @henryiii and @aoloe; thanks to both for the issue and reproducible test case).

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -127,21 +127,21 @@ struct type_info {
 
 /// Each module locally stores a pointer to the `internals` data. The data
 /// itself is shared among modules with the same `PYBIND11_INTERNALS_ID`.
-inline internals *&get_internals_ptr() {
-    static internals *internals_ptr = nullptr;
-    return internals_ptr;
+inline internals **&get_internals_pp() {
+    static internals **internals_pp = nullptr;
+    return internals_pp;
 }
 
 /// Return a reference to the current `internals` data
 PYBIND11_NOINLINE inline internals &get_internals() {
-    auto *&internals_ptr = get_internals_ptr();
-    if (internals_ptr)
-        return *internals_ptr;
+    auto **&internals_pp = get_internals_pp();
+    if (internals_pp && *internals_pp)
+        return **internals_pp;
 
     constexpr auto *id = PYBIND11_INTERNALS_ID;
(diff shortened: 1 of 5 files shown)


aoloe:
bot the test code and the original code that raised the issue work correctly with this pull request.

thanks!

jagerman:
Marking this for v2.2.2; this doesn't change what gets stored in the builtins, and so should be perfectly compatible with v2.2.[01] modules.

jagerman:
Cc @dean0x7d for comments.  What is here works, but maybe you have a nicer idea than the double pointer.

wjakob:
I feel like I'm missing important to understand this PR. When the interpreter shuts down, how does the internals pointer get marked as stale? I just see two leaking memory allocations (one for internals, one for the pointer to it) but no changes to shutdown/GC-related code.

jagerman:
The critical change for that is in `finalize_interpreter`: the `*internals_ptr_ptr` that gets deleted and set to `nullptr` is, with this PR, shared across all modules and the embedded interpreter with the same internals version.  Previously each module/interpret had its own pointer, so the `delete *internals_ptr_ptr` destroyed it, but the `*internals_ptr_ptr = nullptr;` was only setting the local version to `nullptr`: outside modules still had their internals pointer set.

Basically, there is now only one pointer to the internals data, where previously there were `n` pointers to it: each module/interpreter only stores a local pointer to that pointer so that when something deletes it (and sets the one-true-pointer to `nullptr`) they all see it.

wjakob:
Aha, understood now. Thank you for the clarification.

PR #1093, opened 2017-09-16
Info about inconsistent detection of Python version between pybind11 …

yesint:
Info about inconsistent detection of Python version between pybind11 and CMake in FAQ

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -241,3 +241,36 @@ Common gotchas to watch out for involve not ``free()``-ing memory region
 that that were ``malloc()``-ed in another shared library, using data
 structures with incompatible ABIs, and so on. pybind11 is very careful not
 to make these types of mistakes.
+
+Inconsistent detection of Python version in CMake and pybind11
+==============================================================
+
+The functions ``find_package(PythonInterp)`` and ``find_package(PythonLibs)`` provided by CMake 
+are not consistent with detection of Python version in pybind11. If several versions of Python are installed
+in the Linux system (i.e. 2.7 and 3.5) the pybind11 defaults to 3.5, while CMake usually defaults to 2.7.
+The reason for this design decision is unreliable and buggy behavior of ``find_package(PythonInterp)`` and ``find_package(PythonLibs)``
+which are among the worst-designed parts of CMake. We don't want to mimick this behavior in pybind11 in any way.
+
+This difference may cause inconsistencies and errors if *both* mechanisms are used in the same project. Consider the following
+Cmake code:
+
+.. code-block:: cmake
+
+    find_package(PythonInterp)
+    find_package(PythonLibs)
+    find_package(pybind11)
+
+It will detect Python 2.7 and pybind11 will peek it as well.
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -241,3 +241,36 @@ Common gotchas to watch out for involve not ``free()``-ing memory region
...
 to make these types of mistakes.
+
+Inconsistent detection of Python version in CMake and pybind11
+==============================================================
+
+The functions ``find_package(PythonInterp)`` and ``find_package(PythonLibs)`` provided by CMake
+are not consistent with detection of Python version in pybind11. If several versions of Python are installed
+in the Linux system (i.e. 2.7 and 3.5) the pybind11 defaults to 3.5, while CMake usually defaults to 2.7.
+The reason for this design decision is unreliable and buggy behavior of ``find_package(PythonInterp)`` and ``find_package(PythonLibs)``
+which are among the worst-designed parts of CMake. We don't want to mimick this behavior in pybind11 in any way.
```
jagerman: While I don't disagree with this statement, I do find the language a little strong for putting in the main documentation.  Here's a suggestion for these first two paragraphs (feel free to edit):

```
The functions ``find_package(PythonInterp)`` and ``find_package(PythonLibs)`` provided by CMake
for Python version detection are not used by pybind11 due to unreliability and limitations that make
them unsuitable for pybind11's needs.  Instead pybind provides its own, more reliable Python detection
CMake code.  Conflicts can arise, however, when using pybind11 in a project that *also* uses the CMake
Python detection.

Consider the following CMake code when run on a system with both Python 2.7 and 3.6 installed:
```

docs/faq.rst
```diff
@@ -241,3 +241,36 @@ Common gotchas to watch out for involve not ``free()``-ing memory region
...
+This difference may cause inconsistencies and errors if *both* mechanisms are used in the same project. Consider the following
+Cmake code:
+
+.. code-block:: cmake
+
+    find_package(PythonInterp)
+    find_package(PythonLibs)
+    find_package(pybind11)
+
+It will detect Python 2.7 and pybind11 will peek it as well.
```
jagerman: Did you mean "peek" -> "pick"?

docs/faq.rst
```diff
@@ -241,3 +241,36 @@ Common gotchas to watch out for involve not ``free()``-ing memory region
...
+.. code-block:: cmake
+
+    find_package(pybind11)
+    find_package(PythonInterp)
+    find_package(PythonLibs)
+
+will detect Python 3.5 for pybind11 and may crash on ``find_package(PythonLibs)`` afterwards.
+
+It is advised to avoid using ``find_package(PythonInterp)`` and ``find_package(PythonLibs)`` from CMake and rely
+on pybind11 in detecting Python version. if this is not possible CMake machinery should be called *before* including pybind11.
```
jagerman: "... and rely on pybind11's more robust Python version detection instead."

Also the first "i" in "if this is not possible..." should be capitalized.

PR #1094, opened 2017-09-17
Allow implicit conversion from PyTypeObject* to py::handle.

anntzer:
This allows constructs such as

    py::isinstance(obj, PyFoo_Type);

where PyFoo_Type is a PyTypeObject* defined by some other extension
module's header.


jagerman:
One thing that I have thought about a few times, but never gotten around to implementing, is a `py::type` class to encapsulate type objects.  There are a few internal places currently where pybind deals with `PyTypeObject *`s directly which would be a bit cleaner with an encapsulation class along the lines of `py::object`.

It would fit in between `py::object` and `py::class_<>`, and would be constructible.  With some implicit conversions and perhaps minor API additions (i.e. to accept the new type object in `isinstance`) I think you could get essentially what you're looking for, but keep the Python C API isolated from the exposed pybind11 Python C++ API.

anntzer:
That would definitely seem better (but more effort to implement too :-)).

Issue #1097, opened 2017-09-18
CMake interaction with C/C++ std flags


henryiii:
This started in the discussion for #1089. The core issue was that C++ flags mix into C compilation, because PyBind has it's own implementation of C++ flags, and it doesn't include a way to separate C code.

CMake 2.8 was released in 2009, long before C++11 (much less the other standards) were released. Modern CMake does have a way to select C++ versions, and PyBind can be in contention with them.

Method 1: CMake 3.1+: `CMAKE_CXX_STANDARD` (and `CMAKE_CXX_STANDARD_REQUIRED`, `CMAKE_CXX_EXTENSIONS`). The current implementation does include checks for these, and turns off custom flags.
Method 2: CMake 3.1+: Manually listing features with `target_compile_features`. This was supposed to be the "best" way to do things, but listing every feature in C++11 is tedious.
Method 3: CMake 3.8: Meta-compile features were added for C++ standards, finally allowing `target_compile_features(my_target PUBLIC cxx_std_11)` or similar.

The first method is checked and appropriately handled by pybind11. The final method is not well supported by pybind11. I believe that by leaving `PYBIND_CPP_STANDARD` on CMake 3.8+, PyBind should default to this method. This should allow overriding the standard (automatically handled by CMake if you request it on a pybind target), as well as keep the C++/C separate without non-portable generator expressions.

And we can avoid the awful CPP/CXX mixed name by phasing out `PYBIND_CPP_STANDARD` on new targets 😉.

PR #1098, opened 2017-09-18
Add support for CMake compiler features

henryiii:
This is a suggested implementation for #1097.

This causes pybind11 to use CMake's built-in compiler features feature if no `PYBIND11_CPP_STANDARD` is set on CMake 3.8+. This integrates with the rest of the CMake infrastructure if the user wants to use `compiler_features`. It can also be added manually now. For example:

```cmake
cmake_minimum_required(VERSION 3.8)
project(scripting VERSION 1.0 LANGUAGES CXX)

set(PYBIND11_CXX_FEATURES cxx_std_11)
add_subdirectory(pybind11)

add_executable(scripting src/main.cpp)
target_compile_features(scripting PUBLIC cxx_std_14)
set_target_properties(scripting PROPERTIES CXX_EXTENSIONS OFF)

target_link_libraries(scripting pybind11::embed)

pybind11_add_module(fooapi src/foo.cpp)
```
In this example, scripting will be build with `-std=c++14`, and fooapi will be built with `-std=c++11`. No flags will be duplicated, only one std flag will be listed, etc. Before the patch, the standard gets listed twice, would be included on C only compilation, etc.

See below for a description of how the auto-discovery works.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -100,12 +100,32 @@ regular LTO if ``-flto=thin`` is not available.
 Configuration variables
 -----------------------
 
-By default, pybind11 will compile modules with the C++14 standard, if available
-on the target compiler, falling back to C++11 if C++14 support is not
-available.  Note, however, that this default is subject to change: future
-pybind11 releases are expected to migrate to newer C++ standards as they become
-available.  To override this, the standard flag can be given explicitly in
-``PYBIND11_CPP_STANDARD``:
+By default, pybind11 will compile modules with the highest C++ standard available
+on the target compiler, from the set C++17, C++14, and C++11. With CMake 3.8+, this
+uses CMake's compile features mechinism and meta-features to detect your compiler's
+abilities. Older CMake versions fall back to manually trying to compile a mini-program
+with each flag.
+
+If you want to override this behavior, you have three options:
+
+CMake 3.1+: If you set ``CMAKE_CXX_STANDARD`` before including pybind11, it will be respected and no
+futher action will be taken by pybind11.
+
+CMake 3.1+: If you explicitly set a list of compile features in ``PYBIND11_CXX_FEATURES``, that will be used.
+See `CMAKE_CXX_KNOWN_FEATURES <https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html>`_
+for your CMake version.
+
+.. code-block:: cmake
+
+    # These meta-features were added in CMake 3.8:
+    set(PYBIND11_CXX_FEATURES cxx_std_11)
+    set(PYBIND11_CXX_FEATURES cxx_std_14)
+    set(PYBIND11_CXX_FEATURES cxx_std_17) # Experimental C++17 support
+
+    add_subdirectory(pybind11)  # or find_package(pybind11)
+
+CMake 2.8+: If neither of the above is set and you want to explicitly control the flag,
+the standard flag can be given as ``PYBIND11_CPP_STANDARD``:
 
 .. code-block:: cmake
 
@@ -120,9 +140,13 @@ available.  To override this, the standard flag can be given explicitly in
 
     add_subdirectory(pybind11)  # or find_package(pybind11)
 
-Note that this and all other configuration variables must be set **before** the
+Note that these and all other configuration variables must be set **before** the
 call to ``add_subdirectory`` or ``find_package``. The variables can also be set
 when calling CMake from the command line using the ``-D<variable>=<value>`` flag.
+
+Mixed C and C++ projects should use the ``PYBIND11_CXX_FEATURES`` mechanism to avoid
+adding flags to the C compiler. If you need the latest version of CMake, it is trivial
+to install; try ``pip install cmake`` or ``pip install --user cmake``.
 
 The target Python version can be selected by setting ``PYBIND11_PYTHON_VERSION``
 or an exact Python installation can be specified with ``PYTHON_EXECUTABLE``.
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -18,27 +18,116 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
 include(CheckCXXCompilerFlag)
 include(CMakeParseArguments)
 
-if(NOT PYBIND11_CPP_STANDARD AND NOT CMAKE_CXX_STANDARD)
-  if(NOT MSVC)
-    check_cxx_compiler_flag("-std=c++14" HAS_CPP14_FLAG)
+# The following options start out empty and cached; if empty they support "auto" behavior
 
-    if (HAS_CPP14_FLAG)
-      set(PYBIND11_CPP_STANDARD -std=c++14)
-    else()
-      check_cxx_compiler_flag("-std=c++11" HAS_CPP11_FLAG)
-      if (HAS_CPP11_FLAG)
-        set(PYBIND11_CPP_STANDARD -std=c++11)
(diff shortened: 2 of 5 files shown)


jagerman:
Is there a way to detect whether something has set the minimum version to 3.8?  Or in other words, what happens to my pybind project if I have `cmake_minimum_required(VERSION 3.0)` and `cmake` on my system gets upgraded from, say, 3.7 to 3.8?

jagerman:
> The only downside currently is that the default will be C++11 if nothing is set by the user.

That's a fairly significant blocker; the default changing from C++14 to C++11 is definitely going to break some projects that assume the default is (at least) C++14.

henryiii:
> That's a fairly significant blocker; the default changing from C++14 to C++11 is definitely going to break some projects that assume the default is (at least) C++14.

Yes, I've got an idea or two; selecting the default as "the latest the compiler supports" is not the problem, it's allowing people to build C++11 code if PyBind asks for 14. It could be (another) CMake variable, or it could look for "-or/ std=X" but there might be a better way. Still a work in progress, and open to suggestions.

henryiii:
> Is there a way to detect whether something has set the minimum version to 3.8? Or in other words, what happens to my pybind project if I have cmake_minimum_required(VERSION 3.0) and cmake on my system gets upgraded from, say, 3.7 to 3.8?

Sort of, you can check the settings of CMake policies. CMP0067 was added in 3.8, and even covers a C++ standard setting.

henryiii:
So, here are ideas:
* Add a new option, `-DPYBIND_USE_COMPILER_FEATURES`, that toggles the compiler features behavior. I'm not fond of adding extra settings, but this may be the least invasive.
* Simply change the compiler feature request to match the old auto-choice. I'll need to see if it can be downgraded by a project. If so, this might be the best option.
* Add a `pybind11::module11` target that is always C++11, and let the normal target and the `add_` command be the latest possible

Even in the current implementation, nothing is changed if you use `PYBIND_CPP_STANDARD`.

Thoughts?

henryiii:
Okay, I'm pretty happy with this now. Here's the logic, from top to bottom, new parts in bold:

* If `CMAKE_CXX_STANDARD` is set, do nothing
* **`If PYBIND11_CXX_FEATURES` is set, use those compile features (CMake 3.1+)**
* If `PYBIND11_CPP_STANDARD` is set, adds those flags
* **If CMake < 3.8:** try to auto-detect and fill in `PYBIND11_CPP_STANDARD` with the latest standard flag (11, 14, **17,** specialized MSVC support). Cached.
* **If CMake >= 3.8, automatically fill in `PYBIND11_CXX_FEATURES` with highest supported meta-feature (11, 14, 17). Faster configure time.  Cached.**

I've added a one time message if auto-discovery works, telling the user the mode and the mechanism. It won't display if you set one of these variables yourself or if you are rerunning (already in cache).

This allows pybind11 to take advantage of the faster and officially supported CMake methods for C++ standard discovery (CMake pre-populates the compiler feature list). It also improves interaction with other targets, removes double std listings, and might fix #1097, allowing CMake 3.8+ to mix C++ and C targets (please check).

To test this, you'll need to clear your CMakeCache.txt or remove the value from `PYBIND11_CPP_STANDARD`.

If it looks good, I can add a note to the docs and possibly release notes. The other option would be adding compile features for 3.1-3.8, but those would be a list of features pybind11 needs, and CMake would then pick what flags to pass; the meta-features seems simpler and seems to be the primary way CMake is adding feature support for C++17. The old way of individual features was nice because it could detect if you didn't have support, such as G++ 4.7 vs. 4.8. This method still supports someone adding a list of features to `PYBIND11_CXX_FEATURES`, so this seems to be the most general solution.

henryiii:
Added the new C++ discovery to pybind11 CONFIG, as well.

henryiii:
CMake requires a different syntax for IMPORTED targets; I've fixed the adder function to handle that (fixes config targets). Note: In CMake 3.10 or 3.11, this oddity may go away; there's work being done to fix that [here](https://gitlab.kitware.com/cmake/cmake/merge_requests/1264).

henryiii:
Adding CXX flags is now consolidated into one place, in a helper function (`_pybind11_target_cxx_std `) for consistency.

Ready for review.

docs/compiling.rst
```diff
@@ -100,12 +100,32 @@ regular LTO if ``-flto=thin`` is not available.
...
+CMake 3.1+: If you explicitly set a list of compile features in ``PYBIND11_CXX_FEATURES``, that will be used.
+See `CMAKE_CXX_KNOWN_FEATURES <https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html>`_
+for your CMake version.
+
+.. code-block:: cmake
+
+    # These meta-features were added in CMake 3.8:
+    set(PYBIND11_CXX_FEATURES cxx_std_11)
+    set(PYBIND11_CXX_FEATURES cxx_std_14)
+    set(PYBIND11_CXX_FEATURES cxx_std_17) # Experimental C++17 support
```
wjakob: The `# Experimental` comment can be removed I suppose? :)

tools/pybind11Tools.cmake
```diff
@@ -18,27 +18,116 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
...
+if(NOT CMAKE_CXX_STANDARD)
+  # While complile features were introduced in 3.1, only 3.8+ have meta-features
+  # (And C++17 seems to be mostly supported through the meta-feature)
+
+  if(CMAKE_VERSION VERSION_LESS 3.8 AND NOT PYBIND11_CXX_FEATURES)
+    if(NOT PYBIND_CPP_STANDARD)
+      # Only try to get the standard manually if CMake doesn't support compiler features
+      if(MSVC)
+        set(PYBIND11_CPP_STANDARD "/std:c++14")
+        message(STATUS "pybind11 selected C++14 flag, MSVC")
```
wjakob: Can you change all the status messages to: `pybind11: using C++14`, etc.?

tools/pybind11Tools.cmake
```diff
@@ -18,27 +18,116 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
...
+    endif()
+  endif()
+endif()
+
+# Allow the user to override by setting the cache value after the auto-discovery
+# Empty values get filled on first run (PYBIND11_CPP_STANDARD for CMake < 3.8,
+# otherwise PYBIND11_CXX_FEATURES), and then at this point are promoted to the CACHE.
+
+set(PYBIND11_CPP_STANDARD ${PYBIND11_CPP_STANDARD} CACHE STRING
+  "C++ standard flag, e.g. -std=c++11, -std=c++14, /std:c++14.  Defaults to highest supported for CMake 2.8-3.7")
```
wjakob: What happens if nothing is specified and we are on CMake 3.8? Of course it would be good if the highest standard is used even there.

tools/pybind11Tools.cmake
```diff
@@ -18,27 +18,116 @@ find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
...
+
+if(NOT CMAKE_VERSION VERSION_LESS 3.1)
+  #Only provide the option if CMake >= 3.1
+  set(PYBIND11_CXX_FEATURES ${PYBIND11_CXX_FEATURES} CACHE STRING
+    "List of compile features for PyBind, will use highest detected C++ version in CMake 3.8+")
```
wjakob: PyBidn -> pybind11

wjakob:
Also, can you clarify the interaction of this PR with #1089?

PR #1101, opened 2017-09-18
feat: Slice allowing None with std::optional

### LINKED ISSUE #1095: Can't pass None to pybind11::slice
## Issue description

The definition of `pybind11:slice` does not allow one to create a slice object where one of start, stop, step are None, since the only constructor requires all three be of type ``ssize_t``.

From https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L1062:

```c++
    slice(ssize_t start_, ssize_t stop_, ssize_t step_) {
        int_ start(start_), stop(stop_), step(step_);
        m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
        if (!m_ptr) pybind11_fail("Could not allocate slice object!");
    }
```

Slicing notation places uses None whenever the index is variable, e.g. `somelist[:]` means `slice(None, None, None)`. 

Specifically one cannot create a slice that asks for all elements of a list reversed without constructing a slice that has `stop=None`.

```python
>>> l = [1,2,3,4,5]
>>> l[::-1]  # Requires start=None, stop=None, step=-1
[5,4,3,2,1]
>>> l[5:0:-1]
[5,4,3,2]  # stop=0 terminates the list *before* the 0th element
>>> l[5:-1:-1]
[] # stop=-1 terminates the list before the final element
```

Of course PySlice_New can be called but the pybind API is quite nice :).

henryiii:
This is from #1095. `py::slice` now accepts `py::none()` in the constructor for any of its arguments. Added a (slightly messy) test from the issue.

EDIT (@YannickJadoul):
Closes #1095

## Suggested changelog entry:

```rst
* Python objects and (C++17) ``std::optional`` now accepted in ``py::slice`` constructor.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -14,6 +14,10 @@
 #include <utility>
 #include <type_traits>
 
+#if defined(PYBIND11_HAS_OPTIONAL)
+#  include <optional>
+#endif
+
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 /* A few forward declarations */
@@ -1345,11 +1349,20 @@ private:
 class slice : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(slice, object, PySlice_Check)
-    slice(ssize_t start_, ssize_t stop_, ssize_t step_) {
-        int_ start(start_), stop(stop_), step(step_);
+    slice(handle start, handle stop, handle step) {
         m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
-        if (!m_ptr) pybind11_fail("Could not allocate slice object!");
+        if (!m_ptr)
+            pybind11_fail("Could not allocate slice object!");
     }
+
+#ifdef PYBIND11_HAS_OPTIONAL
+    slice(std::optional<ssize_t> start, std::optional<ssize_t> stop, std::optional<ssize_t> step)
+        : slice(indexToObject(start), indexToObject(stop), indexToObject(step)) {}
+#else
+    slice(ssize_t start_, ssize_t stop_, ssize_t step_)
+        : slice(int_(start_), int_(stop_), int_(step_)) {}
+#endif
+
     bool compute(size_t length, size_t *start, size_t *stop, size_t *step,
                  size_t *slicelength) const {
         return PySlice_GetIndicesEx((PYBIND11_SLICE_OBJECT *) m_ptr,
@@ -1363,6 +1376,12 @@ public:
           length, start,
           stop, step,
           slicelength) == 0;
+    }
+
+private:
+    template <typename T>
+    static object indexToObject(T index) {
+        return index ? object(int_(*index)) : object(none());
     }
 };
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1357,7 +1357,7 @@ public:
 
 #ifdef PYBIND11_HAS_OPTIONAL
     slice(std::optional<ssize_t> start, std::optional<ssize_t> stop, std::optional<ssize_t> step)
-        : slice(indexToObject(start), indexToObject(stop), indexToObject(step)) {}
+        : slice(index_to_object(start), index_to_object(stop), index_to_object(step)) {}
 #else
     slice(ssize_t start_, ssize_t stop_, ssize_t step_)
         : slice(int_(start_), int_(stop_), int_(step_)) {}
@@ -1380,7 +1380,7 @@ public:
 
 private:
     template <typename T>
-    static object indexToObject(T index) {
+    static object index_to_object(T index) {
         return index ? object(int_(*index)) : object(none());
     }
 };
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "detail/common.h"
 #include "pybind11.h"
 #include <set>
 #include <unordered_set>
@@ -19,28 +20,20 @@
 #include <deque>
 #include <valarray>
 
-#ifdef __has_include
-// std::optional (but including it in c++14 mode isn't allowed)
-#  if defined(PYBIND11_CPP17) && __has_include(<optional>)
-#    include <optional>
-#    define PYBIND11_HAS_OPTIONAL 1
-#  endif
-// std::experimental::optional (but not allowed in c++11 mode)
-#  if defined(PYBIND11_CPP14) && (__has_include(<experimental/optional>) && \
-                                 !__has_include(<optional>))
-#    include <experimental/optional>
-#    define PYBIND11_HAS_EXP_OPTIONAL 1
-#  endif
-// std::variant
-#  if defined(PYBIND11_CPP17) && __has_include(<variant>)
-#    include <variant>
-#    define PYBIND11_HAS_VARIANT 1
-#  endif
-#elif defined(_MSC_VER) && defined(PYBIND11_CPP17)
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable: 4127) // warning C4127: Conditional expression is constant
+#endif
+
+// See `detail/common.h` for implementation of these guards.
+#if defined(PYBIND11_HAS_OPTIONAL)
 #  include <optional>
+#elif defined(PYBIND11_HAS_EXP_OPTIONAL)
+#  include <experimental/optional>
+#endif
+
+#if defined(PYBIND11_HAS_VARIANT)
 #  include <variant>
-#  define PYBIND11_HAS_OPTIONAL 1
-#  define PYBIND11_HAS_VARIANT 1
 #endif
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -20,11 +20,6 @@
 #include <deque>
 #include <valarray>
 
-#if defined(_MSC_VER)
-#pragma warning(push)
-#pragma warning(disable: 4127) // warning C4127: Conditional expression is constant
-#endif
-
 // See `detail/common.h` for implementation of these guards.
 #if defined(PYBIND11_HAS_OPTIONAL)
 #  include <optional>
--- a/tests/test_sequences_and_iterators.cpp
+++ b/tests/test_sequences_and_iterators.cpp
@@ -15,6 +15,9 @@
 
 #include <algorithm>
 #include <utility>
+#ifdef PYBIND11_HAS_OPTIONAL
+#include <optional>
+#endif  // PYBIND11_HAS_OPTIONAL
 
 template<typename T>
 class NonZeroIterator {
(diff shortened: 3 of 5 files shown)


include/pybind11/pytypes.h
```diff
@@ -1065,11 +1065,24 @@ class weakref : public object {
 class slice : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(slice, object, PySlice_Check)
-    slice(ssize_t start_, ssize_t stop_, ssize_t step_) {
-        int_ start(start_), stop(stop_), step(step_);
-        m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
+    slice(object start_, object stop_, object step_) {
+
```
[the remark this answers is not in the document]
henryiii: I'm traveling, will get to these soon as I can!

include/pybind11/pytypes.h
```diff
@@ -1065,11 +1065,24 @@ class weakref : public object {
...
+        m_ptr = PySlice_New(start_.ptr(),
+                            stop_.ptr(),
+                            step_.ptr());
         if (!m_ptr) pybind11_fail("Could not allocate slice object!");
     }
+
+    slice(ssize_t start_, ssize_t stop_, ssize_t step_) : slice(int_(start_), int_(stop_), int_(step_)) {}
```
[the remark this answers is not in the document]
henryiii: Cast is not yet implemented at this part of PyBind11. It comes in a later header.

This is what I've tried:
```cpp
    template <typename Start, typename Stop, typename Step>
    slice(Start start_, Stop stop_, Step step_) {
         object start_ob = cast(start_), stop_ob = cast(stop_), step_ob = cast(step_);
         m_ptr = PySlice_New(start_ob.ptr(),
                              stop_ob.ptr(),
                              step_ob.ptr());
          if (!m_ptr) pybind11_fail("Could not allocate slice object!");
     }
```

Initially it picks up `object::cast()`, if you force it to `pybind11::cast`, then it is not defined. A forward declaration looks to be tricky.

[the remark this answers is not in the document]
wjakob: This could be solved by adding a simple forward declaration.

YannickJadoul:
This seems interesting to get into the next minor release, I'd say. Still interested, @henryiii? Happy to take over as well, if you prefer.

@henryiii, since this mostly a C++ API, is there a reason to allow `py::object`? Given the lack of `std::optional`, should we allow `ssize_t` or `nullptr_t` as arguments?

henryiii:
Sure, you could take it over if you'd like. The `py::object` was primarily for the the ability to take `py::none` since we don't have an optional type. In boost-histogram, I wrap and handle the indexing in Python, but I do use more than just integers in slices - functions, complex numbers, etc. are valid. Maybe being able to take arbitrary objects would be useful?

YannickJadoul:
> Maybe being able to take arbitrary objects would be useful?

Oh, yes, didn't think that far ahead. Then maybe it's just the mix between C++ and Python types that feels weird to me.

jlucier:
@YannickJadoul @henryiii I too am running into a situation where I'd like to see this functionality. Any status updates or would this be something I could help with?

YannickJadoul:
@jlucier Not a lot of updates, no; it kind of slipped my mind. I've just rebased this PR onto `master`. Any help or input is of course welcome :-)

jlucier:
@YannickJadoul I think the implementation that's been proposed here would be perfect (assuming the feedback for @wjakob can be reconciled).

Right now I'm simply trying to work around the problem, but having a bit of trouble since my `__getitem__` is overloaded. Accepting `py::object` as the parameter in order to capture what would otherwise be `py::slice` is now capturing every call to `__getitem__` from python (since everything is fundamentally `py::object` haha).

Is there a recommended workaround for now until this is released? Can I create my own "converter" for `slice` objects or something like that?

YannickJadoul:
> Right now I'm simply trying to work around the problem, but having a bit of trouble since my `__getitem__` is overloaded. Accepting `py::object` as the parameter in order to capture what would otherwise be `py::slice` is now capturing every call to `__getitem__` from python (since everything is fundamentally `py::object` haha).

You should still be able to accept a `py::slice`, though? The only thing this PR is trying to fix is the constructor from the C++-side; there should be no issue with accepting `py::slice` and just an ordinary call to `slice.compute(...)`?

YannickJadoul:
@henryiii, what about just this?

```c++
    slice(handle start, handle stop, handle step) {
        m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
        if (!m_ptr)
            pybind11_fail("Could not allocate slice object!");
    }

#ifdef PYBIND11_HAS_OPTIONAL
    slice(std::optional<ssize_t> start, std::optional<ssize_t> stop, std::optional<ssize_t> step)
        : slice(indexToObject(start), indexToObject(stop), indexToObject(step)) {}
#else
    slice(ssize_t start_, ssize_t stop_, ssize_t step_)
        : slice(int_(start_), int_(stop_), int_(step_)) {}
#endif

private:
    template <typename T>
    static object indexToObject(T index) {
        return index ? object(int_(*index)) : object();
    }
```

This has the advantage that you can create any kind of slice with Python objects, and adds the little extra that on C++17, you can use `optional` from C++? At the same time, it doesn't cause combinatorial explosion (but allows to extend the API if we deem it necessary, later).

YannickJadoul:
I've pushed this as a commit, such that I don't have to add yet another thing on my stash. This can be removed again, ofc.

jlucier:
> > Right now I'm simply trying to work around the problem, but having a bit of trouble since my `__getitem__` is overloaded. Accepting `py::object` as the parameter in order to capture what would otherwise be `py::slice` is now capturing every call to `__getitem__` from python (since everything is fundamentally `py::object` haha).
> 
> You should still be able to accept a `py::slice`, though? The only thing this PR is trying to fix is the constructor from the C++-side; there should be no issue with accepting `py::slice` and just an ordinary call to `slice.compute(...)`?

@YannickJadoul You're right, I misread this issue almost entirely. For some reason I thought the `slice(ssize_t,ssize_t,ssize_t)` constructor requiring "non-null" arguments for each parameter meant I couldn't accept a `slice` from python that didn't have each filled in. Then I went off finding solutions for a problem I didn't have haha.

Thanks for the help regardless!

tests/test_sequences_and_iterators.cpp
```diff
@@ -93,6 +117,23 @@ TEST_SUBMODULE(sequences_and_iterators, m) {
...
+        .def(py::init<>())
+        .def("__getitem__",
+                [](Thing &t, py::slice slice) -> py::list {
+                    return get_item_list(t, slice);
+                }
+        )
+        .def("reverse",
+                [](Thing &t) -> py::list {
+                    py::slice ordinary_indices(0, static_cast<ssize_t>(t.size()), 1);
+                    py::slice reversed_slice(py::none(), py::none(), py::int_(-1));
```
EricCousineau-TRI, line 130: nit Is it possible to explicitly test `optional<>` spellings, if the feature is enabled?

tests/test_sequences_and_iterators.cpp
```diff
@@ -93,6 +117,23 @@ TEST_SUBMODULE(sequences_and_iterators, m) {
...
+    py::class_<Thing>(m, "Thing")
+        .def(py::init<>())
+        .def("__getitem__",
+                [](Thing &t, py::slice slice) -> py::list {
+                    return get_item_list(t, slice);
+                }
+        )
+        .def("reverse",
+                [](Thing &t) -> py::list {
+                    py::slice ordinary_indices(0, static_cast<ssize_t>(t.size()), 1);
```
EricCousineau-TRI, line 129: nit Intent for this line is a bit unclear.

My read is that it's testing construction options of `py::slice`. If so, I'd suggest just returning `py::slice` directly, and test it via Python directly?

include/pybind11/pytypes.h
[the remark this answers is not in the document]
EricCousineau-TRI: Will push suggested change.

tests/test_sequences_and_iterators.cpp
EricCousineau-TRI: (Will include it suggested change push)

tests/test_sequences_and_iterators.cpp
EricCousineau-TRI: (Will include in suggested change push)

EricCousineau-TRI (changes requested):

tests/test_sequences_and_iterators.cpp
```diff
@@ -93,6 +117,23 @@ TEST_SUBMODULE(sequences_and_iterators, m) {
...
+                    py::slice reversed_slice(py::none(), py::none(), py::int_(-1));
+                    py::list reversed = get_item_list(t, reversed_slice);
```
EricCousineau-TRI, line 131: Needs explicit testing for `std::optional<>`. (In my suggested change, I'm either using it wrong or there's a minor bug.)

include/pybind11/pytypes.h
```diff
@@ -1171,11 +1171,22 @@ class weakref : public object {
...
-        int_ start(start_), stop(stop_), step(step_);
+
+    slice(handle start, handle stop, handle step) {
         m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
-        if (!m_ptr) pybind11_fail("Could not allocate slice object!");
+        if (!m_ptr)
+            pybind11_fail("Could not allocate slice object!");
     }
+
+#ifdef PYBIND11_HAS_OPTIONAL
```
EricCousineau-TRI, line 1181: This is too early of a usage for `PYBIND11_HAS_OPTIONAL`.
It's defined in `stl.h`, but this depends on `pytest.h` indirectly via inclusion of `pybind11.h`.

Suggested fix: Consolidate STL usages iff any STL types are to be used in core API.

EricCousineau-TRI: Done in 5837ece4bfc

EricCousineau-TRI (approved):
Gosh dernit GitHub code review :( Wanted to comment in above, but it seems to have auto-resolved things?

Anywho, I pushed some stuff in 5837ece4bfc. PTAL!

include/pybind11/stl.h
```diff
@@ -19,28 +20,20 @@
...
-#  if defined(PYBIND11_CPP17) && __has_include(<variant>)
-#    include <variant>
-#    define PYBIND11_HAS_VARIANT 1
-#  endif
-#elif defined(_MSC_VER) && defined(PYBIND11_CPP17)
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable: 4127) // warning C4127: Conditional expression is constant
+#endif
+
```
henryiii, line 27:
```suggestion
```

This probably shouldn't be here?

include/pybind11/pytypes.h
```diff
@@ -1345,11 +1349,20 @@ class weakref : public object {
 class slice : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(slice, object, PySlice_Check)
```
Skylion007, line 1351: Should any of these constructors be explicit?

henryiii: There aren't any here with only one argument? (that is, we don't set default arguments for the new constructor / modified constructor)

henryiii:
@wjakob you have a change-requested, could you clear that? This is now a much simpler & different solution from @EricCousineau-TRI.

Skylion007 (approved):

wjakob:
Looks reasonable to me. Very minor: I'd probably write `indexToObject` as `index_to_objects` as snake-case is generally used for functions throughout the codebase following Python.

include/pybind11/pytypes.h
```diff
@@ -1345,11 +1349,20 @@ class weakref : public object {
...
+    slice(handle start, handle stop, handle step) {
         m_ptr = PySlice_New(start.ptr(), stop.ptr(), step.ptr());
-        if (!m_ptr) pybind11_fail("Could not allocate slice object!");
+        if (!m_ptr)
+            pybind11_fail("Could not allocate slice object!");
     }
+
+#ifdef PYBIND11_HAS_OPTIONAL
+    slice(std::optional<ssize_t> start, std::optional<ssize_t> stop, std::optional<ssize_t> step)
+        : slice(indexToObject(start), indexToObject(stop), indexToObject(step)) {}
```
henryiii, line 1360:
```suggestion
        : slice(index_to_object(start), index_to_object(stop), index_to_object(step)) {}
```

include/pybind11/pytypes.h
```diff
@@ -1364,6 +1377,12 @@ class slice : public object {
           stop, step,
           slicelength) == 0;
     }
+
+private:
+    template <typename T>
+    static object indexToObject(T index) {
```
henryiii, line 1383:
```suggestion
    static object index_to_object(T index) {
```

PR #1103, opened 2017-09-20
correct stride in matrix example and test

aburch:
This also matches the Eigen example for the row-major case.


aburch:
I'm slightly tired, but for an array of dimensions { rows, cols }, shoudn't the stride be either { Z*cols, Z } or { Z, Z*rows } depending on the storage format, but never { Z*rows, Z } (with Z = sizeof(...))?

jagerman:
Good catch; you're quite right about the strides, and the solution looks good.  I've pushed a commit to your branch that adds a couple tests that fail due to this under current master.  (Once the CI tests on that pass, this is good to merge).

jagerman:
Merged!  Thanks for the PR.

PR #1110, opened 2017-09-22
attach const-ness to immutable data instead of returned pointer

complyue:
In obtaining pointers to readonly numpy arrays, I found this bug in semantics.

It works for me but I don't know whether it'd break anything.


complyue:
ah, sorry, I get it wrong ...

PR #1131, opened 2017-10-09
Priority overload

### LINKED ISSUE #2537: [BUG] Allow for __str__ to be usable with enums (again)
In version 2.4 the following code did the expected thing, i.e. invoked the defined `__str__` overload for the enum when `str()` was used in Python code on an instance of that enum:
```c++
enum Pet {
    Dog = 0,
    Cat
};

auto pet = py::enum_<Pet>(m, "Pet", py::arithmetic(), "some pet")
    .value("Dog", Dog)
    .value("Cat", Cat)
    .def("__str__", [](Pet p) { return p == Dog ? "Dog" : "Cat"; });
```
This does not work anymore in V2.6 (I have not checked in V2.5, though). V2.6 always returns either `"Pet.Dog"` or `"Pet.Cat"` and ignores the defined `__str__`.

henryiii:
Adds `py::prepend()` as an optional tag when defining a function, which causes an added function to be prepended to the beginning of the overload chain. This can be used to override a default constructor that was provided earlier, for example. When combined with #1122, this will allow the default string (or the old int) constructors for enums to be replaced by users.

No performance penalty, it just prepends instead of appends to the call chain. Provides a possible solution to, and docs for, so closes #2537, so targeting 2.6.0 (related to the original use case, actually :) )

Also adds `get_pointer`/`set_pointer` for `py::capsule`.

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -367,9 +367,7 @@ Attempting to bind ``Pet::set`` will cause an error since the compiler does not
 know which method the user intended to select. We can disambiguate by casting
 them to function pointers. Binding multiple functions to the same Python name
 automatically creates a chain of function overloads that will be tried in
-the order they were defined. If the ``py::prepend()`` tag is added to the
-definition, the function will be placed at the begining of the overload sequence
-instead.
+sequence.
 
 .. code-block:: cpp
 
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -367,7 +367,9 @@ Attempting to bind ``Pet::set`` will cause an error since the compiler does not
 know which method the user intended to select. We can disambiguate by casting
 them to function pointers. Binding multiple functions to the same Python name
 automatically creates a chain of function overloads that will be tried in
-sequence.
+the order they were defined. If the ``py::prepend()`` tag is added to the
+definition, the function will be placed at the begining of the overload sequence
+instead.
 
 .. code-block:: cpp
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -141,7 +141,7 @@ struct argument_record {
 struct function_record {
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
-          is_operator(false), is_method(false), has_args(false), 
+          is_operator(false), is_method(false), has_args(false),
           has_kwargs(false), has_kw_only_args(false), prepend(false) { }
 
     /// Function name
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -142,7 +142,7 @@ struct function_record {
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
           is_operator(false), is_method(false), has_args(false), 
-          has_kwargs(false), has_kw_only_args(false), is_prepended(false) { }
+          has_kwargs(false), has_kw_only_args(false), prepend(false) { }
 
     /// Function name
     char *name = nullptr; /* why no C++ strings? They generate heavier code.. */
@@ -193,7 +193,7 @@ struct function_record {
     bool has_kw_only_args : 1;
 
     /// True if this function is to be inserted at the beginning of the overload resolution chain
-    bool is_prepended : 1;
+    bool prepend : 1;
 
     /// Number of arguments (including py::args and/or py::kwargs, if present)
     std::uint16_t nargs;
@@ -486,7 +486,7 @@ struct process_attribute<module_local> : process_attribute_default<module_local>
 /// Process a 'prepend' attribute, putting this at the beginning of the overload chain
 template <>
 struct process_attribute<prepend> : process_attribute_default<prepend> {
-    static void init(const prepend &, function_record *r) { r->is_prepended = true; }
+    static void init(const prepend &, function_record *r) { r->prepend = true; }
 };
 
 /// Process an 'arithmetic' attribute for enums (does nothing here)
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -483,8 +483,9 @@ struct process_attribute<module_local> : process_attribute_default<module_local>
     static void init(const module_local &l, type_record *r) { r->module_local = l.value; }
 };
 
-/// Process an 'prepend' attribute, putting this at the top of the overload chain
-template <> struct process_attribute<prepend> : process_attribute_default<prepend> {
+/// Process a 'prepend' attribute, putting this at the beginning of the overload chain
+template <>
+struct process_attribute<prepend> : process_attribute_default<prepend> {
     static void init(const prepend &, function_record *r) { r->is_prepended = true; }
 };
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -74,6 +74,9 @@ struct module_local { const bool value; constexpr module_local(bool v = true) : 
 /// Annotation to mark enums as an arithmetic type
 struct arithmetic { };
 
+/// Mark a function for addition at the beginning of the existing overload chain instead of the end
+struct prepend { };
+
 /** \rst
     A call policy which places one or more guard variables (``Ts...``) around the function call.
 
@@ -138,8 +141,8 @@ struct argument_record {
 struct function_record {
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
-          is_operator(false), is_method(false),
-          has_args(false), has_kwargs(false), has_kw_only_args(false) { }
+          is_operator(false), is_method(false), has_args(false), 
+          has_kwargs(false), has_kw_only_args(false), is_prepended(false) { }
 
     /// Function name
     char *name = nullptr; /* why no C++ strings? They generate heavier code.. */
@@ -188,6 +191,9 @@ struct function_record {
 
     /// True once a 'py::kw_only' is encountered (any following args are keyword-only)
     bool has_kw_only_args : 1;
+
+    /// True if this function is to be inserted at the beginning of the overload resolution chain
+    bool is_prepended : 1;
 
     /// Number of arguments (including py::args and/or py::kwargs, if present)
     std::uint16_t nargs;
@@ -475,6 +481,11 @@ struct process_attribute<metaclass> : process_attribute_default<metaclass> {
 template <>
 struct process_attribute<module_local> : process_attribute_default<module_local> {
     static void init(const module_local &l, type_record *r) { r->module_local = l.value; }
+};
+
+/// Process an 'prepend' attribute, putting this at the top of the overload chain
+template <> struct process_attribute<prepend> : process_attribute_default<prepend> {
+    static void init(const prepend &, function_record *r) { r->is_prepended = true; }
 };
 
 /// Process an 'arithmetic' attribute for enums (does nothing here)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -372,10 +372,9 @@ protected:
             if (!m_ptr)
                 pybind11_fail("cpp_function::cpp_function(): Could not allocate function object");
         } else {
-            /* Append at the end of the overload chain */
+            /* Append at the beginning or end of the overload chain */
             m_ptr = rec->sibling.ptr();
             inc_ref();
-            chain_start = chain;
             if (chain->is_method != rec->is_method)
                 pybind11_fail("overloading a method with both static and instance methods is not supported; "
                     #if defined(NDEBUG)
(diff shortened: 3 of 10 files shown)


wjakob:
Although I understand the motivation, I'm not in favor of this modification, which further increases the runtime cost and complexity of an already very complex dispatch routine. I think that we ought to make this part of pybind11 shorter and faster, not the other way.

henryiii:
Does this increase the runtime? All it does is allow someone to attach a function at the beginning of the overload chain instead of the end, but the overload chain is the same length either way. It doesn't even increase the size of the struct, since there's one bit left over for alignment. So in the setup (runtime) phase, it adds one if statement.

This functionality seems useful in a more general case, where a user needs to pre-insert a call. But up to you.

wjakob:
Aha -- I misread the patch in that case. I was assuming that the priority tag was handled in the method dispatch function, but I see now that it is in the part of the code that creates Python function overloads.

I've recently heard from users who found that Boost.Python dispatches methods faster than pybind11 (!?!?), which still has me a bit concerned that we are piling on too much stuff.

wjakob:
One naming suggestion: how about calling it ``py::prepend`` or ``py::prepend_overload`` instead of ``priority_overload``, which is IMO a bit more clear in terms of what happens (preventing question s like: "what does it mean to have two priority overloads?").

henryiii:
I took the name from the original comment from @jagerman, I don't think he meant it to be final. I like `py::prepend`.

Unrelated: Any numbers or profile code for the boost Python comparison?

jagerman:
`prepend` sounds fine to me.  `priority_overload` was just a spur-of-the-moment name.

wjakob:
> Unrelated: Any numbers or profile code for the boost Python comparison?

Unfortunately we don't currently have a good way of tracking performance over time, which is a bit of a concern. It would be nice to have a CI to catch performance regressions similarly to how we catch bugs with Travis and friends. I played around with Unladen Swallow at some point but haven't had enough free clock cycles to turn it into something useful.

tests/test_methods_and_attributes.cpp
```diff
@@ -327,6 +327,18 @@ TEST_SUBMODULE(methods_and_attributes, m) {
     py::class_<MetaclassOverride>(m, "MetaclassOverride", py::metaclass((PyObject *) &PyType_Type))
         .def_property_readonly_static("readonly", [](py::object) { return 1; });
 
+    // test_overload_ordering
+    struct OverloadOrder {
+        int value;
+        OverloadOrder(int value) : value(value) {}
+    };
+    py::class_<OverloadOrder>(m, "OverloadOrder")
```
jagerman: Using a class here seems a little overkilled; it ought to be simpler to just overload an ordinary function:
```C++
m.def("overload_order", [](std::string) { return 1; });
m.def("overload_order", [](std::string) { return 2; });
m.def("overload_order", [](int) { return 3; });
m.def("overload_order", [](int) { return 4; }, py::prepend{});
```

tests/test_methods_and_attributes.py
```diff
@@ -474,3 +474,13 @@ def test_custom_caster_destruction():
 
     # Make sure we still only have the original object (from ..._no_destroy()) alive:
     assert cstats.alive() == 1
+
+
+def test_overload_order_normal():
```
jagerman: You can combine these two into `test_overload_ordering` (I assume it started that way, given the `// test_overload_ordering` comment in the `.cpp`).  The first one isn't a particularly profound test on its own; it's just testing that the normal behaviour works as expected.

include/pybind11/pybind11.h
```diff
@@ -341,9 +340,18 @@ class cpp_function : public function {
...
                     #endif
                 );
-            while (chain->next)
-                chain = chain->next;
-            chain->next = rec;
+
+            if (rec->is_prepended) {
+                // Beginning of chain
+                chain_start = rec;
+                rec->next = chain;
```
jagerman: I've pushed a commit to your branch that should get this working.  (Feel free to merge into your own commit). Description below.

The missing piece is that the current implementation is only half-updating the chain; `chain_start` is only used to generated the docstring, but the actual dispatch begins at the pointer stored inside the `rec_capsule` created by the code path when there is no existing chain:
initial definition a few lines above:
```C++
            capsule rec_capsule(rec, [](void *ptr) { // ...
            // ...

            m_ptr = PyCFunction_NewEx(rec->def, rec_capsule.ptr(), scope_module.ptr());
```
This `rec_capsule`, which points to the initial function, gets passed to the `dispatcher()` function as `self`; the dispatcher then starts the chain from that function record.

So in addition to setting `chain_start`, we're going to have to dig into that capsule and replace its pointer (but *without* destructing the capsule because destruction frees the chained function records).  This should do it:
```C++
            if (rec->is_prepended) {
                // Beginning of chain; we need to replace the capsule's current head-of-the-chain
                // pointer with this one, then make this one point to the previous head of the
                // chain.
                chain_start = rec;
                rec->next = chain;
                auto rec_capsule = reinterpret_borrow<capsule>(((PyCFunctionObject *) m_ptr)->m_self);
                rec_capsule.set_pointer(rec);
            } else {
```
which needs a new method `capsule::set_pointer` wrapping `PyCapsule_SetPointer`:

```C++
    // Replaces a capsule's pointer *without* calling the destructor on the existing one.
    void set_pointer(const void *value) {
        if (PyCapsule_SetPointer(m_ptr, const_cast<void *>(value)) != 0)
            pybind11_fail("Could not set capsule pointer");
    }
```

tests/test_methods_and_attributes.py
```diff
@@ -474,3 +474,13 @@ def test_custom_caster_destruction():
...
+    'Check to see if the normal overload order (first defined) works'
+    assert m.OverloadOrder("this").value == 1
+
+
+def test_overload_order_priority():
+    'Check to see if the prepend overload order works'
+    assert m.OverloadOrder(0).value == 4
```
jagerman: In addition to checking that the right overload was selected, this should also check that the generated docstring for the overloaded method is ordered correct (and not missing anything), and (similarly) that the exception message for invalid arguments is correct.

jagerman:
I've pushed a commit to your branch that ought to get this working (a86451df7472f89b7577256bd19ba01adb26a03b).  Basically the start of the chain was only being half-updated: it also needed to be updated inside the `capsule rec_capsule` that gets bound as the `self` argument for `dispatcher` because `dispatcher` starts from *that* pointer, not the local `chain_start` variable (which only gets used to regenerate the docstring).

henryiii:
FYI, I'll polish off the tests as soon as I get a chance, hopefully around this weekend.

henryiii:
@jagerman That was a very long weekend, but it should be ready for review finally!

tests/test_methods_and_attributes.cpp
```diff
@@ -335,6 +335,12 @@ TEST_SUBMODULE(methods_and_attributes, m) {
     py::class_<MetaclassOverride>(m, "MetaclassOverride", py::metaclass((PyObject *) &PyType_Type))
         .def_property_readonly_static("readonly", [](py::object) { return 1; });
 
+
```
jagerman: Add a comment here: `// test_overload_ordering` to reference the python test name.

henryiii:
Maybe we could also add a mention in the upgrade guide that `__str__` on enums is now defined, so you need to use `py::prepend()` to override?

docs/classes.rst
```diff
@@ -367,7 +367,9 @@ Attempting to bind ``Pet::set`` will cause an error since the compiler does not
 know which method the user intended to select. We can disambiguate by casting
 them to function pointers. Binding multiple functions to the same Python name
 automatically creates a chain of function overloads that will be tried in
-sequence.
+the order they were defined. If the ``py::prepend()`` tag is added to the
+definition, the function will be placed at the begining of the overload sequence
```
rwgk, line 371: If the definition is marked with `py::prepend()`, ...

?

rwgk (approved):
Looks great! I have just some minor naming suggestions.

YannickJadoul (approved):
Should be rather painless, I think, yes :-)

include/pybind11/attr.h
```diff
@@ -189,6 +192,9 @@ struct function_record {
     /// True once a 'py::kw_only' is encountered (any following args are keyword-only)
     bool has_kw_only_args : 1;
 
+    /// True if this function is to be inserted at the beginning of the overload resolution chain
+    bool is_prepended : 1;
```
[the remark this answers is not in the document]
YannickJadoul: We're not using "marked" or "mark" elsewhere, though, so I'm not convinced. Is `py::arithmetic` then also a "mark"? I would personally vote to not introduce extra terminology (even though you're right; `is_prepended` made me raise an eyebrow, but I think it's still reasonably consistent).

docs/classes.rst
YannickJadoul: Same remark about "marked" as above.

include/pybind11/pybind11.h
```diff
@@ -385,9 +384,22 @@ class cpp_function : public function {
...
-            chain->next = rec;
+
+            if (rec->is_prepended) {
+                // Beginning of chain; we need to replace the capsule's current head-of-the-chain
+                // pointer with this one, then make this one point to the previous head of the
+                // chain.
+                chain_start = rec;
+                rec->next = chain;
+                auto rec_capsule = reinterpret_borrow<capsule>(((PyCFunctionObject *) m_ptr)->m_self);
+                rec_capsule.set_pointer(rec);
```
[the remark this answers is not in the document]
henryiii: I feel like this is a little simpler, and providing access to `PyCapsule_SetPointer` is a small plus?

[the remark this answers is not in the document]
henryiii: (To be honest, I believe @jagerman added these details to the original PR, and I tried to do something more like your suggestion at first)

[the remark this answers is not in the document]
YannickJadoul: Fair enough as well. Maybe it's better to not change the contents of an existing struct, yes. Seems `PyCapsule_SetPointer` is reasonably cheap anyway (and not really more expensive than copying the contents):

https://github.com/python/cpython/blob/1ed54435268b285964141fb74d47ceaa33ea79ab/Objects/capsule.c#L133-L149

(though, since it can throw errors, etc, you do need to hold the GIL to call it, ofc, but we do hold the GIL already at that point)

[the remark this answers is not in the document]
YannickJadoul: Apart from that:
1. It might be a bit weird to only expose `set_pointer`, when there are so many other `PyCapsule_` functions? But they can be added when necessary. Alternatively, we don't expose that as public API and just use `PyCapsule_SetPointer` inernally?
2. Given that `PyCapsule_GetPointer` is disguised as `operator T*()`, should this then be `operator=(T*)`?

YannickJadoul (approved):

rwgk (approved):

include/pybind11/pybind11.h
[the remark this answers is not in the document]
henryiii: 1. So many? I think it's basically four pairs, is it not? Get/set pointer, name, destructor, and context? And we already provide some of that.
2. I think that would be fine, as long as someone doesn't expect the destructor to be called when setting it?

docs/classes.rst
wjakob: This text covers basic usage of pybind11 for people who learn how to create bindings for classes for the first time. This new `py::prepend` tag scratches an itch that will only arise in very rare/advanced usage of pybind11, hence I don't think this is a good place to talk about it. The most sensible place to put it would be in an (IIRC currently non-existent) overview of all of the tags that can be specified in a function binding declaration.

include/pybind11/attr.h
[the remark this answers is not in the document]
wjakob: Why not just "prepend"? The attribute is of the "do this please"-type, so `is_` seems an odd prefix.

wjakob (commented):
The implementation looks all good to me, but documentation could be improved IMO.

wjakob:
(Really sorry that this has been stuck here in 2017. The flurry of recent activity has been wonderful -- thank you for all of your help! @henryiii @YannickJadoul @bstaletic @rwgk @EricCousineau-TRI and @ax3l)

wjakob:
My vote would also go for `set_pointer` btw, it promotes clarity.

I feel less strongly about the following: For consistency, we could also have a `get_pointer`. We will need to continue to provide the operator-based getter, but it could be implemented in terms of this new `get_pointer` function.

henryiii:
FWIW, I would be in favor of adding a `.get_pointer()` too, for consistency, and I think it's a bit more discoverable (ie, Xcode would show it in a completion list while conversions don't show up).

docs/classes.rst
henryiii: I've moved it to the place in advanced where the sequence is discussed, is that better?

include/pybind11/attr.h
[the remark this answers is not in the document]
henryiii: Because all other items here were `is_` or `has_`. But I changed it to `prepend`, which matches the tag.

PR #1133, opened 2017-10-09
Upgrading to Xcode 9

henryiii:
Moving the travis build for Xcode 8.3 to Xcode 9. (8.3 is currently broken on Travis)


henryiii:
Issue affects 9.0 also. Still probably a good idea to update after Travis starts working again.

jagerman:
I think you can fix the failure by adding a `brew update` before the `brew install python$PY`, i.e.

`brew update && brew install python$PY`

jagerman:
Merged, which will hopefully get the other builds passing again.

Issue #1138, opened 2017-10-10
pybind11 wrapped method segfaults if a function returns `unique_ptr<T>` when holder is `shared_ptr<T>`


EricCousineau-TRI:
Testing out alternatives w.r.t. #1132, I was testing out the behavior of something like this:
```c++
class A {
 public:
  A(int value)
      : value_(value) {}
  int value() const { return value_; }
 private:
  int value_{};
};

unique_ptr<A> create_instance() {
  return make_unique<A>(50);
}

shared_ptr<A> check_creation(py::function py_factory) {
  shared_ptr<A> obj;
  {
    py::object py_obj = py_factory();
    obj = py::cast<shared_ptr<A>>(py_obj);
    // ERROR: Segfault here. I believe it may be due to a double-free, since in one of the implicit `py::cast<>` calls, the move'd `unique_ptr<>` is destructed while still containing a reference.
  }
  return obj;
}

...

  py::class_<A, std::shared_ptr<A>>(m, "A")
    .def(py::init<int>())
    .def("value", &A::value);

  m.def("create_instance", &create_instance);
  m.def("check_creation", &check_creation);

...

  py::exec(R"(
factory = lambda: m.create_instance()
obj = m.check_creation(factory)
print(obj.value())
)");
```
## Issue

It seems that `move_only_holder_caster` is not aggressive enough when calling `get()`, and it will double-free with a non-empty `unique_ptr<T>`:
https://github.com/pybind/pybind11/blob/86e2ad4f77442c3350f9a2476650da6bee253c52/include/pybind11/cast.h#L1460-L1463
https://github.com/pybind/pybind11/blob/86e2ad4f77442c3350f9a2476650da6bee253c52/include/pybind11/cast.h#L1369-L1372

## Potential Solution
I believe the simple solution would be to teach `holder_helper<holder_type>::get(src)` to use `release()` for `holder_type&&` when `holder_type = unique_ptr<U, ...>`.

## Workaround
The simple workaround is to explicitly cast to `shared_ptr<T>` in C++, since it can be implicitly constructed from `unique_ptr<T>`. This sidesteps `pybind11` destructing a non-empty but unused `unique_ptr<T>`:
```
shared_ptr<A> create_instance() {
  return make_unique<A>(50);
}
```

EricCousineau-TRI:
Nevermind, issue just comes from assuming that type has the same `holder_type`:
[...]

PR #1139, opened 2017-10-10
Address #1138 when mixing holder_type's upon casting an object.

EricCousineau-TRI:
This presently fails `test_move_only_holder`. Will inspect if there is a way around this - perhaps only killing the existing holder if there is no type match between holders?

This addresses #1138.


EricCousineau-TRI:
Updated to pass tests, simplified override for `holder_helper` to stay in the same class.

jagerman:
The core issue in #1138 is that we don't check the holder type at all when initializing an instance: we just assume that a function returning a holder is returning the right kind of holder.  I've just submitted #1161 to detect that and fail if it happens when attempting to register the type and/or function.

There are some other cases this PR doesn't catch that the other one does: for example returning a `std::shared_ptr<A>` that needs a `std::unique_ptr<A>`, or one that returns a `boost::shared_ptr<A>` to a type needing an `std::shared_ptr<A>`, or various other combinations.

More problematic, the change here would also break backwards-compatibility (and indeed breaks the current test suite) by requiring that any move-only holder adds a `release()` method.

Conversion between holder types--e.g. being able to convert an rvalue `unique_ptr` into a `shared_ptr` seems feasible, but I think it's going to need a different path than stealing the pointer and stashing it in the new type.  `std::shared_ptr` *has* a `std::unique_ptr &&` constructor for doing exactly this promotion, but we'd need a nice way to detect that case, and ideally, detect it at compile time rather than run-time.

EricCousineau-TRI:
Sweet! Will close this PR shortly.

Will comment on the other PR.

EricCousineau-TRI:
> the change here would also break backwards-compatibility

Philosophical question here: Can I ask what the actual intent is in delineating holder between copyable and move-only, rather than keeping it more towards shared and unique?
Can I ask if there are existing usages in external code bases, where move-only != unique and copyable != shared?

jagerman:
> Philosophical question here: Can I ask what the actual intent is in delineating holder between copyable and move-only, rather than keeping it more towards shared and unique?

It's not really a philosophical question, it's just that we have fundamentally different code paths when a holder can be copied, with the `copyable_` and `move_only_` base classes capturing the different implementations.  Essentially, we *are* already basically assuming that copyable means shared ownership and move-only implies unique ownership.

But more generally (and the thing that affects backwards compatibility) is that we allow `struct holder_helper` specializations to be used to adapt any random smart pointer type that someone has on hand, but that such a specialization isn't needed if the holder type has a `get()` method that returns the pointer.

Issue #1145, opened 2017-10-14
Support transfer of ownership of Python subclasses of pybind11 C++ classes with `shared_ptr<>`, avoid slicing


EricCousineau-TRI:
Update (2020-12-22): Simplest reproduction can be seen in #1333

---

If a Python subclass of a pybind11 C++ class, with `holder_type = shared_ptr<T>`, has a `shared_ptr<>` that exists in C++ when its Python `ref_count()` goes to zero, Python will destroy the Python extension, and leave the C++ Alias type (e.g. `PyBase`, using `PYBIND11_OVERLOAD*`) functionality bare (and actually ~aliased / truncated~ affected by [object slicing](https://en.wikipedia.org/wiki/Object_slicing)).

This could be prevented if a custom `__del__`, `PyTypeObject::tp_del`, or `PyTypeObject::tp_dealloc` is installed such that the instance could be resurrected if `holder.use_count() > 1` for the given value-holder, and the ownership is released to C++.
(`__del__` is the easiest, as Python explicitly mentions that you can increase the refcount of a destructing object and save it from death.)

This is related to #1132.

ppt-adsk:
Hello,

I am part of the Autodesk Maya team.  We are using pybind11 for an upcoming project, and are hitting this exact problem (Python subclass of a pybind11 C++ class, holder_type = shared_ptr, shared_ptr in C++ with Python ref_count of 0, resulting in an object without the Python derived functionality).

We are not familiar with the details of your proposal, but are extremely interested in a solution to the problem, as for us it currently involves kludgey Python-side referencing of objects to keep their ref_count above 0).  Thanks for trying to move this forward.

EricCousineau-TRI:
Thanks for posting your use case!

I have tried to explain how the main portion of the WIP PR would work here:
https://github.com/EricCousineau-TRI/pybind11/blob/feature/py_move_shared-wip/docs/advanced/classes.rst#virtual-inheritance-and-lifetime
Please let me know if you have any questions about that, and please do let me know if you happen to find any better solutions.

cgerum:
I also ran into this Problem. For now i think I can work around this problem. But I would by Interested in an easier solution. Your proposal seems fine with me. 

Can I be of any assistance moving this forward?

EricCousineau-TRI:
> Can I be of any assistance moving this forward?

Most certainly! If you're up to it, could you try using this branch, and provide your feedback?
[...]

Issue #1132, opened 2017-10-09
Support ownership transfer for `unique_ptr<T>` (move-only holders) from Python to C++ (and back)


EricCousineau-TRI:
I had reviewed some of the prior questions on GitHub, StackOverflow, etc, and found that accepting a `unique_ptr<T>` from Python is not presently possible (ref: [one of the GitHub issues](https://github.com/pybind/pybind11/issues/609#issuecomment-274180833)).

I think that this can be done, and have made a (rough, but not *too* rough) prototype of doing so that I believe is generally comprehensive for most of the simple edge cases.

[branch](https://github.com/pybind/pybind11/compare/stable...EricCousineau-TRI:feature/py_move) | [head commit](https://github.com/EricCousineau-TRI/pybind11/tree/c1455fa) (as of writing)

## Demo Usage

Code sampled from [this pseudo-test case](https://github.com/EricCousineau-TRI/pybind11/blob/fbb199ce2194f1479c124f2703eca23dadb7fd2e/tests/test_unique_ptr_cast.cc#L336).

```c++
class Base {
 public:
  Base(int value) : value_(value) {}
  virtual ~Base() {}
  virtual int value() const { return value_; }
 private:
  int value_{};
};

// Trampoline class.
class PyBase : public py::trampoline<Base> {
 public:
  typedef py::trampoline<Base> TBase;
  using TBase::TBase;
  int value() const override { PYBIND11_OVERLOAD(int, Base, value); }
};

// Make a terminal function.
void terminal_func(unique_ptr<Base> ptr) {
  cout << "Value: " << ptr->value() << endl;
  ptr.reset();  // This will destroy the instance.
  cout << "Destroyed in C++" << endl;
}

// Declare class.
py::class_<Base, PyBase>(m, "Base")
  .def(py::init<int>())
  .def("value", &Base::value);
m.def("terminal_func", &terminal_func);

// Extend and use it in some Python code.
py::exec(R"(
class PyExtBase(m.Base):
    def __init__(self, value):
        m.Base.__init__(self, value)
    def __del__(self):
        print("PyExtBase.__del__")
    def value(self):
        return 10 * m.Base.value(self)

m.terminal_func([m.Base(20)])
)");
```

Should print something like:
```
Value: 200
PyExtBase.__del__
Destroyed in C++
```

## Functionality Overview

[...]

PR #1146, opened 2017-10-14
WIP: Support ownership transfer between C++ and Python with `shared_ptr<T>` and `unique_ptr<T>` for pure C++ instances and single-inheritance instances

EricCousineau-TRI:
This addresses #1132 and #1145 (and accommodates #1138 / incorporates #1139).

All unittests pass on my system (Python 2.7, release + debug).

## Todo

- [ ] Minimize changes to internals: `detail::type_info` and `detail::instance`
- [ ] Implement test cases to examine ownership behavior with and without `pybind11::wrapper<Base>`
- [ ] Squash commits + rebase on `master`, make PR to `master`.

### DIFF
--- a/docs/advanced/cast/overview.rst
+++ b/docs/advanced/cast/overview.rst
@@ -119,6 +119,9 @@ as arguments and return values, refer to the section on binding :ref:`classes`.
 | ``std::string_view``,              | STL C++17 string views    | :file:`pybind11/pybind11.h`   |
 | ``std::u16string_view``, etc.      |                           |                               |
 +------------------------------------+---------------------------+-------------------------------+
+| ``std::unique_ptr<T>``,            | STL (or custom) smart     | :file:`pybind11/cast.h`       |
+| ``std::shared_ptr<T>``, etc.       | pointers.                 |                               |
++------------------------------------+---------------------------+-------------------------------+
 | ``std::pair<T1, T2>``              | Pair of two custom types  | :file:`pybind11/pybind11.h`   |
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::tuple<...>``                | Arbitrary tuple of types  | :file:`pybind11/pybind11.h`   |
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -65,10 +65,10 @@ helper class that is defined as follows:
 
 .. code-block:: cpp
 
-    class PyAnimal : public Animal {
+    class PyAnimal : public py::wrapper<Animal> {
     public:
         /* Inherit the constructors */
-        using Animal::Animal;
+        using py::wrapper<Animal>::wrapper;
 
         /* Trampoline (need one for each virtual function) */
         std::string go(int n_times) override {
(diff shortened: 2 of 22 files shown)


wjakob:
I don't think it's going to be feasible for us to even consider a commit of this scale to support moving ``std::unique_ptr`` (which should be discussed too -- I'm not really sure that this feature makes sense to me).

EricCousineau-TRI:
I understand on the commit scale. I can certainly break this into smaller components to take this out of a WIP stage. (I can also collapse the commits - those were me just prototyping small tests as I was learning some of the internals of `pybind11`.)

For moving `unique_ptr<>`, can I ask what part does not make sense?

This also includes preventing Python-subclass aliasing that would arise from passing `shared_ptr<>`s from Python to C++, which at present would fail silently. (This is my main focus, but I would still like to enable `cast`ing `unique_ptr<>`-held instances.) The solution shares an implementation with the `unique_ptr<>` approach.
Additionally, it addresses #1138 more robustly, and provides some limited mixing of compatible holder types (`unique_ptr<T>` -> `shared_ptr<T>`).

EricCousineau-TRI:
I've squashed the commits in this branch. All of the granular experimental history is in `feature/py_move_shared-wip` on my repo.

The workflow for derived-type aliasing, specifically with `shared_ptr<>`, is now present in a more simplified unittest (hacky with `ConstructorStats` workarounds, but works):
https://github.com/pybind/pybind11/pull/1146/files#diff-5dd500b0d477a5b264d862e0157b4847R31

Still on my TODO is the `unique_ptr<>` modifications captured in `pybind11`-style unittest; however, I'll defer that pending more discussion on if you think it doesn't belong. (I'd like to have it, but I understand that at present it's a tad messy.)

wjakob:
I don't think squashing everything into a single commit addresses the problem, which is really the sheer size of the diff. I'll try to take a look in the next 1-2 weeks regarding the functional aspects (sorry for the delay, I have an unrelated deadline ATM).

EricCousineau-TRI:
Sorry, I didn't mean to imply that squashing was the only problem :)
I will see if I can provide a few more concise unittests to help bubble up the features that I'm focused on; if possible, I will see if I can stage these features, possibly layering the commits in this PR.

Thanks!

jagerman:
One thing that I'm missing is a concise overview of what this feature is all about.  Can I suggest that you write up a summary of a few paragraphs aimed at someone who has some pybind experience, but not a core developer that describes the motivation, feature, and simple implementation.  (The reason I ask for a experienced-but-not-core-developer target is that this isn't just for evaluation of the PR: it would also form the basis of the `docs/*` content that would eventually need to be added before the PR could be merged).

EricCousineau-TRI:
Most certainly!
I will type something up similar to what I had in #1132, but include the case for `shared_ptr<T>`.
For development, I will also try to identify the separate features, and use that as a basis for a PR-train.

Thank you for your patience and for considering this!

EricCousineau-TRI:
I've added a quick draft of the main features of what this PR could introduce at a user-facing level. If you have time, please let me know if you have any comments!
https://github.com/EricCousineau-TRI/pybind11/blob/feature/py_move_shared-wip/docs/advanced/classes.rst

EricCousineau-TRI:
I have updated the documentation, and included details on the holder stuff for ownership transfer as well:
https://github.com/EricCousineau-TRI/pybind11/blob/feature/py_move_shared-wip/docs/advanced/smart_ptrs.rst

EricCousineau-TRI:
@ppt-adsk mentioned that they have ran into a similar issue, at least for inheritance with `shared_ptr`s.

That being said, can I ask if there is any way that I could carve up the overview docs or this PR to make it easier to review?
I can re-vamp the PR to focus solely on `shared_ptr<>`, and defer mixing holder types to the convergence of #1161.

EricCousineau-TRI:
I'm going to decree this PR as stale. I've updated the overview to indicate where to look for more up-to-date discussions.

PR #1154, opened 2017-10-21
Add .template keyword to fix VS2017 /permissive- compilation

### LINKED ISSUE #1107: Compile failure with MSVC and /permissive- (Two-phase lookup)
Pybind11 fails to compile with newer MSVC compilers, under /permissive- mode (extra standards conformance). Disabling the new two-phase lookup (with /Zc:twoPhase-) fixes the issue, but it ought to be able to compile without it. 

`#include <pybind11/pybind11.h>
using namespace pybind11;

double Mult(double a, double b) { return a*b; }

void export_Mult(pybind11::module& m) {
      m.def("mult", &Mult);
}`

produces:
`pybind11\cast.h(1416): error C2187: syntax error: ')' was unexpected here`

patrikhuber:
This fixes #1107.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1415,7 +1415,7 @@ protected:
     bool load_value(value_and_holder &&v_h) {
         if (v_h.holder_constructed()) {
             value = v_h.value_ptr();
-            holder = v_h.holder<holder_type>();
+            holder = v_h.template holder<holder_type>();
             return true;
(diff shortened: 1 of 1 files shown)


jagerman:
I'm going to close this in favour of #1155: it fixes it, but also changes the VS2017 appveyor builds to use `/permissive-` to catch this sort of thing going forward.

Issue #1121, opened 2017-09-29
alias template error with Intel 2016.0.3 compilers


loriab:
## Issue description

In trying to update some projects from 2.0.0 to 2.2.1, I've found that they consistently won't compile with the `icpc (ICC) 16.0.3 2016041` line of compilers. Fine with GCC 5.2 and Intel `icpc (ICC) 17.0.4 20170411`. An error is below:

```
In file included from /home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/attr.h(13),
                 from /home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/pybind11.h(43),
                 from /home/psilocaluser/gits/cmake_example/src/main.cpp(1):
/home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/cast.h(837): error: alias template "pybind11::detail::type_caster_base<type>::cast_op_type" may not be used in the type-id of the alias-declaration
      template <typename T> using cast_op_type = cast_op_type<T>;
                                                 ^

In file included from /home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/attr.h(13),
                 from /home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/pybind11.h(43),
                 from /home/psilocaluser/gits/cmake_example/src/main.cpp(1):
/home/psilocaluser/gits/cmake_example/pybind11/include/pybind11/cast.h(1814): warning #3373: nonstandard use of "auto" to both deduce the type from an initializer and to announce a trailing return type
      static constexpr auto args_pos = constexpr_first<argument_is_args, Args...>() - (int) sizeof...(Args),
                       ^

compilation aborted for /home/psilocaluser/gits/cmake_example/src/main.cpp (code 2)
gmake[2]: *** [CMakeFiles/cmake_example.dir/src/main.cpp.o] Error 2
```

## Reproducible example code

The error can be seen with @dean0x7d's sample cmake+pybind11 repo, https://github.com/pybind/cmake_example (admittedly pinned at 2.2.0). When forced to switch compiler families as below, it gives the error above.

```
>>> icpc --version
icpc (ICC) 16.0.3 20160415
Copyright (C) 1985-2016 Intel Corporation.  All rights reserved.
>>> mkdir build/temp.linux-x86_64-3.5
[...]

PR #1156, opened 2017-10-24
Fix for CUDA (and Intel) compilers from #1121

henryiii:
This is needed to compile with NVCC to bind CUDA code again (worked in 2.1). Should also be applied to master, I believe.


henryiii:
Let me know if I need to make a separate PR to `master`.

henryiii:
Fixed by 32ef69a

PR #1160, opened 2017-10-29
support docstrings in enum:_:value()

### LINKED ISSUE #1136: Docstring for each value of an Enum
## Issue description

I haven't found a way to specify a docstring describing the role of each value of an enum (documentation, FAQ, tests, and `pybind11.h`).

The `value` method seems like a perfect candidate: it could accept an optional docstring similarly to `def`.

## Example code

```
py::enum_<MyEnum>(m, "MyEnum",
        "This docstring is for the overall enum.",
        py::arithmetic())
        .value("EOne", EOne)  // No way to pass per-value docstring here.
        .value("ETwo", ETwo)
        .export_values();
```

wjakob:
This PR fixes #1136 by adding an optional docstring for each value of an enumeration so that they can be documented in a way that is analogous to classes, methods, properties, etc.

I'm not 100% happy happy with the change because it ends up adding even more code that is instantiated for each enumeration. A longer-term change that would be nice to do is to refactor enumeration so that there is a top-level class ``enum_base`` class (analogous to ``pybind11_object``) which contains genereric versions of all methods and thin wrappers that are instantiated for each concrete enumeration.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,9 @@ v2.3.0 (Not yet released)
   for non-MSVC compilers).
   `#934 <https://github.com/pybind/pybind11/pull/934>`_.
 
+* The ``value()``  method of ``py::enum_`` now accepts an optional docstring
+  that will be shown in the documentation of the associated enumeration.
+
 v2.2.1 (September 14, 2017)
 -----------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


jagerman:
LGTM.  I agree that enums could use an overhaul, but this seems a reasonable short-term solution.

Issue #1159, opened 2017-10-27
Build issue on linux(centos)


srinivasavaradan:
Hello, 

I downloaded pybind11 and ran the following steps as mentioned in the document - http://pybind11.readthedocs.io/en/master/basics.html

After installing the prerequisites, I do - 

mkdir build
cd build
cmake ..
make check -j 4

After successfully completing the step 'cmake ..' I get the below errors after step 'make check -j 4'-
 
make[3]: * [tests/test_cmake_build/CMakeFiles/test_subdirectory_embed] Error 1 make[2]: * [tests/test_cmake_build/CMakeFiles/test_subdirectory_embed.dir/all] Error 2 make[2]: *** Waiting for unfinished jobs....

make[1]: * [tests/CMakeFiles/check.dir/rule] Error 2 make: * [check] Error 2

I am not sure what this means or how to fix this. Can someone point me in the right direction or explain what is happening here?. 

Below is the complete log when I run 'make check -j 4'
"
[...]

PR #1162, opened 2017-10-30
Only build embedded tests with a shared libpython

jagerman:
Edit: see below.

~~CMake's find_library can return a static library in some conditions
(certainly if the shared object isn't present; I'm not sure if there
are other conditions as well), which we don't support (and fails).~~

~~This improves the libpython detection by getting the shared library name
directly from Python rather than letting CMake guess (by prepending
`lib` and appending library suffixes).~~

~~This should, at least, prevent the embedded build from failing as in
issue #1159 -- instead it should either find the proper shared library,
or else result in a compilation failure from the missing library.~~

That didn't work on macOS.  It looks like Python just doesn't export enough to find the shared library.  Pushed a new approach that essentially goes back to the previous version, but does it in two steps: first looking only for a shared library, then, if that fails, widening the search to allow a static lib.  The shared library gets exported in a separate variable, which is then used to disable the embedding tests if a shared library wasn't found.

### DIFF
--- a/tests/test_embed/CMakeLists.txt
+++ b/tests/test_embed/CMakeLists.txt
@@ -4,6 +4,12 @@ if(${PYTHON_MODULE_EXTENSION} MATCHES "pypy")
   return()
 endif()
 
+
+if(NOT PYTHON_LIBRARY_SHARED)
+  message(STATUS "Python shared library not found; skipping interpreter tests")
+  return()
+endif()
+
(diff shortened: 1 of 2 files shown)


jagerman:
Scrapping this PR.  The windows version actually does build against a static python lib, which this broke.  I think the edge case of only having a static lib available isn't common enough to worry about, and there may be cases where the static lib actually works.

PR #1163, opened 2017-11-02
Trivial typos

luzpaz:
Non-user facing. 
Found using `codespell -q 3`


jagerman:
The `publically` one is marginally debatable, but I've no problem including it with the rest.

Thanks for the PR; merged.

PR #1164, opened 2017-11-02
typo (follow up to #1163)

luzpaz:


luzpaz:
Not sure why but I see that this didn't committed the way I'm familiar it should. 
What I was trying to fix: 
`./tools/clang/cindex.py:705: paramater  ==> parameter`

jagerman:
`tools/clang` is a submodule (https://github.com/wjakob/clang-cindex-python3), and isn't strictly necessary for pybind.  It looks like that typo is already fixed in the upstream repository (http://llvm.org/svn/llvm-project/cfe/trunk/bindings/python/clang/), but the local submodule hasn't merged upstream changes in a while.

luzpaz:
Thanks, for the explanation. Sorry for the noise. Closing PR

PR #1167, opened 2017-11-03
Patch contributors section.

petrmanek:
See https://www.youtube.com/watch?v=X6oUz1v17Uo for reference.


jagerman:
ST:TNG is great, of course. But have to close the PR: Patrick Stewart has other roles, too.

PR #1171, opened 2017-11-06
__qualname__ and nested class naming fixes

### LINKED ISSUE #1166: Incorrect type names in docstrings for inner enums
## Issue description
I have an inner enum. The type name for the enum in docstrings is not correct. I'm not sure if this is a bug in docstring generation, but I can't see any documentation on how to supply the correct name.

## Reproducible example code
```
namespace py = pybind11;
struct Pet
{
  enum class Kind { Dog };
  Pet() : type(Kind::Dog) {}
  Kind type;
};
void foo(Pet::Kind) {}
PYBIND11_MODULE (hello, m)
{
  py::class_<Pet> pet(m, "Pet");
  pet.def(py::init())
    .def_readwrite("type", &Pet::type);
  py::enum_<Pet::Kind>(pet, "Kind")
    .value("Dog", Pet::Kind::Dog);
  m.def("foo", &foo);
}
```

In python
```
>>> import hello
>>> hello.foo.__doc__
'foo(arg0: hello.Kind) -> None\n'
>>> help(hello.Pet.Kind)
Help on class Kind in module hello:
class Kind(pybind11_builtins.pybind11_object)
...
 |  __eq__(...)
 |      __eq__(self: hello.Kind, arg0: hello.Kind) -> bool
...
```
I would expect to see "hello.Pet.Kind" in the docstrings not "hello.Kind"

jagerman:
A few fixes related to how we set `__qualname__` and how we show the type name in function signatures (fixes #1166).

- `__qualname__` isn't supposed to have the module name at the beginning, but we've been putting it there.  This removes it, while keeping the `Nested.Class` name chaining.

- print `__module__.__qualname__` rather than `type->tp_name`; the latter doesn't work properly for nested classes, so we would get `module.B` rather than `module.A.B` for a class `B` with parent `A`.  This also unifies the Python 3 and PyPy code.  Fixes #1166.

- This now sets a `__qualname__` attribute on the type (as would happen internally to Python 3.3+) for Python <3.3 and PyPy.  While not particularly important to have in earlier Python versions, it's useful for us to be able to extracted the nested name for function signatures (which is why `__qualname__` was invented in the first place).

- Added tests for the above.

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -14,6 +14,15 @@
 NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 NAMESPACE_BEGIN(detail)
 
+#if PY_VERSION_HEX >= 0x03030000
+#  define PYBIND11_BUILTIN_QUALNAME
+#  define PYBIND11_SET_OLDPY_QUALNAME(obj, nameobj)
+#else
+// In pre-3.3 Python, we still set __qualname__ so that we can produce reliable function type
+// signatures; in 3.3+ this macro expands to nothing:
+#  define PYBIND11_SET_OLDPY_QUALNAME(obj, nameobj) setattr((PyObject *) obj, "__qualname__", nameobj)
+#endif
+
 inline PyTypeObject *type_incref(PyTypeObject *type) {
     Py_INCREF(type);
     return type;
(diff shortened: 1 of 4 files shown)


wjakob:
Looks like there is a minor warning on MSVC (use of two separate variables named "t" in nested scopes) that was treated as an error. Other than that, it looks great!

PR #1183, opened 2017-11-14
Link to libpython with flat namespace on OSX

grubbins:
On OSX, linking against `libpython` dynamically from the compiled python module will break at runtime if the python interpreter was statically linked with `libpython`. This problem has been avoided by not linking against `libpython` at all, and then disabling unresolved symbol errors with `-undefined dynamic`.

The downside of that solution is that it suppresses all unresolved symbol errors for that link, even those that are entirely unrelated to use of `pybind11`. This can hide other build setup issues and lead to less obvious runtime errors.

The reason things break on OSX is the two-level namespace: the symbols statically linked into the interpreter and the symbols obtained from `libpython` dynamically are distinct (unlike on Linux). If we disable the two-level namespace with `-flat_namespace`, the problem goes away and we can link to `libpython` and retain unresolved symbol errors.

This change was tested on Sierra with `make check` and seems to work fine; I confirmed that the new linker argument is being applied, and also that if I remove `libpython` from the link then I get unresolved symbol errors as expected.

I am using `pybind11` on another project in which we use our own CMake setup and link against Anaconda python packages, which (as of recently) statically link `libpython` - so I encountered the duplicate symbol problem, and found the `-flat_namespace` solution works perfectly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ if(NOT (CMAKE_VERSION VERSION_LESS 3.0))  # CMake >= 3.0
   if(WIN32 OR CYGWIN)
     target_link_libraries(module INTERFACE $<BUILD_INTERFACE:${PYTHON_LIBRARIES}>)
   elseif(APPLE)
-    target_link_libraries(module INTERFACE "-undefined dynamic_lookup")
+    target_link_libraries(module INTERFACE "-flat_namespace" ${PYTHON_LIBRARIES})
   endif()
 
   add_library(embed INTERFACE)


wjakob:
One of the main reasons for the current approach is that it allows redistributing libraries that were linked against one Python library (e.g. OSX) and then use them with another library (e.g. Brew, or an interpreter that is built into a piece of software like Blender or Autodesk Maya). In this case, we don't want OSX's ld to even look at the other Python library. (Blender will e.g. crash when ld tries to load a second Python library). Since your change reintroduces linking against a specific Python library, I wonder how your approach fares in this situation.

grubbins:
Well, we have to resolve the same set of symbols so if they are not present in the runtime environment then it won't work however you cut it. But you're right that this change does embed the file path of the originally used library and so we get into the question of the search process for that file.

I also had a change of heart regarding the relative cost of flat namespace... this uncovered a different issue elsewhere in my project where two versions of hdf5 (previously independent thanks to the two-level namespace) got incompatibly conflated.

So I think I might retract the whole PR :/

grubbins:
Another way forward might be to explicitly mark all the affected symbols as undefined, but unless we can do that with some kind of linker script, it would be very ugly.

wjakob:
The discussion doesn't make a good case for changing the current behavior, hence I'll close this very old PR.

PR #1184, opened 2017-11-15
Clarify docs for functions taking bytes and not str.

anntzer:
I don't think the point mentioned in the commit is explicitly mentioned anywhere; my reading of the current docs originally suggested (wrongly) to me that there was no way to declare a function as accepting bytes and not str.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ MANIFEST
 *.py[co]
 *.egg-info
 *~
+.*.swp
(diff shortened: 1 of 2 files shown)


wjakob:
I think this comment only applies to Python 3.x (on Python 2.x there should be no difference).

anntzer:
edited accordingly

anntzer:
kindly bumping

PR #1186, opened 2017-11-16
Matching Python 2 int behavior on Python 2

henryiii:
This is from a Gitter discussion. Pybind11's default conversion to `int` always produces a `long` on Python 2 (`int`s and `long`s were unified in Python 3). This patch fixes `int` handling to match Python 2 on Python 2; for short types (`size_t` or smaller), the number will be returned as an `int` if possible, otherwise `long`. Requires Python 2.5+.

An example of where this matters: The `sys.exit` function must have an int. So if you have this:

```cpp
.def("__int__", &Thing::operator int)
```

You currently need this to work;

```python
sys.exit(int(int(thing)))
```

My Python 2 numpy is having issues, so can't test locally yet.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -981,18 +981,11 @@ public:
         if (std::is_floating_point<T>::value) {
             return PyFloat_FromDouble((double) src);
         } else if (sizeof(T) <= sizeof(ssize_t)) {
-#if PY_VERSION_HEX < 0x03000000
             // This returns a long automatically if needed
             if (std::is_signed<T>::value)
-                return PyInt_FromSsize_t((ssize_t) src);
+                return PYBIND11_LONG_FROM_SIGNED(src);
             else
-                return PyInt_FromSize_t((size_t) src);
-#else
-            if (std::is_signed<T>::value)
-                return PyLong_FromSsize_t((ssize_t) src);
-            else
-                return PyLong_FromSize_t((size_t) src);
-#endif
+                return PYBIND11_LONG_FROM_UNSIGNED(src);
         } else {
             if (std::is_signed<T>::value)
                 return PyLong_FromLongLong((long long) src);
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -980,11 +980,19 @@ public:
     static handle cast(T src, return_value_policy /* policy */, handle /* parent */) {
         if (std::is_floating_point<T>::value) {
             return PyFloat_FromDouble((double) src);
-        } else if (sizeof(T) <= sizeof(long)) {
+        } else if (sizeof(T) <= sizeof(ssize_t)) {
+#if PY_VERSION_HEX < 0x03000000
+            // This returns a long automatically if needed
             if (std::is_signed<T>::value)
-                return PyLong_FromLong((long) src);
+                return PyInt_FromSsize_t((ssize_t) src);
             else
-                return PyLong_FromUnsignedLong((unsigned long) src);
+                return PyInt_FromSize_t((size_t) src);
+#else
+            if (std::is_signed<T>::value)
+                return PyLong_FromSsize_t((ssize_t) src);
+            else
+                return PyLong_FromSize_t((size_t) src);
+#endif
         } else {
             if (std::is_signed<T>::value)
                 return PyLong_FromLongLong((long long) src);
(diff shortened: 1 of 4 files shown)


henryiii:
I believe this works now, but Clang 5.0 is broken for everyone right now. Someone will need to restart it tomorrow.

tests/test_builtin_casters.py
```diff
@@ -323,3 +323,12 @@ def test_numpy_bool():
     assert convert(np.bool_(False)) is False
     assert noconvert(np.bool_(True)) is True
     assert noconvert(np.bool_(False)) is False
+
+
+def test_int_long():
+    """In Python 2, a C++ int should return a Python int if possible, not long."""
```
jagerman: Perhaps elaborate a bit on why:

> In Python 2, a C++ int should return a Python int rather than long if possible:  int's can always be used where a long is required, but longs are not always accepted where ints are used (such as the argument to sys.exit()).

include/pybind11/cast.h
```diff
@@ -980,11 +980,19 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
     static handle cast(T src, return_value_policy /* policy */, handle /* parent */) {
         if (std::is_floating_point<T>::value) {
             return PyFloat_FromDouble((double) src);
-        } else if (sizeof(T) <= sizeof(long)) {
+        } else if (sizeof(T) <= sizeof(ssize_t)) {
+#if PY_VERSION_HEX < 0x03000000
```
jagerman: Instead of the `#ifdef` here, let's add these to `detail/common.h` like we have for various other python API calls that differ across python versions:
```C++
#define PYBIND11_LONG_FROM_SIGNED(o) PyLong_FromSize_t((size_t) o)
#define PYBIND11_LONG_FROM_UNSIGNED(o) PyLong_FromSsize_t((ssize_t) o)
```
for Python 3 and
```C++
#define PYBIND11_LONG_FROM_SIGNED(o) PyInt_FromSize_t((size_t) o)     // Returns long if needed.
#define PYBIND11_LONG_FROM_UNSIGNED(o) PyInt_FromSsize_t((ssize_t) o) // Likewise.
```
for Python 2; then this part of the `cast.h` code becomes much cleaner:
```C++
            if (std::is_signed<T>::value)
                return PYBIND11_LONG_FROM_SIGNED(o);
            else
                PYBIND11_LONG_FROM_UNSIGNED(o);
```

henryiii: Okay, I was considering something like that. I'll add when the CI build finishes.

tests/test_builtin_casters.py
henryiii: It is possible (though extremely rare) that a long could be required; you could just check the type, for example. But now you can force a long in C++ by casting to long long.

henryiii: Updated.

include/pybind11/cast.h
henryiii: Done.

henryiii:
Repushed, now that Clang 5 is working again.

wjakob:
This looks great.

tests/test_builtin_casters.py
```diff
@@ -323,3 +323,16 @@ def test_numpy_bool():
...
+
+def test_int_long():
+    """In Python 2, a C++ int should return a Python int rather than long
+    if possible: longs are not always accepted where ints are used (such
+    as the argument to sys.exit()). A C++ long long is always a Python
+    long."""
+
+    import sys
+    must_be_long = type(getattr(sys, 'maxint', 1) + 1)
+    isinstance(m.int_cast(), int)
```
jagerman: Shouldn't these be `assert isinstance(...)`?

jagerman:
Aside from the missing `assert`s in the test, this looks good.

tests/test_builtin_casters.py
henryiii: Oops!

jagerman:
Looks like `apt.llvm.org` is back to being broken, but more importantly, the added tests don't appear to be passing under Python 2.

henryiii:
The problem is `long long == size_t`. Which is nice, that means that everything is covered on most systems though just that one Python API call. The current method of a backup call is still best just in case `size_t` is 32 bits, such as on some 32 bit systems.

Python ints are signed, so the max `unsigned long long` should trigger a python `long` return.

jagerman:
The i386 build should hit it; I think there `long long` is 64 bits but `size_t` is 32.

henryiii:
Can this be merged? I think it is finished from my end.

jagerman:
Yeah, thanks for the reminder (I'm pretty swamped with other things at the moment).

PR #1191, opened 2017-11-20
Fix warning in C++17 mode.

bluescarni:
In C++17, static constexpr members are implicitly inline and it's deprecated to re-declare them outside the class. This fixes a GCC 7.2 ``-Wdeprecated`` warning in C++17 mode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -193,9 +193,9 @@ install:
         ${PYPY:+--extra-index-url https://imaginary.ca/trusty-pypi}
     echo "done."
 
-    wget -q -O eigen.tar.gz https://bitbucket.org/eigen/eigen/get/3.3.3.tar.gz
-    tar xzf eigen.tar.gz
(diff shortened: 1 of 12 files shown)


wjakob:
Thank you -- this looks good to me.

jagerman:
It looks good to me.  I've also pushed another commit to this branch that adds `-Wdeprecated` to the test suite to see if it catches anything else—assuming that passes the CI builds, this should be good to merge.

jagerman:
Aha, that caught some more.  Another commit pushed to address them.

bluescarni:
I got a few more methinks, but I cannot reproduce the eigen3 error on my setup.

jagerman:
Perhaps it was fixed in 3.3.4; I've pushed a change to update it to see if that fixes it.  (Edit: forced pushed: the first attempt was broken).

jagerman:
Okay, they are all solved now.  Upgrading to Eigen 3.3.4 didn't change anything, so I just disabled `-Wdeprecated` in `eigen.h`.

jagerman:
This is now merged into `master` (outside of github to squash the commits properly) in ba33b2fc798418c8c9dfe801c5b9023d3703f417.

PR #1196, opened 2017-11-26
Support keyword-only arguments

### LINKED ISSUE #1194: Indicating that a function takes keyword-only arguments
## Issue description

I was told on gitter that pybind11 currently does not support creating functions that takes keyword-only arguments.  It would be appreciated if it did.

As far as I can tell, support for keyword-only arguments is directly implemented in PyArg_ParseTupleAndKeywords by the "$" specifier (https://docs.python.org/3/c-api/arg.html#other-objects) (which requires keyword-only arguments to have a default; I guess it's fine to have whatever limitations the Python C-API currently has).

On pybind11's side, I guess this can be provided as something like `py::kwonly("foo")` / `"foo"_kwo` where one would have expected `py::arg("foo")` / `"foo"_a`; or with some marker object separating kwonly arguments from non-kwonly ones, similar to `*` on Py3.

## Reproducible example code

N/A

jagerman:
This adds support for a `py::args_kw_only()` annotation that can be specified between `py::arg` annotations to indicate that any following arguments are keyword-only.  This allows you to write:

```C++
    m.def("f", [](int a, int b) { /* ... */ },
          py::arg("a"), py::args_kw_only(), py::arg("b"));
```

and have it work like:

```Python
    def f(a, *, b):
        # ...
```

in Python 3 with respect to have the `a` and `b` arguments are accepted (that is, `a` can be positional or by keyword; `b` can only be specified by keyword).

Fixes #1194

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -362,6 +362,34 @@ like so:
     py::class_<MyClass>("MyClass")
         .def("myFunction", py::arg("arg") = (SomeType *) nullptr);
 
+Keyword-only arguments
+======================
+
+Python 3 introduced keyword-only arguments by specifying an unnamed ``*``
+argument in a function definition:
+
+.. code-block:: python
+
+    def f(a, *, b):  # a can be positional or via keyword; b must be via keyword
+        pass
+
+    f(a=1, b=2)  # good
+    f(b=2, a=1)  # good
+    f(1, b=2)    # good
+    f(1, 2)      # TypeError: f() takes 1 positional argument but 2 were given
+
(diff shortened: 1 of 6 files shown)


henryiii:
This is great! I have looked for this several times. What does it do in Python 2?

MichaelGoulding:
Any update on this?

TheButlah:
What is the status of this PR?

skoslowski:
I am also interested in this feature. Gave it a try, still works great! Pushed a conflict-free version to #2100. I am happy to help resolve any remaining issues and get this merged

henryiii:
#2100 has been merged, so this should be available now! Slightly renamed (`py::kwonly`).

PR #1198, opened 2017-11-27
Add spaces around "=" in signature repr.

anntzer:
PEP8 indicates (correctly, IMO) that when an annotation is present, the
signature should include spaces around the equal sign, i.e.

    def f(x: int = 1): ...

instead of

    def f(x: int=1): ...

(in the latter case the equal appears to bind to the type, not to the
argument).

pybind11 signatures always includes a type annotation so we can always
add the spaces.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -238,7 +238,7 @@ protected:
             } else if (c == '}') {
                 // Write default value if available.
                 if (arg_index < rec->args.size() && rec->args[arg_index].descr) {
-                    signature += "=";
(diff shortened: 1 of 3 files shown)


jagerman:
This sounds good to me.  PEP8 aside, I also agree that it looks better in most cases.

jagerman:
> pybind11 signatures always includes a type annotation so we can always add the spaces.

It seems to me that for some cases (e.g. taking a `py::object` or `py::handle` argument) when we'd be better off not adding the type annotation.  But that's *not* a concern for this PR.

Issue #1208, opened 2017-12-07
Thread kill while running pybind11 module


kreuzberger:
## Issue description

When running a python script in c++ thread and the thread is terminated, application core dumps.

This case is not considered in the catch clauses and could be easily handled 
by adding 
`catch( abi::__forced_unwind& )`
in the catch clauses

kreuzberger:
Who is responsible for closing this issue? seems to be fixed after commit.

PR #1209, opened 2017-12-07
#1208 Bugfix thread kill while running pybind11 module

kreuzberger:
Adding special unwind catch clause provided with gcc std abi


wjakob:
LGTM.

Issue #1207, opened 2017-12-07
Separate optional and variant support from other containers `stl.h`


funchal:
I'd like to propose that we:

- move `std::optional` and `std::variant` conversion support, which are currently in `pybind11/stl.h`, to a new file (say `pybind11/utility.h` since they are in the STL [utility](http://en.cppreference.com/w/cpp/utility) library in C++17).
- include `pybind11/utility.h` from `pybind11/stl.h` so that we maintain full backward compatibility.
- update the docs [Overview](http://pybind11.readthedocs.io/en/stable/advanced/cast/overview.html) page to mention the new header.

My motivation for this proposal is that the conversion for other containers (set/vector/map) in `pybind11/stl.h` comes with a major downside, documented in http://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#automatic-conversion, and I believe that optional/variant can be useful in their own right.

If you like the idea, I'm happy to prepare a pull request. Please let me know. #

wjakob:
Hi @funchal ,

This sounds reasonable to me. Do you want to create a PR? (please also mention this in the documentation and the changelog file).

Best,
Wenzel

funchal:
Here you go: https://github.com/pybind/pybind11/pull/1210

funchal:
Quick reminder, there is still interest in resolving this issue. `stl.h` has too much stuff in it some of it I want and some I don't. I currently have to workaround this would be great to get some solution in place.

PR #1210, opened 2017-12-08
Separate optional and variant support from other STL containers #1207

funchal:
Move `std::optional` and `std::variant` conversion support, which were previously in `pybind11/stl.h`, to a new header file `pybind11/utility.h` since they are in the STL utility library in C++17. Include `pybind11/utility.h` from `pybind11/stl.h` so that we maintain full backward compatibility. Update the docs, tests and changelog to mention the new header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ set(PYBIND11_HEADERS
   include/pybind11/pytypes.h
   include/pybind11/stl.h
   include/pybind11/stl_bind.h
+  include/pybind11/utility.h
 )
 string(REPLACE "include/" "${CMAKE_CURRENT_SOURCE_DIR}/include/"
        PYBIND11_HEADERS "${PYBIND11_HEADERS}")
--- a/docs/advanced/cast/overview.rst
+++ b/docs/advanced/cast/overview.rst
@@ -143,11 +143,11 @@ as arguments and return values, refer to the section on binding :ref:`classes`.
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::unordered_set<T>``          | STL unordered set         | :file:`pybind11/stl.h`        |
 +------------------------------------+---------------------------+-------------------------------+
-| ``std::optional<T>``               | STL optional type (C++17) | :file:`pybind11/stl.h`        |
+| ``std::optional<T>``               | STL optional type (C++17) | :file:`pybind11/utility.h`    |
 +------------------------------------+---------------------------+-------------------------------+
-| ``std::experimental::optional<T>`` | STL optional type (exp.)  | :file:`pybind11/stl.h`        |
+| ``std::experimental::optional<T>`` | STL optional type (exp.)  | :file:`pybind11/utility.h`    |
 +------------------------------------+---------------------------+-------------------------------+
-| ``std::variant<...>``              | Type-safe union (C++17)   | :file:`pybind11/stl.h`        |
+| ``std::variant<...>``              | Type-safe union (C++17)   | :file:`pybind11/utility.h`    |
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::function<...>``             | STL polymorphic function  | :file:`pybind11/functional.h` |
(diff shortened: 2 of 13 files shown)


jagerman:
This rename sounds good to me in principle, but I have one concern: `utility.h` is a pretty generic name; it would be nice to attach the notion of this being stl `<utility>` support as opposed to some sort of pybind11 utilities.

Thoughts (to you and @wjakob and anyone else who wants to offer an opinion on the matter) on going a bit bigger on this reorganization by putting this into `pybind11/stl/utility.h`, moving the container support into something like `stl/containers.h`, and the container binding from `stl_bind.h` into `pybind11/stl/bind.h`.  `pybind11/stl.h` could then be a meta-header that just includes all the various `pybind11/stl/*.h` feature headers, and `pybind11/stl_bind.h` would become a deprecated include for `pybind11/stl/bind.h`.

We already added the `detail` directory in 2.2, and the plan in #708 (and related in #864) will almost certainly reorganize by introducing other directories.

funchal:
@jagerman Thanks for having a look at this. I thought about your comment for a bit, and here's my opinion for what it's worth:
- There's a reason why `pybind11/stl.h` isn't included automatically by `pybind11.h`, as it introduces possibly unwanted (and perhaps even arguably confusing) container conversion behavior. In some sense, `pybind11/stl_bind.h` is an alternative to  `pybind11/stl.h` that uses api binds instead of conversion. We should be careful about keeping this distinction clear, if/when we go ahead with the header re-organisation you proposed. Perhaps `bind/stl.h` and `cast/stl.h`?
- I might be wrong, but I feel like `std::variant` and `std::optional` are in a bit of a different situation. Similarly to `std::tuple`/`std::pair` etc, their behavior appears to be totally fine imho, and perhaps they should just be included automatically by `pybind11.h`, like `std::tuple`/`std::pair`.
- I agree with the reorganization but there are diminishing returns. In particular, I believe that reorganizing api headers (those that enable different combinations of features) is way more important than reorganizing `detail` internals. It also seems more important than breaking down headers into various individual feature headers `pybind11/stl/*.h` etc as you also suggested, since I find it unlikely that these would be individually included (most users would just use the meta-header).

jagerman:
> There's a reason why `pybind11/stl.h` isn't included automatically by `pybind11.h`, as it introduces possibly unwanted (and perhaps even arguably confusing) container conversion behavior.

The main reason is, I believe, just to keep the compilation a bit faster but not pulling in extra headers needed for stl containers in cases where they aren't needed.  Including doesn't exactly change the behaviour: without it, trying to return a stl container just results in a failure because pybind won't know how the cast the type to Python (because it'll get to the generic type caster, but `std::vector<Foo>` is not a registered type).  It's not clear to me that the non-`stl.h` behaviour is ever particular useful.

> I might be wrong, but I feel like `std::variant` and `std::optional` are in a bit of a different situation. 

I agree with you that `variant` and `optional` are essentially equivalent in their level of purpose to `tuple` and `pair`.  Those, however, are in `cast.h` sort of by accident: the function argument dispatcher used to use a `std::tuple`, with [special methods](https://github.com/pybind/pybind11/blob/v1.8/include/pybind11/cast.h#L647) in the tuple caster specific to loading argument lists.  That's no longer the case (as of 2.0, IIRC)—argument loading now goes via the `detail::argument_loader` class which is entirely separate.  So if `std::tuple` support were being added *today*, it might well be in `stl.h` (and `std::pair` goes along with `std::tuple`--as of 2.2 they are handled by the same code).

> I agree with the reorganization but there are diminishing returns. In particular, I believe that reorganizing api headers (those that enable different combinations of features) is way more important than reorganizing detail internals.

Agreed.  That's basically why we moved things into `detail/` for 2.2, but not other subdirectories.  Partly whether this happens depends on whether we do the caster overhaul for 2.3, which will benefit from it.

YannickJadoul:
Adding this to finally be resolved in 2.7.0 :-) Maybe through a new, non-conflicting PR, though.

PR #1213, opened 2017-12-14
Update PyPI URLs.

QuLogic:
As it says, URLs are pointing to a now-outdated fork.


jagerman:
Thanks for the PR, merged.

PR #611, opened 2017-01-22
Add support for positional args with args/kwargs

jagerman:
This commit rewrites the function dispatcher code to support mixing regular arguments with `py::args`/`py::kwargs` arguments.  It also simplifies the argument loader noticeably as it no longer has to worry about args/kwargs: all of that is now sorted out in the dispatcher, which now simply appends a tuple/dict if the function takes `py::args`/`py::kwargs`, then passes all the arguments in a single tuple.

Some (intentional) restrictions:
- you may not bind a function that has args/kwargs somewhere other than the end.  This somewhat matches Python, but mainly keeps the dispatch code a little cleaner by being able to not worry about where to inject the `py::args`/`py::kwarg`s in the argument list.  Fully matching what Python allows (i.e. named, but not positional, arguments after `*args`) seems not worth bothering with.
- If you specify an argument both positionally and via a keyword argument, you get a TypeError (matching what happens in Python).

PR #1216, opened 2017-12-15
Fix leak in var arg handling

zdevito:
When using the mixed position + vararg path, pybind over inc_ref's
the vararg positions. Printing the ref_count() of `item` before
and after this change you see:

Before change:

```
refcount of item before assign 3
refcount of item after assign 5
```

After change
```
refcount of item before assign 3
refcount of item after assign 4
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -581,7 +581,7 @@ protected:
                         extra_args = tuple(args_size);
                         for (size_t i = 0; i < args_size; ++i) {
                             handle item = PyTuple_GET_ITEM(args_in, args_copied + i);
-                            extra_args[i] = item.inc_ref().ptr();
+                            extra_args[i] = reinterpret_borrow<object>(item);
                         }
                     }
                     call.args.push_back(extra_args);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -580,8 +580,7 @@ protected:
                         size_t args_size = n_args_in - args_copied;
                         extra_args = tuple(args_size);
                         for (size_t i = 0; i < args_size; ++i) {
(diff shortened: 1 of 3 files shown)


jagerman:
Can you add a test case to the PR that triggers the warning (without the fix) and passes (with the fix).  It looks from your description like you already have something along those lines.

include/pybind11/pybind11.h
```diff
@@ -581,7 +581,7 @@ class cpp_function : public function {
                         extra_args = tuple(args_size);
                         for (size_t i = 0; i < args_size; ++i) {
                             handle item = PyTuple_GET_ITEM(args_in, args_copied + i);
-                            extra_args[i] = item.inc_ref().ptr();
+                            extra_args[i] = reinterpret_borrow<object>(item);
```
jagerman: The `reinterpret_borrow` and the temporary variable aren't actually needed; the two lines can be simplified to just:
```C++
extra_args[i] = PyTuple_GET_ITEM(args_in, args_copied + i);
```

jagerman:
Bug confirmed (this was introduced in the dispatcher overhaul in #611).  See my suggested simplification, add a simple test case (so that it can't reoccur), and I'll merge.

jagerman:
Actually, you can ignore my suggestions; I've written some test code, and I'll push it (and the minor code simplification) to this PR soon.

zdevito:
Sounds good, thanks for writing the test code!

jagerman:
Pushed the simplification and test cases.  (I also ended up rebasing your branch against current master to incorporate the MSVC fix I just pushed there).  Assuming it all builds successfully I'll go ahead with the merge soon.

jagerman:
Pushed a PyPy test fix; hopefully that does it.

jagerman:
Merged!  Thanks for the PR.

PR #1220, opened 2017-12-22
Require cmake version 3 in pybind11Config.cmake

joakim-hove:
It seems that the `pybind11Config.cmake` file has a hard requirement on cmake version greater than 3.0; this PR will just make that explicit.

If there is interest in merging this PR I would be happy to update the PR with documentation fixes, raising the minimum cmake version from 2.8.12 to 3.0.


henryiii:
Fixed in the CMake refactor. :)

PR #1223, opened 2017-12-22
Segfault fix in cpp_function::dispatcher of an overloaded function with *args/**kwargs

YannickJadoul:
Fixing a segfault in the cpp_function::dispatcher of an overloaded function with args/kwargs, during the second pass over the overloads (with conversions).

Inside the overload loop, new `py::object`s are created for the *args and **kwargs arguments (`extra_args` and `kwargs`), but these go out of scope by the time a second pass with conversions over the different overloads is executed.
At the destruction of the `argument_loader`, its nested casters, and the `py::args` object, if Python already garbage collected, things go wrong.

A quick, 3-line solution is to keep a `std::vector<py::object>` and add the `extra_args` and `kwargs` instances to keep them alive for long enough.

Attached is a trace of the segfault in GDB (running `python3-gdb`; `python3` just gets corrupted instead of segfaulting and shouts `Fatal Python error: GC object already tracked` somewhere later):
[args-crash-gdb-bt.txt](https://github.com/pybind/pybind11/files/1583437/args-crash-gdb-bt.txt)

I don't know if there might be a better/cleaner/... solution than this 'keepalive' vector, but this at least seems to be solving the segfaults/corruption and does not change the flow of the dispatcher.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -452,6 +452,7 @@ protected:
             // py::arg().noconvert()).  This lets us prefer calls without conversion, with
             // conversion as a fallback.
             std::vector<function_call> second_pass;
+            std::vector<object> args_kwargs_keepalive;
 
             // However, if there are no overloads, we can just skip the no-convert pass entirely
             const bool overloaded = it != nullptr && it->next != nullptr;
@@ -586,6 +587,7 @@ protected:
                     }
                     call.args.push_back(extra_args);
                     call.args_convert.push_back(false);
+                    args_kwargs_keepalive.push_back(extra_args);
                 }
 
                 // 4b. If we have a py::kwargs, pass on any remaining kwargs
@@ -594,6 +596,7 @@ protected:
                         kwargs = dict(); // If we didn't get one, send an empty one
                     call.args.push_back(kwargs);
                     call.args_convert.push_back(false);
+                    args_kwargs_keepalive.push_back(kwargs);
                 }
 
                 // 5. Put everything in a vector.  Not technically step 5, we've been building it


jagerman:
I can't look into this in detail right now, but my initial thought (which may well be wrong) is that it seems as though they should be kept alive by being inside the `call.args` vector.  If adding them to a second vector is fixing this, is it perhaps just masking somewhere that we missed an `inc_ref` (or equivalently that we `stole` somewhere where we should have `borrowed`)?

YannickJadoul:
As far as I saw, [`call.args`](https://github.com/pybind/pybind11/blob/master/include/pybind11/cast.h#L1799) is a `std::vector<handle>`, and won't do any reference counting?

jagerman:
Ah yes, I see.  (I forgot: we wanted a handle vector there to avoid a bunch of unnecessary inc_ref/dec_refs in here).

jagerman:
I agree that we need another keep-alive reference here, but I don't really like the idea of introducing another vector here—we'd be introducing extra mallocs into the function dispatcher (which is performance-critical code).  I think putting the extra keepalive `object`s as two new members in the `function_call` should work; it makes `function_call` slightly larger, but that way we're just allocating slightly larger `function_call` objects rather than adding new allocations.

Try out 7c306b1 -- I think that should solve it.

YannickJadoul:
Ok, agreed, that's a fine solution too. And the error in the specific case I had is also solved by it (as I just tested), so ... thanks!

jagerman:
(I'm going to keep this open until I merge that commit into master just to keep it on my radar).

jagerman:
Merged the fix into master (48e1f9aadc6ed715742e41a3cd3693a234a96246).

YannickJadoul:
Thanks for the quick fix!

PR #1225, opened 2017-12-26
Make register_dtype() accept any field containers

aldanor:
Using `initializer_list` was kind of dumb to start with, as it can only be empty-constructed, thus disallowing any sort of programmatic munging. This patch makes it not care about the field descriptors container type.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -18,9 +18,9 @@
 #include <cstring>
 #include <sstream>
 #include <string>
-#include <initializer_list>
 #include <functional>
 #include <utility>
+#include <vector>
 #include <typeindex>
 
 #if defined(_MSC_VER)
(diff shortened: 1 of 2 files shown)


jagerman:
Is this going to break existing code using an initializer list currently?  (IIRC, initializer lists can't be deduced to a template type).  Would a `detail::any_container<field_descriptor>` work here?

aldanor:
Yep, I totally forgot about `any_container<>` -- fixed it (also added a little test confirming that it works both ways now).

aldanor:
(Looks all green.)

include/pybind11/numpy.h
```diff
@@ -1083,7 +1083,7 @@ inline PYBIND11_NOINLINE void register_structured_dtype(
     // - https://github.com/numpy/numpy/pull/7798
     // Because of this, we won't use numpy's logic to generate buffer format
     // strings and will just do it ourselves.
-    std::vector<field_descriptor> ordered_fields(fields);
+    std::vector<field_descriptor> ordered_fields(*fields);
```
jagerman: You should be able to use just `std::move(fields)` here instead of `*fields` (to avoid a vector copy).

aldanor: Hmm I guess it was actually always copying them but yea -- `fields` is not used past this point so we could just move them; fixed.

aldanor:
Travis build apparently failed, but it's not test-related.

PR #1226, opened 2017-12-26
WIP Add arg().disown() (superseding the consume call_policy)

torokati44:
I finally found some time to work a little bit on #971. This PR supersedes it. _(Should I close it?)_ I reworked it, taking some of @jagerman's advice. However, it is still not nearly complete.

- [x] It is now done with `py::arg`, instead of as a call policy.
- [x] The holder is now destructed, and the reference is set to be not owning.
- [x] Releasing is done via a `holder_helper` static method.
- [ ] Moving `owned` to `value_and_holder` is not yet done.
- [ ] The default `holder_helper::release` does not give a meaningful error when there is no `holder_type::release()`
- [ ] `holder_type` is still assumed to be `std::unique_ptr<void*>`

For the last one: I have no idea how to get the proper `holder_type` for the args. I tried to look in the direction of `class_<arg>::holder_type`, but that would have to be forward declared, and I'm no even sure what to supply as the `type` template parameter.

Is this in the right place in `dispatcher()`?
Also, I saw that #1146 referenced #971, is this PR still relevant despite that one?

Any comments are still much appreciated.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -613,6 +613,39 @@ protected:
                     call.args_convert.swap(second_pass_convert);
                 }
 
+
+                //printf("calling %s %s, args %d %lu\n", func.name, func.signature, func.nargs, func.args.size());
+                for (size_t i = 0; i < func.args.size(); ++i) {
+                    if (!func.args[i].disown) {
+                        //printf("NOT disowning arg %lu\n", i);
+                        continue;
+                    } else {
+                        //printf("DISOWNING arg %lu\n", i);
+                    }
+
+                    auto &arg = call.args[i];
+                    if (!arg)
+                        pybind11_fail("nothing to disown?!");
+
+                    if (arg.is_none())
+                        continue; /* Nothing to disown */
+
+                    auto inst = reinterpret_cast<detail::instance *>(arg.ptr());
+                    auto value_and_holder = values_and_holders(inst).begin();
+
+
+                    using holder_type = std::unique_ptr<void*>; // how do I get the proper one?
+
+                    auto &holder = value_and_holder->holder< holder_type  >();
+
+
+                    holder_helper<holder_type>::release(holder);
+
+                    value_and_holder->set_holder_constructed(false);
+                    inst->owned = false;
+                    holder.~holder_type();
+                }
+
                 // 6. Call the function.
                 try {
                     loader_life_support guard{};
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -124,10 +124,11 @@ struct argument_record {
     const char *descr; ///< Human-readable version of the argument value
     handle value;      ///< Associated Python object
     bool convert : 1;  ///< True if the argument is allowed to convert when loading
+    bool disown : 1;   ///< TODO docs
     bool none : 1;     ///< True if None is allowed when loading
 
-    argument_record(const char *name, const char *descr, handle value, bool convert, bool none)
-        : name(name), descr(descr), value(value), convert(convert), none(none) { }
+    argument_record(const char *name, const char *descr, handle value, bool convert, bool disown, bool none)
+        : name(name), descr(descr), value(value), convert(convert), disown(disown), none(none) { }
 };
 
 /// Internal data structure which holds metadata about a bound function (signature, overloads, etc.)
@@ -353,8 +354,8 @@ template <> struct process_attribute<is_new_style_constructor> : process_attribu
 template <> struct process_attribute<arg> : process_attribute_default<arg> {
     static void init(const arg &a, function_record *r) {
         if (r->is_method && r->args.empty())
-            r->args.emplace_back("self", nullptr, handle(), true /*convert*/, false /*none not allowed*/);
-        r->args.emplace_back(a.name, nullptr, handle(), !a.flag_noconvert, a.flag_none);
+            r->args.emplace_back("self", nullptr, handle(), true /*convert*/, false /*disown*/, false /*none not allowed*/);
+        r->args.emplace_back(a.name, nullptr, handle(), !a.flag_noconvert, a.flag_disown, a.flag_none);
     }
 };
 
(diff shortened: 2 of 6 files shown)


torokati44:
FWIW, even in its current wonky form, this has proven to be "working for me". At least for now.

EricCousineau-TRI:
> Also, I saw that #1146 referenced #971, is this PR still relevant despite that one?

I'd say yes, this PR is still quite relevant (and I'm actually still looking to have it land in some form, as it in it's own state has been useful to us :) ). It looks like you're providing a much more generic mechanism to transfer ownership, which is awesome!

> For the last one: I have no idea how to get the proper holder_type for the args.

In my WIP branch from #1146, I type-erase mechanisms that deal with holders, store those in `detail::type_info` (unfortunately needed, but meh), and then call them when needed:
https://github.com/RobotLocomotion/pybind11/blob/drake/include/pybind11/pybind11.h#L1277

I believe that ultimately, you'd need this sort of mechanism to properly mix holder types (which my implementation misses, since I don't store `typeid`, but @jagerman has in his #1161).

And naiive question: How does this pass compilation with `shared_ptr` and all that (things that don't have a `release`)? Is it because you've implicitly recast the `unique_ptr<void*>`?

include/pybind11/pybind11.h
```diff
@@ -613,6 +613,39 @@ class cpp_function : public function {
...
+                        pybind11_fail("nothing to disown?!");
+
+                    if (arg.is_none())
+                        continue; /* Nothing to disown */
+
+                    auto inst = reinterpret_cast<detail::instance *>(arg.ptr());
+                    auto value_and_holder = values_and_holders(inst).begin();
+
+
+                    using holder_type = std::unique_ptr<void*>; // how do I get the proper one?
```
EricCousineau-TRI: @torokati44 After seeing your stuff, and reworking my PR, I believe this function should be what you need for type-erased holder transfer:
https://github.com/EricCousineau-TRI/pybind11/blob/379afa73c7fedded57e4fedc9c3b077cb7292643/include/pybind11/cast.h#L482

You could use it to something like the tune of:
```
if (!reclaim_existing_if_needed(inst, tinfo, existing_holder)) {
  throw cast_error("Cannot release ownership of this object");  // If you have an incompatible holder type, like `shared_ptr`.
}
```

The caveat is, you'll need to figure out what the holder type is for the given argument. I believe that should be elsewhere in this area of code.
(Also, ensure that you get `tinfo` from the argument type and NOT the holder, as `holder->type` may be null!)

torokati44:
I was actually asking if my PRs were relevant, seeing that yours' are progressing. :)
Thank you for your tips, I'll take a look at them once I have the time/capacity to work on this again.

EricCousineau-TRI:
> I was actually asking if my PRs were relevant, seeing that yours' are progressing. :)

Hup, sorry about that! (Was in technical mode)
Ideally, I think the ownership transfer should based on the holder: If you want to take an object from `pybind` and then pass it to something else in C++ (e.g. something that takes `unique_ptr`, or something that takes a raw pointer), then it'd be best to have a `unique_ptr<>` as an argument, rather than annotate the argument (which I believe would end up yielding the same functionality).

That being said, this is based on my usage of it, which looks like this:
https://github.com/RobotLocomotion/drake/blob/5dbb882/bindings/pydrake/systems/framework_py.cc#L186

Can I ask if you have a code snippet or open-source project that you could point to, using your code?

torokati44:
> Can I ask if you have a code snippet or open-source project that you could point to, using your code?

I'm trying to create Python bindings for the OMNeT++ API. (I don't think I can share it just yet.)

It has ownership transfer of objects from the simulation kernel and the user code, in both directions, passed as raw pointer function arguments.
See for example [handleMessage](https://omnetpp.org/doc/omnetpp/api/classomnetpp_1_1cSimpleModule.html#a59e577d050358fd9f737678394505aca) and [send](https://omnetpp.org/doc/omnetpp/api/classomnetpp_1_1cSimpleModule.html#a5584b421899bb435d1ef83def8c61453).

If `cSimpleModule` is to be subclassed in Python, this would mean ownership transfer between C++ and Python, and vice versa. In one direction (C++ -> Python) it works well already, and I'm trying to make the other direction (for `send()`) work with this PR.

As the API above is pretty much set in stone, I can't just switch to `unique_ptr`. Well, I could put together a wrapper of some sorts, but I'd rather not, if it can be avoided, as that wouldn't be very elegant.

Ownership transfer (or the lack thereof) via return value (again, as raw pointers) also only works one way (AFAIK; C++ -> Python), but one issue at a time... :)

EricCousineau-TRI:
Ah, I see. In that case, it sounds like you'd be interested @jagerman's proposal for adding a `load_options`-like option to `py::arg()`:
https://github.com/pybind/pybind11/issues/864#issuecomment-313972586
https://github.com/pybind/pybind11/issues/1241#issuecomment-356699163

torokati44:
I have been able to achieve (almost) what I wanted with current vanilla `pybind11`; by using a custom holder that works similar to `std::shared_ptr`, except `.release()` acts on all holder instances that refer to the same value: https://gist.github.com/torokati44/adc99fd7279449c9000803b128cc36c2

I'm sure it violates a bunch of programming principles, and the names could be better, etc., but it gets the job done (somehow). At least until a better solution (either a call policy, an `arg()` option, or whatever) is found to take ownership away from Python.

The only inconvenience is that wherever this kind of ownership transfer happens with a function argument or return value, I have to bind a tiny wrapper lambda instead of the real function, that calls `holder.release()` before/after the actual function call. (And obviously instances of subclasses defined in Python can't be kept alive after Python loses a reference to them.)

_I had silently hoped that someone with more experience with this kind of template-heavy code will finish this instead of me, or provide a better alternative, but so far this hasn't happened. Oh well._

I don't think I'll ever finish this, rather just go ahead with my custom holder. Thanks for the help anyway, @EricCousineau-TRI!

PR #1229, opened 2017-12-29
fix return from std::map bindings to __delitem__

### LINKED ISSUE #1228: cast error for __delitem__ bindings to map and unordered_map
## Issue description
Cannot delete items from STL bindings for `std::map` and `std::unordered_map`

Bindings to `__delitem__` return an iterator which pybind11 doesn’t know how to cast. Fortunately `__delitem__` does not need to return anything.

vanossj:
tests and fixes #1228

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -587,7 +587,7 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
                auto it = m.find(k);
                if (it == m.end())
                    throw key_error();
-               return m.erase(it);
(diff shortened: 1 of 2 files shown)


jagerman:
The fix looks good to me.  Was map deletion *completely* broken in all cases?  I'm surprised that we really didn't have any map `del` tests--thanks very much for fixing that test suite oversight (along with the core issue).

vanossj:
I didn't try a ton of cases, but I didn't find any that worked either.

jagerman:
Merged, thanks for the PR!

Issue #11, opened 2015-11-16
Passing both python functions and std::function as arguments interchangeably


Vayu:
I noticed that including `pybind11/functional.h` has some (probably) unintended side-effects.

If I include `pybind11/functional.h` I cannot pass `std::function` instances which I get from C++ code back to C++ functions. On the other hand without `pybind11/functional.h`, I cannot pass Python methods to C++ functions expecting `std::function`.

Is it possible to get both working, or are they mutually exclusive? I'm attaching a short example

```C++
#include <pybind11/pybind11.h>
#include <pybind11/functional.h>  // comment/uncomment this

struct Functor {
    bool operator() (int a, int b) const {
      return a < b;
    }
    std::function<bool(int, int)> cast() {
      return std::function<bool(int, int)>(*this);
    }
};

struct Caller {
    bool call(std::function<bool(int, int)> func, int a, int b) {
      return func(a, b);
    }
};

namespace py = pybind11;

PYBIND11_PLUGIN(tmp) {
    py::module m("tmp", "pybind11 tmp plugin");

    py::class_<Caller>(m, "Caller").def(py::init<>())
                                   .def("call", &Caller::call);

    py::class_<std::function<bool(int, int)>>(m, "FuncType").def(py::init<>());

    py::class_<Functor>(m, "Functor").def(py::init<>())
                                     .def("__call__", [](const Functor& f, int a, int b) { return f(a, b); })
                                     .def("cast", &Functor::cast);
    return m.ptr();
}
```

```Python
import tmp

cppfunc = tmp.Functor()
caller = tmp.Caller()

def pyfunc(a, b):
    return a < b

try:
    print 'Calling C++ functor {0!r}'.format(cppfunc),
    res = cppfunc(1, 2)
    print 'OK'
except Exception as error:
    print 'FAILED with error {1!r}'.format(cppfunc, error)

try:
    print 'Calling C++ with Python function {0!r}'.format(pyfunc),
    caller.call(pyfunc, 1, 2)
    print 'OK'
except Exception as error:
    print 'FAILED with error {1!r}'.format(cppfunc, error)

stdfunc = cppfunc.cast()
try:
    print 'Calling C++ with C++ function {0!r}'.format(stdfunc),
    caller.call(stdfunc, 1, 2)
    print 'OK'
except Exception as error:
    print 'FAILED with error {1!r}'.format(cppfunc, error)
```

Without `pybind11/functional.h`
```
[...]

PR #1237, opened 2018-01-06
Feature/unique ptr arg pr1

EricCousineau-TRI:
This is the first PR in order to enable passing `unique_ptr` back and forth between C++.
This is a more manageable bite-size chunk of #1146 (supporting #11

This relates #971, #1226.

I'm moving forward with this because we have found this feature useful in Drake, and have a potential mechanism to even allow this to be REPL-friendly. (This change does require minimal changes in existing code, which is preferable to switching to `shared_ptr` and never being able to release a pointer from its clutches.)
https://github.com/RobotLocomotion/drake/blob/f6f23c5/bindings/pydrake/systems/framework_py.cc#L148
https://github.com/EricCousineau-TRI/pybind11/blob/25b08a4/tests/test_smart_ptr.py#L319

### DIFF
--- a/docs/advanced/smart_ptrs.rst
+++ b/docs/advanced/smart_ptrs.rst
@@ -1,5 +1,19 @@
-Smart pointers
-##############
+.. _holders:
+
+Smart pointers and holders
+##########################
+
+Holders
+=======
+
+The binding generator for classes, :class:`class_`, can be passed a template
+type that denotes a special *holder* type that is used to manage references to
+the object.  If no such holder type template argument is given, the default for
+a type named ``Type`` is ``std::unique_ptr<Type>``, which means that the object
+is deallocated when Python's reference count goes to zero. It is possible to
+switch to other types of reference counting wrappers or smart pointers, which
+is useful in codebases that rely on them, such as ``std::shared_ptr<Type>``, or
+even a custom type.
 
 std::unique_ptr
 ===============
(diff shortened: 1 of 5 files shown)


EricCousineau-TRI:
After the first portion of this PR, I realized that it wouldn't be too much more to enable `unique_ptr` as arguments.

Please note that I have curated each of the commits to make sure that each commit makes sense by itself.

This also includes a commit to fix #1238, because I encountered this issue while developing this branch.

EricCousineau-TRI:
Hmm... CI failures seem possibly unrelated to this PR?

On Travis CI, there are errors where the assertion errors differ by a space.
On AppVeyor, there is a warning about using `if constexpr` (which I'd really prefer not to, as it's C++17), that I guess is then escalated to an error? (I can patch this with a temporary, but just wondering if this warning should be suppressed?)

May I ask if there are any ideas on the Travis CI NumPy stuff?

EricCousineau-TRI:
@jagerman Can I ask if you've had a chance to take a peek at this PR? Might you have any suggestions at a design level that might make this fit into a future `pybind` release?

rwgk:
> This would be awesome, any chance to get this addressed ?

https://github.com/pybind/pybind11/blob/smart_holder/README_smart_holder.rst

The branch exists for ~3 years now (and is used in Google production code since then). The last noteworthy change was in the summer of 2021. I'm keeping it up-to-date with master, and it is fully compatible with master (see the README).

gentlegiantJGC:
Is this resolved now that the smart holder branch has been merged?

rwgk:
> Is this resolved now that the smart holder branch has been merged?

Yes, I think so, based on just glancing through. Let's close this, since you asked about it.

Issue #1231, opened 2017-12-30
Unmodified C++ object after calling a python callback


benEnsta:
## Issue description

Hi,

I have a C++ code which calls a python function. This function modifies an object passed as its argument. When this object has been instantiated from python, modifications done by the python function are available to the c++ side.
However, an object, directly instantiated from the C++ side, is not modified after the function call.

The following example illustrates this behavior. With the function example_NOK, the object o1 is unmodified after the python function call. This is not the case with example_OK.

Is there a way to fix this issue and have my C++ object modified ?
Thanks for your help.

## Reproducible example code

```c++
#include <pybind11/pybind11.h>
#include <pybind11/functional.h>
#include <string>

namespace py = pybind11;

// custom type
class MyType {
public:
  MyType(int val) :val(val){}
  int val;
};

void example_OK(std::function<void(MyType&)> f, MyType& o){
  py::print("[example_OK] initial o = ", o);
  f(o);
  py::print("[example_OK] final o = ", o);
}

void example_NOK(std::function<void(MyType&)> f){
  MyType o1(10);
  py::print("[example_NOK] initial o = ", o1);
  f(o1);
  py::print("[example_NOK] final o = ", o1);
}

PYBIND11_MODULE(cmake_example, m) {
    py::class_<MyType>(m, "MyType")
      .def(py::init<int>())
      .def("__str__", [](MyType& o){return std::to_string(o.val);})
      .def_readwrite("val", &MyType::val)
    ;

    m.def("example_NOK", &example_NOK);
    m.def("example_OK", &example_OK);
}
```
python code:

```python
import cmake_example as m

def f(o):
    o.val = -3

t = m.MyType(10)
m.example_OK(f, t)
m.example_NOK(f)
```

Ìt produces the following output
```bash
[example_OK] initial o =  10
[example_OK] final o =  -3
[example_NOK] initial o =  10
[example_NOK] final o =  10
```
In both case, we expected that  the final value of the object was -3.

EricCousineau-TRI:
Do you know if the code works if you change your function signature to `std::function<void (MyType*)>`, and pass it a pointer?

I may have ran into a similar issue where `pybind` was copying the object rather than referencing it (but got a compile-time error because the type was non-copyable).
Here is where I encountered this, and the shim I put in place:
https://github.com/RobotLocomotion/drake/blob/0215e565aa38aaa9271d7757174a1c5ec61bcd6c/bindings/pydrake/systems/framework_py.cc#L102

[...]

Issue #1200, opened 2017-11-30
Pass a reference to C++ container to the Python callback


uentity:
Hi!

I faced this issue while implementing a Python's `os.walk`-like function for my tree structure. I hoped to mimic the ability of callback to edit passed list of directories in-place and guide further search.

Suppose we have the following in C++:
```C++
#include <pybind11/stl_bind.h>

PYBIND11_MAKE_OPAQUE(std::vector<double>);
namespace py = pybind11;
...

py::bind_vector<std::vector<double>>(m, "d_list", py::module_local(false));
// this will work as expected because I can explicitly point that return value is a reference
m.def("test_static_d", []() -> std::vector<double>& {
	static std::vector<double> v{0, 0};
	return v;
}, py::return_value_policy::reference);

// callback will fail to edit passed vector, because a copy is involved somewhere?
m.def("test_callback", [](std::function<void(std::vector<double>&)> f) {
	std::vector<double> v{0, 0};
	f(v);
	for(const auto& l : v) {
		std::cout << l << ' ';
	}
	std::cout << std::endl;
});
```

And now the Python code:
```Python
In [2]: a = example.test_static_d()

In [3]: a.append(1)

In [4]: a
Out[4]: d_list[0, 0, 1]

In [5]: b = example.test_static_d()

In [6]: b
Out[6]: d_list[0, 0, 1]
# so far so good
In [7]: b is a
Out[7]: True

# callback seems to modify a copy of vector
def f(v) :
    v.append(1.)
    print(v)

# C++ function don't get updated vector after callback invoke
In [1]: example.test_callback(f)
d_list[0, 0, 1]
0 0 
```

Is there a way to specify that callback should take an opaque container by reference (in order to modify it in-place on Python side)?

EricCousineau-TRI:
Possibly related to #1231: This looks like it might also be a symptom of argument reference policies - it is most like likely passing a copy of your vector to your function.

Can you try making your signature into `std::function<void (std::vector<double>*)>`?

jagerman:
The main issue here in how return-value-policies propagate (or rather, *don't* propagate) through `std::function` bindings.  You can see this in `functional.h`: the `value = [func](...)` lambda simply calls the `py::function` argument with its arguments, which ends up doing ordinary argument casting or arguments--which defaults to copying.  Realistic, what we want in a case like this is to be able to call (cutting away error handling etc. for exposition):

```C++
py::function f = reinterpret_borrow<py::function>(src);
[...]

PR #1240, opened 2018-01-07
 Use `return_value_policy::reference` for mutable lvalues (#1200).

EricCousineau-TRI:
This addresses #1200 (and #1231) by ensuring that argument casting is more C++-ish: Do not default to copying for mutable lvalue references.

The problem in the above issues is that in casting a Python function to a C++ function and calling it directly on a C++ object, there's a possibility that the C++ object will have not been seen by `pybind`. Because of this, the prior policy was to copy the previously unseen object.

This PR *only* changes function argument casting to alleviate this. It keeps the behavior of `cast<>()` the same, for reasons stated in `test_methods_and_attributes.py` unittest.

\cc @uentity @benEnsta

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -92,15 +92,17 @@ The following table provides an overview of available policies:
 +--------------------------------------------------+----------------------------------------------------------------------------+
 | :enum:`return_value_policy::automatic`           | **Default policy.** This policy falls back to the policy                   |
 |                                                  | :enum:`return_value_policy::take_ownership` when the return value is a     |
-|                                                  | pointer. Otherwise, it uses :enum:`return_value_policy::move` or           |
-|                                                  | :enum:`return_value_policy::copy` for rvalue and lvalue references,        |
-|                                                  | respectively. See above for a description of what all of these different   |
-|                                                  | policies do.                                                               |
+|                                                  | pointer or :enum:`return_value_policy:reference` when the return value is  |
+|                                                  | a mutable lvalue reference. Otherwise, it uses                             |
+|                                                  | :enum:`return_value_policy::move` or :enum:`return_value_policy::copy` for |
+|                                                  | rvalue and const lvalue references, respectively. See above for a          |
+|                                                  | description of what all of these different policies do.                    |
 +--------------------------------------------------+----------------------------------------------------------------------------+
 | :enum:`return_value_policy::automatic_reference` | As above, but use policy :enum:`return_value_policy::reference` when the   |
-|                                                  | return value is a pointer. This is the default conversion policy for       |
-|                                                  | function arguments when calling Python functions manually from C++ code    |
-|                                                  | (i.e. via handle::operator()). You probably won't need to use this.        |
+|                                                  | return value is a pointer or mutable lvalue reference. This is the default |
+|                                                  | conversion policy for function arguments when calling Python functions     |
+|                                                  | manually from C++ code (i.e. via handle::operator()). You probably won't   |
+|                                                  | need to use this.                                                          |
 +--------------------------------------------------+----------------------------------------------------------------------------+
 
 Return value policies can also be applied to properties:
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -789,6 +789,12 @@ public:
         return cast(&src, policy, parent);
     }
 
+    static handle cast(itype &src, return_value_policy policy, handle parent) {
+        if (policy == return_value_policy::automatic || policy == return_value_policy::automatic_reference)
+            policy = return_value_policy::reference;
+        return cast(&src, policy, parent);
+    }
+
     static handle cast(itype &&src, return_value_policy, handle parent) {
         return cast(&src, return_value_policy::move, parent);
     }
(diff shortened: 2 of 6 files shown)


jagerman:
This seems like a backwards-compatibility-breaking change, which makes it a non-starter for 2.x.

Regarding #1200, I'll comment in that issue.

EricCousineau-TRI:
> This seems like a backwards-compatibility-breaking change, which makes it a non-starter for 2.x.

I'd say it's a *potential* breaking change. However, this behavior was not captured by any of the other unittests (at least as seen on my system), and required zero changes in those unittests to make them pass.

TBH, I couldn't imagine a scenario where this is desired functionality, and would deem the current behavior as a bug. Copying completely makes sense for const lvalue (no modification) and rvalue (terminal lifetime), but not for mutable lvalues.

wjakob:
Agreed with @jagerman. I don't think we'll want to change something so fundamental about return value policies (and I quite frankly don't see the benefit here).

EricCousineau-TRI:
Gotcha, thank y'all for looking at it!

Regarding value, I personally have value in it in that (a) it is closer to C++'s behavior and seems like expected (since I and at least two others were surprised by it) and (b) the alternative is to manually annotate input arguments, which seems like this might grow rather verbose.

If the route for (b) were taken and I wanted to eventually abandon our fork, I'd probably end up writing wrappers around all overloads for `def`, `def_static`, `cpp_function`, etc., to automatically infer these flags (because, at least at this time of writing, I'd never want the original behavior), which seems like a lot more work than the 6 lines of code added to the codebase to realize the change.

If backwards-incompatibility is the main issue, is it something that could be considered for 3.x?
(Is there a roadmap or something for that?)

jagerman:
The biggest reason for defaulting to `copy` is that it's a *safe* operation (from a Python caller's perspective), while `reference` isn't, and we generally try to make potentially unsafe things opt-in.  Requiring you to be explicit with `reference` is essentially making sure you take that extra step to say "yeah, I know this is potentially unsafe, but I know what I'm doing."  For example:
```C++
class Element { /* ... */ };
class HasElement {
private:
    int i_;
public:
    int &get() { return i_; }
};
```
In C++, I need to know at the point of calling whether the returned value can be referenced or not.  If I do it wrong, I get into trouble:
```C++
    Element &e = HasElement().get(); // bad
    Element e = HasElement().get(); // No problem (invokes a copy)
```
Now in the Python interface for something like this, you'd normally bind it with `reference_internal`, so that the equivalent becomes safe:
```Python
    e = HasElement().get() // Safe; e holds a reference to the temporary
```
*or* you make sure the Python code only ever gets references that can't be deleted.

As an Python `__call__` argument (which is really where both these issues are coming up via `std::function`), there's a similar non-safety that can come up with you try to use `reference` with an argument that has temporary lifetime: namely that Python code might try to keep a reference, which might not be valid.  For example:
```Python
def set_element(e):
    global last_element
    last_element = e
foo(set_element)
```
combined with:
```C++
m.def("foo", [](py::function f) { f(HasElement().get()); });
```
Since right now the default is `automatic_reference`, that ends up copying.  You *can* tell it to reference with:
```C++
m.def("foo", [](py::function f) { f<py::return_value_policy::reference>(HasElement().get()); });
// (I suspect that probably needs a `template` and/or `operator ()` to actually compile)
```
but again, that's only something you do if you know that the returned lvalue reference is safe.

TL;DR: I do think we should *let* a `reference` policy be specified for how `std::function` casts its arguments; it's mainly a matter of how to make it work nicely.

EricCousineau-TRI:
I don't think I agree with you, as it seems like you're wanting to by-default redirect / break a C++ mechanism (mutable lvalue references) with `pybind` (defaulting to `copy`) rather than maintain it (`reference`) or augment it (`reference_internal`). (By break, I mean that you're not guaranteed to be modifying the instance you were passed for callback arguments, since they not may be registered in `pybind`, yadda yadda.).

I see what you mean by your example for a callback `[](py::function f) { f(HasElement().get()); }`, where a user may invoke a callback which is impure, and maintains a reference to a soon-to-be-invalidated temporary.

However, that seems like a *very* unusual edge case: the callback-invoking method uses a temporary value in C++, calls a method directly using `py::function` which is *directly* passed the return of a method returning a lvalue reference.

What I would see as more usual is that the user accepts a callback from Python, but then casts it to a `std::function<...>`, and then passes that to another, pure C++ callsite, where that callsite may have the temporary lvalue reference because, hey, it's pure C++. In this case, if the callback does not intend on modifying the value, then argument will be `T` or `const T&`, and then all the copies will be had on the pybind side, no harm no foul.
But, if the callback does modify the value, then it will be the same exact problem: The developer will have to explicitly dictate that they want a reference in the `py::cpp_function`, a downstream user will pass an impure callback that stores a global reference, and they'll be back at square one with the same segfault (with extra work on top of that).
(Additionally, if there is a segfault (at some point), it may be clearer that something has gone awry with referencing, rather than running the code to find out that your binding has not worked as expected. May that's just the masochist in me speaking, though...)

(As a minor aside, wouldn't `int&` be copied anyways, since it's a primitive / built-in pybind type? I know it's not super relevant to the discussion as it's a conceptual example, but I actually don't know.)

All that being said, ultimately a smaller workaround for me would be to wrap whatever methods, and replace any mutable rvalue references (arguments and return values) with pointers (possibly with a non-null check), as that'd be the behavior I'd want.

Thank you for the explanations and taking time to respond - sorry for the stubborness, just throwing in my 2 cents!

EricCousineau-TRI:
Oh goodness, I feel rather silly. After focusing on the issue posted here, I realized my problem was actually this and one more thing: the default behavior of `return_value_policy::copy` for *const* lvalue references, for non-copyable, non-movable classes.

I will post more about this in a follow-up issue.

That being said, I still stand by my original statements. However, they now have much less value to me, so I'd say I'd be fine with closing this PR :)

PR #1243, opened 2018-01-09
Issue DEBUG-only failure if STL wrappers attempt to allocate size of -1

steven-johnson:
Not sure if this is the best way to approach this issue, from either code or design point of view. The intent here is that if user code omits a __len__ definition for a sequence-like object, we fail with a helpful error message, rather than just "bad_alloc".


jagerman:
Seems reasonable to me.  Can you add a test case for this to the PR?

Issue #1204, opened 2017-12-04
"anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]"


anntzer:
## Issue description

Compiling a file that includes pybind11/detail/common.h (it is private but obviously included by other public headers) with clang -Wpedantic triggers the warning
```
In file included from src/main.cpp:1:
In file included from /usr/include/python3.6m/pybind11/pybind11.h:43:
In file included from /usr/include/python3.6m/pybind11/attr.h:13:
In file included from /usr/include/python3.6m/pybind11/cast.h:13:
In file included from /usr/include/python3.6m/pybind11/pytypes.h:12:
/usr/include/python3.6m/pybind11/detail/common.h:386:9: warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]
        struct 
```

## Reproducible example code

Add `#include <pybind11/detail/common.h>` to python_example and compile with
```
CC=clang CFLAGS=-Wpedantic CXX=clang setup.py build_ext
```

Psirus:
I've run into this as well. Since I'm using the provided CMake macros (thank you for this, btw), I would be fine with a solution like this:

~~~diff
diff --git a/tools/pybind11Tools.cmake b/tools/pybind11Tools.cmake
index a7c471a..b8bfb1c 100644
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -130,7 +130,7 @@ function(pybind11_add_module target_name)
 
   add_library(${target_name} ${lib_type} ${exclude_from_all} ${ARG_UNPARSED_ARGUMENTS})
 
-  target_include_directories(${target_name}
+  target_include_directories(${target_name} SYSTEM
     PRIVATE ${PYBIND11_INCLUDE_DIR}  # from project CMakeLists.txt
     PRIVATE ${pybind11_INCLUDE_DIR}  # from pybind11Config
     PRIVATE ${PYTHON_INCLUDE_DIRS})
~~~

Psirus:
Was just a suggestion in case getting the headers warning free was too intrusive. Obviously I'd prefer a solution that works for everyone as well.

wjakob:
Thank you -- yes, please do.

jagerman:
Fixed by #1248

PR #1245, opened 2018-01-11
Ignore `-Wnested-anon-types` when using Clang

Psirus:
As it says in the title, this change will add the `-Wnested-anon-types` to the list of ignored warning flags under Clang. Solves #1204.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -38,6 +38,9 @@
 #  if __GNUC__ >= 7
 #    pragma GCC diagnostic ignored "-Wnoexcept-type"
 #  endif
+#elif defined(__clang__)
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wnested-anon-types"
 #endif
 
 #include "attr.h"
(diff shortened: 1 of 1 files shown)


jagerman:
In this case I think it might be cleaner to listen to the warning rather than ignoring it, i.e. by adding:

```C++
struct nonsimple_values_and_holders {
    void **values_and_holders;
    uint8_t *status;
};
```
just before `struct instance {` in `detail/common.h`, then using it in the union by changing the union from:
```C++
    union {
        void *simple_value_holder[1 + instance_simple_holder_in_ptrs()];
        struct { /* ... */ } nonsimple;
    };
```
to:
```C++
    union {
        void *simple_value_holder[1 + instance_simple_holder_in_ptrs()];
        nonsimple_values_and_holders nonsimple;
    };
```

(Assuming that's the only thing raising this warning, of course).

Psirus:
Fine by me, it's just that most of the pybind code is way over my head and I didn't want to break anything.

Psirus:
Solved in an improved way in #1248.

PR #1248, opened 2018-01-11
Use a named rather than anon struct in instance

jagerman:
The anonymous struct nested in a union triggers a -Wnested-anon-type
warning ("anonymous types declared in an anonymous union are an
extension") under clang (#1204).  This names the struct and defines it
out of the definition of `instance` to get around to warning (and makes
the code slightly simpler).


jagerman:
Cc @Psirus, @anntzer - I think this PR should avoid the warning, but confirmation would be nice before I merge it.

anntzer:
works for me, thanks!

Psirus:
Same for me. Thank you very much.

Issue #1247, opened 2018-01-11
Regression from v2.1.1 to v2.2.1


daveah:
## Issue description

When moving from v2.1.1 to v2.2.1 a wrapper around a class with both std::vector and std::initializer_list templated constructors storing data in a boost variant no longer compiles.

## Reproducible example code

```
#include "pybind11/pybind11.h"
#include <vector>
#include <initializer_list>
#include <boost/blank.hpp>
#include <boost/variant.hpp>

namespace py = pybind11;

class Foo {
public:
  template <typename TT>
  Foo(TT init) : Foo(std::vector<TT>(std::move(init))) {}
  template <typename TT>
  Foo(std::vector<TT> init) : _data(std::move(init)) {}
  template <typename TT>
  Foo(std::initializer_list<TT> init) : Foo(std::vector<TT>(init)) {}

private:
  using var = boost::variant<boost::blank, std::vector<uint32_t>>;

  var _data;
};

PYBIND11_PLUGIN(_foo) {
  py::module mm("_foo", "Foo");
  py::class_<Foo> foo(mm, "Foo");
  foo.def(py::init<uint32_t>());
  foo.def(py::init<std::vector<uint32_t>>());
  return mm.ptr();
};
```

Using boost 1.59, gcc 6.3.1 and python 3.5 this compiles fine with 2.1.1, but not with 2.2.1.

jagerman:
This was caused by #1015, which switched to using brace initialization so that you can initialize simple aggregate types with `py::init<>`, e.g.
```C++
struct Agg { int a; int b; };
// ...
m.def(py::init<int, int>());
```
Unfortunately, this causes one fairly rare corner case compatibility issue (also discussed in #1015) which is just exactly what you ran into: when the class has an `initializer_list` constructor it wins for brace initialization over a regular constructor.  All of this means we end up invoking the `Foo(std::initializer_list<std::vector<uin32_t>> init)` template specialization, which, in your code, tries to invoke the `Foo(std::vector<std::vector<uint32_t>> init)` constructor, which of course fails (`_data` expects a uint32_t vector, not a uint32 vector-vector).

Long story short, it needs a workaround [as mentioned in the docs -- see the Note here](http://pybind11.readthedocs.io/en/stable/advanced/classes.html#brace-initialization).

One option--the one mentioned in the docs--is to change your constructor definition to the new-in-2.2 style lambda init:
```C++
[...]

PR #1249, opened 2018-01-11
Use stricter brace initialization

jagerman:
This updates the `py::init` constructors to only use brace
initialization for aggregate initiailization if there is no constructor
with the given arguments.

This, in particular, fixes the regression in #1247 where the presence of
a `std::initializer_list<T>` constructor started being invoked for
constructor invocations in 2.2 even when there was a specific
constructor of the desired type.

The added test case demonstrates: without this change, it fails to
compile because the `.def(py::init<std::vector<int>>())` constructor
tries to invoke the `T(std::initializer_list<std::vector<int>>)`
constructor rather than the `T(std::vector<int>)` constructor.

By only using `new T{...}`-style construction when a `T(...)`
constructor doesn't exist, we should bypass this by while still allowing
`py::init<...>` to be used for aggregate type initialization (since such
types, by definition, don't have a user-declared constructor).

### DIFF
--- a/include/pybind11/detail/init.h
+++ b/include/pybind11/detail/init.h
@@ -52,6 +52,16 @@ bool is_alias(Cpp<Class> *ptr) {
 template <typename /*Class*/>
 constexpr bool is_alias(void *) { return false; }
 
+// Constructs and returns a new object; if the given arguments don't map to a constructor, we fall
+// back to brace aggregate initiailization so that for aggregate initialization can be used with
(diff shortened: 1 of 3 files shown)


jagerman:
Cc @wjakob , @dean0x7d .  I *think* this approach should restore the backwards-compatibility while still letting us have aggregate initializers, but I may be overlooking something.

wjakob:
The crazy corner cases that users of this project unearth don't cease to amaze. Your fix looks very reasonable to me!

Issue #1239, opened 2018-01-07
Visual Studio 2017 compiler error with Conformance Mode enabled


chhenning:
With v15.5.2 I just noticed the compiler flag "/permissive-". If enabled just "#include <pybind11/pybind11.h>" will generate the error:
```
error C2884: 'pybind11::detail::_': introduced by using-declaration conflicts with local function 'pybind11::detail::_'
```
The solution is to set Conformance Mode to "No".

For whatever reason this flag was enabled when I created a new project.

jagerman:
Does this happen with current `master`?  There are some fixes in current master (which will also be included in 2.2.2) to fix compilation errors under `/permissive-`, and our appveyor builds are also building (successfully) with `/permissive-` since a582d6c7.

That said, I don't recall seeing the error you mention, so perhaps this is something new.

chhenning:
I just refreshed my master branch and the error still happens.
```
#include <pybind11\pybind11.h>

int main()
{
    return 0;
}
```

Please note that my Visual Studio version is 15.5.2. Not sure your appveyor has the latest of the greatest...

jagerman:
The Appveyor 2017 build is usually updated fairly quickly -- and should be 15.5.2 according to https://www.appveyor.com/docs/build-environment/#visual-studio-2017.

I'm updating my local MSVC now (which looks like it's going to 15.5.3) and will see if I can reproduce.

jagerman:
Aha, looks like we just started getting the same error in the appveyor builds (e.g. https://ci.appveyor.com/project/jagerman/pybind11/build/1.0.1029/job/b10hn4mcu525mon4).  I'll investigate and push a fix shortly.

jagerman:
This seems to be caused by the `using detail::_;` in `pybind11.h`, but that's no longer in `master` since 56613945a, and indeed I can't reproduce it on current master (are you sure you `master` branch is in sync with our `master` branch?).  The errors I linked to above are coming from the v2.2 branch, to which I've added a commit for 2.2.2 (in #1250) that fixes it.

chhenning:
Sorry, I just noticed I'm on stable branch. And you are correct I do not have the issue with master branch.

jagerman:
Closing: this is fixed in both master and the upcoming 2.2.2.

PR #1250, opened 2018-01-12
Make a 2.2.2 release

jagerman:
I think we've accumulated more than enough fixes to make a 2.2.2 release; this PR (against the v2.2 branch) is just the appropriate commits cherry-picked from `master` (and doesn't add anything new aside from the changelog and version variables bump).

If anyone knows of some critical fix that I missed in this PR, or some other issue that needs to be fixed for 2.2.2, now is the time to speak up to get it included.


jagerman:
Pushed a couple fixes for a new MSVC `/permissive-` error reported in #1239.  (That this warning comes up in the first place smells like an MSVC bug, but the workaround is easy and allows removing a bunch of now-unnecessary `detail::` qualifications).

jagerman:
Added 89cb495 (and updated the changelog), which is an equivalent to #1253 (for issue #1251), backported to 2.2 so as to leave the patients list as a vector (and walking through it to look for and avoid storing duplicate patients).

wjakob:
This all looks good to me, I will push out the release shortly.

wjakob:
Done. 2d0507db43cd5a117f7843e053b17dffca114107 adds the changelog to the main branch.

PR #1253, opened 2018-01-13
Track patients with unordered_set rather than vector

### LINKED ISSUE #1251: Using `py::keep_alive<N, P>()` may cause `internals.patients[nurse]` to grow without bounds
Looking at some of the internals, it looks like `return_value_policy::reference_internal` is only "activated" when an instance is created:
https://github.com/pybind/pybind11/blob/48e1f9a/include/pybind11/cast.h#L552

However, `py::keep_alive<N, P>()` looks like it is always activated:
https://github.com/pybind/pybind11/blob/48e1f9a/include/pybind11/cast.h#L552
https://github.com/pybind/pybind11/blob/48e1f9a/include/pybind11/attr.h#L442
https://github.com/pybind/pybind11/blob/48e1f9a/include/pybind11/pybind11.h#L1470
https://github.com/pybind/pybind11/blob/48e1f9a/include/pybind11/detail/class.h#L289

A simple fix would be to add a simple check:
```
#include <algorithm>
...
auto& p = internals.patients[nurse];
if (std::find(p.begin(), p.end(), patient) != p.end()) ...
```

This would impact performance a little, but would prevent a packrat problem for long(er)-running Python programs.

jagerman:
The stored vector of pybind-registered-type patients can grow without
bound if a function is called with the same patient multiple times.

This commit changes the pybind internal patient storage to an
`unordered_set` to avoid the issue.

Fixes #1251 

I've also pushed a separate fix for 2.2.2 (in #1250) that solves this by walking through the vector to avoid storing a duplicate.  (It's not as nice a fix, but it doesn't require changing the `internals` structure for 2.2.2).

### DIFF
--- a/tests/test_call_policies.cpp
+++ b/tests/test_call_policies.cpp
@@ -8,6 +8,7 @@
 */
 
 #include "pybind11_tests.h"
+#include "constructor_stats.h"
 
 struct CustomGuard {
     static bool enabled;
@@ -60,6 +61,21 @@ TEST_SUBMODULE(call_policies, m) {
         .def("returnChildKeepAlive", &Parent::returnChild, py::keep_alive<1, 0>())
         .def("returnNullChildKeepAliveChild", &Parent::returnNullChild, py::keep_alive<1, 0>())
         .def("returnNullChildKeepAliveParent", &Parent::returnNullChild, py::keep_alive<0, 1>());
+
+    // test_keep_alive_single
+    m.def("add_patient", [](py::object /*nurse*/, py::object /*patient*/) { }, py::keep_alive<1, 2>());
+    m.def("get_patients", [](py::object nurse) {
+        py::set patients;
+        for (PyObject *p : pybind11::detail::get_internals().patients[nurse.ptr()])
+            patients.add(py::reinterpret_borrow<py::object>(p));
+        return patients;
+    });
+    m.def("refcount", [](py::handle h) {
+#ifdef PYPY_VERSION
+        ConstructorStats::gc(); // PyPy doesn't update ref counts until GC occurs
+#endif
+        return h.ref_count();
+    });
 
 #if !defined(PYPY_VERSION)
     // test_alive_gc
--- a/tests/test_call_policies.py
+++ b/tests/test_call_policies.py
@@ -1,6 +1,6 @@
 import pytest
 from pybind11_tests import call_policies as m
-from pybind11_tests import ConstructorStats
+from pybind11_tests import ConstructorStats, UserType
 
 
 def test_keep_alive_argument(capture):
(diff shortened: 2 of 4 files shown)


wjakob:
Hi @jagerman -- sorry for not getting to this (now-ancient PR) earlier. This is just to say that your change looks good to me. We could merge it to master and consider it for inclusion to the next minor version. Would you be able to address the conflicts?

EricCousineau-TRI:
I'm briefly trying to see if I can update this PR, resolving merge conflicts, per discussion in #1251.

(For now, will do fast-forward merge)

tests/test_call_policies.py
```diff
@@ -69,6 +69,35 @@ def test_keep_alive_return_value(capture):
...
+    assert [m.refcount(nurse), m.refcount(p1), m.refcount(p2)] == [b, b + 1, b + 1]
+    m.add_patient(nurse, p2)
+    assert m.get_patients(nurse) == {p1, p2}
+    assert [m.refcount(nurse), m.refcount(p1), m.refcount(p2)] == [b, b + 1, b + 1]
+    m.add_patient(nurse, p2)
+    m.add_patient(nurse, p1)
+    assert m.get_patients(nurse) == {p1, p2}
+    assert [m.refcount(nurse), m.refcount(p1), m.refcount(p2)] == [b, b + 1, b + 1]
+    del nurse
+    assert [m.refcount(p1), m.refcount(p2)] == [b, b]
```
EricCousineau-TRI, line 98: I would like to see the behavior of `m.has_patients(nurse)` after it's deleted. Will add a brief test.

tests/test_call_policies.cpp
```diff
@@ -61,6 +62,21 @@ TEST_SUBMODULE(call_policies, m) {
...
 
+    // test_keep_alive_single
+    m.def("add_patient", [](py::object /*nurse*/, py::object /*patient*/) { }, py::keep_alive<1, 2>());
+    m.def("get_patients", [](py::object nurse) {
+        py::set patients;
+        for (PyObject *p : pybind11::detail::get_internals().patients[nurse.ptr()])
+            patients.add(py::reinterpret_borrow<py::object>(p));
+        return patients;
+    });
+    m.def("refcount", [](py::handle h) {
```
EricCousineau-TRI, line 73: This seems like it could live in Pure Python. Will move this there since `pytest.gc_collect()` exists.

tests/test_call_policies.py
EricCousineau-TRI: Done.

EricCousineau-TRI:
Filed #2289 per @YannickJadoul's request:
https://github.com/pybind/pybind11/issues/1251#issuecomment-656763382

My pref is on this PR, if we think it'll be more efficient.
However, I will defer to y'all on which one y'all want!

YannickJadoul:
Same thing as before: depends on whether/when we want to break the internal ABI ;-)

PR #1254, opened 2018-01-16
Update docs on std::out_of_range exception mapping

hovren:
Docs currently says that `std::out_of_range` is mapped to `ValueError`, but source says `IndexError`.
Given the documentation for `std::out_of_range` it seems like `IndexError` is the correct one, so the docs should be updated.


wjakob:
Merged (I did so directly from my account due to conflicts, and since it was just a 1-line change)

PR #1255, opened 2018-01-16
Add missing includes for better Bazel compatibility

steven-johnson:
Bazel has a "strict" build model that requires all C++ header files be compilable on their own, and thus must explicitly #include all headers they require (even if de facto header inclusion order means they'd get them "for free"). This adds a couple of headers that are needed (but missing) by this model.


eacousineau:
Nice! Albeit a tad late, might you have a chance to rebase this onto latest `master`? (Hopefully CI should be happy)

steven-johnson:
PTAL

wjakob:
LGTM.

PR #1270, opened 2018-02-06
Correct VS version in FAQ

schlamar:
There is no VS 2009 :-)


jagerman:
I wonder if the relevant section is still true: Is the official Windows Python binary still compiled with ancient MSVC?

schlamar:
Yes, Python 2.7 is still compiled with VS 2008. It is not possible to switch because of binary compatibility to existing extensions.

Issue #1212, opened 2017-12-10
error LNK2019: unresolved external symbol __imp_strdup


looooo:
## Issue description
I currently try to use pybind11 in FreeCAD. But there I run into a difficult problem. I have no clue to which library this problem refers to. The error occurs during linking a target which has pybind11 imported. The problem occurs also if there is no binding-code at all, but pybind11 included.

This happens on **windows**. Everything works on linux.

This is the error-message:
```
Creating library src\Mod\MeshPart\App\flatmesh.lib and object src\Mod\MeshPart\App\flatmesh.exp
MeshFlatteningPy.cpp.obj : error LNK2019: unresolved external symbol __imp_strdup referenced in function "protected: void __cdecl pybind11::cpp_function::initialize_generic(struct pybind11::detail::function_record *,char const *,class type_info const * const *,unsigned __int64)" (?initialize_generic@cpp_function@pybind11@@IEAAXPEAUfunction_record@detail@2@PEBDPEBQEBVtype_info@@_K@Z)
Mod\MeshPart\flatmesh.pyd : fatal error LNK1120: 1 unresolved externals
```

I also opened an isuue at the conda-forge feedstock some time ago, as I thought it might be related to the conda-package...: https://github.com/conda-forge/pybind11-feedstock/issues/23

wjakob:
Any kind of error message with ``_imp_*`` means that some of the DLL import/export settings are messed up. Potentially you're not linking to the correct standard library (MSVC has various different types -- DLL, Static, Threaded/Non-Threaded). Unfortunately this is not something for which we can provide support here.

looooo:
regarding: https://stackoverflow.com/questions/7582394/strdup-or-strdup we should use _strdup for visual c++. 
So maybe defining:
```#define strdup _strdup```
for windows?

PR #1279, opened 2018-02-14
replace strdup -> _strdup

looooo:
solves issue #1212


wjakob:
``_strdup`` is a non-standard MSVC intrinsic. Closed, see #1212.

PR #1301, opened 2018-02-27
Add test case for memory leak when using lambdas

yeganer:
Using a function as an argument to a C++ object causes a reference cycle
that the garbage collector is somehow not able to break, causing the
corresponding C++ object to leak.

### DIFF
--- a/tests/test_embed/test_interpreter.cpp
+++ b/tests/test_embed/test_interpreter.cpp
@@ -1,4 +1,5 @@
 #include <pybind11/embed.h>
+#include <pybind11/functional.h>
 
 #ifdef _MSC_VER
 // Silence MSVC C++17 deprecation warning from Catch regarding std::uncaught_exceptions (up to catch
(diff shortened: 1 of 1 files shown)


rwgk:
I'd say there is a leak somewhere, but where?

I'd expect that `local` is automatically cleared when the C++ `local` variable goes out of scope, unless the `local`(s) passed via `py::exec` is/are kept alive elsewhere.

We know with a high degree of certainty that `py::dict` doesn't leak. Something that could easily be demonstrated by removing the `py::exec` call.

General: I do know that `Py_Initialize` and `Py_Finalize` should only be called once each in a process:

https://github.com/rwgk/stuff/blob/dc76a8713a2fdaf520893d96255da56fe8a401ab/noddy/embedded_noddy_main_run.c

Python does not fully clean up after itself. I believe it could maybe be better than it is, but a full cleanup seems infeasible. See "Bugs and caveats" here:

One way to get more certainty about the leak would be to reduce `py::exec` to the raw Python C API equivalent (seems pretty straightforward, if someone wants to try). The outcome could well be that pybind11 is fine and that the leak is in Python. Or not, idk.

PR #1303, opened 2018-02-28
Improve macro type handling for types with commas

jagerman:
- `PYBIND11_MAKE_OPAQUE` now takes `...` rather than a single argument and
  expands it with `__VA_ARGS__`; this lets templated, comma-containing
  types get through correctly.
- Adds a new macro `PYBIND11_TYPE()` that lets you pass the type into a
  macro as a single argument, such as:
  ```C++
  PYBIND11_OVERLOAD(PYBIND11_TYPE(R<1,2>), PYBIND11_TYPE(C<3,4>), func)
  ```
  Unfortunately this only works for one macro call: to forward the
  argument on to the next macro call (without the processor breaking it
  up again) requires also adding the `PYBIND11_TYPE(...)` to type macro
  arguments in the `PYBIND11_OVERLOAD_...` macro chain.
- updated the documentation with these two changes, and use them at a couple
  places in the test suite to test that they work.

### DIFF
--- a/docs/advanced/cast/stl.rst
+++ b/docs/advanced/cast/stl.rst
@@ -175,9 +175,6 @@ in Python, and to define a set of available operations, e.g.:
         }, py::keep_alive<0, 1>()) /* Keep vector alive while iterator is used */
         // ....
 
-Please take a look at the :ref:`macro_notes` before using the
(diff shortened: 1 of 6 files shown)


wjakob:
Super-neat -- the change looks good to me.

PR #1316, opened 2018-03-10
Reimplement version check and combine init macros

### LINKED ISSUE #1314: Replace sscanf by sscanf_s on Windows?
## Issue description

PYBIND11_PLUGIN currently uses sscanf to parse the output of Py_GetVersion().  On MSVC, the use of sscanf triggers a warning:
> warning C4996: 'sscanf': This function or variable may be unsafe. Consider using sscanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.

Obviously 1) the warning is spurious (the literal format used is totally safe) and 2) the programmer could, indeed, set _CRT_SECURE_NO_WARNINGS; but still, perhaps pybind11 could just use sscanf_s there?

## Reproducible example code

N/A

jagerman:
This reimplements the version check to avoid sscanf (which has
reportedly started throwing warnings under MSVC, even when used
perfectly safely -- #1314).  It also extracts the mostly duplicated
parts of PYBIND11_MODULE/PYBIND11_PLUGIN into separate macros.

Fixes #1314

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -211,6 +211,31 @@ extern "C" {
 #define PYBIND11_TOSTRING(x) PYBIND11_STRINGIFY(x)
 #define PYBIND11_CONCAT(first, second) first##second
 
+#define PYBIND11_CHECK_PYTHON_VERSION \
+    {                                                                          \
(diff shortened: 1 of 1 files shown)


anntzer:
I can confirm that this fixes the issue.

wjakob:
This looks good to me!

Issue #1312, opened 2018-03-09
Returning NULL with an exception set from a py::init ctor


anntzer:
## Issue description

It is apparently not allowable to return a null pointer with an exception set from a py::init constructor.  It would be nice if that was indeed possible (raising the exception on the Python side, of course).

## Reproducible example code

```

#include <pybind11/pybind11.h>

struct T { };

namespace py = pybind11;

PYBIND11_MODULE(python_example, m) {
    py::class_<T>(m, "T")
    .def(
      py::init(
        [](int x) -> T* {
          if (x == 0) {
            PyErr_SetString(PyExc_RuntimeError, "oops");
            return nullptr;
          } else {
            return new T{};
          }
        }));
}
```
calling `T(0)` yields "TypeError: pybind11::init(): factory function returned nullptr", when I was hoping for it to throw RuntimeError("oops").

Of course, in this case, I could just have thrown a c++ exception and rely on it being converted; the actual use case is when the constructor is calling back into the Python C-API, which may set an exception.

Originally posted on gitter.

jagerman:
Changing `return nullptr` to `throw py::error_already_set()` should do what you want.

jagerman:
Sounds good to me.  Perhaps a wording tweak:
"... initialized and just needs to be propagated to Python"
to:
"... initialized.  This exception is used to propagate a Python exception generated in C++ code back to Python."

jagerman:
Closed via #1318.

PR #1318, opened 2018-03-11
Clarify error_already_set documentation.

anntzer:
Per discussion in https://github.com/pybind/pybind11/issues/1312#issuecomment-372118330.

The table was wrapped to 79 characters as the second column seemed overly tight otherwise.

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -11,45 +11,45 @@ exceptions:
 
 .. tabularcolumns:: |p{0.5\textwidth}|p{0.45\textwidth}|
 
-+--------------------------------------+------------------------------+
-|  C++ exception type                  |  Python exception type       |
-+======================================+==============================+
-| :class:`std::exception`              | ``RuntimeError``             |
-+--------------------------------------+------------------------------+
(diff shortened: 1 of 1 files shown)


docs/advanced/exceptions.rst
```diff
@@ -11,45 +11,45 @@ exceptions:
...
+| :class:`pybind11::key_error`         | ``KeyError`` (used to indicate out   |
+|                                      | of bounds access in ``__getitem__``, |
+|                                      | ``__setitem__`` in dict-like         |
+|                                      | objects, etc.)                       |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::error_already_set` | Indicates that the Python exception  |
+|                                      | flag has already been set; this      |
+|                                      | exception is used to propagate the   |
+|                                      | Python exception generated in C++    |
+|                                      | code back to Python.                 |
```
jagerman: The wording seems a little bit awkward to me.  How about:

Indicates that the Python exception flag has already been set via Python API calls from C++ code; this C++ exception is used to propagate such a Python exception back to Python.

anntzer: edited accordingly

PR #1321, opened 2018-03-12
Adds pydoc2cppdoc.py and example documentation.

paulross:
If you are migrating a Python project to pybind11 it is a bit painful cutting and pasting the Python documentation strings into the pybind11 code.

So I have written a little tool that extracts all the Python docstrings for a module and writes them out in a form suitable as a header file that can be referenced by the pybind11 code.

I have and example project that used it here: https://github.com/paulross/xmlwriter


wjakob:
Thank you for the PR. I think it addresses a bit too rare of a use case though to warrant including in pybind11 though -- most users would rather be interested in easier methods of moving comments from C++ headers to pybind11 binding code.

PR #1322, opened 2018-03-14
WIP: Add runtime selection of GIL implementation

### LINKED ISSUE #1276: Double-nested gil_scoped_release in a pythread causes fatal interpreter error

KyleFromKitware:
Fixes #1276.

Example usage:

```c++
#include <pybind11/embed.h>

#include <iostream>

int main() {
  pybind11::scoped_interpreter interp;
  pybind11::select_gil_impl<pybind11::basic_gil_impl>();

  pybind11::gil_scoped_release release;
  std::cout << "Running C++ code" << std::endl;

  return 0;
}
```

The `select_gil_impl()` call notifies pybind that this particular program *requires* the `basic_gil_impl` implementation. If another executable or library has already called `select_gil_impl()` with a different, incompatible implementation, an exception is thrown. If no library or executable ever calls `select_gil_impl()`, the `basic_gil_impl` is used by default. (Whether we should use basic or advanced by default is an ongoing discussion. I personally am in favor of basic being the default.)

If we want to use the advanced API with the thread disassociation feature:

```c++
#include <pybind11/embed.h>

#include <iostream>

int main() {
  pybind11::scoped_interpreter interp;
  pybind11::select_gil_impl<pybind11::advanced_gil_impl>();

  pybind11::advanced_gil_impl::release release(true);
  std::cout << "Running C++ code" << std::endl;

  return 0;
}
```

`advanced_gil_impl` doesn't currently do any checks to see if `basic_gil_impl` has already been used. That has been left up to the `gil_scoped_release` internals.

Important note: this PR *does* introduce two incompatible changes requiring a major version update:

1. The advanced API is no longer used by default. This is easy to change.
2. You can no longer call `gil_scoped_release` with the boolean disassociation flag. You have to use `advanced_gil_impl::release` instead.

This is the first draft of a WIP. It should serve as a good first sketch of what we want this to look like. Comments welcome, I look forward to hearing your feedback.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ set(PYBIND11_HEADERS
   include/pybind11/detail/class.h
   include/pybind11/detail/common.h
   include/pybind11/detail/descr.h
+  include/pybind11/detail/gil_internals.h
   include/pybind11/detail/init.h
   include/pybind11/detail/internals.h
   include/pybind11/detail/typeid.h
@@ -58,6 +59,7 @@ set(PYBIND11_HEADERS
   include/pybind11/embed.h
   include/pybind11/eval.h
   include/pybind11/functional.h
+  include/pybind11/gil.h
   include/pybind11/numpy.h
   include/pybind11/operators.h
   include/pybind11/pybind11.h
--- a/include/pybind11/detail/gil_internals.h
+++ b/include/pybind11/detail/gil_internals.h
@@ -1,0 +1,138 @@
+/*
+    pybind11/detail/gil_internals.h: Containers for GIL implementations
+
+    Copyright (c) 2018 Kitware Inc. <kyle.edwards@kitware.com>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include "common.h"
+#include "internals.h"
+
+#include "../gil.h"
+
+NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+
+class basic_gil_impl;
+
+// Change this to change the default GIL implementation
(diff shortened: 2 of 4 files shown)


wjakob:
See higher-level comments on #1276.

Issue #1294, opened 2018-02-22
vectorizing a method that operates on an array


nschloe:
I have a pybind11ified function that operates on a vector of dimension 1,
```c++
#include <pybind11/pybind11.h>
#include <pybind11/numpy.h>

namespace py = pybind11;

void distill(py::array_t<double, py::array::c_style | py::array::forcecast> p) {
  auto buf1 = p.request();

  auto r = p.mutable_unchecked<1>();
  for (ssize_t i = 1; i < r.shape(0); i++) {
    r(i) = 3.14 * (r(i) + r(i-1));  // whatever
  }
}

PYBIND11_MODULE(_accupy, m) {
  m.def("distill", &distill);
}
```
I would now like to change the function to have it operate on the first dimension of numpy arrays of arbitrary dimensionality. The first thing that comes to mind is adding an inner loop
```c++
for (ssize_t j = 0; j < r.shape(1); j++) {
}
```
to the above and wrapping the method in Python code à la
```python
def distill(p):
    q = p.reshape(p.shape[0], numpy.prod(p.shape[1:]))
    out = _mymod.distill(q)
    return out.reshape(p.shape)
```
This however wouldn't work with arrays of one dimension anymore. Another option:
```python
distill = numpy.vectorize(_mymod.distill, signature="(m)->()")

distill(a.T).T
```
– Also a bit weird to fall back to vectorize after we've already been in C++.

Is there a canonical way for vectorizing pybind11 methods that operate on an array?

calumhrmurray:
Is this what you are looking for?

http://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#vectorizing-functions

For me it was as simple as,

.def("function_name",py::vectorize(&class_name::function_name))

nschloe:
This only works for functions that natively take scalar values, doesn't it?

calumhrmurray:
I think I misunderstood your problem. You want to give an array to a vectorized c++ function? Whereas my solution was giving an array to a vectorized pybind11 c++ function? 

Why not just have a double function in c++ and then vectorize it?

In my experience a py::vectorize tool works very quickly. In my particular use a comparison of vectorized c++ code versus vectorizing with pybind11, the pybind11 vectorization came out quicker. This could be that my c++ was not great though.

nschloe:
The operation is supposed to work only in the first dimension of the array. Check out the example in the original post:
```c++
[...]

PR #1325, opened 2018-03-17
[WIP] numpy_ufunc: Allow custom ufuncs to be defined 

EricCousineau-TRI:
I ended up implementing this for the following feature:
https://github.com/RobotLocomotion/pybind11/pull/12

This is a very minor adjustment for a WIP submission; things left to do:
- [ ] Simplify logic for defining builtin dtype loops; at present, it's nasty because I originally only focused on user-defined dtypes
- [ ] Add unittests - there are some in the above fork's PR for operator overloading, but nothing explicitly focused on ufuncs really
- [ ] See if structured dtypes are supported
- [ ] Either fix `detail/inteference.h` to be C++11-friendly, or ditch it and inline the function signature inference - I've seen this mechanism more or less copied+pasted in several locations in the code, and had pulled this from my implementation [here](https://github.com/RobotLocomotion/drake/blob/b9eb279cdbe2095d660fb8ad0be14c643967ced2/bindings/pydrake/util/wrap_function.h#L85) in preparation for wrapping things like `def` without having to repeating myself too much

May relate https://github.com/pybind/pybind11/issues/1294?

### DIFF
--- a/include/pybind11/detail/inference.h
+++ b/include/pybind11/detail/inference.h
@@ -1,0 +1,141 @@
+/*
+    pybind11/detail/inference.h -- Simple inference for generic functions
+
+    Copyright (c) 2018 Eric Cousineau <eric.cousineau@tri.global>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include "common.h"
+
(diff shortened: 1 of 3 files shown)


chaitan94:
This is really interesting! :smile: 

I know it's been quite a while now since this PR was raised - but would it be possible for you to add a code sample showing an example/intended usage or a test case of this new feature? I wanted to test this out and I think having that would add a lot of clarity to the PR in general.

PR #1327, opened 2018-03-20
Fix potential crash when calling an overloaded function

oremanj:
The crash would occur if:
- dispatcher() uses two-pass logic (because the target is overloaded and some arguments support conversions)
- the first pass (with conversions disabled) doesn't find any matching overload
- the second pass does find a matching overload, but its return value can't be converted to Python

The code for formatting the error message assumed `it` still pointed to the selected overload,
but during the second-pass loop `it` was nullptr. Fix by setting `it` correctly if a second-pass
call returns a nullptr `handle`. Add a new test that segfaults without this fix.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -650,8 +650,15 @@ protected:
                         result = PYBIND11_TRY_NEXT_OVERLOAD;
                     }
 
-                    if (result.ptr() != PYBIND11_TRY_NEXT_OVERLOAD)
+                    if (result.ptr() != PYBIND11_TRY_NEXT_OVERLOAD) {
+                        if (!result) {
+                            // The error reporting logic below expects 'it' to be valid, as it would be
+                            // if we'd encountered this failure in the first-pass loop.
+                            for (it = overloads; it != &call.func; it = it->next)
+                                ;
+                        }
                         break;
+                    }
                 }
             }
         } catch (error_already_set &e) {
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -420,8 +420,8 @@ protected:
         using namespace detail;
 
         /* Iterator over the list of potentially admissible overloads */
-        function_record *overloads = (function_record *) PyCapsule_GetPointer(self, nullptr),
-                        *it = overloads;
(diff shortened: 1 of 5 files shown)


include/pybind11/pybind11.h
```diff
@@ -650,8 +650,15 @@ class cpp_function : public function {
                         result = PYBIND11_TRY_NEXT_OVERLOAD;
                     }
 
-                    if (result.ptr() != PYBIND11_TRY_NEXT_OVERLOAD)
+                    if (result.ptr() != PYBIND11_TRY_NEXT_OVERLOAD) {
+                        if (!result) {
+                            // The error reporting logic below expects 'it' to be valid, as it would be
+                            // if we'd encountered this failure in the first-pass loop.
+                            for (it = overloads; it != &call.func; it = it->next)
```
jagerman: ISTM this could be simplified down to just:
```C++
if (!result)
    it = &call.func;
```
(unless I'm forgetting something about this code).

tests/test_class.cpp
```diff
@@ -21,6 +21,10 @@ struct NoBraceInitialization {
     std::vector<int> vec;
 };
 
+struct NotRegistered {};
+struct StringWrapper { std::string str; };
+NotRegistered test_error_after_conversions(StringWrapper) { return {}; }
+
```
jagerman: These classes should be moved with the module definitions, and the function rewritten as a lambda, unless there's a good reason to keep them outside.  (We try to enforce that style in the test suite, where possible, to keep the test cases as localized as possible within each test suite file, to makes it easier to read).

include/pybind11/pybind11.h
oremanj: D'oh, right you are - I was getting confused about function_call vs function_record.

oremanj:
@jagerman Ping on this? The CI failure appears unrelated, e.g. just-uploaded https://github.com/pybind/pybind11/pull/1362 has the same error.

oremanj:
@wjakob @jagerman ping? I have a number of other PRs I'm sitting on which I have been holding off on contributing because this one has been stalled so long... for a while I was assuming there just wasn't development going on, but it looks like that's been more active lately.

wjakob:
Apologies for the delay -- this looks good now.

PR #1329, opened 2018-03-20
 numpy: Provide concrete size aliases, test equivalence for `dtype(...).num`

### LINKED ISSUE #1328: npy_format_descriptor may not be accurate for platform-dependent lengths of `short`, `long`, and `long long`
Ran into this via an esoteric route: When tinkering with user-defined dtypes, it seems that implementation of `np.ones` uses `np.copyto(dst, 1, casting='unsafe')`, where the `1` gets interpreted as `np.array(1, dtype=np.int64)` (at least, on my platform).

When I tried registered a caster using `npy_descriptor_format<int64_t>::dtype()` using the `type_num` field, I realized its value was for `NPY_LONGLONG`, whereas `NPY_INT64` corresponds to `NPY_LONG`.

Will try to see if I can create a failing test, and then submit a tentative fix PR based on that.

EricCousineau-TRI:
Closes #1328 

See the linked issue for context.

Note that this is split into two commits:
1. Introduces a simple explicit test on dtypes to guarantee that they match for concrete dtypes, and then some assurances on platform-dependent sizes. (No changes to `numpy.h`)
2. Mirrors the concrete type aliases (e.g. `NPY_INT32`). This solves an issue where `py::dtype::of<int64_t>().attr("num")` did not match to NumPy's definition of `NPY_INT64`.

TBH, I can understand some hesitation for incorporating the second commit. However, I would like to see the first commit make it in - took a bit to understand this issue and debug.

Todo:
- [x] Rebase once review is done.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -14,6 +14,7 @@
 #include <numeric>
 #include <algorithm>
 #include <array>
+#include <cstdint>
 #include <cstdlib>
 #include <cstring>
 #include <sstream>
@@ -108,6 +109,18 @@ inline numpy_internals& get_numpy_internals() {
     return *ptr;
 }
 
+template <typename T> struct same_size {
+    template <typename U> using as = bool_constant<sizeof(T) == sizeof(U)>;
+};
+
+// Lookup a type according to its size, and return a value corresponding to the NumPy typenum.
+template <typename Concrete, typename... Check, typename... Int>
+constexpr int platform_lookup(Int... codes) {
+    using code_index = std::integral_constant<int, constexpr_first<same_size<Concrete>::template as, Check...>()>;
+    static_assert(code_index::value != sizeof...(Check), "Unable to match type on this platform");
+    return std::get<code_index::value>(std::make_tuple(codes...));
+}
+
 struct npy_api {
     enum constants {
         NPY_ARRAY_C_CONTIGUOUS_ = 0x0001,
@@ -126,7 +139,23 @@ struct npy_api {
         NPY_FLOAT_, NPY_DOUBLE_, NPY_LONGDOUBLE_,
         NPY_CFLOAT_, NPY_CDOUBLE_, NPY_CLONGDOUBLE_,
         NPY_OBJECT_ = 17,
-        NPY_STRING_, NPY_UNICODE_, NPY_VOID_
+        NPY_STRING_, NPY_UNICODE_, NPY_VOID_,
+        // Platform-dependent normalization
+        NPY_INT8_ = NPY_BYTE_,
+        NPY_UINT8_ = NPY_UBYTE_,
+        NPY_INT16_ = NPY_SHORT_,
+        NPY_UINT16_ = NPY_USHORT_,
+        // `npy_common.h` defines the integer aliases. In order, it checks:
+        // NPY_BITSOF_LONG, NPY_BITSOF_LONGLONG, NPY_BITSOF_INT, NPY_BITSOF_SHORT, NPY_BITSOF_CHAR
+        // and assigns the alias to the first matching size, so we should check in this order.
+        NPY_INT32_ = platform_lookup<std::int32_t, long, int, short>(
+            NPY_LONG_, NPY_INT_, NPY_SHORT_),
+        NPY_UINT32_ = platform_lookup<std::uint32_t, unsigned long, unsigned int, unsigned short>(
+            NPY_ULONG_, NPY_UINT_, NPY_USHORT_),
+        NPY_INT64_ = platform_lookup<std::int64_t, long, long long, int>(
+            NPY_LONG_, NPY_LONGLONG_, NPY_INT_),
+        NPY_UINT64_ = platform_lookup<std::uint64_t, unsigned long, unsigned long long, unsigned int>(
+            NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
     };
 
     typedef struct {
@@ -1004,8 +1033,8 @@ private:
     // NB: the order here must match the one in common.h
     constexpr static const int values[15] = {
         npy_api::NPY_BOOL_,
-        npy_api::NPY_BYTE_,   npy_api::NPY_UBYTE_,   npy_api::NPY_SHORT_,    npy_api::NPY_USHORT_,
-        npy_api::NPY_INT_,    npy_api::NPY_UINT_,    npy_api::NPY_LONGLONG_, npy_api::NPY_ULONGLONG_,
+        npy_api::NPY_BYTE_,   npy_api::NPY_UBYTE_,   npy_api::NPY_INT16_,    npy_api::NPY_UINT16_,
+        npy_api::NPY_INT32_,  npy_api::NPY_UINT32_,  npy_api::NPY_INT64_,    npy_api::NPY_UINT64_,
         npy_api::NPY_FLOAT_,  npy_api::NPY_DOUBLE_,  npy_api::NPY_LONGDOUBLE_,
         npy_api::NPY_CFLOAT_, npy_api::NPY_CDOUBLE_, npy_api::NPY_CLONGDOUBLE_
     };
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -14,6 +14,67 @@
 
 #include <cstdint>
 
+// Size / dtype checks.
+struct DtypeCheck {
+    py::dtype numpy{};
+    py::dtype pybind11{};
+};
+
+template <typename T>
+DtypeCheck get_dtype_check(const char* name) {
+    py::module np = py::module::import("numpy");
+    DtypeCheck check{};
+    check.numpy = np.attr("dtype")(np.attr(name));
+    check.pybind11 = py::dtype::of<T>();
+    return check;
+}
+
+std::vector<DtypeCheck> get_concrete_dtype_checks() {
+    return {
+        // Normalization
+        get_dtype_check<std::int8_t>("int8"),
+        get_dtype_check<std::uint8_t>("uint8"),
+        get_dtype_check<std::int16_t>("int16"),
+        get_dtype_check<std::uint16_t>("uint16"),
+        get_dtype_check<std::int32_t>("int32"),
+        get_dtype_check<std::uint32_t>("uint32"),
+        get_dtype_check<std::int64_t>("int64"),
+        get_dtype_check<std::uint64_t>("uint64")
+    };
+}
+
+struct DtypeSizeCheck {
+    std::string name{};
+    int size_cpp{};
+    int size_numpy{};
+    // For debugging.
+    py::dtype dtype{};
+};
+
+template <typename T>
+DtypeSizeCheck get_dtype_size_check() {
+    DtypeSizeCheck check{};
+    check.name = py::type_id<T>();
+    check.size_cpp = sizeof(T);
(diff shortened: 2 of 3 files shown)


EricCousineau-TRI:
Patch between running `python -m pyteste -s test_numpy_array.py` on first commit vs. second commit:
```patch
diff --git a/tmp/a.txt b/tmp/b.txt
index cda8c6e..23fd223 100644
--- a/tmp/a.txt
+++ b/tmp/b.txt
@@ -18,9 +18,7 @@ test_numpy_array.py <DtypeSizeCheck name='short' size_cpp=2 size_numpy=2 dtype=i
 <DtypeCheck numpy=int32 pybind11=int32>
 <DtypeCheck numpy=uint32 pybind11=uint32>
 <DtypeCheck numpy=int64 pybind11=int64>
-NOTE: typenum mismatch for <DtypeCheck numpy=int64 pybind11=int64>: 7 != 9
 <DtypeCheck numpy=uint64 pybind11=uint64>
-NOTE: typenum mismatch for <DtypeCheck numpy=uint64 pybind11=uint64>: 8 != 10
 .................................
 
-========================== 33 passed in 0.08 seconds ===========================
+========================== 33 passed in 0.07 seconds ===========================
```

include/pybind11/numpy.h
```diff
@@ -108,6 +108,15 @@ inline numpy_internals& get_numpy_internals() {
...
 }
 
+// Lookup a type according to its size, and return a value corresponding to the NumPy typenum.
+template <typename Concrete,
+    typename Check1, typename Check2, typename Check3 = Check2>
+constexpr int platform_lookup(int check1, int check2, int check3 = -1) {
+    return (sizeof(Concrete) == sizeof(Check1)) ? check1 :
+               (sizeof(Concrete) == sizeof(Check2)) ? check2 :
+                 (sizeof(Concrete) == sizeof(Check3)) ? check3 : -1;
+}
```
jagerman: Here's a template that makes this a little more generic, along with a compile-time check for the error case (rather than returning -1).  It also requires passing the arguments as an `{ ... }` array because that seemed simpler.
```C++
using namespace pybind11::detail;

template <typename T> struct same_size { template <typename U> using as = bool_constant<sizeof(T) == sizeof(U)>; };

template <typename Concrete, typename... Check>
constexpr int platform_lookup(std::array<int, sizeof...(Check)> codes) {
    using code_index = std::integral_constant<int, constexpr_first<same_size<Concrete>::template as, Check...>()>;
    static_assert(code_index::value != sizeof...(Check), "Unable to match type on this platform");
    return codes[code_index::value];
}

int main() {
    // fails (short not matched):
    //int bar = platform_lookup<short, int, unsigned, unsigned long, int>({4, 9, 11, 13});
    int foo = platform_lookup<std::uint32_t, unsigned int, unsigned long>({8,9});
}
```

include/pybind11/numpy.h
```diff
@@ -126,7 +135,20 @@ struct npy_api {
...
+        NPY_UINT8_ = NPY_UBYTE_,
+        NPY_INT16_ = NPY_SHORT_,
+        NPY_UINT16_ = NPY_USHORT_,
+        NPY_INT32_ = platform_lookup<int32_t, short, int, long>(
+            NPY_SHORT_, NPY_INT_, NPY_LONG_),
+        NPY_UINT32_ = platform_lookup<uint32_t, unsigned short, unsigned int, unsigned long>(
+            NPY_USHORT_, NPY_UINT_, NPY_ULONG_),
+        NPY_INT64_ = platform_lookup<int64_t, int, long, long long>(
+            NPY_INT_, NPY_LONG_, NPY_LONGLONG_),
+        NPY_UINT64_ = platform_lookup<uint64_t, uint, unsigned long, unsigned long long>(
```
jagerman: `uint` isn't a standard C or C++ type (even if commonly available); that should be `unsigned int`, or if you're feeling lazy, just `unsigned` (though some people strongly dislike the latter).

The `int32_t` etc. types are more portable, but should still technically be prefixed with `std::` (we should also have a `#include <cstdint>` in here to properly bring them in).

jagerman: It looks like `uint` isn't available on MSVC.

jagerman:
It looks like the i386 tests broke, as well (one of the tests is coming up with a 4 instead of an 8).

include/pybind11/numpy.h
EricCousineau-TRI: Done. Thanks!

include/pybind11/numpy.h
EricCousineau-TRI: Unfortunately, C++11 doesn't like multiple statements in the function - it deems it non-`constexpr` :(

Will briefly tinker with it to see if I can add some indirection to preserve the `constexpr`-ness.

EricCousineau-TRI:
Reverted my change to see if this is something I introduced vs. something else to be fixed. 

This discrepancies here (a) between signed and unsigned and (b) between the numpy size (4) and its type (`int64`) seem *very* disturbing though...
```
<DtypeSizeCheck name='unsigned long' size_cpp=4 size_numpy=4 dtype=uint32>
<DtypeSizeCheck name='long long' size_cpp=8 size_numpy=4 dtype=int64>
```

include/pybind11/numpy.h
EricCousineau-TRI: Hup, was wrong: Just needed to make the `array<>` argument `const array<>`.
Thanks!

jagerman:
I found the easiest thing was to set up a free appveyor account and attach it to your own repository; then you can start, restart, etc. builds as required.

I'm not exactly sure what we need to do to fix Failure A, but that certainly isn't caused by your PR.

Failure B is more serious: it looks like MSVC hates my template generalization.  You could try this version:
```C++
template <int, typename...> struct matches_sizeof_impl : std::integral_constant<int, -1> {};
template <int N, typename T, typename A, typename... B>
struct matches_sizeof_impl<N, T, A, B...> : std::integral_constant<int,
    sizeof(T) == sizeof(A) ? N : matches_sizeof_impl<N+1, T, B...>::value> {};
template <typename T, typename... A> using matches_sizeof_index = matches_sizeof_impl<0, T, A...>;

template <typename Concrete, typename... Check, typename code_index = matches_sizeof_index<Concrete, Check...>>
constexpr int platform_lookup(const std::array<int, sizeof...(Check)> &codes) {
    static_assert(code_index::value >= 0, "Unable to match type on this platform");
    return codes[code_index::value];
}
```

It's getting a bit messier, but at least it keeps that compilation failure if the type isn't found.

jagerman:
Here's a better implementation; this makes `first_true_index` generic enough now that it could go into `pybind11/detail/common.h`:

```C++
template <int N, bool... Bool> struct first_true_impl : std::integral_constant<int, -1> {};
template <int N, bool B, bool... BMore> struct first_true_impl<N, B, BMore...> :
    std::integral_constant<int, B ? N : first_true_impl<N+1, BMore...>::value> {};
/// Takes a parameter pack of bool values, return the index of the first one that is true or -1 if
/// none of the template arguments are true.
template <bool... B> using first_true_index = first_true_impl<0, B...>;

template <typename Concrete, typename... Check, typename code_index = first_true_index<sizeof(Concrete) == sizeof(Check)...>>
constexpr int platform_lookup(const std::array<int, sizeof...(Check)> &codes) {
    static_assert(code_index::value >= 0, "Unable to match type on this platform");
    return codes[code_index::value];
}

**Comment by EricCousineau-TRI** [comment]
Done - sorry about the delay!

**Comment by EricCousineau-TRI** [comment]
Just rebased - sorry for the drawn out delay!

Actually ended up needing this, as implicit conversions for `py::type_caster` for arrays requires that the registered dtype for conversion be precisely `np.int64`; `dtype::of<int>` yields `np.int32`, and `dtype::of<std::int64_t>` does not yield the same typenum as `np.int64`.

**Comment by olegsinyavskiy** [comment]
@EricCousineau-TRI, great work! I've run into this issue recently. I got into that by passing pybind int64_t array into OpenCV functions. Without this fix you'll get "data type = 9 is not supported" Exception. Any plans on merging this soon?

**Comment by EricCousineau-TRI** [comment]
Lemme see if I can fix these pesky MSVC issues...

**Comment by EricCousineau-TRI** [comment]
K, finally found the magic incantation to make MSVC happy.

@jagerman Might you have a chance to give this another glance?

**Comment by eacousineau** [comment]
Hup! It looks like AppVeyor is happy now, huzzah! (I don't think I did anything special, aside from have Wenzel's update?)

@jagerman or @wjakob Can I ask if this might be good to merge?
(It may or may not also relate a more recent issue, #1806)

**Comment by wjakob** [comment]
This looks good to me -- merged! :)

**Comment by EricCousineau-TRI** [comment]
Awesome, thanks!

File: include/pybind11/numpy.h [hunk 3]
```diff
@@ -1004,8 +1033,8 @@ struct npy_format_descriptor<T, enable_if_t<satisfies_any_of<T, std::is_arithmet
     // NB: the order here must match the one in common.h
     constexpr static const int values[15] = {
         npy_api::NPY_BOOL_,
-        npy_api::NPY_BYTE_,   npy_api::NPY_UBYTE_,   npy_api::NPY_SHORT_,    npy_api::NPY_USHORT_,
-        npy_api::NPY_INT_,    npy_api::NPY_UINT_,    npy_api::NPY_LONGLONG_, npy_api::NPY_ULONGLONG_,
+        npy_api::NPY_BYTE_,   npy_api::NPY_UBYTE_,   npy_api::NPY_INT16_,    npy_api::NPY_UINT16_,
+        npy_api::NPY_INT32_,  npy_api::NPY_UINT32_,  npy_api::NPY_INT64_,    npy_api::NPY_UINT64_,
```
YannickJadoul, line 1037: Is this actually correct? Normally, it should be the other way around, `int16`, `int32`, and `int64` being aliases, and `l` should mean `long`, `q` `longlong`, ...?

See e.g.:
- https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types
- https://github.com/numpy/numpy/blob/6ef5ec39cdfaf77aa4600ec2e3bf9f679a4fd527/numpy/core/include/numpy/npy_common.h#L545
- https://github.com/numpy/numpy/blob/6ef5ec39cdfaf77aa4600ec2e3bf9f679a4fd527/numpy/core/include/numpy/ndarraytypes.h#L121
- https://github.com/numpy/numpy/blob/6ef5ec39cdfaf77aa4600ec2e3bf9f679a4fd527/numpy/core/code_generators/generate_umath.py#L204

EricCousineau-TRI: Good point! I've (explicitly) cross-ref'd your mention in the issue:
https://github.com/pybind/pybind11/issues/1908#issuecomment-660291752

The dumb question is, what's the correct testing fix? (it's been a while since I wrote this...)

YannickJadoul: I don't know. Still trying to figure out. I actually came across this by accident and wanted to make sure I'm correct, before starting to dig further into this! :-)

To make things more complex, I just found this as well:

```pycon
>>> np.dtype('i1')
dtype('int8')
>>> np.dtype('i2')
dtype('int16')
>>> np.dtype('i4')
dtype('int32')
>>> np.dtype('i8')
dtype('int64')
>>> np.dtype('u8')
dtype('uint64')
```

But yes, maybe let's discuss further in #1908 (though I'm not yet convinced that that one is actually an issue); easier to have things grouped in one place.

PR #1334, opened 2018-03-23
Fix enum value's __int__ returning non-int when underlying type is bool or of char type

### LINKED ISSUE #1331: py::enum_'s `__int__` and `__hash__` do not behave correctly when the underlying type of enum class is char
## Issue description

I am exposing an enum class to python, and this is my code:
``` cpp
// enum class to expose
enum class Player : char { 
    White = -1, None = 0, Black = 1 
};
```

```cpp
// pybind11 wrapper
py::enum_<Player>(mod, "Player", "Gomoku player types")
    .value("white", Player::White)
    .value("none",  Player::None)
    .value("black", Player::Black)
    //.def(py::init<int>()) this line run into another error
    .def("__int__",   [](Player p) { return static_cast<int>(p); })
    .def("__float__", [](Player p) { return static_cast<double>(p); })
    .def("__hash__",  [](Player p) { return static_cast<int>(p); })
    .def("__neg__",   [](Player p) { return -p; })
    .def_static("calc_score", [](Player p, Player w) { return getFinalScore(p, w); });

```

As I was testing the module in python, I ran into such a strange issue:
```bash
>>> int(Player.black)
TypeError: __int__ returned non-int (type str)

>>> hash(Player.black)
TypeError: __hash__ method should return an integer

>>> float(Player.black)
1.0
```
In fact, similar issues have occurred when I pass the `py::arithmetic()` extra parameter to py::enum_.

Then I checked the inner member functions and discovered these:
```bash
>>> Player.black.__int__()
'\x01'
>>> Player.white.__int__()
'ÿ'
>>> Player.white.__hash__()
'ÿ'
```

So, why does a `static_cast<int>` resulted in a str in python?

**Intuitively,  I change the underlying type of `enum class Player` from `char` to `int`, and:**
```bash
>>> Player.black.__int__()
1
>>> Player.white.__hash__()
-1
>>> int(Player.none)
0
```
Thus, it is the underlying type `char` of `enum class player` that causes all these bugs.

### LINKED ISSUE #1820: Enum with underlying char not supported in 2.3.0?
Just upgrade from 2.2.4 to 2.3.0,  nothing else changed,  but got error when import package in python as below

`libc++abi.dylib: terminating with uncaught exception of type pybind11::error_already_set: TypeError: __int__ returned non-int (type str)`

it seems enum definition not supported.   eg.

enum class A: char { First='1',   Second = '2', }

Vigilans:
### Introduction

This PR solves the problem of enum value's `__int__` returning a non-int value when its underlying type is bool or of char type.

### Detail

This PR extends the `enum_` class with following type alias:

* `Underlying` type: the actual underlying type of enum (The original `Scalar` type);
* `Scalar` type: the integer type with **length** and **signedness** equivalent to the underlying type.

The enum should always use `Scalar` as the inner integer type when dealing with python side.

So,
* Scalar for underlying type of normal integer would stay the same; 
* Scalar for underlying type of bool or char type would use the int type of same length and signedness, specified through the `equivalent_integer` struct.

### Applicable issues
Fix #1331
Fix #1820
Fix https://github.com/osmcode/pyosmium/issues/114

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1549,6 +1549,19 @@ struct enum_base {
     handle m_parent;
 };
 
+template <bool is_signed, size_t length> struct equivalent_integer {};
+template <> struct equivalent_integer<true,  1> { using type = int8_t;   };
+template <> struct equivalent_integer<false, 1> { using type = uint8_t;  };
+template <> struct equivalent_integer<true,  2> { using type = int16_t;  };
+template <> struct equivalent_integer<false, 2> { using type = uint16_t; };
+template <> struct equivalent_integer<true,  4> { using type = int32_t;  };
+template <> struct equivalent_integer<false, 4> { using type = uint32_t; };
+template <> struct equivalent_integer<true,  8> { using type = int64_t;  };
+template <> struct equivalent_integer<false, 8> { using type = uint64_t; };
+
+template <typename IntLike>
+using equivalent_integer_t = typename equivalent_integer<std::is_signed<IntLike>::value, sizeof(IntLike)>::type;
+
 NAMESPACE_END(detail)
 
 /// Binds C++ enumerations and enumeration classes to Python
@@ -1559,13 +1572,17 @@ public:
     using Base::attr;
     using Base::def_property_readonly;
     using Base::def_property_readonly_static;
-    using Scalar = typename std::underlying_type<Type>::type;
+    using Underlying = typename std::underlying_type<Type>::type;
+    // Scalar is the integer representation of underlying type
+    using Scalar = detail::conditional_t<
+        detail::is_std_char_type<Underlying>::value,
+        detail::equivalent_integer_t<Underlying>, Underlying>;
 
     template <typename... Extra>
     enum_(const handle &scope, const char *name, const Extra&... extra)
       : class_<Type>(scope, name, extra...), m_base(*this, scope) {
         constexpr bool is_arithmetic = detail::any_of<std::is_same<arithmetic, Extra>...>::value;
-        constexpr bool is_convertible = std::is_convertible<Type, Scalar>::value;
+        constexpr bool is_convertible = std::is_convertible<Type, Underlying>::value;
         m_base.init(is_arithmetic, is_convertible);
 
         def(init([](Scalar i) { return static_cast<Type>(i); }));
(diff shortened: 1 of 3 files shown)


jagerman:
Is `unsigned char` (or, for that matter, `signed char`, which is also distinct from `char` and `unsigned char`) necessary here?  That `is_std_char_type` template is there to capture the types that are used for strings, which is pretty much just `char`.  It shouldn't be catching `unsigned char` and `signed char` in the first place--those should fall through to the arithmetic converter.

Vigilans-Yea:
@jagerman 

I see. It corresponds to the compile error which seems to tell me that `unsigned char` is used as an arithmetic type, instead of an element of string, which is listed inside the `is_std_char_type`.
Thank you for your explanation≡ω≡

Vigilans:
Sorry for my foolishness of not getting your point and suspended this PR for one year long...

I've updated the PR to keep pace with the master, and fixed it with a more robust mechanism, along with more tests, to solve both issues of char and bool types.

My original post for context reference:

> I followed the suggestion mentioned in the issue and now the bug is eliminated.
> 
> BTW, It should be noted that the definition of `detailed::is_std_char_type` does not check  `unsigned char`:
> ```cpp
> template <typename CharT> using is_std_char_type = any_of<
>     std::is_same<CharT, char>, /* std::string */
>     std::is_same<CharT, char16_t>, /* std::u16string */
>     std::is_same<CharT, char32_t>, /* std::u32string */
>     std::is_same<CharT, wchar_t> /* std::wstring */
> >;
> ```
> And when I tried to add `unsigned char` to the type list, compile errors occurred. Therefore, I alternatively turn to add it to the definition of `enum_<Type>::ScalarInt`:
> ```cpp
> using ScalarInt = detail::conditional_t<
>     detail::is_std_char_type<Scalar>::value || std::is_same<Scalar, unsigned char>::value,
>     detail::conditional_t<std::is_signed<Scalar>::value, int, unsigned>,
>     Scalar>;
> ```

Vigilans:
Hi @wjakob @jagerman, now CI has been fixed, would you please take some time to review it? Sorry for suspending this PR for too long...

ax3l:
HI @Vigilans - looking through unfixed problems with propose solutions, do you think it would be possible to rebase this pull request for a new round of review? Sorry for the long delay on this!

PR #1343, opened 2018-03-31
Initial meson port

jeandet:
Hi this work is in preliminary shape, it would allow to easily integrate pybind11 in projects built with [Meson](http://mesonbuild.com/index.html) using [Wrap](http://mesonbuild.com/Wrap-dependency-system-manual.html).
The PR isn't ready to be merged, please do not hesitate to give me feedbacks on this.
It would be awesome to get it merged once ready :).


wjakob:
The official build system for the test suite is and will remain CMake -- we just don't have the clock cycles for supporting myriads of other competing build systems. Sorry for rejecting the PR -- I appreciate that this must have been quite a bit of work.

PR #1345, opened 2018-04-02
[READY] Implement an enum_ property "name"

bstaletic:
The property returns the enum_ value as a string.
For example:

```
>>> import module
>>> module.enum.VALUE
enum.VALUE
>>> str(module.enum.VALUE)
'enum.VALUE'
>>> module.enum.VALUE.name
'VALUE'
```

This is actually the equivalent of Boost.Python "name" property.

@jagerman Pointed out #1122 as it is somewhat related. The reason I didn't propose this change there is because I believe that smaller changes are easier to both reason about and review.

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -488,6 +488,24 @@ The entries defined by the enumeration type are exposed in the ``__members__`` p
     >>> Pet.Kind.__members__
     {'Dog': Kind.Dog, 'Cat': Kind.Cat}
 
+The ``name`` property returns the name of the enum value as a unicode string.
+
+.. note::
+
+    It is also possible to use ``str(enum)``, however these accomplish different
+    goals. The following shows how these two approaches differ.
+
+    .. code-block:: pycon
+
+        >>> p = Pet( "Lucy", Pet.Cat )
+        >>> pet_type = p.type
+        >>> pet_type
+        Pet.Cat
+        >>> str(pet_type)
+        'Pet.Cat'
+        >>> pet_type.name
+        'Cat'
+
 .. note::
 
     When the special tag ``py::arithmetic()`` is specified to the ``enum_``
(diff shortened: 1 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -1381,6 +1382,13 @@ template <typename Type> class enum_ : public class_<Type> {
             }
             return pybind11::str("{}.???").format(name);
         });
+        def_property_readonly("name", [m_entries_ptr](Type value) -> pybind11::str {
+            for (const auto &kv : reinterpret_borrow<dict>(m_entries_ptr)) {
+                if (pybind11::cast<Type>(kv.second[int_(0)]) == value)
+                    return pybind11::str("{}").format(kv.first);
```
jagerman: Going through `format` here is unnecessary: this can just be: `return pybind11::str(kv.first);`

bstaletic:
I have added the documentation and addressed the review comment.

docs/classes.rst
```diff
@@ -488,6 +488,24 @@ The entries defined by the enumeration type are exposed in the ``__members__`` p
     >>> Pet.Kind.__members__
     {'Dog': Kind.Dog, 'Cat': Kind.Cat}
 
+The ``name`` property returns the name of the enum value as a unicode string.
+
+.. note::
+
+    It is also possible to use ``py::str(enum)``, however these accomplish different
```
jagerman: Typo: that should just be `str(enum)` (i.e. no `py::`), since this is a Python reference.

jagerman:
One minor doc fix, but then it looks good to me.

One other thing to add: Python 3 (or Python 2 enum34) support the `.name` property as well, so this addition makes sense to me. (It also fits with #781 assuming the current `py::enum_` remains the fallback when `enum34` isn't available).

bstaletic:
The `py::` has been removed.
This shows that I shouldn't be making PR's after having not slept the previous night.

PR #1346, opened 2018-04-02
Fix missing word typo

patrikhuber:
I think that there's the word "for" missing for that sentence to be correct.
Please double-check that the sentence means what it's supposed to mean. :-)


jagerman:
Good catch, thanks for the PR.

PR #1347, opened 2018-04-03
Update C++ macros for C++17 and MSVC Z mode

henryiii:
Adds:

* Correct value for C++17
* Initial value for C++20
* Support for the new MSVC `/Zc:__cplusplus` mode (see [this forum, final post](https://developercommunity.visualstudio.com/content/problem/139261/msvc-incorrectly-defines-cplusplus.html))

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -27,15 +27,16 @@
 #  endif
 #endif
 
-#if !defined(_MSC_VER) && !defined(__INTEL_COMPILER)
+#if !(defined(_MSC_VER) && __cplusplus == 199711L) && !defined(__INTEL_COMPILER)
 #  if __cplusplus >= 201402L
 #    define PYBIND11_CPP14
-#    if __cplusplus > 201402L /* Temporary: should be updated to >= the final C++17 value once known */
+#    if __cplusplus >= 201703L
 #      define PYBIND11_CPP17
 #    endif
 #  endif
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && __cplusplus == 199711L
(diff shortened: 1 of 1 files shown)


henryiii:
I can fix the trailing space error after #1348 is accepted, so that the build will pass.

include/pybind11/detail/common.h
```diff
@@ -27,19 +27,26 @@
...
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && __cplusplus == 199711L
 // MSVC sets _MSVC_LANG rather than __cplusplus (supposedly until the standard is fully implemented)
+// Unless you use the /Zc:__cplusplus flag on Visual Studio 2017 15.7 Preview 3 or newer
 #  if _MSVC_LANG >= 201402L
 #    define PYBIND11_CPP14
 #    if _MSVC_LANG > 201402L && _MSC_VER >= 1910
 #      define PYBIND11_CPP17
+#      if __MSVC_LANG > 201703L && _MSC_VER >= 1910
+#        define CLI11_CPP20
```
jagerman: `PYBIND11_CPP20`?

henryiii: Oops, copy and paste error.

henryiii:
I think it's best to make sure the feature works, at least, if it provided. I agree that `has_include` and feature macros work, but there are cases where they are not used properly (experimental optional does not always define `__cpp_lib_experimental_optional`, for example) or are not avaiable (especially before c++17).

I'll drop C++20, but corrected C++17 and better MSVC support is still probably useful.

wjakob:
Looks good -- merged.

PR #1348, opened 2018-04-03
Fix for conda failures on Windows

henryiii:
PRs seem to be failing, this line should fix that.


jagerman:
Thanks for looking into this, that failure has been annoying for a while!

PR #1352, opened 2018-04-08
strdup is deprecated in MSVC. Use the ISO C++ conformant _strdup instead

zidnae:
https://msdn.microsoft.com/en-us/library/ms235454.aspx

strdup is deprecated in Visual Studio 2015 and 2017. _strdup is the ISO C++ conformant replacement provided by Microsoft.

I was trying to compile the Pet class example from the documentation using Clang for Windows and was getting a linker error for an undefined symbol "strdup". Adding this define fixed it.

Was using Clang version 5.0 installed via conda using the conda-forge package. I was using the standard library and linker that was packaged with Visual Studio 2017 community edition.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -11,7 +11,7 @@
 #pragma once
 
 #if defined(_MSC_VER)
-#  define strdup _strdup
+#  define strdup _strdup         //The POSIX strdup is deprecated. Use the ISO C++ conformant _strdup instead
 #  pragma warning(push)
 #  pragma warning(disable: 4100) // warning C4100: Unreferenced formal parameter
 #  pragma warning(disable: 4127) // warning C4127: Conditional expression is constant
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #if defined(_MSC_VER)
+#  define strdup _strdup
 #  pragma warning(push)
 #  pragma warning(disable: 4100) // warning C4100: Unreferenced formal parameter
 #  pragma warning(disable: 4127) // warning C4127: Conditional expression is constant


jagerman:
@wjakob - thoughts on replacing the `char *`s inside `function_record` with `std::string`s?  (i.e. letting `std::string` do the `strdup`s)

wjakob:
An early pre-release version of pybind11 used std::strings everywhere, and it was the resulting object code bloat that had me switch to strdup wherever possible (there is non-zero overhead that goes beyond the constructor call). IIRC just that gave on the order of a 10% reduction for the example bindings back then.

I'm very familiar with the MSVC security warnings, and this one is particularly idiotic (the referenced _strdup "ISO standard" is a standard that only Microsoft implements).

My suggestion would be to just apply the usual workarounds, which is to disable these warnings.
```
add_definitions(/D "_CRT_SECURE_NO_WARNINGS" /D "_SCL_SECURE_NO_WARNINGS")
```
This could be added to the ``pybind11_module`` CMake function.

Thoughts?

jagerman:
Yeah, the warning does seem idiotic.  Can we force the warnings off within just pybind via a `#pragma`?

jagerman:
Re-reading the original: it doesn't seem like this is just a warning, but there's an actual linker error; are those flags sufficient to provide the function?

zidnae:
The issue is a linker issue that occurs when compiling with Clang using Microsoft's standard library and linker. For the given example
```cpp

int add(int i, int j) {
    return i + j;
}

PYBIND11_MODULE(example, m) {

    m.def("add", &add, "A function which adds two numbers");
}
```

The following command:
`clang++ -O3 -Wall -shared -std=c++14 -v -LC:\Users\Jamie\Anaconda3\libs -IC:\Users\Jamie\Anaconda3\Include -IC:\Users\Jamie\AppData\Roaming\Python\Python36\Include  example.cpp -o example.pyd`

Produces the following error:
```bash
clang version 5.0.0 (tags/RELEASE_500/final)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Users\Jamie\Anaconda3\Library\bin
 "C:\\Users\\Jamie\\Anaconda3\\Library\\bin\\clang++.exe" -cc1 -triple x86_64-pc-windows-msvc19.13.26129 -emit-obj -mincremental-linker-compatible -disable-free -disable-llvm-verifier -discard-value-names -main-file-name example.cpp -mrelocation-model pic -pic-level 2 -mthread-model posix -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 -momit-leaf-frame-pointer -v -dwarf-column-info -debugger-tuning=gdb -resource-dir "C:\\Users\\Jamie\\Anaconda3\\Library\\lib\\clang\\5.0.0" -I "C:\\Users\\Jamie\\Anaconda3\\Include" -I "C:\\Users\\Jamie\\AppData\\Roaming\\Python\\Python36\\Include" -internal-isystem "C:\\Users\\Jamie\\Anaconda3\\Library\\lib\\clang\\5.0.0\\include" -internal-isystem "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Tools\\MSVC\\14.13.26128\\include" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\Include\\10.0.16299.0\\ucrt" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.16299.0\\shared" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.16299.0\\um" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.16299.0\\winrt" -O3 -Wall -std=c++14 -fdeprecated-macro -fdebug-compilation-dir "C:\\Users\\Jamie\\pybind11" -ferror-limit 19 -fmessage-length 317 -fms-extensions -fms-compatibility -fms-compatibility-version=19.13.26129 -fdelayed-template-parsing -fobjc-runtime=gcc -fcxx-exceptions -fexceptions -fdiagnostics-show-option -fcolor-diagnostics -vectorize-loops -vectorize-slp -o "C:\\Users\\Jamie\\AppData\\Local\\Temp\\example-0e1e64.o" -x c++ example.cpp
clang -cc1 version 5.0.0 based upon LLVM 5.0.0 default target x86_64-pc-windows-msvc
ignoring nonexistent directory "C:\Users\Jamie\AppData\Roaming\Python\Python36\Include"
#include "..." search starts here:
#include <...> search starts here:
 C:\Users\Jamie\Anaconda3\Include
 C:\Users\Jamie\Anaconda3\Library\lib\clang\5.0.0\include
 C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.13.26128\include
 C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\ucrt
 C:\Program Files (x86)\Windows Kits\10\include\10.0.16299.0\shared
 C:\Program Files (x86)\Windows Kits\10\include\10.0.16299.0\um
 C:\Program Files (x86)\Windows Kits\10\include\10.0.16299.0\winrt
End of search list.
 "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Tools\\MSVC\\14.13.26128\\bin\\HostX64\\x64\\link.exe" -out:example.pyd -defaultlib:libcmt "-libpath:C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Tools\\MSVC\\14.13.26128\\lib\\x64" "-libpath:C:\\Program Files (x86)\\Windows Kits\\10\\Lib\\10.0.16299.0\\ucrt\\x64" "-libpath:C:\\Program Files (x86)\\Windows Kits\\10\\Lib\\10.0.16299.0\\um\\x64" "-libpath:C:\\Users\\Jamie\\Anaconda3\\libs" -nologo -dll -implib:example.lib "C:\\Users\\Jamie\\AppData\\Local\\Temp\\example-0e1e64.o"
   Creating library example.lib and object example.exp
example-0e1e64.o : warning LNK4217: locally defined symbol __std_terminate imported in function "?dtor$40@?0?PyInit_example@4HA" (?dtor$40@?0?PyInit_example@4HA)
example-0e1e64.o : warning LNK4217: locally defined symbol _CxxThrowException imported in function "void __cdecl pybind11::pybind11_fail(char const *)" (?pybind11_fail@pybind11@@YAXPEBD@Z)
example-0e1e64.o : error LNK2019: unresolved external symbol strdup referenced in function "protected: void __cdecl pybind11::cpp_function::initialize_generic(struct pybind11::detail::function_record *,char const *,class type_info const * const *,unsigned __int64)" (?initialize_generic@cpp_function@pybind11@@IEAAXPEAUfunction_record@detail@2@PEBDPEBQEBVtype_info@@_K@Z)
example.pyd : fatal error LNK1120: 1 unresolved externals
clang++.exe: error: linker command failed with exit code 1120 (use -v to see invocation)
```

Adding the define gets rid of the error and allows the code to compile (though the two linker warnings still remain)

Issue #1317, opened 2018-03-11
Motivation for "type does not have a non-default holder type while its base does"?


BorisSchaeling:
What's the motivation for the `if (default_holder != base_info->default_holder)` check in [add_base()](https://github.com/pybind/pybind11/blob/15f36d2b2d07a8ce00dcc00cfe40770dec9ec53b/include/pybind11/attr.h#L262)?

I'm asking as I came across this check when I tried to pybind this C++ code: 

    class Base
    {
    protected:
      virtual ~Base() = default;
    };

    class Derived : public Base
    {
    };

As the Base class has a protected destructor, I used [nodelete](http://pybind11.readthedocs.io/en/stable/advanced/classes.html?highlight=nodelete#non-public-destructors): 

    PYBIND11_MODULE(Test, m) {
      class_<Base, std::unique_ptr<Base, nodelete>> base{ m, "Base" };
      class_<Derived, Base> derived{ m, "Derived" };
    }

The code compiles but when you try to import the module you get this error message because the if-check fails: 

> ImportError: generic_type: type "Derived" does not have a non-default holder type while its base "Base" does

Now I could just use for example std::shared_ptr as a holder type for the Derived class to satisfy the condition (as then Derived uses a non-default holder type, too). But I think I better understand what pybind11 is trying to protect me from with this if-check. Or does the if-check make no sense in my use case (as it's actually meant to protect users in another use case)?

trelau:
I'm also curious about this. I ended up defining a ``Deleter`` template to use in the ``Derived`` classes like:

```cpp
template<typename T> struct Deleter { void operator() (T *o) const { delete o; } };

class_<Derived, std::unique_ptr<Derived, Deleter<Derived>>,  Base> derived{ m, "Derived" };
```

Not sure if this is the best approach, but it seems to "work" on the Python side.

jagerman:
The motivation is to catch a fairly common case where someone accidentally tries to use a `std::shared_ptr<Derived>` with a `std::unique_ptr<Base>` (or vice versa)—we can't cast a derived holder to a base holder in such a case, which breaks our ability to handle inheritance.

[...]

PR #1353, opened 2018-04-08
Enable unique_ptr holder with mixed Deleters between base and derived types

trelau:
Recognize "std::unique_ptr<T, D>" as a default holder even if "D" doesn't match between base and derived types. Currently, if "D" doesn't match you get a "type does not have a non-default holder type while its base does" error.

Reference this issue: https://github.com/pybind/pybind11/issues/1317

And this solution: https://github.com/pybind/pybind11/issues/1317#issuecomment-378088064

An example where this will remove some boilerplate code is described [here](http://pyocct.readthedocs.io/en/latest/dev.html#non-public-destructors).

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1069,7 +1069,7 @@ public:
         record.holder_size = sizeof(holder_type);
         record.init_instance = init_instance;
         record.dealloc = dealloc;
-        record.default_holder = std::is_same<holder_type, std::unique_ptr<type>>::value;
+        record.default_holder = detail::is_instantiation<std::unique_ptr, holder_type>::value;
 
         set_operator_new<type>(&record);
 
--- a/tests/test_smart_ptr.cpp
+++ b/tests/test_smart_ptr.cpp
@@ -186,6 +186,32 @@ TEST_SUBMODULE(smart_ptr, m) {
         .def(py::init<int>())
         .def_readwrite("value", &MyObject4::value);
 
+    // test_unique_deleter
+    // Object with std::unique_ptr<T, D> where D is not matching the base class
+    // Object with a protected destructor
+    class MyObject4a {
+    public:
+        MyObject4a(int i) {
+            value = i;
+            print_created(this);
+        };
+        int value;
+    protected:
+        virtual ~MyObject4a() { print_destroyed(this); }
+    };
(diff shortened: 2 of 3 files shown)


jagerman:
LGTM.  As I mentioned in #1317, this better matches the intentions of the check (which is really just meant to keep you from accidentally mixing `unique_ptr`s and `shared_ptr`s).

EricCousineau-TRI:
One potential concern: What happens when you have a non-empty deleter that actually changes the memory layout of `unique_ptr`? It seems like this now enables you to mix `unique_ptr`s along the inheritance chain, which could cause a segfault without some more `typeid` checks (e.g. #1161)?

jagerman:
I don't think it's a problem since `unique_ptr<T>`s aren't castable to `unique_ptr<Base>`/`unique_ptr<Derived>` in the first place.  The check here is to make sure that you don't mix holders that can cast to base/derived holders up and down the inheritance chain (i.e. `shared_ptr`) with holders that can't (i.e. `unique_ptr`).

EricCousineau-TRI:
Just did a quick test on this branch:
https://github.com/EricCousineau-TRI/pybind11/pull/1/files#diff-2afbeba333b4cc339dabd6a379ea943cR155

Takeaways:
* Generally, static compilation checks for castability between the `Deleter`s of `unique_ptr` will ensure users don't carelessly cast `unique_ptr`s.
* If a user starts mixing `unique_ptr`s and, for whatever reason, uses a nontrivial deleter (something that takes up memory, e.g. `std::function<>` for type-erased destructors) at somewhere along the inheritance chain, they may be in for a nasty surprise if they are not consistent with their API.

The above caveat seems like a rare edge case, and would seem like a general design mistake, so I think this PR is fine as-is; could be addressed at a later point, perhaps with #1161.

trelau:
Rebased on master. Any idea why a single configuration is failing for xcode?

trelau:
Rebased on master and passed all builds. Ready for review @jagerman @wjakob

justbuchanan:
+1 for this PR. What else needs to be done before it can be merged? I'de be happy to help if possible.

trelau:
anything in particular holding back this PR? happy to help if there are any requests.

wjakob:
It looks good to me. The reason for the delay is my (very) limited time budget for maintaining this project.

PR #1356, opened 2018-04-10
Update numpy.rst

thomashrabe:
Changing from auto to py::buffer_info is more intuitive. It took me few hours to understand the missing type was what I was looking for... Maybe replacing all auto typing with their real type is better for the tutorial code.


wjakob:
Looks good, thanks!

lrigazio:
your Matrix example has possibly an issue ~ i had to swap rows with cols in buffer_info to make it work (it was crashing in py ~ and giving wrong strides)

>> { sizeof(float) * m.rows(),             /* Strides (in bytes) for each index */

change to

>> {sizeof(float) * m.cols(),              /* Strides (in bytes) for each index */

Issue #1113, opened 2017-09-26
Exceptions in module initialization should be propagated


anntzer:
## Issue description

Currently, Python exceptions triggered during module initialization cause program termination due to an uncaught `pybind11::error_already_set`.  For example, adding
```
py::object s = py::cast("a"); py::print(s.attr("foo"));
```
to the module initialization code in `python_example/src/main.cpp`, building, and trying to import `python_example` causes the above-mentioned issue.

Compare with adding such code e.g. to the `add` function in the same file; then, calling `add()` results in a normal AttributeError being thrown at the Python level.

Likewise, in pure Python, errors thrown at import time are reported normally (e.g. `a.py`: `import b`; `b.py`: `"a".foo` gives an AttributeError when importing `a`).

If it is technically difficult to rethrow the exception (I am not familiar with the internals of module initialization), it would at least be helpful to reproduce the text (the `__str__`, I guess) of the inner exception to stderr, to help with debugging.

## Reproducible example code

See above.

uwydoc:
confirmed on v2.2.x and master, no such problem with v2.1.x.

the import actually result in crash, so no message is printed. the crash comes from the destructor of `error_already_set` instance, during the [construction of gil object](../blob/v2.2/include/pybind11/pybind11.h#L1861), i cannot figure out why, i think it was caused by wrong order of resource releasing.

anyway, after referring other places where `error_already_set` is used, and the implementation of `PYBIND11_PLUGIN` in v2.1.x, `e.restore` is invoked in (almost) all exception handling codes of `error_already_set`. i think `e.restore()` is missing before line \#290 in file [detail/common.h](../blob/v2.2/include/pybind11/detail/common.h#L290). after adding it, proper error message is printed.

i think it is a trivial bug and needs confirm.

ps. `error_already_set::clear` is deprecated in v2.2.x, it seems `restore` is the _proper_ method to invoke. however, `restore` just restore the previous saved exception without clearing it with `PyErr_Clear`, so i am not sure whether `clear` is ok deprecate.

PR #1362, opened 2018-04-17
Properly report exceptions thrown during module initialization.

anntzer:
If an exception is thrown during module initialization, the
error_already_set destructor will try to call `get_internals()` *after*
setting Python's error indicator, resulting in a `SystemError: ...
returned with an error set`.

Fix that by temporarily stashing away the error indicator when
initializing the internals.

(I *think* any proper unit test would require creating a new module just for that purpose, and I'm not particularly familiar with the testing machinery used by pybind, so guidance would be welcome here.  But "it works for me"...)

xref #1113.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1894,6 +1894,7 @@ class gil_scoped_release { };
 
 error_already_set::~error_already_set() {
     if (type) {
+        error_scope scope;
         gil_scoped_acquire gil;
         type.release().dec_ref();
         value.release().dec_ref();


jagerman:
I'm having trouble reproducing the error; do you have a simple example code that fails?

jagerman:
I think the `error_scope` fix here is good, but is in the wrong place--it should instead, I think, be backed up one level, into the `error_already_set` destructor rather than `get_internals`.  That is, adding the error scope here instead:

```C++
error_already_set::~error_already_set() {
    if (type) {
        error_scope scope;
        gil_scoped_acquire gil;
        type.release().dec_ref();
        value.release().dec_ref();
        trace.release().dec_ref();
    }
}
```

It works with your example in #1113; do you want to test it out some more (and update the PR if it all checks out)?

anntzer:
It's not tally clear to me why it'd make more sense to stash it in this scope instead, but sure, it works.  Force-pushed the fix as suggested.

jagerman:
> It's not totally clear to me why it'd make more sense to stash it in this scope instead, but sure, it works. Force-pushed the fix as suggested.

Well, it seems to me that the issue here isn't that `get_internals()` is doing something wrong, but rather that `error_already_set` is setting an error but then making a call that can lead to Python code (via internals).

PR #1367, opened 2018-04-20
pybind11Tools: Use interface target library when able

EricCousineau-TRI:
It'd be nice if the interface target libraries were used via `pybind11_add_module`.

\cc @jamiesnape


henryiii:
Fixed as part of the CMake refactor. :)

PR #1368, opened 2018-04-20
WIP: Fix iostream when used with nogil

henryiii:
When using the iostream wrappers, if you release the gil, you get a segfault. This small patch ensures that mixing them is safe.

Edit: Waiting on #1286, or #1211, or #1556 (probably more)


wjakob:
LGTM.

PR #1369, opened 2018-04-20
Fix pip10 buggy behavoir

henryiii:
`pip` in Pip10 is broken. You can't use `pip` on windows, but have to do `python -m pip`. (Also true with Ubuntu and probably others). Pip officially is forcing `python -m pip` to do a pip upgrade on Windows (since it replaces the running executable - I assume it has to replace it because of the `import main` disaster...).

There's a not in PATH warning that turns into an error... (The warning is not a bad idea, the fact that it errors out on Appveyor is not so great). Added a flag to stop the error (we use `python -m pytest` so the warning is not needed anyway).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -41,8 +41,8 @@ install:
     if ($env:PYTHON) {
       if ($env:PLATFORM -eq "x64") { $env:PYTHON = "$env:PYTHON-x64" }
       $env:PATH = "C:\Python$env:PYTHON\;C:\Python$env:PYTHON\Scripts\;$env:PATH"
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Resolves #1372

(Sorry, hadn't seen this prior; thanks for pointing it out, @jamiesnape!)

henryiii:
Note that this is still using Pip 10, just fixing the things that are no longer supported in Pip 10. Like the pip command... 😡 🙄

wjakob:
Thanks for the fix!

Issue #1370, opened 2018-04-20
SEGV to create numpy ndarray with empty shape


sonots:
## Issue description

SEGV occurs to create numpy ndarray with empty shape.

## Reproducible example code

```
py::array(py::dtype("f"), {}, {});
```

SEGV occurs at https://github.com/pybind/pybind11/blob/060936fed2ad0bc14251a8703b6139eb7ea62a25/include/pybind11/numpy.h#L762

It looks the reason is because `size_t i = ndim - 1` at the above line

https://github.com/pybind/pybind11/blob/060936fed2ad0bc14251a8703b6139eb7ea62a25/include/pybind11/numpy.h#L761

overflows when `ndim` is 0, and `i` becomes `18446744073709551615`.

sonots:
workaround:

```
py::array py_array = py::array(py::dtype("f"), {1}, {});
py_array.resize({});
```

PR #1371, opened 2018-04-20
Fix SEGV to create empty shaped numpy array

sonots:
Fix https://github.com/pybind/pybind11/issues/1370


jagerman:
Looks good.  Thanks for the PR.

PR #1373, opened 2018-04-20
Ensure PythonLibsNew_FOUND is set in FindPythonLibsNew module

jamiesnape:
Since the module is named `FindPythonLibsNew`, `PythonLibsNew_FOUND` should be set appropriately.


henryiii:
This is also useful if you use CMake's FeatureSummary module; currently it misreports about PythonLibsNew because `PythonLibsNew_FOUND` is not set. 👍

EricCousineau-TRI:
@wjakob @jagerman Any chance we might be able to land this?

wjakob:
LGTM!

PR #1376, opened 2018-04-24
Typo

davidcaron:
Small function name typo in docs


jagerman:
Good catch, thanks for the PR.

PR #1384, opened 2018-04-29
Misc. typos

luzpaz:
Found via `codespell`


jagerman:
Thanks for the PR.  (Looking over your diff I noticed one more ("trempoline" -> "trampoline") that I pushed to your branch.)

PR #1387, opened 2018-05-04
Simplified example allowing more robust usage

tdegeus:
* Simplified example, also allowing more robust usage
* Fixed minor spelling issues


jagerman:
Thanks for the PR.  Looks like the documentation didn't get updated from when we moved superclasses into the templates.

PR #1394, opened 2018-05-11
add signed overload for `py::slice::compute`

### LINKED ISSUE #1390: py::stride::compute does not accept signed type
`py::stride` used to have `size_t` to store start/stop/step values. Nowadays it uses `ssize_t` (issue #776, PR #782). 

The issues is that `py::slice::compute` still accepts `size_t*` arguments instead of `ssize_t*`

https://github.com/pybind/pybind11/blob/086d53e8c66a84d0ec723d5435918c76edd878e8/include/pybind11/pytypes.h#L1073-L1079

C-style pointers casts must be read here as `reinterpret_cast<>`.

To obtain signed strides user should have something like 

```C++
ssize_t start, stop, step, slicelength;
if (!slice.compute(
            N,
            reinterpret_cast<size_t*>(&start),
            reinterpret_cast<size_t*>(&stop),
            reinterpret_cast<size_t*>(&step),
            reinterpret_cast<size_t*>(&slicelength)
)){
}
... // use extracted values
```

So `reinterpret_cast<>` appears twice: in user code and in the library (in `py::slice::compute`) converting back and forth `ssize_t*` and `size_t*`.

The non-breaking solution would be to add an overload for `ssize_t*` to `py::slice`. 

```C++
bool compute(ssize_t length, ssize_t *start, ssize_t *stop, ssize_t *step, 
              ssize_t *slicelength) const { 
     return PySlice_GetIndicesEx((PYBIND11_SLICE_OBJECT *) m_ptr, 
                                 length, start, 
                                 stop,  step, 
                                 slicelength) == 0; 
 } 
```

sizmailov:
closes #1390

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1077,6 +1077,13 @@ public:
                                     (ssize_t *) stop, (ssize_t *) step,
                                     (ssize_t *) slicelength) == 0;
     }
+    bool compute(ssize_t length, ssize_t *start, ssize_t *stop, ssize_t *step,
+      ssize_t *slicelength) const {
+      return PySlice_GetIndicesEx((PYBIND11_SLICE_OBJECT *) m_ptr,
+          length, start,
+          stop, step,
+          slicelength) == 0;
+    }
 };
 
 class capsule : public object {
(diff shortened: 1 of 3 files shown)


wjakob:
LGTM, thanks!

PR #1405, opened 2018-05-22
Container: ipython completions

ax3l:
Add ipython 5+ completions for the `__getitem__` method in `dict` / C++ maps.

This adds interactive completion in ipython (and related kernels, e.g. in Notebooks).
From my tests, this currently works only for dictionaries with string keys and is ignored for all others.

### Example

https://github.com/openPMD/openPMD-api/pull/191

### Reference

https://ipython.readthedocs.io/en/5.x/config/integrating.html

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -593,6 +593,16 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
 
     cl.def("__len__", &Map::size);
 
+    // interactive ipython key completions, currently works with string keys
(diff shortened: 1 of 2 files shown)


wjakob:
IMO this is a bit of a fringe case -- it will increase the size of bindings for anybody who uses ``std::map``s while only addressing a fairly specific use case.

ax3l:
I agree, it's currently also only supported for strings in `ipython`.

Just wanted to share it for the interested reader, I am good to close this, users can add it on their own where it's helpful.

Issue #1406, opened 2018-05-22
FindPythonLibsNew.cmake not working with MSYS Python interpreter


fteicht:
FindPythonLibsNew.cmake tests for windows system (CMAKE_HOST_WIN32 is true) then it assumes that the Python interpreter works with windows-like directories hierarchy (lines 141-143).
However, it does not work when using MSYS' python interpreter which assumes a Unix-like directories hierarchy.
Note that compiling python bindings with MSYS' mingw compiler requires to use MSYS' python interpreter. For instance, anaconda's interpreter's include files do not need to be compatible with MSYS' compiler due to version mismatch.

xiaowen64:
this doesn't work on MSYS2

CMake Error at tools/FindPythonLibsNew.cmake:158 (message):
  Python libraries not found
Call Stack (most recent call first):
  tools/pybind11Tools.cmake:16 (find_package)
  CMakeLists.txt:33 (include)

samuelpmish:
same issue-- my underlying C++ code gets about a 2x speedup when compiling with gcc (via MinGW) versus visual studio, but I can't generate python bindings for it, because of these errors and incompatibility with MinGW.

What would need to change to support the MinGW toolchain on Windows?

wojdyr:
FTR MSYS2 is now packaging patched pybind11
https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-pybind11
(but it still doesn't work out of the box for me)

wjakob:
@wojdyr , @samuelpmish , @xiaowen64, @fteicht : would you be able to test @kanonet's PR #1851 and provide feedback on whether the issue is completely resolved after that?

kanonet:
What will happen if they do not answer? Will my fix get released? Is there anything I can do?

wjakob:
closed via #1851

PR #1407, opened 2018-05-22
Corrected bug in FindPythonLibsNew.cmake not working with MSYS

fteicht:
Solves the following issue:
https://github.com/pybind/pybind11/issues/1406

The trick consists in searching for python libraries first, which will work on linux and MSYS. Then, if it does not work and we are on a win32 host, assume a windows-like directories hierarchy and test for python library existence with such a hierarchy.


henryiii:
This was fixed about a year so, see the linked issue. Let us know if there's still a problem, especially after #2370 goes in!

PR #1413, opened 2018-05-26
Allow python builtins to be used as callbacks

davidhewitt:
At the moment when passing python builtins like `sum()` as callback functions to c++, the std::function caster will fail with `RuntimeError: Unable to extract capsule contents`. This is because of the machinery to extract stateless c++ function pointers to avoid the c++ -> python -> c++ overhead.

I've added a simple check that the definition being extracted is indeed a capsule and an accompanying test that I can now use `sum()` as a callback. Hope you find this useful.


jagerman:
The change looks good to me.  There's one minor change needed in the added python test to fix the flake warning (two blank lines needed before the added test).

davidhewitt:
Great, I'll find a moment to tidy that up at the weekend!

davidhewitt:
@jagerman should be all good now. Thanks

davidhewitt:
@jagerman just wanted to make sure this wasn't forgotten - I still run into this every so often.

TobiasBruell:
I just ran into the same problem. Many thanks!

lycantropos:
what about progress on this one?

YannickJadoul:
> what about progress on this one?

We're in the middle of cleaning up issues and PRs. Give us a little bit of time, please.

If you want to help this PR progress, fixing the merge conflicts would help us greatly! :-)

rwgk (approved):
Looks great! Is the original author still around? If someone rebases on current master and the CI comes back green I think we should merge this straightaway.

Skylion007:
@rwgk @henryiii I added an xfail for PyPy, but I would appreciate if you either of you could see if there is an easy fix / workaround we are missing. Like maybe wrapping it in a lambda or something?

include/pybind11/functional.h
```diff
@@ -43,20 +43,30 @@ struct type_caster<std::function<Return(Args...)>> {
...
+                    }
+                    rec = rec->next;
                 }
-                rec = rec->next;
+            } else {
+               // Usually indicates that it is a builtin function.
+#if            defined(PYPY_VERSION)
+               // PyPy will segfault otherwise when passing in raw builtin functions.
+               return false;
+               //pybind11_fail("Passing raw builtin functions not supported with PyPy. Wrap in function.");
```
Skylion007, line 68: Do a fail her also caused a segfault. Weird.

Skylion007:
Now it segfaults on every PyPy EXCEPT Ubuntu. So weird...

include/pybind11/functional.h
```diff
@@ -43,27 +43,37 @@ struct type_caster<std::function<Return(Args...)>> {
...
+                    }
+                    rec = rec->next;
                 }
-                rec = rec->next;
+            } else {
+               // Usually indicates that it is a builtin function.
+#if            defined(PYPY_VERSION)
+               // PyPy will segfault otherwise when passing in raw builtin functions.
+               // This will lead to a TypeError instead.
+               return false;
```
Skylion007, line 68: So it looks like this test case ALWAYS segfaulted on non-ubuntu PyPy then which means that that the code is still better than before. I am really stumped on how to avoid this segfault though.

include/pybind11/functional.h
```diff
@@ -43,27 +43,37 @@ struct type_caster<std::function<Return(Args...)>> {
...
+#endif
             }
         }
 
         // ensure GIL is held during functor destruction
         struct func_handle {
             function f;
-            func_handle(function&& f_) : f(std::move(f_)) {}
+            func_handle(function &&f_) noexcept : f(std::forward<function>(f_)) {}
```
Skylion007, line 76: These are bugs, should be forwarded. Not moved.

davidhewitt:
👋 original author still here! Sorry it's taken me a couple of days to reply. It looks like you've got this one under control without me, so I won't push anything unless you ask me to.

These days I don't use C++ much any more, I tend to use Rust instead. I'm still very much working in the Python + native domain though, I am a maintainer of [PyO3](https://github.com/PyO3/pyo3).

I still watch progress here closely; I like to think that both projects can benefit from sharing ideas and pushing the ecosystem forward. Sadly we're still some way behind in PyO3 so I don't think there's much learning from PyO3 that would interest you yet. 

Do feel free to ping me any time (on this issue or any other) if you want my input as a "neighbour" 😄

include/pybind11/functional.h
rwgk: I just counted, we have 19 xfail for PyPy already.

rwgk:
I imported this PR into the Google environment and ran ASAN & MSAN. Both are happy. I'm now submitting this PR to our global testing system. I'll report on the results tomorrow.

include/pybind11/functional.h
henryiii: PyPy has bugs. If we aren't adding a new one or making it worse (which it appears we are not), then I'm happy with it if it passes Google's testing.

rwgk:
Unfortunately there is a problem, triggered by code that's not open source, so I cannot easily point or share. It's a compilation failure (using some very modern clang version). I'll share the full traceback non-publicly. Here is the result of `grep functional` in the traceback:

```
pybind11/functional.h(90,74): error: use of undeclared identifier 'func_handle'
pybind11/functional.h(99,17): error: no matching conversion for functional-style cast from 'func_handle' to 'func_wrapper'
pybind11/functional.h(88,16): note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'func_handle' to 'const func_wrapper' for 1st argument
pybind11/functional.h(88,16): note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'func_handle' to 'func_wrapper' for 1st argument
pybind11/functional.h(88,16): note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided
pybind11/functional.h(99,17): error: no matching conversion for functional-style cast from 'func_handle' to 'func_wrapper'
pybind11/functional.h(88,16): note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'func_handle' to 'const func_wrapper' for 1st argument
pybind11/functional.h(88,16): note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'func_handle' to 'func_wrapper' for 1st argument
pybind11/functional.h(88,16): note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided
```

rwgk:
WAIT: after reporting the above I tried to reproduce the errors interactively, but it works!
I'm confused...

rwgk:
It turns out to be some kind of Windows build.
I never had such failures before and don't know how to reproduce interactively.

rwgk:
In the meantime the author of https://github.com/pybind/pybind11/pull/1883 gave me some pointers, including how to reproduce the error interactively. **Theory**: we could be up against a bug in the Windows clang compiler. I'll try to get the attention of the team maintaining that compiler.

include/pybind11/functional.h
Skylion007: Actually, since the type is deduced and copy-initialized in a member-initiailzer list, the original code does appear to be actually be correct: https://stackoverflow.com/a/31213816/2444240

rwgk: Oh good! I'll try this again with the clang windows compiler.

Skylion007:
I'll split off the bugprone checks into their own PR.

PR #1431, opened 2018-06-15
Fix auto type deduction failures when var name is an existing type name

### LINKED ISSUE #1381: Unable to compile with PGI C++ due to `auto` deduction failures
Hello,

The latest free version of PGI compilers was released today. I tried to compile with it a simple pybind11 program but I get a bunch of errors. By "simple program" I mean just including `<pybind11/pybind11.h>`. Here is the output with all the errors: https://pastebin.com/jeCNBwWx
All the errors are cause by lines where something is cast to `auto`. After going through the files and manually typing the types, I was able to compile the simple example found at http://pybind11.readthedocs.io/en/master/basics.html#creating-bindings-for-a-simple-function

The changes I performed are the following:

`include/pybind11/detail/class.h`:
multiple lines: replace `auto` in `reinterpret_cast<detail::instance *>` lines with `detail::instance*`
517: `auto` -> `object`
554: `auto` -> `PyObject*`
561: `auto` -> `PyTypeObject*`

`include/pybind11/detail/init.h`:
27: `auto` -> `pybind11::detail::descr<1UL, pybind11::detail::value_and_holder>`

`include/pybind11/pybind11.h`:
741: `auto` -> `dict`

It appears that the PGI C++ compiler fails to deduct types when using the `auto` keyword in some places. To me, the pybind11 code I edited seems standard C++, but at the same time I am not a C++ expert. Can someone verify that the problem here is the PGI compiler and not some non-standard C++ used in pybind11?

Best Regards,
George

chuckatkins:
Fixes #1381 

`name` is certainly the messiest and most pervasive of the bunch here.  I'm happy to change any of the variable names as needed.

While this does address the PGI type deduction issues, pybind11 still won't correctly build with PGI without #1428.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -814,7 +814,7 @@ template <typename type> class type_caster_base : public type_caster_generic {
     using itype = intrinsic_t<type>;
 
 public:
-    static constexpr auto name = _<type>();
+    static constexpr auto name_ = _<type>();
 
     type_caster_base() : type_caster_base(typeid(type)) { }
     explicit type_caster_base(const std::type_info &info) : type_caster_generic(info) { }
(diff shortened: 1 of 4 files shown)


jagerman:
It seems from the CI builds this still needs some work.

chuckatkins:
Indeed.  I thought I had it but i pushed premature and I've got a bit more churning on this to do.  Since it might be a few days before I finish it I'll close this PR for now and reopen it after I test on the necessary platforms

jagerman:
Feel free to leave it open, it doesn't hurt anything.  Just ping me when it's ready and I'll take a look.

PR #1434, opened 2018-06-20
CMake: Remember Python Version

ax3l:
It is useful not only to remember the python libs and includes but also the interpreter version in cache.

If users call pybind11 throught `add_subdirectories` they will otherwise have no access to the selected interpreter version. The interpreter version is useful for downstream projects, e.g. to select default `lib/pythonX.Y/site-packages/` install paths.

Fix https://github.com/pybind/cmake_example/issues/13


ax3l:
cc @dean0x7d @SylvainCorlay

ax3l:
Is it possible the python 3.6.6 is broken in `dev`, too?

ax3l:
@jagerman is the CI broken on Python 3.6.6?

ax3l:
@chuckatkins any thoughts on this one as well?

ax3l:
@jagerman @dean0x7d would be glad if we can improve in-source pybind11 options with this addition :)

PR #1435, opened 2018-06-20
Use std::addressof to obtain holder address instead of operator&

tyler92:
Usage of `operator&` for custom smart pointers leads to unexpected behaviour. In my case segmentation fault occurs when i try to run code:

```
// C++ code

template<class T>
class my_custom_ptr
{
    // ...
    T** operator&() { /* return address of internal pointer */ }
    // ...
};

my_custom_ptr<some_object> get_object()
{
    return my_custom_ptr<some_object>( new some_object );
}

PYBIND11_MODULE(module, m)
{
    m.def( "get_object", &get_object )
    py::class_<some_object, my_custom_ptr<some_object>>( m, "some_object" );
}

// Python code
import module
obj = module.get_object()
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1423,7 +1423,7 @@ public:
 
     explicit operator type*() { return this->value; }
     explicit operator type&() { return *(this->value); }
-    explicit operator holder_type*() { return &holder; }
+    explicit operator holder_type*() { return std::addressof(holder); }
 
     // Workaround for Intel compiler bug
     // see pybind11 issue 94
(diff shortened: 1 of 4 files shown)


jagerman:
The fix looks good to me; this is indeed an oversight.  Can you turn your example into a concrete test (ideally one that throws an exception if the `operator &` gets called rather than segfaulting!) and add it to `test_smart_ptr` in the test suite?

tyler92:
Yes, of course. It will take some time.

tyler92:
Done!

jagerman:
Merged, thanks for the PR!

PR #1442, opened 2018-06-29
Untangle cast logic to not implicitly require castability

DennisOSRM:
The current code requires implicitly that integral types are cast-able to floating point. In case of strongly-typed integrals (e.g. [as explained here](http://www.ilikebigbits.com/blog/2014/5/6/type-safe-identifiers-in-c)) this is not always the case.

The patch at hand uses compile-time magic to make sure that cast-ability is only requested if the type supports it.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1028,6 +1028,12 @@ public:
     }
 
     template<typename U = T>
+    static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, short>::value, handle>::type
+    cast(U src, return_value_policy /* policy */, handle /* parent */) {
+        return PyLong_FromLong((long) src);
+    }
+
+    template<typename U = T>
     static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, int>::value, handle>::type
     cast(U src, return_value_policy /* policy */, handle /* parent */) {
         return PyLong_FromLong((long) src);
@@ -1041,6 +1047,12 @@ public:
 
     template<typename U = T>
     static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, unsigned char>::value, handle>::type
+    cast(U src, return_value_policy /* policy */, handle /* parent */) {
+        return PyLong_FromUnsignedLong((unsigned long) src);
+    }
+
+    template<typename U = T>
+    static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, unsigned short>::value, handle>::type
     cast(U src, return_value_policy /* policy */, handle /* parent */) {
         return PyLong_FromUnsignedLong((unsigned long) src);
     }
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1009,21 +1009,51 @@ public:
         return true;
     }
 
-    static handle cast(T src, return_value_policy /* policy */, handle /* parent */) {
-        if (std::is_floating_point<T>::value) {
-            return PyFloat_FromDouble((double) src);
-        } else if (sizeof(T) <= sizeof(ssize_t)) {
-            // This returns a long automatically if needed
-            if (std::is_signed<T>::value)
-                return PYBIND11_LONG_FROM_SIGNED(src);
-            else
-                return PYBIND11_LONG_FROM_UNSIGNED(src);
-        } else {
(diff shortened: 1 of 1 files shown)


include/pybind11/cast.h
```diff
@@ -1009,21 +1009,76 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
-            if (std::is_signed<T>::value)
-                return PyLong_FromLongLong((long long) src);
-            else
-                return PyLong_FromUnsignedLongLong((unsigned long long) src);
-        }
+    template <class A, class B>
+    struct is_explicitly_convertible
+    {
+        enum {value = std::is_same<A, B>::value || (std::is_constructible<B, A>::value && !std::is_convertible<A, B>::value)};
+    };
```
jagerman: This struct can be simplified considerably:
```C++
template <class A, class B> using is_explicitly_convertible = bool_constant<
    std::is_same<A, B>::value || (std::is_constructible<B, A>::value && !std::is_convertible<A, B>::value>;
```

DennisOSRM: `std::bool_constant` is a relatively new addition in C++17 and would require a break. Nevertheless, it is a great suggestion and really valuable feedback. I went ahead and used `std::integral_constant<bool, ..>` instead in 416a43f.

include/pybind11/cast.h
```diff
@@ -1009,21 +1009,76 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
+    static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, long long>::value, handle>::type
+    cast(U src, return_value_policy /* policy */, handle /* parent */) {
+        return PyLong_FromLongLong((long long) src);
+    }
+
+    template<typename U = T>
+    static typename std::enable_if<!std::is_floating_point<U>::value && is_explicitly_convertible<U, unsigned long long>::value, handle>::type
+    cast(U src, return_value_policy /* policy */, handle /* parent */) {
+        return PyLong_FromUnsignedLongLong((unsigned long long) src);
     }
```
jagerman: Splitting this up into an exhaustive list of types doesn't strike me as a good idea: aside from being easy to miss some in the standard (for example, you're missing `signed char`, `char16_t`, `char32_t`, `wchar_t`), it also excludes custom types that a compiler might support or a third-party library might add `is_integral` specializations for.

I'm also having a bit of trouble seeing what this accomplishes (a test case to demonstrate the current failure would help).  The referenced type-safe-handles type doesn't seem immediately applicable to me as it doesn't seem like it will be caught in this code at all (this is already inside a `std::is_arithemetic` type caster specialization).

include/pybind11/cast.h
DennisOSRM: > But I'm curious why the not-implicitly-convertible check is needed here.

It makes sure the overload doesn't become ambiguous.

include/pybind11/cast.h
DennisOSRM: > Splitting this up into an exhaustive list of types doesn't strike me as a good idea: aside from being easy to miss some in the standard (for example, you're missing signed char, char16_t, char32_t, wchar_t), it also excludes custom types that a compiler might support or a third-party library might add is_integral specializations for.

Agree, an exhaustive list is not beautiful at all and perhaps it is not complete. Support for signed char should be there, though.

> I'm also having a bit of trouble seeing what this accomplishes (a test case to demonstrate the current failure would help). The referenced type-safe-handles type doesn't seem immediately applicable to me as it doesn't seem like it will be caught in this code at all (this is already inside a std::is_arithemetic type caster specialization).

The current code in master branch requires any type to be cast-able to floating point as well as to castable to an integral type. Otherwise the code is malformed and won't compile. Arguably this patch can be improved, but it uses a SFINAE-like approach to make sure types don#t have to cast-able to fp and int numbers.

DennisOSRM:
@jagerman Please have a look at the latest revision. It should simplify things a lot.

DennisOSRM:
friendly ping

jagerman:
The change looks good to me.  I'll merge tomorrow if no one else has any comments.

DennisOSRM:
cool, thanks @jagerman

DennisOSRM:
friendly ping

jagerman:
Oops, sorry, got tired up with other things and forgot about this.  Merged; thanks for the PR!

PR #1447, opened 2018-07-05
Fix auto type deduction failures when var name is an existing type name

### LINKED ISSUE #1381: Unable to compile with PGI C++ due to `auto` deduction failures

chuckatkins:
This is the continuation of #1431 and fixes #1381.

While this does address the PGI type deduction issues, pybind11 still won't correctly build with PGI without #1428.


chuckatkins:
Noite that once #1428 is merged I'll rebase this and it should be ready to go.

jagerman:
Have you reported this bug to PGI?  We can work around it, of course, but it's definitely a compiler bug.

wjakob:
Hi @chuckatkins,

I've merged the first commit (a899dd0). The second (Utilize CMake's language standards abstraction when possible) is superseded by a larger PR that is still in the queue.

Regarding the third (work around PGI compiler bugs by making changes throughout pybind11) -- sorry, no :(. If this was one of the major 3 (GCC, Clang, MSVC), I would be more open to this sort of thing.

Best,
Wenzel

ax3l:
@chuckatkins can you confirm [PGI 18.7+](https://github.com/pybind/pybind11/issues/1381#issuecomment-411805050) fixed the compiler bug?

PR #1448, opened 2018-07-05
DEN-4895 when cross-compiling from a 64-bit host to a 32-bit target, …

caseykelso:
…this check fails as the host has a 64-bit python interpreter but the compiler is 32-bit


wjakob:
Sorry, this PR is just too messy to be considered in its current form (and it hardcodes things like the Python 2.7 lib location)

Issue #1451, opened 2018-07-11
Defining enum with mutiple identical values should cause an error


cysiek:
It's quite easy to make the following mistake:

```
enum SimpleEnum
    {
        ONE, TWO, THREE
    };

    PYBIND11_MODULE(MyPythonModule, module)
    {
        py::enum_<SimpleEnum>(module, "SimpleEnum")
        .value("ONE", SimpleEnum::ONE)          //NOTE: all value function calls are called with the same first parameter value
        .value("ONE", SimpleEnum::TWO)
        .value("ONE", SimpleEnum::THREE)
        .export_values();
}
```

This code compiles fine without any error/warning. When the module is loaded in Python, you can access only first value of SimpleEnum - "ONE". Would it be possible to make this code throw an error or at least warning at compile time (or eventually during loading python module) ?

Other information:
Pybind version: 2.2.3, Win 10 x64, Visual Studio 2015 x64, Python: Anaconda Python 3.6.5

jagerman:
Should be a simple enough check if you want to submit a PR (and turn your original post into a test case):
```C++
    pybind11::str pyname(name);
    if (m_entries.contains(pyname))
        throw value_error("...");
    // ... (current code)
```

cysiek:
I've created a pull request - https://github.com/pybind/pybind11/pull/1453
Please note that it's my first PR to open source project, so some "obvious" things might be wrong ;) Feel free to give me any feedback.

cysiek:
I've fixed formatting issues, but still travis fails for XCode 9 C++ ( [https://travis-ci.org/pybind/pybind11/builds/404210394](url) ). For me it looks a bit like configuration issue, @jagerman @bstaletic could you have a look at it?

bstaletic:
PR merged. Closing.

PR #1453, opened 2018-07-12
check for already existing enum value added; added test

cysiek:
Fix for https://github.com/pybind/pybind11/issues/1451 (Defining enum with mutiple identical values should cause an error)
Now trying to add already existing value throws an exception.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1465,7 +1465,7 @@ public:
         this->attr(name) = v;
         auto name_converted = pybind11::str(name);
         if (m_entries.contains(name_converted))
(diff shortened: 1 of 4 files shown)


bstaletic:
That error on Travis is definitely unrelated to your changes.

As for your PR, it looks good. I personally like `pyname` more than `name_converted`, but that's a matter of taste.

bstaletic:
I can't say anything about this getting merged because I don't have write access to the repo.
The travis fix is in #1454.

cysiek:
So... now this pull request will be accepted? Or something is missing?

wjakob:
Merged, thank you!

Issue #1444, opened 2018-07-05
pybind11 uses deprecated PyThread_set_key_value API


jbarlow83:
## Issue description

As of Python 3.7, the PyThread_* API was deprecated and superceded by the Thread Specific Storage API.

Details from Python documentation:
* https://docs.python.org/3/c-api/init.html#thread-specific-storage-tss-api
* https://www.python.org/dev/peps/pep-0539/

Compiling against Python 3.7 will produce numerous deprecation warnings for every file that includes pybind11 headers. Despite this, everything seems to be working okay. 

To suppress the warnings, pybind11 should select the new API on Python 3.7.

jbarlow83:
Closed by #1454

jonatanolofsson:
Could we get a new release with this in?

PR #1454, opened 2018-07-13
Switching deprecated Thread Local Storage (TLS) usage in Python 3.7 to Thread Specific Storage (TSS)

YannickJadoul:
Python 3.7 deprecated the use of the 'Thread Local Storage (TLS) API', resulting in a bunch of warnings when compiling (see #1444). This PR switches between using the old TLS API and new TSS API, based on the Python version number.

See https://python.readthedocs.io/en/latest/c-api/init.html#thread-specific-storage-tss-api and https://www.python.org/dev/peps/pep-0539/ for details on the new API.

Potential issues:
- Parts of the code where TLS was used were already subject to conditional compilation, macros and `#if`s. But now a few more are added to switch between the old and new API (though only at a handful of places). Is it worth adding a wrapper API, to get rid of some of the preprocessor directives within the code?
- The new `Py_tss_t *` key type of `internals::tstate` needs to be allocated (`PyThread_tss_alloc()`), but since the internals struct gets statically allocated once (and there is no destructor or other cleanup), it never gets `PyThread_tss_free`d (neither is the entry removed with `PyThread_tss_delete`). But then the old version never called `PyThread_delete_key`, either, and the struct has a static lifetime. Should there be a way of cleaning this up (e.g., through a `Py_AtExit` callback)?

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -21,15 +21,13 @@ inline PyObject *make_object_base_type(PyTypeObject *metaclass);
 // The old Python Thread Local Storage (TLS) API is deprecated in Python 3.7 in favor of the new
 // Thread Specific Storage (TSS) API.
 #if PY_VERSION_HEX >= 0x03070000
-    #define PYBIND11_TLS_KEY_TYPE Py_tss_t *
-    #define PYBIND11_TLS_KEY_NULL_VALUE nullptr
+    #define PYBIND11_TLS_KEY_INIT(var) Py_tss_t *var = nullptr
     #define PYBIND11_TLS_GET_VALUE(key) PyThread_tss_get((key))
     #define PYBIND11_TLS_REPLACE_VALUE(key, value) PyThread_tss_set((key), (tstate))
     #define PYBIND11_TLS_DELETE_VALUE(key) PyThread_tss_set((key), nullptr)
 #else
     // Usually an int but a long on Cygwin64 with Python 3.x
-    #define PYBIND11_TLS_KEY_TYPE decltype(PyThread_create_key())
-    #define PYBIND11_TLS_KEY_NULL_VALUE 0
+    #define PYBIND11_TLS_KEY_INIT(var) decltype(PyThread_create_key()) var = 0
     #define PYBIND11_TLS_GET_VALUE(key) PyThread_get_key_value((key))
     #if PY_MAJOR_VERSION < 3
         #define PYBIND11_TLS_REPLACE_VALUE(key, value) do { PyThread_delete_key_value((key)); PyThread_set_key_value((key), (value)); } while (false)
@@ -100,7 +98,7 @@ struct internals {
     PyTypeObject *default_metaclass;
     PyObject *instance_base;
 #if defined(WITH_THREAD)
-    PYBIND11_TLS_KEY_TYPE tstate = PYBIND11_TLS_KEY_NULL_VALUE;
+    PYBIND11_TLS_KEY_INIT(tstate);
     PyInterpreterState *istate = nullptr;
 #endif
 };
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -18,6 +18,27 @@ inline PyTypeObject *make_static_property_type();
 inline PyTypeObject *make_default_metaclass();
 inline PyObject *make_object_base_type(PyTypeObject *metaclass);
 
+// The old Python Thread Local Storage (TLS) API is deprecated in Python 3.7 in favor of the new
+// Thread Specific Storage (TSS) API.
+#if PY_VERSION_HEX >= 0x03070000
+    #define PYBIND11_TLS_KEY_TYPE Py_tss_t *
+    #define PYBIND11_TLS_KEY_NULL_VALUE nullptr
+    #define PYBIND11_TLS_GET_VALUE(key) PyThread_tss_get((key))
+    #define PYBIND11_TLS_REPLACE_VALUE(key, value) PyThread_tss_set((key), (tstate))
+    #define PYBIND11_TLS_DELETE_VALUE(key) PyThread_tss_set((key), nullptr)
+#else
+    // Usually an int but a long on Cygwin64 with Python 3.x
+    #define PYBIND11_TLS_KEY_TYPE decltype(PyThread_create_key())
+    #define PYBIND11_TLS_KEY_NULL_VALUE 0
+    #define PYBIND11_TLS_GET_VALUE(key) PyThread_get_key_value((key))
+    #if PY_MAJOR_VERSION < 3
+        #define PYBIND11_TLS_REPLACE_VALUE(key, value) do { PyThread_delete_key_value((key)); PyThread_set_key_value((key), (value)); } while (false)
+    #else
+        #define PYBIND11_TLS_REPLACE_VALUE(key, value) PyThread_set_key_value((key), (value))
+    #endif
+    #define PYBIND11_TLS_DELETE_VALUE(key) PyThread_set_key_value((key), nullptr)
+#endif
+
 // Python loads modules by default with dlopen with the RTLD_LOCAL flag; under libc++ and possibly
 // other STLs, this means `typeid(A)` from one module won't equal `typeid(A)` from another module
 // even when `A` is the same, non-hidden-visibility type (e.g. from a common include).  Under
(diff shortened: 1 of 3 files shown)


YannickJadoul:
That failing test seems to be due to `brew` installing Python 3.7 and the `python3.6` command not being found.

bstaletic:
I tried to change the `.travis.yml` to use python3.7, but that had a rather surprising behaviour. It raised these deprecation warnings that you mention, only they were treated as [errors](https://travis-ci.org/bstaletic/pybind11/jobs/404409827).

So to properly fix CI, we need this PR *and* [this change](https://github.com/bstaletic/pybind11/commit/91a9855e97d154a6f0914a217c3219b40f294130).

YannickJadoul:
Hmmmm, interesting. Seems to be [this line in `.travis.yml`](https://github.com/pybind/pybind11/blob/master/.travis.yml#L35) that's doing that (together with something in the `tests/CMakeLists.txt`).

Either way, I'll incorporate those changes in this one, then. Thanks! :-)

YannickJadoul:
Hurray, checks are passing! Thanks, @bstaletic!

So this PR would also solve this problem the other recent PRs that are getting CI failures (i.e., #1455, #1453, #1448, ...)

include/pybind11/detail/internals.h
```diff
@@ -166,8 +170,15 @@ PYBIND11_NOINLINE inline internals &get_internals() {
...
         PyThreadState *tstate = PyThreadState_Get();
-        internals_ptr->tstate = PyThread_create_key();
-        PyThread_set_key_value(internals_ptr->tstate, tstate);
+        #if PY_VERSION_HEX >= 0x03070000
+            internals_ptr->tstate = PyThread_tss_alloc();
+            if (!internals_ptr->tstate || PyThread_tss_create(internals_ptr->tstate))
+                pybind11_fail("get_internals: could not successfully initialize the TSS key!");
+            PyThread_tss_set(internals_ptr->tstate, tstate);
+        #else
+            internals_ptr->tstate = PyThread_create_key();
```
wjakob: Need to check for failure status

wjakob:
This looks great! — I just added one minor comment.

wjakob:
Regarding this question:

> Is it worth adding a wrapper API, to get rid of some of the preprocessor directives within the code?

Pybind11 already has a few ``#define PYBIND11_*``s for cases where function names differ but the API is otherwise unchanged. This is AFAIK the case for most of the TSS calls except for the one that resets to ``NULL``, so this could simplify the code quite a bit.

> The new Py_tss_t * key type of internals::tstate needs to be allocated (PyThread_tss_alloc()), but since the internals struct gets statically allocated once (and there is no destructor or other cleanup), it never gets PyThread_tss_freed (neither is the entry removed with PyThread_tss_delete). But then the old version never called PyThread_delete_key, either, and the struct has a static lifetime. Should there be a way of cleaning this up (e.g., through a Py_AtExit callback)?

This could be interesting, but I'd propose to do it in a separate PR.

bstaletic:
> This could be interesting, but I'd propose to do it in a separate PR.

I tried playing with Valgrind and ASAN. Running the benchmaarks for my project, there were consistent memory leaks (nothing alarming, but it was there) pointing to something in the python library. Considering that `valgrind python -c 'x=5'` (and even some slightly less trivial code)  showed no signs of memory leaks, I wondder if it would be possible to track these memory leaks down and if they are caused by pybind.

Not calling `PyThread_tss_free` or `PyThread_delete_key` sounds like something ASAN/Valgrind would mark as a potential leak.

The whole point of such an excercise would be to allow projects to use ASAN as a part of the CI.

YannickJadoul:
@wjakob Done, thanks for the suggestions.

> I just added one minor comment.

Done that. Apart from that, in principle, the `set` call could also fail, [according to the API docs](https://docs.python.org/3/c-api/init.html#c.PyThread_tss_set). Should we still check those (because that did not happen before with the old TLS API, either) ?

> [...] so this could simplify the code quite a bit

Yes, done so as well. Just noting that the macro functionality is not needed, so we could also have nice C++ typedefs and functions instead of macros.

@wjakob and @bstaletic Yeah, ok, that seems indeed like a separate effort to clean up all allocations and have a clean ASAN/Valgrind run. So, let's add this cleanup to that list of things to do in such a PR, then?

include/pybind11/detail/internals.h
```diff
@@ -18,6 +18,27 @@ inline PyTypeObject *make_static_property_type();
 inline PyTypeObject *make_default_metaclass();
 inline PyObject *make_object_base_type(PyTypeObject *metaclass);
 
+// The old Python Thread Local Storage (TLS) API is deprecated in Python 3.7 in favor of the new
+// Thread Specific Storage (TSS) API.
+#if PY_VERSION_HEX >= 0x03070000
+    #define PYBIND11_TLS_KEY_TYPE Py_tss_t *
+    #define PYBIND11_TLS_KEY_NULL_VALUE nullptr
```
jagerman: Do the type and null value need to be separate here?  It looks from the way they are used that you could simplify it to something along the lines of:

```C++
#define PYBIND11_TLS_KEY_INIT(var) Py_tss_t *var = nullptr;
// or for <3.7:
#define PYBIND11_TLS_KEY_INIT(var) decltype(PyThread_create_key()) var = 0;
```

bstaletic:
> ASAN/Valgrind

Yes, that's definitely something that should be done in a different PR.

include/pybind11/detail/internals.h
YannickJadoul: Hmmm, myeah, true. Depends if at some point you'd need to have the type separate.

But I can still change that, if you want? (Or feel free to push that commit to this branch yourself, as you should be allowed by GitHub.)

jagerman: Considering that these are entirely internal macros (i.e. if you're using them outside of pybind11, shame on you) we can also break them up later if needed -- but for the moment it doesn't look like they are.

jagerman: (Pushed a commit)

YannickJadoul:
Thanks!
Weird how Travis CI was apparently not triggered by your commit and push, but since Travis succeeded before, this should be OK if AppVeyor succeeds...

jagerman:
> Weird how Travis CI was apparently not triggered by your commit and push, but since Travis succeeded before, this should be OK if AppVeyor succeeds...

Travis-ci has flagged me for abuse for some unknown reason, and thus ignores my commits.  It's a bit annoying.

YannickJadoul:
Or they just know you don't write bugs? ;-)

wjakob:
> Travis-ci has flagged me for abuse for some unknown reason, and thus ignores my commits. It's a bit annoying.

Wow, super-strange. I'll go ahead and merge this now.

include/pybind11/detail/internals.h
```diff
@@ -18,6 +18,27 @@ inline PyTypeObject *make_static_property_type();
...
+    // Usually an int but a long on Cygwin64 with Python 3.x
+    #define PYBIND11_TLS_KEY_TYPE decltype(PyThread_create_key())
+    #define PYBIND11_TLS_KEY_NULL_VALUE 0
+    #define PYBIND11_TLS_GET_VALUE(key) PyThread_get_key_value((key))
+    #if PY_MAJOR_VERSION < 3
+        #define PYBIND11_TLS_REPLACE_VALUE(key, value) do { PyThread_delete_key_value((key)); PyThread_set_key_value((key), (value)); } while (false)
+    #else
+        #define PYBIND11_TLS_REPLACE_VALUE(key, value) PyThread_set_key_value((key), (value))
+    #endif
+    #define PYBIND11_TLS_DELETE_VALUE(key) PyThread_set_key_value((key), nullptr)
```
cyfdecyf: We should use `PyThread_delete_key_value` for `PYBIND11_TLS_DELETE_VALUE`.

`PyThread_set_key_value` is strange that it does not change key value to nullptr when it's already set to non-null value, tested on Python 2.7.12.

I notice this because my Python callback function (overriding virtual function by `PYBIND11_OVERLOAD_PURE`) always segfaults upon the second call.

YannickJadoul: @cyfdecyf And did it do that before this PR? Because I thought the code from before and after would be the same, on < 3.7 Python versions?

YannickJadoul: Hmmm, you're right, something weird is happening, there, using this macro :|

wjakob: @YannickJadoul, @cyfdecyf : could you check if PR #1517 fixes the issue?

cyfdecyf: Sorry for the late reply. I'll try to construct a test case for this issue.

jamesr66a: @wjakob I can confirm your patch fixes a regression we're seeing in PyTorch https://github.com/pytorch/pytorch/pull/11473#issuecomment-420155829. We have a release this Friday, do you have an ETA on when that fix can be merged? Thanks!

wjakob: I will merge it today and push out a new patch release today.

PR #1455, opened 2018-07-15
stl.h: propagate return value policies to type-specific casters

wjakob:
Return value policies for containers like those handled in in 'stl.h'
are currently broken.

The problem is that detail::return_value_policy_override<C>::policy()
always returns 'move' when given a non-pointer/reference type, e.g.
'std::vector<...>'.

This is sensible behavior for custom types that are exposed via
'py::class_<>', but it does not make sense for types that are handled by
other type casters (STL containers, Eigen matrices, etc.).

This commit changes the behavior so that
detail::return_value_policy_override only becomes active when the type
caster derives from type_caster_generic.

Furthermore, the override logic is called recursively in STL type
casters to enable key/value-specific behavior.

### DIFF
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -8,6 +8,7 @@
 */
 
 #include "pybind11_tests.h"
+#include "constructor_stats.h"
 #include <pybind11/stl.h>
 
 // Test with `std::variant` in C++17 mode, or with `boost::variant` in C++11/14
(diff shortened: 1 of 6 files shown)


tests/test_stl.cpp
```diff
@@ -235,4 +236,21 @@ TEST_SUBMODULE(stl, m) {
 
     // test_stl_pass_by_pointer
     m.def("stl_pass_by_pointer", [](std::vector<int>* v) { return *v; }, "v"_a=nullptr);
+
+    class Placeholder {
+    public:
+        Placeholder() { print_created(this); }
+        Placeholder(const Placeholder &) = delete;//{ print_created(this); }
```
jagerman: Debugging?

wjakob:
Thanks -- merged now.

damian-cell (commented):
Duplicate of #

damian-cell:
> Return value policies for containers like those handled in in 'stl.h'
> are currently broken.
> 
> The problem is that detail::return_value_policy_override::policy()
> always returns 'move' when given a non-pointer/reference type, e.g.
> 'std::vector<...>'.
> 
> This is sensible behavior for custom types that are exposed via
> 'py::class_<>', but it does not make sense for types that are handled by
> other type casters (STL containers, Eigen matrices, etc.).
> 
> This commit changes the behavior so that
> detail::return_value_policy_override only becomes active when the type
> caster derives from type_caster_generic.
> 
> Furthermore, the override logic is called recursively in STL type
> casters to enable key/value-specific behavior.

PR #1462, opened 2018-07-18
VS 15.8.0 Preview 4.0 has a bug with alias templates

MichaelGoulding:
This should hopefully be resolved in 15.9.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -476,7 +476,7 @@ template <typename...> struct void_t_impl { using type = void; };
 template <typename... Ts> using void_t = typename void_t_impl<Ts...>::type;
 
 /// Compile-time all/any/none of that check the boolean value of all template types
-#ifdef __cpp_fold_expressions
+#if (!defined(_MSC_VER) || ( _MSC_VER >= 1916 )) && defined(__cpp_fold_expressions)
 template <class... Ts> using all_of = bool_constant<(Ts::value && ...)>;
 template <class... Ts> using any_of = bool_constant<(Ts::value || ...)>;
 #elif !defined(_MSC_VER)


include/pybind11/detail/common.h
```diff
@@ -476,7 +476,7 @@ template <typename...> struct void_t_impl { using type = void; };
 template <typename... Ts> using void_t = typename void_t_impl<Ts...>::type;
 
 /// Compile-time all/any/none of that check the boolean value of all template types
-#ifdef __cpp_fold_expressions
+#if (!defined(_MSC_VER) || ( _MSC_VER >= 1916 )) && defined(__cpp_fold_expressions)
```
jagerman: Minor nitpick, but can you rewrite this to put the `defined(__cpp_fold_expressions)` conditions first, to make it easier to see what the intention is here?

i.e. "if the compiler has fold expressions and isn't a buggy MS compiler" vs. "if the compiler isn't a buggy MS compiler and has fold expressions"

I haven't been keeping up with MSVC versioning -- does `_MSC_VER == 1915` uniquely identify the buggy release, and are we sure this will be fixed for 15.8.1 Preview 4.1 Late Summer Creator Edition (or whatever MS decides to name the next one)?

MichaelGoulding: The developers told me they have the fix "scheduled tentatively for the 15.9 release" which would be compiler version 19.16.

wjakob:
I'm a bit skeptical whether it makes sense for us to commit compiler workarounds for pre-release compilers. I'd hate it if we accumulated lots of cruft over the years for software that is only relevant for a very short time window.

samertm:
FWIW, the latest update of Visual Studio 2017, 15.8.2, also has this issue.

The VS team is aware and has a fix ready for an upcoming release, but they didn't specify when to expect it: https://developercommunity.visualstudio.com/content/problem/284636/-cpp-fold-expressions-is-now-defined-but-pybind11.html

PR #1463, opened 2018-07-19
Make py::isinstance<py::str> less permissive

### LINKED ISSUE #1461: py::isinstance<py::str>(py_bytes_object) should not return true
## Issue description

In pybind 2.2.3, the template specialization `py::isinstance<py::str>()` returns True when used on `py::bytes` objects. This is incorrect in both Py2 and 3 (since `py::str` is `unicode` in Py2).

`py::instance(obj, py::str().get_type())` works correctly.

## Reproducible example code

```c++

    m.def("test_isinstance",
        []() {
            py::bytes b = "123";
            py::print(py::isinstance<py::str>(b));
        }
    );

```

```python

>>> pybind11.test_isinstance()
True

```

jagerman:
py::str's uses a permissive check function that allows both str (Py 3)/
unicode (Py 2), *and* bytes (Py 3)/str (Py 2) so that a `py::str` can
be created from either, but this makes the default `isinstance`
implementation overly permissive.

Fixes #1461


bstaletic:
If I got this right, this won't break code that does the following:
```c++
std::string GetStringFromPythonObject( pybind11::object o ) {
        // If it is any kind of string like python object just cast it to std::string
        if ( pybind11::isinstance< bytes >( o ) || pybind11::isinstance< str >( o ) ) {
                return o.cast< std::string >();
        }
        // Else, go through pybind11::str()
        return pybind11::str( o ).cast< std::string >();
}
```
Is that correct?

jagerman:
It looks like this PR does break Python 2, where `py::isinstance<str>` is being used to test whether the value is convertible to a string.  This is probably at least somewhat historic: we used to have a `check()` method which `isinstance` replaced, but the check was intentionally more permissive for `str`s.

This might be trickier to fix without breaking things.

wjakob:
This had some weird conflicts, I ended up applying the patch directly myself.

wjakob:
Whoops, I missed the fact that this does indeed break the test suite on 2.7. Reverting & reopening.

Issue #863, opened 2017-05-18
Feature idea: allow exposing read-only buffers


bmerry:
#### Issue description

This is something that occurred to me when reading the Python docs on memoryview, but not something I have any immediate need for - which is why I've labelled it a "feature idea" rather than a "feature request".

At present, `buffer_info` doesn't have any indication of whether the memory should be treated as read-only or read-write. There may be a use case for a class to expose a buffer while not allowing writes to it. It looks like it shouldn't be too hard to add a flag to buffer_info and then add check in `pybind11_buffer` to raise a `BufferError` if the request has `PyBUF_WRITABLE` is set but the exposed buffer is read-only.

bmerry:
Just spotted that this might be duplicated by #481, although it might not be quite the same thing.

aldanor:
Yes this makes sense, should be a simple flag addition; in fact I had it implemented in one of the branches at some point but it wasn't merged in.

I believe numpy should respect this flag (and set writeable flag accordingly), but this better be double checked.

firerogue:
Hi, I am trying to bind a project to Python using this library, and have just run in to this issue. I have an object with the function `const void* get_data()` that points to a couple megabytes of data. Right now I have to copy this into a `std::vector<uint8_t>` and send that to python, but this is very expensive. I was hoping to use `py::buffer` to get around this expensive copy, but can't because my data is read-only

complyue:
My +1 for I'm using pybind to expose mmap-ed array data to arm massive data on shared storage to multiple compute nodes, certainly majority of the mmaps are readonly.

If I don't touch pybind internal/detail like this:
```c++
  vector<array> bas; // bundle arrays
// ...
      // construct bundle array inplace
      bas.emplace_back(
          dt, // dtype
          // the explicit initializer_list construction is a necessary hint to compile,
          // plain {} won't work, even with explicit, element-wise type cast
          std::initializer_list<ssize_t>{scnt, slen}, // shape
          static_cast<void *>(mp + offset), // ptr
          mm // base
      );
      if (readonly_a) {
        // no public pybind api to manipulate array flags so far, have to use internal details until
        // https://github.com/aldanor/pybind11/pull/2 finds its way into release
        py::detail::array_proxy(bas.back().ptr())->flags &= ~py::detail::npy_api::NPY_ARRAY_WRITEABLE_;
      }
```
[...]

PR #1466, opened 2018-07-23
support for readonly buffers (#863)

skoslowski:
This adds a readonly flag to buffer_info objects. I consider this a first draft since it only covers the Buffer Protocol portion. Support for readonly in memoryview, array, ... will follow (as separate PRs)

I wanted to get some feedback on this part first:

- `readonly` added as default `false` parameter to not break things
- the `readonly` attribute is inserted at the end - consistent with the constructors, but deviating from the order in `Py_Buffer`.
- added a `const T*` constructor with 'readonly=true' - the same could be done for the others, but may require special casing types like nullptr_t


nedrebo:
Related https://github.com/pybind/pybind11/issues/1993

wjakob:
Merged now, apologies that this took a while.

PR #1476, opened 2018-08-01
Doc update: Remove duplicate feature from list

vanossj:
pybind11 is pretty awesome without padding the feature list


wjakob:
Good catch. There is also a duplicate in README.md -- can you remove that as well?

wjakob:
Thanks! :)

PR #1479, opened 2018-08-01
Allow ostream-redirection of more than 1024 characters

### LINKED ISSUE #1478: scoped_ostream_redirect only works for the first 1024 characters
## Issue description

When using `scoped_ostream_redirect`, or even `detail::pythonbuf` directly via a `std::ostream`, only the first 1024 characters are written. This is the size of the buffer in `detail::pythonbuf`, which leads me to believe there's some issue with the buffering.

## Reproducible example code

```c++
#include <iostream>
#include <string>
#include <pybind11/iostream.h>

namespace py = pybind11;

PYBIND11_MODULE(sample, m)
{
    m.def("foo", [] {
        std::string chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.\n";
        // 64 chars long
        std::cout << "Size = " << chars.size() << std::endl;

        pybind11::scoped_ostream_redirect redirect;

        // Only the first 16 lines are written
        for (std::size_t i = 0; i < 20; ++i) {
            std::cout << chars;
        }
    });
}
```

```python
import sample
sample.foo()
```

Output:

```
Size = 64
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.
```

Expected 4 more lines.

Quincunx271:
Fixes #1478 

In short, `detail::pythonbuf::overflow()` had its branch for the return statement backwards, since `bool(sync())` is `true` if and only if `sync()` fails; `sync()` returns `0` on success. This fixes it by using `sync() == 0` as the condition.

### DIFF
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -34,7 +34,7 @@ private:
             *pptr() = traits_type::to_char_type(c);
             pbump(1);
         }
-        return sync() ? traits_type::not_eof(c) : traits_type::eof();
+        return sync() == 0 ? traits_type::not_eof(c) : traits_type::eof();
     }
 
     int sync() {
--- a/tests/test_iostream.py
+++ b/tests/test_iostream.py
@@ -53,6 +53,16 @@ def test_captured(capsys):
     assert stdout == ''
     assert stderr == msg
 
+def test_captured_large_string(capsys):
+    # Make this bigger than the buffer used on the C++ side: 1024 chars
+    msg = "I've been redirected to Python, I hope!"
+    msg = msg * (1024 // len(msg) + 1)
+
+    m.captured_output_default(msg)
+    stdout, stderr = capsys.readouterr()
+    assert stdout == msg
+    assert stderr == ''
+
 
 def test_guard_capture(capsys):
     msg = "I've been redirected to Python, I hope!"


Quincunx271:
The CI errors seem to be a warning about using a deprecated numpy type (`matrix`), which has nothing to do with this change. All non-numpy tests seem to be passing.

wjakob:
There is something I don't understand about the original implementation, which is somewhat related to the patch here. It looks like ``sync()`` can't return any value other than zero -- so what's the purpose of the ternary expression?

Quincunx271:
If someone wrote a class which inherits from `pythonbuf`, `sync()` could return a non-zero value.

wjakob:
Ah, gotcha -- it's a virtual function in the C++ standard. This looks good to me then.

PR #1496, opened 2018-08-22
Support more natural syntax for vector extend

clrusby:
Currently pybind only supports the syntax as shown in the test on line 36:
v_int2.extend(m.VectorInt([4, 5]))

It would be more natural to support the additional syntax as shown on line 37:
v_int2.extend([6, 7])

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1293,6 +1293,21 @@ inline size_t len(handle h) {
     return (size_t) result;
 }
 
+inline size_t len_hint(handle h) {
+#if PY_VERSION_HEX >= 0x03040000
+    ssize_t result = PyObject_LengthHint(h.ptr(), 0);
+#else
+    ssize_t result = PyObject_Length(h.ptr());
+#endif
+    if (result < 0) {
+        // Sometimes a length can't be determined at all (eg generators)
+        // In which case simply return 0
+        PyErr_Clear();
+        return 0;
+    }
+    return (size_t) result;
+}
+
 inline str repr(handle h) {
     PyObject *str_value = PyObject_Repr(h.ptr());
     if (!str_value) throw error_already_set();
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -122,7 +122,7 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
 
     cl.def(init([](iterable it) {
         auto v = std::unique_ptr<Vector>(new Vector());
-        v->reserve(len(it));
+        v->reserve(len_hint(it));
         for (handle h : it)
            v->push_back(h.cast<T>());
         return v.release();
@@ -131,6 +131,28 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
     cl.def("extend",
        [](Vector &v, const Vector &src) {
            v.insert(v.end(), src.begin(), src.end());
+       },
+       arg("L"),
+       "Extend the list by appending all the items in the given list"
+    );
+
+    cl.def("extend",
+       [](Vector &v, iterable it) {
+           const size_t old_size = v.size();
+           v.reserve(old_size + len_hint(it));
+           try {
+               for (handle h : it) {
+                   v.push_back(h.cast<T>());
+               }
+           } catch (const cast_error &) {
+               v.erase(v.begin() + static_cast<typename Vector::difference_type>(old_size), v.end());
+               try {
+                   v.shrink_to_fit();
+               } catch (const std::exception &) {
+                   // Do nothing
+               }
+               throw;
+           }
        },
        arg("L"),
        "Extend the list by appending all the items in the given list"
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,11 @@ v2.3.0 (Not yet released)
 * The ``value()``  method of ``py::enum_`` now accepts an optional docstring
   that will be shown in the documentation of the associated enumeration.
 
+* Vectors now support construction from generators, as well as ``extend()`` from a
+  list or generator.
+  `#1496 <https://github.com/pybind/pybind11/pull/1496>`_.
+
+
 v2.2.3 (April 29, 2018)
 -----------------------------------------------------
 
--- a/tests/test_stl_binders.py
+++ b/tests/test_stl_binders.py
@@ -11,6 +11,10 @@ def test_vector_int():
     assert len(v_int) == 2
     assert bool(v_int) is True
 
+    # test construction from a generator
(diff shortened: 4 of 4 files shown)


include/pybind11/stl_bind.h
```diff
@@ -129,8 +129,10 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
     }));
 
     cl.def("extend",
-       [](Vector &v, const Vector &src) {
-           v.insert(v.end(), src.begin(), src.end());
+       [](Vector &v, iterable it) {
```
bstaletic: Do we really want `iterable`? It doesn't make much sense to extend a vector of strings with a dictionary even if its keys are strings. With this PR `v_int2.extend({1:1,2:2})` would extend the `v_int2` with `[1, 2]`.

Consider strings too, since they are iterable:
```
>>> from foo import VectorString
>>> bar = VectorString()
>>> bar.extend('bar')
>>> bar
VectorString[b,a,r]
```

I personally think that we should only allow python `list` and C++-like `VectorFoo` to extend `VectorFoo`.
Also, there's a performance benefit in not doing the `cast<T>` when `VectorFoo` is passed, so an overload for python types should at least be considered.

clrusby:
Given this is what happens if you call extend on a native python list, I’d say the behaviour is consistent so yes we do want it to act like this.

Point taken re: performance hit, so maybe we add this as an additional overloaded function rather than replacing the existing implementation completely? I’ll submit an update to the PR to do this in the next day or so.

Chris

bstaletic:
Another thing that I didn't mention is error handling. Passing a list with mixed elements results in a vague exception that tells the user to recompile in debug mode. Get on the gitter chat to see the discussion about it from maybe an hour ago with me and @yeganer.

clrusby:
Ok so performance update has been made.  With regards error handling - I don't see that this is any different from other places in pybind, for instance the vector constructor.  So this should probably be fixed in the general case rather than just here.  (So saying, presumably the reason why the error message is different in debug and not in release mode is because of performance, so this might not be a great idea).

clrusby:
Hi again,

Just found the error handling discussion and see what you mean.  The simplest (and hence almost certainly best) option is to go via a temporary vector, which is what I've done.

I also saw your comments on passing in generators.  I figured I'd have a stab at addressing this too, see my most recent commit.  I thought I'd create a variant of len called len_hint so that I could take advantage of new functionality in python 3.4 (PyObject_LengthHint).

Chris

bstaletic:
Take a look at my take on the whole thing: https://github.com/bstaletic/pybind11/commit/20c8306968d804598025183f3bbfcfb09d23694c

include/pybind11/stl_bind.h
```diff
@@ -136,6 +136,28 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
        "Extend the list by appending all the items in the given list"
     );
 
+    cl.def("extend",
+       [](Vector &v, iterable it) {
+           const size_t old_size = v.size();
+           v.reserve(old_size + len_hint(it));
```
bstaletic: I'd favour the more accurate `len` over `len_hint` if `len` is available. Something like this, perhaps:

    v.reserve(old_size + (hasattr(it, "__len__") ? len(it) : len_hint(it)));

Why am I using a `?:` here? Because, according to @yeganer it makes it clearer that this is just preallocation and that everything inside `reserve()` is just an implementation detail, compared to `if (...) { v.reserve(...); } else { v.reserve(...) }`.

I think the same should be done on line 125.

include/pybind11/pytypes.h
```diff
@@ -1282,6 +1282,21 @@ inline size_t len(handle h) {
     return (size_t) result;
 }
 
+inline size_t len_hint(handle h) {
+#if PY_VERSION_HEX >= 0x03040000
+    ssize_t result = PyObject_LengthHint(h.ptr(), 0);
+#else
+    ssize_t result = PyObject_Length(h.ptr());
```
bstaletic: Falling back to a more accurate `PyObject_Length()` is questionable. How about this:

- >=Python3.4, everything is fine just do what you're doing now.
- <Python3.4, there is no `PyObject_Length()`, so just `return 0`.

include/pybind11/stl_bind.h
clrusby: Not necessary, the documentation for PyObject_LengthHint states that: Return an estimated length for the object o. First try to return its __actual__ length, then an estimate using __length_hint__(), and finally return the default value.

include/pybind11/pytypes.h
clrusby: I assume you mean <Python3.4 there is no PyObject_LengthHint?

include/pybind11/stl_bind.h
bstaletic: I'm definitely missing something. Your `len_hint` will return `PyObject_LengthHint()` and won't try `PyObject_Length()`, so why is it not necessary to call `len` here first?

include/pybind11/pytypes.h
bstaletic: Right, my bad.

include/pybind11/stl_bind.h
clrusby: PyObject_LengthHint returns the actual length if available (i.e. via a call to '__len__').  If not it tries '__length_hint__()' and failing that returns the default value (that I'm setting to 0).  This seems to be precisely the behaviour we want?

bstaletic: Thanks for clearing that up. I wasn't aware that `PyObject_LengthHint()` calls `__len__()` first. Yes, this is what we want.

clrusby:
Great - I think we are now in agreement that this code is now good to ship?  I don't think the failing build is anything to do with these changes.  How do we proceed from here?

Thanks

bstaletic:
>  I think we are now in agreement that this code is now good to ship?

Yes, I'm fine with the changes made here. If `len` is documented, `len_hint` should be as well, but I have not looked at it.

>  I don't think the failing build is anything to do with these changes.

My fork has been failing in the same way. On Travis it seems that the new numpy broke something. On Appveyor, it's Conda... again.

> How do we proceed from here?

That's the hard part. Now this should get the attention of the maintainers who seem to be found nowhere lately.

bstaletic:
Rebasing onto master should fix CI.

clrusby:
Had to make a minor change to the test so the flake8 checks passed, but now we’re good.

Chris

bstaletic:
Changelog entry is missing.

clrusby:
@bstaletic changelog updated, thanks for pointing this out.

clrusby:
@wjakob any thoughts on this PR? Is it suitable for merging into a future version?

wjakob:
This looks good to me, thanks!

PR #1502, opened 2018-08-27
added py::ellipsis() method for slicing of multidimensional NumPy arrays

wjakob:
This PR adds a new py::ellipsis() method which can be used in
conjunction with NumPy's generalized slicing support. For instance,
the following is now valid (where `a` is a NumPy array):
```
py::array b = a[py::make_tuple(0, py::ellipsis(), 0)];
```


tests/test_numpy_array.py
```diff
@@ -408,3 +408,7 @@ def test_array_create_and_resize(msg):
     a = m.create_and_resize(2)
     assert(a.size == 4)
     assert(np.all(a == 42.))
+
```
ax3l: Flake8 error in CI:
```
./tests/test_numpy_array.py:412:1: E302 expected 2 blank lines, found 1
def test_index_using_ellipsis():
```

tests/test_numpy_array.cpp
```diff
@@ -295,4 +295,8 @@ TEST_SUBMODULE(numpy_array, sm) {
         std::fill(a.mutable_data(), a.mutable_data() + a.size(), 42.);
         return a;
     });
+
+    sm.def("index_using_ellipsis", [](py::array a) {
+        return a[py::make_tuple(0, py::ellipsis(), 0)];
```
ax3l: must be excluded in python 2 tests

wjakob:
@ax3l : thanks!

PR #1507, opened 2018-08-30
add wrapper for callable builtin function

vanossj:
Wraps builtin function [callable](https://docs.python.org/3.7/library/functions.html#callable).

Same behavior as `py::isinstance<py::function>(obj)` bit a bit more clear, especially concerning class types.

Tests could be moved to an existing file.


wjakob:
Hm, I think it would make more sense for `callable` to be a wrapper type akin to `function`, etc. I'm not enthusiastic about merging this and will close the PR.

PR #1513, opened 2018-09-06
Add get method to py::dict

woodychow:
It works just like Python except that the default argument is not optional

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1173,6 +1173,14 @@ public:
     void clear() const { PyDict_Clear(ptr()); }
     bool contains(handle key) const { return PyDict_Contains(ptr(), key.ptr()) == 1; }
     bool contains(const char *key) const { return PyDict_Contains(ptr(), pybind11::str(key).ptr()) == 1; }
+    template <typename T> object get(handle key, T &&defaultv) const {
+        PyObject* ret = PyDict_GetItem(ptr(), key.ptr());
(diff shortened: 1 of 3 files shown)


jagerman:
I'm sort of ambivalent about including this (the C++ isn't onerous: `py::object o = dict.contains(key) ? dict[key] : py::cast(default);`, but I think that if we accepted it it would make more sense to move this to `class accessor`, something along the lines of:

```C++
    template <typename T>
    object get_with_default(T &&defaultv) const {
        return obj.contains(key) ? *this : object_or_cast(std::forward<T>(defaultv)
    }
```

(Then you can use it for non-`dict` objects--include `dict`s that are held in C++ `object`s--as well).

woodychow:
boost::python has the same function. I want to minimize the changes needed to transition from boost::python to pybind11. Also, it seems more intuitive to just mirror the existing Python function name and usage.

Could you please elaborate more on the code you wrote? What would be the example usage of it? Is it like
```cpp
py::accessor(obj, key).get_with_default(defaultv)
```
?

jagerman:
`accessor` is an internal class that a pybind user should never need to deal with directly, but is the return type of `operator []` on a `py::object` (including `py::dict`).  So in other words, it would allow you to write:

```C++
    py::object o = /* ... */;
    py::object val = o["abc"].get_with_default("z");
```

woodychow:
Thanks. I see.

The point of this MR is to mirror the existing Python and boost::python API, which I think is more intuitive than inventing a new API.

Skylion007:
Alternative to #2779

PR #1514, opened 2018-09-06
Support arbitrary type in item accessor's subscript operator

woodychow:


wjakob:
IIRC this is subsumed by the second subpatch in PR #1511. Can you confirm?

woodychow:
I see. Thanks very much!

PR #1517, opened 2018-09-09
fix regression discussed by @cyfdecyf in #1454

wjakob:
The PYBIND11_TLS_DELETE_VALUE in PR #1454 changed behavior wrt. Python 2.x. This commit restores the old behavior.


wjakob:
@cyfdecyf: I was not able to reproduce the issue when calling pure virtual functions multiple times in Python 2.7.x, thus it's hard for me to say if this PR addresses the issue. Could you please give it a try? For the record, it would also be great to know what OS you're running this on.

wjakob:
Thank you @cyfdecyf, I added your test.

PR #1524, opened 2018-09-14
compile time check that there are no argument annotations for properties

### LINKED ISSUE #1519: py::arg in def_property results in invalid pointers
## Issue description
Its possible to add py::arg to def_property (and co). It seems to work until its time to clean up, then there are invalid pointer problems.

When using a setter I can see the reason for providing and arg, especially if the property setter only takes a specific type. But it should be supported or fail while compiling.

## Reproducible example code
```c++

namespace py = pybind11;

struct Thing{
    Thing() = default;
    int x=0;
};

PYBIND11_MODULE(property_args, m) {
    py::class_<Thing> thing(m, "Thing");
    thing.def(py::init<>());
    thing.def_property("x", [](Thing &instance) {
            return instance.x;
        }, [](Thing &instance, py::int_ new_x) {
            instance.x = py::cast<int>(new_x);
        },
        "x"_a, // remove this line and everything fine
        "property doc");
}
```

Compiles just fine. Works when using it (ipython example), but crashes when exiting.

```python
In [1]: import property_args
In [2]: t = property_args.Thing()
In [3]: t.x
Out[3]: 0
In [4]: t.x?
Type:        property
String form: <property object at 0x7fa8c81747c8>
Docstring:   property doc
In [5]: t.x = 10
In [6]: t.x
Out[6]: 10
In [7]: exit
munmap_chunk(): invalid pointer
```

vanossj:
fixes #1519 

When a property is made, all attributes are processed. However for properties, these attributes are processed outside of the `cpp_function` constructor. The `cpp_function` class duplicates and frees strings for attributes like `doc` and `arg`. When these attributes are processed and added to the `function_record` outside `cpp_function` the strings are not duplicated, but then are still `free`d by the `cpp_function` destruction, causing the error.

The `def_property_static` correctly `free`s and `strdup`s the doc attribute, but does not check if an `arg` attribute as processed. Since adding an argument annotation to either the setter or getter would never be displayed to the user, it makes more sense to check that there aren't any at compile time than to `free` and `strdup` all new `arg`s.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1257,6 +1257,8 @@ public:
     /// Uses cpp_function's return_value_policy by default
     template <typename... Extra>
     class_ &def_property_static(const char *name, const cpp_function &fget, const cpp_function &fset, const Extra& ...extra) {
+        static_assert( 0 == detail::constexpr_sum(std::is_base_of<arg, Extra>::value...),
+                      "Argument annotations are not allowed for properties");
         auto rec_fget = get_function_record(fget), rec_fset = get_function_record(fset);
         auto *rec_active = rec_fget;
         if (rec_fget) {


bstaletic:
This lacks a test that proves that this change indeed works.

wjakob:
That looks fine, but why only for static properties?

vanossj:
Its for all properties. All the `def_property` and family functions eventually call `def_property_static` which is where the `Extra` doc strings and annotations get processed.

wjakob:
Indeed, I forgot about that :). LGTM then.

Issue #1224, opened 2017-12-25
numpy arrays scalar


fbriol:
Hello I need different functions taking as arguments a value of « numpy arrays scalar » type.

It is possible to write prototypes like this:

    PYBIND11_MODULE(np_scalar, m) {
        m.def("overloaded", [](py::array_t<double>) {
            return "double";
        });
        m.def("overloaded", [](py::array_t<std::int8_t>) {
            return "int8";
        });
        m.def("overloaded", [](py::array_t<std::int16_t>) {
            return "int16";
        });
    }

However, this method does not allow to intercept scalars, for example:

    >>> import np_scalar
    >>> np_scalar.overloaded(numpy.array([1], dtype='int8'))
    'int8'
    >>> np_scalar.overloaded(numpy.int8(1))
    'double'

How do we intercept numpy arrays scalar?

Thank you very much for your help.

chhenning:
I'm not an expert but can you do:
```
m.def("overloaded", [](py::array a)
{
    auto buf = a.request();
    auto fmt = buf.format;

    if (fmt == "int32")
    {
        return "int32";
    }
    else if (fmt == "float64")
    {
        return "float64";
    }
    else
    {
        return "unknown";
    }
});
```

fbriol:
This doesn't work because this prototype only accepts array, not scalar.

    m.def("overloaded", [](py::array a)
    {
        auto buf = a.request();
        auto fmt = buf.format;

        if (fmt == "i")
        {
            return "int32";
        }
        else if (fmt == "h")
        {
            return "int16";
        }
        else
        {
            return "unknown";
        }
    });

    >>> import np_scalar
    >>> np_scalar.overloaded(numpy.array([1], dtype="int32"))
    'int32'
    >>> np_scalar.overloaded(numpy.int32(1))
    .../...
    TypeError: overloaded(): incompatible function arguments. The following argument types are 
    supported:
        1. (arg0: array) -> str

vanossj:
I was just doing a similar thing. It *should* work with `py::buffer` instead of `py::array`
However numpy doesn't expose the buffer info of scalers correctly, numpy/numpy#10265, it always reports them as byte arrays.

```c++
m.def("overloaded", [](py::buffer a)
{
    auto buf = a.request();
    if (buf.ndim == 0)
    {
        return "scaler of type " + buf.format;
    }
    else
    {
        return "ndarray of type " + buf.format;
    }
});
```

```python
>>> import np_scalar
[...]

PR #1530, opened 2018-09-17
Test: Numpy Scalar Creation

ax3l:
I found that te numpy array tests already contained an empty-shaped array test, but none with data in it.

Following PEP 3118, scalars have an empty shape and ndim 0. This works already and is now also documented/covered by a test.

Side note for readers: as documented [here](https://github.com/pybind/pybind11/issues/1224#issuecomment-354357392), Numpy < 1.15 is bugged when handling scalars.


ax3l:
ping @jagerman @wjakob this PR documents usage of scalar numpy arrays

wjakob:
LGTM, thanks!

Issue #1532, opened 2018-09-17
Incompatible CMake compile options for MSVC


paamand:
pybind11Tools.cmake line 214 and 220 adds the "/MP" compile option. However, this option is incompatible with the default compile options /Gm (minimal rebuild) for MSVC. Result is compiler warning.

Solution: Only add the /MP option to DEBUG build.
Replace line 212 through 221 with
```
target_compile_options(${target_name} PRIVATE /bigobj)
if(CMAKE_VERSION VERSION_LESS 3.11)
  target_compile_options(${target_name} PRIVATE $<$<CONFIG:RELEASE>:/MP>)
else()
  # Only set these options for C++ files.  This is important so that, for
  # instance, projects that include other types of source files like CUDA
  # .cu files don't get these options propagated to nvcc since that would
  # cause the build to fail.
  target_compile_options(${target_name} PRIVATE $<$<CONFIG:RELEASE>:$<$<COMPILE_LANGUAGE:CXX>:/MP>>)
endif()
```

paamand:
Fixed in pull request #1533 
https://github.com/pybind/pybind11/pull/1533

bstaletic:
The pull request has been merged. Closing.

PR #1533, opened 2018-09-17
Issue #1532: Incompatible config options, /MP vs /Gm for MSVC in DEBUG

paamand:
Added a cmake generator expression for /MP flag to only be set in RELEASE.

### DIFF
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -211,13 +211,13 @@ function(pybind11_add_module target_name)
     # needed for bigger binding projects due to the limit to 64k addressable sections
     target_compile_options(${target_name} PRIVATE /bigobj)
     if(CMAKE_VERSION VERSION_LESS 3.11)
-      target_compile_options(${target_name} PRIVATE $<$<CONFIG:RELEASE>:/MP>)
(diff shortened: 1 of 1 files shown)


wjakob:
Can you comment on why? Also, note that there are other release mode builds (e.g. with debug info, min. size) where the parallel build makes sense.

wjakob:
Ah, I see the title now. The question then is which of the (four?) CMake build modes adds /Gm.

BTW: What I don't understand is why the AppVeyor build succeeds if the flags are incompatible as you say.

paamand:
It's only a warning. MSVC prints the warning about ignoring /Gm, which is probably why it suceedes?
From [https://msdn.microsoft.com/en-us/library/bb385193.aspx](https://msdn.microsoft.com/en-us/library/bb385193.aspx)
"The /Gm Compiler Option
By default, a project build enables the /Gm compiler option (incremental builds) for debug builds, and disables it for release builds. Therefore, the /MP compiler option is automatically disabled in debug builds because it conflicts with the default /Gm compiler option."

wjakob:
Could you change the condition so that /MP is added if the mode is /not/ debug, rather only in release mode? There are other builds (e.g. MinSizeRel) where the parallelism is helpful.

wjakob:
Thanks!

Issue #1531, opened 2018-09-17
Compiler warnings on #include <pybind11/operators.h>


paamand:
When operators.h and iostream.h are both included
```
#include <pybind11/operators.h>
#include <pybind11/iostream.h>
```
you will get a warning about declaration of 'self'
```
iostream.h(197): warning C4459: declaration of 'self' hides global declaration
operators.h(41): note: see declaration of 'pybind11::detail::self'
```

The culprit is this line in operators.h:
`static const self_t self = self_t();`

I don't know enough about the line to suggest an edit.

bstaletic:
Do you by any chance have `using namespace pybind11` or `using pybind11::self` in that translation unit? Or maybe even `using namespace std`? Fully qualifying `pybind11::self` should avoid the warning.

paamand:
I don't, but operators.h has. So either remove the namespace for operators.h or rename the local variable in iostream.h. I suggest the latter, as it is a very minor fix:
`iostream.h@197: .def("__exit__", [](detail::OstreamRedirect &self, args) { self.exit(); });`
becomes
`iostream.h@197: .def("__exit__", [](detail::OstreamRedirect &self_, args) { self_.exit(); });`

Created pull request.
https://github.com/pybind/pybind11/pull/1535

bstaletic:
Once again, PR merged. Closing

PR #1535, opened 2018-09-18
Update iostream.h: Changed a local varname 'self' to 'self_'

paamand:
Avoiding conflicts in namespace pybind11::self.
https://github.com/pybind/pybind11/issues/1531


paamand:
Comment was intended to highlight the potential bad practice of exposing the "self" from operators.h by namespace import.
Removed comment.

PR #1537, opened 2018-09-19
Numpy Scalar Array Overloads?

ax3l:
Numpy scalar arrays (ndim==0, shape=()) should match the same `array_t`, `array` and `buffer` overloads as their nD counterparts.

Yet, for some reason they do not.

Also, a `np.uint64(42)` matches a `py::buffer` interface for me in Python 3.6.6 yet not in Python 2.7.13

All tested with Numpy 1.15.1 (due to issues with proper scalar handling prior to 1.15.0).

Any idea what is wrong here?
(This PR is just opened for demonstration so we all see the same behavior.)

### DIFF
--- a/tests/test_numpy_array.py
+++ b/tests/test_numpy_array.py
@@ -259,6 +259,15 @@ def test_overload_resolution(msg):
     assert m.overloaded(np.array([1], dtype='complex')) == 'double complex'
     assert m.overloaded(np.array([1], dtype='csingle')) == 'float complex'
 
+    # Scalar arrays:
+    assert m.overloaded(np.float64(1)) == 'double'
+    assert m.overloaded(np.float32(1)) == 'float'
+    assert m.overloaded(np.ushort(1)) == 'unsigned short'
+    assert m.overloaded(np.intc(1)) == 'int'
+    assert m.overloaded(np.longlong(1)) == 'long long'
+    assert m.overloaded(np.complex(1)) == 'double complex'
+    assert m.overloaded(np.csingle(1)) == 'float complex'
+
     # No exact match, should call first convertible version:
     assert m.overloaded(np.array([1], dtype='uint8')) == 'double'
 


ax3l:
cc @vanossj maybe you have an idea on this?

ax3l:
cc @jagerman do you have any idea what might be wrong here? Especially the Python 3 issue is troubling us.

vanossj:
I've been working on this in the background. For matching scalars to array overload, the problem is that the py::array class uses `PyArray_Check_` to see if the the overloaded function is a valid match. `PyArray_Check_` is false for scalar types, even though those could be coerced into arrays easily. 

I've managed to get it so that a scalar would match to an overload with `py::array`. But the scalar doesn't yet match an overload to the the more specific 'py::array_t`. That's my next step.

vanossj:
maybe someone can help clarify behaviour (and desired behaviour, if that differs), but I think that pybind11 will match the first valid overloaded function, not necessarily the best overload. If you pass in a `np.float64(0)` which overload would it hit?

```c++
void overload(py:array x);
void overload(py:array_t<double> x);
void overload(py:buffer x);
```

all of those are valid overloads (if scalar to array conversion works as expected), i don't know if pybind11 is suppose to choose the `array_t` overload, or just go with the first one.

what if we add more overloads
```c++
void overload(float x);
void overload(double x);
```
The `double` overload should match, but i could see the case for `float`, if we used np.float32(), then both should definitely match.

ax3l:
> pybind11 will match the first valid overloaded function, not necessarily the best overload

@vanossj that's exactly my problem, since this is inconsequently handled. For "regular" (ndim>0) numpy arrays, pybind11 *does* match the best overload (good!). But for scalar numpy arrays it does not and instead takes the first.

> void overload(float x);

That's a different example, one on fundamental python types, and similar to string and integer handling. Here it's totally ok that pybind11 matches the first one found since there can not be a "best" match to the builtin Python type: there is only one integer, one float and one basic string type in basic Python.

vanossj:
The problem with numpy scalars is that they are and aren't arrays.

Which overload should `np.float64` hit?

```c++
void overload(py:array_t<double> x);
void overload(double x);
```

I don't have a definitive answer. Maybe its best to let the programmer decide by changing the definition order.

vanossj:
numpy scalars objects do not support the buffer interface in python 2. So you're py::buffer overload not working in python 2.7 is a numpy issue not a pybind11 issue.
```python
memoryview(np.int16()) # works in python3, not python2
```

ax3l:
> Which overload should np.float64 hit?

array_t before the generic float. I think that's quite well for a distinct numpy type.

vanossj:
In one of the libraries I use, I would choose the other way. I use numpy scalars so they maintain their native size and signed behaviour, but I want them to behave as numbers, not as arrays.

ax3l:
But the thing is, they *are* zero-d arrays by numpy definition :-)

I have exactly the same use case as you, I basically want to keep their exact type, and am ok to catch them and separate them in `array_t` if need be, by `ndim()`, as I already do it via the `py::buffer` interface.

We can still do this if we make numpy.scalars match the `array_t` interface.

vanossj:
Scalars like `np.uint64()` are not 0d arrays. They may behave like 0D arrays a lot of the time, but at the C level implementation they are not the same.

The numpy c-api [`PyArray_CheckScalar()`](https://docs.scipy.org/doc/numpy-1.15.0/reference/c-api.array.html#c.PyArray_CheckScalar) hints at the underlying difference.

`np.uint64()`, `np.in16()`, ... these are all subclasses of `PyGenericArr_Type`

np.array(1, dypte=np.uint64), np.array(1, dypte=np.uint64), ... these are all subclasses of `PyArray_Type` (aka `np.ndarray`)

Both of these ways make a numpy scalar, but `PyGenericArr_Type` and `PyArray_Type` are not directly related.

```python
isinstance(np.uint64(), np.ndarray) #False
```

In pybind, `array`/`array_t` are C++ bindings to the `PyArray_Type` objects. Pybind cannot use the `array_t` interface to represent C `PyGenericArr_Type` objects, it must first create a new `PyArray_Type` object, that has the same dtype and value, and represent that object. Its no longer linked to the original `PyGenericArr_Type` object.

My first commits at #1550 seemed to let scalars like `x=np.uint64(12)` match `array_t<uint64_t>`, but it had a subtle flaw, it was always copying that `x` into a new object, `y=np.array(12, dtype=np.uint64)`. Then the overloaded `array_t<uint64_t>` function would receive object `y`. This seemed to work if you only read the value of the scalar you passed, but you would never be able to mutate it and see that change back in object `x`.

ax3l:
Thank you for the details. I was already concerned when I saw the numpy implementation that those are distinct types, made to *look and behave identical* to the user, that this would be troublesome.

In that case I agree that this will be tricky, we should probably at least note in the docs that `array_t` does not include numpy scalars with 0-d dimension then (since they are distinct, different types) and mention the `py::buffer` approach for handling both in Python 3 as we do.

For Python 2, do you have a hint how we could handle those types in overloads (since it's not a buffer) to avoid they cast just to the generics? I guess one can somehow catch them typesafe in `PyGenericArr_Type`?

PR #1539, opened 2018-09-21
normalizes all possible output directories for "external_module"

TinyTinni:
Hi,
MSVC outputs targets into an additional directory depending on the
selected build type inside the IDE i.e. 
on a Release build, the target "external_module" goes to: "<output_dir>\\**Release**\\external_module<etc>.pyd".**

This will remove the extra directory for every possible build type in the target "external_module", so that the test target "cpptest"  can import "external_module" with the in cmake configured working directory.

Settting the property `LIBRARY_OUTPUT_DIRECTORY ` is not enough for MSVC (and, i guess, XCode does the same).

### DIFF
--- a/tests/test_embed/CMakeLists.txt
+++ b/tests/test_embed/CMakeLists.txt
@@ -36,6 +36,11 @@ add_custom_target(cpptest COMMAND $<TARGET_FILE:test_embed>
 
 pybind11_add_module(external_module THIN_LTO external_module.cpp)
 set_target_properties(external_module PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+foreach(config ${CMAKE_CONFIGURATION_TYPES})
+    string(TOUPPER ${config} config)
+    set_target_properties(external_module PROPERTIES LIBRARY_OUTPUT_DIRECTORY_${config} ${CMAKE_CURRENT_SOURCE_DIR})
+endforeach()
+
 add_dependencies(cpptest external_module)
 
 add_dependencies(check cpptest)


wjakob:
Why is this needed? All the continuous integration tests already pass on windows as far as I can see.

TinyTinni:
You are right, and as far as i can see now, I am unable to explain, why they are working.

First, it started, that on my local machine the cpptest didn't worked (couldn't import external_module, because it was in the Debug subdirectory,)

Looking at the past buildings on appveyor [\[i.e. last success build before PR\]](https://ci.appveyor.com/project/wjakob/pybind11/build/1.0.2987/job/4ijhks04nm7sygvn), the output seems to be the same like on my local machine:

> external_module.vcxproj -> C:\\projects\\pybind11\\tests\\test_embed\\_Debug_\\external_module.cp36-win_amd64.pyd
>   test_embed.vcxproj -> C:\\projects\\pybind11\\tests\\test_embed\\_Debug_\\test_embed.exe

Therefore, the output_directory property without config [\[on this line\]](https://github.com/pybind/pybind11/blob/master/tests/test_embed/CMakeLists.txt#L38) has the same behavior as described in my first post (adds $CONFIG_TYPE subdirectory to the output_dir).

So both, the binary and the module are in the same sub-directory, but the test command ("cpptest" target) says, that "test_embed.exe" should be executed in `${CMAKE_CURRENT_SOURCE_DIR}`[\[source_ref\]](https://github.com/pybind/pybind11/blob/master/tests/test_embed/CMakeLists.txt#L35), which is clearly not `tests\test_embed\Debug` but `tests\test_embed\`. However, in this directory wasn't placed any "external_module" (not by the compiler, not by cmake via copy) and python should fail to find it, since it doesn't look in sub-directories, as far as i know.

Even when i copy the build command from appveyor:
`- cmake --build . --config %CONFIG% --target cpptest -- /m /v:m`
I get the error about the non found external_module on my local machine on a fresh cloned version:
>  FAILED:
>    REQUIRE( py::module::import("external_module").attr("A")(123).attr("value").cast<int>() == 123 )
>  due to unexpected exception with message:
>    ModuleNotFoundError: No module named 'external_module'  
>
> test cases:  8 |  7 passed | 1 failed
> assertions: 30 | 29 passed | 1 failed

Maybe the error is on my side.
But, i can't see, whether inside the CMake code, nor in the appveyor script, where you are pointing to the "external_module" file in the _Debug_ sub-directory when executing test_embed.exe via cpptest target.
Also, i can't find, inside the script, any extension of the `PYTHONPATH` variable for a module lookup.

Therefore, I cannot explain, why it is currently working in appveyor and i cannot explain why it should work on my local machine or any other machine without a special env setting.

wjakob:
Thank you for checking -- I don't think it makes sense for us to consider PR whose need is unclear, so I'll close this for now. If others run into the same issue, we'll take another look.

PR #1544, opened 2018-09-26
Update basics.rst

radicalrafi:
you will need to install pybind11 using pip3


wjakob:
This is incorrect, the build system will (and should!) use the local version of pybind11.

radicalrafi:
Deleted the branch , I made the edit as I found this issue in my current device I assume that using Anaconda causes this .
Thanks and Sorry for the trouble :smiley:

PR #1549, opened 2018-10-02
Added a debug flag to the PYBIND11_INTERNALS_VERSION

### LINKED ISSUE #1548: PYBIND11_INTERNALS_ID doesn't take build-type or compiler flags into account
## Issue description

Two pybind11 - modules with the same PYBIND11_INTERNALS_VERSION aren't necessarily binary compatible, since the build flags aren't included in the PYBIND11_INTERNALS_ID, and at least MSVC subtly changes the memory layout depending on the build type.

I am developing two pybind11-modules at the same time, using cmake and VS2017 to build and debug them. And they need to be built with the same build type, or the internals - struct (created and registered by whichever module is imported first) isn't binary compatible, and the first call to `(*internals_pp)->registered_exception_translators.push_front()` chrashes.

crest01:
potentially resolves #1548

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -141,7 +141,7 @@ struct type_info {
 #define PYBIND11_INTERNALS_VERSION 2
 
 #if defined(_DEBUG)
-#   define PYBIND11_BUILD_TYPE "debug"
+#   define PYBIND11_BUILD_TYPE "_debug"
 #else
 #   define PYBIND11_BUILD_TYPE ""
 #endif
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/internals.h
```diff
@@ -140,17 +140,23 @@ struct type_info {
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version
 #define PYBIND11_INTERNALS_VERSION 2
 
+#if defined(_DEBUG)
+#   define PYBIND11_BUILD_TYPE "debug"
```
wjakob: preceding underscore? (``_debug``)

crest01: Yeah, thats MSVC specific.
And i suspect that whole problem might only appear with MSVC anyways

crest01: ah. That does make more sense.

crest01:
This does indeed fix #1548 for me, but i only tried the debug/release combination. Not sure if any other compiler flags have a similar effect.

wjakob:
Merged, thanks!

PR #1550, opened 2018-10-02
WIP: Numpy scalars overload to arrays

vanossj:
Allows numpy scalar objects (ex `np.int16`, `np.float32`) to match a `py::array` and `py::array_t` arguments in bound functions.

Note that only numpy scalars will match, NOT any PEP 3118 scalar (ex `ctypes.c_int16`)

Includes tests from @ax3l  in PR #1537 

Numpy scalars will call the first overloaded binding that they can be converted into: POD, array_t, array, or buffer. If an existing user has bindings like 
```c++
sm.def("overloaded8", [](py::array_t<double>) { return "double"; });
sm.def("overloaded8", [](double) { return "double POD"; });
```

This patch would change behaviour. If maintaining backwards comparability for this case is important.  `array` could have a variodic template for a flag type `py::allow_scalar_conversion`

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -145,6 +145,9 @@ struct npy_api {
     bool PyArrayDescr_Check_(PyObject *obj) const {
         return (bool) PyObject_TypeCheck(obj, PyArrayDescr_Type_);
     }
+    bool PyArray_IsScalar_(PyObject *obj) const {
+        return (bool) PyObject_TypeCheck(obj, PyGenericArrType_Type_);
+    }
 
     unsigned int (*PyArray_GetNDArrayCFeatureVersion_)();
     PyObject *(*PyArray_DescrFromType_)(int);
@@ -157,6 +160,7 @@ struct npy_api {
     PyTypeObject *PyArray_Type_;
     PyTypeObject *PyVoidArrType_Type_;
     PyTypeObject *PyArrayDescr_Type_;
+    PyTypeObject *PyGenericArrType_Type_;
     PyObject *(*PyArray_DescrFromScalar_)(PyObject *);
     PyObject *(*PyArray_FromAny_) (PyObject *, PyObject *, int, int, int, PyObject *);
     int (*PyArray_DescrConverter_) (PyObject *, PyObject **);
@@ -171,6 +175,7 @@ private:
         API_PyArray_GetNDArrayCFeatureVersion = 211,
         API_PyArray_Type = 2,
         API_PyArrayDescr_Type = 3,
+        API_PyGenericArrType_Type = 10,
         API_PyVoidArrType_Type = 39,
         API_PyArray_DescrFromType = 45,
         API_PyArray_DescrFromScalar = 57,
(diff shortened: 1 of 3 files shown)


vanossj:
numpy scalars are not ndarrays. They can be converted to an ndarray (with ndim 0), but they are always converted. Currently setting py:arg().no_convert() doesn't prevent scalar to ndarray conversion in this patch

vanossj:
pybind11 overloads works in [two passes ](https://pybind11.readthedocs.io/en/stable/advanced/functions.html?highlight=noconvert()#overload-resolution-order). Either an object is a perfect match, first pass, or its implicitly convertible, second pass.

consider the following objects:

 code  | numpy type | PEP 3118 scalar?
---------|----------------|-----------------------
`x=np.float64(1)` | nump.float64 | yes 
`y=np.array(1, dtype=float64)` | nump.ndarray | yes 
`z=np.array([1], dtype=float64)` | nump.ndarray | no 

Since `x` is not a `numpy.ndarray` type it will *never* be a perfect match to a `py::array_t` argument. So then the second pass will decide if it can be converted to a `py::array_t` type. `x` could be converted to an `array_t<float>` or `array_t<double>` or `array_t<int>`, so the second pass will call the first of those bindings, not necessarily `array_t<double>`.

`y` and `z` are both `ndarrays` and could match exactly to `py::array_t` overloaded methods

@ax3l because of the peculiarities of numpy scalar implementation and the limited granularity of pybind11 overloading, I don't think the behaviour you want in #1537 is possible. You'll probably have to resort to catching all scalars in a single `py::buffer` overloaded function

ax3l:
@vanossj do you think we can reopen this and make this a python 3+ only feature?

vanossj:
what about something like:
```c++
void overloaded(py::array_t<float> x, bool recursion=false) {
    py::print("got float array");
}

void overloaded(py::array_t<double> x, bool recursion=false) {
    py::print("got float array");
}

void overloaded(py::buffer b, bool recursion=false) {
    if(recursion) {
        throw std::runtime_error("If we recurse here, that means there is a missing array_t overload");
    }
    py::buffer_info info = b.request();
    if (info.ndim == 0) {
        py::array x(b);
        overloaded(x, true); // should get caught by array_t overload
    }
    else {
        throw std::runtime_error("Not a scalar, but not caught by an array_t overloaded(), missing case");
    }
}


PYBIND11_MODULE(example, m) {
    m.def("overloaded", py::overload_cast<py::array_t<float>, bool>(&overloaded), py::arg().noconvert(), py::arg("recursion_flag")=false );
    m.def("overloaded", py::overload_cast<py::array_t<double>, bool>(&overloaded), py::arg().noconvert(), py::arg("recursion_flag")=false );

    // this needs to be the last one
    m.def("overloaded", py::overload_cast<py::buffer, bool>(&overloaded), py::arg().noconvert(), py::arg("recursion_flag")=false );
}
```

The `py::buffer` overload should be the last one, it catches scalars, converts them to arrays, then recalls `overloaded()`, This time it should match to one of the `array_t` overloads.

ax3l:
Yep, that's how I am sorting out things right now as well via a `py::buffer` in Python 3 (and by not supporting scalar arrays at all in Python 2).

But this looks wrong to me, it should be catched in `py::array_t` as well in Python 3, no?

Issue #1554, opened 2018-10-05
clang7 doesn't need experimental anymore 


fantasticfears:
## Issue description

`#include <pybind11/stl.h>` can't compile with clang++-7.0 because optional is no longer experimental.

I tried to put macro detection code in `details/common.h` but it doesn't work.

```
#if defined(__clang__) && __clang_major__ >= 7
#  define PYBIND11_CPP17
#endif
```

## Reproducible example code

```cpp
#include "pybind11/stl.h"

int main()
{
  return 0;
}
```

fantasticfears:
I have encountered this once again with g++ 7.3 shipping with `<optional>`. 

https://github.com/pybind/pybind11/commit/ff6bd092d482b0c4e51aeb7367c3e5ecbc18a9de

```
#  if defined(PYBIND11_CPP14) && (__has_include(<experimental/optional>) && \
                                 !__has_include(<optional>))
```

The problem comes with header files existed at the same time.

fantasticfears:
I used the old package between 2.2.0 ~ 2.2.4. It's fixed in 2.2.4.

PR #1555, opened 2018-10-05
Fix for C++ standard macros

fantasticfears:
This is a possible fix for #1554.

First commit is all about missing macro `PYBIND11_CPP17` where `PYBIND11_CPP14` are.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -45,6 +45,10 @@
 #  endif
 #endif
 
+#if defined(PYBIND11_CPP17) && !defined(PYBIND11_CPP14)
+#  define PYBIND11_CPP14
+#endif
+
 // Compiler version assertions
 #if defined(__INTEL_COMPILER)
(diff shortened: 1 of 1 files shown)


wjakob:
It would be much cleaner to just add 

```cpp
#if defined(PYBIND11_CPP17) || !defined(PYBIND11_CPP14)
#  define PYBIND11_CPP14
#endif
```
in one place. I did not understand your change related to clang -- can you explain what this does?

fantasticfears:
It seems that I used the old package from package manager. It should be fixed somewhere between 2.2.0 ~ 2.2.4. It's rebased!

wjakob:
So -- having taken another look at this PR, I'm not sure I understand why it is needed. All the places which set ``PYBIND11_CPP17`` in ``include/pybind11/detail`` are guaranteed to also set ``PYBIND11_CPP14``..

fantasticfears:
You are right. I think it's part of building system problems. Closing 👍

Issue #1525, opened 2018-09-14
Deadlock with std::function argument


davidhewitt:
## Issue description

I can easily reproduce a deadlock which seems to be related to `gil_scoped_acquire` and `std::function<>` arguments, when being called from a thread other than the main thread.

## Reproducible example code

In particular, this binding code:

```c++
#include <pybind11/pybind11.h>
#include <pybind11/functional.h>

PYBIND11_MODULE(deadlock, m)
{
  m.def("deadlock", [] (std::function<void ()> f) { f(); });
}
```

Deadlocks with the following Python code:

```python
import deadlock
from concurrent.futures import ThreadPoolExecutor

with ThreadPoolExecutor(max_workers=2) as e:
    e.submit(deadlock.deadlock, lambda: None)
    e.submit(deadlock.deadlock, lambda: None)
```

Or also with this:

```python
import deadlock
from concurrent.futures import ThreadPoolExecutor

with ThreadPoolExecutor(max_workers=1) as e:
    e.submit(deadlock.deadlock, lambda: None)

with ThreadPoolExecutor(max_workers=1) as e:
    e.submit(deadlock.deadlock, lambda: None)
```

Something about two calls seems significant, I can't reproduce with just one.

I can spend some time on the weekend writing a patch if you are able to point me to what is probably going wrong :)

Thanks
David

lambdaknight:
You might check out PR #1595. That fixes a deadlock issue I was getting with functions getting called from a non-main thread. Maybe it'll fix yours?

davidhewitt:
I've just seen #1211 has been merged, and I can confirm that the examples I give above now work correctly without deadlocking on master. Exciting!

Issue #1308, opened 2018-03-06
Strange deadlock with pybind11 + Flask w/ app debug mode


skystrife:
## Issue description

The exact setup for this to happen seems really bizarre to me, but here it goes: I get a deadlock when using Flask with its default debug app reloader with my pybind11 module. I've reduced the example down to the smallest thing I can think of that still replicates.

The ingredients:

1. `#include <pybind11/functional.h>`
2. A pybind11-bound function that takes a `std::function`
3. A class defined in the module
5. A Flask app running with `app.run(debug=True)`
6. A route in that Flask app that calls the function from 2.

What really confuses me is that if I remove the class in step 3, the deadlock goes away. Similarly, if the function in step 2 doesn't take something complicated (like just an `int`), the deadlock goes away. If the call to run the app is instead just `app.run()`, the deadlock goes away (similarly if `use_reloader=False` is passed). Something in the combination of all three of these things causes the deadlock when the passed-in function is invoked. Using a default `std::function` parameter doesn't seem to change anything.

Versions:

```
$ python --version
Python 3.6.4

$ gcc --version
gcc (GCC) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

pybind11 `stable` branch (see submodule in reproduction repo)

## Reproducible example code

https://github.com/skystrife/pybind11-bug

fgsalomon:
We are on the same boat. We have a module with lots of callbacks using `std::function` that was working great in our desktop applications. Now, we wanted to implement a REST service on top of it and encountered this deadlock situation. In our case passing `use_reloader=False` didn't work. We have replaced Flask with Falcon and we didn't have this issue. Perhaps is a Flask bug?

Our versions:
```
gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

flask --version
Flask 1.0.2
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:51:32) 
```

PR #1556, opened 2018-10-06
Resolve threading issue with gil_scoped_acquire

davidhewitt:
It was possible for python threads to execute c++ code which would subsequently call gil_scoped_acquire, incorrectly conclude that the gil was not held, and then deadlock.

The most basic example I can build of the failing code is:

```c++

PYBIND11_MODULE(pybind11test, m)
{
  m.def("deadlock", [] (std::function<void ()> f) { f(); });
}
```
and then invoked in python as so:

```python
import pybind11test

import threading

thread1 = threading.Thread(target=pybind11test.deadlock, args=(lambda: None,))
thread1.start()

thread2 = threading.Thread(target=pybind11test.deadlock, args=(lambda: None,))
thread2.start() 

thread1.join()
thread2.join()
```

I believe that this patch will resolve both #1525 and #1308.

To me this is a critical problem, as I often work with python threads. I would be very grateful if this can be reviewed and merged in time for the pybind11 2.3 release.

I am happy to make any changes you feel necessary, including writing a different solution altogether if you have a better one in mind.

Thanks in advance,
David

### DIFF
--- a/tests/test_callbacks.py
+++ b/tests/test_callbacks.py
@@ -105,3 +105,21 @@ def test_function_signatures(doc):
 
 def test_movable_object():
     assert m.callback_with_movable(lambda _: None) is True
+
+
+def test_no_deadlock():
+    # This code would deadlock - see issue 1525
+
+    def f():
+        pass
+
+    import threading
+
+    thread1 = threading.Thread(target=m.callback_void_std_function, args=(f,))
+    thread1.start()
+
+    thread2 = threading.Thread(target=m.callback_void_std_function, args=(f,))
+    thread2.start()
+
+    thread1.join()
+    thread2.join()
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -47,6 +47,13 @@ v2.3.0 (Not yet released)
 * ``pybind11_add_module()``: allow including Python as a ``SYSTEM`` include path.
   `#1416 <https://github.com/pybind/pybind11/pull/1416>`_.
 
+* Resolved an issue where std::function arguments to functions could create a deadlock
+  when called from threads other than the main thread.
+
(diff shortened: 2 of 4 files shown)


tests/test_callbacks.py
```diff
@@ -105,3 +105,21 @@ def test_function_signatures(doc):
 
 def test_movable_object():
     assert m.callback_with_movable(lambda _: None) is True
+
+
+def test_no_deadlock():
+    # This code would deadlock - see issue 1525
```
davidhewitt: I wanted to write a test case, but I am unsure if adding code than can potentially deadlock the test suite is wise. Other solutions warranted, can remove the test case altogether if preferred.

davidhewitt: Looks like at least on CI the deadlock is handled cleanly!

davidhewitt:
Looks like this has only worked for Python 3, not 2 (my dev box is Python 3).

I'd like to give this another shot that works for both platforms. Would be great if anyone familar with this part of the codebase can weigh in on whether I'm heading on the right path before I spend too much time on this.

awaizman1:
Hi David,

This PR is critical to me too, I found out that running my C-extensions which acquire gil from non-main python thread dead-lock.

Do you know when this PR will be accepted?

Thanks a lot!

uentity:
@davidhewitt I'm sure that in the first place we should consider helping each other (and giving ideas) to resolve issues and walk around corner cases we face as users of this great library.
And also I believe in rule that maintainer's interest has strong positive correlation with users' interest :smiley:

So can't wait for this PR to be finished.

davidhewitt:
I've just seen that the master branch has recieved a merge from #1211 which solves exactly what I do here. As such, I'm going to close this PR. Wish I'd managed to discover #1211 previously!

PR #1558, opened 2018-10-07
common: Prevent module re-import duplication in Python3 

### LINKED ISSUE #1559: Python3: Modules can be duplicated if they are imported while being constructed
## Issue description

With my system's version of Python3 (Ubuntu 16.04, Python 3.5.2), if I import a given module while that module is being constructed, it will duplicate the module object.

This generally happens when I have a utility function for binding / registration that is generally used outside of the module, but I used it within the module.

Given that you can import a module from itself while it's being constructed (which is admittedly a bit of a race condition), it'd be nice if it at least didn't duplicate the module.

## Reproducible example code

```c++

class Class { };

PYBIND11_MODULE(example, m) {
  py::class_<Class>(m, "Class");

   py::str name = m.attr("__name__");
   // This 
   py::module::import(std::string(name).c_str());
}
```

## Workaround

Within our code base, I defined a simple (but hacky) macro:
```c++
#if PY_MAJOR_VERSION >= 3
#define PYDRAKE_PREVENT_PYTHON3_MODULE_REIMPORT(variable) \
  { \
    static py::handle variable##_original; \
    if (variable##_original) { \
      variable = py::reinterpret_borrow<py::module>(variable##_original); \
      return; \
    } else { \
      variable##_original = variable; \
    } \
  }
#else  // PY_MAJOR_VERSION >= 3
#define PYDRAKE_PREVENT_PYTHON3_MODULE_REIMPORT(variable)
#endif  // PY_MAJOR_VERSION >= 3
```
This can be put at the very beginning of the module definition.

Submitted a PR here if it's useful to a larger audience.

EricCousineau-TRI:
Resolves #1559

With my system's version of Python3 (Ubuntu 16.04, Python 3.5.2), if I import a given module while that module is being constructed, it will duplicate the module object.

This captures this behavior in a test, and provides a potential workaround.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -171,7 +171,18 @@
 #define PYBIND11_PLUGIN_IMPL(name) \
     extern "C" PYBIND11_EXPORT PyObject *PyInit_##name()
 
+#define PYBIND11_CHECK_REIMPORT(m)                                             \
+    {                                                                          \
+        static int module_key;                                                 \
+        auto& modules = pybind11::get_or_create_shared_data<                   \
+            std::unordered_map<int*, pybind11::handle>>("__pybind11_modules"); \
+        pybind11::handle& original = modules[&module_key];                     \
+        if (original) return original.inc_ref().ptr();                         \
+        original = m;                                                          \
+    }
+
 #else
+// PY_MAJOR_VERSION <= 2
 #define PYBIND11_INSTANCE_METHOD_NEW(ptr, class_) PyMethod_New(ptr, nullptr, class_)
 #define PYBIND11_INSTANCE_METHOD_CHECK PyMethod_Check
 #define PYBIND11_INSTANCE_METHOD_GET_FUNCTION PyMethod_GET_FUNCTION
@@ -198,6 +209,7 @@
         (void)pybind11_init_wrapper();                      \
     }                                                       \
     PyObject *pybind11_init_wrapper()
+#define PYBIND11_CHECK_REIMPORT(m)
 #endif
 
 #if PY_VERSION_HEX >= 0x03050000 && PY_VERSION_HEX < 0x03050200
@@ -285,6 +297,7 @@ extern "C" {
     PYBIND11_PLUGIN_IMPL(name) {                                               \
         PYBIND11_CHECK_PYTHON_VERSION                                          \
         auto m = pybind11::module(PYBIND11_TOSTRING(name));                    \
+        PYBIND11_CHECK_REIMPORT(m)                                             \
         try {                                                                  \
             PYBIND11_CONCAT(pybind11_init_, name)(m);                          \
             return m.ptr();                                                    \
--- a/tests/test_modules.cpp
+++ b/tests/test_modules.cpp
@@ -95,4 +95,9 @@ TEST_SUBMODULE(modules, m) {
 
         return failures;
     });
+
+    // Try re-import.
+    // If this fails, then it will fail at duplicate registration.
+    py::str name = m.attr("__name__");
+    m.attr("self_import_at_construction_time") = py::module::import(std::string(name).c_str());
 }
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -70,3 +70,8 @@ def test_duplicate_registration():
     """Registering two things with the same name"""
 
     assert m.duplicate_registration() == []
+
+
+def test_self_import_at_construction_time():
+    """Ensures that a pybind11 module can import itself at construction time."""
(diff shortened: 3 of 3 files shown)


wjakob:
I can't seem to match up your description to the contents of the PR. In particular, PYBIND11_CHECK_REIMPORT only has a non-empty definition when compiling for Python 2.x. In any case, it would be important to fix the issue both 2.x and 3.x versions of Python (is one not affected for some reason?).

EricCousineau-TRI:
Updated to use `get_or_create_shared_data` to handle interpret restart, leveraging a static variable to kind-of get a unique key for a plugin.

include/pybind11/detail/common.h
```diff
@@ -171,7 +171,18 @@
...
     extern "C" PYBIND11_EXPORT PyObject *PyInit_##name()
 
+#define PYBIND11_CHECK_REIMPORT(m)                                             \
+    {                                                                          \
+        static int module_key;                                                 \
+        auto& modules = pybind11::get_or_create_shared_data<                   \
+            std::unordered_map<int*, pybind11::handle>>("__pybind11_modules"); \
+        pybind11::handle& original = modules[&module_key];                     \
+        if (original) return original.inc_ref().ptr();                         \
+        original = m;                                                          \
```
rwgk, line 181: Uncertain myself, but asking anyway: could it make sense to move this code to an inline function in the `pybind11::detail` namespace?
It would also be nice to add a terse comment (here or in the function) with a hint why this is needed, so that people don't have to dig up this PR to find out.

rwgk (approved):
I agree with @henryiii (slack channel) that this would be good to merge.

YannickJadoul:
This does feel like a rather heavy fix (always keeping track of all modules in a shared variable) for a small window of time where things can go wrong (during import of the module). Just curious if there's no other way; this is not a pybind11-specific problem, so how do other frameworks (or even the raw C API) approach this?

Secondary part of this concern: storing `py::handle`s means that at some point they could start referring to garbage-collected data. Are we safe, here?

wjakob:
I agree with Yannick, this strikes me as to heavy of a change given the benefit that it provides. It's not just the runtime cost of creating yet another hash table (negligible), but the fact that we have to generate and include code for another hash table variant with template arguments `int*, pybind11::handle`.

Can't we do this with a simple global flag to prevent the kind of re-entrant behavior you mention?

EricCousineau-TRI:
Agreed with y'all's points on heavy-handedness and non-reentrancy of the current solution.

Rather than doing that, perhaps the right thing is ensuring that, when available, Python 3 modules use multi-phase initialization by default, as Yannick pointed out in #2548. I would vote for either re-catering this PR towards that issue, or closing this out and re-opening for that?

(I think this would slip on the release schedule, but we're fine in Drake for doing this manually, and not sure if other people have encountered this as a pain point?)

EricCousineau-TRI:
> Can't we do this with a simple global flag to prevent the kind of re-entrant behavior you mention?

Not sure if I understand how this current solution would be solvable via a global flag. @wjakob If you can, can you speak to this?

include/pybind11/detail/common.h
EricCousineau-TRI: Yeah, I can delegate most of this code to a function, and could add a comment, pending other discussions - thanks!

YannickJadoul:
> perhaps the right thing is ensuring that, when available, Python 3 modules use multi-phase initialization by default, as Yannick pointed out in #2548. I would vote for either re-catering this PR towards that issue, or closing this out and opening a new PR addressing that more specifically?

:+1: Sounds good to me. But maybe better to start a new PR whenever someone has a bit of time. That PR should probably also follow/coordinate with #2552 & #2251?

> (I think this would slip on the release schedule, but we're fine in Drake for doing this manually, and not sure if other people have encountered this as a pain point?)

If you would need an easy workaround included in pybind11, I'd be fine with that, but so far, this is the only time I heard about problems in this case. So if you don't mind keeping your own workaround, it might be better to solve this with the multi-phase initialization in 2.7.0 or so?

EricCousineau-TRI:
> So if you don't mind keeping your own workaround, it might be better to solve this with the multi-phase initialization in 2.7.0 or so?

Yup, sounds great! I'm totally fine with us keeping our workaround!

EricCousineau-TRI:
> That PR should probably also follow/coordinate with #2552 & #2251?

Er, did you mean #2551?

YannickJadoul:
> Er, did you mean #2551?

Yes, sorry! (#2251's title is appropriately "Typo" :-D )

PR #1560, opened 2018-10-08
Weird inline/noinline specification breaks gcc.

amosbird:


wjakob:
This specification is perfectly good and there valid reasons (to allow multiple linkage and yet not inline the functions into the function body of callers), and it has always worked on GCC in the past. Please file an issue ticket with more details.

PR #1563, opened 2018-10-10
Allow recursive checkout without https

henryiii:
Pybind11 breaks a master project recursively cloning when it is added as a submodule via ssh and https is not available, because it has a submodule that does not respect the method it was checked out as (https vs. ssh). This makes the submodule relative, so that it will respect the checkout method of the master project.


wjakob:
Nice, I did not realize this is possible.

Issue #1552, opened 2018-10-04
Child Class Method Not Being Called as Expected, Base Class Method Called Instead


bgoodman44:
Hello,

I have a simple test case for polymorphism that isn't working quite right..  I have two C++ classes, Client and Dispatcher.  Dispatcher has a method called  "Dispatch" that takes a Client* as an input, and calls the the clients virtual ProcessEvent() method.  I've made both of these classes available to python using the code below.  I've derived from the Client class as the python class SomeClient, and overridden its ProcessEvent function.  When I call "Dispatch" from anywhere EXCEPT inside the SomeClient::ProcessEvent method, everything works as expected, i get a report from the child class.  But when I call Dispatch from within SomeClient::ProcessEvent, the base classes ProcessEvent is called.  Any ideas what could be going wrong here?  I expect to get an infinite recursion that results in SomeClient::ProcessEvent being called forever.

C++:
```
#include <iostream>
#include "pybind11/pybind11.h"

namespace py = pybind11;
class Dispatcher;
class Client
{
public:
    Client(Dispatcher* disp): PtrD(disp)
    {
        std::cout << "In Client::Client\n";
    }
    virtual ~Client(){};
    virtual void ProcessEvent()
    {
        std::cout << "THIS SHOULDN'T HAPPEN --In Client::ProcessEvent\n";
    }
    Dispatcher* PtrD;
};
```
```
class Dispatcher
{
public:
    Dispatcher()
    {
        std::cout << "In Dispatcher::Dispatcher\n";
    }
    virtual ~Dispatcher(){};

    void Dispatch(Client* client)
    {
        std::cout << "Dispatcher::Dispatch called by " << client << std::endl;
        client->ProcessEvent();
    }
};
```
```
class DispatcherTrampoline : public Dispatcher
{
public:
    using Dispatcher::Dispatcher;
};
```
```
class ClientTrampoline : public Client
{
public:
    using Client::Client;

    void ProcessEvent() override
    {
        PYBIND11_OVERLOAD(void,Client,ProcessEvent,);
    }
};
```
```
PYBIND11_MODULE(libTestCase,mod)
{
    py::class_<Client,ClientTrampoline> cli(mod,"Client");
    cli.def(py::init<Dispatcher* >());
    cli.def("ProcessEvent",&Client::ProcessEvent);
    cli.def_readwrite("PtrD",&Client::PtrD);

[...]

Issue #1546, opened 2018-09-27
shared_ptrs to Python-derived instances outlive the associated Python object


cdyson37:
## Issue description
Storing shared_ptrs to Python-derived instances for later execution causes virtual functions to fail if the Python instance has subsequently been destroyed.

Tested on master: e7761e3 / Python 3.6 / gcc 7.2

### C++
```c++
#include <pybind11/pybind11.h>

namespace py = pybind11;

struct Base
{
  virtual void f (int i) const = 0;
  virtual ~Base () { };
};

struct Trampoline : Base
{
  void f (int i) const override
  {
    PYBIND11_OVERLOAD_PURE_NAME (
      void,
      Base,
      "f",
      impl,
      /* args */
      i);
  }
};

struct Holder
{
  Holder (std::shared_ptr<Base> ptr)
  : ptr (std::move (ptr))
  { }

  std::shared_ptr<Base> ptr;
};

void takes_holder (const Holder &h)
{
  h.ptr->f (42);
}

PYBIND11_MODULE (testmodule, m)
{
  py::class_<Base, Trampoline, std::shared_ptr<Base>>
    base_wrap (m, "Base");

  base_wrap.def (py::init<> ());

  py::class_<Holder> holder_wrap (m, "Holder");
  holder_wrap.def (py::init<std::shared_ptr<Base>> ());

  m.def ("takes_holder", &takes_holder);
}
```

### Python
```python
import testmodule

class PythonDerived (testmodule.Base):
    def __init__ (self):
        testmodule.Base.__init__ (self)

    def f (self, i):
        print ("Called with: {}".format (i))


pd = PythonDerived ()
holder = testmodule.Holder (pd)
del pd # Comment this out and it works
testmodule.takes_holder (holder)
```
### Output
```
Traceback (most recent call last):
  File "./test.py", line 16, in <module>
    testmodule.takes_holder (holder)
RuntimeError: Tried to call pure virtual function "Base::f"
```
### Output (with 'del' removed)
```
Called with: 42
```
## Workaround:
Add the following (tweaked by @iwanders - thanks)
```c++
namespace pybind11::detail {

  template<>
  struct type_caster<std::shared_ptr<Base>>
  {
    PYBIND11_TYPE_CASTER (std::shared_ptr<Base>, _("Base"));

    using BaseCaster = copyable_holder_caster<Base, std::shared_ptr<Base>>;

    bool load (pybind11::handle src, bool b)
    {
      BaseCaster bc;
[...]

Issue #1389, opened 2018-05-09
Ownership of python objects inheriting from C++ classes


florianwechsung:
Hi,

I have an issue with the ref counting of python objects that are passed to C++. 
I have a C++ class `Vector` that the use can inherit from (hence I also have a `PyVector` trampoline class). The user-defined python class should then be passed back to C++ where some actions on this class are performed. Unfortunately, it seems that python takes complete ownership of the class, so that it gets deleted even if the C++ side still needs it.

MFE:

```
#include <pybind11/pybind11.h>
#include <iostream>
#include <memory>

namespace py = pybind11;

class Vector : public std::enable_shared_from_this<Vector> {
    public:
        Vector() {}
        virtual void print() {
            std::cout << "Vector" << std::endl;
        }
};

class PyVector : public Vector {
    public:
        virtual void print() override {
            PYBIND11_OVERLOAD(void, Vector, print,);
        }
};

class Foo {
    public:
        Foo(std::shared_ptr<Vector> vec) : _vec(vec){}
        std::shared_ptr<Vector> _vec;
        void print() {
            _vec -> print();
        }
};

PYBIND11_MODULE(cmake_example, m) {

    py::class_<Vector, std::shared_ptr<Vector>,
        PyVector>(m, "Vector")
            .def(py::init<>());

    py::class_<Foo, std::shared_ptr<Foo>>(m, "Foo")
        .def(py::init<std::shared_ptr<Vector>>())
        .def("print", &Foo::print);

}
```

Python code:
```
import cmake_example as m

class MyVector(m.Vector):

    def print(self):
        print("MyVector")

    def __del__(self):
        print("I was deleted")

def make_foo():
    x = MyVector()
    foo = m.Foo(x)
    foo.print()
    return foo

myfoo = make_foo()
myfoo.print()

```
Output:
```
MyVector
I was deleted
Vector
```
Expected output:
```
MyVector
MyVector
I was deleted
```

jagerman:
This is, unfortunately, a limitation of pybind: we keep the C++ instance alive, but can't keep the Python instance alive.

[...]

PR #1566, opened 2018-10-14
Fix Python object lifetimes associated with shared_ptrs

### LINKED ISSUE #1333: [BUG] Problem when creating derived Python objects from C++ (inheritance slicing)
EDIT(eric): Adding a tracking list.

- Enumerate possible solutions and drawbacks, e.g.
  - [ ] #2757 Using `py::wrapper<>` and GC ressurrection
  - [ ] Others...
- Identify best path forward, implement.

---

## Issue description

Hello,
Consider deriving a C++ class in Python and overloading virtual functions, including a 'Clone' function.
When passing a derived Python object to C++ without holding a reference (like calling Clone()), the information about the Python type is lost. In this case the wrong base-class function is called. More complex cases may result in a SEGFAULT.

Is there any way to work around/fix this without causing memory leaks?

Thanks!

### test.cpp
```cpp

// test.cpp
#include <Python.h>

#include <iostream>
#include <memory>

using std::cout;
using std::endl;
using std::unique_ptr;
using std::shared_ptr;
using std::make_shared;

namespace py = pybind11;

class Base {
public:
    virtual void Print() { 
        cout << "Base!" << endl;
    }

    virtual shared_ptr<Base> Clone() {
        return make_shared<Base>();
    }
};

class PyBase : public Base {
public:
    using Base::Base; // Inherit constructors
    void Print() override { PYBIND11_OVERLOAD(void, Base, Print, ); }
    shared_ptr<Base> Clone() override { PYBIND11_OVERLOAD(shared_ptr<Base>, Base, Clone, ); }
};

PYBIND11_MODULE(libtest, m) {
    py::class_<Base, py::wrapper<PyBase>, shared_ptr<Base>>(m, "Base")
      .def(py::init<>())
      ;

    m.def("Test", [] ( shared_ptr<Base> b ) {
          shared_ptr<Base> d = b->Clone();
          // Python object of d is already dead!
          d->Print();  // prints "Base!"
    });
}

```

### test.py
```py
from libtest import *

class Derived(Base):
    def __init__(self):
        super().__init__()

    def Print(self):
        print("Derived!")

    def Clone(self):
        return Derived()

Test(Derived())

```

cdyson37:
This is a straw man solution to #1546, based on the patch included in the report. It also fixes  #1333 and #1389 (note that "I was deleted" doesn't appear unless you conduct the test inside function scope).

Probably a pre-requisite for #1552 but does not fix it (I have to be careful wording this or github gets the wrong idea).

Questions:
* This means that every py::object->std::shared_ptr<T> cast keeps the py::object alive. This appears to be sufficient to solve the problem, but it might not be entirely necessary. For instance, some shared_ptr<T> casts might be associated with the original base class, and not require the py::object to be kept around. Suggestions for limiting the scope of the fix welcome.
* Does the test look okay? I'm happy to rename, move or change it in whatever way people think best.

### DIFF
--- a/tests/test_virtual_functions.py
+++ b/tests/test_virtual_functions.py
@@ -375,3 +375,18 @@ def test_issue_1454():
     # Fix issue #1454 (crash when acquiring/releasing GIL on another thread in Python 2.7)
     m.test_gil()
     m.test_gil_from_thread()
+
+def test_issue_1546():
+    # Fix issue #1546 (Python object not kept alive by shared_ptr)
+    somelist = []
+    class Derived(m.SharedPtrBase):
+        def __init__(self):
+            super(Derived, self).__init__()
+
+        def f(self, value):
+            somelist.append (value)
+            print("Right one", 42)
+
+    holder = m.SharedPtrHolder(Derived())
+    holder.run(42);
+    assert somelist == [42]
--- a/tests/test_virtual_functions.py
+++ b/tests/test_virtual_functions.py
@@ -388,5 +388,7 @@ def test_issue_1546():
             print("Right one", 42)
 
     holder = m.SharedPtrHolder(Derived())
+    # At this point, our 'Derived' instance has gone out of scope in Python but
+    # is being kept alive by a shared_ptr inside 'holder'.
     holder.run(42);
     assert somelist == [42]
--- a/tests/test_virtual_functions.py
+++ b/tests/test_virtual_functions.py
@@ -376,19 +376,21 @@ def test_issue_1454():
     m.test_gil()
     m.test_gil_from_thread()
 
+
 def test_issue_1546():
     # Fix issue #1546 (Python object not kept alive by shared_ptr)
     somelist = []
+
     class Derived(m.SharedPtrBase):
         def __init__(self):
             super(Derived, self).__init__()
 
         def f(self, value):
-            somelist.append (value)
+            somelist.append(value)
             print("Right one", 42)
 
     holder = m.SharedPtrHolder(Derived())
     # At this point, our 'Derived' instance has gone out of scope in Python but
     # is being kept alive by a shared_ptr inside 'holder'.
-    holder.run(42);
+    holder.run(42)
     assert somelist == [42]
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -50,6 +50,9 @@ v2.3.0 (Not yet released)
 * ``pybind11/stl.h`` does not convert strings to ``vector<string>`` anymore.
   `#1258 <https://github.com/pybind/pybind11/issues/1258>`_.
 
+* fixed shared_ptrs to Python-derived instances outliving the associated Python object
+  `#1546 <https://github.com/pybind/pybind11/issues/1546>`_.
+
 v2.2.4 (September 11, 2018)
 -----------------------------------------------------
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1503,37 +1503,36 @@ protected:
 template <typename T>
 class type_caster<std::shared_ptr<T>>
 {
-    PYBIND11_TYPE_CASTER (std::shared_ptr<T>, _(PYBIND11_STRING_NAME));
+    PYBIND11_TYPE_CASTER(std::shared_ptr<T>, _(PYBIND11_STRING_NAME));
 
     // Re-use copyable_holder_caster
     using BaseCaster = copyable_holder_caster<T, std::shared_ptr<T>>;
 
-    bool load (pybind11::handle src, bool b)
+    bool load(pybind11::handle src, bool b)
     {
         BaseCaster bc;
-        bool success = bc.load (src, b);
-        if (!success)
-        {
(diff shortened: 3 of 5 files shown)


iwanders:
Thanks for this solution to the problem, I used it in a multithreaded situation and ran into segfaults during destruction from the C++ side. Turns out that if the last pointer gets destroyed from another thread, the Python object may be cleaned up without having the GIL. This will caused reproducible segfaults in my case. Modifying line 1552 to provide a de-allocator lambda like:
```cpp
 auto py_obj_ptr = std::shared_ptr<py::object> (new py::object{py_obj}, [](auto py_object_ptr) {
  pybind11::gil_scoped_acquire gil;
  delete py_object_ptr;
});
```
Will ensure we hold the GIL when we delete the Python object. In my case it did not matter whether that object had a `__del__` method, the segfault occurs regardless. With this small modification this solution works great. :tada:

cdyson37:
@iwanders - good point! I've incorporated that. For a horrible moment I thought I would have to move gil_scoped_acquire to its own header to work around a circular dependency, but at instantiation time the definition of that class has already been seen, so I settled for a forward declaration.

mmodenesi:
Is this line of work still active? I'm really interested in a solution to this problem and before I copy paste or taylor my own hack, I would like to know what are the chances of something like this being incorporated to mainstream. Thanks so much!

tests/test_virtual_functions.py
```diff
@@ -375,3 +375,22 @@ def test_issue_1454():
...
+    class Derived(m.SharedPtrBase):
+        def __init__(self):
+            super(Derived, self).__init__()
+
+        def f(self, value):
+            somelist.append(value)
+            print("Right one", 42)
+
+    holder = m.SharedPtrHolder(Derived())
+    # At this point, our 'Derived' instance has gone out of scope in Python but
```
eacousineau, line 393: Just to be sure, could you add a `pytest.gc_collect()` here?

tests/test_virtual_functions.py
```diff
@@ -375,3 +375,22 @@ def test_issue_1454():
...
+    # is being kept alive by a shared_ptr inside 'holder'.
+    holder.run(42)
+    assert somelist == [42]
```
eacousineau, line 396: Is there a way that you can check and ensure that all objects get destroyed when the go out of scope?

See usages of `ConstructorStats` for how to do so - I'm happy to point out more examples!

eacousineau:
This looks great!
Is there anything I can do to help?

FWIW I had started out writing a repro on `master` here (no fix):
https://github.com/pybind/pybind11/compare/master...EricCousineau-TRI:issue-1389-alias-repro
This was before I actually read the issue thread and found this PR :P I dunno if it's helpful, as it's basically a longer version of this (but just based on the doc example of `Cat` / `Animal`).

eacousineau:
@wjakob @rwgk Any chance we could snag some of y'all's time to look at this PR?

eacousineau:
FTR Here are some (if not all?) of the issues that could be *truly* fixed by this PR (I think)

#1120
#1145
#1333
#1389
#1546
#1552
#1774
#1902
#1937

\cc @pvallet

tests/test_virtual_functions.py
eacousineau: (The thing I'm still not sure of is if this creates a sort of reference cycle, and effectively cannot be GC'd until the interpreter dies...)

wjakob:
I agree that this is a severe problem that we should definitely fix. One aspect to consider is that only a small subset of users that rely on ``std::shared_ptr`` will also provide trampolines (I would generally consider this a fairly advanced feature of pybind11). The current PR penalizes everyone, and will therefore need further adjustments.

I think it may be possible to find out whether the underlying type uses a trampoline or not and only run the special case then. See in particular the `has_alias` variable in the `pybind11::class_` constructor. This might need to be stashed so that it's available when you need it.

Other aspects: running the full base caster to make a `std::shared_ptr` that is then destructed later strikes me as inefficient about the current implementation.

Quite a bit of this code doesn’t depend on the template type. One issue with a template library like pybind11 is that this code will be replicated thousands of times when creating bindings for large codebases. Thus, whenever possible, it’s very useful to factor out things into non-templated functions that are called from more specific template classes.

cdyson37:
Hi @wjakob thanks for taking a look! As I say in the original PR, it's just a test and a straw man solution; no doubt it can be much improved. Likewise thank you to @eacousineau for taking a look and for suggestions.

Where should we go from here? Do you or one of the more regular developers want to have a go at improving this fix, or coming up with a different one? Or I can have a go at improving it? I'm more than happy to leave it to an expert though.

virtuald:
Hadn't noticed this previously, but superceded by https://github.com/pybind/pybind11/pull/2839

cdyson37:
Let's close this in favour of https://github.com/pybind/pybind11/pull/2839 - thanks to everyone who commented, hopefully the other PR will be merged soon!

fangchaooo:
but I still find this issue https://github.com/pybind/pybind11/issues/5560

Issue #1339, opened 2018-03-28
Enum Operators should be marked with py::is_operator


davidhewitt:
## Issue description

The `enum_` builtin operators do not have the `is_operator()` marker, e.g. 

https://github.com/pybind/pybind11/blob/6d0b4708c6eb1be31b943419a73aec00095a2d64/include/pybind11/pybind11.h#L1410

This results in TypeError being thrown when comparing an enum_ and something else, rather than False.

Perhaps the definitions can be anyway now be simplified to e.g. `def(py::self == py::self)`  etc, which includes `is_operator()` implicitly?

## Reproducible example code
N/A

tarcisiofischer:
I also noticed this. Here is an example:

C++ Code:
```
    {
        enum class Color {
            red,
            green,
            blue
        };
        py::enum_<Color>(m, "Color")
            .value("red", Color::red)
            .value("green", Color::green)
            .value("blue", Color::blue)
            .export_values();
    }
```

Python code: `m.Color.red == 1`
Expected result: `False`
Obtained result:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __eq__(): incompatible function arguments. The following argument types are supported:
    1. (self: m.Color, arg0: m.Color) -> bool

Invoked with: Color.red, 1
```

@davidhewitt Are you working on this?

davidhewitt:
I'm not; I took a brief look at the source code and I got the impression from the tests that this might be by design:

https://github.com/pybind/pybind11/blob/55dc131944c764ba7e30085b971a9d70531114b3/tests/test_enum.py#L78

Personally, if the pybind owners are willing for this to change I think making the behaviour non-throwy is an improvement for consistency. But if they have reasons for this to be a special case then I can accept that.

tarcisiofischer:
Just leaving a new bump for this issue. Is there anyone available to tell if this is expected behavior? If not, I can tackle the problem and open a pull request, but I need to know if it'll be accepted and merged.

wjakob:
Did you give ``py::is_arithmetic()`` a try? It makes the comparisons more relaxed.

tarcisiofischer:
Note that, if I'm not mistaken, the other way around will return `False`, as expected:

```
a = MyEnum.RED  # a py::enum_
b = Foo()  # Some other type
a == b      # I'm expecting False, but instead TypeError is raised

b == a      # False
```

[...]

PR #1571, opened 2018-10-22
Fix scoped enums comparison for equal/not equal cases (#1339)

tarcisiofischer:
Fixes the expected behavior for #1339.


wjakob:
Beautiful, thanks!

PR #1578, opened 2018-10-25
Add PYBIND11_ prefix to the THROW macro to prevent name collisions.

mwu-tow:
The https://github.com/pybind/pybind11/commit/54eb8193e512f35edca99753155c3a804ed5c176 commit introduced a `THROW` macro. Unfortunately, some of my projects use macro by that name as well, causing compilation errors (pybind's one is recognized as redefinition) with current pybind11.

This PR fixes the issue by prefixing the macro with proper prefix, so user code won't break it that easily.


wjakob:
Good catch, thanks!

PR #1581, opened 2018-10-30
Make `overload_cast_impl` available in C++11 mode.

schmeisser:
Narrow the scope of the `#if defined(PYBIND11_CPP14)` block around overload_cast to only
cover the parts where C++14 is stricly required. Thus, the implementation in
`pybind11::details::overload_cast_impl` is still available in C++11 mode.

In C++11 mode `overload_cast_impl` can still be used like `overload_cast` with just an additional set of parentheses.


bstaletic:
Can you update the docs to reflect your changes and write a test case to show that this really works?

schmeisser:
I've updated a test which previously used `static_cast`s in C++11 mode to use a mixture of `static_cast` and `overload_cast_impl`. Also added a corresponding sentence to the docs.

bstaletic:
Great! Thanks for updating the PR.

wjakob:
Looks great, thanks!

PR #1584, opened 2018-10-31
Update documentation for initialize_interpreter

joshkel:
Add a caveat about the program terminating if initializing the interpreter fails. (This bit me during development, so I thought it would be worth pointing out.)

Add a detailed link to Python 3 documentation.


wjakob:
Looks good, thanks!

PR #1595, opened 2018-11-06
Fix async Python functors invoking from multiple C++ threads (#1587)

### LINKED ISSUE #1587: Crash when Python lambdas (with captured state) are called from multiple C++ threads (fix inside)
# Intro
Current `pybind11` functional `type_caster` support invoking callbacks passed from Python in async way, i.e. from multiple C++ threads. It implements that by holding GIL while functor is being executed according to following essential code from `type_caster::load()` that initializes `value` member:
```C++
...
		value = [func](Args... args) -> Return {
			gil_scoped_acquire acq;
			object retval(func(std::forward<Args>(args)...));
			/* Visual studio 2015 parser issue: need parentheses around this expression */
			return (retval.template cast<Return>());
		};
```
Notice the sentence `gil_scoped_acquire acq;` that captures and releases GIL in RAII fashion.

# Problem
Problem with code above is that destruction of `value` (and captured Python functor `func`) happens **after** GIL has been released. Then, if functor `func` is for example Python lambda that captures some variables, these variables are being freed (reference counter decremented) when GIL is no longer held.

Notice that all this process of functor invoke and destruction can execute in some worker C++ thread and that leads to UB (immediate terminate in my experience).

Problem isn't arising If `func` is pure function or a stateless lambda.

# Solution
I've made a workaround to this issue by replacing the code above with the following:
```C++
...
		// dynamically allocated lambda that actually invokes passed functor
		auto f = new auto([func](Args... args) -> Return {
			object retval(func(std::forward<Args>(args)...));
			/* Visual studio 2015 parser issue: need parentheses around this expression */
			return (retval.template cast<Return>());
		});
		if(!f) return false;

		// ensure GIL is released AFTER functor destructor is called
		value = [f](Args... args) -> Return {
			gil_scoped_acquire acq;
			(*f)(std::forward<Args>(args)...);
			delete f;
		};
```
Basically what it does -- it keeps captured GIL until functor is finished and completely destructed. This approach **completely cures the problem**.
However the downside is that it dynamically allocates the inner lambda (that actually invokes `func`). With C++17 lambda will be constructed in-place without copy/move involved. But still this proposal may be sub-optimal.

So, I'm calling for core devs here for looking into this issue because it leads to rather severe limitations of Python callbacks usage.

uentity:
Ensure GIL is released after functor destructor finished (not only
during functor execution as in previous implementation).

This fixes #1587

### DIFF
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -54,9 +54,20 @@ public:
             }
         }
 
-        value = [func](Args... args) -> Return {
+        // ensure GIL is held during functor destruction
+        struct func_handle {
+            function f;
+            func_handle(function&& f_) : f(std::move(f_)) {}
+            func_handle(const func_handle&) = default;
+            ~func_handle() {
+                gil_scoped_acquire acq;
+                function kill_f(std::move(f));
+            }
+        };
+
+        value = [hfunc = func_handle(std::move(func))](Args... args) -> Return {
             gil_scoped_acquire acq;
-            object retval(func(std::forward<Args>(args)...));
+            object retval(hfunc.f(std::forward<Args>(args)...));
             /* Visual studio 2015 parser issue: need parentheses around this expression */
             return (retval.template cast<Return>());
         };
--- a/tests/test_callbacks.cpp
+++ b/tests/test_callbacks.cpp
@@ -10,6 +10,7 @@
 #include "pybind11_tests.h"
 #include "constructor_stats.h"
 #include <pybind11/functional.h>
+#include <thread>
 
 
 int dummy_function(int i) { return i + 1; }
@@ -146,4 +147,22 @@ TEST_SUBMODULE(callbacks, m) {
     py::class_<CppBoundMethodTest>(m, "CppBoundMethodTest")
         .def(py::init<>())
         .def("triple", [](CppBoundMethodTest &, int val) { return 3 * val; });
+
+    // test async Python callbacks
+    using callback_f = std::function<void(int)>;
+    m.def("test_async_callback", [](callback_f f, py::list work) {
+        // make detached thread that calls `f` with piece of work after a little delay
+        auto start_f = [f](int j) {
+            auto invoke_f = [f, j] {
+                std::this_thread::sleep_for(std::chrono::milliseconds(50));
+                f(j);
+            };
+            auto t = std::thread(std::move(invoke_f));
+            t.detach();
+        };
+
+        // spawn worker threads
+        for (auto i : work)
+            start_f(py::cast<int>(i));
+    });
 }
--- a/tests/test_callbacks.py
+++ b/tests/test_callbacks.py
@@ -1,5 +1,6 @@
 import pytest
 from pybind11_tests import callbacks as m
+from threading import Thread
 
 
 def test_callbacks():
@@ -105,3 +106,31 @@ def test_function_signatures(doc):
 
 def test_movable_object():
     assert m.callback_with_movable(lambda _: None) is True
+
+
+def test_async_callbacks():
+    # serves as state for async callback
+    class Item:
+        def __init__(self, value):
+            self.value = value
+
+    res = []
+
+    # generate stateful lambda that will store result in `res`
+    def gen_f():
+        s = Item(3)
+        return lambda j: res.append(s.value + j)
+
+    # do some work async
+    work = [1, 2, 3, 4]
+    m.test_async_callback(gen_f(), work)
+    # wait until work is done
+    from time import sleep
+    sleep(0.5)
+    assert sum(res) == sum([x + 3 for x in work])
+
+
+def test_async_async_callbacks():
+    t = Thread(target=test_async_callbacks)
+    t.start()
+    t.join()


wjakob:
I'm a bit confused by the change, can you clarify? Specifically, I'm wondering if/why it is safe to ``std::move()`` the function within the lambda function. It seems to me that for lambdas with state, this operation will only work the first time. Afterwards, the lambda closure object will contain an invalid function object, which will potentially trigger undefined behavior the next time the function is called. What am I missing?

wjakob:
FWIW what's also missing here is a testcase that would have lead to breakage in the previous implemenation.

uentity:
@wjakob Your first comment is absolutely right, I missed that point, s.t. all functors became single-shot (can be called only once and fail/crash 2nd time).

Fixed that in next commit. Had to introduce local "function handle" struct with custom destructor. Seems that my code is working now.

Considering testcase: I can write a Python test with statefull lambda, and it should crash Python (and I suppose we can't catch/prevent it in testing code) without this PR and work fine with it.

uentity:
@wjakob I've made a test case and squashed `functional` changes into a single commit.
Please, take a look.

[NOTE] test leads to segfault (on my machine) if run without latest commit (with previous `functional` version).

uentity:
@wjakob there seems to be a problem with `gil_scoped_aquire`.
If async callbacks are launched from separate (non-main) Python thread, then following code will lead to deadlock:
```C++
~func_handle() {
	gil_scoped_acquire acq;
	function kill_f(std::move(f));
}
```
If I replace it with conventional Python API for acquiring GIL (straight from documentation):
```C++
~func_handle() {
	PyGILState_STATE gstate;
	gstate = PyGILState_Ensure();
	{
		function kill_f(std::move(f));
	}
	PyGILState_Release(gstate);
}
```
 then everything works without issues.
Consider the latest couple of commits.

davidhewitt:
Hello @wjakob @uentity, I've authored a separate PR #1556 which resolves the deadlock problem with `gil_scoped_acquire`. You might wish to check that out!

Below is my summary of what I understand of the problems these PRs are trying to solve.

`std::function<...>` extracted from pybind11 typically is some signature like `std::function<int(int)>` which has no obvious association to python types and can easily be passed to portions of the codebase which are not aware to things like the GIL. People doing this then expect these objects to be thread safe, however:

1. If C++ code spawns a thread and copies the function handle to give to that new thread, then in that new thread the function will correctly capture the GIL during calls but not during destruction. That is what this PR solves.
2. If Python spawns a thread and calls a pybind11-defined function which takes one of these `std::function` types as an argument, then on the first construction of a `gil_scoped_acquire` object we get a deadlock. That is what #1556 solves.

I would love to contribute in any way possible to help resolve these!

uentity:
Hi @davidhewitt!
If you look at my last commit on this topic, then you can figure out that I was trying to remedy the consequences of problem you described in item 2. It's great that you've fixed `gil_scoped_aquire` in first place!

With you PR #1556 applied I can bring back my original implementation of `func_handle` destructor in this PR, i.e.:
```C++
~func_handle() {
	gil_scoped_acquire acq;
	function kill_f(std::move(f));
}
```

Very soon after I thought my issue #1587 is solved I've faced the issue of your PR #1566 and fixed it by replacing `gil_scoped_acquire` with Python GIL API calls (see above). But my fix isn't perfect, because if doesn't follow RAII (for example, destructor of functor `f` throws...).

BTW can you give any comment on comparative "heaviness" of capturing GIL solutions I noted [earlier](https://github.com/pybind/pybind11/pull/1595#issuecomment-438768465)? If I look at `gil_scoped_acquire` implementation, I have a (deceptive?) feeling that it is "heavy" :-)

davidhewitt:
I understand `gil_scoped_acquire` is slightly more complicated because it is designed to allow python to be migrated to a different OS thread. I think because of this it's probably advisable to use `gil_scoped_acquire` everywhere. Plus RAII :)

NB I see that #1211 has just been merged which was exactly the same thing that I was trying to solve. So I think that `gil_scoped_acquire` will work properly for you in this PR if you rebase.

include/pybind11/functional.h
```diff
@@ -54,9 +54,20 @@ struct type_caster<std::function<Return(Args...)>> {
...
 
-        value = [func](Args... args) -> Return {
+        // ensure GIL is held during functor destruction
+        struct func_handle {
+            function f;
+            func_handle(function&& f_) : f(std::move(f_)) {}
+            func_handle(const func_handle&) = default;
+            ~func_handle() {
+                gil_scoped_acquire acq;
+                function kill_f(std::move(f));
```
davidhewitt: It might be more clear to explicitly reset the function handle and decrease the reference count: `f.release().dec_ref();`

uentity: Agreed. Done.

uentity:
@wjakob style check is failing because of non-existent URL:
```
$ curl -fsSL ftp://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.12.linux.bin.tar.gz | tar xz
curl: (6) Could not resolve host: ftp.stack.nl
```

include/pybind11/functional.h
uentity: @davidhewitt causes strange test failure on VS with Python 3.6 on x86 platform, so reverted back to original version.

hammer498:
I was also affected by this and your patch has unblocked me.  Thank you @uentity!

uentity:
@hammer498 you're welcome =)

wjakob:
This looks good to me now.

PR #1606, opened 2018-11-12
FAQ addition

akroizer:
Add possible solution to ImportError issue


wjakob:
super-minor: there are some trailing spaces that cause the CI bot to fail. Could you remove those?

PR #1608, opened 2018-11-13
Fix Intel C++ compiler warnings on Windows

Baljak:
_MSC_VER is also defined when using ICC on Windows so Intel compiler check has to be done before MSVC to disable warnings properly.

In addition, warning 1786 related to strdup is also disabled (same as 4996 for MSVC).

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -10,7 +10,17 @@
 
 #pragma once
 
-#if defined(_MSC_VER)
+#if defined(__INTEL_COMPILER)
+#  pragma warning push
+#  pragma warning disable 68    // integer conversion resulted in a change of sign
+#  pragma warning disable 186   // pointless comparison of unsigned integer with zero
+#  pragma warning disable 878   // incompatible exception specifications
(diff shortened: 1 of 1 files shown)


wjakob:
Why did you change the syntax from ``warning(disable:)`` to ``warning disable``?
We *definitely* shouldn't disable the warnings globally.

Baljak:
It seems `warning disable` is the prefered syntax on Intel compiler. See [https://software.intel.com/en-us/forums/intel-c-compiler/topic/746949](here) and [https://stackoverflow.com/questions/35228509/how-to-disable-warning-using-a-pragma-with-intel-c-compiler](here).

For the change related to "pragma warning pop", it was already done like that for Intel C++ compiler on Linux. I only changed it to be the same on Windows. If we don't do that, warning aren't disabled, (another Intel compiler weird issue?)

wjakob:
Ah, that's a good point. I remember the issue now -- the Intel compiler can generate warnings for any template if the place of usage does not mask the warning as well.

wjakob:
Merged, thanks!

PR #1609, opened 2018-11-15
Adds type_caster support for std::deque.

voxmea:
* Adds std::deque to the types supported by list_caster in stl.h.
* Adds a new test_deque test in test_stl.{py,cpp}.
* Updates the documentation to include std::deque as a default
  supported type.


wjakob:
Looks good, thanks!

PR #1617, opened 2018-11-21
Add convenient "c_str" method for "str"

martinRenou:
I need a way to retrieve the string value without having to make a copy by casting to an `std::string`, something like `c_str` is perfect for my use case.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -898,15 +898,9 @@ public:
     explicit str(handle h) : object(raw_str(h.ptr()), stolen_t{}) { }
 
     operator std::string() const {
-        object temp = *this;
-        if (PyUnicode_Check(m_ptr)) {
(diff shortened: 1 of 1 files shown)


jagerman:
This appears to break numerous tests; it's also exposing a raw C-str returned from (and owned by!) Python in a fairly low-level way, which can lead to easy-to-make mistakes, and is generally something we try not to explicitly expose in pybind (nothing stops you from making your own call to the C API, of course, by casting the `py::str` or `py::object` to a `(PyObject *)`.  You just need to be careful when doing that.

But an easier solution, if you are using (or can use) C++17, is just to take the argument as a `std::string_view` (or if this is coming from some attribute rather than an argument, by casting the object to a `std::string_view`).

martinRenou:
Thanks for your answer, I'm not sure I understand why tests are broken. I think I'll go for a custom call to the Python C API directly in my code. I wanted to avoid using the Python C API in my code in order to keep a human-readable code using only pybind11! Anyway, thanks, I close the PR.

PR #1633, opened 2018-12-12
Fixed small typo

ohm314:
I think this particular method binding should not be done with `PYBIND11_OVERLOAD_PURE` but instead `PYBIND11_OVERLOAD`.


wjakob:
LGTM.

PR #1639, opened 2018-12-18
Rename `module` to `module_`

### LINKED ISSUE #1593: Potential collision with C++20 `module` keyword
Hi everyone,

I just wanted to bring to your attention that the name `module` is likely to become a [keyword](https://en.cppreference.com/w/cpp/keyword) in C++20, which I believe would cause a collision with `pybind11::module`. Maybe you already know this, maybe not, but I thought I'd better write something here just in case.

Here is an interesting discussion from Vulkan facing similar problems: https://github.com/KhronosGroup/Vulkan-Docs/issues/568

Related links:
https://www.reddit.com/r/cpp/comments/4b9qru/what_exactly_is_the_proposed_module_system_how/
https://clang.llvm.org/docs/Modules.html

EricCousineau-TRI:
Avoids potential conflict with C++20 keyword

Resolves #1593
Seemed simple enough, just a quick pass using QtCreator and some follow-up passes. Doesn't update `*.rst` docs.

\cc @dalboris

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -526,17 +526,17 @@ inline PyObject* make_new_python_type(const type_record &rec) {
 #endif
     }
 
-    object module;
+    object module_;
     if (rec.scope) {
         if (hasattr(rec.scope, "__module__"))
-            module = rec.scope.attr("__module__");
+            module_ = rec.scope.attr("__module__");
         else if (hasattr(rec.scope, "__name__"))
-            module = rec.scope.attr("__name__");
+            module_ = rec.scope.attr("__name__");
     }
 
     auto full_name = c_str(
 #if !defined(PYPY_VERSION)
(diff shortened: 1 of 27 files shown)


EricCousineau-TRI:
Seems like other PRs are suffering from the same issues, e.g. #1641.

@jagerman TravisCI failures just seem like a URL flake. Any chance you have an idea on the MSVC front?
(Regarding this PR, I'm fine letting it stagnate for a while, but just wanted to emphasize that other PRs may be blocked by this.)

bstaletic:
CI is fixed, but some tests are flaky, so try rebasing onto master.

EricCousineau-TRI:
Done - thanks!

henryiii:
Looks like module is okay to keep: https://en.cppreference.com/w/cpp/keyword (it’s a special identifier).

EricCousineau-TRI:
Sweet! Closing.

PR #1641, opened 2018-12-18
Add getters for exception type, value and traceback

martinRenou:
In order to retrieve the exception `type`, `value` and `trace`, we can use `sys.exc_info()` in Python. But using pybind11 it's not that easy currently.
When an exception occurs in Python code it's directly converted to `py::error_already_set` and it holds the error. Which means that if we want to retrieve the `type`, `value` and `trace` we first need to restore the error using the `.restore()` method, and then call `PyErr_Fetch`, which is not ideal.
I added accessors to objects describing the error in `py::error_already_set`.

Please tell me what you think :)

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1987,12 +1987,12 @@ class gil_scoped_release { };
 #endif
 
 error_already_set::~error_already_set() {
-    if (type) {
+    if (m_type) {
         error_scope scope;
         gil_scoped_acquire gil;
-        type.release().dec_ref();
-        value.release().dec_ref();
-        trace.release().dec_ref();
+        m_type.release().dec_ref();
+        m_value.release().dec_ref();
+        m_trace.release().dec_ref();
     }
 }
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -324,7 +324,7 @@ public:
     /// Constructs a new exception from the current Python error indicator, if any.  The current
     /// Python error indicator will be cleared.
     error_already_set() : std::runtime_error(detail::error_string()) {
-        PyErr_Fetch(&type.ptr(), &value.ptr(), &trace.ptr());
+        PyErr_Fetch(&m_type.ptr(), &m_value.ptr(), &m_trace.ptr());
     }
 
     error_already_set(const error_already_set &) = default;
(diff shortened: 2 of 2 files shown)


YannickJadoul:
@martinRenou Shouldn't/couldn't this be `const object &get_type() const { return type; }`, as C++-style to avoid refcounting in the copy constructor when it's not necessary to copy the `object` object?

YannickJadoul:
I'm looking into that now, actually. On Travis, it's got something to do with a dead link, it seems. Nothing in this PR, I'd guess :-)

YannickJadoul:
Although ... the current build failure is also partly due to this PR, I think, since I'm getting error messages on "duplicate const".
Probably this `const object& get_type() { return type; } const` instead of `const object& get_type() const { return type; }` typo?

martinRenou:
Oh yeah sorry... I did that too quickly with vim without really checking what I was doing

YannickJadoul:
Great, thanks! :-)
Rebased #1651 onto this, already.

About the failing tests: #1642 is fixing the broken build and solving the Travis part.

zhihaoy:
Unless someone whats to bikeshed it, I think this is good, because I've started to use it (I thought about letting vcpkg to patch it, at least in our team's vcpkg clone).

martinRenou:
Any chance to merge this?

wjakob:
This seems reasonable. However, just a comment regarding the naming: we haven't so far used `get_` as prefix for getters in this library.

wjakob:
actually, scratch that :) -- there are quite a few prefixed `get_`.

martinRenou:
I can change it to `type`, `value` and `trace` if you prefer :)

wjakob:
Hm, I think I would prefer them to be named just `value()`, `trace()`, etc. (you will probably have to change the name of the members then).

There is one slight inconsistency with `handle::get_type` which actually *does* use a `get_` prefix, but elsewhere we generally omit it.

martinRenou:
@wjakob Hope it's fine now :)

martinRenou:
Any chance to get this in? @wjakob. I guess the travis job not passing is unrelated.

apollo13:
I am running with this patch locally, works fine so far

wjakob:
Merged, thanks!

martinRenou:
Thanks :)

PR #1648, opened 2018-12-25
Allow list insertion

ManuelSchneid3r:


wjakob:
This looks like a good addition, but the argument order violates what is used in Python (`insert(index, value)`). Can you update the PR accordingly?

wjakob:
Closed due to inactivity (and superseded by #1888)

PR #1651, opened 2018-12-31
Implementation of Python's 'with' statement as py::with

YannickJadoul:
As discussed with @martinRenou and @bstaletic on Gitter, a `with` statement complying with the [semantics of PEP 343](https://www.python.org/dev/peps/pep-0343/#specification-the-with-statement) might be a useful utility to add to the pybind11 toolbox.

Since C++ will crash if an exception gets thrown while another one is already active, it seems impossible to have a safe RAII implementation of this. But for almost all purposes, this version, taking a function or lambda, should be sufficient.

This PR depends on @martinRenou's PR #1641 adding getters to `py::error_already_set`.

### DIFF
--- a/docs/advanced/pycpp/utilities.rst
+++ b/docs/advanced/pycpp/utilities.rst
@@ -21,6 +21,48 @@ expected in Python:
     auto args = py::make_tuple("unpacked", true);
     py::print("->", *args, "end"_a="<-"); // -> unpacked True <-
 
+Executing code in a Python `with` statement
+===========================================
+
+Where in C++, the scope of local variables and their destructor is used to manage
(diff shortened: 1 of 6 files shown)


wjakob:
Hi Yannick,

this is certainly a neat idea and was clearly a lot of work. However, for me the C++ form of this feature is simply too awkward, and I find the implementation too complex to justify shipping a rather "niche" feature in header files that will be included downstream by all projects using pybind11. So I'll close this for now. If the implementation was substantially simpler, in a separate header file that is not used by default, and didn't require changes to central pybind11 logic, I might consider it.

Best,
Wenzel

PR #1656, opened 2019-01-07
function_record: fix uninit member

ax3l:
fix an uninitialized public member.

found with [coverity](https://scan.coverity.com/dashboard) in a [downstream project](https://github.com/openPMD/openPMD-api).

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -182,7 +182,7 @@ struct function_record {
     bool is_method : 1;
 
     /// Number of arguments (including py::args and/or py::kwargs, if present)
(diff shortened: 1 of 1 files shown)


ax3l:
@zhihaoy good points, looks better now and is more modern.

wjakob:
`nargs` is initialized in `initialize_generic`, it does not need a default initializer.

zhihaoy:
Please re-open this, initializing this is entirely not hurt, because in layout it is between two initialized members, so a sane compiler will just initialized it -- if a compiler wants to not to initialize it the compiler has to do more work!

PR #1659, opened 2019-01-07
Overload Signature: Potential Nullptr Deref

ax3l:
Fix a potential nullptr in error handling. The `it` variable is already checked in the `try` block for non-nullptr and might therefore be `nullptr` in this branch as well.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -763,7 +763,10 @@ protected:
         } else if (!result) {
             std::string msg = "Unable to convert function return value to a "
                               "Python type! The signature was\n\t";
-            msg += it->signature;
+            if (it == nullptr)
+                msg += "not found (nullptr).";
+            else
+                msg += it->signature;
             append_note_if_missing_header_is_suspected(msg);
             PyErr_SetString(PyExc_TypeError, msg.c_str());
             return nullptr;


include/pybind11/pybind11.h
```diff
@@ -763,7 +763,10 @@ class cpp_function : public function {
         } else if (!result) {
             std::string msg = "Unable to convert function return value to a "
                               "Python type! The signature was\n\t";
-            msg += it->signature;
+            if (it == nullptr)
+                msg += "no signature found (nullptr)";
```
zhihaoy: Please change `msg` to say so; the full phrase

> The signature was
> no signature found

reads weird.

ax3l: makes sense, updated!

wjakob:
The catch block can never be reached while `it` is `nullptr`. This is a case where coverity lacks context and is too conservative.

ax3l:
Actually coverity took the following logic: in line 460 with `for (; it != nullptr; it = it->next) {` we check that `it` is not a `nullptr`.

Coverity consequently assumes `it` *can* be a `nullptr` and warns at the line marked in this PR.

wjakob:
During function dispatch, there are 3 basic outcomes:

1. No matching overload can be found
2. An exception is thrown
3. A matching overload is found and executes, but we can't convert the return value.

The case you wanted to patch is related to #3, and it can only happen if #1 and #2 were not taken (and in this case, ``it != nullptr``).

PR #1660, opened 2019-01-07
value_and_holder: uninit members

ax3l:
fix some uninitialized members in `value_and_holder` for some of the constructors.

`value_and_holder` has two constructors that leave the class half-init and has several public methods that operate unchecked on those member vars (deref, etc.)


wjakob:
Fine.

PR #1661, opened 2019-01-07
dict_readonly: member init

ax3l:
fix missing member initialization in pytypes: read-only dict.

`dict_readonly` has a `dereference` method that would return uninit data.


wjakob:
fine..

PR #1664, opened 2019-01-10
pybind11_getbuffer: useless safe nullptr check

ax3l:
Alternative implementation for #1657: if we know that `obj` is never a `nullptr` [1], we should not `nullptr`-check it *after* already de-referencing it five lines earlier.

[1] https://github.com/pybind/pybind11/pull/1657#issuecomment-452090058


wjakob:
LGTM.

PR #1671, opened 2019-01-23
compatibility with pytest 4.0, fix #1670

nim65s:
Cf. https://docs.pytest.org/en/latest/deprecations.html#pytest-namespace

PR #1672, opened 2019-01-23
Silent numpy warnings, fix CI

nim65s:
Hi,

This PR follows #882, requires #1671, and fixes Travis builds.

Appveyor is still erroring because another warning:
> python : DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -185,7 +185,7 @@ def gc_collect():
     gc.collect()
 
 
-def pytest_namespace():
+def pytest_configure():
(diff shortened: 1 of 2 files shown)


bstaletic:
Now why is Appveyor erroring out when pip raises a warning?

wjakob:
Thanks for the fix! I'll see if I can get AppVeyor to pass as well..

PR #1676, opened 2019-01-28
Improve documentation related to inheritance.

iwanders:
I ran into a situation where my C++ signature was `bool(std::vector<int>&)`. Instead of using this function signature I thought it would be more Pythonic to either return a `list` or a `None` from the Python method itself. Figuring out how to do this took longer than I hoped since `get_overload` was not present in the documentation. I thought I'd do something back to this project by improving the documentation:

* Adds section to the reference detailing the trampoline macro's and `get_overload`
* Adds section to advanced classes page describing how to use `get_overload` to make a `bool(int&)` C++ method look at the return type of the associated Python method.

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -80,10 +80,10 @@ helper class that is defined as follows:
         }
     };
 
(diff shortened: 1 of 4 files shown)


wjakob:
Beautiful! Thank you very much for the PR!

PR #1678, opened 2019-01-31
Apply c++ standard flag only to files of CXX language

loriab:
The `-std=c++xx` flag should only be applied to `CXX` files. Normally, applying the `-std` to all files is harmless because the compiler recognizes but issues a warning on the invalid flag. However, in the sad case where a  module to be built contains Fortran files _and_ the std is quite new _and_ the fortran compiler is quite old (e.g., 14 and gfortran 4.8.5), the warning turns to error: `Warning: command line option '-std=c++14' is valid for C++/ObjC++ but not for Fortran` --> `error: unrecognized command line option '-std=c++14'`.

Of the two changes below, the pybind11Config is what worked for my project, but it seemed reasonable to switch the other as well. I haven't built with old cmake versions, but 3.3 is when the docs say cmake learned `COMPILE_LANGUAGE` generator.


loriab:
Thank you for fixing the CI. I believe this PR is now fit for consideration.

wjakob:
FYI: there are two potentially related PRs: #1098 and #1089. Not sure what is the best way to get all of this merged..

loriab:
It looks like #1089 is identical to half (tools file) of this PR. The other half of this PR (config file) is what my project needed. Probably just a difference between the (many) ways to incorporate pb11 into a project.

The other PR has rather more discussion, so if that one incorporates the config file changes from here, I'm glad to close this. Or, this PR has both -- your choice.

With respect to #1098, I've never played with cmake compile options, but it doesn't seem to address what language of files the C++ standard is applied to, so I'd venture the PR is separate.

wjakob:
I decided to merge this one since it is a superset of #1089.

PR #1682, opened 2019-02-01
Disable testing when using BUILD_TESTING

pfultz2:
CMake uses this variable to enable/disable tests in their [CTest module](https://cmake.org/cmake/help/latest/module/CTest.html)(which is why the variable name is highlighted by github as a special value).

More importantly, it allows pybind to be installed directly with [cget](https://cget.readthedocs.io/en/latest/) with `cget install pybind11/pybind` since cget also sets `BUILD_TESTING` to `Off` when not testing a package.

It will still default to `On` when pybind is not a master project, and `Off` when it is not.


wjakob:
Seems reasonable, but can't you do this without introducing a redundant option (which already exists internally in CMake). This will probably not work, but something like:
```
option(PYBIND11_TEST    "Build pybind11 test suite?"     ${BUILD_TESTING} AND ${PYBIND11_MASTER_PROJECT})
```

wjakob:
ping?

pfultz2:
> Seems reasonable, but can't you do this without introducing a redundant option

That option only exists if you include `CTest`, and it will default to `On` always. This will set the default based on whether its a subproject.

This is also similar to the `BUILD_SHARED_LIBS` variable. It is recommended to set it by default to `On` in the project when you want the library to be built as a shared library(but the user can override the variable to get a static library), from [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html):

> This variable is often added to projects as an OPTION so that each user of a project can decide if they want to build the project using shared or static libraries.

henryiii:
I thought I'd fixed issues like this, but my logic for `BUILD_TESTING` was not quite right.

The normal/simplest/best way to do this for a project that uses CTest (which I think we should look at using later) would be:

```cmake
if(PYBIND11_MASTER_PROJECT)
  include(CTest) # defined BUILD_TESTING
  if(BUILD_TESTING)
    add_subdirectory(tests)
  endif()
endif()
```

The downside is that you can't build the tests if this is not the master project, but in general that's not a good idea and probably not really supported anyway.

PR #1683, opened 2019-02-03
PYBIND11_TLS_REPLACE_VALUE should use macro argument value in Python …

jquesnelle:
The `PYBIND11_TLS_REPLACE_VALUE` macro doesn't actually use the `value` argument when using the new Python TLS API (3.7+). Instead, it always uses `tstate` as the argument. This works because the two places `PYBIND11_TLS_REPLACE_VALUE` is used in the code the argument *is* `tstate`.

1. https://github.com/pybind/pybind11/blob/e2b884c33bcde70b2ea562ffa52dd7ebee276d50/include/pybind11/pybind11.h#L1890
2. https://github.com/pybind/pybind11/blob/e2b884c33bcde70b2ea562ffa52dd7ebee276d50/include/pybind11/pybind11.h#L1963

However, for correctness, and for anyone piggybacking on pybind's cross-version TLS support (like me!) it should use the `value` argument


wjakob:
Nice catch!

PR #1494, opened 2018-08-19
Enable dynamic attribute support for Pypy >= 6

inducer:
According to my experiments, the code that's disabled by the conditional on Pypy appears to work just fine on (at least) Pypy 6.0.

PR #1720, opened 2019-03-08
[WIP] Test PyPy 3.6 on travis

lopuhin:
Adapted from https://github.com/pybind/pybind11/pull/1494/ (it would enable pypy support for pytorch).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,12 @@ matrix:
     addons:
       apt:
         packages: [libblas-dev, liblapack-dev, gfortran]
+  # Test a PyPy 3.6 build
+  - os: linux
+    env: PYPY=7.0 PYTHON=3.6 CPP=11 GCC=4.8
+    addons:
+      apt:
+        packages: [libblas-dev, liblapack-dev, gfortran]
   # Build in 32-bit mode and tests against the CMake-installed version
   - sudo: true
     services: docker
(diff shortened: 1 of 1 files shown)


mattip:
It seems PyPy lacks an implementation for `_Py_fopen_obj` (which is part of the Py_LIMITED_API, so I am not sure PyPy will ever implement it) and `PyRun_FileEx` (which is itself a simplified version of `PyRun_FileExFlags`, still lacking on PyPy).

wjakob:
Hi @mattip and @lopuhin,

how about putting an include guard and simply disabling the run-from-file features in `eval.h` (it being a fairly niche feature.)? It could simply throw an exception on that platform.

Best,
Wenzel

bstaletic:
There are only two errors here:

```
error: ‘_Py_fopen_obj’ was not declared in this scope
error: ‘PyRun_FileEx’ was not declared in this scope
```

The first one truly doesn't exist in pypy. The second one exists as `PyRun_File` and should be caught by `#if PY_VERSION_HEX < 0x03000000 && defined(PYPY_VERSION)` but it isn't.

Because of the first error, how about detecting if we're compiling for PYPY and then doing something like:

```
template <eval_mode mode = eval_statements>
object eval_file(str fname, object global = globals(), object local = object()) {
#ifndef pypy_macro
// current implementation
#else 
#error PyPy doesn't support eval_file(), use eval() instead
#endif 
}
```

An alternative is reading the file into a string in `eval_file()` and then invoking `eval()`.

wjakob:
Hi -- any news here? It would be cool to support recent versions of PyPy in the CI.

rwgk:
Hi Konstantin, there was a recent upgrade to use PyPy 7.3 on travis, e.g. here is a random recent result (bottom of page): https://travis-ci.org/github/pybind/pybind11/builds/698654908
Is this PR still relevant?

lopuhin:
@rwgk great, thank you! No it's not relevant any more, any newer PyPy works

PR #1726, opened 2019-03-11
doc: Add "Getting Help" section

### LINKED ISSUE #1724: doc: Make Gitter + StackOverflow links more prominent?
I've seen quite a few issues that aren't really bug reports or feature requests; the issues templates does direct people to use Gitter, but it seems that there are still quite a few issues that get by.

Would it help to make a "Getting Help" section in the RTD docs, perhaps right under "About this project", to point to Gitter / StackOverflow?

(I'm posting this assuming that this is not a huge issue, but perhaps could still be optimized. If this is a complete non-issue, feel free to close this!)

eacousineau:
Resolves #1724 (if it helps)

Modeling this after what we did on Drake: https://drake.mit.edu/getting_help.html

Meta:
~Is there an easy way to search Gitter chatrooms? I've struggled to figure this out, and it seems like a *really* rich source of info if it were easily searchable...~ See response below.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,12 +1,6 @@
-Make sure you've completed the following steps before submitting your issue -- thank you!
-
-1. Check if your question has already been answered in the [FAQ](http://pybind11.readthedocs.io/en/latest/faq.html) section.
-2. Make sure you've read the [documentation](http://pybind11.readthedocs.io/en/latest/). Your issue may be addressed there.
-3. If those resources didn't help and you only have a short question (not a bug report), consider asking in the [Gitter chat room](https://gitter.im/pybind/Lobby).
-4. If you have a genuine bug report or a more complex question which is not answered in the previous items (or not suitable for chat), please fill in the details below.
-5. Include a self-contained and minimal piece of code that reproduces the problem. If that's not possible, try to make the description as clear as possible.
-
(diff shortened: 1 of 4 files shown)


bstaletic:
> Is there an easy way to search Gitter chatrooms? I've struggled to figure this out, and it seems like a really rich source of info if it were easily searchable...

In the upper left corner there are, in order, downwards, Gitter logo, "All conversations" and finally "Search Gitter".

eacousineau:
Ah, sweet, thanks! I had seen that before, but I thought that search all of Gitter, because the ordering of results are first "Rooms & People", then "Chat Messages"; however, searching as you mentioned, the posts do appear to be constrained to `pybind11/Lobby`.
(I do kinda wish that they had a better search interface though...)

wjakob:
Hi Eric,

this is generally a good idea, but I'd like to include as much as possible of this information in the ISSUE_TEMPLATE.md — I want people to be forced to read that rather than having just a single line that is quick to delete.

This project has become very popular and is getting seriously hammered with issue reports (328 and counting!). Part of the reason is that many people fire off tickets for "obvious" questions without reading the manual.

Best,
Wenzel

eacousineau:
Sounds good! If possible, it'd be great to see a StackOverflow link, and maybe a quick tip on searching for Gitter (given that I, uh, struggled with the obvious :P) on the issue template.

I'll go ahead and close this PR, and try that out.

PR #1732, opened 2019-03-18
Make static member functions, added with `def_static`, `staticmethod` descriptor instances

YannickJadoul:
This PR makes sure Sphinx sees and documents static member functions as such (i.e., with an extra "*static*" annotation, cfr. [the Python docs](https://docs.python.org/3/library/stdtypes.html#str.maketrans)).

This also corresponds directly to what `@staticmethod` does in Python and [how Boost.Python does this](https://github.com/boostorg/python/blob/4b01139720128d45e6887b78271baf4eb201e24d/src/object/class.cpp#L709). One caveat, just like in pure Python, is that the `staticmethod` wrapper disappears when accessing the attribute, because of the [descriptor protocol implementation in `staticmethod`](https://docs.python.org/3.7/howto/descriptor.html#static-methods-and-class-methods).

If this is considered useful, we could also still add `@classmethod`?

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1112,7 +1112,7 @@ public:
                 "def_static(...) called with a non-static member function pointer");
         cpp_function cf(std::forward<Func>(f), name(name_), scope(*this),
                         sibling(getattr(*this, name_, none())), extra...);
-        attr(cf.name()) = cf;
+        attr(cf.name()) = staticmethod(cf);
         return *this;
     }
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -742,6 +742,8 @@ inline bool PyEllipsis_Check(PyObject *o) { return o == Py_Ellipsis; }
 
 inline bool PyUnicode_Check_Permissive(PyObject *o) { return PyUnicode_Check(o) || PYBIND11_BYTES_CHECK(o); }
 
+inline bool PyStaticMethod_Check(PyObject *o) { return o->ob_type == &PyStaticMethod_Type; }
+
 class kwargs_proxy : public handle {
 public:
     explicit kwargs_proxy(handle h) : handle(h) { }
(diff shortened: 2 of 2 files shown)


wjakob:
Hi @YannickJadoul,

This generally looks reasonable, but I'm wondering if this is a breaking change? (i.e. one that will require a major version bump of pybind11)
 
AFAIK making something a static method will change the associated calling conventions, correct?

Best,
Wenzel

YannickJadoul:
@wjakob Good point, I hadn't really considered that.

However, calling conventions will not be broken. Even the object returned by `Class.some_staticmethod` will stay the same. The only thing this will change is the type of `SomeClass.__dict__["some_staticmethod"]` (which is apparently what Sphinx is looking at to add the extra _static_ to the API documentation).

This corresponds to the effect `@staticmethod` has in pure Python does. Additionally, since the C-implemented method types in pybind11 are already different from the types of the method you'd get in pure Python, the types of pybind11/CPython extensions methods are already inherently duck-typed, so I think this is safe.

I've just written the following example to compare the effects of the change:

```c++

namespace py = pybind11;

class DeepThought {
public:
	static bool isAnswer(int x) { return x == 42; }
};

PYBIND11_MODULE(example, m) {
	py::class_<DeepThought>(m, "DeepThought")
		.def(py::init<>())
		.def_static("is_answer", &DeepThought::isAnswer)
		.def_static("is_answer_staticmethod", &DeepThought::isAnswer);

	m.attr("DeepThought").attr("is_answer_staticmethod") = PyStaticMethod_New(m.attr("DeepThought").attr("is_answer_staticmethod").ptr());
}
```

```pycon
Python 3.7.1 (default, Oct 22 2018, 11:21:55) 
[GCC 8.2.0] on linux
>>> import example
>>> example.DeepThought.is_answer(41)
False
>>> example.DeepThought.is_answer_staticmethod(42)
True
>>> example.DeepThought().is_answer(41)
False
>>> example.DeepThought().is_answer_staticmethod(42)
True
>>> example.DeepThought.is_answer
<built-in method is_answer of PyCapsule object at 0x7fc6009c02d0>
>>> example.DeepThought.is_answer_staticmethod
<built-in method is_answer_staticmethod of PyCapsule object at 0x7fc6009c02a0>
>>> example.DeepThought.__dict__["is_answer"]
<built-in method is_answer of PyCapsule object at 0x7fc6009c02d0>
>>> example.DeepThought.__dict__["is_answer_staticmethod"]
<staticmethod object at 0x7fc5fed3e1d0>
```

(Also tested in 2.7, and it shows the same results)

wjakob:
Ok, this looks all good to me then! — Thanks!

PR #1734, opened 2019-03-20
Correctly report Python exceptions set but not thrown during import.

anntzer:
Consider the following example:

        PyErr_SetString(PyExc_ImportError, "oops");
    }

Currently, importing this module will result in

    SystemError: initialization of python_example raised unreported exception

because the Python exception is not converted to a C++ one
(py::error_already_set).  Fix that by checking the error flag, so that
one instead gets

    ImportError: oops

Real use cases include numpy's `import_array()` or pycairo's
`import_cairo()`, which can set the exception flag (and return -1) on
failure.  Currently the correct way to handle them with pycairo is

    if (import_foo() < 0) { m.ptr() = nullptr; return; }

and this patch removes the need for `m.ptr() = nullptr;`.

No tests for now for the same reason as in https://github.com/pybind/pybind11/pull/1362 (I *think* any proper unit test would require creating a new module just for that purpose, and I'm not particularly familiar with the testing machinery used by pybind, so guidance would be welcome here. But "it works for me"...).

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -258,7 +258,8 @@ extern "C" {
     PYBIND11_PLUGIN_IMPL(name) {                                               \
         PYBIND11_CHECK_PYTHON_VERSION                                          \
         try {                                                                  \
-            return pybind11_init();                                            \
+            auto m = pybind11_init();                                          \
+            return PyErr_Occurred() ? nullptr : m;                             \
         } PYBIND11_CATCH_INIT_EXCEPTIONS                                       \
     }                                                                          \
     PyObject *pybind11_init()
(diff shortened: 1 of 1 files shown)


anntzer:
Kindly bumping.  To the best of my understanding, the test failures seem spurious.

wjakob:
Hi @anntzer,

the pybind11 way involves turning such errors into exceptions that are propagated and correctly, and setting PyErr_SetString directly within pybind11 code violates that contract. If you call into other libraries with that behavior, you'll have to check for errors and throw py::error_already_set.

Best,
Wenzel

anntzer:
Ah, I missed the possibility to just throw py::error_already_set in that case, thanks for pointing that out.

PR #1743, opened 2019-03-26
Adding fix for copy warning on clang 7 / macOS 10.14.4

henryiii:
This occurs if you have a high level of warnings set. See <https://reviews.llvm.org/D43322?id=134514>.


wjakob:
That's actually a neat warning! Can you rebase this onto the latest master just to have CI run on it?

henryiii:
It’s really handy! I found a bug in my code with this warning (wrong return type). However, the CI runs on a recent clang and errors on warnings, so the fix is already in as part of the CI fix #1744.

henryiii:
We need to try GCC 9 (still in beta, though) - it has a ton of new warnings and reporting features.

Issue #1745, opened 2019-03-27
Add more attribute getters like object_api<>::doc()?


ultrasilicon:
## Issue description
I was looking for a way to set `__version__` property for my module, but I didn't find a direct way of doing this in the docs. I worked around by doing: 
``` c++
PYBIND11_MODULE(MyModule, m)
{
    m.doc() = "My Module";
    m.attr("__version__") = "1.1.0";
    ...
}
```

Will it be better if we can do:
``` c++
PYBIND11_MODULE(MyModule, m)
{
    m.doc() = "My Module";
    m.version() = "1.1.0";
    m.author() = "Sponge Bob";
    ...
}
```

YannickJadoul:
#1746 was closed with an explanation on why this wasn't included into pybind11.

PR #1746, opened 2019-03-27
Added convenient attribute getters

ultrasilicon:
#1745 
Like function `str_attr_accessor object_api<D>::doc()` is a convenient attribute getter for Python property `__doc__`, the corresponding function for `__version__`, `__author__` and `__license__` are added.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -146,8 +146,17 @@ public:
     PYBIND11_DEPRECATED("Use py::str(obj) instead")
     pybind11::str str() const;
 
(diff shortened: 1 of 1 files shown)


eacousineau:
These seem more like module-specific attributes than object-specific attributes. Consider adding them to `py::module`?

wjakob:
Hi @ultrasilicon,

generally a nice idea, but a bit too specific/niche to warrant including in pybind11 IMO. I try to limit the amount of things that go into the library that aren't widely needed, and this functionality is already available by typing a few extra characters.

Best,
Wenzel

PR #1748, opened 2019-03-29
Fix casting of time points with non-system-clock duration with VS

uentity:
Add explicit `time_point_cast` to time point with duration of system
clock. Fixes Visual Studio compile error.

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -128,7 +128,7 @@ public:
         // Lazy initialise the PyDateTime import
         if (!PyDateTimeAPI) { PyDateTime_IMPORT; }
 
-        std::time_t tt = system_clock::to_time_t(src);
+        std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src));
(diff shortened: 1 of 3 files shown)


wjakob:
Hi @uentity,

I'm surprised by this PR because the test suite correctly builds on MSVC and presumably tests this functionality as well. If there is truly an issue, could you provide a testcase that fails with current pybind11?

Thanks,
Wenzel

uentity:
Hi @wjakob,
Done. Without this PR `test_chrono.cpp` won't compile on Visual Studio 2017. Though I tested only in my local environment, I'm pretty sure compile issue will be reproducible anywhere.

uentity:
@wjakob style check is failing not because of my changes:
```
Warning, treated as error:
/home/travis/build/pybind/pybind11/docs/advanced/classes.rst:328:malformed hyperlink target.
```

bstaletic:
@uentity The error has been fixed. You can just rebase the PR.

uentity:
@bstaletic Done.

wjakob:
Thanks!

PR #1753, opened 2019-04-03
Fixing warnings about conversions in GCC 7+

henryiii:
Another minor warning fix (that still is very nice to have, given this can spew lots of text to the screen). Would be nice to have backported to the 2.2.x branch unless 2.3 is imminent.


wjakob:
I don't see how this could possibly break the build, so I'll go ahead and merge.

PR #1757, opened 2019-04-06
Fix for issue in latest conda

henryiii:
Attempt to fix AppVeyor issue. There's a bug in Conda 4.6.11; this avoids the bug by allowing the transaction to happen in the preinstalled conda instead (4.5.4). Note that conda still updates conda.


henryiii:
@wjakob I think this will fix the windows build!

wjakob:
Ok, sounds reasonable to me.

PR #1760, opened 2019-04-09
doc: Add note about casting from `None` to `T*`

sizmailov:
I get confused with default cast rules for types:

**Use case:**
I want to pass optional argument of type without natural sentinel value. Type `T*` has additional sentinel value compared to `T`. So minimal function interface would look like `void func(T* t)`.

**Problem:** 
`None` passed to T* argument treated differently for different `T`, namely:

`nullptr` passed to `double*` argument results to `TypeError`
`nullptr` passed to `Foo*` is OK
`nullptr` passed to `py::object*` is OK
`nullptr` passed to `std::vector<Foo>*` argument results to `TypeError` with `#include <pybind11/stl.h>`
`nullptr` passed to `std::vector<Foo>*` argument is OK with `#include <pybind11/stl_bind.h>` and `py::bind_vector<...>`
nullptr passed to const char* argument is OK

**Solution:**
Explicitly mention this behavior in documentation.

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -470,10 +470,11 @@ The default behaviour when the tag is unspecified is to allow ``None``.
 .. note::
 
     Even when ``.none(true)`` is specified for an argument, ``None`` will be converted to a
-    ``nullptr`` *only* for custom types. Pointers to built-in types (``double *``, ``int *``, ...) and
-    STL types (``std::vector<T> *``, ...; if ``pybind11/stl.h`` is included) are copied when converted
-    to C++ (see :doc:`/advanced/cast/overview`) and will not allow ``None`` as argument.
-    To pass optional argument of these copied types consider using ``std::optional<T>``
+    ``nullptr`` *only* for custom and :ref:`opaque <opaque>` types. Pointers to built-in types
+    (``double *``, ``int *``, ...) and STL types (``std::vector<T> *``, ...; if ``pybind11/stl.h``
+    is included) are copied when converted to C++ (see :doc:`/advanced/cast/overview`) and will
+    not allow ``None`` as argument.  To pass optional argument of these copied types consider
+    using ``std::optional<T>``
 
 Overload resolution order
 =========================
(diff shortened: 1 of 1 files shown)


YannickJadoul:
What about something like:

"Even when ``.none(true)`` is specified for an argument, ``None`` will only converted to a ``nullptr`` for custom types. Pointers to built-in types (``double *``, ``int *``, ...) and STL types (``std::vector<T> *``, ...; if ``pybind11/stl.h`` is included) are copied when converted to C++ (see :doc:`/advanced/cast/overview`) and will not allow `None` as argument. To pass optional argument of these copied types consider using ``std::optional<T>``, or expose STL classes as custom classes with the help of ``pybind11/stl_bind.h``."

I'd even suggest to maybe take out the "expose STL classes ... pybind11/std_bind"?

sizmailov:
Thanks for comment! I slightly changed first phrase and removed `stl_bind` part. I still feel uncomfortable with this kind of counter-intuitive diversity, but at least now it's explicitly stated.

YannickJadoul:
Great, as far as I'm concerned!

Let's see what other comments this PR gets. I agree it's kind of a weird compromise where pointers to these types are allowed, but changing the values will not change them and `nullptr` are not allowed. Then again, there's probably some gooed reasons for these decisions that were made :-)

sizmailov:
@YannickJadoul I get an idea, please correct me if I'm wrong. Documentation defines *opaque* type as a type which might be converted between Python and C++ without creating a copy. All user-defined types are opaque. 
Can you confirm that problem exists for non-opaque types only? If so the note should be reworded using *opaque* terminology to be more accurate.

YannickJadoul:
@sizmailov I think you're right indeed, but I don't know enough about "opaque types" to confirm this (and since that term is mostly used in [this section](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#making-opaque-types)), I'd maybe add it as "... custom and opaque types ..."  ?

wjakob:
This looks great, thanks!

PR #1767, opened 2019-04-12
Added __contains__ to stl bindings for maps

flippmoke:
Currently there is no `__contains__` method defined for STL container bindings. If you do not define a `__contains__` it will by default fall back to `__iter__` for checking if the key exists as described [here](https://docs.python.org/3/reference/expressions.html#membership-test-operations). For many C++ map containers types this can be very expensive.


wjakob:
LGTM!

PR #1769, opened 2019-04-19
Mark exceptions with export to fix exceptions in shared libraries

yeswalrus:
In a setup with two shared libraries, A and B, if A throws an exception, B should be able to catch it. To do this, the exception must have external linkage, otherwise A and B will both have their own separate copies of the type.  See https://gcc.gnu.org/wiki/Visibility for more information, but the short version is that all exceptions defined in libraries should be declared with `__attribute__ ((visibility("default")))`

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -654,7 +654,7 @@ using expand_side_effects = bool[];
 NAMESPACE_END(detail)
 
 /// C++ bindings of builtin Python exceptions
-class builtin_exception : public std::runtime_error {
+class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
 public:
     using std::runtime_error::runtime_error;
     /// Set the error using the Python C API
@@ -662,7 +662,7 @@ public:
 };
 
 #define PYBIND11_RUNTIME_EXCEPTION(name, type) \
-    class name : public builtin_exception { public: \
+    class PYBIND11_EXPORT name : public builtin_exception { public: \
         using builtin_exception::builtin_exception; \
         name() : name("") { } \
         void set_error() const override { PyErr_SetString(type, what()); } \
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -654,12 +654,20 @@ using expand_side_effects = bool[];
 NAMESPACE_END(detail)
 
 /// C++ bindings of builtin Python exceptions
+//
+#if defined(_MSC_VER)
+#  pragma warning(push)
+#  pragma warning(disable: 4275) // warning C4275: An exported class was derived from a class that wasn't exported
+#endif
 class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
 public:
     using std::runtime_error::runtime_error;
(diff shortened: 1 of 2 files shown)


eacousineau:
Thanks!
FTR Motivating Gitter discussion: https://gitter.im/pybind/Lobby?at=5cb937183b6cb0686a177d53

YannickJadoul:
Nice!

Something's wrong on Windows, though :-(

> `non dll-interface class 'std::runtime_error' used as base for dll-interface class 'pybind11::builtin_exception'`

Any idea what this problem could be and how it could be solved?

yeswalrus:
https://stackoverflow.com/questions/24511376/how-to-dllexport-a-class-derived-from-stdruntime-error
Looks like it's noise and can safely be silenced.

eacousineau:
I'm not sure if you'll be able to have this land by hoisting the visibility of `py::object` and `py::handle`. I believe those are intentionally sealed to minimize binary size (if LTO works, I guess?) and prevent version clashes.

Can I ask why you hoisted the visibility on those types?

yeswalrus:
Because compilers will throw warnings if I don't, specifically [MSVC C4251](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251?view=vs-2019).

the error_already_set exception contains py::object instances, which derives from py::handle.
It *might* be the case that we can simply silence the warning, but as I understand it, that might be dangerous. I don't believe that this will impact binary size, but that's worth double checking

yeswalrus:
master sozie-pybind11_tests.cpython-36m-x86_64-linux-gnu.so: `1975048`
branch sozie-pybind11_tests.cpython-36m-x86_64_linux-gnu.so:`1979240`

So it looks like it does increase binary size by a small amount. However, I think it' sstill more correct to do this - any time an object passes between shared libraries, it ought to be marked as exported, and these explicitly do.

eacousineau:
Aye, that makes sense.

Out of my dogma-fueled paranoia on type privacy, I may suggest reworking `error_already_set` to not use `pybind11`s Python API, and just use the C API directly (e.g. `PyObject*` to store, `Py_XINCREF` on `T()` and `Py_XDECREF` on `~T()`), just to minimize how much is neccessary to export the error to preserve RTTI.

However, I'd also suggest waiting until @wjakob weighs in to actually pull the trigger on making that change.

wjakob:
This looks generally good to me. A few questions:

1. do `pyobject_tag`, `object` and `handle` really need PYBIND11_EXPORT? Does ist still work if this is removed? The goal of this PR is to ensure that the `std::type_info` of an exception class is exported so that inter-module throws are caught correctly, and I don't see how this is related. If this is just to silence a warning, can you `#pragma disable` it instead?

2. Let's leave `error_already_set` as is, please :).

wjakob:
ping?

wjakob:
@yeswalrus, are you still planning to submit this PR? If so, could you respond to my feedback above?

Thanks,
Wenzel

Skylion007:
This was fixed in a later PR: #2999

PR #1772, opened 2019-04-25
Allow std::string as `attr` method argument

martinRenou:


bstaletic:
This would require constructing a `std::string` temporary when passing a C style string literal and thus hurt performance in this case. Is there any reason for this PR apart from avoiding `std::string::c_str()` in user code?

eacousineau:
> This would require [...]

Not to mention that other parts of the code base would require similar updates, like `.def()`, [`process_attribute` stuff](https://github.com/pybind/pybind11/blob/35045eeef8969b7b446c64b192502ac1cbf7c451/include/pybind11/attr.h#L319-L324), etc.

I (believe that I) agree with the implicit pain point of this PR, that it's a wee bit annoying to add `.c_str()` to everything, but the API should generally be consistent, and `const char*` is a usable lowest common denominator - esp. when it comes to specialization / SFINAE for `process_attribute` on `Extra...`.

sizmailov:
pybind stores raw pointers (`char *`) in number of places. The live time of pointed array must be longer than pybind object with references. Life time of different pybind objects is different and ranges from single statement to (roughly) whole module lifetime. Without storing a copy of passed `std::string` one can't guarantee that stored pointer is still valid. On the other hand char literals have unlimited life time and 100% safe. 

New *duplicate* `const std::string&` interface will require careful analysis to avoid performance hits and hard-to-find bugs. 

Next bug example is synthetic, but it's just a demonstration. 

<details>

<summary>

`const std::string& str` might be bounded to temporary string. `pybind::detail::accessor` stores raw pointer to `str.c_str()` which will be invalidated at temporary string destruction. Further access to stored pointer would be segmentation fault.

</summary>

*Bug example:*
```C++
std::string key_prefix = "prefix_";
auto a = foo.attr(key_prefix+"suffix"); // after semicolon raw pointer becomes invalid
...
a = 3.14; // access violation 
```

*No bug example 1:*
```C++
std::string key_prefix = "prefix_";
py::object a = foo.attr(key_prefix+"suffix"); // ok, `pybind::detail::accessor` casted to object and destroyed
...
a = 3.14; // ok
```


*No bug example 2:*
```C++
std::string key_prefix = "prefix_";
foo.attr(key_prefix+"suffix") = 3.14; // ok, temporary destroyed after whole statement evaluation 
```

</details>

martinRenou:
I guess that, instead of replacing `.attr(const char*)` by `.attr(const std::string&)`, we can add an overload that accepts an `std::string`. And we could do the same for other methods/functions in the public API that take `const char*` as argument.
Concerning the lifetime issue, I guess we would need to make a copy of it for safety, but that would slow down a bit the code. 

Anyway, we can close this one if that's not really useful/safe to merge. We could open an issue about it in order to keep the idea around.

wjakob:
I tend to agree with the other opinions here this calls for a more systematic change (at the same time I am somewhat concerned about potentially duplicating an large number of functions). `std::string_view` is the perfect solution to this problem, but sadly it only exists in C++17. One potential option would be a custom pybind11-local string_view replacement (which is certainly realizable in older C++ standards like C++11).

PR #1784, opened 2019-05-08
Use FindPython3.cmake rather than the deprecated FindPythonInterp.cmake

mfdeakin:
FindPythonInterp is deprecated and doesn't always find the Python executable (most notably the default installation in Fedora 30), so pybind11 should use FindPython3 or FindPython (if supporting Python2 is still desired; FindPython is version agnostic).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 # All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.12.4)
 
 if (POLICY CMP0048)
   # cmake warns if loaded from a min-3.0-required parent dir, so silence the warning:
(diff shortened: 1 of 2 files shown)


mfdeakin:
This is likely premature - FindPythonInterp was only deprecated last July, and the replacement implementations don't support PyPy yet

eacousineau:
Not completely sure, but this *might* be covered by @jamiesnape's PR: https://github.com/pybind/pybind11/pull/1373

wjakob:
I'll close this for now, potentially to be revisited in the future.

spflueger:
Hi,
currently we are experiencing problems with cmake 3.16, pybind11 2.4.3 and a python 3.8 venv. The problem is related to the fact that pybind11's CMake code cannot find the venv python. Going to the master branch fixes the problem (maybe because of PR #1974).
I think switching to `FindPython` would definitely make things easier with newer cmake and python versions. (cmake 3.10, pybind11 2.4.3, and python 3.8 venv works fine oddly enough)

mfdeakin:
I've updated my branch to default to using FindPython, but falling back to FindPythonInterp when FindPython fails (such as with PyPy). I think that's sufficient to address my own concerns

henryiii:
This was fixed in #2370 , and CMake 3.18.2+ supports PyPy. :) (best also with PyPy 7.3.2+, as well)

PR #1788, opened 2019-05-16
Minor tweaks to tools/mkdoc.py

drmoose:
This PR contains an assortment of minor changes to `tools/mkdoc.py` intended to make it easier to integrate into automated build workflows:

* Adds compatibility with Clang 8.x
* Accepts the `-o` option, so the output can be written to a file instead of stderr
* Detects and loads the clang system include dir (e.g. `/usr/lib/clang/8.0.0/include`) on Linux
* Tries to get as much of the functionality as possible out of `if __name__ == '__main__'` and into functions that can be called by other python scripts.

Because this is a grab bag of minor tweaks, I've tried to organize the commits so that it's easy to pick and choose -- if there's something I've done here that you don't like, I'm happy to rebase that commit out of the branch.

### DIFF
--- a/tools/mkdoc.py
+++ b/tools/mkdoc.py
@@ -342,8 +342,17 @@ if __name__ == '__main__':
             break
     try:
         if out_path:
-            with open(out_path, 'w') as out_file:
-                mkdoc(args, out_file)
+            try:
+                with open(out_path, 'w') as out_file:
+                    mkdoc(args, out_file)
+            except:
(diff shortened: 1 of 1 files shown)


wjakob:
This looks really great! I'm happy with all of these changes.

wjakob:
(The reorganization into small commits was nice, so I rebased your commits rather than squashing them)

PR #1795, opened 2019-05-31
example not working

ice0:


wjakob:
A few lines above, it says:

```
For brevity, all code examples assume that the following two lines are present:
.. code-block:: cpp
    namespace py = pybind11;
``

**Comment by ice0** [comment]
Yes, but your example, which I am changing, does not follow your own guide.
Code from guide:
```

int add(int i, int j) {
    return i + j;
}

PYBIND11_MODULE(example, m) {

    m.def("add", &add, "A function which adds two numbers");
}
```

Comment from guide:
> With a simple modification code, it is possible to inform Python about the names of the arguments (“i”  and “j” in this case).
```
m.def("add", &add, "A function which adds two numbers", py::arg("i"), py::arg("j"));
```
No it doesn't work.
If you disagree, please show how to make the code above to work using this modification.

**Comment by ax3l** [comment]
I checked this and the example above actually shows a "here is a complete `.cpp` file". So although the first hint is a bit separate it makes sense how it currently looks like. Alternatively, one could move the general comment a bit lower.

**Comment by ice0** [comment]
We can also change the code like this:
```
m.def("add", &add, "A function which adds two numbers", 
      pybind11::arg("i"), pybind11::arg("j"));
```
or just fix the first example.

ax3l:
Yes, yet the convention is used throughout all following examples but the first `.cpp` file, which does not access the namespace :)

PR #1801, opened 2019-06-06
Added note in documentation regarding make install

nstelzen:
Hello, 
For my application I was not able to use the typical cmake `add_subfolder` approach, and needed to use a external location for pybind11. 
Perhaps it is very obvious to some, but in order for `find_package` to work correctly one needs to run a `make install`. I spent an embarrassingly long amount of time figuring this out, so I added a little note in the documentation to help other noobs like me.

In case you are curious, my application uses a Docker development environment which contains all of the external libraries and dependencies. The source code folder is mounted as a volume in docker, and the compilation is done in the container. An external library on the container can not be a subfolder of the mounted volume, which is why the `add_subfolder` approach doesn't work in this case.

Thanks a lot for a great project, I was very happy to find it!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "restructuredtext.preview.sphinx.disabled": true,
+    "restructuredtext.linter.disabled": true
+}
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

PR #1833, opened 2019-07-09
a workaround to avoid a c++14 feature

ToruNiina:
Hi,

Thank you for the amazing work. I'm one of the users of this great library.

Today I found that the latest version of this library includes a c++14 feature, lambda capture initializer. You can see it by the following minimum code.

```cpp
// sample.cpp

PYBIND11_MODULE(example, /*m*/) {
}
```

and 

```console
$ g++-9 -std=c++11 -c -Wall -Wpedantic -Wextra sample.cpp
In file included from sample.cpp:2:
./include/pybind11/functional.h: In member function 'bool pybind11::detail::type_caster<std::function<_Res(_ArgTypes ...)> >::load(pybind11::handle, bool)':
./include/pybind11/functional.h:68:18: warning: lambda capture initializers only available with '-std=c++14' or '-std=gnu++14'
   68 |         value = [hfunc = func_handle(std::move(func))](Args... args) -> Return {
      |                  ^~~~~
```

As the name suggests, pybind11 requires only C++11 (at least now). Although most of the existing compilers can compile this c++14 extension, I think it is good to have a workaround to make it c++11 conformant.

So here I added a temporary function-object that behaves equivalently to the lambda expression.

### DIFF
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -65,12 +65,19 @@ public:
             }
         };
 
-        value = [hfunc = func_handle(std::move(func))](Args... args) -> Return {
-            gil_scoped_acquire acq;
-            object retval(hfunc.f(std::forward<Args>(args)...));
-            /* Visual studio 2015 parser issue: need parentheses around this expression */
-            return (retval.template cast<Return>());
+        // to emulate 'move initialization capture' in C++11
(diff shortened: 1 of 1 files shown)


wjakob:
Looks great, thank you!

PR #1836, opened 2019-07-11
Make sure `detail::get_internals` acquires the GIL before making Python calls

### LINKED ISSUE #1364: gil_scoped_acquire can make unsafe CPython calls without holding the GIL
## Issue description

The `gil_scoped_acquire` guard can in some cases require the GIL, because it calls detail::get_internals(), which calls into the Python CAPI if` get_internals` hasn't already been called with the GIL held in the current module.

Reproduced with Pybind 2.2.2, CPython 2.7.13; issue still appears in pybind11 master.

## Reproducible example code

Contrived example:

```
// noddy.cpp
#include <Python.h>

PyMODINIT_FUNC
initnoddy(void) 
{
    Py_BEGIN_ALLOW_THREADS  // release the GIL
    {
        pybind11::gil_scoped_acquire guard;  // <-- first pybind call in this module
        // Bang!
    }

    Py_END_ALLOW_THREADS  // re-acquire the GIL
}
```

Edit: Original report used the incorrect type for the `guard` variable; have fixed.

saran-t:
This is only necessary if `get_internals` is called for the first time in a given module when the running thread is in a GIL-released state.

Fixes #1364

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -171,6 +171,14 @@ PYBIND11_NOINLINE inline internals &get_internals() {
     if (internals_pp && *internals_pp)
         return **internals_pp;
 
+    // Ensure that the GIL is held since we will need to make Python calls.
+    // Cannot use py::gil_scoped_acquire here since that constructor calls get_internals.
+    struct gil_scoped_acquire {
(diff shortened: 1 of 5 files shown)


wjakob:
closed following #1364.

Issue #1274, opened 2018-02-10
Structured dtype field ordering now handled differently in NumPy 1.14.0


aldanor:
This works on NumPy 1.13:

```cpp
struct Foo { int x; int y; };
PYBIND11_NUMPY_DTYPE(Foo, y, x);
```

but on 1.14 it now throws:

```python
ImportError: NumPy: invalid buffer descriptor!
```

I guess the reason is this:

> Second, assignment between structured arrays will now occur "by position"
instead of "by field name". The Nth field of the destination will be set to the
Nth field of the source regardless of field name, unlike in numpy versions 1.6
to 1.13 in which fields in the destination array were set to the
identically-named field in the source array or to 0 if the source did not have
a field.

> Correspondingly, the order of fields in a structured dtypes now matters when
computing dtype equality.

```python
# 1.13
>>> d1 = np.dtype({'names': ['x', 'y'], 'formats': ['i4', 'i4'], 'offsets': [0, 8]})
>>> d2 = np.dtype({'names': ['y', 'x'], 'formats': ['i4', 'i4'], 'offsets': [8, 0]})
>>> d1 == d2
True
```

```python
# 1.14
>>> d1 = np.dtype({'names': ['x', 'y'], 'formats': ['i4', 'i4'], 'offsets': [0, 8]})
>>> d2 = np.dtype({'names': ['y', 'x'], 'formats': ['i4', 'i4'], 'offsets': [8, 0]})
>>> d1 == d2
False
```

wjakob:
Closed by #1837.

PR #1837, opened 2019-07-14
Dtype field ordering for NumPy 1.14

socec:
As reported in #1274, the fields are ordered "by position" instead of "by field name" as of NumPy 1.14. This PR sorts the fields by position before registering the dtype.

Perhaps it's worth debating if this goes against what the user wanted. An alternative approach would be to throw a more meaningful exception for Numpy >= 1.14, since we know the expected field order (e.g. "NumPy: Invalid field order while registering dtype, expecting: ..."). This would force the user to correctly order the fields without pybind doing it in the background. But the user would also have to be aware of that change in NumPy 1.14. IMHO the user mostly wants to just register the dtype and the proposed solution will do it across current versions.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1066,8 +1066,14 @@ inline PYBIND11_NOINLINE void register_structured_dtype(
     if (numpy_internals.get_type_info(tinfo, false))
         pybind11_fail("NumPy: dtype is already registered");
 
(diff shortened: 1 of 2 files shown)


aldanor:
(Closes #1274 if merged)

wjakob:
This looks good to me, thanks for the fix!

PR #1842, opened 2019-07-15
Add support for __await__, __aiter__, and __anext__ protocols

### LINKED ISSUE #1841: __await__ support for py::class_
I'd like to define a `py::class_` that supports Python 3 await.  For Python-defined classes, it is sufficient to define an `__await__` method that returns a suitable iterable type.  However, for Python C API-defined types, it appears to be necessary to set the `tp_as_async` member of the type.

Can you suggest how this can be done, or if not possible without modifying the pybind11 source code, suggest a recommended approach to take in implementing those modifications?

jbms:
Fixes #1841


wjakob:
This looks good to me, assuming that CI passes.

jbms:
The initial version did not pass, but I fixed the exclusion logic for Python < 3.5 and now it passes.

Issue #1844, opened 2019-07-17
Time Conversions - Python datetime.date to C++ system_clock::time_point


phil-zxx:
As explained here https://github.com/pybind/pybind11/blob/master/docs/advanced/cast/chrono.rst, there is a conversion option of `datetime.datetime → std::chrono::system_clock::time_point` from `Python` to `C++`.

However, I am wondering whether it is possible to also do the conversion for the simple `date` object, i.e. convert `datetime.date → std::chrono::system_clock::time_point` from `Python` to `C++`?

phil-zxx:
Seems like we just have to add an exra `else if` block in `include\pybind11\chrono.h` as follows (since `PyDate_Check` conveniently exists already):

```cpp
if (PyDateTime_Check(src.ptr())) {
    /* ... */
}
else if (PyDate_Check(src.ptr())) {
    std::tm cal;
    cal.tm_sec   = 0;
    cal.tm_min   = 0;
    cal.tm_hour  = 0;
    cal.tm_mday  = PyDateTime_GET_DAY(src.ptr());
    cal.tm_mon   = PyDateTime_GET_MONTH(src.ptr()) - 1;
    cal.tm_year  = PyDateTime_GET_YEAR(src.ptr()) - 1900;
    cal.tm_isdst = -1;

    value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
    return true;
}
```

wjakob:
feel free to create a PR for this (along with test cases, please.).

phil-zxx:
Where are test cases located? That is, where would I need to add them?

phil-zxx:
Submitted

YannickJadoul:
#1848 was merged

PR #1848, opened 2019-07-18
Added ability to convert from datetime.date to system_clock::time_point

phil-zxx:
Added ability to convert from Python datetime.date to C++ system_clock::time_point

As raised here: https://github.com/pybind/pybind11/issues/1844

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -106,8 +106,11 @@ public:
         if (!PyDateTimeAPI) { PyDateTime_IMPORT; }
 
         if (!src) return false;
+
+        std::tm cal;
+        microseconds msecs;
+
         if (PyDateTime_Check(src.ptr())) {
-            std::tm cal;
             cal.tm_sec   = PyDateTime_DATE_GET_SECOND(src.ptr());
             cal.tm_min   = PyDateTime_DATE_GET_MINUTE(src.ptr());
             cal.tm_hour  = PyDateTime_DATE_GET_HOUR(src.ptr());
@@ -115,11 +118,8 @@ public:
             cal.tm_mon   = PyDateTime_GET_MONTH(src.ptr()) - 1;
             cal.tm_year  = PyDateTime_GET_YEAR(src.ptr()) - 1900;
             cal.tm_isdst = -1;
-
-            value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
-            return true;
+            msecs        = microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
         } else if (PyDate_Check(src.ptr())) {
-            std::tm cal;
             cal.tm_sec   = 0;
             cal.tm_min   = 0;
             cal.tm_hour  = 0;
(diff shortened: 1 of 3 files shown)


include/pybind11/chrono.h
```diff
@@ -119,6 +119,19 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
...
+            cal.tm_min   = 0;
+            cal.tm_hour  = 0;
+            cal.tm_mday  = PyDateTime_GET_DAY(src.ptr());
+            cal.tm_mon   = PyDateTime_GET_MONTH(src.ptr()) - 1;
+            cal.tm_year  = PyDateTime_GET_YEAR(src.ptr()) - 1900;
+            cal.tm_isdst = -1;
+
+            value = system_clock::from_time_t(std::mktime(&cal));
+            return true;
+        }
```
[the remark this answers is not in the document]
phil-zxx: Would `PyDateTime_TIME_GET_HOUR`, etc. even work? Given the input is a `datetime.date` (and thus does not have any time information).

include/pybind11/chrono.h
```diff
@@ -119,6 +119,19 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
             value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
             return true;
         }
+        else if (PyDate_Check(src.ptr())) {
```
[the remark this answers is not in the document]
wjakob: yep.

include/pybind11/chrono.h
```diff
@@ -118,6 +118,30 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
 
             value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
             return true;
+        } else if (PyDate_Check(src.ptr())) {
+            std::tm cal;
```
wjakob: Could you move the std::tm object to the parent scope?

include/pybind11/chrono.h
```diff
@@ -118,6 +118,30 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
...
+            cal.tm_sec   = PyDateTime_TIME_GET_SECOND(src.ptr());
+            cal.tm_min   = PyDateTime_TIME_GET_MINUTE(src.ptr());
+            cal.tm_hour  = PyDateTime_TIME_GET_HOUR(src.ptr());
+            cal.tm_mday  = 1;   // This date (day, month, year) = (1, 0, 70)
+            cal.tm_mon   = 0;   // represents 1-Jan-1970, which is the first
+            cal.tm_year  = 70;  // earliest available date for Python's datetime
+            cal.tm_isdst = -1;
+
+            value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_TIME_GET_MICROSECOND(src.ptr()));
```
wjakob: Could you move this to the parent scope? (along with the other 2 cases where `value` is assigned.)

I suggest that all 3 cases write a `microsecond` variable that is simply zero in the `date`-only case.

phil-zxx: There is `PyDateTime_DATE_GET_MICROSECOND` and `PyDateTime_TIME_GET_MICROSECOND`, so I think we have to keep this within each `if` block?

include/pybind11/chrono.h
phil-zxx: Move all `std::tm cal;` to the line before `if (PyDateTime_Check(src.ptr())) {`?

include/pybind11/chrono.h
wjakob: yep, but these can assign to a variable that is kept in the outer scope. The `from_time_t` call is then shared by all 3 branches.

PR #1852, opened 2019-07-21
Expose BufferError among other pybind11 exceptions

mosra:
A very minor thing, exposing python's [BufferError](https://docs.python.org/3/library/exceptions.html#BufferError) among other exception types. Useful when dealing with `py::buffer_protocol{}` interfaces. Not sure if any tests need to be added for this, I guess not :)

Thank you!


wjakob:
This looks fine. For posterity, can you give an example of a potential use case that involves throwing a hypothetical py::buffer_error from C++?

mosra:
Is a link to another repo fine? This is one use case: https://github.com/mosra/magnum-bindings/blob/1ec72697322715ce1d79cdd2eb22f11c1a6fc0af/src/python/corrade/containers.cpp#L121-L122

```cpp
 if(buffer.strides && buffer.strides[0] != buffer.itemsize)
    throw py::buffer_error{Utility::formatString("expected stride of {} but got {}", 
        buffer.itemsize, buffer.strides[0])};
```

Basically, any error handling in `py::init()` that takes a `py::buffer`, or in in a `def_buffer()` would be throwing these.

wjakob:
Ok, sounds good :)

PR #1860, opened 2019-07-26
numpy: fix refcount leak to dtype singleton

### LINKED ISSUE #1858: pybind11::array argument leaks reference count of dtype singleton
## Issue description

Calling a function with a `const pybind11::array&` argument appears to leak refcounts to some Python object. However, it's not the object passed in nor there is a memory leak, so possibly the refcount of some global singleton is not decremented properly.

<s>Numpy itself has a known refcount leak (https://github.com/numpy/numpy/issues/8503), but only with dtype=object arrays --- which are not involved in the case below. Unless, pybind11 uses numpy object arrays internally?</s>
The leaking refcount seems to be that of a dtype singleton.

## Reproducible example code

Python debug build is needed for `sys.gettotalrefcount()`, but you can see the dtype leak also on normal Python:

```shell
set -e
PYTHON=python3.7-dbg

cat <<__EOF__ > refleak.cxx
#include <pybind11/numpy.h>

/* calling this leaks sys.gettotalrefcount() */
void doit(const pybind11::array& a)
{
    return;
}

PYBIND11_MODULE(refleak, m) {
    m.def("doit", doit);
}
__EOF__

cat <<__EOF__ > run.py
import refleak
import numpy as np
import sys

if not hasattr(sys, 'gettotalrefcount'):
    sys.gettotalrefcount = lambda: float('nan')

x = np.array([1])
dtype = np.dtype(np.float_)

before_dtype = sys.getrefcount(dtype)
before = sys.gettotalrefcount()
for j in range(10000):
    refleak.doit(x)
after = sys.gettotalrefcount()
after_dtype = sys.getrefcount(dtype)

import pybind11, numpy
print("numpy:", numpy.__version__)
print("pybind11:", pybind11.__version__)
print("leaked (dtype):", after_dtype - before_dtype, "refs")
print("leaked (total):", after - before, "refs")
print("OK" if after < before + 5000 else "FAIL")
__EOF__

cat <<__EOF__ > setup.py
from setuptools import Extension, setup
import numpy, pybind11
include_dirs = [numpy.get_include(), pybind11.get_include(True), pybind11.get_include(False)]
ext_modules = [Extension("refleak", ["refleak.cxx"], include_dirs=include_dirs)]
setup(name='foo', ext_modules=ext_modules)
__EOF__

$PYTHON setup.py build_ext -i -g
$PYTHON run.py
```
gives here (same for pybind11 2.3.0 and 4a3464fd88ae8666; numpy 1.16.4 or current master)
```
numpy: 1.16.4
pybind11: 2.3.0
leaked (dtype): 10000 refs
leaked (total): 10004 refs
FAIL
```

pv:
PyArray_DescrFromType returns a new reference, not borrowed one,
so handle it accordingly.

Closes #1858

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1044,7 +1044,7 @@ public:
 
     static pybind11::dtype dtype() {
         if (auto ptr = npy_api::get().PyArray_DescrFromType_(value))
-            return reinterpret_borrow<pybind11::dtype>(ptr);
+            return reinterpret_steal<pybind11::dtype>(ptr);
         pybind11_fail("Unsupported buffer format!");
     }
 };
--- a/tests/test_numpy_array.py
+++ b/tests/test_numpy_array.py
@@ -434,3 +434,14 @@ def test_array_create_and_resize(msg):
 def test_index_using_ellipsis():
     a = m.index_using_ellipsis(np.zeros((5, 6, 7)))
     assert a.shape == (6,)
+
+
+@pytest.unsupported_on_pypy
+def test_dtype_refcount_leak():
+    from sys import getrefcount
+    dtype = np.dtype(np.float_)
+    a = np.array([1], dtype=dtype)
+    before = getrefcount(dtype)
+    m.ndim(a)
+    after = getrefcount(dtype)
+    assert after == before


wjakob:
Great, thank you!

PR #1872, opened 2019-08-06
Fixed clang-cindex-python3 submodule

### LINKED ISSUE #1831: tools/clang using relative submodule breaks non-github superprojects
https://github.com/pybind/pybind11/blob/502ffe50a9f22c04637bbc8ec0019488458ba948/.gitmodules#L3

Using a relative path here breaks cases where pybind11 is itself used as a submodule in another repo _that is not hosted on github_.

The issue is raised when doing a recursive submodule update on the super project, as relative paths are resolved based on the superprojects URL (and there is no way that I know of to exclude a single submodule from the recursive update, since command line options override settings from .gitmodules).

Since you use this as an internal tool (#807), would you consider changing back to an absolute url, such as `git@github.com:wjakob/clang-cindex-python3.git`, or perhaps removing the submodule?

jean-airoldie:
Closes #1831

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tools/clang"]
 	path = tools/clang
-	url = ../../wjakob/clang-cindex-python3
+	url = https://github.com/wjakob/clang-cindex-python3


wjakob:
This will re-break #1563. Can you please work out with @henryiii if there is some better way to make everyone happy? Note that removing this submodule is not an option since a number of people/projects use the header file documentation extractor at this point.

henryiii:
Both situations break things; since this is a github hosted project, correctly handling github hosted projects might take priority, and changing this will break CI. You can always clone this onto your other platform and make this one change; since the .gitmodules does not change often, it would be fairly easy to keep up to date.

However, there might be a way around this if we take the proposed solution; using `git config --global url."git@github.com:".insteadOf "https://github.com/"` - that still has to be injected into CI, and will be a slightly ugly change, but *might* be an acceptable workaround.

henryiii:
To me it sounds almost like a git bug; a relative submodule in a submodule should be relative to *that* submodule's URL...

jean-airoldie:
From the git-scm doc:
> `submodule.<name>.url`
> Defines a URL from which the submodule repository can be cloned. This may be either an absolute URL ready to be passed to git-clone[1] or (if it begins with ./ or ../) a location relative to the superproject’s origin repository.

So this kinda sounds like a git bug indeed.

jean-airoldie:
Would hosting the tool directly in this repo be acceptable? Last this it was updated was 4years ago, so it wouldn't be too much of a headache to maintain.

wjakob:
It's a part of Clang and has a different (UIUC), although compatible license. I definitely wouldn't want to merge it directly into pybind11.

I'm warming up to the idea of moving this out into a separate repository, but this is a breaking change for everybody who relies on this stuff and thus would have to wait until the next major version release (pybind11 3.x.x.) which is still some ways off into the future.

PR #1877, opened 2019-08-12
Detect whether we are running in a Conda environment and adjust get_include()

sdebionne:
Try to address [this issue](https://github.com/conda-forge/pybind11-feedstock/pull/32#discussion_r293723814) with the conda-forge feedstock.

For the records, Conda packages install headers in `prefix/include` on Linux and `prefix\Library\include` on Windows.

### DIFF
--- a/pybind11/__init__.py
+++ b/pybind11/__init__.py
@@ -9,10 +9,18 @@ def get_include(user=False):
     # Are we running in a virtual environment?
     virtualenv = hasattr(sys, 'real_prefix') or \
         sys.prefix != getattr(sys, "base_prefix", sys.prefix)
+    
+    # Are we running in a conda environment?
+    conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
 
     if virtualenv:
         return os.path.join(sys.prefix, 'include', 'site',
                             'python' + sys.version[:3])
+    elif conda:
+        if os.name == 'nt':
+            return os.path.join(sys.prefix, 'Library', 'include')
+        else:
+            return os.path.join(sys.prefix, 'include')
     else:
         dist = Distribution({'name': 'pybind11'})
         dist.parse_config_files()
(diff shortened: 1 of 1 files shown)


pybind11/__init__.py
```diff
@@ -9,10 +9,18 @@ def get_include(user=False):
     # Are we running in a virtual environment?
     virtualenv = hasattr(sys, 'real_prefix') or \
         sys.prefix != getattr(sys, "base_prefix", sys.prefix)
+
```
ax3l: the PEP8 check fails since this empty line contains whitespaces

sdebionne: Fixed

wjakob:
Hi @sdebionne, @ax3l,

this looks reasonable -- do you want me to merge it, or should I wait until it is proven to work in an actual CI build (as suggested by @ax3l here: https://github.com/conda-forge/pybind11-feedstock/pull/32#issuecomment-521709969)

Thanks,
Wenzel

ax3l:
I added the current PR's changeset to the conda-forge feedstock now.
Yes, let's wait for it to pass before merging this here as well.

pybind11/__init__.py
```diff
@@ -10,9 +10,17 @@ def get_include(user=False):
     virtualenv = hasattr(sys, 'real_prefix') or \
         sys.prefix != getattr(sys, "base_prefix", sys.prefix)
 
+    # Are we running in a conda environment?
+    conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+
     if virtualenv:
```
ax3l: @sdebionne maybe that's a stupid question but let me ask it anyway:
can one be inside conda and as well, due to it's pip support, within a virtualenv? If so, how to handle this gracefully?

ax3l:
As the tests turned out, the `get_include()` paths were already broken under Linux as well when taken from a conda install. Good job @sdebionne fixing them as well! Thanks again!

@wjakob ready for merge :)

wjakob:
Okay, merged then.

wjakob:
btw: I assume that you'll want this PR to be included in the next patch release 2.3.1, right?

ax3l:
Yes, please :)

PR #1879, opened 2019-08-13
Sanitize FindPythonLibsNew.cmake

tvatter:
This PR fixes https://github.com/pybind/cmake_example/issues/24

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -76,16 +76,20 @@ endif()
 # VERSION. VERSION will typically be like "2.7" on unix, and "27" on windows.
 execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c"
     "from distutils import sysconfig as s;import sys;import struct;
-print('.'.join(str(v) for v in sys.version_info));
-print(sys.prefix);
-print(s.get_python_inc(plat_specific=True));
-print(s.get_python_lib(plat_specific=True));
-print(s.get_config_var('SO'));
-print(hasattr(sys, 'gettotalrefcount')+0);
-print(struct.calcsize('@P'));
-print(s.get_config_var('LDVERSION') or s.get_config_var('VERSION'));
-print(s.get_config_var('LIBDIR') or '');
-print(s.get_config_var('MULTIARCH') or '');
+save_stdout = sys.stdout
+sys.stdout = open('trash', 'w')
+values = ['.'.join(str(v) for v in sys.version_info),
+              sys.prefix,
+              s.get_python_inc(plat_specific=True),
+              s.get_python_lib(plat_specific=True),
+              s.get_config_var('EXT_SUFFIX'),
+              hasattr(sys, 'gettotalrefcount')+0,
+              struct.calcsize('@P'),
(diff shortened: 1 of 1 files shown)


wjakob:
For posterity, can you explain in more detail here what the actual problem is?

I'm not 100% happy with this PR yet: I found the `print()` statements much easier to understand than the big multi-line statement. Also, the script definitely shouldn't create files on disk ("trash").

tvatter:
In a conda environment with Python 3.7.4 and a bunch of packages installed, I noticed that the first call to `get_config_var` produces "garbage" output, which then messes with the rest of the script.

By capturing the unwanted output when calling the functions and printing afterwards, we present something like this from happening.

Also, `';'.join([str(item) for item in values])` directly converts into the list format from `cmake`, which makes lines 108 and 109 unnecessary.

Regarding the creation of files on disk, I'm pretty new to Python, but I can have a look at another way of "sinking" the unwanted output somewhere.

tvatter:
For the creation of files on disk, would a replace of `trash` by `os.devnull` be OK?

tvatter:
Regarding the multiple print vs how it is done in the PR: now the code conveys exactly what it is doing, namely (a) creating a list of values, (b) converting the Python list to cmake format, (c) outputing the list.

wjakob:
Does os.devnull work on all platforms (Windows?)

tvatter:
It is supposed to.

wjakob:
Ok, I see. Could you add a comment saying that one of the commands prints output to stdout just so that the context is clear? Please also fix the CI issues.

wjakob:
Just for clarification: if you look through the CI results, you'll see that many tests now fail with errors (ImportError: No module named pybind11_tests) which I assume is a side effect of the changes you have made here.

tvatter:
I guess too, but I have no idea what makes it fail on Python 2.7 (Linux/OSX) and Windows, as the changes are completely straightfoward. I will have a look today.

wjakob:
ping? :)

tvatter:
Last week, I tried to figure out what was going on:

https://travis-ci.org/pybind/pybind11/jobs/571536066#L1049

The issue is that I cannot reproduce locally, as it works with Python 2.7 for me, and I have no windows computer. I will borrow my wife's computer tonight to figure out the windows thing, but I really don't get what's happening on ubuntu since I have a very similar config :/

wjakob:
Somehow the suffix of the python binary gets corrupted. It is called ``pybind11_testsNone``, so it should not be a surprise that the python test suite is unable to import it.

wjakob:
Dear @tvatter,

I was wondering if there were any news here? Apart from the fix, I was also wondering if you could provide some more information about the circumstances where this problem appears. 

I could not the spurious output produced by ``s.get_config_var`` on Anaconda with Python 3.7 on MacOS.

Best,
Wenzel

tvatter:
I'm so sorry, I've not been to actually reproduce the issue, and I've been completely swamped by teaching since the semester started. If having this open bothers you, don't hesitate to close and I'll come back to this whenever I have more bandwidth available. Apologies.

wjakob:
Ok, I'll clear this for now then.

PR #1882, opened 2019-08-14
Add negative indexing support to stl_bind.

ali-beep:


wjakob:
This is a nice feature addition, but the variable update in an conditional expression is a no-go for me :). I think that it would be cleanest to factor out the arithmetic & throw statement into a small lambda function that takes the array size and requested entry index.

include/pybind11/stl_bind.h
```diff
@@ -115,6 +115,14 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
     using SizeType = typename Vector::size_type;
     using DiffType = typename Vector::difference_type;
 
+    auto wrap_i = [](DiffType i, SizeType n) {
+        if (i < 0)
+            i += n;
+        if (i < 0)
```
wjakob: Wouldn't it be better to make this check `if (i < 0 || i > n)` and then remove the exceptions elsewhere?

wjakob:
Merged, thanks!

PR #1884, opened 2019-08-14
Added set::contains and generalized dict::contains

superbobry:
Dynamically resolving `__contains__` on each call is wasteful since set has a public `PySet_Contains function`.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1224,8 +1224,9 @@ public:
     detail::dict_iterator begin() const { return {*this, 0}; }
     detail::dict_iterator end() const { return {}; }
     void clear() const { PyDict_Clear(ptr()); }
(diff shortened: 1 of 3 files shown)


wjakob:
This looks like a good change to me, but you'll have to fix the test suite regressions.

wjakob:
Also, it would be good to have at least 1 test case covering the new `set` contains method.

superbobry:
Sorry about that, added tests.

I also plan to add a separate `dict` accessor for `const char*`. Should I do it in this PR or open a new one?

wjakob:
NP -- please do it in the same PR. It would be good if `const std::string &` is also handled the same way.

superbobry:
Apparently the `*String` functions are only left for backward compatibility so I've decided not to use them.

The PR is ready for review.

wjakob:
Merged, thanks.

PR #1886, opened 2019-08-16
test pair-copyability on C++17 upwards

### LINKED ISSUE #1885: False positives from is_copy_constructible with C++17
When PYBIND11_CPP17 is defined, `detail::is_copy_constructible` may give false positives for complex structures, leading to compiler errors.

```c++
py::cast(std::map<int, std::unique_ptr<int>>{}); // does compile
py::cast(std::map<int, std::map<int, std::unique_ptr<int>>>{}); // does not compile
```

Undefining the macro, respectively unguarding the third variant of `is_copy_constructible` solves this problem.
Tested with gcc and clang(libc++) and pybind v2.3.0.

sgsaenger:
The stdlib falsely detects containers like `M=std::map<T, U>` as copyable, even when one of T and U is not copyable.
Therefore we cannot rely on the stdlib dismissing `std::pair<T, M>` by itself, even on C++17.

Fixes #1885

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -797,12 +797,11 @@ template <typename Container> struct is_copy_constructible<Container, enable_if_
         negation<std::is_same<Container, typename Container::value_type>>
     >::value>> : is_copy_constructible<typename Container::value_type> {};
 
-#if !defined(PYBIND11_CPP17)
-// Likewise for std::pair before C++17 (which mandates that the copy constructor not exist when the
-// two types aren't themselves copy constructible).
+// Likewise for std::pair
+// (after C++17 it is mandatory that the copy constructor not exist when the two types aren't themselves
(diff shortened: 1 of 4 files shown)


wjakob:
This looks good -- however, it would be good to have a test for a non-copyable data structure with pairs (e.g. `std::map` as you suggested) to ensure that this doesn't break in the future.

sgsaenger:
Upon writing the tests, i've noticed that the same problems seem to arise with `std::is_copy_assignable`, which is used in `detail::map_assignment` by `bind_map`.
I've just copied the constructible-fix in `cast.h` and created a custom `is_copy_assignable`.

sgsaenger:
Any updates on this, are the tests acceptable?

wjakob:
Looks good to me—thanks!

sgsaenger:
Thank you!

PR #1887, opened 2019-08-16
Added .empty() to all collection types

superbobry:
I did not add new tests because the implementation uses (supposedly already tested `size`) and is otherwise trivial.


wjakob:
I was surprised to learn that this function was missing :)

PR #1888, opened 2019-08-17
Added support for list insertion.

kingofpayne:
`pybind11::list` supports `append` method but lacks `insert`.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1266,7 +1266,8 @@ public:
         PyList_Append(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr());
     }
     template <typename T> void insert(size_t index, T &&val) const {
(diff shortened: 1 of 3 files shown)


wjakob:
Looks good, but this function still needs to be covered by a test.

wjakob:
What you said above about the signedness being a major change is a good point. What I propose then is that you remove the `static_cast` to create an unsigned version of this API. The transition to signed-everywhere-by-default is a separate issue that doesn't need to be tackled now.

kingofpayne:
Do you mean I make a signed-only version of `insert` for the API ?. Or do you want both signed (without `static_cast`) and unsigned (with `static_cast`) ? (There might be a typo in your previous message ? If y remove the `static_cast` but keep the method unsigned then the CI will fail). Can you please confirm, I would be very happy to make any change.

wjakob:
Ah, I realize now that my request made no sense. The function wants a signed integer as input, and your cast simply avoids a potential warning that would break CI. I'll merge this as-is then.

PR #1890, opened 2019-08-22
Add const to buffer:request()

slarew:
This PR simply marks `buffer::request()` as `const`.  Without the `const`, code cannot have a const py::buffer and request a buffer_info.

`request()` does not alter the object so it should be conceptually safe.


wjakob:
Looks good, thanks!

PR #1912, opened 2019-09-06
 Avoid conversion to `int_` rhs argument of enum eq/ne

### LINKED ISSUE #1830: enum instance equality broken in pybind11 2.3
## Issue description

In pybind11 2.3, comparing an enum instance with a non-enum attempts to cast the non-enum to int, resulting in broken behaviors when the non-enum is not castable to an int, or a string that can be cast to an int.

## Reproducible example code

```c++

enum Foo { a = 1 };

    pybind11::enum_<Foo>(m, "Foo")
        .value("a", Foo::a);
}
```
compile using the python_example example repo.

With pybind11 2.2.4:
```
$ python -c 'from python_example import Foo; print(Foo.a == None)'
False
$ python -c 'from python_example import Foo; print(Foo.a == "1")'
TypeError: __eq__(): incompatible function arguments. The following argument types are supported:
    1. (self: python_example.Foo, arg0: python_example.Foo) -> bool
    2. (self: python_example.Foo, arg0: int) -> bool

Invoked with: Foo.a, '1'
```
With pybind11 2.3:
```
$ python -c 'from python_example import Foo; print(Foo.a == None)'
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
$ python -c 'from python_example import Foo; print(Foo.a == "1")'
True
```

I would prefer that equality checks always return True or False (consistently with the stdlib's enum.Enum), although I can understand raising a TypeError too.  On the other hand, returning True when comparing with the string "1" is clearly incorrect.

sizmailov:
Fixes #1830

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1469,9 +1469,17 @@ struct enum_base {
                 },                                                                     \
                 is_method(m_base))
 
+        #define PYBIND11_ENUM_OP_CONV_LHS(op, expr)                                    \
+            m_base.attr(op) = cpp_function(                                            \
+                [](object a_, object b) {                                              \
+                    int_ a(a_);                                                        \
+                    return expr;                                                       \
+                },                                                                     \
+                is_method(m_base))
+
         if (is_convertible) {
-            PYBIND11_ENUM_OP_CONV("__eq__", !b.is_none() &&  a.equal(b));
-            PYBIND11_ENUM_OP_CONV("__ne__",  b.is_none() || !a.equal(b));
+            PYBIND11_ENUM_OP_CONV_LHS("__eq__", !b.is_none() &&  a.equal(b));
+            PYBIND11_ENUM_OP_CONV_LHS("__ne__",  b.is_none() || !a.equal(b));
 
             if (is_arithmetic) {
                 PYBIND11_ENUM_OP_CONV("__lt__",   a <  b);
@@ -1501,6 +1509,7 @@ struct enum_base {
             }
         }
 
+        #undef PYBIND11_ENUM_OP_CONV_LHS
         #undef PYBIND11_ENUM_OP_CONV
         #undef PYBIND11_ENUM_OP_STRICT
 
--- a/tests/test_enum.cpp
+++ b/tests/test_enum.cpp
@@ -13,11 +13,13 @@ TEST_SUBMODULE(enums, m) {
     // test_unscoped_enum
     enum UnscopedEnum {
         EOne = 1,
-        ETwo
+        ETwo,
+        EThree
     };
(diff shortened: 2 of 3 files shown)


wjakob:
This change looks reasonable, but shouldn't it also be implemented consistently for the other operations (e.g. less than operator for arithmetic types, etc.).

sizmailov:
I think the right way is to mock python behavior. Here is test snippet with int, string and custom types:

<details>

<summary>test snippet</summary>

```python
class A(object): 
  def __repr__(self): return "A()"
class B(object): 
  def __repr__(self): return "B()"

def test_binop(op, f):
  try:
    result = f()
    print("`{} {} {}` = `{}`({})".format(repr(a),op,repr(b),result,type(result)))
  except Exception as e:
    print("`{} {} {}` raises {}: {}".format(repr(a),op,repr(b),type(e), e))

def test_unop(op, f):
  try:
    result = f()
    print("`{} {}` = `{}`({})".format(op,repr(a),result,type(result)))
  except Exception as e:
    print("`{} {}` raises {}: {}".format(op,repr(a),type(e), e))

for a,b in [(A(), B()),
            (1, 2), 
            (1, A()),
            (A(), 1),
            ]:
  test_binop("!=", lambda: a!=b)
  test_binop("==", lambda: a==b)
  test_binop("<", lambda: a<b)
  test_binop("<=", lambda: a<=b)
  test_binop(">", lambda: a>b)
  test_binop("&", lambda: a&b)
  test_binop("|", lambda: a|b)
  test_binop("^", lambda: a^b)

for a in [A(), 1, "string"]:
 test_unop("~", lambda: ~a)
```

 </details>

snippet on repl.it: [(python2)](https://repl.it/repls/AmbitiousCuddlyAdaware) [(python3)](https://repl.it/repls/PepperyRudeTypes)

`A() < 1` is `False` in python2 and `TypeError` in python3. 

Current behavior is in line with python3. I've updated tests to check behavior of enum arithmetic and behavior on operations with unsupported types. 

Should we branch on python version or leave it as is? 

While current behavior is in line with pytnon3, the exception message is a bit off-putting:
```
>       y ^ object()
E       TypeError: int() argument must be a string, a bytes-like object or a number, not 'object'
```
Additional check will duplicate checks inside of `py::int_`, so it's a bit of redundant work on non-exceptional path. extra `try-catch` feels like heavy-weight, but it should not affect non-exceptional path. Should it be added?

sizmailov:
For some reason enumeration elements order in docstring is not determined. Replaced exact docstring comparison with every-line-exists-in-docstring check.

wjakob:
I don't think the extra check is needed, and defaulting to Python 3 behavior sounds good to me.

PR #1933, opened 2019-09-24
FIX: do not use un-documented CPython functions

### LINKED ISSUE #1932: compilation fails with cpython master due to removed function
## Issue description

Use with cpython master branch (of scipy) fails with

```
  creating build/temp.linux-x86_64-3.9/scipy/fft/_pocketfft
    compile options: '-DPOCKETFFT_PTHREADS -I/home/tcaswell/.virtualenvs/bleeding/include/site/python3.9 -I/home/tcaswell/.virtualenvs/bleeding/include -I/home/tcaswell/.pybuild/bleeding/include/python3.9 -c'
    extra options: '-std=c++14 -fvisibility=hidden'
    g++: scipy/fft/_pocketfft/pypocketfft.cxx
    In file included from scipy/fft/_pocketfft/pypocketfft.cxx:15:
    /home/tcaswell/.virtualenvs/bleeding/include/site/python3.9/pybind11/pybind11.h: In member function ‘void pybind11::gil_scoped_acquire::dec_r
    /home/tcaswell/.virtualenvs/bleeding/include/site/python3.9/pybind11/pybind11.h:1953:13: error: ‘PyThreadState_DeleteCurrent’ was not declare
     1953 |             PyThreadState_DeleteCurrent();
          |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
          |             PyThreadState_Delete
    error: Command "g++ -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DPOCKETFFT_PTHREADS -I/home/tcaswell/.virthon3.9/site-packages/numpy/core/include -I/home/tcaswell/.virtualenvs/bleeding/include -I/home/tcaswell/.pybuild/bleeding/include/python3.9 -c s/_pocketfft/pypocketfft.o.d -std=c++14 -fvisibility=hidden" failed with exit status 1
```

This is due to cpython removing this function for py3.9 : https://github.com/python/cpython/pull/15315 / https://bugs.python.org/issue37878

This function is used at https://github.com/pybind/pybind11/blob/f3109d8419b3ed56cf9795f6343ab08b0c27746a/include/pybind11/pybind11.h#L1939-L1957

## Reproducible example code

Trying to install scipy from source with cpython master.

tacaswell:
In python 3.9 PyThreadState_DeleteCurrent was removed, instead use
PyThreadState_Delete and pass it the current tstate we are working on.

Re-order the calls so that we PyThreadState_Delete can have access to
the state carried by tstate before we clear it.

closes #1932

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1949,8 +1949,8 @@ public:
                 if (!release)
                     pybind11_fail("scoped_acquire::dec_ref(): internal error!");
             #endif
+            PyThreadState_Delete(tstate);
             PyThreadState_Clear(tstate);
-            PyThreadState_DeleteCurrent();
             PYBIND11_TLS_DELETE_VALUE(detail::get_internals().tstate);
             release = false;
         }


tacaswell:
This makes it compile again, but I am skeptical that this actually fixes the problem as I suspect that the current thread is not the same as the one pointed to by `tstate`.  Feel free to close this if it is just noise.

tacaswell:
The still existing function does not let you delete the thread you are sitting in so does not work.

PR #1936, opened 2019-09-26
Add vcpkg installation instructions

grdowns:
pybind11 is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single command they can use to build pybind11 and include it into their projects


wjakob:
This is reasonable, but please put your addition at the end of the list. I have never heard of vcpkg and don't think it should be the first item of 'Build Systems' :).

henryiii:
Closes #1959

bstaletic (approved):

henryiii:
Thanks!

PR #1948, opened 2019-10-06
Fix version mismatch typos in .travis.yml

dalboris:
Hi everyone! I just noticed these typos, they might confuse people looking at the Travis builds. Thanks again for your work on pybind11!


wjakob:
Indeed! Thank you.

PR #1950, opened 2019-10-08
Adapt to python3.8 C API change

sizmailov:
Do `Py_DECREF(type)` on *all* python objects on deallocation

Tested with python 3.7.3 and python 3.8.0rc1


wjakob:
Super-cool -- thank you very much for tracking down the issue and fixing it.

sizmailov:
Note: This patch leads to segfault in python 3.8.0b4 (and I in earlier 3.8.0* versions too). Update to 3.8.0rc1 solves the problem.

PR #1951, opened 2019-10-09
Allow to vectorize functions with void return type

sizmailov:
Currently `py::vectorize` requires vectorized function to have non-void return type. It mimics  `numpy.vectorize` which *requires* function to have return (otherwise array of `None` is created). 

@henryiii found that `py::vectorize` does not accept functions with `void` return type and suggested a patch to allow it in pybind. 

This PR adds specialization of `py::details::vectorize_helper` for `void` return type. 

While I think the problem should be addressed I'm confident that it cannot be merged as is because
 - it takes +120 extra LOC in `numpy.h` (=> increase build time for everyone who use numpy)
 - 95% of primary template code is repeated in specialization

While first point might be solved by moving specialization to a separate header (suggest a name?) it would in turn decrease cohesion and maintainability. 

Second point can be solved by refactoring primary template, but it seems to be a non-trivial change and I suspect it would still require comparable amount of template code (~100 extra LOC compared to current version)

@henryiii @wjakob What would be your suggestions here?

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1595,6 +1595,125 @@ private:
     }
 };
 
+// Specialization for void return type
+template<typename Func, typename... Args>
+struct vectorize_helper<Func, void, Args...> {
+private:
+  static constexpr size_t N = sizeof...(Args);
+  static constexpr size_t NVectorized = constexpr_sum(vectorize_arg<Args>::vectorize...);
+  static_assert(NVectorized>=1,
+      "pybind11::vectorize(...) requires a function with at least one vectorizable argument");
+
+public:
+  template<typename T>
+  explicit vectorize_helper(T&& f)
+      : f(std::forward<T>(f)) { }
+
+  void operator()(typename vectorize_arg<Args>::type... args) {
+      run(args...,
+          make_index_sequence<N>(),
+          select_indices<vectorize_arg<Args>::vectorize...>(),
+          make_index_sequence<NVectorized>());
+  }
+
+private:
+  remove_reference_t<Func> f;
+
+  // Internal compiler error in MSVC 19.16.27025.1 (Visual Studio 2017 15.9.4), when compiling with "/permissive-" flag
+  // when arg_call_types is manually inlined.
+  using arg_call_types = std::tuple<typename vectorize_arg<Args>::call_type...>;
+  template<size_t Index> using param_n_t = typename std::tuple_element<Index, arg_call_types>::type;
+
+  // Runs a vectorized function given arguments tuple and three index sequences:
+  //     - Index is the full set of 0 ... (N-1) argument indices;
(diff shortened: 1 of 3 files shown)


henryiii:
Note: This is useful for accumulators (reductions) and histograms - really any class that has a method that is called on each element of an array.

I don't think a few extra LOC are going to affect compile times much (new STL headers can add many thousands), and it doesn't get instantiated unless used. But the maintenance part for heavily repeated code is a potential problem. So I'd be interested in the refactoring idea.

henryiii:
What's the current status?

sizmailov:
Again, I think it would be good addition to pybind, but I stuck with nasty implementation. 

The basic problem is that `void` is incomplete type, so code diverges and eventually gets duplicated.

Based on @YannickJadoul and @bstaletic advises on gitter chat I've implemented non-specialized implementation using SFINAE technique, but it didn't result to much sorter version. 

I agree here with @YannickJadoul: currently the PR is too ugly and likely not going to be merged. 

He also suggested that problem can be reduced to some degree by technique used in `py::cpp_function`, but I didn't have a chance to try it (code links: [1](https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L129-L131), [2](https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L159-L160))

Unfortunately I will be not able to work on it in next two weeks. May be someone end up with better solution in meantime.

include/pybind11/numpy.h
```diff
@@ -1534,31 +1534,49 @@ struct vectorize_helper {
...
+    template <bool flag=!is_noreturn, typename std::enable_if<flag>::type* =nullptr,
+              size_t... Index>
+    object apply_to_single_value(const std::array<void *, N> &params, const index_sequence<Index...>&) {
+        return cast(f(*reinterpret_cast<param_n_t<Index> *>(params[Index])...));
+    }
 
-        return std::move(result);
+    template <bool flag=is_noreturn, typename std::enable_if<flag>::type* =nullptr,
+              size_t... Index>
+    py::none apply_to_single_value(const std::array<void *, N> &params, const index_sequence<Index...>&) {
```
YannickJadoul: Need to remove the `py::` namespaces, as the namespace alias is not defined by pybind11

YannickJadoul:
@sizmailov, @henryiii, I played around a bit and I found an alternative that seems less intrusive to the existing code: https://github.com/YannickJadoul/pybind11/tree/vectorize-no-return (main commit: https://github.com/YannickJadoul/pybind11/commit/087410166621528ad449c9395e30b616a04516b5)

 If you're still interested, have a look and tell me what you think. Maybe we can combine elements from different solutions into an even better PR :)

sizmailov:
Superseded by #1969

Issue #1604, opened 2018-11-10
"no matching function for call to 'operator delete'" in projected targeting C++17, Clang 7.0


cosnxdx:
## Issue description
Build targeting C++17 with Clang version 7.0 fails with error:

> error: no matching function for call to 'operator delete' ::operator delete(p,s,std::align_val_t(a));
> 
> note: candidate function not viable: no known conversion from 'pybind11::size_t' (aka 'unsigned long') to 'ImNewDummy' for 2nd argument
> inline void  operator delete(void*, ImNewDummy, void*)

replacing operator delete(p, s, std::align_val_t(a)) with operator delete(p) solves the build issue

wjakob:
Looks like you are working with a class that defines a custom deletion operator. We'll need a minimal self-contained example before this issue can be discussed in more detail.

jdumas:
I can reproduce this by compiling the pybind11 repo in C++17 mode:
```
cmake -DCMAKE_CXX_COMPILER=clang++ -DPYBIND11_CPP_STANDARD=-std=c++17 ..
```

ndoxx:
Hi,
I think I solved this problem by enabling sized deallocation under Clang, which is not enabled by default (see https://bcain-llvm.readthedocs.io/projects/clang/en/release_37/ReleaseNotes/). 

Use the option -fsized-deallocation. The macro __cpp_sized_deallocation should then be defined in the C++ header "new", and the operator delete with signature void operator delete(void*, std::size_t, std::align_val_t) will be declared.

NikolausDemmel:
> Use the option -fsized-deallocation. 

Fixes it for me as well. Anyone know a cross-platform (cmake) way to check when this should be added?

jquesnelle:
I ran into this too. It might make sense to detect this (Clang/C++17) 
in the pybind cmake and add `-fsized-deallocation` to the build interface since pybind explicitly is invoking sized deallocation. More background on Clang: https://reviews.llvm.org/D8467

jwnimmer-tri:
There's a chance that my RobotLocomotion/pybind11#30 has resolved the problem.  (You are welcome to cherry-pick it here upstream, if you like.)

e00E:
I ran into this problem when compiling the first example in the documentation with clang:
```c++
#include <pybind11/pybind11.h>

int add(int i, int j) {
    return i + j;
}

PYBIND11_MODULE(example, m) {
[...]

PR #1957, opened 2019-10-18
guard sized deallocation with features test macro

skoslowski:
This adds a additional compile-time check for sized deallocation support; fixing #1604:

- clang has sized deallocation disabled by default. [C++ Support in Clang](https://clang.llvm.org/cxx_status.html#n3778)
- checking the language standard version is not sufficient, so also check [`__cpp_sized_deallocation`](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17)
- with this patch the work-around mentioned in #1604 using `-fsized-deallocation` is no longer necessary / required

## Reproducible example code

build the unit tests with clang and C++17  (here: cmake 3.14.5, clang 8.0.0):
```
$ cmake -S . -B build -DCMAKE_CXX_COMPILER=clang++ -DPYBIND11_CPP_STANDARD=-std=c++17
$ cmake --build build
[...]
../include/pybind11/pybind11.h:1008:9: error: no matching function for call to 'operator delete'
[...]
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1003,7 +1003,7 @@ void call_operator_delete(T *p, size_t s, size_t) { T::operator delete(p, s); }
 
 inline void call_operator_delete(void *p, size_t s, size_t a) {
     (void)s; (void)a;
-#if defined(PYBIND11_CPP17)
+#if defined(PYBIND11_CPP17) && defined(__cpp_sized_deallocation)
     if (a > __STDCPP_DEFAULT_NEW_ALIGNMENT__)
         ::operator delete(p, s, std::align_val_t(a));
     else


wjakob:
Don't forget about the aligned `operator new` in cast.h. (Line 579)

wjakob:
See also https://github.com/RobotLocomotion/pybind11/pull/30.

skoslowski:
Good point, I wasn't aware of https://github.com/RobotLocomotion/pybind11/pull/30, but it seems a comprehensive solution. It does potentially change behavior for anyone not on `PYBIND11_CPP17`, but has `__cpp_aligned_new`.

I'd be happy to reproduce these patches here, keeping the existing C++17 check for compatibility, if that is preferred. Another option would be to cherry-pick the original PR.

wjakob:
I've cherry-picked the commit from RobotLocomotion and will close this instead.

PR #1965, opened 2019-10-22
Better error report

HDembinski:
This adds better error messages when py::cast is not used correctly.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -535,10 +535,11 @@ public:
                 if (copy_constructor)
                     valueptr = copy_constructor(src);
                 else {
+                    std::string type_name(tinfo->cpptype->name());
+                    detail::clean_type_id(type_name);
                     std::ostringstream os;
                     os << "return_value_policy = copy, but object "
-                       << detail::clean_type_id(tinfo->cpptype->name())
-                       << " is non-copyable!";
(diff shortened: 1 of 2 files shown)


HDembinski:
This job merely had a timeout, could someone with privileges please restart it?
https://travis-ci.org/pybind/pybind11/jobs/601280214

wjakob:
I would consider an error message like that, but only if compiled in debug mode. (Analogous to other uses of fancy string manipulation throughout the library, which include a note of the sort " (please recompile in debug mode for more details)". This is to avoid code bloat.)

wjakob:
build restarted.

HDembinski:
Ok, makes sense. My solution is still not perfect though, because the type name reported is without qualifiers, eq. `const T&` is reported as `T`. This is unfortunately how the typeid operator works, it drops all those qualifiers. Unfortunately, the qualifiers are critical to debug issues with the caster.

Keeping the qualifiers will be a more involved patch, which seems better suited for a separate PR.

HDembinski:
This version now only prints the type name in a debug build.

wjakob:
Looks good, thanks! Regarding getting the extra flags: I think the easiest and most cross-platform way is to wrap the type ``T`` into some dummy template ``foo<T>``, stringify that, and then strip of the outer parts.

Issue #1966, opened 2019-10-22
CI build passes despite failures


HDembinski:
See
https://travis-ci.org/pybind/pybind11/jobs/601262398

This should have failed since some tests fail, but the build wrongly passed.

wjakob:
closed via #1967.

PR #1967, opened 2019-10-22
make builds with python tests and cpp tests fail if either one fails

HDembinski:


HDembinski:
This addresses #1966

wjakob:
Good catch -- thank you!

PR #1969, opened 2019-10-24
Allow to vectorize functions with void return type - alternative implementation

YannickJadoul:
As discussed with @sizmailov and @henryiii, this is a smaller/less-intrusive implementation of PR #1951.

Quoting @sizmailov:

>  Currently py::vectorize requires vectorized function to have non-void return type. It mimics  numpy.vectorize which requires function to have return (otherwise array of None is created). @henryiii found that py::vectorize does not accept functions with void return type and suggested a patch to allow it in pybind.

Quoting @henryiii:

> Note: This is useful for accumulators (reductions) and histograms - really any class that has a method that is called on each element of an array.


include/pybind11/numpy.h
```diff
@@ -1482,6 +1482,54 @@ struct vectorize_arg {
     using type = conditional_t<vectorize, array_t<remove_cv_t<call_type>, array::forcecast>, T>;
 };
 
+
+template <typename Func, typename Return, typename... Args>
```
henryiii, line 1486: Maybe add a comment above each version of `vectorize_returned_array`, to explain what it does? It's not that obvious from quick glance without the context of the PR.

henryiii (approved):

bstaletic (approved):
This looks good to me. Any perceived uglyness, I guess, stems from `void` not being a Regular type.

include/pybind11/numpy.h
```diff
@@ -1583,27 +1634,26 @@ struct vectorize_helper {
...
+                         Return *out,
+                         size_t size,
+                         const std::vector<ssize_t> &output_shape,
                          index_sequence<Index...>, index_sequence<VIndex...>, index_sequence<BIndex...>) {
 
-        buffer_info output = output_array.request();
-        multi_array_iterator<NVectorized> input_iter(buffers, output.shape);
+        multi_array_iterator<NVectorized> input_iter(buffers, output_shape);
 
-        for (array_iterator<Return> iter = array_begin<Return>(output), end = array_end<Return>(output);
```
YannickJadoul, line 1600: `array_begin` and `array_end` are in the `detail` namespace, and futher unused. Should we remove those?

YannickJadoul (commented):
Meh. Somehow not entirely convinced, but I don't immediately see a better way. So apart from the one note about `array_begin` and `array_end`, I guess we can just merge this? :-)

YannickJadoul:
> Maybe add a comment above each version of `vectorize_returned_array`, to explain what it does? It's not that obvious from quick glance without the context of the PR, you have to notice the Return missing.

Also, sorry it took so long to review my own old PR, @henryiii! Thanks for adding these extra fixes :-)

wjakob (approved):
This looks fantastic, thank you for managing to add void return values so succinctly (IIRC it was proposed a few times so far, but all of those previous PRs were quite intrusive.)

include/pybind11/numpy.h
henryiii: I think it would be fine to remove the whole block, including `array_iterator `:

```cpp
template  <class T>
using array_iterator = typename std::add_pointer<T>::type;

template <class T>
array_iterator<T> array_begin(const buffer_info& buffer) {
    return array_iterator<T>(reinterpret_cast<T*>(buffer.ptr));
}

template <class T>
array_iterator<T> array_end(const buffer_info& buffer) {
    return array_iterator<T>(reinterpret_cast<T*>(buffer.ptr) + buffer.size);
}
```

YannickJadoul: It's a confusing set of concepts anyway, because it *only* works for contiguous arrays. Thanks! I'm removing this, waiting for CI, then merging :-)

henryiii:
Tests "passed", just our standard 3.9 macOS segfault flake. Too bad 3.9 comes out on Monday without reporting this. :'(

YannickJadoul:
> Too bad 3.9 comes out on Monday without reporting this. :'(

Oh, yeah. Slipped my mind. But I also would have no clue where to start looking :-/

PR #1972, opened 2019-10-30
now has == operator and string prints for maps

siddpr:
Added 
    * comparison operator to map
    * to_string prints to map

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -1,3 +1,4 @@
+
 /*
     pybind11/std_bind.h: Binding generators for STL data types
 
@@ -508,6 +509,7 @@ NAMESPACE_BEGIN(detail)
 
 /* Fallback functions */
 template <typename, typename, typename... Args> void map_if_insertion_operator(const Args &...) { }
+template <typename, typename, typename... Args> void map_if_equal_operator(const Args &...) { }
 template <typename, typename, typename... Args> void map_assignment(const Args &...) { }
 
 // Map assignment when copy-assignable: just copy the value
@@ -546,10 +548,12 @@ void map_assignment(enable_if_t<
            }
     );
 }
-
+template <typename T>
+std::string to_string(const T& val){ return pybind11::str(pybind11::object(pybind11::cast(val)));}
 
 template <typename Map, typename Class_> auto map_if_insertion_operator(Class_ &cl, std::string const &name)
--> decltype(std::declval<std::ostream&>() << std::declval<typename Map::key_type>() << std::declval<typename Map::mapped_type>(), void()) {
+-> decltype(std::declval<std::ostream&>() << to_string(std::declval<typename Map::key_type>())
+                                          << to_string(std::declval<typename Map::mapped_type>()), void()) {
 
     cl.def("__repr__",
            [name](Map &m) {
@@ -559,7 +563,7 @@ template <typename Map, typename Class_> auto map_if_insertion_operator(Class_ &
             for (auto const &kv : m) {
                 if (f)
                     s << ", ";
-                s << kv.first << ": " << kv.second;
+                s << to_string(kv.first) << ": " << to_string(kv.second);
                 f = true;
             }
             s << '}';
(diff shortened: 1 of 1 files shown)


include/pybind11/stl_bind.h
```diff
@@ -559,7 +562,7 @@ template <typename Map, typename Class_> auto map_if_insertion_operator(Class_ &
             for (auto const &kv : m) {
                 if (f)
                     s << ", ";
-                s << kv.first << ": " << kv.second;
+                s << to_string(kv.first) << ": " << to_string(kv.second);
```
wjakob: Hm, going to Python here strikes me really as overkill. We should rather expect the user to provide an `operator<<` if they want to be able to stringify their objects. Can you please remove this part of the change?

include/pybind11/stl_bind.h
```diff
@@ -592,8 +603,26 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
 
     cl.def(init<>());
 
+    cl.def(init([](const Map& d){
+            auto m = std::unique_ptr<Map>(new Map());
```
wjakob: Please use stack allocation and a normal ``Return`` at the end. This will be handled by C++'s return value optimization.

include/pybind11/stl_bind.h
```diff
@@ -592,8 +603,26 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
...
+            for (auto const &kv : d) {
+              m->insert(std::make_pair(kv.first, kv.second));
```
wjakob: Wrong indentation, remove braces.

include/pybind11/stl_bind.h
```diff
@@ -592,8 +603,26 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
...
+            }
+            return m.release();
+    }));
+
+    cl.def(init([](const iterable& it) {
+          auto m = std::unique_ptr<Map>(new Map());
```
wjakob: ditto

include/pybind11/stl_bind.h
```diff
@@ -592,8 +603,26 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
...
+            for (auto const& kv : pybind11::dict(it))
+                  m->insert(std::make_pair(kv.first.cast<KeyType>(), kv.second.cast<MappedType>()));
```
wjakob: Making a python dictionary and then iterating over that doesn't seem very efficient. We should directly get the keys/values from the iterable.

include/pybind11/stl_bind.h
siddpr: agreed

include/pybind11/stl_bind.h
```diff
@@ -569,6 +569,14 @@ template <typename Map, typename Class_> auto map_if_insertion_operator(Class_ &
     );
 }
 
+template <typename Map, typename Class_> auto map_if_equal_operator ( Class_ &cl)
+-> decltype(std::declval<typename Map::key_type>() == std::declval<typename Map::key_type>() and
+  std::declval<typename Map::mapped_type>() == std::declval<typename Map::mapped_type>(), void()) {
+
+    cl.def("__eq__", [](const Map& lhs, const Map& rhs) {
```
wjakob: Both `std::map` and `std::unordered_map` actually provide their own equality test (that is potentially more efficient). See https://en.cppreference.com/w/cpp/container/map/operator_cmp and https://en.cppreference.com/w/cpp/container/unordered_map/operator_cmp. It would be preferable to rely on these implementations rather than recreating an equality test here.

include/pybind11/stl_bind.h
```diff
@@ -592,9 +600,26 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args&&.
 
     cl.def(init<>());
 
+    cl.def(init([](const Map& d) {
+        Map m;
```
wjakob: Both map (https://en.cppreference.com/w/cpp/container/map/map) and unordered_map (https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map) that should be used instead of re-creating a custom copy constructor here.

include/pybind11/stl_bind.h
siddpr: @wjakob Can you please point me to something for this? Casting key and value fail here

PR #1974, opened 2019-10-30
CMake: Add Python 3.8 to pybind11Tools

ax3l:
Add Python 3.8 to considered versions in CMake for additional hints.
https://cmake.org/cmake/help/v3.15/module/FindPythonLibs.html


wjakob:
Good catch!

PR #1976, opened 2019-10-31
Fixing SystemError when nb_bool/nb_nonzero sets a Python exception in type_caster<bool>::load

YannickJadoul:
If the error indicator is set by a failing conversion in `type_caster<bool>::load` (e.g., by trying to convert a `numpy.ndarray` of 2 or more values, resulting in `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`), the error is not noticed and Python complains that a non-NULL value is returned while an error has been set (`SystemError: [...] returned a result with an error set`).

This PR should fix that.

NOTE: Without the changes to `cast.h`, the newly added test fails:
```
  what():  SystemError: <class 'type'> returned a result with an error set
```
Is that not unexpected as well? Shouldn't all `error_already_set` exceptions be caught by pybind11?

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1173,6 +1173,9 @@ public:
                 value = (bool) res;
                 return true;
             }
+            else {
+                PyErr_Clear();
+            }
         }
         return false;
     }
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -318,11 +318,15 @@ def test_numpy_bool():
     import numpy as np
     convert, noconvert = m.bool_passthrough, m.bool_passthrough_noconvert
 
+    def cant_convert(v):
+        pytest.raises(TypeError, convert, v)
+
     # np.bool_ is not considered implicit
     assert convert(np.bool_(True)) is True
     assert convert(np.bool_(False)) is False
     assert noconvert(np.bool_(True)) is True
     assert noconvert(np.bool_(False)) is False
+    cant_convert(np.zeros(2, dtype='int'))
 
 
 def test_int_long():


YannickJadoul:
(Thanks for @tdegeus for reporting on Gitter and sticking along while debugging, btw!)

YannickJadoul:
> Is that not unexpected as well? Shouldn't all `error_already_set` exceptions be caught by pybind11?

OK, this is because `py::repr` and `py::str` are called in `cpp_function::dispatcher` when trying to report that the function could not be called, but this is outside the big try-catch. Somewhere in there, Python notices that there's been an error, returns `NULL` as result of `PyObject_Repr`, and `pybind11` throws an error, causing the Python interpreter to crash.
Not entirely clean/expected, but independent from the problem in this PR.

Ready to merge, as far as I'm concerned, then!

tdegeus:
Thanks @YannickJadoul !

wjakob:
What strikes me as odd about this PR is why this is something specific to the `bool` caster? Can't these types of errors crop up elsewhere as well?

YannickJadoul:
@wjakob The main problem is calling into the C API without checking the result (i.e., if it's `-1`, in this case). If that call goes wrong, `PyErr_Occurred()` will be not-NULL, but pybind11 continues because it expects errors to throw a `error_already_set` exception.

When other casters call into the C API, they do check the result for errors and clear if necessary:
https://github.com/pybind/pybind11/blob/a6355b00f84d997a9ddcf209b6464447432be78a/include/pybind11/cast.h#L1024
https://github.com/pybind/pybind11/blob/a6355b00f84d997a9ddcf209b6464447432be78a/include/pybind11/cast.h#L1219

wjakob:
Ok, fair enough -- merged, thanks!

YannickJadoul:
The two Travis CI jobs are failing because something (`brew` in one case) times out; it shouldn't be a problem with this PR.

Issue #787, opened 2017-04-07
Can't use holders around types with custom converters


jagerman:
This fails for any `T` with a custom type converter (e.g. `std::string`, `Eigen::MatrixXd`, `MyCustomConvertingType`, etc.):

```
m.def("f", [](std::shared_ptr<T>) {})
```

The issue is that the holder type casters assume that `type_caster<T>` is a `type_caster_base<T>`, and make use of `type_caster_base<T>` constructors and internals.  #786 makes it into a compile-time failure (before that it will result in strange run-time failures as we end up invoking a generic caster on an unregistered type).

jagerman:
I don't see a nice way to implement this for all casters.

One possible idea for thought: type casters could opt-in to supporting casting to/from holders by providing `static bool load_into(holder_type<T> &, handle src, bool convert)` and `static cast_from(const holder_type<T> &)`; then we can SFINAE on the existence of those methods when trying to cast a holder around a non-generic type.

uentity:
@jagerman, I rebuild my code against latest master branch with your commit and static asserts didn't triggered. I think that's because I don't actually specify class holder. I use custom casting directly for `shared_ptr< bs_numpy_array >`, because It makes sense in one place -- make and hold a reference to an array on C++ side during corresponding numpy array lifetime in Python.

Here you can see what I do: [https://github.com/uentity/blue-sky-re/blob/master/kernel/include/bs/python/nparray.h](url)

jagerman:
(URL tag arguments flipped; I do that all the time).

jagerman:
Ah, you have your own type caster for a shared_ptr specifically, which is going to bypass the built-in generic holder casters I added the static asserts to.

jagerman:
I think it's because of this part:
```C++
	operator sp_array*() { return &value; }
	operator sp_array&() { return value; }
	operator Array*() { return value.get(); }
	operator Array&() { return *value; }

	template <typename _T> using cast_op_type = pybind11::detail::cast_op_type<_T>;
```
so pybind could be getting a raw pointer to the shared_ptr, and trying to wrap *that* in a unique pointer.  I think you can change this to just:
```C++
	operator sp_array() { return value; }
	template <typename> using cast_op_type = sp_array;
```
to just make a copy of the shared pointer (rather than a reference or pointer) when it needs to be accessed.

As for why you don't hit the `static_assert`, I think it's the nested shared pointers here, and I'll try to figure out a fix for that as well.

[...]

PR #1985, opened 2019-11-14
Smart Pointers for non custom types

flippmoke:
Added a way for non custom types to take advantage of smart pointers which hopefully solves #787.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -16,6 +16,7 @@
 #include "detail/internals.h"
 #include <array>
 #include <limits>
+#include <memory>
 #include <tuple>
 #include <type_traits>
 
@@ -1453,9 +1454,11 @@ template <typename... Ts> class type_caster<std::tuple<Ts...>>
 
 /// Helper class which abstracts away certain actions. Users can provide specializations for
 /// custom holders, but it's only necessary if the type has a non-standard interface.
-template <typename T>
+template <typename type, typename holder_type>
 struct holder_helper {
-    static auto get(const T &p) -> decltype(p.get()) { return p.get(); }
+    static auto get(const holder_type &p) -> decltype(p.get()) { return p.get(); }
+    static auto get(holder_type &p) -> decltype(p.get()) { return p.get(); }
+    static holder_type create(type && val) { return holder_type(new type(std::forward<type>(val))); }
 };
 
 /// Type caster for holder types like std::shared_ptr, etc.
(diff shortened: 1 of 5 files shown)


flippmoke:
Some notes:

I know this change should update documentation and I am willing to do so, but would love to get feedback on implementation before updating docs. 

For backwards compatibility with individuals using `holder_helper` I could create other new structs as helpers here, but it felt strange as I tried to design the interface. This was especially true due to the const vs non const `get` that was implemented. 

I did consider making the different versions of `copyable_holder_caster` using SFINAE, but I couldn't seem to get it work properly. If someone thinks that would be a better approach I can attempt to do it. 

I didn't add as many tests for situations that are a bit odd and will add some tests for them again if this approach is useful. One example would be doing something like the following which would create the possibility that both methods are created I think -- I am not sure how SFINAE would work here.

```
py::bind_vector<std::vector<Object>, std::shared_ptr<std::vector<Object>>>(m, "ObjectVec");
```

flippmoke:
I just realized my implementation has some issues with the correct `return_value_policy` and handling correctly the lifetime of objects. I am going to attempt to correct this, so still a WIP.

flippmoke:
I think this fixed the return value policy but I think some review here if the `return_value_policy_override` is going to be unexpected as I think it might be in some situations. For example:

If you have a custom binding for a class `MyObject`, but if it does not use a `shared_ptr` for internal storage, a return value of `std::shared_ptr<MyObject>` would have its value moved out (I think). This might be unexpected for some users.

flippmoke:
@wjakob This is ready for a full review, but if this is is this something you don't see the library implementing let me know and I can close the PR and approach this in my work in some other way. Thanks!

DmitriGoloubentsev:
@flippmoke Did you find a workaround for this? I'm surprised this PR didn't get deserved attention. 

Using shared_ptr<std::vector<double>> is pretty common, but pybind fails for me on this.

rwgk:
Oh, interesting, I never noticed this PR before. I need to look closer.

But in the meantime, for years, I worked on this, and it's used by Google since spring 2021:

https://github.com/pybind/pybind11/tree/smart_holder

Just use that. It has everything that the `master` branch has. All you need to do is git clone from the `smart_holder` branch and then use `py::classh` instead of `py::class_` (and only when it actually matters for your use cases).

For the past couple months I've been working on getting it ready for merging into master. At the moment I'm reviewing the handling of `const unique_ptr<T> &`, I didn't give that enough attention before; but that's only a pretty obscure and small part of the `unique_ptr`/`shared_ptr` functionality.

DmitriGoloubentsev:
@rwgk Do you have plans to merge it in master?

rwgk:
> @rwgk Do you have plans to merge it in master?

#5339, created yesterday. Waiting for feedback.

rwgk:
P.S. The title of this PR is unfortunate: what is "custom"? I misunderstood it at first glance.

I recommend working on the PR description, to concisely and carefully state what this PR is about, with a simple example.

Not sure what to write in the title exactly, but I'd definitely try to avoid a vague term like "custom".

Usually I write the PR description first, then extract a title from that.

PR #1986, opened 2019-11-14
Prevent CMake Error on Custom Python Builds

mattdawkins:
Without this change, on custom python builds (for example, the debug CPython build I just did on windows) the sys.prefix variable can sometimes be empty in the above python call, leading to only 5 arguments being passed into the cmake regex call here, result in a cmake configure time error. Adding quotes ensures 6 arguments are passed in.


mattdawkins:
CMake Error at tools/FindPythonLibsNew.cmake:137 (string):
      string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
      command.
    Call Stack (most recent call first):
      tools/pybind11Tools.cmake:16 (find_package)
      CMakeLists.txt:33 (include)

wjakob:
LGTM, thanks!

PR #1988, opened 2019-11-15
Aligned allocation fix for clang-cl

bluescarni:
This is a tentative fix for an issue I have encountered while trying to build a pybind11 extension using clang-cl on Windows.

As a bit of background, clang-cl is essentially a drop-in binary-compatible replacement for the ``cl.exe`` compiler from MSVC, based on LLVM/clang. This means that one can install clang-cl on top of an existing MSVC installation, and then clang-cl can be used as a compiler in conjunction with the standard C++ library provided with MSVC.

The issue I ran into is that, on my specific setup, the type ``std::align_val_t`` is not available, but pybind11 is trying anyway to employ it. This happens because of the following:

- the MSVC standard library version I am using (MSVC 2015) does *not* provide ``std::align_val_t``, but, at the same time,
- I am compiling my code in C++17 mode, which results in clang-cl enabling the standard definition ``__cpp_aligned_new``,
- which is then checked by pybind11 in order to establish whether ``std::align_val_t`` (and aligned allocation more generally) is available,

hence the compilation error.

This is admittedly a bit of a corner-case setup, but, as far as I know, it is the only setup currently available on conda/conda-forge which allows to compile and package moderately-complicated C++17 software on Windows. conda-forge uses by default an older version of MSVC, which is not practically usable for C++17 development. Using the clang-cl compiler allows to sidestep the deficiencies of the MSVC compiler while at the same time retaining binary compatibility with it.

The proposed fix essentially runs a version check on the ``_MSC_VER`` variable, rather than trusting the presence of ``__cpp_aligned_new``. According to this webpage,

https://docs.microsoft.com/en-us/cpp/build/reference/zc-alignednew?view=vs-2019

aligned allocation is available on MSVC since Visual Studio 2017 version 15.5 (``_MSC_VER`` 1912).


wjakob:
A bit of a corner-case indeed, but the patch is thankfully very minimalistic. I'll merge it.

PR #1989, opened 2019-11-17
cast: Add C++ -> pytype caster

eacousineau:
Regarding Gitter conversation:
https://gitter.im/pybind/Lobby?at=5dd190b250010612b2d6e84a

I'm not sure if this is a completely desirable API, but seems OK at a first glance?
Although I couldn't figure out a way to easily test that the borrowing constructor was indeed called (wasn't sure of an easy way to implement refcounting...).

\cc @bstaletic

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1693,8 +1693,8 @@ T cast(const handle &handle) {
 }
 
 // pytype -> pytype (calls converting constructor)
-template <typename T, detail::enable_if_t<detail::is_pyobject<T>::value, int> = 0>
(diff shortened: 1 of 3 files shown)


eacousineau:
One of the major caveats is that this is by no means conservative: due to how `PYBIND11_OBJECT_CVT` works, it will fail fast if the type does not adhere to the check.
In the given `vector<float>` example from Gitter, [`PyList_Check`](https://docs.python.org/3.8/c-api/list.html#c.PyList_Check) could cause failure, which looks like it could happen if `stl_bind.h` is used (where `bind_vector` does not subclass from `PyList_Type` -- for good reason!).

PR #1990, opened 2019-11-17
Add AutoWIG to list of binding generators

virtuald:
I'm not associated with the AutoWIG project in any way, but I ran across it yesterday and played around with it a bit today. It does what it claims to do -- namely, spit out pybind11 bindings (the default appears to be boost.python). Unfortunately, the documentation is non-existent and to use it effectively it seems you really need to be a conda/scons user, but perhaps this little bump will help others find it and improve it.

I grabbed the text verbatim from their github site.

CC @pfernique (AutoWIG author)

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -289,9 +289,9 @@ code by introspecting existing C++ codebases using LLVM/Clang. See the
 .. [binder] http://cppbinder.readthedocs.io/en/latest/about.html
 
 [AutoWIG]_ is a Python library that wraps automatically compiled libraries into
-high-level languages. Our approach consists in parsing C++ code using LLVM/Clang
-technologies and generating the wrappers using the Mako templating engine. Our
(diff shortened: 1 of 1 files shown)


wjakob:
This looks fine, but the use of first perspective ("[..] our [..]") makes for a strange read in the documentation.

virtuald:
Good point, edited.

pfernique:
@virtuald it's true that the inner documentation is not really good but if you want more details, you can look at the article published (cited in the README) and I forgot to put a link to the following repository https://github.com/StatisKit/FP17 that contains all examples of the article.

Boost.Python is the default since I encountered some problems with my PyBind11 generator (In most cases it works but more work need to be done to references all problems that could be encountered and their solution)

virtuald:
@pfernique I think autowig is very promising, which is why I'd love to see more people trying to use it :)

wjakob:
Merged!

PR #1998, opened 2019-11-20
Add memoryview constructors for PyMemoryView_FromMemory

pwuertz:
I was looking for a way to expose a simple `memoryview` to python, i.e. pointer + size without additional overhead. [PyMemoryView_FromMemory()](https://docs.python.org/3/c-api/memoryview.html#c.PyMemoryView_FromMemory) seems to be well suited for this, but pybind11 has no matching constructor for it.

I added two alternative constructors for `py::memoryview` using the FromMemory creator, allowing for read/write access or read-only if `const` classified.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1353,6 +1353,17 @@ public:
             pybind11_fail("Unable to create memoryview from buffer descriptor");
     }
 
+    #if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 3)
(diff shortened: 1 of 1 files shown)


pwuertz:
Rebased to current master

virtuald:
This was implemented in e248869893163b95cffb4a63e899e5d933f3ef20

YannickJadoul:
Indeed; see #2223. Sorry this wasn't dealt with in time, @pwuertz :-/
Thanks for pointing this out, @virtuald!

PR #2000, opened 2019-11-21
Add FAQ entry for dealing with long functions interruption

CharlesB2:
I've spent some time figuring out how to correctly interrupt a long-running function when a SIGINT occurs, so I thought this entry could be useful.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -262,6 +262,13 @@ throwing an exception that gets translated to KeyboardInterrupt
 
 .. code-block:: cpp
 
+    class interruption_error: public std::exception {
+    public:
+        const char* what() const noexcept{
+            return "Interruption signal caught.";
+        }
+    };
+
     PYBIND11_MODULE(example, m)
     {
         m.def("long running_func", []()
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -254,17 +254,18 @@ How can I properly handle Ctrl-C in long-running functions?
 Ctrl-C is received by the Python interpreter, and holds it until the GIL
 is released, so a long-running function won't be interrupted.
 
-To interrupt from inside your function, you can use the ``PyErr_CheckSignals()`` function,
-that will tell if a signal has been raised on the Python side.
-It's a flag check so its impact is negligible. If a signal is on, interrupt execution by
-throwing an exception that gets translated to KeyboardInterrupt
-(see :doc:`advanced/exceptions` section):
+To interrupt from inside your function, you can use the ``PyErr_CheckSignals()``
+function, that will tell if a signal has been raised on the Python side.  This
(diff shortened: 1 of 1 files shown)


bjodah:
I've dabbled with interrupts so I appreciate this. What would be an example of `interruption_error()`? Is there any such default exception in C++ stdlib? Or do one need to create one and specify it's translation to Python's `KeyboardInterrupt` using pybind11?

CharlesB2:
Sorry I didn't include it, it's a user-defined exception since C++ stdexcept has no equivalent:

```
class interruption_error: public exception {
public:
    const char* what() const noexcept{
        return "Interruption signal caught."; 
    }
};
```

docs/faq.rst
```diff
@@ -248,6 +248,40 @@ that that were ``malloc()``-ed in another shared library, using data
...
 
+How can I properly handle Ctrl-C in long-running functions?
+===========================================================
+
+Ctrl-C is received by the Python interpreter, and holds it until the GIL
+is released, so a long-running function won't be interrupted.
+
+To interrupt from inside your function, you can use the ``PyErr_CheckSignals()`` function,
+that will tell if a signal has been raised on the Python side.
+It's a flag check so its impact is negligible. If a signal is on, interrupt execution by
```
wjakob: "It's [..]" -> "This function merely checks a flag, [..]. When a signal has been received, you can explicitly interrupt"

CharlesB2: Thanks, updated wording

wjakob:
Thanks!

eacousineau:
Nice! From tinkering with this, I think this could possibly be simplified; will PR a small change.

Reading the docs, it looks like the exception would have already been raised if this function returns non-zero:
https://docs.python.org/3.6/c-api/exceptions.html?highlight=pyerr_checksignals#c.PyErr_CheckSignals
https://docs.python.org/2.7/c-api/exceptions.html?highlight=pyerr_checksignals#c.PyErr_CheckSignals

So rather than create a redundant `KeyboardInterrupt`, it looks like you can just throw `py::error_already_set`. The following seems to work for me:
```c++
  m.def("stuff", []() -> void {
    while (true) {
      if (PyErr_CheckSignals() != 0) {
      }
      using namespace std::chrono_literals;
      std::this_thread::sleep_for(50ms);
    }
  });
```
```py
try:
  m.stuff()
  assert False
except KeyboardInterrupt:
  print("[ Done ]")
```
Full code here: https://github.com/eacousineau/repro/blob/f791b817ae88743cff66f7214f40a487ae99d843/python/pybind11/custom_tests/test_tmp.cc

eacousineau:
Posted: https://github.com/pybind/pybind11/pull/2007

CharlesB2:
Great, much simpler with `py:error_already_set`!

PR #2016, opened 2019-12-05
Polymorphic type hook base

rwgk:
Allows users to specialize polymorphic_type_hook with std::enable_if.

Creating pull request on behalf of Ken Oslund (https://github.com/kenoslund-google), the original author of the changes.

Currently user specializations of the form

template struct polymorphic_type_hook<itype, std::enable_if_t<...>> { ... };

will fail if itype is also polymorphic, because the existing specialization will also
be enabled, which leads to 2 equally viable candidates. With this change, user provided
specializations have higher priority than the built in specialization for polymorphic types.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -835,19 +835,25 @@ NAMESPACE_END(detail)
 // You may specialize polymorphic_type_hook yourself for types that want to appear
 // polymorphic to Python but do not use C++ RTTI. (This is a not uncommon pattern
 // in performance-sensitive applications, used most notably in LLVM.)
+//
+// polymorphic_type_hook_base allows users to specialize polymorphic_type_hook with
+// std::enable_if. User provided specializations will always have higher priority than
+// the default implementation and specialization provided in polymorphic_type_hook_base.
 template <typename itype, typename SFINAE = void>
-struct polymorphic_type_hook
+struct polymorphic_type_hook_base
 {
     static const void *get(const itype *src, const std::type_info*&) { return src; }
 };
 template <typename itype>
-struct polymorphic_type_hook<itype, detail::enable_if_t<std::is_polymorphic<itype>::value>>
+struct polymorphic_type_hook_base<itype, detail::enable_if_t<std::is_polymorphic<itype>::value>>
 {
     static const void *get(const itype *src, const std::type_info*& type) {
         type = src ? &typeid(*src) : nullptr;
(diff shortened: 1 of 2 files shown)


kenoslund-google:
Currently user specializations of the form

template <typename itype> struct polymorphic_type_hook<itype, std::enable_if_t<...>> { ... };

will fail if itype is also polymorphic, because the existing specialization will also
be enabled, which leads to 2 equally viable candidates. With this change, user provided
specializations have higher priority than the built in specialization for polymorphic types.

wjakob:
Hi @rwgk, @kenoslund-google,

thank you for this change, it looks very reasonable to me. Two things are still missing though:

1. An description + example in the documentation on how to use polymorphic types in this way
2. A testcase that would have failed with the previous version of pybind11.

I plan to merge the PR once these are added.

Thanks,
Wenzel

rwgk:
Hi @wjakob, thanks for your feedback!

I updated the existing test_tagbased_polymorphic.cpp. It's a one-line change, to add a default virtual destructor. There are new source code comments explaining why. Without the cast.h change in this PR, the modified test_tagbased_polymorphic.cpp fails to build. The full error is in the commit message.

We (@kenoslund-google and I) believe that the existing documentation doesn't need updating, because this PR is essentially just fixing a bug/oversight.

(I rebased this PR before committing the change to the test, which pulled in 14 commits. I hope I did the right thing.)

kenoslund-google:
Ralf- I'm not sure the rebase really did what you want. When you look at the files changed, there are a lot of extraneous diffs, which are not related to this change. I think you want to move the original commit (c3bfaf1d5a5c2845fd9591c14d7b9490f58eaf6c) to be on top of/ after all those 14 other commits you pulled in, then fold that and the new commit (26181a8b8444ad90d2a70114e4f11c21344ca36d) into a single change.

rwgk:
I was afraid ... I was trying to do the equiv of a g4 sync, following some
instructions I found googling. Let's see what Jacobs says? I think multiple
commits is fine, from a remark someone made in December, but I'm less
certain how to do the g4 sync equiv correctly.

rwgk:
Hi @wjakob, I just rebased this PR again and cleaned up the commits (the accidental diffs from before are gone). It's a very small PR, just two files changed. We have an open source project that depends on this PR (https://github.com/google/tink). Is there a chance you could merge this PR, to unblock that project?

wjakob:
Merged, thank you!

Issue #1610, opened 2018-11-16
register_exception does not work after re-initializing an embedded interpreter


bkloster-sma:
Consider the following example, which adds an exception type `SomeException` to module `foo` , re-initializes the interpreter, and tries to do the same again:

```
PYBIND11_EMBEDDED_MODULE(foo, mod) { /* Nothing */ }

struct SomeException : public std::runtime_error {};

int main() {

    {
        pybind11::scoped_interpreter interpreter;
        auto foo = pybind11::module::import("foo");
        pybind11::register_exception<SomeException>(foo, "SomeException");
        assert(!foo.attr("SomeException").is_none());
    }

    {
        pybind11::scoped_interpreter interpreter;
        auto foo = pybind11::module::import("foo");
        pybind11::register_exception<SomeException>(foo, "SomeException");
        assert(!foo.attr("SomeException").is_none()); // This throws an exception
    }

}
```
The second call to `register_exception` fails silently, because the first `pybind11::exception` object is still alive due to its static lifetime in [`pybind11::detail::get_exception_object`](https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1777), making `register_exception` think that the type has already been added.

The same behavior can be observed when trying to add the same exception type to multiple modules:

```
PYBIND11_EMBEDDED_MODULE(foo, mod) { /* Nothing */ }
PYBIND11_EMBEDDED_MODULE(bar, mod) { /* Nothing */ }

struct SomeException : public std::runtime_error {};

int main() {
    pybind11::scoped_interpreter interpreter;
    auto foo = pybind11::module::import("foo");
    pybind11::register_exception<SomeException>(foo, "SomeException");
    pybind11::register_exception<SomeException>(bar, "SomeException");
    assert(!foo.attr("SomeException").is_none());
    assert(!bar.attr("SomeException").is_none()); // This throws an exception
}
```

rear1019:
The first case described is exactly the bug I encountered some days ago. I will shortly make a pull request with a fix. It should also fix the second case.

PR #2017, opened 2019-12-05
Fix: register_exception<>() fails after reinitialization of interpreter

rear1019:
See issue #1610 for reference. Test cases for issues described therein are included.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1821,7 +1821,25 @@ exception<CppException> &register_exception(handle scope,
                                             const char *name,
                                             PyObject *base = PyExc_Exception) {
     auto &ex = detail::get_exception_object<CppException>();
-    if (!ex) ex = exception<CppException>(scope, name, base);
+    if (!ex) {
+        /*
+         * This is the first time register_exception<>() is called:
+         * get_exception_object<>() returns a reference to a static,
+         * default-initialized (= empty) exception object. Perform
+         * initialization. Note that the exception is added to the module
+         * “scope” inside the constructor.
+         */
+        ex = exception<CppException>(scope, name, base);
+    } else {
+        /*
+         * register_exception<>() has already been called, i.e. the exception is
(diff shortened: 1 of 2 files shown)


bstaletic (approved):
@rear1019 Thanks for the pull request and sorry for such a late response.

include/pybind11/pybind11.h
```diff
@@ -1821,7 +1821,25 @@ exception<CppException> &register_exception(handle scope,
...
+        ex = exception<CppException>(scope, name, base);
+    } else {
+        /*
+         * register_exception<>() has already been called, i.e. the exception is
+         * initialized. Add the exception to the module “scope”
+         */
+        if (hasattr(scope, name))
+            pybind11_fail("Error during initialization: multiple incompatible "
+                          "definitions with name \"" + std::string(name) + "\"");
+        scope.attr(name) = ex;
```
YannickJadoul, line 1841: Side-effect: this also allows to register the same exception in multiple scopes. Do we want that? Isn't it better to error on this and tell the user to manually write `scope.attr(name) = ex;`?

Meanwhile, `ex`'s scope will still refer to the old object from before the re-initialization, no? Is this safe?

tests/test_embed/test_interpreter.cpp
```diff
@@ -282,3 +282,55 @@ TEST_CASE("Reload module from file") {
     result = module.attr("test")().cast<int>();
     REQUIRE(result == 2);
 }
+
+namespace rerun_register_exception
```
YannickJadoul, line 286: Very minor detail: do we need this namespace here? Why did you add it?
It might be good practice, but it's also not used for the other tests, so maybe consistency is more important?

YannickJadoul (commented):
The question in general is if it's a good idea to reinitialize the interpreter; Python's docs at least warn it's not a great idea.

But now we have some tests, and (if the remarks are handled) I don't see much of a problem with including this, so why not.

PR #2019, opened 2019-12-09
Fix a memory leak when creating Python3 modules.

T045T:


T045T:
Is there an easy way to repro the Python3.5 Travis failure? I'd like to run that with a debugger attached.

bstaletic:
So, unrelated to this PR, I've been playing with valgrind and pybind this morning. I can confirm that this solves the 104 byte leak.

As for the travis/python3.5 environment, take a look at 

https://github.com/pybind/pybind11/blob/master/.travis.yml#L70-L76

wjakob:
Hi @T045T, @bstaletic,

would you mind explaining in a bit more detail what the original problem is?

Best,
Wenzel

bstaletic:
@wjakob 

I'm not sure *how* this PR fixes the leak, but I can provide valgrind output to confirm that it does.

Running `PYTHONMALLOC=malloc valgrind python -c 'import ycm_core'` where `ycm_core` is an extension module [defined here](https://github.com/ycm-core/ycmd/blob/master/cpp/ycm/ycm_core.cpp), results in the following leak:

```
==14848== 104 bytes in 1 blocks are definitely lost in loss record 1,846 of 2,687
==14848==    at 0x4838DEF: operator new(unsigned long) (vg_replace_malloc.c:344)
==14848==    by 0x5A387C1: PyInit_ycm_core (in /home/bstaletic/work/ycmd/ycm_core.so)
==14848==    by 0x4C4FB89: _PyImport_LoadDynamicModuleWithSpec (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4C4FD69: ??? (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4B5E8C6: ??? (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4B6DE90: PyVectorcall_Call (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4BC2ADF: _PyEval_EvalFrameDefault (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4B8D039: _PyEval_EvalCodeWithName (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4B8DDEE: _PyFunction_Vectorcall (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4BC197C: _PyEval_EvalFrameDefault (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4B8DD5A: _PyFunction_Vectorcall (in /usr/lib/libpython3.8.so.1.0)
==14848==    by 0x4BBD88E: _PyEval_EvalFrameDefault (in /usr/lib/libpython3.8.so.1.0)
```

So pybind is indeed leaking 104 bytes:

```
==14543==    definitely lost: 104 bytes in 1 blocks
==14543==    indirectly lost: 0 bytes in 0 blocks
==14543==      possibly lost: 260,043 bytes in 1,413 blocks
==14543==    still reachable: 566,855 bytes in 4,552 blocks
```

The "possibly lost" and "still reachable" are from CPython itself, but the "definitely lost" is pybind's fault. Again, I don't know how this pull request works, but the above valgrind command runs clean with this PR.

T045T:
So, `PyModule_Create()` really expects to be called with a pointer to a static `PyModuleDef`. That is, I don't think it takes ownership. Creating the `PyModuleDef` with `new` and not freeing it anywhere was what caused the memory leak.

The original code then called `Py_INCREF()`, which works together with `Py_DECREF()` to implement [reference counting](https://docs.python.org/3/c-api/refcounting.html).
When the reference count reaches zero, the pointer is freed.
I figured it would be safest to use this mechanism to eventually deallocate the pointer, because I was not 100% certain that things elsewhere *don't* call `Py_INCREF()` or `Py_DECREF()` on it.

It seemed like nothing called `Py_DECREF()` on our module definition, however. The memory analysis tools kind of proved that.
That's why I added the `m_free` callback. That one is called when the module that `PyModule_Create()`, well, creates, is freed again. The argument is supposed to be a pointer to the module.

So this is how I called `Py_DECREF()` - now why did I change `new` to `PyMem_New()`? Because `Py_DECREF()` uses a special deallocation function that uses `free`, not `delete`, I needed the pointer to be created via `malloc`, which is exactly what `PyMem_New()` does.

So there is a risk that using `Py_INCREF()` and `Py_DECREF()` here is needless cargo culting, and I could have just used `new` and `delete` instead, but I'm not confident that that would not break.

T045T:
I also just pushed an updated version that might resolve the Segfault in the Travis builds – in case either of the pointers is `NULL` (`Py_XDECREF()` checks for `NULL`).

bstaletic:
@T045T In case you haven't, rebase onto latest master, since some travis fixes have been merged.

include/pybind11/pybind11.h
```diff
@@ -794,11 +794,16 @@ class module : public object {
...
         if (!options::show_user_defined_docstrings()) doc = nullptr;
 #if PY_MAJOR_VERSION >= 3
-        PyModuleDef *def = new PyModuleDef();
+        PyModuleDef *def = PyMem_New(PyModuleDef, 1);
         std::memset(def, 0, sizeof(PyModuleDef));
         def->m_name = name;
         def->m_doc = doc;
         def->m_size = -1;
+        def->m_free = [](void* module) {
+          if (module != nullptr) {
```
wjakob: pybind11 codebase uses 4 spaces for indentation, not 2

include/pybind11/pybind11.h
```diff
@@ -794,11 +794,16 @@ class module : public object {
...
+            Py_XDECREF( PyModule_GetDef((PyObject*) module));
```
wjakob: extra space after '('

T045T: Fixed. Note that neither of these were reported by `tools/check-style.sh`

wjakob:
Merged, thanks!

wjakob:
Hmm, I realize now that Python 3.5 segfaults likely due to this PR. I'll revert the commit for now until this issue is resolved.

T045T:
Fair. I hope to get to the bottom of that by the end of the week.

bstaletic:
@T045T I was able to repro the segfault with either gcc or clang. The only requirement was a debug version of python.

Cmake configuration output:
```
bstaletic@Gallifrey build  (git)-[tstate-leak]-% cmake -DCMAKE_BUILD_TYPE=Debug -DPYTHON_EXECUTABLE=/usr/bin/python3.5dm -DPYBIND11_PYTHON_VERSION=3.5 -DPYBIND11_CPP_STANDARD=-std=c++14 -DPYBIND11_WERROR=ON -DDOWNLOAD_CATCH=ON ..
-- The CXX compiler identification is Clang 9.0.0
-- Check for working CXX compiler: /usr/sbin/clang++
-- Check for working CXX compiler: /usr/sbin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: /usr/bin/python3.5dm (found suitable version "3.5.9", minimum required is "3.5")
-- Found PythonLibs: /usr/lib/libpython3.5dm.so
-- Building tests WITHOUT Eigen
-- Found Boost: /usr/lib64/cmake/Boost-1.71.0/BoostConfig.cmake (found suitable version "1.71.0", minimum required is "1.56")
-- Performing Test HAS_FLTO_THIN
-- Performing Test HAS_FLTO_THIN - Success
-- LTO enabled
-- Downloading catch v1.9.3...
-- Building interpreter tests using Catch v1.9.3
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- pybind11 v2.4.dev4
-- Configuring done
-- Generating done
-- Build files have been written to: /home/bstaletic/work/pybind11/build
```

After running `make pytest` and failing, I could do the following:

- `cd tests`
- `python3.5dm -m pytest test_eval.py`

```
bstaletic@Gallifrey tests  (git)-[tstate-leak]-% python3.5dm -m pytest test_eval.py
===================================================================== test session starts =====================================================================
platform linux -- Python 3.5.9, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
rootdir: /home/bstaletic/work/pybind11/tests, inifile: pytest.ini
collected 1 item

test_eval.py .                                                                                                                                          [100%]

====================================================================== 1 passed in 0.01s ======================================================================
Fatal Python error: Segmentation fault

Current thread 0x00007f0792657740 (most recent call first):
zsh: segmentation fault  python3.5dm -m pytest test_eval.py
```

Running the same in gdb:

```
bstaletic@Gallifrey tests  (git)-[tstate-leak]-% gdb python3.5dm --quiet
Reading symbols from python3.5dm...
(No debugging symbols found in python3.5dm)
(gdb) run -m pytest test_eval.py
Starting program: /usr/bin/python3.5dm -m pytest test_eval.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Detaching after fork from child process 9962]
/usr/lib/../share/gcc-9.2.0/python/libstdcxx/v6/xmethods.py:731: SyntaxWarning: list indices must be integers or slices, not str; perhaps you missed a comma?
  refcounts = ['_M_refcount']['_M_pi']
===================================================================== test session starts =====================================================================
platform linux -- Python 3.5.9, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
rootdir: /home/bstaletic/work/pybind11/tests, inifile: pytest.ini
collected 1 item

test_eval.py .                                                                                                                                          [100%]

====================================================================== 1 passed in 0.01s ======================================================================

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7cecb24 in _Py_ForgetReference (op=op@entry=0x55555595f530) at Objects/object.c:1766
1766        if (op == &refchain ||
(gdb) bt
#0  0x00007ffff7cecb24 in _Py_ForgetReference (op=op@entry=0x55555595f530) at Objects/object.c:1766
#1  0x00007ffff7cec367 in _Py_Dealloc (op=0x55555595f530) at Objects/object.c:1794
#2  0x00007ffff7cd32e6 in free_keys_object (keys=0x555555ad9b40) at Objects/dictobject.c:351
#3  0x00007ffff7cd409b in dict_dealloc (mp=0x7ffff757f598) at Objects/dictobject.c:1654
#4  0x00007ffff7dc9e8e in _PyImport_Fini () at Python/import.c:301
#5  0x00007ffff7ddaaed in Py_Finalize () at Python/pylifecycle.c:605
#6  Py_Finalize () at Python/pylifecycle.c:519
#7  0x00007ffff7ddbb49 in Py_Exit (sts=sts@entry=0) at Python/pylifecycle.c:1474
#8  0x00007ffff7ddf8e7 in handle_system_exit () at Python/pythonrun.c:617
#9  0x00007ffff7ddfded in handle_system_exit () at Python/pythonrun.c:685
#10 PyErr_PrintEx (set_sys_last_vars=set_sys_last_vars@entry=1) at Python/pythonrun.c:627
#11 0x00007ffff7de015b in PyErr_Print () at Python/pythonrun.c:523
#12 0x00007ffff7dfe9de in RunModule (modname=<optimized out>, set_argv0=<optimized out>) at Modules/main.c:210
#13 0x00007ffff7dff8a0 in RunMainFromImporter (sys_path0=<optimized out>) at Modules/main.c:735
#14 Py_Main (argc=4, argv=<optimized out>) at Modules/main.c:749
#15 0x0000555555555194 in main ()
(gdb)
```

bstaletic:
Here's a minimal repro:

- Prerequisites:
  - Python 3.5 compiled like [this](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python35)
  - `foo.cpp` that contains the following code:

```c++
PYBIND11_MODULE(foo, m){}
```
- Steps to repro:
  -     c++ -shared foo.cpp -o foo.so -fPIC `python3.5dm-config --cflags --ldflags` -isystem pybind/include
  -     python -c 'import foo'

- Expected output: empty
- Actual output: `Segmentation fault`
- Backtrace: same as above

T045T:
I cannot repro this, either via CMake or your minimal repro case.

These are the steps I used to set up the dev environment:

```bash
# install prerequisites
sudo apt install -y make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl git valgrind

# set up pyenv
git clone https://github.com/pyenv/pyenv.git ~/.pyenv
echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc
echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc
echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bashrc
source ~/.bashrc

# install Python 3.5.9 with the same build settings as the arch config
PYTHON_CONFIGURE_OPTS="--enable-shared --with-threads --with-computed-gotos --enable-ipv6 --with-valgrind --with-system-expat --with-dbmliborder=gdbm:ndbm --with-system-ffi  OPT=\"-fPIC\"" pyenv install 3.5.9

cd $HOME/pybind11
pyenv local 3.5.9

# build pybind
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Debug -DPYTHON_EXECUTABLE=$(pyenv root)/versions/3.5.9/bin/python3 -DPYBIND11_PYTHON_VERSION=3.5 -DPYBIND11_CPP_STANDARD=-std=c++14 -DPYBIND11_WERROR=ON -DDOWNLOAD_CATCH=ON ..

# run test
make pytest -j4

# Alternatively, manually compile repro example
cd $HOME/pybind11
echo -e "#include <pybind11/pybind11.h>\nPYBIND11_MODULE(foo, m){}" > foo.cpp
c++ -shared foo.cpp -o foo.so -fPIC `$(pyenv root)/versions/3.5.9/bin/python3.5-config --cflags --ldflags` -isystem include

python -c 'import foo'
```

bstaletic:
That's strange... I don't see anything different compared to what I have and I can consistently repro.

I've actually started with a "debian:stretch" docker container, since that's what Travis is actually running. The following is what travis is executing for the problematic build:

```sh
export CXX=g++-6 CC=gcc-6
docker pull debian:stretch
containerid=$(docker run --detach --tty \
  --volume="$PWD":/pybind11 --workdir=/pybind11 \
  --env="CC=$CC" --env="CXX=$CXX" \
  debian:stretch)
SCRIPT_RUN_PREFIX="docker exec --tty $containerid"
$SCRIPT_RUN_PREFIX sh -c 'for s in 0 15; do sleep $s; apt-get update && apt-get -qy dist-upgrade && break; done'
cmake --version
$SCRIPT_RUN_PREFIX sh -c "for s in 0 15; do sleep \$s; \
 apt-get -qy --no-install-recommends install \
 python3.5-dbg python3-scipy-dbg python3.5-dev python3-pytest python3-scipy \
 libeigen3-dev libboost-dev cmake make g++-6 && break; done"
$SCRIPT_RUN_PREFIX cmake \
  -DCMAKE_BUILD_TYPE=Debug \
  -DPYTHON_EXECUTABLE=/usr/bin/python3.5dm \
  -DPYBIND11_PYTHON_VERSION=3.5 \
  -DPYBIND11_CPP_STANDARD=-std=c++14 \
  -DPYBIND11_WERROR=ON \
  -DDOWNLOAD_CATCH=ON \
  .
$SCRIPT_RUN_PREFIX make pytest -j 2 VERBOSE=1
$SCRIPT_RUN_PREFIX make cpptest -j 2
```

The `cpptest` step shouldn't be needed or even reached, since the segfault is in pytest.
Note that I haven't tried running this script all at once, I executed the commands manually.

bstaletic:
@T045T Actually, forget that docker. I know what's wrong with your build step...

I had to edit locally the PKGBUILD I linked. Add `--with-pydebug` to `PYTHON_CONFIGURE_OPTS`.

T045T:
That did it, thanks! Confirmed that changing to `new`/`delete` fixes the segfault, and works on at least 3.5 and 3.7 (I'll let Travis verify that nothing else broke).

PR #2022, opened 2019-12-11
Travis fixes

bstaletic:
Currently, breathe has a bug that breaks pybind CI. See https://github.com/michaeljones/breathe/issues/462

The other thing that is broken is `brew upgrade python` on osx 9. Switching the Travis image to 9.4 works.

cc @YannickJadoul


wjakob:
Thank you!

PR #2025, opened 2019-12-12
Fixes #1295: Handle debug interpreter

### LINKED ISSUE #1295: Why is linking to Python debug library disabled for Windows?
From detail/common.h

```
/// Include Python header, disable linking to pythonX_d.lib on Windows in debug mode
#if defined(_MSC_VER)
#  if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION < 4)
#    define HAVE_ROUND 1
#  endif
#  pragma warning(push)
#  pragma warning(disable: 4510 4610 4512 4005)
#  if defined(_DEBUG)
#    define PYBIND11_DEBUG_MARKER
#    undef _DEBUG
#  endif
#endif
```
Since I need debug build #undef _DEBUG caused me a lot of grief until I dicovered it.

On top of wanting to be able to debug my own code, I don't mind being able to step through Python for Windows dll code as well. In fact, in a few cases it was very useful.

Could anyone please explain why would one want to #undef _DEBUG?

JGamache-autodesk:
If a debug interpreter is detected, we allow linking with
pythonXX_d.lib under windows.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -103,7 +103,7 @@
 #  endif
 #  pragma warning(push)
 #  pragma warning(disable: 4510 4610 4512 4005)
-#  if defined(_DEBUG)
+#  if defined(_DEBUG) && !defined(Py_DEBUG)
 #    define PYBIND11_DEBUG_MARKER
 #    undef _DEBUG
 #  endif


wjakob:
Looks good, thanks!

PR #2026, opened 2019-12-13
Add C++20 char8_t/u8string support

VemundH:
This adds support for C++20 char8_t and std::u8string.
Using feature test macro __cpp_lib_char8_t to detect compiler support.
This also fixes test compilation error in C++20 mode.


wjakob:
Looks good, thank you!

PR #2031, opened 2019-12-16
Disable exception handling based on environment variable.

fabiant-graphcore:
This makes it much easier to debug C++ code behind pybind11 using GDB, since GDB will be able to catch the errors before pybind11 handles them.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -451,7 +451,7 @@ protected:
                 return none().release().ptr();
         }
 
-        try {
+        auto local_dispatch = [&] {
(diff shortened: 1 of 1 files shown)


wjakob:
It's not legal to throw C++ exceptions into a C-only codebase such as Python, hence I don't think that this is an option for merging. (Also, I'm not enthusiastic about reading environment variables from pybind11..)

fabiant-graphcore:
@wjakob Any other ideas to simplify using GDB on pybind11 programs then?

wjakob:
I usually enter "catch throw" or "b __cxa_throw" to trigger GDB in the moment when an exception is thrown.

PR #2038, opened 2019-12-19
Fix OS X aligned new for older MACOSX targets

wolfv:
It seems like Clang can compile the aligned new and delete, but then issues an error if compiling with OS X targets aiming at anything below OS X 10.12 (I assume the standard library of these older targets doesn't ship with the necessary functions).

It says: `error: call to unavailable function 'operator delete': introduced in macOS 10.12`
E.g. in this build log: https://travis-ci.com/wolfv/pygal/jobs/268938454

This happened to me on conda when compiling with `-std=c++17` enabled (as conda-forge seems to target an OS X version of 10.9 or 10.8 at the moment).

This is the LLVM PR https://reviews.llvm.org/D34556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -140,6 +140,10 @@ matrix:
     name: Python 3.7, c++14, AppleClang 9, Debug build
     osx_image: xcode9.4
     env: PYTHON=3.7 CPP=14 CLANG DEBUG=1
+  - os: osx
+    name: Python 3.7, c++17, AppleClang 9, Min OSX 10.9, Debug build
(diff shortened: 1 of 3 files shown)


virtuald:
I don't think this is needed on master anymore, due to https://github.com/pybind/pybind11/commit/759221f5c56939f59d8f342a41f8e2d2cacbc8cf? 

... however, it would be good to have a new pybind11 release that included that fix.

wjakob:
Closing, pybind 2.5 was released in the meantime.

Issue #2040, opened 2019-12-23
Handling rvalue function arguments


rhaschke:
I'm trying to wrap a [C++ function that accepts an rvalue argument (`std::unique_ptr<T>&&`)](https://github.com/pybind/pybind11/commit/98ea003e66c30d528ebacf48c0f3ca474c56f74f#diff-79ad37e791d4444c807b52332cb2153cR22-R26) and consumes it. In the [docs](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-unique-ptr) it is mentioned that `unique_ptrs` are not allowed as arguments, because the instance might still be used.
However, when passing the pointer as an rvalue reference, this should work anyway - the holder pointer still exists and re-accessing it in python should throw. That's the way it worked in boost::python at least. [There, it was possible to release auto_ptrs](https://github.com/pybind/pybind11/commit/6cfebdc7f16589734955214c17900bbd696766bb#diff-05d5257ade95412f3224ce0adff61725R23-R27). Accessing the python object subsequently, was just throwing:
```
Python argument types in
    test_move_arg_bp.access(Item)
did not match C++ signature:
    access(Item)
```

An example illustrating the pybind11 issue and a running `boost::python` example is available [here](https://github.com/pybind/pybind11/compare/master...rhaschke:rvalue-argument).

EricCousineau-TRI:
I believe this may be the same request as #1132 - can you confirm or, if it isn't, clarify?
(My old issue does kinda dive off the deep-end for inheritance too, though)

Is the real motivation here `unique_ptr<T>`, or do you have non-pointer-types that you want to move too?

rhaschke:
> Is the real motivation here unique_ptr<T>, or do you have non-pointer-types that you want to move too?

Our use case just needs to move `unique_ptr<T>`. But, to be honest, I don't (anymore) see the difference to rvalue function arguments in general. But I didn't touch this issue for nearly a year now.

EricCousineau-TRI:
Makes sense in pure C++, but yeah, it's a bit more nuanced with `pybind11` and native C++ types:
https://pybind11.readthedocs.io/en/stable/advanced/cast/index.html

To answer some of your questions here:
https://github.com/pybind/pybind11/pull/2047#issuecomment-570091389

> Why does `cast_op<T>` [...]

Can't answer that one easily :( I think we may have to mull that over a bit (\cc @YannickJadoul @wjakob)

[...]

PR #2046, opened 2019-12-30
Towards a solution for rvalue holder arguments in wrapped functions

rhaschke:
In #2040 I was complaining about a fundamental issue in pybind11, namely that rvalue function arguments (of custom types and particularly of `std::unique_ptr`) are not supported, i.e. the following function cannot be wrapped:
```c++
void consume(std::unique_ptr<CustomType>&& holder) {
  CustomType sink(std::move(*holder.release());
}
```
As this is a major blocker for my project to transition from boost::python to pybind11, I tried to investigate the issue in more detail.
It looks like, that a major design decision was to _explicitly_ forbid transferring ownership from python to C++. This allows maintaining a [_raw_ `value` pointer](https://github.com/pybind/pybind11/blob/f9f3bd711f266c435fa7cd8198e5696e3fb37eef/include/pybind11/detail/common.h#L400-L422) (in addition to the `holder`) for fast and direct access to the raw pointer. (I noticed that the holder's `get()` is only rarely called.)

If you are not willing to support this important use case, you should explicitly state this fact very prominently on some main documentation page(s). However, I hope that you _will_ add this feature in the future as pybind11 is a great replacement for boost::python, [providing many benefits](https://pybind11.readthedocs.io/en/stable/intro.html#goodies) and exhibiting much better code readability.

The above-mentioned design decision becomes noticeable - amongst other things - by the fact that [`move_only_holder_casters`](https://github.com/pybind/pybind11/blob/f9f3bd711f266c435fa7cd8198e5696e3fb37eef/include/pybind11/cast.h#L1559-L1569) do not support _loading_ of arguments and thus conversion from python to C++ is not possible for move-only holder types. 
The commits of this PR replace the existing implementation of `move_only_holder_caster` with that of `copyable_holder_caster` (with some minor tweaks), which essentially enables the requested feature. 
However, a major caveat remains: The holder pointer remains accessible after moving, while it should become `None` and thus raise an exception if accessed again. I tried to accomplish this [here](https://github.com/pybind/pybind11/commit/16bb1838ac59066a1e674844ed5e81c9ccc099a6#diff-e0d8df3cdf4b37acbf5e8712fb6ce80cR1532), but it didn't have the desired effect - I guess because pybind11 mainly accesses object instances by their raw value pointers instead of going through the holder.

Another major caveat comes from the overload-resolution process: During this process, several function overloads are tried to match in sequence, each one requiring to load the python arguments into the `type_caster`. However, this loading process already [_moves_ the holder into the caster](https://github.com/pybind/pybind11/commit/16bb1838ac59066a1e674844ed5e81c9ccc099a6#diff-e0d8df3cdf4b37acbf5e8712fb6ce80cR1566) and thus is only possible once.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -287,3 +287,11 @@ code by introspecting existing C++ codebases using LLVM/Clang. See the
 [binder]_ documentation for details.
 
 .. [binder] http://cppbinder.readthedocs.io/en/latest/about.html
+
+[AutoWIG]_ is a Python library that wraps automatically compiled libraries into
+high-level languages. It parses C++ code using LLVM/Clang technologies and
+generates the wrappers using the Mako templating engine. The approach is automatic,
+extensible, and applies to very complex C++ libraries, composed of thousands of
+classes or incorporating modern meta-programming constructs.
+
+.. [AutoWIG] https://github.com/StatisKit/AutoWIG
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1995,6 +1995,8 @@ private:
 
     template <size_t... Is>
     bool load_impl_sequence(function_call &call, index_sequence<Is...>) {
+        // BUG: When loading the arguments, the actual argument type (pointer, lvalue reference, rvalue reference)
+        // is already lost (argcasters only know the intrinsic type), while the behaviour should differ for them, e.g. for rvalue references.
         for (bool r : {std::get<Is>(argcasters).load(call.args[Is], call.args_convert[Is])...})
             if (!r)
                 return false;
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1527,9 +1527,8 @@ public:
 
     bool load(handle& src, bool convert) {
         bool success = base::template load_impl<move_only_holder_caster<type, holder_type>>(src, convert);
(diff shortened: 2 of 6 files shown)


rhaschke:
> I noticed that the holder's `get()` is only rarely called.

Trying to understand why this is the case, I learned that the default caster (`type_caster_base<T>`) has no notion of the expected holder type and never goes through the holder's `get()`!

rhaschke:
I think I fixed one remaining issue: Instead of moving the holder potentially multiple times during loading in overload resolution, I just "load" a _pointer_ to the holder and perform actual moving only when actually casting to `holder_type&&`.
However, the main issue of invalidating the python instance remains.

rhaschke:
I might have solved the 2nd remaining issue, namely invalidating the `PyObject` instance after moving, as well. In addition to invalidating and (re)checking the instance pointer, we probably need to check that there are no other python object instances accessing (member) elements of the to-be-moved object instance. Is `has_patients` a way to do so?

wjakob:
Transferring ownership from Python to C++ is a super-dangerous and rather unnatural (non-pythonic) operation and intentionally not supported in pybind11. It would be very hard to convince me to make a change here, and the current discussion & tentative patch don't make a strong enough case to support a very narrow use case IMO. Happy to discuss further, but I'm closing this for now. Sorry that it's taken me a while to respond.

PR #2047, opened 2019-12-30
[WIP] Towards a solution for custom-type rvalue arguments

rhaschke:
pybind11 doesn't support rvalue function arguments for custom types, i.e. the following function cannot be wrapped:
```c++
void consume(CustomType&& object) {
  CustomType sink(std::move(object));
}
```
This issue is (more or less) unrelated to rvalue holder arguments as requested e.g. in #2040 or #2583.
While moving holder arguments essentially transfers ownership from python to C++, which is admittedly difficult, supporting functions like above shouldn't pose a conceptual problem:
The original object instance remains existing in python, but only its _internals_ are moved over to C++.

Currently, this use case is not supported, because the `cast_op_type` provided by `type_caster_base<T>` doesn't support moving. [Replacing it with `movable_cast_op_type<T>`](https://github.com/pybind/pybind11/commit/fb0ecaa0ec28528fac6138048831eaafb0eab9e8#diff-e0d8df3cdf4b37acbf5e8712fb6ce80cR883) and [providing a corresponding cast operator](https://github.com/pybind/pybind11/commit/fb0ecaa0ec28528fac6138048831eaafb0eab9e8#diff-e0d8df3cdf4b37acbf5e8712fb6ce80cR887) already solves the problem for my example.

However, this solution makes moving preferred over copying, which isn't desired either.
**EDIT:** A major rework of this PR solves this issue, see https://github.com/pybind/pybind11/pull/2047#issuecomment-724256796 and https://github.com/pybind/pybind11/pull/2047#issuecomment-791916065 for more details.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -644,6 +644,10 @@ public:
     // static_cast works around compiler error with MSVC 17 and CUDA 10.2
     // see issue #2180
     explicit operator type&() { return *(static_cast<type *>(this->value)); }
+
+    // holders only support copying, not moving
+    template <typename T> using cast_op_type = detail::cast_op_type<T>;
+
     explicit operator holder_type*() { return std::addressof(holder); }
     explicit operator holder_type&() { return holder; }
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -50,14 +50,15 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 template <typename type, typename SFINAE = void> class type_caster : public type_caster_base<type> { };
 template <typename type> using make_caster = type_caster<intrinsic_t<type>>;
 
-// Shortcut for calling a caster's `cast_op_type` cast operator for casting a type_caster to a T
+// Shortcuts for calling a caster's `cast_op_type` cast operator for casting a type_caster to a T
+// cast_op operating on an lvalue-reference caster, enables moving only if required by the actual function argument
 template <typename T> typename make_caster<T>::template cast_op_type<T> cast_op(make_caster<T> &caster) {
     return caster.operator typename make_caster<T>::template cast_op_type<T>();
 }
+// cast_op operating on an rvalue-referenced caster enforces an rvalue-reference for the cast_op type as well
 template <typename T> typename make_caster<T>::template cast_op_type<typename std::add_rvalue_reference<T>::type>
 cast_op(make_caster<T> &&caster) {
-    return std::move(caster).operator
-        typename make_caster<T>::template cast_op_type<typename std::add_rvalue_reference<T>::type>();
+    return caster.operator typename make_caster<T>::template cast_op_type<typename std::add_rvalue_reference<T>::type>();
 }
 
 template <typename type> class type_caster<std::reference_wrapper<type>> {
@@ -101,7 +102,7 @@ public:
         } \
         operator type*() { return &value; } \
         operator type&() { return value; } \
-        operator type&&() && { return std::move(value); } \
+        operator type&&() { return std::move(value); } \
         template <typename T_> using cast_op_type = pybind11::detail::movable_cast_op_type<T_>
 
 
@@ -795,8 +796,8 @@ class type_caster<T, enable_if_t<is_pyobject<T>::value>> : public pyobject_caste
 // - type_caster<T>::operator T&() must exist
 // - the type must be move constructible (obviously)
 // At run-time:
-// - if the type is non-copy-constructible, the object must be the sole owner of the type (i.e. it
-//   must have ref_count() == 1)h
+// - if the type is non-copy-constructible, the object must be the sole owner of the type
+//   (i.e. it must have ref_count() == 1)
 // If any of the above are not satisfied, we fall back to copying.
 template <typename T> using move_is_plain_type = satisfies_none_of<T,
     std::is_void, std::is_pointer, std::is_reference, std::is_const
@@ -1163,7 +1164,7 @@ private:
 
     template <typename Return, typename Func, size_t... Is, typename Guard>
     Return call_impl(Func &&f, index_sequence<Is...>, Guard &&) && {
-        return std::forward<Func>(f)(cast_op<Args>(std::move(std::get<Is>(argcasters)))...);
+        return std::forward<Func>(f)(cast_op<Args>(std::get<Is>(argcasters))...);
     }
 
     std::tuple<make_caster<Args>...> argcasters;
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -855,12 +855,6 @@ template <typename T, typename SFINAE> type_caster<T, SFINAE> &load_type(type_ca
     }
     return conv;
 }
-// Wrapper around the above that also constructs and returns a type_caster
-template <typename T> make_caster<T> load_type(const handle &handle) {
-    make_caster<T> conv;
-    load_type(conv, handle);
-    return conv;
-}
 
 PYBIND11_NAMESPACE_END(detail)
 
@@ -870,7 +864,9 @@ T cast(const handle &handle) {
     using namespace detail;
     static_assert(!cast_is_temporary_value_reference<T>::value,
             "Unable to cast type to reference: value is local to type caster");
-    return cast_op<T>(load_type<T>(handle));
+    make_caster<T> conv;
+    load_type(conv, handle);
+    return cast_op<T>(conv);
 }
 
 // pytype -> pytype (calls converting constructor)
@@ -906,7 +902,9 @@ detail::enable_if_t<!detail::move_never<T>::value, T> move(object &&obj) {
 #endif
 
     // Move into a temporary and return that, because the reference may be a local value of `conv`
-    T ret = std::move(detail::load_type<T>(obj).operator T&());
+    detail::make_caster<T> conv;
+    load_type(conv, obj);
+    T ret = std::move(conv.operator T &());
     return ret;
 }
 
--- a/include/pybind11/eigen.h
+++ b/include/pybind11/eigen.h
@@ -346,7 +346,7 @@ public:
 
     operator Type*() { return &value; }
     operator Type&() { return value; }
-    operator Type&&() && { return std::move(value); }
+    operator Type&&() { return std::move(value); }
     template <typename T> using cast_op_type = movable_cast_op_type<T>;
 
 private:
--- a/tests/test_copy_move.py
+++ b/tests/test_copy_move.py
@@ -20,9 +20,7 @@ def test_move_and_copy_casts():
 
     cstats = m.move_and_copy_cstats()
     c_m, c_mc, c_c = (
-        cstats["MoveOnlyInt"],
-        cstats["MoveOrCopyInt"],
-        cstats["CopyOnlyInt"],
+        cstats[name] for name in ["MoveOnlyInt", "MoveOrCopyInt", "CopyOnlyInt"]
     )
 
     # The type move constructions/assignments below each get incremented: the move assignment comes
@@ -48,9 +46,7 @@ def test_move_and_copy_loads():
 
     cstats = m.move_and_copy_cstats()
     c_m, c_mc, c_c = (
-        cstats["MoveOnlyInt"],
-        cstats["MoveOrCopyInt"],
-        cstats["CopyOnlyInt"],
+        cstats[name] for name in ["MoveOnlyInt", "MoveOrCopyInt", "CopyOnlyInt"]
     )
 
     assert m.move_only(10) == 10  # 1 move, c_m
@@ -74,15 +70,43 @@ def test_move_and_copy_loads():
     assert c_m.alive() + c_mc.alive() + c_c.alive() == 0
 
 
+def test_move_copy_class_loads():
+    """Call some functions that load custom type arguments and count the number of moves/copies"""
+
+    cs = m.move_and_copy_cstats()["MoveOnlyInt"]
+    o = m.MoveOnlyInt(3)
+    assert m.MoveOnlyInt.val(o) == 3
+    assert o.value == -1  # value becomes -1 after moving
+    o = m.MoveOnlyInt(4)
+    assert m.MoveOnlyInt.lref(o) == 4
+    assert m.MoveOnlyInt.rref(o) == 4
+    assert o.value == -1  # value becomes -1 after moving
+    assert (cs.copy_constructions, cs.move_constructions) == (0, 2)
+
+    cs = m.move_and_copy_cstats()["MoveOrCopyInt"]
+    o = m.MoveOrCopyInt(3)
+    assert m.MoveOrCopyInt.val(o) == 3
+    assert m.MoveOrCopyInt.lref(o) == 3
+    assert m.MoveOrCopyInt.rref(o) == 3
+    assert o.value == -1  # value becomes -1 after moving
+    assert (cs.copy_constructions, cs.move_constructions) == (1, 1)
+
+    cs = m.move_and_copy_cstats()["CopyOnlyInt"]
+    o = m.CopyOnlyInt(3)
+    assert m.CopyOnlyInt.val(o) == 3
+    assert m.CopyOnlyInt.lref(o) == 3
+    assert m.CopyOnlyInt.rref(o) == 3  # copies as well
+    assert o.value == 3  # value hasn't change
+    assert (cs.copy_constructions, cs.move_constructions) == (2, 0)
+
+
 @pytest.mark.skipif(not m.has_optional, reason="no <optional>")
 def test_move_and_copy_load_optional():
     """Tests move/copy loads of std::optional arguments"""
 
     cstats = m.move_and_copy_cstats()
     c_m, c_mc, c_c = (
-        cstats["MoveOnlyInt"],
-        cstats["MoveOrCopyInt"],
-        cstats["CopyOnlyInt"],
+        cstats[name] for name in ["MoveOnlyInt", "MoveOrCopyInt", "CopyOnlyInt"]
     )
 
     # The extra move/copy constructions below come from the std::optional move (which has to move
--- a/tests/test_copy_move.py
+++ b/tests/test_copy_move.py
@@ -26,14 +26,15 @@ def test_move_and_copy_casts():
     # The type move constructions/assignments below each get incremented: the move assignment comes
     # from the type_caster load; the move construction happens when extracting that via a cast or
     # loading into an argument.
-    assert m.move_and_copy_casts(3) == 18
+    assert m.move_and_copy_casts(3) == 6 * 3
     assert c_m.copy_assignments + c_m.copy_constructions == 0
     assert c_m.move_assignments == 2
     assert c_m.move_constructions >= 2
     assert c_mc.alive() == 0
-    assert c_mc.copy_assignments + c_mc.copy_constructions == 0
+    assert c_mc.copy_assignments == 0
+    assert c_mc.copy_constructions == 2
     assert c_mc.move_assignments == 2
-    assert c_mc.move_constructions >= 2
+    assert c_mc.move_constructions == 0
     assert c_c.alive() == 0
     assert c_c.copy_assignments == 2
     assert c_c.copy_constructions >= 2
@@ -50,21 +51,23 @@ def test_move_and_copy_loads():
     )
 
     assert m.move_only(10) == 10  # 1 move, c_m
-    assert m.move_or_copy(11) == 11  # 1 move, c_mc
+    assert m.move(10) == 10  # 0 move (just ref), c_mc
+    assert m.copy(11) == 11  # 1 copy, c_mc
     assert m.copy_only(12) == 12  # 1 copy, c_c
-    assert m.move_pair((13, 14)) == 27  # 1 c_m move, 1 c_mc move
-    assert m.move_tuple((15, 16, 17)) == 48  # 2 c_m moves, 1 c_mc move
+    assert m.move_pair((13, 14)) == 27  # 1 c_m move, 1 c_mc copy
+    assert m.move_tuple((15, 16, 17)) == 48  # 2 c_m moves, 1 c_mc copy
     assert m.copy_tuple((18, 19)) == 37  # 2 c_c copies
-    # Direct constructions: 2 c_m moves, 2 c_mc moves, 1 c_c copy
+    # Direct constructions: 2 c_m moves, 2 c_mc copies, 1 c_c copy
     # Extra moves/copies when moving pairs/tuples: 3 c_m, 3 c_mc, 2 c_c
     assert m.move_copy_nested((1, ((2, 3, (4,)), 5))) == 15
 
     assert c_m.copy_assignments + c_m.copy_constructions == 0
     assert c_m.move_assignments == 6
     assert c_m.move_constructions == 9
-    assert c_mc.copy_assignments + c_mc.copy_constructions == 0
-    assert c_mc.move_assignments == 5
-    assert c_mc.move_constructions == 8
+    assert c_mc.copy_assignments == 0
+    assert c_mc.copy_constructions == 5
+    assert c_mc.move_assignments == 6
+    assert c_mc.move_constructions == 3
     assert c_c.copy_assignments == 4
     assert c_c.copy_constructions == 6
     assert c_m.alive() + c_mc.alive() + c_c.alive() == 0
@@ -111,8 +114,9 @@ def test_move_and_copy_load_optional():
 
     # The extra move/copy constructions below come from the std::optional move (which has to move
     # its arguments):
-    assert m.move_optional(10) == 10  # c_m: 1 move assign, 2 move construct
-    assert m.move_or_copy_optional(11) == 11  # c_mc: 1 move assign, 2 move construct
+    assert m.move_optional(10) == 10  # c_m: 1 move assign, +1 move construct
+    assert m.mc_move_optional(10) == 10  # c_mc: 1 move assign, +0 move construct
+    assert m.mc_copy_optional(11) == 11  # c_mc: 1 move assign, 1 copy construct
     assert m.copy_optional(12) == 12  # c_c: 1 copy assign, 2 copy construct
     # 1 move assign + move construct moves each of c_m, c_mc, 1 c_c copy
     # +1 move/copy construct each from moving the tuple
@@ -122,9 +126,10 @@ def test_move_and_copy_load_optional():
     assert c_m.copy_assignments + c_m.copy_constructions == 0
     assert c_m.move_assignments == 2
     assert c_m.move_constructions == 5
-    assert c_mc.copy_assignments + c_mc.copy_constructions == 0
-    assert c_mc.move_assignments == 2
-    assert c_mc.move_constructions == 5
+    assert c_mc.copy_assignments == 0
+    assert c_mc.copy_constructions == 2
+    assert c_mc.move_assignments == 3
+    assert c_mc.move_constructions == 4
     assert c_c.copy_assignments == 2
     assert c_c.copy_constructions == 5
     assert c_m.alive() + c_mc.alive() + c_c.alive() == 0
--- a/include/pybind11/detail/type_caster_base.h
+++ b/include/pybind11/detail/type_caster_base.h
@@ -760,8 +760,7 @@ public:
  * Determine suitable casting operator for pointer-or-lvalue-casting type casters.  The type caster
  * needs to provide `operator T*()` and `operator T&()` operators.
  *
- * If the type supports moving the value away via an `operator T&&() &&` method, it should use
- * `movable_cast_op_type` instead.
+ * If the type supports moving, it should use `movable_cast_op_type` instead.
  */
 template <typename T>
 using cast_op_type =
@@ -769,20 +768,36 @@ using cast_op_type =
         typename std::add_pointer<intrinsic_t<T>>::type,
         typename std::add_lvalue_reference<intrinsic_t<T>>::type>;
 
+// A by-value argument will use lvalue references by default (triggering copy construction)
+// except the type is move-only and thus should use move semantics.
+template <typename T> using cast_op_type_for_byvalue =
+    conditional_t<std::is_copy_constructible<intrinsic_t<T>>::value,
+        typename std::add_lvalue_reference<intrinsic_t<T>>::type,
+        typename std::add_rvalue_reference<intrinsic_t<T>>::type >;
+
 /**
- * Determine suitable casting operator for a type caster with a movable value.  Such a type caster
- * needs to provide `operator T*()`, `operator T&()`, and `operator T&&() &&`.  The latter will be
- * called in appropriate contexts where the value can be moved rather than copied.
+ * Determine suitable casting operator for movable types.
+ * While C++ decides about copy vs. move semantics based on the type of the passed argument,
+ * Python doesn't know about lvalue and rvalue references and thus cannot decide based on the
+ * passed argument. Instead, the wrapper needs to define the desired semantics of argument passing
+ * explicitly:
+ * - By specifying an rvalue reference (T&&), move semantics is enforced.
+ * - By default, lvalue references (T&) are passed, triggering copy semantics if necessary.
+ *   An exception are move-only types: These are passed via move semantics.
  *
- * These operator are automatically provided when using the PYBIND11_TYPE_CASTER macro.
+ * The corresponding caster needs to provide the following operators:
+ * `operator T*()`, `operator T&()`, and `operator T&&()`.
+ * These operators are automatically provided when using the PYBIND11_TYPE_CASTER macro.
  */
 template <typename T>
 using movable_cast_op_type =
     conditional_t<std::is_pointer<typename std::remove_reference<T>::type>::value,
         typename std::add_pointer<intrinsic_t<T>>::type,
-    conditional_t<std::is_rvalue_reference<T>::value,
-        typename std::add_rvalue_reference<intrinsic_t<T>>::type,
-        typename std::add_lvalue_reference<intrinsic_t<T>>::type>>;
+        conditional_t<std::is_rvalue_reference<T>::value,
+            typename std::add_rvalue_reference<intrinsic_t<T>>::type,
+            conditional_t<std::is_lvalue_reference<T>::value,
+                typename std::add_lvalue_reference<intrinsic_t<T>>::type,
+                cast_op_type_for_byvalue<T> > > >;
 
 // std::is_copy_constructible isn't quite enough: it lets std::vector<T> (and similar) through when
 // T is non-copyable, but code containing such a copy constructor fails to actually compile.
@@ -917,7 +932,7 @@ public:
 
     operator itype*() { return (type *) value; }
     operator itype&() { if (!value) throw reference_cast_error(); return *((itype *) value); }
-    operator itype&&() && { if (!value) throw reference_cast_error(); return std::move(*((itype *) value)); }
+    operator itype&&() { if (!value) throw reference_cast_error(); return std::move(*((itype *) value)); }
 
 protected:
     using Constructor = void *(*)(const void *);
--- a/include/pybind11/detail/type_caster_base.h
+++ b/include/pybind11/detail/type_caster_base.h
@@ -912,10 +912,12 @@ public:
             nullptr, nullptr, holder);
     }
 
-    template <typename T> using cast_op_type = detail::cast_op_type<T>;
+    // allow moving of custom types
+    template <typename T> using cast_op_type = detail::movable_cast_op_type<T>;
 
     operator itype*() { return (type *) value; }
     operator itype&() { if (!value) throw reference_cast_error(); return *((itype *) value); }
+    operator itype&&() && { if (!value) throw reference_cast_error(); return std::move(*((itype *) value)); }
 
 protected:
     using Constructor = void *(*)(const void *);
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -421,6 +421,7 @@ TEST_SUBMODULE(class_, m) {
     // test_exception_rvalue_abort
     struct PyPrintDestructor {
         PyPrintDestructor() = default;
+        PyPrintDestructor(const PyPrintDestructor&) = default;
         ~PyPrintDestructor() {
             py::print("Print from destructor");
         }
--- a/tests/test_copy_move.cpp
+++ b/tests/test_copy_move.cpp
@@ -111,10 +111,10 @@ TEST_SUBMODULE(copy_move_policies, m) {
     // test_move_and_copy_casts
     m.def("move_and_copy_casts", [](py::object o) {
         int r = 0;
-        r += py::cast<MoveOrCopyInt>(o).value; /* moves */
+        r += py::cast<MoveOrCopyInt>(o).value; /* copies */
         r += py::cast<MoveOnlyInt>(o).value; /* moves */
         r += py::cast<CopyOnlyInt>(o).value; /* copies */
-        auto m1(py::cast<MoveOrCopyInt>(o)); /* moves */
+        auto m1(py::cast<MoveOrCopyInt>(o)); /* copies */
         auto m2(py::cast<MoveOnlyInt>(o)); /* moves */
         auto m3(py::cast<CopyOnlyInt>(o)); /* copies */
         r += m1.value + m2.value + m3.value;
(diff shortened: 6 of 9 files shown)


rhaschke:
Keeping the copy-only cast operator for `copyable_holder_caster`, I was able to solve most unit test issues. However, I had to slightly modify the `ConstructorStats`, because moving is preferred over copying now. **Please check the validity of these changes before merging.**

wjakob:
Dear Robert,

just a quick heads-up: I'm super busy with a conference deadline and don't have many free clock cycles in the next month. A quick bit of feedback is that this PR (and the sub-component #2048) seem very reasonable, but I'm scared by #2046 (both in terms of the heavy modifications and all of the things that could possibly go wrong). Moving a holder from Python to C++ and making it disappear in Python-land seems like a really counter-intuitive and even dangerous operation because the language provides no semantics for anything like it. I've also added @jagerman who authored some of the parts you are modifying.

Best,
Wenzel

rhaschke:
Dear Wenzel,

thanks for your feedback. I will be busy to the end of the month as well, I just used the holidays to look into pybind11 and migrate a project of mine from boost::python.
I can fully understand your concerns regarding #2046. Moving ownership from Python to C++ is somewhat risky. However, I think, if tackled correctly, it can be made safe: The python object instance doesn't disappear, but the value pointer and its holder are _invalidated_. All that is needed, is to carefully validate them before accessing them.
Lacking support for move semantics is a serious shortcoming for a C++ to python wrapper library and it is an actual blocker for my migration process.
It would be great if we could jointly come up with a safe and clean solution for this issue. #2046 is only a very preliminary approach, exploring potential solutions. It would be great if we could discuss open issues in a skype conference or the like.

I have to admit that I am new to template meta programming and I don't yet fully understand the architecture of pybind11. Particularly, I have the following questions currently:
- Why does `cast_op<T>` choose the target type based on the corresponding caster (`make_caster<T>::cast_op_type`) instead of the actual function argument type, which should be `T`?
I understand that this allows the caster to steer the casting process. However, it also means that move-semantics is preferred over copy-semantics, as soon as `movable_cast_op_type<T>` is used instead of `cast_op_type<T>`. For example, this results in the move constructor being called for [ExampleMandA::add1](https://github.com/pybind/pybind11/pull/2047/files#diff-b965cdf8e31c93de7cef9482b7fac9d2R35) with [this PR's main modification](https://github.com/pybind/pybind11/pull/2047/files#diff-e0d8df3cdf4b37acbf5e8712fb6ce80cR912), which is probably not desired.
In my point of view, the caster should suppress undesired semantics by not providing corresponding implementations, but the casting process itself should exactly follow the function signature.
- What is the purpose of [registering instances](https://github.com/pybind/pybind11/blob/2fda9d5dd876de2fc0c530e4a375b1f8c52e6758/include/pybind11/detail/internals.h#L99)?
- In general, is there some documentation about the architectural design of pybind11?

Best, Robert

BerengerBerthoul:
Hi,
I have the same problem with creating an C++ object in python, then moving it into some other C++ object, still in python. I don't care so much about what the python object is supposed to mean after it has been moved from, what is important is that the C++ underlying object is fine. What is the status of this PR and #2046?In particular, as asked by @rhaschke, is it something pybind11 is willing to support, or discard as not well-defined? From @rhaschke, it seems that boost does it?

rhaschke:
It would be great if pybind11 would officially support this scenario which is common when wrapping modern C++ libs.
boost::python doesn't support move semantics as well, but I was able to find a workaround using _deprecated_ auto_ptrs.
My proposed solution for pybind11 works for me, but I'm - of course - not aware of all the bells and whistles.

YannickJadoul:
The problem is that "moving" or "transferring ownership" is a concept that's not really known to Python (and might not result in the best Python API?). So I guess that's the reason there's a lot of hesitation, here as well as in other projects such as Boost.Python.

However, it should (in principle) already be possible to do so, no? You could create a wrapper object (maybe even a class template), have the class/`unique_ptr` you want to move as a member, and write a custom type_caster? The custom type caster could be reasonably simple, basically just wrapping and unwrapping, I believe.

YannickJadoul:
> However, it should (in principle) already be possible to do so, no?

Apologies. Apparently that's not possible, right now, and pybind11 already chokes when a function expects a rvalue reference.

YannickJadoul:
This does work for me, though, because you cán move from non-const lvalue references:

```C++
#include <iostream>

namespace py = pybind11;


class X {
public:
        X(int x) : m_x(x) { std::cout << "X::X(" << m_x << ")" << std::endl; }
        X(const X&) = delete;
        X(X &&other) : m_x(std::move(other.m_x)) { other.m_x = -1; std::cout << "X::X(X&&) (m_x = " << m_x << ")" << std::endl; }
        X &operator=(const X&) = delete;
        X &operator=(X&&) = default;
        ~X() { std::cout << "X::~X() (m_x = " << m_x << ")" << std::endl; }

private:
        int m_x;
};

void sink(X &&x) { X(std::move(x)); };

PYBIND11_MODULE(example, m) {
        py::class_<X>(m, "X");

        m.def("source", [](int x) { return X(x); });
        m.def("sink", [](X& x) { sink(std::move(x)); });
}
```

```pycon
>>> import example
>>> x = example.source(42)
X::X(42)
X::X(X&&) (m_x = 42)
X::~X() (m_x = -1)
>>> example.sink(x)
X::X(X&&) (m_x = 42)
X::~X() (m_x = 42)
>>> x = None
X::~X() (m_x = -1)
```

I assume you could write a custom type caster checking if the `X` object is valid or not (i.e., in this case, whether it contains `-1`), and thus whether you can still call methods on it or pass it to C++ from Python.

BerengerBerthoul:
`The problem is that "moving" or "transferring ownership" is a concept that's not really known to Python`
Honestly, I don't really get the problem. Ok, python programmers should not need to know the intricacies of move semantics. But python functions modifying their arguments do exist and are not especially non-pythonic. So here we are in presence of a function modifying its arguments so that the argument is not usable anymore because it has been consumed. I am sure that this pattern arises as soon as you need to transfer a ressouce.
To sum it up : I don't think it is hard to understand that an object is not valid anymore because it has been transfered somewhere else.

BerengerBerthoul:
Maybe on a more controversial note: there is always many wrong uses of a feature. But if they are good use cases (and I think there are, if you guys are not conviced I am willing to give more details), the library should support them. If not, then the user is left to code his or her own half-baked, non-standard workaround.

YannickJadoul:
> But python functions modifying their arguments do exist and are not especially non-pythonic.

True, yes, but there's a difference between changing an object and invalidating it? That's not exactly common in Python, no, an object that entirely invalid/untouchable?

Another issue is that Python's variables work differently to C++. Every Python variable is basically already a shared pointer (with reference counting). So by "invalidating" one variable, you're also invalidating all other variables pointing to this same object. Yes, not insurmountable to explain, especially if it's can be checked and a nice error can be shown, but there's really a few places where C++'s memory management just doesn't fit with Python and vice versa :-/

> Maybe on a more controversial note: there is always many wrong uses of a feature. But if they are good use cases (and I think there are, if you guys are not conviced I am willing to give more details), the library should support them. If not, then the user is left to code his or her own half-baked, non-standard workaround.

I might agree here, though encouraging good design by default is still a good plan and will take away questions ;-) I'm not saying you're wrong, btw, or that this can never be part of pybind11; merely explaining why it's been a tricky thing to have in pybind11 (and one where a custom solution might be advisable, actually, since it'll often depend on the actual application on how to handle this?).

But I'm curious about your use-case, yes. I found that often, there's a nice workaround that results in a more "pythonic" API.

Btw, you're definitely not the first ones to come up with this problem. I know that for example @EricCousineau-TRI has played around with this before (see e.g., #1146, I believe? Lots of further discussion there, almost ancient history by now, for pybind11's lifetime).

EricCousineau-TRI:
> [...] I know that for example @EricCousineau-TRI has played around with this before (see e.g., #1146, I believe? Lots of further discussion there, almost ancient history by now, for pybind11's lifetime).

FWIW We're still using a flavor of #1146 in our fork:
https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/README_DRAKE.md

I can't say it was the worst decision to allow moving `unique_ptr<T>` from Python to C++, but I can't say it's the best either.
If you care to see some of the nuances, see:
https://github.com/RobotLocomotion/drake/issues/14016
https://github.com/RobotLocomotion/drake/issues/13058
https://github.com/RobotLocomotion/drake/blob/387c0f9a52e6bc7ceafe067d6e7f1c42fba4e650/bindings/pydrake/common/value_pybind.h#L73-L77

>  [...] Ok, python programmers should not need to know the intricacies of move semantics. [...]

For the case of `unique_ptr`, at least our implementation does - we have to (kind of poorly) tell users that "hey, Python doesn't own this, so Python can't hand it over to C++".

EricCousineau-TRI:
To the point of this PR, though, it seems pretty straightforward at first blush.

The net gain I see here is that users don't have to write odd wrappers to bind functions with rvalues; the fallout of move semantics does seem like it'd be the downstream dev's responsibility, which IMO are foot-guns that are still available in nominal pybind? (e.g. `.def("delete", [](T* x) { delete x; })`).

I'd be happy to review (albeit slowly), and see if we can try to create simulated edge cases that we could encode in unittests (e.g. if there ever is a point where a Python program is affected by move semantics).

(I do think that we should de-scope supporting anything like `unique_ptr` in this PR, though)

EricCousineau-TRI:
Oh, upon reading #2040, it looks like the goal *is* to handle `unqiue_ptr`. On that point, yeah, I'd say this may be a more tad complex. Will comment in that issue to try and see some more deets.

BerengerBerthoul:
First, I am very happy to see that people are responding quickly to the comments, it's good to have an active community!

I actually missed this line from  @YannickJadoul 
`m.def("sink", [](X& x) { sink(std::move(x)); });`
I think it solves the problem in my case. @rhaschke isn't it enought for your use case?

> True, yes, but there's a difference between changing an object and invalidating it? That's not exactly common in Python, no, an object that entirely invalid/untouchable?
Well actually, a moved-from object is not invalid but in a "valid but unspecified state". So you can call any method that does not requires an invariant. E.g. you can call `size()` on a moved-from vector : the result is unspecified but this is not "undefined behavior". On the other hand, it seems practical to just think of the object as invalid. Just to say that the line between invalid and modified is not that well defined.

> you're also invalidating all other variables pointing to this same object
Yes, but same thing if you are resizing an array to 0, this may be surprising to the other variables. I think this is conceptually the same usual kind of behavior

> But I'm curious about your use-case, yes. I found that often, there's a nice workaround that results in a more "pythonic" API.
My use case is the following : I have CFD problem where I am building the configuration step by step in Python. The classes describing the configuration itself are in C++. So let's say I want to build the initial solution fields inside a FlowSolution object, and then transfer the FlowSolution to the Configuration object. Then I want to use a move since the object is heavy, and I don't plan to use it anymore (if it were to be the case, I could call a getter on Configuration)

EricCousineau-TRI:
@BerengerBerthoul Since Yannick's workaround works for you, it seems like this PR would valuable to you (you're not expecting `unique_ptr` to work in this instance). In that light, I'd be happy for this PR to move forward as-is (no `unique_ptr` stuff).

@rhaschke Would you be up for moving this forward w/o anything explicitly for `unique_ptr`?
If you don't have time, I'm happy to move it forward.

rwgk:
> So I guess that's the reason there's a lot of hesitation, here as well as in other projects such as Boost.Python.

Move semantics didn't exist when Boost.Python was written (2000-2003 timeframe).

YannickJadoul:
> Move semantics didn't exist when Boost.Python was written (2000-2003 timeframe).

Right, of course! Thanks for that reminder. So yes, maybe this can be a project towards 2.7.0? Maybe we can make this safer with some opt-in flag, or so?

rhaschke:
@EricCousineau-TRI, I would be happy to move this forward, although, eventually, I would need support for moving unique_ptrs.
I think the key question is how to tell the user that an instance became invalid...
@YannickJadoul: What do you think is missing to progress with this PR?

EricCousineau-TRI:
@rhaschke For me on this PR, I think if you edit the overview to note that this is an independent, strict improvement towards #2040 (and thus narrow the scope of this specific PR), and rebase, I'd be happy to review it.

Will let Yannick chime in as well.

EricCousineau-TRI:
>  I think the key question is how to tell the user that an instance became invalid... [...]

My point about scoping is that this is irrelevant to `pybind11`, other than just adding some minor caution tape saying "If you rely on move semantics in your publicly exposed Python-C++ API, you should take care in your own API that invalidated objects have clear error messages for users (e.g. do not cause crazy segfaults)."

YannickJadoul:
> @YannickJadoul: What do you think is missing to progress with this PR?

For me, to decently review this, I'd need a few hours (after 2.6.0 is out, when I have some time) to fully get all the details of the type casters again. E.g., this first change of `detail::cast_op_type<T>;` into `detail::movable_cast_op_type<T>;` feels ... well, I don't feel confident to say a lot about this without studying and reminding myself about the whole casting mechanism again.
There are a few other problems with the casters (e.g., #2245/#2303/#2527, or #2336), so casters might need a redesign once we start aiming at 3.0, anyway.

Apart from that, stealing objects from Python is rather tricky (given the way Python treats objects), so once there is something, I would like some input from @wjakob on this. Of course, once you allow objects to be stolen from pybind11, maybe there isn't a big difference anymore with stealing the holders. At the same time, stealing objects is already possible, since you can get a reference to the object? (not sure if that's an argument for or against this PR :-) )

I can see this can all be useful for power users and a bit of a gap between C++ and Python, but at the same time, I still don't believe it's a great Python API, so making this opt-in and slightly harder than necessary might push users towards creating more Python-like APIs. Any thoughts on that?

rwgk:
> I can see this can all be useful for power users and a bit of a gap between C++ and Python

Hi @YannickJadoul , I think the motivation behind this PR, and #2046, goes much deeper. See also my question #2583.

> so making this opt-in and slightly harder than necessary might push users towards creating more Python-like APIs. Any thoughts on that?

I actually started out wanting to eliminate the `unique_ptr` passing feature from PyCLIF. But when I ran into the concrete use case from which #2583 is distilled, I developed serious doubts. Short term, it would probably take me at least a week or two to change the existing Google code that needs the `unique_ptr` passing, and I expect unhappy customers immediately, because the Python wrappers force them to deviate from modern C++ best practices. Long-term, I expect a steady trickle of new unhappy customers, especially if adding Python bindings is an afterthought, but forces them to rethink their original C++ design.

About disadvantages, I can see two: 1. an extra `if` to check if the held object was invalidated (runtime hit, although my guess is it's in the noise). 2. it could be difficult to debug where in a process the held object was invalidated, in case that's unexpected.

If I balance those disadvantages with the advantage of enabling wrapping of modern C++ code, I come out strongly on the "it's worth it" side, but I also think your "harder than necessary" idea is a good one. Maybe it could even be leveraged to address the 2. disadvantage above, via something like
```
function_expecting_unique_ptr(obj)  # easy, but ...
function_expecting_unique_ptr(obj.disown())  # harder than necessary, but flags code that gives up ownership
```
In the C++ bindings we'd need some mechanism to add the special `.disown()` method, and the casters would have to reject a plain `obj` but accept `obj.disown()`.

rhaschke:
I rebased onto recent master. 
I agree with @YannickJadoul's doubts expressed in https://github.com/pybind/pybind11/pull/2047#issuecomment-712325915: Changing `detail::cast_op_type<T>` into `detail::movable_cast_op_type<T>` wasn't a good idea as this makes **moving preferred over copying** as indicated in https://github.com/pybind/pybind11/pull/2047#issuecomment-569857805 already. I added a new unit test illustrating the problem: https://github.com/pybind/pybind11/pull/2047/commits/60ef62c211a42f6dd947e0f8e1fb32477c4b09af

I think the casters should be reworked to consider the actual argument type (pointer, lvalue reference, rvalue reference, constness?). As the argument casters only know about the _intrinsic_ type, this detailed information is already lost, although the casting behavior should differ for these type flavors. Playing a little with the casting operators [here](https://github.com/pybind/pybind11/pull/2047/commits/29713a9bc8eccd18343b8b09115765ae14269cad#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R921-R923), I couldn't find a working solution.

YannickJadoul:
@rhaschke, yeah, it seems to be one of the things in pybind11 that grew a bit out of control :-( There hás to be some way to fix it, though? Worst case, it'll will involve some compile-time macro opting-in to the new approach.

For the record: I'm not per se against adding/enabling these moves; it would make pybind11 more complete, in some way. But I'm just voicing the reasons why this is currently not really supported and why we should be really careful here. I do very much get @wjakob's carefulness/hesitance, when it comes to this.

rhaschke:
I completely reworked this PR to address the remaining issue that moving was always preferred over copying regardless of the actual function signature. The main reason is that `cast_op(caster&&)` is enforcing move semantics by always adding an rvalue reference as soon as the caster itself is an rvalue reference:
https://github.com/pybind/pybind11/blob/b72cebeb228e839caa13fa0269bd15676e130def/include/pybind11/cast.h#L953-L957
The alternative `cast_op(caster&)` implementation that _is_ selecting the move/copy semantics based on the actual function signature, was essentially disabled in #851 by [enforcing rvalue references for all argument casters](https://github.com/pybind/pybind11/pull/851/files#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R1520).
This PR [reenables the lvalue-reference `cast_op()` for calling C++ functions](https://github.com/pybind/pybind11/pull/2047/commits/a95b70a4e32fe138f1c69f400e404ef273a48b00#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R2021), thus allowing to chose between move and copy semantics based on the function argument type:
* `f(T&&)` triggers move semantics for `T` (if available)
* `f(T)` triggers copy semantics for `T` (and [falls back to move semantics if copying is not supported](https://github.com/pybind/pybind11/pull/2047/files#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R787))
* `f(T&)` uses lvalue-references as before.

Furthermore, it was important to actually enable moving if move semantics is chosen. The original implementation [returning an rvalue reference to the internal `value` of the caster](https://github.com/pybind/pybind11/blob/813d7e868746f106f9ebb98decae55219d97a722/include/pybind11/cast.h#L549-L550) wasn't actually performing the move as can be seen from the [failing unit tests](https://github.com/pybind/pybind11/pull/2047/files?file-filters%5B%5D=.py#diff-08f3f7ee359b10e22b190ffcc2f139f969389f9e2c337a104030dacbbcfefe59R85) before 0412b945cfe5c9b88b6696a0e32dae5525c17144. Instead of casting to an rvalue reference, I'm now [casting to `T`](https://github.com/pybind/pybind11/pull/2047/commits/0412b945cfe5c9b88b6696a0e32dae5525c17144#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R788) and [explicitly performing the move](https://github.com/pybind/pybind11/pull/2047/commits/0412b945cfe5c9b88b6696a0e32dae5525c17144#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74R925). The same technique was already used to implement move semantics for `tuple_caster`:
https://github.com/pybind/pybind11/blob/b72cebeb228e839caa13fa0269bd15676e130def/include/pybind11/cast.h#L1438-L1446

As before, due to preferring copy over move semantics for plain `T` arguments, I had to [adapt some `ConstructorStats` in the unit tests](https://github.com/pybind/pybind11/pull/2047/commits/9700bcabf30cecdc97812b8ad05e2590245e2ff8). There are some other adaptions to the unit tests as well, which I am not yet sure about. Furthermore, there are some failing CI tests... So there is still some work to do.

rhaschke:
I shortly looked into a few failing tests. One issue seems to be a discrepancy between gcc and clang: While my local clang was succeeding, gcc is failing on the `itype()` cast for abstract types:
```
pybind11/cast.h:925:5: error: invalid abstract return type ‘test_submodule_callbacks(pybind11::module_&)::AbstractBase’
  925 |     operator itype() { if (!value) throw reference_cast_error(); return itype(std::move(*((itype *) value))); }
      |     ^~~~~~~~
pybind11/tests/test_callbacks.cpp:120:11: note:   because the following virtual functions are pure within ‘test_submodule_callbacks(pybind11::module_&)::AbstractBase’:
  120 |     class AbstractBase {
      |           ^~~~~~~~~~~~
pybind11/tests/test_callbacks.cpp:123:30: note: 	‘virtual unsigned int test_submodule_callbacks(pybind11::module_&)::AbstractBase::func()’
  123 |         virtual unsigned int func() = 0;
```
I have no idea how to handle this. Any help is highly appreciated.

YannickJadoul:
@rhaschke Can you perhaps `enable_if` this part with `std::is_abstract`?

rhaschke:
> @rhaschke Can you perhaps `enable_if` this part with `std::is_abstract`?

Yes, I tried. Didn't work for me. But I'm not an expert on the SFINAE syntax. Would be great if somebody more experienced could have a look.

> Note: I haven't actually looked at the code; so I don't know why it worked before and is failing now.

The reason is that I changed the cast operator from `operator T&&()` to `operator T()` to enable actual moving as explained in 2. in https://github.com/pybind/pybind11/pull/2047#issuecomment-724256796.

rhaschke:
I added a sandbox branch (#2647) comprising a simple sandbox program to understand the behavior of move/copy semantics in native C++. This revealed, that just providing a cast `operator T&&()` is not sufficient to actually trigger the move constructor.
Compare the behavior of the executable `tests/main_rhaschke` using either [`operator T&&()`](https://github.com/rhaschke/pybind11/blob/72bed417e8f6041661e88a47e2c3553311694b5f/tests/main_rhaschke.cpp) and [`operator T()`](https://github.com/rhaschke/pybind11/blob/sandbox-2047/tests/main_rhaschke.cpp). Only in the latter case, move and copy constructors are called as expected, thus mimicking the behavior of C++.

rhaschke:
@YannickJadoul: I think I fixed the SFINAE syntax and https://github.com/pybind/pybind11/pull/2047/commits/2cb13e6db54f3b982f214bc145da4510b74a6e59 compiles on gcc. However, it fails on clang. [Removing the protection](https://github.com/pybind/pybind11/pull/2047/commits/cadc0c1cf1bfca364472abb66cc250200612afdc) works on clang, but fails on gcc (https://github.com/pybind/pybind11/pull/2047/commits/cadc0c1cf1bfca364472abb66cc250200612afdc). Do you have any advice how to proceed?

EricCousineau-TRI:
IMO, this still seems like a strict improvement in its current scope.

@rhaschke Any chance you have time to update this and merge it in?

include/pybind11/pybind11.h
```diff
@@ -90,7 +90,7 @@ class cpp_function : public function {
     /// but with an added `&`.
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     cpp_function(Return (Class::*f)(Arg...)&, const Extra&... extra) {
-        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(args...); },
+        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(std::forward<Arg>(args)...); },
```
bstaletic, line 93: Shouldn't this be `Arg&&... args`, just like in `functional.h`?

include/pybind11/pybind11.h
```diff
@@ -106,7 +106,7 @@ class cpp_function : public function {
     /// but with an added `&`.
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     cpp_function(Return (Class::*f)(Arg...) const&, const Extra&... extra) {
-        initialize([f](const Class *c, Arg... args) -> Return { return (c->*f)(args...); },
+        initialize([f](const Class *c, Arg... args) -> Return { return (c->*f)(std::forward<Arg>(args)...); },
```
bstaletic, line 109: Same here. `Arg&&... args`?

tests/test_copy_move.py
```diff
@@ -48,47 +46,75 @@ def test_move_and_copy_loads():
...
+    # Direct constructions: 2 c_m moves, 2 c_mc copies, 1 c_c copy
     # Extra moves/copies when moving pairs/tuples: 3 c_m, 3 c_mc, 2 c_c
     assert m.move_copy_nested((1, ((2, 3, (4,)), 5))) == 15
 
     assert c_m.copy_assignments + c_m.copy_constructions == 0
     assert c_m.move_assignments == 6
-    assert c_m.move_constructions == 9
-    assert c_mc.copy_assignments + c_mc.copy_constructions == 0
-    assert c_mc.move_assignments == 5
+    assert c_m.move_constructions == 14
```
bstaletic, line 65: There's 55% more moves. Should this be a concern? Moves are usually quite cheap, but 55% still seems like a lot. Is this even representative of "real" code?

include/pybind11/eigen.h
```diff
@@ -346,7 +346,7 @@ struct type_caster<Type, enable_if_t<is_eigen_dense_plain<Type>::value>> {
 
     operator Type*() { return &value; }
     operator Type&() { return value; }
-    operator Type&&() && { return std::move(value); }
+    operator Type() { return std::move(value); }
```
bstaletic, line 349: Do we want to replace `operator Type&&` with `operator Type` or have both? `PYBIND11_TYPE_CASTER` defines both.

include/pybind11/cast.h
```diff
@@ -772,20 +772,42 @@ using cast_op_type =
...
         typename std::add_lvalue_reference<intrinsic_t<T>>::type>;
 
+template <typename T> using cast_op_type_for_moving =
+    conditional_t<std::is_move_constructible<intrinsic_t<T>>::value,
+        // return an rvalue ref via T(std::move(value)) which triggers a move construction
+        intrinsic_t<T>,
+        // this doesn't work for abstract types: cast these to an rvalue reference
+        typename std::add_rvalue_reference<intrinsic_t<T>>::type>;
+
+// A plain argument type will use lvalue references (triggering copy semantics)
```
bstaletic, line 782: By "plain", you're talking about "by value" arguments?

bstaletic (commented):
> Further, it was important to actually enable moving if move semantics is chosen.

That sounds like a good idea.

>> Can you perhaps enable_if this part with std::is_abstract?

> Yes, I tried. Didn't work for me. But I'm not an expert on the SFINAE syntax. Would be great if somebody more experienced could have a look.

I can take a look, but... where? Has it been fixed already?

&nbsp;

I also wanted to take a look at the CI failures, but you have to love Microsoft. The logs have expired...

include/pybind11/pybind11.h
rhaschke: I agree and I actually tried to apply the perfect-forwarding pattern in these 4 class-method forwarders as well, but it failed terribly.

include/pybind11/cast.h
rhaschke: Exactly.

include/pybind11/cast.h
```diff
@@ -2007,7 +2041,7 @@ class argument_loader {
 
     template <typename Return, typename Func, size_t... Is, typename Guard>
     Return call_impl(Func &&f, index_sequence<Is...>, Guard &&) && {
-        return std::forward<Func>(f)(cast_op<Args>(std::move(std::get<Is>(argcasters)))...);
+        return std::forward<Func>(f)(cast_op<Args>(std::get<Is>(argcasters))...);
```
rhaschke, line 2044: This change is crucial: Not moving the argcasters here, avoids adding an rvalue reference to all cast_op targets.

include/pybind11/eigen.h
rhaschke: You are right. We should keep both, one for move-forwarding and one for move construction.

include/pybind11/cast.h
```diff
@@ -915,10 +937,15 @@ template <typename type> class type_caster_base : public type_caster_generic {
...
-    template <typename T> using cast_op_type = detail::cast_op_type<T>;
+    // allow moving of custom types
+    template <typename T> using cast_op_type = detail::movable_cast_op_type<T>;
 
     operator itype*() { return (type *) value; }
     operator itype&() { if (!value) throw reference_cast_error(); return *((itype *) value); }
+    // move caster, actually triggering a move construction from value
+    operator itype() { if (!value) throw reference_cast_error(); return std::move(*((itype *) value)); }
+    // fallback move caster, just passing rvalue reference as is
+    operator itype&&() { if (!value) throw reference_cast_error(); return std::move(*((itype *) value)); }
```
bstaletic, line 948: This overload set seems weird. With a maybe too high WTF per line of code ratio. I'm not saying it's wrong, but I'm sure everyone would benefit from some clarifications. For context:


> - `f(T&&)` triggers move semantics for `T` (if available)
> - `f(T)` triggers copy semantics for `T` (and falls back to move semantics if copying is not supported)
> - `f(T&)` uses lvalue-references as before.


Here are the questions:

1. What's the difference between `operator itype` and `operator itype&&`? Both look the same.
2. If we want `f(T)` to be a copy (which seems reasonable), why is `operator itype` using `std::move()`?
3. Should the operators be ref-qualified, as in `operator itype&&() &&`?

include/pybind11/cast.h
```diff
@@ -915,10 +937,15 @@ template <typename type> class type_caster_base : public type_caster_generic {
             nullptr, nullptr, holder);
     }
 
-    template <typename T> using cast_op_type = detail::cast_op_type<T>;
+    // allow moving of custom types
+    template <typename T> using cast_op_type = detail::movable_cast_op_type<T>;
 
     operator itype*() { return (type *) value; }
```
bstaletic, line 943: Is this a typo? Shouldn't this be `return (itype*) value;`?

bstaletic:
This comiles on my laptop: https://github.com/bstaletic/pybind11/commit/c2e7c475d9dad8b1fa6bc988610b83de77e881c1

include/pybind11/pybind11.h
rhaschke: I think, improving the perfect forwarding should be another PR. I reverted the corresponding commit as it isn't relevant for this PR.

tests/test_copy_move.py
rhaschke: I was able to improve this via f130cddba71f88450ce0275436415b980626c8ec.

include/pybind11/cast.h
rhaschke: Fixed.

include/pybind11/cast.h
rhaschke: Probably. I fixed the typo without getting any functional change.

rhaschke:
> This comiles on my laptop: [bstaletic@c2e7c47](https://github.com/bstaletic/pybind11/commit/c2e7c475d9dad8b1fa6bc988610b83de77e881c1)

Great. Thanks! I will apply those changes asap.

include/pybind11/eigen.h
```diff
@@ -346,7 +346,8 @@ struct type_caster<Type, enable_if_t<is_eigen_dense_plain<Type>::value>> {
 
     operator Type*() { return &value; }
     operator Type&() { return value; }
-    operator Type&&() && { return std::move(value); }
+    operator Type&&() { return std::move(value); }
+    operator Type() { return std::move(value); }
```
bstaletic, line 350: Does `operator Type()` here need the SFINAE treatement?

bstaletic (commented):
Looks pretty good to me. There's one thing I don't understand though.

https://github.com/pybind/pybind11/pull/2047#discussion_r581065535

`T t1 = t2;` doesn't move. Our `operator type()` does call `std::move`. Won't that cause copiable types to be moved?

include/pybind11/cast.h
rhaschke: > 1. What's the difference between `operator itype` and `operator itype&&`? Both look the same.

Both variants are needed to actually trigger a move construction in both of the following cases:
- for a function `foo(T&&)`, explicitly trigger a move construction via `operator itype`
  Just passing the rvalue reference, wouldn't trigger a move construction!
- for a function `foo(T)` with a non-copy-constructible type, implicitly trigger a move construction
  via `operator itype&&`

Introducing the extra cast operator is just a technical trick to actually achieve the correct move-construction behavior.
I also tried [including this logic into `cast_op<T>()`](https://github.com/rhaschke/pybind11/blob/36ba36e892aeac3adb48e59d5edc313be2521713/include/pybind11/cast.h#L976-L981) instead. While this approach will remove the need to define the redundant `operator itype()`, my naive implementation returns a reference to a local temporary. Alternatively changing the return type based on the two cases, will make `caster::cast_op_type` inconsistent to what `cast_op` actually returns. Maybe, you, @bstaletic, have a better idea to implement it there.

> 2. If we want `f(T)` to be a copy (which seems reasonable), why is `operator itype` using `std::move()`?

In C++, the `f(T)` uses a copy constructor for lvalue references and a move constructor for rvalue references.
Returning `std::move()` from `operator itype()` essentially yields `itype(std::move(arg))`, which realizes the move construction.

> 3. Should the operators be ref-qualified, as in `operator itype&&() &&`?

I think this extra rvalue-qualifier was used in the past to restrict the move operation to situations where the caster itself is an rvalue. However, due to [`cast_op<T>(caster&&)`](https://github.com/pybind/pybind11/pull/2047/files#diff-b8e97eea5f84a84248db1bc72afaccfd8ba28cad023917d52bdc6bbc3669cd74L953-L957), the cast_op_type is always an rvalue reference if the caster is an rvalue reference. There is no need to require an rvalue caster also for the actual casting operation.

I improved the source code comments, see f93d0f01f25472dd40f542d450f87fe518ae952f.

include/pybind11/eigen.h
rhaschke: The Eigen types always are move-constructible. Hence, no need for protection, I think.

rhaschke:
> This comiles on my laptop: [bstaletic@c2e7c47](https://github.com/bstaletic/pybind11/commit/c2e7c475d9dad8b1fa6bc988610b83de77e881c1)

This commit doesn't compile for me, neither with gcc nor with clang :disappointed:

rhaschke:
I rebased onto master and reworked the PR thoroughly - once again. Particularly, I gave up the idea to automatically trigger a move construction when passing into an rvalue argument. Now, both lvalue and rvalue arguments don't trigger any constructor, they just pass the reference (as in C++). However, in contrast to the previous `movable_cast_op_type`, which preferred moving over copying, it is the other way around now: copying is preferred over moving. Thus, moving needs to be explicitly requested by the wrapping API, specifying an rvalue argument: `foo(T&& arg)`.

Actually, I suggest to use this new `movable_cast_op_type` as the default as it allows moving as long as the type allows and the wrapper asks for it. To forbid moving, the old default should be named `copyonly_cast_op_type`. Additionally, we could have a `move_preferred_cast_op_type`. @YannickJadoul, @wjakob: what do you think about this?

rhaschke:
I'm not sure why the clang 3.6/3.7 builds [fail with an error referring to gcc includes](https://github.com/pybind/pybind11/pull/2047/checks?check_run_id=2045964797).
The [failure of CentOS7](https://github.com/pybind/pybind11/pull/2047/checks?check_run_id=2045965670) seems to be unrelated: NVIDIA HPC SDK cannot be installed.

rwgk:
> The [failure of CentOS7](https://github.com/pybind/pybind11/pull/2047/checks?check_run_id=2045965670) seems to be unrelated: NVIDIA HPC SDK cannot be installed.

Yes, that's a known flake, I saw it the first time ~2-3 weeks ago. Seems to be intermittently failing download attempts. If you try again it'll probably work.

rwgk:
> I'm not sure why the clang 3.6/3.7 builds [fail with an error referring to gcc includes](https://github.com/pybind/pybind11/pull/2047/checks?check_run_id=2045964797).

I'm pretty sure that gcc error is very easy to work around, I ran into it several times, too.
Here is an example workaround:

https://github.com/pybind/pybind11/blob/dbcb6d3a9f57e379069e2b157bc76fc6fca18d1b/tests/test_class_sh_with_alias.cpp#L17

You may only need some of the 4 lines there, but I just put in all, to not have to go through a time-consuming back-and-forth with the CI.

rwgk:
> Actually, I suggest to use this new movable_cast_op_type as the default as it allows moving as long as the type allows and the wrapper asks for it. To forbid moving, the old default should be named copyonly_cast_op_type. Additionally, we could have a move_preferred_cast_op_type.

I think ideally one `cast_op_type` is best. Having multiple will create confusion for sure. What you currently call `movable_cast_op_type` seems just right, but please correct me in case my understanding is wrong:

* If the wrapped function arg does not have `&&`, it'll use copying, if it does, it'll move. — Sounds great. Looks like "What You See Is What You Get". (It was unintuitive that it didn't work like that before IMO.) — Question: does the Python refcount play a role here?
* IIUC if the original C++ function arg has a `&&`, but I don't want to move the object, I can hide the `&&` behind a lambda function. — That's very obvious and intuitive, too.

It would be nice to develop a new "Understanding copy and move semantics for wrapped function arguments" section along with this PR, that explains the new behavior in a systematic way (for people new to pybind11), and contrasts it to the old behavior (for people who have existing wrappers), with tips what to watch out for when upgrading.

rhaschke:
> If the wrapped function arg does not have `&&`, it'll use copying, if it does, it'll move. — Sounds great. Looks like "What You See Is What You Get". (It was unintuitive that it didn't work like that before IMO.)

Not exactly. If the wrapped function arg is an lvalue (&) or rvalue (&&) reference, the argument is just passed as a reference. Thus, neither copying nor moving is performed (yet). Copying or moving only occurs if the argument is declared by-value. 

> Question: does the Python refcount play a role here?

No, unfortunately not. I thought about using the refcount to automatically decide between lvalue and rvalue references. That would be actually the most elegant solution, because unique python objects (with refcount = 1) could be automatically passed as rvalue references. However, this still wouldn't allow to explicitly move an argument.
But, more importantly, I didn't find a technical solution for this, because the `cast_op` would need to decide its output type dynamically at runtime based on the refcount, which is simply not possible.

> If the original C++ function arg has a `&&`, but I don't want to move the object, I can hide the `&&` behind a lambda function. — That's very obvious and intuitive, too.

I think trying to do so, will result in a compile error, because the lambda function cannot pass an lvalue reference to the rvalue reference of the underlying function. There you need to use `std::move`, which eventually yields a move construction again...

rwgk:
Looks like I'm still miles off with my understanding. Sorry.

I did experiment myself a little bit while working on the smart_holder_type_caster, but apparently I misunderstood a lot what actually happens. FWIW, I started out with this (which I got mostly from PR #2705):

https://github.com/pybind/pybind11/pull/2834/files#diff-f854d6c77a3f645bec46a4b1161cc3f0292d62a86bc322a526bbd95d77e1a8d9R97

I got pretty far with this, but eventually I wound up removing the first two operators, simply driven by wanting to be compatible with `type_caster_base`, and avoiding errors with some compilers:

* I first had to take out `operator T&&() &&`: https://github.com/pybind/pybind11/pull/2672/commits/50e81bce070877def6098662ca89ce3c07473a6e
* Then I also removed `operator T()`, to fix a gcc error: https://github.com/pybind/pybind11/pull/2672/commits/bfaec80519b664496d2c11192184a067e0be1095

I'm not sure this is helping here directly, but I think it would be great to have something similar to `test_type_caster_bare_interface` that systematically and minimalistically exercises all "return" and "pass" situations, with comments what happens; a unit test that doubles as a demo / documentation of all the nuances.

rhaschke:
> Looks like I'm still miles off with my understanding. Sorry.

Vice versa, I couldn't really understand the aim of your referenced PRs. Shall we have a short video chat to clarify?

> It would be great to have something similar to `test_type_caster_bare_interface` that systematically and minimalistically exercises all "return" and "pass" situations.

I agree and I think, that's what was done in tests/test_copy_move.py : `test_move_and_copy_casts()` and `test_move_and_copy_loads()`. I don't see how the load tests in #2834 actually test for copying/moving.

zwimer:
Being able to use forwarding references of custom classes would be a game changer :)

rwgk:
Off the top of my head: I think we ran into issues with trampolines that are very tricky to resolve. Possibly, if the desired feature had been taken into account early on, it may have been easier to accommodate it, but now it's "too late".

Unless someone high-powered & with a good block of time jumps on this, I don't think this will get done anytime soon.

My own assessment: a nice to have. Certainly not a game changer.

rhaschke:
This feature is provided by the `smart_holder` branch. Closing this issue.

rwgk:
> This feature is provided by the `smart_holder` branch. Closing this issue.

The latest smart_holder branch (as of #5257) is using `type_caster_base` pretty much as-is. That's best seen by looking at the equivalent diff against master:

https://github.com/pybind/pybind11/pull/5213/files#diff-95c66a8bb93a0f1ab41557995b5b9ebd0f743f0d2852573863f328e18f38d193

pybind11/detail/smart_holder_type_casters.h is gone completely.

Is it still true now that "this feature is provided by the `smart_holder` branch"?

rhaschke:
> Is it still true now that "this feature is provided by the `smart_holder` branch"?

Maybe you are right. The key feature of the smart_holder branch I'm using are rvalue holder arguments.

PR #2048, opened 2019-12-31
Allow perfect forwarding of method args

### LINKED ISSUE #1694: Rvalue reference parameters work in constructors, but not in other functions
```cpp
struct example {
    example(std::string&&) {}
    void name(std::string&&) {}
};

pybind11::class_<example>(m, "example")
    .def(pybind11::init<std::string&&>()) // this works
    .def<void(example::*)(std::string&&)>("name", &example::name) // this doesn't
;
```
Error message:

/usr/local/include/pybind11/pybind11.h:72:74: error: rvalue reference to type 'basic_string<...>' cannot bind to lvalue of type 'basic_string<...>'
        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(args...); },

rhaschke:
This fixes #1694. I wanted to contribute something after filing issues #2046 and #2047.
The fix just adds perfect forwarding for method args to support both, lvalue and rvalue arguments.
The [unit test required a minor adaptation](https://github.com/pybind/pybind11/commit/9df27bd983b55a7f39951da2863f5cba1823f79b#diff-290c8b2315630c8705b7373cbf1b8438L61-R62) as [`ExampleMandA::add1()`](https://github.com/pybind/pybind11/blob/2fda9d5dd876de2fc0c530e4a375b1f8c52e6758/tests/test_methods_and_attributes.cpp#L34) now copies and moves (instead of copying twice).

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -75,14 +75,14 @@ public:
     /// Construct a cpp_function from a class method (non-const)
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     cpp_function(Return (Class::*f)(Arg...), const Extra&... extra) {
-        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(args...); },
+        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(std::forward<Arg>(args)...); },
                    (Return (*) (Class *, Arg...)) nullptr, extra...);
     }
 
     /// Construct a cpp_function from a class method (const)
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     cpp_function(Return (Class::*f)(Arg...) const, const Extra&... extra) {
-        initialize([f](const Class *c, Arg... args) -> Return { return (c->*f)(args...); },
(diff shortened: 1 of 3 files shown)


wjakob:
Looks like a simple enough change. I'm curious though what kinds of types will actually be affected by this? `std::string`? What about STL types like `std::vector<..>`? Does this commit by itself lead to any changed behavior wrt. custom types bound using `py::class_<>`?

wjakob:
This is a smaller component of #2047, correct?

rhaschke:
Yes, #2048 building on top of this one. However, I think they are independent. Not sure, though.

rhaschke:
> I'm curious what kinds of types will actually be affected by this?

This PR just enables perfect forwarding. It's not related to particular types, but to the type specifiers (e.g. & or &&) in method arguments.

> Does this commit by itself lead to any changed behavior wrt. custom types bound using py::class_<>?

No, the PR just enables rvalue arguments in methods. If the custom type supports rvalue reference casting (only with #2048), it will be possible to forward them to such method arguments.

bstaletic:
@wjakob For example, this would fix #1694

wjakob:
Seems quite safe, let's merge it and see if anyone complains.

rhaschke:
Great to see some progress here. Thanks for merging.

PR #2052, opened 2020-01-02
Inconsistent holder types cause crash

rhaschke:
Currently, pybind11 doesn't check for compatibility of holder types. A holder caster accesses the corresponding `value_and_holder` slot without any type checking, which can cause serious crashes if the actually requested holder type doesn't match the declared one.
The first commit of this PR illustrates the issue as a new test case and the second commit suggests a potential solution: namely storing the `typeinfo` of the declared holder and comparing it to the requested one.
The failing tests emphasize another issue: there is no dynamic type casting for holders yet.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -231,6 +231,7 @@ struct type_record {
 
     // Pointer to RTTI type_info data structure
     const std::type_info *type = nullptr;
+    const std::type_info *holder_type = nullptr;
 
     /// How large is the underlying C++ type?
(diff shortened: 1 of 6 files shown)


rhaschke:
Alternatively, one could validate the compatibility of holder types in the constructor of the caster.

rhaschke:
Rebased onto master.

rhaschke:
Closing in favor of #1161, which checks only once during the definition of a new wrapper function, while my code checked for each cast.

PR #2053, opened 2020-01-02
Fix compilation with MinGW only

Baljak:
PR #1851 broke compilation with MinGW without MSYS. See https://github.com/pybind/pybind11/pull/1851#issuecomment-570244451.

This fixes it while still retaining the intent of the original PR. Since both MSYS and MYS2 define a MSYSTEM environment variable, we just need to check if it is defined or not.


kanonet:
I tested this and can confirm, that this fix does not break compiling using msys2. So its all fine from my side.
BTW it can even get shortened to `if(CMAKE_HOST_WIN32 AND NOT DEFINED ENV{MSYSTEM})` this would still work on my side. Thank you Baljak!

Baljak:
Thanks for the review @kladonet. I simplified the check as you suggested.

wjakob:
Great, thanks everyone!

PR #2057, opened 2020-01-05
Fix inconsistent && qualification in argument_loader's call functions

rhaschke:
As discussed in #2044, all `call` methods should be consistently declared as `&&` to indicate that `*this` will be invalidated by the calls due to moving the `argcasters` members.


wjakob:
Thanks!

Issue #1512, opened 2018-09-02
Float Overloads Specialize to Least Precision


ax3l:
## Issue description

When overloading for float parameters, such as:
```C++
    m.def("bar", [](float const& value) {
        std::cout << "float! " << value << std::endl;
    });
    m.def("bar", [](double const& value) {
        std::cout << "double! " << value << std::endl;
    });
    m.def("bar", [](long double const& value) {
        std::cout << "long double! " << value << std::endl;
    });
```

the corresponding call from a python `builtin.float` will always select the least precise floating point type:
```python
from cmake_example import bar

bar(1.23)
# float! 1.23

type(1.23)
# float

bar(1.23e38)
# float! 1.23e+38

type(1.23e38)
# float

bar(1.23e39)
# float! inf

type(1.23e39)
# float
```

## Reproducible example code

See above.

Contrary, in `Boost.Python` always the *most precise* floating point type was selected:
```C++
void bar1( float const& value ) {
    std::cout << "float! " << value << std::endl;
}
void bar2( double const& value ) {
    std::cout << "double! " << value << std::endl;
}
void bar3( long double const& value ) {
    std::cout << "longdouble! " << value << std::endl;
}

// ...

    def("bar", bar1);
    def("bar", bar2);
    def("bar", bar3);
```

```python
# ...

bar(1.23)
# longdouble! 1.23
```

## Proposed change

besides that making all three overloads possible for `builtin.float` but only selecting one is still fine, we should change the priorities of the selected overloads similar to the handling in `Boost.Python` to the largest/most-precise type.

for explicit control of passing `float32/64/128` we should fix the bug reported in #1224

YannickJadoul:
Python's `float`s are C++ `double`s, right? So `long double` should never be matched?

But yes, it seems a bit weird. We could consider casting a Python `float` to a C++ `float` to be a "conversion", so it would only get converted on the second pass, but how much existing code would we then break?

bstaletic:
With scalar types, pybind will always pick the first one it finds. Order matters a lot and `long doubles` are selected if encountered first.

YannickJadoul:
That's the current situation, yes. But isn't it weird that a `double` gets narrowed to `float` without any warning?

If that's fine, we can close this issue, and maybe just make a note in the docs. Any further thoughts, 2 years later, @ax3l?

bstaletic:
[...]

PR #2060, opened 2020-01-08
NumPy scalars

### LINKED ISSUE #2036: Single precision float treated as doubles, np.float32 incompatible?
(on 2.4.3)
Looks like there's a mix of a few issues in the example below:

- First, least precision floats would be picked if you don't carefully order your overloads (as @ax3l mentioned in #1512), sounds like that's something that may be done automatically
- There's no builtin float32, so both floats and doubles show up as 'float' on Python side; this would have been fine unless...
- ... passing `np.float32()` actually worked. You may expect it would have worked but because it doesn't inherit from `builtin.float` it won't. Is it worth special-casing scalar float32 case?

I do sort of understand where this is coming from, but someone who doesn't know py11 internals may assume that "well, np.float32[:] arrays work, and np.float64[:] arrays work, and I can pass np.float64 and builtin.float for double-type arguments, so np.float32 should work as well?", but then it gets converted to a different type altogether.

```python

namespace py = pybind11;

auto f_double(double x) { return "double"; }
auto f_float(float x) { return "float"; }
auto f_int(int x) { return "int"; }

PYBIND11_MODULE(py11, m) {
    // double/float: these two overloads have to be ordered this way... (see #1512)
    m.def("f", &f_double, py::arg("x"));
    m.def("f", &f_float, py::arg("x"));
    // int
    m.def("f", &f_int, py::arg("x"));
}
```

Let's give this a spin:

```python
>>> f?
Overloaded function.
1. f(x: float) -> str  # <-- ?
2. f(x: float) -> str
3. f(x: int) -> str

>>> f(1)
'int'

>>> f(1.)
'double'  # note: would have been float if two overloads were swapped

>>> f(np.float64(1.))
'double'  # so far so good

>>> f(np.float32(1.))
<ipython-input-7-4d9dada8d141>:1: DeprecationWarning: an integer is required (got type numpy.float32).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  f(np.float32(1.))
'int'  # not so good
```

aldanor:
Closes #2036, #1512 

- Add `py::numpy_scalar<>` and `py::make_scalar()` as strict type casters between C and NumPy arithmetic types + tests + docs
- Fix a bug where `PyArray_DescrNewType` API code was wrong (although it's not really used at the moment, but still)

### DIFF
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -234,6 +234,13 @@ them mapping to respective C++ counterparts.
     NumPy type and nothing else (e.g., ``py::numpy_scalar<int64_t>`` will not
     accept built-in ``int`` or any other type for that matter).
 
+.. note::
+
+    Native C types are mapped to NumPy types in a platform specific way: for
+    instance, ``char`` may be mapped to either ``np.int8`` or ``np.uint8``
+    depending on the platform. If you want to ensure specific NumPy types,
+    it is recommended to use fixed-width aliases from ``<cstdint>>``.
+
 Vectorizing functions
 =====================
 
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -205,6 +205,35 @@ prevent many types of unsupported structures, it is still the user's
 responsibility to use only "plain" structures that can be safely manipulated as
 raw memory without violating invariants.
 
+Scalar types
+============
+
+In some cases we may want to accept or return NumPy scalar values such as
+``np.float32`` or ``np.uint16``. It is especially important in case of
+C-side single-precision floats which by default will be bound to Python's
+double-precision builtin floats, causing mismatch in float precision.
+
+Luckily, there's a helper type for this occasion - ``py::numpy_scalar``:
+
+.. code-block:: cpp
+
+    m.def("square_float", [](py::numpy_scalar<float> value) {
+        float v = value;
+        return py::make_scalar(v * v);
+    });
+
+This type is trivially convertible to and from the type it wraps; currently
+supported scalar types are NumPy arithmetic types: ``bool_``, ``int8``,
+``int8``, ``int16``, ``int32``, ``int64``, ``uint8``, ``uint16``, ``uint32``,
+``uint64``, ``float32``, ``float64``, all of them mapping to respective C++
+equivalents.
+
+.. note::
+
+    This is a strict type, it will only allow input arguments of the specified
+    NumPy type and nothing else (e.g., ``py::numpy_scalar<int64_t>`` will not
+    accept built-in ``int`` or any other type for that matter).
+
 Vectorizing functions
 =====================
 
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -225,8 +225,8 @@ Luckily, there's a helper type for this occasion - ``py::numpy_scalar``:
 This type is trivially convertible to and from the type it wraps; currently
 supported scalar types are NumPy arithmetic types: ``bool_``, ``int8``,
 ``int8``, ``int16``, ``int32``, ``int64``, ``uint8``, ``uint16``, ``uint32``,
-``uint64``, ``float32``, ``float64``, all of them mapping to respective C++
-equivalents.
+``uint64``, ``float32``, ``float64``, ``complex64``, ``complex128``, all of
+them mapping to respective C++ counterparts.
 
 .. note::
 
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -519,6 +519,11 @@ struct numpy_scalar {
     numpy_scalar& operator=(value_type value) { this->value = value; return *this; }
 };
 
+template<typename T>
+numpy_scalar<T> make_scalar(T value) {
+    return numpy_scalar<T>(value);
+}
+
 class dtype : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(dtype, object, detail::npy_api::get().PyArrayDescr_Check_);
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -39,6 +39,8 @@ NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 class array; // Forward declaration
 
+template<typename> struct numpy_scalar; // Forward declaration
+
 NAMESPACE_BEGIN(detail)
 template <typename type, typename SFINAE = void> struct npy_format_descriptor;
 
@@ -471,6 +473,19 @@ template <typename T, ssize_t Dim>
 struct type_caster<unchecked_mutable_reference<T, Dim>> : type_caster<unchecked_reference<T, Dim>> {};
 
 NAMESPACE_END(detail)
+
+template<typename T>
+struct numpy_scalar {
+    using value_type = T;
+
+    value_type value;
+
+    numpy_scalar() = default;
+    numpy_scalar(value_type value) : value(value) {}
+
+    operator value_type() { return value; }
+    numpy_scalar& operator=(value_type value) { this->value = value; return *this; }
+};
 
 class dtype : public object {
 public:
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -256,6 +256,31 @@ private:
     }
 };
 
+template<typename T> struct numpy_scalar_info {};
+
+#define DECL_NPY_SCALAR(ctype_, name_, typenum_) \
+    template<> struct numpy_scalar_info<ctype_> { \
+        static constexpr auto name = _(name_); \
+        static constexpr int typenum = npy_api::typenum_##_; \
+    }
+
+DECL_NPY_SCALAR(bool, "bool", NPY_BOOL);
+
+DECL_NPY_SCALAR(int8_t, "int8", NPY_INT8);
+DECL_NPY_SCALAR(int16_t, "int16", NPY_INT16);
+DECL_NPY_SCALAR(int32_t, "int32", NPY_INT32);
+DECL_NPY_SCALAR(int64_t, "int64", NPY_INT64);
+
+DECL_NPY_SCALAR(uint8_t, "uint8", NPY_UINT8);
+DECL_NPY_SCALAR(uint16_t, "uint16", NPY_UINT16);
+DECL_NPY_SCALAR(uint32_t, "uint32", NPY_UINT32);
+DECL_NPY_SCALAR(uint64_t, "uint64", NPY_UINT64);
+
+DECL_NPY_SCALAR(float, "float32", NPY_FLOAT);
+DECL_NPY_SCALAR(double, "float64", NPY_DOUBLE);
+
+#undef DECL_NPY_SCALAR
+
 inline PyArray_Proxy* array_proxy(void* ptr) {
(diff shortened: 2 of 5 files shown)


aldanor:
This is definitely not the most efficient way of doing it as it creates dtypes etc... but at least it's fully strict and correct.

I guess instead of messing with dtypes you could extract Python types for all require numpy scalar types and cache them somewhere, and then do a simple (non-numpy) type equivalence check or something of the sort.

docs/advanced/pycpp/numpy.rst
```diff
@@ -205,6 +205,35 @@ prevent many types of unsupported structures, it is still the user's
...
+This type is trivially convertible to and from the type it wraps; currently
+supported scalar types are NumPy arithmetic types: ``bool_``, ``int8``,
+``int8``, ``int16``, ``int32``, ``int64``, ``uint8``, ``uint16``, ``uint32``,
+``uint64``, ``float32``, ``float64``, all of them mapping to respective C++
+equivalents.
+
+.. note::
+
+    This is a strict type, it will only allow input arguments of the specified
+    NumPY type and nothing else (e.g., ``py::numpy_scalar<int64_t>`` will not
```
ax3l:
```suggestion
    NumPy type and nothing else (e.g., ``py::numpy_scalar<int64_t>`` will not
```

docs/advanced/pycpp/numpy.rst
```diff
@@ -205,6 +205,35 @@ prevent many types of unsupported structures, it is still the user's
...
+Luckily, there's a helper type for this occasion - ``py::numpy_scalar``:
+
+.. code-block:: cpp
+
+    m.def("square_float32", [](py::numpy_scalar<float> value) {
+        float v = value;
+        return py::make_scalar(v * v);
+    });
+
+This type is trivially convertible to and from the type it wraps; currently
```
ax3l: Why not the complex types as well?
https://docs.scipy.org/doc/numpy/user/basics.types.html

aldanor: `np.complex64` we probably can (`std::complex<float>`), but `np.complex128` is now platform-dependent (matches precision of built-in complex in Python).

aldanor:
> Would it be possible to provide a type to match all py::numpy_scalar types, similar to py::buffer, as well? This would reduce the number of overloads needed in user code.

I'm not sure it's possible the way it's currently done since it just matches typenums in a very strict way.. you mean something like `py::numpy_polymorphic_scalar`? But then to extract anything out of it you'll need a whole layer of API which resembles `array`'s API.

aldanor:
Ok, tried to improve type caster a little, using a basic isinstance check, hopefully it's sufficient? (the arithmetic types and dtypes are now also cached statically)

A little check:

```c++
m.def("f32_in", [](py::numpy_scalar<float>) { return 0; });
m.def("f32_out", []() { return py::numpy_scalar<float>(0.); })
```

Before:

```python
>>> %timeit f32_in(f)
430 ns ± 8.79 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

>>> %timeit f32_out()
162 ns ± 0.424 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

After:

```python
>>> %timeit f32_in(f)
355 ns ± 2.64 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

>>> %timeit f32_out()
149 ns ± 0.819 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

docs/advanced/pycpp/numpy.rst
ax3l: I would add `np.csingle` and `np.cdouble`, no?

aldanor:
@ax3l Note that the current type names in type signatures are actually only correct on a standard 64-bit platform, so they need to be changed to avoid confusion. I.e., float is not float32 in general and double is not float64.

While we can surely just map `py::numpy_scalar<float>` to `NPY_FLOAT` and same with double (and complex numbers) the way it's currently done, what do we write in type signature? Note that `float` will overlap with builtin float and it's not nice. Also, the size of those floats/doubles is embedded in the resulting binary, so I guess it would be nice to show it (like `float16` or `float64` or whatever it is)? Any ideas?

Also, I guess we want to bind `long double` as well then? (and `complex<float>`, `complex<double>` - but then it's back to the same question again as above).

docs/advanced/pycpp/numpy.rst
aldanor: See the detailed comment below re: `csingle`, `cdouble`, `float` and `double`.

aldanor:
Ok, added long doubles and all three types of complex numbers + updated the tests.

aldanor:
Looks green now. (damn all those msvc tests, that took a bit...)

docs/advanced/pycpp/numpy.rst
```diff
@@ -205,6 +205,35 @@ prevent many types of unsupported structures, it is still the user's
...
+In some cases we may want to accept or return NumPy scalar values such as
+``np.float32`` or ``np.uint16``. It is especially important in case of
+C-side single-precision floats which by default will be bound to Python's
+double-precision builtin floats, causing mismatch in float precision.
+
+Luckily, there's a helper type for this occasion - ``py::numpy_scalar``:
+
+.. code-block:: cpp
+
+    m.def("square_float32", [](py::numpy_scalar<float> value) {
```
ax3l: just to be general:
```suggestion
    m.def("square_float", [](py::numpy_scalar<float> value) {
```

include/pybind11/numpy.h
```diff
@@ -148,14 +147,23 @@ struct npy_api {
...
             NPY_LONG_, NPY_LONGLONG_, NPY_INT_),
-        NPY_UINT64_ = platform_lookup<std::uint64_t, unsigned long, unsigned long long, unsigned int>(
+        NPY_UINT64_ = platform_lookup<8, unsigned long, unsigned long long, unsigned int>(
             NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
+        // The same type of lookups with floats; the order is: double, float, long double
+        NPY_FLOAT32_ = platform_lookup<4, double, float, long double>(
+            NPY_DOUBLE_, NPY_FLOAT_, NPY_LONGDOUBLE_),
+        NPY_FLOAT64_ = platform_lookup<8, double, float, long double>(
+            NPY_DOUBLE_, NPY_FLOAT_, NPY_LONGDOUBLE_),
+        NPY_COMPLEX64_ = platform_lookup<4, double, float, long double>(
```
ax3l: is this correct, not
```suggestion
        NPY_COMPLEX64_ = platform_lookup<8, std::complex<double>, std::complex<float>, std::complex<long double>>(
```
Anyway, probably does not matter.

include/pybind11/numpy.h
```diff
@@ -250,6 +267,64 @@ struct npy_api {
...
+
+#define DECL_NPY_SCALAR(ctype_, typenum_) \
+    template<> struct numpy_scalar_info<ctype_> { \
+        static constexpr auto name = npy_format_descriptor_name<ctype_>::name; \
+        static constexpr int typenum = npy_api::typenum_##_; \
+    }
+
+DECL_NPY_SCALAR(bool, NPY_BOOL);
+
+DECL_NPY_SCALAR(int8_t, NPY_INT8);
```
ax3l: I am not entirely sure if specializing for fundamental types `short`, `int`, `long` and `long long` isn't the better choice to cover all types (and then mapping to concrete bit representations).

include/pybind11/numpy.h
aldanor: This is closer to how checks look like in npy_config header. Anyways, it's the same thing since size of `complex<>` is always double the size of the underlying type.

include/pybind11/numpy.h
aldanor: Yea, I'm not sure about that either, it's all a bit dodgy. But I guess if done this way it's at least guaranteed they don't overlap.

ax3l: Yes, but it does not guarantee that on platforms where `int` is implemented with the same size as `long` both types will be covered by the API.

aldanor: As I've said, I don't have very strong preference, it's somewhat dodgy either way, but we could do something like this (and hope that the stars align in the right direction):

```c++
DECL_NPY_SCALAR(signed char, NPY_BYTE);
DECL_NPY_SCALAR(short, NPY_SHORT);
DECL_NPY_SCALAR(int, NPY_INT);
DECL_NPY_SCALAR(long, NPY_LONG);
DECL_NPY_SCALAR(long long, NPY_LONGLONG);

DECL_NPY_SCALAR(unsigned char, NPY_UBYTE);
DECL_NPY_SCALAR(unsigned short, NPY_USHORT);
DECL_NPY_SCALAR(unsigned int, NPY_UINT);
DECL_NPY_SCALAR(unsigned long, NPY_ULONG);
DECL_NPY_SCALAR(unsigned long long, NPY_ULONGLONG);
```

Because there's 5 of each signed/unsigned standard types in C, there will be guaranteed to be an overlap (i.e. almost always `long` and `long long`). But I guess that's the way you have to do it "by the book".

Unless I'm missing something, that's probably ok, we will have 10 type casters for 10 distinct C types which will map to 8 (or less) distinct numpy types.

aldanor: Technically, there's another special type here - `char`, which may be either `NPY_BYTE` or `NPY_UBYTE` depending on signedness.

Anyways, I've update the mappings and added the `char` as well + a note in the doc.

aldanor:
And now with native  int mappings, [here are](https://travis-ci.org/pybind/pybind11/jobs/635192923) the test failures :) Hmm...

include/pybind11/numpy.h
ax3l: actually there are even three *distinct* char types: `char`, `signed char` and `unsigned char` ^^
https://en.cppreference.com/w/cpp/language/types#Character_types

aldanor: > actually there are even three distinct char types: unsigned char, signed char and char ^^

Have you read the last diff? :)

aldanor:
So... why do the tests fail? Hm..

docs/advanced/pycpp/numpy.rst
```diff
@@ -205,6 +205,42 @@ prevent many types of unsupported structures, it is still the user's
...
+    This is a strict type, it will only allow input arguments of the specified
+    NumPy type and nothing else (e.g., ``py::numpy_scalar<int64_t>`` will not
+    accept built-in ``int`` or any other type for that matter).
+
+.. note::
+
+    Native C types are mapped to NumPy types in a platform specific way: for
+    instance, ``char`` may be mapped to either ``np.int8`` or ``np.uint8``
+    depending on the platform. If you want to ensure specific NumPy types,
+    it is recommended to use fixed-width aliases from ``<cstdint>>``.
```
ax3l: typo:
```suggestion
    it is recommended to use fixed-width aliases from ``<cstdint>``.
```

cosama:
Have been using pybind11 for a project and it works really well, great work on getting here.

I don't know enough about the `numpy` internals, to help here much I think but it would be very useful for that project to have also a untyped `py::numpy_scalar` in analogue to `py::array`. At least it might make sense to name the current implementation `py::numpy_scalar_t` similar as `py::array_t` so that  a `py::numpy_scalar` could be added in the future.

@aldanor: Thanks for working on this.

PR #2067, opened 2020-01-13
Allow nonconstructible holders

### LINKED ISSUE #1178: Shared Pointers and Protected Constructors
I am trying to wrap a C++ class of the following type. The class has a private destructor but should be containable in a shared_ptr due to the public inheritance from enable_shared_from_this.

The code also throws an error if I try to use the following pybind wrapper:
py::class_<MyClass, std::unique_ptr<MyClass, py::nodelete>> clsMyClass(mod, "MyClass");

#include <pybind11/stl.h>

namespace py = pybind11;
namespace {

class MyClass : public std::enable_shared_from_this<MyClass> {
protected:
	MyClass() {}
	~MyClass() {}
};

PYBIND11_MODULE(myclass, mod) {
    py::class_<MyClass, std::shared_ptr<MyClass>> clsMyClass(mod, "MyClass");
}
}

pd-robgee:
Fixes #1178

This PR implements the fix from 53be81931f35313f70affc9826bdfed9820cce2c.
That fix by itself leads to an uncaught exception instead of a Python RuntimeError, this is fixed in this PR.
Furthermore test cases have been added.


wanfranck:
Sorry to interrupt, but is there any progress on this issue? I feel like I'm facing the problem which could be easily fixed as soom as this PR will be merged. Best regards.

pd-robgee:
Unfortunately the project maintainers didn't/don't seem very interested in incorporating this suggestion into their code base. As can be seen from this PRs conversation page I requested a review from three suggested reviewers, but to this day none of them have even commented anything at all.

I have accepted this fate and worked around it by incorporating this patch via the build system in the project I used pybind11 in. I have since moved on to other projects and I'm no longer involved in that project. Therefore, attempting to get this PR merged is no longer a high priority for me. I'm sorry @wanfranck, but thank you for your interest in this patch anyway!

EricCousineau-TRI:
@rwgk Can you advise on how this may (or may not) be more admissible via the `smart_holder` based workflow?

At first blush, concerned about performance and code impacts given core part of code (broadly #2760; acutely, should have scoped benchmark).

pinchartl:
The [libcamera](https://www.libcamera.org/) Python [bindings](https://github.com/tomba/libcamera/commit/17fce573afb24e57cfea486d0a627bb64c2b145d) are blocked by this issue. A review would be appreciated. If a different approach is needed, we would like to know.

rwgk:
> The [libcamera](https://www.libcamera.org/) Python [bindings](https://github.com/tomba/libcamera/commit/17fce573afb24e57cfea486d0a627bb64c2b145d) are blocked by this issue. A review would be appreciated. If a different approach is needed, we would like to know.

Have you tried the smart_holder branch? It's basically drop-in, use `classh<T>` instead of `class_<T, shared_ptr<T>>` where needed (fine to only use where actually needed). At Google we are using the smart_holder code with private destructors.

PR #2072, opened 2020-01-15
Fix conda include reporting [do not merge]

henryiii:
The Conda-forge package is reporting the wrong directory. From local testing, the standard way of accessing the directory seems to be working correctly; it's just the injected special logic that is causing issues.

Is the virtualenv logic needed, or could it be removed as well?

Testing this in Conda-forge before making the PR non-draft.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -287,3 +287,11 @@ code by introspecting existing C++ codebases using LLVM/Clang. See the
 [binder]_ documentation for details.
 
 .. [binder] http://cppbinder.readthedocs.io/en/latest/about.html
+
+[AutoWIG]_ is a Python library that wraps automatically compiled libraries into
+high-level languages. It parses C++ code using LLVM/Clang technologies and
+generates the wrappers using the Mako templating engine. The approach is automatic,
+extensible, and applies to very complex C++ libraries, composed of thousands of
+classes or incorporating modern meta-programming constructs.
+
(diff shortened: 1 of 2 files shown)


henryiii:
This fixed the problem with Conda here: https://github.com/conda-forge/pybind11-feedstock/pull/43

henryiii:
The failure with CI is due to a ruby error in homebrew, and is unrelated.

henryiii:
The thing I have not tested is that if a pip install in a conda environment works properly with this change.

ax3l:
Thank you for the report.

As added [here](https://github.com/conda-forge/pybind11-feedstock/pull/43#issuecomment-574957768), I cannot reproduce the problem with conda (on conda-forge) that you report and this diff likely breaks the includes.

Also note that upcoming releases got again a logic change in https://github.com/pybind/pybind11/pull/1995 - maybe you are mismatching versions between development and latest stable with corresponding install logic locally?

henryiii:
Yes, this is targeted at stable, it may be fixed in master.

ax3l:
Got it, it's a conda-forge, Unix-only install problem.
Fix proposed in https://github.com/conda-forge/pybind11-feedstock/pull/44

wjakob:
The logic in `__init__.py` has changed quite a bit in the meantime, and https://github.com/conda-forge/pybind11-feedstock/pull/44 landed. Can this be closed?

henryiii:
I think so. I'll open an issue if anything is still broken.

PR #2074, opened 2020-01-16
stl_bind: add binding for std::vector::clear

fwjavox:
This commit adds bindings for the std::vector::clear function when using pybind11::bind_vector.


wjakob:
Great, thanks!

PR #2083, opened 2020-01-18
Update binding example for virtual function doc.

AntoinePrv:
Hello,

Just suggesting a small change in the documentation.
In section "_Combining virtual functions and inheritance_", the bindings with trampoline class don't mention the inheritance to parent class.
It may be straightforward for some people, but, as a beginner, it threw me off for a bit.
I suggest showing the bindings as they would look in a complete example.


wjakob:
Merged, thanks!

Issue #2078, opened 2020-01-17
Conversion between std::chrono and numpy.datetime64


phil-zxx:
Using `#include <pybind11/chrono.h>` we can define
```py
def(py::init([](const std::chrono::system_clock::time_point& obj)
```
so that we can convert between `std::chrono` and python's `datetime`.

Is there a similar concept for conversions between `std::chrono` and numpy's `datetime64`? I had a look at `pybind11/numpy.h` but couldn't find anything.

I am happy implement it and submit a PR, if someone could point me in the right direction for how to do it. Many thanks!

phil-zxx:
I've tried it myself, please see PR.

PR #2085, opened 2020-01-19
Added ability to convert from numpy.datetime64 to system_clock::time_point

phil-zxx:
Added ability to convert from Python `numpy.datetime64` to C++ `system_clock::time_point`

As raised here: https://github.com/pybind/pybind11/issues/2078

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include "pybind11.h"
+#include "numpy.h"
 #include <cmath>
 #include <ctime>
 #include <chrono>
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -11,7 +11,6 @@
 #pragma once
 
 #include "pybind11.h"
-#include "numpy.h"
 #include <cmath>
 #include <ctime>
 #include <chrono>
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
Does this work with different units that `numpy.datetime64` may have, such as `datetime64[ns]`? See https://github.com/pybind/pybind11/issues/2078.

m4ce:
any chance we could have this PR reviewed and merged? I am looking at the ability to convert chrono::system_clock::time_point to numpy.datetime64. Will this PR also work when returning a strucuted numpy array from C++ to python where a field is a time_point?

stevesimmons:
Hi guys, is solving this on anyone's radar? I'm asking because I'd very much like the awkward-1.0#367 issue mentioned above by @jpivarski to be solved. I don't have the time right now to solve these two issues myself. Just wondering if anyone else is likely to tackle them soon...

jpivarski:
https://github.com/scikit-hep/awkward-1.0/issues/367 for cross-linking

SimonHeybrock:
@stevesimmons @jpivarski For `scipp`, we have since implemented bindings based on `py::array` in a manual way, supporting various time units and avoiding copies. I don't know if something like this could be brought into `pybind11` itself, but maybe this is useful for you. The two main bits are: https://github.com/scipp/scipp/blob/02aa4be8d59b84caf74d2bd37bef990e952564cf/python/bind_data_access.h#L56-L79 and https://github.com/scipp/scipp/blob/02aa4be8d59b84caf74d2bd37bef990e952564cf/python/numpy.h#L46-L63.

PR #2087, opened 2020-01-20
Fix the use of MSVC in an MSYS environment

### LINKED ISSUE #2079: Compilation on windows Fails after PR #2053
Hello everyone,

First, thank you everyone for this great tool ! 

## Issue description

Current system:

* Windows 10
* MinGW
* conda 4.8.1 installed on windows
* python version 3.7.4 (the one in conda)
* cmake 3.14.0 installed on windows
* compiler MSVC 14.24.28314

After PR #2053 was merged, our building pipelines on windows started failing. The output message error is `fatal error LNK1181: cannot open input file 'python37.lib`.
After tracking the problem, I found out that `PYTHON_LIBRARY` in file `tools/FindPythonLibsNew.cmake` was not set to the correct path.

* Before PR #2053 `PYTHON_LIBRARY` = `C:/Users/user/Anaconda3/libs/python37.lib`
* After PR #2053 `PYTHON_LIBRARY` = `python37.lib`

Debugging each step of the previous file resulted that:

* `PYTHON_LIBDIR` is empty
* After the patch, it tries to look in a Linux location for the library
* `PYTHON_LIBRARY` is set to `python37.lib`

I found a solution changing some lines in `tools/FindPythonLibsNew.cmake`, without changing the behaviour for Linux or Mac.

I would like to know if you'll be interested for a PR, if yes the only thing is I don't have many more windows system to test on it. To I don't know if other Windows user with their build system could test ?

Or maybe do you have any other suggestion that could help us of course :) ?

## Reproducible example code

1. windows system
2. install mingw
3. install conda for windows
4. import a C++ project using cmake for building process, and of course pybind

In the `CmakeLists.txt` file, we use `add_subdirectory()` to indicate where the folder of pybind11 is located.

Expected error: `fatal error LNK1181: cannot open input file 'python37.lib`

Let me know if you have any questions/suggestion.

Cheers,
Julián

Baljak:
Since MSVC can also be used in an MSYS environment (for example the one provided with Git for Windows), we also need to check if the MinGW compiler is used alongside the `MSYSTEM` environment variable. This expand the MSYS check made in PR #2053 

@MonkeyBreaker Could you check that this fixes your issue?
@kanonet Could you check as well that pybind11 + MSYS still work in your cases?

Fix #2079


MonkeyBreaker:
@Baljak 

I confirm that it now work's on my side

Great !

wjakob:
Shouldn't `MSVC` appear somewhere in the condition?

Baljak:
@wjakob I agree, I made the condition like this to allow more compilers than MSVC.

kanonet:
I can confirm, that this does work for me with MSYS2+MinGW. I think this is good as you wrote it, this way it does only change behavior for Msys+MinGW. Did someone test MinGW without MSYS?
This is how my original PR #1851 should have looked, to bad it did not. Thx for fixing it!

Baljak:
@kanonet I did and it works. Actually I made PR #2048 because I'm using MinGW alone. Now every configuration should be OK.

MonkeyBreaker:
Hi !

Any idea when this PR will be merged on master ?
Nothing urgent, as a temporary solution we checkout on branch `v2.4`.

wjakob:
Merged!

wjakob:
Thanks for tracking this down :)

reds-heig:
Thank you ! :)

MonkeyBreaker:
Thank you !

Issue #1922, opened 2019-09-12
get_type_overload may encounter false cache hits if derived instances are GC'd


EricCousineau-TRI:
## Issue description

This is the upstream form of https://github.com/RobotLocomotion/drake/issues/11424 that @BetsyMcPhail root-caused in this discussion: https://github.com/RobotLocomotion/drake/issues/11424#issuecomment-527991296

> High-level overview and proposed solution
> 
> Pybind11 caches functions that aren't overloaded in Python to avoid costly Python dictionary lookups.
> 
> The non-overload cache is a set which uses (type.ptr, function name) as a key. Once an item is added to the cache, it is never removed.
> 
> The bug occurs when a type that has been added to the cache (e.g type = <class 'custom_test.TestCustom.test_deprecated_protected_aliases..OldSystem'>
> type.ptr() = 0x2746c58) is destroyed and a new type is created with the exact same pointer (e.g. type = <class 'custom_test.TestCustom.test_leaf_system_overrides..TrivialSystem'>, type.ptr() = 0x2746c58) that should not be in the cache.
> 
> The first attempt at fixing the issue was to use self.ptr instead of type.ptr in the cache key. Testing demonstrated that this was not enough to solve the issue.
> 
> In the end, removing all associated cached items when an instance is deregistered seems to fix the issue.

She submitted a potential fix for it in our fork here:
https://github.com/RobotLocomotion/pybind11/pull/32

## Reproducible example code

I've done a quick minimal repro based on master @ 12e8774b here:
https://github.com/EricCousineau-TRI/pybind11/commit/97260868f5fb0f1a0ba81c38b173bee99b6c71c4

Extra keyword: garbage collect virtual inheritance polymorphism

wjakob:
Hi @EricCousineau-TRI,

oh oh, that sounds like it was very painful to track down :(. That's a very subtle issue indeed and something that was not considered during development. Unfortunately haven't bee able to come up with nice way of resolving the problem.

One option of course would be to remove the caching mechanism. That's a no-go because it will route every virtual function call through Python (& the GIL). Some C++ project issue vast amounts of these, potentially in parallel from multiple threads. Such a change would essentially break these projects.

[...]

PR #2092, opened 2020-01-23
Update 'inactive_overload_cache' key

BetsyMcPhail:
Instead of just the function name, use `__qualname__.name`.

Solves issue #1922  

\@wjakob

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -83,9 +83,9 @@ template <typename value_type>
 using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
 
 struct overload_hash {
-    inline size_t operator()(const std::pair<const PyObject *, const char *>& v) const {
+    inline size_t operator()(const std::pair<const PyObject *, std::string>& v) const {
         size_t value = std::hash<const void *>()(v.first);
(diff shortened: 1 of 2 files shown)


EricCousineau-TRI:
@YannickJadoul @rwgk I was updating our fork, and forgot that we hadn't landed this (see linked issue for more deets).

Is this a candidate for 2.6.0?

EricCousineau-TRI:
Came across this when crafting: https://github.com/RobotLocomotion/pybind11/pull/47

YannickJadoul:
This should probably use `detail::get_fully_qualified_tp_name` (from #2520), but apart from that, a quick look says "yes, I think so?"

YannickJadoul:
At the same time, it's still quite late to add, though? :-/

YannickJadoul:
Also, what if the new class has exactly the same name as the old one?

bstaletic:
Also, also, the changes to the hash function are really wrong. You don't want to combine hashes with just an xor.

henryiii:
We can work on it for 2.7 - I think it's too late for 2.6 for anything with a chance of breaking.

EricCousineau-TRI:
Sounds good! We'll address this for release 2.7.0, not 2.6.0 - thanks!

EricCousineau-TRI:
Closed in lieu of superseding PR, #2772

PR #2095, opened 2020-01-25
Fix docstrings of enum methods

YannickJadoul:
#1511 broke the docstrings of the `__repr__`, `__add__`, etc. methods of `enum_` types. E.g., the docstring of a `__repr__` now looks like this:

```
>>> example.AnEnum.__repr__.__doc__
'(self: handle) -> str\n'
```

This PR fixes that by passing the name to the `cpp_function` constructors, in the same was as `class_::def` would do:

```
>>> example.AnEnum.__repr__.__doc__
'__repr__(self: handle) -> str\n'
```

I also converted `__setstate__` and `__getstate__` to `pickle`, to get rid of the warnings about old- vs. new-style constructors.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1418,7 +1418,7 @@ struct enum_base {
                         return pybind11::str("{}.{}").format(type_name, kv.first);
                 }
                 return pybind11::str("{}.???").format(type_name);
-            }, is_method(m_base)
+            }, name("__repr__"), is_method(m_base)
         );
 
         m_base.attr("name") = property(cpp_function(
@@ -1429,7 +1429,7 @@ struct enum_base {
                         return pybind11::str(kv.first);
                 }
                 return "???";
-            }, is_method(m_base)
+            }, name("name"), is_method(m_base)
         ));
 
         m_base.attr("__doc__") = static_property(cpp_function(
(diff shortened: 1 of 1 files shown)


wjakob:
All the changes look good for me, except the changes regarding ``hash`` and ``getstate``. I remember at the time intentionally getting rid of pickle() to avoid unneeded code, which you now added back. Can we stick with how it was previously?

YannickJadoul:
The reason I did this is that by adding the `name` to the original `__setstate__`, it would start complaining about "using an old-style placement-new '__setstate__' which has been deprecated", and I thought the one function wouldn't matter and even avoid a second call from C++ -> Python -> C++ by calling  `__int__` is called in `__getstate__`.

But this latest commit should fix it.

One minor concern is that the docstrings still show `int_` instead of `int`, but if I can easily fix that, I'll open a new PR.

wjakob:
LGTM, thanks!

wjakob:
FYI: I removed the use of `std::move` in ``__setstate__``, that seemed a bit overkill for an enum.

YannickJadoul:
> FYI: I removed the use of `std::move` in `__setstate__`, that seemed a bit overkill for an enum.

Oh right, it's just a primitive underlying type, anyway. That was just a remainder of copy-pasting the `__setstate__` implemenation, I suppose.

Issue #2076, opened 2020-01-16
pybind11 ignores exceptions raised by len(arg), calls std::vector::reserve() with bad value


nmusolino:
##  Issue description

When converting a sequence-like argument to an `std::vector<T>`, pybind11 ignores exceptions raised by `len(arg)` and erroneously calls `std::vector<T>::reserve()` with a very large value.  The result is a confusing exception of the following form:
```
Traceback (most recent call last):
  File "./pybind11_repro.py", line 11, in <module>
    my_module.f(a)
ValueError: vector::reserve
```

## Reproducible example code

C++ code:
```
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>

#include <vector>

namespace py = pybind11;
using namespace py::literals;

namespace my_module {

void f(const std::vector<int>&) {}

}  // end namespace my_module

PYBIND11_MODULE(my_module, m){
    m.def("f", &my_module::f);
};
```

This C++ code was compiled with:
```
g++-8 -Wall -shared -std=c++17 -fPIC -o my_module$(python3-config --extension-suffix) $(python3-config --ldflags) pybind11_repro.cpp -I.<pybind11 include dir> -I/usr/local/include/python3.8  
```

Python code:
```
import numpy
import my_module

my_module.f([1, 2])  # Okay

a = numpy.array(2.0)
my_module.f(a)       # raises: ValueError: vector::reserve
```

I used a zero-dimensional numpy array to demonstrate this problem, but it actually occurs with any object that is sequence-like (has `__getitem__`) and raises from `__len__`:
```
class BadLen(): 
    def __getitem__(self, i): 
        return i 
           
    def __len__(self): 
        raise Exception() 

my_module.f(BadLen())    # raises: ValueError: vector::reserve
```

## Result and expected result
When running the Python code above, a C++ exception is thrown by `std::vector<int>::reserve()`.  pybind11 propagates this to the Python caller, but this is confusing and does not indicate why the operation failed.
```
Traceback (most recent call last):
  File "./pybind11_repro.py", line 11, in <module>
    my_module.f(a)
ValueError: vector::reserve
```

This exception occurred because `len(a)` in the example above raises a Python exception (see Diagnosis section below):
```
>>> len(a)
Traceback (most recent call last):
  File "pybind11_repro.py", line 9, in <module>
    len(a)
TypeError: len() of unsized object
```
I think that pybind11 should propagate this *original* Python exception to the caller, instead of triggering an unrelated exception at a later point in time.

[...]

PR #2096, opened 2020-01-26
Propagate exceptions in sequence::size() (#2076)

nmusolino:
Adds a check for error after a call to `PySequence_Size` in `pybind11::sequence::size()`.

As detailed in #2076, before this change, `sequence::size()` could return an erroneous value of `SIZE_MAX` in a scenario when `PySequence_Size` indicated an exception by returning `-1`.

The unit test in this MR, when run with `pytypes.h` from before the change, results in the following error: `SystemError: <built-in method sequence_length of PyCapsule object at 0x...> returned a result with an error set`.  After this MR is applied, the test completes successfully.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1242,7 +1242,12 @@ private:
 class sequence : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(sequence, object, PySequence_Check)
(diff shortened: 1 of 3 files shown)


nmusolino:
@wjakob, I think this MR is ready for review.  Would you be able to take a look?

wjakob:
Merged, thanks!

PR #2099, opened 2020-01-27
eigen: Don't require conformability on length-0 dimensions

### LINKED ISSUE #2098: eigen: Function taking `Eigen::Ref<const MatrixXd>` cannot accept 0xN matrices (N > 1)
## Issue description

When binding a function that takes `Eigen::Ref<const MatrixXd>`, `type_caster::load` fails when `src` has a size of `(0, N)`, where N > 1.

(For row-major, I assume that this conversion fails if the matrix is Nx0.)

Relates @allanleal's comment in #738.

## Reproducible example code

Using 4f72ef84, see code here:
https://github.com/EricCousineau-TRI/repro/blob/e10a223690ca5548d250996c5f49548adebe9cd6/python/pybind11/custom_tests/test_eigen_drake_issue12633.cc

Output:

```
<built-in method pass_through of PyCapsule object at 0x7fe90b53dd80> - 0, 1
<built-in method pass_through of PyCapsule object at 0x7fe90b53dd80> - 0, 2
<built-in method pass_through_ref of PyCapsule object at 0x7fe90b53df90> - 0, 1
<built-in method pass_through_ref of PyCapsule object at 0x7fe90b53df90> - 0, 2
  what():  TypeError: pass_through_ref(): incompatible function arguments. The following argument types are supported:
    1. (arg0: numpy.ndarray[float64[m, n], flags.f_contiguous]) -> numpy.ndarray[float64[m, n], flags.f_contiguous]

Invoked with: array([], shape=(0, 2), dtype=float64)

At:
  <string>(9): check
  <string>(12): <module>
```

EricCousineau-TRI:
Resolves #2098

Not completely sure if there should be some special-casing for writeable references...

---
This change is [Reviewable](https://reviewable.io/reviews/pybind/pybind11/2099)

### DIFF
--- a/include/pybind11/eigen.h
+++ b/include/pybind11/eigen.h
@@ -100,9 +100,9 @@ template <bool EigenRowMajor> struct EigenConformable {
         return
             !negativestrides &&
             (props::inner_stride == Eigen::Dynamic || props::inner_stride == stride.inner() ||
-                (EigenRowMajor ? cols : rows) == 1) &&
+                (EigenRowMajor ? cols : rows) <= 1) &&
             (props::outer_stride == Eigen::Dynamic || props::outer_stride == stride.outer() ||
-                (EigenRowMajor ? rows : cols) == 1);
+                (EigenRowMajor ? rows : cols) <= 1);
     }
     operator bool() const { return conformable; }
 };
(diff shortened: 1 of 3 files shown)


wjakob:
Looks great, thank you -- merged!

PR #2100, opened 2020-01-28
Support keyword-only arguments (rebased)

### LINKED ISSUE #1194: Indicating that a function takes keyword-only arguments

skoslowski:
This is a duplicate of #1196 rebased to current master resolving the merge conflicts on the existing PR.

(Fixes #1194)

This adds support for a `py::args_kw_only()` annotation that can be
specified between `py::arg` annotations to indicate that any following
arguments are keyword-only.  This allows you to write:

    m.def("f", [](int a, int b) { /* ... */ },
          py::arg("a"), py::args_kw_only(), py::arg("b"));

and have it work like Python 3's:

    def f(a, *, b):
        # ...

with respect to how `a` and `b` arguments are accepted (that is, `a` can
be positional or by keyword; `b` can only be specified by keyword).

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -362,6 +362,34 @@ like so:
     py::class_<MyClass>("MyClass")
         .def("myFunction", py::arg("arg") = (SomeType *) nullptr);
 
+Keyword-only arguments
+======================
+
+Python 3 introduced keyword-only arguments by specifying an unnamed ``*``
+argument in a function definition:
+
+.. code-block:: python
+
+    def f(a, *, b):  # a can be positional or via keyword; b must be via keyword
+        pass
+
+    f(a=1, b=2)  # good
+    f(b=2, a=1)  # good
+    f(1, b=2)    # good
+    f(1, 2)      # TypeError: f() takes 1 positional argument but 2 were given
+
+Pybind11 provides a ``py::args_kw_only`` object that allows you to implement
+the same behaviour by specifying the object between positional and keyword-only
+argument annotations when registering the function:
+
+.. code-block:: cpp
+
+    m.def("f", [](int a, int b) { /* ... */ },
+          py::arg("a"), py::args_kw_only(), py::arg("b"));
+
+Note that, as in Python, you cannot combine this with a ``py::args`` argument.
+This feature does *not* require Python 3 to work.
+
 .. _nonconverting_arguments:
 
 Non-converting arguments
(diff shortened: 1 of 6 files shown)


TheButlah:
+1 for this feature,

cpuhrsch:
I'd love this as well!

cpuhrsch:
Ping: Are you planning on landing this soon? :)

cpuhrsch:
@gst - are there any updates to this?

cpuhrsch:
Pinging the creator of this PR @skoslowski

skoslowski:
@CharlesB2, I am not a maintainer of pybind11, so there is nothing I can do. This PR only exists to help get the original code merged.

I have been wanting this for a while now and it has dragged on for too long. So, for now, I have fallen back to using a private fork with this patch included.

izdeby:
This would be a great addition to pybind. Is there anyway this can be merged, @gst?

gst:
I dont have merge privilege here otherwise I'd have already merged ;)

@wjakob is the last person having done a commit 10 days ago.. ping ?

jagerman:
I like this addition, but some minor bikeshedding: Thoughts on changing the tag to something shorter, perhaps `py::kwonly{}`?  The name feels a little verbose to me currently:

```C++
m.def("f", [](int a, int b) { /* ... */ },
        py::arg("a"), py::args_kw_only(), py::arg("b"));
```
vs:
```C++
m.def("f", [](int a, int b) { /* ... */ },
        py::arg("a"), py::kwonly(), py::arg("b"));
```

gst:
agree with you `py::kwonly` looks better.

skoslowski:
I am also in favor of `kwonly()`. If there are no objections, I'll amend the original commit.

wjakob:
This is a nice feature, and the patch looks fantastic. Apologies that it's taken me a while to get to this. Many thanks to @jagerman and @skoslowski.

henryiii:
Wow, didn't know this went in!!!

Issue #2062, opened 2020-01-09
Memory leaks in MFC  dialog base application when use scoped_interpreter


ruoleng:
The example code in VC 2017 give a leak tip in debug mode.
Output windows show text like this:

Detected memory leaks!
Dumping objects ->
{266} normal block at 0x00C9B4E8, 4 bytes long.
Data: <    > 00 00 00 00 
Object dump complete.

I had try the code in Win32 Application,   it's OK,  no leak.

The version of Visual Stuido 2017  I'm using  is 15.9.16.

It seem the  "internals_pp" pointer did not release.
In  "internals.h"  file  of  function "inline internals **&get_internals_pp() ". 

## Reproducible example code

namespace py = pybind11;
py::scoped_interpreter guard{};
auto dtPassTo = py::dict();
auto global = py::dict(py::module::import("__main__").attr("__dict__"));
dtPassTo["msg"] = py::cpp_function(
	[](std::string strMsg) {
	::MessageBoxA(nullptr, strMsg.c_str(), "", MB_OK);
});

py::exec("msg('test')", global, dtPassTo);

ruoleng:
Well, I did a little dig. Then I found in 'internals.h' file, the function 'PYBIND11_NOINLINE inline internals &get_internals() '  there is a code block (line 272 ~ 274 ):

        if (!internals_pp) internals_pp = new internals*();
        auto *&internals_ptr = *internals_pp;
        internals_ptr = new internals();

Find out internals_pp is a pointer that point to another pointer. But only release the   *internals_pp pointer, not release the internals_pp pointer.
The release code in line 163 ~ 166  the embed.h file, which is the block of function ‘inline void finalize_interpreter() '.

After I add release internals_pp pointer , there is not more leak.

Before fix, release code is :

if (*internals_ptr_ptr) {
    delete *internals_ptr_ptr;
    *internals_ptr_ptr = nullptr;
}

After fix , release code is :

if (*internals_ptr_ptr) {
    delete *internals_ptr_ptr;
    *internals_ptr_ptr = nullptr;
}

if (internals_ptr_ptr) {
    delete internals_ptr_ptr;
    internals_ptr_ptr = nullptr;
}

erwincoumans:
I can confirm there is a memory leak, even without MFC.
Where is this internals_pp  and internals_ptr  deleted?

[...]

PR #2125, opened 2020-02-28
Remove memory leaks, with this patch it doesn't leak anymore (Windows…

erwincoumans:
… Visual Studio, checked using Visual Leak Detector, https://kinddragon.github.io/vld)

This patch uses Py_AtExit to make sure memory is deallocated at the very end. 
It may conflict with other platforms/configurations (embedded etc). 

The static PyModuleDef def2; assumes there is only one single PYBIND11_MODULE. 
In case more modules are allowed, that part of the patch can be removed and a more fancy solution for that deallocation is needed.

These leaks are similar to Issue #2062 
However, in my case I'm seeing memory leaks just using a simple plain Python plugin using pybind11 (not embedded).

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -212,6 +212,24 @@ inline internals **&get_internals_pp() {
     return internals_pp;
 }
 
+inline void delete_internals_pp() {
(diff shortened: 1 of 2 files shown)


erwincoumans:
Closing, this may be a Windows only issue (looking at the CI errors)

PR #2126, opened 2020-03-01
Add enum value to enum repr

auscompgeek:
This changes enum reprs to look like `<Enum.name: value>` similarly to
the Python enum module.

This keeps the str of enums as `Enum.name`, like the Python enum module.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1486,6 +1486,16 @@ detail::initimpl::pickle_factory<GetState, SetState> pickle(GetState &&g, SetSta
 }
 
 PYBIND11_NAMESPACE_BEGIN(detail)
+
+inline str enum_name(handle arg) {
+    dict entries = arg.get_type().attr("__entries");
+    for (const auto &kv : entries) {
+        if (handle(kv.second[int_(0)]).equal(arg))
+            return pybind11::str(kv.first);
+    }
+    return "???";
+}
+
 struct enum_base {
     enum_base(handle base, handle parent) : m_base(base), m_parent(parent) { }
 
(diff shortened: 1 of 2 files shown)


include/pybind11/pybind11.h
```diff
@@ -1435,29 +1445,22 @@ struct enum_base {
         auto static_property = handle((PyObject *) get_internals().static_property_type);
 
         m_base.attr("__repr__") = cpp_function(
-            [](handle arg) -> str {
+            [](object arg) -> str {
```
henryiii, line 1448: Why handle to object? You don't need reference counting here, do you?

YannickJadoul: Argh, good catch. I should've look more closely at the details rather than eyeballing the structural changes and shrugging "OK"...

YannickJadoul: OK, so I just tried this out, and the reason seems to be because there's no `int_` conversion from `handle` (you can `reinterpret_steal`/`reinterpret_borrow`, but those will not do the `int(...)` conversion).

YannickJadoul: See e.g. `__getstate__` and `__hash__`, also doing that. I don't immediately see a better solution, then? (though if feels there ought to be one, changing the interface of the Python types; but that's a different discussion/PR)

bstaletic: Maybe pass the `object` by reference, to avoid reference counting?

YannickJadoul: @bstaletic Hmmm, interesting.

Thinking about it a bit futher: will it, though? I believe (though the `cast`, `make_caster`, `cast_op`, `PYBIND11_TYPE_CASTER`, ... code is quite convoluted) that the `object` will be created and moved from the caster anyway, if you take it by value. So it shouldn't matter?

YannickJadoul: I vote to keep it this way, merge this PR, and create an issue/discussion on why you can't call the `int_` conversion on a handle (like you can for `str`).

bstaletic: I'm okay with that.

henryiii (approved):

YannickJadoul:
Successfully rebased, it seems :-) Does @bstaletic still want to say something (as you were involved in the discussion?)

Otherwise, this should be ready, @henryiii?

YannickJadoul:
Another rebase, after #2510. If CI turns green, I'll merge, then.

YannickJadoul:
Thanks, @auscompgeek!

Issue #1919, opened 2019-09-11
Pybind always move objects of type std::optional


armandoeng:
If a container has an optional attribute, that attribute will always be moved.

In the code below, I found that the optional_caster override the policy. For reasons unknown to me, it is always overriding to move. Removing that line solves the problem.

## Example code

The custom_optional_caster is equals to the optional_caster, but with a removed line.
``` cpp
#include <pybind11/pybind11.h>
#include <pybind11/operators.h>

#include <optional>
//#undef PYBIND11_CPP17
#include <pybind11/stl.h>

namespace py = pybind11;

PYBIND11_MAKE_OPAQUE(std::vector<double>);

namespace pybind11{
namespace detail{

// This type caster is intended to be used for std::optional and std::experimental::optional
template<typename T> struct custom_optional_caster {
    using value_conv = make_caster<typename T::value_type>;

    template <typename T_>
    static handle cast(T_ &&src, return_value_policy policy, handle parent) {
        if (!src)
            return none().inc_ref();
//        policy = return_value_policy_override<typename T::value_type>::policy(policy);
        return value_conv::cast(*std::forward<T_>(src), policy, parent);
    }

    bool load(handle src, bool convert) {
        if (!src) {
            return false;
        } else if (src.is_none()) {
            return true;  // default-constructed value is already empty
        }
        value_conv inner_caster;
        if (!inner_caster.load(src, convert))
            return false;

        value.emplace(cast_op<typename T::value_type &&>(std::move(inner_caster)));
        return true;
    }

    PYBIND11_TYPE_CASTER(T, _("Optional[") + value_conv::name + _("]"));
};

#ifndef PYBIND11_CPP17
template<typename T> struct type_caster<std::optional<T>>
: public custom_optional_caster<std::optional<T>> {};

template<> struct type_caster<std::nullopt_t>
    : public void_caster<std::nullopt_t> {};

#endif

}
}

struct FooA{
    std::vector<double> my_vector {};
};

struct FooB {
    std::optional<FooA> my_container{};
};

void exportOptionalIssue(py::module &m) {
    py::class_<FooA>(m, "FooA")
        .def(py::init<>())
        .def_readwrite("my_vector", &FooA::my_vector);

    py::class_<FooB>(m, "FooB")
        .def(py::init<>())
        .def_readwrite("my_container", &FooB::my_container);
}
```

The tests were made to be used with pytest. This first shows that FooA works well.
``` python
def test_FooA():
    fooA = FooA()
[...]

PR #2134, opened 2020-03-14
Bugfix/issue #1919 optional always move

fatvlady:
Added test which covers move-on-use bug for `std::optional`.
Fix is to correctly forward reference to `return_value_policy_override`.

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -266,8 +266,9 @@ template<typename T> struct optional_caster {
     static handle cast(T_ &&src, return_value_policy policy, handle parent) {
         if (!src)
             return none().inc_ref();
-        using dereference_type = decltype(*std::forward<T_>(src));
-        policy = return_value_policy_override<dereference_type>::policy(policy);
+        if (!std::is_lvalue_reference<T>::value) {
+            policy = return_value_policy_override<T>::policy(policy);
+        }
         return value_conv::cast(*std::forward<T_>(src), policy, parent);
     }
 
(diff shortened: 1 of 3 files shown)


fatvlady:
Please, link to #1919

fatvlady:
Travis problem with Pypy 2 caused by `numpy == 1.18.1` installation on build agent which requires `python >= 3.5`. Installing `numpy == 1.15.4` first and `scipy` after fixes the problem.

m4ce:
Also came across this issue - can we get this PR merged?

fatvlady:
@wjakob Could you please comment on this/approve? Seems like quite an easy fix with full backward compatibility.

wjakob:
Sorry for taking a long time to look at this. One quick observation: if you study the other STL casters, you will note that they all guard the RVP override with a check of the form:
```
if (!std::is_lvalue_reference<T>::value)
```
Is there any reason for not porting that check here?

fatvlady:
Hi @wjakob, thanks for the reply.
There's no serious reason that I can recall now. Probably my thinking was involving several points:
- `return_value_policy_override` already checks for lvalue reference, no need for duplicate logic
- I was not sure whether it is possible to have `caster<T>::cast<T&>` or vice versa, however your point re other casters implies that it is not the case in current codebase

Now, if we rely on the fact that optional dereference copies over the const-ref modifiers, most elegant solution would be to have
```
policy = return_value_policy_override<T>::policy(policy);
```
without any additional checks. Tests are passing, will commit.

fatvlady:
^ decided to add check just to be compliant with other casters, does not change anything

wjakob:
Ok, this is very straightforward to merge now -- thanks for making the changes.

PR #2151, opened 2020-04-05
Add is_final to disallow inheritance from Python

virtuald:
While not a requirement to bind final C++ classes, the underlying class is probably final for a reason so you might as well disallow python classes from inheriting also?


virtuald:
It appears that [pypy might not honor this flag](https://bitbucket.org/pypy/pypy/issues/2742/cpyext-class-subclassable-even-if)? What's the best way to approach this then -- ignore it and document it? That seems easiest to me.

We might be able to do something in the metaclass to fix it, but I'm not sure the best way to go about that.

wjakob:
LGTM, thanks!

EricCousineau-TRI:
BTW Sorry to have discovered this so late, but this is awesome - thank y'all!

PR #2152, opened 2020-04-05
Throw TypeError when subclasses forget to call __init__

### LINKED ISSUE #2103: Use pybind11 metaclass to detect when __init__ has not been called from subclass
My users are high school robotics teams, and the RobotPy project is wrapping a C++ library that uses a lot of inheritance. They keep inheriting from C++ objects, add a constructor without calling the super constructor, it crashes, and they don't really know why at first. It would be great if pybind11 could catch this and throw a python exception instead with a useful error message.

As noted in this stack overflow answer (https://stackoverflow.com/a/46120342), from a metaclass it's possible to call a function after `__init__` is called (at least from Python anyways). Has anyone tried customizing the pybind11 metaclass to do this sort of detection?

virtuald:
Forgetting to call `__init__` in inherited classes is one of the biggest issues my users run into (and it's frustrating because it typically causes a segfault), I'd love to see this merged. It would make pybind11 so much more usable!

... this probably causes a minor performance impact when creating new objects. I would expect it to be in the noise, though I haven't looked at the timing yet.

- Fixes #2103

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -149,6 +149,11 @@ memory for the C++ portion of the instance will be left uninitialized, which
 will generally leave the C++ instance in an invalid state and cause undefined
 behavior if the C++ instance is subsequently used.
 
+.. versionadded:: 2.5.1
+
+   The default pybind11 metaclass will throw a ``TypeError`` when it detects
+   that ``__init__`` was not called by a derived class.
+
 Here is an example:
 
 .. code-block:: python
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -156,6 +156,31 @@ extern "C" inline PyObject *pybind11_meta_getattro(PyObject *obj, PyObject *name
 }
 #endif
 
+/// metaclass `__call__` function that is used to create all pybind11 objects.
+extern "C" inline PyObject *pybind11_meta_call(PyObject *type, PyObject *args, PyObject *kwargs) {
+
+    // use the default metaclass call to create/initialize the object
+    PyObject *self = PyType_Type.tp_call(type, args, kwargs);
+    if (self == nullptr) {
+        return nullptr;
+    }
+
+    // This must be a pybind11 instance
+    auto instance = reinterpret_cast<detail::instance *>(self);
+
+    // Ensure that the base __init__ function(s) were called
+    for (auto vh : values_and_holders(instance)) {
+        if (!vh.holder_constructed()) {
+            PyErr_Format(PyExc_TypeError, "%.200s.__init__() must be called when overriding __init__",
+                         vh.type->type->tp_name);
+            Py_DECREF(self);
+            return nullptr;
+        }
+    }
+
+    return self;
+}
+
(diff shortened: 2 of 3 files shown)


wjakob:
This change looks great, and I am happy to merge it. Performance impact should be minimal as you say. But can you please fix pep8? (See the failing "STYLE" test).

wjakob:
In particular, there is a long line with the error message string that should be split.

virtuald:
FIxed, sorry for the delay!

virtuald:
Ready for merge @wjakob , thanks!

virtuald:
Still ready. :)

virtuald:
Ping.

virtuald:
Ping? :(

wjakob:
Thanks @virtuald, and for the reminder @YannickJadoul.

virtuald:
@YannickJadoul the docs I added indirectly say what you want?

> The default pybind11 metaclass will throw a ``TypeError`` when it detects
   that ``__init__`` was not called by a derived class.

Implying that if you set your own metaclass then you'll get different behavior... ?

YannickJadoul:
> @YannickJadoul the docs I added indirectly say what you want?

OK, yes, that's a bit implicit, but good enough. Using custom metaclasses comes with a lot of caveats anyway, currently.

rwgk:
> It could also provide `__init__` without making a bare `pyspiel.Game` itself constructable:

Hi @YannickJadoul, my current solution is this:
```
    py::class_<Game, std::shared_ptr<Game>> game(m, "Game");
    game.def("num_distinct_actions", &Game::NumDistinctActions)
+      .def(py::init([](const std::string& game_string) {
+       return std::const_pointer_cast<Game>(LoadGame(game_string));
+      }))
```

There is a list of known/allowed names to pick from. It works with one I picked randomly. I asked for advice about creating a "bare" game, to use your word for it. Let's see what they say.

To have an `__init__` without args I could add an overload with a hard-wired `game_string`, e.g. based on their recommendation:
```
+      .def(py::init([]() {
+       return std::const_pointer_cast<Game>(LoadGame("backgammon"));
+      }))
```

Mostly out of curiosity, I also tried this:
```
+      .def(py::init([]() { return std::shared_ptr<Game>{}; }))
```
It builds fine but `pyspiel.Game()` segfaults straightaway, in the `__init__` call itself (not the destructor, I convinced myself). I don't know what dereferences the pointer and why.

YannickJadoul:
> It builds fine but `pyspiel.Game()` segfaults straightaway, in the `__init__` call itself (not the destructor, I convinced myself). I don't know what dereferences the pointer and why.

Nice, you bumped into a bug! :-D I managed to figure out what goes wrong; see #2430.

> To have an `__init__` without args I could add an overload with a hard-wired `game_string`, e.g. based on their recommendation:

Right, I understand. I was mostly trying to point out the trampoline class, which allows to add the `__init__` only when subclassing. So `pyspiel.Game` itself would still not be default-constructible, and it would still force the use of these factory methods.

virtuald:
Ha, I found a hacky way to determine if the constructor was overridden or not. I'll make a PR.

PR #2156, opened 2020-04-08
Improve finding python with cmake >= 3.12.0

felfert:
Hi,

With this change, during setup, cmake uses the new "Python" cmake module instead of the old "PythonInterp" **if** the cmake version is >= 3.12.0. This has the advantage of properly detecting python3 even on systems where both python2 and python3 are installed and python2 preceedes python3 in the PATH environment variable.

I have the above situation here, and therefore would appreciate if you could merge that. It also applies to v.2.5 without changes.

Cheers
 -Fritz

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -55,17 +55,27 @@ if(PYTHONLIBS_FOUND AND PYTHON_MODULE_EXTENSION)
     return()
 endif()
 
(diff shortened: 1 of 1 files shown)


felfert:
Well, it appears that cmake 3.12.4 was broken again (i'm currenly using cmake 3.17.0) So I close this for now, check the various versions and resubmit when I got the proper version determined.

PR #2158, opened 2020-04-09
shadowing parameter

Dekken:


jagerman:
This is such a common C++ pattern that I put the fault here on gcc for warning on this with `-Wshadow` enabled:
```C++
struct A {
    int x;
    A(int x) : x{x} {}
};
```
because this really *isn't* shadowing: it is entirely defined and unambiguous what the two `x`s are in that constructor, and is really common practice.  And indeed this *is* an open gcc bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147

It's quite sad because it means gcc's `-Wshadow` warning is pretty much useless.  Clang (correctly) understands that constructor arguments like this are not problematic at all, and doesn't `-Wshadow` warn on them unless you explicitly turn on an extra `-Wshadow-field-in-constructor`.

Until GCC fixes this warning bug the solution is probably just to consider gcc's `-Wshadow` useless.

wjakob:
Agreed, I don't think this warrants making changes in pybind11.

zchrissirhcz:
I think shadow variable checking should be enabled. It's a better coding style/convention.

Issue #1690, opened 2019-02-12
doc: Add debugging tips?


eacousineau:
Doing a quick search of "debugging" doesn't bring up too much in the `pybind11` docs.

For the repo that I work on, we've found it useful to document some debugging tips that are relevant to Python C extensions:
https://drake.mit.edu/python_bindings.html#debugging-with-the-python-bindings

Perhaps some mention of using `trace`, perhaps a small example with GDB would go a long ways?
For GDB, it'd be nice to show (a) debugging a module and (b) debugging a C++ binary with an embedded interpreter.
I find that for hacking directly on `pybind11` (in things like #1325 and #1146), it's uber useful to plop into Python via an embedded interpreter, rather than debugging through the `python` executable itsefl.

SuperShinyEyes:
I'd like to know how to debug the C++ code that are invoked by Python code in a similar fashion to pdb. Is there any canonical way of doing this?

jdumas:
Would love to see some examples as well. Currently I'm running some C++ code through `python`, a `std:runtime_error()` is thrown in C++, but when I type `bt` in `gdb` it shows `No stack`, so I'm a bit stuck on what to do.

EricCousineau-TRI:
Low-speed follow-up, but I added some basic stuff in our fork here:
https://github.com/RobotLocomotion/pybind11/pull/33#issuecomment-613626122

Will PR some of it here, but not sure where exactly to put the docs.

bstaletic:
Here's an idea. Why not put the debugging tips in the community-editable github wiki?

wjakob:
I've turned the Wiki on just now, let's see if people find it useful.

EricCousineau-TRI:
Hm... Concern about the Wiki is it's kinda unclear when to add to docs vs. Wiki; but yeah, I'll do a first pass on Wiki!

EricCousineau-TRI:
Made a first low-effort pass :sweat_smile: 
https://github.com/pybind/pybind11/wiki/Debugging-segfaults-and-hard-to-decipher-pybind11-bugs

PR #2165, opened 2020-04-14
test: Add `pytest.traced`, use unbuffered lines, and make `check-style` more strict

EricCousineau-TRI:
Towards #1690

See this comment: https://github.com/pybind/pybind11/issues/1690#issuecomment-613633656

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -230,7 +230,7 @@ else()
 endif()
 
 # A single command to compile and run the tests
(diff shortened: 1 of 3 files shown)


wjakob:
Hi @EricCousineau-TRI ,

quite a few details are missing for me to consider this PR (which should probably be split into seprate ones, one for each unrelated change). What does the change to `check-style.sh` accomplish (in which case is it stricter?) Note that there is an unresolved TODO in there. What about "traced" (there doesn't seem to be any usage of this new function). And what about the unbuffered `-u` flag? (We did not have any issues with this before, so why is it needed now?)

Thanks,
Wenzel

wjakob:
Ping? :)

PR #2175, opened 2020-04-20
pdoc3 integration

knarfS:
I wanted to document my embedded python module but on the one hand Sphinx was to complicated and didn't work for me and on the other hand pydoc wasn't able to generate a good looking documentation.
I found [pdoc3](https://pdoc3.github.io/pdoc/), which I was able to run from inside my embedded python interpreter and it was generating a [nice looking documentation for my project](https://knarfs.github.io/doc/smuview/0.0.4/python_bindings_api.html).

This PR contains three things:
1. An option to disable the list of the enum members in the docstring for the enum.
2. An optional [``__pdoc__`` module attribute](https://pdoc3.github.io/pdoc/doc/pdoc/#overriding-docstrings-with-__pdoc__) for the clean documentation of the enum members with pdoc3.
3. An initial chapter in the documentation about documenting with pdoc3.

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -304,3 +304,45 @@ the default settings are restored to prevent unwanted side effects.
 
 .. [#f4] http://www.sphinx-doc.org
 .. [#f5] http://github.com/pybind/python_example
+
+
+Generating documentation using pdoc3
+====================================
+
+pdoc3 [#f8]_ is an alternative for generate documentation without the need of a rather complex
+configuration like for Sphinx. It is especially usefull when documenting embedded python modules.
+
+It is recommended to enable the auto generation of function signaures and for a nice looking
+documentation of enums, you propably want to disable the ``enum_members_docstring`` option and
+enable the ``populate_enum_pdoc`` option:
+
+.. code-block:: cpp
+
+    PYBIND11_MODULE(example, m) {
+        py::options options;
+        options.enable_function_signatures();
+        options.disable_enum_members_docstring();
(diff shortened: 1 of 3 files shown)


wjakob:
I would say this is a little too fringe to me in terms of justify adding complexity to pybind11. It also doesn't seem like a good idea to me to specialize pybind11 for generating documentation for all sorts of different documentation generators (where will this end?). The same thing could be done in a much cleaner way with a few sprinkled lines of Python e.g. in `pdoc`, which could auto-detect that pybind11-bound module is being processed.

knarfS:
Thanks for your response @wjakob.

Would you consider to accept just commit 0925e91 in this or a similar form?
Background is that pybind11 is hard coding a list of all enum members into the docstring of the enum class itself. When any doc generator wants to document the enum (class) members by itself, all members would show up twice in the documentation.
It makes sense to let the doc generator do the job to make the enum members reference-able inside the documentation and to provide a consistent look of the documentation.

Please let me know what you think.

Regarding the `__pdoc__` attribute, I can do something like this:
```cpp
PYBIND11_EMBEDDED_MODULE(mymodule, m) {
  m.add_object("__pdoc__", py::dict(), false);

  py::enum_<MyEnum> my_enum(m, "MyEnum, "The is an enum of my types.");
  my_enum.value("A", MyEnum::A);
  m.attr("__pdoc__")["MyEnum.A"] = "This is the enum type A.";
  my_enum.value("B", MyEnum::B);
  m.attr("__pdoc__")["MyEnum.B"] = "This is the enum type B.";
}
```

PR #2176, opened 2020-04-21
Documentation Custom Type Caster: Change __init__(self) to __new__(cls)

orgarten:
```__init__(self)``` cannot return values. 

According to https://stackoverflow.com/questions/2491819/how-to-return-a-value-from-init-in-python ```__new__(cls)``` should be used, which works.

### DIFF
--- a/docs/advanced/cast/custom.rst
+++ b/docs/advanced/cast/custom.rst
@@ -23,7 +23,7 @@ The following Python snippet demonstrates the intended usage from the Python sid
 .. code-block:: python
 
     class A:
-        def __int__(self):
+        def __new__(cls):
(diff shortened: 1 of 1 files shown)


wjakob:
Good catch, thanks!

YannickJadoul:
Why was this changed/merged? The original code says `__int__`, not `__init__`.

wjakob:
Indeed! I was misled by the summary and thought that this is a no-brainer. Anyways, reverted.

orgarten:
Woah, thanks for pointing that out. That's embarrassing but explains why it didn't work for me either...Sorry for the confusion!

PR #2184, opened 2020-04-26
Support & Doc CMAKE_CXX_STANDARD

ax3l:
This is the 2nd commit of #1428 with added documentation. Thanks to @chuckatkins for providing this! (The first commit of the same PR is already in mainline.)

`CMAKE_CXX_STANDARD` is a CMake community standard to set the C++ standard of a build. Document it in favor of the previous variable, which stays as a legacy flag for existing projects.

https://cmake.org/cmake/help/v3.17/variable/CMAKE_CXX_STANDARD.html


wjakob:
Awesome, thanks!

ax3l:
My pleasure, thanks for merging!

Issue #2139, opened 2020-03-23
FindPythonLibsNew.cmake module is not suitable for cross-compilation environment


ahesford:
## Issue description

The `FindPythonLibsNew.cmake` module launches a Python interpreter to determine some key properties for building of `pybind11` and any dependent projects that may rely on cmake for configuration. In a cross-compilation environment, this causes at least three issues:

1. If target-architecture executables are not compatible with the host platform, the cmake module will either fail to execute the Python interpreter or will execute an interpreter on the host. Note that it is generally not possible to invoke the target Python interpreter in a cross-compilation environment because the target executable may not be compatible with the host platform.

1. When the host and target platforms have different pointer sizes and the cmake module attempts to run the Python interpreter from the host, `FindPythonLibsNew.cmake` may will fail with a `FATAL_ERROR` complaining that the bitness of Python and the chosen compiler do not agree if the module falls back to the host interpreter.

1. The variable `PYTHON_MODULE_EXTENSION` variable, set from `distutils.sysconfig.get_config_var('SO')` (note: it seems that the `SO` variable is deprecated and should be replaced with `EXT_SUFFIX`) and assigned to the `SUFFIX` target property in `pybind11Tools.cmake`, may be wrong when the extension of the host interpreter includes architecture information from the host and the host interpreter is invoked by the module.

If the host and target Python interpreters have compatible version numbers and configurations, all but the `PYTHON_SIZEOF_VOID_P` and `PYTHON_MODULE_EXTENSION` variables may be propertly determined from executing the host interpreter. Nevertheless, relying on executing the Python interpreter to determine key configuration values is always prone to error in cross-compilation environments.

## What is to be done?

Removing the dependency on `FindPythonLibsNew.cmake` and instead relying on `FindPython.cmake` that ships with cmake at least moves the Python configuration discovery problem outside of the pybind11 project back to cmake, where the incentive to support cross-platform and cross-compilation environments is much greater.

At a minimum, removing the `PYTHON_SIZEOF_VOID_P` check avoids the immediate problem. Allowing manual override of other variables would allow cmake projects that depend on pybind11 to force proper behavior in a cross-compilation environment. (I assume this is mostly an issue for packagers.)

Issue #1718, opened 2019-03-04
arm cross compilation issue when need to compile to different python


lironmo:
hellos,
i need to compile my app to different architecture and SBCs (like raspberry pi pi / banana pi /rock64/ etc)
and i see that even if i defined the python arm libs the code: find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED) ignore my flags and try to detect the interpreter by it's own and failed.

YannickJadoul:
This will probably get fixed by PR #2370 (in progress)? If not, please reopen.

henryiii:
Let's let #2370 close it. I'm happy to add to the list the get closed. ;)

When #2370 goes in, just use either the new Python support by finding Python yourself (using the new FindPython package in CMake) or use `PYTHON_NOPYTHON=ON` to disable the integration with Python entirely, which allows you to handle Python yourself.

Issue #1330, opened 2018-03-21
Cross compilation workflow


photex:
Howdy!

Say I have a cross-compiled python setup and want to build against that. Currently pybind appears to be tightly coupled to the system python in it's cmake configuration.

How could this be approached? I'm happy to work on this and submit a PR.

My use case is building native modules for `aarch64-linux-gnu` in a relatively controlled environment.

bstaletic:
A workaround would be to forget about Pybind11's cmake and just add the header directory  in your own cmake (or whatever).

Monomix:
Did you ever find a way to get this up and running? I'm in a similar situation: I'd like to build an extension for a 32-bit ARM target using as close to the standard workflow as possible, but am running into issues because the system python is 64-bit.

gocarlos:
No, i did not... right now im just using systemd and dbus to launch python scripts on the target board

mdenna-nviso:
Hi, I also have the same issue, I'm cross compiling for ARM on an intel machine, and pybind11 finds the Python installation for the host, so of course the build fails:

```
CMake Error at deps-base/pkg-pybind11/src/tools/FindPythonLibsNew.cmake:125 (message):
  Python config failure: Python is 64-bit, chosen compiler is 32-bit
Call Stack (most recent call first):
  deps-base/pkg-pybind11/src/tools/pybind11Tools.cmake:16 (find_package)
  deps-base/pkg-pybind11/src/CMakeLists.txt:33 (include)

```
Any indication on how to direct pybind11 to the python installation for the target?

Monomix:
I ended up just commenting out the bit-width check in FindPythonLibsNew.cmake - hacky, but the only solution I could find.

The order of inclusion of pybind11 and Python in CMake also has a few gotchas that need to be noted, see https://github.com/pybind/pybind11/blob/master/docs/faq.rst#user-content-inconsistent-detection-of-python-version-in-cmake-and-pybind11.

mdenna-nviso:
Hi, thanks for the above hint about inclusion order! 
Yes commentig out the check can work in some cases, but fails in others, as the target platform might have a different python version from the host or no python installation at all.

romansavrulin:
On build machine you can work-around this by doing

```
dpkg --add-architecture i386
apt install -y python2.7:i386
```
You maybe need to remove amd64 package of python too. 

[...]

PR #2196, opened 2020-04-30
WIP: Use CMake's own FindPythonLibs instead of vendored FindPythonLibsNew.

codypiersall:
This enables easy cross-compilation and pybind11 doesn't have to carry
around an old version of FindPythonLibs.

### DIFF
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -13,7 +13,12 @@ if(NOT PYBIND11_PYTHON_VERSION)
 endif()
 
 set(Python_ADDITIONAL_VERSIONS 3.9 3.8 3.7 3.6 3.5 3.4)
-find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
+# this minimum was determined by testing, and could potentially be made lower
+if(CMAKE_VERSION VERSION_LESS 3.10.2)
+    find_package(PythonLibsNew ${PYBIND11_PYTHON_VERSION} REQUIRED)
+else()
+    find_package(PythonLibs ${PYBIND11_PYTHON_VERSION} REQUIRED)
+endif()
 
 include(CheckCXXCompilerFlag)
 include(CMakeParseArguments)


codypiersall:
I think this PR could use some feedback -- I am not a CMake expert, and I worry the law of [Chesterton's Fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence) might apply here.  I'm sure there is a reason that the FindPythonLibsNew file was added, but I'm not sure what it is.

I tested that this enables cross compilation using an SDK generated with the Yocto project and still allows native compilation, but I haven't vetted this thoroughly yet.

codypiersall:
One thing to note is that this would likely bump the minimum CMake version required.  I have only tested it on CMake 3.10.3.

This addresses #2139 directly, and is the fix proposed in that issue.  Can you think of drawbacks @ahesford?

Also related to #1718 and #1330

wjakob:
Simply getting rid of the old approach to find Python definitely will not fly, since this library must continue to function on legacy systems with old versions of CMake. What might be an acceptable compromise is to decide on a CMake version where we deem `FindPythonLibs` sufficiently non-broken to automatically switch over the CMake-provided script.

codypiersall:
Makes sense -- I should be able to udpate this PR soon.  If CMake version 3.10 or greater, I'll use the system FindPythonLibs, and fall back to FindPythonLibsNew on older CMake.

codypiersall:
I force-pushed the change--now on newer CMake versions the version of PythonLibs shipped with CMake will be used, and the version of `PythonLibsNew` has not been touched.

wjakob:
Rather than just trying on one version and using that, think this will need a bit more investigative work on your part (or somebody else who is willing to help) -- understanding what was broken in the old version of the CMake-supplied scripts, and when it was fixed.

codypiersall:
Sounds like it's time to do a little spelunking for me.  Unfortunately the original commit message that adds the version to this repository is

> Include a more robust FindPythonLibs module for CMake

I'll go through CMake's source too to see if I can figure out why this was done, and what about it was more robust.

codypiersall:
From a bit of surfing the net, it looks like this is a pretty common library to be included in CMake-based projects.  For example, it's in [pyarrow](https://github.com/apache/arrow/blob/master/cpp/cmake_modules/FindPythonLibsNew.cmake), a [random repo from 6 years ago](https://gitlab.enseeiht.fr/abcd/abcd/blob/b904ab9d17f39c2bf2626f7d8c8972b5dc7a3fff/python/numpy/libs/numpy/cmake/FindPythonLibsNew.cmake), and other hits as well.  Looks like something that's been cop/pasted around for a while now.  Still trying to figure out where it *actually* came from and what problem it was solving that FindPythonLibs.cmake failed to solve.

[This PR](https://bitbucket.org/fenics-project/dolfin/pull-requests/29/more-robust-findpythonlibsnewcmake-script/diff) from years ago on a different project gives the only solid info I have found: 

> *The reason for FindPythonLibsNew.cmake is that **cmake's builtin FindPythonLibs pretty much never did what I wanted, and usually found the wrong libraries, had trouble finding them, etc.** This version gets the libraries for the python interpreter it found instead of doing an independent search, so you control which Python you want by configuring FindPythonInterp, and then FindPythonLibsNew will always set the libraries for that particular python. **This might make cross-compiling slightly more difficult,** and it doesn't handle 100% of cases with debug python builds, but it works way better for me.

(Emphasis mine)

The PR has a dead link to the original.

@dean0x7d, if you've got the time, do you remember why the original FindPythonLibsNew.cmake was added?  I saw that you were the one who originally added it.

codypiersall:
Also just want to point out that the version I used in this PR as it is right now has been [deprecated since 3.12](https://cmake.org/cmake/help/v3.17/module/FindPythonLibs.html)

ahesford:
Honestly, I don't know the right path here. I also am not a CMake expert. The `FindPython{,2,3}` modules suggested as a replacement do not seem to accomplish what's going on in the custom module as far as discovering properties of the Python interpreter. It's not clear to me how to define some of these properties without either hard-cording them for each target system or executing the target Python interpreter at build time.

I'm coming at this from the perspective of a Void Linux packaging, where packages can be cross-compiled for any supported architecture using the `xbps-src` build system. All official packages for all architectures are built on `x86` host systems, with 32-bit packages being built from an `i686` system image and 64-bit packages being built from an `x86_64` system. The value of `PYTHON_SIZEOF_VOID_P` should be correct in these instances, but the shared-object extension is guaranteed to be wrong for non-native packages.

Hard-coding the values is probably feasible (but undesirable) for supported architectures in Void, but I wonder if it's worthwhile. The CMake modules provide some convenience to dependent projects, but is otherwise expendible. I'll probably sidestep this issue by forgoing CMake modules in the package.

codypiersall:
@ahesford thanks for the comment.  I think one important thing I was missing was that the Python architecture does not have to match the host architecture.  When I say "host" here, I mean in terms of cross-compilation, so the _host_ is not the machine doing the build necessariliy, but the machine where the executables will run.  In my case my _build_ machine is x86-64 but the _host_ is 32-bit ARM.

It's very common in Windows to run a 32-bit Python executable in a 64-bit OS, and I suspect that reasons like that is why `FindPythonLibsNew` was made, possibly?  I'm really not sure.  And I'm unable to test it at the moment.

--------

Another strategy possibly worth considering is to only call `FindPythonLibs` instead of `FindPythonLibsNew` when we're cross-compiling, because in that situation `FindPythonLibsNew` can't possibly work correctly without doing some workarounds.

ahesford:
It doesn't look like `FindPythonLibs` offers everything `FindPythonLibsNew` defines, but the now-preferred `FindPython` seems to offer most things with different names. I think one of the biggest things missing would be `PYTHON_SIZEOF_VOID_P`, which is used to trigger a fatal error when bitness of the Python runtime and C++ compiler disagree. It's not clear to me that this check is critical at the CMake level. It's convenient to know during build whether there is a discrepancy that will produce broken extensions, but the failure mode is so obvious that any reasonable testing regimen would catch this issue.

It seems to me a reasonable course of action would be:
1. Use `FindPython` where possible as the preferred default,
2. Modify `FindPythonLibsNew` to:
  * Stop trying to discover variables that are not offered by `FindPython`, and
  * Set variable names that agree with those in `FindPython`

This makes `FindPythonLibsNew` nothing but a fallback for old CMake versions that can't support proper built-in discovery.

If there is a lot of momentum behind the existing variable names in `FindPythonLibsNew`, those names could be retained and a thin wrapper developed to prefer `FindPython` where possible but map its variables to those set by `FindPythonLibsNew`. (It's not clear how the "momentum" would be gauged, since their use depends on individual dependants.) I still think the pointer-size validator has to go to make cross-compilation work.

codypiersall:
@ahesford,

ack your last comment.  

The biggest problem with FindPythonLibsNew is that the way it finds out about the Python interpreter is by running it, which it certainly won't be able to do in a cross-compilation environment.  Regardless of anything else we do, when cross-compiling we can't do the check that is being done here:

https://github.com/pybind/pybind11/blob/a54eab92d265337996b8e4b4149d9176c2d428a6/tools/FindPythonLibsNew.cmake#L77-L92

Your suggestions makes sense to me.

codypiersall:
Also want to mention that my PR is totally broken still for the cross-compile case.

codypiersall:
My naive implementation here keeps failing in new and exciting ways; added WIP to the title to reflect this.

wjakob:
One thing to keep in mind is that cross-compilation is an incredibly fringe use case. Running the Python interpreter to query it for its properties is *the thing that one should do* in 99.9% of cases. When this project was started, FindPythonLibs didn't do that, causing a huge amount of pain, eventually motivating our custom solution here. I am not sure what it does now. We should be very careful to adapt any cmake-provided feature if it relies on other more fragile ways of finding the Python interpreter and associated include paths. Another point is that any changes to this component must continue to work for Python 2.7.

codypiersall:
I wouldn't call cross-compilation "incredibly fringe".  For embedded systems it's *very* normal, and "embedded" systems include dual core 1GHz ARM chips these days :-).  

But I definitely take your point that it is *much less common*, and that native compilation *must* continue working as well as it does today for everyone, and that if anything done in this PR messes up that the PR will not be accepted.

> Running the Python interpreter to query it for its properties is the thing that one should do in 99.9% of cases

Ack.  I need to dive in to CMake's source and figure out how they do things and see what the mismatch is.

codypiersall:
I won't be able to look into much for a couple weeks (in a sprint at work) but here's a list of stuff that I (or someone) need to do still:

1. Determine exactly in what scenarios the implementation of `FindPythonLibsNew` works better than the Cmake-provided `FindPythonLibs` or `FindPython`.  e.g. find a scenario where `FindPythonLibs` fails and `FindPythonLibsNew` does The Right Thing™.  I suspect this may have to do with things like automatically finding out if the Python being linked against is a debug build; this probably cannot be done simply without actually running the Python executable.
2. Consider using `FindPython` instead of `FindPythonLibs`.  `FindPythonLibs` has been deprecated since 3.12: https://cmake.org/cmake/help/v3.12/module/FindPythonLibs.html.  Or maybe just pick the appropriate one based on the running CMake version.
3. Ensure that the variables that need to get set are set correctly.  In particular, this PR fails to (at least) set `PYTHON_MODULE_EXTENSION`, which causes extensions to be built without exensions.
4. This PR still calls `FindPythonLibsNew` unconditionally in pybind11Config.cmake.in: https://github.com/pybind/pybind11/blob/master/tools/pybind11Config.cmake.in#L86.  I overlooked that initially.
5. *Possibly* consider using `FindPythonLibs` instead of `FindPythonLibsNew` only in the case of [cross-compilation](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html).

Number 1. on that list is probably the most important.

----------------

I'm using pybind11 in a Yocto-based project, and I wrote a Bitbake recipe that solves these problems for my use case, but it is not robust to be included upstream—for example, it hardcodes "`.so`" as the `PYTHON_MODULE_EXTENSION`

codypiersall:
I'm still interested in this in theory, but unfortunately I'm unable to work on this any time soon, and rather than just leaving it open I'm going to close it.  I've solved it locally for the project I'm working on (patched pybind11 in a bitbake recipe for my Yocto setup), but as I mentioned it's not a robust fix.

Thanks everyone who put time and energy into this issue!

ahesford:
For now, I'm packaging pybind11 for Void without the cmake modules. That gets around the cross-compilation issue for our use case, at the expense of breaking any downstream packages that might assume the pybind11 modules exist. For now, the only dependent in the Void repo is pyopencl, which I also maintain and builds just fine without cmake stuff.

If somebody ever comes complaining to me about missing cmake modules, I'll come poking around here to see if the issue has been revived or resolved. In the meantime, this doesn't seem to be a major blocker for anything I'm doing.

PR #2203, opened 2020-05-06
Replace deprecated find_package(PythonInterp)

kolibri91:
I upgraded `python` to 3.8 and I got problems by using `pybind11` v2.4.3.
It turns out that `pybind11` still creates the binaries for `python3.7` but my project tried to package them with `python3.8`.

It seems as the root cause of the problem are deprecated commands in _tools/FindPythonLibsNew.cmake_.

According https://cmake.org/cmake/help/git-stage/module/FindPythonInterp.html?highlight=pythoninterp:
- `find_package(PythonInterp)` is deprecated
- `find_package(Python COMPONENTS Interpreter)` should be used instead

I kept the whole logic the same, I just replaced the necessary variables with the new ones.

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -52,20 +52,23 @@
 
 # Checking for the extension makes sure that `LibsNew` was found and not just `Libs`.
 if(PYTHONLIBS_FOUND AND PYTHON_MODULE_EXTENSION)
-    return()
+  return()
 endif()
 
 # Use the Python interpreter to find the libs.
+if(NOT PythonLibsNew_FIND_VERSION)
+  set(PythonLibsNew_FIND_VERSION "")
+endif()
 if(PythonLibsNew_FIND_REQUIRED)
-    find_package(Python COMPONENTS Interpreter ${PythonLibsNew_FIND_VERSION} REQUIRED)
+  find_package(Python COMPONENTS Interpreter ${PythonLibsNew_FIND_VERSION} REQUIRED)
 else()
-    find_package(Python COMPONENTS Interpreter ${PythonLibsNew_FIND_VERSION})
+  find_package(Python COMPONENTS Interpreter ${PythonLibsNew_FIND_VERSION})
 endif()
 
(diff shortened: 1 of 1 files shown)


kolibri91:
@wjakob Please let me know, if I should merge those changes to another branch where the CI is running.

henryiii:
You have to require a very recent version of CMake to use find_package(Python). It should be optional, I believe. I would love to have this as an option, though, as the old version was *very* buggy, which is why PyBind11 patches it, and why CMake made a new find module and now responds to bug reports about the old one with 'use the new one'.

wjakob:
I would be open to this change, if implemented with the right checks wrt. CMake versions. Some users of pybind11 are still on ancient versions of CMake (2.8.x).

henryiii:
I would make it manually selectable, such as `set(PYBIND11_NEWPYTHON ON)` before accessing PyBind11, similar to the way FindBoost works. You don't want it based on CMake version only, as it works differently and you wouldn't want the behavior to change based on CMake version (at least, current CMake version).

I'd really love this, though, as the old FindPython* packages had a lot of bugs that CMake will not fix, due to the new version, so PyBind11 patches - which I've recently had break another other package that was also patching it too.

The new versions also fix things PyBind11 hasn't fixed, like conda environments, etc.

wjakob:
> The new versions also fix things PyBind11 hasn't fixed, like conda environments, etc.
That's good to now, and a reason for moving into this direction.

YannickJadoul:
@kolibri91, I think @henryiii is currently working on this in PR #2370. Is this PR still adding something else, or why did you reopen it (after closing it yourself) ?

kolibri91:
Sorry, the close happens by accident.

I will take a look into PR #2370 and check if its would solve my problem too.

YannickJadoul:
OK, no problem! Just throwing it out there, that lots of the CMake infrastructure is being worked on :-)

kolibri91:
Duplicate of #2370

YannickJadoul:
Thank you so much for the work you put in, though, @kolibri91! I'm sorry it somehow got superseded!

PR #2212, opened 2020-05-14
Docstring generation tweaks

sizmailov:
This PR contains a couple of trivial tweaks which make generated docstring a bit better. 

|  this PR | master | 
| ------------- |---------|
|  <pre>`a() -> numpy.ndarray[numpy.int32]`</pre> | <pre>`a() -> numpy.ndarray[int32]`</pre> | 
|  <pre>`b1(arg0: Iterator) -> None`<br>`b2(arg0: Iterable) -> None`</pre> | <pre>`b1(arg0: iterator) -> None`<br>`b2(arg0: iterable) -> None`</pre> | 
|  <pre>`c() -> numpy.ndarray`</pre> | <pre>`c() -> array`</pre> | 
|  <pre>`d() -> int`</pre> | <pre>`d() -> int_`</pre> | 

```C++
#include "pybind11/embed.h"
#include "pybind11/numpy.h"

namespace py = pybind11;

  std::array<int, 3> data = {1, 2, 3};
  m.def("a", [&]{ return py::array_t({3, 1}, data.data()); });
  m.def("b1", [](py::iterator&) { });
  m.def("b2", [](py::iterable&) { });
  m.def("c", []{ return py::array(); });
  m.def("d", []{ return py::int_(0); });
}

int main() {
  py::exec(R"(
      from example import *
      print(a.__doc__, end="")
      print(b1.__doc__, end="")
      print(b2.__doc__, end="")
      print(c.__doc__, end="")
      print(d.__doc__, end="")
)");
}
```
</details>

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1604,6 +1604,7 @@ template <typename base, typename deleter> struct is_holder_type<base, std::uniq
 
 template <typename T> struct handle_type_name { static constexpr auto name = _<T>(); };
 template <> struct handle_type_name<bytes> { static constexpr auto name = _(PYBIND11_BYTES_NAME); };
+template <> struct handle_type_name<int_> { static constexpr auto name = _("int"); };
 template <> struct handle_type_name<iterable> { static constexpr auto name = _("Iterable"); };
 template <> struct handle_type_name<iterator> { static constexpr auto name = _("Iterator"); };
 template <> struct handle_type_name<args> { static constexpr auto name = _("*args"); };
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1604,6 +1604,8 @@ template <typename base, typename deleter> struct is_holder_type<base, std::uniq
 
 template <typename T> struct handle_type_name { static constexpr auto name = _<T>(); };
 template <> struct handle_type_name<bytes> { static constexpr auto name = _(PYBIND11_BYTES_NAME); };
+template <> struct handle_type_name<iterable> { static constexpr auto name = _("Iterable"); };
+template <> struct handle_type_name<iterator> { static constexpr auto name = _("Iterator"); };
 template <> struct handle_type_name<args> { static constexpr auto name = _("*args"); };
 template <> struct handle_type_name<kwargs> { static constexpr auto name = _("**kwargs"); };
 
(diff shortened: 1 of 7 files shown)


sizmailov:
To justify intuition about low frequency of "not-immediate-return" `make_iterator` usage I've turned to github search API for `py::make_iterator`. Search is limited to 1000 files, so can't tell only for first 1000 found files (~5726 in total [[web search](https://github.com/search?l=C%2B%2B&q=py%3A%3Amake_iterator&type=Code)]).

In 1000 files there are 229 unique strings with `return py:make_iterator`. 
The only non-comment occurrence of `py::make_iterator` without preceding `return` was found in (pybind?) test code [here](https://github.com/chisuhua/graph/blob/bf82af49979297a1722832dc1468f4e7374862b8/tests/pybind/test_sequences_and_iterators.cpp#L510). This should also be no problem since cast should happen automatically. 

`py::make_map_iterator` search find only two unique strings with preceding return in another 1000 files (did a separate search). No other usages. Looks like this function is much less popular. 

While strictly speaking PR introduce breaking change it seems like it breaks no existing code. 

If you think this argument is not applicable and change to `py::make_iterator` should go to major pybind release (3.0), then we can move this part to a separate PR and left here two other non-conflicting commits.

rwgk:
Hi Sergei, I'm helping out reviewing PRs, to take some load off Wenzel. Could you please split this PR into 3? E.g. commits 1-4 stay in this PR, 5, 6 each get their separate PR. That will allow me and Wenzel to act quicker on 1-4, and give us separate contexts to discuss the points you're raising. I'll look again asap after the split.

sizmailov:
Hi, Ralf! Thanks for looking into that. I've split PR in three as you suggested.

include/pybind11/cast.h
```diff
@@ -1604,6 +1604,9 @@ template <typename base, typename deleter> struct is_holder_type<base, std::uniq
 
 template <typename T> struct handle_type_name { static constexpr auto name = _<T>(); };
 template <> struct handle_type_name<bytes> { static constexpr auto name = _(PYBIND11_BYTES_NAME); };
+template <> struct handle_type_name<int_> { static constexpr auto name = _("int"); };
+template <> struct handle_type_name<iterable> { static constexpr auto name = _("Iterable"); };
```
rwgk, line 1608: Is it feasible/meaningful to add tests for Iterable & Iterator in this PR?

sizmailov: done!

sizmailov:
fixed silenced flake8 spacing complaints

(travis.org seems to become quite unstable recently, probably migration to .com is required)

wjakob:
These all look good to me, thanks!

PR #2213, opened 2020-05-15
Add lvalue ref-qualified cpp_function constructors

clemenssielaff:
Hi there,

I tried and failed to bind a method with an explicit lvalue ref-qualifier (see [c++ reference](https://en.cppreference.com/w/cpp/language/member_functions)). I understand that we would not want to bind rvalue ref-qualified methods, but the lvalue counterparts should be safe..?

This PR adds two `cpp_function` constructor overloads for lvalue ref-qualified methods: one for const, the other for not.
I have also added a simple test method with an explicit lvalue ref-qualifier. The code would fail to compile without the additional constructors, now it compiles and works like it should (like a regular, non ref-qualified method).

Let me know what you think and thank you for all your hard work! pybind11 has been a joy to program against and it is used pretty much everywhere I worked.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -80,6 +80,8 @@ public:
     }
 
     /// Construct a cpp_function from a class method (non-const, lvalue ref-qualifier)
+    /// Is a copy of the overload for non-const functions without explicit ref-qualifier
+    /// but with an added `&`.
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     cpp_function(Return (Class::*f)(Arg...)&, const Extra&... extra) {
         initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(args...); },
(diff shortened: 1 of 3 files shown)


clemenssielaff:
Hi Ralf,

thank you for your reply! I have added the suggested comment and moved my test additions out into their own test class and -case.

Agreed that the code duplication is unfortunate, but I do not know of a better way that is shorter and/or more readable. I am open to suggestions though.

include/pybind11/pybind11.h
```diff
@@ -80,6 +80,8 @@ class cpp_function : public function {
     }
 
     /// Construct a cpp_function from a class method (non-const, lvalue ref-qualifier)
+    /// Is a copy of the overload for non-const functions without explicit ref-qualifier
```
rwgk, line 83: Maybe delete "Is" here and below?
// A copy of ...

clemenssielaff: ... and done :+1:

wjakob:
I wish there was a way to merge all of these constructors using templating but, alas, there is not AFAIK.

PR #2217, opened 2020-05-18
operators: Explicitly expose `py::hash(py::self)`

EricCousineau-TRI:
Add warnings about extending STL

When using `pybind11` for [Drake](https://drake.mit.edu), I found that I had to write:
```
cls.def(py::detail::hash(py::self));
```
instead of what the unittest wrote:
https://github.com/pybind/pybind11/blob/a54eab92d265337996b8e4b4149d9176c2d428a6/tests/test_operator_overloading.cpp#L110

This explicitly exposes the overload in the `pybind11` namespace.
Additionally, I was mislead into thinking that you may be able to overload `std::abs`. Per these docs, it does not appear to be valid (nor could I get it to work straight off the bat):
https://en.cppreference.com/w/cpp/language/extending_std

Follow-up to https://github.com/pybind/pybind11/pull/1034


EricCousineau-TRI:
@rwgk Thanks! However, it was still a tad hard for me to "add" in the new value b/c of whitespace changes.
Would it be OK if I instead curated my commits?
e.g. make minimal set of whitespace changes, no functional changes, and then add the functional change on top of that? Then bisecting / git history would be preserved!

rwgk:
Only what you feel is easy and reasonable. It's nice to minimize
distracting whitespace diffs but only if the effort to achieve that is
reasonable.

EricCousineau-TRI:
Took about 3min to split the commits apart. Lemme know what you think!

wjakob:
Looks great, thanks @EricCousineau-TRI  and @rwgk for the review!

PR #2218, opened 2020-05-18
cmake: Expose `PYBIND11_TEST_OVERRIDE`

EricCousineau-TRI:
Primarily for the ccmake curses GUI

I've found myself using this quite a bit, and would love to have it a bit more accessible.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -8,6 +8,7 @@
 cmake_minimum_required(VERSION 2.8.12)
 
 option(PYBIND11_WERROR  "Report all warnings as errors"  OFF)
+set(PYBIND11_TEST_OVERRIDE "" CACHE STRING "Tests from ;-separated list of *.cpp files will be built instead of all tests")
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
@rwgk I'm not sure if it would make sense, as `set(... CACHE ...)` / `option(...)` cache vars are pretty standard for CMake configuration.
https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry
https://cmake.org/cmake/help/latest/command/option.html

@jamiesnape If you have time, can I ask your opinion (on the CMake side)?

jamiesnape:
I do not think the CMake constructs need documenting, but the docstring in the `set(... CACHE)` could be improved, I think. Override tests is a little vague. Something like "tests from ;-separated list of *.cpp files will be built instead of all tests"?

EricCousineau-TRI:
Done - thanks!

wjakob:
LGTM, thanks!

Issue #1501, opened 2018-08-26
Use-after-free in object::memoryview for the format description


aguinet:
## Issue description

`object:memoryview` pass a temporary buffer for the format to `PyMemoryView_FromBuffer`. By reading the source code of this function, it does not "`strdup`" this format string, so it supposes it's always valid.

It is not really clear in the documentation how ownership works for the format. I figured this out because I had broken format issues with memory views in https://github.com/aguinet/dragonffi/ . A temporary fix by using `strdup` here https://github.com/aguinet/dragonffi/blob/master/third-party/pybind11/pytypes.h#L1255 fix the problem, but leaks memory.

## Reproducible example code

Hard to reproduce, as it's an use-after-free problem. Basically, creates a memoryview with a given format, and in some cases the format within the Python object will be wrong!

YannickJadoul:
I think this was fixed by #2223? If not, please reopen!

PR #2223, opened 2020-05-22
Fix undefined memoryview format

kyamagu:
Current `py::memoryview` constructor accepts `py::buffer_info`, and use `info.format.c_str()` as a format to PyBuffer object. However, this `char*` pointer is used after `info` goes out of stack, leading to undefined behavior.

This PR attempts to fix this, using static variable inside the constructor. `test_memoryview` calls the constructor two times with different format and check that the first memoryview is not affected by the second call.

Note the format strings might be replaced by instantiated `py::format_string<T>` types, but initializer becomes complicated.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1332,40 +1332,60 @@ public:
 
 class memoryview : public object {
 public:
-    explicit memoryview(const buffer_info& info) {
-        static Py_buffer buf { };
-        // Py_buffer uses signed sizes, strides and shape!..
-        static std::vector<Py_ssize_t> py_strides { };
-        static std::vector<Py_ssize_t> py_shape { };
-        static std::vector<std::string> formats = {
-            "?", "b", "B", "h", "H", "i", "I", "q", "Q", "f", "d", "g"
-        };
-        buf.buf = info.ptr;
-        buf.itemsize = info.itemsize;
-        auto it = std::find(formats.begin(), formats.end(), info.format);
-        buf.format = (it == formats.end()) ?
-            nullptr : const_cast<char*>(it->c_str());
-        buf.ndim = (int) info.ndim;
-        buf.len = info.size;
-        py_strides.clear();
-        py_shape.clear();
-        for (size_t i = 0; i < (size_t) info.ndim; ++i) {
-            py_strides.push_back(info.strides[i]);
-            py_shape.push_back(info.shape[i]);
-        }
-        buf.strides = py_strides.data();
-        buf.shape = py_shape.data();
-        buf.suboffsets = nullptr;
-        buf.readonly = info.readonly;
-        buf.internal = nullptr;
-
-        m_ptr = PyMemoryView_FromBuffer(&buf);
-        if (!m_ptr)
-            pybind11_fail("Unable to create memoryview from buffer descriptor");
-    }
-
     PYBIND11_OBJECT_CVT(memoryview, object, PyMemoryView_Check, PyMemoryView_FromObject)
+    explicit memoryview(char *mem, ssize_t size, bool writable = false)
+        : object(PyMemoryView_FromMemory(mem, size, (writable) ? PyBUF_WRITE : PyBUF_READ), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate memoryview object!");
+    }
+    explicit memoryview(const buffer_info& info);
 };
+
+inline memoryview::memoryview(const buffer_info& info) {
+    // TODO: two-letter formats are not supported.
+    static const char* formats[] = {
+        pybind11::format_descriptor<bool>::value,
+        pybind11::format_descriptor<int8_t>::value,
+        pybind11::format_descriptor<uint8_t>::value,
+        pybind11::format_descriptor<int16_t>::value,
+        pybind11::format_descriptor<uint16_t>::value,
+        pybind11::format_descriptor<int32_t>::value,
+        pybind11::format_descriptor<uint32_t>::value,
+        pybind11::format_descriptor<int64_t>::value,
+        pybind11::format_descriptor<uint64_t>::value,
+        pybind11::format_descriptor<float>::value,
+        pybind11::format_descriptor<double>::value,
+        pybind11::format_descriptor<long double>::value,
+    };
+    if (info.view()) {
+        // Note: PyMemoryView_FromBuffer never increments obj reference.
+        m_ptr = (info.view()->obj) ?
+            PyMemoryView_FromObject(info.view()->obj) :
+            PyMemoryView_FromBuffer(info.view());
+    }
+    else {
+        size_t length = sizeof(formats) / sizeof(char*);
+        auto format = std::find(formats, formats + length, info.format);
+        if (format == (formats + length))
+            pybind11_fail("Invalid format string");
+        std::vector<Py_ssize_t> shape(info.shape.begin(), info.shape.end());
+        std::vector<Py_ssize_t> strides(info.strides.begin(), info.strides.end());
+        Py_buffer view;
+        view.buf = info.ptr;
+        view.obj = nullptr;
+        view.len = info.size * info.itemsize;
+        view.readonly = info.readonly;
+        view.itemsize = info.itemsize;
+        view.format = const_cast<char*>(*format);
+        view.ndim = static_cast<int>(info.ndim);
+        view.shape = shape.data();
+        view.strides = strides.data();
+        view.suboffsets = nullptr;
+        view.internal = nullptr;
+        m_ptr = PyMemoryView_FromBuffer(&view);
+    }
+    if (!m_ptr)
+        pybind11_fail("Unable to create memoryview from buffer descriptor");
+}
 /// @} pytypes
 
 /// \addtogroup python_builtins
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -13,7 +13,6 @@
 #include "buffer_info.h"
 #include <utility>
 #include <type_traits>
-#include <algorithm>
 
 NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
@@ -1357,16 +1356,20 @@ public:
     /** \rst
         Creates ``memoryview`` from static buffer.
 
-        The caller is responsible for managing the lifetime of ``ptr`` and
-        ``format``. This method is meant for providing a ``memoryview`` for
-        C/C++ buffer not managed by Python.
+        This method is meant for providing a ``memoryview`` for C/C++ buffer not
+        managed by Python. The caller is responsible for managing the lifetime
+        of ``ptr`` and ``format``, which MUST outlive the memoryview constructed
+        here.
+
+        See also: Python C API documentation for `PyMemoryView_FromBuffer`_.
+
+        .. _PyMemoryView_FromBuffer: https://docs.python.org/c-api/memoryview.html#c.PyMemoryView_FromBuffer
 
         :param ptr: Pointer to the buffer.
         :param itemsize: Byte size of an element.
         :param format: Pointer to the null-terminated format string. For
             homogeneous Buffers, this should be set to
-            format_descriptor<T>::value.
-        :param ndim: Number of dimensions.
+            ``format_descriptor<T>::value``.
         :param shape_in: Shape of the tensor (1 entry per dimension).
         :param strides_in: Number of bytes between adjacent entries (for each
             per dimension).
@@ -1374,31 +1377,34 @@ public:
             written to.
      \endrst */
     static memoryview frombuffer(
-        void *ptr, ssize_t itemsize, const char* format, ssize_t ndim,
+        void *ptr, ssize_t itemsize, const char* format,
         detail::any_container<ssize_t> shape_in,
-        detail::any_container<ssize_t> strides_in, bool readonly=false);
+        detail::any_container<ssize_t> strides_in, bool readonly = false);
 
     template<typename T>
     static memoryview frombuffer(
         T *ptr, detail::any_container<ssize_t> shape_in,
-        detail::any_container<ssize_t> strides_in, bool readonly=false) {
+        detail::any_container<ssize_t> strides_in, bool readonly = false) {
         return memoryview::frombuffer(
             reinterpret_cast<void*>(ptr), sizeof(T),
-            format_descriptor<T>::value, static_cast<ssize_t>(shape_in->size()),
-            shape_in, strides_in, readonly);
+            format_descriptor<T>::value, shape_in, strides_in, readonly);
     }
 
 #if PY_MAJOR_VERSION >= 3
     /** \rst
         Creates ``memoryview`` from static memory.
 
-        The caller is responsible for managing the lifetime of ``mem``. This
-        constructor is meant for providing a ``memoryview`` for C/C++ buffer not
-        managed by Python.
+        This method is meant for providing a ``memoryview`` for C/C++ buffer not
+        managed by Python. The caller is responsible for managing the lifetime
+        of ``mem``, which MUST outlive the memoryview constructed here.
+
+        See also: Python C API documentation for `PyMemoryView_FromBuffer`_.
+
+        .. _PyMemoryView_FromMemory: https://docs.python.org/c-api/memoryview.html#c.PyMemoryView_FromMemory
      \endrst */
-    static memoryview frommemory(char *mem, ssize_t size, bool writable = false) {
+    static memoryview frommemory(char *mem, ssize_t size, bool readonly = false) {
         PyObject* ptr = PyMemoryView_FromMemory(
-            mem, size, (writable) ? PyBUF_WRITE : PyBUF_READ);
+            mem, size, (readonly) ? PyBUF_READ : PyBUF_WRITE);
         if (!ptr)
             pybind11_fail("Could not allocate memoryview object!");
         return memoryview(object(ptr, stolen_t{}));
@@ -1407,17 +1413,17 @@ public:
 };
 
 inline memoryview memoryview::frombuffer(
-    void *ptr, ssize_t itemsize, const char* format, ssize_t ndim,
+    void *ptr, ssize_t itemsize, const char* format,
     detail::any_container<ssize_t> shape_in,
     detail::any_container<ssize_t> strides_in, bool readonly) {
-    std::vector<Py_ssize_t> shape(std::move(shape_in));
-    std::vector<Py_ssize_t> strides(std::move(strides_in));
-    if (ndim != static_cast<ssize_t>(shape.size()) ||
-        ndim != static_cast<ssize_t>(strides.size()))
-        pybind11_fail("memoryview: ndim doesn't match shape and/or strides length");
+    size_t ndim = shape_in->size();
+    if (ndim < 1)
+        pybind11_fail("memoryview: shape length (ndim) must be greater than zero");
+    if (ndim != strides_in->size())
+        pybind11_fail("memoryview: shape length doesn't match strides length");
     ssize_t size = 1;
-    for (size_t i = 0; i < static_cast<size_t>(ndim); ++i)
-        size *= shape[i];
+    for (size_t i = 0; i < ndim; ++i)
+        size *= (*shape_in)[i];
     Py_buffer view;
     view.buf = ptr;
     view.obj = nullptr;
@@ -1426,8 +1432,8 @@ inline memoryview memoryview::frombuffer(
     view.itemsize = itemsize;
     view.format = const_cast<char*>(format);
     view.ndim = static_cast<int>(ndim);
-    view.shape = shape.data();
-    view.strides = strides.data();
+    view.shape = shape_in->data();
+    view.strides = strides_in->data();
     view.suboffsets = nullptr;
     view.internal = nullptr;
     PyObject* obj = PyMemoryView_FromBuffer(&view);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1377,14 +1377,34 @@ public:
             written to.
      \endrst */
     static memoryview frombuffer(
-        void *ptr, ssize_t itemsize, const char* format,
+        void *ptr, ssize_t itemsize, const char *format,
         detail::any_container<ssize_t> shape,
         detail::any_container<ssize_t> strides, bool readonly = false);
+
+    static memoryview frombuffer(
+        const void *ptr, ssize_t itemsize, const char *format,
+        detail::any_container<ssize_t> shape,
+        detail::any_container<ssize_t> strides) {
+        return memoryview::frombuffer(
+            const_cast<void*>(ptr), itemsize, format, shape, strides, true);
+    }
 
     template<typename T>
     static memoryview frombuffer(
         T *ptr, detail::any_container<ssize_t> shape,
-        detail::any_container<ssize_t> strides, bool readonly = false);
+        detail::any_container<ssize_t> strides, bool readonly = false) {
+        return memoryview::frombuffer(
+            reinterpret_cast<void*>(ptr), sizeof(T),
+            format_descriptor<T>::value, shape, strides, readonly);
+    }
+
+    template<typename T>
+    static memoryview frombuffer(
+        const T *ptr, detail::any_container<ssize_t> shape,
+        detail::any_container<ssize_t> strides) {
+        return memoryview::frombuffer(
+            const_cast<T*>(ptr), shape, strides, true);
+    }
 
 #if PY_MAJOR_VERSION >= 3
     /** \rst
@@ -1407,6 +1427,10 @@ public:
         if (!ptr)
             pybind11_fail("Could not allocate memoryview object!");
         return memoryview(object(ptr, stolen_t{}));
+    }
+
+    static memoryview frommemory(const void *mem, ssize_t size) {
+        return memoryview::frommemory(const_cast<void*>(mem), size, true);
     }
 #endif
 };
@@ -1440,15 +1464,6 @@ inline memoryview memoryview::frombuffer(
     return memoryview(object(obj, stolen_t{}));
 }
 #endif  // DOXYGEN_SHOULD_SKIP_THIS
-
-template<typename T>
-inline memoryview memoryview::frombuffer(
-    T *ptr, detail::any_container<ssize_t> shape,
-    detail::any_container<ssize_t> strides, bool readonly) {
-    return memoryview::frombuffer(
-        reinterpret_cast<void*>(ptr), sizeof(T),
-        format_descriptor<T>::value, shape, strides, readonly);
-}
 /// @} pytypes
 
 /// \addtogroup python_builtins
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1370,25 +1370,21 @@ public:
         :param format: Pointer to the null-terminated format string. For
             homogeneous Buffers, this should be set to
             ``format_descriptor<T>::value``.
-        :param shape_in: Shape of the tensor (1 entry per dimension).
-        :param strides_in: Number of bytes between adjacent entries (for each
+        :param shape: Shape of the tensor (1 entry per dimension).
+        :param strides: Number of bytes between adjacent entries (for each
             per dimension).
         :param readonly: Flag to indicate if the underlying storage may be
             written to.
      \endrst */
     static memoryview frombuffer(
         void *ptr, ssize_t itemsize, const char* format,
-        detail::any_container<ssize_t> shape_in,
-        detail::any_container<ssize_t> strides_in, bool readonly = false);
+        detail::any_container<ssize_t> shape,
+        detail::any_container<ssize_t> strides, bool readonly = false);
 
     template<typename T>
     static memoryview frombuffer(
-        T *ptr, detail::any_container<ssize_t> shape_in,
-        detail::any_container<ssize_t> strides_in, bool readonly = false) {
-        return memoryview::frombuffer(
-            reinterpret_cast<void*>(ptr), sizeof(T),
-            format_descriptor<T>::value, shape_in, strides_in, readonly);
-    }
+        T *ptr, detail::any_container<ssize_t> shape,
+        detail::any_container<ssize_t> strides, bool readonly = false);
 
 #if PY_MAJOR_VERSION >= 3
     /** \rst
@@ -1397,6 +1393,8 @@ public:
         This method is meant for providing a ``memoryview`` for C/C++ buffer not
         managed by Python. The caller is responsible for managing the lifetime
         of ``mem``, which MUST outlive the memoryview constructed here.
+
+        This method is not available in Python 2.
 
         See also: Python C API documentation for `PyMemoryView_FromBuffer`_.
 
@@ -1412,18 +1410,17 @@ public:
 #endif
 };
 
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
 inline memoryview memoryview::frombuffer(
     void *ptr, ssize_t itemsize, const char* format,
-    detail::any_container<ssize_t> shape_in,
-    detail::any_container<ssize_t> strides_in, bool readonly) {
-    size_t ndim = shape_in->size();
-    if (ndim < 1)
-        pybind11_fail("memoryview: shape length (ndim) must be greater than zero");
-    if (ndim != strides_in->size())
+    detail::any_container<ssize_t> shape,
+    detail::any_container<ssize_t> strides, bool readonly) {
+    size_t ndim = shape->size();
+    if (ndim != strides->size())
         pybind11_fail("memoryview: shape length doesn't match strides length");
-    ssize_t size = 1;
+    ssize_t size = ndim ? 1 : 0;
     for (size_t i = 0; i < ndim; ++i)
-        size *= (*shape_in)[i];
+        size *= (*shape)[i];
     Py_buffer view;
     view.buf = ptr;
     view.obj = nullptr;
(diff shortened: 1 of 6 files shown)


include/pybind11/pytypes.h
```diff
@@ -1336,9 +1337,14 @@ class memoryview : public object {
         // Py_buffer uses signed sizes, strides and shape!..
         static std::vector<Py_ssize_t> py_strides { };
```
rwgk, line 1338: @aldanor 
Question for the original author, who introduced this line (in a different file, later moved here):
https://github.com/pybind/pybind11/commit/a7e62e1ca621c7a7c1b8956873a16c255f2939b6

Why are these std::vector objects static?
Just as an optimization? To minimize the allocation overhead?
Did you consider how this plays with threads? (I don't know if it's possible that multiple threads could get here at the same time, but it's a worry if you have static arrays like here.)

include/pybind11/pytypes.h
```diff
@@ -1336,9 +1337,14 @@ class memoryview : public object {
...
         static std::vector<Py_ssize_t> py_shape { };
+        static std::vector<std::string> formats = {
```
rwgk, line 1340: This is not future proof: if codes are added in future Python versions, this function will fail in highly unobvious and confusing ways.
https://github.com/python/cpython/blob/master/Objects/memoryobject.c#L1162
I see formats, after curation there, can have two characters. Is that a concern here?

include/pybind11/pytypes.h
```diff
@@ -1336,9 +1337,14 @@ class memoryview : public object {
...
+            "?", "b", "B", "h", "H", "i", "I", "q", "Q", "f", "d", "g"
+        };
         buf.buf = info.ptr;
         buf.itemsize = info.itemsize;
-        buf.format = const_cast<char *>(info.format.c_str());
+        auto it = std::find(formats.begin(), formats.end(), info.format);
```
rwgk, line 1345: I someone accidentally mistypes a format character, that error will be silently ignored here and the wrong buf.format (nullptr) assigned.

tests/test_pytypes.py
```diff
@@ -261,3 +261,13 @@ def test_number_protocol():
 def test_list_slicing():
     li = list(range(100))
     assert li[::2] == m.test_list_slicing(li)
+
+
+def test_memoryview():
+    import array
+    view = m.test_memoryview(b'abc')
+    if sys.version_info[0] == 3:
+        # Python 2.7 array does not implement the new buffer protocol
```
rwgk, line 270: This seems to contradict the documentation:

https://docs.python.org/2/c-api/buffer.html

Quoting: "The new buffer API has been backported to Python 2.6, and the memoryview object has been backported to Python 2.7. It is strongly advised to use them rather than the old APIs, unless you are blocked from doing so for compatibility reasons."

Maybe you need a more if condition here?

kyamagu: `array.array` does not offer the new buffer interface, and throws a TypeError.

The documentation states:

> Two examples of objects that support the buffer interface are strings and arrays.
> An array can only expose its contents via the old-style buffer interface. This limitation does not apply to Python 3, where memoryview objects can be constructed from arrays, too.

Unless pybind11 supports the old style buffer interface, it is impossible to accept `array.array` in python 2.7.

include/pybind11/pytypes.h
kyamagu: What is the policy in pybind11 in this situation? Throwing an exception to handle this?

include/pybind11/pytypes.h
kyamagu: Understood. This should perhaps support unknown format that may appear in future. Currently there seems only `@` can appear in the format as a prefix, so the length is at most 2.

The main problem here is that `const char*` must reside after `py::buffer_info` goes out of scope, and this constructor needs to take care of this. As cpython's `get_native_fmtstr ` is not public, there needs a list of supported formats somewhere inside `pybind11`. Naive approach is to have a static set of strings that caches a given string, but this won't have any format validation.

Relevant: https://bugs.python.org/issue15821

include/pybind11/pytypes.h
```diff
@@ -1336,9 +1337,14 @@ class memoryview : public object {
         // Py_buffer uses signed sizes, strides and shape!..
         static std::vector<Py_ssize_t> py_strides { };
         static std::vector<Py_ssize_t> py_shape { };
+        static std::vector<std::string> formats = {
+            "?", "b", "B", "h", "H", "i", "I", "q", "Q", "f", "d", "g"
+        };
```
aldanor, line 1342: Hm, wondering if this could be just `static char* format` given they are all length-1?

aldanor: Ah, apparently it can be length 2 as noted above; well, could probably use something like a `static const char* format[]`?

rwgk:
The code you're trying to fix has a far more serious bug: it doesn't keep a Python reference to the object it's providing a view for (`buf.obj` is not set). One indication is `sys.getrefcount(a)` before and after you create the memoryview view with your test_memoryview() function. It's unchanged. If you use the built-in memoryview the reference count increases by 1. Another way to show the issue is to use weakref, e.g. I tried this with Python 3.6:
```
    l = [38, -84, 15]                                                           
    a = array.array('i', l)                                                     
    w = weakref.ref(a)                                                          
    if 0:  # SWITCH                                                                      
      mv = m.test_memoryview(a)                                                 
    else:                                                                       
      mv = memoryview(a)                                                        
    def check(mv):                                                              
      list(mv) == l                                                             
      print('PASS')                                                             
    check(mv)                                                                   
    del a                                                                       
    print('w():',  w())                                                         
    check(mv)                                                                   
```
I ran this with the MSAN memory sanitizer.
With SWITCH 0 I got:
```
  PASS
  w(): array('i', [38, -84, 15])
  PASS
```
With SWITCH 1:
```
  PASS
  w(): None
  ==46858==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x5635170d106a in PyLong_FromLong third_party/python_runtime/v3_6/Objects/longobject.c:239:5
    #1 0x563517157602 in unpack_single third_party/python_runtime/v3_6/Objects/memoryobject.c:1701:12
    #2 0x563517156120 in memory_item third_party/python_runtime/v3_6/Objects/memoryobject.c:2249:16
    #3 0x563516e1cddb in PySequence_GetItem third_party/python_runtime/v3_6/Objects/abstract.c:1641:16
    #4 0x56351709a613 in iter_iternext third_party/python_runtime/v3_6/Objects/iterobject.c:63:14
    #5 0x5635170a794c in listextend third_party/python_runtime/v3_6/Objects/listobject.c:854:26
    #6 0x5635170b1a97 in list_init third_party/python_runtime/v3_6/Objects/listobject.c:2313:24
    #7 0x563517264069 in type_call third_party/python_runtime/v3_6/Objects/typeobject.c:915:19
    #8 0x563516e46e2f in _PyObject_FastCallDict third_party/python_runtime/v3_6/Objects/abstract.c:2331:18
    #9 0x563516e49e3a in _PyObject_FastCallKeywords third_party/python_runtime/v3_6/Objects/abstract.c:2496:14
    #10 0x5635177de676 in call_function third_party/python_runtime/v3_6/Python/ceval.c:4861:17
    ... truncated ...
```

NOTE:
To get around the issue you identified, I locally patched the code to simply leak the format string:
```
--- a/google3/third_party/pybind11/include/pybind11/pytypes.h
+++ b/google3/third_party/pybind11/include/pybind11/pytypes.h
@@ -1333,7 +1333,8 @@ public:
         static std::vector<Py_ssize_t> py_shape { };
         buf.buf = info.ptr;
         buf.itemsize = info.itemsize;
-        buf.format = const_cast<char *>(info.format.c_str());
+        std::string* format_leaking = new std::string{info.format};
+        buf.format = const_cast<char *>(format_leaking->c_str());
         buf.ndim = (int) info.ndim;
         buf.len = info.size;
         py_strides.clear();
```

I think the bigger issue needs to be fixed first. There isn't much gain from fixing just the issue with the format handling; maybe the bigger fix resolves it as a sideeffect.

@aldanor Did you have a reason for not using `info.view`, with `Py_INCREF(info.view)` before passing it to  `PyMemoryView_FromBuffer(info.view)`?

kyamagu:
Great, the fundamental issue is that the current implementation is not following the expected exporter behavior:
https://docs.python.org/3/c-api/buffer.html#buffer-protocol

I will leave this PR open at this moment

rwgk:
Hi @kyamagu, thanks for pointing out the issue. I'll reach out to @aldanor to discuss how we can fix the issues.

aldanor:
@rwgk Apologies, been a bit busy so only got to replying now. Tbh I don't remember much about this, I think it was needed to make something else work (I was working mostly on making structured numpy dtypes work in pybind at the moment) so not enough attention was paid to this.

All of the points above sound about right, all this static variable business in both `memoryview` and `buffer_info` should go away (except for maybe static format strings which can be just stack allocated), and the refcounts have to be carefully respected. It would also be nice to add a few refcount tests like the ones outlined above by @rwgk?

include/pybind11/pytypes.h
wjakob: I looked briefly at this and don't understand it either. AFAIK the danger is that PyMemoryView_FromBuffer expects the `Py_buffer` to remain valid after the call. (Actually, the precise semantics of this function seem to be a source of confusion in the wider Python community: https://bugs.python.org/issue15821). 

But making the `std::vector` static here won't address this issue. It merely ensures that the pointers survive the return of the function, but then they'll be garbled by the next function call. I think that this whole function needs a fundamental redesign.

wjakob:
related: #1501

wjakob:
This may end up needing a fancier construction such as a weak reference that waits for the memoryview to expire before it `free()`s the `Py_buffer*`.

wjakob:
@kyamagu, do you plan to do further work on this PR?

kyamagu:
@wjakob I am not working on this issue at this moment, as the refcount issue seems to need a fix first. Should I go further and revise this PR?

wjakob:
Both of these are really the same issue, as far as I can see.

kyamagu:
Alright, look into this

rwgk:
Hi Kota, thanks a lot for taking on this fix! I already looked at your code a little bit and at first sight it looks good. I'm planning to repeat my test from May 23. It's a holiday weekend here, it may take me a few days, but I'll try asap.

tests/test_pytypes.py
```diff
@@ -261,3 +261,24 @@ def test_number_protocol():
 def test_list_slicing():
     li = list(range(100))
     assert li[::2] == m.test_list_slicing(li)
+
+
+@pytest.mark.parametrize('method, args, format', [
+    (m.test_memoryview_fromobject, (b'abc', ), 'B'),
+    (m.test_memoryview_frombuffer_reference, (b'abc',), 'B'),
+    (m.test_memoryview_frombuffer_new, tuple(), 'b'),
```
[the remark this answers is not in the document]
rwgk: Just quick FYI: it's simply a signed vs unsigned char platform-dependent issue. I started working on a slightly more thorough test, using the opportunity to exercise different format descriptors. I'll send it when I'm finished (asap).

tests/test_pytypes.cpp
```diff
@@ -307,4 +307,21 @@ TEST_SUBMODULE(pytypes, m) {
...
+
+    m.def("test_memoryview_frombuffer_reference", [](py::buffer b) {
+        return py::memoryview(b.request());
+    });
+
+    m.def("test_memoryview_frombuffer_new", []() {
+        const char* buf = "abc";
+        const char* buf2 = "\x00\x00\x00\x00";
+        auto mv = py::memoryview(py::buffer_info(buf, 3, 1));
+        // Call twice with a different buffer to check the view content.
```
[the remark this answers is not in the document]
kyamagu: Ok, this test was to check if any field is not overwritten by multiple calls for the `static std::vector` issue in the original implementation. Now that static variables are only used in format string, and maybe not necessary. Will remove this.

rwgk:
Hi Kota, below are modified versions of your tests. That code passes in the Google environment with Python 2.7 & 3.7 & ASAN & MSAN, and under standard Debian with 2.7 & 3.6.
I'm still looking at a detail of your implementation.

`test_pytypes.cpp`:
```
    m.def("test_memoryview_frombuffer_new", [](bool unsigned_if_true) {         
          static const int16_t si16[] = { 3, 1, 4, 1, 5 };                        
          static const uint16_t ui16[] = { 2, 7, 1, 8 };                          
          auto smv = py::memoryview(py::buffer_info(si16, 5, true));              
          auto umv = py::memoryview(py::buffer_info(ui16, 4, true));              
          return unsigned_if_true ? umv : smv;                                    
    });                                                                         
```

`test_pytypes.py`:
```
@pytest.mark.parametrize('method, arg, fmt, expected_view', [                   
    (m.test_memoryview_fromobject, b'red', 'B', b'red'),                        
    (m.test_memoryview_frombuffer_reference, b'green', 'B', b'green'),          
    (m.test_memoryview_frombuffer_new, False, 'h', [3, 1, 4, 1, 5]),            
    (m.test_memoryview_frombuffer_new, True, 'H', [2, 7, 1, 8]),                
])                                                                              
def test_memoryview(method, arg, fmt, expected_view):                           
    view = method(arg)                                                          
    assert isinstance(view, memoryview)                                         
    assert view.format == fmt                                                   
    if isinstance(expected_view, bytes) or sys.version_info[0] >= 3:            
        view_as_list = list(view)                                                 
    else:                                                                       
        # Using max to pick non-zero byte (big-endian vs little-endian).          
        view_as_list = [max([ord(c) for c in s]) for s in view]                   
    assert view_as_list == list(expected_view)                                  
                                                                                
                                                                                
@pytest.mark.parametrize('method', [                                            
    m.test_memoryview_fromobject,                                               
    m.test_memoryview_frombuffer_reference,                                     
])                                                                              
def test_memoryview_refcount(method):                                           
    buf = b'\x0a\x0b\x0c\x0d'                                                   
    ref_before = sys.getrefcount(buf)                                           
    view = method(buf)                                                          
    ref_after = sys.getrefcount(buf)                                            
    assert ref_before < ref_after                                               
    assert list(view) == list(buf)                                              
```

include/pybind11/pytypes.h
```diff
@@ -1331,35 +1332,62 @@ class buffer : public object {
...
+        : object(PyMemoryView_FromMemory(mem, size, (writable) ? PyBUF_WRITE : PyBUF_READ), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate memoryview object!");
+    }
+#endif
+    explicit memoryview(const buffer_info& info);
 };
+
+inline memoryview::memoryview(const buffer_info& info) {
+    // TODO: two-letter formats are not supported.
+    static const char* formats[] = {
```
rwgk, line 1347: Observations, I have to defer to @wjakob  judgement for what to do (if anything):

Here the compiler has to work very hard just to put together the array of char pointers:
```static const char* formats[] = {"?", "b", "B", "h", "H", "i", "I", "q", "Q", "f", "d", "g"};```
That's basically the same as the hard-coded string in `detail/common.h`:
```static constexpr const char c = "?bBhHiIqQfdg"[...];```

It's only a subset of the format strings accepted by the current Python implementation:

https://github.com/python/cpython/blob/e51dd9dad6590bf3a940723fbbaaf4f64a3c9228/Objects/memoryobject.c#L1150

Very unfortunately, this authoritative Python function is hidden away (`static`).
My own choice would be to copy the function into pybind11 (probably with modifications), with a long comment to explain why. I'd also carefully check every released Python version to see if the list evolved over time, to track that in our copy, if necessary.

Alternatively, if we decide we only want the subset of format strings, I'd do a slight refactor of `detail/common.h` to make the list available without the very involved detour through the C++ template engine.

**Comment by kyamagu** [comment]
@rwgk Oops, just working on a different part now, will take your comment above into

**Comment by rwgk** [comment]
Thanks Kota! Just today we started a new way to maintain pybind11, making team-based decisions. I'll consult with the other maintainers regarding my "Observations" above, to see what they think.

**Comment by wjakob** [comment]
I took a brief look: the array of standard formats is a bit on the verbose side — the more serious problem is that it will not work for more complex dtypes that are very important to some users. My suggestion (to be discussed/tested if that actually works) would be something like the following:

1. strdup the format string
2. create a capsule that will free the string, when it is freed
3. create a tuple with 2 entries: the first is the capsule, the second is the owner (`Py_buffer::obj`) of the provided buffer or `None`
4. specify the tuple as `obj` of the newly created Py_buffer.

The owner reference will then both keep the capsule and the actual owner of the data region alive until deallocation.

**Comment by kyamagu** [comment]
We can skip the case when the owner is not `nullptr`, because when owner is not null, we can use `PyMemoryView_FromObject` instead of `PyMemoryView_FromBuffer` to let the reference object manage the format string. The problem arise only when there is no owner. Maybe no tuple is necessary.

FYI, CPython internally defines ManagedBuffer Object (`_PyManagedBufferObject `) to manage lifetime of format string when needed (`PyMemoryView_GetContiguous` implementation).
https://github.com/python/cpython/blob/master/Objects/memoryobject.c

Alternatively, changing the type of `buffer_info:: format` from `std::string` to `const char*` would resolve the problem, but that breaks the API backward compatibility and also ask the caller to manage the lifetime of the format string.

**Comment by rwgk** [comment]
Great to learn that we don't need to build the tuple!
I strongly feel it's best to solve the problem in pytypes.h, instead of pushing it back to users.
Offline I asked Wenzel if we could just build a Python bytes object to store the string, but he thinks we may not have a guarantee that the string isn't moved around (maybe by PyPy). The capsule idea seems pretty straightforward, too. Do you want to give it a try? I'll retest quickly.

**Comment by kyamagu** [comment]
@rwgk Ok, I will try making a capsule implementation. Might take some time

**Comment by kyamagu** [comment]
Hmm, it seems impossible to go with the capsule approach. `Py_Buffer::obj` is always rewritten to NULL inside `PyMemoryView_FromBuffer` implementation, and is never reference counted:
https://github.com/python/cpython/blob/master/Objects/memoryobject.c#L764

Therefore, the following immediately segfaults.

```cpp
char* format = reinterpret_cast<char*>(PyMem_Malloc(strlen(info.format.c_str()) + 1));
if (!format)
    pybind11_fail("Could not allocate format string!");
strcpy(format, info.format.c_str());
capsule obj(format, &PyMem_Free);
std::vector<Py_ssize_t> shape(info.shape.begin(), info.shape.end());
std::vector<Py_ssize_t> strides(info.strides.begin(), info.strides.end());
Py_buffer view;
view.buf = info.ptr;
view.obj = obj;
view.len = info.size * info.itemsize;
view.readonly = info.readonly;
view.itemsize = info.itemsize;
view.format = format;
view.ndim = static_cast<int>(info.ndim);
view.shape = shape.data();
view.strides = strides.data();
view.suboffsets = nullptr;
view.internal = nullptr;
m_ptr = PyMemoryView_FromBuffer(&view);
```

If we really need to manage lifetime of the format string, perhaps we need to give up using `PyMemoryView_FromBuffer` and instead use `PyMemoryView_FromObject` with custom object supporting buffer protocol and destructor to release format string. [`_PyManagedBufferObject`](https://github.com/python/cpython/blob/master/Objects/memoryobject.c#L26) in CPython has the functionality but unfortunately is private. Will check if this approach is feasible.

rwgk:
Hi Kota, your observation indeed matches the documentation:

https://docs.python.org/3/c-api/buffer.html#c.Py_buffer
> As a special case, for temporary buffers that are wrapped by PyMemoryView_FromBuffer() or PyBuffer_FillInfo() this field is NULL. In general, exporting objects MUST NOT use this scheme.

I feel very uneasy about essentially emulating the _PyManagedBufferObject functionality just to get around the problem with the format string.

While looking around I saw that the Py_buffer documentation refers to the struct module for the format syntax:

> const char *format
> A NUL terminated string in struct module style syntax describing the contents of a single item.

According to the struct module documentation the format syntax is very stable.
Between Python 2.7 and 3.8, just three characters were added (n, N in 3.3 and e in 3.6).
That seems very easy to track, and in the worst case if we miss an addition it's a super easy fix.
While the table we need may look significant at first sight, it translates to very little machine code and it's a very easy job for the compiler to process it. On balance I think that's the way to go.
Below is a super compact solution that I think will do the trick (it compiles and runs standalone).

I'll ask Wenzel to take another look here.

```
#include <stdio.h>

static inline const char *
get_native_fmtstr(const char *fmt)
{
    // Table as found in Python 3.8 struct module documentation.
    // Note: More characters than in Python 3.8 memoryobject.c table.
    static const char packed[] =
        "@x\0" "@c\0" "@b\0" "@B\0" "@?\0" "@h\0" "@H\0" "@i\0" "@I\0"
        "@l\0" "@L\0" "@q\0" "@Q\0" "@n\0" "@N\0" "@e\0" "@f\0" "@d\0"
        "@s\0" "@p\0" "@P";

    int at = 0;
    if (fmt[0] == '@') {
        at = 1;
        fmt++;
    }
    if (fmt[0] == '\0' || fmt[1] != '\0') {
        return NULL;
    }
    for (int i = 1; i < sizeof(packed); i += 3) {
        if (packed[i] == fmt[0]) {
            return &packed[at ? i - 1 : i];
        }
    }
    return NULL;
}

int main(int argc, const char* argv[]) {
  if (argc > 1) {
      const char* fmt_str = get_native_fmtstr(argv[1]);
      if (fmt_str) {
          printf("FOUND: %s\n", fmt_str);
      }
  }
  return 0;
}
```

rwgk:
Hi Kota, do you know, is the get_native_fmtstr always enforced in memoryobject.c, in the code path we're calling from the code we're working on? Or is it possible to pass other (arbitrary?) strings?

kyamagu:
@rwgk As far as I see, `get_native_fmtstr` is only used inside `memory_cast`, which is called when `memoryview.cast` in python side: 
https://github.com/python/cpython/blob/master/Objects/memoryobject.c

So, probably there is no code path that calls `get_native_fmtstr` when creating a new memoryview.

One other thing is that the old python 2.7 `memoryview` implementation differs a lot from python 3.3+ implementation. We need to be careful not relying too much on the internal of CPython implementation.
https://github.com/python/cpython/blob/2.7/Objects/memoryobject.c

rwgk:
Very good, thanks a lot Kota!
Although, that brings us back to wanting to manage the lifetime of the `char *` pointee ... but I'm getting more confused than clear unfortunately: if `view.obj` isn't used (I consider this certain now), what ensures that the `view.buf` pointee is valid for the lifetime of the `memoryview` we're creating?

I'm beginning to think the `PyMemoryView_FromBuffer()` interface is only safe for `view.buf` pointees at a static address for the lifetime of the process. Does that make sense?

Looking again at the 3.8 documentation:

> PyObject *PyMemoryView_FromMemory(char *mem, Py_ssize_t size, int flags)
> Return value: New reference.
> Create a memoryview object using mem as the underlying buffer. flags can be one of PyBUF_READ or PyBUF_WRITE.
> 
> New in version 3.3.
> 
> PyObject *PyMemoryView_FromBuffer(Py_buffer *view)
> Return value: New reference.
> Create a memoryview object wrapping the given buffer structure view. For simple byte buffers, PyMemoryView_FromMemory() is the preferred function.

That reinforces my belief: if the "preferred function" clearly only take a `char *` pointer, `mem` must live for the lifetime of the process, because clearly the created `memoryview` has no way of knowing if `mem` was freed.

If that still makes sense: What do we have to do to not build a trap? Maybe instead of trying to build a `memoryview`, throw an exception to explain? Maybe point to a new small section in the docs to be helpful?

kyamagu:
> the created memoryview has no way of knowing if mem was freed.

Yes, I think that's the scenario when there is no obj. This is the case when the user is trying to provide a `memoryview` for a C/C++ buffer that is not managed by python. In this case, the user is responsible for the lifetime of the buffer. This should be clearly documented. And, in this case, it is probably reasonable to also ask the user to manage the lifetime of format string too, by allocating `const char*` somewhere (needs `buffer_info` to keep `const char*` instead of `std::string` for this).

When there is obj set, Python is always managing the lifetime of both buffer and format.

rwgk:
Hi Kota, sorry for the delay in responding. It's great that we now understand the situation. I agree now, both `view.buf` and `view.format` are the user's responsibility; `view.buf` out of necessity, `view.format` because it's just a minor extra requirement considering the necessity for `view.buf`.

What still bothers me: it is quite a trap, even if well documented, because the C++ interface doesn't suggest there could be danger. On the contrary, when I see `memoryview` I associate that we the promise that everything is cleanly managed, automatically. We are basically hiding the danger behind `pybind11::buffer`, `pybind11::buffer_info`, and `pybind11::memoryview`. To not build a trap, my first choice would be to throw an exception where we're now calling `PyMemoryView_FromBuffer`, pointing users to a different interface designed to make the behavior obvious. For example: `object static_memoryview(...)`. For that we could document that `buf` and `format` must live for the duration of the process; anything else is inherently unsafe, although of course users can still take risks if they want to, but I wouldn't even suggest that in the documentation. Essentially, it could just be a thin wrapper around `PyMemoryView_FromBuffer`, with an overload that emulates Python >3.3 `PyMemoryView_FromMemory`. — What do you think?

kyamagu:
@rwgk Okay, splitting the `pybind11::memoryview` constructor makes sense. I think the confusion comes partly from the mixed use of `pybind11::buffer_info` that were created from different sources:

1. Those created by `pybind11::request()`: this buffer_info holds reference to `Py_Buffer` managed by python.
2. All other cases: `pybind11::buffer_info` is constructed without `Py_Buffer::obj`, to provide a view over C/C++ buffer.

You're suggesting to prohibit the second scenario to call the current `memoryview` constructor, right?

```c++
explicit memoryview(const buffer_info& info)
```

For the second scenario, I think we should skip taking `buffer_info` at all, because the current buffer_info definition heavily relies on `std::string` to manage format string. This might lead to nearly-duplicate constructor APIs, but can avoid confusion between `std::string` and `const char*`. So, perhaps something like the following signatures to add:

```c++
memoryview(void *ptr, ssize_t itemsize, const char* format, ssize_t ndim,
    detail::any_container<ssize_t> shape_in, detail::any_container<ssize_t> strides_in, bool readonly=false)

template<typename T>
memoryview(T *ptr, detail::any_container<ssize_t> shape_in, detail::any_container<ssize_t> strides_in, bool readonly=false
```

rwgk:
Hi Kota, yes, that's very close that what I was thinking, too. I really like your interface. The only question is:
A. regular constructor
B. static member function of `memoryview` (factory function inside the class)
C. free factory function
All will work technically, with the two signatures you're suggesting. B. & C. have two advantages (IMO):
1. with good naming we can alert code authors (and readers) to the fact that they are responsible for managing the lifetimes of `ptr`, even if they don't look at the documentation.
2. it's possible to pin-point usages without sophisticated tooling: simply search for e.g. `memoryview::static_pointee` (if we choose B.).
If we choose A., it's difficult to pin-point the user-managed cases. Note that I work at Google, I'm looking at an ocean of source code. Easy discoverability can make a huge difference for projects involving large scale changes.

kyamagu:
@rwgk Ok, I've rewritten the constructor and introduce static factory methods `frombuffer` and `frommemory`, named after CPython API. There might be better naming, but this should be aligned to pybind11 coding style.

include/pybind11/pytypes.h
```diff
@@ -1331,35 +1332,109 @@ class buffer : public object {
...
+            PyMemoryView_FromBuffer(info.view());
         if (!m_ptr)
             pybind11_fail("Unable to create memoryview from buffer descriptor");
     }
 
-    PYBIND11_OBJECT_CVT(memoryview, object, PyMemoryView_Check, PyMemoryView_FromObject)
+    /** \rst
+        Creates ``memoryview`` from static buffer.
+
+        The caller is responsible for managing the lifetime of ``ptr`` and
```
rwgk, line 1360: Small suggestion: swapping sentences ("what" first, then "how"), adding an emphasis, and pointing to relevant Python C API (is there a better way for that?):

> This method is meant for providing a ``memoryview`` for C/C++ buffer not managed by Python. The caller is responsible for managing the lifetime of ``ptr`` and  ``format``, which MUST outlive the ``memoryview`` constructed here.
> 
> See also: Python C API documentation for `PyMemoryView_FromBuffer`.
> 

Similarly below for `frommemory`.

include/pybind11/pytypes.h
```diff
@@ -1331,35 +1332,109 @@ class buffer : public object {
...
+#endif
 };
+
+inline memoryview memoryview::frombuffer(
+    void *ptr, ssize_t itemsize, const char* format, ssize_t ndim,
+    detail::any_container<ssize_t> shape_in,
+    detail::any_container<ssize_t> strides_in, bool readonly) {
+    std::vector<Py_ssize_t> shape(std::move(shape_in));
+    std::vector<Py_ssize_t> strides(std::move(strides_in));
+    if (ndim != static_cast<ssize_t>(shape.size()) ||
```
rwgk, line 1415: Could this be moved to the top, with `shape_in->size()` and `shape_out->size()`?
Could you just use e.g. `shape_in->data` below instead of copying/moving into your own `std::vector`?
`any_container` contains a `std::vector` already and has `const std::vector<T> *operator->() const`.

include/pybind11/pytypes.h
```diff
@@ -1331,35 +1332,109 @@ class buffer : public object {
...
+        ndim != static_cast<ssize_t>(strides.size()))
+        pybind11_fail("memoryview: ndim doesn't match shape and/or strides length");
+    ssize_t size = 1;
```
rwgk, line 1418: Wondering about the corner case that `ndim == 0`. In that case `size` comes out as `1`, but I'm guessing should be `0`?
Might be prudent to unit test this corner case including valgrind or MSAN test to ensure nothing bad happens downstream; I'm guessing it'll be fine, but if not it would be best to add a `pybind11_fail` here.

include/pybind11/pytypes.h
kyamagu: Ok. Maybe we can add a hyperlink

https://docs.python.org/c-api/memoryview.html#c.PyMemoryView_FromBuffer
https://docs.python.org/c-api/memoryview.html#c.PyMemoryView_FromMemory

include/pybind11/pytypes.h
kyamagu: Ok

include/pybind11/pytypes.h
kyamagu: Sure, will fix

kyamagu:
@rwgk Hi, the new commit adds tests for failure cases and an additional fix for PyPI tests.

There is still an error for docs generation; the current version of breathe + sphinx + doxygen toolchain cannot correctly handle inlined `memoryview::frombuffer` definition and produces warnings. It is still possible to build docs by removing `-W` option from sphinx, but Travis CI fails due to this warning. Do you know any workaround?

rwgk:
Thanks Kota! I'll prioritize retesting with MSAN tomorrow. I'll also alert
the other maintainers to the doc gen issue.

wjakob:
Thanks Kota, I would try to see whether #if !defined(DOXYGEN) helps to get the CI to pass. Of course that means that the function won’t be documented.

Speaking of that, we don’t have any documentation for this class. Especially given the tricky template definitions, would you be able to say a few words about typical use cases somewhere in this section? -> https://pybind11.readthedocs.io/en/stable/advanced/pycpp/index.html

rwgk:
Hi Kota, I re-did full MSAN/ASAN testing and didn't find any problems :-)

I made a few small suggested changes. The complete diffs as tested are here (is there a better way to share that with you)?

https://drive.google.com/file/d/1M2iABgJ-9htyaNXj2iN4wLmQ_wqnSl9v/view?usp=sharing

Please feel free to pick-and-choose.

I experimented with moving the `memoryview::frombuffer` outside the class scope, as a potential workaround for the doc generation issue, and it builds & runs fine. I don't know if it actually helps with the doc generation. Even if it does, the code is a little more noisy. I don't know what's the higher value, less noisy code, or complete documentation.

I also experimented allowing ndim == 0. MSAN/ASAN don't report issues with Python 2 & 3. The Python 2 behavior is a bit weird (see test), but Python 3 is completely sane. Since that's the future my vote is to allow ndim == 0. An additional thought is that a shape element could be 0, which boils down to the same situation, a pointer to a 0-size array. We're also not guarding against that, which I think is good.

For test_memoryview, I undid my change turning `args` into `arg`, to avoid the `py::none unused` argument for the new function you added.

kyamagu:
Thanks, @rwgk 's suggestion makes sense, `ndim=0` seems a valid case. Will prepare another commit.

> is there a better way to share that with you?

Maybe you can directly edit in this PR? I have never used this functionality before, though.
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

@wjakob Alright, I'll check the doc

rwgk:
> Maybe you can directly edit in this PR?

Ah, that's interesting, thanks, I'll keep that in mind for next time. I think we're nearly done polishing this PR!

kyamagu:
Ok, I've figured out a workaround to docs generation using a macro to skip lines: https://www.doxygen.nl/manual/faq.html#faq_code

In addition, I've added const variants of `frombuffer` and `frommemory`. Let me know if that is too much.

include/pybind11/pytypes.h
```diff
@@ -1331,35 +1331,139 @@ class buffer : public object {
...
+    view.itemsize = itemsize;
+    view.format = const_cast<char*>(format);
+    view.ndim = static_cast<int>(ndim);
+    view.shape = shape->data();
+    view.strides = strides->data();
+    view.suboffsets = nullptr;
+    view.internal = nullptr;
+    PyObject* obj = PyMemoryView_FromBuffer(&view);
+    if (!obj)
+        pybind11_fail("Unable to create memoryview from buffer structure");
```
rwgk, line 1463: Sorry I forgot to mention last time:
1. I tried this with ptr == nullptr (ndim == 0) and that fails here. Adding that to the unit tests would be great.
2. Maybe use `throw error_already_set()`? To not clobber the original error generated by Python?

kyamagu: Ok, that sounds great

kyamagu:
Oops, py27 doesn't throw when nullptr given, will change the test

wjakob:
This looks great, you have my blessing. One last minor request: in converting from SnakeCase to underscore_case, you didn't add underscores, can we have those? E.g.: `FromBuffer` -> `from_buffer`, etc.

rwgk:
@kyamagu Merged! Thanks a lot for the great work!

PR #2228, opened 2020-05-25
Improve the documentation of partitionning/extending extension modules.

jblespiau:
Why? I have been looking for this for hours, and even though I had read
the section, I had not understood it, as "everything 'just' works" was
not explicit enough for me.

I have seen a similar question in the Gitter chat, with a great example.

This proves that I was not alone, and I feel this is an advanced topic
that could benefit from being a little more expicit.

What? Thus, this explicits the problem with referecing variables
defined in another scope, and this also gives a full example
with the different imports.

Hopefuly, the will save some time to some people.

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -110,9 +110,10 @@ Partitioning code over multiple extension modules
 
 It's straightforward to split binding code over multiple extension modules,
 while referencing types that are declared elsewhere. Everything "just" works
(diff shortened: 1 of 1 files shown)


wjakob:
Sorry, I'm disinclined towards this PR -- the current coverage is IMO well-written, and the new material added is basically a kind of monkey-patching that seems quite dangerous. I don't think the pybind11 manual should be advertising such design patterns.

PR #2233, opened 2020-05-28
Fix CI, prepare test on Python 3.9 beta

henryiii:
* Fixed broken Sphinx job in CI by pinning Sphinx < 3. We can possibly fix this later on the latest Breathe + Sphinx.
* Enabled NumPy on the 3.8 test
* Added new 3.9 Beta test, expected fail for now.
* Dropped the old PyPy 5.8 test because it was broken, added two PyPy 7.3 tests (2.7 and 3.6); these do (expected) fail, but we should be able to start fixing the tests, I hope. The 2.7 mode looks close (3-4 test failures), the 3.6 mode fails to build.

### DIFF
--- a/include/pybind11/operators.h
+++ b/include/pybind11/operators.h
@@ -157,7 +157,7 @@ PYBIND11_UNARY_OPERATOR(float,      float_,       (double) l)
 #undef PYBIND11_BINARY_OPERATOR
 #undef PYBIND11_INPLACE_OPERATOR
 #undef PYBIND11_UNARY_OPERATOR
-NAMESPACE_END(detail);
+NAMESPACE_END(detail)
 
 using detail::self;
 
--- a/include/pybind11/operators.h
+++ b/include/pybind11/operators.h
@@ -157,7 +157,7 @@ PYBIND11_UNARY_OPERATOR(float,      float_,       (double) l)
 #undef PYBIND11_BINARY_OPERATOR
 #undef PYBIND11_INPLACE_OPERATOR
 #undef PYBIND11_UNARY_OPERATOR
-NAMESPACE_END(detail)
+NAMESPACE_END(detail);
 
 using detail::self;
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -514,7 +514,6 @@ private:
         return cache;
     }
 
-private:
     handle obj;
     key_type key;
     mutable object cache;
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -514,6 +514,7 @@ private:
         return cache;
     }
 
+private:
     handle obj;
     key_type key;
     mutable object cache;
--- a/tests/test_stl_binders.py
+++ b/tests/test_stl_binders.py
@@ -67,6 +67,7 @@ def test_vector_int():
     v_int2.clear()
     assert len(v_int2) == 0
 
+
 # related to the PyPy's buffer protocol.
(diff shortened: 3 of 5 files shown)


tests/test_stl_binders.py
```diff
@@ -67,6 +67,7 @@ def test_vector_int():
     v_int2.clear()
     assert len(v_int2) == 0
 
+
```
rwgk, line 70: Accident?

include/pybind11/pytypes.h
```diff
@@ -514,7 +514,6 @@ class accessor : public object_api<accessor<Policy>> {
         return cache;
     }
 
-private:
```
rwgk, line 517: Accident?
If not, could you please add a comment why the members below cannot be private anymore?

include/pybind11/operators.h
```diff
@@ -157,7 +157,7 @@ PYBIND11_UNARY_OPERATOR(float,      float_,       (double) l)
 #undef PYBIND11_BINARY_OPERATOR
 #undef PYBIND11_INPLACE_OPERATOR
 #undef PYBIND11_UNARY_OPERATOR
-NAMESPACE_END(detail)
+NAMESPACE_END(detail);
```
rwgk, line 160: Accident?
If not, could you add please a comment why the semicolon is needed here?

tests/test_stl_binders.py
henryiii: Yes. Still in progress, playing with fixing PyPy. Will fix!

include/pybind11/pytypes.h
henryiii: This was a duplicated private, there's one just 10 lines above or so. However, it was removed to try to fix sphinx+breathe errors, which I finally fixed by forcing and old sphinx and old breathe.

include/pybind11/operators.h
henryiii: This removed one sphinx+breathe error; Doxygen isn't guessing the correct nature of NAMESPACE_END, and I wasn't able to get it to expand this. Finally gave up and reverted to and old Sphinx+Breathe, which works (or at least doesn't spit out warning then error out).

henryiii: (I will revert these changes after I get a new PyPy working or find a way to get the old one to download correctly)

henryiii:
This *should* have fixed Doxygen, but did not:

```python
breathe_doxygen_config_options = {
    'ENABLE_PREPROCESSING'   : 'YES',
    'MACRO_EXPANSION'        : 'YES',
}
```

henryiii:
Okay, I think this is ready for review. Changes:
* Fixed broken Sphinx job in CI by pinning Sphinx < 3. We can possibly fix this later on the latest Breathe + Sphinx.
* Enabled NumPy on the 3.8 test
* Added new 3.9 test
* Dropped the old PyPy 5.8 test because it was broken, added two PyPy 7.3 tests (2.7 and 3.6); these do (expected) fail, but we should be able to start fixing the tests, I hope. The 2.7 mode looks close (3-4 test failures), the 3.6 mode fails to build.

tests/test_stl_binders.py
henryiii: Actually, no, it was a flake8 request.

wjakob:
This looks fantastic! Thank you very much @henryiii. I'm also glad that you updated PyPy. The 5.x release was super-flaky and sometimes caused test failures, maybe the new version will address this.

Issue #2180, opened 2020-04-23
Build failed with Visual Studio 2017 and CUDA 10.2


SimeonEhrig:
## Issue description

The nvcc compiler crashes at a minimal example, when I compile with VS 2017 and CUDA 10.2 on Windows 10.

```bash
cmake -G"Visual Studio 15 2017 Win64" ..
-- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.18363.
-- The CXX compiler identification is MSVC 19.16.27039.0
-- The CUDA compiler identification is NVIDIA 10.2.89
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check for working CUDA compiler: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.2/bin/nvcc.exe
-- Check for working CUDA compiler: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.2/bin/nvcc.exe -- works
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Found PythonInterp: C:/Users/user/miniconda3/envs/pybind11_conda/python.exe (found suitable version "3.8.2", minimum required is "3.6")
-- Found PythonLibs: C:/Users/user/miniconda3/envs/pybind11_conda/libs/python38.lib (found suitable version "3.8.2", minimum required is "3.6")
-- Found PythonInterp: C:/Users/user/miniconda3/envs/pybind11_conda/python.exe (found version "3.8.2")
-- Found PythonLibs: C:/Users/user/miniconda3/envs/pybind11_conda/libs/Python38.lib
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/user/project/simple_cuda/build_conda
(pybind11_conda) PS C:\Users\user\project\simple_cuda\build_conda> cmake --build .
Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
[...]

PR #2240, opened 2020-06-03
Fix compiler error with MSVC 17 and CUDA 10.2

SimeonEhrig:
Fixes the compiler error described in this issue: #2180

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1498,7 +1498,9 @@ public:
     }
 
     explicit operator type*() { return this->value; }
-    explicit operator type&() { return *(this->value); }
+    // static_cast works around compiler error with MSVC 17 and CUDA 10.2
+    // see issue #2180
+    explicit operator type&() { return *(static_cast<type *>(this->value)); }
     explicit operator holder_type*() { return std::addressof(holder); }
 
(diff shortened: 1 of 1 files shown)


atilag:
I can confirm that solved the problem (we were facing the same exact issue).
Would love to see this merged :)

wjakob:
I'll merge this once you've had a chance to add an extra comment. (As we do in a few other places).

Side comment: compiling pybind11 code with nvcc sounds really scary (even if it's just for the "host" part).

SimeonEhrig:
I added the comment.

> Side comment: compiling pybind11 code with nvcc sounds really scary (even if it's just for the "host" part).

In our case, pybind11 is not the scary part to compile ;-) We use a highly [templated C++ library](https://github.com/alpaka-group/alpaka/) which also supports gcc, clang and icc as host compiler for nvcc.

wjakob:
Thanks!

SimeonEhrig:
> Thank you for digging into this and finding a work-around, Simeon!
> 
> I co-reported your reproducer as Nvidia Bug 3013788 so they have a chance to get their splitter/frontend into order for upcoming CUDA releases.

Thanks for reporting the bug.

PR #2241, opened 2020-06-03
[common.h] Mark entry point as "unused".

tkoeppe:
This change defines a new, portable macro `PYBIND11_MAYBE_UNUSED` to mark declarations as unused, and annotates the `PYBIND11_MODULE` entry point with this attribute.

The purpose of this annotation is to facilitate dead code detection, which might otherwise consider the module entry point function dead, since it isn't otherwise used. (It is only used via FFI.)

The actual definition of the GCC/Clang attribute is copied from Abseil, https://github.com/abseil/abseil-cpp/blob/da3a87690c56f965705b6a233d25ba5a3294067c/absl/base/attributes.h#L551.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -92,6 +92,14 @@
 #  define PYBIND11_DEPRECATED(reason) __attribute__((deprecated(reason)))
 #endif
 
+#if defined(PYBIND11_CPP17)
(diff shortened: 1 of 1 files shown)


tkoeppe:
Apologies for the force-pushes, I received some feedback offline. The PR should be stable now, everyone who asked me to make it agrees with its present form.

wjakob:
Looks great, thanks!

PR #2244, opened 2020-06-05
Render typed iterators (e.g. Iterator[int]) in docstrings

sizmailov:
This PR makes pybind11 generate much more informative typed iterators.

| this PR |
| ------------- |
| <pre>`__iter__(self: example.VectorInt) -> Iterator[int]`<br>`__iter__(self: example.MapTupleToComplex) -> Iterator[Tuple[int, str]]`<br>`items(self: example.MapTupleToComplex) -> Iterator[Tuple[Tuple[int, str], complex]]`</pre> |

| master |
| ------------- |
| <pre>`__iter__(self: example.VectorInt) -> iterator`<br>`__iter__(self: example.MapTupleToComplex) -> iterator`<br>`items(self: example.MapTupleToComplex) -> iterator`</pre> | 

Strictly speaking its a potential breaking change but seems like no existing code is affected, see below.

This PR makes `py::make_iterator` and `py::make_key_iterator` return `py::detail::iterator_state` instance instead of `py::iterator`.
It doesn't affect regular use of those functions (immediate return from `__iter__` lambda),
but requires changes in user code with implicit assignments/conversions, e.g.:

```c++
// next line is ok now, but will break with this PR
// fix on user side: apply py::cast() explicitly
py::iterator it = make_iterator(...); 
```

**Search for code that would break:**
To justify intuition about low frequency of "not-immediate-return" make_iterator usage I've turned to github search API for py::make_iterator. Search is limited to 1000 files, so can't tell only for first 1000 found files (~5726 in total [[web search](https://github.com/search?l=C%2B%2B&q=py%3A%3Amake_iterator&type=Code)]).

In 1000 files there are 229 unique strings with return `py:make_iterator`.
The only non-comment occurrence of `py::make_iterator` without preceding return was found in (pybind?) test code [here](https://github.com/chisuhua/graph/blob/bf82af49979297a1722832dc1468f4e7374862b8/tests/pybind/test_sequences_and_iterators.cpp#L510). This should also be no problem since cast should happen automatically.

`py::make_map_iterator` search find only two unique strings with preceding return in another 1000 files (did a separate search). No other usages. Looks like this function is much less popular.

While strictly speaking a breaking change was introduced it seems like it breaks no existing code.
If you think this argument is not applicable and change to `py::make_iterator` should go to major pybind release (e.g. 3.0).

```C++
#include <map>
#include <complex>

#include "pybind11/embed.h" 

namespace py = pybind11;

using map_tuple_to_complex = std::map<std::tuple<int,std::string>, std::complex<double>>;
PYBIND11_MAKE_OPAQUE(std::vector<int>);
PYBIND11_MAKE_OPAQUE(map_tuple_to_complex);

  py::bind_map<map_tuple_to_complex>(m, "MapTupleToComplex");
}

int main() {
  py::exec(R"(
      from example import *
      print(VectorInt.__iter__.__doc__, end="")
      print(MapTupleToComplex.__iter__.__doc__, end="")
      print(MapTupleToComplex.items.__doc__, end="")
)");
}
```
</details>

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1680,6 +1680,23 @@ struct iterator_state {
     bool first_or_done;
 };
 
+template<typename Iterator, typename Sentinel, return_value_policy Policy>
+struct type_caster<iterator_state<Iterator, Sentinel, false, Policy>> :
+    type_caster_base<iterator_state<Iterator, Sentinel, false, Policy>> {
+    using ValueType = decltype(*std::declval<Iterator>());
+public:
+    static constexpr auto name = _("Iterator[") + make_caster<ValueType>::name + _("]");
+};
+
+template<typename Iterator, typename Sentinel, return_value_policy Policy>
+struct type_caster<iterator_state<Iterator, Sentinel, true, Policy>> :
+    type_caster_base<iterator_state<Iterator, Sentinel, true, Policy>> {
+    using ValueType = decltype((*std::declval<Iterator>()).first);
+public:
+    static constexpr auto name = _("Iterator[") + make_caster<ValueType>::name + _("]");
+};
+
+
 NAMESPACE_END(detail)
 
 /// Makes a python iterator from a first and past-the-end C++ InputIterator.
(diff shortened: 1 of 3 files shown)


wjakob:
I'm against merging this -- first, because of the incompatibility, and because it adds a bunch of code for what is a relative niche feature. This is generally a dangerous rabbit hole: we could increase the complexity of pybind11 to generate near-perfect type annotations, but I don't think that is in the interest of the majority of the users of this project. My goal post is: do an okay job in a minimal fashion.

wjakob:
(Side note: sorry for rejecting it after you've clearly put quite a bit of work into the PR and explanations above).

PR #2246, opened 2020-06-08
[common.h] Mark another entry point as "unused".

tkoeppe:
For rationale, see #2241, eeb1044818af5b70761deae602c49eba439164dc; there is a second entry point function defined by the `PYBIND11_MODULE` macro that also needs to be annotated as unused.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -185,9 +185,10 @@
 #define PYBIND11_STR_TYPE ::pybind11::str
 #define PYBIND11_BOOL_ATTR "__bool__"
 #define PYBIND11_NB_BOOL(ptr) ((ptr)->nb_bool)
-// Providing a separate declaration to make Clang's -Wmissing-prototypes happy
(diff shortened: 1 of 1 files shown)


tkoeppe:
Hello! Sorry, during #2241 I had missed the fact that there were two distinct declarations, `pybind11_init_foo` and `PyInit_foo`, and I only annoated one, but forgot this other one.

I only see the results of the dead-code analysis offline after some time (i.e. I can't test it live), so I missed this piece.

tkoeppe:
(And could you maybe restart that one broken Travis test? The failure seems unrelated. And the AppVeyor tests seem to be stuck getting started?)

tkoeppe:
@rwgk: I had also missed the change for the Py2 path, could you kindly re-review?

wjakob:
Thanks!

tkoeppe:
Likewise! I can finally confirm that this is effective, all our uses of pybind are now no longer considered "dead". Thanks a lot!

Issue #1665, opened 2019-01-11
Feature request: numpy csr_matrix without Eigen


Dekken:
I've gotten [something working](https://github.com/Dekken/statick/blob/master/example.cpp#L47)

[built like](https://github.com/Dekken/statick/blob/master/numpy.h#L413)

[used like](https://github.com/Dekken/statick/blob/master/example.py)

"check_" is forced to pass, so it's not ideal but it all seems to work anyway. I don't know what the conventions are here but I hope something like this could be merged eventually.

Ref:https://github.com/pybind/pybind11/issues/1022

Dekken:
if someone could advise me how to construct the proxy objects for this that's be great

Dekken:
@wjakob @jagerman @anyone

Dekken:
[...](https://i.giphy.com/media/VqYyKYygtFvz2/giphy.webp)

eloyfelix:
this would indeed be a nice feature. @Dekken did you manage to do it? Your links don't work anymore.

Dekken:
I have it working, I moved some things around.

https://github.com/Dekken/statick/blob/master/inc/statick/pybind/csr.hpp
https://github.com/Dekken/statick/blob/master/inc/statick/pybind/numpy.hpp
https://github.com/Dekken/statick/blob/master/inc/statick/pybind/linear_model.hpp#L23

Let me try isolate and clean things up.

eloyfelix:
wow, that would be great. Many thanks!

@wjakob wouldn't this be a nice feature to add to pybind11 core?

PR #2247, opened 2020-06-11
numpy csr_matrix interop - Work in progress

Dekken:
PR for https://github.com/pybind/pybind11/issues/1665

I expect many changes, this is just to start with something, and see if tests pass.


Dekken:
if I need to use some functions in addition to 
``` https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L203 ```

Should I update this code, or add it to another version?
I'm still wrapping my head around this function pointer mapping business, any tips are appreciated.

Dekken:
I assume the function codes are taken from here https://github.com/numpy/numpy/blob/master/numpy/core/code_generators/numpy_api.py

Dekken:
I believe the csr_matrix is a part of "scipy" rather than "numpy".

Is it acceptable to add "scipy" to the test environments?

PR #2249, opened 2020-06-12
Respect rvp for registered_instances.

jskimko:
In `type_caster_generic::cast`, the rvp will be ignored if a returned object exists in `registered_instances`. This results in non-WYSIWYG behavior for `copy` and `move` policies.

This change adds a conditional around the registered_instances logic.

#### Example
```
namespace py = pybind11;                                                        
                                                                                
struct A {                                                                      
    A(int i) : x(i) { printf("A(%d) %p\n", x, this); }                          
    A(const A& a) : x(a.x) { printf("copy() %p\n", this); }                     
    ~A() { printf("~A() %p\n", this); }                                         
                                                                                
    void foo() { printf("foo() %p %d\n", this, x); }                            
    void inc() { printf("inc() %p %d\n", this, ++x); }                          
    int x;                                                                      
};                                                                              
                                                                                
A g(3);                                                                         
A& get() { printf("get() %p\n", &g); return g; }                                
void foo() { g.foo(); }                                                         
void inc() { g.inc(); }                                                         
                                                                                
PYBIND11_MODULE(example, m) {                                                   
    m.doc() = "an example module.";                                             
                                                                                
    py::class_<A>(m, "A")                                                       
        .def("foo", &A::foo)                                                    
        .def("inc", &A::inc);                                                   
                                                                                
    m.def("foo", foo);                                                          
    m.def("inc", inc);                                                          
                                                                                
    m.def("get_ref", &get, py::return_value_policy::reference);                 
    m.def("get_copy", &get, py::return_value_policy::copy);                     
}
```

```
from example import *                                                              
                                                                             
ref = get_ref()                       
# get() 0x2aaaac7f9004                    
copy = get_copy()
# get() 0x2aaaac7f9004               
# copy() 0x69f540                                                                  
                                                                                   
foo()
# foo() 0x2aaaac7f9004 3                                                                       
ref.foo()                                                                          
# foo() 0x2aaaac7f9004 3
copy.foo()                                                                         
# foo() 0x69f540 3     
                                                                                     
inc()                                                                              
# inc() 0x2aaaac7f9004 4

ref.foo()                                                                          
# foo() 0x2aaaac7f9004 4
copy.foo()
# foo() 0x69f540 3
```

#### Tests
~~This causes `test_methods_and_attributes` and `test_operator_overloading` to fail on `assert cstats.copy_constructions == XXX`. This suggests that previously, even with a `copy` or `move` policy, there was intent to have reference-like behavior (i.e. avoid a copy), or that the wrong policy was specified.~~  
Fixed 21679eb.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -508,11 +508,13 @@ public:
         if (src == nullptr)
             return none().release();
 
-        auto it_instances = get_internals().registered_instances.equal_range(src);
-        for (auto it_i = it_instances.first; it_i != it_instances.second; ++it_i) {
-            for (auto instance_type : detail::all_type_info(Py_TYPE(it_i->second))) {
-                if (instance_type && same_type(*instance_type->cpptype, *tinfo->cpptype))
-                    return handle((PyObject *) it_i->second).inc_ref();
+        if (policy != return_value_policy::copy && policy != return_value_policy::move) {
+            auto it_instances = get_internals().registered_instances.equal_range(src);
+            for (auto it_i = it_instances.first; it_i != it_instances.second; ++it_i) {
+                for (auto instance_type : detail::all_type_info(Py_TYPE(it_i->second))) {
+                    if (instance_type && same_type(*instance_type->cpptype, *tinfo->cpptype))
+                        return handle((PyObject *) it_i->second).inc_ref();
+                }
             }
         }
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -871,7 +871,7 @@ public:
 
     static handle cast(const itype &src, return_value_policy policy, handle parent) {
         if (policy == return_value_policy::automatic || policy == return_value_policy::automatic_reference)
-            policy = return_value_policy::copy;
+            policy = return_value_policy::reference;
         return cast(&src, policy, parent);
     }
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1859,7 +1859,9 @@ private:
     arg_v(arg &&base, T &&x, const char *descr = nullptr)
         : arg(base),
           value(reinterpret_steal<object>(
-              detail::make_caster<T>::cast(x, return_value_policy::automatic, {})
+              std::is_rvalue_reference<decltype(x)>::value
+              ? detail::make_caster<T>::cast(x, return_value_policy::copy, {})
+              : detail::make_caster<T>::cast(x, return_value_policy::automatic, {})
           )),
           descr(descr)
 #if !defined(NDEBUG)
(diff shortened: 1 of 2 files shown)


jskimko:
In-place operators appears to be using `copy` instead of `reference`.

include/pybind11/operators.h
```diff
@@ -103,6 +105,7 @@ template <typename B, typename L, typename R> struct op_impl<op_##id, op_l, B, L
     static char const* name() { return "__" #id "__"; }                                \
     static auto execute(L &l, const R &r) -> decltype(expr) { return expr; }           \
     static B execute_cast(L &l, const R &r) { return B(expr); }                        \
+    static return_value_policy rvp() { return return_value_policy::reference; }        \
```
jskimko, line 108: If there is a way to check the return type of `execute` to narrow the application of `rvp::reference`, please advise. I.e., similar to `decltype(expr)` but without the arguments. However, even in cases where the return value is not a reference, this will still invoke copy construction, so this is more of a self-documentating-code problem.

wjakob:
There was never a promise of "WYSIWYG" (not sure what that means here)

Return value policies *only* apply if the object has not yet been registered with Python. FWIW making this kind of change (even if there was broad consensus to do so) would break tons of packages that depend on the current behavior. I don't think it's worth the pain..

include/pybind11/cast.h
```diff
@@ -869,7 +871,7 @@ template <typename type> class type_caster_base : public type_caster_generic {
 
     static handle cast(const itype &src, return_value_policy policy, handle parent) {
         if (policy == return_value_policy::automatic || policy == return_value_policy::automatic_reference)
-            policy = return_value_policy::copy;
+            policy = return_value_policy::reference;
```
jskimko, line 874: This code path occurs when:
(1) the return type is a const/non-const lvalue reference and no rvp is specified.
(2) a reference value is provided as a default for `py::arg`.
(3) any others?

The previous `rvp::copy` policy was ignored due to the early-return logic. With the new conditional guard, we now use `rvp::reference` and end up checking the value.

Case (1) should be `rvp::reference`. Case(2) is handled in 6745c6a981e761d95ddbcd772c720461092f918e.

YannickJadoul:
> There was never a promise of "WYSIWYG" (not sure what that means here)
> 
> Return value policies _only_ apply if the object has not yet been registered with Python. FWIW making this kind of change (even if there was broad consensus to do so) would break tons of packages that depend on the current behavior. I don't think it's worth the pain..

I indeed think this might be quite a breaking change, especially for the cases where now a default is used but that are counting to get the existing Python object. `return_value_policy` is quite a brittle, complex bridge between the C++ and Python way of thinking about memory, so unless we can show that all existing cases will work as expected, this couldn't be merged before a version 3, anyway.

Besides that: if you want to return a copy, why wouldn't you just return by value rather than by reference? Returning a pointer or reference seems to indicate you want to copy as little as possible and that the actual instance of the object gets shared, no?

Issue #1568, opened 2018-10-17
Having a class which holds the same pointer inside can crash


novoselov-ab:
## Issue description

Having a class which holds the same pointer inside can crash.
The essence of the problem is that `register_instance` and `deregister_instance` search all instances with the same pointer and then select with matching type: `Py_TYPE(self) == Py_TYPE(it->second)`. In certain situations there could be multiple instance of the same type and same pointer.

## Reproducible example code

This class can cause a problem:
```cpp
    struct SamePointer {};
    static SamePointer samePointer;

    py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer").def(py::init<>([&]() {
        return &samePointer;
    }));
```

To reproduce you need regular simple class:
```cpp
    struct Empty {};
    py::class_<Empty> (m, "Empty")
        .def(py::init<>());
```

And this python code
```py
import m
x = []
for _ in range(100):
    x.append(m.SamePointer())

for j in range(2):
    for i in range(100):
        if i % 2 == j:
            x[i] = m.Empty()
```

Creating other python classes will reuse the same python object, which was meant to be freed by python, but still considered valid by pybind. 
This line will be triggered `pybind11_fail("pybind11_object_dealloc(): Tried to deallocate unregistered instance!");`

---
Maybe it is a bad idea to write a class this way, but it was non obvious and took a lot of time to catch a failure and debug.

Blistic:
I hit a slight variant of this problem myself when returning cached shared_ptr's of pybind11 objects.  I fixed the issue by changing the ```deregister_instance_impl``` if condition to check the exact pointer address itself, rather than just the type:
```
if (self == it->second) {
  registered_instances.erase(it);
  return true;
} 
```
After making this change it fixes my version of the problem, as well as the reproducible example code of @novoselov-ab and all of the other pybind11 tests pass on my system.  The type check was explicitly put in by @jagerman in commit 1f8a100d38e1a9bc96586cef63a53a79e8d343a6 and it would have been more straightforward to put in an address equality check so I assume there is a good reason for checking the type equality that I do not properly understand?  I just cannot find anything that breaks on my system when I use the more specific address equality check, and it fixes the current issue.

jagerman:
If I'm remembering correctly, the issue here is that the address equality check will break multiple inheritance because a pointer to the base instance won't necessarily have the same address as the pointer to the derived instance; consider:

```C++
#include <iostream>

class Base1 { int b1; };
[...]

Issue #1238, opened 2018-01-07
Pybind incorrectly deregisters instances (by type comparison), yields errors when C++ recycles memory


EricCousineau-TRI:
I ran into an interesting (read: frustrating) bug that occurred when I had code that looked like this:
```
obj = m.UniquePtrHeld(1)
m.unique_ptr_terminal(obj)  # Destroys the value held by `obj`
obj = m.UniquePtrHeld(1)
```

After C++ destroyed the first instance, the second instance was allocated into the same address, when casting the instance back, `pybind` registered a new instance (in `get_internals().registered_instances`) when the same ptr-value, but a different instance.
In `deregister_instance_impl`, it finds all values that match the given ptr, and only checks if the `Py_TYPE` is the same.

The fix in this case is just to compare instances (if there wasn't a strong reason for getting the type). An example change:
https://github.com/EricCousineau-TRI/pybind11/blob/feature/unique_ptr_arg_pr1-wip/include/pybind11/detail/class.h#L225

jagerman:
I'm not entirely sure I see what the issue is.  Is it that there are two entries in `registered_instances`, but that the old one is no longer valid?  It sounds like that old instance needs to be deleted at some earlier point.

As for the proposed fix, there is a strong reason for the type comparison, namely that two separate objects can share the same address, typically when one object is the first member of a class (or first member of a first member, etc.).

In other words, something like this can trigger it:
```C++
struct A { /* ... */ };
struct B {
    A a;
    /* ... */
};
py::class_<A>(m, "A");
py::class_<B>(m, "B")
    .def("a", [](B &self) { return &self.a; }, py::return_value_policy::reference_internal);
```
in which case both the `A` and `B` instances will share the same address, hence the check *also* on the type for all registered instance handling. In theory, you could probably delete all of the instances when deleting the outermost object (e.g. the "B"), but I don't think there's a trivial way to figure out which one is the outermost.

EricCousineau-TRI:
To preface, this is for my `unique_ptr` PR, so it is not a behavior that'd normally occur in `pybind` (unless ownership transfer flags are introduced, per #1226 / #1200).

> namely that two separate objects can share the same address...

Ah, that makes sense! Thank you! I guess a slightly more costly solution then is to compare instance and type.

> Is it that there are two entries in registered_instances, but that the old one is no longer valid?

Yup!

> It sounds like that old instance needs to be deleted at some earlier point.

[...]

PR #2252, opened 2020-06-14
Bugfix/Check actual value when deregistering pybind11 instance

rfhaque:
In ```deregister_instance_impl``` the code iterates through all instances associated with a pointer in the multimap ```registered_instances``` and deregisters the first instance that matches the type. This may fail in cases where multiple instances of the same type are registered to the same underlying pointer. The fix is to additionally check if the instance to be deregistered is in fact the same. The following code (potentially) reproduces the issue

```
namespace py = pybind11;

template <typename T>
class shared_holder {
  struct deleter {
    bool owned;
    deleter(bool owned = true) : owned(owned) {}
    void operator()(T* obj) const {
      if(owned) delete obj;
    }
  };
  std::unique_ptr<T, deleter> impl;
public:
  shared_holder(T* p, bool owned = true) : impl(p, owned) {}
  shared_holder(T& p) : impl(&p, false) {}
  shared_holder(const shared_holder& r) : 
    impl((const_cast<shared_holder&>(r)).impl.release(), r.impl.get_deleter().owned) {}
  T* get() const { return impl.get(); }
};

PYBIND11_DECLARE_HOLDER_TYPE(T, shared_holder<T>);

struct A {
  int x;
  A(int x) : x(x) { printf("Default A: %p\n", this); }
  A(const A& a) : x(a.x) { printf("Copy A: %p\n", this); }
  ~A() { printf("Destroy A: %p\n", this); }
};

struct B {
  int x;
  B(int x) : x(x) { printf("Default B: %p\n", this); }
  B(const B& a) : x(a.x) { printf("Copy B: %p\n", this); }
  ~B() { printf("Destroy B: %p\n", this); }
};

A& foo(A& a) {
  return a;
}

PYBIND11_MODULE(example, m) {
  py::class_<A, shared_holder<A>>(m, "A")
  .def(pybind11::init([](A* a) {
    printf("Shared A: %p\n", a); 
    return shared_holder<A>(a, false); 
  } ))
  .def(pybind11::init<int>());
  py::class_<B, shared_holder<B>>(m, "B")
  .def(pybind11::init([](B* a) {
    printf("Shared B: %p\n", b); 
    return shared_holder<B>(b, false);
  } ))
  .def(pybind11::init<int>());
  m.def("foo", foo);
}
```

```
from example import A, B
from example import foo

class AB(A, B):
  def __init__(self, val = 0, a = None):
    if a is None:
      A.__init__(self, val)
      B.__init__(self, val)
    else:
      A.__init__(self, a)
      B.__init__(self, a)

a = AB(val=100)
print "------------------------------------------"
b = AB(a=a)
print "------------------------------------------"
b = AB(a=a)
print "------------------------------------------"

print "Calling foo"
foo(a)
print "Calling foo...Done"
```

```
Default A: 0x707550
Default B: 0x7075b0
------------------------------------------
Shared A: 0x707550
Shared B: 0x7075b0
------------------------------------------
Shared A: 0x707550
Shared B: 0x7075b0
------------------------------------------
Calling foo
*** Error in `./a.out': double free or corruption (fasttop): ***
```

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -250,7 +250,7 @@ inline bool deregister_instance_impl(void *ptr, instance *self) {
     auto &registered_instances = get_internals().registered_instances;
     auto range = registered_instances.equal_range(ptr);
     for (auto it = range.first; it != range.second; ++it) {
-        if (Py_TYPE(self) == Py_TYPE(it->second)) {
+        if (self == it->second) {
             registered_instances.erase(it);
             return true;
         }
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -420,6 +420,16 @@ TEST_SUBMODULE(class_, m) {
     py::class_<PyPrintDestructor>(m, "PyPrintDestructor")
         .def(py::init<>())
         .def("throw_something", &PyPrintDestructor::throw_something);
+
+    struct SamePointer {};
+    static SamePointer samePointer;
+    py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer")
+        .def(py::init([]() { return &samePointer; }))
+        .def("__del__", [](SamePointer&) { py::print("__del__ called"); });
+
+    struct Empty {};
+    py::class_<Empty>(m, "Empty")
+        .def(py::init<>());
 }
 
 template <int N> class BreaksBase { public:
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -371,3 +371,18 @@ def test_non_final_final():
 def test_exception_rvalue_abort():
     with pytest.raises(RuntimeError):
         m.PyPrintDestructor().throw_something()
+
+
+# https://github.com/pybind/pybind11/issues/1568
+def test_multiple_instances_with_same_pointer(capture):
+    n = 100
+    instances = [m.SamePointer() for _ in range(n)]
+    for i in range(n):
+        # We need to reuse the same allocated memory for with a different type,
+        # to ensure the bug in `deregister_instance_impl` is detected. Otherwise
+        # `Py_TYPE(self) == Py_TYPE(it->second)` will still succeed, even though
+        # the `instance` is already deleted.
+        instances[i] = m.Empty()
+    # No assert: if this does not trigger the error
+    #   pybind11_fail("pybind11_object_dealloc(): Tried to deallocate unregistered instance!");
+    # and just completes without crashing, we're good.


henryiii:
Duplicate of #1592

include/pybind11/detail/class.h
```diff
@@ -223,7 +223,7 @@ inline bool deregister_instance_impl(void *ptr, instance *self) {
     auto &registered_instances = get_internals().registered_instances;
     auto range = registered_instances.equal_range(ptr);
     for (auto it = range.first; it != range.second; ++it) {
-        if (Py_TYPE(self) == Py_TYPE(it->second)) {
+        if (Py_TYPE(self) == Py_TYPE(it->second) && self == it->second) {
```
YannickJadoul, line 226: If `self == it->second`, wouldn't `Py_TYPE(self) == Py_TYPE(it->second)` be a useless comparison?

(`Py_TYPE` is `#define Py_TYPE(ob)             (((PyObject*)(ob))->ob_type)` in CPython's headers)

YannickJadoul: And given this, it's important to figure out why `Py_TYPE` was used in the first place?

I do think [this concern](https://github.com/pybind/pybind11/issues/1568#issuecomment-435686117) was adequately addressed by [this reply](https://github.com/pybind/pybind11/issues/1568#issuecomment-435701894).

YannickJadoul (commented):
If I understand correctly, this is a good fix, but:
1. Would it be easy to add some tests? See also #1568 and #2394 for a simpler reproducer.
2. #1592 was earlier. Should we give credit to @Blistic for fixing this first, and merge that (identical) PR?

Apart from that, I believe this only occurs with constructors, because if you're returning a pointer that's associated with an existing instance, it will return that existing instance?

henryiii:
> Should we give credit to @Blistic for fixing this first, and merge that (identical) PR?

You can also assign shared authorship. We can go with the one that adds tests first and make it a race. ;)

YannickJadoul:
> 1. Would it be easy to add some tests? See also #1568 and #2394 for a simpler reproducer.

Regarding this: I just checked, and it's complex to consistently reproduce (and dependent on Python implementation details, I believe). Reason is that the problem depends on memory getting corrupted (the `Py_TYPE` inside the `instance` object, to be precise). If that corruption doesn't happen, pybind11 still finds as many entries in `registered_instances` with the correct type as before.

<details>
  <summary>Extended demonstration</summary>

Debug output changes (i.e., printing the arguments of `register_instance_impl` and `deregister_instance_impl`, and contents of `get_internals().registered_instances`:
```diff
diff --git a/include/pybind11/detail/class.h b/include/pybind11/detail/class.h
index 8d36744f..a3644a13 100644
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -243,13 +243,21 @@ inline void traverse_offset_bases(void *valueptr, const detail::type_info *tinfo
 }
 
 inline bool register_instance_impl(void *ptr, instance *self) {
+    for (auto &p : get_internals().registered_instances)
+        printf("* %p, %p (%s)\n", p.first, p.second, Py_TYPE(p.second)->tp_name);
     get_internals().registered_instances.emplace(ptr, self);
+    printf("register_instance_impl(%p %p)\n", ptr, self);
     return true; // unused, but gives the same signature as the deregister func
 }
 inline bool deregister_instance_impl(void *ptr, instance *self) {
+    for (auto &p : get_internals().registered_instances)
+        printf("* %p, %p (%s)\n", p.first, p.second, Py_TYPE(p.second)->tp_name);
     auto &registered_instances = get_internals().registered_instances;
     auto range = registered_instances.equal_range(ptr);
+    printf("deregister_instance_impl(%p, %p)\n", ptr, self);
     for (auto it = range.first; it != range.second; ++it) {
+        // if (Py_TYPE(self) == Py_TYPE(it->second) && self == it->second) {
+        printf("- %p (%s), %p (%s)\n", self, Py_TYPE(self)->tp_name, it->second, Py_TYPE(it->second)->tp_name);
         if (Py_TYPE(self) == Py_TYPE(it->second)) {
             registered_instances.erase(it);
             return true;
```

Interactive shell:
```pycon
Python 3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 8.3.0] on linux
>>> import example as skel
>>> s1 = skel.S(0)
register_instance_impl(0x7fadcbfef580 0x7fadcdd0d470)
>>> s1 = skel.S(0)
* 0x7fadcbfef580, 0x7fadcdd0d470 (example.S)
register_instance_impl(0x7fadcbfef580 0x7fadcc726bb0)
* 0x7fadcbfef580, 0x7fadcc726bb0 (example.S)
* 0x7fadcbfef580, 0x7fadcdd0d470 (example.S)
deregister_instance_impl(0x7fadcbfef580, 0x7fadcdd0d470)
- 0x7fadcdd0d470 (example.S), 0x7fadcc726bb0 (example.S)
>>> del s1
* 0x7fadcbfef580, 0x7fadcdd0d470 (str)
deregister_instance_impl(0x7fadcbfef580, 0x7fadcc726bb0)
- 0x7fadcc726bb0 (example.S), 0x7fadcdd0d470 (str)
terminate called after throwing an instance of 'std::runtime_error'
  what():  pybind11_object_dealloc(): Tried to deallocate unregistered instance!
Aborted (core dumped)
```

</details>

henryiii:
It's okay to add a test that is flaky if there's a problem but always works if it's fixed. (IIRC)

YannickJadoul:
> Reason is that the problem depends on memory getting corrupted

So, to clarify: this is some UB; so we probably can't confidently test it, being sure it will go wrong?
Then again, maybe we still should? Happy to try figure something out, if you want.

YannickJadoul:
@EricCousineau-TRI Cfr. #1238: did you have a consistent reproducer?

YannickJadoul:
Suggested test (stolen from @Blistic in https://github.com/pybind/pybind11/pull/1592#issue-228178289) and fix for my own remark:

```diff
diff --git a/include/pybind11/detail/class.h b/include/pybind11/detail/class.h
index 8d36744f..4b09fa51 100644
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -250,7 +250,7 @@ inline bool deregister_instance_impl(void *ptr, instance *self) {
     auto &registered_instances = get_internals().registered_instances;
     auto range = registered_instances.equal_range(ptr);
     for (auto it = range.first; it != range.second; ++it) {
-        if (Py_TYPE(self) == Py_TYPE(it->second)) {
+        if (self == it->second) {
             registered_instances.erase(it);
             return true;
         }
diff --git a/tests/test_class.cpp b/tests/test_class.cpp
index 5369cb06..e3ec4601 100644
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -390,6 +390,16 @@ TEST_SUBMODULE(class_, m) {
     py::class_<PyPrintDestructor>(m, "PyPrintDestructor")
         .def(py::init<>())
         .def("throw_something", &PyPrintDestructor::throw_something);
+
+    struct SamePointer {};
+    static SamePointer samePointer;
+    py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer")
+        .def(py::init([]() { return &samePointer; }))
+        .def("__del__", [](SamePointer&) { py::print("__del__ called"); });
+
+    struct Empty {};
+    py::class_<Empty>(m, "Empty")
+        .def(py::init<>());
 }
 
 template <int N> class BreaksBase { public:
diff --git a/tests/test_class.py b/tests/test_class.py
index 4214fe79..4f60126a 100644
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -331,3 +331,18 @@ def test_non_final_final():
 def test_exception_rvalue_abort():
     with pytest.raises(RuntimeError):
         m.PyPrintDestructor().throw_something()
+
+
+# https://github.com/pybind/pybind11/issues/1568
+def test_multiple_instances_with_same_pointer(capture):
+    n = 100
+    instances = [m.SamePointer() for _ in range(n)]
+    for i in range(n):
+        # We need to reuse the same allocated memory for with a different type,
+        # to ensure the bug in `deregister_instance_impl` is detected. Otherwise
+        # `Py_TYPE(self) == Py_TYPE(it->second)` will still succeed, even though
+        # the `instance` is already deleted.
+        instances[i] = m.Empty()
+    # No assert: if this does not trigger the error
+    #   pybind11_fail("pybind11_object_dealloc(): Tried to deallocate unregistered instance!");
+    # and just completes without crashing, we're good.
```

YannickJadoul:
Nice! All turning red (without the fix!); so the tests are working :-)

EricCousineau-TRI:
> @EricCousineau-TRI Cfr. #1238: did you have a consistent reproducer?

I believe I did, but it was specific to our fork using `unique_ptr` ownership transfer from Python to C++, so not achievable with upstream pybind11.

EricCousineau-TRI (approved):
Approving given that Yannick tested this w/o the change to `class.h`.

Thanks!

YannickJadoul:
> Thanks!

Thanks :-) Given the subtle nature of this bug, I'll wait a bit and see if we get one or a few more reviews.

henryiii (approved):

wjakob:
The fix here will resolve a long-standing problem (yay!), but I suspect that it might also uncover some other problems related to usage of multiple inheritance that we don't adequately test at the moment. Or perhaps not. I suggest that we merge it and see what happens.

YannickJadoul:
Let's see, then, yes. As far as I could tell, https://github.com/pybind/pybind11/issues/1568#issuecomment-435701894 addressed this, and I had another check, a few months ago. But if things start going wrong, I'm happy to get back into this and start debugging!

YannickJadoul:
Thanks, @rfhaque and @Blistic! :-)

Issue #2051, opened 2020-01-01
Question about argument order for PYBIND11_OVERLOAD_NAME


jonchammer:
## Issue description

There seems to be a discrepancy in the documentation with regards to the `PYBIND11_OVERLOAD_NAME` macro. In this [this](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#overriding-virtual-functions-in-python) section, the documentation says:

> There are also two alternate macros PYBIND11_OVERLOAD_PURE_NAME and PYBIND11_OVERLOAD_NAME which take a string-valued name argument between the Parent class and Name of the function slots, which defines the name of function in Python.

However, the reference for that macro (and indeed the actual source code) implies that the C name comes before the Python name (rather than the other way around).

Based on my own testing, it seems like the quoted advice is correct and that the python name should come first, but I'd like to verify to clean up any potential confusion. This is a basic example, where the function `bar()` in C++ would be mapped to the function `python_bar()` in Python:

## Example
```C++
struct Foo
{
    virtual void bar() {}
};

// Trampoline class for Foo
struct PyFoo : public Foo
{
    void bar() override
    {
        // Compiles
        PYBIND11_OVERLOAD_NAME(
            void, Foo, "python_bar", bar,
        );

        // Does not compile
        // PYBIND11_OVERLOAD_NAME(
        //     void, Foo, bar, "python_bar",
        // );
    }
};

// ...
py::class_<Foo, PyFoo>(m, "Foo")
        .def(py::init<>())
        .def("python_bar", &Foo::bar);
```

Thanks for all the work that goes into this project! It's appreciated.

methylDragon:
I know this is a bit late, but I had the same issue.

You are absolutely correct! I've sent a relevant PR here: https://github.com/pybind/pybind11/pull/2254

YannickJadoul:
Resolve by that PR, then! Thanks for reporting, @jonchammer, and for fixing, @methylDragon!

PR #2254, opened 2020-06-15
Rectify docs for PYBIND11_OVERLOAD_NAME

methylDragon:
This PR fixes the issue mentioned in https://github.com/pybind/pybind11/issues/2051

It is a small change, and a minor error, but it caused quite a bit of headache and confusion until I dove deep into the macro expansion.

The docs confuse the `name` and `fn` macro parameters.

## Explanation
Notice in these lines there is a call to `PYBIND11_OVERLOAD_INT`. Specifically with the third argument being `name`.

We can see that the `fn` argument in this case is used in `cname::fn(__VA_ARGS__)` to call the appropriate C++ class method of name `fn`.
https://github.com/pybind/pybind11/blob/63df87fa490d49244b76249854559ec8db22f119/include/pybind11/pybind11.h#L2160-L2162

But in `PYBIND11_OVERLOAD_INT`, we can see that the macro expects the `name` argument to be the name of the Python function/method name.
https://github.com/pybind/pybind11/blob/63df87fa490d49244b76249854559ec8db22f119/include/pybind11/pybind11.h#L2130-L2141

In fact, in the reference doc for `PYBIND11_OVERLOAD_INT` we see that this is the case also.

https://github.com/pybind/pybind11/blob/63df87fa490d49244b76249854559ec8db22f119/include/pybind11/pybind11.h#L2117-L2124

## Proposed solutions
While it would be nice to flip around the arguments, I think a less API breaking fix would just be to fix the docs.

This particular documentation error doesn't affect `PYBIND11_OVERLOAD` or `PYBIND11_OVERLOAD_PURE`, because in that case both bound method names on the C++ and Python side are identical. That's probably the reason why this issue has flown under the radar for so long.

Thanks for this extremely useful binding library!

- CH3EERS

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2152,8 +2152,8 @@ template <class T> function get_overload(const T *this_ptr, const char *name) {
         PYBIND11_OVERLOAD_NAME(
             std::string, // Return type (ret_type)
             Animal,      // Parent class (cname)
-            toString,    // Name of function in C++ (name)
-            "__str__",   // Name of method in Python (fn)
+            "__str__",   // Name of method in Python (name)
+            toString,    // Name of function in C++ (fn)
         );
       }
 \endrst */


wjakob:
Indeed! Good catch :)

PR #2256, opened 2020-06-18
Adding test_isinstance_string_types, with asserts simply matching cur…

rwgk:
…rent behavior.

Basis for discussing actual desired behavior.

Background:
* https://github.com/pybind/pybind11/pull/1463/commits/77059f00096b093c1b18223ade7b6fce6a79e819
* https://github.com/pybind/pybind11/pull/2198

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -928,12 +928,21 @@ public:
 private:
     /// Return string representation -- always returns a new reference, even if already a str
     static PyObject *raw_str(PyObject *op) {
+#ifdef PYBIND11_STR_RAW_STR_PY2_EMULATE_UNICODE_CONSTRUCTOR_NOT_IMPLICIT_ENCODE
 #if PY_MAJOR_VERSION < 3
         PyObject *str_value = PyObject_Unicode(op);
 #else
         PyObject *str_value = PyObject_Str(op);
 #endif
         if (!str_value) throw error_already_set();
+#else
+        PyObject *str_value = PyObject_Str(op);
+        if (!str_value) throw error_already_set();
+#if PY_MAJOR_VERSION < 3
+        PyObject *unicode = PyUnicode_FromEncodedObject(str_value, "utf-8", nullptr);
+        Py_XDECREF(str_value); str_value = unicode;
+#endif
+#endif
         return str_value;
     }
 };
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -782,7 +782,9 @@ PYBIND11_NAMESPACE_END(detail)
         Name(handle h, stolen_t) : Parent(h, stolen_t{}) { } \
         PYBIND11_DEPRECATED("Use py::isinstance<py::python_type>(obj) instead") \
         bool check() const { return m_ptr != nullptr && (bool) CheckFun(m_ptr); } \
-        static bool check_(handle h) { return h.ptr() != nullptr && CheckFun(h.ptr()); }
+        static bool check_(handle h) { return h.ptr() != nullptr && CheckFun(h.ptr()); } \
+        template <typename Policy_> \
+        Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
 
 #define PYBIND11_OBJECT_CVT(Name, Parent, CheckFun, ConvertFun) \
     PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
@@ -792,9 +794,7 @@ PYBIND11_NAMESPACE_END(detail)
     { if (!m_ptr) throw error_already_set(); } \
     Name(object &&o) \
     : Parent(check_(o) ? o.release().ptr() : ConvertFun(o.ptr()), stolen_t{}) \
-    { if (!m_ptr) throw error_already_set(); } \
-    template <typename Policy_> \
-    Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
+    { if (!m_ptr) throw error_already_set(); }
 
 #define PYBIND11_OBJECT(Name, Parent, CheckFun) \
     PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -738,8 +738,6 @@ inline bool PyIterable_Check(PyObject *obj) {
 inline bool PyNone_Check(PyObject *o) { return o == Py_None; }
 inline bool PyEllipsis_Check(PyObject *o) { return o == Py_Ellipsis; }
 
-inline bool PyUnicode_Check_Permissive(PyObject *o) { return PyUnicode_Check(o) || PYBIND11_BYTES_CHECK(o); }
-
 inline bool PyStaticMethod_Check(PyObject *o) { return o->ob_type == &PyStaticMethod_Type; }
 
 class kwargs_proxy : public handle {
@@ -885,7 +883,7 @@ class bytes;
 
 class str : public object {
 public:
-    PYBIND11_OBJECT_CVT(str, object, detail::PyUnicode_Check_Permissive, raw_str)
+    PYBIND11_OBJECT_CVT(str, object, PyUnicode_Check, raw_str)
 
     str(const char *c, size_t n)
         : object(PyUnicode_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
@@ -947,8 +945,6 @@ inline namespace literals {
  \endrst */
 inline str operator"" _s(const char *s, size_t size) { return {s, size}; }
 }
-
-template <> inline bool isinstance<str>(handle obj) { return PyUnicode_Check(obj.ptr()); }
 
 /// \addtogroup pytypes
 /// @{
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -928,12 +928,12 @@ public:
 private:
     /// Return string representation -- always returns a new reference, even if already a str
     static PyObject *raw_str(PyObject *op) {
-        PyObject *str_value = PyObject_Str(op);
-        if (!str_value) throw error_already_set();
 #if PY_MAJOR_VERSION < 3
-        PyObject *unicode = PyUnicode_FromEncodedObject(str_value, "utf-8", nullptr);
-        Py_XDECREF(str_value); str_value = unicode;
+        PyObject *str_value = PyObject_Unicode(op);
+#else
+        PyObject *str_value = PyObject_Str(op);
 #endif
+        if (!str_value) throw error_already_set();
         return str_value;
     }
 };
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -948,6 +948,8 @@ inline namespace literals {
 inline str operator"" _s(const char *s, size_t size) { return {s, size}; }
 }
 
+template <> inline bool isinstance<str>(handle obj) { return PyUnicode_Check(obj.ptr()); }
+
 /// \addtogroup pytypes
 /// @{
 class bytes : public object {
@@ -1009,6 +1011,12 @@ inline str::str(const bytes& b) {
         pybind11_fail("Could not allocate string object!");
     m_ptr = obj.release().ptr();
 }
+
+#if PY_MAJOR_VERSION < 3
+using native_str = bytes;
+#else
+using native_str = str;
+#endif
 
 /// \addtogroup pytypes
 /// @{
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1012,12 +1012,6 @@ inline str::str(const bytes& b) {
     m_ptr = obj.release().ptr();
 }
 
-#if PY_MAJOR_VERSION < 3
-using native_str = bytes;
-#else
-using native_str = str;
-#endif
-
 /// \addtogroup pytypes
 /// @{
 class none : public object {
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -144,7 +144,7 @@ template <typename Type, typename Value> struct list_caster {
     using value_conv = make_caster<Value>;
 
     bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src) || isinstance<str>(src))
+        if (!isinstance<sequence>(src) || isinstance<str>(src) || isinstance<native_str>(src))
             return false;
         auto s = reinterpret_borrow<sequence>(src);
         value.clear();
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -144,7 +144,7 @@ template <typename Type, typename Value> struct list_caster {
     using value_conv = make_caster<Value>;
 
     bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src) || isinstance<str>(src) || isinstance<native_str>(src))
+        if (!isinstance<sequence>(src) || isinstance<bytes>(src) || isinstance<str>(src))
             return false;
         auto s = reinterpret_borrow<sequence>(src);
         value.clear();
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1629,7 +1629,25 @@ struct pyobject_caster {
     template <typename T = type, enable_if_t<std::is_same<T, handle>::value, int> = 0>
     bool load(handle src, bool /* convert */) { value = src; return static_cast<bool>(value); }
 
+#ifdef PYBIND11_DISABLE_IMPLICIT_STR_FROM_BYTES
     template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
+#else
+    template <typename T = type, enable_if_t<std::is_same<T, str>::value, int> = 0>
+    bool load(handle src, bool /* convert */) {
+        if (isinstance<bytes>(src)) {
+            PyObject *str_from_bytes = PyUnicode_FromEncodedObject(src.ptr(), "utf-8", nullptr);
+            if (!str_from_bytes) throw error_already_set();
+            value = reinterpret_steal<type>(str_from_bytes);
+            return true;
+        }
+        if (!isinstance<type>(src))
+            return false;
+        value = reinterpret_borrow<type>(src);
+        return true;
+    }
+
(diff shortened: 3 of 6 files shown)


rwgk:
@wjakob @YannickJadoul Request for Review

This PR has two commits:
1. d50fcadcc5f38b53db977a8aa93acdadd2e003d3 Adding tests for existing behavior. This is to establish a baseline, to more easily see how the behavior changes with the 2. Commit.
2. c78cc6aa0a6bdbce5050f98c15129f476aed6234 Makes pybind11::isinstance<str> behavior compatible with native Python isinstance() behavior.

Commit 2. has a side-effect, that could be seen as **desirable**, or **undesirable**: it disables implicit `decode` when passing `bytes` to a unicode argument.

Note that Commit 2. also fixes https://github.com/pybind/pybind11/pull/2198 (which is what got me started with this PR).

About naming: the meaning of the term `str` is highly ambiguous, and using it to explain is bound to be confusing.
Instead, I'm using the terms `actual bytes` and `actual unicode`, irrespective of the Python version.
I'm using the term "native name" for the name as used by the Python interpreter.

For Python 2:
  * `actual bytes` has the native names `bytes` and `str`.
  * `actual unicode` has the native name `unicode`.

For Python 3:
  * `actual bytes` has the native name `bytes`.
  * `actual unicode` has the native name `str`.

For pybind11, _irrespective_ of the Python version:
  * `actual bytes` is `pybind11::bytes`.
  * `actual unicode` is `pybind11::str`.

The essential part of Commit 2. is a one-line insertion copied from @jagerman : https://github.com/pybind/pybind11/commit/77059f00096b093c1b18223ade7b6fce6a79e819

The desired effect is that it makes `isinstance<str>` compatible native `isinstance(actual_unicode)`. See the corresponding one-line change in `test_pytypes.py`.

But three other tests are affected.

* In `test_pytypes.py` we now have:
    ```
    with pytest.raises(TypeError):                                              
        m.pass_to_pybind11_unicode(actual_bytes)  # NO implicit decode          
    ```
    Before Commit 2. the function call succeeds, with implicit decoding. This behavior change would need **to be added to the Release Notes**.

* Because of the behavior change, `test_eval_file` in `test_eval.py` breaks. It is easily fixed with the equivalent of `six.ensure_text()`.

* This test in `test_stl.py` fails:
    ```
    assert m.func_with_string_or_vector_string_arg_overload('A') == 3  # returns 2 without the fix
    ````
    The fix is a one-line change to `list_caster` in `stl.h`, which amounts to: don't convert a sequence object to a list if it is `actual unicode` or native `str`. To make that one-line change easy to read, I introduced a `native_str` alias in `pytypes.h` (this would need **to be documented**).

**Is the behavior change good or bad?**

My opinion: Good. "Explicit is better than implicit." for this case. It's easy to add `.decode()`, or `six.ensure_text()` if Python 2 compatibility matters. Also note:

  * Passing to `std::string` (and relatives) still supports passing `actual bytes` or `actual unicode` directly. With this PR, all choices are available: accept either, accept only `actual bytes`, accept only `actual unicode`.
  * The added explicit `.decode()` are compatible with pybind11 as-is, if it matters to some users that their code works with old and new versions of pybind11.
  * This PR works out nicely for both Python 2 and Python 3.

Detail: if accepted, I'd want to merge the commits separately, for the same reasons that I have them separately in this PR.

bstaletic:
> But three other tests are affected.

Are those affected on python3 as well as python2?

> Because of the behavior change, `test_eval_file` in `test_eval.py` breaks. It is easily fixed with the equivalent of `six.ensure_text()`.

Can you provide a link to `ensure_text()` and what it does?

> The fix is a one-line change to list_caster in stl.h, which amounts to: don't convert a sequence object to a list if it is actual unicode or native str. To make that one-line change easy to read, I introduced a native_str alias in pytypes.h (this would need to be documented).

What is "native str" in this context? Do you mean `unicode` on py2 and `str` on py3? How is that different from what you have called "actual unicode"? Or do you mean `bytes` on py2 and `str` on py3?

rwgk:
> > But three other tests are affected.
> 
> Are those affected on python3 as well as python2?

`test_eval.py` and `test_stl.py` were passing with Python 3, but failing with Python 2. My new test in `test_pytypes.py` is affected for both Python versions.
Good question, sorry I forgot to clarify. 

> Can you provide a link to `ensure_text()` and what it does?

https://pypi.org/project/six/
https://github.com/benjaminp/six/blob/c0be8815d13df45b6ae471c4c436cce8c192245d/six.py#L923

> What is "native str" in this context? Do you mean `unicode` on py2 and `str` on py3? How is that different from what you have called "actual unicode"? Or do you mean `bytes` on py2 and `str` on py3?

The new `pybind11::native_str` is the C++ side equivalent of native `str` as seen in the interpreter, therefore depends on the Python version. Did you see the new code block in `pytypes.h`?

```
#if PY_MAJOR_VERSION < 3
using native_str = bytes;
#else
using native_str = str;
#endif 
```

Note: `pybind11::str` is always `actual_unicode`.

bstaletic:
Regarding "native str", is the proposed behaviour of `list_caster` correct? You're banning `bytes` and `unicode` on python2, but only `unicode` on python3.

rwgk:
> Regarding "native str", is the proposed behaviour of `list_caster` correct? You're banning `bytes` and `unicode` on python2, but only `unicode` on python3.

Good question again: I figured that was the original intention, to prevent accidental conversion of "regular" str literals to a list of integers. Thinking: with Python 3 it makes sense to allow `bytes`, with Python 2 not so much.

include/pybind11/stl.h
```diff
@@ -144,7 +144,7 @@ template <typename Type, typename Value> struct list_caster {
     using value_conv = make_caster<Value>;
 
     bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src) || isinstance<str>(src))
+        if (!isinstance<sequence>(src) || isinstance<str>(src) || isinstance<native_str>(src))
```
wjakob, line 147: redundant check on Python 3.x?

include/pybind11/pytypes.h
```diff
@@ -950,6 +950,8 @@ inline namespace literals {
 inline str operator"" _s(const char *s, size_t size) { return {s, size}; }
 }
 
+template <> inline bool isinstance<str>(handle obj) { return PyUnicode_Check(obj.ptr()); }
```
wjakob, line 953: This will change the behavior of pybind11, no?

wjakob:
I'm generally okay with these changes, but compatibility is going to be problem. Disabling an implicit bytes -> unicode conversion will break packages that depend on this behavior. Given that we have chosen to adopt SemVer, it seems to me that this kind of change will necessarily fall in the category "postpone to next major version bump".

bstaletic:
> > Regarding "native str", is the proposed behaviour of `list_caster` correct? You're banning `bytes` and `unicode` on python2, but only `unicode` on python3.
> 
> Good question again: I figured that was the original intention, to prevent accidental conversion of "regular" str literals to a list of integers. Thinking: with Python 3 it makes sense to allow `bytes`, with Python 2 not so much.

I still don't understand. Why does it make sense to allow `bytes` on python3, but not on python2? I would understand, were pybind11 dealing in `py::native_bytes` and `py::native_str` everywhere. It isn't, so having `py::native_str` in this one place seems very surprising.

include/pybind11/stl.h
rwgk: Yes, exactly.
I did it this way mostly because I thought it's the clearest way of putting the condition, for the purpose of this code review. ("optimize later")

To also answer @bstaletic question here:
> I still don't understand. Why does it make sense to allow `bytes` on python3, but not on python2? I would understand, were pybind11 dealing in `py::native_bytes` and `py::native_str` everywhere. It isn't, so having `py::native_str` in this one place seems very surprising.

TL;DR: What would be the behavior that you prefer to see here?

With Python 2 the interpreter (!) will tell you `bytes is str` is `True`, therefore you cannot enable conversion of `bytes` to `list` without also enabling conversion of [Python 2 native `str`] to `list`. That's a feature you can only have with Python 3. — Another way to put it: do you NOT want to use the feature in Python 3 because you cannot have it in Python 2?

We can do anything else we want here, what you see is what I assumed was the original intent.
(At some point long time ago, when there was only Python 2, I had a similar problem and disabled `str` to `list` conversions to prevent confusing accidents.)
Once we are certain what we want, we could optimize, e.g. by putting the #ifdef here and not even having the `native_str` type.

include/pybind11/pytypes.h
rwgk: Yes. I wrote the long Request for Review comment for exactly this reason. I'm pointing this out in at the beginning ("desirable/undesirable") and at the end ("good or bad?"), hoping to provide a start for collectively reasoning out if, how, and when we want to change the behaviors.

There are actually "3 and a half" behaviors that are (unfortunately) convoluted:

1. isinstance<str> (reminder: `str` is always `actual unicode` in pybind11), keep or match native?
2. implicit decode when passing `bytes` to `pybind11::str`, keep or disable?
3. what do we want to do about `bytes` to `list` conversions in A. Python 2, B. Python 3?

What would be your preferred answers for these 3 cases **independently**?
I guess we can always engineer a way to achieve any combination of preferred answers, even making the changes independently in multiple pybind11 releases.

rwgk:
> > > Regarding "native str", is the proposed behaviour of `list_caster` correct? You're banning `bytes` and `unicode` on python2, but only `unicode` on python3.
> > 
> > 
> > Good question again: I figured that was the original intention, to prevent accidental conversion of "regular" str literals to a list of integers. Thinking: with Python 3 it makes sense to allow `bytes`, with Python 2 not so much.
> 
> I still don't understand. Why does it make sense to allow `bytes` on python3, but not on python2? I would understand, were pybind11 dealing in `py::native_bytes` and `py::native_str` everywhere. It isn't, so having `py::native_str` in this one place seems very surprising.

Answer under https://github.com/pybind/pybind11/pull/2256#issuecomment-662764517

include/pybind11/stl.h
bstaletic: > TL;DR: What would be the behavior that you prefer to see here?

Either:

- Ban everything that's string-like in the `list_caster`
- Or allow `bytes` in both py2 and py3

> With Python 2 the interpreter (!) will tell you `bytes is str` is `True`, therefore you cannot enable conversion of `bytes` to `list` without also enabling conversion of [Python 2 native `str`] to `list`. 

Why is py2 `str` (or py2 `bytes`, but those are the exact same thing) to `list` problematic if py3 `bytes` to `list` isn't?

> Another way to put it: do you NOT want to use the feature in Python 3 because you cannot have it in Python 2?

Rarely, but if it helps me reason about what is going on in the code, I see nothing wrong with that.

rwgk: > Ban everything that's string-like in the list_caster

That's current behavior. Two consequences:
1. Effectively you do not want to use a feature that's feasible with Python 3 because it's infeasible in Python 2.
2. Effectively a rejection of #2198. 

> Or allow bytes in both py2 and py3

That breaks the existing test in test_stl.py: `assert m.func_with_string_or_vector_string_arg_overload('A') == 3`
We'd have to change that to u'A', `list('A')`, or `iter('A')` to pass again.

The behavior currently implemented in this PR means:
* If you care about Python 3 (and the future) only: you're fine!
* If you still care about Python 2 even past EOL, you'll have to use explicit `list(s)` or `iter(s)` (which is also Python 3 compatible).

I feel that's the best compromise, but I'll go with the votes of the team.
@YannickJadoul and Wenzel, what are your votes on this specific `list_caster` question?

bstaletic: > That breaks the existing test in test_stl.py: assert `m.func_with_string_or_vector_string_arg_overload('A') == 3`

Is this an actual concern, considering that this pull request already contains a breaking change.

rwgk: > > That breaks the existing test in test_stl.py: assert `m.func_with_string_or_vector_string_arg_overload('A') == 3`
> 
> Is this an actual concern, considering that this pull request already contains a breaking change.

I'd say so, I think it's worth getting each of the 3 questions settled in the best way possible.

The only fully intentional change is for `isinstance<str>`. We could preserve the behaviors for the other 2, although I've not looked into what it would take to preserve the implicit `bytes` to `actual unicode` conversions. I'm waiting for Yannick's and Wenzel's takes before investing time there.

YannickJadoul:
> Then again, why does the current implementation not do that?

Let me answer my own question:

https://github.com/pybind/pybind11/blob/7e0a4fb7d0c548231b2dc104054170cb6bebea96/include/pybind11/pytypes.h#L888-L890

So why not change `str::check_` rather than the `isinstance` specialization?

include/pybind11/pytypes.h
```diff
@@ -1012,6 +1014,12 @@ inline str::str(const bytes& b) {
     m_ptr = obj.release().ptr();
 }
 
+#if PY_MAJOR_VERSION < 3
+using native_str = bytes;
+#else
+using native_str = str;
+#endif
+
```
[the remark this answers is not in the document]
rwgk: Ah, very good, thanks for pointing that out! I would never have guessed from the name what it does, but now I know :-)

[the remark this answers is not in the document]
YannickJadoul: If you/we want to provide this as API to pybind11 users, we should provide a typedef, though, rather than have them use this macro.
If so, I'm just not entirely convinced about `native_str` as a name (though I don't now, immediately have a better suggestion).

rwgk:
> So why not change `str::check_` rather than the `isinstance` specialization?

Jumping to my most important question first: what are my options for re-enabling the implicit conversion from `py::bytes` to `py::str` after replacing `detail::PyUnicode_Check_Permissive` with `PyUnicode_Check`?

Motivation for asking: I want to backtrack from making 3 behavior changes to only one. Concretely:

* I want to fix `isinstance<str>` which is currently highly misleading (seems plain wrong to me).
* I want to keep the existing implicit conversion from `py::bytes` to `py::str` (because Wenzel suggested that can only change with a version change) **but how?**
* I want to keep the existing behavior for the `list_caster`, which is super easy to tweak that way.

Secondary observations & questions:

I tried @YannickJadoul 's suggestion (as I understood it), by removing the `isinstance<str>` specialization and changing
```
PYBIND11_OBJECT_CVT(str, object, detail::PyUnicode_Check_Permissive, raw_str)
```
to
```
PYBIND11_OBJECT_CVT(str, object, PyUnicode_Check, raw_str)
```

Python 3: **All** tests pass, with the rest of this PR as-is. (To clarify: NO implicit bytes-str conversion, as before.)

Python 2: Only one test failed: `test_pytypes.py` `test_constructors()`. Fails for `str`. Looking at it, I'm confused why it even works with Python 2 pybind11 as-is. I'll work on trying to understand.

While trying ideas for fixes I stumbled over another problem, for which I created a mock PR to easily show: #2340
The question there: Are we looking at that as a bug or feature?

bstaletic:
> Motivation for asking: I want to backtrack from making 3 behavior changes to only one. Concretely:

> - I want to fix isinstance<str> which is currently highly misleading (seems plain wrong to me).
> - I want to keep the existing implicit conversion from py::bytes to py::str (because Wenzel suggested that can only change with a version change) but how?

Wouldn't this require a major version bump either way, because the first bullet point is a breaking change? I agree that it is the wrong behaviour, but I remember @jagerman talking about compatibility concerns regarding the first bullet point.

&nbsp;

I might be repeating myself, but one more thing. Pybind11 is currently consistent when talking about string-like types. `py::str`, `py::bytes`, `py::isinstance<T>` do not change meaning at all when switching from python2 to python3. I feel that this consistency is important and saves us time when providing support for users.
If others don't think I'm right about this, that's fine. However, I reserve my right to say "told you so" eventually.

rwgk:
> Pybind11 is currently consistent when talking about string-like types. `py::str`, `py::bytes`, `py::isinstance<T>` do not change meaning at all when switching from python2 to python3.

That's some sort of _consistent_ yes, but also highly misleading and outright useless. It's consistent only in some narrowly defined scope. If you look at the bigger picture: one major achievement of the very expensive Python 2 -> 3 transition is a clear delineation between `str` and `bytes` (Python 3 terminology). pybind11 as-is undoes that achievement and is _inconsistent_ with Python semantics.

If anyone out there currently relies on `isinstance<str>` being true for both `bytes` & `str` (or `str` aka (!) `bytes` & `unicode`, respectively, in Python 2), it's probably an oversight, and fixing the bug on our end is more likely to uncover hidden bugs on their end than causing them an actual problem. If they really want `isinstance<str> || isinstance<bytes>` they can just add the second condition, making it obvious and explicit what they want, and it's even compatible with previous pybind11 versions. Compare the small and probably very rare inconvenience of having to add that condition to the ongoing confusion.

bstaletic:
>  If they really want `isinstance<str> || isinstance<bytes>` they can just add the second condition, making it obvious and explicit what they want

I definitely agree with this. In fact, the above is what my own code is already doing. I was just pointing out that if we want to fix this, we will have to make a breaking change no matter what.

rwgk:
> I definitely agree with this. In fact, the above is what my own code is already doing. I was just pointing out that if we want to fix this, we will have to make a breaking change no matter what.

Cool, thanks!
We have #2348 now, thanks to @YannickJadoul , which IIUC allows us to cherry-pick changes that we consider safe for minor version number increments, while holding back other changes for major version number increments. I just have to learn / get used to the workflow now :-)

rwgk:
Hi @YannickJadoul I followed the instructions you gave me via chat: this PR is now based on the `str-bytes-cleanup` branch and set to merge into it. Thanks for the instructions, they worked for me without a problem.

BUT: Please do NOT review and definitely not merge yet. My next step will be to make the `detail::PyUnicode_Check_Permissive` to `PyUnicode_Check` change.

YannickJadoul:
> Hi @YannickJadoul I followed the instructions you gave me via chat: this PR is now based on the `str-bytes-cleanup` branch and set to merge into it. Thanks for the instructions, they worked for me without a problem.
> 
> BUT: Please do NOT review and definitely not merge yet. My next step will be to make the `detail::PyUnicode_Check_Permissive` to `PyUnicode_Check` change.

Great. Just shout when it's time to discuss consequences and dig up/judge old decisions ;-)

rwgk:
This PR is obsoleted by #2380.

Dropping this PR because the convolution of 3 behavior changes lead to long discussions and no approvals :-)

PR #2380 changes only exactly one behavior. See PR description there.

Issue #2258, opened 2020-06-21
Vectorizing POD types


chaitan94:
## Issue description

In the [documentation for vectorization](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#vectorizing-functions), this is mentioned:

> Only arithmetic, complex, and POD types passed by value or by const & reference are vectorized; all other arguments are passed through as-is.

However, it seems that if we use a POD type, `py::vectorize` does not work as expected. When I check the tests, it seems we do test that [non-POD types pass through without vectorization](https://github.com/pybind/pybind11/blob/8c0cd94465fbc1dbc34217e5a614edc784f0913e/tests/test_numpy_vectorize.cpp#L49), but there are no tests/examples for how to vectorize POD types.

## Reproducible example code

```cpp
// Passthrough test: pod types should be vectorized
struct PODClass {
    uint32_t value;
};

py::class_<PODClass>(m, "PODClass")
    .def(py::init<>())
    .def_readwrite("value", &PODClass::value);

PYBIND11_NUMPY_DTYPE(PODClass, value);

m.def("pod_passthrough", py::vectorize(
    [](PODClass a) {
        return a;
    }
));

static_assert(std::is_pod<PODClass>::value);  // To ensure we actually do have a POD type
```

This compiles. When we try to use in python, however:

```python
        o1 = m.PODClass()
        o1.value = 10
        o2 = m.PODClass()
        o2.value = 20
        a = np.array([o1, o2])
>       m.pod_passthrough(a)
E       TypeError: pod_passthrough(): incompatible function arguments. The following argument types are supported:
E           1. (arg0: numpy.ndarray[pybind11_tests.numpy_vectorize.PODClass]) -> object
E       
E       Invoked with: array([<pybind11_tests.numpy_vectorize.PODClass object at 0x7f662db9de30>,
E              <pybind11_tests.numpy_vectorize.PODClass object at 0x7f6638a3d970>],
E             dtype=object)
```

Not sure if this is expected or I am doing something wrong.

EricCousineau-TRI:
If you're looking for custom dtypes, I've enumerated what (I think) the options are here:
#2259

You may want `dtype=object`, as that's the easiest to deal with: #1152

chaitan94:
Not sure if I understood you correctly. Do you mean I should try the following?

```python
vec = np.array([object_1, object_2], dtype=np.object)
```

[...]

PR #2260, opened 2020-06-21
test_numpy_dtypes: Add test for py::vectorize()

EricCousineau-TRI:
Shows test for #2258

~Feel free to close if this doesn't add value.~

With the latest push of 337dc9c, I think this has value?

\cc @chaitan94

### DIFF
--- a/tests/test_numpy_dtypes.cpp
+++ b/tests/test_numpy_dtypes.cpp
@@ -259,7 +259,26 @@ TEST_SUBMODULE(numpy_dtypes, m) {
     catch (...) { return; }
 
     // typeinfo may be registered before the dtype descriptor for scalar casts to work...
-    py::class_<SimpleStruct>(m, "SimpleStruct");
+    py::class_<SimpleStruct>(m, "SimpleStruct")
+        // Explicit construct with braces to ensure zero-valued initialization.
+        .def(py::init([]() { return SimpleStruct{}; }))
+        .def_readwrite("bool_", &SimpleStruct::bool_)
+        .def_readwrite("uint_", &SimpleStruct::uint_)
+        .def_readwrite("float_", &SimpleStruct::float_)
+        .def_readwrite("ldbl_", &SimpleStruct::ldbl_)
+        .def("astuple", [](const SimpleStruct& self) {
+            return py::make_tuple(self.bool_, self.uint_, self.float_, self.ldbl_);
+        })
+        .def_static("fromtuple", [](const py::tuple tup) {
+            if (py::len(tup) != 4) {
+                throw py::cast_error("Invalid size");
+            }
+            return SimpleStruct{
+                tup[0].cast<bool>(),
+                tup[1].cast<uint32_t>(),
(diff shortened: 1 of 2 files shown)


tests/test_numpy_dtypes.py
```diff
@@ -287,6 +291,56 @@ def test_scalar_conversion():
...
+    assert isinstance(s_recarray_scalar, np.ndarray)
+    assert s_recarray_scalar.dtype == simple_dtype
+    with pytest.raises(TypeError) as excinfo:
+        m.f_simple(s_recarray_scalar)
+    assert 'incompatible function arguments' in str(excinfo.value)
+    # Explicitly convert to m.SimpleStruct.
+    assert m.f_simple(
+        m.SimpleStruct.fromtuple(s_recarray_scalar.item())) == 20
+
+    # Show that an array of dtype=object does *not* convert.
```
chaitan94, line 333: Exactly. This is what https://github.com/pybind/pybind11/issues/2258 is about. I feel ideally `m.f_simple_vectorized(np.array([s]))` should work directly without any issues. If not, the docs should at least mention either the `totuple`/`fromtuple` or the `creat_recarray`-like workarounds.

tests/test_numpy_dtypes.cpp
```diff
@@ -259,7 +259,26 @@ TEST_SUBMODULE(numpy_dtypes, m) {
     catch (...) { return; }
 
     // typeinfo may be registered before the dtype descriptor for scalar casts to work...
-    py::class_<SimpleStruct>(m, "SimpleStruct");
+    py::class_<SimpleStruct>(m, "SimpleStruct")
+        // Explicit construct with braces to ensure zero-valued initialization.
+        .def(py::init([]() { return SimpleStruct{}; }))
```
henryiii, line 264: This doesn't work on GCC 4.8 (manylinux1)

chaitan94: Haven't tested this, but I think this should work:
```suggestion
        .def(py::init([]() {
            SimpleStruct s = {0};
            return s;
         }))
```
Alternatively, the ` = {0}` part can be skipped, leaving just `SimpleStruct s;`

bstaletic:
The new test is failing after rebase.

henryiii:
I think that finally makes GCC 4.8 / CentOS 7 (manylinux1) happy.

bstaletic (approved):

henryiii (approved):

EricCousineau-TRI:
Awesome, sorry for the delay here, but merging now!

EricCousineau-TRI:
Thank y'all!

tests/test_numpy_dtypes.py
EricCousineau-TRI: Thanks for the comment! I think a docs change is currently out-of-scope for this PR (as it's a strict improvement).

Might you be up for trying out a doc PR? (if not, I can circle back to it, but it may take me a bit)

PR #2160, opened 2020-04-09
travis: Ensure Python 3.8 is used, disable failing PyPy 2.7 build

EricCousineau-TRI:
Resolves #2159

Lemme know if I should take out the disabling of PyPy.

Thanks to @carlsonmark for the help!

---
This change is [Reviewable](https://reviewable.io/reviews/pybind/pybind11/2160)

PR #2265, opened 2020-06-25
add uninstall target for cmake

### LINKED ISSUE #2264: [Feature Request]  Create uninstall target for CMake
## Issue description

Right now there is no uninstall target for CMake, which makes it hard to uninstall pybind11 after installing it from source. I believe having the ability to sudo make uninstall is very helpful and much wanted.

## Reproducible example code

N/A

acxz:
Resolves #2264

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,8 +247,8 @@ if(PYBIND11_INSTALL)
     DESTINATION ${PYBIND11_CMAKECONFIG_INSTALL_DIR})
 endif()
 
-# uninstall target
-if(NOT TARGET uninstall)
+# Uninstall target
+if(PYBIND11_MASTER_PROJECT)
   configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
                  "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake" IMMEDIATE @ONLY)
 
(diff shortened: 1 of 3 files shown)


acxz:
@wjakob can you take a look at this PR?

acxz:
based on disc on #2160 gonna close/reopen this PR

henryiii:
One tiny change - could you add a check to make sure this only gets added when built as the main project? `PYBIND11_MASTER_PROJECT`, IIRC. If a parent project wants to add this, they should do it themselves. Uninstall targets are a little tricky (since they take a list of files and delete them), which is why CMake doesn't have built in support for it even now.

I'll help if it conflicts with #2338.

CMakeLists.txt
```diff
@@ -155,3 +155,14 @@ if (PYBIND11_INSTALL)
     endif()
   endif()
 endif()
+
+# uninstall target
+if(NOT TARGET uninstall)
```
henryiii, line 160:
```suggestion
if(PYBIND11_MASTER_PROJECT)
```

(you actually don't need to check for the target existing in this case, since master project doesn't have this target)

acxz: Thanks for reviewing @henryiii . This is a good suggestion, didn't even think about it. I'll commit it in.

acxz:
The CI errors seem unrelated

henryiii:
Rebased after #2338 and force-pushed.

henryiii:
Thanks!

PR #2267, opened 2020-06-26
Fix uninitialized-variables cmake warnings

nilsleiffischer:
Fix at least some of the warning that appear when running `pybind11_add_module` in cmake with the `--warn-uninitialized` flag.


wjakob:
Looks good, thanks!

PR #2273, opened 2020-06-28
Render `py::none` as `None` in docstrings

### LINKED ISSUE #2270: Replace the return type of make_iterator from "iterator" to "typing.Iterator"
## Issue description

Currently when declaring a method that returns the result of `py::make_iterator()`, the type signature in the docstring uses the return type `iterator`, which is not a valid Python type. In the example below we see the docstring:

```
__iter__(self: numbers.Numbers) -> iterator\n
```

I was hoping that the type could instead be switched to either `typing.Iterator` or `Iterator`.

```
__iter__(self: numbers.Numbers) -> typing.Iterator\n
```

The use case where this is coming up:
I'm generating type stubs of a pybind module using stubgen. When running mypy with the generated stubs I see the following error:

```
error: name 'iterator' is not defined
```

## Reproducible example code

```cpp

namespace py = pybind11;

static const std::vector<string> numbers = {"1", "2", "3"};

class Numbers {
    std::vector<int>::iterator begin() { return numbers.begin(); }
    std::vector<int>::iterator end() { return numbers.end(); }
}

PYBIND11_MODULE(numbers, m) {
    py::class<Numbers>(m, "Numbers")
        .def("__iter__",
            [](const Numbers& self) {
                return py::make_iterator(self.begin(), self.end());
            },
            py::keep_alive<0, 1>()
        )
    ;
}

```

```python
>>> import numbers
>>> numbers.Numbers.__iter__.__doc__
'__iter__(self: numbers.Numbers) -> iterator\n'
```

AWhetter:
Closes #2270

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1609,6 +1609,7 @@ template <> struct handle_type_name<bytes> { static constexpr auto name = _(PYBI
 template <> struct handle_type_name<int_> { static constexpr auto name = _("int"); };
 template <> struct handle_type_name<iterable> { static constexpr auto name = _("Iterable"); };
 template <> struct handle_type_name<iterator> { static constexpr auto name = _("Iterator"); };
+template <> struct handle_type_name<none> { static constexpr auto name = _("None"); };
 template <> struct handle_type_name<args> { static constexpr auto name = _("*args"); };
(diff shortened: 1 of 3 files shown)


AWhetter:
Thanks for the fast review @wjakob :+1:

PR #2277, opened 2020-07-03
Operator methods include a type signature of the additional overload

### LINKED ISSUE #2269: Include "object" as an argument type in the type signature of operator methods
## Issue description

Currently when declaring an operator method, the type signature in the docstring is not changed to include the additional type. In the example below we see the docstring:

```
__eq__(self: my_class.MyClass, arg0: my_class.MyClass) -> bool
```

however I would expect it to either include the additional type signature that is more generic:

```
__eq__(*args, **kwargs)
Overloaded function.

1. __eq__(self: my_class.MyClass, arg0: my_class.MyClass) -> bool

2. __eq__(self: my_class.MyClass, arg0: object) -> bool
```

or to completely override the type signature:

```
__eq__(self: my_class.MyClass, arg0: object) -> bool
```

The use case where this is coming up:
I'm generating type stubs of a pybind module using stubgen. When running mypy with the generated stubs I see the following error:

```
Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
```

## Reproducible example code

```cpp
# include <pybind11/pybind11.h>

namespace py = pybind11;

class MyClass
{
    MyClass(int id) : m_id(id);
    int m_id;
}

PYBIND11_MODULE(my_class, m) {
    py::class<MyClass>(m, "MyClass")
        .def("__eq__",
            [](const MyClass& self, const MyClass& other) {
                return self.m_id == other.m_id;
            },
            py::is_operator()
        )
    ;
}
```

```python
>>> import my_class
>>> my_class.MyClass.__eq__.__doc__
'__eq__(self: my_class.MyClass, arg0: my_class.MyClass) -> bool\n'
```

AWhetter:
Closes #2269


include/pybind11/pybind11.h
```diff
@@ -250,6 +250,7 @@ class cpp_function : public function {
         /* Generate a proper function signature */
         std::string signature;
         size_t type_index = 0, arg_index = 0;
+        std::string self_type("object");
```
AWhetter, line 253: I've included a default value here because, although unusual, it is a valid to use `py::is_operator()` on a function. In which case this variable would never be set by the loop below.
Using `object` as the default value makes sense because an operator method/function will accept any type when it falls back to returning `NotImplemented`: https://github.com/pybind/pybind11/blob/d54d6d8c613914a7063897d5ddddc62348ffe74a/include/pybind11/pybind11.h#L738-L740

include/pybind11/pybind11.h
```diff
@@ -408,6 +417,12 @@ class cpp_function : public function {
                 if (options::show_function_signatures()) signatures += "\n";
             }
         }
+        if (rec->is_operator) {
+            signatures += "\n";
+            signatures += std::to_string(++index) + ". ";
+            signatures += rec->name;
+            signatures += "(self: " + self_type + ", other: object) -> NotImplemented\n";
```
AWhetter, line 424: Although the type signature of an operator method is actually `(*args: Any, **kwargs: Any) -> NotImplemented` because pybind falls back to returning `NotImplemented` when all other type signatures fail, mypy is specifically looking for the signature implemented here. We could add an additional type signature `(*args: Any, **kwargs: Any) -> NotImplemented` for correctness as well but that feels like overkill to me?

include/pybind11/pybind11.h
AWhetter: This comment is no longer relevant.

AWhetter:
Is there anything that I can do to help progress this pull request?

bstaletic (approved):
Thanks for the ping. There's just a ton of issues and pull requests open and stuff slips through sometimes.

This PR looks good to me.

YannickJadoul:
Let's run such a thing past @wjakob, though.

wjakob:
(More of a general rant, not specifically about this commit -- apologies @AWhetter)
Not a big fan of this change -- it adds code to every pybind11 project that few users will ever care about (I e.g. don't want to pay for it in my own projects). IMO pybind11 docstrings are pretty good. Getting them to "flawless quality" is a dangerous project -- an an almost bottomless rabbithole: each patch just adds a little bit of bloat, there are many further such improvements that folks will submit in the future. I'd rather stop this here. For those who want flawless docstrings, my recommendation would be to fork the library and go the whole way there.

include/pybind11/pybind11.h
```diff
@@ -435,6 +444,16 @@ class cpp_function : public function {
...
         }
+        if (rec->is_operator) {
+            signatures += "\n";
+            signatures += std::to_string(++index) + ". ";
+            signatures += rec->name;
+            signatures += "(";
+            if (rec->is_method) {
+                signatures += "self: " + self_type + ", ";
+            }
+            signatures += "*args, **kwargs) -> NotImplemented\n";
```
YannickJadoul, line 455: Scrutinizing this again: `NotImplemented` is the value, and not the type, right. `type(NotImplemented)` prints `<class 'NotImplementedType'>` for me.

Also, is this the way it's done? Can you give me a reference of type annotations in pure Python code, perhaps?

AWhetter:
I will say that I'm not hugely invested in this change. While I made this for completeness and correctness, for the use case described in the original issue I don't think it's unreasonable to expect someone to have to go through the type stubs generated by stubgen after running it on a pybind module and having to manually curate the output. It does mean that the process goes from being automated to being more manual, but I think it's an acceptable workflow nonetheless.

Overall I still think I would prefer the original issue to be addressed. I think it's worthwhile documenting that you can supply anything to operator methods for users' sake. But it's not the end of the world if we decide not to address it.

@wjakob On the topic of bloat, do you mean bloat in terms of extra run time code, in code complexity, or in terms of final binary size?
I can reduce the size of the code change by half if I don't establish a type for `self`, which the type checking doesn't care about anyway. I was aiming for consistency with the existing docstring output. But maybe that's a wortwhile tradeoff if we're looking to simplify this change?

include/pybind11/pybind11.h
AWhetter: I think I had incorrectly assumed that being a singleton, that `NotImplemented` was treated like `None` and that you're allowed to supply it. In fact, mypy doesn't use or need it at all (https://github.com/python/mypy/issues/6710#issuecomment-485580032) and so it isn't often used. Given that comment, `Any` seems like the correct type to return.
Alternatively, we could combine the two signatures into one and simply say that the type of the second argument is always object, because it will always be accepted and mypy will assume that NotImplemented will be returned regardless. It's less useful for a user that way, but it resolves the original issue without much extra complexity in the implementation.

YannickJadoul:
@AWhetter, as far as I understood the main concern is a steady increase of the binary size, by a series of small, insignificant increases such as this one.

AWhetter:
That makes sense. As mentioned in my previous comment, I can make this change smaller, but I get the impression that even that is too high a cost. I'm happy to close this and close the issue as wontfix if we think that's best?

YannickJadoul:
I wouldn't mind keeping this open for now, and collect it somewhere, in case we would later find a way to opt-in to this.

EricCousineau-TRI:
For objective measurement, c.f. #2760

PR #2283, opened 2020-07-08
Change NAMESPACE_BEGIN and NAMESPACE_END macros into PYBIND11_NAMESPACE_BEGIN and PYBIND11_NAMESPACE_END

### LINKED ISSUE #2199: NAMESPACE_BEGIN macro redefinition
## Issue description

The NAMESPACE_BEGIN is very generic name and therefor there could be the name clash (for instance CryptoPP does use the NAMESPACE_BEGIN also).

It could be fixed renaming the NAMESPACE_BEGIN -> PYBIND11_NAMESPACE_BEGIN. I think this is the only macro that hasn't the PYBIND11 prefix.

YannickJadoul:
Straightforward PR, implementing suggestion from #2199.

Closes #2199.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,9 +147,9 @@ matrix:
     osx_image: xcode7.3
     env: PYTHON=2.7 CPP=14 CLANG CMAKE=1
   - os: osx
-    name: Python 3.7, c++14, AppleClang 9, Debug build
+    name: Python 3.8, c++14, AppleClang 9, Debug build
     osx_image: xcode9.4
-    env: PYTHON=3.7 CPP=14 CLANG DEBUG=1
+    env: PYTHON=3.8 CPP=14 CLANG DEBUG=1
   # Test a PyPy 2.7 build
   - os: linux
     dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -243,7 +243,7 @@ before_install:
       PY_CMD=python$PYTHON
       if [ "$TRAVIS_OS_NAME" = "osx" ]; then
         if [ "$PY" = "3" ]; then
-          brew update && brew unlink python@2 && brew upgrade python
+          brew update && brew unlink python@2 && (brew upgrade python || brew install python)
         else
           curl -fsSL https://bootstrap.pypa.io/get-pip.py | $PY_CMD - --user
         fi
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -12,7 +12,7 @@
 
 #include "cast.h"
 
-NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 /// \addtogroup annotations
 /// @{
@@ -113,7 +113,7 @@ struct call_guard<T, Ts...> {
 
 /// @} annotations
 
-NAMESPACE_BEGIN(detail)
+PYBIND11_NAMESPACE_BEGIN(detail)
 /* Forward declarations */
 enum op_id : int;
 enum op_type : int;
@@ -524,5 +524,5 @@ constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
     return named == 0 || (self + named + has_args + has_kwargs) == nargs;
 }
 
-NAMESPACE_END(detail)
-NAMESPACE_END(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_END(detail)
+PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -11,7 +11,7 @@
 
 #include "detail/common.h"
 
-NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 /// Information record describing a Python buffer object
 struct buffer_info {
@@ -94,7 +94,7 @@ private:
     bool ownview = false;
 };
 
-NAMESPACE_BEGIN(detail)
+PYBIND11_NAMESPACE_BEGIN(detail)
 
 template <typename T, typename SFINAE = void> struct compare_buffer_info {
     static bool compare(const buffer_info& b) {
@@ -110,5 +110,5 @@ template <typename T> struct compare_buffer_info<T, detail::enable_if_t<std::is_
     }
 };
 
-NAMESPACE_END(detail)
-NAMESPACE_END(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_END(detail)
+PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -36,8 +36,8 @@
 #  define PYBIND11_HAS_U8STRING
 #endif
 
-NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
-NAMESPACE_BEGIN(detail)
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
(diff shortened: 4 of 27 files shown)


wjakob:
This is a no-brainer, feel free to merge once the CI passes.

bstaletic:
So CI is broken. Partially because PyPy is broken since yesterday, like we discussed.

The other error is:

```
+brew upgrade python
Error: python not installed
```

YannickJadoul:
> The other error is:
> 
> ```
> +brew upgrade python
> Error: python not installed
> ```

Sigh. How does Travis keep on changing what's default installed with homebrew??

.travis.yml
```diff
@@ -243,7 +243,7 @@ before_install:
       PY_CMD=python$PYTHON
       if [ "$TRAVIS_OS_NAME" = "osx" ]; then
         if [ "$PY" = "3" ]; then
-          brew update && brew unlink python@2 && brew upgrade python
+          brew update && brew unlink python@2 && brew upgrade python || brew install python
```
bstaletic, line 246: This line is wrong. If `brew update` fails, this will still try to execute `brew install python`. Put the last two commands in parenthesis.

YannickJadoul: Alright. That won't matter on these silly failures, though?

YannickJadoul: Fixed.

bstaletic:
Messages from `brew install python`:

```
==> Caveats
Python has been installed as
  /usr/local/opt/python@3.8/bin/python3
You can install Python packages with
  /usr/local/opt/python@3.8/bin/pip3 install <package>
They will install into the site-package directory
  /usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages
```

So it installed python 3.8, not 3.7. This could explain the error:

    ld: library not found for -lpython3.7m

henryiii:
Homebrew finally finished the Python 3.7 to 3.8 transition, was merged yesterday, so this isn't really Travis' fault.

YannickJadoul:
Thanks, @bstaletic and @henryiii. Not behind my laptop, currently. Feel free to push a fix to my branch!

.travis.yml
```diff
@@ -149,7 +149,7 @@ matrix:
   - os: osx
     name: Python 3.7, c++14, AppleClang 9, Debug build
```
YannickJadoul, line 150: Description is still 3.7

henryiii: How did I miss that? I'll have to checkout & force push now...

YannickJadoul:
Thanks Henry!

Something fishy still happened before, btw, because CMake was called with `-DPYBIND11_PYTHON_VERSION=3.7` and reported
```
-- Found PythonInterp: /usr/local/bin/python3.7 (found suitable version "3.7.5", minimum required is "3.7") 
-- Found PythonLibs: python3.7m
```

That's why I didn't look further at HomeBrew's installation...

.travis.yml
YannickJadoul: I can do it in a few hours as well!

henryiii:
`python` became `python@3.7`, `python@3.8` became `python` and was allowed to override the links in `/usr/local/`. It looks to me that 3.7 didn't become keg-only, though I rather would have thought it should have. python3.7 is still installed and still available.

.travis.yml
henryiii: Done.

YannickJadoul:
Ah, right, thanks for the explanation! I don't know brew's internals thát well :)

bstaletic:
Someone could cancel the second to last appveyor run and save this PR ~20 minutes.

YannickJadoul:
Appveyor doesn't give collaborators access. It's on wjakob's account and we'd need an explicit invitation (which is probably not worth the trouble if we're switching to GHA soon enough).

henryiii:
@wjakob already approved, so merged. Thanks!

YannickJadoul:
3 minutes too late to merge! Thanks for the final fixes, @henryiii!

patrikhuber:
Hi,

I updated from an older pybind to the recent master today and was greeted with a bunch of compiler error messages. If I see it correctly, this is a change that came after the 2.5.0 release so I'm probably ahead of the curve. I couldn't find anything in the docs about it. I would strongly recommend to mention this in the Upgrade Guide https://pybind11.readthedocs.io/en/stable/upgrade.html.

Also the typecaster example in the docs https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html uses the manual `namespace pybind11 { namespace detail {`. In fact, a search in the documentation for `NAMESPACE_BEGIN` or `PYBIND11_NAMESPACE_BEGIN` yields no (useful) results. I would recommend adding the macro to the docs and in particular to the custom type caster example?

Thank you all for working on and improving pybind11, it's one of the most awesome libraries/tools out there! :-)

patrikhuber:
I'm actually now wondering what this macro is for - as you still need to specify the actual name of the namespace, e.g.:
```
PYBIND11_NAMESPACE_BEGIN(pybind11)
```
I can't see any benefits? It's not shorter (longer even) and you still need to remember and "manually" type the name of the actual namespace, so you might as well just type
```
namespace pybind11 {
namespace detail {
```

It would be useful if you could just do:
```
PYBIND11_NAMESPACE_BEGIN
// ... my typecaster code ...
PYBIND11_NAMESPACE_END
```
In that way, as a user/consumer of pybind11 I don't have to remember what namespace to define my stuff in, and if you pybind11 folks ever decide to change the namespace, it's future proof (I thought that's what the purpose of the macro is, but now see that that's currently not the case).

YannickJadoul:
This is an undocumented macro, though, so I'm not sure pybind11 is responsible for breaking undocumented stuff? ;-) (Then again, I also don't really know why the macros are actually there?)

There's also not a new release yet, and I don't think it's clear yet whether it will be a patch, minor, or major release, so it's hard to write this part of the docs (and anyway, it won't show up on the "stable" version of RTD!). Though I guess it could be a good plan to keep a list of changes to master, there (@henryiii did start a "pending" changelog, I believe).

> I can't see any benefits?

Not sure why it's there, but there must be a reason?


> It would be useful if you could just do:
> 
> ```
> PYBIND11_NAMESPACE_BEGIN
> // ... my typecaster code ...
> PYBIND11_NAMESPACE_END
> ```

That wouldn't really make sense, because it's in `pybind11::detail`, and adding detail by defaults will give other problems (because obviously, some things  (actually, all "public" things) are nót in `detail`. The casters are the weird one out because you are allowed to create your own caster, but I believe pybind11 doesn't guarantee that the interface won't break, so that (and historical reasons) is why it's in `detail`.

*PS: Expecting @bstaletic to pop up any time now, shouting "Hyrum's law".*

bstaletic:
A few thoughts.

1. `NAMESPACE_BEGIN` and `NAMESPACE_END` are horrible macro names that should never appear in any code base.
2. The docs don't mention it because there's no reason to.
3. The docs use `namespace pybind11 { namespace detail {` because that's what is officially supported.
4. Does pybind11 have users who depend on the exact definition of header guards?

I'm honestly surprised that anyone would try to depend on something like `NAMESPACE_BEGIN`. As it just doesn't come up in the docs or issues, except in some diffs, one would *almost* have to look for these macros and use something out of contract.

That said...

> PS: Expecting @bstaletic to pop up any time now, shouting "Hyrum's law".

Exactly!

henryiii:
> Hyrum's law

That's exactly what I thought of when I saw this. :)

henryiii:
I didn't remember the upgrade guide, thanks! Will make sure CMake related things go in there soon!

patrikhuber:
I see! Makes complete sense. I agree then it should not be mentioned in the docs or upgrade guide, as it's internal-use only. I've removed it from my type-casters and changed it to `namespace pybind11 { namespace detail {`.

I don't remember why I used that macro or where I found it, but I don't think I was specifically looking for it. If I had to venture a guess I would say that it (mistakenly, in that case) might have been part of some pybind11 example typecaster code somewhere a few years ago, and I copied it from there. But anything is possible.

Thanks a lot, and I guess "sorry for using undocumented stuff" :-).

YannickJadoul:
Thanks for the friendly reminder that we need to start thinking about this upgrade guide ;-)

Issue #2109, opened 2020-02-07
Missing argument error messages should match python's


virtuald:
## Issue description

When a user neglects to provide the correct number of arguments, pybind11 gives a different error message than that which python users might expect. In my opinion, the python error message is clearer.

## Reproducible example code

```c++
int add(int i, int j) {
    return i + j;
}
```
```py
def pyadd(i, j):
    return i + j
```

So if I call the python snippet:
```
>>> pyadd(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: pyadd() missing 1 required positional argument: 'j'
```

Whereas the pybind11 snippet is:

```
>>> add(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: add(): incompatible function arguments. The following argument types are supported:
    1. (i: int, j: int) -> int

Invoked with: 1
```

Now, this isn't a great example because the types are really short. But as a user, the first thing that I'm scanning for is the types and whether my types match -- it takes some effort to figure out the number is incorrect. Here's a more real example from pybind11:

```
Traceback (most recent call last):
  File "C:\Users\first\AppData\Local\Programs\Python\Python38\lib\site-packages\wpilib\_impl\start.py", line 106, in start
    self.robot.startCompetition()
  File "robot.py", line 84, in teleopPeriodic
    self.drive.arcadeDrive(self.trig, self.con0.getX())
TypeError: getX(): incompatible function arguments. The following argument types are supported:
    1. (self: wpilib._wpilib.XboxController, hand: wpilib.interfaces._interfaces.GenericHID.Hand) -> float

Invoked with: <wpilib._wpilib.XboxController object at 0x0000022286350F30>
```

There's a lot of punctuation in that error message, and the actual issue (missing argument) is obscured by everything. This gets worse as the number of arguments increases.

bstaletic:
I'm not sure about this. If pybind matches python's format, the type information is lost. Considering that we're talking C++, type information is important.

YannickJadoul:
The reason pybind11 gives this error message, is because it's also doing overload resolution, I think. This becomes clearer if you would bind both `int add(int, int)` and some (hypothetical) `std::string add(std::string)`; then pybind11 doesn't know if you missed a second argument or passed the wrong type for the argument.

But if you have a suggestion to change these error message that takes all of these possibilities into account (and can be implemented reasonably straightforwardly), we could definitely consider and discuss it further!

[...]

PR #2284, opened 2020-07-09
WIP: Provide simple hints for function call argument mismatches in some cases

virtuald:
Partial fix for https://github.com/pybind/pybind11/issues/2109, this will sometimes add error hints to the user for each overload that doesn't match. It should only impact performance if the function call is incorrect -- but exceptions are exceptional so this shouldn't be unexpected.

Here's what you get in a single overload case, mismatched parameters, additional error information highlighted:

```patch
 TypeError: test(): incompatible function arguments. The following argument types are supported:
     1. (a: int, b: int) -> int
+    ~> takes 2 positional arguments but 3 were given
 
 Invoked with: 1, 2, 3
```

Or let's say there are multiple overloads that don't match:

```patch
 TypeError: test(): incompatible function arguments. The following argument types are supported:
     1. (a: int, b: int) -> int
+    ~> takes 2 positional arguments but 3 were given
     2. (a: int, b: int, c: int, d: int) -> int
+    ~> missing 1 required positional argument
 
 Invoked with: 1, 2, 3
```

Looking at ceval.c in CPython, a full fix would be a lot of work! But these additional error messages seem like they would make it a little easier for the user to determine what went wrong at a glance, similar to what Python's error messages normally do.

Feedback wanted for message formatting, other ideas, sufficiency, etc. If all is well, I can add some tests somewhere.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -768,6 +768,39 @@ protected:
                 }
                 if (!wrote_sig) msg += it2->signature;
 
+                // When reasonable, provide guidance for simple reasons why each
+                // overload doesn't match. Matches the dispatch code above, but
+                // duplicated for performance reasons
+
+                const function_record &func = *it2;
+                size_t num_args = func.nargs;    // Number of positional arguments that we need
+                if (func.has_args) --num_args;   // (but don't count py::args
+                if (func.has_kwargs) --num_args; //  or py::kwargs)
+                size_t pos_args = num_args - func.nargs_kwonly;
+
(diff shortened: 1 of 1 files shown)


YannickJadoul:
I think I could be convinced to include this. The only two arguments against would be the compiled size overhead (we could get an estimate, though; there are some plans to work on this infrastructure) and the complexity of the `dispatcher` function (but it might already need a refactoring?).

Further feedback: what's the plan to handle optional arguments? E.g., passing `1, 2, 3` to `(a: int, b: int, c: int, d: int, e: int = 0, f: int = 42)` ?

PR #2289, opened 2020-07-10
Don't add duplicate patients

### LINKED ISSUE #1251: Using `py::keep_alive<N, P>()` may cause `internals.patients[nurse]` to grow without bounds

EricCousineau-TRI:
Fixes #1251
Either an alternative or precursor for #1253

This is the forward-port of 56c1edb from v2.2.2, as mentioned here:
https://github.com/pybind/pybind11/issues/1251#issuecomment-656741405

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -317,9 +317,13 @@ extern "C" inline int pybind11_object_init(PyObject *self, PyObject *, PyObject 
 inline void add_patient(PyObject *nurse, PyObject *patient) {
     auto &internals = get_internals();
     auto instance = reinterpret_cast<detail::instance *>(nurse);
+    auto &current_patients = internals.patients[nurse];
     instance->has_patients = true;
+    for (auto &p : current_patients)
+        if (p == patient)
+            return;
     Py_INCREF(patient);
-    internals.patients[nurse].push_back(patient);
+    current_patients.push_back(patient);
 }
 
 inline void clear_patients(PyObject *self) {
--- a/tests/test_call_policies.py
+++ b/tests/test_call_policies.py
@@ -1,6 +1,6 @@
 import pytest
 from pybind11_tests import call_policies as m
-from pybind11_tests import ConstructorStats
+from pybind11_tests import ConstructorStats, UserType
 
 
 def test_keep_alive_argument(capture):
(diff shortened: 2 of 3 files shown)


tests/test_call_policies.py
```diff
@@ -69,6 +69,35 @@ def test_keep_alive_return_value(capture):
     """
 
 
+def test_keep_alive_single():
+    """Issue #1251 - patients are stored multiple times when given to the same nurse"""
+
+    nurse, p1, p2 = UserType(), UserType(), UserType()
```
EricCousineau-TRI, line 75: Same as #1253 - I would like to modify this test to make `refcount()` Pythonic, and check behavior after `nurse` is GC'd.

EricCousineau-TRI: Done. Cherry-picked 3fb4858be7afe6ed75e22c0026c480c3ea5f55f9 from #1253.

include/pybind11/detail/class.h
```diff
@@ -317,9 +317,13 @@ extern "C" inline int pybind11_object_init(PyObject *self, PyObject *, PyObject
 inline void add_patient(PyObject *nurse, PyObject *patient) {
     auto &internals = get_internals();
     auto instance = reinterpret_cast<detail::instance *>(nurse);
+    auto &current_patients = internals.patients[nurse];
     instance->has_patients = true;
+    for (auto &p : current_patients)
```
YannickJadoul, line 322: This could be `if (std::find(current_patients.begin(), current_patients.end(), patient) != current_patients.end())`, I suppose. Normally that would be cleaner, but now it's actually lengthier. So probably not worth it.

Issue #1704, opened 2019-02-22
Segfault from py::str()


davidhewitt:
In the sample below:

https://github.com/pybind/pybind11/blob/25abf7efba0b2990f5a6dfb0a31bc65c0f2f4d17/include/pybind11/pytypes.h#L928-L936

I'm pretty sure the `#if PY_MAJOR_VERSION < 3` and `if (!str_value) throw error_already_set();` lines are the wrong way around. As it stands, on Python 3 then if the `PyObject_Str()` call fails then a nullptr becomes the ptr held by a `py::str` object, leading to possible segfaults.

PR #2290, opened 2020-07-11
[READY] Check for NULL in raw_str on Python3

bstaletic:
Fixes https://github.com/pybind/pybind11/issues/1704


YannickJadoul:
Thanks for the quick fix, @bstaletic, and for reporting in #1704, @davidhewitt!

Issue #2278, opened 2020-07-06
error: no matching function for call to ‘pybind11::class_<Operations<int> >


matthrx:
Hello,

I am trying to use the pybind11 library with a very basic example. I am trying to apply pybind11 to a template class doing basic operations (for the example). I did research for a long time in order to know whether someone had faced a similar mistake but so far they are fruitless, that's why I decided to open this issue.

Here is the template class Operations (i don't include the cpp file as function are extremely simple)
![subtract](https://user-images.githubusercontent.com/49817172/86586404-1eb72d00-bf88-11ea-96c6-963c78a6b554.png)

And here is the file used to bind it to a python library 
![binding](https://user-images.githubusercontent.com/49817172/86586559-5a51f700-bf88-11ea-949f-3845df23beb9.png)

When I attempt to install it, I have the following error : 
error: no matching function for call to ‘pybind11::class_<Operations<int> >::def(std::__cxx11::basic_string<char>, int (Operations<int>::*)(int, int))’
   26 |       .def("subtract"+type, &operations::subtract);

Thank you for your time

PS : It is the first time I open an issue, I am sincerely sorry if I was not supposed to do it for this kind of error that I think is only because I did something wrong. Obviously, I looked into the faq.html of pybind11 and found nothing relevant for my case. I didn't add the CMakeLists.txt and the setup.py, I will, if you think the error might come from them.

YannickJadoul:
The issue seems to be that `class_::def` is expecting a `const char *` as first `name` argument, and you're giving it a `std::string`. As far as I can quickly check, I believe `name` is copied inside of the call to `def`, so `("add"+type).c_str()` might work, but best be careful and check.

> PS : It is the first time I open an issue, I am sincerely sorry if I was not supposed to do it for this kind of error that I think is only because I did something wrong. Obviously, I looked into the faq.html of pybind11 and found nothing relevant for my case. I didn't add the CMakeLists.txt and the setup.py, I will, if you think the error might come from them.

No worries; thanks for all the information and self-contained exampled. Since you're asking yourself, two small tips:
1) If you post code as text (between ```) instead of screenshots, it's a lot easier for other to investigate by copying your code.
2) For this kind of question, rather than an actual/potential issue with pybind11, there's [Gitter](https://gitter.im/pybind/Lobby)

PR #2291, opened 2020-07-11
Set __hash__ to None for types that defines __eq__ but not __hash__

### LINKED ISSUE #2191: Overriding __eq__ does not set __hash__ to None

sizmailov:
PR patches `py::class_<>::def` as discussed in  https://github.com/pybind/pybind11/pull/2287

fixes #2191

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1047,6 +1047,13 @@ inline void call_operator_delete(void *p, size_t s, size_t a) {
     #endif
 }
 
+inline void add_class_method(object& cls, const char *name_, const cpp_function &cf) {
+    cls.attr(cf.name()) = cf;
+    if (strcmp(name_, "__eq__") == 0 && !cls.attr("__dict__").contains("__hash__")) {
+      cls.attr("__hash__") = none();
+    }
+}
+
 PYBIND11_NAMESPACE_END(detail)
 
 /// Given a pointer to a member function, cast it to its `Derived` version.
@@ -1144,7 +1151,7 @@ public:
     class_ &def(const char *name_, Func&& f, const Extra&... extra) {
         cpp_function cf(method_adaptor<type>(std::forward<Func>(f)), name(name_), is_method(*this),
                         sibling(getattr(*this, name_, none())), extra...);
-        attr(cf.name()) = cf;
+        add_class_method(*this, name_, cf);
         return *this;
     }
 
--- a/tests/test_operator_overloading.cpp
+++ b/tests/test_operator_overloading.cpp
@@ -187,6 +187,38 @@ TEST_SUBMODULE(operators, m) {
         .def(py::self *= int())
         .def_readwrite("b", &NestC::b);
     m.def("get_NestC", [](const NestC &c) { return c.value; });
+
+
+    // test_overriding_eq_reset_hash
+    // #2191 Overriding __eq__ should set __hash__ to None
+    struct Comparable {
+        int value;
+        bool operator==(const Comparable& rhs) const {return value == rhs.value;}
+    };
+
+    struct Hashable : Comparable {
+        explicit Hashable(int value): Comparable{value}{};
+        size_t hash() const { return static_cast<size_t>(value); }
+    };
+
+    struct Hashable2 : Hashable {
+        using Hashable::Hashable;
+    };
+
+    py::class_<Comparable>(m, "Comparable")
+        .def(py::init<int>())
+        .def(py::self == py::self);
+
+    py::class_<Hashable>(m, "Hashable")
+        .def(py::init<int>())
+        .def(py::self == py::self)
+        .def("__hash__", &Hashable::hash);
+
+    // define __hash__ before __eq__
+    py::class_<Hashable2>(m, "Hashable2")
+        .def("__hash__", &Hashable::hash)
+        .def(py::init<int>())
+        .def(py::self == py::self);
 }
 
 #ifndef _MSC_VER
--- a/tests/test_operator_overloading.py
+++ b/tests/test_operator_overloading.py
@@ -126,3 +126,19 @@ def test_nested():
     assert abase.value == 42
     del abase, b
     pytest.gc_collect()
+
+
+def test_overriding_eq_reset_hash():
+
+    assert m.Comparable(15) is not m.Comparable(15)
+    assert m.Comparable(15) == m.Comparable(15)
+
+    with pytest.raises(TypeError):
+        hash(m.Comparable(15))  # TypeError: unhashable type: 'm.Comparable'
+
+    for hashable in (m.Hashable, m.Hashable2):
+        assert hashable(15) is not hashable(15)
+        assert hashable(15) == hashable(15)
+
+        assert hash(hashable(15)) == 15
+        assert hash(hashable(15)) == hash(hashable(15))


include/pybind11/pybind11.h
```diff
@@ -1144,7 +1151,7 @@ class class_ : public detail::generic_type {
     class_ &def(const char *name_, Func&& f, const Extra&... extra) {
         cpp_function cf(method_adaptor<type>(std::forward<Func>(f)), name(name_), is_method(*this),
                         sibling(getattr(*this, name_, none())), extra...);
-        attr(cf.name()) = cf;
```
sizmailov, line 1147: Does anyone know why this line was not `attr(name_) = cf;` ? Just curious.

bstaletic: It used to be `attr(name_) = cpp_function( ... )` until [this commit](https://github.com/pybind/pybind11/commit/570822102c037600781da816d1a083db031c8cde#diff-6645501a414a69e6b16abfe4bee9025fR676).

sizmailov: Indeed, it's intended change. Had to do `blame` myself. Thanks!

include/pybind11/pybind11.h
```diff
@@ -1047,6 +1047,13 @@ inline void call_operator_delete(void *p, size_t s, size_t a) {
     #endif
 }
 
+inline void add_class_method(object& cls, const char *name_, const cpp_function &cf) {
```
YannickJadoul, line 1050: Just curious: If this PR does reduce the binary size, this `inline` is just ignored by the compiler and is only necessary to avoid the ODR?

sizmailov: Correct.

YannickJadoul:
Oh, what about docs, cfr. #2287? Do we need to mention this now or is it just "fixed". Or are we merging that as reworked #2278?

bstaletic:
> Oh, what about docs, cfr. #2287? Do we need to mention this now or is it just "fixed". Or are we merging that as reworked #2278?

Unless I'm missing something, we can just treat #2191 as resolved and leave the docs alone.

YannickJadoul:
> Unless I'm missing something, we can just treat #2191 as resolved and leave the docs alone.

Fine with me. I just saw @jbarlow83's remark (https://github.com/pybind/pybind11/pull/2287#discussion_r453143724).
@ixje, you brought up this issue in the first place; do you think further docs are needed, or is just being consistent with Python good enough?

tests/test_operator_overloading.cpp
```diff
@@ -187,6 +187,38 @@ TEST_SUBMODULE(operators, m) {
...
+    // test_overriding_eq_reset_hash
+    // #2191 Overriding __eq__ should set __hash__ to None
+    struct Comparable {
+        int value;
+        bool operator==(const Comparable& rhs) const {return value == rhs.value;}
+    };
+
+    struct Hashable : Comparable {
+        explicit Hashable(int value): Comparable{value}{};
+        size_t hash() const { return value % 32; /* full of collisions */ }
```
rwgk, line 201: Is "full of collisions" important to the test? The test only uses one value (15).

tests/test_operator_overloading.py
```diff
@@ -126,3 +126,18 @@ def test_nested():
...
+    assert m.Comparable(15) == m.Comparable(15)
+
+    with pytest.raises(TypeError):
+        hash(m.Comparable(15))  # TypeError: unhashable type: 'm.Comparable'
+
+    for hashable in (m.Hashable, m.Hashable2):
+        assert hashable(15) is not hashable(15)
+        assert hashable(15) == hashable(15)
+
+        assert hash(hashable(15)) == hash(hashable(15))
```
rwgk, line 143: Is this to prove that the correct hash implementation is used?
Does `hash(hashable(15)) == hash(hashable(15 + 32))` make sense?
To prove that the `% 32` is actually used.

YannickJadoul: If the default `id`-based comparison would be used, this assertion would fail (I believe? Unless the memory location of the temporary `hashable(15)` gets recycled?

rwgk: Yeah, I agree, but it is so cheap to be more certain, I'd definitely seize the opportunity.

ixje:
> @ixje, you brought up this issue in the first place; do you think further docs are needed, or is just being consistent with Python good enough?

I did the docs PR when we still had the believe that the impact of actually solving it was too high. I'll close that one in favour of this, as it is no longer required. Thanks everybody for resolving this ❤️

tests/test_operator_overloading.cpp
sizmailov: Not really. Can be reduced to `return value`. I'll update PR.

tests/test_operator_overloading.py
sizmailov: The `__hash__` contract implies that equal object has same hash. This is what is checked  here (along with hashibility of the objects). 

>  Does hash(hashable(15)) == hash(hashable(15 + 32)) make sense?

Only as verification of concrete implementation. The exact hash values are implementation detail. 

@YannickJadoul, you are correct, default hash would fail here.

YannickJadoul: > @YannickJadoul, you are correct, default hash would fail here.

Did you try this? Because I could imagine Python creating a `hashable(15)` at memory location `x`, computing the `hash`, deallocating the `hashable` object as it goes to refcount 0, the creating `hashable(15)` at the same memory location, and computing the same `id`-based hash?

sizmailov: Yes, I've tried to mock it with 
```python
        assert hashable(15) is not hashable(15)
        assert hashable(15) == hashable(15)
        del hashable.__hash__
        assert hash(hashable(15)) == hash(hashable(15))
```
and it fails with 

```
>           assert hash(hashable(15)) == hash(hashable(15))
E           AssertionError: assert -9223363257492360940 == 8779362415047
E            +  where -9223363257492360940 = hash(<pybind11_tests.operators.Hashable object at 0x7fc1ac731148>)
E            +    where <pybind11_tests.operators.Hashable object at 0x7fc1ac731148> = <class 'pybind11_tests.operators.Hashable'>(15)
E            +  and   8779362415047 = hash(<pybind11_tests.operators.Hashable object at 0x7fc1ac731c70>)
E            +    where <pybind11_tests.operators.Hashable object at 0x7fc1ac731c70> = <class 'pybind11_tests.operators.Hashable'>(15)
```

YannickJadoul: I actually just tried that. I knew I'd seen this Python trap before:

```pycon
>>> class X: pass
... 
>>> id(X()), id(X())
(140256801227776, 140256801227776)
>>> hash(X()) == hash(X())
True
>>> x = X()
>>> y = X()
>>> id(x), id(y)
(140256801227776, 140256801227832)
>>> hash(x) == hash(y)
False
```

The same holds for running as a script:

```python
class X: pass

print(id(X()), id(X()))
print(hash(X()) == hash(X()))

x, y = X(), X()

print(id(x), id(y))
print(hash(x) == hash(y))
```

```
$ python3 test.py 
140000331939064 140000331939064
True
140000331939064 140000331959096
False
```

sizmailov: Hm. I don't mind to test against actual values of `hash(hashable(...))` to verify that we use the definition we want.

sizmailov: I've added `assert hash(hashable(15)) == 15` so we know for sure the default `id`-based implementation is not involved.

YannickJadoul: OK, that's also fine. I was just thinking to not have them as temporary objects, but this does the same, I suppose.

YannickJadoul: > Yes, I've tried to mock it with
> 
> ```python
>         assert hashable(15) is not hashable(15)
>         assert hashable(15) == hashable(15)
>         del hashable.__hash__
>         print(hashable.__hash__)
>         assert hash(hashable(15)) == hash(hashable(15))
> ```
> 
> and it fails with
> 
> ```
> <slot wrapper '__hash__' of 'object' objects>
> 
> ...
> 
> >           assert hash(hashable(15)) == hash(hashable(15))
> E           AssertionError: assert -9223363257492360940 == 8779362415047
> E            +  where -9223363257492360940 = hash(<pybind11_tests.operators.Hashable object at 0x7fc1ac731148>)
> E            +    where <pybind11_tests.operators.Hashable object at 0x7fc1ac731148> = <class 'pybind11_tests.operators.Hashable'>(15)
> E            +  and   8779362415047 = hash(<pybind11_tests.operators.Hashable object at 0x7fc1ac731c70>)
> E            +    where <pybind11_tests.operators.Hashable object at 0x7fc1ac731c70> = <class 'pybind11_tests.operators.Hashable'>(15)
> ```

Ah, wait. Hadn't seen that. Interesting, so here it doesn't do that? Maybe because of the AST rewrite pytest does?

sizmailov: Whatever it is I don't think it's a reliable behavior, i.e. it might change from version to version from one day of week to another. I think now we are backed with better test case.

YannickJadoul: Agreed! :-)

YannickJadoul:
I think this was approved by everyone? I'm merging.

Thank you very much, once again, @sizmailov! :-)

Issue #2159, opened 2020-04-09
ci: Travis Job for Python 3.8 on Linux does not actually run Python 3.8?


EricCousineau-TRI:
## Issue description

If you look at any job `Python 3.8, c++17, gcc 7 (w/o numpy/scipy)` in Travis CI builds, they don't seem to actually be using Python 3.8. (Not sure if this is due to a patch I submitted...)

Came across this is in [our fork](https://github.com/RobotLocomotion/pybind11), which effectively uses the same CI scripts here.

## Reproducible example code

See this build:

https://travis-ci.org/github/pybind/pybind11/jobs/640390164#L467
> -- Found PythonInterp: /opt/pyenv/shims/python3 (found version "3.6.7")

carlsonmark:
Do you think that changing `PY_CMD=python3` to `PY_CMD=python3.8` would be sufficient to fix that? Or is there more todo?

EricCousineau-TRI:
Yup, I think that would be it!

Testing it out in my fork, can PR the patch if I can confirm it works:
https://github.com/RobotLocomotion/pybind11/pull/40

(I may also disable one of the PyPy jobs, as that seems to be freezing on a download?)

carlsonmark:
OK. I think if that doesn't work, then maybe some clever use of `update-alternatives` may help. I'm not really familiar with any of that though. I'm just starting to look into travis-ci myself, here.

Good luck!

carlsonmark:
Looks like you will need to add this package too:
```
          - python3.8-distutils
```
Which allows `python3.8 -m pip` to work on xenial.

carlsonmark:
My mistake, I had installed python3.8 from snap a while ago and uninstalled it, so my environment is not clean.

This may be the way to install pip in python3.8 using deadsnakes: https://github.com/deadsnakes/issues/issues/95

```
curl https://bootstrap.pypa.io/get-pip.py | python3.8
```
But you might want to read that page first in case you think it may be a bad idea.

EricCousineau-TRI:
Hm... Lemme try out `python3 -m ensurepip`, as that tends to work when build CPython from source. Thanks!

chaitan94:
Possibly related: https://github.com/pybind/pybind11/issues/1090

henryiii:
This has all been wiped out and replaced by GHA, and those do run the correct versions. :)

PR #2292, opened 2020-07-11
Fix compilation of tests with `/std:c++17`

### LINKED ISSUE #2089: Pybind fails to build with MSVC 19.16 in C++17 mode
I used pybind v 2.4.3

> ninja

Error:
        with
        [
            _Ty=void
        ]

Great library, BTW! 😃 👍

YannickJadoul:
Seems it needs to be closed to get it passed correctly?

Closes #2089

So this first commit should fail, given #2089.


YannickJadoul:
Am I going crazy? The debug runs print `PYBIND11_CPP_STANDARD=/std:c++17` from CMake, but the actual compile command contains `/std:c++14`???

YannickJadoul:
Seriously!?

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

> $<COMPILE_LANGUAGE:languages>
> ...
> Note that with Visual Studio Generators and Xcode there is no way to represent target-wide compile definitions or include directories separately for C and CXX languages. Also, with Visual Studio Generators there is no way to represent target-wide flags separately for C and CXX languages. Under these generators, expressions for both C and C++ sources will be evaluated using CXX if there are any C++ sources and otherwise using C. A workaround is to create separate libraries for each source file language instead:

henryiii:
The correct solution probably would be to use compile_features; when we make this 3.7+ only we should be able to change this properly.

YannickJadoul:
> The correct solution probably would be to use compile_features; when we make this 3.7+ only we should be able to change this properly.

Yes, I had thought of that as well. But depends on dropping old CMake support, then.

> But yes, this is a general issue that always needs a workaround for these IDE's.

Good point. I'll check if there's more of these around!

(Should I also check for Xcode, btw, or not?)

henryiii:
If it doesn't work on Xcode, I'd likely know pretty quickly. ;)

I can't think of any more of these that are likely to be a problem, but I wouldn't have remembered the one above (I had to fix in in CLI11 a long time ago). I know of some that probably don't affect us (like the `...` syntax in version support being broken in MSVC's forked copy of CMake 3.11)

YannickJadoul:
Thanks! I found 2 more occurences of `$<COMPILE_LANGUAGE:CXX>`: one for the embedding target (same fix), and one here: https://github.com/pybind/pybind11/blob/8b9eb964d904c1dab33ed819497fb997f25e473b/tools/pybind11Tools.cmake#L257

Not sure what the last one does, but it's already MSVC-specific. Any idea what to do? Just ignore, since it's only about parallel builds?

henryiii:
Maybe tag it with a TODO? Is there a reason that `cmake --build -j<N>` doesn't work for parallel builds? Manually sticking specific compile options is usually a bad idea. And, due to the COMPILE_LANGUAGE, this one should be a no-op.

YannickJadoul:
> Maybe tag it with a TODO? Is there a reason that `cmake --build -j<N>` doesn't work for parallel builds? Manually sticking specific compile options is usually a bad idea. And, due to the COMPILE_LANGUAGE, this one should be a no-op.

No clue. Never really touched the CMake internals before. I'll just apply the same fix anyway. If someone passes by that knows more about CMake, we can still change it? :-)

YannickJadoul:
Builds are still using `/std:c+14`!? Am I somehow going crazy? I thought I had found the issue :-(

bstaletic:
I know the title says `/std:c++17`, but could we maybe include this as well: https://github.com/pybind/pybind11/issues/2159

YannickJadoul:
> I know the title says `/std:c++17`, but could we maybe include this as well: #2159

Yes, good point. I was planning to see if supporting VS 2019 would still be possible, but this is also something to add!
Once I get things to fail, that is ...

.appveyor.yml
```diff
@@ -57,14 +56,15 @@ install:
     7z x 3.3.3.zip -y > $null
     $env:CMAKE_INCLUDE_PATH = "eigen-eigen-67e894c6cd8f;$env:CMAKE_INCLUDE_PATH"
 build_script:
+- set VERBOSE=1
+- cmake --version
 - cmake -G "%CMAKE_GENERATOR%" -A "%CMAKE_ARCH%"
-    -DPYBIND11_CPP_STANDARD=/std:c++%CPP%
+    -DPYBIND11_CPP_STANDARD=/std:c++%CPP
```
bstaletic, line 62: You stripped the trailing `%` here. `CPP` isn't treated as a variable in this case.

YannickJadoul: Thanks. Committed too quickly!

tools/pybind11Tools.cmake
```diff
@@ -218,8 +220,10 @@ function(pybind11_add_module target_name)
   # Make sure C++11/14 are enabled
   if(PYBIND11_CPP_STANDARD)
     if(CMAKE_VERSION VERSION_LESS 3.3 OR CMAKE_GENERATOR MATCHES "Visual Studio")
+      message("HERE, I hope?")
       target_compile_options(${target_name} PUBLIC ${PYBIND11_CPP_STANDARD})
     else()
+      message("NOT HERE, I hope?")
```
LeslieGerman, line 226: Have you left these messages in by intent ? (on line 223, too)

LeslieGerman:
@YannickJadoul 
> Thanks! I found 2 more occurences of `$<COMPILE_LANGUAGE:CXX>`: one for the embedding target (sa
> target_compile_options(${target_name} PRIVATE $<$<NOT:$<CONFIG:Debug>>:$<$<COMPILE_LANGUAGE:CXX>:/MP>>) 
> Not sure what the last one does ...

[/MP: Build with Multiple Processes](https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2017)

According to my experiences in previous years, it may or may not improve compilation time. See also [MS own Guideline](https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2017#guidelines).

If compilation speed is not a key factor here, you can safely omit this option IMO.

tools/pybind11Tools.cmake
YannickJadoul: Yes. This is in the commit with message "Debugging non-failure of /std:c++17", and this PR is not ready, yet.

YannickJadoul: Yes. This is in the commit with message "Debugging non-failure of /std:c++17", and this PR is not ready, yet.

YannickJadoul:
> [/MP: Build with Multiple Processes](https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2017)

Yes, that I can look up. My question was just wondering out loud why it's in pybind11's CMake and why not in debug, etc. (Which I might be able to look up through some `git blame`, yes.)

bstaletic:
What's the status of this? Are we just going to mandate `/permissive-`?

YannickJadoul:
Yes, but I haven't gotten to cleaning up this PR and adding it to the docs. Feel free to take over, if you want?

PR #2293, opened 2020-07-12
[READY] Small fixes in numpy.h

### LINKED ISSUE #1599: array_t constructor takes shape as size_t, not ssize_t
## Issue description

The array_t constructor takes its input shape as unsigned size_t (== npt_uintp), even though the numpy API (e.g. PyArray_NewFromDescr) uses signed ssize_t (== npy_intp) and pybind11 itself uses signed sizes in most places (e.g. `ssize_t shape(ssize_t dim)`).

## Reproducible example code

Use the following snippet into the setup provided in the pybind11/python_example repo:
```
#include <pybind11/numpy.h>

namespace py = pybind11;

py::array_t<int> make_array(ssize_t size) {
    return py::array_t<int>{{size}};
}

    m.def("make_array", &make_array);
}
```
Compilation (pybind11 2.2.4, gcc 8.2.1) yields
```
src/main.cpp: In function ‘pybind11::array_t<int> make_array(ssize_t)’:
src/main.cpp:7:35: warning: narrowing conversion of ‘size’ from ‘ssize_t’ {aka ‘long int’} to ‘pybind11::size_t’ {aka ‘long unsigned int’} inside { } [-Wnarrowing]
     return py::array_t<int>{{size}};
```

bstaletic:
- `PyArray_NewFromDescr_` should have been using `Py_intptr_t const *`
  - https://github.com/numpy/numpy/blob/18a6e3e505ee416ddfc617f3e9afdff5a031c2c2/numpy/core/src/multiarray/ctors.h#L5-L8
- `PyArray_GetArrayParamsFromObject_` should be using `NPY_BOOL`
  - https://github.com/numpy/numpy/blob/18a6e3e505ee416ddfc617f3e9afdff5a031c2c2/numpy/core/src/multiarray/ctors.c#L1350-L1355
  - https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.dtype.html#c.npy_bool
- `PyArray_DescrNewFromType` is at offset 96
  - https://github.com/numpy/numpy/blob/36e017194c32a53cf2965a513243cb3c348095df/numpy/core/code_generators/numpy_api.py#L141
- `array_t` constructor that takes a size and a pointer should take `ssize_t`.
  - Fixes #1599

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -181,8 +181,9 @@ struct npy_api {
     unsigned int (*PyArray_GetNDArrayCFeatureVersion_)();
     PyObject *(*PyArray_DescrFromType_)(int);
     PyObject *(*PyArray_NewFromDescr_)
-        (PyTypeObject *, PyObject *, int, Py_intptr_t *,
-         Py_intptr_t *, void *, int, PyObject *);
+        (PyTypeObject *, PyObject *, int, Py_intptr_t const *,
+         Py_intptr_t const *, void *, int, PyObject *);
+    // Unused. Not removed because that affects ABI of the class.
     PyObject *(*PyArray_DescrNewFromType_)(int);
     int (*PyArray_CopyInto_)(PyObject *, PyObject *);
     PyObject *(*PyArray_NewCopy_)(PyObject *, int);
@@ -193,9 +194,10 @@ struct npy_api {
     PyObject *(*PyArray_FromAny_) (PyObject *, PyObject *, int, int, int, PyObject *);
     int (*PyArray_DescrConverter_) (PyObject *, PyObject **);
     bool (*PyArray_EquivTypes_) (PyObject *, PyObject *);
-    int (*PyArray_GetArrayParamsFromObject_)(PyObject *, PyObject *, char, PyObject **, int *,
-                                             Py_ssize_t *, PyObject **, PyObject *);
+    int (*PyArray_GetArrayParamsFromObject_)(PyObject *, PyObject *, unsigned char, PyObject **, int *,
+                                             Py_intptr_t *, PyObject **, PyObject *);
     PyObject *(*PyArray_Squeeze_)(PyObject *);
+    // Unused. Not removed because that affects ABI of the class.
     int (*PyArray_SetBaseObject_)(PyObject *, PyObject *);
     PyObject* (*PyArray_Resize_)(PyObject*, PyArray_Dims*, int, int);
 private:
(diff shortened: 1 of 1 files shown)


YannickJadoul:
> `PyArray_DescrNewFromType` is at offset 96

How was this never noticed? This means that `PyArray_DescrNewFromType` is never used?

> `array_t` constructor that takes a size and a pointer should take `ssize_t`.

This was changed in #782, but it seems that was a bit too enthusiast? Would you mind going over the other changes in this PR and see if more things that should not made signed now still are?

YannickJadoul:
> > `array_t` constructor that takes a size and a pointer should take `ssize_t`.
> 
> This was changed in #782, but it seems that was a bit too enthusiast? Would you mind going over the other changes in this PR and see if more things that should not made signed now still are?

There has actually been discussion on that: https://github.com/pybind/pybind11/pull/782#pullrequestreview-35659549

bstaletic:
>> `PyArray_DescrNewFromType` is at offset 96

> How was this never noticed? This means that PyArray_DescrNewFromType is never used?

It is never used. Also, the compiler has no way of warning us here because this works by grabbing a `capsule`, casting it to a `void**` and then using offsets to cast `void*` to function pointers. It completely side-steps any possibility for type safety.



> > > `array_t` constructor that takes a size and a pointer should take `ssize_t`.
> > 
> > 
> > This was changed in #782, but it seems that was a bit too enthusiast? Would you mind going over the other changes in this PR and see if more things that should not made signed now still are?
> 
> There has actually been discussion on that: [#782 (review)](https://github.com/pybind/pybind11/pull/782#pullrequestreview-35659549)

Hmm... but we're geting a cast to signed anyway. That's why the warning happens.

YannickJadoul:
> It is never used. Also, the compiler has no way of warning us here because this works by grabbing a `capsule`, casting it to a `void**` and then using offsets to cast `void*` to function pointers. It completely side-steps any possibility for type safety.

Right, I thought so. Should we maybe comment out the things that aren't used (and not tested)?

> Hmm... but we're geting a cast to signed anyway. That's why the warning happens.

Ah, wait, you're changing it tó `ssize_t`. Right, this was forgotten in the `array_t`, but I was looking at `array`. (In my defence, that's only 2 bytes difference.)

bstaletic:
> > It is never used. Also, the compiler has no way of warning us here because this works by grabbing a `capsule`, casting it to a `void**` and then using offsets to cast `void*` to function pointers. It completely side-steps any possibility for type safety.
> 
> Right, I thought so. Should we maybe comment out the things that aren't used (and not tested)?
> 

I'd rather drop it completely than comment it out. As this pull request shows, the more of numpy's multiarray C API we use, the worse. Should I check the rest and remove all that aren't used?

bstaletic:
> I'd rather drop it completely than comment it out. As this pull request shows, the more of numpy's multiarray C API we use, the worse. Should I check the rest and remove all that aren't used?

After a short discussion with @YannickJadoul we agreed not to touch `PyArray_DescrNewFromType` besides fixing the offset, because removing it would affect the ABI of `py::detail::npy_api`.

wjakob:
Oh gosh, good catch about the bad offset. All of these changes look good to me, please feel free to merge yourself when ready.

YannickJadoul:
All is green; let's go!

YannickJadoul:
Thanks, @bstaletic!

PR #2299, opened 2020-07-14
fix: external catch -> catch, fixes #2168

### LINKED ISSUE #2168: [2.5.0] catch is now called catch2
To use the system catch lib for testing, a little patch is needed.
```
diff -Nuar pybind11-2.5.0.org/tools/FindCatch.cmake pybind11-2.5.0/tools/FindCatch.cmake
--- pybind11-2.5.0.org/tools/FindCatch.cmake	2020-04-16 07:33:50.871360486 +0200
+++ pybind11-2.5.0/tools/FindCatch.cmake	2020-04-16 08:14:30.524482052 +0200
@@ -38,7 +38,7 @@
 endfunction()
 
 # Look for catch locally
-find_path(CATCH_INCLUDE_DIR NAMES catch.hpp PATH_SUFFIXES catch)
+find_path(CATCH_INCLUDE_DIR NAMES catch.hpp PATH_SUFFIXES catch2)
 if(CATCH_INCLUDE_DIR)
   _get_catch_version()
 endif()
```

tuxmaster5000:


YannickJadoul:
Is this tested? Is there any way we can see that this PR won't cause problems with older builds (and catch similar errors in the future, perhaps) ?

henryiii:
We use `DOWNLOAD_CATCH` instead of building with an external catch. I don't think there's much to test. Most users don't run the tests, and those that do, usually should use DOWNLOAD_CATCH. Testing libraries usually are built with similar settings; googletest doesn't even allow installs.

bstaletic:
We definitely shouldn't keep the broken option in our cmake files. Being able to use external, system dependencies helps distro maintainers package pybind. I don't see anything really controversial about this pull request.

GTest does allow installations. From my distro's gtest build script: https://github.com/archlinux/svntogit-community/blob/packages/gtest/trunk/PKGBUILD#L36-L41

henryiii:
GTest traditionally always deleted the install target, that may have changed.

henryiii:
And agree that having a way to target external installs is good, just saying I don't know if it's worth testing, and it's only for tests, distributers are not supposed to distribute our tests.

bstaletic:
Okay, to me your first comment sounded like you were against targeting an external Catch2. I also don't think there's anything to test here. People who depend on external Catch2 will complain if it changes name ever again.

henryiii:
Nope, wasn't against it. By the way, though, we currently download Catch 1.9.something, not Catch 2.

bstaletic:
https://github.com/pybind/pybind11/blob/master/tests/test_embed/CMakeLists.txt#7

Maybe it's time to upgrade? 1.9.3 is more than 3 years old.

henryiii:
> Maybe it's time to upgrade? 1.9.3 is more than 3 years old.

In #2324

YannickJadoul:
I just meant "tested locally", not that we need to add it to our test suite. And a bit curious about the possibility to still use the old version.

But seems to be approved, then! Thanks, @tuxmaster5000! :-)

PR #2303, opened 2020-07-16
Fix casters of STL containers with `char*`

### LINKED ISSUE #2245: passing containers of char * (e.g. std::vector<char *>) does not work
## Issue description

pybind11 does not seem to correctly pass containers of `char *` to Python. I'm using the example of  `std::array<const char *, N>`, but `std::vector`s have the same behavior.  The `char *` received by C++ seems to be pointing to a random memory location.  The memory is still within the process, as there are no segfaults (that I have seen, anyway).

## Reproducible example code

This demo class has two separate constructors.  One takes the type `std::array<const char *, 2>`, and the other takes the types `const char *, const char*`.  The `std::array` version does the wrong thing, and the `char *`, when printed, just prints out garbage.  The `char *, char *` version does the right thing, and prints the passed in string as expected.

I'm using pybind v2.5.0.  I went through the [docs on strings](https://pybind11.readthedocs.io/en/latest/advanced/cast/strings.html) but didn't notice anything.  Just guessing, but I would *guess* the problem might be shallow copies of pointers somewhere, since this only happens with containers.  But I really have no clue.

This problem does not surface whenever `std::string` is used, which is another reason I think it may be related to shallow copies of pointers (since `std::string` will copy its data when copied).

```c++
#include <vector>
#include <stdio.h>

#include <pybind11/stl.h>
namespace py = pybind11;

class Demo {
public:
    Demo(std::array<const char *, 2> strs) {
        printf("Got %s %s\n", strs[0], strs[1]);
    }
    Demo(const char * str1, const char * str2) {
        printf("Got %s %s\n", str1, str2);
    }
};

PYBIND11_MODULE(demo, m) {
    // the real deal here
    py::class_<Demo>(m, "Demo")
        .def(py::init<std::array<const char *, 2>>())
        .def(py::init<const char *, const char *>())
    ;
}
```

After building, the following Python code demonstrates the issue:

```python
import demo

# prints random garbage
demo.Demo(['string 1', 'string 2'])

# prints the right thing
demo.Demo('string1', 'string2')
```

YannickJadoul:
This cóuld fix #2245, I believe. Would you mind testing, if it does, @ codypiersall?

If it does, the question is if this is a price we're willing to pay for it?

On the one hand, the casters for `tuple`, `pair`, etc, also hold on to their subcasters. On the other hand, it might be a heavy price to "just" solve the issue with `char *`? Maybe there is a better way to keep the things `char*` point to alive?

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -73,8 +73,11 @@ template <typename Type, typename Key> struct set_caster {
             return false;
         auto s = reinterpret_borrow<pybind11::set>(src);
         value.clear();
+        subcasters.clear();
+        subcasters.reserve(s.size());
         for (auto entry : s) {
-            key_conv conv;
+            subcasters.emplace_back();
+            auto &conv = subcasters.back();
             if (!conv.load(entry, convert))
                 return false;
             value.insert(cast_op<Key &&>(std::move(conv)));
@@ -96,6 +99,9 @@ template <typename Type, typename Key> struct set_caster {
     }
 
     PYBIND11_TYPE_CASTER(type, _("Set[") + key_conv::name + _("]"));
+
+private:
+    std::vector<key_conv> subcasters;
 };
 
 template <typename Type, typename Key, typename Value> struct map_caster {
@@ -107,9 +113,13 @@ template <typename Type, typename Key, typename Value> struct map_caster {
             return false;
         auto d = reinterpret_borrow<dict>(src);
         value.clear();
+        subcasters.clear();
+        subcasters.reserve(d.size());
         for (auto it : d) {
-            key_conv kconv;
-            value_conv vconv;
+            subcasters.emplace_back();
+            auto &conv_pair = subcasters.back();
+            auto &kconv = conv_pair.first;
+            auto &vconv = conv_pair.second;
             if (!kconv.load(it.first.ptr(), convert) ||
                 !vconv.load(it.second.ptr(), convert))
                 return false;
@@ -138,6 +148,9 @@ template <typename Type, typename Key, typename Value> struct map_caster {
     }
 
     PYBIND11_TYPE_CASTER(Type, _("Dict[") + key_conv::name + _(", ") + value_conv::name + _("]"));
+
+private:
+    std::vector<std::pair<key_conv, value_conv>> subcasters;
 };
 
 template <typename Type, typename Value> struct list_caster {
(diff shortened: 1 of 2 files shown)


include/pybind11/stl.h
```diff
@@ -342,6 +369,7 @@ struct variant_caster<V<Ts...>> {
         auto caster = make_caster<U>();
         if (caster.load(src, convert)) {
             value = cast_op<U>(caster);
+            subcaster = std::move(caster);
```
YannickJadoul, line 372: Actually, are casters movable?

YannickJadoul:
An alterantive to solve #2245 would be to just disallow `type_caster<char*>` as a nested type, btw (either by extending `type_casters` to tell whether they can be used a subcasters, or by ad-hoc not allowing `char*` to be nested in these `stl.h` casters).

include/pybind11/stl.h
bstaletic: They are.

YannickJadoul:
OK, this still has a major problem, actually. For types that aren't movable or aren't empty when moved out (like `std::array`), this will take more memory (linear in the depth of the nesting), since: e.g. a `type_caster` for `std::vector<T>` now contains `std::vector<T> value;` as well as `std::vector<type_caster<T>>`, where `type_caster<T>` possibly has the same duplication again.

So `array<array<array<array<array<int>>>>>` will contain 6 times the amount of memory actually needed before. So we still need to find a better solution.

EricCousineau-TRI:
I think the following things are at play here:
- you have type conversion, so there are temporaries: https://pybind11.readthedocs.io/en/stable/advanced/cast/index.html
- those temporaries don't own their memory (the caster does in this case)
- for the duration of the function call, those casters have to stay alive
- outside of that, lifetime gets sketchy...

I don't think you can get around "all the things need to stay alive", at least for the above cases.

The only two solutions I see:
1. Expressly forbid it. Make an explicit error for `py::cast<container<char*>>(...)` that says this is a a bad idea.
2. Specialize container casters to have the parallel `caster` containers *only* when necessary. Thus, `vector<vector<...<int>...>>` isn't affected.
  - This works for loading *in a very constrained scope* (e.g. immediate function calls). Outside of that, say for `.def_readwrite`, it gets sketchy real fast.

Do you imagine other possibilities?

EricCousineau-TRI:
Er, also, is it linear or exponential? I think it may be exponential, like `O(N^depth)`?

EricCousineau-TRI:
And looking at @kenoslund-google's overview in #2527, yeah, I think that's what they're implying the specialization approach.

PR #2310, opened 2020-07-20
style: add clang-format file

henryiii:
This is an attempt at using clang-format for the C++ style. If this is accepted, then a) we can drop the old check_style.sh script (will be done in this PR), b) all open PR's are likely to need to run `pre-commit` to become compatible again before merging. Due to point b), we probably will want to wait till a specific point to merge this - so for now, this should be seen more as a discussion for the format style file we would like to adopt (if we go this way).

A few points about how I came up with the style:
* It is based on LLVM as a start
* I was trying to mimic the old style as closely as possible, rather than making ideal choices - maybe some should change.
    * Quite a few places are inconsistent, so had to pick style over the other

And, about the style itself:
* PyBind follows LLVM's `int *x` convention, vs. `int* x` - I've seen good arguments for the latter, though the former is (only) better if you are listing multiple items, which you should not do. We should verify we want to keep it
* I've left the 2-space indent for public/private/protected.

```yaml
AlignConsecutiveAssignments: true # Varies in the source, sometimes aligned, sometimes not
AlwaysBreakTemplateDeclarations: Yes # Varies in the source, but more often than not
BinPackArguments: false # Varies in the source, but this looks better, IMO
BinPackParameters: false
BreakBeforeBinaryOperators: All # Mostly/always followed
BreakConstructorInitializers: BeforeComma 
ColumnLimit: 99 # Wasn't always followed - matching flake8 here
IndentPPDirectives: AfterHash # Often was not indented the same, though - 2 spaces instead of 4
IndentWidth: 4
Language: Cpp
SpaceAfterCStyleCast: true # Varied a little bit, but this seemed to be the most common
# SpaceInEmptyBlock: true # too new, so { } will be changed to {}
Standard: Cpp11
TabWidth: 4
```

Edit: Ignore failures for now, this is for a discussion of the style, should have skipped CI.

Also, since this is a fairly large change, I think we should also move again with a check_style merged into pre-commit check so we can continue with adding GHA. We can also consider applying black at roughly the same time, assuming the format looks okay, when this PR is ready to be accepted?

### DIFF
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -104,8 +104,10 @@ PYBIND11_NAMESPACE_END(detail)
     .. code-block:: cpp
 
         {
-            py::scoped_ostream_redirect output{std::cerr, py::module_::import("sys").attr("stderr")};
-            std::cerr << "Hello, World!";
+            py::scoped_ostream_redirect output{
+                std::cerr,
+                py::module::import("sys").attr("stderr")
+            };
         }
  \endrst */
 class scoped_ostream_redirect {
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,21 @@
+---
+# See all possible options and defaults with:
+# clang-format --style=llvm --dump-config
+BasedOnStyle: LLVM
+AccessModifierOffset: -4
+AlignConsecutiveAssignments: true
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBinaryOperators: All
+BreakConstructorInitializers: BeforeColon
+ColumnLimit: 99
+IndentCaseLabels: true
+IndentPPDirectives: AfterHash
+IndentWidth: 4
+Language: Cpp
+SpaceAfterCStyleCast: true
+# SpaceInEmptyBlock: true # too new
+Standard: Cpp11
+TabWidth: 4
+...
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -176,10 +176,35 @@ name, pre-commit):
 pre-commit install
 ```
 
+### Clang-Format
+
+As of v2.6.2, pybind11 ships with a [`clang-format`][clang-format]
+configuration file at the top level of the repo (the filename is
+`.clang-format`). Currently, formatting is NOT applied automatically, but
+manually using `clang-format` for newly developed files is highly encouraged.
+To check if a file needs formatting:
+
+```bash
+clang-format -style=file --dry-run some.cpp
+```
+
+The output will show things to be fixed, if any. To actually format the file:
+
+```bash
+clang-format -style=file -i some.cpp
+```
+
+Note that the `-style-file` option searches the parent directories for the
+`.clang-format` file, i.e. the commands above can be run in any subdirectory
+of the pybind11 repo.
+
 ### Clang-Tidy
 
-To run Clang tidy, the following recipe should work. Files will be modified in
-place, so you can use git to monitor the changes.
+[`clang-tidy`][clang-tidy] performs deeper static code analyses and is
+more complex to run, compared to `clang-format`, but support for `clang-tidy`
+is built into the pybind11 CMake configuration. To run `clang-tidy`, the
+following recipe should work. Files will be modified in place, so you can
+use git to monitor the changes.
 
 ```bash
 docker run --rm -v $PWD:/pybind11 -it silkeh/clang:10
@@ -198,7 +223,7 @@ cmake -S . -B build-iwyu -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE=$(which include-what-y
 cmake --build build
 ```
 
-The report is sent to stderr; you can pip it into a file if you wish.
+The report is sent to stderr; you can pipe it into a file if you wish.
 
 ### Build recipes
 
@@ -325,6 +350,8 @@ if you really want to.
 
 
 [pre-commit]: https://pre-commit.com
+[clang-format]: https://clang.llvm.org/docs/ClangFormat.html
+[clang-tidy]: https://clang.llvm.org/extra/clang-tidy/
 [pybind11.readthedocs.org]: http://pybind11.readthedocs.org/en/latest
 [issue tracker]: https://github.com/pybind/pybind11/issues
 [gitter]: https://gitter.im/pybind/Lobby
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,12 +43,7 @@ ignore =
     docs/**
     tools/**
     include/**
-    .appveyor.yml
-    .cmake-format.yaml
-    .gitmodules
-    .pre-commit-config.yaml
-    .readthedocs.yml
-    .clang-tidy
+    .*
     pybind11/include/**
     pybind11/share/**
     CMakeLists.txt


bstaletic:
My two cents: Formatting style doesn't really matter as long as it is consistent. At that point, just pick any style and move on. I only have one strong opinion. Do not indent with 2 spaces. Since that's not the case, I'm fine with whatever.

wjakob:
Looks suprisingly good, I am impressed. Two adjustment requests:
```
BreakConstructorInitializers: BeforeColon
IndentCaseLabels: true
```
Is there some way to avoid the 2-space indentation of public/private/..? I don't think we previously did that in the codebase.

include/pybind11/functional.h
```diff
@@ -17,15 +17,16 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 
 template <typename Return, typename... Args>
 struct type_caster<std::function<Return(Args...)>> {
-    using type = std::function<Return(Args...)>;
-    using retval_type = conditional_t<std::is_same<Return, void>::value, void_type, Return>;
-    using function_type = Return (*) (Args...);
+    using type          = std::function<Return(Args...)>;
+    using retval_type   = conditional_t<std::is_same<Return, void>::value, void_type, Return>;
+    using function_type = Return (*)(Args...);
```
YannickJadoul, line 22: > ```yaml
> AlignConsecutiveAssignments: true # Varies in the source, sometimes aligned, sometimes not
> ```

Ugh, really?

I mean... there's two types of consecutive assignments: one where you are assigning bunch of similar things (like enum values, or fields of a date/time struct), and another one where you just happen to assign two things consecutively.
Somehow, it feels exaggerated or even confusing in the latter case to align them?

include/pybind11/iostream.h
```diff
@@ -95,35 +91,32 @@ PYBIND11_NAMESPACE_END(detail)
     .. code-block:: cpp
 
         {
-            py::scoped_ostream_redirect output{std::cerr, py::module::import("sys").attr("stderr")};
-            std::cerr << "Hello, World!";
+            py::scoped_ostream_redirect output{std::cerr,
+ py::module::import("sys").attr("stderr")}; std::cerr << "Hello, World!";
```
YannickJadoul, line 95: What happened here? This seems wrong.

include/pybind11/stl_bind.h
```diff
@@ -19,73 +19,87 @@ PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
 
 /* SFINAE helper class used by 'is_comparable */
-template <typename T>  struct container_traits {
-    template <typename T2> static std::true_type test_comparable(decltype(std::declval<const T2 &>() == std::declval<const T2 &>())*);
-    template <typename T2> static std::false_type test_comparable(...);
-    template <typename T2> static std::true_type test_value(typename T2::value_type *);
-    template <typename T2> static std::false_type test_value(...);
-    template <typename T2> static std::true_type test_pair(typename T2::first_type *, typename T2::second_type *);
-    template <typename T2> static std::false_type test_pair(...);
```
YannickJadoul, line 28: Are these short helpers really not an exception to the rule? Do they need to take up twice as many lines?

include/pybind11/stl_bind.h
```diff
@@ -19,73 +19,87 @@ PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
...
 
     cl.def(self == self);
     cl.def(self != self);
 
-    cl.def("count",
-        [](const Vector &v, const T &x) {
-            return std::count(v.begin(), v.end(), x);
-        },
+    cl.def(
+        "count",
```
YannickJadoul, line 95: Ugh? But but but ... "def count:" is nice if it's on one line, no? :-(

include/pybind11/functional.h
henryiii: If they are not related, you can add a line in between. These are all type settings, so I don't think it's that harmful to align them.

include/pybind11/stl_bind.h
bstaletic: You can play with "priorities" of breaks and have a fine grained control over where clang-format inserts new lines. I played with these priorities once for two hours and didn't achieve what I want, but that's a different story.

include/pybind11/functional.h
henryiii: (I'm also fine not aligned, too - no strong opinions on style, really, just want consistency and some way to automate)

henryiii:
> don't allow for any exceptions

You can turn clang-format off if you want a specific style in a specific spot; there's nothing wrong with that. It then also gives the reader a signal that you really care about the exact formatting of a block of code. However, for a good format, I think it's often not really that helpful - once you get used to a style, deviations from it are usually not helpful. While they may seem great to the person writing them at the time, they generally just serve to confuse the person reading it because it is inconsistent. Remember you cannot communicate to C++ (or Python) via style, so it is *entirely* something you are trying to communicate to the programmer on top of whatever you are actually communicating to the program.

include/pybind11/functional.h
YannickJadoul: Yeah, OK, they're related (so probably extra lines are weird), but there's nothing parallel in form in the assignments (i.e., they're not all assigning 1, 2, 3, 4, ...).

But yes, maybe that's the disadvantage to the advantage of having everything automated...

include/pybind11/stl_bind.h
henryiii: The first one is really very long - it's a bit more readable when shortened in a normal width window. What needs to be done here is to add a blank line in between each item.

YannickJadoul:
> While they may seem great to the person writing them at the time, they generally just serve to confuse the person reading it because it is inconsistent.

That might be true, yes. I'm sure I'll adapt to it as well. Just putting this out there.

> Remember you cannot communicate to C++ (or Python) via style, so it is _entirely_ something you are trying to communicate to the programmer on top of whatever you are actually communicating to the program.

Exactly! But how can you communicate anything more through formatting if everything needs to follow the fixed rules? If I see `template <...>` on the same line as `struct`, I read: ah, right, this is just a short one-line helper-thing, rather than a full class.

But yeah, that's maybe too philosophical. Everyone else seems excited to get this in, so thanks for configuring all this! :-)

include/pybind11/iostream.h
henryiii: This is inside a string, and seems to be poorly formatted in either case. I've fixed it and it's fine now.

include/pybind11/stl_bind.h
YannickJadoul: Sure, yes. Just meant as another example where a human might have thought about this, and a formatter doesn't understand the semantics. pybind11's own docs suggest the previous style: https://pybind11.readthedocs.io/en/stable/classes.html#binding-lambda-functions

henryiii:
> right, this is just a short one-line helper-thing, rather than a full class.

But shouldn't that be communicated through the code itself, the name, or a comment? What happens if you have a conceptually "short one-line helper" that happens to be really long or a few more lines?

include/pybind11/stl_bind.h
henryiii: The problem with putting them on one line is what do you do with the next line? For example,

```cpp
my_really_fantastic_and_great_object_of_some_sort.def("count",
                                                      a_lambda_or_somethign_else_1,
                                                      a_lambda_or_somethign_else_2
);
```

Gives you potentially very little space, and most editors won't get the spacing right for you when you are writing it (since it's a variable indent).

```cpp
my_really_fantastic_and_great_object_of_some_sort.def("count",
    a_lambda_or_somethign_else_1,
    a_lambda_or_somethign_else_2
);
```

makes the "count" much harder to notice/read with the other items.

```cpp
my_really_fantastic_and_great_object_of_some_sort.def(
    "count",
    a_lambda_or_somethign_else_1,
    a_lambda_or_somethign_else_2
);
```

is the only one that is consistent and readable if the arguments and/or object name are sometimes lengthy.


There are quite a few packing controls, though. I think I use the experimental auto packing flag somewhere else.

YannickJadoul:
> But shouldn't that be communicated through the code itself, the name, or a comment? What happens if you have a conceptually "short one-line helper" that happens to be really long or a few more lines?

Also, definitely, but if you can make it easier to immediately spot, just by the shape of the code... But anything's a habit, I suppose.

include/pybind11/stl_bind.h
YannickJadoul: Yeah, sure, I can see that. But here it's just `cl.def`, short enough, which might influence judgment?

At any rate, I find it a weird inconsistency with the docs.

bstaletic: For what it is worth, I also prefer not breaking before the first argument if there's enough space, like in this case.

henryiii:
I've removed the "application" commit and made this a manual hook only, so it can be applied if someone wants to, but will not be required everywhere (yet). I don't think it should be applied until we are in a fairly PR-free state - and maybe new code could be expected to pass it, etc.

rwgk:
@henryiii pointed me here, thanks!

Merging this without hook (for now) would be great, then at least we can all run with the same config on new or modified code. Doing that for a while will give us experience what works well and what may need tweaking. At some opportune moment we could go through systematically, then activate the hooks.

How will this relate do `clang-tidy` as documented in the Clang-Tidy section in `.github/CONTRIBUTING.md`?

Things I'm also wondering about: 1. how exactly do I run `clang-format` manually on a specific file?  2. guidance/recommendations for using `// NOLINT` (I guess).
Would it make sense to explain this is a small section in `.github/CONTRIBUTING.md`?

henryiii:
If this passes, I'm in favor of merging, though we should keep all whitespace change PRs separated, and list style-only changes in `.git-blame-ignore-revs`.

henryiii:
Also happy to have someone add docs, points 1 and 2 are correct above. ;)

include/pybind11/iostream.h
```diff
@@ -104,8 +104,10 @@ PYBIND11_NAMESPACE_END(detail)
     .. code-block:: cpp
 
         {
-            py::scoped_ostream_redirect output{std::cerr, py::module_::import("sys").attr("stderr")};
-            std::cerr << "Hello, World!";
```
rwgk, line 108: Is this to avoid knows the known flakyness?

rwgk:
> Also happy to have someone add docs, points 1 and 2 are correct above. ;)

I'd be happy to contribute a small section, although I'm still unclear about this question (above):

> How will this relate do `clang-tidy` as documented in the Clang-Tidy section in `.github/CONTRIBUTING.md`?

Could you please give me a rough idea? I can work on polishing the wording.

henryiii:
> How will this relate do `clang-tidy` as documented in the Clang-Tidy section in `.github/CONTRIBUTING.md`?

These are two separate checks. The docs could probably sit side by side, due to the similarity in name, but they don't have anything else in common - clang-tidy has to know how to compile the code, clang-format does not, etc.

I would put a clear disclaimer that we are not reformatting old code with this style yet. Probably will happen right after #2445 , at a guess.

include/pybind11/iostream.h
henryiii: I think it tripped on this originally and made a mess. But this is a nicer format now.

YannickJadoul: @henryiii `std::cerr << "Hello, World!";` did disappear, though

henryiii: Hmm, not sure what happened, probably a bad merge on my part. It did get updates, so I think the "old" formatting you see here now is better than when the PR started.

henryiii: Pushed a fix to master.

YannickJadoul: Great, thanks!

PR #2311, opened 2020-07-20
ci: adding pre-commit check-style.sh

henryiii:
Run check-style through pre-commit. Slight change to check-style, you now list the files you want to run on, and it doesn't do things that are already checked. Updated description, removed style check from Travis. Also included the upstream fix for the tab pre-commit hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,5 +24,14 @@ repos:
   rev: 3.8.2
   hooks:
   - id: flake8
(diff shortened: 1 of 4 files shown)


henryiii:
I'm trying to get the tabs hook improved upstream in https://github.com/Lucas-C/pre-commit-hooks/issues/14 - will make this un-drafted when that happens.

wjakob:
LGTM!

Issue #2201, opened 2020-05-05
CMake 3.12+ Python Find Routines, IPO/LTO


ax3l:
CMake 3.12.0+ deprecates the old find modules for Python and unifies them into a new call find module:
- https://cmake.org/cmake/help/v3.17/module/FindPythonInterp.html
- https://cmake.org/cmake/help/v3.12/module/FindPython.html

We should consider mid to long-term to update our CMake scripts accordingly, as soon as we can require CMake 3.12+ ([released July 2018](https://blog.kitware.com/cmake-3-12-0-available-for-download/)).

Furthermore, CMake 3.9.0+ introduces an abstraction for IPO/LTO, which can simplify our code base.
- https://cmake.org/cmake/help/v3.9/module/CheckIPOSupported.html

RUrlus:
This is also helpful for those using [Pyenv](https://github.com/pyenv/pyenv), the old find module doesn't detect the active python version but rather the newest one. The new version does correctly identify the active version .

henryiii:
I highly recommend supporting this (and will try to help tackle this in the near future), but even if we jump to supporting 3.12+ only tomorrow, I think we will need to support both methods for a long time. If a package using PyBind11 does anything else with CMake + Python, either another package or even just using the variables they expect to be defined, they will probably have to update to the new module as well (unless they mix better than I think they do).

Quickly thinking about it, I would have a `PYBIND_FIND_PYTHON` variable that defaults to "OFF" for now, and later (3.0?) will default to "ON", that selects between the two implementations. I think the new implementation is probably quite simple, as quite a bit of what we do is already fixed or added to CMake now. Also, we might need to make the minimum version for `PYBIND_FIND_PYTHON=ON` a little higher than 3.12, since there were lots of fixes and additions in the subsequent versions as they polished off the new modules.

Another option would be to simply respect an existing find package run - so if you do:

```cmake
find_package(Python COMPONENTS Development)
find_package(pybind11)
```

then it picks up the discovered new-style Python instead of self discovering with the old mechanism. That might be the best way, actually.

I haven't used the new module much yet, since basically whenever I need Python + CMake, it's probably for a PyBind11 module. :)

dvj:
[...]

PR #2312, opened 2020-07-21
CMake improvement: support windows python with msys & mingw

lkeegan:
Change to `FindPythonLibsNew.cmake` behaviour:
  - on windows, look for windows python lib (including when using mingw and msys)
  - if python lib is not found and we are using msys, then look for system python lib

Previous behaviour was to skip the first step when using msys & mingw, so windows python libs are not found (see example [travis job](https://travis-ci.org/github/lkeegan/pybind11-msys-tests/jobs/710303307#L1548)):
```
-- Found PythonInterp: C:/Python38/python.exe (found version "3.8.2") 
-- Found PythonLibs: python38
```
i.e. doesn't find the python library

New behaviour with this PR (see example [travis job](https://travis-ci.org/github/lkeegan/pybind11-msys-tests/jobs/710303510#L1554)):
```
-- Found PythonInterp: C:/Python38/python.exe (found version "3.8.2") 
-- Found PythonLibs: C:/Python38/libs/python38.lib
```

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -147,7 +147,7 @@ string(REGEX REPLACE "\\\\" "/" PYTHON_PREFIX "${PYTHON_PREFIX}")
 string(REGEX REPLACE "\\\\" "/" PYTHON_INCLUDE_DIR "${PYTHON_INCLUDE_DIR}")
 string(REGEX REPLACE "\\\\" "/" PYTHON_SITE_PACKAGES "${PYTHON_SITE_PACKAGES}")
 
-if(CMAKE_HOST_WIN32 AND NOT (MINGW AND DEFINED ENV{MSYSTEM}))
+if(CMAKE_HOST_WIN32)
     set(PYTHON_LIBRARY
         "${PYTHON_PREFIX}/libs/python${PYTHON_LIBRARY_SUFFIX}.lib")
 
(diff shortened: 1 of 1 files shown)


bstaletic:
So this looks reasonable, but someone with more experience with msys/mingw needs to give his blessing.

YannickJadoul:
> So this looks reasonable, but someone with more experience with msys/mingw needs to give his blessing.

That's not me, but I don't know if there is a regular contributor that has more experience?

There's talks to ditch all this `FindPythonLibsNew.cmake` in favour of a newer `FindPythonLibs` in CMake >= 3.12 (#2201), but meanwhile, I guess this could go in?

tools/FindPythonLibsNew.cmake
```diff
@@ -147,7 +147,7 @@ string(REGEX REPLACE "\\\\" "/" PYTHON_PREFIX "${PYTHON_PREFIX}")
 string(REGEX REPLACE "\\\\" "/" PYTHON_INCLUDE_DIR "${PYTHON_INCLUDE_DIR}")
 string(REGEX REPLACE "\\\\" "/" PYTHON_SITE_PACKAGES "${PYTHON_SITE_PACKAGES}")
 
-if(CMAKE_HOST_WIN32 AND NOT (MINGW AND DEFINED ENV{MSYSTEM}))
```
YannickJadoul, line 150: The `MINGW` part of the condition here disappeared, now, though, and doesn't come back later. Why is that?

henryiii:
> ditch all this

I don't think we will be able to ditch it, at least for a while, but there will (hopefully) be a much better way to do it soon. But agree that we should still get fixes like this in!

tools/FindPythonLibsNew.cmake
lkeegan: If I understand correctly, this `MINGW` condition was here so that when using windows+msys+mingw it would *only* look for msys system python, while when using windows+msys+any other compiler, (e.g. MSVC) it would *only* look for windows python (<https://github.com/pybind/pybind11/pull/2087>)

Now the logic on windows is to first look for the windows python lib, and if that fails and we are using msys, then look for the msys system python lib. So the `MINGW` condition is gone - the logic is the same regardless of which compiler is being used with msys.

YannickJadoul:
> I don't think we will be able to ditch it,

Ow, then I got slightly too excited. Even better to get this in then.

tools/FindPythonLibsNew.cmake
YannickJadoul: OK, yes. And - please bear with me; I'm really not experienced with msys - does it make sense that `MSVC` would go looking for msys system Python, then (if no Windows Python is found)?

lkeegan: That's a good point - I don't think that would make sense. Then it would be better to put the `MINGW` condition back in.

YannickJadoul: Alright, great, thanks! Final thing: which of the two should get precedence when using mingw, the Windows Python or the msys Python?

lkeegan: I think this will be determined by which python binary comes first in the users `PATH` (or whatever they set explicitly with `-DPYTHON_EXECUTABLE=...`)

YannickJadoul: OK, then, then I'm in! :-)

YannickJadoul:
> Me too. :)

OK, and it's green, so let's merge it then? :-)

Thanks, @lkeegan!

lkeegan:
thanks!

PR #2314, opened 2020-07-22
test_builtin_casters.py test_string_view: Python 2 c++17, c++2a compa…

rwgk:
…tibility.

Tested with 2.7.18rc1, built with Py_UNICODE_SIZE 4.
Change also tested with Python 3.8.


rwgk:
Thanks a lot Yannick for pointing out the other test using the same approach, I missed that. Since there is a precedence for what I'm doing I'll merge this PR now, to get a baseline for my work on PR #2256.

Issue #1205, opened 2017-12-04
Documentation outdated for numpy example


chhenning:
On the [page](http://pybind11.readthedocs.io/en/master/advanced/pycpp/numpy.html) I believe the example for initializing a Matrix with a numpy array is using the outdated "__init__" notation. Look for the "/* Bind MatrixXd (or some other Eigen type) to Python */" comment to find the code.

When I run the code I'm greeted with the following warning:

    FutureWarning: pybind11-bound class 'chh.Matrix' is using an old-style placement-new '__init__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.

YannickJadoul:
Thanks for reporting, @chhenning!

PR #2316, opened 2020-07-23
Use new style __init__ in numpy docs

bstaletic:
Fixes https://github.com/pybind/pybind11/issues/1205


YannickJadoul:
Just some docs. Merging. Thanks, Boris! :-)

PR #2317, opened 2020-07-23
 Avoid attr("__repr__") in initialize_generic

### LINKED ISSUE #2028: Cannot use a python class as a default argument
## Issue description

Using any `py::object` as a default argument seems to be supported. But in the particular case when the object is a class, it fails at import type.

## Reproducible example code

This works fine:
```
m.def(
    "test1",
    [](py::object a) { py::print(py::str(a)); },
    "a"_a = py::module::import("decimal").attr("__version__")
);

$ python
>>> import my_module;
>>> my_module.test1();
1.70
```
But this fails (only difference is `Decimal` in place of `__version__`)
```
m.def(
    "test2",
    [](py::object a) { py::print(py::str(a)); },
    "a"_a = py::module::import("decimal").attr("Decimal")
);

$ python
>>> import my_module;
ImportError: TypeError: descriptor '__repr__' of 'decimal.Decimal' object needs an argument
```
I would expect to simply recieve the `Decimal` class in `a`, but it seems like pybind11 is trying to construct a `Decimal` object, or something like that.

bstaletic:
If the default argument value is a class, and not an instance of a
class, `a.value.attr("__repr__")` raises a `ValueError`. Switching to
`repr(a.value)` makes this use case work.

Fixes #2028

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -228,7 +228,7 @@ protected:
             if (a.descr)
                 a.descr = strdup(a.descr);
             else if (a.value)
-                a.descr = strdup(a.value.attr("__repr__")().cast<std::string>().c_str());
+                a.descr = strdup(repr(a.value).cast<std::string>().c_str());
         }
 
         rec->is_constructor = !strcmp(rec->name, "__init__") || !strcmp(rec->name, "__setstate__");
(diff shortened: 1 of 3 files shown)


bstaletic:
Appveyor forgot how to use git and didn't clone the repository in one build. It's just a flake.

bstaletic:
Green CI!

YannickJadoul:
Green, green, green. Everything green! Thanks, Boris! :-)

PR #2320, opened 2020-07-23
Allow unions in is_accessible_base_of

### LINKED ISSUE #1283: detail::is_accessible_base_of seems to fail on union types
Not sure if this is a bug or intended behavior: it looks like detail::is_accessible_base_of seems give false negative for `union` types. Here the minimal example to reproduce and an error.

Thanks,

```C++
union A
{
	int i;
	float f;
};
PYBIND11_MODULE(example, m) {
	A a, b;
	a = b;  // work fine!

	pybind11::class_< A > (m, "A")
        .def(pybind11::init<>() )
		.def("assign", (A & (A::*)(const A &)) &A::operator=); // trigger static assert
		;
}
```

Error:
```
In file included from union.cpp:1:
/home/benchmark/src/pybind11/include/pybind11/pybind11.h:1000:5: error: static_assert failed "Cannot bind an inaccessible base class method; use a lambda definition instead"
    static_assert(detail::is_accessible_base_of<Class, Derived>::value,
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/benchmark/src/pybind11/include/pybind11/pybind11.h:1085:25: note: in instantiation of function template specialization 'pybind11::method_adaptor<A, A &, A, const A &>' requested here
                        ^
union.cpp:16:4: note: in instantiation of function template specialization 'pybind11::class_<A>::def<A &(A::*)(const A &)>' requested here
                .def("assign", (A & (A::*)(const A &)) &A::operator=); // trigger static assert
                 ^
1 error generated.
```

bstaletic:
Fixes #1283 

Questions:

1. Do we want to support raw unions?
2. Is the placement of parens making things more or less clear, in this case?
3. How should I test this?


YannickJadoul:
>   1. Do we want to support raw unions?

Question for @wjakob: do we?

Quoting [myself and @bstaletic](https://github.com/pybind/pybind11/issues/1283#issuecomment-663017640):

>> We do support binding `std::variant` and variant-like types. I guess it's not too far fetched to support unions?
>
> Myeah. Just ... I don't know. Raw unions (without being embedded in a class and some safe-access methods) are very non-Pythonic? So just curious if they were ever meant to be supported.

>   2. Is the placement of parens making things more or less clear, in this case?

To me it doesn't matter. Fine to add them. Even better to add a comment on why `is_base_of` doesn't cover `is_same`, I'd say.

>   3. How should I test this?

How do you mean? Can't we just bind a union in the tests and see if it compiles and works?

bstaletic:
> Even better to add a comment on why `is_base_of` doesn't cover `is_same`, I'd say.

Done.

> Can't we just bind a union in the tests and see if it compiles and works?

We can, but what should I actually assert?

YannickJadoul:
> Done.

:+1: Perfect

> We can, but what should I actually assert?

Ah, yes. For this PR the actual test is that is just compiles, but if we support unions, you could assert just the basic functionality, I guess? Something like this?

```py
union.int_field = 42
assert union.int_field == 42
union.float_field = 3.14159
assert union.float_field = 3.14159
```

The more I think about it, the more I think unions are "too dangerous" for Python and that if someone really needs this, the union can/should just be wrapped in some class that turns things type-safe, though. :-/

bstaletic:
> you could assert just the basic functionality, I guess? Something like this?

Also done.

wjakob (commented):
One issue with unions (and a use case for which they were often used in the past) is as a means of conversion, e.g. between floats and their binary representations e.g. via your IntFloat class. This is all undefined behavior these days, and an explicit `memcpy` must be used.

That said, as long as it is clear that pybind11-bound enums should never be used for such conversion, I don't see a major issue in shipping this change. the change to `is_accessible_base_of` by @bstaletic is really harmless.

Before giving the thumbs-up to merge, my request would be to explain this feature and the caveats with a few lines of text in the docs.

bstaletic:
> One issue with unions (and a use case for which they were often used in the past) is as a means of conversion, e.g. between floats and their binary representations e.g. via your IntFloat class. This is all undefined behavior these days, and an explicit memcpy must be used.

Sure, but I am not doing something like that in this PR. I'm explicitly writing to a union's data member before reading it.

> Before giving the thumbs-up to merge, my request would be to explain this feature and the caveats with a few lines of text in the docs.

Great. I'm on it.

bstaletic:
>> Before giving the thumbs-up to merge, my request would be to explain this feature and the caveats with a few lines of text in the docs.

> Great. I'm on it.

Done.

PR #2324, opened 2020-07-24
ci: Add Clang testing and C++11/17

henryiii:
Adding Clang version testing. Also adds C++11 and C++17 on the normal tests.

* C++17 + Clang further fixes; it will error out in user code so added it where it will help users
* Lots of warnings fixed; worked my way back to Clang 3.6. I think only std::function::target used in a test is missing for 3.5; 3.4 & 3.3 looked bad.
* Updates Catch to a newer version


tests/test_embed/CMakeLists.txt
```diff
@@ -43,3 +43,13 @@ endforeach()
...
 
 add_dependencies(check cpptest)
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
+  if(CMAKE_CXX_STANDARD VERSION_LESS 17)
+    target_compile_options(test_embed PUBLIC -Wno-deprecated-register)
+    target_compile_options(external_module PUBLIC -Wno-deprecated-register)
+  else()
+    target_compile_options(test_embed PUBLIC -Wno-register)
+    target_compile_options(external_module PUBLIC -Wno-register)
```
bstaletic, line 53: What about just adding both flags, without checking the standard version?

tests/test_cmake_build/subdirectory_target/CMakeLists.txt
```diff
@@ -13,3 +13,11 @@ set_target_properties(test_cmake_build PROPERTIES PREFIX "${PYTHON_MODULE_PREFIX
 
 add_custom_target(check ${CMAKE_COMMAND} -E env PYTHONPATH=$<TARGET_FILE_DIR:test_cmake_build>
                   ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/../test.py ${PROJECT_NAME})
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
+  if(CMAKE_CXX_STANDARD VERSION_LESS 17)
+    target_compile_options(test_cmake_build PUBLIC -Wno-deprecated-register)
```
bstaletic, line 19: Again, you can just use `LESS` above. Also, maybe skip the check and add both flags? Not sure about the latter.

Actually, if we know both of these are going to be numbers, can we use normal math symbols, like `>=`?

tests/CMakeLists.txt
```diff
@@ -152,10 +152,10 @@ function(pybind11_enable_warnings target_name)
     endif()
   endif()
 
-  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
-    if(CMAKE_CXX_STANDARD STREQUAL "14")
+  if(CMAKE_CXX_STANDARD AND CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
```
bstaletic, line 155: What happens if `CMAKE_CXX_STANDARD` is not defined?

tools/pybind11Tools.cmake
```diff
@@ -224,6 +224,14 @@ function(pybind11_add_module target_name)
     endif()
   endif()
 
+  # Python 2 doesn't really support C++17, Clang warns/errors over it
+  if(CMAKE_CXX_STANDARD
+     AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"
+     AND PYTHON_VERSION VERSION_LESS 3.0
+     AND NOT CMAKE_CXX_STANDARD VERSION_LESS 17)
```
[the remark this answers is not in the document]
henryiii: I tend to avoid `VERSION_*_EQUAL` since they were added in CMake 3.7 (and I haven't broken old CMake support quite yet). But, you are right, this is just a number, so `GREATER_EQUAL` would be just fine.

tests/test_embed/CMakeLists.txt
henryiii: I believe you can get a warning/error for an unrecognized flag.

tests/CMakeLists.txt
henryiii: Then it is empty and this does not get added. I'm not being helpful here if `CMAKE_CXX_STANDARD` is not used. In the near future, the C++ handling will be improved and this likely will become a generator expression.

tests/test_call_policies.cpp
```diff
@@ -46,6 +46,7 @@ TEST_SUBMODULE(call_policies, m) {
     class Parent {
     public:
         Parent() { py::print("Allocating parent."); }
+        Parent(const Parent& parent) = default;
```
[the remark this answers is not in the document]
henryiii: It only cared about having these two added, rather than all five. Since it's only in the tests, I wasn't worried about following the rule of five or checking to see which ones were okay to be deleted instead of defaulted.

include/pybind11/numpy.h
```diff
@@ -550,7 +550,7 @@ class array : public buffer {
         forcecast = detail::npy_api::NPY_ARRAY_FORCECAST_
     };
 
-    array() : array({{0}}, static_cast<const double *>(nullptr)) {}
+    array() : array(0, static_cast<const double *>(nullptr)) {}
```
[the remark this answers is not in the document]
henryiii: One version of Clang didn't allow `{{0}}`, and anther version didn't allow `{0}`. The new version triggers:

```cpp
template <typename T>
explicit array(ssize_t count, const T *ptr, handle base = handle()) : array({count}, {}, ptr, base) { }
```

which has the same effect as the constructor we were trying to directly trigger (it calls it).

tests/test_class.cpp
```diff
@@ -227,6 +227,8 @@ TEST_SUBMODULE(class_, m) {
         static void *operator new(size_t s, void *ptr) { py::print("C placement-new", s); return ptr; }
         static void operator delete(void *p, size_t s) { py::print("C delete", s); return ::operator delete(p); }
         virtual ~AliasedHasOpNewDelSize() = default;
+        AliasedHasOpNewDelSize() = default;
+        AliasedHasOpNewDelSize(const AliasedHasOpNewDelSize&) = default;
```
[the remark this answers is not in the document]
henryiii: Because the depreciated behavior was to add it, so I added it. I might have been able to delete it.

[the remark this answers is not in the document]
henryiii: I think we can usually delete the copy constructors, but we require the default constructors for several of these (if they are structs, they get a default constructor automatically but show a deprecation warning).

tests/test_call_policies.cpp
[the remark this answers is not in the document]
bstaletic: We used to have an explicit default constructor and destructor. The copy constructor (and copy assignment operator) was (were) implicitly defined, which raised a deprecation warning. Conclusion:

- These classes have never had the move constructor or the move assignment operator.
- To restore the old implicit behaviour, we should default the copy assignment operator as well.

Our options, as I see it:

- Match the old implicit behaviour and default the copy assignment operator too.
- Obey the rule of 5 and default the move constructor and move assignment.
- Just ignore this, since these are only tests and not user facing code.

Feel free to take any action you like.

tests/test_class.cpp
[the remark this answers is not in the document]
bstaletic: The implicit default constructor doesn't exist as soon as you mention any other constructor. That's why you had to default it explicitly. Which is fine.

tests/test_embed/CMakeLists.txt
bstaletic: Right. Clang warns about that, but gcc doesn't. Hah, we can add `-Wno-register`, `-Wno-deprecated-register` *and* `-Wno-unknown-warning-option`. Just kidding, of course.

tools/pybind11Tools.cmake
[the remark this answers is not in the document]
bstaletic: To be honest, all my rambling about special members can be ignored, but I think that simpler cmake is important. Cmake is horrible to read as it is.

henryiii:
Dropped `VERSION_`, deleted copy constructors instead of defaulted them.

tools/pybind11Tools.cmake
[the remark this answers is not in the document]
henryiii: Fully agree about CMake, it should be as clean as possible. I'll be revamping it soon to drop 3.6 and below, which will help.

> Cmake is horrible to read as it is

CMake 2.8 is ugly. Modern CMake is not. If we dumped all compatibility with older versions and forced everyone to use the latests tools like `find_package(Python)`, we could have a fairly beautiful CMakeLists. But we can't do that - but maybe we can still improve things a lot.

tests/test_cmake_build/subdirectory_target/CMakeLists.txt
henryiii: No, math symbols are not used in CMake.

tests/test_gil_scoped.cpp
```diff
@@ -14,6 +14,8 @@
 class VirtClass  {
 public:
     virtual ~VirtClass() {}
```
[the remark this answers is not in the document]
henryiii: Done, why not. :)

tests/test_class.cpp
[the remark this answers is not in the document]
henryiii: It is smart enough to complain about both, actually, when you have a destructor.

tests/test_call_policies.cpp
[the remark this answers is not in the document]
henryiii: I'm deleting the copy assignment, because we don't use it. We need an explicitly defaulted constructor because we have a destructor and we are default constructing these. We aren't using moves, and I think compilers only warn about the old rule of 3 and not the new rule of 5, so for now, I think this is fine (warning driven development is more acceptable for tests than for library code ;) ). I can only detect the cases that fail the warning, so finding all move constructors/assignments would be tricky, anyway.

tests/test_cmake_build/subdirectory_target/CMakeLists.txt
bstaletic: But the following does print `less`.

```cmake
if(2 > 3)
  message("greater")
else()
  message("less")
endif()
```

henryiii: It's not listed here: https://cmake.org/cmake/help/latest/command/if.html#command:if

henryiii: And I don't see it listed here either: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmConditionEvaluator.cxx

henryiii: And when I run the above code, I get:

```bash
$ cmake -P example.cmake
CMake Error at example.cmake:1 (if):
  if given arguments:

    "2" ">" "3"

  Unknown arguments specified
```

If I replace `>` with `LESS`, I get:


```bash
$ cmake -P example.cmake
greater
```

henryiii: (PS: I should have replaced it with GREATER, but not the point. :) )

bstaletic: > It's not listed here: https://cmake.org/cmake/help/latest/command/if.html#command:if

I read that section about 5 times, thinking I might have missed something.

> cmake -P example.cmake

I was saving the example as `CMakeLists.txt` and running it as `cmake .`. 

> If I replace `>` with `LESS`, I get:

Good enough for me. I've got nothing more to complain about. Thanks for the pull request!

henryiii: Your complaints are always welcome. :)

tests/test_embed/CMakeLists.txt
YannickJadoul: Wait, gcc doesn't know `-Wno-register`? I do build with that.

tests/test_cmake_build/installed_embed/CMakeLists.txt
```diff
@@ -8,8 +8,17 @@ message(STATUS "Found pybind11 v${pybind11_VERSION}: ${pybind11_INCLUDE_DIRS}")
...
 
 add_custom_target(check $<TARGET_FILE:test_cmake_build> ${PROJECT_SOURCE_DIR}/../test.py)
+
+if(CMAKE_CXX_STANDARD AND CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
+  if(CMAKE_CXX_STANDARD LESS 17)
+    target_compile_options(test_cmake_build PUBLIC -Wno-deprecated-register)
+  else()
+    target_compile_options(test_cmake_build PUBLIC -Wno-register)
+  endif()
+endif()
```
YannickJadoul, line 24: Are these necessary? Aren't they added by `pybind11Tools.cmake`?

henryiii: This target was added with `add_executable`, so we don't control it. The proper fix in the future will be to add these flags to the `pybind::...` targets, but that will be part of the CMake upgrade & cleanup.

YannickJadoul: Ah, yes, this is all embedding, sorry!

tests/test_embed/CMakeLists.txt
bstaletic: I was talking about warning on, literally, `-Wno-whatever`. `man gcc` does mention `-Wregister` so you're fine.

YannickJadoul: Alright! Follow-up question, then: why is there a test for `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` when this could also be gcc? Or does gcc not complain about `register`; I thought it did?

bstaletic:
```       -Wregister (C++ and Objective-C++ only)
           Warn on uses of the "register" storage class specifier, except when it is part of the GNU Explicit Register Variables extension.  The use of the
           "register" keyword as storage class specifier has been deprecated in C++11 and removed in C++17.  Enabled by default with -std=c++17.
```

So the warning will be turned on by default when a user switches on C++17. GCC 11 will default to C++17.

henryiii: ~~Does this show up though? We are compiling with GCC + C++17 with warnings as errors.~~ I bet we are not setting ` CMAKE_CXX_EXTENSIONS=OFF`, actually. I'd rather leave it as is (not any worse than before) then tackle this during the CMake overhaul.

YannickJadoul: OK!

wjakob:
Looks all good to me, thanks!

PR #2327, opened 2020-07-25
ci: cmake Configure test

henryiii:
A few basic config checks to make sure we support what we say we do. Not building, just config. (note: the 2.8 build actually doesn't run the tests properly If you do have it run, the cpp tests "pass" in 0 seconds. We will drop the 2.8 test when we drop 2.8 support (very soon)).

We may split this test later into a "build test suite" and a "use as sub package / config" test (based on the current tests in the test suite so we can have a higher requirement for the test suite than for user code. This should let us simplify the test cmake code we have to maintain, and help us support nicer optional features (since we can then use them ourselves).

### DIFF
--- a/.github/workflows/configure.yml
+++ b/.github/workflows/configure.yml
@@ -1,0 +1,81 @@
+name: Configure
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable
+      - v*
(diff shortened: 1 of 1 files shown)


henryiii:
This only adds a few tests, no other changes, and my next PR requires it, so I think it's safe to merge.

PR #2328, opened 2020-07-25
add support for std::integral_constant

jehelset:
add a stl-caster for `std::integral_constant<>`, some tests and doc.

### DIFF
--- a/docs/advanced/cast/stl.rst
+++ b/docs/advanced/cast/stl.rst
@@ -5,9 +5,10 @@ Automatic conversion
 ====================
 
 When including the additional header file :file:`pybind11/stl.h`, conversions
-between ``std::vector<>``/``std::deque<>``/``std::list<>``/``std::array<>``,
-``std::set<>``/``std::unordered_set<>``, and
-``std::map<>``/``std::unordered_map<>`` and the Python ``list``, ``set`` and
+between ``std::integral_constant<>``,
+``std::vector<>``/``std::deque<>``/``std::list<>``/``std::array<>``,
+``std::set<>``/``std::unordered_set<>``, and ``std::map<>``/``std::unordered_map<>``
+and the Python ``int`` data type, and the ``list``, ``set`` and
 ``dict`` data structures are automatically enabled. The types ``std::pair<>``
 and ``std::tuple<>`` are already supported out of the box with just the core
 :file:`pybind11/pybind11.h` header.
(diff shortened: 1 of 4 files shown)


bstaletic:
What's the use case for this? `integral_constant` is a type that only ever has a single integer value.

jehelset:
useful f.ex. as the tag-type in tag-based dispatching, see f.ex. [boost](https://www.boost.org/community/generic_programming.html), under `Tag Dispatching`. 

a type that is not out of place in a C++11 api, and therefore would be nice to see supported out of the box.

bstaletic:
That answers the question "what is `integral_constant` used for in C++". It does not answer the question "what can you do with `integral_constant` in python that you can't do with `int`". The link to boost.org does not discuss python or Boost.Python at all. That is to say, I still don't see a valid use case for including this in pybind11.

jehelset:
yes, the link is not about python. the link discusses tag-based dispatching, a use-case of `std::integral_constant`, ie, a reason for why this type can be a part of a C++-api. when exposing such a C++-api in python with pybind11, support for `std::integral_constant` would let the api be exposed in a more direct fashion: 

[gits](https://gist.github.com/jehelset/2a319f3f71aedea652dd7091cd3d4bbb)

it's not about doing something with the integral-constant in python, as you can see it is passed as plain integer-value when leaving C++ in any case, [stl.hpp](https://github.com/jehelset/pybind11/blob/std_integral_constant/include/pybind11/stl.h#L64),  it's about deferring the mapping from a python-int to a `std::integral_constant` to pybind11, and exposing the api as directly as possible with minimal binding code.

YannickJadoul:
I agree with @bstaletic. `std::integral_constant` is a C++-construction to make some template magic work, since the type *is* the value. IMO, it 1) should not be part of a public C++ API, 2) has no place in a pythonic Python API, and 3) is such a rare use case that I don't see a reason to make it part of pybind11.
Apart from that, you can just put this caster in a separate header in your project, and include/use it; but pybind11's philosophy is to try keeping the library small but extensible, so we'd really need a more compelling, Python-based argument to include it for everyone.

I'll close this. If it turns out more users could use this functionality in pybind11, we can reconsider. So if anyone reads this and needs the functionality, do please comment.

Thank you very much for doing the effort and turning it into a PR, though! :-) I'm sorry, but we do need to guard the scope what's included in pybind11 a bit.

PR #2330, opened 2020-07-26
ci: transition to GHA

henryiii:
This PR _mostly_ transitions the test system to GHA. Features:

* Appveyor reduced from 7 to 2 builds, for Visual Studio 2015
    * Testing 32-bit Visual Studio 2017 in GHA now too
* Travis builds moved entirely to GHA
    * New version of doxygen test uses Alpine, should be much more stable than the old test
* New `-DDOWNLOAD_EIGEN` option for tests, CMake 3.11+ only
    - GHA tests now test Eigen and SciPy
* Badges updated
* Added GCC builds (starting with 7, the older docker images are not supported anymore and are too old to use reasonably)
* Added CentOS builds (CentOS 7 is the only reasonable way to get GCC 4.8 AFAIK, and that's the compiler in manylinux1)
* Nicer naming scheme
* Attempt to fix caching (will speed up macOS PyPy builds, which are the slowest when Appveyor is not backlogged)

If we drop VS 2015, then we could drop Appveyor. However, that version was the default version for several Python's in the Python 3 series. This should still tremendously help the slowest bottleneck in testing.


tests/CMakeLists.txt
```diff
@@ -97,6 +97,8 @@ set(PYBIND11_CROSS_MODULE_GIL_TESTS
   test_gil_scoped.py
 )
 
+option(DOWNLOAD_EIGEN "Download EIGEN (requires CMake 3.14+)" OFF)
```
henryiii, line 100: Oops, technically 3.11.

tests/CMakeLists.txt
```diff
@@ -129,7 +155,7 @@ if(PYBIND11_TEST_FILES_EIGEN_I GREATER -1)
     message(STATUS "Building tests with Eigen v${EIGEN3_VERSION}")
   else()
     list(REMOVE_AT PYBIND11_TEST_FILES ${PYBIND11_TEST_FILES_EIGEN_I})
-    message(STATUS "Building tests WITHOUT Eigen")
+    message(STATUS "Building tests WITHOUT Eigen, use -DDOWNLOAD_EIGEN on CMake 3.11+ to download")
```
bstaletic, line 158: Should `DOWNLOAD_EIGEN` default to `ON`, but still first try to find Eigen on the system?

henryiii: I matched `DOWNLOAD_CATCH` for now.

wjakob:
This looks really great! +1 for testing a truly ancient GCC version as well :-).

Issue #1666, opened 2019-01-13
How to return a std::tuple wrapped in std::shared_ptr???


lazylazypig:
Hi, all
I'd like to bind a function that return a std::tuple wrapped in smart pointer (like std::shared_tr). following is my sample code:

## C++ Portion

#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
#include <pybind11/functional.h>
#include <tuple>
#include <string>
#include <iostream>

namespace py = pybind11;

using PlateInfo = std::tuple<std::tuple<int, int, int, int>, std::string>;

std::shared_ptr<PlateInfo> getPlate(bool isOK)
{
	if (!isOK) nullptr;
	return std::make_shared<PlateInfo>(std::make_tuple(std::make_tuple(100, 100, 200, 200), "abcdefg"));
}

PYBIND11_MODULE(example, m)
{
	m.def("getPlate", &getPlate);
}

the code above can be compile without no error. but, in the python code, it failed :(

## Python Portion

import example

plate_info = example.getPlate(True)

it is just a very simple use case, by I got following errors:

Traceback (most recent call last):
  File "test.py", line 59, in <module>
    plate_info = example.getPlate(True)
TypeError: Unable to convert function return value to a Python type! The signature was
        (arg0: bool) -> std::tuple<std::tuple<int,int,int,int>,std::basic_string<char,std::char_traits<char>,std::allocator<char> > >

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.

as you seen, I have included the needed header like <pybind11/stl.h>. have i done something wrong? and, how can i fixed it? thanks:)

haolu23:
have the same question. it is fine to just return a tuple, but not shared_ptr of tuple.

lazylazypig:
@haolu23 unfortunately, there seem to be no way for this usage. so i prefer to return std::optional/boost::optional of tuple instead. the code pieces showed bellow:

#include <pybind11/pybind11.h>
#include <pybind11/stl.h>

#ifdef USE_BOOST_OPTIONAL
#include <boost/optional.hpp>
using boost::optional;
using boost::make_optional;

namespace pybind11 {
namespace detail {
template <typename T>
struct type_caster<boost::optional<T>> : optional_caster<boost::optional<T>> {};
} // detail
[...]

PR #2334, opened 2020-07-27
cast pointer to std::tuple and std::pair

wojdyr:
This PR copies `handle cast(T *src, ...)` from PYBIND11_TYPE_CASTER to tuple_caster.
It enables bindings to functions returning a pointer to std::tuple or std::pair - in the same way as pybind11 handles now pointers to int, std::string, std::array, and everything else.

I don't have a deep understanding of pybind11 internals - please double check this patch before applying.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1421,6 +1421,17 @@ public:
         return cast_impl(std::forward<T>(src), policy, parent, indices{});
     }
 
+    // copied from the PYBIND11_TYPE_CASTER macro
+    template <typename T>
+    static handle cast(T *src, return_value_policy policy, handle parent) {
+        if (!src) return none().release();
+        if (policy == return_value_policy::take_ownership) {
+            auto h = cast(std::move(*src), policy, parent); delete src; return h;
+        } else {
+            return cast(*src, policy, parent);
+        }
+    }
+
     static constexpr auto name = _("Tuple[") + concat(make_caster<Ts>::name...) + _("]");
 
     template <typename T> using cast_op_type = type;
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -117,12 +117,16 @@ TEST_SUBMODULE(builtin_casters, m) {
         return std::make_pair(RValueCaster{}, std::make_tuple(RValueCaster{}, std::make_pair(RValueCaster{}, RValueCaster{}))); });
     m.def("lvalue_nested", []() -> const decltype(lvnested) & { return lvnested; });
 
+    static std::pair<int, std::string> int_string_pair{2, "items"};
+    m.def("int_string_pair", []() { return &int_string_pair; });
+
     // test_builtins_cast_return_none
     m.def("return_none_string", []() -> std::string * { return nullptr; });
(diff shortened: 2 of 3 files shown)


YannickJadoul:
I'm still a little bit in doubt on whether we actually want this. After all, `std::optional<std::tuple<...>>` is just as easy to write, no?

On the other hand, consistency with the other casters might not be bad...

wojdyr:
If the existing C++ code returns a pointer, converting it to std::optional in the bindings is more involved (and requires C++17) than just writing `m.def("func", &func)`

bstaletic:
> I'm still a little bit in doubt on whether we actually want this. 

On the account of "principle of least surprise" (alternatively [WTF/s](https://commadot.com/wp-content/uploads/2009/02/wtf.png)), I think we do want this. Currently these casters seem inconsistent with the rest.

YannickJadoul:
Fun plot twist. This also doesn't compile:

```c++
m.def("f", [](std::pair<int, double> *p){ std::cout << p->first << ", " << p->second << std::endl; });
```

Apart from that, I would agree that this should be more similar, yes.

YannickJadoul:
> So does it make sense to allow returning `std::pair<T, U>*`/`std::tuple<Ts...>*` when you can't take them as arguments?

Actually. Let me answer my own question:

No, it doesn't. But it's already like that for the other types as well, so there's no inconsistency between different casters.

wojdyr:
I implemented only the C++ -> Python conversion, because I don't know how to do the Python -> C++ part.
Someone smarter is needed for this.

YannickJadoul:
> I implemented only the C++ -> Python conversion, because I don't know how to do the Python -> C++ part.
> Someone smarter is needed for this.

We're struggling ourselves with that: see e.g. https://github.com/pybind/pybind11/issues/2245

Can I ask: since you want to be able to return pointers to things that will get copied anyway. How do you see accepting non-const references and pointers to `std::pair`, `std::tuple`, `std::string`, `std::vector` (under `stl.h`), ... (i.e., anything that is copied and wouldn't really be a reference/pointer to the Python object) ? Is that necessary to support?
We were just seriously thinking of ditching that, to solve https://github.com/pybind/pybind11/issues/2245.

wojdyr:
@YannickJadoul I think it's useful to get automatic bindings even in case of non-const references or pointers, but I'm not the best person to discuss it.

YannickJadoul:
> @YannickJadoul I think it's useful to get automatic bindings even in case of non-const references or pointers, but I'm not the best person to discuss it.

Thanks; any bit of input is interesting. And I see your concern.
The thing why we're considering this is because it would solve an issue and that these references and pointers are misleading (because it's a reference/pointer to a temporary, copied object).

Anyhow, thanks! :-)

wjakob:
Does the following shorter approach not work?

```
    static handle cast(T *src, return_value_policy policy, handle parent) {
        if (!src) return none().release();
    }
```

wjakob:
Ooops, scratch that previous comment. LGTM.

YannickJadoul:
Thanks, @wojdyr! :-)

YannickJadoul:
*(Quickly still cross-referencing to #1666, which was already closed)*

PR #2335, opened 2020-07-27
Check scope's __dict__ instead of using hasattr when registering classes and exceptions

### LINKED ISSUE #1624: how to overwrite parent class's enum with same name?
## Issue description

I'm binding a complex c++ library to python, in this lib, it has many inheritance, and for example
```
class A {
public:
     enum AEnum{...}
}

class B: public A {
public:
     enum AEnum{...}
}
```
in child class, there is a enum with exact same name with parent class, but it has different values. so in my binding code, I can bind these enum to each class, it compiles fine, but when in python, if I use `import testmodule`, then it throws an error about that `AEnum` is already defined.

So How could I export these different enums with same name, but in different class?

PS: I didn't use `py::enum_<xx>().export_values()`, I don't want any of these enum values in my root module.

YannickJadoul:
This allows registering the same name in a derived class scope, and fixes #1624

Disadvantage:
- The check is slower.
- Wouldn't work with custom `__getattribute__`/`__setattribute__` (maybe now it already wouldn't, either?).

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1012,7 +1012,7 @@ public:
     PYBIND11_OBJECT_DEFAULT(generic_type, object, PyType_Check)
 protected:
     void initialize(const type_record &rec) {
-        if (rec.scope && hasattr(rec.scope, rec.name))
+        if (rec.scope && hasattr(rec.scope, "__dict__") && rec.scope.attr("__dict__").contains(rec.name))
             pybind11_fail("generic_type: cannot initialize type \"" + std::string(rec.name) +
                           "\": an object with that name is already defined");
 
@@ -1930,7 +1930,7 @@ public:
         std::string full_name = scope.attr("__name__").cast<std::string>() +
                                 std::string(".") + name;
         m_ptr = PyErr_NewException(const_cast<char *>(full_name.c_str()), base.ptr(), NULL);
-        if (hasattr(scope, name))
+        if (hasattr(scope, "__dict__") && scope.attr("__dict__").contains(name))
             pybind11_fail("Error during initialization: multiple incompatible "
                           "definitions with name \"" + std::string(name) + "\"");
         scope.attr(name) = *this;
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -391,3 +391,29 @@ def test_base_and_derived_nested_scope():
     assert m.BaseWithNested.Nested != m.DerivedWithNested.Nested
     assert m.BaseWithNested.Nested.get_name() == "BaseWithNested::Nested"
     assert m.DerivedWithNested.Nested.get_name() == "DerivedWithNested::Nested"
+
+
+def test_register_duplicate_class():
+    import types
+    module_scope = types.ModuleType("module_scope")
+    with pytest.raises(RuntimeError) as exc_info:
+        m.register_duplicate_class_name(module_scope)
+    expected = ('generic_type: cannot initialize type "Duplicate": '
+                'an object with that name is already defined')
+    assert str(exc_info.value) == expected
+    with pytest.raises(RuntimeError) as exc_info:
+        m.register_duplicate_class_type(module_scope)
+    expected = 'generic_type: type "YetAnotherDuplicate" is already registered!'
+    assert str(exc_info.value) == expected
+
+    class ClassScope:
+        pass
+    with pytest.raises(RuntimeError) as exc_info:
+        m.register_duplicate_nested_class_name(ClassScope)
+    expected = ('generic_type: cannot initialize type "DuplicateNested": '
+                'an object with that name is already defined')
+    assert str(exc_info.value) == expected
+    with pytest.raises(RuntimeError) as exc_info:
+        m.register_duplicate_nested_class_type(ClassScope)
+    expected = 'generic_type: type "YetAnotherDuplicateNested" is already registered!'
+    assert str(exc_info.value) == expected
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -388,4 +388,6 @@ def test_multiple_instances_with_same_pointer(capture):
 # https://github.com/pybind/pybind11/issues/1624
 def test_base_and_derived_nested_scope():
     assert issubclass(m.DerivedWithNested, m.BaseWithNested)
-    assert m.DerivedWithNested.Nested != m.BaseWithNested.Nested
+    assert m.BaseWithNested.Nested != m.DerivedWithNested.Nested
+    assert m.BaseWithNested.Nested.get_name() == "BaseWithNested::Nested"
+    assert m.DerivedWithNested.Nested.get_name() == "DerivedWithNested::Nested"
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -383,3 +383,9 @@ def test_multiple_instances_with_same_pointer(capture):
     # No assert: if this does not trigger the error
     #   pybind11_fail("pybind11_object_dealloc(): Tried to deallocate unregistered instance!");
     # and just completes without crashing, we're good.
+
+
+# https://github.com/pybind/pybind11/issues/1624
+def test_base_and_derived_nested_scope():
+    assert issubclass(m.DerivedWithNested, m.BaseWithNested)
+    assert m.DerivedWithNested.Nested != m.BaseWithNested.Nested
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -393,6 +393,7 @@ def test_base_and_derived_nested_scope():
     assert m.DerivedWithNested.Nested.get_name() == "DerivedWithNested::Nested"
 
 
+@pytest.mark.skip("See https://github.com/pybind/pybind11/pull/2564")
 def test_register_duplicate_class():
     import types
     module_scope = types.ModuleType("module_scope")
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -440,8 +440,10 @@ TEST_SUBMODULE(class_, m) {
 
     py::class_<BaseWithNested> baseWithNested_class(m, "BaseWithNested");
     py::class_<DerivedWithNested, BaseWithNested> derivedWithNested_class(m, "DerivedWithNested");
-    py::class_<BaseWithNested::Nested>(baseWithNested_class, "Nested");
-    py::class_<DerivedWithNested::Nested>(derivedWithNested_class, "Nested");
+    py::class_<BaseWithNested::Nested>(baseWithNested_class, "Nested")
+        .def_static("get_name", []() { return "BaseWithNested::Nested"; });
+    py::class_<DerivedWithNested::Nested>(derivedWithNested_class, "Nested")
+        .def_static("get_name", []() { return "DerivedWithNested::Nested"; });
 }
 
 template <int N> class BreaksBase { public:
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -406,6 +406,7 @@ TEST_SUBMODULE(class_, m) {
     struct IsNonFinalFinal {};
     py::class_<IsNonFinalFinal>(m, "IsNonFinalFinal", py::is_final());
 
+    // test_exception_rvalue_abort
     struct PyPrintDestructor {
         PyPrintDestructor() = default;
         ~PyPrintDestructor() {
@@ -417,6 +418,7 @@ TEST_SUBMODULE(class_, m) {
         .def(py::init<>())
         .def("throw_something", &PyPrintDestructor::throw_something);
 
+    // test_multiple_instances_with_same_pointer
     struct SamePointer {};
     static SamePointer samePointer;
     py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer")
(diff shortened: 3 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -917,7 +917,7 @@ class generic_type : public object {
     PYBIND11_OBJECT_DEFAULT(generic_type, object, PyType_Check)
 protected:
     void initialize(const type_record &rec) {
-        if (rec.scope && hasattr(rec.scope, rec.name))
+        if (rec.scope && hasattr(rec.scope, "__dict__") && rec.scope.attr("__dict__").contains(rec.name))
```
YannickJadoul, line 920: Potentially, we could drop `hasattr(rec.scope, "__dict__")`, if we're happy throwing an exception when the user tries to use a scope that doesn't have a `__dict__` attribute?

bstaletic:
> Wouldn't work with custom __getattribute__/__setattribute__ (maybe now it already wouldn't, either?).

It used to work. `PyObject_GetAttrString(o, "attr")` calls

- [`tp_getattr`](https://github.com/python/cpython/blob/d583738a87c3019dcfe06ed4a0002d1d6c9e9762/Objects/object.c#L792)
- Then defers to [`PyObject_GetAttr(o1, unicode_string_object)`](https://github.com/python/cpython/blob/d583738a87c3019dcfe06ed4a0002d1d6c9e9762/Objects/object.c#L796)
- Which ends up checking [`tp_getattro`](https://github.com/python/cpython/blob/d583738a87c3019dcfe06ed4a0002d1d6c9e9762/Objects/object.c#L891)
- If `tp_getattro` is null, it checks [`tp_getattr`](https://github.com/python/cpython/blob/d583738a87c3019dcfe06ed4a0002d1d6c9e9762/Objects/object.c#L896) again, for good measure.

This is so horribly expensive. By the way, there's a detail I omitted. `PyObject_HasAttrString(o, "str")`. Because it just defers to `PyObject_GetAttrString(o, "str")` and if that throws just calls `PyErr_Clear()` to swallow it.

Oh, by the way, `tp_getattr` is deprecated.

include/pybind11/pybind11.h
rwgk: The idea of throwing an exception in this situation worries me, a lot.

I'm surprised to see the strong runtime performance concerns from both Yannick and Boris. Is it correct that this code only runs when a pybind11 module is imported? Is that so slow that people complain already? What's the incremental extra runtime overhead? My first guess is it's no measurable, given all the other things happening at import time (files are opened, dlls loaded), but maybe I'm overlooking something?

bstaletic: > I'm surprised to see the strong runtime performance concerns from both Yannick and Boris.

It's not really a concern. It is only run once at import time, so it hardly matter.

> What's the incremental extra runtime overhead?

I wasn't measuring `import whatever`, but rather this specific patch in isolation.

```
>>> timeit.timeit(stmt='B().__dict__.get("X")', setup='class A:\n  def X(self):pass\nclass B(A):pass')
0.17064448199380422
>>> timeit.timeit(stmt='hasattr(B(), "X")', setup='class A:\n  def X(self):pass\nclass B(A):pass')
0.12798299700079951
```

Again, this is executed exactly once. It really shouldn't be a concern. I just wanted to satisfy my own curiosity.

YannickJadoul: Not too worried about it, either! Just want to point it out.
The reasoning is/was, if `hasattr(rec.scope, "__dict__")` fails, most likely the later `scope.attr(name) = *this;` will also fail/throw an exception?

tests/test_class.py
```diff
@@ -323,3 +323,8 @@ def test_non_final_final():
         class PyNonFinalFinalChild(m.IsNonFinalFinal):
             pass
     assert str(exc_info.value).endswith("is not an acceptable base type")
+
+
+def test_base_and_derived_nested_scope():
+    assert issubclass(m.DerivedWithNested, m.BaseWithNested)
+    assert m.DerivedWithNested.Nested != m.BaseWithNested.Nested
```
rwgk, line 330: I consider this a very weak assert, because there could be bugs or even features making this unequal by chance.
Is there a simple way to make this more conclusive? E.g. modeling #1624 more closely? Or adding a method `int foo() { return 1; }` for base nested and `return 2` for the derived nested?

YannickJadoul: Methods should not be affected by this, I believe? I've never seen or had any complaints where you can't override methods in a derived class.

The main test is actually that the module imports, because in #1624, pybind11 refuses to override the `Nested` class in the base class. So if the import succeeds and the two `Nested` classes are different, we don't regress on this PR, I believe. (The test fails without the fixes.)

> I consider this a very weak assert, because there could be bugs or even features making this unequal by chance.

Should I `assert m.DerivedWithNested.Nested is not m.BaseWithNested.Nested`? I guess that would result in the same check on the two nested classes being distinct, but that can't be overridden by `__eq__` in the metaclass.

YannickJadoul: I guess I could still add some static property to both classes, though.

rwgk: Up to you, if you feel this is sufficiently strong I'm fine, but in my experience trying to be stingy in test code will bite you sooner or later. The basic idea: find a reasonably simple way to be absolutely sure some other bug or misunderstanding cannot fool us here. Often that means you're not purely testing what you actually want to test, but I see that as a safety cushion.

YannickJadoul: I see what you mean.
In this case, I'm still not convinced it will catch something more, but it can't really hurt either, so ... done :-)

I still want to get a look at @bstaletic's links as well, to see if accessing `__dict__` is really the correct solution.

rwgk: That looks like a great safety cushion to me: just 4 more lines, difficult to image that we could still get fooled.

YannickJadoul: Right, thanks for the review! :-)

include/pybind11/pybind11.h
rwgk: I think I misunderstood this code initially (github is helpfully hiding the next line...). Would this be logically equivalent?
```
if (rec.scope && hasattr(rec.scope, rec.name) && hasattr(rec.scope, "__dict__") && rec.scope.attr("__dict__").contains(rec.name))
```
Now suspecting: the existing `hasattr` returns true, but we want to raise an error only if we find that the attribute is defined in `__dict__`?

YannickJadoul: Hmmm, that's an interesting one! Yes, it might be better, this way.

You'll have figured out by now, but this is the failing condition, btw. This is when pybind11 throws an error.
I also figured out the reason it throws an error: this is the only place the created class/enum type gets stored. In the internals, only a (non-owning) `PyObject*` gets stored. If a pybind11 class would get overwritten and get garbage collected, `detail::internals`'s `type_map<type_info *> registered_types_cpp; // std::type_index -> pybind11's type information` map a C++ type to a Python type that doesn't exist anymore

YannickJadoul: Thinking out loud (it's getting late on this side of the ocean, so I might be talking nonsense):

Would `if (rec.scope && hasattr(rec.scope, rec.name) && (!hasattr(rec.scope, "__dict__") || rec.scope.attr("__dict__").contains(rec.name)))` even be better?

I.e.:
- if we have no `scope`: weird, but OK, nevermind, we can't override anything
- if `hasattr` is false: fine, not problem, we can `setattr` the new class
- if `hasattr` is true, and we don't have a `__dict__`: aaaargh, something weird is going on in this type; and the `getattr` (and probably `setattr` as well) mechanism has been overwritten; so, custom case: go ahead, but you are probably specialized enough to find a way around it
- if `hasattr` is true, and we do have a `__dict__`: check if the dict contains it, if not, we probably found it in a parent class's scope

YannickJadoul: (Writing that out: yeah, maybe the mix of boolean expressions could be rewritten more clearly, indeed)

rwgk: > we probably found it in a parent class's scope

The "probably" worries me a little.
This is beginning to look tricky.
Something totally crazy, with a vague hope that'll lead us somewhere good:

existing_obj = getattr(scope, name, None)
existing_refcount = getrefcount(existing_obj)
new_obj = setattr(scope, name, nested_obj)
if existing_obj is not None and getrefcount(existing_obj) < existing_refcount:
  raise

not good: makes assumptions about the refcounting system.
but the idea is to figure out if we actually squash an existing object (and create dangling internal pointers IIUC).

YannickJadoul: > > we probably found it in a parent class's scope
> 
> The "probably" worries me a little.

Or somewhere else, but yes, it worries me as well.
Then again, for sane scope types, this is OK, I would say. pybind11 has a lot of assumptions where "By default and in sane behavior, this is OK; if you do crazy stuff, you're hopefully wise enough to know what dangers there are."

The refcounting seems dangerous, indeed. It should work, but it feels fishy. And PyPy might not find it as nice, either (since it's basically a CPython implementation detail).

YannickJadoul: An other option is to make things overwriteable by keeping the classes somewhere else alive (hard to avoid leaks?) ? (Or, maybe even worse - because registered classes could "suddenly" become not convertible anymore - remove the `PyObject*` references from the internals when a `class_`-constructed Python type gets gc'd?)

rwgk: I totally agree with you. I was wondering about "pybind11 has a lot of assumptions", you just explained it, thanks! I agree with that, too. Your four-bullet logic makes sense AFAICT. If it's compatible with the general pybind11 philosophy, I think all we need is good hard tests (maybe complete already?) and good comments in the sources. Idea: move the logic into a separate function, so we can add comprehensive comments without cluttering the bigger context where the code lives right now?

rwgk: > An other option is to make things overwriteable by keeping the classes somewhere else alive (hard to avoid leaks?) ? (Or, maybe even worse - because registered classes could "suddenly" become not convertible anymore - remove the `PyObject*` references from the internals when a `class_`-constructed Python type gets gc'd?)

(side remark: I submitted my previous comment before I saw this.)

I'd venture to a more sprawling change like this only after deciding for sure your current local solution is infeasible.

YannickJadoul: OK, I'll see if I can clean this up!

> I totally agree with you. I was wondering about "pybind11 has a lot of assumptions", you just explained it, thanks! I agree with that, too.

One last argument against that/worry: before, we were too strict on not wanting to overwrite an attribute (we wouldn't allow it even if it should be possible; only if `__getaatr__` and `__setattry__` are completely out of sync, we did something that could result in a crash). Now, we'd be slightly too lenient: we do allow it in some weird case where there's a dict but `__setattr__` doesn't overwrite in the dict (which could  get some pybind11 type instance garbage collected and cause a segfault).

bstaletic: > - if `hasattr` is true, and we don't have a `__dict__`: aaaargh, something weird is going on in this type; and the getattr (and probably setattr as well) mechanism has been overwritten;

This is the case that worries me. Would this PR break this case?

YannickJadoul: Potentially, yes. Because the previous `py::class_` would be replaced, could (maybe) get garbage collected, and pybind11 would hold a bad `PyObject*` pointer to the old class.

YannickJadoul: Not sure how to solve this, though. Checking `hasattr`/`getattr` is too wide (i.e., checks base classes), while `__dict__` is too strict (would miss the crazy case where `__getattr__`/`__setattr__`, or `__getattribute__`/`__setattribute__`, would be overwritten).

bstaletic: Maybe we shouldn't fix it? If it's this hard to do it right, perhaps we shouldn't "fix" it and break things that currently work.

YannickJadoul: Interesting suggestion, yes. I must say the current situation feels wrong, though, where it matters in which order you register a class first in the base class or first in a derived class.

And the above shouldn't break too much. there's a thing where, if `getattr` and `setattr` are overwritten, things might already be broken anyway. In any case, the `scope` should be a class or a module, no?

bstaletic: > In any case, the `scope` should be a class or a module, no?

Do you want the good news or the bad news? The bad news?

```
>>> def f():print('f')
...
>>> def g():print('g')
...
>>> f.g = g
>>> del g
>>> f.g()
g
>>> f()
f
```

The good news is that using functions as scopes isn't allowed with pybind. I tried. I really really tried to make pybind accept a function object as a scope.

YannickJadoul: I'm indeed not sure if this is good or bad news :-/

henryiii:
Needs rebasing.

YannickJadoul:
Rebased. I do still want to give this solution another look, btw, now that we're a few months later and potentially wiser.

rwgk (approved):

wjakob:
This change looks fine to me -- runtime cost of pybind11 module initialization is usually not much of a concern (it's very fast.), so the extra check is negligible. One question though: you added some tests that ensure that the new kinds of scoped class declarations work, but are we ever checking that one cannot define the same class twice within the same module? IIRC no, in which case it would be good to add an extra test for that.

YannickJadoul:
OK, if everyone's fine with not supporting weird metaclasses overriding `__getattribute__` and `__setattribute__` and bypassing `__dict__`, I'm fine. 

On potential improvement (mentioned above) would be this:

> Would `if (rec.scope && hasattr(rec.scope, rec.name) && (!hasattr(rec.scope, "__dict__") || rec.scope.attr("__dict__").contains(rec.name)))` even be better?
> 
> I.e.:
> 
>    * if we have no `scope`: weird, but OK, nevermind, we can't override anything
> 
>    * if `hasattr` is false: fine, not problem, we can `setattr` the new class
> 
>    * if `hasattr` is true, and we don't have a `__dict__`: aaaargh, something weird is going on in this type; and the `getattr` (and probably `setattr` as well) mechanism has been overwritten; so, custom case: go ahead, but you are probably specialized enough to find a way around it
> 
>    * if `hasattr` is true, and we do have a `__dict__`: check if the dict contains it, if not, we probably found it in a parent class's scope

But this is too complex?

I'll still quickly add a test! :-)

YannickJadoul:
> One question though: you added some tests that ensure that the new kinds of scoped class declarations work, but are we ever checking that one cannot define the same class twice within the same module? IIRC no, in which case it would be good to add an extra test for that.

I searched for the error message, couldn't find it in the tests, and added `test_register_duplicate_class`.

henryiii (approved):

YannickJadoul:
OK, this is finally green, and I'm fed up with this. I'm merging this.

I still have another idea, but that'll be for a separate issue/next release :-)

YannickJadoul:
Thanks everyone for all the reviews and help! :-)

PR #2338, opened 2020-07-29
feat: CMake update, part 1

### LINKED ISSUE #1097: CMake interaction with C/C++ std flags

### LINKED ISSUE #1710: pybind11 incorrectly adds "-fvisibility" flag to nvcc
Since `nvcc` is a compiler driver and not a compiler in itself, it doesn't accept flags like `-fvisibility` directly, but passes them on via phase-options like `-Xcompiler=`. If you pass the flag directly, `nvcc` will return with an error (`nvcc fatal   : Unknown option 'fvisibility'`).

When you set up a typical pybind11 project that uses a mix of C++ and CUDA with 

```
cmake_minimum_required(VERSION 3.9)
project(py_cuda LANGUAGES CXX CUDA)
add_subdirectory(pybind11)
pybind11_add_module(...)
```

and the module consists of C++ (cpp/h) and CUDA files (cu/cuh), `nvcc` will return the error mentioned above during the build.

I know that pybind11 doesn't officially support `nvcc`, but the fix for this issue probably very small. The culprit for the behaviour is the following line: https://github.com/pybind/pybind11/blob/master/CMakeLists.txt#L108

`target_compile_options`  doesn't discriminate on the language being compiled or the compiler in use, which is probably why you placed it inside `NOT MSVC` condition. Interestingly `pybind11_add_module` already implements this correctly using set_target_properties and the respective target-property: https://github.com/pybind/pybind11/blob/master/tools/pybind11Tools.cmake#L158

Maybe those approaches could be merged? I would simply replace the line in the `CMakeLists.txt` with

```
set_target_properties(module PROPERTIES CXX_VISIBILITY_PRESET "hidden")
set_target_properties(module PROPERTIES CUDA_VISIBILITY_PRESET "hidden")
```

Am I missing something?

This happens in the following environment:
cmake version 3.10.2
The CXX compiler identification is GNU 7.3.0
The CUDA compiler identification is NVIDIA 9.1.85

But i think it will happen in all Linux environments and maybe even on MacOS.

### LINKED ISSUE #1733: Missing file pybind11Targets.cmake?
I have what looks like a successful installation of pybind11, but building another package on top of it causes a message that pybind11Targets.cmake is missing. Which it is.

There are one or two more tickets about this, but they kinda seem to have "gone away of their own accord". 

My installation is nothing special:

```
cmake \ 
    -D CMAKE_INSTALL_PREFIX=${PYBIND11_INSTALL} \ 
    -D PYTHON_EXECUTABLE=`which python3` \ 
    -D PYBIND11_INSTALL=ON \ 
    -D PYBIND11_MASTER_PROJECT=ON \
    -D PYBIND11_TEST=OFF \ 
   ${PYBIND11_SRC}```

(Some of those options are pure voodoo programming....)

### LINKED ISSUE #1903: link time optimization disabled due to OLD behaviour of CMake policy 0069
Pybind11 requires CMake of version 2.8.12 or newer. Because this is lower than 3.9, it sets the OLD behaviour of the policy 0069.

https://cmake.org/cmake/help/latest/policy/CMP0069.html#policy:CMP0069

With the OLD behaviour, LTO is not applied and this is warning displayed during the build:

```
-- Configuring done
CMake Warning (dev) at [...]/pybind11/tools/pybind11Tools.cmake:131 (add_library):
  Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
  enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  INTERPROCEDURAL_OPTIMIZATION property will be ignored for target [...]
```

Pybind11 explicitly states in the documentation that it needs LTO to make good binary code so I think this is important.

I think this would be a good solution:

```
if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.9)
  cmake_policy(SET CMP0069 NEW)
endif()
```

### LINKED ISSUE #1991: CXX flags passed to nvcc using CMake make compilation fail
## Issue description

Compilation fails when using CMake and pybind11_add_module for project containing CUDA code. The -flto flag is passed to NVCC and causes the following output:
`nvcc fatal   : Unknown option 'flto'`

## Reproducible example code
In CMakeLists.txt:

cmake_minimum_required(VERSION 3.14)
project(projectname LANGUAGES CXX CUDA)
find_package(pybind11 REQUIRED)

pybind11_add_module(modulename src/cpp_source.cpp src/cuda_source.cu)

## Possible solution

In `pybind11/tools/pybind11Tools.cmake`, line [104](https://github.com/pybind/pybind11/blob/bd24155b8bf798f9f7022b39acd6d92e52d642d6/tools/pybind11Tools.cmake#L104):
-use `$<$<COMPILE_LANGUAGE:CXX>:` to set the flag only for GCC
-pass it to nvcc as `-Xcompiler=-flto`

### LINKED ISSUE #2018: Debug build does not set `-DPy_DEBUG` with cmake
### TLDR
I think the cmake variable `PYTHON_IS_DEBUG` is not visible inside the scope of the function `pybind11_add_module`.

### What

Trying to create a debug build, using the debug version of Python with `cmake`:

- `-DCMAKE_BUILD_TYPE=Debug`
- `-DPYTHON_EXECUTABLE=/usr/lib/python3.6-dgb`

Build goes fine, but importing the resulting module gives an `Import Error` about the undefined symbol `PyModule_Create2`.

From what I have read, this indicates that we are linking correctly against the debug version of Python, but the includes are wrong, since `PyModule_Create2` has a different name in the debug version of Python.

### Solution?

- https://github.com/pybind/pybind11/blob/37352491225358b97ce302273bf2d887a477efb0/tools/pybind11Tools.cmake#L142-L147 is not working as it should.
- I have verified that `PYTHON_IS_DEBUG` is set correctly in https://github.com/pybind/pybind11/blob/master/tools/FindPythonLibsNew.cmake#L115, and it is also correct in `pybind11Tools.cmake` (outside the function scope)...
- ... but **not when the function is called**.

I am having trouble understanding `cmake`'s scoping rules. If I set `PYTHON_IS_DEBUG` in the cmake CACHE it works. I cannot tell whether this is the correct solution though?

### Versions

- cmake 3.10.2
- pybind11 2.4.3
- python3.6-dbg
- Ubuntu

### LINKED ISSUE #2089: Pybind fails to build with MSVC 19.16 in C++17 mode

### PR DESCRIPTION
**Pull request by henryiii**
This is a complete CMake update, part 1 of approximately 2. The key themes are better use of CMake built-in features, better consistency between subdirectory and config mode, and a more deterministic config file. This will enable new features in subsequent PRs.

## CMake 3.7 or better is now required

This enables many simplifications, and a more unified approach to targets. Many if statements dropped, though a few more added for enhanced support for later CMake versions.

## Reworked C++ support to properly handle C++ version in CMake instead manually

CMake 2.8, which came out three years before 2011, does not have C++11 support (hopefully not a surprise, though given the number of people that seem to want to use a new compiler with a CMake version that significantly predates it, I'm not sure). Now it does, so PyBind11 plays nice with C++ support! FINALLY! Key points:

* `PYBIND11_CPP_STANDARD` is deprecated, and issues a warning. It _does_ still work, though, just grabbing the last two digits and setting `CMAKE_CXX_STANDARD` with it.
* If a compiler has a reasonable default (like many do now), we get to just use that, instead of injecting a flag.
* `CMAKE_CXX_STANDARD` is not required - CMake knows that pybind requires at least C++11 support, and will automatically add at least the C++11 flag if needed to any target that links to it (via compile features) if the compiler has a default less than C++11.
* If a target requires a higher level, that will be used _instead_ (not duplicated, like before), either through compile features (per target) or via `CMAKE_CXX_STANDARD` (global) or CXX_STANDARD (per target, not inherited)
* Because it's built into CMake, it correctly supports all compilers, and correctly mixes languages (that is, it is escaped correctly for CUDA!)
* We get to remove the old caveats about setting the C++ standard before PyBind11 is used - it can be placed anywhere now.

Closes #1097

## If-less config files
The config file generated will not depend on the _host_ system - this is a requirement for the CMake config files to be distributed in a pure python wheel - which can't know what the host compiler / python version is beforehand. Generator expressions are used, which are evaluated during the build, rather than before.

Note: there are two ifs that affect the configs: If you generate the file with CMake 3.13+, it cannot be read by 3.7-3.12. However, this is _much_ better than before, were you couldn't change much of anything after generating a Config file, and crucially, the 3.13+ one has a lower potential for bugs (it enforces that the dynamic lookup does not get de-duplicated accidentally). The main planned user of this feature is PyPI package, and there should be no problem getting CMake 3.13+ using pip.

Key first step for #2266!

PS: Normal if's *in* the Config.in are fine, just not if's making the targets themselves in the main file.

## SYSTEM changes

The current handling of SYSTEM is not consistent with CMake standards. CMake treats all IMPORTED targets as SYSTEM, so this was/is always on when using configuration mode. This now treats Python always as SYSTEM (may fix the warning we struggled with for a while), and treats PyBind11 as SYSTEM when being used as a subdirectory, consistent with the config mode.  The SYSTEM argument to `pybind11_add_module` is no longer needed and is inconsistent with `target_link_libraries`. A warning is printed if given. You can always set `NO_SYSTEM_FROM_IMPORTED` if you want to make imported targets non-system.

## Smaller features

* The library prints the CMake version when used as a master project. It also no longer unconditionally prints the PyBind11 version when included as a submodule.
* CMake polices are set to the newest version tested, reducing warnings and hack-arounds
    - I'm supporting the built-in, modified CMake 3.11 that shipped in some versions of MSVC instead of using the native syntax for supporting a version range.
    - Closes #1903
* The version number is now a standard CMake VERSION - this has the downside that it must be made up of valid numbers
    - Adding a new dev flag
    - Available as `pybind11_VERSION*` and `PROJECT_VERSION*(`, picked up correctly by CMake wherever it's needed (for example, by CPack)
* The standard `BUILD_TESTING` flag is now supported, along with `PYBIND11_TEST`
    - Both have distinct uses unless you remove building tests when this is not the master project, which is not a bad idea IMO
    - I didn't use `include(CTest)` here, which sets `BUILD_TESTING` up for you, but I could.
* The visibility=hidden flag attempt has been removed from the `pybind11::module` target; it caused issues (closes #1710, combined with flto change closes #1991), it went against the description (simplest possible flags), it clashes with the correct way to do this in the `pybind11_add_module` function, and is better handled by a nice example (added to the docs).
* flto now uses CMake instead of custom flag checking for CMake 3.9+, closes #1991, partially #2201 - added `PYPBIND11_CLASSIC_LTO` to force the old search method and manual flags
* Python 2.7 warning suppression added to targets (I could be talked out of that one, but it nicely fixes the issues we were having at the target level - the new SYSTEM handling helps, but likely still breaks in C++17 mode without this fix)
* Moved the version macros up to the top of the detail/config file to make them easier to find and update
* Don't override additional Python versions if variable already set, closes #2089
* Building against a debug Python less than 3.8 should work now, closes #2018
* Targets are always installed when everything else is installed, closes #1733
* Spacing changed to 4-space indent, 2-space keyword that I've seen used a bit and like, but maybe 2-space is more standard? I usually use 4 space everywhere, but the CMake docs does use 2-space (just checked)

This has gotten large enough, so I'll be working on new FindPython support in the next PR. Also, we currently might be a bit sloppy about putting build products in the source tree - I'll investigate that in the near future, too. This is mostly focused on correctness, which should make alternate compilers, CUDA, and such easier to support (since we are relying on CMake to support them, rather than trying to hardcode everything ourselves), but we do get 5 or so issues closed with this (not counting 1-2 of mine). This also should enable CMake configs in the python wheel, which when mixed with PEP 518 builds and scikit-build, could make something truly beautiful. :)

As a further test, I dropped this into boost-histogram (CMake mode) and awkward1 (both submodule systems) and verified that no changes were needed, it is backward compatible and still builds correctly locally.

----

Upgrade guide:

Several details of the CMake support have been deprecated; warnings will be shown if you need to change something. The changes are:

* `PYBIND11_CPP_STANDARD=<platform-flag>` is deprecated, please use `CMAKE_CXX_STANDARD=<number>` instead, or any other valid CMake CXX or CUDA standard selection method, like `target_compile_features`.
* If you do not request a standard, PyBind11 targets will compile with the compiler default, but not less than C++11, instead of forcing C++14.
* Direct `pybind11::module` usage should always be accompanied by at least `set(CMAKE_CXX_VISIBILITY_PRESET hidden)` or similar - it used to try to manually force this compiler flag (but not correctly on all compilers or with CUDA).
* `pybind11_add_module`'s `SYSTEM` argument is deprecated - this now behaves like other imported libraries in both config and submodule mode, and behaves like a `SYSTEM` library by default.
* CMake 3.9+ will use CMake's built-in `INTERPROCEDURAL_OPTIMIZATION` instead of a custom flag discovery algorithm - set `PYPBIND11_CLASSIC_LTO` to force the old system on all CMake versions. `THIN_LTO` has no effect when the new system is being used, it will select the `thin_lto` as applicable.

### CODE REVIEW

File: docs/compiling.rst [hunk 1]
```diff
@@ -201,6 +194,17 @@ to an independently constructed (through ``add_library``, not
...
 
+    If you want to add these in yourself, you can use:
+
+    .. code-block:: cmake
+
+        cmake_minimum_required(3.9)
+        set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+
+    or set teh corisponding property (without the ``CMAKE_``) on the targets
```
henryiii, line 205:
```suggestion
    or set the corresponding property (without the ``CMAKE_``) on the targets
```

docs/compiling.rst
```diff
@@ -201,6 +194,17 @@ to an independently constructed (through ``add_library``, not
     Studio (``/bigobj``).  The :ref:`FAQ <faq:symhidden>` contains an
     explanation on why these are needed.
 
+    If you want to add these in yourself, you can use:
+
+    .. code-block:: cmake
+
+        cmake_minimum_required(3.9)
+        set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
henryiii, line 202: What about `VISIBILITY_INLINES_HIDDEN`, should we put that in here too?

docs/compiling.rst
```diff
@@ -100,25 +96,22 @@ regular LTO if ``-flto=thin`` is not available.
...
+By default, pybind11 will compile modules with the compiler default or the
+minimum standard required by PyBind11, whichever is higher.  You can set the
+standard explicitly with
+`CMAKE_CXX_STANDARD <https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html>`_:
 
 .. code-block:: cmake
 
     # Use just one of these:
-    set(CMAKE_CXX_STANDARD 11)
     set(CMAKE_CXX_STANDARD 14)
```
henryiii, line 107: What about doing the standard formula here instead:

```cmake

set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)

```

I would expect users would be able to replace the 14 easily, and this is what you see in most CMake (non-library) projects.

YannickJadoul: Would be OK with me (potentially, you put `set(CMAKE_CXX_STANDARD 14)  # Change to 11, 14, 17 as wanted`

Do we need the last line, specifically for pybind11? If not, you might put a comment next to it that it's optional, and what it does?

YannickJadoul:
>     * `PYBIND11_CPP_STANDARD` is deprecated, and issues a warning. It _does_ still work, though, just grabbing the last two digits and setting `CMAKE_CXX_STANDARD` with it.

Beware of `PYBIND11_CPP_STANDARD=/std:c++latest`, though. This has been the way to get C++17 before C++17 was fully implemented in MSVC, I believe. Now it should give you sóme C++20 features, I believe.

Were there now talks to preserve some legacy support for old CMake? I'd happily drop that, actually. CMake 3.7 is +3.5 years old (https://cmake.org/pipermail/cmake/2016-November/064540.html), so we can't keep up support forever.

docs/compiling.rst
```diff
@@ -128,7 +121,7 @@ For example:
 
     cmake -DPYBIND11_PYTHON_VERSION=3.6 ..
     # or
-    cmake -DPYTHON_EXECUTABLE=path/to/python ..
+    cmake -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") ..
```
YannickJadoul, line 124: Maybe keep the old example as well?

docs/compiling.rst
YannickJadoul: Do we need to add symbol hiding ourselves? I would somehow like if this kept being the default, as it is now. I think lots of users might have no idea about it, but would produce huge extension modules/wheels. Other, running into problems with visibility, probably are doing more intricate stuff, so should be able to do research and disable "hidden" visibility, I'd say.

include/pybind11/detail/common.h
```diff
@@ -9,6 +9,11 @@
 
 #pragma once
 
+#define PYBIND11_VERSION_MAJOR 2
+#define PYBIND11_VERSION_MINOR 6
+#define PYBIND11_VERSION_PATCH 0
+#define PYBIND11_VERSION_TYPE "dev"
```
YannickJadoul, line 15: I guess it makes sense that we're now working on 2.6.0, but shouldn't we leave this to @wjakob, to do all the release and version number acrobatics?

tests/CMakeLists.txt
```diff
@@ -5,16 +5,26 @@
...
+if(${CMAKE_VERSION} VERSION_LESS 3.18)
+    cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
+else()
+    cmake_policy(VERSION 3.18)
+endif()
+
+# There's no harm in including a project in a project
+project(pybind11_tests CXX)
 
 option(PYBIND11_WERROR  "Report all warnings as errors"  OFF)
```
YannickJadoul, line 21: Wouldn't this be a property of the target pybind1 creates, as well?

tests/test_cmake_build/CMakeLists.txt
```diff
@@ -1,26 +1,22 @@
...
-  # 3.1 needed for cmake -E env for testing
-  return()
-endif()
 
-include(CMakeParseArguments)
 function(pybind11_add_build_test name)
   cmake_parse_arguments(ARG "INSTALL" "" "" ${ARGN})
 
-  set(build_options "-DCMAKE_PREFIX_PATH=${PROJECT_BINARY_DIR}/mock_install"
+  set(build_options "-DCMAKE_PREFIX_PATH=${pybind11_BINARY_DIR}/mock_install"
```
YannickJadoul, line 7: This should never be executed with `${PROJECT_BINARY_DIR}` != `${pybind11_BINARY_DIR}`, imo, but yeah, it's just safer, I guess?

tests/test_embed/CMakeLists.txt
```diff
@@ -43,13 +37,3 @@ endforeach()
...
-
-if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND PYTHON_VERSION VERSION_LESS 3.0)
-  if(CMAKE_CXX_STANDARD LESS 17)
-    target_compile_options(test_embed PUBLIC -Wno-deprecated-register)
-    target_compile_options(external_module PUBLIC -Wno-deprecated-register)
-  else()
-    target_compile_options(test_embed PUBLIC -Wno-register)
-    target_compile_options(external_module PUBLIC -Wno-register)
-  endif()
-endif()
```
YannickJadoul, line 55: So nice, to see duplicated code getting removed!

henryiii:
> Were there now talks to preserve some legacy support for old CMake? I'd happily drop that, actually.

@wjakob made this the minimum for now, based on Debian oldstable (see <https://cliutils.gitlab.io/modern-cmake/chapters/intro/installing.html#full-list> ) .

include/pybind11/detail/common.h
henryiii: I had to replace `PYBIND11_VERSION_PATCH dev1` with something. Until it's released, this shouldn't matter too much. I could have done 2.5.1, or 2.5.0 dev, it didn't matter, I just needed something with an integer in PATCH.

tests/test_cmake_build/CMakeLists.txt
henryiii: The tests are now a subproject (as is supposed to be done in if you can run the subdirectory (tests) separately, which you can), so this is no longer true - PROJECT is the tests project. So this needs to be explicit.

tests/CMakeLists.txt
henryiii: You can run this separately, and subdirectories can and often should be projects too.

YannickJadoul:
> Personally, I think it's easier to tell users to update CMake than to maintain old code and the bugs in older versions which cause issues (I see quite a few in our issue tracker that would have been fixed by updates to CMake), and your CMake should always be newer than your compiler, and CMake is not required if you just want to use the library, you can pip install it, it's extremely backward compatible, etc... But I do understand wanting to keep older CMake support, and I support 3.4+ currently with CLI11, so I guess I'm not one to talk. ;)

No, I completely agree. If it's been 3.5 years, we should be able to drop some of the burden!
Just curious, because I had interpreter some discussions as "let's keep some fallback", but I should probably read better ;-)

henryiii:
On: docs/compiling.rst

Before, we hacked in the compile flags, and it showed - we were breaking on mixed language builds and compilers where we didn't know the correct flags to add. The point of `pybind11_add_module` is to add theses sorts of extra flags, and there's no easy way to remove them; even if they are breaking the build. There is a simple CMake function to add this, so we should use that, rather than trying to hack it in ourselves.

If CMake added an `INTERFACE_CXX_VISIBILITY_PRESET` and `INTERFACE_CUDA_VISIBILITY_PRESET`, this could be done via a target, but until then, we should probably not try to inject manual flags here ourselves, and use the CMake feature instead. (That might be something to ask for in CMake, by the way).

tests/test_cmake_build/CMakeLists.txt
YannickJadoul: Ah, right, thanks!

YannickJadoul:
> On: docs/compiling.rst

Sorry, because of GitHub's clipping of the changelog, I completely missed that this is for the `pybind11::module` (I guess what you link against for embedding?), and not for `pybind11_add_module`. This is fine, as this is arguably advanced use. Sorry for the confusion!

henryiii:
`pybind11::embed` is for embedding. `pybind11::module` is you want to build things yourself, and not rely on a magic function. It is meant to be for advanced users, and it even says clearly in the docs it does not add extra flags - it was just trying to add this one anyway before and now it doesn't.

docs/compiling.rst
henryiii: Summary from below (not that GitHub lets me comment on this again): the was only for the "advanced" target (regular does proper visibility hiding still), and there are strong downsides to trying to set visibility through an interface target (unless CMake adds first class support for it), and it damages the "good" usage of it via the `pybind11_add_module` function.

I should note  here that 3.9+ is only needed for the line below, about the LTO; CMake 3.7 handles the VISIBILITY just fine.

docs/compiling.rst
henryiii: A: Added

B: No, this is just a strictness setting - CMake will do something like `-std=g++14` without it; projects still build correctly with the compiler specific extensions. I don't think it needs an explicit comment, since you often see these together like this - but I would have added one, I just didn't remember this till after I pushed.

henryiii: And the STANDARD_REQUIRED one is optional too, it just is allowed to "decay" if your compiler doesn't support it, when  many users prefer a config error.

henryiii:
I addressed the `PYBIND11_CPP_STANDARD=/std:c++latest` with an explicit error now, telling a user to update to `CMAKE_CXX_STANDARD`. It will only issue a warning and continue if it detects a valid C++ version.

henryiii:
I found a nice CMake format tool, this is more like the old format, and pre-commit checked too! :)

docs/compiling.rst
henryiii: Added notes.

YannickJadoul: Thanks! :-)

henryiii:
Added more protections for the Py2.7 flag, and added a flag for windows C++17 + Py 2.7. I didn't protect the Win flag with CXX; if you use it with CUDA, this might break it (and to enable Windows 2.7 support only, that would be a shame). I don't use CUDA + MSVC, though, so not sure. Fixed a small typo in master that was not making warnings errors.

.cmake-format.yaml
```diff
@@ -0,0 +1,69 @@
+parse:
```
wjakob, line 1: First time I heard about .cmake-format, that's neat! One related question: we're beginning to "clog up" the root directory with many dot-files related to CI infrastructure. Can we move them to '.github' as well?

CMakeLists.txt
```diff
@@ -2,156 +2,243 @@
...
 
-if (POLICY CMP0048)
-  # cmake warns if loaded from a min-3.0-required parent dir, so silence the warning:
-  cmake_policy(SET CMP0048 NEW)
-endif()
-
-# CMake versions < 3.4.0 do not support try_compile/pthread checks without C as active language.
-if(CMAKE_VERSION VERSION_LESS 3.4.0)
-  project(pybind11)
+# VERSION 3.7...3.18, but some versions of VS have a patched CMake 3.11 that do not work properly
```
wjakob, line 10: This is neat way to deal with all of the policy warnings, I'll remember that for my next CMake project! It wasn't clear to me when reading these lines what the VS-specific issue is, and how those lines address it. Can you clarify?

CMakeLists.txt
```diff
@@ -2,156 +2,243 @@
...
   set(PYBIND11_MASTER_PROJECT ON)
+  message(STATUS "pybind11 v${pybind11_VERSION} ${PYBIND11_VERSION_TYPE}")
+  message(STATUS "CMake ${CMAKE_VERSION}")
+
+  if(CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  endif()
+else()
+  set(PYBIND11_MASTER_PROJECT OFF)
```
wjakob, line 49: Can we have the
```
message(STATUS "pybind11 v${pybind11_VERSION} ${PYBIND11_VERSION_TYPE}")
```
even when pybind11 is not the master project? (I find that a quite useful reminder when working on larger projects that an older pybind11 is being used somewhere).

CMakeLists.txt
```diff
@@ -2,156 +2,243 @@
...
+              cxx_range_for
+              cxx_right_angle_brackets
+              cxx_strong_enums
+              cxx_user_literals
+              cxx_variadic_templates)
+else()
+  # This was added in CMake 3.8, but we are keeping a consistent breaking point for the config file
+  # at 3.13. A config generated by CMake 3.13+ can ony be read in 3.13+ due to the SHELL usage
+  # later, so this is safe to do.
+  target_compile_features(pybind11 INTERFACE cxx_std_11)
```
wjakob, line 160: Regarding this line: how will this work when the user wants to use newer versions of the standard in their library? Does CMake somehow take the max(..) of the `INTERFACE` value?

CMakeLists.txt
```diff
@@ -2,156 +2,243 @@
...
+  pybind11 ${pybind11_system} INTERFACE $<BUILD_INTERFACE:${PYBIND11_INCLUDE_DIR}>
+                                        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+# Only add Python for build - must be added during the import for config since it has to be
+# re-discovered.
+target_include_directories(pybind11 SYSTEM INTERFACE $<BUILD_INTERFACE:${PYTHON_INCLUDE_DIRS}>)
+
+if(CMAKE_VERSION VERSION_LESS 3.13)
+  target_compile_features(
+    pybind11
+    INTERFACE cxx_auto_type
```
wjakob, line 146: What's the reason for providing this long list? I imagine that any of them, like `cxx_variadic_templates` will enforce a C++11 compilation flag to be passed, no?

CMakeLists.txt
```diff
@@ -2,156 +2,243 @@
...
+else()
+  # SHELL (3.12+) forces this to remain together, and link_options was added in 3.13+
+  target_link_options(module INTERFACE "$<$<PLATFORM_ID:Darwin>:SHELL:-undefined dynamic_lookup>")
 endif()
 
-if (PYBIND11_INSTALL)
+message(STATUS "CXX Compiler version: ${CMAKE_CXX_COMPILER_VERSION}")
+
+# Workaround for Python 2.7 and C++17 (C++14 as a warning) incompatibility
+set(clang_4plus
```
wjakob, line 185: Generator expressions are not the most readable representation, and I had a hard time figuring the end result here. Can you provide a human-readable sumary that goes beyond the `# Workaround ...`?

docs/compiling.rst
```diff
@@ -100,25 +96,22 @@ regular LTO if ``-flto=thin`` is not available.
...
 -----------------------
 
-By default, pybind11 will compile modules with the C++14 standard, if available
-on the target compiler, falling back to C++11 if C++14 support is not
-available.  Note, however, that this default is subject to change: future
-pybind11 releases are expected to migrate to newer C++ standards as they become
-available.  To override this, the standard flag can be given explicitly in
-`CMAKE_CXX_STANDARD <https://cmake.org/cmake/help/v3.17/variable/CMAKE_CXX_STANDARD.html>`_:
+By default, pybind11 will compile modules with the compiler default or the
+minimum standard required by PyBind11, whichever is higher.  You can set the
```
wjakob, line 100: `PyBind11` -> `pybind11`

docs/compiling.rst
```diff
@@ -127,8 +120,12 @@ For example:
...
 
     cmake -DPYBIND11_PYTHON_VERSION=3.6 ..
-    # or
-    cmake -DPYTHON_EXECUTABLE=path/to/python ..
+
+    # Another method:
+    cmake -DPYTHON_EXECUTABLE=/path/to/python ..
+
+    # You will often see this idiom:
+    cmake -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") ..
```
wjakob, line 128: "Idiom" strikes me as something mainly used for spoken language.

include/pybind11/detail/common.h
wjakob: Fine with me. But just to clarify: we can no longer put `dev` into the patch definition here? That will maybe need some clarifications for `docs/release.rst`.

wjakob: One potential issue as well: many of the places that reference `PYBIND11_VERSION_MAJOR` etc. to generate a combined version string now do not concatenate that with `PYBIND11_VERSION_TYPE`.

tools/FindPythonLibsNew.cmake
```diff
@@ -1,84 +1,74 @@
-# - Find python libraries
-# This module finds the libraries corresponding to the Python interpreter
-# FindPythonInterp provides.
-# This code sets the following variables:
+# * Find python libraries This module finds the libraries corresponding to the Python interpreter
```
wjakob, line 1: just to clarify: only whitespace changes in this file?

.cmake-format.yaml
henryiii: Putting them in `.github` is probably a bad idea, as the files in there have special meanings for GitHub - like `.github/CONTRIBUTING.md`. issue templates, pull request templates, actions, workflows, etc. But this one could go anywhere, such as in `tools`. The problem is that if you try to run `cmake-format` from the command line, it will pick this up only if it's in the main folder. However, I would expect this and all the other formatters to only be run by `pre-commit`, so that's not a big deal. The `.pre-commit-config.yaml` file needs to be in the main folder. I rather think the `.clang-format` file will need to be in the main folder (IIRC, `-style=file` is the parameter you run, which doesn't include a path).

These are all just dot files, so they only make the main GitHub page longer (which is still a little irritating, but a common issue, especially for Python).

henryiii: (My general rule is that I always use the default name/location if something expects it, so tools "just work" if someone tries to run them; but if there's not special meaning to a name/path, I put it in a folder to keep the page clean)

CMakeLists.txt
henryiii: I've written about it here: https://cliutils.gitlab.io/modern-cmake/chapters/basics.html (I realize I need to update the text, I talk about the "upcoming release" of 3.12). The [bug report and resolution is here](https://developercommunity.visualstudio.com/content/problem/294070/cmake-312-version-requirements-definition-fails-fo.html). Do we only support the latest release of Windows compilers (IE, is it okay if an earlier release has an issue in the same family, like 15.3 vs. 15.4)? If that's the case, we can simplify this slightly. I have no idea how widespread this still is.

TL;DR version:

In CMake 3.12, they added the `VERSION 3.7...3.18` syntax - from CMake 3.12 on, this automatically sets the policies to the highest possible value, up to the maximum listed. By careful choice of syntax, it happens to be completely valid syntax (though you don't get the nice policy bump) in CMake < 3.12 on all official CMake versions! The problem is MSVC had a built-in fork of CMake < 3.12 that didn't support this.

CMakeLists.txt
henryiii: Generally, it's not a good idea to have messages in a subdirectory, there's no way to stop this if you don't want it (for example, if you use FeatureSummary at the end instead), and it's a difference between find_package - but it is helpful, related directly to PyBind11, and just one line - we can put it back outside the if and then add a `PYBIND11_QUIET` if someone really wants to be able to turn it off (not me).

CMakeLists.txt
henryiii: `target_compile_features` says "this needs at least the features of C++11" to compile". If your compiler doesn't support the feature, you get a nice, configure time error (more applicable to the non-meta feature method). If you have a target with multiple features, it selects the flag needed to get all of them (so yes, basically the max). If you have a `CXX_STANDARD` set, then that is used for the flag, or (again) a nice configure time error is given you try to set a standard too low.

In CMake 3.11+, you can set COMPILE_OPTIONS on individual files!

tools/FindPythonLibsNew.cmake
```diff
@@ -1,84 +1,74 @@
...
-#=============================================================================
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+# FITNESS FOR # A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+# WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# =============================================================================
```
YannickJadoul, line 40: Same here.

Is it somehow possible to stop the formatter from doing this?

CMakeLists.txt
henryiii: That's a mix of being lazy/correct and these lists: (gcc example) https://gitlab.kitware.com/cmake/cmake/-/blob/v3.8.0/Modules/Compiler/GNU-CXX-FeatureTests.cmake - features were "ranked" based on when a compiler started supporting them. For the simplest list, I look through these lists, pick the one closest to the top that we require, then do that for several different compilers, and remove duplicates. Something like `cxx_auto_type` will be too low on that list - we won't get a nice configure time error if this is run on GCC 4.7 or earlier, and it may trigger a -std=c++0x flag instead.

People got sick of listing all the features, and post C++11 people really started thinking in terms of standards instead of compliers, so they added the `cxx_std_11` meta-features in 3.8, and `cuda_std_11` features in 3.17.

The lists are here: https://gitlab.kitware.com/cmake/cmake/-/tree/v3.18.0/Modules/Compiler (the ones that say CXX-FeatureTest, GCC, AppleClang, Clang, and ICC would probably be close enough)- I'm not sure exactly which C++11 features we use, though I know we don't require GCC 4.9. I can try to make a shorter list. Do we require any of the features introduced in gcc 4.8.1, for example?

CMakeLists.txt
henryiii: Sure.

docs/compiling.rst
henryiii: I randomly use pybind11, Pybind11, and PyBind11 - just like I randomly use numpy, Numpy, and NumPy... I'll grep. :)

tools/FindPythonLibsNew.cmake
```diff
@@ -1,84 +1,74 @@
...
-#  PYTHON_MODULE_EXTENSION    - lib extension, e.g. '.so' or '.pyd'
-#  PYTHON_MODULE_PREFIX       - lib name prefix: usually an empty string
-#  PYTHON_SITE_PACKAGES       - path to installation site-packages
-#  PYTHON_IS_DEBUG            - whether the Python interpreter is a debug build
+# PYTHONLIBS_FOUND           - have the Python libs been found PYTHON_PREFIX - path to the Python
+# installation PYTHON_LIBRARIES           - path to the python library PYTHON_INCLUDE_DIRS        -
+# path to where Python.h is found PYTHON_MODULE_EXTENSION    - lib extension, e.g. '.so' or '.pyd'
+# PYTHON_MODULE_PREFIX       - lib name prefix: usually an empty string PYTHON_SITE_PACKAGES - path
+# to installation site-packages PYTHON_IS_DEBUG            - whether the Python interpreter is a
+# debug build
```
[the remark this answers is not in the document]
henryiii: Yep! It tries to reformat moments by default. Nice for wrapping, bad for structures it doesn't recognize. I'll tone it back a bit here. I tried to check for this, but missed this one.

tools/FindPythonLibsNew.cmake
henryiii: Yes, there's an option to get it to leave comments alone, but it tries to "fix" comments by default. It does recognize list structures and a few things like that.

tools/FindPythonLibsNew.cmake
henryiii: Yes, I didn't touch this.

include/pybind11/detail/common.h
henryiii: Or `dev<number>`, sorry. I'm used to [PEP 440](https://www.python.org/dev/peps/pep-0440/) numbering, but this seems to already be just a "dev" attached to a normal PATCH number, which is easy to parse in CMake.

YannickJadoul: It would actually be nice to notice in a CMake build when you're not building with an officially released version, though? Or not?

henryiii: I've thought about that. But you don't get the benefit of an automatic "dev" tag if you are not using CMake, then.

henryiii:
Dropping the changlog here - will fill out and finish in part 2:

```rest
v2.6.0 (IN PROGRESS)
-----------------------------------------------------

* CMake 3.7 or newer is now required, with newer versions strongly preferred.
  `#2338 <https://github.com/pybind/pybind11/pull/2338>`_.

Several details of the CMake support have been deprecated; warnings will be
shown if you need to change something. The changes are:

* ``PYBIND11_CPP_STANDARD=<platform-flag>`` is deprecated, please use
  ``CMAKE_CXX_STANDARD=<number>`` instead, or any other valid CMake CXX or CUDA
  standard selection method, like ``target_compile_features``.

* If you do not request a standard, PyBind11 targets will compile with the
  compiler default, but not less than C++11, instead of forcing C++14.

* Direct ``pybind11::module`` usage should always be accompanied by at least
  ``set(CMAKE_CXX_VISIBILITY_PRESET hidden)`` or similar - it used to try to
  manually force this compiler flag (but not correctly on all compilers or with
  CUDA).

* ``pybind11_add_module``'s ``SYSTEM`` argument is deprecated - this now behaves
  like other imported libraries in both config and submodule mode, and behaves
  like a ``SYSTEM`` library by default.

* CMake 3.9+ will use CMake's built-in ``INTERPROCEDURAL_OPTIMIZATION`` instead
  of a custom flag discovery algorithm - set ``PYPBIND11_CLASSIC_LTO`` to force
  the old system on all CMake versions. ``THIN_LTO`` has no effect when the new
  system is being used, it will select the ``thin_lto`` as applicable.
```

include/pybind11/detail/common.h
henryiii: I've restored the old C++ (except for moving the lines up a bit), CMake now parses patch into path and type. We could change this later (as currently, if someone were to do a comparison on the patch version in a macro, it won't work for the dev version - not that that would be common), but that's can be addressed later if desired.

CMakeLists.txt
henryiii: I've shortened the list based on looking at a few of the feature lists in several compilers.

docs/compiling.rst
henryiii: I forgot to address this one...

tools/FindPythonLibsNew.cmake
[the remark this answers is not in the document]
henryiii: I've turned off comment reformatting for now. We might be able to reactivate it at some point, I assume it can parse help headers like this one, just have to see what's off. But that's for later.

CMakeLists.txt
henryiii: Done.

docs/compiling.rst
henryiii: Changed.

docs/compiling.rst
henryiii: This was the only one.

CMakeLists.txt
wjakob: That's a really useful reference, thanks for sharing.

henryiii:
Timed out error on appveyor, can be ignored. I don't have permission to restart the appveyor builds.

wjakob:
I restarted it

PR #464, opened 2016-10-24
Add default and converting constructors for all concrete Python types

dean0x7d:
There is currently a functionality split between `py::str()` and `py::object::str()` which can be surprising:

``` c++
const char *c_string = ...;
auto s1 = pt::str(c_string); // convert to str
std::string std_string = ...;
auto s2 = pt::str(std_string); // convert to str
py::object obj = ...;
auto s3 = pt::str(obj); // --> no conversion <--
auto s4 = obj.str(); // convert to str
```

This PR makes case `s3` do the expected conversion (if needed). Case `s4` is deprecated.

### Update

This was extended to all concrete Python types represented in pybind11. New `reinterpret_*` functions were also added to streamline taking raw `PyObject *`. For example:
```c++
// New constructors
auto l1 = list(); // default construct empty list
auto l2 = list(obj); // convert `obj` to a list (throw on error)

// DEPRECATED
auto l3 = list(ptr, true/false); // borrow or steal `PyObject *` without any conversion

// In favor of
auto lb = reinterpret_borrow<list>(ptr);
auto ls = reinterpret_steal<list>(ptr);
```

### Deprecations

- `obj.str()` is deprecated in favor of `str(obj)`.
- `T obj = ...; obj.check()` is deprecated in favor of `isinstance<T>(obj)`.
- The `T(handle, bool)` constructor is deprecated in favor of `reinterpret_borrow<T>()` and `reinterpret_steal<T>()`.
- `obj.repr()` is **removed** in favor of `repr(obj)`. There is no deprecation warning since this hasn't actually made it to a stable version yet.

### Implementation

Making this work required changing the usual type check order, e.g.:

``` c++
auto l = list(ptr, true);
if (l.check())
   // ...
```

The above check happens after a `py::object` type is created, however this would not work correctly with a converting constructor since `py::str` would be able to make a string representation of any object which would essentially make its `.check()` always true. This would also make using `py::str` as a function parameter very difficult since it would shadow every other overload.

To overcome this, type creation and checking are reversed with the following syntax:

``` c++
if (isinstance<list>(ptr)) {
    auto l = reinterpret_borrow<list>(ptr);
    // ...
}
```

The `py::isinstance<T>(obj)` syntax was chosen because it can work uniformly for types derived from `py::object` as well as user types wrapped with `py:class_`.

``` c++
class Pet { ... };
py::class_<Pet>(...);

m.def("is_pet", [](py::object obj) {
    return py::isinstance<Pet>(obj); // works as expected
});
```

[...]

PR #2340, opened 2020-07-29
Fixing pybind11::bytes() ambiguous overload issue and adding missing bytes type to test_constructors().

rwgk:
Two of the additions work, but `pybind11::bytes(obj)` generates this compilation error (tested with c++11 & c++2a):

```
/usr/local/google/home/rwgk/forked/pybind11/tests/test_pytypes.cpp:214:23: error: ambiguous conversion for functional-style cast from 'pybind11::detail::item_accessor' (aka 'accessor<accessor_policies::generic_item>') to 'py::bytes'
            "bytes"_a=py::bytes(d["bytes"]),
                      ^~~~~~~~~~~~~~~~~~~~
/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/detail/../pytypes.h:957:21: note: candidate constructor
    PYBIND11_OBJECT(bytes, object, PYBIND11_BYTES_CHECK)
                    ^
/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/detail/../pytypes.h:957:21: note: candidate constructor
/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/detail/../pytypes.h:987:15: note: candidate constructor
inline bytes::bytes(const pybind11::str &s) {
              ^
1 error generated.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -784,7 +784,9 @@ PYBIND11_NAMESPACE_END(detail)
         Name(handle h, stolen_t) : Parent(h, stolen_t{}) { } \
         PYBIND11_DEPRECATED("Use py::isinstance<py::python_type>(obj) instead") \
         bool check() const { return m_ptr != nullptr && (bool) CheckFun(m_ptr); } \
-        static bool check_(handle h) { return h.ptr() != nullptr && CheckFun(h.ptr()); }
+        static bool check_(handle h) { return h.ptr() != nullptr && CheckFun(h.ptr()); } \
+        template <typename Policy_> \
+        Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
 
 #define PYBIND11_OBJECT_CVT(Name, Parent, CheckFun, ConvertFun) \
     PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
@@ -794,9 +796,7 @@ PYBIND11_NAMESPACE_END(detail)
     { if (!m_ptr) throw error_already_set(); } \
     Name(object &&o) \
     : Parent(check_(o) ? o.release().ptr() : ConvertFun(o.ptr()), stolen_t{}) \
-    { if (!m_ptr) throw error_already_set(); } \
-    template <typename Policy_> \
-    Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
+    { if (!m_ptr) throw error_already_set(); }
 
 #define PYBIND11_OBJECT(Name, Parent, CheckFun) \
     PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -197,6 +197,7 @@ TEST_SUBMODULE(pytypes, m) {
     // test_constructors
     m.def("default_constructors", []() {
         return py::dict(
+            "bytes"_a=py::bytes(),
             "str"_a=py::str(),
             "bool"_a=py::bool_(),
             "int"_a=py::int_(),
@@ -210,6 +211,7 @@ TEST_SUBMODULE(pytypes, m) {
 
     m.def("converting_constructors", [](py::dict d) {
         return py::dict(
+            "bytes"_a=py::bytes(d["bytes"]),
             "str"_a=py::str(d["str"]),
             "bool"_a=py::bool_(d["bool"]),
             "int"_a=py::int_(d["int"]),
@@ -225,6 +227,7 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("cast_functions", [](py::dict d) {
         // When converting between Python types, obj.cast<T>() should be the same as T(obj)
         return py::dict(
+            "bytes"_a=d["bytes"].cast<py::bytes>(),
             "str"_a=d["str"].cast<py::str>(),
             "bool"_a=d["bool"].cast<py::bool_>(),
             "int"_a=d["int"].cast<py::int_>(),
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -188,23 +188,29 @@ def test_accessors():
 
 def test_constructors():
     """C++ default and converting constructors are equivalent to type calls in Python"""
-    types = [str, bool, int, float, tuple, list, dict, set]
+    types = [bytes, str, bool, int, float, tuple, list, dict, set]
     expected = {t.__name__: t() for t in types}
+    if str is bytes:  # Python 2.
+        # Note that bytes.__name__ == 'str' in Python 2.
+        # pybind11::str is unicode even under Python 2.
+        expected["bytes"] = bytes()
+        expected["str"] = u""  # flake8 complains about unicode().
     assert m.default_constructors() == expected
 
     data = {
-        str: 42,
-        bool: "Not empty",
-        int: "42",
-        float: "+1e3",
-        tuple: range(3),
-        list: range(3),
-        dict: [("two", 2), ("one", 1), ("three", 3)],
-        set: [4, 4, 5, 6, 6, 6],
-        memoryview: b'abc'
+        "bytes": b'41',  # Currently no supported or working conversions.
+        "str": 42,
+        "bool": "Not empty",
+        "int": "42",
+        "float": "+1e3",
+        "tuple": range(3),
+        "list": range(3),
+        "dict": [("two", 2), ("one", 1), ("three", 3)],
+        "set": [4, 4, 5, 6, 6, 6],
+        "memoryview": b'abc'
     }
-    inputs = {k.__name__: v for k, v in data.items()}
-    expected = {k.__name__: k(v) for k, v in data.items()}
+    inputs = {k: v for k, v in data.items()}
+    expected = {k: eval(k)(v) for k, v in data.items()}
 
     assert m.converting_constructors(inputs) == expected
     assert m.cast_functions(inputs) == expected


YannickJadoul:
Clue: the situation for `py::str` is almost exactly the same, except that `py::bytes` has `PYBIND11_OBJECT`, while `py::str` has `PYBIND11_OBJECT_CVT`.

YannickJadoul:
Next clue: this makes things compile:

```diff
diff --git a/include/pybind11/pytypes.h b/include/pybind11/pytypes.h
index 536835f7..d1c528a6 100644
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -802,7 +802,9 @@ PYBIND11_NAMESPACE_END(detail)
     PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
     /* This is deliberately not 'explicit' to allow implicit conversion from object: */ \
     Name(const object &o) : Parent(o) { } \
-    Name(object &&o) : Parent(std::move(o)) { }
+    Name(object &&o) : Parent(std::move(o)) { } \
+    template <typename Policy_> \
+    Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
 
 #define PYBIND11_OBJECT_DEFAULT(Name, Parent, CheckFun) \
     PYBIND11_OBJECT(Name, Parent, CheckFun) \
```

What I don't get is:
- Why does `py::bytes` not use `PYBIND11_OBJECT_CVT`? Is there a reason `py::bytes` cannot have a converting constructor?
- Why does `PYBIND11_OBJECT` not call `check_`? The constructor taking an `object` basically says "OK, I trust this is the type you say it is", without checking, almost like a C++ `reinterpret_cast` (except segfault safe, because Python is dynamically typed and only the `PyObject*` gets stored anyway). Maybe there are some good reasons for having this, but I feel this should have a much more dangerously looking construct (`reinterpret_as`?) that shouts "I assume you know what you're doing".

rwgk:
> * Why does `PYBIND11_OBJECT` not call `check_`?

One thing that comes to mind: in Python 2 `int` and `long` are different types. But that's the exception, not the rule.

YannickJadoul:
> One thing that comes to mind: in Python 2 `int` and `long` are different types. But that's the exception, not the rule.

`int_` is still using `PYBIND11_OBJECT_CVT`, and uses the `PYBIND11_LONG_CHECK` macro for that (different between [Python 2](https://github.com/pybind/pybind11/blob/1491c94c9d742d2d7bfccfc57b3944d66e5e048a/include/pybind11/detail/common.h#L200) and [Python 3](https://github.com/pybind/pybind11/blob/1491c94c9d742d2d7bfccfc57b3944d66e5e048a/include/pybind11/detail/common.h#L173), for exactly the reason you describe).

https://github.com/pybind/pybind11/blob/1491c94c9d742d2d7bfccfc57b3944d66e5e048a/include/pybind11/pytypes.h#L1070-L1072

rwgk:
I morphed this PR from a demo for the `pybind11::bytes()` ambiguous overload issue into a full PR that I want to merge.

Thanks @YannickJadoul for pointing out the fix!

You will see that I'm changing `PYBIND11_OBJECT_COMMON`. It isn't mentioned in the documentation, and isn't used directly anywhere in the Google codebase, which I take as an indication that it is probably rarely used externally.

YannickJadoul:
Nice, getting some consistency/sanity into all of this :-)
Just a heads-up: I don't know when I will be able to do a full review. It might take a few days.

All in all, this seems fine to me, but I still want to see these two things answered (I'll dive into all blames and PRs) before I'm confident in saying that this is the way to go:

> What I don't get is:
> 
>    * Why does `py::bytes` not use `PYBIND11_OBJECT_CVT`? Is there a reason `py::bytes` cannot have a converting constructor?
> 
>    * Why does `PYBIND11_OBJECT` not call `check_`? The constructor taking an `object` basically says "OK, I trust this is the type you say it is", without checking, almost like a C++ `reinterpret_cast` (except segfault safe, because Python is dynamically typed and only the `PyObject*` gets stored anyway). Maybe there are some good reasons for having this, but I feel this should have a much more dangerously looking construct (`reinterpret_as`?) that shouts "I assume you know what you're doing".

If anyone has any insights into this, though, do say so!

rwgk:
> Nice, getting some consistency/sanity into all of this :-)
> Just a heads-up: I don't know when I will be able to do a full review. It might take a few days.
> 
> All in all, this seems fine to me, but I still want to see these two things answered (I'll dive into all blames and PRs) before I'm confident in saying that this is the way to go:
> 
> > What I don't get is:
> > 
> > * Why does `py::bytes` not use `PYBIND11_OBJECT_CVT`? Is there a reason `py::bytes` cannot have a converting constructor?
> > * Why does `PYBIND11_OBJECT` not call `check_`? The constructor taking an `object` basically says "OK, I trust this is the type you say it is", without checking, almost like a C++ `reinterpret_cast` (except segfault safe, because Python is dynamically typed and only the `PyObject*` gets stored anyway). Maybe there are some good reasons for having this, but I feel this should have a much more dangerously looking construct (`reinterpret_as`?) that shouts "I assume you know what you're doing".
> 
> If anyone has any insights into this, though, do say so!

Thanks Yannick, I strongly feel that these questions are best addressed in separate PRs.
I'm really glad to finally get some solid ground under my feet with this PR. This PR is an important stepping stone for me to make progress with the other PRs I opened, and eventually #2198 which got me started on a small odyssey.

YannickJadoul:
> Thanks Yannick, I strongly feel that these questions are best addressed in separate PRs.

Oh, I see. I was/am a bit afraid of making things even more complicated, potentially forgetting to tie up loose ends. If `py::bytes` is not using `PYBIND11_OBJECT_CVT` by mistake, but there was a good reason to have `PYBIND11_OBJECT` not have a conversion from `detail::accessor<Policy_>`, we might be making things worse.


> I'm really glad to finally get some solid ground under my feet with this PR. This PR is an important stepping stone for me to make progress with the other PRs I opened, and eventually #2198 which got me started on a small odyssey.

Right, I understand. Things have gotten complex very fast!
I do actually believe this is a correct fix.
My main concern is that this code was written and maintained by smart people before, and I'm suspicious about overlooking some complex reasoning why things are the way they are. Every "quick fix" I've wanted to do in pybind11 turned out to not be that straightforward at all ;-)

Regardless of that, I have a suggestion: what if we'd make a longer-running `str-bytes-consistency` branch in the pybind11 repo to collect all str/bytes/unicode/... changes? That would allow us to merge things fix by fix, see progress, but also still keep an eye on the global picture (and provide @wjakob with a comprehensive, complete overview at the end).

rwgk:
> Regardless of that, I have a suggestion: what if we'd make a longer-running `str-bytes-consistency` branch in the pybind11 repo to collect all str/bytes/unicode/... changes?

I have mixed feelings.
Thoughts:
* I look at this PR as a harmless bug fix, as close as it gets.
* It makes all PYBIND11_OBJECT macros _consistent_. — Note that "PYBIND11_OBJECT" does not appear in the documentation and therefore this family of macros can safely be considered an implementation detail; it appears in only 3 .h files.
* It adds to an existing test.

That's really all.

Branching adds a lot of friction. Generally, I don't have great experience with branching. Note also Bertrand Meyer's Agile class (edx), it calls out branching as one of the top bad ideas.

To be completely thorough: the `"str": u""` assignment for Python 2 in the modified test makes that part no longer rely on the chimera nature of `pybind11::str` (#2343 is nothing more than a proof of that, with just a small tweak to existing tests; I'm planning to drop that PR completely).

The chimera issue will be fixed with #2256, modified by following your suggestion to replace `detail::PyUnicode_Check_Permissive` with `PyUnicode_Check`. The only thing I still have to figure out for #2256 is to NOT disable implicit `pybind11::bytes` to `pybind11:str` conversion at the same time. If I can achieve that, #2256 will again be a non-behavior-changing pure bug fix.

#2256 will be the only additional step after this PR needed to make `pybind11:str` true to the intent that it encapsulates Python 3 `str` (it is not a chimera anymore), which will mean we're in a much better place, from which we can then safely experiment with using `PYBIND11_OBJECT_CVT` for `pybind11::bytes`, and adding calls to `check_` (although I expect they will be redundant because we will not have a "permissive" `CheckFun` anymore).

I think branching will make this only more difficult, especially branching for this pure bug fix, and hopefully already the next step (#2256) will make things fall into place.

YannickJadoul:
Related PR, introducing `PYBIND11_OBJECT_CVT` for `py::str` but `PYBIND11_OBJECT` for `py::bytes`: https://github.com/pybind/pybind11/pull/464 (https://github.com/pybind/pybind11/commit/e18bc02fc9bc17f537aa3d57ebd08c277b6a3959)

YannickJadoul:
Alright, Ralf and I have been going back in time 3 or 4 years, and found this in pybind11's history:
 
>    * Why does `py::bytes` not use `PYBIND11_OBJECT_CVT`? Is there a reason `py::bytes` cannot have a converting constructor?

https://github.com/pybind/pybind11/pull/464/ introduced a rework of `PYBIND11_OBJECT_CVT` and `PYBIND11_OBJECT`. Quoting https://github.com/pybind/pybind11/pull/464#issuecomment-257034932, the reason to not have `py::bytes` use `PYBIND11_OBJECT_CVT` is:

> The only odd class is bytes. There are significant differences between Python 2.x and 3.x which make a general object -> bytes conversion pretty messy. There is already a str -> bytes converting constructor which should be enough for most cases, so I didn't want to complicate this too much.

We'll see in a later PR how easy it is to increase consistency here
 
>     * Why does `PYBIND11_OBJECT` not call `check_`? The constructor taking an `object` basically says "OK, I trust this is the type you say it is", without checking, almost like a C++ `reinterpret_cast` (except segfault safe, because Python is dynamically typed and only the `PyObject*` gets stored anyway). Maybe there are some good reasons for having this, but I feel this should have a much more dangerously looking construct (`reinterpret_as`?) that shouts "I assume you know what you're doing".

This still needs to be investigated (another PR to try things out), but there is no reason that `PYBIND11_OBJECT` wouldn't have a constructor taking a `detail::accessor<Policy>`, just like `PYBIND11_OBJECT_CVT`, as far as I understand. The PR that introduced that is https://github.com/pybind/pybind11/pull/1076, and since it's a small PR fixing a bug report, a reasonable explanation is that the `PYBIND11_OBJECT` constructor taking `detail::accessor<Policy>` was overlooked. So +1 for adding it to `PYBIND11_OBJECT` as well (through `PYBIND11_OBJECT_COMMON`, in particular, as this PR does it).

YannickJadoul:
Merging into `str-bytes-cleanup` instead of `master`, so we have a long-lived tracking PR to see all changes to `str`/`bytes`

Issue #2061, opened 2020-01-09
AttributeError: 'module' object has no attribute 'argv'


pavelschon:
## Issue description

`sys.argv` is not available when using embedded interpreter.

## Reproducible example code

```c++
#include <pybind11/embed.h>

namespace py = pybind11;

int main( int argc, char** argv )
{
    py::scoped_interpreter guard{};

    py::print(py::module::import("sys").attr("argv"));
}

```

```
$ ./test
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  AttributeError: 'module' object has no attribute 'argv'
Aborted
```

## Workaround
Workaround is possible, however, it requires to call python C-API function and linkage to libpython.so.

```c++
#include <pybind11/embed.h>

namespace py = pybind11;

int main( int argc, char** argv )
{
    py::scoped_interpreter guard{};

    PySys_SetArgv(argc, argv);

    py::print(py::module::import("sys").attr("argv"));
}
```
```
$ ./test
['./test']
```
I feel that pybind11 should provide API to set _argc+argv_ without need to call Python C-API.
Related issue: #1918

pavelschon:
Just an idea: pass argc, argv to constructor of scoped_interpreter:
```c++
py::scoped_interpreter guard{argc, argv};
```

bstaletic:
I can't reproduce this. Can you recheck again?

YannickJadoul:
Also related to #1768 (though there `argv` exists, but is empty.

YannickJadoul:
It might deserve a note in the docs, at least. But an extra constructor for `guard` doesn't seem wrong either, to me.

YannickJadoul:
> It might deserve a note in the docs, at least. But an extra constructor for `scoped_interpreter` doesn't seem wrong either, to me.

bstaletic:
@YannickJadoul I'd like to make sure there is a problem that needs solving first. If it works, like it worked for me, I don't think we should add another constructor "just because".

pavelschon:
> @YannickJadoul I'd like to make sure there is a problem that needs solving first. If it works, like it worked for me, I don't think we should add another constructor "just because".

Yes, there is a problem: sys.argv is not set by pybind11 automatically as one may expect.
Workaround is to explicitly call  PySys_SetArgv(argc, argv);.

I proposed a solution: add alternate constructor to py::scoped_interpreter accepting argc+argv.

YannickJadoul:
[...]

Issue #1768, opened 2019-04-13
Cannot render open ai gym environment


adam-r-kowalski:
I’m having an issue with wrapping open ai gym. I created a minimal reproducible example that demonstrates what I mean

```c++
#include <pybind11/embed.h>

namespace py = pybind11;

auto main() -> int {
  auto interpreter = py::scoped_interpreter{};
  auto gym = py::module::import("gym");
  auto env = gym.attr("make")("CartPole-v0");
  auto state = env.attr("reset")();
  env.attr("render")();
  return 0;
}
```

This will crash, however only if I try to call the render method. If I only ask for the current state and print that out everything seems to work fine.

```
libc++abi.dylib: terminating with uncaught exception of type pybind11::error_already_set: IndexError: list index out of range

At:
  /Users/adamkowalski/miniconda3/lib/python3.7/site-packages/pyglet/window/__init__.py(560): __init__
  /Users/adamkowalski/miniconda3/lib/python3.7/site-packages/pyglet/gl/__init__.py(208): _create_shadow_window
  /Users/adamkowalski/miniconda3/lib/python3.7/site-packages/pyglet/window/__init__.py(1896): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(728): exec_module
  <frozen importlib._bootstrap>(677): _load_unlocked
  <frozen importlib._bootstrap>(967): _find_and_load_unlocked
  <frozen importlib._bootstrap>(983): _find_and_load
  /Users/adamkowalski/miniconda3/lib/python3.7/site-packages/pyglet/gl/__init__.py(239): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(728): exec_module
  <frozen importlib._bootstrap>(677): _load_unlocked
  <frozen importlib._bootstrap>(967): _find_and_load_unlocked
  <frozen importlib._bootstrap>(983): _find_and_load
  /Users/adamkowalski/miniconda3/lib/python3.7/site-packages/gym/envs/classic_control/rendering.py(25): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(728): exec_module
  <frozen importlib._bootstrap>(677): _load_unlocked
  <frozen importlib._bootstrap>(967): _find_and_load_unlocked
  <frozen importlib._bootstrap>(983): _find_and_load
[...]

PR #2341, opened 2020-07-29
Call PySys_SetArgv when initializing interpreter.

drmoose:
This PR is to suggest automatically calling `PySys_SetArgvEx` (or an equivalent function) upon initialization of a `pybind11::scoped_interpreter`.

## Summary of Changes
- Both `initialize_interpreter` and `scoped_interpreter` now have three additional arguments:
  - `int argc = 0` and `char** argv = nullptr` -- argument count and values, meant to be passed in directly from the calling program's `int main()`.
  - `bool add_current_dir_to_path`
- When `add_current_dir_to_path` is true (the default, to try to maintain some semblance of backwards compatibility), the directory that gets added to the path will be `dirname(argv[0])` rather than just `.`. If all arguments are supplied their default values (or `argv[0] == ''`) this is equivalent to the current behavior.

## Rationale

1. `sys.argv` is an `AttributeError` unless `PySys_SetArgvEx` gets called. This creates problems for GUI python code (such as matplotlib) since there's code that reads `sys.argv` in the Debian/Ubuntu version of both [`tkinter`](https://salsa.debian.org/cpython-team/python3-stdlib/-/blob/master/3.6/Lib/tkinter/__init__.py#L2018) and [gdk](https://salsa.debian.org/gnome-team/pygobject/-/blob/debian/master/gi/overrides/Gdk.py#L440). In both cases, there's no reason the value supplied needs to be `int main()`s real `argv`, the property just needs to exist.

2. The existing `initialize_interpreter` code adds the current working directory to `sys.path`, which is very similar to [CVE-2008-5983](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2008-5983). cpython itself solved this problem by introducing the `PySys_SetArgvEx` function which takes an extra argument that specifies whether to modify the path.

3. There's a significant amount of work involved in getting the arguments in the right format and figuring out exactly which overload to call. It took me a day of reading through the cpython git history to find all of these edge cases, so I'd like to save others the effort. In particular,
    -  `PySys_SetArgvEx` itself only exists in python 2.6.6-2.7.x and python >= 3.1.3. For other versions, one must call the old `PySys_SetArgv` and then correct the CVE-2008-5983 behavior by hand.
    - `PySys_SetArgvEx` only supports null arguments (special-casing them to `sys.argv=[""]`) in python 3.8+
    - In python 3.x, `PySys_SetArgv` and `PySys_SetArgvEx` take a `wchar_t**` instead of a `char**` as their parameter, and the conversion is nontrivial
    - Python itself provides a conversion function (`Py_DecodeLocale`) but only in 3.5+
    - `mbstowcs(nullptr, ...)` isn't reliable on BSD
    - Without reading Python's source code, it's difficult to tell who owns what memory when passing a converted value.


include/pybind11/embed.h
```diff
@@ -87,29 +87,126 @@ struct embedded_module {
...
+        if (nullptr == safe_argv[0]) {
+            delete[] safe_argv;
+            return;
+        }
+        safe_argv[0][0] = '\0';
+        argc = 1;
+    }
+#if PY_MAJOR_VERSION >= 3
+    // SetArgv* on python 3 takes wchar_t, so we have to convert.
+    wchar_t** widened_argv = new wchar_t*[static_cast<unsigned>(argc)];
```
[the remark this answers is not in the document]
drmoose: Correct. Although I couldn't reproduce it on any of my local compilers, github's clang gave an error there. argc is guaranteed to be positive by the if guard above, so the cast is safe.

include/pybind11/embed.h
```diff
@@ -87,29 +87,126 @@ struct embedded_module {
...
+    for (int ii = 0; ii < argc; ++ii) {
+#  if PY_MINOR_VERSION >= 5
+        // From Python 3.5 onwards, we're supposed to use Py_DecodeLocale to
+        // generate the wchar_t version of argv.
+        widened_argv[ii] = Py_DecodeLocale(safe_argv[ii], nullptr);
```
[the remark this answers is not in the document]
drmoose: Ah, but that would make the loop condition a signed comparison error

include/pybind11/embed.h
```diff
@@ -87,29 +87,126 @@ struct embedded_module {
...
+        for (int ii = 0; ii < argc; ++ii)
+            if (nullptr != widened_argv[ii])
+                FREE_WIDENED_ARG(widened_argv[ii]);
+        delete[] widened_argv;
+    }
+#  undef FREE_WIDENED_ARG
+#else
+    // python 2.x
+#  if PY_MINOR_VERSION < 6 || (PY_MINOR_VERSION == 6 && PY_MICRO_VERSION < 6)
+#    define NEED_PYRUN_TO_SANITIZE_PATH 1
```
bstaletic, line 158: Does this need to be a macro?

include/pybind11/embed.h
```diff
@@ -87,29 +87,126 @@ struct embedded_module {
...
+    }
+#if PY_MAJOR_VERSION >= 3
+    // SetArgv* on python 3 takes wchar_t, so we have to convert.
+    wchar_t** widened_argv = new wchar_t*[static_cast<unsigned>(argc)];
+    for (int ii = 0; ii < argc; ++ii) {
+#  if PY_MINOR_VERSION >= 5
+        // From Python 3.5 onwards, we're supposed to use Py_DecodeLocale to
+        // generate the wchar_t version of argv.
+        widened_argv[ii] = Py_DecodeLocale(safe_argv[ii], nullptr);
+#    define FREE_WIDENED_ARG(X) PyMem_RawFree(X)
```
bstaletic, line 114: Should this be `PyMem_RawFree()` or `PyMem_Free()`?

include/pybind11/embed.h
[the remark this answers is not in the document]
drmoose: Do you think it would be cleaner if I did

```cpp
size_t argc_unsigned = static_cast<unsigned>(argc)
```

at the top of the function and then just used `size_t`s throughout?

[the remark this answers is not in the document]
drmoose: :facepalm: ...nope. If I do that I get narrowing conversion errors on the PySys_SetArgv calls.

include/pybind11/embed.h
drmoose: `PyMem_RawFree()` was recommended specifically by the [`Py_DecodeLocale` docs](https://docs.python.org/3/c-api/sys.html#c.Py_DecodeLocale)

include/pybind11/embed.h
drmoose: I'm not sure. My worry was that if I made it a `bool`, a smart optimizer would be able to see that the `PyRun_SimpleString` call is unreachable code.

include/pybind11/embed.h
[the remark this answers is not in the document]
bstaletic: This is one of the impossible things to get completely right without using some casts. The array indices are `size_t`, yet `argv` is an `int`.

I'm not going to nag about this, but there is some inconsistency in this patch.

include/pybind11/embed.h
bstaletic: I really, really don't like using macros for flow control.

drmoose:
> Apologies for the initial two comments outside the review.

I'm not sure what this means, but I appreciate all the comments, especially this one:

> There's a lot of manual memory management going on here. There are even some non default deletes, like `PyMem_RawFree()`. Then there's checking against `nullptr` after a `new` expression, which is pointless, since `new` throws `std::bad_alloc`. This should all be fixed with liberal use of `std::unique_ptr`.

The superfluous null checking was just a dumb habit on my part (because everything in the python API needs to be null checked, I was sort-of on autopilot). The manual deletion was unbelievably ugly but it was there because I didn't have a better idea. I've rewritten the function to avoid it. Even though I'm not confident I've done this correctly (having learned _just now, from you_ that there's an array-special-case `unique_ptr<T[]>`), I'm much happier with how this looks now.

include/pybind11/embed.h
drmoose: I changed how this code is gated in the rewrite; it's still gated by the `PY_*_VERSION` macros (because I'm not sure how to _not_ do that), but I avoid the `#define`...`#undef`, which I agree was hideous.

include/pybind11/embed.h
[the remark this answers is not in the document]
drmoose: Improved slightly (`ii` is now `size_t`) but still a little awkward.

include/pybind11/embed.h
```diff
@@ -87,29 +89,112 @@ struct embedded_module {
...
+    // Use of PyMem_RawFree here instead of PyMem_Free is as recommended by the python
+    // API docs: https://docs.python.org/3/c-api/sys.html#c.Py_DecodeLocale
+    struct pymem_rawfree_deleter {
+        void operator()(void* ptr) const {
+            PyMem_RawFree(ptr);
+        }
+    };
+    std::vector< std::unique_ptr<wchar_t[], pymem_rawfree_deleter> > widened_argv_entries;
+#  else
+    std::vector< std::unique_ptr<wchar_t[]> > widened_argv_entries;
```
drmoose, line 119: @bstaletic is there a way to avoid the `std::vector` I've added here and still get smart-pointer deletion?

drmoose:
I'm not sure what's going on with the CI... the mac stage changed from "skipped" to "failed", but there aren't any errors shown when I click details. Everything seems fine on my test mac.

bstaletic:
@drmoose Sorry for the late reply. Check this out: https://gist.github.com/bstaletic/ba0afff792396ea438937948b2d80582

It's still not completely clean, but I believe it is a lot more readable.

drmoose:
Sorry it took me a while to get back to this. I've converted the changes from your gist into commits and pushed to a [`set_argv_bstaletic` branch on my fork](https://github.com/drmoose/pybind11/commits/set_argv_bstaletic). I've also cherry-picked some (but not all) of those changes to this PR. Here's the commits and my rationale:

* fb3de9f @bstaletic  Define wide_char_arg_deleter outside set_interpreter_argv.
* 3b8438e @bstaletic  Do sys.path workaround in C++ rather than eval.
* 0566160 @bstaletic  Factor out wchar conversion to a separate function.

These three are unambiguously cleaner than what I had before. Although I had to do a little bit of subsequent tweaking to get something that compiles, this was great.

* 6db8202 @bstaletic  Always use the CVE-2008-5983 fallback rather than #ifdef'ing around it.

This one scares me a bit, since it always uses the obsolete approach. Since Python have added `PySys_SetArgvEx` I feel like _using_ it where possible will be more future-proof. However, the multiline `#if`-gate I have picking between `PySys_SetArgv` and `PySys_SetArgvEx` is ridiculous and unreadable (which I imagine is why you decided to remove it in your version).  1d7f2b3 is an attempt at a more-readable version of the check.

* 0e928c7 @bstaletic  Use ifdefs instead of auto for pysys_argv decltype.

Here, I suspect we just don't agree about whether `auto` is a good thing. I've left this one off the PR because in my view it adds more lines of code without really clarifying what's going on. If you're just vehemently opposed to `auto`, though, I don't really have any strong objection.

bstaletic:
> I've also cherry-picked some (but not all) of those changes to this PR. Here's the commits and my rationale:

Yeah, I know my gist wasn't 100% clean.

> - @bstaletic Always use the CVE-2008-5983 fallback rather than #ifdef'ing around it.

Yikes. Your concern about the CVE is justified.

> However, the multiline #if-gate I have picking between PySys_SetArgv and PySys_SetArgvEx is ridiculous and unreadable (which I imagine is why you decided to remove it in your version).

Right. `#ifdef` in the middle of a function makes the function considerably harder to read. I was just trying to avoid that.

>  1d7f2b3 is an attempt at a more-readable version of the check.

It *is* cleaner, but since you created this PR, there have been some changes. Pybind supports only python 2.7 and 3.5+. Therefore, you can just assume that `PySys_SetArgvEx` exists and just use it freely.

> - 0e928c7 @bstaletic Use ifdefs instead of auto for pysys_argv decltype.

> Here, I suspect we just don't agree about whether auto is a good thing.

`auto` is a good thing. I'm just not sure if it's a good thing in this case. Leave `auto` for now. We can hear other people's opinions as well.

include/pybind11/embed.h
```diff
@@ -87,29 +89,103 @@ struct embedded_module {
...
+/// Python 2.x/3.x-compatible version of `PySys_SetArgv`
+inline void set_interpreter_argv(int argc, char** argv, bool add_current_dir_to_path) {
+    // Before it was special-cased in python 3.8, passing an empty or null argv
+    // caused a segfault, so we have to reimplement the special case ourselves.
+    char** safe_argv = argv;
+    std::unique_ptr<char*[]> argv_guard;
+    std::unique_ptr<char[]> argv_inner_guard;
+    if (nullptr == argv || argc <= 0) {
+        argv_guard = std::unique_ptr<char*[]>(safe_argv = new char*[1]);
+        argv_inner_guard = std::unique_ptr<char[]>(safe_argv[0] = new char[1]);
```
bstaletic, line 130: For the two lines above you can call `.reset()` like this:

    argv_guard.reset(safe_argv = new char*[1]);

include/pybind11/embed.h
```diff
@@ -87,29 +89,103 @@ struct embedded_module {
...
+    size_t argv_size = static_cast<size_t>(argc);
+    // SetArgv* on python 3 takes wchar_t, so we have to convert.
+    std::unique_ptr<wchar_t*[]> widened_argv(new wchar_t*[argv_size]);
+    std::vector< std::unique_ptr<wchar_t[], wide_char_arg_deleter> > widened_argv_entries;
+    for (size_t ii = 0; ii < argv_size; ++ii) {
+        widened_argv_entries.emplace_back(widen_chars(safe_argv[ii]));
+        if (!widened_argv_entries.back()) {
+            // A null here indicates a character-encoding failure or the python
+            // interpreter out of memory. Give up.
+            return;
```
bstaletic, line 144: Should we raise an exception here, instead of silently doing nothing?

include/pybind11/embed.h
```diff
@@ -87,29 +89,103 @@ struct embedded_module {
...
+            widened_argv[ii] = widened_argv_entries.back().get();
+    }
+
+    auto pysys_argv = widened_argv.get();
+#else
+    // python 2.x
+    auto pysys_argv = safe_argv;
+#endif
+
+    PySys_SetArgvEx(argc, pysys_argv, add_current_dir_to_path ? 1 : 0);
```
bstaletic, line 155: Why don't you just pass `add_current_dir_to_path`? A `bool` is losslessly and implicitly convertible to `int`.

Also, we didn't allow users the choice here before. Any reason not to hardocde it to `1`?

bstaletic (changes requested):
Thanks for the PR. Mostly looks good, just a bit of polishing.

include/pybind11/embed.h
drmoose: 8358be4

include/pybind11/embed.h
drmoose: I think silently doing nothing is actually OK here. Consider:
- All this means is that `PySys_SetArgvEx` won't be called, but current pybind doesn't call it, so probably 95% of users won't notice or care.
- Assuming locales are configured correctly on your machine, you are probably out of memory. As such, your exception handler is not likely to work.
- Assuming locales are configured _incorrectly_, a failure here will be harder to debug than the inevitable failures in the user's python code.

That said, I'm not married to the decision. It just seemed like a reasonable fallback to revert to the old behavior.

include/pybind11/embed.h
drmoose: > Why don't you just pass `add_current_dir_to_path`? A `bool` is losslessly and implicitly convertible to `int`.

Because I misremembered the implicit conversion rules and thought `(int)true` is `-1`.  13753aa

> Also, we didn't allow users the choice here before. Any reason not to hardocde it to 1?

`1` is explicitly requesting the CVE-2008-5983 behavior. My & my company's code will always use `false` (and indeed, we're already using a locally-patched pybind to fix it.) The only reason I left `true` as the default was for backwards-compatibility.

bstaletic (approved):
Okay, this gets a greenlight from me. However, I'd like someone else to take a look as well.

Either way, thanks for the pull request.

YannickJadoul (commented):
This seems like a huge amount of work, so thanks for that, @drmoose!

However, this is also quite a bit of code, and I'm not entirely convinced it should be added to pybind11. But please do try to convince me (and others), @drmoose and @bstaletic.

A few questions that first come to mind, when seeing this:
- `sys.argv` can be set directly, I believe (through `py::module::import("sys").attr("argv") = ...`). What's the reason to call `PySys_SetArgvEx`?
- Why do we want to support the `char **argv, int argc` signature, and not just `std::vector<const char*>` or even `std::vector<std::string>`. I can see that C/C++'s `main` takes these arguments, of course, but wouldn't it almost always be bad practice to immediately pass on these arguments of `main` to Python? Wouldn't you, in a situation where Python is embedded, expect the main, embedding program to handle the arguments before deciding to pass on a few to Python? Or maybe the embedding program wants to just set a fixed set of arguments? If not, it's seems more like "wrapping" CPython, rather than embedding it. It's also kind of annoying to produce a `char**`, when you're not getting it as `main` parameters (but from `char**` to `std::vector<std::string>` is reasonably straightforward).
- This argument is further strengthened by the `updatepath` in `PySys_SetArgvEx`, which tries out whether `argv[0]` is a script file (which, if you're passing on `main`s arguments, is wrong, because it'll be the executable's name?). This also somehow doesn't sense to me, except when you're thinly "wrapping" Python rather than embedding it in a larger program (and even then, not really, because it's the executable). The question here is what the default assumption should be; I'd say pybind11 should assume as little as possible, here, as long as it's possible for the user to achieve this if necessary.
- Why is `Py_DecodeLocale` being used; what advantage does it offer? I'd say it's up to the embedding program to decide on the encoding, rather than use Python's defaults here? Shouldn't we stick to pybind11's ["utf-8 everywhere"](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#passing-python-strings-to-c)?
- Do we really want to set `sys.argv == ['']` by default? I'm actually fine with that, but this was once rejected by @wjakob (https://github.com/pybind/pybind11/issues/1768#issuecomment-503474136). See also discussions in #1768 and #2061.
- On a related note: does this need to be part of the constructor? (For some reason, the order of the arguments seems funky to me, thought I know it's to keep backwards compatibility with the API without these arguments, ofc). Couldn't this be a method?

bstaletic:
So, I don't need this, but let me try to answer @YannickJadoul's questions:

> But please do try to convince me (and others),

Getting this right, as opposed to hacking it together is hard. Pybind11 is meant to abstract that and make interoperability "seamless", according to the readme.

> - sys.argv can be set directly, I believe (through py::module::import("sys").attr("argv") = ...). What's the reason to call PySys_SetArgvEx

Respecting user locale. What if `main(int argc, char** argv)`  is not carrying any form of unicode? Then you can't just add `argv[n]` to `sys.args` because of re-encoding that you'd have to do manually first.

> - Why do we want to support the char **argv, int argc signature, and not just std::vector<const char*> or even std::vector<std::string>.

The first isn't buying you much and the latter simply wouldn't work. Eventually, you have to pass an `int` and a `wchar_t**` to `SetArgv`.

> I can see that C/C++'s main takes these arguments, of course, but wouldn't it almost always be bad practice to immediately pass on these arguments of main to Python? Wouldn't you, in a situation where Python is embedded, expect the main, embedding program to handle the arguments before deciding to pass on a few to Python? Or maybe the embedding program wants to just set a fixed set of arguments? 

That's kind of a non-sequitor. When embedding the interpreter, you're free to "preprocess" the arguments, before passing them to `py::scoped_interpreter`. Also, can you think of any reason why you'd want to hide an argument from `sys.argv`?

> It's also kind of annoying to produce a char**, when you're not getting it as main parameters (but from char** to std::vector<std::string> is reasonably straightforward).

If you're embedding python, you control `main()`. Right?

> - This argument is further strengthened by the updatepath in PySys_SetArgvEx

No idea what `updatepath` is, so I'll skip this section.

> - Why is Py_DecodeLocale being used; what advantage does it offer? **I'd say it's up to the embedding program to decide on the encoding**, rather than use Python's defaults here?

Emphasis mine. This is the part that you got wrong. It's no one's decision other than the OS's. If your OS holds arguments encoded in EBCDIC (god forbid) you have no choice but to re-encode it. That's what `Py_DecodeLocale` is providing us.

> - Do we really want to set sys.argv == [''] by default?

That's what Python 3.8 does, so yes. We're just being consistent.

> - On a related note: does this need to be part of the constructor? (For some reason, the order of the arguments seems funky to me, thought I know it's to keep backwards compatibility with the API without these arguments, ofc). Couldn't this be a method?

I don't see what that would buy us, other than a slightly more awkward API.

drmoose:
@bstaletic's response is more concise and better than anything I might've written. Nevertheless, here I go:

> > -  This argument is further strengthened by the updatepath in PySys_SetArgvEx

> No idea what updatepath is, so I'll skip this section.

`updatepath` is what `PySys_SetArgvEx` calls the boolean parameter at the end, which I've called `add_current_dir_to_path`. And, the mention of a "script file" in the official python docs is misleading because they don't anticipate our use case. [`pathconfig.c`](https://github.com/python/cpython/blob/8f42748ded5e978fe8a924115179d45a74a6363b/Python/pathconfig.c#L668) treats any file that exists (as opposed to the `-m`/`-c` flags) as a "script" in context.


> this is also quite a bit of code. I'm not entirely convinced it should be added to pybind11.

Whether it's quite a bit of code depends on your perspective -- to us as humans, yes it is. But, that's because there are so many `#ifdef`-gated cases to cover. After preprocessing, the compiler sees a relatively short utility function that converts from `char**` to `wchar_t**` and then passes that value to another function. Doing that conversion right is not easy, and since each python minor version has changed the correct approach, your first attempt likely won't compile on anyone else's computer. [The `Py_Initialize` docs](https://docs.python.org/3/c-api/init.html#c.Py_Initialize) tell us to call `PySys_SetArgvEx` after initializing, so anyone who wants to embed a python will potentially run into having to do this same amount of work.

That's an argument for the logic being in _a library_ but not necessarily _in pybind11, specifically_.

I'll admit, I wasn't aware of #1768 and #2061 when I did this (I always forget to remove the default `is:open` from github searches). #2061 seems to end by inviting the OP to file effectively this exact PR. I respectfully disagree with the claim in #1768 that this is "not a pybind-specific issue,"  because pybind makes a point of calling `Py_*` functions so users don't have to in general and [_requires_ users to use its alternate interpreter initialization](https://pybind11.readthedocs.io/en/master/advanced/embedding.html#interpreter-lifetime) in particular.

But, the [existing pybind11 initialization behavior](https://github.com/pybind/pybind11/blob/v2.5.0/include/pybind11/embed.h#L111) is clearly wrong, both from a [CVE-2008-5983](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2008-5983) perspective and from a "[this is not what Python does](https://asciinema.org/a/AgFIhcLB0FtKD4jNd1YkqdAmQ)"  perspective ([gist](https://gist.github.com/drmoose/1b6a35a951153975a3e1658f26c6c240)). My thinking was, if I'm going to propose fixing it, I might as well try to fix it properly.

> `std::vector<std::string>`

This is a good idea for an overload, and I'd be willing to write & add that for cases where you want to manually provide arguments to the embedded python. But, I don't think it should be _the only_ overload.

`int, char**` is useful both in the "wrapping CPython" case _and_ in the case of embedding a python interpreter in a program that parses its own args (in that case, you could pass the argv unmodified but hardcode argc to 1, so python sees `sys.argv == ['/path/to/your/binary']`). I don't know enough about the pybind11 userbase to adequately predict which of the two overloads would be of more use to more people. I personally (writing simulation code in c++ that uses pybind to send data to matplotlib) will nearly always be using `1, argv, false`.

> (For some reason, the order of the arguments seems funky to me, thought I know it's to keep backwards compatibility with the API without these arguments, ofc).

I agree it's weird. My personal preference would be for argc, argv to be first (as the author of #2061 suggests).

> Couldn't this be a method?

It _could_, but arguably your python interpreter is invalid (and probably unusable) until the method is called, which would violate the RAII principle.

YannickJadoul:
> Getting this right, as opposed to hacking it together is hard. Pybind11 is meant to abstract that and make interoperability "seamless", according to the readme.

Right, though we also have a tradeoff to take care of, on whether we can promise to maintain this. To some degree, pybind11 should also just be extensible enough to add external code for specific use cases. But OK, I think I could be convinced about this being in pybind11.
I'd still like to get some input from @wjakob on where to draw the line, though.

> > * sys.argv can be set directly, I believe (through py::module::import("sys").attr("argv") = ...). What's the reason to call PySys_SetArgvEx
> 
> Respecting user locale. What if `main(int argc, char** argv)` is not carrying any form of unicode? Then you can't just add `argv[n]` to `sys.args` because of re-encoding that you'd have to do manually first.

But we should also support the cases where not `int argc, char** argv` is passed on. The embedded Python interpreter might only be a small part of the main program. E.g., two use cases I can think of: calling a specific Python library as part of a larger C++ calculation, or adding an interactive shell to an existing program. In both cases, "just" passing on `argc` and `argv` of the main program doesn't seem right to me, as this will contain the C++ executable and potentially some flags that are entirely irrelevant to the Python interpreter.

Moreover, if `sys.argv` are "faked" by C++ and not passed on from `main`, we'd need to follow UTF-8 everywhere, again, I believe.
 
> > * Why do we want to support the char *_argv, int argc signature, and not just std::vector<const char_> or even std::vectorstd::string.
> 
> The first isn't buying you much and the latter simply wouldn't work. Eventually, you have to pass an `int` and a `wchar_t**` to `SetArgv`.

Good point, yes. But I still think I want this thing to have a C++-friendly interface. Again, IMO this needs to be usable from `main`, but also from other points in the C++ interface. (`std::string` could be UTF-8 encoded, btw; that would already solve the `wchar_t` thing. Or we could have `std::wstring`, ofc.)

We could also just set `sys.argv` in there, instead of calling `SetArgv`?

> That's kind of a non-sequitor. When embedding the interpreter, you're free to "preprocess" the arguments, before passing them to `py::scoped_interpreter`. Also, can you think of any reason why you'd want to hide an argument from `sys.argv`?

Because some arguments might be meant for the much-bigger C++ program, and not meant to be processed or visible from Python.
Point in case, PyCharm prints `['/home/yannick/bin/pycharm/helpers/pydev/pydevconsole.py', '40575', '34033']` for `argv`, rather than a bunch of JVM arguments that have absolutely no meaning in Python.

> If you're embedding python, you control `main()`. Right?

Yup. But you don't need to construct `embedded_interpreter` in `main`, ofc. That would really limit the design possibilities.

> > * This argument is further strengthened by the updatepath in PySys_SetArgvEx
> 
> No idea what `updatepath` is, so I'll skip this section.

The docs  [say this](https://docs.python.org/3/c-api/init.html#c.PySys_SetArgvEx
>  If updatepath is non-zero, the function also modifies sys.path according to the following algorithm:
>
> - If the name of an existing script is passed in argv[0], the absolute path of the directory where the script is located is prepended to sys.path.
> - Otherwise (that is, if argc is 0 or argv[0] doesn’t point to an existing file name), an empty string is prepended to sys.path, which is the same as prepending the current working directory (".").

This is rather nice when just wrapping Python, but also quite dangerous to have as (hidden) default, if you ask me. (Minor detail, and definitely fixable, but `add_current_dir_to_path` is potentially misleading here, as well?)

> > * Why is Py_DecodeLocale being used; what advantage does it offer? **I'd say it's up to the embedding program to decide on the encoding**, rather than use Python's defaults here?
> 
> Emphasis mine. This is the part that you got wrong. It's no one's decision other than the OS's. If your OS holds arguments encoded in EBCDIC (god forbid) you have no choice but to re-encode it. That's what `Py_DecodeLocale` is providing us.

Right, OK, this part I did miss indeed. This is nice to have, indeed, but once again only for `argc, argv` in `main`, right? This is only to "undo" the OS's encoding of arguments to main? (See same argument above.)

Could we somehow split this up into two parts? One part with a static utility method where you get get `argv` decoded, and one part where the arguments are actually set. (Happy to do the work here, btw. I realize I'm quite late to the party, after @drmoose did all this work...)

> > * Do we really want to set sys.argv == [''] by default?
> 
> That's what Python 3.8 does, so yes. We're just being consistent.

Perfectly fine with this. Again, given the previous discussions in issues, this is something we probably ought to get @wjakob's input on?

> > * On a related note: does this need to be part of the constructor? (For some reason, the order of the arguments seems funky to me, thought I know it's to keep backwards compatibility with the API without these arguments, ofc). Couldn't this be a method?
> 
> I don't see what that would buy us, other than a slightly more awkward API.

Making slightly more opt-in, in the context of previous comments?

YannickJadoul:
> @bstaletic's response is more concise and better than anything I might've written. Nevertheless, here I go:

Whoops, I was writing a reply to @bstaletic, at the same time, it seems :-)

> `updatepath` is what `PySys_SetArgvEx` calls the boolean parameter at the end, which I've called `add_current_dir_to_path`. And, the mention of a "script file" in the official python docs is misleading because they don't anticipate our use case. [`pathconfig.c`](https://github.com/python/cpython/blob/8f42748ded5e978fe8a924115179d45a74a6363b/Python/pathconfig.c#L668) treats any file that exists (as opposed to the `-m`/`-c` flags) as a "script" in context.

:+1:
OK. Now I'm wondering if "any file that exist" makes things better or worse? At any rate, it would suddenly not add `'.'` anymore, if `argv[0]` is a file?

> > this is also quite a bit of code. I'm not entirely convinced it should be added to pybind11.
> 
> Whether it's quite a bit of code depends on your perspective -- to us as humans, yes it is. But, that's because there are so many `#ifdef`-gated cases to cover. After preprocessing, the compiler sees a relatively short utility function that converts from `char**` to `wchar_t**` and then passes that value to another function. Doing that conversion right is not easy, and since each python minor version has changed the correct approach, your first attempt likely won't compile on anyone else's computer. [The `Py_Initialize` docs](https://docs.python.org/3/c-api/init.html#c.Py_Initialize) tell us to call `PySys_SetArgvEx` after initializing, so anyone who wants to embed a python will potentially run into having to do this same amount of work.

I was mostly thinking about the human perspective of maintenance, indeed. I agree that the actual code that's being executed is far from a problematic amount :-)

The docs in principle don't tell you you have to call `PySys_SetArgvEx`, though. They just say it's the thing to call to set `sys.argv`. But yes, it's an annoying piece of code to write, so from that perspective it ís something useful to include.
 
> That's an argument for the logic being in _a library_ but not necessarily _in pybind11, specifically_.

It's not an easy tradeoff, no, but you do have a point this might be useful to many people. So from that perspective, yes. (I'm also not really using the embedding functionality, so bear with me, please. While reviewing this, I'm trying to get an idea of how general this is and should be :-) )

> I'll admit, I wasn't aware of #1768 and #2061 when I did this (I always forget to remove the default `is:open` from github searches). #2061 seems to end by inviting the OP to file effectively this exact PR. I respectfully disagree with the claim in #1768 that this is "not a pybind-specific issue," because pybind makes a point of calling `Py_*` functions so users don't have to in general and [_requires_ users to use its alternate interpreter initialization](https://pybind11.readthedocs.io/en/master/advanced/embedding.html#interpreter-lifetime) in particular.

No worries, issues are hard to look through. I agree here, though (as I think I said in #1768 :-) )
 
> But, the [existing pybind11 initialization behavior](https://github.com/pybind/pybind11/blob/v2.5.0/include/pybind11/embed.h#L111) is clearly wrong, both from a [CVE-2008-5983](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2008-5983) perspective and from a "[this is not what Python does](https://asciinema.org/a/AgFIhcLB0FtKD4jNd1YkqdAmQ)" perspective ([gist](https://gist.github.com/drmoose/1b6a35a951153975a3e1658f26c6c240)). My thinking was, if I'm going to propose fixing it, I might as well try to fix it properly.

Yes, OK, that's a good point. I guess it won't be a breaking change because the default is still the same?
Though from the perspective of that vulnerability, the default ought to be not adding the current path?

> > `std::vector<std::string>`
> 
> This is a good idea for an overload, and I'd be willing to write & add that for cases where you want to manually provide arguments to the embedded python. But, I don't think it should be _the only_ overload.

No, not per se. What about the idea above, though, where you'd have a `static std::vector<std::string> decode_main_arguments(int argc, char **argv)` or so?

> `int, char**` is useful both in the "wrapping CPython" case _and_ in the case of embedding a python interpreter in a program that parses its own args (in that case, you could pass the argv unmodified but hardcode argc to 1, so python sees `sys.argv == ['/path/to/your/binary']`). I don't know enough about the pybind11 userbase to adequately predict which of the two overloads would be of more use to more people. I personally (writing simulation code in c++ that uses pybind to send data to matplotlib) will nearly always be using `1, argv, false`.

Fair enough. My main issue is that this is slightly too inviting to just directly pass on the argument to `main` without thinking about it (though maybe that's fine/not our problem)? And wouldn't `{std::string(argv[0])}` not do the same job, then? (You're calling Python anyway, so the one-time overhead of copying the string seems to be negligible. Unless Python really wants to hold on to that `char**` ofc. That would make memory management messier :-( )
 
> > (For some reason, the order of the arguments seems funky to me, thought I know it's to keep backwards compatibility with the API without these arguments, ofc).
> 
> I agree it's weird. My personal preference would be for argc, argv to be first (as the author of #2061 suggests).
> 
> > Couldn't this be a method?
> 
> It _could_, but arguably your python interpreter is invalid (and probably unusable) until the method is called, which would violate the RAII principle.

OK, that's a good point as well. I guess there's not a lot we can do about it, then, except for keeping it like this.

YannickJadoul:
(Short extra note: I am starting to better see the use and design reasons for this, btw, so thanks for putting forward this argumentation :-) )

drmoose:
> Right, though we also have a tradeoff to take care of, on whether we can promise to maintain this.
> I was mostly thinking about the human perspective of maintenance, indeed.

My point about the "perspective" was meant to address precisely this concern -- it's long, but it's more maintainable than it looks because it's really a bunch of copies of the same short function, targeting different Pythons. If the cpython team decide to change their favored approach again, another version of the function will need to be `#ifdef`d in, but they seems to have settled on something stable in 3.5. The paths targeting old pythons can be left alone.

In a distant, post-apocalyptic future where there's no longer any point in supporting Python < 3.5, half of this code can be deleted harmlessly.


> And wouldn't {std::string(argv[0])} not do the same job, then? 

Only in some countries. If our implementation assumes `std::string` is UTF-8 encoded, anybody whose `LC_ALL` is set to something non-utf8 is potentially sending a string in their local codepage to be decoded as UTF-8. 


> What about the idea above, though, where you'd have a static `std::vector<std::string> decode_main_arguments(int argc, char **argv)` or so?

I suspect you may be underrating how insanely complicated this would be to implement. To go from the locale-encoded `argv` element to a utf-8 encoded `std::string`, we need...
  - `Py_Initialize` to have already been called so we have access to `Py_DecodeLocale` (this creates a chicken-egg problem with trying to pass the result to `scoped_interpreter`'s constructor)
  - A utility that re-narrows the `wchar_t*` to `char*` in utf-8 (or else make the api be `std::wstring` which hampers the goal of C++ usability)
  - Memory management so that we're not leaking any of the intermediate encodes

Then, we'd need to convert the utf-8 to `wchar_t*`.

Specifying utf-8 as the `std::string` encoding irrespective of the user's locale means we can't use `Py_DecodeLocale` or anything from posix to do the encoding (because `setlocale` is static and global), so we'll have to do `PyUnicode_AsWideCharString(PyUnicode_FromStringAndSize(str.c_str(), ...), ...)` (cleaning up all the extra allocations that occur as side effects). For this part at least, we can leverage existing pybind conversion functions because this round of conversions would happen after `Py_Initialize`.

Even getting all that memory management and pointer dancing correct, you'd be doing extra locale conversions (which are potentially lossy/crashy depending on the combination of locales at play) and (at a guess) quadrupling the quantity of incomprehensible code that needs to be maintained.

I'm willing to try if it's something you feel strongly about, but it may take me a while. And it'll be ugly on a level that makes the current PR look like a Monet by comparison.

> At any rate, it would suddenly not add '.' anymore, if argv[0] is a file?
> I guess it won't be a breaking change because the default is still the same?

Correct. It'll add the path to `argv[0]`. But, since that change depends on the pybind11 user supplying an argv, it's still backwards compatible.

> Though from the perspective of that vulnerability, the default ought to be not adding the current path?

Also correct. If anybody's using pybind embed with code that runs setuid, they're potentially opening their users to a privilege escalation vulnerability. I'm more than happy to flip the default in my branch if everyone's OK with a breaking change.

> (Minor detail, and definitely fixable, but `add_current_dir_to_path` is potentially misleading here, as well?)

Do you have a better suggestion? Because I'm not happy with this name either. I suppose we could just call it `updatepath` to match Python (although that seems annoyingly vague to me, which is why I didn't initially).

bstaletic:
Instead of writing another essay, I'll try to be short this time.

- Asking @wjakob for input is always a safe call.
- I do not think that, in @drmoose's words, the interpreter is invalid/unusable before setting `sys.argv`. Otherwise, this PR wouldn't be happening now, but many years ago. That isn't to say that I don't think this PR is useful. Quite the contrary.
- Making this functionality available from an arbitrary entry point, without access to the actual arguments passed to `main()` seems to be a *wildly* different problem than what this PR is doing. I'm not sure we need to solve both problems. For "here's a vector of strings that are arguments", just converting that to `py::list` and assigning to `module("sys").attr("argv")` seems to be good enough.
  - In other words, if a user wants to pass (a subrange of) main's argument to `sys.argv`, system argument encoding is a pain in the ass and this PR is the way to go. For "pulling a list of args out of one's ass", the current tools should be enough.
- We already have a separated utility (or a building block) for "just decoding", though only in python 3.5+ and it return `wchar_t*`.
- I don't think this should be a separate member function, though I could be persuaded otherwise.
- `add_current_dir_to_path` name is fine. The fine details regarding when it is "misleading" should be in the docs. There's no point encoding a history of CPython in a variable name.

...Still longer than expected...

drmoose:
Two minor points of clarification and then I'll shut up.

> I do not think that, in @drmoose's words, the interpreter is invalid/unusable before setting sys.argv. 

It isn't unusable _now_ but would be if we initialized without doing `module::import("sys").attr("path").cast<list>().append(".");` by default and then expected the user to call a separate argv setup function. Certainly not insurmountable.


> `add_current_dir_to_path`

How about `add_program_dir_to_path`... `current` implies `.` which is precisely what _doesn't_ happen anymore.

bstaletic:
> How about `add_program_dir_to_path`... `current` implies `.` which is precisely what doesn't happen anymore.

Right. That sounds like a good idea.

YannickJadoul:
> Only in some countries. If our implementation assumes `std::string` is UTF-8 encoded, anybody whose `LC_ALL` is set to something non-utf8 is potentially sending a string in their local codepage to be decoded as UTF-8.

OK, yes. But then the question is if it's pybind11's job to decode `argv`? I'll agree it'll come in handy, but I still have a feeling this should not be directly in the main API (as it will very likely cause confusion/problems when it's the only way to pass something). Maybe something with a named constructor (i.e., static factory method) would work? But after your full explanation, I'll first play around myself a bit :-)

> I'm willing to try if it's something you feel strongly about, but it may take me a while. And it'll be ugly on a level that makes the current PR look like a Monet by comparison.

Right, OK, I hadn't seen the full extent of this proposal :-) (One extra note: the part going from UTF-8 `std::string` to Python should be reasonably painless, though. There's already a bunch of code in pybind11 that does that in the string casters, so...). Maybe let's not go here, then.

> Also correct. If anybody's using pybind embed with code that runs setuid, they're potentially opening their users to a privilege escalation vulnerability. I'm more than happy to flip the default in my branch if everyone's OK with a breaking change.

I do have a feeling this will break a lot of use cases :-( Maybe just an extra-large warning to the docs?

YannickJadoul:
Short status: I think some input of others (especially users that embed Python!) would be good. If a few extra pairs of eyes take a look, we either have a few new suggestions or the reassurance that this would be OK. But I'm starting to see why this PR is there and why you made these choices, @drmoose. Thanks for patiently explaining! :-)

drmoose:
Maybe the OPs from #1768 and #2601 have feedback? @adam-r-kowalski @pavelschon

pavelschon:
I'm wondered how @drmoose implemented the idea from #2061 even without knowing my ticket :)

It's the natural thing to pass _argc+argv_ to the constructor of _scoped_interpeter_, following RAII and still allowing the developer a full control over what goes in.

rwgk:
@drmoose are you still around?
I'm amazed how much attention the details got already. I'm not able to comb through it all.
High-level question: is there an important reason to provide the `int argc, char** argv` (not even `const`?) interface, vs e.g. `std::vector<std::string>`? The motivation for asking: that would seem easier to use and be less likely to accidentally introduce memory bugs.

drmoose:
@rwgk It's meant to be compatible with the signature of `int main(int argc, char** argv)`, because in order for the paths to work correctly, `argv[0]` must be set, and the only reliable, cross-platform way I know to get that value is to save the one passed into `main` when the program begins.

So, most of the time, the person trying to use pybind embed starts with a `char**`. If the API changed to `vector<string>` they'd need to convert, but converting from `char**` to `vector<string>` is nontrivial because of locales (`argv` may be in the end user's local codepage, `std::string` is expected to be UTF-8).

rwgk:
Sounds good, thanks! Would it still work for your purposes if we added `const` as below (tested with c++11, ..., c++20)?

```
#include <iostream>

void foo(int argc, const char *const cargv[]) {
    for (int i = 0; i < argc; i++) {
        std::cout << cargv[i] << std::endl;
    }
}

int main(int argc, char *argv[]) {
    foo(argc, argv);
    const char *const cargv[]{"hello", "world"};
    foo(2, cargv);
    return 0;
}
```

include/pybind11/embed.h
```diff
@@ -83,29 +85,103 @@ struct embedded_module {
...
+}
+
+/// Python 2.x/3.x-compatible version of `PySys_SetArgv`
+inline void set_interpreter_argv(int argc, char** argv, bool add_program_dir_to_path) {
+    // Before it was special-cased in python 3.8, passing an empty or null argv
+    // caused a segfault, so we have to reimplement the special case ourselves.
+    char** safe_argv = argv;
+    std::unique_ptr<char*[]> argv_guard;
+    std::unique_ptr<char[]> argv_inner_guard;
+    if (nullptr == argv || argc <= 0) {
```
henryiii, line 124:
```suggestion
    if (argv == nullptr || argc <= 0) {
```

I prefer the variable first (such as in the second half of this line, in fact).

include/pybind11/embed.h
```diff
@@ -83,29 +85,103 @@ struct embedded_module {
...
+        argv_guard.reset(safe_argv = new char*[1]);
+        argv_inner_guard.reset(safe_argv[0] = new char[1]);
+        safe_argv[0][0] = '\0';
```
henryiii, line 127: Does this function really modify the char** it is passed? Here it's modifying a new one, for example. It doesn't seem directly obvious why this is not `const`. I found `const char**` preferable in some cases in CLI11, IIRC.

You can always be more strict; if you have a `char**`, you can call a `const char**` on it. But you can't (without feelling really bad about yourself about casting, anyway) go the other way.

henryiii:
This patch adds const, okay to commit?

```diff
diff --git a/include/pybind11/embed.h b/include/pybind11/embed.h
index 432185e4..246122c3 100644
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -96,7 +96,7 @@ struct wide_char_arg_deleter {
     }
 };

-inline wchar_t* widen_chars(char* safe_arg) {
+inline wchar_t* widen_chars(const char* safe_arg) {
 #if PY_VERSION_HEX >= 0x030500f0
     wchar_t* widened_arg = Py_DecodeLocale(safe_arg, nullptr);
 #else
@@ -115,18 +115,16 @@ inline wchar_t* widen_chars(char* safe_arg) {
 }

 /// Python 2.x/3.x-compatible version of `PySys_SetArgv`
-inline void set_interpreter_argv(int argc, char** argv, bool add_program_dir_to_path) {
+inline void set_interpreter_argv(int argc, const char* const* argv, bool add_program_dir_to_path) {
     // Before it was special-cased in python 3.8, passing an empty or null argv
     // caused a segfault, so we have to reimplement the special case ourselves.
-    char** safe_argv = argv;
-    std::unique_ptr<char*[]> argv_guard;
-    std::unique_ptr<char[]> argv_inner_guard;
-    if (nullptr == argv || argc <= 0) {
-        argv_guard.reset(safe_argv = new char*[1]);
-        argv_inner_guard.reset(safe_argv[0] = new char[1]);
-        safe_argv[0][0] = '\0';
+    bool special_case = (nullptr == argv || argc <= 0);
+
+    const char* const empty_argv[] {"\0"};
+    const char* const* safe_argv = special_case ? empty_argv : argv;
+    if(special_case)
         argc = 1;
-    }
+
 #if PY_MAJOR_VERSION >= 3
     auto argv_size = static_cast<size_t>(argc);
     // SetArgv* on python 3 takes wchar_t, so we have to convert.
@@ -174,7 +172,7 @@ PYBIND11_NAMESPACE_END(detail)
  \endrst */
 inline void initialize_interpreter(bool init_signal_handlers = true,
                                    int argc = 0,
-                                   char** argv = nullptr,
+                                   const char* const* argv = nullptr,
                                    bool add_program_dir_to_path = true) {
     if (Py_IsInitialized() != 0)
         pybind11_fail("The interpreter is already running");
@@ -258,7 +256,7 @@ class scoped_interpreter {
 public:
     scoped_interpreter(bool init_signal_handlers = true,
                        int argc = 0,
-                       char** argv = nullptr,
+                       const char* const* argv = nullptr,
                        bool add_program_dir_to_path = true) {
         initialize_interpreter(init_signal_handlers, argc, argv, add_program_dir_to_path);
     }
```

Skylion007:
bump @drmoose, if no objections will commit the const patch.

henryiii:
I'll commit the patch, and we can roll back if there are objections. But you can always be more strict, and can't be less, so I can't see why this would do anything but help.

include/pybind11/embed.h
```diff
@@ -83,29 +85,101 @@ struct embedded_module {
...
+    const char* const empty_argv[] {"\0"};
+    const char* const* safe_argv = special_case ? empty_argv : argv;
+    if (special_case)
+        argc = 1;
+
+#if PY_MAJOR_VERSION >= 3
+    auto argv_size = static_cast<size_t>(argc);
+    // SetArgv* on python 3 takes wchar_t, so we have to convert.
+    std::unique_ptr<wchar_t*[]> widened_argv(new wchar_t*[argv_size]);
+    std::vector< std::unique_ptr<wchar_t[], wide_char_arg_deleter> > widened_argv_entries;
```
Skylion007, line 132: Any particular reason this is an std::vector instead of c or std::array?

include/pybind11/embed.h
```diff
@@ -83,29 +85,101 @@ struct embedded_module {
...
+
+    const char* const empty_argv[] {"\0"};
+    const char* const* safe_argv = special_case ? empty_argv : argv;
+    if (special_case)
+        argc = 1;
+
+#if PY_MAJOR_VERSION >= 3
+    auto argv_size = static_cast<size_t>(argc);
+    // SetArgv* on python 3 takes wchar_t, so we have to convert.
+    std::unique_ptr<wchar_t*[]> widened_argv(new wchar_t*[argv_size]);
```
Skylion007, line 131: Shouldn't this ideally be a std::array vs a carray?

include/pybind11/embed.h
Skylion007: and if you are going to use it, you should call reserve() to reserve the space.

drmoose: We can't use `std::array` because `argv_size` is not known at compile time. Using a C type would just require more manual memory management. Your commit adding `.reserve()` to the vector is the correct solution.

drmoose:
@henryiii Sorry I didn't reply earlier; I wanted to get back to a computer before commenting because I wanted to make sure that there was a way to add `const` to the signature without having to `const_cast` it away (potential UB) for the Python 2 code path. It seems in the meantime you've solved that problem yourself with 318495e so :+1: and thank you.

include/pybind11/embed.h
```diff
@@ -83,29 +85,106 @@ struct embedded_module {
...
+        mbstowcs(widened_arg, safe_arg, count + 1);
+    }
+#endif
+    return widened_arg;
+}
+
+/// Python 2.x/3.x-compatible version of `PySys_SetArgv`
+inline void set_interpreter_argv(int argc, const char* const* argv, bool add_program_dir_to_path) {
+    // Before it was special-cased in python 3.8, passing an empty or null argv
+    // caused a segfault, so we have to reimplement the special case ourselves.
```
Skylion007, line 120: Nit: shouldn't we add an debug assert that argc >= 0? We are casting a signed int to an unsigned size_t

include/pybind11/embed.h
```diff
@@ -83,29 +85,106 @@ struct embedded_module {
...
+        }
+        widened_argv[ii] = widened_argv_entries.back().get();
+    }
+
+    auto pysys_argv = widened_argv.get();
+#else
+    // python 2.x
+    std::vector<std::string> strings{safe_argv, safe_argv+argv_size};
+    std::vector<char*> char_strings{argv_size};
+    for (std::size_t i=0; i<argv_size; ++i)
```
Skylion007, line 149: Nit: this doesn't look like it's clang-formatted properly @henryiii

Skylion007 (approved):

include/pybind11/embed.h
rwgk: I think it's covered by the `argc <= 0` below, combined with `argc = 1`.

rwgk (approved):

Issue #1878, opened 2019-08-12
Throwing a c++ exception in a rvalue causes an abort


madebr:
## Issue description

When throwing a c++ exception inside a method of an object that is an rvalue (=has no name associated) AND printing in the destructor, then the program aborts.

## Reproducible example code

pybind11 c++ python module definition:
```
#include <pybind11/pybind11.h>
#include <exception>
#include <iostream>

namespace py = pybind11;

class callback {
public:
    callback() {}
    ~callback() {
        std::cout << "c++: callback::~callback()" << std::endl;
        py::print("Here does the crash happen");
   
    }

    void on_fatal() {
        std::cout << "c++: callback::on_fatal" << std::endl;
        throw std::runtime_error("whatever");
    }
};

PYBIND11_MODULE(pymodule, m) {
    py::class_<callback> class_callback(m, "Callback");
    class_callback
        .def(py::init<>())
        .def("on_fatal", &callback::on_fatal);
}
```

When using the python module as follows, the python interpreter aborts:
```
import pymodule
pymodule.Callback().on_fatal()
```

The output is:
```
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  SystemError: <built-in method join of str object at 0x7fbe724d4bf0> returned a result with an error set
Aborted (core dumped)
```

By assigning the just created object to a variable, the abort disappears:
```
import pymodule
cb = pymodule.Callback()
cb.on_fatal()
```

arpit15:
https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1831 comment might be useful.

madebr:
Thanks for the pointer. The error happens here: https://github.com/pybind/pybind11/blob/c9d32a81f40ad540015814edf13b29980c63e39c/include/pybind11/pybind11.h#L1822

YannickJadoul:
Confirmed to still be an issue in 2.5.0. The issue is that somehow, a Python error is already set, yet the `Callback` object is collected and accesses Python again (without throwing an error, this time), confusing Python/pybind11.

PR #2342, opened 2020-07-30
Add error_scope to py::class_::dealloc() to protect destructor calls

jbarlow83:
Fixes issue #1878

@YannickJadoul as discussed in gitter

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1388,6 +1388,13 @@ private:
 
     /// Deallocates an instance; via holder, if constructed; otherwise via operator delete.
     static void dealloc(detail::value_and_holder &v_h) {
+        // We could be deallocating because we are cleaning up after a Python exception.
+        // If so, the Python error indicator will be set. We need to clear that before
+        // running the destructor, in case the destructor code calls more Python.
+        // If we don't, the Python API will exit with an exception, and pybind11 will
+        // throw error_already_set from the C++ destructor which is forbidden and triggers
+        // std::terminate().
+        error_scope scope;
         if (v_h.holder_constructed()) {
             v_h.holder<holder_type>().~holder_type();
             v_h.set_holder_constructed(false);
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -323,3 +323,9 @@ def test_non_final_final():
         class PyNonFinalFinalChild(m.IsNonFinalFinal):
             pass
     assert str(exc_info.value).endswith("is not an acceptable base type")
+
+
+# https://github.com/pybind/pybind11/issues/1878
+def test_exception_rvalue_abort():
+    with pytest.raises(RuntimeError):
+        m.PyPrintDestructor().throw_something()
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -379,6 +379,17 @@ TEST_SUBMODULE(class_, m) {
     // test_non_final_final
     struct IsNonFinalFinal {};
     py::class_<IsNonFinalFinal>(m, "IsNonFinalFinal", py::is_final());
+
+    struct PyPrintDestructor {
+        PyPrintDestructor() {}
+        ~PyPrintDestructor() {
+            py::print("Print from destructor");
+        }
+        void throw_something() { throw std::runtime_error("error"); }
+    };
+    py::class_<PyPrintDestructor>(m, "PyPrintDestructor")
+        .def(py::init<>())
+        .def("throw_something", &PyPrintDestructor::throw_something);
 }
 
 template <int N> class BreaksBase { public:


jbarlow83:
I'll do a separate PR in a few days with improvements to documentation when using destructors.

tests/test_class.py
```diff
@@ -323,3 +323,9 @@ def test_non_final_final():
         class PyNonFinalFinalChild(m.IsNonFinalFinal):
             pass
     assert str(exc_info.value).endswith("is not an acceptable base type")
+
+
+# https://github.com/pybind/pybind11/issues/1878
+def test_exception_rvalue_abort():
+    with pytest.raises(RuntimeError):
+        m.PyPrintDestructor().throw_something()
```
YannickJadoul, line 331: Can we assert that the `PyPrintDestructor()` actually gets called before the error gets handled? We need to make sure we're testing the situation we think we're testing.

include/pybind11/pybind11.h
```diff
@@ -1388,6 +1388,13 @@ class class_ : public detail::generic_type {
 
     /// Deallocates an instance; via holder, if constructed; otherwise via operator delete.
     static void dealloc(detail::value_and_holder &v_h) {
+        // We could be deallocating because we are cleaning up after a Python exception.
+        // If so, the Python error indicator will be set. We need to clear that before
+        // running the destructor, in case the destructor code calls more Python.
+        // If we don't the Python API will exit with an exception, and pybind11 will
```
rwgk, line 1394: Comma after don't:
If we don't, the ...

tests/test_class.py
rwgk: Sounds tricky, because you'd have to somehow know what is "before" and insert something there.
My opinion: if the test fails without the change to pybind11.h we're good.

bstaletic: Yes, the "before" does fail as described in #1878

YannickJadoul: OK, if it's fine with you two, why not :-)

YannickJadoul: Actually, it's probably OK here, but in an `assert`, pytest rewrites the AST (as we've seen before), so it is slightly tricky: https://github.com/pybind/pybind11/pull/2291/files#r453234693

jbarlow83: Since `throw_something()` is actually `PyPrintDestructor.throw_something(self)` under the hood, it's impossible for the `PyPrintDestructor()` to be not constructed before the exception is thrown. Python needs a reference to `self` before it can make the method call.

Given the subtlety of #1878 I think any additional assertions etc. may mean we create another reference to PyPrintDestructor which would change the timing of the destructor and not trigger the error.

rwgk: I just spent about one hour to convince myself that this PR is safe to merge beyond a reasonable doubt. See approach below. (In the unlikely case that there are issues that escaped the automatic and my manual testing, it's a super easy rollback.)

I will merge this PR now. @jbarlow83 Thank you very much for the contribution!

Approach:

* I checked out current pybind11 `master`.
* I put the changes in this PR on top.
* I built with Python 2.7.18rc1 and 3.8.4rc1, using `clang++` with `-std=c++11` and `-std=c++2a` (i.e. 4 builds).
* All tests PASS in all 4 builds.
* I then commented out the 1-line change in `pybind11.h` and rebuilt the 4 combinations.
* Without the 1-line change, `test_class.py` FAILS in all 4 builds. The error messages differ between Python 2 and 3, but NOT between `-std=c++11` and `-std=c++2a`. The two kinds of error messages are pasted below.
_______________________________________________________________________________________

PYTHON 2 ERROR
```
===================================================================================== FAILURES ======================================================================================
____________________________________________________________________________ test_exception_rvalue_abort ____________________________________________________________________________

>           m.PyPrintDestructor().throw_something()
E           SystemError: error return without exception set

test_class.py:331: SystemError
------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------
Print from destructor
============================================================================== short test summary info ==============================================================================

```

_______________________________________________________________________________________

PYTHON3 ERROR
```
test_class.py ....................terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  SystemError: <built-in method join of str object at 0x7f23d0a84cf0> returned a result with an error set
Fatal Python error: Aborted

Current thread 0x00007f23d0f9d740 (most recent call first):
  File "/usr/local/google/home/rwgk/clone/pybind11/tests/test_class.py", line 331 in test_exception_rvalue_abort
  File "/usr/local/lib/python3.8/dist-packages/_pytest/python.py", line 182 in pytest_pyfunc_call
<long traceback truncated>
```
_______________________________________________________________________________________

For completeness:
```
$ clang++ --version
clang version 9.0.1-11 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

YannickJadoul:
Thanks, @jbarlow83!

PR #2343, opened 2020-07-30
additional assert demonstrating that pybind::str can also by bytes in…

rwgk:
…deed (proving isinstance<str> right)

### DIFF
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -190,7 +190,8 @@ def test_constructors():
     """C++ default and converting constructors are equivalent to type calls in Python"""
     types = [str, bool, int, float, tuple, list, dict, set]
     expected = {t.__name__: t() for t in types}
-    assert m.default_constructors() == expected
+    default_constructed = m.default_constructors()
+    assert default_constructed == expected
 
     data = {
         str: 42,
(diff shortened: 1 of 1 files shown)


tests/test_pytypes.py
```diff
@@ -190,7 +190,8 @@ def test_constructors():
     """C++ default and converting constructors are equivalent to type calls in Python"""
     types = [str, bool, int, float, tuple, list, dict, set]
     expected = {t.__name__: t() for t in types}
-    assert m.default_constructors() == expected
+    default_constructed = m.default_constructors()
+    assert default_constructed == expected
```
YannickJadoul, line 194: What's the reason for this change?

YannickJadoul:
Alright, this makes sense from the code's perspective, since both `isinstance` as well as the converting constructor from `const object&`/`object&&` use the static `check_` method of the `py::object` subclasses:

```c++
template <typename T, detail::enable_if_t<std::is_base_of<object, T>::value, int> = 0>
bool isinstance(handle obj) { return T::check_(obj); }
```

and

```c++
#define PYBIND11_OBJECT_CVT(Name, Parent, CheckFun, ConvertFun) \
    Name(const object &o) \
    : Parent(check_(o) ? o.inc_ref().ptr() : ConvertFun(o.ptr()), stolen_t{}) \
    { if (!m_ptr) throw error_already_set(); } \
    Name(object &&o) \
    : Parent(check_(o) ? o.release().ptr() : ConvertFun(o.ptr()), stolen_t{}) \
    { if (!m_ptr) throw error_already_set(); } \
    template <typename Policy_> \
    Name(const ::pybind11::detail::accessor<Policy_> &a) : Name(object(a)) { }
```

This seems wrong to me, though. Especially because there is a `str(const bytes&)` constructor as well, that actually does the conversion:

```
inline str::str(const bytes& b) {
    char *buffer;
    ssize_t length;
    if (PYBIND11_BYTES_AS_STRING_AND_SIZE(b.ptr(), &buffer, &length))
        pybind11_fail("Unable to extract bytes contents!");
    auto obj = reinterpret_steal<object>(PyUnicode_FromStringAndSize(buffer, (ssize_t) length));
    if (!obj)
        pybind11_fail("Could not allocate string object!");
    m_ptr = obj.release().ptr();
}
```

but `d["str"]` (with `py::dict d`), doesn't return a `str`, but just some proxy that's implicitly convertible to `py::object`, and only the "dynamic Python type" is actually `bytes`, so the C++ overload resolution never works here.

YannickJadoul:
(I'm assuming this is a demonstrating PR, and not something you actually want to get merged, right?)

tests/test_pytypes.py
rwgk: Reused further below for the new asserts.

rwgk:
> (I'm assuming this is a demonstrating PR, and not something you actually want to get merged, right?)

Yes, this PR is just a quick separate demonstration of the bug. I'm thinking of merging this change with 4a4684bc5c6a08bb9045ac372fc92b440d639452 under #2256, and get that merged as a baseline for fixing.

tests/test_pytypes.py
YannickJadoul: Should've spotted that.

rwgk:
PR #2366 just merged includes another way to expose that `pybind11::str` can hold `bytes`. Dropping this PR (it was meant to be a demonstration only from the beginning).

PR #2347, opened 2020-07-31
fix: allow lto to be off in CMake 3.9+

### LINKED ISSUE #1415: GCC 4.8 LTO Crashes
## Issue description

Several places on the internet see build issues of pybind11 with GCC 4.8, probably in a linker bug during LTO:
- https://stackoverflow.com/questions/47819210/pybind11-lto-wrapper-failed
- https://github.com/spack/spack/issues/7618

```
lto1: internal compiler error: in build_abbrev_table, at dwarf2out.c:7478
```

Have you seen this with pybind11 before?

Is there a way to *externally* pass `NO_EXTRAS` in calls to `pybind11_add_module` via generic pybind11 CMake options? This would allow users to build pybind11 projects on GCC 4.8 without triggering the bug and without re-writing CMake files of packages (e.g. packaging solutions such as conda or spack).

Maybe this is also just a mismatch between dependencies in those reports in the compile chain?

## Reproducible example code

I saw this recently on Travis-CI but can't reproduce it in docker containers (yet).

Works on Ubuntu Trusty:
```bash
$ docker pull ubuntu:14.04
$ docker run -it ubuntu:14.04

> apt-get update
> apt-get install -y python-dev python-pip git cmake g++ make
> pip install -U pytest

> git clone https://github.com/pybind/pybind11.git
# cd pybind11 && git checkout v2.2.2 && cd ..
> mkdir build
> cd build

> cmake ../pybind11
# gcc 4.8.4
> make
# ...
````

Surprisingly works on Centos 7:
```bash
$ docker pull centos:7
$ docker run -it centos:7

> yum -y update
> yum install -y epel-release
> yum install -y python-devel python-pip git cmake gcc gcc-c++ make
> pip install -U pytest
> git clone https://github.com/pybind/pybind11.git
> mkdir build
> cd build

> cmake ../pybind11
# gcc 4.8.5
> make
# ...
```

henryiii:
If `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is defined to `OFF`, respect that (CMake 3.9+). Closes #1415.


henryiii:
By the way, I could respect this being set to OFF even before 3.9, to have a universal way to turn it off in emergencies.

henryiii:
I'm going to have to address this in a slightly different way - will either be part of the FindPython work or added just after it.

henryiii:
PR #2370 does this in a slightly different way.

PR #2353, opened 2020-08-01
tests: avoid putting build products into source directory

henryiii:
As @rwgk has noted, and as sometimes comes up on the Gitter, pybind11 doesn't respect build and source separation - it dumps the test binaries into the source directories, meaning you can't have multiple build directories at once; or, at least, you have to recompile each time you want to change directories (which eliminates the main reason to have build directories in the first place). This does some cleanup, and makes sure that the source directories do not get artifacts! Only one Python file is copied in; the rest is done via PyTest config.

Also added a warning if someone tries an in-place build, which is a bad idea in general, and should be avoided.

* All build products go in the binary dir
    * PyTest's ini file solves the old issue that required this ugly hack
* Warns if in place builds are used (but they are still fully supported & tested)
* Warns if separate build directory is being used but output files have already polluted the source directory
* Drops the continue-on-error attempt in GHA, didn't do anything.
* If build products are in the source (in place build or leftover), you can run pytest directly on Python 2 (was broken before)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,16 +50,10 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
   ### Warn if not an out-of-source builds
   if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
     set(lines
-        "You are building in-place, which is highly discouraged in CMake. "
-        "Please consider cleaning your source directory and building in a separate folder. "
-        "For example:\n"
-        "rm -r CMakeCache.txt CMakeFiles/ cmake_uninstall.cmake pybind11Config.cmake pybind11ConfigVersion.cmake tests/CMakeFiles/\n"
-    )
-    if(CMAKE_VERSION VERSION_LESS 3.14)
-      list(APPEND lines "mkdir build\ncd build\ncmake ..\n cmake --build .")
-    else()
-      list(APPEND lines "cmake -S . -B build\ncmake --build build")
-    endif()
+        "You are building in-place. If that is not what you intended to "
+        "do, you can clean the source directory with:\n"
+        "rm -r CMakeCache.txt CMakeFiles/ cmake_uninstall.cmake pybind11Config.cmake "
+        "pybind11ConfigVersion.cmake tests/CMakeFiles/\n")
     message(AUTHOR_WARNING ${lines})
   endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,23 @@ if(NOT pybind11_FIND_QUIETLY)
 endif()
 
 # Check if pybind11 is being used directly or via add_subdirectory
-if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  ### Warn if not an out-of-source builds
+  if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
+    set(lines
+        "You are building in-place, which is highly discouraged in CMake. "
+        "Please consider cleaning your source directory and building in a separate folder. "
+        "For example:\n"
+        "rm -r CMakeCache.txt CMakeFiles/ cmake_uninstall.cmake pybind11Config.cmake pybind11ConfigVersion.cmake tests/CMakeFiles/\n"
+    )
+    if(CMAKE_VERSION VERSION_LESS 3.14)
+      list(APPEND lines "mkdir build\ncd build\ncmake ..\n cmake --build .")
+    else()
+      list(APPEND lines "cmake -S . -B build\ncmake --build build")
+    endif()
+    message(AUTHOR_WARNING ${lines})
+  endif()
+
   set(PYBIND11_MASTER_PROJECT ON)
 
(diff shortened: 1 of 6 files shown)


henryiii:
Current problems: Everything works, except on Windows, where I get:

Python 2.7:
```
Failed to import pybind11_tests from pytest:
   ImportError: AttributeError: gil_scoped
```


Python 3.5 and 3.6:
```
import pybind11_tests  # noqa: F401 imported but unused
  ImportError: SystemError: <class 'property'> returned a result with an error set
```

Python 3.7 is fine. (I'm also taking advantage of this to test a few extra configurations, but not related to the actual PR)

henryiii:
Further debugging: There’s a hack in the `conftest.py` that hides the real errors on importing. Removing the hack gives clearer errors:

Python 2.7 can’t import anything from `pybind11_tests` . `ImportError: cannot import name gil_scoped` and such for each submodule.

Python 3.5 and 3.6 produce `ImportError: generic_type: type "ConstructorStats" is already registered!`  every time.
And of course, all other OSs, and other Pythons, work fine. And building in place works fine, even on Windows + Python 2.7, 3.5, and 3.6.

---

Python 2.7:

```
 _____________________ ERROR collecting test_gil_scoped.py _____________________
  ImportError while importing test module 'D:\a\pybind11\pybind11\tests\test_gil_scoped.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  ..\..\tests\test_gil_scoped.py:4: in <module>
      from pybind11_tests import gil_scoped as m
  E   ImportError: cannot import name gil_scoped
  ______________________ ERROR collecting test_iostream.py ______________________
  ImportError while importing test module 'D:\a\pybind11\pybind11\tests\test_iostream.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  ..\..\tests\test_iostream.py:2: in <module>
      from pybind11_tests import iostream as m
  E   ImportError: cannot import name iostream
...
```

Python 3.5 and 3.6:

```
_______________________ ERROR collecting test_async.py ________________________
  ImportError while importing test module 'D:\a\pybind11\pybind11\tests\test_async.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  ..\..\tests\test_async.py:4: in <module>
      from pybind11_tests import async_module as m
  E   ImportError: generic_type: type "ConstructorStats" is already registered!
  ______________________ ERROR collecting test_buffers.py _______________________
  ImportError while importing test module 'D:\a\pybind11\pybind11\tests\test_buffers.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  ..\..\tests\test_buffers.py:8: in <module>
      from pybind11_tests import buffers as m
  E   ImportError: generic_type: type "ConstructorStats" is already registered!
...
```

henryiii:
I'm not totally sure what I fixed to make this start working, but it's working perfectly now after rebasing on my latest fixes.

henryiii:
I've added a warning if the source directory is dirty with previous output files (the first time you build with this, you'll likely have a dirty source directory from previous runs).

CMakeLists.txt
```diff
@@ -44,7 +44,23 @@ include(CMakeDependentOption)
 message(STATUS "pybind11 v${pybind11_VERSION} ${pybind11_VERSION_TYPE}")
 
 # Check if pybind11 is being used directly or via add_subdirectory
-if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
```
YannickJadoul, line 47: Fine to add this, if you want, but is it pybind11's job to educate people on how to use CMake?
As long as pybind11 works with in-place builds, I'm not sure if it's necessary to add this code.

henryiii: The problem is it's really easy to accidentally make an inplace build when you didn't mean to. For example, let's say you are changing a define and forget to include the build directory path:

```cmake
cmake -DCMAKE_BUILD_TYPE=Debug
```

Looks like an innocent mistake, but since you are passing something (doesn't matter what), CMake will go ahead and make an in-place build with your setting. (CMake does protect you from a raw "`cmake`" call, but not when you add anything to it).

henryiii: I can change the text - in-place is supported (as it's the best way to recover exactly what used to happen with out-of-source builds), though not recommended and not required.

henryiii: How about: "You are building in-place. If that is not what you intended to do, you can clean the source directory with:"

henryiii: I changed the text, also dropped the suggestion for how to build (though I rather liked that).

wjakob:
Hi @henriii,

supporting proper out-of-tree builds sounds great to me, and those changes all look good to me. I would consider it important that no warning is generated when pybind11 is used (via add_subdirectory) by another project, and that user/project chooses to do an in-tree build. 

An aside question: do you know what CMake's policy regarding in-tree builds is? I vaguely remember seeing a deprecation warning at some point, which seemed a bit pushy (In-tree builds are perfectly OK in some cases)

Thanks,
Wenzel

wjakob (approved):

henryiii:
> But my understanding from reading the code is that no warning is generated in that case

Yes, this is only for when this is the main project. As a subproject, it's up to the implementer of that project, not us. We didn't add anything to the source directory in those cases, so no change there :)

> do you know what CMake's policy regarding in-tree builds is

As far as I know, it will always be 100% up to the package authors. Many packages do manually disable in-source builds, or generate warnings, but there are some cases where it's easier/simpler, and CMake itself isn't going to stop supporting those.

PR #2360, opened 2020-08-03
Enable py::ellipsis on Python 2

YannickJadoul:
Not sure why #1502 disabled `py::ellipsis` for Python 2. `...` doesn't seem to exist as a shortcut for `Ellipsis` (the singleton instance of the `ellipsis` type; similar to `None` and `NoneType`), but `Ellipsis` still exists (https://docs.python.org/2/library/constants.html#Ellipsis), as well as the C Python interface (https://docs.python.org/2/c-api/slice.html#c.Py_Ellipsis).

Might not be véry useful in Python 2, but I see no reason not to include it, and it broke some tests in #2349 because I forgot to disable that part of the test for Python 2.

### DIFF
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -371,6 +371,8 @@ Ellipsis
 Python 3 provides a convenient ``...`` ellipsis notation that is often used to
 slice multidimensional arrays. For instance, the following snippet extracts the
 middle dimensions of a tensor with the first and last index set to zero.
+In Python 2, the syntactic sugar ``...`` is not available, but the singleton
+``Ellipsis`` (of type ``ellipsis``) can still be used directly.
 
 .. code-block:: python
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -736,9 +736,7 @@ inline bool PyIterable_Check(PyObject *obj) {
 }
 
 inline bool PyNone_Check(PyObject *o) { return o == Py_None; }
-#if PY_MAJOR_VERSION >= 3
 inline bool PyEllipsis_Check(PyObject *o) { return o == Py_Ellipsis; }
-#endif
 
 inline bool PyUnicode_Check_Permissive(PyObject *o) { return PyUnicode_Check(o) || PYBIND11_BYTES_CHECK(o); }
 
@@ -1020,13 +1018,11 @@ public:
     none() : object(Py_None, borrowed_t{}) { }
 };
 
-#if PY_MAJOR_VERSION >= 3
 class ellipsis : public object {
 public:
     PYBIND11_OBJECT(ellipsis, object, detail::PyEllipsis_Check)
     ellipsis() : object(Py_Ellipsis, borrowed_t{}) { }
 };
-#endif
 
 class bool_ : public object {
 public:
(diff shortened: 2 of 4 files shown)


docs/advanced/pycpp/numpy.rst
```diff
@@ -371,6 +371,8 @@ Ellipsis
 Python 3 provides a convenient ``...`` ellipsis notation that is often used to
 slice multidimensional arrays. For instance, the following snippet extracts the
 middle dimensions of a tensor with the first and last index set to zero.
+In Python 2, the syntactic sugar ``...`` is not available, but the singleton
+``Ellipsis`` (of type ``ellispsis``) can still be used directly.
```
rwgk, line 375: typo

YannickJadoul: Thanks! :-)

YannickJadoul:
Thanks for the quick review!

> I see @ax3l wrote "must be excluded in python 2 tests" in a comment 2 years ago.

I requested a review from @ax3l as well, but my hypothesis is that @wjakob saw that `...` didn't work in Python 2 (which it doesn't, but `Ellipsis` still seems to be working), put preprocessor guards around the `py::ellipsis` declaration, and @ax3l saw the the tests failing because the test was run on Python 2 as well.

henryiii:
`...` works everywhere in Python 3, while in Python 2, it only worked inside brackets - but `Ellipsis` works everywhere in both versions.

YannickJadoul:
> `...` works everywhere in Python 3, while in Python 2, it only worked inside brackets - but `Ellipsis` works everywhere in both versions.

Oooooh, nice, OK! That explains a lot. I though I had used it in Python 2 with numpy!

henryiii:
Yes, it was originally added for NumPy, and was special inside slicing, just like `:` is, but it started being used as a handy token in a few places, so it was enabled everywhere in Python 3. Now it's the recommended choice for Python stubs. :)

YannickJadoul:
Python 2 on Windows failing seems to have been a fluke?
Also, apparently @ax3l is on holiday, so let's wait for @bstaletic and merge? I don't see how it could break/hurt, but if necessary, we can still revert.

henryiii:
Windows (maybe 2.7 only) iostreams miss the output once in a while. I think we need to investigate further in the future, but not related to this PR.

YannickJadoul:
> Windows (maybe 2.7 only) iostreams miss the output once in a while. I think we need to investigate further in the future, but not related to this PR.

Thanks for the reassurance. Never noticed before, but good to know :-)

rwgk:
> Python 2 on Windows failing seems to have been a fluke?
> Also, apparently @ax3l is on holiday, so let's wait for @bstaletic and merge? I don't see how it could break/hurt, but if necessary, we can still revert.

Sounds good to me.
General: With @henryiii 's super comprehensive test suite, for super simple fixes like this, I think it's fair to go ahead and merge, to free up our bandwidth for the many other things we still have piled up.

YannickJadoul:
Yes, but if @ax3l or @wjakob were not on holiday, I would've been perfectly happy waiting a few days to make sure we didn't miss any reason this was only added for Python 3  :-) (Actually, even now; it's just annoying because #2349 fails; it's perfectly reviewable, though.)

bstaletic:
>  Also, apparently @ax3l is on holiday, so let's wait for @bstaletic and merge? I don't see how it could break/hurt, but if necessary, we can still revert.

Oh, I've seen this already. Looks good to me. Feel free to merge.

PR #2362, opened 2020-08-03
pytypes: Add Gotchas section about default-constructed wrapper types and py::none()

### LINKED ISSUE #2361: `py::str x = py::none()` is a bit confusing; it converts None to a string, rather than setting the value to None
## Issue description

If I have a C++ function with `py::str` as an argument, I wanted to be able to default it to `py::none()`. However, I was later surprised in my Python code because I was getting `'None'` (a string version) rather than purely `None`.

## Reproducible example code

See this patch here:
https://github.com/pybind/pybind11/commit/9c9b1f5bd6d6a4765b3eeaba8b9f96eece151873

The test fails with the following:
```
    def test_str_with_cpp_default_none():                                                              
>       assert m.test_str_with_cpp_default_none() == (None, "Hello")                                   
E       AssertionError: assert ('None', 'Hello') == (None, 'Hello')                                    
E         At index 0 diff: 'None' != None
```

I got bit in the following downstream code:
https://github.com/RobotLocomotion/drake/blob/feb254ea0a711c6596d7b8bf720672e0a5acd329/bindings/pydrake/common/deprecation_pybind.h#L21-L22

The workarounds I could think of:
* Use `py::object`, and then let Python code enforce the safe of string-ification.
* Use `std::optional<std::string>` to make it more C++-y

## Suggestions

Two options:

* Somehow distinguish assignment-at-construction `py::str x = py::none()` (use None), assignment `py::str x; x = py::none()` and make these set the object truly to None; for conversion, `py::str(py::none())`, allow this to be stringified.
  * I doubt this is possible in C++ without doing weird implicit converting operators.
* Warn about this in the docs here: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html#
  * Something like "Warning about assigning `py::none()` to objects in C++; This may happen with defautl arguments ...".

EricCousineau-TRI:
Resolves #2361 (through docs)


EricCousineau-TRI:
This is the "document the status quo" solution. I'm a bit fearful about changing the behavior just yet, but will speak more to that in the issue.

EricCousineau-TRI:
Man... trying to explain this contract and compare against other languages was a bit of a fun exercise...

docs/advanced/pycpp/object.rst
```diff
@@ -168,3 +173,77 @@ Generalized unpacking according to PEP448_ is also supported:
...
+strict **non-nullable** reference types. They may not store a copy when
+assigned to, but they cannot store ``None``. For statically typed languages,
+this is in contrast  Java's ``String`` or ``List<E>``, or C#'s ``string`` or
+``List<T>``, which are strict nullable refernce types, and C++'s
+``std::string``, which is simply a value type, or
+``std::optional<std::string>``, which is a nullable value type.
+
+.. note::
+
+    You actually *can* make ``py::str``, ``py::dict``, etc. nullable by
```
EricCousineau-TRI, line 198: Er, I dunno if this is actually true. I know that the "is default constructed" predicates are used throughout the pybind codebase for easy checks (I think), but I'm not actually sure how `type_caster`s will actually handle those (e.g. fail fast, admit weird leaky behavior, or work fine without a hitch).

docs/advanced/pycpp/object.rst
```diff
@@ -168,3 +173,77 @@ Generalized unpacking according to PEP448_ is also supported:
...
+signatures (either pure C++, or in bound signatures). However, there are some
+"gotchas" for ``py::none()`` and how it interacts with these types. In best
+case scenarios, it will fail fast (e.g. with default arguments); in worst
+cases, it will silently work but corrupt the types you want to work with.
+
+In general, the pytypes like ``py::str``, ``py::dict``, etc., are
+strict **non-nullable** reference types. They may not store a copy when
+assigned to, but they cannot store ``None``. For statically typed languages,
+this is in contrast  Java's ``String`` or ``List<E>``, or C#'s ``string`` or
+``List<T>``, which are strict nullable refernce types, and C++'s
```
EricCousineau-TRI, line 192: Misspelled: "refernce"

docs/advanced/pycpp/object.rst
bstaletic: The default constructor doesn't construct a valid python object. I'm thinking that this might cause more confusion that it will help. Maybe we shouldn't mention this at all.

tests/test_pytypes.cpp
```diff
@@ -319,6 +319,24 @@ TEST_SUBMODULE(pytypes, m) {
...
     });
 
+    // See #2361
+    // These four tests should reflect the text in `object.rst`, the
+    // "Interaction with None" section.
+    m.def("test_str_with_default_arg_none" ,[](py::str value) {
+        return value;
+    }, py::arg("value") = py::none());
+    m.def("test_str_assign_none", []() {
+        py::str is_this_none = py::none();
```
YannickJadoul, line 329: This is not an assignment? This is a implicit conversion, I believe.

YannickJadoul:
I also find this very confusing, actually. `None` is not a `str` or a `dict` or anything in Python (just check `isinstance`), so it's only because Python is dynamically types that it feels like you can assign `None` to a `str` variable.

What's happening:
- In a pybind11-bound function, `isinstance(None, str)` fails so pybind11 says it doesn't match the signature (`type_caster` basically uses `isinstance`).
- In a pybind11-bound function with default arguments, pybind11 just converts default arguments (`py::arg(...) = ...`) as  `py::object` and adds these values to the arguments if the user doesn't pass enough arguments. Same happens as in the first bullet, `None` (created by `py::none()`) is not a `str` nor `dict` nor ... so the actual overload resolution fails. *(Yes, this is a bit weird, that you can add default arguments that can never match the type of the overload! I think it's just hard to fix, since these are all different phases of pybind11.)*
- In a normal C++ function (or lambda), there is an implicit conversion available from `py::object` to `py::str` or `py::dict` or ... (see https://github.com/pybind/pybind11/issues/2361#issuecomment-668530356). The C++ constructor/implicit conversion of `str` and `dict` check if the object respectively is a `str` or `dict`, and if not, calls a conversion function. In case of `str`, `str(None)` works in Python, while `dict(None)` results in an error.

So actually, I don't see what's wrong/broken in pybind11, and what should need explaining. This addition to the docs seems disproportionally large to me, for what's basically consistent behavior (pybind11-bound functions are consistent with Python, C++ functions are consistent with C++)?

This PR seems to add to the confusion between the actual "C++ interface for Python" with the "overload resolution of bound functions to Python".

I also strongly object against mentioning Java and C# in our docs. This is about Python and C++, and they both have a different way of handling variable. Basically, Java and C# (non-primitive) variables are all pointers and those pointers can be a null pointer. Python variables work differently, since you bind a name to an object. And the singleton `None` object has a different type than `str`, so they are not compatible. This is also why `mypy` distinguishes between `str` and `Optional[str]` (see https://mypy.readthedocs.io/en/stable/kinds_of_types.html#strict-optional).

NB: `std::optional<py::str>` should work, no? Otherwise, just 2 overloads (one with `py::str`, one with `py::none`) should also do the job.

docs/advanced/pycpp/object.rst
YannickJadoul: Indeed, this is only on a C++ level. There's a difference between `None` and "no object".

tests/test_pytypes.cpp
rwgk: To make a connection: #2366 will cover this, too (I still need to work on it though).
I think it's best to repeat the test in your context to also have complete coverage of the dimension you're sampling, but you could reuse the more general `convert_to_pybind11_str` function.
Don't let that hold up this PR if you're faster than me. I'll keep an eye out and could replace `test_str_assign_none` with the more general function later.

docs/advanced/pycpp/object.rst
EricCousineau-TRI: I think it should be mentioned somewhere... but yeah, perhaps here is not the best place. Will mull it over.

tests/test_pytypes.cpp
EricCousineau-TRI: > This is not an assignment? This is a implicit conversion, I believe.

Erp, well, actually, I think just a constructor. Implicit conversion I think would only get triggered if it were really an assignment...
Will change.

> To make a connection: #2366 will cover this, too (I still need to work on it though).

Sweet! Yeah, I will probably keep this here, and possibly try to exercise it among different types.

EricCousineau-TRI:
> So actually, I don't see what's wrong/broken in pybind11, and what should need explaining [...]

I'm down for reducing the volume of the text, but I do feel that it's an edge case worth mentioning. We've spent time hashing out this behavior here and in the issue, so I believe that is sufficient evidence that it does need explaining, but perhaps more briefly.

> I also strongly object against mentioning Java and C# in our docs.

I'm good with that. I intended to capture this as a rather complex contract, but yes, relating to Java or C# may confuse the point further.

> This is also why mypy distinguishes between str and Optional[str] (see https://mypy.readthedocs.io/en/stable/kinds_of_types.html#strict-optional).

Good point!

> NB: `std::optional<py::str>` should work, no? Otherwise, just 2 overloads (one with py::str, one with py::none) should also do the job.

Yup! Hence only capturing this as documentation, rather than vying for a change in functionality.

docs/advanced/pycpp/object.rst
EricCousineau-TRI: Removed.

docs/advanced/pycpp/object.rst
EricCousineau-TRI: Done. Still mentioned, but separated it into a separate section to make it less confusing.

tests/test_pytypes.cpp
EricCousineau-TRI: Nope, you're right. It is an implicit copy constructor.

EricCousineau-TRI: Done.

docs/advanced/pycpp/object.rst
```diff
@@ -13,6 +13,11 @@ Available types include :class:`handle`, :class:`object`, :class:`bool_`,
 :class:`iterable`, :class:`iterator`, :class:`function`, :class:`buffer`,
 :class:`array`, and :class:`array_t`.
 
+.. warning::
+
+    Be sure to review the :ref:`gotchas` before using this heavily in your C++
```
virtuald, line 18: Shouldn't this be

```
:ref:`gotchas <pytypes_gotchas>`
```

EricCousineau-TRI: Oops! Yeah, I misspelled the refname. Fixed now.

FWIW ``:ref:`pytypes_gotchas``` worked too:
image

docs/advanced/pycpp/object.rst
```diff
@@ -168,3 +173,23 @@ Generalized unpacking according to PEP448_ is also supported:
...
+``(PyObject*)nullptr``. To check for this, either use ``(bool)my_wrapper`` or
+``my_wrapper.ptr() == nullptr``.
+
+Assigning py::none() to wrappers
+--------------------------------
+
+You may be tempted to use types like ``py::str`` and ``py::dict`` in C++
+signatures (either pure C++, or in bound signatures), and assign them default values of ``py::none()``. However, in a best case scenario, it will fail fast
+because ``None`` is not convertible to that type (e.g. ``py::dict``), or in a worse case scenario, it will silently work but corrupt the types you want to
+work with (e.g. ``py::str(py::none())`` will yield ``"None"`` in Python).
```
YannickJadoul, line 195: Do these lines still need to be made the same length? I'm not a huge fan of line length limitations, but the rest of the docs do it, so for the sake of consistency, it might be nice?

docs/advanced/pycpp/object.rst
```diff
@@ -168,3 +173,23 @@ Generalized unpacking according to PEP448_ is also supported:
...
+.. _pytypes_gotchas:
+
+Gotchas
+=======
+
+Default-Constructed Wrappers
+----------------------------
+
+When a wrapper type is default-constructed, it is **not** a valid Python object (i.e. it is not ``py::none()``). It is simply the same as
+``(PyObject*)nullptr``. To check for this, either use ``(bool)my_wrapper`` or
```
YannickJadoul, line 186: Minor detail, but can we get rid of the C casts?

E.g., "It is simply the same as a ``PyObject*`` null pointer", and "To check for this, use the (implicit) conversion of the wrapper object to ``bool``: ``static_cast<bool>(my_wrapper)``." (Do we need to mention the `ptr()` way, if there's a nice C++ operator?)

tests/test_pytypes.cpp
YannickJadoul: > Nope, you're right. It is an implicit copy constructor.
> 
> I always forget the difference between `T value = y`, `T value(y)`, and `T value{y}`, and how copy elision works...

Yup, same here. It's reasonably confusing how to call it, but anyway, it's clear what it does! :-)

docs/advanced/pycpp/object.rst
henryiii: I second avoiding C-style casts.

EricCousineau-TRI: Done. Good points!

docs/advanced/pycpp/object.rst
EricCousineau-TRI: Done. Oops!

rwgk (approved):
Thanks Eric!

PR #2363, opened 2020-08-04
fix: add missing signature

henryiii:
All the other array constructors have an optional handle, but this one was missing.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -858,7 +858,7 @@ public:
         if (!m_ptr) throw error_already_set();
     }
 
-    explicit array_t(const buffer_info& info) : array(info) { }
+    explicit array_t(const buffer_info& info, handle base = handle()) : array(info, base) { }
 
     array_t(ShapeContainer shape, StridesContainer strides, const T *ptr = nullptr, handle base = handle())
         : array(std::move(shape), std::move(strides), ptr, base) { }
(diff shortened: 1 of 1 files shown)


YannickJadoul:
I was about to say "Isn't that because it's a constructor from `buffer_info`?", but apparently `buffer_info` doesn't contain a `base` either? I'm not a buffer-protocol expert, though, so I don't know if it should be in there. If not, this seems fine.

Maybe do the same for the `array_t` constructor, though?

henryiii:
FYI, here's where I ran into this:

https://github.com/scikit-hep/boost-histogram/blob/026e5e5c3ec709083fadc3ead61408d030d03d64/include/bh_python/register_histogram.hpp#L125-L132

henryiii:
Someone told me when I found this a while back it was probably just an oversight, but I don't know where that conversation was.

henryiii:
By the way, buffer info is information about the data, but it does not contain ownership information. The handle contains ownership information, so you can be sure the owning object is not destroyed while the array is still present. The other constructors support this, it's just missing from the direct buffer info constructor.

YannickJadoul:
Alright; `Py_buffer` does contain a `object` field, but `py::buffer_info` doesn't. I don't know enough about the buffer protocol to assess the details of what should do what, but adding these optional `base` arguments to the signature seems fine to me!

PR #2364, opened 2020-08-04
feat: py::type::of<T>() and py::type::of(h)

### LINKED ISSUE #1692: Support `PyTypeObject`
I couldnt find if PyBind supports type objects.  If not, then something like this is useful:

```
  struct typeobject : py::handle {
    typeobject(PyTypeObject *type)
      : py::handle((PyObject*)type)
    {
    }
    typeobject static float_() { return typeobject(&PyFloat_Type); }
    typeobject static int_() { return typeobject(&PyInt_Type); }
    typeobject static str() { return typeobject(&PyString_Type); }
  };
```

or 

```
  py::handle typeobject(PyTypeObject *type) {
    return py::handle((PyObject*)type);
  }

  template<typename T>
  py::handle typeobject();

  template<>
  py::handle typeobject<py::float_>() { return typeobject(&PyFloat_Type); }

  template<>
  py::handle typeobject<py::int_>() { return typeobject(&PyInt_Type); }

  template<>
  py::handle typeobject<py::str>() { return typeobject(&PyString_Type); }
```

for example handling type arguments
`module.def("foo", py::arg("bar") = typeobject::str());`

henryiii:
This adds `py::type::of<T>()`, which provides a way to get the Python type associated with a C++ type. Currently there are two possible ways to do this (pointed out by @YannickJadoul and @bstaletic on Gitter about a year ago, IIRC), but they require usage of internal structures or the detail namespace. This provides a natural, pythonic way to get the type object without reverting to internal details.

This is very useful in templated code; for example, in boost-histogram, it is needed to get the templated storage from a templated histogram.

EDIT (@YannickJadoul):
Closes #1692

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1240,9 +1240,9 @@ You can get the type object from a C++ class that has already been registered us
 
 .. code-block:: python
 
-    auto T_py = py::type::of<T>();
+    py::type T_py = py::type::of<T>();
 
-You can directly use ``py::type(ob)`` to get the type object from any python
+You can directly use ``py::type::of(ob)`` to get the type object from any python
 object, just like ``type(ob)`` in Python.
 
 .. versionadded:: 2.6
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1247,6 +1247,6 @@ object, just like ``type(ob)`` in Python.
 
 .. note::
 
-    Other types, like ``py::type::of<int>``, do not work, see :ref:`type-conversions`.
+    Other types, like ``py::type::of<int>()``, do not work, see :ref:`type-conversions`.
 
 .. versionadded:: 2.6
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1232,3 +1232,17 @@ appropriate derived-class pointer (e.g. using
     more complete example, including a demonstration of how to provide
     automatic downcasting for an entire class hierarchy without
     writing one get() function for each class.
+
+Accessing the type object
+=========================
+
+You can get the type object from a C++ class that has already been registered using:
+
+.. code-block:: python
+
+    auto T_py = py::type::of<T>();
+
+You can directly use ``py::type(ob)`` to get the type object from any python
+object, just like ``type(ob)`` in Python.
+
+.. versionadded:: 2.6
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1245,4 +1245,8 @@ You can get the type object from a C++ class that has already been registered us
 You can directly use ``py::type::of(ob)`` to get the type object from any python
 object, just like ``type(ob)`` in Python.
 
+.. note::
+
+    Other types, like ``py::type::of<int>``, do not work, see :ref:`type-conversions`.
+
 .. versionadded:: 2.6
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2204,6 +2204,27 @@ object object_api<Derived>::call(Args &&...args) const {
 
 PYBIND11_NAMESPACE_END(detail)
 
+
+/** \ingroup python_builtins
+    \rst
+    Return the registered type object for a C++ class, given as a template parameter.
+    py::type<T>() returns the Python type object previously registered for T.
+\endrst */
+template<typename T>
+handle type() {
+    static_assert(
+      std::is_base_of<detail::type_caster_generic, detail::make_caster<T>>::value,
+      "This currently only works for registered C++ types. The type here is most likely type converted (using type_caster)."
+    );
+
+    return detail::get_type_handle(typeid(T), true);
+}
+
+inline handle type(handle h) {
+    PyObject* obj = (PyObject *) Py_TYPE(h.ptr());
+    return handle(obj);
+}
+
 #define PYBIND11_MAKE_OPAQUE(...) \
     namespace pybind11 { namespace detail { \
         template<> class type_caster<__VA_ARGS__> : public type_caster_base<__VA_ARGS__> { }; \
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2209,7 +2209,7 @@ template<typename T>
 type type::of() {
    static_assert(
       std::is_base_of<detail::type_caster_generic, detail::make_caster<T>>::value,
-      "This currently only works for registered C++ types. The type here is most likely type converted (using type_caster)."
+      "py::type::of<T> only supports the case where T is a registered C++ types. The type here is most likely type converted (using type_caster)."
     );
 
     return type((PyTypeObject*) detail::get_type_handle(typeid(T), true).ptr());
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2205,25 +2205,16 @@ object object_api<Derived>::call(Args &&...args) const {
 PYBIND11_NAMESPACE_END(detail)
 
 
-/** \ingroup python_builtins
-    \rst
-    Return the registered type object for a C++ class, given as a template parameter.
-    py::type<T>() returns the Python type object previously registered for T.
-\endrst */
 template<typename T>
-handle type() {
-    static_assert(
+type type::of() {
+   static_assert(
       std::is_base_of<detail::type_caster_generic, detail::make_caster<T>>::value,
       "This currently only works for registered C++ types. The type here is most likely type converted (using type_caster)."
     );
 
-    return detail::get_type_handle(typeid(T), true);
+    return type((PyTypeObject*) detail::get_type_handle(typeid(T), true).ptr());
 }
 
-inline handle type(handle h) {
-    PyObject* obj = (PyObject *) Py_TYPE(h.ptr());
-    return handle(obj);
-}
 
 #define PYBIND11_MAKE_OPAQUE(...) \
     namespace pybind11 { namespace detail { \
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2209,10 +2209,10 @@ template<typename T>
 type type::of() {
    static_assert(
       std::is_base_of<detail::type_caster_generic, detail::make_caster<T>>::value,
-      "py::type::of<T> only supports the case where T is a registered C++ types. The type here is most likely type converted (using type_caster)."
+      "py::type::of<T> only supports the case where T is a registered C++ types."
     );
 
-    return type((PyTypeObject*) detail::get_type_handle(typeid(T), true).ptr());
+    return type((PyObject*) detail::get_type_handle(typeid(T), true).ptr(), borrowed_t());
 }
 
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -750,6 +750,10 @@ inline bool PyIterable_Check(PyObject *obj) {
     }
 }
 
+inline PyObject* Py_TYPE_Convert(PyObject *obj) {
+    return (PyObject*) Py_TYPE(obj);
+}
+
 inline bool PyNone_Check(PyObject *o) { return o == Py_None; }
 inline bool PyEllipsis_Check(PyObject *o) { return o == Py_Ellipsis; }
 
@@ -891,16 +895,15 @@ private:
     object value = {};
 };
 
+
+
 class type : public object {
 public:
-    PYBIND11_OBJECT_DEFAULT(type, object, PyType_Check);
-
-    // Explicit omitted here since PyTypeObject required (rather than just PyObject)
-    type(PyTypeObject* type_ptr) : object((PyObject*) type_ptr, borrowed_t()) {}
+    PYBIND11_OBJECT_CVT(type, object, PyType_Check, detail::Py_TYPE_Convert);
 
     /// Giving a handle/object gets the type from it
     static type of(const handle& h) {
-        return type(Py_TYPE(h.ptr()));
+        return type(detail::Py_TYPE_Convert(h.ptr()), borrowed_t());
     }
 
     /// Convert C++ type to py::type if prevously registered. Does not convert standard types, like int, float. etc. yet.
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -891,20 +891,21 @@ private:
     object value = {};
 };
 
-class type : public handle {
+class type : public object {
 public:
+    PYBIND11_OBJECT_DEFAULT(type, object, PyType_Check);
+
     // Explicit omitted here since PyTypeObject required (rather than just PyObject)
-    type(PyTypeObject* type_ptr) : handle((PyObject*) type_ptr) {}
+    type(PyTypeObject* type_ptr) : object((PyObject*) type_ptr, borrowed_t()) {}
 
     /// Giving a handle/object gets the type from it
-    explicit type(const handle& h) : handle((PyObject *) Py_TYPE(h.ptr())) {}
+    static type of(const handle& h) {
+        return type(Py_TYPE(h.ptr()));
+    }
 
     /// Convert C++ type to py::type if prevously registered. Does not convert standard types, like int, float. etc. yet.
     template<typename T>
     static type of();
-
-    /// Custom check function that also ensures this is a type
-    bool check() const { return ptr() != nullptr && PyType_Check(ptr());}
 };
 
 class iterable : public object {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -19,6 +19,7 @@ PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 /* A few forward declarations */
 class handle; class object;
 class str; class iterator;
+class type;
 struct arg; struct arg_v;
 
 PYBIND11_NAMESPACE_BEGIN(detail)
@@ -888,6 +889,22 @@ private:
 
 private:
     object value = {};
+};
+
+class type : public handle {
+public:
+    // Explicit omitted here since PyTypeObject required (rather than just PyObject)
+    type(PyTypeObject* type_ptr) : handle((PyObject*) type_ptr) {}
+
+    /// Giving a handle/object gets the type from it
+    explicit type(const handle& h) : handle((PyObject *) Py_TYPE(h.ptr())) {}
+
+    /// Convert C++ type to py::type if prevously registered. Does not convert standard types, like int, float. etc. yet.
+    template<typename T>
+    static type of();
+
+    /// Custom check function that also ensures this is a type
+    bool check() const { return ptr() != nullptr && PyType_Check(ptr());}
 };
 
 class iterable : public object {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -750,10 +750,6 @@ inline bool PyIterable_Check(PyObject *obj) {
     }
 }
 
-inline PyObject* Py_TYPE_Convert(PyObject *obj) {
-    return (PyObject*) Py_TYPE(obj);
-}
-
 inline bool PyNone_Check(PyObject *o) { return o == Py_None; }
 inline bool PyEllipsis_Check(PyObject *o) { return o == Py_Ellipsis; }
 
@@ -899,14 +895,13 @@ private:
 
 class type : public object {
 public:
-    PYBIND11_OBJECT_CVT(type, object, PyType_Check, detail::Py_TYPE_Convert);
+    PYBIND11_OBJECT_COMMON(type, object, PyType_Check)
 
-    /// Giving a handle/object gets the type from it
-    static type of(const handle& h) {
-        return type(detail::Py_TYPE_Convert(h.ptr()), borrowed_t());
-    }
+    explicit type(handle h): type((PyObject*) Py_TYPE(h.ptr()), borrowed_t{}) {}
+    explicit type(object ob): type((PyObject*) Py_TYPE(ob.ptr()), borrowed_t{}) {}
 
-    /// Convert C++ type to py::type if prevously registered. Does not convert standard types, like int, float. etc. yet.
+    /// Convert C++ type to py::type if previously registered. Does not convert
+    //standard types, like int, float. etc. yet.
     template<typename T>
     static type of();
 };
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -901,7 +901,8 @@ public:
     explicit type(object ob): type((PyObject*) Py_TYPE(ob.ptr()), borrowed_t{}) {}
 
     /// Convert C++ type to py::type if previously registered. Does not convert
-    //standard types, like int, float. etc. yet.
+    // standard types, like int, float. etc. yet.
+    // See https://github.com/pybind/pybind11/issues/2486
     template<typename T>
     static type of();
 };
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -895,10 +895,9 @@ private:
 
 class type : public object {
 public:
-    PYBIND11_OBJECT_COMMON(type, object, PyType_Check)
+    PYBIND11_OBJECT(type, object, PyType_Check)
 
-    explicit type(handle h): type((PyObject*) Py_TYPE(h.ptr()), borrowed_t{}) {}
-    explicit type(object ob): type((PyObject*) Py_TYPE(ob.ptr()), borrowed_t{}) {}
+    static type of(handle h) { return type((PyObject*) Py_TYPE(h.ptr()), borrowed_t{}); }
 
     /// Convert C++ type to py::type if previously registered. Does not convert
     // standard types, like int, float. etc. yet.
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -147,16 +147,8 @@ TEST_SUBMODULE(class_, m) {
             return py::type::of<Invalid>();
     });
 
-    m.def("get_type", [](py::handle h) {
-        return py::type::of(h);
-    });
-
-    m.def("get_type_direct", [](py::object ob) {
+    m.def("get_type", [](py::object ob) {
         return py::type(ob);
-    });
-
-    m.def("get_type_implicit", [](py::object ob) -> py::type {
-        return ob;
     });
 
     // test_mismatched_holder
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -139,6 +139,7 @@ TEST_SUBMODULE(class_, m) {
     // test_type
     m.def("check_type", [](int category) {
         // Currently not supported (via a fail at compile time)
+        // See https://github.com/pybind/pybind11/issues/2486
         // if (category == 2)
         //     return py::type::of<int>();
         if (category == 1)
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -134,6 +134,23 @@ TEST_SUBMODULE(class_, m) {
         );
     });
 
+    struct Invalid {};
+
+    // test_type
+    m.def("check_type", [](int category) {
+        // Currently not supported (via a fail at compile time)
+        // if (category == 2)
+        //     return py::type<int>();
+        if (category == 1)
+            return py::type<DerivedClass1>();
+        else
+            return py::type<Invalid>();
+    });
+
+    m.def("compute_type", [](py::handle h) {
+        return py::type(h);
+    });
+
     // test_mismatched_holder
     struct MismatchBase1 { };
     struct MismatchDerived1 : MismatchBase1 { };
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -140,11 +140,11 @@ TEST_SUBMODULE(class_, m) {
     m.def("check_type", [](int category) {
         // Currently not supported (via a fail at compile time)
         // if (category == 2)
-        //     return py::type<int>();
+        //     return py::type::of<int>();
         if (category == 1)
-            return py::type<DerivedClass1>();
+            return py::type::of<DerivedClass1>();
         else
-            return py::type<Invalid>();
+            return py::type::of<Invalid>();
     });
 
     m.def("compute_type", [](py::handle h) {
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -148,7 +148,7 @@ TEST_SUBMODULE(class_, m) {
     });
 
     m.def("compute_type", [](py::handle h) {
-        return py::type(h);
+        return py::type::of(h);
     });
 
     // test_mismatched_holder
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -147,8 +147,16 @@ TEST_SUBMODULE(class_, m) {
             return py::type::of<Invalid>();
     });
 
-    m.def("compute_type", [](py::handle h) {
+    m.def("get_type", [](py::handle h) {
         return py::type::of(h);
+    });
+
+    m.def("get_type_direct", [](py::object ob) {
+        return py::type(ob);
+    });
+
+    m.def("get_type_implicit", [](py::object ob) -> py::type {
+        return ob;
     });
 
     // test_mismatched_holder
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -148,8 +148,16 @@ TEST_SUBMODULE(class_, m) {
             return py::type::of<Invalid>();
     });
 
-    m.def("get_type", [](py::object ob) {
-        return py::type(ob);
+    m.def("get_type_of", [](py::object ob) {
+        return py::type::of(ob);
+    });
+
+    m.def("as_type", [](py::object ob) {
+        auto tp = py::type(ob);
+        if (py::isinstance<py::type>(ob))
+            return tp;
+        else
+            throw std::runtime_error("Invalid type");
     });
 
     // test_mismatched_holder
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -39,15 +39,25 @@ def test_type():
     # assert m.check_type(2) == int
 
 
-def test_type_py():
-    assert m.get_type(1) == int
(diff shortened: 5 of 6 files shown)


include/pybind11/cast.h
```diff
@@ -2198,6 +2198,17 @@ object object_api<Derived>::call(Args &&...args) const {
...
 PYBIND11_NAMESPACE_END(detail)
 
+
+/** \ingroup python_builtins
+    \rst
+    Return the registered type object for a C++ class, given as a template parameter.
+    py::type<T>() returns the Python type object previously registered for T.
+\endrst */
+template<typename T>
+handle type() {
```
YannickJadoul, line 2208: Should this be `handle` or `object`? I have a feeling that often, most user-facing cases return `object` rather than `handle`, or not?

include/pybind11/cast.h
```diff
@@ -2198,6 +2198,17 @@ object object_api<Derived>::call(Args &&...args) const {
...
+    return detail::get_type_handle(typeid(T), true);
```
EricCousineau-TRI, line 2209: Is there a way to have this return something like `py::none()` in C++ when the user doesn't want a fail-fast?

EricCousineau-TRI:
Thanks!!! I've had some code that, uh, dipped into the internals for a while now :sweat_smile: 
https://github.com/RobotLocomotion/drake/blob/26bf5dca63fe0c63df940ef30c2e3b5196c4339a/bindings/pydrake/common/cpp_param_pybind.cc#L101-L102

include/pybind11/cast.h
EricCousineau-TRI: Yeah, from a quick perusal, that's what I generally see too.
FWIW, the pattern seems to be return `object`, but take `handle` as an argument.

tests/test_class.cpp
```diff
@@ -134,6 +134,16 @@ TEST_SUBMODULE(class_, m) {
         );
     });
 
+    struct Invalid {};
+
+    // test_type
+    m.def("check_type", [](bool valid) {
+        if (valid)
+            return py::type<DerivedClass1>();
+        else
```
EricCousineau-TRI, line 143: To Yannick's point, could you add a test case with a comment for type-coverted class, e.g. `int`?

include/pybind11/cast.h
YannickJadoul: Yes, I believe a `PyObject*` has an implicit conversion to `handle`, so it's +- compatible with some raw C API type, but given that we manage refcounting for the user, returning `object` makes sense. If the user needs a raw `PyObject*` again, `release()` is there?

YannickJadoul:
> I am totes fine with this only dealing with registered types. Perhaps, then renaming it to something slightly more obtuse, like `py::generic_type`, or something like that to align with the hierarchy of casters using `type_caster_generic`?

I like the `py::type` name, though, so I'd prefer to keep it. But how hard would it be to make using another type for this fail compiling? Could we have some `enable_if` or `static_assert` checking `std::is_base_of<py::detail::make_caster<...>, py::detail::generic_type_caster<the_same_kind_of_decay_as_in_make_caster<...<>>>::value`?

EricCousineau-TRI:
Yeah, I think a fail fast is achievable. Perhaps with something like this in the function body?
```
static_assert(
  is_generic_type_v<T>,
  "This only works for registered C++ types. The type here is most likely type converted (using type_caster).")
```

YannickJadoul:
> `is_generic_type_v<T>`

Except that we don't have that, I believe? So it'll be something like checking that a subclass of the default `type_caster_base` or `type_caster_generic` is used? (Actually, I think it's `type_caster_base` that does this mapping between `T` and the Python type, no?)

include/pybind11/cast.h
henryiii: Is it possible to delete a user-defined type? Refcounting seemed to be unnecessary if it's something that can't be deleted.

include/pybind11/cast.h
henryiii: Not being able to convert a type is something that doesn't really exist in Python, so I selected the error throwing version of `get_type_handle`, but I think simply changing this to false would return none. Is that better than error checking? I don't think pybind11 is close to working / plans to work with error catching off?

henryiii:
If we have it fail for now with non-user registered types, it could always be added later - changing a failure to a success is always allowed. :)

include/pybind11/cast.h
YannickJadoul: That is also true, yes. Yes, I think these types can go out of scope, but pybind11 breaks at that point, anyway ;-) 
But then we should document how to convert it to `py::object` if needed (I guess this should be a `reinterpret_borrowed`?) Are there other functions that return a `handle` in such case that aren't in the `detail` namespace, though?

My use case now also works with `handle` though, so maybe this is good enough, yes.

```c++
template <typename Mixin>
void addMixinBase(py::handle &binding) {
	auto typeHandle = py::detail::get_type_handle(typeid(Mixin), true);
	binding.attr("__bases__") = py::make_tuple(typeHandle) + binding.attr("__bases__");
}
```

YannickJadoul: > but pybind11 breaks at that point, anyway ;-)

In case anyone's interested, I recently worked out this example, in the context of #2335:

<details>
  <summary>Hiding this example because it's too long/off-topic</summary>

```c++

namespace py = pybind11;

class X {};
class Y {};

PYBIND11_MODULE(example, m)
{
        py::class_<X>(m, "X");
        py::class_<Y>(m, "Y");
        m.def("foo", []() { return X(); });
        m.def("bar", []() { return Y(); });
}
```

```pycon
>>> import example
>>> example.X
<class 'example.X'>
>>> example.Y
<class 'example.Y'>
>>> example.foo()
<example.X object at 0x7f7987a83c70>
>>> example.bar()
<example.Y object at 0x7f7987a83d70>
>>> import gc
>>> gc.collect()
0
>>> gc.collect()
0
>>> gc.collect()
0
>>> example.X = example.Y
>>> example.X
<class 'example.Y'>
>>> gc.collect()
0
>>> gc.collect()
0
>>> gc.collect()
0
>>> gc.get_referrers(type(example.foo()))
[{'__name__': 'example', '__doc__': None, '__package__': None, '__loader__': None, '__spec__': None, 'X': <class 'example.X'>, 'Y': <class 'example.Y'>, 'foo': <built-in method foo of PyCapsule object at 0x7f7987a80600>, 'bar': <built-in method bar of PyCapsule object at 0x7f7987a80660>, '__file__': '/home/yannick/tmp/sandbox/build/example.cpython-38-x86_64-linux-gnu.so'}, (<class 'example.X'>, <class 'pybind11_builtins.pybind11_object'>, <class 'object'>), <slot wrapper '__init__' of 'example.X' objects>]
>>> del gc.get_referrers(type(example.foo()))[0]['X']
>>> gc.collect()
0
>>> gc.collect()
5
>>> gc.collect()
0
>>> gc.collect()
0
>>> example.bar()
<example.Y object at 0x7f7987a02ef0>
>>> example.foo()
<example.X object at 0x7f7987a02e70>
>>> example.foo()
<example.X object at 0x7f7987a02ef0>
>>> 42
Segmentation fault (core dumped)
```
</details>

include/pybind11/cast.h
YannickJadoul: A C++ type could not (yet) be exposed, right?
But I think throwing an error should definitely be the default. Maybe the other option could/should stay in `detail`, even, and a try-catch could be the proposed public solution?

include/pybind11/cast.h
henryiii: I'm not against it being an object, this was just my thinking for handle. It works for my use case, too:

```cpp
.def_property_readonly_static(
    "_storage_type",
    [](py::object) {
        return py::detail::get_type_handle(
            typeid(typename histogram_t::storage_type), true);
    });
```

YannickJadoul: `py::object_api<T>::get_type()` also returns a `handle` and is public. I'm OK with keeping it like this, then, I think.

include/pybind11/cast.h
YannickJadoul: But then yes, it's more efficient, so it could be possible. We mainly need to see that the API makes sense, indeed.

Another possible overload, btw: should we also have a `py::type(py::handle)` or so? Similar to all the `py::cast` variants?

YannickJadoul:
> If we have it fail for now with non-user registered types, it could always be added later - changing a failure to a success is always allowed. :)

Exactly! As long as we do it safely (at compile time), this would be perfect!

tests/test_class.py
```diff
@@ -25,9 +25,18 @@ def test_instance(msg):
...
+    with pytest.raises(RuntimeError) as execinfo:
+        m.check_type(0)
+
+    assert ('pybind11::detail::get_type_info: unable to find type info '
+            'for "test_submodule_class_(module&)::Invalid"' in str(execinfo.value))
+
+    with pytest.raises(RuntimeError) as execinfo:
+        assert m.check_type(2) == int
+
+    assert ('pybind11::detail::get_type_info: unable '
```
YannickJadoul, line 38: Wait, this is a runtime error, and not a compilation error? We can figure out at compile time that this is going to fail, right?

henryiii: I'm only addressing one point at a time - currently, yes. In the future, probably no. ;)

YannickJadoul: (Sorry, just got an email with changes and saw you did effort to test for runtime errors. Too impatient, then ;-) )

include/pybind11/cast.h
henryiii: Going the other way is not possible, you can't return a C++ type, but having `py::handle py::type(py::handle a)` might make sense, as a way to do python's `type(a)` in C++.

henryiii:
```cpp
inline handle type(handle h) {
    PyObject* obj = (PyObject *) Py_TYPE(h.ptr());
    return handle(obj);
}
```

works, would that make sense to add?

henryiii: (Based on https://github.com/python/cpython/blob/eaa551702d80fd67219c48ee6a13ffb571ca360b/Objects/typeobject.c#L990-L1001 )

YannickJadoul: Yes, that's what I meant! Not to C++ ofc; sorry to be unclear!
Even better, maybe: `return obj.get_type()`?

henryiii: Free functions are generally better than member functions in C++ - simpler/more composable, etc. Maybe adding this then eventually deprecating / removing `h.get_type()`?

henryiii: It would simplify the mixin, and keep the API closer to Python. `h.str()` is already deprecated in favor of `py:str()`.

YannickJadoul: `handle::get_type` does make sense to me, so I'm not sure if it's worth deprecating. If anything, I'd say to maybe implement `py::type(handle)` through `handle::get_type()`, just as a convenience, Python-like function.

henryiii: It is only used internally, not publicly in any docs or tests, so to me it seems it would read better (more python like) and simplify the mixin.

For example, randomly picking one of the 10 places it's used:

```cpp
if (!obj.get_type().is(matrix_type))
// Becomes
if (!py::type(obj).is(matrix_type))
```
For comparison, the Python would be:

```python
if not type(obj) is matrix_type
```

YannickJadoul: OK, that's a good argument! I think I am/can be convinced, then. I just don't want to be the only one to mandate an API deprecation.

@rwgk, @bstaletic, @EricCousineau-TRI?

henryiii: We can also move the discussion into an issue, though I do need to come up with an implementation for `py::type(h)` - either the one above, or relying on `h.get_type()` (unless we also wait for that).

YannickJadoul: Yeah, sorry, with this amount of comments from me, nothing ever gets merged. What about merging this PR (I believe everyone was in favor, anyway), and moving the other proposal to a next PR?

wjakob:
This looks great, but would you mind adding an explicit discussion in the documentation (the auto-extracted docstrings IIRC do end up in the documentation as well, but rather hidden..)

henryiii:
Waiting pending discussion on #2388

docs/advanced/classes.rst
```diff
@@ -562,7 +562,7 @@ crucial that instances are deallocated on the C++ side to avoid memory leaks.
 
 Destructors that call Python
 ============================
-
+earch);45M
```
jagerman, line 565: That doesn't look right :thinking:

henryiii:
```suggestion

```

henryiii: :#O Oops! Will restore the lost space tomorrow. Wasn't even something I was editing...

henryiii:
By the way, I'm open to ideas on how to get something like `py::type::of<int>()` to work. :) Also, should we use `py::type::of(h)` for the from handle form? And should this be a handle subclass or object?

include/pybind11/pytypes.h
```diff
@@ -890,6 +891,22 @@ class iterator : public object {
     object value = {};
 };
 
+class type : public handle {
+public:
+    // Explicit omitted here since PyTypeObject required (rather than just PyObject)
+    type(PyTypeObject* type_ptr) : handle((PyObject*) type_ptr) {}
```
YannickJadoul, line 897: Should we not try doing this the standard way, with the `PYBIND11_OBJECT` or `PYBIND11_OBJECT_CVT` macros? Those will add `check`, etc. (It there's a reason to nót use it, some docs on why would be nice to have documented.)

henryiii: That would make this an `py::object` subclass instead of a `py::handle` sublcass. Is that what we want for type? (final question in the comment on the PR)

EricCousineau-TRI:
> By the way, I'm open to ideas on how to get something like py::type::of<int>() to work. :)

I think there are only two real ways to do this, and one "fake" way:
(1) Don't do it! Let downstream users decide how they want to interact with `py::detail::type_caster` / RTTI.
  * For an example, see my reference to Drake in this comment: https://github.com/pybind/pybind11/pull/2364#issuecomment-671078885
(2) Update the `py::detail::type_caster` to indicate a compile-time conversion. This may be informative (e.g. the concrete type) or may not (e.g. it just returns something like `PyObjectType`).
  * This falls apart for run-time branching, so I'd suggest don't?
(3) "Fake" way: Actually convert a default-constructible object and get the type from that.
  * Fragile *and* expensive. Do. Not. Want.

> Also, should we use py::type::of(h) for the from handle form? And should this be a handle subclass or object?

I forgot where we discussed it, but I think taking an input should be `py::handle`, whereas returning an output should be `py::object`?
Similar to how `py::dtype::of` returns `py::dtype`, it seems like you could just return `py::type`.
Ref

include/pybind11/pytypes.h
EricCousineau-TRI: Yeah, I think this should just follow suite with `py::dtype` and inherit either directly from `py::object` or `py::object_api`?

EricCousineau-TRI: FWIW `py::dtype` uses `PYBIND11_OBJECT_DEFAULT`, so I do think that this should use the type-checking constructor helper, as well as inherit from `py::object`.

docs/advanced/classes.rst
```diff
@@ -1232,3 +1232,17 @@ appropriate derived-class pointer (e.g. using
...
     writing one get() function for each class.
+
+Accessing the type object
+=========================
+
+You can get the type object from a C++ class that has already been registered using:
+
+.. code-block:: python
+
+    auto T_py = py::type::of<T>();
```
wjakob, line 1243: show type here (e.g. `py::type`) just for clarity?

include/pybind11/cast.h
```diff
@@ -2203,6 +2203,18 @@ object object_api<Derived>::call(Args &&...args) const {
 
 PYBIND11_NAMESPACE_END(detail)
 
+
+template<typename T>
+type type::of() {
+   static_assert(
```
wjakob, line 2209: nice idea to check

wjakob: I'd change the error message though: ``py::type::of<T> only supports the case where T is a regis[..]`` instead of the generic "This"

wjakob (approved):
two very minor suggestions.

docs/advanced/classes.rst
henryiii: It looked like duplicated info, but it's really not, I guess, since py::type::function doesn't necessarily have to return a py::type.

include/pybind11/pytypes.h
```diff
@@ -890,6 +891,23 @@ class iterator : public object {
     object value = {};
 };
 
+class type : public object {
+public:
+    PYBIND11_OBJECT_DEFAULT(type, object, PyType_Check);
+
+    // Explicit omitted here since PyTypeObject required (rather than just PyObject)
+    type(PyTypeObject* type_ptr) : object((PyObject*) type_ptr, borrowed_t()) {}
```
YannickJadoul, line 899: Do we want/need this? I'm a bit weary of having (too many) C API types in the pybind11 API.

But ofc, with the weird C-style inheritance, it might be useful, yes. Though forcing the user to just call `type((PyObject*) something)` might be just as good? It would push the C API ugliness to the user (who already chose to interact with the C API without pybind11?).

include/pybind11/pytypes.h
```diff
@@ -890,6 +891,23 @@ class iterator : public object {
     object value = {};
 };
 
+class type : public object {
+public:
+    PYBIND11_OBJECT_DEFAULT(type, object, PyType_Check);
```
YannickJadoul, line 896: `DEFAULT`? Then you do allow for a default constructor with a `PyObject*` `nullptr`? Is there any use to allowing that, and not just have `PYBIND11_OBJECT`?

Also, this could be `PYBIND11_OBJECT_CVT` (given the right conversion function; since `Py_TYPE` is a macro, we might need a minimal wrapper), if you want to also allow `py::type(obj)` rather than `py::type::of(obj)`?

tests/test_class.cpp
```diff
@@ -134,6 +134,23 @@ TEST_SUBMODULE(class_, m) {
...
+        // Currently not supported (via a fail at compile time)
+        // if (category == 2)
+        //     return py::type::of<int>();
+        if (category == 1)
+            return py::type::of<DerivedClass1>();
+        else
+            return py::type::of<Invalid>();
+    });
+
+    m.def("compute_type", [](py::handle h) {
```
YannickJadoul, line 150: Very minor thing, but "compute"? Isn't "get" closer to what's happening?

include/pybind11/pytypes.h
YannickJadoul: That wóuld allow for non-`explicit` conversions `py::type thisType = obj;`, though. So maybe not what we want, then? :-/ But the extra constructor just like with `py::str` would be nice, though.

docs/advanced/classes.rst
YannickJadoul: It's meant to look like duplicate info, the way we're designing this API, though? ;-)

YannickJadoul:
> By the way, I'm open to ideas on how to get something like `py::type::of<int>()` to work. :)

I like @EricCousineau-TRI's second version, I think. It would make it completely optional for old casters so it wouldn't break users' code, we can easily add it to existing casters, and it's flexible. I think it's +- what I vaguely had in mind. And it does seem to be closely related to `type_casters`?

> This falls apart for run-time branching, so I'd suggest don't?

How do you mean? Having an optional `static handle pybind11::detail::type_caster::get_type_object()` would work, though, no? (not sure about the return type; should it be `handle`/`PyTypeObject*`/...?) If it exists, `py::type::of<T>` works; otherwise it doesn't.

bstaletic (commented):
@henryiii Sorry for taking so long to review this. I only have one or two minor comments.

include/pybind11/pytypes.h
YannickJadoul: Then just an `explicit type(handle)` constructor?

henryiii:
I don't think I got the reference counting right; it seems to be deleting the class it accesses. Will investigate in a day or two, unless someone sees an obvious mistake sooner. :)

include/pybind11/pytypes.h
```diff
@@ -749,6 +750,10 @@ inline bool PyIterable_Check(PyObject *obj) {
     }
 }
 
+inline PyObject* Py_TYPE_Convert(PyObject *obj) {
+    return (PyObject*) Py_TYPE(obj);
```
YannickJadoul, line 754:
```suggestion
    return handle((PyObject*) Py_TYPE(obj)).inc_ref().ptr();
```

include/pybind11/pytypes.h
```diff
@@ -890,6 +895,22 @@ class iterator : public object {
     object value = {};
 };
 
+
+
+class type : public object {
+public:
+    PYBIND11_OBJECT_CVT(type, object, PyType_Check, detail::Py_TYPE_Convert);
```
YannickJadoul, line 902: @bstaletic wasn't a fan of the implicit conversion from `object` to `type`, though?

bstaletic: I could be convinced otherwise, but it does seem odd that this would be possible:

```c++
auto f(py::type) {}

int main() { py::object i = py::int_(3); f(i); }
```

YannickJadoul: I think I agree here. I'd be happy with this as well, for example:
```c++
PYBIND11_OBJECT(type, object, PyType_Check)

explicit type(handle object) : type((PyObject*) Py_TYPE(obj), borrowed_t{}) {}
```

EricCousineau-TRI:
> > This falls apart for run-time branching, so I'd suggest don't?
>
> How do you mean? [...]

@YannickJadoul I mean when the resultant type of a type cast may change at run-time for Python objects (maybe "branching" was the wrong word).

The simplest case would be `std::variant` - there will be run-time branching of the type, and what they may want may heavily depend on their use case.

Should this return something like `Typing.Variant[...]`, where `...` is filled in with the inner type-of expressions, only for Python 3.6+?
Or should it fail fast because it's not a type they can deal with? Or ...?

And that's a relatively easy case. What about when users may actually want to distinguish between `int32` and `uint32`?
AFAIK, using `type_caster<uint32_t>` and `type_caster<int32_t>` would end up returning `int`; then users would need to write their own wrapper if they wanted something like `numpy.uint32` or `torch.uint32`.

I mean, having `py::type::of<int>()` could help, I just dunno if that should be lumped into this PR without more discussion, as I think it's more nuanced.

EricCousineau-TRI:
More concretely: I think having `py::type::of<T>` only work for generic types is a strict improvement, there's space for adding in `py::type::of<>` later, so why delay this PR?

bstaletic (approved):

EricCousineau-TRI:
Filed #2486 ;)

YannickJadoul:
> The simplest case would be `std::variant` - there will be run-time branching of the type, and what they (users) may want may heavily depend on their use case.

Right, but then for `std::variant`, you can never do that? `py::type::of<T>` doesn't get an instance, anyway?

> Should this return something like `Typing.Variant[...]`, where `...` is filled in with the inner type-of expressions, only for Python 3.6+?

That's an interesting suggestion, though, but I'm not sure. It's not really a type, right? Can you create an object using this?

> And that's a relatively easy case. What about when users may actually want to distinguish between `int32` and `uint32`?

But these do get mapped to the same Python type? I don't see the problem, here? There's a mismatch between C++ and Python, but you can still say "If you pass a C++ type `T`, you'll get a Python type `py::type::of<T>`. At least it's better than nothing.
And it provides a way to get the `int` type object, through `py::type::of<int>()`? (Maybe there's also a better API, indeed.)

> I mean, having `py::type::of<int>()` could help, I just dunno if that should be lumped into this PR without more discussion, as I think it's more nuanced.

Agreed! Let's get the easy stuff in, just making sure it doesn't stop us from achieving the more complicated stuff :-)

docs/advanced/classes.rst
```diff
@@ -1232,3 +1232,17 @@ appropriate derived-class pointer (e.g. using
...
+=========================
+
+You can get the type object from a C++ class that has already been registered using:
+
+.. code-block:: python
+
+    py::type T_py = py::type::of<T>();
+
+You can directly use ``py::type::of(ob)`` to get the type object from any python
+object, just like ``type(ob)`` in Python.
```
EricCousineau-TRI, line 1246: nit Is it worth adding `.. note::` here that other types like `py::type::of<int>` does not work, with a brief xref to this section?
https://pybind11.readthedocs.io/en/stable/advanced/cast/index.html

tests/test_class.cpp
```diff
@@ -134,6 +134,23 @@ TEST_SUBMODULE(class_, m) {
         );
     });
 
+    struct Invalid {};
+
+    // test_type
+    m.def("check_type", [](int category) {
+        // Currently not supported (via a fail at compile time)
+        // if (category == 2)
```
EricCousineau-TRI, line 142: nit Can you reference #2486 here?

include/pybind11/pytypes.h
```diff
@@ -890,6 +891,21 @@ class iterator : public object {
...
+
+class type : public object {
+public:
+    PYBIND11_OBJECT_COMMON(type, object, PyType_Check)
+
+    explicit type(handle h): type((PyObject*) Py_TYPE(h.ptr()), borrowed_t{}) {}
+    explicit type(object ob): type((PyObject*) Py_TYPE(ob.ptr()), borrowed_t{}) {}
+
+    /// Convert C++ type to py::type if previously registered. Does not convert
+    //standard types, like int, float. etc. yet.
```
EricCousineau-TRI, line 904: nit Can you reference #2486 here?

tests/test_class.py
```diff
@@ -26,6 +26,29 @@ def test_instance(msg):
...
+def test_type():
+    assert m.check_type(1) == m.DerivedClass1
+    with pytest.raises(RuntimeError) as execinfo:
+        m.check_type(0)
+
+    assert 'pybind11::detail::get_type_info: unable to find type info' in str(execinfo.value)
+    assert 'Invalid' in str(execinfo.value)
+
+    # Currently not supported
+    # assert m.check_type(2) == int
```
EricCousineau-TRI, line 38: nit Can you reference #2486 here?

EricCousineau-TRI (approved):
Looks awesome!

docs/advanced/classes.rst
henryiii: Added, please check.

EricCousineau-TRI: Thanks!

EricCousineau-TRI (approved):

henryiii:
> Right, but then for std::variant, you can never do that? py::type::of<T> doesn't get an instance, anyway?

Without verifying it could be done, I think you could have `py::type::of<std::variant<...>>(instance)` working in that case, but only if all the inner types were supported (note that there is an argument above). But I don't think supporting `<int>` means you have to support `<variant>`?

> Let's get the easy stuff in

Unless it was extremely easy, was not planning to add to this PR :)

include/pybind11/pytypes.h
```diff
@@ -890,6 +891,22 @@ class iterator : public object {
...
 };
 
+
+
+class type : public object {
+public:
+    PYBIND11_OBJECT_COMMON(type, object, PyType_Check)
+
+    explicit type(handle h): type((PyObject*) Py_TYPE(h.ptr()), borrowed_t{}) {}
+    explicit type(object ob): type((PyObject*) Py_TYPE(ob.ptr()), borrowed_t{}) {}
```
YannickJadoul, line 901: This isn't necessary, I think? `object` should be a subclass of `handle`, and there's no extra refcounting going on?

YannickJadoul: Oh, yes, some slicing, since it's not by reference. But there are precedents for that, I think?

YannickJadoul (approved):

include/pybind11/pytypes.h
YannickJadoul: If you/we choose to keep the `object` overload, better to make it a `const object &`, then. It avoids an unnecessary refcount.

YannickJadoul: Ah, no, `PYBIND11_OBJECT` already adds that overload.

https://github.com/pybind/pybind11/blob/cc982ac1cd3ec31040c0cfff3a6dd0183f7c1cac/include/pybind11/pytypes.h#L813-L817

How do we fix this such that it works in the intended way? When do we want `py::type(...)` to "cast" a `py::object` to a subclass, and when do we want it to actually call `type(...)`? Does this mean that @henryiii's original `py::type::of` is necessary, and that we can't have the `py::type(obj)` interface I've been dreaming of? :-(

henryiii: We use `PYBIND11_OBJECT_COMMON` instead of `PYBIND11_OBJECT`.

henryiii: (I'll try dropping this overload soon)

henryiii: Can this be a `const handle &`?

YannickJadoul: > We use `PYBIND11_OBJECT_COMMON` instead of `PYBIND11_OBJECT`.

🤦 How did I miss that?

> Can this be a `const handle &`?

Probably, but I don't think it's necessary. `handle` really is a fancy and more beautiful way of writing `PyObject *`.

YannickJadoul:
Sorry to open that discussion again (after first telling @henryiii to change it), but, I'm now realizing the following:

If `py::type(obj)` always gets the type of an object; we have no way of actually casting a `py::object` (being returned from a Python call or so) to be converted? Is that OK, or did I speak too soon, when saying `py::type(handle)` would be a great API?

henryiii:
I think I preferred `py::type::of(h)`, even though it's a little further from the Python API, both from symmetry with `py::type::of<T>()` and because it keeps you from casting a known type into a type. (I had tests for this, the behavior of `py::type::of` was not the same as `py::type` if you have it a valid class).

henryiii:
It looks like `PYBIND11_OBJECT` doesn't throw if the constructor doesn't pass the check function. So only manually checking the ->check is the only way to ensure the object really is a type if you do `py::type(object)`? I take it there's a reason constructers should throw in this case?

YannickJadoul:
> It looks like `PYBIND11_OBJECT` doesn't throw if the constructor doesn't pass the check function. So only manually checking the ->check is the only way to ensure the object really is a type if you do `py::type(object)`? I take it there's a reason constructers should not throw in this case? I can manually add it instead of using `PYBIND11_OBJECT` if there's not a reason.

I worked on that here in #2349, but it got stalled behind the `__qualname__` unification with Python 2 & PyPy.

henryiii:
Good, then leave as is, then we can take advantage of that if it gets merged?

YannickJadoul:
> I think I preferred `py::type::of(h)`, even though it's a little further from the Python API, both from symmetry with `py::type::of<T>()` and because it keeps you from casting a known type into a type. (I had tests for this, the behavior of `py::type::of` was not the same as `py::type` if you gave it a valid class).

I really liked `py::type(...)`, but I'm now realizing that will be ambiguous and confusing :-( So yes, sorry for having you make that change to then revert it! `py::type::of` will be better.

YannickJadoul:
> Good, then leave as is, then we can take advantage of that if it gets merged?

Good for me. I'll try to pick that up again, soon'ish...

include/pybind11/cast.h
```diff
@@ -2204,6 +2204,18 @@ object object_api<Derived>::call(Args &&...args) const {
...
+      std::is_base_of<detail::type_caster_generic, detail::make_caster<T>>::value,
+      "py::type::of<T> only supports the case where T is a registered C++ types."
+    );
+
+    return type((PyObject*) detail::get_type_handle(typeid(T), true).ptr(), borrowed_t());
```
EricCousineau-TRI, line 2215: Ahhh, nuts!
I forgot to ask: Would you be OK with exposing a "type erased" version of this?
e.g. `py::type::of(const std::type_info&)`?

It would be useful for us in code that mentioned before, e.g. here:
https://github.com/RobotLocomotion/drake/blob/a7e2a49cb673e73c40a327478850f56a3e5fde9d/bindings/pydrake/common/cpp_param_pybind.cc#L102-L111

I can do it as a follow-up if you'd like to get this merged in!

EricCousineau-TRI (approved):

YannickJadoul (approved):

include/pybind11/cast.h
henryiii: This would add a new stdlib import?

YannickJadoul: Are we sure about releasing these implementation details?
Is there a reason you can't make this a template function in your client code, @EricCousineau-TRI?

Then again, I guess these implementation details are not bound to change too soon.

YannickJadoul: This also might add complications if you expect `typeid(int)` or so to work, though?

Banning this "power-use" to the `detail` namespace, and not publicly exposing it (and potentially getting bugs, etc) - i.e., what we have now - could just work as well. This `detail` might not soon change, but I'm wondering if it's a good idea to add to the public API :-/

henryiii: I think he's just asking for a runtime version of `of<T>` using `std::type_info`? I don't think that's exposing an implementation detail?

henryiii: Given that's what's returned by `typeid(T)` anyway, it should not be an extra dependency, I can add that, I think. And one can be based on the other with no duplication.

henryiii: Okay, maybe eventually then. I think this is cleared to merge, then?

YannickJadoul: OK, we can see. Just wanted to express a few doubts; not veto'ing anything. But let's get that in another PR, then, afaic :-)

wjakob:
Thanks everyone!

PR #2365, opened 2020-08-04
Receive PyModuleDef pointer in module's constructor 

sunyj:
Optionally receiving a pre-defined  `PyModuleDef` pointer in `module`'s constructor helps some embedded Python applications a lot.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -818,7 +818,12 @@ public:
     PYBIND11_OBJECT_DEFAULT(module, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    explicit module(const char *name, const char *doc = nullptr, PyModuleDef *def = nullptr) {
+    explicit module(const char *name, const char *doc = nullptr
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Could you please expand your reasoning? pybind11 tries to abstract the C API away from you and pybind11's API never takes C API parameters, so there needs to be a very good reason for this to be added to pybind11.

Meanwhile, if you really want this, why don't you construct your `py::module` like this?

```C++
PyObject *module_ptr = PyModule_Create(def);
auto m = reinterpret_steal<py::module>(module_ptr);
```

sunyj:
> Could you please expand your reasoning? pybind11 tries to abstract the C API away from you and pybind11's API never takes C API parameters, so there needs to be a very good reason for this to be added to pybind11.
> 
> Meanwhile, if you really want this, why don't you construct your `py::module` like this?
> 
> ```c++
> PyObject *module_ptr = PyModule_Create(def);
> auto m = reinterpret_steal<py::module>(module_ptr);
> ```

In fact, this is exactly what I needed.  Thanks.

PR #2366, opened 2020-08-05
Adding tests specifically to exercise pybind11::str::raw_str.

rwgk:
Initially this was meant to be a PR preparing for follow-up PR #2367. The discussion on the second PR exposed behavior changes I was unaware of initially. I will drop #2367 for now because it's non-essential at the moment, to not distract further from the far more important `pybind11::str` change.

### DIFF
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -220,6 +220,38 @@ def test_constructors():
         assert noconv2[k] is expected[k]
 
 
+def test_pybind11_str_raw_str():
+    # specifically to exercise pybind11::str::raw_str
+    cvt = m.convert_to_pybind11_str
+    assert cvt(u"Str") == u"Str"
+    assert cvt(b'Bytes') == u"Bytes" if str is bytes else "b'Bytes'"
+    assert cvt(None) == u"None"
+    assert cvt(False) == u"False"
+    assert cvt(True) == u"True"
+    assert cvt(42) == u"42"
+    assert cvt(2**65) == u"36893488147419103232"
+    assert cvt(-1.50) == u"-1.5"
+    assert cvt(()) == u"()"
+    assert cvt((18,)) == u"(18,)"
+    assert cvt([]) == u"[]"
+    assert cvt([28]) == u"[28]"
+    assert cvt({}) == u"{}"
+    assert cvt({3: 4}) == u"{3: 4}"
+    assert cvt(set()) == u"set([])" if str is bytes else "set()"
+    assert cvt({3, 3}) == u"set([3])" if str is bytes else "{3}"
+
+    valid_orig = u"Ǳ"
+    valid_utf8 = valid_orig.encode("utf-8")
+    valid_cvt = cvt(valid_utf8)
+    assert type(valid_cvt) == bytes  # Probably surprising.
+    assert valid_cvt == b'\xc7\xb1'
+
+    malformed_utf8 = b'\x80'
+    malformed_cvt = cvt(malformed_utf8)
+    assert type(malformed_cvt) == bytes  # Probably surprising.
+    assert malformed_cvt == b'\x80'
+
+
 def test_implicit_casting():
     """Tests implicit casting when assigning or appending to dicts and lists."""
     z = m.get_implicit_casting()
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -237,6 +237,8 @@ TEST_SUBMODULE(pytypes, m) {
         );
     });
 
+    m.def("convert_to_pybind11_str", [](py::object o) { return py::str(o); });
+
     m.def("get_implicit_casting", []() {
         py::dict d;
         d["char*_i1"] = "abc";


tests/test_pytypes.py
```diff
@@ -220,6 +220,27 @@ def test_constructors():
         assert noconv2[k] is expected[k]
 
 
+def test_pybind11_str_raw_str():
+    # specifically to exercise pybind11::str::raw_str
+    cvt = m.convert_to_pybind11_str
+    assert cvt(u"Str") == u"Str"
+    assert cvt(b"Bytes") == u"Bytes" if str is bytes else "b'Bytes'"
```
YannickJadoul, line 227: Should we maybe use `sys.version_info.major < 3` instead of `str is bytes`?

Or start using some kind of compatibility library (cfr. https://python-future.org/compatible_idioms.html#strings-and-bytes, or `six`)? Another thing that could make things simpler/more Python 3: `from __future__ import unicode_literals` would allow us to remove all `u` prefixes.

tests/test_pytypes.py
```diff
@@ -220,6 +220,27 @@ def test_constructors():
         assert noconv2[k] is expected[k]
 
 
+def test_pybind11_str_raw_str():
+    # specifically to exercise pybind11::str::raw_str
+    cvt = m.convert_to_pybind11_str
+    assert cvt(u"Str") == u"Str"
```
YannickJadoul, line 226: This could also be:

```python
pybind11_str = unicode if sys.version_info.major < 3 else str
for test in [u"Str",
             None,
             False,
             True,
             ...]:
    assert cvt(test) == pybind11_str(test)
```

Or do we need to be explicit here, line by line?

tests/test_pytypes.py
```diff
@@ -220,6 +220,27 @@ def test_constructors():
...
+    assert cvt(2**65) == u"36893488147419103232"
+    assert cvt(-1.50) == u"-1.5"
+    assert cvt(()) == u"()"
+    assert cvt((18,)) == u"(18,)"
+    assert cvt([]) == u"[]"
+    assert cvt([28]) == u"[28]"
+    assert cvt({}) == u"{}"
+    assert cvt({3: 4}) == u"{3: 4}"
+    assert cvt(set()) == u"set([])" if str is bytes else "set()"
+    assert cvt({3, 3}) == u"set([3])" if str is bytes else "{3}"
```
YannickJadoul, line 241: Another one, sorry. There're no funky unicode characters/string/weirdly encoded strings as bytes here!

YannickJadoul:
> rwgk wants to merge 1 commit into `pybind:master` from `rwgk:test_pybind11_str_raw_str`

Is this correct, or do you want to merge into `str-bytes-cleanup`?

rwgk:
> > rwgk wants to merge 1 commit into `pybind:master` from `rwgk:test_pybind11_str_raw_str`
> 
> Is this correct, or do you want to merge into `str-bytes-cleanup`?

It's intentional: no behavior change, just hardening tests, and a very-small-scale internal cleanup in the companion PR. This will help keep the branch simpler for Wenzel to review when he's back. I actually hope to extract more of the new pure test code into master ahead of time. That way Wenzel can easily focus on what's actually changing.

The whole point of bringing in a team of maintainer is to take load off Wenzel's shoulders.

Thanks for all the suggestions. I think I will take them all (hopefully later today), except in the context of str-vs-bytes, the `str is bytes` condition makes things super obvious, as opposed to looking at `sys.version_info`; that latter takes an extra mental leap to realize "ah, str is bytes".

YannickJadoul:
> It's intentional: no behavior change,

Ah, OK; my thoughts behind this were that it would be nice to keep an overview of all the things that changed to the str-bytes handling. Just as well for @wjakob (OK, yes, this is perfectly fine to not pass his sight, agreed) as for us to check "are all the small tests and changes added now consistent, altogether?".

> except in the context of str-vs-bytes, the `str is bytes` condition makes things super obvious, as opposed to looking at `sys.version_info`; that latter takes an extra mental leap to realize "ah, str is bytes".

Ah, OK. Then we have a different mental model :-D For me it's exactly the opposite way around. It's the `str is bytes` that always take me an extra mental leap: "Ah, yes, this is ... Python 2, so `str` and `unicode` rather than `bytes` and `str`". But either way works, I suppose. Not going to bikeshed this, then ;-)

tests/test_pytypes.py
YannickJadoul: And while you're add it, I don't know how much work it is, but some checks on types with custom `__str__`/`__unicode__` methods might be interesting?

tests/test_pytypes.py
EricCousineau-TRI: There were other parts of test code that explicitly check for 2 vs. 3, I believe. Let me find it.

EricCousineau-TRI: Examples:
https://github.com/pybind/pybind11/blob/0af7fe6c1943e6a9043e4e01c4bc9059108a6c98/tests/test_builtin_casters.py#L119
https://github.com/pybind/pybind11/blob/0af7fe6c1943e6a9043e4e01c4bc9059108a6c98/tests/test_buffers.py#L11

tests/test_pytypes.py
EricCousineau-TRI: I actually prefer it being explicit, just to make it as clear as possible, and permit specific branching.

tests/test_pytypes.py
EricCousineau-TRI: FWIW I've filed #2376

tests/test_pytypes.py
YannickJadoul: Also fine. Let's leave it up to @rwgk, then :-)

rwgk:
Thanks for the reviews!

I made one very minor change compared to the reviewed code, pulled up from downstream changes in my fork, so that the follow-up reviews for Wenzel are simpler.
I will merge this now.

Note in the meantime I refactored or decided to drop all the other PRs I pushed up, based on all the feedback I received. I will clean that up now with explanations (I have a final end state for reviews already in my fork). The main driving force is to prepare a PR changing `pybind::str` that is as easy as possible to review by Wenzel.

PR #2368, opened 2020-08-06
fix: some build-system cleanup

henryiii:
Taking the unrelated changes from #2353 to see if that can go in stand alone.

* Some cleanup of CMakeLists
* Make all target names unique
* Include `.cmake.in` files when formatting
* Update CI to do more testing (shorten Configure to Config, too)
* `PYBIND11_CPP_STANDARD` was not being handled correctly, now it does what I claimed it did before


tests/test_cmake_build/subdirectory_target/CMakeLists.txt
```diff
@@ -1,22 +1,32 @@
 cmake_minimum_required(VERSION 3.7)
+
+# VERSION 3.7...3.18, but some versions of VS have a patched CMake 3.11
+# that do not work properly with this syntax, so using the following workaround:
+if(${CMAKE_VERSION} VERSION_LESS 3.18)
```
rwgk, line 5: IIUC the comment suggests this is a workaround for VS, but the `if` condition in platform agnostic. Would it make sense to add `and platform_is_windows`?

henryiii: The problem is that there's a nice syntax for this:

```cmake
cmake_minimum_required(VERSION 3.7...3.18)
```

That was _allowed_ on all official versions of CMake, and works on CMake 3.12+ (which is when they added it).
But MSVC ships with a custom fork of CMake, and they fiddled with the versioning - and in the process, broke this, since they didn't know what CMake planned to add. IIRC, 15.6 I think is the one where it was fixed.

So I'm just doing this without the nice syntax, but noting that this is really just the equivalent of that range. Once we drop CMake 3.11 and below, we can switch to the range syntax.

rwgk: > The problem is that there's a nice syntax for this:
> 
> ```cmake
> cmake_minimum_required(VERSION 3.7...3.18)
> ```
> 
> That was _allowed_ on all official versions of CMake, and works on CMake 3.12+ (which is when they added it).
> But MSVC ships with a custom fork of CMake, and they fiddled with the versioning - and in the process, broke this, since they didn't know what CMake planned to add. IIRC, 15.6 I think is the one where it was fixed.
> 
> So I'm just doing this without the nice syntax, but noting that this is really just the equivalent of that range. Once we drop CMake 3.11 and below, we can switch to the range syntax.

Ah, got it, thanks!

Minor suggestion for modifying the comment:

# The `cmake_minimum_required(VERSION 3.7...3.18)` syntax does not work with some versions of VS that have a patched CMake 3.11. This forces us to emulate the behavior using the following workaround:

henryiii: I like that better, thanks!

PR #2371, opened 2020-08-07
[RFC] Render typed iterators in docstrings (alternative)

sizmailov:
This is an alternative to #2244 without backward incompatibility drawback. 

Interestingly, binary size is not affected by this change 
```
# g++ (Ubuntu 8.3.0-6ubuntu1) 8.3.0
du -b pybind11_tests.cpython-37m-x86_64-linux-gnu.so
3244544
```

Examples (`compare.cpp`): 

| master |
| ------------- |
| <pre>`__iter__(self: example.VectorInt) -> Iterator`<br>`__iter__(self: example.MapTupleToComplex) -> Iterator`<br>`items(self: example.MapTupleToComplex) -> Iterator`</pre> | 

| this PR |
| ------------- |
| <pre>`__iter__(self: example.VectorInt) -> Iterator[int]`<br>`__iter__(self: example.MapTupleToComplex) -> Iterator[Tuple[int, str]]`<br>`items(self: example.MapTupleToComplex) -> Iterator[Tuple[Tuple[int, str], complex]]`</pre> |

```C++
#include <map>
#include <complex>

#include "pybind11/embed.h" 

namespace py = pybind11;

using map_tuple_to_complex = std::map<std::tuple<int,std::string>, std::complex<double>>;
PYBIND11_MAKE_OPAQUE(std::vector<int>);
PYBIND11_MAKE_OPAQUE(map_tuple_to_complex);

  py::bind_map<map_tuple_to_complex>(m, "MapTupleToComplex");
}

int main() {
  py::exec(R"(
      from example import *
      print(VectorInt.__iter__.__doc__, end="")
      print(MapTupleToComplex.__iter__.__doc__, end="")
      print(MapTupleToComplex.items.__doc__, end="")
)");
}
```
</details>

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1712,6 +1712,22 @@ struct iterator_state {
     bool first_or_done;
 };
 
+template<typename Iterator, typename Sentinel, return_value_policy Policy>
+struct type_caster<iterator_state<Iterator, Sentinel, false, Policy>> :
+    type_caster_base<iterator_state<Iterator, Sentinel, false, Policy>> {
+    using ValueType = decltype(*std::declval<Iterator>());
+public:
+    static constexpr auto name = _("Iterator[") + make_caster<ValueType>::name + _("]");
+};
+
+template<typename Iterator, typename Sentinel, return_value_policy Policy>
+struct type_caster<iterator_state<Iterator, Sentinel, true, Policy>> :
+    type_caster_base<iterator_state<Iterator, Sentinel, true, Policy>> {
+    using ValueType = decltype((*std::declval<Iterator>()).first);
+public:
+    static constexpr auto name = _("Iterator[") + make_caster<ValueType>::name + _("]");
+};
+
 PYBIND11_NAMESPACE_END(detail)
 
 /// Makes a python iterator from a first and past-the-end C++ InputIterator.
@@ -1720,7 +1736,8 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
           typename Sentinel,
           typename ValueType = decltype(*std::declval<Iterator>()),
           typename... Extra>
-iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
+detail::iterator_state<Iterator, Sentinel, false, Policy>
+    make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
     typedef detail::iterator_state<Iterator, Sentinel, false, Policy> state;
 
     if (!detail::get_type_info(typeid(state), false)) {
@@ -1738,8 +1755,7 @@ iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
                 return *s.it;
             }, std::forward<Extra>(extra)..., Policy);
     }
-
-    return cast(state{first, last, true});
+    return state{first, last, true};
 }
 
 /// Makes an python iterator over the keys (`.first`) of a iterator over pairs from a
@@ -1749,7 +1765,8 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
           typename Sentinel,
           typename KeyType = decltype((*std::declval<Iterator>()).first),
           typename... Extra>
-iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
+detail::iterator_state<Iterator, Sentinel, true, Policy>
+    make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
     typedef detail::iterator_state<Iterator, Sentinel, true, Policy> state;
 
     if (!detail::get_type_info(typeid(state), false)) {
@@ -1768,20 +1785,24 @@ iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
             }, std::forward<Extra>(extra)..., Policy);
     }
 
-    return cast(state{first, last, true});
+    return state{first, last, true};
 }
 
 /// Makes an iterator over values of an stl container or other container supporting
 /// `std::begin()`/`std::end()`
 template <return_value_policy Policy = return_value_policy::reference_internal,
-          typename Type, typename... Extra> iterator make_iterator(Type &value, Extra&&... extra) {
+          typename Type, typename... Extra>
+detail::iterator_state<decltype(std::begin(std::declval<Type&>())), decltype(std::end(std::declval<Type&>())), false, Policy>
+make_iterator(Type &value, Extra&&... extra) {
     return make_iterator<Policy>(std::begin(value), std::end(value), extra...);
 }
 
 /// Makes an iterator over the keys (`.first`) of a stl map-like container supporting
 /// `std::begin()`/`std::end()`
 template <return_value_policy Policy = return_value_policy::reference_internal,
-          typename Type, typename... Extra> iterator make_key_iterator(Type &value, Extra&&... extra) {
+          typename Type, typename... Extra>
+detail::iterator_state<decltype(std::begin(std::declval<Type&>())), decltype(std::end(std::declval<Type&>())), true, Policy>
+make_key_iterator(Type &value, Extra&&... extra) {
     return make_key_iterator<Policy>(std::begin(value), std::end(value), extra...);
 }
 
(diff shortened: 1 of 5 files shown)


include/pybind11/pybind11.h
```diff
@@ -1720,7 +1736,8 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
           typename Sentinel,
           typename ValueType = decltype(*std::declval<Iterator>()),
           typename... Extra>
-iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
+detail::iterator_state<Iterator, Sentinel, false, Policy>
+[[deprecated]] make_iterator_ng(Iterator first, Sentinel last, Extra &&... extra) {
```
sizmailov, line 1740: I'm not 100% happy about deprecation. Probably the annotations are unnecessary and should be removed.

include/pybind11/pybind11.h
```diff
@@ -1768,23 +1785,62 @@ iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
...
+}
+
+/// Makes a python iterator from a first and past-the-end C++ InputIterator.
+template <return_value_policy Policy = return_value_policy::reference_internal,
+    typename Iterator,
+    typename Sentinel,
+    typename ValueType = decltype(*std::declval<Iterator>()),
+    typename... Extra>
+[[deprecated("Superseded by make_iterator_ng")]]
+auto make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
```
bstaletic, line 1798: Deduced return type is available since C++14. You'll have to use `auto make_iterator(...) -> decltype(function body) {}`

include/pybind11/pybind11.h
```diff
@@ -1768,23 +1785,64 @@ iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
...
+    return state{first, last, true};
+}
+
+/// Makes a python iterator from a first and past-the-end C++ InputIterator.
+template <return_value_policy Policy = return_value_policy::reference_internal,
+    typename Iterator,
+    typename Sentinel,
+    typename ValueType = decltype(*std::declval<Iterator>()),
+    typename... Extra>
+[[deprecated("Superseded by make_iterator_ng")]]
```
sizmailov, line 1797: Deprecation of long-standing function is a hostile action to users, since it would lead to warnings in user code, which would force everyone to use _ng version of functions.

This can be viewed as the cost of making the change fully backward-compatible. 
Having two similar functions is not a good thing, so one should be marked as deprecated in favor of superior one. 

I'm not 100% happy about this. Is there other viable option?

bstaletic, line 1797: The `[[deprecated]]` attribute is C++14. Use `PYBIND11_DEPRECATED` macro instead.

EricCousineau-TRI: I think if I were to consider this (using a suffix and deprecate the old) vs. #2244 (updating the function), my money would be still be on #2244 :(

In terms of incompatibility, I think that could be solved by using a (deprecated?) implicit conversion from `py::detail::iterator_state<...>` to `py::iterator`.

However, per @wjakob's points about complexity, I'm not sure if this PR avoids the complexity from before, due to the deprecation (which may lead to another name swap + deprecation from `make_iterator_ng` back to `make_iterator`?).

The best option (that I can think of) is to mirror the setup used in `numpy.h` for `py::array` and `py::array_t`:

Let `iterator` be the fully type-erased Python interface type, while `iterator_t<ValueType>` can add a (minimal) additional "C++ type-friendly" filter on top of `iterator` (e.g. the casting name, forced casting, etc.). If you do that, then perhaps you only need `iterator_t<ValueType>`, rather than carrying around type info in `detail::iterator_state<Iterator, Sentinel, bool, Policy>` and use all of the `decltype(...)` stuff? Then you can update the `stl.h` stuff to use `iterator_t<ValueType>` vs. just `iterator` or propogating types from there.
It may induce some additional non-functional indirection, but requires less type change?

Thoughts? @bstaletic or @wjakob if y'all are available?

bstaletic: > I think if I were to consider this (using a suffix and deprecate the old) vs. #2244 (updating the function), my money would be still be on #2244 :(
> 
> In terms of incompatibility, I think that could be solved by using a (deprecated?) implicit conversion from py::detail::iterator_state<...> to py::iterator.
> 
> However, per @ wjakob's points about complexity, I'm not sure if this PR avoids the complexity from before, due to the deprecation (which may lead to another name swap + deprecation from make_iterator_ng back to make_iterator?).

I agree with you up to this point.

> The best option [...]

Your idea here doesn't sound bad to me, but I believe @wjakob should have the last word.

YannickJadoul: I also quite like the idea of fixing the actual problem, rather than adding onto the confusion. Then again, I can see the point that this "only" fixes the docstring (and you could already use `py::arg_v` to do so I believe?).

At any rate, if we're thinking about making a breaking 3.0.0 release, it would be cool if we could include a fixed version. A potential route would be to first use a new name + deprecation, then rename in a (distant future) pybind11 3.0.0?

YannickJadoul:
> Interestingly, binary size is not affected by this change

How many instantiations of `py::iterator` are in this test? I would assume the binary size to go up if many different iterator types are instantiated?

include/pybind11/pybind11.h
sizmailov: > Then again, I can see the point that this "only" fixes the docstring

This PR only fixes annotation of return type. Functions with typed iterator argument still need to use type-erased `py::iterator`. I don't think `py::arg_v` can help here.

I like the idea of `py::iterator_t<ValueType>`. The only issue I think of is overloads support. Dispatcher needs to tell the difference between `f(py::iterator_t<int>)` vs `f(py::iterator_t<bool>)`. I think it's not possible to tell which is which until the actual iteration happens, unlike in case of `py::array_t`. This issue can be "avoided" by banning overloads that different only in `py::iterator_t` argument(s). Alternatively, we can make a scaffold to extract first element (_which still might be not enough_) to dispatch and then iterate using mocked-up iterator, but this sounds like an implementation nightmare and potential nasty debug sessions for users.

sizmailov:
> How many instantiations of py::iterator are in this test? I would assume the binary size to go up if many different iterator types are instantiated?

Can't tell from release binary. Everything related to pybnd11 seems to be optimized completely :confused: 

```bash
$ readelf -Ws pybind11_tests.cpython-37m-x86_64-linux-gnu.so | c++filt | grep pybind
   387: 00000000001f2170   153 FUNC    GLOBAL DEFAULT   12 std::_Function_base::_Base_manager<pybind11::detail::type_caster<std::function<void ()>, void>::load(pybind11::handle, bool)::func_wrapper>::_M_manager(std::_Any_data&, std::_Any_data const&, std::_Manager_operation)
   437: 000000000027a500    38 OBJECT  GLOBAL DEFAULT   14 typeinfo name for pybind11::detail::broadcast_trivial
   445: 00000000002485b0   153 FUNC    GLOBAL DEFAULT   12 std::_Function_base::_Base_manager<pybind11::detail::type_caster<std::function<int (int)>, void>::load(pybind11::handle, bool)::func_wrapper>::_M_manager(std::_Any_data&, std::_Any_data const&, std::_Manager_operation)
   456: 0000000000248470   153 FUNC    GLOBAL DEFAULT   12 std::_Function_base::_Base_manager<pybind11::detail::type_caster<std::function<void (int)>, void>::load(pybind11::handle, bool)::func_wrapper>::_M_manager(std::_Any_data&, std::_Any_data const&, std::_Manager_operation)
   457: 0000000000316800    16 OBJECT  GLOBAL DEFAULT   20 typeinfo for pybind11::detail::broadcast_trivial
   467: 0000000000260ac0   343 FUNC    GLOBAL DEFAULT   12 PyInit_pybind11_tests
```

include/pybind11/pybind11.h
YannickJadoul: > This PR only fixes annotation of return type. Functions with typed iterator argument still need to use type-erased `py::iterator`. I don't think `py::arg_v` can help here.

Right, jumped to conclusions too quickly, there! We don't actually have a `py::ret_v`, I suppose, to support a manual/custom return type annotation?

> I like the idea of `py::iterator_t<ValueType>`. The only issue I think of is overloads support.

I thought we already have something like this (where the internal types are only checked at runtime), but I can't find it. This will always be a problem, though, so I wouldn't be opposed to just add it and document that it won't do anything for overloads (but it would do the casting for you, once you ask the next element). pybind11 has more corners where overloads aren't perfect and order matters, so ...
Extracting the first element would be reasonably unexpected, I believe. And as you say: it wouldn't provide any guarantees, so let's have users do that themselves, if they really want to disambiguate.

PR #2376, opened 2020-08-09
tests: Consolidate version (2 vs. 3) and platform (CPython vs. PyPy) checks

EricCousineau-TRI:
Minor updates that may help discussion in #2366

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -205,7 +205,19 @@ def pytest_configure():
         from pybind11_tests.eigen import have_eigen
     except ImportError:
         have_eigen = False
-    pypy = platform.python_implementation() == "PyPy"
+
+    # Provide simple `six`-like aliases.
+    pytest.PY2 = (sys.version_info.major == 2)
+    pytest.PY3 = (sys.version_info.major == 3)
+    pytest.CPYTHON = (platform.python_implementation() == "CPython")
+    pytest.PYPY = (platform.python_implementation() == "PyPy")
+
+    if not pytest.PY2 and not pytest.PY3:
+        print("pybind11 is only supported on Python 2.x or 3.x")
+        sys.exit(1)
+    if not pytest.CPYTHON and not pytest.PYPY:
+        print("pybind11 is only supported for CPython or PyPy")
+        sys.exit(1)
 
     skipif = pytest.mark.skipif
     pytest.suppress = suppress
@@ -215,13 +227,13 @@ def pytest_configure():
                                              reason="eigen and/or numpy are not installed")
     pytest.requires_eigen_and_scipy = skipif(
         not have_eigen or not scipy, reason="eigen and/or scipy are not installed")
-    pytest.unsupported_on_pypy = skipif(pypy, reason="unsupported on PyPy")
-    pytest.bug_in_pypy = pytest.mark.xfail(pypy, reason="bug in PyPy")
-    pytest.unsupported_on_pypy3 = skipif(pypy and sys.version_info.major >= 3,
+    pytest.unsupported_on_pypy = skipif(pytest.PYPY, reason="unsupported on PyPy")
+    pytest.bug_in_pypy = pytest.mark.xfail(pytest.PYPY, reason="bug in PyPy")
+    pytest.unsupported_on_pypy3 = skipif(pytest.PYPY and pytest.PY3,
                                          reason="unsupported on PyPy3")
-    pytest.unsupported_on_pypy_lt_6 = skipif(pypy and sys.pypy_version_info[0] < 6,
+    pytest.unsupported_on_pypy_lt_6 = skipif(pytest.PYPY and sys.pypy_version_info[0] < 6,
                                              reason="unsupported on PyPy<6")
-    pytest.unsupported_on_py2 = skipif(sys.version_info.major < 3,
+    pytest.unsupported_on_py2 = skipif(pytest.PY2,
                                        reason="unsupported on Python 2.x")
     pytest.gc_collect = gc_collect
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -208,16 +208,8 @@ def pytest_configure():
 
     # Provide simple `six`-like aliases.
     pytest.PY2 = (sys.version_info.major == 2)
-    pytest.PY3 = (sys.version_info.major == 3)
     pytest.CPYTHON = (platform.python_implementation() == "CPython")
     pytest.PYPY = (platform.python_implementation() == "PyPy")
-
-    if not pytest.PY2 and not pytest.PY3:
-        print("pybind11 is only supported on Python 2.x or 3.x")
-        sys.exit(1)
-    if not pytest.CPYTHON and not pytest.PYPY:
-        print("pybind11 is only supported for CPython or PyPy")
-        sys.exit(1)
 
     skipif = pytest.mark.skipif
     pytest.suppress = suppress
@@ -229,7 +221,7 @@ def pytest_configure():
         not have_eigen or not scipy, reason="eigen and/or scipy are not installed")
     pytest.unsupported_on_pypy = skipif(pytest.PYPY, reason="unsupported on PyPy")
     pytest.bug_in_pypy = pytest.mark.xfail(pytest.PYPY, reason="bug in PyPy")
-    pytest.unsupported_on_pypy3 = skipif(pytest.PYPY and pytest.PY3,
+    pytest.unsupported_on_pypy3 = skipif(pytest.PYPY and not pytest.PY2,
                                          reason="unsupported on PyPy3")
     pytest.unsupported_on_pypy_lt_6 = skipif(pytest.PYPY and sys.pypy_version_info[0] < 6,
                                              reason="unsupported on PyPy<6")
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -115,12 +115,19 @@ def test_bytes_to_string():
     """Tests the ability to pass bytes to C++ string-accepting functions.  Note that this is
     one-way: the only way to return bytes to Python is via the pybind11::bytes class."""
     # Issue #816
-    bytes_ = bytes if pytest.PY2 else str
 
-    assert m.strlen(bytes_("hi")) == 2
-    assert m.string_length(bytes_("world")) == 5
-    assert m.string_length(bytes_("a\x00b")) == 3
-    assert m.strlen(bytes_("a\x00b")) == 1  # C-string limitation
+    def to_bytes(s):
+        if pytest.PY2:
+            b = s
+        else:
+            b = s.encode("utf8")
+        assert isinstance(b, bytes)
+        return b
+
+    assert m.strlen(to_bytes("hi")) == 2
+    assert m.string_length(to_bytes("world")) == 5
+    assert m.string_length(to_bytes("a\x00b")) == 3
+    assert m.strlen(to_bytes("a\x00b")) == 1  # C-string limitation
 
     # passing in a utf8 encoded string should work
     assert m.string_length(u'💩'.encode("utf8")) == 4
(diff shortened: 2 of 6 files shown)


EricCousineau-TRI:
FWIW I dunno how much I love injecting stuff into `pytest`, but :shrug:, it at least sticks with convention.
Perhaps at some point it would make things more obvious to consolidate these into an explicit utility module, like `pybind11_test_utilities` or something.

tests/conftest.py
```diff
@@ -205,7 +205,19 @@ def pytest_configure():
         from pybind11_tests.eigen import have_eigen
     except ImportError:
         have_eigen = False
-    pypy = platform.python_implementation() == "PyPy"
+
+    # Provide simple `six`-like aliases.
+    pytest.PY2 = (sys.version_info.major == 2)
+    pytest.PY3 = (sys.version_info.major == 3)
+    pytest.CPYTHON = (platform.python_implementation() == "CPython")
+    pytest.PYPY = (platform.python_implementation() == "PyPy")
```
bstaletic, line 213: Should we really be putting stuff in someone else's module?

tests/conftest.py
```diff
@@ -205,7 +205,19 @@ def pytest_configure():
...
+
+    if not pytest.PY2 and not pytest.PY3:
+        print("pybind11 is only supported on Python 2.x or 3.x")
```
bstaletic, line 216: Maybe just check `sys.version_info.major >= 2`? I'm not really sure how necessary this check is.

YannickJadoul: This might indeed break if they somehow decide Python version numbers are decimal and will jump from 3.9 to 4.0?

tests/conftest.py
YannickJadoul: Same here. I agree with @bstaletic, that this should probably not be part of `pytest`. `PY2`, etc, in full-caps, should be clear enough that no one uses the same name? It's already better than using lower-case `pypy`, anyway.

YannickJadoul: > it at least sticks with convention.

Oh, is this a convention?

henryiii: I think @EricCousineau-TRI means the six convention. I don't think injecting things into pytest is convention; and if it is, it's a bad convention. One reason conftest runs before but in the same namespace as tests is to allow you _not_ to have to communicate through pytest. If you want to replicate the six convention, I would do this:

```python
class six:
    PY2 = sys.version_info.major == 2
```
I would actually _not_ add PY3, as that would force the correct usage of PY2 - the things currently checking against `PY3` should actually be checking `not PY2`. Python 4 won't change the things we are checking PY3 for.

tests/conftest.py
henryiii: 3.10 is next, but 4.0 won't be such a large change as 3.0.

Please remove this "fail fast" check - if someone tries Python 4.0. they should see what is going wrong so they can fix it, not have the tests fail to run. I think you should remove the CPython/PyPy check too - if someone tries to run the test, they should see what goes wrong. It should fail to compile, generally, so they shouldn't get here.

I generally don't like fail-fast checks (I spent hours trying to debug an error that was hidden behind a fail-fast check last week), and you should never fail-fast for a version of something that doesn't exist yet.

tests/conftest.py
YannickJadoul: If we're going for `six` as a name anyway, why not just add it to `requirements.txt` and depend on it?

YannickJadoul: > I would actually _not_ add PY3, as that would force the correct usage of PY2 - the things currently checking against `PY3` should actually be checking `not PY2`. Python 4 won't change the things we are checking PY3 for.

Good point as well, btw! :+1:

henryiii: We could do that, but unlike a normal package, we don't have an "install" to do the testing, so this would mean everyone running the tests would have to install six. It's [pretty common](https://github.com/tomerfiliba/plumbum/blob/6788d15b017d7d6a3cf1834c9b098db2719762a9/plumbum/lib.py#L34-L94) to implement just the bit of six you need - and for pybind11, this seems like the simplest way to do this (since we really only require a single dependency, pytest, currently to run the tests).

rwgk: A agree, making `six` a requirement causes more harm (lost time) than good, especially this late in the game.

To @YannickJadoul and @bstaletic 's concerns: I feel sticking with the precedent (`pytest.pypy`) is the right thing here. Going through the trouble establishing a separate module for this is again more trouble than it is worth it IMO. I think it's important to be able to make meaningful small steps in the right direction, without tearing open a whole new construction site every time.

I really like @YannickJadoul 's suggestion to only introduce `PY2` but NOT `PY3`. Hopefully we'll never have to go through something like the 2-to-3 transition again.

@YannickJadoul : if `pytest.PY2` is centrally available I definitely prefer that over my `str is bytes` test. (My thinking was: if there is nothing central, I use a test that it most directly related to what I'm dealing with, but I think the central variable is best.)

YannickJadoul: Great, let's get this in, then!

I'm still not a huge fan of adding to pybind11 (I do think adding a separate module and importing it could be cleaner; and it should be painless, just a few constants in a separate file), but it does work, so it's not a crucial point for me.

> I really like @YannickJadoul 's suggestion to only introduce `PY2` but NOT `PY3`. Hopefully we'll never have to go through something like the 2-to-3 transition again.

@henryiii's idea, btw ;-)

tests/test_builtin_casters.py
```diff
@@ -115,13 +115,12 @@ def test_bytes_to_string():
     """Tests the ability to pass bytes to C++ string-accepting functions.  Note that this is
     one-way: the only way to return bytes to Python is via the pybind11::bytes class."""
     # Issue #816
-    import sys
-    byte = bytes if sys.version_info[0] < 3 else str
+    bytes_ = bytes if pytest.PY2 else str
```
henryiii, line 118: Are you sure this is needed? See: https://six.readthedocs.io/#six.binary_type

> Python 2.6 and 2.7 include bytes as a builtin alias of str, so six’s version is only necessary for Python 2.5 compatibility.

tests/conftest.py
virtuald: It would be more appropriate to use a [pytest fixture](https://docs.pytest.org/en/stable/fixture.html) for this instead of modifying the namespace?

```python
# conftest.py

@pytest.fixture
def compat():
  return SomeObject()

```

```python

def test_a_thing(compat):
   if compat.PY2:
      # do something
```

You could also add boolean PY2/etc fixtures:

```
@pytest.fixture
def PY2():
   return True if whatever else False

def test_a_thing(PY2):
   if PY2:
      blah

File: tests/conftest.py [hunk 4]
```diff
@@ -215,13 +227,13 @@ def pytest_configure():
                                              reason="eigen and/or numpy are not installed")
     pytest.requires_eigen_and_scipy = skipif(
         not have_eigen or not scipy, reason="eigen and/or scipy are not installed")
-    pytest.unsupported_on_pypy = skipif(pypy, reason="unsupported on PyPy")
-    pytest.bug_in_pypy = pytest.mark.xfail(pypy, reason="bug in PyPy")
-    pytest.unsupported_on_pypy3 = skipif(pypy and sys.version_info.major >= 3,
+    pytest.unsupported_on_pypy = skipif(pytest.PYPY, reason="unsupported on PyPy")
+    pytest.bug_in_pypy = pytest.mark.xfail(pytest.PYPY, reason="bug in PyPy")
+    pytest.unsupported_on_pypy3 = skipif(pytest.PYPY and pytest.PY3,
```
henryiii, line 232: These probably should be a custom markers, not injected into the pytest namespace. https://docs.pytest.org/en/stable/example/markers.html#marking-platform-specific-tests-with-pytest

tests/conftest.py
henryiii: Fully agree with @virtuald (though markers on tests would be even better in many cases). PyTest uses fixtures and markers, not injection into the pytest namespace. If "pytest.PY3" was ever added, this might break.

YannickJadoul: Yes, but to be honest, custom markers/fixtures seem to be too complex/elaborate for what this is (currently) doing, no? If the alternatives are between `pytest` and custom markers, I think I'd (currently) prefer the first one.
But even better, I'd just go for some auxiliary module and `from auxiliary_module import *`? I don't think there's any reason to use the `pytest_configure` hook for this, is there?

YannickJadoul:
Any further suggestions how to continue with this? It would be nice to get this in so new PRs can use this.

henryiii:
Since we already inject stuff into pytest with `pytest_configure`, I'd be okay to put this in for now, then do a proper cleanup later. I bet that's what @EricCousineau-TRI meant by convention. There are several bad practices here that need to be fixed, but that's not the focus of this PR. The only change I think we should require now is avoiding PY3 - I can add that, though.

henryiii:
I've pushed the PY2 only change, and dropped the fail fasts. @EricCousineau-TRI, does this look fine? I'll try to refactor later for pytest best practices, but it shouldn't hold up this PR.

henryiii:
Sorry, had to rebase & force-push, another check that needs to be converted to PY2 was added, and it was detected by flake8.

tests/conftest.py
EricCousineau-TRI: I'm OK with removing the fail-fast, but I dunno how I feel about there being no warning at all. Maybe I'll try to add a warning in after this merges.

I guess I can feel your pain about the fail-fast, but I do think fail-fasts can help in certain contexts.

tests/conftest.py
EricCousineau-TRI: Nah, I think the injection into `pytest` is weird, but my vote is we keep things relatively simple in this PR, then go the auxiliary module route (`from pybind_test_helpers` or whatevs) in the future.

YannickJadoul:
Great, thanks, @henryiii! As far as I'm concerned, this can go in soon enough, then?

EricCousineau-TRI:
> @EricCousineau-TRI, does this look fine?

Yup! Just going through to address the open discussions.

tests/test_builtin_casters.py
EricCousineau-TRI: Lemme check. FWIW I just converted existing code, didn't think too deeply about it.

tests/conftest.py
henryiii: If it fails, the warning will be the actual failure. And if it works, hooray!

The one I almost didn't remove was the CPython + PyPy one, but I have no idea how this would even reach the Python stage, it shouldn't compile.

tests/test_builtin_casters.py
EricCousineau-TRI: Done. The logic in the function seemed wrong - it didn't really test `bytes` in Python 3.

tests/conftest.py
EricCousineau-TRI: Er, yeah, but failing fast in contract-checking can help, e.g.
"Hey, I'm a function that will gain absolutely zero from duck typing, and I want a specific type... but I dont' have `assert isinstance(...)`. Now *you* get to figure out why this random field or dictionary string is missing!"

So yeah, that's what I mean about "in certain contexts". For build systems, warnings are perhaps best, but some users completely ignore warnings too...

EricCousineau-TRI:
K, pending CI, this should be good to go!

tests/test_builtin_casters.py
YannickJadoul: Pinging @rwgk, since it's a test involving str/bytes ;-) This can just be merged, though; looks OK to me.

EricCousineau-TRI:
Mac CI going slow :( Looks like there're only 2 nodes available for provisioning... :crying_cat_face:

YannickJadoul:
Hurray! :-)

PR #2377, opened 2020-08-10
docs: contrib/issue templates

henryiii:
This sets up some GitHub specifics to make the main pages a little cleaner, add nice issue templates and a prominent link to the gitter, and adds information requested by @rwgk about setting up development environments and building with CMake.

A second PR can be made to move `CONTRIBUTING.md ` to `.github/CONTRIBUTING.md `, to save a little space on the main page. We can add a nice link to it from `README.md`.


YannickJadoul:
Is it urgent to move `CONTRIBUTING.md` to `.github`? I've often seen it at the root?

Apart from that, I'm OK with this, but this isn't too urgent and I can imagine @wjakob wants to have a say/some input in this? So I'd suggest waiting before merging this until we can get some feedback?

henryiii:
It was @wjakob's suggestion that we clean up the main folder a bit, though have no idea if likes `CONTRIBUTING.md` in the main folder or not. I am slightly in favor of the `.github` location for it, since it still shows up in all the proper GitHub prompts (that's true in either place) and keeps the main folder slightly cleaner. We can always link to it from the README, etc., if it needs to be more obvious.

No rush, agreed.

YannickJadoul:
(Just assigning to @wjakob, so he can have a clear overview, in a few weeks)

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,163 @@
...
+cmake -S . -B build -DPYTHON_EXECUTABLE=$(which python) -DDOWNLOAD_CATCH=ON -DDOWNLOAD_EIGEN=ON
+cmake --build build -j4
+```
+
+Tips:
+
+* You can use `virtualenv` (from PyPI) instead of `venv` (which is Python 3
+  only).
+* You can select any name for your environment folder.
+* If you use a different shell, there are other activate scripts.
```
bstaletic, line 73: Do we really need to teach people how to use their shell and virtual environments?

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,163 @@
...
+* `-DUSE_PYTHON_INSTALL_DIR=ON`: Try to install into the python dir
+
+
+<details><summary>A few standard CMake tricks: (click to expand)</summary><p>
+
+* Use `cmake --build build -v` to see the commands used to build the files.
+* Use `cmake build -LH` to list the CMake options with help.
+* Use `ccmake` if available to see a curses (terminal) gui, or `cmake-gui` for
+  a completely graphical interface (not present in the PyPI package).
+* Use `-G` and the name of a generator to use something other than `make`, like
```
bstaletic, line 108: Doesn't the default depend on the OS? I'm certain `-G` doesn't default to `GNU Makefiles` on Windows. Maybe not even on macOS.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,163 @@
...
+  `Xcode` or `Ninja` (automatic multithread!).
+* Open the `CMakeLists.txt` with QtCreator to generate for that IDE.
+* Use `cmake --build build -j12` to build with 12 cores (for example).
+* If you are using the `llvm` tool-suite, you can use
+  `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to generate the .json file that the
```
bstaletic, line 113: I don't know about MSVC, but cmake can generate the compilation database when you're compiling with gcc.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,10 +1,15 @@
+---
+name: Bug Report
+about: File an issue about a bug
+title: "[BUG] "
```
bstaletic, line 4: Is it possible to use labels instead of "tags"?

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,10 +1,15 @@
...
 
-1. Check if your question has already been answered in the [FAQ](http://pybind11.readthedocs.io/en/latest/faq.html) section.
-2. Make sure you've read the [documentation](http://pybind11.readthedocs.io/en/latest/). Your issue may be addressed there.
-3. If those resources didn't help and you only have a short question (not a bug report), consider asking in the [Gitter chat room](https://gitter.im/pybind/Lobby).
-4. If you have a genuine bug report or a more complex question which is not answered in the previous items (or not suitable for chat), please fill in the details below.
-5. Include a self-contained and minimal piece of code that reproduces the problem. If that's not possible, try to make the description as clear as possible.
+1. Make sure you've read the [documentation](http://pybind11.readthedocs.io/en/latest/). Your issue may be addressed there.
+2. Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby).
+3. Include a self-contained and minimal piece of code that reproduces the problem. If that's not possible, try to make the description as clear as possible.
```
bstaletic, line 12: Should we add the "search the issue tracker first" to the list?

.github/CONTRIBUTING.md
henryiii: macOS is make too, but yes. There's also an environment variable that controls it (if you like Ninja best, for example). On windows, doens't it default to nmake?

.github/CONTRIBUTING.md
henryiii: I didn't mean it couldn't generate it, that's just where I always use it. Will fix.

.github/ISSUE_TEMPLATE/bug-report.md
henryiii: Yes, but sometimes labels mean "this is a verified bug that needs to be fixed" rather than something that's just been reported. Of course, you could have more labels.

.github/CONTRIBUTING.md
bstaletic: > On windows, doens't it default to nmake?

Could be. I'm not a Windows user.

.github/CONTRIBUTING.md
henryiii: I can make the env folder tip a bit more specific. I can drop the part about other shells, but I'm a fish user, so I actually always run `venv/bin/activate.fish`, which is why I had the tip here. But if someone is using fish/csh, they hopefully know what they are doing, I guess.

.github/ISSUE_TEMPLATE/bug-report.md
bstaletic: That's a good point.

henryiii: Changed the others. This one is still open. `labels: [...]` would set it.

.github/CONTRIBUTING.md
YannickJadoul: As far as I know, it defaults to the latest VS version it can find. No idea what happens if there's no VS.

henryiii: I just avoided listing the default explicitly.

.github/ISSUE_TEMPLATE/bug-report.md
YannickJadoul: > Yes, but sometimes labels mean "this is a verified bug that needs to be fixed" rather than something that's just been reported. Of course, you could have more labels.

I've been trying to do that, if I come across a bug I can confirm but don't have time to resolve.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+`--target` can be spelled `-t` in CMake 3.15+. You can also run individual tests with these targets:
+
+* `pytest`: Python tests only
+* `cpptest`: C++ tests only
+* `test_cmake_build`: Install / subdirectory tests
+
+If you want to build just a subset of tests, use
+`-DPYBIND11_TEST_OVERRIDE="test_callbacks.cpp;test_picking.cpp"`. If this is
```
rwgk, line 131: typo: test_pickling

`-DPYBIND11_TEST_OVERRIDE` easily gets me on the wrong track ("are we testing overrides?").
Maybe `PYBIND11_SELECTED_TESTS`?

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,27 @@
...
+[issue tracker]: https://github.com/pybind/pybind11/issues
+[Gitter chat room]: https://gitter.im/pybind/Lobby
+
+*After reading, remove this checklist and the template text in parentheses below.*
+
+## Issue description
+
+(Provide a short description, state the expected behavior and what actually happens.)
+
+## Reproducible example code
```
rwgk, line 25: Would it make sense to encourage sending a PR with a new, failing test? That could be used as a starting point for the unit test after fixing.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+sections on how to contribute code and bug reports.
+
+### Reporting bugs
+
+Before submitting a question or bug report, please take a moment of your time
+and ensure that your issue isn't already discussed in the project documentation
+provided at
+[http://pybind11.readthedocs.org/en/latest](http://pybind11.readthedocs.org/en/latest).
+
+Assuming that you have identified a previously unknown problem or an important
```
rwgk, line 11: Maybe replace this entire paragraph with a link to the issue template?

.github/CONTRIBUTING.md
henryiii: That's just the existing name, made a bit more public. I'd be open to change it, but probably in another PR (unless it really makes sense to do it here since it wasn't very public before).

henryiii:
I've moved it back - we can move it later.

CONTRIBUTING.md
```diff
@@ -17,18 +15,19 @@ isolate the function(s) that cause breakage, submit matched and complete C++
 and Python snippets that can be easily compiled and run on my end.
```
rwgk, line 15: Replace the last part of the sentence with "... compiled and run in isolation"?
Mainly because "my" isn't correct anymore.

.github/CONTRIBUTING.md
henryiii: I added better links.

wjakob:
Looks all good to me, and moving the github-related markdown files to a `.github` sub-directory is great.

henryiii:
I'll do the move in a separate PR.

PR #2382, opened 2020-08-11
Fix warning C26817 on copying in `for (auto vh : value_and_holder(...))`

### LINKED ISSUE #2379: Warning when compiling under Visual Studio 16.7.0
## Issue description

Warning C26817
Potentially expensive copy of variable 'vh' in range-for loop. Consider making it a const reference (es.71).
include\pybind11\cast.h:461

MichaelGoulding:
C26817: Potentially expensive copy of variable 'vh' in range-for loop. Consider making it a const reference (es.71).

EDIT (@YannickJadoul): Fixes #2379

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -458,7 +458,7 @@ PYBIND11_NOINLINE inline handle get_object_handle(const void *ptr, const detail:
     auto &instances = get_internals().registered_instances;
     auto range = instances.equal_range(ptr);
     for (auto it = range.first; it != range.second; ++it) {
-        for (auto vh : values_and_holders(it->second)) {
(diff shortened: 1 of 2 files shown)


bstaletic:
In theory, `value_and_holder` is big enough to be passed by reference. In practice, gcc and clang seem to output the same assembly.

Besides the `auto vh` in `cast.h`, there's also one in `detail/class.h`. We should fix both.

YannickJadoul:
Green, green, green, green, green, ...

Thanks for reporting and fixing, @MichaelGoulding!

MichaelGoulding:
Thanks, for fixing up the last bit. @YannickJadoul

PR #2387, opened 2020-08-12
ci: include Python 3.9 RC1

henryiii:
The patch we submitted to Python should be included in yesterday's Python 3.9 RC1, so this should pass now. Fingers crossed!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,12 +58,6 @@ jobs:
             arch: x64
             max-cxx-std: 17
 
-            # Currently can't build due to warning, fixed in CPython > 3.9b5
-          - runs-on: macos-latest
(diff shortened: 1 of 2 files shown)


henryiii:
No "real" changes, just adding Python 3.9rc1 on macOS (though 3.9 was building and testing NumPy before - we haven't traditionally done that (and it's slow), so I think that's fine to drop, it was really just an oversight before).

henryiii:
It already helped us get a PR into Python 3.9 before the release :) https://github.com/python/cpython/pull/21600

henryiii:
The sooner we catch bugs, either in our code or in Python (as linked above), the better for everyone. The goal is to be Python 3.9 ready on the day of release.

PR #2112, opened 2020-02-12
Use 'raise from' in initialization

virtuald:
We ran into an issue where a unicode decode exception occurred during initialization, and it was difficult to figure out what was causing the exception without looking directly at the object. Unfortunately, pybind11 currently swallows all exceptions by only retaining the text. This would change it to use the equivalent of `raise ImportError('something') from e`.

There are several things to be considered here:

* [x] I think this mechanism was introduced in python 3.3?, need to add appropriate #ifdefs, not 100% sure the best way to go about this 
* [x] Needs tests I assume?
* [ ] Probably should apply this in other places where exceptions might be swallowed, but I think that's more appropriate for another PR?

## Suggested changelog entry:

```rst
* Added ``py::raise_from`` to enable chaining exceptions
```

Issue #1716, opened 2019-03-03
crash when matching overload not found


germasch:
## Issue description

This issue was observed as a [ADIOS2](https://github.com/ornladios/ADIOS2) [issue](https://github.com/ornladios/ADIOS2/issues/1233). pybind11 is used to generate the python bindings, but I believe it's not specific to ADIOS2. In same cases, calling a C++ function with the wrong arguments causes the python interpreter to crash.

## Reproducible example code

So I know I'm supposed to provide a simple test case, which would be easier if I didn't have to deal with the hassles of building it. Actually, your doc says, to run the tests that come with pybind11, do
```
mkdir build
cd build
cmake ..
make check -j 4
```
But `make check` doesn't work (make does).

Anyway, here's what I cut it down to:
```cxx
void open_test(int i, bool b)
{
  std::cerr << "open_test " << i << " " << b << std::endl;
}

PYBIND11_MODULE(adios2, m)
{
    m.def("open_test", &open_test);
}
```

```py
import numpy as np
import adios2

myArray = np.array([0, 1., 2., 3., 4., 5., 6., 7., 8., 9.])

adios2.open_test(2, True) # <-- OK
adios2.open_test(2, myArray) # <-- crashes
```

Output:
```
[kai@macbook build (python_cleanup *)]$ python ../examples/hello/bpWriter/x.py
open_test 2 1
libc++abi.dylib: terminating with uncaught exception of type pybind11::error_already_set: SystemError: <class 'type'> returned a result with an error set

At:
  /opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/numpy/core/arrayprint.py(1416): array_repr
  ../examples/hello/bpWriter/x.py(8): <module>
[...]
```

I think I roughly figured out what's happening:
* dispatcher goes through available overload and tries to call them
* if none matches,  an error message is supposed to be printed, listing the available overloads, and
the actual arguments.
* composing that message is where the uncaught exception appears. Specifically, it happens in pytypes.h:repr()
```cxx
PyObject *str_value = PyObject_Repr(h.ptr());
    if (!str_value) throw error_already_set();
```
when `repr` is called to make a string out of the (actual) numpy.array argument in order to compose the "incompatible arguments" message.

`repr` on the numpy array fails because there is already an active exception, which makes things fail in  the first line below in `arrayprint.c`.
```
    if type(arr) is not ndarray:
        class_name = type(arr).__name__
    else:
        class_name = "array"
[...]

PR #2389, opened 2020-08-12
fix: throwing repr caused a segfault

### LINKED ISSUE #2383: Segfault when __repr__ throws
## Issue description

If a `__repr__` throws, then passing an invalid object causes a segfault.

## Reproducible example code

```diff
diff --git a/tests/test_exceptions.cpp b/tests/test_exceptions.cpp
index 372d0ae..8244322 100644
--- a/tests/test_exceptions.cpp
+++ b/tests/test_exceptions.cpp
@@ -218,4 +218,7 @@ TEST_SUBMODULE(exceptions, m) {
         }
     });

+    // Test repr that cannot be displayed
+    m.def("simple_bool_passthrough", [](bool x){return x;});
+
 }
diff --git a/tests/test_exceptions.py b/tests/test_exceptions.py
index b794e3d..64bdc63 100644
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -179,3 +179,14 @@ def test_nested_throws(capture):
     with pytest.raises(m.MyException5) as excinfo:
         m.try_catch(m.MyException, pycatch, m.MyException, m.throws5)
     assert str(excinfo.value) == "this is a helper-defined translated exception"
+
+
+# This can often happen if you wrap a pybind11 class in a Python wrapper
+def test_invalid_repr():
+
+    class MyRepr(object):
+        def __repr__(self):
+            raise AttributeError("Example error")
+
+
+    m.simple_bool_passthrough(MyRepr())
```

henryiii:
Closes #2383 , thanks to hint from @jbarlow83 .

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -776,7 +776,11 @@ protected:
             for (size_t ti = overloads->is_constructor ? 1 : 0; ti < args_.size(); ++ti) {
                 if (!some_args) some_args = true;
                 else msg += ", ";
-                msg += pybind11::repr(args_[ti]);
+                try {
+                    msg += pybind11::repr(args_[ti]);
+                } catch (const error_already_set&) {
+                    msg += pybind11::repr(args_[ti].get_type());
+                }
             }
             if (kwargs_in) {
                 auto kwargs = reinterpret_borrow<dict>(kwargs_in);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -779,7 +779,7 @@ protected:
                 try {
                     msg += pybind11::repr(args_[ti]);
                 } catch (const error_already_set&) {
-                    msg += pybind11::repr(args_[ti].get_type());
+                    msg += "<repr raised Error>";
                 }
             }
             if (kwargs_in) {
@@ -791,7 +791,12 @@ protected:
                     for (auto kwarg : kwargs) {
                         if (first) first = false;
                         else msg += ", ";
-                        msg += pybind11::str("{}={!r}").format(kwarg.first, kwarg.second);
+                        msg += pybind11::str("{}=").format(kwarg.first);
+                        try {
+                            msg += pybind11::repr(kwarg.second);
+                        } catch (const error_already_set&) {
+                            msg += "<repr raised Error>";
+                        }
                     }
                 }
             }
--- a/tests/test_exceptions.cpp
+++ b/tests/test_exceptions.cpp
@@ -194,4 +194,7 @@ TEST_SUBMODULE(exceptions, m) {
         }
     });
 
+    // Test repr that cannot be displayed
+    m.def("simple_bool_passthrough", [](bool x) {return x;});
+
 }
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -149,3 +149,16 @@ def test_nested_throws(capture):
     with pytest.raises(m.MyException5) as excinfo:
         m.try_catch(m.MyException, pycatch, m.MyException, m.throws5)
     assert str(excinfo.value) == "this is a helper-defined translated exception"
+
+
+# This can often happen if you wrap a pybind11 class in a Python wrapper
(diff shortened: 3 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -776,7 +776,11 @@ class cpp_function : public function {
             for (size_t ti = overloads->is_constructor ? 1 : 0; ti < args_.size(); ++ti) {
                 if (!some_args) some_args = true;
                 else msg += ", ";
-                msg += pybind11::repr(args_[ti]);
+                try {
+                    msg += pybind11::repr(args_[ti]);
+                } catch (const error_already_set&) {
+                    msg += pybind11::repr(args_[ti].get_type());
```
henryiii, line 782: The repr produced is pretty ugly, but better than a segfault. Maybe should indicate it's an instance of the class?

```
>       m.simple_bool_passthrough(MyRepr())
E       TypeError: simple_bool_passthrough(): incompatible function arguments. The following argument types are supported:
E           1. (arg0: bool) -> bool
E
E       Invoked with: <class 'test_exceptions.test_invalid_repr.<locals>.MyRepr'>

test_exceptions.py:162: TypeError
```

jbarlow83: Yes, it should probably say it's an instance.

This is the sort of situation where Python would chain exceptions ("While handling this error, another error occurred..."). but there's no C API for that. Although I wonder if it's worth indicating that an exception occurred in `repr()`.

jbarlow83:
The call to `pybind11::str` at line 795 can probably trigger a segfault too, if `__str__` raise as an error.

include/pybind11/pybind11.h
henryiii: It doesn't seem like this information is always available - some compilers are not showing it. Maybe just a "<repr raised error>" message would work?

jbarlow83: Maybe `discard_as_unraisable()` does make sense here. It would get the stack trace and exception from `__repr__` out, and would not be compiler-dependent.

henryiii: Okay, we could use it then; I've added an unchanging  `<repr raised Error>` currently but would be happy to change it.

YannickJadoul: We could potentially put the exception as string into `<repr raised str(exc)>`, but maybe we'll start some infinite recursion, then?

virtuald: There is a C API for chaining exceptions, see #2112

jbarlow83: What I meant was, for their own reasons, the Python core developers chose to not automate chaining exceptions in C or provide an API to make it easier to do.
https://www.python.org/dev/peps/pep-0490/
It's hard to tell what to make of the situation, in terms of what they think C extensions to do.

We potentially have exceptions for each argument to the function, and each of those could involve an individual chain of exceptions - potentially there's a tree of exceptions. I think chaining these exceptions to each other could muddy the waters because it could end up linking the exceptions from attempts to repr() different arguments. Write unraisable here gets the exception out, and then we print a message complaining that we can't repr() the arguments.

henryiii:
I currently just have a simple `<repr raised Error>`, should we do more? It's still better than a segfault. ;)

YannickJadoul:
Nah, that's probably fine enough. If this happens, it's because there's a mistake anyway, right? We just don't want stuff to segfault.

YannickJadoul:
Is this related to #1716, though?

henryiii:
I think it would solve that segfault, but the error message would not be as ideal, I'd think?

YannickJadoul:
> I think it would solve that segfault, but the error message would not be as ideal, I'd think?

Reading again, it might not be solved, actually? It seems like `type_caster` for `bool` leaves the error indicator set in #1716?

wjakob:
nice, I vaguely remember running into this in a project of mine, but I never got around to fixing it.

henryiii:
I think this can go in, then, and the messaging can be improved later if it's needed?

PR #2392, opened 2020-08-14
Add check if `str(handle)` correctly converted the object, and throw py::error_already_set if not

YannickJadoul:
Only problematic on Python 2, I think. Discovered here: https://github.com/pybind/pybind11/pull/2380#discussion_r469649144

### DIFF
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -104,10 +104,19 @@ def test_str(doc):
 
     assert m.str_from_object(A()) == "this is a str"
     assert m.repr_from_object(A()) == "this is a repr"
+    assert m.str_from_handle(A()) == "this is a str"
 
     s1, s2 = m.str_format()
     assert s1 == "1 + 2 = 3"
     assert s1 == s2
+
+    malformed_utf8 = b"\x80"
+    with pytest.raises(UnicodeDecodeError) as excinfo:
+        assert m.str_from_object(malformed_utf8)
+    assert 'invalid start byte' in str(excinfo.value)
+    with pytest.raises(UnicodeDecodeError) as excinfo:
+        assert m.str_from_handle(malformed_utf8)
+    assert 'invalid start byte' in str(excinfo.value)
 
 
 def test_bytes(doc):
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -111,12 +111,14 @@ def test_str(doc):
     assert s1 == s2
 
     malformed_utf8 = b"\x80"
-    with pytest.raises(UnicodeDecodeError) as excinfo:
-        assert m.str_from_object(malformed_utf8)
-    assert 'invalid start byte' in str(excinfo.value)
-    with pytest.raises(UnicodeDecodeError) as excinfo:
-        assert m.str_from_handle(malformed_utf8)
-    assert 'invalid start byte' in str(excinfo.value)
+    if env.PY2:
+        with pytest.raises(UnicodeDecodeError):
+            assert m.str_from_object(malformed_utf8)
+        with pytest.raises(UnicodeDecodeError):
+            assert m.str_from_handle(malformed_utf8)
+    else:
+        assert m.str_from_object(malformed_utf8) == "b'\\x80'"
+        assert m.str_from_handle(malformed_utf8) == "b'\\x80'"
 
 
 def test_bytes(doc):
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -918,7 +918,7 @@ public:
         Return a string representation of the object. This is analogous to
         the ``str()`` function in Python.
     \endrst */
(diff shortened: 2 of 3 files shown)


YannickJadoul:
Also note that by checking always checking the result after calling `raw_ptr`, we can more the `if (!str_value) throw error_already_set();` inside the Python 2-only preprocessor block, and get rid of one full pointer comparison (!) on Python 3, when calling the constructor from `py::object`! ;-)

YannickJadoul:
Of course. I tried this locally based on #2380, but `detail::PyUnicode_Check_Permissive` first still needs to die, such that `py::str` cannot contain a `bytes`-object anymore!

YannickJadoul:
Rebased onto #2380, but waiting for #2380 to get merged before this can get rebased and merged.

tests/test_pytypes.py
```diff
@@ -102,11 +102,25 @@ def __repr__(self):
...
     assert m.str_from_object(A()) == "this is a str"
     assert m.repr_from_object(A()) == "this is a repr"
+    assert m.str_from_handle(A()) == "this is a str"
 
     s1, s2 = m.str_format()
     assert s1 == "1 + 2 = 3"
     assert s1 == s2
 
+    malformed_utf8 = b"\x80"
+    if sys.version_info.major < 3:
```
YannickJadoul, line 112: TODO: Replace with `PY2`

Also, this is quite weird behavior (i.e., quite different between Python 2 and 3), but it's the same that's now already there.

henryiii:
@YannickJadoul if you used `pre-commit install` this couldn't happen. ;)

henryiii:
This might benefit from being rebased on #2397

tests/test_pytypes.py
```diff
@@ -205,6 +222,11 @@ def test_constructors():
     }
     inputs = {k.__name__: v for k, v in data.items()}
     expected = {k.__name__: k(v) for k, v in data.items()}
+    if pytest.PY2:  # Similar to the above. See comments above.
```
henryiii, line 225: All these will need to be `env.PY2` after #2397.

include/pybind11/cast.h
```diff
@@ -1631,6 +1631,14 @@ struct pyobject_caster {
 
     template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
     bool load(handle src, bool /* convert */) {
+#ifndef PYBIND11_DISABLE_IMPLICIT_STR_FROM_BYTES
```
henryiii, line 1634: JK, why does this have an escape hatch? It seems to not have been added as an option to the CMakeLists. (an addition which should wait till after #2370 if it gets added, anyway)

tests/test_pytypes.py
henryiii: Or, if this goes in first, then @2397 needs to be rebased and modified for this, which is also fine.

YannickJadoul:
> @YannickJadoul if you used `pre-commit install` this couldn't happen. stuck_out_tongue_winking_eye

OK, you win this one ;-)

include/pybind11/cast.h
YannickJadoul: Hmmm, wait, this isn't mine. This is coming from @rwgk's #2380 (which I had to rebase on to get some tests working, at some point). Let me check this again (and probably rebase onto the latest `str-bytes-cleanup`, though I thought I'd done so...).

YannickJadoul: OK, fixed. Not sure what went wrong, but thanks for catching that!

tests/test_pytypes.py
YannickJadoul: This wasn't mine either, but I've pushed 441bc04d44e8889488ac9a4c719df1caa1a11a44 directly to `str-bytes-cleanup`!

include/pybind11/cast.h
YannickJadoul: Full disclosure on this macro: https://github.com/pybind/pybind11/pull/2380#discussion_r469575686
So, as far as I know, it's not supposed to be public, and it's still up for debate on what to do with ti :-)

tests/test_pytypes.py
YannickJadoul: Done.

YannickJadoul:
Fixed this. It's quite a minor PR, so a few more quick checks (@bstaletic & @henryiii, the things you mentioned should be fixed).
@rwgk, maybe you should give this a quick final check (since it involves `py::str`), and merge into `str-bytes-cleanup` at will if you're happy with it?

wjakob:
LGTM.

rwgk:
> Looks great, thanks Yannick!
> I'll hold off merging only because I'm still unclear why the current `str-bytes-cleanup` branch fails to pass some tests in the Google environment (I didn't get a chance to work on it at all yesterday).
> (For completeness: the failing tests are NOT pybind11 unit tests.)

For future reference: the test failure was due to accidental `str` vs `bytes` mixup, masked by the double-nature of `pybind11::str`.

PR #2393, opened 2020-08-14
ci: include Boost

henryiii:
Add Boost to the CI checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,8 +80,16 @@ jobs:
         python-version: ${{ matrix.python }}
         architecture: ${{ matrix.arch }}
 
+    - name: Setup Boost
+      if: runner.os != 'macOS'
+      run: echo "::set-env name=BOOST_ROOT::$BOOST_ROOT_1_72_0"
+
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Er, just to check, what was the motivation for this? (issue number, or Slack convo?)

I'm watching paint dry in #2376, and it seems like the Boost download can take 1m - 3m in different Mac builds:
https://github.com/pybind/pybind11/pull/2376/checks?check_run_id=985786507#step:5:9

EricCousineau-TRI:
Ah, benchmarking, gotcha!
https://github.com/pybind/pybind11/blob/cba4a98546ed0e054fb85d1a7eb70a37154d8fd4/docs/benchmark.rst

henryiii:
I think it also tests boost::variant (probably v1 and not v2). We might be able to cache homebrew.

henryiii:
We can drop macOS, and leave Boost only for unix/Windows (since they don't plan to add Boost to the macOS image).

EricCousineau-TRI:
Aye. Submitted naive PR: #2395

PR #2395, opened 2020-08-14
ci: Remove "Setup Boost (macOS)" step

EricCousineau-TRI:
Per convo: https://github.com/pybind/pybind11/pull/2393#issuecomment-674186838 - slows down the macOS builds for very little gain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,10 +84,6 @@ jobs:
       if: runner.os != 'macOS'
       run: echo "::set-env name=BOOST_ROOT::$BOOST_ROOT_1_72_0"
 
-    - name: Setup Boost (macOS)
-      if: runner.os == 'macOS'
-      run: brew install boost
-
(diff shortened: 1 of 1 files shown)


henryiii:
I miss GitLab's merge when checks pass option. :'(

EricCousineau-TRI:
Ooh, hadn't seen that in GitLab! But yeah, that would be awesome in cases like these...

YannickJadoul:
We tried out https://github.com/apps/probot-auto-merge on `cibuildwheel`, at some point. It's ... yeah, a replacement, but there should be a better way.

henryiii:
Did that work well enough?

YannickJadoul:
It seems so: joerick/cibuildwheel#383 It worked, but it's maybe a bit weird to have this bot merge it, rather than it just being a feature of the platform...

bstaletic:
https://mergify.io/ seems like an alternative to probot-auto-merge.

PR #2396, opened 2020-08-14
Using recently added `pytest.PY2` instead of `str is bytes`.

rwgk:
Important gain: uniformity & therefore easier cleanup when we drop PY2 support.
Very slight loss: it was nice to have `str is bytes` as a reminder in this specific context.

### DIFF
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -224,7 +224,7 @@ def test_pybind11_str_raw_str():
     # specifically to exercise pybind11::str::raw_str
     cvt = m.convert_to_pybind11_str
     assert cvt(u"Str") == u"Str"
-    assert cvt(b'Bytes') == u"Bytes" if str is bytes else "b'Bytes'"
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Failure here is for Python 3.9-dev on macOS:
https://github.com/pybind/pybind11/pull/2396/checks?check_run_id=986425511#step:14:56

```
     def test_python_to_cpp_to_python_from_thread_multiple_parallel():
        """Makes sure there is no GIL deadlock when running in a thread multiple times in parallel.
    
        It runs in a separate process to be able to stop and assert if it deadlocks.
        """
>       assert _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=True) == 0
E       assert -11 == 0
E        +  where -11 = _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=True)

test_gil_scoped.py:65: AssertionError
```

I say this is good to merge, as the failure is unrelated to this PR?

EricCousineau-TRI:
I'ma go ahead and pull the "merge" trigger.

rwgk:
Awesome, thanks Eric!

PR #2397, opened 2020-08-15
tests: cleanup and ci hardening

henryiii:
Refactor the tests. Use a consistent failing/skipping scheme.

* Attempt to allow 3.9-dev to fail (not sure how it works)
* Fail-fast diagnostic removed, fails fast (once) with the correct error message without a custom wrapper.
* Using xfail where possible (skip means not-applicable, xfail means it should ideally pass)
* Use importorskip instead of ugly custom skips
* PyTest 3.1+ required instead of 3.0 (current is 6.0 for Py3, 4.6.9 for Py2) - better warning reporting
* Use info.py to simplify skips/xfails/checks
* Update links to PyPy issues
* Several (7 or so) skipped PyPy tests are now passing (especially those requiring NumPy) (reported by xfail!)
* Strings used for the skipif/xfails in most cases so that PyTest reports the condition causing the skip/xfail directly in the report.

Skips/xfails are more consistent, no more `pytest.bug_in_pypy`, `pytest.unsupported_on_pypy_lt_6`, etc. (PyPy 7+ required, so ignoring check for old PyPy verisions). No more injection into the pytest module. xfails are reported at the end and most are strictly required. We can use `reasons` or `raises` to improve the test reporting.


tests/info.py
```diff
@@ -0,0 +1,12 @@
...
+import sys
+
+LINUX = sys.platform.startswith("linux")
+MACOS = sys.platform.startswith("darwin")
+WIN = sys.platform.startswith("win32") or sys.platform.startswith("cygwin")
+
+CPYTHON = platform.python_implementation() == "CPython"
+PYPY = platform.python_implementation() == "PyPy"
+
+PY2 = sys.version_info.major == 2
```
YannickJadoul, line 12: If we're vendoring `six`, I think it already contains `PY2`?

tests/test_eval.py
```diff
@@ -15,7 +19,7 @@ def test_evals(capture):
     assert m.test_eval_failure()
 
 
-@pytest.unsupported_on_pypy3
+@pytest.mark.xfail("info.PYPY and not info.PY2", raises=RuntimeError)
```
YannickJadoul, line 22: Do these all need to be strings, if `info` is imported anyway?

tests/info.py
YannickJadoul: (Ah, wait, this is not vendoring `six` anymore)

.pre-commit-config.yaml
```diff
@@ -13,6 +13,7 @@ repos:
   - id: requirements-txt-fixer
   - id: trailing-whitespace
   - id: fix-encoding-pragma
+    exclude: ^tests/six.py$
```
YannickJadoul, line 16: Not necessary anymore?

henryiii: Yes, thanks!

tests/test_eval.py
henryiii: Yes, if it's a string, pytest knows why it failed. Otherwise, a `reason=` is required.

YannickJadoul: Right, OK, that's valuable, indeed; less syntax highlighting but more info...

tests/pytest.ini
```diff
@@ -1,11 +1,16 @@
...
 norecursedirs = test_cmake_build test_embed
 addopts =
     # show summary of skipped tests
     -rs
+    # Report xfails (uncomment to use)
+    # -rx
     # capture only Python print and C++ py::print, but not C output (low-level Python errors)
     --capture=sys
+    # warnings for deprecated features
+    -Wd
```
bstaletic, line 12: Why not `-Wa`?

tests/info.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
bstaletic, line 1: Is `info.py` a good name for this? I know that this comment can devolve into bikeshedding... C++ uses `<version>` for this kind of stuff.

tests/test_eval.py
```diff
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 import os
+
 import pytest
+
+import info  # noqa: F401
+
```
bstaletic, line 7: These empty lines seem unnecessary if you ask me.

tests/info.py
henryiii: It mostly needed to be short and simple and not a common existing package. `version.PY2` is longer, not quite as clear, and doesn't actually contain the pybind11 version, which is what you'd rather expect. I'm not sold on the name, though.

tests/test_eval.py
henryiii: I'm using [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) import groups.

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
> 
> Imports should be grouped in the following order:
> 
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
> You should put a blank line between each group of imports.

tests/info.py
bstaletic: I agree that `version.py` is an even worse name. I still don't like `info.py`. Anyway, I'm not going to block the PR on this.

henryiii: `plat.py`?

bstaletic: Or maybe `env.py`? `env` is a much more common abbreviation for environment than `plat` is for platform. On top of that, the standard library has `platform.py`.

tests/test_eval.py
bstaletic: Thanks for the reminder.

henryiii:
I believe ~8 tests are now passing that used to fail (and were hidden by skip instead of xfail), mostly due to better NumPy support in recent PyPy's.

tests/test_methods_and_attributes.py
```diff
@@ -261,7 +261,7 @@ def test_property_rvalue_policy():
 
 
 # https://foss.heptapod.net/pypy/pypy/-/issues/2447
-@pytest.mark.xfail("env.PYPY")
+@pytest.mark.xfail("env.PYPY", strict=True)
```
bstaletic, line 264: Why are we adding `strict=True` instead of using `xfail_strict=True` in `pytest.ini`?
I'm probably missing something simple.

henryiii: I have a test where I want `strict=False`, because it seems to randomly fail. And I didn't realize that the ini option only changes the default (oops!).

YannickJadoul:
> I believe 7 or so tests are now passing in PyPy that used to fail (and were hidden by skip instead of xfail), mostly due to better NumPy support in recent PyPy's.

That's great to hear! -)

henryiii:
If I get two approvals, I'll squash and merge. :)

PR #2399, opened 2020-08-16
tests: loosen time test, not valid on some systems

### LINKED ISSUE #2398: Flaky timing Windows test in CI
There's occasionally a diff of a 1 second in the test that compares two times, mostly on Windows, probably due to the shared infrastructure. We should probably loosen it up just a bit, or maybe compute a bound by running the python clock both before and after, etc.

https://github.com/pybind/pybind11/blob/cd856992121b0a98bbe6e15d9a886f2b43f3d972/tests/test_chrono.py#L8-L25

henryiii:
Fix #2398

Test time diff < 2 seconds instead of < 0.5 seconds since `time.time()` only has 1s accuracy on some systems.

### DIFF
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -15,14 +15,12 @@ def test_chrono_system_clock():
     # The numbers should vary by a very small amount (time it took to execute)
     diff = abs(date1 - date2)
 
-    # There should never be a days/seconds difference
+    # There should never be a days difference
     assert diff.days == 0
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Sounds good to me! The actual goal of this test is to make sure it doesn't just produce random data and that the C++ and Python implementation match each other, I suppose?

(Alternatively, you could try `<= 1`, but I guess it's fine either way.)

PR #2402, opened 2020-08-17
docs: move CONTRIBUTING

henryiii:
Followup to #2377, moves contributing.

Also modernizes _some_ of README, including adding a prominent link to the contributing document. ~~The most notable point is the update to the PyPy version - currently we test PyPy/PyPy3 7.3. If we state an earlier version here, we should probably test it. Open to doing something different though.~~


YannickJadoul:
> The most notable point is the update to the PyPy version - currently we test PyPy/PyPy3 7.3.

Happy with that, but I feel this is something that should pass @wjakob's eye, since it's changing one of the promises of pybind11.

henryiii:
Agreed. I don't like promising anything we don't test, so if we promise earlier, we should add a test.

henryiii:
I've made the statement less promisy and more truthful (tested on 7.3) so this can move forward. @wjakob has already signed off on the CONTRIBUTING move, which was the focus of this PR.

YannickJadoul:
I'd still like an answer to the minimal PyPy version we're supporting. We could potentially get rid of a bunch of code, if we pick a sane, modern, by-PyPy-supported version.

But that shouldn't stop this PR, indeed. Maybe create an issue?

PR #2409, opened 2020-08-19
Changing pybind11::str to exclusively hold PyUnicodeObject

rwgk:
Before this PR, `pybind11::str` can hold `PyUnicodeObject` or `PyBytesObject`, which is probably surprising and was never documented. As a side-effect, `pybind11::isinstance<str>()` is `true` for both `pybind11::str` and `pybind11::bytes`. This PR changes the `pybind11::str` implementation to be in line with the documented behavior, but provides an escape hatch to go back to the legacy behavior, via the `PYBIND11_STR_LEGACY_PERMISSIVE` macro. This macro will be removed in future releases.

This PR changes `pybind11::str` so that it can only hold `PyUnicodeObject`, and `pybind11::isinstance<str>()` is true only for `pybind11::str`, but false for `pybind11::bytes`. However, for Python 2 only (!), the `pybind11::str` caster is modified to implicitly decode `bytes` to `PyUnicodeObject`. Without this implicit conversion, Python code currently used with Python 2 & 3 would need to be cluttered with `six.ensure_text()` or similar, only to be un-cluttered later after Python 2 support is dropped.

This PR was exhaustively tested in the Google environment (hundreds of thousands of indirect dependencies). A one-page summary of  user code fixes needed is [here](https://docs.google.com/document/d/1TPL-J__mph_yHa1quDvsO12E_F5OZnvBaZlW9IIrz8M/), along with fixes needed for other PRs. The number of fixes needed in connection with this PR was similar to that for other PRs. Two types of required fixes are expected to be common:

* Accidental use of `pybind11::str` instead of `pybind11::bytes`, masked by the legacy permissive behavior. These are probably very easy to fix.

* Reliance on `pybind11::isinstance<str>(obj)` being `true` for `bytes`. This is likely to be easy to fix in most cases by adding `|| pybind11::isinstance<bytes>(obj)`, but a fix may be more involved, e.g. if `pybind11::isinstance<T>` appears in a `template` (we found one such case in the Google environment).

### DIFF
--- a/docs/upgrade.rst
+++ b/docs/upgrade.rst
@@ -10,6 +10,31 @@ modernization and other useful information.
 
 .. _upgrade-guide-2.6:
 
+v2.7
+====
+
+*Before* v2.7, ``py::str`` can hold ``PyUnicodeObject`` or ``PyBytesObject``,
+and ``py::isinstance<str>()`` is ``true`` for both ``py::str`` and
+``py::bytes``. Starting with v2.7, ``py::str`` exclusively holds
+``PyUnicodeObject`` (`#2409 <https://github.com/pybind/pybind11/pull/2409>`_),
+and ``py::isinstance<str>()`` is ``true`` only for ``py::str``. To help in
+the transition of client code, the ``PYBIND11_STR_LEGACY_PERMISSIVE`` macro
+is provided as an escape hatch to go back to the legacy behavior. This macro
+will be removed in future releases. Two types of required client-code fixes
+are expected to be common:
+
+* Accidental use of ``py::str`` instead of ``py::bytes``, masked by the legacy
+  behavior. These are probably very easy to fix, by changing from
+  ``py::str`` to ``py::bytes``.
+
+* Reliance on py::isinstance<str>(obj) being ``true`` for
+  ``py::bytes``. This is likely to be easy to fix in most cases by adding
+  ``|| py::isinstance<bytes>(obj)``, but a fix may be more involved, e.g. if
+  ``py::isinstance<T>`` appears in a template. Such situations will require
+  careful review and custom fixes.
+
+
+
 v2.6
 ====
 
--- a/docs/upgrade.rst
+++ b/docs/upgrade.rst
@@ -18,10 +18,10 @@ and ``py::isinstance<str>()`` is ``true`` for both ``py::str`` and
 ``py::bytes``. Starting with v2.7, ``py::str`` exclusively holds
 ``PyUnicodeObject`` (`#2409 <https://github.com/pybind/pybind11/pull/2409>`_),
 and ``py::isinstance<str>()`` is ``true`` only for ``py::str``. To help in
-the transition of client code, the ``PYBIND11_STR_LEGACY_PERMISSIVE`` macro
+the transition of user code, the ``PYBIND11_STR_LEGACY_PERMISSIVE`` macro
 is provided as an escape hatch to go back to the legacy behavior. This macro
-will be removed in future releases. Two types of required client-code fixes
-are expected to be common:
+will be removed in future releases. Two types of required fixes are expected
+to be common:
 
 * Accidental use of ``py::str`` instead of ``py::bytes``, masked by the legacy
   behavior. These are probably very easy to fix, by changing from
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1656,7 +1656,10 @@ struct pyobject_caster {
 
     template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
     bool load(handle src, bool /* convert */) {
-#if defined(PYBIND11_STR_NON_PERMISSIVE) && !defined(PYBIND11_STR_CASTER_NO_IMPLICIT_DECODE)
+#if PY_MAJOR_VERSION < 3 && !defined(PYBIND11_STR_LEGACY_PERMISSIVE)
+        // For Python 2, without this implicit conversion, Python code would
+        // need to be cluttered with six.ensure_text() or similar, only to be
+        // un-cluttered later after Python 2 support is dropped.
         if (std::is_same<T, str>::value && isinstance<bytes>(src)) {
             PyObject *str_from_bytes = PyUnicode_FromEncodedObject(src.ptr(), "utf-8", nullptr);
             if (!str_from_bytes) throw error_already_set();
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1656,6 +1656,14 @@ struct pyobject_caster {
 
     template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
     bool load(handle src, bool /* convert */) {
+#if defined(PYBIND11_STR_NON_PERMISSIVE) && !defined(PYBIND11_STR_CASTER_NO_IMPLICIT_DECODE)
+        if (std::is_same<T, str>::value && isinstance<bytes>(src)) {
+            PyObject *str_from_bytes = PyUnicode_FromEncodedObject(src.ptr(), "utf-8", nullptr);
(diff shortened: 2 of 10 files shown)


wjakob:
Those changes look good to me. Putting string-specific code into the `py::object` caster is a no-go in principle, but I see that it is masked via a condition that can be tested at compile time, and it's for Python 2.7 only (i.e. will eventually be removed). Can you document this new flag in the documentation (e.g. in the porting guide?)

rwgk:
> Can you document this new flag in the documentation (e.g. in the porting guide?)

Done. Adding to `docs/upgrade.rst` is the only change I made. Thanks Wenzel!

docs/upgrade.rst
```diff
@@ -10,6 +10,31 @@ modernization and other useful information.
...
 
+v2.7
+====
+
+*Before* v2.7, ``py::str`` can hold ``PyUnicodeObject`` or ``PyBytesObject``,
+and ``py::isinstance<str>()`` is ``true`` for both ``py::str`` and
+``py::bytes``. Starting with v2.7, ``py::str`` exclusively holds
+``PyUnicodeObject`` (`#2409 <https://github.com/pybind/pybind11/pull/2409>`_),
+and ``py::isinstance<str>()`` is ``true`` only for ``py::str``. To help in
+the transition of client code, the ``PYBIND11_STR_LEGACY_PERMISSIVE`` macro
```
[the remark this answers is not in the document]
YannickJadoul: Oh, nevermind, it is actually 2 lines further down; I guess I was just expecting/scanning for "2.8"...

docs/upgrade.rst
```diff
@@ -10,6 +10,31 @@ modernization and other useful information.
...
+is provided as an escape hatch to go back to the legacy behavior. This macro
+will be removed in future releases. Two types of required client-code fixes
```
YannickJadoul, line 23: Again, all code is "client-code" in this context, to me.

docs/upgrade.rst
```diff
@@ -10,6 +10,31 @@ modernization and other useful information.
...
+are expected to be common:
+
+* Accidental use of ``py::str`` instead of ``py::bytes``, masked by the legacy
+  behavior. These are probably very easy to fix, by changing from
```
YannickJadoul, line 27: You might mention this is probably a bug?

include/pybind11/cast.h
```diff
@@ -1623,6 +1623,17 @@ struct pyobject_caster {
...
+        // For Python 2, without this implicit conversion, Python code would
+        // need to be cluttered with six.ensure_text() or similar, only to be
+        // un-cluttered later after Python 2 support is dropped.
+        if (std::is_same<T, str>::value && isinstance<bytes>(src)) {
+            PyObject *str_from_bytes = PyUnicode_FromEncodedObject(src.ptr(), "utf-8", nullptr);
+            if (!str_from_bytes) throw error_already_set();
+            value = reinterpret_steal<type>(str_from_bytes);
+            return true;
+        }
+#endif
```
YannickJadoul, line 1636: Alternatively, if we don't want this here (cfr. @wjakob's remark), we could specialize `pyobject_caster<str>` or `type_caster<str>`?
But maybe that's too much for code that will disappear in a hopefully not all too distant future.

include/pybind11/stl.h
```diff
@@ -144,7 +144,7 @@ template <typename Type, typename Value> struct list_caster {
     using value_conv = make_caster<Value>;
 
     bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src) || isinstance<str>(src))
+        if (!isinstance<sequence>(src) || isinstance<bytes>(src) || isinstance<str>(src))
```
YannickJadoul, line 147: I think we actually don't want this change. Doing so (or well, not doing so) would fix #1807.

I'm OK with making and discussing this change (or undoing your change) in a separate PR, though.

YannickJadoul (approved):
A few minor details, but this looks good to me! As we've discussed in plenty of other issues/PRs before, I would argue that most or almost all use cases that will be affected can be considered bugs.
pybind11 is already pretty clear about the names `str` and `bytes` having "Python 3 semantics", I would say.

The one thing is this special case for Python 2 (and the changes necessary to `pyobject_caster`, as pointed out by @wjakob). The good news is Python 2 already does something similar here:

```pycon
[GCC 7.5.0] on linux2
>>> 'abc'
'abc'
>>> unicode('abc')
u'abc'
>>> print('\xc3\xb1'.decode('utf-8'))
ñ
>>> '\xc3\xb1'
'\xc3\xb1'
>>> unicode('\xc3\xb1')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
>>> 
```

So the current implementation matches behavior, except that it's using codec "ascii" instead of "utf-8". We could match this, but I'm also fine keeping UTF-8, since pybind11 is again very [clear on following UTF-8 everywhere](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#passing-python-strings-to-c).

I also want to quickly remind ourselves of why this `detail::PyUnicode_Check_Permissive` was implemented in the first place. I kind of remember we already dug back in history, at some point, but I forgot the conclusion. I'll see if I can find it and add a cross-reference from this PR.

include/pybind11/stl.h
YannickJadoul: See also #2198.

YannickJadoul:
> I'll see if I can find it and add a cross-reference from this PR.

OK, this is the best there is to be found, it seems: https://github.com/pybind/pybind11/commit/5612a0c1c22649865a3296d9acaa78f6b414c680
There's not a lot I can distill from that, except that pybind11 has changed a lot since then, and that the original author from back then (@wjakob) said he's OK with the current changes, so ... not too worried about this :-)

YannickJadoul:
Diving further with @bstaletic into this commit https://github.com/pybind/pybind11/commit/5612a0c1c22649865a3296d9acaa78f6b414c680, there's a piece of dead code, now:

https://github.com/pybind/pybind11/blob/17c22b9e0f56276f212b44f6cf0e6846a01623b2/include/pybind11/pytypes.h#L960-L972

This was introduced in 5612a0c1c22649865a3296d9acaa78f6b414c680, but the second part is now a dead end. Is this something to check with @wjakob, and maybe fix in a follow-up PR?

rwgk:
Thanks @YannickJadoul, sorry for the late reply. Just a quick note: I'll wait until after the 2.6.2 release before putting on a few finishing touches here (e.g. I'll try get rid of the dead code you pointed out).

YannickJadoul:
> Thanks @YannickJadoul, sorry for the late reply. Just a quick note: I'll wait until after the 2.6.2 release before putting on a few finishing touches here (e.g. I'll try get rid of the dead code you pointed out).

No worries. Sounds good to me. (Though, is there a 2.6.2 planned; I'm not sure we currently have a lot of bugfixes waiting for release?)
Should we go ahead and merge #2198 as a way to fix #1807 already in 2.6.2?

rwgk:
> Should we go ahead and merge #2198 as a way to fix #1807 already in 2.6.2?

We had a lot of back and forth about this particular change (me, you, @bstaletic taking different positions), which I why I'm preserving current behavior in this PR.

I'll comment on #2198 to log what I think is the best change.

YannickJadoul:
Well, yeah, I also noted it here: https://github.com/pybind/pybind11/pull/2409#discussion_r524663224

I'm not sure I see anything wrong with the change in #2198. I just though it had stalled because we figured out there was a much bigger issue.

rwgk:
@YannickJadoul, I changed "client code" to "user code" the first time I want to be unambiguous about what "transition" refers to, and delete the second mention of "client code" a couple lines down. You're right, the context is clear enough there.

A follow-up PR for the dead code you discovered would be nice. I'm leaving the code in the PR as it was for the past ~2.5 months. It has been in use internally for the entire time (and an earlier version even since August 2020).

PR #2414, opened 2020-08-19
fix: respect PYTHON_VERSION if set in classic mode

henryiii:
If PYBIND11_PYTHON_VERSION is not set, but PYTHON_VERSION is, on the first run, use that for the hint. We could add a warning or error instead - but the current behavior is a bit odd and unhelpful.

### DIFF
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -14,6 +14,9 @@ endif()
 
 # If this is the first run, PYTHON_VERSION can stand in for PYBIND11_PYTHON_VERSION
 if(NOT DEFINED PYBIND11_PYTHON_VERSION AND DEFINED PYTHON_VERSION)
+  message(WARNING "Set PYBIND11_PYTHON_VERSION to search for a specific version, not "
+                  "PYTHON_VERSION (which is an output). Assuming that is what you "
+                  "meant to do and continuing anyway.")
   set(PYBIND11_PYTHON_VERSION
       "${PYTHON_VERSION}"
       CACHE STRING "Python version to use for compiling modules")
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -12,10 +12,17 @@ if(pybind11_FIND_QUIETLY)
   set(_pybind11_quiet QUIET)
 endif()
 
-# Add a CMake parameter for choosing a desired Python version
-if(NOT PYBIND11_PYTHON_VERSION)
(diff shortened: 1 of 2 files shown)


tools/pybind11Tools.cmake
```diff
@@ -12,10 +12,20 @@ if(pybind11_FIND_QUIETLY)
   set(_pybind11_quiet QUIET)
 endif()
 
-# Add a CMake parameter for choosing a desired Python version
-if(NOT PYBIND11_PYTHON_VERSION)
+# If this is the first run, PYTHON_VERSION can stand in for PYBIND11_PYTHON_VERSION
+if(NOT DEFINED PYBIND11_PYTHON_VERSION AND DEFINED PYTHON_VERSION)
+  message(WARNING "Set PYBIND11_PYTHON_VERSION to search for a specific version, not "
+                  "PYTHON_VERSION (which is an output). Assuming that is what you "
+                  "meant to do and continuing anyway.")
```
henryiii, line 19: @bstaletic Something like this?

bstaletic: Exactly.

bstaletic (approved):
With the added warning, I'm fine with this. Thanks @henryiii

tools/pybind11Tools.cmake
```diff
@@ -12,10 +12,20 @@ if(pybind11_FIND_QUIETLY)
   set(_pybind11_quiet QUIET)
 endif()
 
-# Add a CMake parameter for choosing a desired Python version
-if(NOT PYBIND11_PYTHON_VERSION)
+# If this is the first run, PYTHON_VERSION can stand in for PYBIND11_PYTHON_VERSION
+if(NOT DEFINED PYBIND11_PYTHON_VERSION AND DEFINED PYTHON_VERSION)
```
YannickJadoul, line 16: So is this part only executed in "classic mode" (as the PR title suggests)?

I'm wondering if having set `PYBIND11_PYTHON_VERSION` or `PYTHON_VERSION` ought to give an error/warning when `FindPython` is used? (since it won't have any effect, right?)

YannickJadoul (approved):
Looks good, yes. More of a fix/convenience, indeed, so do merge when you're happy with it, AFAIC!

tools/pybind11Tools.cmake
henryiii: It will give an "unused variable specified" warning in CMake already.

YannickJadoul: Alright!

PR #2429, opened 2020-08-23
Adapt code example in advanced/classes.rst to new handling of forgetting to call the superclass __init__

YannickJadoul:
Cfr. discussion in #2152

TODO: Check how easy it is to display different error when there is no `__init__`.

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -159,7 +159,7 @@ Here is an example:
 
     class Dachshund(Dog):
         def __init__(self, name):
-            Dog.__init__(self) # Without this, undefined behavior may occur if the C++ portions are referenced.
+            Dog.__init__(self) # Without this, a TypeError is raised.
             self.name = name
         def bark(self):
             return "yap!"


bstaletic (approved):
I agree with @rwgk

docs/advanced/classes.rst
```diff
@@ -159,7 +159,7 @@ Here is an example:
```
YannickJadoul, line 159: @rwgk, @bstaletic, I was still wondering if the fragment before needs to be adapted: https://github.com/pybind/pybind11/pull/2429/files#diff-47ee721743764a0358552e687404eb8eR145-R150

It's not entirely right (because it doesn't happen anymore), but it's also not wrong. Is it good enough to keep like this?

rwgk: Good point, I agree, changing that sentence in this PR would be best.
Trying:

Otherwise a TypeError is raised for safety, to not leave the memory for the C++ portion of the instance uninitialized, which would generally cause undefined behavior if the instance is used subsequently.

YannickJadoul: My main problem was not how to formulate it, but how to best structure it with this `.. versionchanged:: 2.6` line, actually?

rwgk: Hm, seems fine to me, maybe with tiny tweaks?

> The default pybind11 metaclass was changed to throw a `TypeError` when it detects that `__init__` was not called by a derived class (PR #2152), while prior versions left the C++ portion of the instance uninitialized.

(I'd include the PR number for easy reference for those who know what to do with it; it's easily ignored for those who don't.)

YannickJadoul: > Otherwise a TypeError is raised for safety, to not leave the memory for the C++ portion of the instance uninitialized, which would generally cause undefined behavior if the instance is used subsequently.

Actually, reading this back and trying to fit in these changes, I think it's fine the way it's formulated now. I'm now realizing it's still important to state the whole thing about UB, because it's still the underlying problem (and is still crucial when a non-default metaclass is used). The only difference is that pybind11 now catches this in the default metaclass. So the structure seems fine to me:

> - Call subclass `__init__`
> - Because you need to initialize the underlying class.
> - But the default metaclass will help you (from 2.6 onwards).

Sorry for the confusion!

> (I'd include the PR number for easy reference for those who know what to do with it; it's easily ignored for those who don't.)

@henryiii already included this in the changelog! :-) I'm assuming that people that know what to do with a pybind11 PR would also find the changelog? (I also don't see the convention elsewhere in the docs, so it might be nice to constrain the development reference to the reflog.)

YannickJadoul:
> TODO: Check how easy it is to display different error when there is no `__init__`.

I've been trying out some stuff, but this seems quite hard, as the default `__init__` (which complains that no constructor has been defined) is set in `tp_init`, while dynamically added methods override it.

This is already quite ugly, but doesn't work in my tests:

```diff
diff --git a/include/pybind11/detail/class.h b/include/pybind11/detail/class.h
index 8d36744f..ecd0348b 100644
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -171,8 +171,16 @@ extern "C" inline PyObject *pybind11_meta_call(PyObject *type, PyObject *args, P
     // Ensure that the base __init__ function(s) were called
     for (const auto &vh : values_and_holders(instance)) {
         if (!vh.holder_constructed()) {
-            PyErr_Format(PyExc_TypeError, "%.200s.__init__() must be called when overriding __init__",
-                         vh.type->type->tp_name);
+            PyObject *self_init = PyObject_GetAttrString((PyObject*)vh.type->type, "__init__");
+            if (!self_init)
+                PyErr_Clear();
+            auto pybind11_object_init = PyObject_GetAttrString(get_internals().instance_base, "__init__");
+            auto message = (self_init == pybind11_object_init) ? 
+                           "%.200s has no __init__ and can therefore not be used as a base class in Python" :
+                           "%.200s.__init__() must be called when overriding __init__";
+            PyErr_Format(PyExc_TypeError, message, vh.type->type->tp_name);
+            Py_XDECREF(self_init);
+            Py_DECREF(pybind11_object_init);
             Py_DECREF(self);
             return nullptr;
         }
```

So yes, agreed, @bstaletic and @rwgk; let's just get this in and worry about the rest later :-)

PR #2430, opened 2020-08-23
Throw exception on returning a unique_ptr or shared_ptr nullptr (or any other holder type) from py::init, rather than crashing

YannickJadoul:
One overload of the implementation of `py::init` was missing this check for a null pointer that's present in the other two. This should fix it.

Crash found by @rgkw in https://github.com/pybind/pybind11/pull/2152#issuecomment-678816799.

### DIFF
--- a/include/pybind11/detail/init.h
+++ b/include/pybind11/detail/init.h
@@ -132,6 +132,7 @@ void construct(value_and_holder &v_h, Alias<Class> *alias_ptr, bool) {
 template <typename Class>
 void construct(value_and_holder &v_h, Holder<Class> holder, bool need_alias) {
     auto *ptr = holder_helper<Holder<Class>>::get(holder);
+    no_nullptr(ptr);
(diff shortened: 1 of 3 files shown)


YannickJadoul:
> Thanks Yannick for jumping in plugging this hole!

Got curious ;-)

Let's get one more review (@henriii, @bstaletic, or maybe @EricCousineau-TRI? Don't know who's familiar with this part of the code). Then we should be fine, given how minor this change is :-)

bstaletic (approved):
Looks fine to me.

YannickJadoul:
Ow, I had tagged the wrong @henryiii, in the [comment before](https://github.com/pybind/pybind11/pull/2430#issuecomment-679043209). But given @bstaletic's review, I'll merge tomorrow morning, if no one speaks up :-)

YannickJadoul:
OK, merging :-) If there's still a problem we overlooked, it can be fixed later.

PR #2431, opened 2020-08-23
Add note that VS2017 requires /permissive- to build in C++17 mode

### LINKED ISSUE #2089: Pybind fails to build with MSVC 19.16 in C++17 mode

YannickJadoul:
As discussed on #2089, this (finally) adds a note to the docs that C++17 mode in VS2017 requires the /permissive- flag.

Closes #2089
Closes #2292﻿

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,6 @@ jobs:
       matrix:
         runs-on: [ubuntu-latest, windows-latest, macos-latest]
         arch: [x64]
-        max-cxx-std: [17]
         python:
         - 2.7
         - 3.5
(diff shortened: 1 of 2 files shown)


YannickJadoul:
I can't find any other place to add this. I'd have thought this was something to add to the FAQ page, but no existing section seemed fitting. But I'm open to suggestions.

bstaletic:
Why not the README? It's where all the supported compilers are mentioned.

YannickJadoul:
> Why not the README? It's where all the supported compilers are mentioned.

The readme currently only contains the minimal version of the compilers (and very summarized; "Microsoft Visual Studio 2015 Update 3 or newer"). Neither VS 2017 nor C++17 are minimal requirements for pybind11, so I was hesitant to throw in those extra lines there? You can still compile perfectly in C++14 mode in VS 2017.

henryiii:
I've simplified the matrix by dropping a few extra arguments. We can still get the "max C++" setting by manually placing the CXX standard flag in args2. I'm about to make the setup a little more complicated by putting the dev flag back in.

YannickJadoul:
Amazing! Thanks again, @henryiii! :-)

PR #2438, opened 2020-08-26
Fix bug roundtripping datetime.time objects after midnight in eastern hemisphere timezones (#2417)

### LINKED ISSUE #2417: [BUG] Eastern hemisphere timezones break `std::chrono::time_point<std::chrono::system_clock, Duration>` caster
`test_chrono_system_clock_roundtrip_time` fails on my system between midnight and 1 o'clock.
Demonstrated by @henryiii in these tests (when run from a timezone in the eastern hemisphere): https://github.com/pybind/pybind11/compare/master...henryiii:fix/timebug

Further investigation shows that a negative number of microseconds is passed to `PyDateTime_FromDateAndTime` (in `cast`), resulting in an C API error that's not handled.

Current hypothesis of what goes wrong:
- Converting a Python `datetime.time` to C++ adds 1 January, 1970 as date in a `std::tm` struct.
- This is then converted to a `<chrono>` type through `system_clock::from_time_t(std::mktime(...))`, turning the clock back an hour to convert to UTC (I'm the CET time zone UTC+01:00 - ignoring summer time) .
- This results in a "negative" time. When converting back to Python, `(duration_cast<us_t>(src.time_since_epoch() % seconds(1))).count()` results in a negative number.

YannickJadoul:
Closes #2417; see description there of what goes wrong.

Many, many thanks to @henryiii and @bstaletic for helping with debugging and fixing this one!

EDIT: Added reviewable.io at @bstaletic's request. Let's give it a try?

---
This change is [Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438)

### DIFF
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from pybind11_tests import chrono as m
 import datetime
+import pytest
 
 
 def test_chrono_system_clock():
@@ -70,9 +71,17 @@ def test_chrono_system_clock_roundtrip_date():
     assert time2.microsecond == 0
 
 
-def test_chrono_system_clock_roundtrip_time():
-    time1 = datetime.datetime.today().time()
-
+@pytest.mark.parametrize("time1", [
+    datetime.datetime.today().time(),
+    datetime.time(0, 0, 0),
+    datetime.time(0, 0, 0, 1),
+    datetime.time(0, 28, 45, 109827),
+    datetime.time(0, 59, 59, 999999),
+    datetime.time(1, 0, 0),
+    datetime.time(5, 59, 59, 0),
+    datetime.time(5, 59, 59, 1),
+])
+def test_chrono_system_clock_roundtrip_time(time1):
     # Roundtrip the time
     datetime2 = m.test_chrono2(time1)
     date2 = datetime2.date()
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -3,6 +3,8 @@ from pybind11_tests import chrono as m
 import datetime
 import pytest
 
+import env  # noqa: F401
+
 
 def test_chrono_system_clock():
 
@@ -71,6 +73,11 @@ def test_chrono_system_clock_roundtrip_date():
     assert time2.microsecond == 0
 
 
+SKIP_TZ_ENV_ON_WIN = pytest.mark.skipif(
+    "env.WIN", reason="TZ environment variable only supported on POSIX"
+)
+
+
 @pytest.mark.parametrize("time1", [
     datetime.datetime.today().time(),
     datetime.time(0, 0, 0),
@@ -82,12 +89,14 @@ def test_chrono_system_clock_roundtrip_date():
     datetime.time(5, 59, 59, 1),
 ])
 @pytest.mark.parametrize("tz", [
-    "Europe/Brussels",
-    "Asia/Pyongyang",
-    "America/New_York",
+    None,
+    pytest.param("Europe/Brussels", marks=SKIP_TZ_ENV_ON_WIN),
+    pytest.param("Asia/Pyongyang", marks=SKIP_TZ_ENV_ON_WIN),
+    pytest.param("America/New_York", marks=SKIP_TZ_ENV_ON_WIN),
 ])
 def test_chrono_system_clock_roundtrip_time(time1, tz, monkeypatch):
-    monkeypatch.setenv("TZ", "/usr/share/zoneinfo/{}".format(tz))
+    if tz is not None:
+        monkeypatch.setenv("TZ", "/usr/share/zoneinfo/{}".format(tz))
 
     # Roundtrip the time
     datetime2 = m.test_chrono2(time1)
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -81,7 +81,14 @@ def test_chrono_system_clock_roundtrip_date():
     datetime.time(5, 59, 59, 0),
     datetime.time(5, 59, 59, 1),
 ])
-def test_chrono_system_clock_roundtrip_time(time1):
+@pytest.mark.parametrize("tz", [
+    "Europe/Brussels",
+    "Asia/Pyongyang",
+    "America/New_York",
+])
+def test_chrono_system_clock_roundtrip_time(time1, tz, monkeypatch):
+    monkeypatch.setenv("TZ", "/usr/share/zoneinfo/{}".format(tz))
+
     # Roundtrip the time
     datetime2 = m.test_chrono2(time1)
     date2 = datetime2.date()
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -150,13 +150,22 @@ public:
         // Lazy initialise the PyDateTime import
         if (!PyDateTimeAPI) { PyDateTime_IMPORT; }
 
-        std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src));
+        // Declare these special duration types so the conversions happen with the correct primitive types (int)
+        using us_t = duration<int, std::micro>;
+
+        // Get out microseconds, and make sure they are positive, to avoid bug in eastern hemisphere time zones
+        // (cfr. https://github.com/pybind/pybind11/issues/2417)
+        auto us = duration_cast<us_t>(src.time_since_epoch() % seconds(1));
+        if (us.count() < 0)
+            us += seconds(1);
+
+        // Subtract microseconds BEFORE `system_clock::to_time_t`, because:
+        // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
+        // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
+        std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
         // this function uses static memory so it's best to copy it out asap just in case
         // otherwise other code that is using localtime may break this (not just python code)
         std::tm localtime = *std::localtime(&tt);
-
-        // Declare these special duration types so the conversions happen with the correct primitive types (int)
-        using us_t = duration<int, std::micro>;
 
         return PyDateTime_FromDateAndTime(localtime.tm_year + 1900,
                                           localtime.tm_mon + 1,
@@ -164,7 +173,7 @@ public:
                                           localtime.tm_hour,
                                           localtime.tm_min,
                                           localtime.tm_sec,
-                                          (duration_cast<us_t>(src.time_since_epoch() % seconds(1))).count());
+                                          us.count());
     }
     PYBIND11_TYPE_CASTER(type, _("datetime.datetime"));
 };
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -150,11 +150,9 @@ public:
         // Lazy initialise the PyDateTime import
         if (!PyDateTimeAPI) { PyDateTime_IMPORT; }
 
-        // Declare these special duration types so the conversions happen with the correct primitive types (int)
-        using us_t = duration<int, std::micro>;
-
         // Get out microseconds, and make sure they are positive, to avoid bug in eastern hemisphere time zones
         // (cfr. https://github.com/pybind/pybind11/issues/2417)
+        using us_t = duration<int, std::micro>;
         auto us = duration_cast<us_t>(src.time_since_epoch() % seconds(1));
         if (us.count() < 0)
             us += seconds(1);
--- a/tests/test_chrono.cpp
+++ b/tests/test_chrono.cpp
@@ -10,6 +10,7 @@
 
 #include "pybind11_tests.h"
 #include <pybind11/chrono.h>
+#include <chrono>
 
 TEST_SUBMODULE(chrono, m) {
     using system_time = std::chrono::system_clock::time_point;


bstaletic (commented):
First of all, thanks for giving Reviewable a chance. Plain github reviews get really annoying (to me) as soon as pulls are not trivial.

With @yannickjadoul's permission, I did take the opportunity to talk about some basic Reviewable features. Everyone reviewing this PR, please open any of the reviewable links in this PR and review it there. (The github comments will appear in the main thread in reviewable. Reviewable comments do appear in github, but might also appear on the github diff page.)

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438#-:-MFgtUIr8-dgFRcmZZNo:brc77j2)* status: all files reviewed, 3 unresolved discussions (waiting on @henryiii and @YannickJadoul)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFguxGkBIDdNapT86Cs:blv1n2e) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
> ```objc
>         if (!PyDateTimeAPI) { PyDateTime_IMPORT; }
> 
>         // Declare these special duration types so the conversions happen with the correct primitive types (int)
> ```

I know this comment was here before, but to me it's more confusing than helpful. Should we remove it?

Non-blocking, because it was there. (Reviewable has these little buttons where you can see/set "disposition" about others'/your comments.)
___
*[include/pybind11/chrono.h, line 160 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgw_on9fVsCwonIzuB:-MFgw_oo4CqW6QaV4Z7Z:b-1d7bn0) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L160)):*
> ```objc
>         auto us = duration_cast<us_t>(src.time_since_epoch() % seconds(1));
>         if (us.count() < 0)
>             us += seconds(1);
> ```

Is this actually correct? If we had produced `-3us`, then adding `1s` produces `999'997us`. Do we want that or `3us`?
___
*[include/pybind11/chrono.h, line 165 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgxHKJ4leyTDU6XYx7:-MFgxHKJ4leyTDU6XYx8:b7aomb6) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L165)):*
> ```objc
>         // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
>         // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
>         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
> ```

The above comment might be a good enough reason to completely avoid the C library and rely completely on Chrono.

YannickJadoul (commented):
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438)* status: all files reviewed, 3 unresolved discussions (waiting on @bstaletic and @henryiii)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFjtktu1OrG3t1Cr1bG:b-34xmpj) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
<details><summary><i>Previously, bstaletic (Boris Staletic) wrote…</i></summary><blockquote>

I know this comment was here before, but to me it's more confusing than helpful. Should we remove it?

Non-blocking, because it was there. (Reviewable has these little buttons where you can see/set "disposition" about others'/your comments.)
</blockquote></details>

To be honest, I just kept this, indeed. From the comment, I believe the type of the microseconds argument from [`PyDateTime_FromDateAndTime`](https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_FromDateAndTime) is matched.
I'm not a `<chrono>` expert, as you know, but it does sound like there was a reason to create this type, so I'm hesitant to just remove it?
___
*[include/pybind11/chrono.h, line 160 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgw_on9fVsCwonIzuB:-MFjuiM6Di57tULU_Zau:bjraklv) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L160)):*
<details><summary><i>Previously, bstaletic (Boris Staletic) wrote…</i></summary><blockquote>

Is this actually correct? If we had produced `-3us`, then adding `1s` produces `999'997us`. Do we want that or `3us`?
</blockquote></details>

I think it is. What's happening is that a time on the eastern hemisphere becomes a date + time from before 1970 in UTC (though `std::mktime` in the `load` method, which interprets a `tm` struct as local time). At that point we still add the microseconds which is always a positive amount in Python's `datetime`, but the total stays negative. So we need to take the modulo of the time vs. seconds, but make sure it's positive

Example:

- Suppose it's 1 minute, 2 seconds, 3 microseconds after midnight, in GMT+1. That's 62 000 003 µs after midnight. So without date we just add a dummy 1970/01/01 (because Python has separate types for date/time/datetime, but C++ not), and it becomes 1970/01/01 00:01:02.000003 (or, 62 000 003 µs after the UNIX epoch).
- `std::mktime` converts that GMT+1 date and time to UTC. Which becomes 1969-12-31 23:01. So, that's 3537999997 before UNIX epoch or -3 537 999 997 after UNIX epoch (=  62 000 003 µs - 3 600 000 000 µs (= 1 hour)).
- Converting back, you take out the microseconds, so -3 537 999 997 µs % 1000 000 = -999 997 µs. But Python wants a positive amount of microseconds, so we need to make it positive, so you add another second (1 000 000 µs). Just taking the absolute value, 999 997 µs, would not give you the thing you put in, because things work differently in the negative numbers (i.e.,  -3 537 999 997 is 3 MORE than -3 538 000 000, so it's -3 538 s + 3 µs).
- Then you subtract it, because we first added it, and because "If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.". But after subtracting the amount of microseconds, you know it's now divisible by 1000 000 (i.e., a second), and ready for `tm` and `localtime` again (which will convert the "almost an hour before 1970" back to our local GMT+1 time).
___
*[include/pybind11/chrono.h, line 165 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgxHKJ4leyTDU6XYx7:-MFjzd5I9H1LdSmLAOyl:b-ebjhuk) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L165)):*
<details><summary><i>Previously, bstaletic (Boris Staletic) wrote…</i></summary><blockquote>

The above comment might be a good enough reason to completely avoid the C library and rely completely on Chrono.
</blockquote></details>

Yes, if we can, that would be great, agreed! But time and calenders are horribly complex.

The only thing I found (when we were looking into this together, yesterday evening), would be to add years, months, days to 1970. But this is not the same, because they do not take the calendar into account: a duration of a month or year is defined as the average time, independent of which point in time it's being added to (I thínk; happy to be proven wrong!).

Example: the difference between the first of January and the first of February is 1 "calendar month", just like the difference between the first of February and the first of March. But in duration, one is 31 days, the other 28 (or 29) days. So you can't just add the average duration and hope to get at the right date. Because if you would add the average duration of a month (30.42 days in a regular year) to the 1st of January, you'd end up somewhere just before noon on the 31st of January, and not midnight the 1st of February. That's why adding calendar-months/years is "context-dependent" (by lack of a better term). I don't think `std::chrono::duration` takes that into account (but I could be wrong).

Happy to be proven wrong on this, though. If `<chrono>` handles this, I'd very gladly ditch C's old structs and functions!

YannickJadoul (commented):
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438)* status: all files reviewed, 3 unresolved discussions (waiting on @bstaletic and @henryiii)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFk5ZS53hmcax-7EVER:b9g2a33) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
<details><summary><i>Previously, henryiii (Henry Schreiner) wrote…</i></summary><blockquote>

You need these, but they are pretty standard things to add for chrono, so not sure if the comment is all that helpful.
</blockquote></details>

OK, I'll remove the comment, then!
___
*[include/pybind11/chrono.h, line 165 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgxHKJ4leyTDU6XYx7:-MFk5dKOCfAtE2oMoDvD:b-c9k4k5) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L165)):*
<details><summary><i>Previously, henryiii (Henry Schreiner) wrote…</i></summary><blockquote>

We already use some things from chrono, so it seems like we shouldn't be trying to mix?
</blockquote></details>

It would be nice, yes. Do you know how to convert a "calendar" time stamp to a chrono time_point (taking leap years and month lengths etc into account) ? I'd love to ditch the old stuff and go full modern C++ :-)
___
*[tests/test_chrono.py, line 83 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFk2zd38wYCKChUpphK:-MFk5xllB_VWl5AQZNLU:b-71kw58) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/tests/test_chrono.py#L83)):*
<details><summary><i>Previously, henryiii (Henry Schreiner) wrote…</i></summary><blockquote>

This doesn't actually fail when run without the changes to pybind11. Is there a way to trigger this failure?
</blockquote></details>

Setting your timezone to UTC/GMT+x will make all tests between 00:00:00 and 0x:00:00 fail, if they have a non-zero amount of milliseconds (otherwise, the bug doesn't get triggered).

The example on https://en.cppreference.com/w/cpp/chrono/c/mktime seems to show you can change the `TZ` environment variable on POSIX? `setenv("TZ", "/usr/share/zoneinfo/America/New_York", 1); // POSIX-specific`
So if you want to join me here, `"/usr/share/zoneinfo/Europe/Brussels"` would hopefully demonstrate.

YannickJadoul (commented):
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @bstaletic and @henryiii)

___
*[tests/test_chrono.py, line 83 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFk2zd38wYCKChUpphK:-MFk8wzgCC0_mTI5eSlL:b-b0nn2u) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/tests/test_chrono.py#L83)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

Setting your timezone to UTC/GMT+x will make all tests between 00:00:00 and 0x:00:00 fail, if they have a non-zero amount of milliseconds (otherwise, the bug doesn't get triggered).

The example on https://en.cppreference.com/w/cpp/chrono/c/mktime seems to show you can change the `TZ` environment variable on POSIX? `setenv("TZ", "/usr/share/zoneinfo/America/New_York", 1); // POSIX-specific`
So if you want to join me here, `"/usr/share/zoneinfo/Europe/Brussels"` would hopefully demonstrate.
</blockquote></details>

Oh, yes. And to get the `5, 59, 59, 1` to fail, you need to be over in `"/usr/share/zoneinfo/Asia/Pyongyang"` or `"/usr/share/zoneinfo/Asia/Tokyo"` or `"/usr/share/zoneinfo/Australia/Sydney"` or ... well, your favorite holiday destination that's at least GMT+6 on the first of January.

henryiii:
Oops! Sorry, didn't mean to push to this branch. 😳

tests/test_chrono.py
```diff
@@ -81,7 +81,14 @@ def test_chrono_system_clock_roundtrip_date():
     datetime.time(5, 59, 59, 0),
     datetime.time(5, 59, 59, 1),
 ])
-def test_chrono_system_clock_roundtrip_time(time1):
+@pytest.mark.parameterize("tz", [
+    "/usr/share/zoneinfo/Europe/Brussels",
+    "/usr/share/zoneinfo/Asia/Pyongyang",
+    "/usr/share/zoneinfo/America/New_York",
+])
```
YannickJadoul, line 88: Windows won't be a fan of this, though? Though I guess it won't hurt either...

henryiii: It shouldn't really do anything on Windows, I would expect.

henryiii:
Here's what I get before:

```
test_chrono.py .....FFF.....FFF..F...............
```

Looks like I pushed the change to chrono.h again (too much `git add -u`), so let's wait to revert that change until the tests fail.

henryiii:
Okay, I'm done force pushing. I'm happy with the tests failing before but not after (at least locally). :) Great catch and fix!

tests/test_chrono.py
YannickJadoul: Except for fooling us (and others) into thinking it ís tested, and running the same tests 3 times. But yes, OK.

We could add `pytest.param(..., marks=pytest.mark.skipif("...", reason="TZ environment variable not supported on Windows")),`?

YannickJadoul (commented):
Thanks for strengthening these tests, @henryiii! Better to have this fail everywhere, rather than just on this side of the Atlantic ;-)

Another thing I'm still worried about: time still gets turned back/forward to be stored in the chrono types, depending on the time zone. Should we be doing that? What's the meaning of Python's datetime wrt. timezones? (might be a separate issue though; this bugfix is still valid, independently)

> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438#-:-MFkG0ySBsMwTwX3gEU-:betx7px)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @bstaletic, @henryiii, and @YannickJadoul)

___
*[include/pybind11/chrono.h, line 165 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgxHKJ4leyTDU6XYx7:-MFkG3KHB5b6XyP6xM7f:b-c8eotu) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L165)):*
<details><summary><i>Previously, henryiii (Henry Schreiner) wrote…</i></summary><blockquote>

You mean in C++20? https://en.cppreference.com/w/cpp/chrono#Calendar
</blockquote></details>

Ah, another pybind20 feature ;-)
___
*[tests/test_chrono.py, line 83 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFk2zd38wYCKChUpphK:-MFkGHRX245qa2p-aN-v:b-ou14eb) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/tests/test_chrono.py#L83)):*
<details><summary><i>Previously, henryiii (Henry Schreiner) wrote…</i></summary><blockquote>

Okay, added to tests (fixed previous unintentional push). Can you check it locally without your change to pybind11's chrono to verify it partially fails there?
</blockquote></details>

OK, will do!

henryiii:
We need to fix that Windows time length...

YannickJadoul:
> We need to fix that Windows time length...

What do you mean?

henryiii:
> What do you mean?

The CI flake issue where windows sometimes reports a slightly larger time difference. Doesn't need to be in this PR, but we should fix it in the near future.

bstaletic (commented):
Mostly, :lgtm:. Only one comment regarding Windows/`TZ`

> Reviewed 1 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438#-:-MFldIfxCDASlzaz1ZQc:b-2s0nxx)* status: all files reviewed, 4 unresolved discussions (waiting on @henryiii and @YannickJadoul)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFlbLyUAx7QJ_tngnbP:bo56dwz) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

OK, I'll remove the comment, then!
</blockquote></details>

The comment wasn't removed. Just commenting because Reviewable thinks this is resolved, since @yannickjadoul started the previous comment with "OK", which is a magic word in Reviewable.
___
*[include/pybind11/chrono.h, line 165 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgxHKJ4leyTDU6XYx7:-MFlacn8-BRtrk5IDMkt:b-93sl72) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L165)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

Ah, another pybind20 feature ;-)
</blockquote></details>

Yeah, there's no `std::chrono::year` in C++11... I'm sure ditching the C library and going full C++11 Chrono is possible, but might not be very friendly to maintainers. Happy to leave that part for another time.
___
*[tests/test_chrono.py, line 88 at r3](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgsZKy0cfd-lJwRGAd-r3-88:-MFlcrKnBVGiBsUawMA_:b-xffbem) ([raw file](https://github.com/pybind/pybind11/blob/af8efcf0f239a8894b0d1a8d79140941111b0378/tests/test_chrono.py#L88)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

Except for fooling us (and others) into thinking it ís tested, and running the same tests 3 times. But yes, OK.

We could add `pytest.param(..., marks=pytest.mark.skipif("...", reason="TZ environment variable not supported on Windows")),`?
</blockquote></details>

That `skipif` is probably a good idea.

YannickJadoul (commented):
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438)* status: all files reviewed, 4 unresolved discussions (waiting on @bstaletic and @YannickJadoul)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFle5wj1Db23lNjkZSO:b-e67ta1) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
<details><summary><i>Previously, bstaletic (Boris Staletic) wrote…</i></summary><blockquote>

The comment wasn't removed. Just commenting because Reviewable thinks this is resolved, since @yannickjadoul started the previous comment with "OK", which is a magic word in Reviewable.
</blockquote></details>

HEEEEEY, can I decide on my own magic words, Reviewable? :-D

YannickJadoul (commented):
Think I replied to everything :-)
What to do now, @bstaletic, with Reviewable? Am I waiting for another review of the final changes? Do I need to "LGTM" my own PR? Any other magic buttons to click or try? (e.g., what's up with this "Mark as reviewed'?)

> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438#-:-MFlxLVDAuOzlDigMKZa:bukuzmd)* status: all files reviewed, 2 unresolved discussions (waiting on @bstaletic and @henryiii)

___
*[include/pybind11/chrono.h, line 153 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFguxGjCKLRUV3hxzq7:-MFlvqDt15psCQUdp00p:bb6un1j) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/include/pybind11/chrono.h#L153)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

HEEEEEY, can I decide on my own magic words, Reviewable? :-D
</blockquote></details>

Done
___
*[tests/test_chrono.py, line 83 at r1](https://reviewable.io/reviews/pybind/pybind11/2438#-MFk2zd38wYCKChUpphK:-MFlx48q3F-xugWXyKzA:bjcbebf) ([raw file](https://github.com/pybind/pybind11/blob/4cca28001c01d3fa2d3ebdd940c7aa9ea044e107/tests/test_chrono.py#L83)):*
<details><summary><i>Previously, YannickJadoul (Yannick Jadoul) wrote…</i></summary><blockquote>

OK, will do!
</blockquote></details>

Done! Also fails for me (except in New York), without the changes, so thanks for turning these tests deterministic!
___
*[tests/test_chrono.py, line 88 at r3](https://reviewable.io/reviews/pybind/pybind11/2438#-MFgsZKy0cfd-lJwRGAd-r3-88:-MFlvn8O5R5PEFob3iIK:bb6un1j) ([raw file](https://github.com/pybind/pybind11/blob/af8efcf0f239a8894b0d1a8d79140941111b0378/tests/test_chrono.py#L88)):*
<details><summary><i>Previously, bstaletic (Boris Staletic) wrote…</i></summary><blockquote>

That `skipif` is probably a good idea.
</blockquote></details>

Done

bstaletic (commented):
Definitely looks good. :lgtm_strong: 

Two things about reviewable now.

1. lgtm_strong isn't invalidated after a push.
2. The one making a blocking comment needs to acknowledge that the change has been made before the discussion is resolved. In this case there's one unresolved discussion because @henryiii didn't ACK.

> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/pybind/pybind11/2438#-:-MFpAua47sIorqPPayae:b79o8q1)* status: all files reviewed, 1 unresolved discussion (waiting on @henryiii)

YannickJadoul:
Hurray, we managed working with Reviewable! :-D

Issue #2263, opened 2020-06-24
docstring type hint incorrect for argument depending on class declaration order


ixje:
## Issue description

The generated type-hint for the argument in the initialiser of class B (see below) is invalid and shows the internal namespace instead of the module. Output of `main.py` (again see below)
```
__init__(...)
    __init__(self: cmake_example.B, arg0: testns::A) -> None
```  

Note that the code is functional (although useless). To get the correct type hints we have to define `py::class_<A>(m, "A")` before `py::class_<B>(m, "B")`
```
__init__(...)
    __init__(self: cmake_example.B, arg0: cmake_example.A) -> None
```

**expected behaviour**
I'd expected to see type hints resolved correctly, regardless of class declaration order.

## Reproducible example code

cmakelists.txt
```cmake
cmake_minimum_required(VERSION 2.8.12)
project(cmake_example)

set(PYTHON_EXECUTABLE ${CMAKE_CURRENT_SOURCE_DIR}/venv/bin/python3)

include(FetchContent)
FetchContent_Declare(
        pybind11
        GIT_REPOSITORY https://github.com/pybind/pybind11.git
        GIT_TAG        v2.5.0
)
FetchContent_MakeAvailable(pybind11)
pybind11_add_module(cmake_example src/main.cpp)
```
main.cpp
```cpp
#include <pybind11/pybind11.h>

namespace py = pybind11;
namespace testns {
    class A {
        A() =default;
    };

    class B {
    public:
        B(A* a) {}
    };
};

using namespace testns;


PYBIND11_MODULE(cmake_example, m) {
    py::class_<B>(m, "B")
            .def(py::init([](A* a) {
                return B(a);
            }));

    py::class_<A>(m, "A")
            .def(py::init());
}
```
main.py
```python
import cmake_example

if __name__ == '__main__':
    print(help(cmake_example.B.__init__))
```

bstaletic:
I don't think this is possible to resolve. While in the `def(py::init(...))`, you have no idea if the user will ever bind `A`. Even if the user will do that, how do you know that `testns::A` will be `cmake_example.A`? Maybe the user does `py::class_<A>(m, "Z")`, in which case the correct docstring would contain `cmake_example.Z`. I also don't see a way to "channel back" the information about `A` bindings to `B` (the key word being "back").

ixje:
That's too bad. I was hoping for some sort of second pass to be able to solve it.

YannickJadoul:
[...]

PR #2441, opened 2020-08-27
[DOC] avoid C++ types in docstrings

sizmailov:
Here is a small paragraph about docstring generation in pybind11. Also replicates @YannickJadoul 's example from #2263. 

I'm not sure if `misc.rst` is the best place for it. 

<details><summary>render</summary>

image

</details>

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -322,7 +322,8 @@ its C++ type name will be used instead to generate the signature in the docstrin
                                               ^^^^^^^
 
 
-This limitation can be circumvented by "forward declaring" bound classes:
+This limitation can be circumvented by ensuring that C++ classes are registered with pybind11
+before they are used as a parameter to a function:
 
 .. code-block:: cpp
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -311,7 +311,7 @@ Avoiding C++ types in docstrings
 ================================
 
 Docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
-At this point argument and return types should be known to pybind11.
+At this point parameter and return types should be known to pybind11.
 If a custom type is not exposed yet through a ``py::class_`` constructor or a custom type caster,
 its C++ type name will be used instead to generate the signature in the docstring:
 
@@ -323,7 +323,7 @@ its C++ type name will be used instead to generate the signature in the docstrin
 
 
 This limitation can be circumvented by ensuring that C++ classes are registered with pybind11
-before they are used as a parameter to a function:
+before they are used as a parameter or return type of a function:
 
 .. code-block:: cpp
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -305,13 +305,15 @@ the default settings are restored to prevent unwanted side effects.
 .. [#f4] http://www.sphinx-doc.org
 .. [#f5] http://github.com/pybind/python_example
 
-.. _avoid-cpp-types-in-docstrings:
+.. _avoiding-cpp-types-in-docstrings:
 
-Avoid C++ types in docstrings
-=============================
+Avoiding C++ types in docstrings
+================================
 
-Most of docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
-At this point arguments/return types should be known to pybind11, otherwise C++ types would be used instead:
+Docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
+At this point argument and return types should be known to pybind11.
+If a custom type is not exposed yet through a ``py::class_`` constructor or a custom type caster,
+its C++ type name will be used instead to generate the signature in the docstring:
 
 .. code-block:: text
 
@@ -320,7 +322,7 @@ At this point arguments/return types should be known to pybind11, otherwise C++ 
                                               ^^^^^^^
 
 
-This limitation can be tolerated by "forward declaring" bound classes:
+This limitation can be circumvented by "forward declaring" bound classes:
 
 .. code-block:: cpp
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -304,3 +304,31 @@ the default settings are restored to prevent unwanted side effects.
 
 .. [#f4] http://www.sphinx-doc.org
 .. [#f5] http://github.com/pybind/python_example
+
+.. _avoid-cpp-types-in-docstrings:
+
+Avoid C++ types in docstrings
+=============================
+
+Most of docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
+At this point arguments/return types should be known to pybind11, otherwise C++ types would be used instead:
+
+.. code-block:: text
+
+     |  __init__(...)
+     |      __init__(self: example.Foo, arg0: ns::Bar) -> None
+                                              ^^^^^^^
+
+
+This limitation can be tolerated by "forward declaring" bound classes:
+
+.. code-block:: cpp
+
+    PYBIND11_MODULE(example, m) {
+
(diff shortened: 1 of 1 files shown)


docs/advanced/misc.rst
```diff
@@ -304,3 +304,31 @@ the default settings are restored to prevent unwanted side effects.
 
 .. [#f4] http://www.sphinx-doc.org
 .. [#f5] http://github.com/pybind/python_example
+
+.. _avoid-cpp-types-in-docstrings:
+
+Avoid C++ types in docstrings
+=============================
+
+Most of docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
```
YannickJadoul, line 313: Are there docstrings that aren't?

docs/advanced/misc.rst
```diff
@@ -304,3 +304,31 @@ the default settings are restored to prevent unwanted side effects.
 
 .. [#f4] http://www.sphinx-doc.org
 .. [#f5] http://github.com/pybind/python_example
+
+.. _avoid-cpp-types-in-docstrings:
+
+Avoid C++ types in docstrings
```
YannickJadoul, line 310: Very minors: maybe "Avoiding ...", since this is on how to avoid those types, rather than "don't do this"?

docs/advanced/misc.rst
```diff
@@ -304,3 +304,31 @@ the default settings are restored to prevent unwanted side effects.
...
+=============================
+
+Most of docstrings are generated at the time of the declaration, e.g. when ``.def(...)`` is called.
+At this point arguments/return types should be known to pybind11, otherwise C++ types would be used instead:
```
YannickJadoul, line 314: Maybe "argument and return types"?

What about: "... be known to pybind11, otherwise C++ types would be used > "...  be known to pybind11. If a custom type is not exposed yet through a `py::class_` constructor or a custom type caster, its C++ type name will be used instead to generate the signature in the docstring:"

docs/advanced/misc.rst
```diff
@@ -304,3 +304,31 @@ the default settings are restored to prevent unwanted side effects.
...
+
+.. code-block:: text
+
+     |  __init__(...)
+     |      __init__(self: example.Foo, arg0: ns::Bar) -> None
+                                              ^^^^^^^
+
+
+This limitation can be tolerated by "forward declaring" bound classes:
```
YannickJadoul, line 323: "can be circumvented"? (or potentially "can be worked around", but I think "circumvented" is the expression)

YannickJadoul (approved):
A few things I would personally phrase slightly differently (up to you to decide if you like it better or not), but looks succinct and good to me!

`misc.rst` seems find. I don't immediately see another place where we talk about docstrings and signature. It's right underneath the only other section about it ("Generating documentation using Sphinx"), so perfect for me!

Thanks for taking the time to create this, @sizmailov! :-)

docs/advanced/misc.rst
sizmailov: `Enum.__doc__` is read only property, so it's content evaluated dynamically. I thought there was something else...

sizmailov: Looks like enum is the only exception and probably it's irrelevant in this context. Let's remove "Most of " part ?

YannickJadoul: Right, OK, yes! Just curious.

What about "Almost all docstrings", then, to emphasize this is practically always?

sizmailov: (ok, py::module is a another one, but again, it's irrelevant enough to not mention)

docs/advanced/misc.rst
sizmailov: Sounds good to me!

YannickJadoul:
Great, thanks! Let's see if someone else still has something to add. If not, this can go in as-is, afaic :-)

docs/advanced/misc.rst
```diff
@@ -304,3 +304,33 @@ the default settings are restored to prevent unwanted side effects.
...
+its C++ type name will be used instead to generate the signature in the docstring:
+
+.. code-block:: text
+
+     |  __init__(...)
+     |      __init__(self: example.Foo, arg0: ns::Bar) -> None
+                                              ^^^^^^^
+
+
+This limitation can be circumvented by "forward declaring" bound classes:
```
bstaletic, line 325: I don't like the use of "forward declaration" here. How about

> ...by separating the bindings of the class itself from the bindings of the respective class' contents:

It's just that "forward declaration" has a very specific meaning in C++ and this isn't talking about that.

sizmailov: That's why it's in quotes. Quotes makes reader to wonder what that means and turn to example to understand. I don't think analogy to _forward declaration in C++_ creates a confusion to reader. 

I had to re-read the suggested replacement to understand what it means (because "bindings of the class" at first read includes "class content bindings" as well). It takes me 5 more seconds to think about before I look at example and finally understand.

While you are formally correct, I think it's better to keep it close to current form.

virtuald: How about instead

> ... by ensuring that C++ classes are registered with pybind11 before they are used as a parameter to a function

bstaletic: I don't think quotes is enough of a hint for "nevermind terminology, look at the example". I know my suggestion isn't perfect either.

@virtuald's suggestion reads better in my opinion, but I would be happy to hear other opinions.

sizmailov: I like @virtuald's suggestion.

YannickJadoul: I had noticed this, but couldn't immediately think of a better way of phrasing it either. So I let this go and thought the analogy does kind of make sense.

But @bstaletic's or @virtuald's also both sound good to me. The main problem is we don't seem to have a fixed terminology.

To match the docs [here](https://pybind11.readthedocs.io/en/stable/classes.html), and combine all suggestions, what about:

> ... by creating the class bindings before creating bindings for functions or methods with a signature that contains the corresponding C++ class:

or (but this is probably too long)

> ... by creating the class bindings (and thus registering them with pybind11) before creating bindings for functions or methods with a signature that contains the corresponding C++ class:

I guess the "forward declaration" becomes clear from the example, then?
Potentially, we could add a method binding that isn't just the constructor? And/or a binding for a free-standing function?

YannickJadoul: (Kind of [relevant xkcd](https://xkcd.com/927/)? "Here's another suggestion.")

sizmailov: I think the goal is to make it easy to read. The phrase "class bindings" usually includes binding of class members/functions in pybind11 docs. Therefore I find wordinig that use it more confusing than those that doesn't. 

I've update the PR with initial @virtuald's wording which seems to fine to everybody.

YannickJadoul: I was slightly afraid that "registering with pybind11" would not be clear to users, but if it's fine with @bstaletic, fine with me as well.

Also, to be really pedantic (we're here anyway :smile:): "as a parameter or return type of a function" ? I promise that's my last remark/request, and that I'll press "merge" after @bstaletic then approves ;-)

sizmailov: Ok, I've updated the PR. 
(Please squash all the commits on merge to avoid history pollution)

YannickJadoul: Thanks!
And jep, that's pybind11's default, these days :-)

YannickJadoul: @bstaletic: last call. We're all good now?

bstaletic (approved):
@sizmailov Thanks for the PR

sizmailov:
Thanks everybody for quick and thoughtful reviews with great attention to even subtle things!

YannickJadoul:
More like, thanks @sizmailov for making this PR and fixing all our minor remarks ;-) Thanks a lot!

PR #2448, opened 2020-08-30
ci: harden chrono test, mark another macos 4.9 dev failure

henryiii:
This should help with a little of the flakiness seen with the timing test

### DIFF
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -25,7 +25,7 @@ def test_chrono_system_clock():
 
     # Since datetime.datetime.today() calls time.time(), and on some platforms
     # that has 1 second accuracy, we compare this way
-    assert int(diff.seconds) <= int(diff_python.seconds)
+    assert diff.seconds <= diff_python.seconds
 
 
 def test_chrono_system_clock_roundtrip():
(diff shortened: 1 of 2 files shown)


YannickJadoul (approved):

tests/test_chrono.py
```diff
@@ -9,21 +9,23 @@
...
     diff = abs(date1 - date2)
 
     # There should never be a days difference
     assert diff.days == 0
 
     # Since datetime.datetime.today() calls time.time(), and on some platforms
-    # that has 1 second accuracy, we should always be less than 2 seconds.
-    assert diff.seconds < 2
+    # that has 1 second accuracy, we compare this way
+    assert int(diff.seconds) <= int(diff_python.seconds)
```
YannickJadoul, line 28: Why `int` here, though?

henryiii: I didn't check to make sure it was forced to be an int already. Is it?

henryiii: I think it is. The docs say it is `in range(60)`.

henryiii, line 28:
```suggestion
    assert diff.seconds <= diff_python.seconds
```

YannickJadoul: OK. The idea was mainly that even if it was something else, `<=` still ought to work? :-) But good!

PR #2449, opened 2020-08-31
docs: fix broken README link

dsaxton:
https://github.com/pybind/pybind11/pull/2402 moved the contributing docs but the link in the README seems to point to the old location


YannickJadoul:
Thanks for noticing and turning this into a PR, @dsaxton! :-)

I guess I can merge these 8 bytes of README without getting sued by the other maintainers.

Issue #1669, opened 2019-01-17
Create sequences using initializer list


cowo78:
## Issue description
To me it would be useful to be able to construct a few containers (i.e. tuple, list) using initializer lists. I would supply a patch.

## Reproducible example code
Something like this:

    pybind11::tuple create_tuple(std::initializer_list<pybind11::object> l)
    {
        pybind11::tuple retval {l.size()};
        size_t index {0};
        for (const pybind11::object& item : l)
            retval[index++] = item;
        return retval;
    }

YannickJadoul:
Sure, if you want to submit a PR, we can happily look at it and consider including it. If it doesn't blow up the binary size, there's a good chance it can be included.

I'm going to close this issue, now, though, in case you wouldn't be interested anymore. Feel free to reopen (or just open a PR).

PR #2451, opened 2020-09-01
Create sequences using initializer list

cowo78:
To me it would be useful to be able to construct a few containers (i.e. tuple, list) using initializer lists (see #1669)

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -11,6 +11,7 @@
 
 #include "detail/common.h"
 #include "buffer_info.h"
+#include <initializer_list>
 #include <utility>
 #include <type_traits>
 
@@ -1216,6 +1217,11 @@ public:
     explicit tuple(size_t size = 0) : object(PyTuple_New((ssize_t) size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate tuple object!");
     }
+    explicit tuple(std::initializer_list<object> init_list) : tuple(init_list.size()) {
+        size_t index {0};
+        for (const pybind11::object& item : init_list)
+            detail::tuple_accessor(*this, index++) = item;
+    }
     size_t size() const { return (size_t) PyTuple_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
@@ -1275,6 +1281,11 @@ public:
     PYBIND11_OBJECT_CVT(list, object, PyList_Check, PySequence_List)
     explicit list(size_t size = 0) : object(PyList_New((ssize_t) size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate list object!");
+    }
+    explicit list(std::initializer_list<object> init_list) : list(init_list.size()) {
+        size_t index {0};
+        for (const pybind11::object& item : init_list)
+            detail::list_accessor(*this, index++) = item;
     }
     size_t size() const { return (size_t) PyList_Size(m_ptr); }
     bool empty() const { return size() == 0; }
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -237,6 +237,16 @@ TEST_SUBMODULE(pytypes, m) {
         );
     });
 
+    // Tuples and lists can also be constructed using an initializer list of pybind11::object subclasses
+    m.def("initializer_list", []() {
+        return py::dict(
+            "tuple_ints"_a = py::tuple({py::int_(1), py::int_(2), py::int_(3)}),
+            "tuple_floats"_a = py::tuple({py::float_(2.2), py::float_(3.1), py::float_(4.5), py::float_(5.4)}),
+            "list_ints"_a = py::list({py::int_(1), py::int_(2), py::int_(3)}),
+            "list_floats"_a = py::list({py::float_(2.2), py::float_(3.1), py::float_(4.5), py::float_(5.4)})
+        );
+    });
+
     m.def("convert_to_pybind11_str", [](py::object o) { return py::str(o); });
 
     m.def("get_implicit_casting", []() {
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -221,6 +221,26 @@ def test_constructors():
     for k in noconv2:
         assert noconv2[k] is expected[k]
 
+    init_list = m.initializer_list()
+    expected = ("tuple_ints", tuple), ("list_ints", list)
+    for key, type_ in expected:
+        ints = init_list[key]
+        assert(isinstance(ints, type_))
+        assert(len(ints) == 3)
+        assert(ints[0] == 1)
+        assert(ints[1] == 2)
+        assert(ints[2] == 3)
+
+    expected = ("tuple_floats", tuple), ("list_floats", list)
+    for key, type_ in expected:
+        floats = init_list[key]
+        assert(isinstance(floats, type_))
+        assert(len(floats) == 4)
+        assert(floats[0] == 2.2)
+        assert(floats[1] == 3.1)
+        assert(floats[2] == 4.5)
+        assert(floats[3] == 5.4)
+
 
 def test_pybind11_str_raw_str():
     # specifically to exercise pybind11::str::raw_str


YannickJadoul (approved):
Seems useful & harmless to me.

tests/test_pytypes.cpp
```diff
@@ -237,6 +237,16 @@ TEST_SUBMODULE(pytypes, m) {
         );
     });
 
+    // Tuples and lists can also be constructed using an initializer list of pybind11::object subclasses
+    m.def("initializer_list", []() {
+        return py::dict(
+            "tuple_ints"_a = py::tuple({py::int_(1), py::int_(2), py::int_(3)}),
+            "tuple_floats"_a = py::tuple({py::float_(2.2), py::float_(3.1), py::float_(4.5), py::float_(5.4)}),
+            "list_ints"_a = py::list({py::int_(1), py::int_(2), py::int_(3)}),
+            "list_floats"_a = py::list({py::float_(2.2), py::float_(3.1), py::float_(4.5), py::float_(5.4)})
```
YannickJadoul, line 246: Since it's `initializer_list<object>`, is it possible to try out mixing `int`s and `float`s?

include/pybind11/pytypes.h
```diff
@@ -1216,6 +1217,11 @@ class tuple : public object {
     explicit tuple(size_t size = 0) : object(PyTuple_New((ssize_t) size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate tuple object!");
     }
+    explicit tuple(std::initializer_list<object> init_list) : tuple(init_list.size()) {
+        size_t index {0};
+        for (const pybind11::object& item : init_list)
+            detail::tuple_accessor(*this, index++) = item;
+    }
```
YannickJadoul, line 1224: With this, we could think about deprecating setting elements of a tuple, since tuples are supposed to be immutable once constructed?

tests/test_pytypes.py
```diff
@@ -221,6 +221,26 @@ def test_constructors():
...
+
+    expected = ("tuple_floats", tuple), ("list_floats", list)
+    for key, type_ in expected:
+        floats = init_list[key]
+        assert(isinstance(floats, type_))
+        assert(len(floats) == 4)
+        assert(floats[0] == 2.2)
+        assert(floats[1] == 3.1)
+        assert(floats[2] == 4.5)
+        assert(floats[3] == 5.4)
```
YannickJadoul, line 242: Sorry, I clicked "approve" too quickly, and missed this.

This could just be `assert initializer_list["tuple_floats"] == (2.2, 3.1, 4.5, 5.4)`, right, removing the for loop and just writing it in 4 lines? We can just take advantage of pytest's AST rewriting to show what goes wrong. Same for  `ints`.
Also, no parentheses for the `assert` statement, please.

tests/test_pytypes.cpp
cowo78: Sure, see updated PR.

include/pybind11/pytypes.h
cowo78: Makes sense to me.

tests/test_pytypes.py
cowo78: Sure. See updated PR.

include/pybind11/pytypes.h
YannickJadoul: Nothing for this PR ofc :-)

YannickJadoul:
Thanks for the updates, @cowo78!

Let's get a few more reviews in, and then this should be able to go in, as far as I'm concerned.

bstaletic:
I would have preferred something more general. Why not allow `std::array` or `boost::unordered_map`? In C++20 we can do something like this:

```c++
explicit tuple(std::ranges::forward_range auto&& rng) : tuple(std::ranges::distance(rng)) {
    size_t i = 0;
    for(object&& o : rng) detail::tuple_accessor(*this, index++) = item;
}
```

Now I know C++20 is really *really* new, but any thoughts on doing something like that with `enable_if` and SFINAE blackmagic?

YannickJadoul:
Is there a problem with accepting any `<typename Container>` that has `container.begin()` and `container.end()` of the right type? C++20 seems really recent :-/

bstaletic:
If you accept any type with `begin()` and `end()`, you'll end up taking `distance()` of `istream_view` thus exhausting the view and then would try to iterate it. The minimal requirement needs to be "any container whose iterators are forward iterators".

YannickJadoul:
Ok, right. What I tried to say: I like that idea so much that it would be a pity to only have it on C++20 ;)

Would you like to look into this, @cowo78, or should we merge this independently and have @bstaletic set up another PR? (Or @bstaletic could add to this PR, ofc)

bstaletic:
Woo! My `enable_if`-foo isn't too bad!

```c++
#include <type_traits>
#include <iterator>
#include <utility>
template<typename... Ts> struct make_void { typedef void type;};
template<typename... Ts> using void_t = typename make_void<Ts...>::type;
template <typename T, typename = void>
struct is_iterable : std::false_type {};
template <typename T>
struct is_iterable<T, void_t<decltype(std::declval<T>().begin()),
                             decltype(std::declval<T>().end())>>
    : std::true_type {};

template <typename T,
	  typename std::enable_if<is_iterable<T>::value &&
				  std::is_base_of<std::forward_iterator_tag,
				                  typename std::iterator_traits<typename T::iterator>::iterator_category>::value,
				  int>::type = 0>
void f(T&&) {}

int main() {
	f(std::vector<int>{});
	f(std::list<int>{});
	f(std::set<int>{});
	f(std::initializer_list<int>{});
	f(std::map<int,int>{});
	//f(pybind11::dict{}); // pybind11 types don't have iterator typedef.
}
```

cowo78:
> Ok, right. What I tried to say: I like that idea so much that it would be a pity to only have it on C++20 ;)
> 
> Would you like to look into this, @cowo78, or should we merge this independently and have @bstaletic set up another PR? (Or @bstaletic could add to this PR, ofc)

I won't argue about C++-20 since I did not (yet) look into the new features. I will try to provide something that allows using containers whose iterators are forward iterators.

Since we're dancing it would be also interesting to construct the containers using heterogeneous types i.e.
`py::tuple(1, 1.3, "something")`

bstaletic:
> I won't argue about C++-20 since I did not (yet) look into the new features. I will try to provide something that allows using containers whose iterators are forward iterators.

Thanks!

> Since we're dancing it would be also interesting to construct the containers using heterogeneous types i.e.
py::tuple(1, 1.3, "something")

In C++17 that would be something like this:

```c++
template<typename ...Ts>
tuple(Ts&&... ts) : tuple(sizeof...(Ts)) {
    // C++17 fold expressions. C++11 would need some recursive template instantiations and would tank compile times.
    size_t index = 0;
    (..., (detail::tuple_accessor(*this, index++) = py::cast(std::forward<Ts>(ts))));
}
```

Questions:

1. Do we want `py::tuple` constructible from any iterable? Like `folly::FBString`?
2. Do we want `py::tuple` constructible from `T...`? If we use this one only, `tuple(std::vector<int>{1,2,3})` compiles, but `len(tuple) == 1` and `tuple == ([1,2,3],)`.
3. Do we want both or only one of these?

Even though the variadic template version is more efficient than an equivalent `initializer_list` one, I think we should just go with the "any forward range" idea.

cowo78:
> > Since we're dancing it would be also interesting to construct the containers using heterogeneous types i.e.
> > py::tuple(1, 1.3, "something")
> 
> In C++17 that would be something like this:
> 
> ```c++
> template<typename ...Ts>
> tuple(Ts&&... ts) : tuple(sizeof...(Ts)) {
>     // C++17 fold expressions. C++11 would need some recursive template instantiations and would tank compile times.
>     size_t index = 0;
>     (..., (detail::tuple_accessor(*this, index++) = py::cast(std::forward<Ts>(ts))));
> }
> ```
> 
> Questions:

My 2c:
 
>     1. Do we want `py::tuple` constructible from any iterable? Like `folly::FBString`?

Yes.

>     2. Do we want `py::tuple` constructible from `T...`? If we use this one only, `tuple(std::vector<int>{1,2,3})` compiles, but `len(tuple) == 1` and `tuple == ([1,2,3],)`.

No. We want `T, Ts...` (at least 2 template args).

>     3. Do we want both or only one of these?

Both, so that you have the following options:

- from initializer_list<py::object>
- from initializer_list<C++ type>
- from C++ container 
- from heterogeneous C++ types passed as arguments

> Even though the variadic template version is more efficient than an equivalent `initializer_list` one, I think we should just go with the "any forward range" idea.

Does not hurt to keep both, as it also gives an option to limit binary size.

YannickJadoul:
Two things:

- I propose to keep this PR small, so we can reasonably quickly merge the part we all agree on. So, let's just try to implement @bstaletic's more general proposal of having any forward-iterable class instead of `initializer_list`?
- Constructing a `tuple` from variadic template arguments already exists: [`py::make_tuple`](https://github.com/pybind/pybind11/blob/4c36fb7b1236fce25e00b63f357ccc36dc006662/include/pybind11/cast.h#L1822-L1848). So again, let's keep this simple and just implement the improvement to the original proposal, I'd say.

YannickJadoul:
Clarification: by that I mean, we only add a constructor from `SomeContainer<py::object>`, as we shouldn't confuse the `py::tuple` interface for something that does a bunch of conversions; there's `py::make_tuple` for that, and if that needs to be adapted, it's another PR.
`SomeContainer<py::object>` should also match `std::initializer_list<py::object>`, I think, so no need to add that overload?

cowo78:
> Two things:
> 
>     * I propose to keep this PR small, so we can reasonably quickly merge the part we all agree on. So, let's just try to implement @bstaletic's more general proposal of having any forward-iterable class instead of `initializer_list`?

My 2c:

- AFAIK move semantics do not work with initializer list (so will need a separate constructor for containers)
- I want to use an initializer list with, i.e. py::int_ and py::float_ (need an initializer_list<py::object>)
- I want to use an initializer list with plain C++ object (need a templated initializer_list)

So, yes, stuff is bloating up here. It may very well be that my knowledge is not up to the task but I did not figure out how to do something more compact.

>     * Constructing a `tuple` from variadic templated arguments already exists: [`py::make_tuple`](https://github.com/pybind/pybind11/blob/4c36fb7b1236fce25e00b63f357ccc36dc006662/include/pybind11/cast.h#L1822-L1848). So again, let's keep this simple and just implement the improvement to the original proposal, I'd say.

Right, I forgot that.

cowo78:
> > I want to use an initializer list with plain C++ object (need a templated initializer_list)
> 
> I don't agree with adding that here, I think. It was also not part of your original proposal. `py::tuple` is the C++ interface to a Python tuple. That might be something for another PR, to maybe add to `py::make_tuple`. Or you just write your own function for that, in your personal project, ofc.

Well, it wasn't in the original proposal just because I used py::object for a start. I just think that both versions are useful. Your shout however. 

> > So, yes, stuff is bloating up here. It may very well be that my knowledge is not up to the task but I did not figure out how to do something more compact.
> 
> No, I'm not claiming that. I just want to keep things separated a bit.

Again, I'm just a simple user so your shout here.

bstaletic:
>> AFAIK move semantics do not work with initializer list (so will need a separate constructor for containers)

> Oh, I don't know about that. @bstaletic will, though, I expect @bstaletic?

Sort of correct. An `initializer_list<T>` basically holds `const T&` and doesn't own the data. This means that `std::move(il.back())` is the same as `il.back()`.  This doesn't mean that things suddenly break.

YannickJadoul:
> Well, it wasn't in the original proposal just because I used py::object for a start. I just think that both versions are useful. Your shout however.

I'm also not the only one going over this, ofc. But here's my suggestion: adapt this PR to @bstaletic's suggestion (`ForwardIterableContainer<T>`, hopefully also working for `initializer_list`, let's see), and propose the other thing in a new PR if you want. I'm currently not a hug fan to merge it, but this will give us and the other maintainers some room to discuss what to do with it.

YannickJadoul:
Actually, I'm quickly losing the value of constructing a `py::tuple` from an `initializer_list`, as well. There's `py::make_tuple`, which does exactly what you want, already (and it also works with `py::object`).

cowo78:
> This pull request is quickly losing focus and we're getting side-tracked. That aside, I'm not a fan of `initializer_list<T>`, because we have `py::make_tuple`. Like @YannickJadoul said, `py::tuple::tuple(...)` isn't meant to do random conversions. Anyway, this overload makes `initializer_list<py::object>` useless.
> 
> As for `initializer_list<py::object>`, it's functionality is covered by both, `initializer_list<T>` and my idea of "any forward range".
> 
> Also, my "any forward range" should probably have a constraint that `is_base_of<handle, T::value_type>::value` is true.

Well, I think one can make use of `py::tuple({1,2,3})`. Currently this feature is not available (and py::make_tuple does not work this way with GCC 10.2.0) so the logic behind this PR.

`initializer_list<py::object>` and `initializer_list<T>` do not 100% overlap as with the templated version you cannot do `py::tuple({py::int_(1), py::float_(2.2)}).

The "forward range" concept of course is much better than the constrained "initializer_list" but it seems that you cannot use a templated rvalue constructor with an initializer list.

All of this said I think we can go back to a simple question: do we **really** need an initializer_list constructor since its functionality can basically be achieved with existing `py::make_tuple`? Since it's mostly sugar to provide a more intuitive (at least for me) usage pattern I'm just fine with either decision.

YannickJadoul:
> Well, I think one can make use of `py::tuple({1,2,3})`. Currently this feature is not available (and py::make_tuple does not work this way with GCC 10.2.0) so the logic behind this PR.

I'm, again, really not sure it's not `py::tuple`'s job to do conversion. This is `py::make_tuple`'s job, I think (though I'm not sure). At any rate, none of the other C++ classes for Python types do this, so we want to keep this as simple as possible. The original idea (with `py::object`s) is probably easy to merge in. But converting from a bunch of other things really needs another discussion, probably best in another PR.

> `initializer_list<py::object>` and `initializer_list<T>` do not 100% overlap as with the templated version you cannot do `py::tuple({py::int_(1), py::float_(2.2)}).
> 
> The "forward range" concept of course is much better than the constrained "initializer_list" but it seems that you cannot use a templated rvalue constructor with an initializer list.

I'm going to pass this to @bstaletic, who tried out more. Can we not achieve this, @bstaletic?

> All of this said I think we can go back to a simple question: do we **really** need an initializer_list constructor since its functionality can basically be achieved with existing `py::make_tuple`? Since it's mostly sugar to provide a more intuitive (at least for me) usage pattern I'm just fine with either decision.

I think it's nice to have a Python-interface to construct from a list of `py::objects`, though, so I wouldn't mind adding that constructor. The other one is then up for later debate, but I think these two issues can be separate?

cowo78:
I reverted all other constructor and left only a `std::initializer_list<pybind11::object>` based constructor for tuples, lists and sets.

include/pybind11/pytypes.h
```diff
@@ -1213,9 +1214,21 @@ class capsule : public object {
...
 public:
     PYBIND11_OBJECT_CVT(tuple, object, PyTuple_Check, PySequence_Tuple)
+    /** \rst
+        Creates an empty ``tuple`` with given ``size`` to be later filled in.
+     \endrst */
     explicit tuple(size_t size = 0) : object(PyTuple_New((ssize_t) size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate tuple object!");
     }
+    /** \rst
+        Creates a ``tuple`` from an initializer list of object instances.
```
joukewitteveen, line 1224: Maybe 
```suggestion
        Creates a ``tuple`` from an initializer list of ``py::object`` instances.
```
just to be explicit? In addition, the added documentation on `make_tuple` in #2840 could be augmented with some remarks on the new ways to initialize a tuple. I didn't bother to document `list` and `set` there, because they are mutable (pybind11 offers ways to modify tuples, but that's besides the point).

tests/test_pytypes.cpp
```diff
@@ -9,6 +9,13 @@
 
 #include "pybind11_tests.h"
 
+#include <array>
+#include <deque>
+#include <list>
+#include <set>
+#include <string>
+#include <vector>
+
```
joukewitteveen, line 18:
```suggestion
```

I assume these are leftovers?

include/pybind11/pytypes.h
cowo78: Sure. Explicit is better than implicit.

tests/test_pytypes.cpp
cowo78: Yes, but leave the `#include <string>` as strings are used.

Issue #996, opened 2017-08-11
`py::list` iterator not assignable


crisluengo:
This does not compile on GCC 5.4:
```cpp
py::list out(1);
auto it = out.begin();
*it = py::cast(0);         // fails to compile
```
These other statements are OK:
```cpp
out[0] = py::cast(0);      // works OK
out.append(py::cast(0));   // works OK
```

The error message indicates that *it is const. I guess it's related to this:
```
using list_iterator = generic_iterator<iterator_policies::sequence_fast_readonly>;
```

To me this is a bug, either in the documentation or in the code. Is there a specific reason for these iterators to be read-only? Am I supposed to create lists by appending elements one by one?

PR #2456, opened 2020-09-02
Drop support for PyPy < 7.3.1 and clean up old PyPy workarounds

### LINKED ISSUE #2436: Explicit minimum supported PyPy?
I'd like to explicitly have a minimum PyPy version stated, supported, and tested.

My proposal would be to support 7.3.0+ for now. Reasoning:

* A PyPy dev ( @mattip ) stated "PyPy users are pretty good about updating" (indicating there's no need to support 1+ year old PyPy's for now) in https://github.com/pypa/wheel/issues/372
* Older versions of PyPy are buggy
* Wheel support on PyPI has been very recent, 7.2 or 7.3 I believe
* The [separate wheel repository](https://github.com/antocuni/pypy-wheels) for NumPy, etc. only supports 7.3 (though they have a legacy archive with 7.1 and 7.2)
* NumPy's PyPI wheels will be 7.3 only when they are released
* `cibuildwheel` is 7.3 only
* 7.3.x wheels work on any 7.3.x release (the first time they've added this ABI promise)
* It's not trivial to test with less than 7.2, since PyPy is often just the latest version, rather than coming in versions, and no wheels are available.

In the past, such as in 2.5.0, our support was hit or miss; we don't support the latest PyPy3, for example, in 2.5.0.

If we choose an older minimum version, we need to add CI for it.

YannickJadoul:
Commit by commit, let's test which parts can go and make tests succeed, and which preprocessor branches we'll need to keep.

Suggestions/commits welcome.

Closes #2436.

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -223,7 +223,7 @@ avoids this issue involves weak reference with a cleanup callback:
 
 .. code-block:: cpp
 
-    // Register a callback function that is invoked when the BaseClass object is colelcted
+    // Register a callback function that is invoked when the BaseClass object is collected
     py::cpp_function cleanup_callback(
         [](py::handle weakref) {
             // perform cleanup here -- this function is called with the GIL held
@@ -237,9 +237,9 @@ avoids this issue involves weak reference with a cleanup callback:
 
 .. note::
 
-    PyPy (at least version 5.9) does not garbage collect objects when the
-    interpreter exits. An alternative approach (which also works on CPython) is to use
-    the :py:mod:`atexit` module [#f7]_, for example:
+    PyPy does not garbage collect objects when the interpreter exits. An alternative
+    approach (which also works on CPython) is to use the :py:mod:`atexit` module [#f7]_,
+    for example:
 
     .. code-block:: cpp
 
--- a/tests/test_multiple_inheritance.cpp
+++ b/tests/test_multiple_inheritance.cpp
@@ -193,14 +193,12 @@ TEST_SUBMODULE(multiple_inheritance, m) {
         .def_readwrite_static("static_value", &VanillaStaticMix2::static_value);
 
 
-#if !(defined(PYPY_VERSION) && (PYPY_VERSION_NUM < 0x06000000))
     struct WithDict { };
     struct VanillaDictMix1 : Vanilla, WithDict { };
     struct VanillaDictMix2 : WithDict, Vanilla { };
     py::class_<WithDict>(m, "WithDict", py::dynamic_attr()).def(py::init<>());
     py::class_<VanillaDictMix1, Vanilla, WithDict>(m, "VanillaDictMix1").def(py::init<>());
     py::class_<VanillaDictMix2, WithDict, Vanilla>(m, "VanillaDictMix2").def(py::init<>());
-#endif
 
     // test_diamond_inheritance
     // Issue #959: segfault when constructing diamond inheritance instance
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,7 @@ Goodies
 In addition to the core functionality, pybind11 provides some extra
 goodies:
 
--  Python 2.7, 3.5+, and PyPy (tested on 7.3) are supported with an
+-  Python 2.7, 3.5+, and PyPy/PyPy3 7.3 are supported with an
    implementation-agnostic interface.
 
 -  It is possible to bind C++11 lambda functions with captured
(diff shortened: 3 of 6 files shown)


bstaletic:
This is a peculiar piece of code...

https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L730-L736

Add to that https://github.com/pybind/pybind11/issues/996

tests/test_multiple_inheritance.cpp
```diff
@@ -193,14 +193,12 @@ TEST_SUBMODULE(multiple_inheritance, m) {
         .def_readwrite_static("static_value", &VanillaStaticMix2::static_value);
 
 
-#if !(defined(PYPY_VERSION) && (PYPY_VERSION_NUM < 0x06000000))
```
bstaletic, line 196: Huh? If this was meant for PYPY < 6, then shouldn't this be thrown out, instead of allowed on 7?

henryiii: Oops, misread that. Thanks!

bstaletic: Actually, I misread this.

`NOT(PYPY AND PYPY < 6)` would be `NOT PYPY OR PYPY >= 6` which would mean either CPython or PyPy 6+. Since we dropped support for PyPy < 7, your original commit was right.

a) Sorry
b) Feel free to drop the last commit.

henryiii: Undone. :)

bstaletic:
`docs/intro.rst` and `docs/advanced/misc.rst` have references to PyPy 5.7 and PyPy 5.9, respectively.

YannickJadoul:
Good catch, @bstaletic! I was going to tell you your can just push to this branch if you want, but I guess you can't. I gave you permissions on my fork now, though. So feel free to... (Feel free to not do that as well, btw! I can get to it later this weekend.)

henryiii:
> but I guess you can't

Can't all maintainers edit your branch, since it's a PR here? Or did you uncheck that option?

YannickJadoul:
> Can't all maintainers edit your branch, since it's a PR here? Or did you uncheck that option?

No, it's checked, but @bstaletic is (officially) not a maintainer.

docs/advanced/misc.rst
```diff
@@ -237,9 +237,9 @@ avoids this issue involves weak reference with a cleanup callback:
 
 .. note::
 
-    PyPy (at least version 5.9) does not garbage collect objects when the
-    interpreter exits. An alternative approach (which also works on CPython) is to use
-    the :py:mod:`atexit` module [#f7]_, for example:
+    PyPy does not garbage collect objects when the interpreter exits. An alternative
```
YannickJadoul, line 240: We might want to re-confirm, this, though. I did not do that, yet.

YannickJadoul:
Thanks, @bstaletic! Fixed those extra PyPy reference, and rebased so that we can take advantage of #2462.

docs/advanced/misc.rst
bstaletic: We have tests for that. Some tests explicitly call `gc.collect()` more than once because of PyPy, so it shouldn't be hard to verify.

YannickJadoul: I'll give it a quick try, but it's actually documented: https://doc.pypy.org/en/latest/cpython_differences.html#differences-related-to-garbage-collection-strategies

> Last note: CPython tries to do a `gc.collect()` automatically when the program finishes; not PyPy. (It is possible in both CPython and PyPy to design a case where several `gc.collect()` are needed before all objects die. This makes CPython’s approach only work “most of the time” anyway.)

Should we add this to the docs, or are we fine as is?

YannickJadoul: (I actually can't find any tests on the garbage collector passing by when the interpreter exits, though. So not sure how to apply @bstaletic's suggestion.)

bstaletic: What about calling `PyGC_Collect()`? Would that work across the board?

YannickJadoul: Good plan, but seems PyPy doesn't have `PyGC_Collect` :-(

YannickJadoul:
Reminder to ourselves: https://github.com/pybind/pybind11/issues/2436#issuecomment-685671487

docs/advanced/misc.rst
bstaletic: Other than `gc.collect()`, which isn't C, there's `PyGC_Collect()` and a bunch of "didn't mean to make it public" functions that got removed in python 3.9. On the other hand `PyGC_Collect()` was never documented.

Conclusion: CPython is a horrible mess.

YannickJadoul: The PyPy devs, struggling to keep up with massive amounts of C API every release, are the last ones you need to convince of that statement, I'm afraid ;-)

rwgk (approved):

PR #2461, opened 2020-09-04
fix: support nvcc and test

henryiii:
Adding a CI test for NVCC, along with fixes and if statements. Everyone is welcome to suggest better fixes, especially for deactivated tests!

* Added CI test for NVCC (use this recipe to run locally in docker, as well)
* Moved logic (I'm calling it `invalid_convert_integral`) into compile time to fix warning (pointless comparison with 0) and provide better performance (will add comments and fix my one existing comment if that's the way we go)
* Worked around weird issue with `vectorize_helper` complaining about private usage internally
*` test_constants_and_functions.cpp` skipped due to "incompatible exception specifications"
* Added body to `void *operator new(size_t bytes)` (fixes warning on Windows, too)
* NVCC joins Intel for `NonCopyable` not supported.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1007,26 +1007,6 @@ template <typename CharT> using is_std_char_type = any_of<
 >;
 
 
-template<typename T, typename U, enable_if_t<!std::is_integral<T>::value, int> = 0>
-bool invalid_convert_integral(U) {
-    return false;
-}
-
-template<typename T, typename U, enable_if_t<std::is_integral<T>::value && !std::is_unsigned<T>::value, int> = 0>
-bool invalid_convert_integral(U value) {
-    return sizeof(U) != sizeof(T)
-           && (   value < (U) (std::numeric_limits<T>::min)()
-               || value > (U) (std::numeric_limits<T>::max)()
-           );
-}
-
-template<typename T, typename U, enable_if_t<std::is_integral<T>::value && std::is_unsigned<T>::value, int> = 0>
-bool invalid_convert_integral(U value) {
-    return sizeof(U) != sizeof(T)
-           &&  value > (U) (std::numeric_limits<T>::max)();
-}
-
-
 template <typename T>
 struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_type<T>::value>> {
     using _py_type_0 = conditional_t<sizeof(T) <= sizeof(long), long, long long>;
@@ -1055,10 +1035,11 @@ public:
                 : (py_type) PYBIND11_LONG_AS_LONGLONG(src.ptr());
         }
 
+        // Python API reported an error
         bool py_err = py_value == (py_type) -1 && PyErr_Occurred();
 
-        // Only compare > 0 if not unsigned (warns on some compilers, like nvcc)
-        if (py_err || invalid_convert_integral<T>(py_value)) {
+        // Check to see if the conversion is valid (integers should match exactly)
+        if (py_err || (std::is_integral<T>::value && py_value != (py_type) (T) py_value)) {
             bool type_error = py_err && PyErr_ExceptionMatches(
 #if PY_VERSION_HEX < 0x03000000 && !defined(PYPY_VERSION)
                 PyExc_SystemError
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1039,7 +1039,8 @@ public:
         bool py_err = py_value == (py_type) -1 && PyErr_Occurred();
 
         // Check to see if the conversion is valid (integers should match exactly)
-        if (py_err || (std::is_integral<T>::value && py_value != (py_type) (T) py_value)) {
+        // Signed/unsigned checks happen elsewhere
+        if (py_err || (std::is_integral<T>::value && sizeof(py_type) != sizeof(T) && py_value != (py_type) (T) py_value)) {
             bool type_error = py_err && PyErr_ExceptionMatches(
 #if PY_VERSION_HEX < 0x03000000 && !defined(PYPY_VERSION)
                 PyExc_SystemError
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1006,6 +1006,27 @@ template <typename CharT> using is_std_char_type = any_of<
     std::is_same<CharT, wchar_t> /* std::wstring */
 >;
 
+
+template<typename T, typename U, enable_if_t<!std::is_integral<T>::value, int> = 0>
+bool invalid_convert_integral(U) {
+    return false;
+}
+
+template<typename T, typename U, enable_if_t<std::is_integral<T>::value && !std::is_unsigned<T>::value, int> = 0>
+bool invalid_convert_integral(U value) {
+    return sizeof(U) != sizeof(T)
+           && (   value < (U) (std::numeric_limits<T>::min)()
+               || value > (U) (std::numeric_limits<T>::max)()
+           );
+}
+
+template<typename T, typename U, enable_if_t<std::is_integral<T>::value && std::is_unsigned<T>::value, int> = 0>
+bool invalid_convert_integral(U value) {
+    return sizeof(U) != sizeof(T)
+           &&  value > (U) (std::numeric_limits<T>::max)();
+}
+
+
 template <typename T>
 struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_type<T>::value>> {
     using _py_type_0 = conditional_t<sizeof(T) <= sizeof(long), long, long long>;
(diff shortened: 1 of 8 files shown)


tests/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../tools")
 
 option(PYBIND11_WERROR "Report all warnings as errors" OFF)
 option(DOWNLOAD_EIGEN "Download EIGEN (requires CMake 3.11+)" OFF)
+option(PYBIND11_CUDA_TESTS "Enable building CUDA tests (requires CMake 3.12+)" OFF)
```
YannickJadoul, line 33: Is there a reason our CMake explicitly needs this? Can't it be set during configuation from the outside, by passing `-DCMAKE_CUDA_STANDARD=...`, etc?

include/pybind11/cast.h
```diff
@@ -1006,6 +1006,27 @@ template <typename CharT> using is_std_char_type = any_of<
...
+    return sizeof(U) != sizeof(T)
+           && (   value < (U) (std::numeric_limits<T>::min)()
+               || value > (U) (std::numeric_limits<T>::max)()
+           );
+}
+
+template<typename T, typename U, enable_if_t<std::is_integral<T>::value && std::is_unsigned<T>::value, int> = 0>
+bool invalid_convert_integral(U value) {
+    return sizeof(U) != sizeof(T)
+           &&  value > (U) (std::numeric_limits<T>::max)();
```
bstaletic, line 1026: Is this completely correct?

`U == int8_t`, `T == uint32_t`, `value == 127` => `(U)numeric_limits::max() == (U)65535 == (U)0xff'ff'ff'ff == -1` => `127 > -1` and we end up saying this is invalid, when 127 can be represented in `uint32_t` losslessly.

tests/CMakeLists.txt
```diff
@@ -104,6 +113,16 @@ if((PYBIND11_TEST_FILES_ASYNC_I GREATER -1) AND (PYTHON_VERSION VERSION_LESS 3.5
   list(REMOVE_AT PYBIND11_TEST_FILES ${PYBIND11_TEST_FILES_ASYNC_I})
 endif()
 
+# Skip tests for CUDA check:
+# /pybind11/tests/test_constants_and_functions.cpp(125):
+#   error: incompatible exception specifications
```
bstaletic, line 118: What does "incompatible exception specification" mean?

tests/test_copy_move.cpp
```diff
@@ -175,14 +175,17 @@ TEST_SUBMODULE(copy_move_policies, m) {
     m.attr("has_optional") = false;
 #endif
 
-    // #70 compilation issue if operator new is not public
+    // #70 compilation issue if operator new is not public - body added but not needed on most compilers
```
bstaletic, line 178: This comment deserves a longer explanation. Which compilers behave oddly, is the first thing that popped into my mind when reading this code.

I know we know which ones today, but will we still remember in a year?

include/pybind11/numpy.h
```diff
@@ -1483,7 +1483,14 @@ struct vectorize_arg {
 
 template <typename Func, typename Return, typename... Args>
 struct vectorize_helper {
+
+// NVCC for some reason breaks if NVectorized is private
```
bstaletic, line 1487: What exactly breaks?

tests/CMakeLists.txt
henryiii: Yes, because you have to convince CMake that the `.cpp` files are actually CUDA files. Plus a few other things further down, like disabling a test that doesn't run on CUDA.

This only changes the test CMake, though, not the normal/public one.

include/pybind11/cast.h
henryiii: I noticed this too, by the time I finished with this I was pretty sure the (existing) logic is flawed. But I left it exactly the same for now, I just replicated what was there.

tests/CMakeLists.txt
henryiii: That's printed by NVCC when running. It doesn't like `noexcept(false)` or `throw()`, I believe, but the error message line is off (I think), so it's hard to tell what it is unhappy with.

tests/CMakeLists.txt
henryiii: Also, warnings specifications change completely with CUDA.

tests/test_copy_move.cpp
henryiii: MSVC and nvcc don't like having a definition with no body, even if it is never used. Having a body is better than manually trying to hide the warning, IMO, and when you start trying to hide the warning from multiple compilers, that explodes. (will add something like this to the note)

henryiii: I could probably move this out of the test definition, then the compilers would be okay with having a declaration without a definition. What they are complaining about is that these are local, so there's never a way to add the method body.

include/pybind11/numpy.h
henryiii: It claims NVectorize is private in this context and quits.

```
[ 73%] Building CUDA object tests/CMakeFiles/pybind11_tests.dir/test_opaque_types.cpp.o
cd /build/tests && /usr/local/cuda/bin/nvcc  -Dpybind11_tests_EXPORTS -I/pybind11/include -isystem=/usr/include/python3.8  -O1 -DNDEBUG -Xcompiler=-fPIC -Xcompiler=-fvisibility=hidden   -Werror all-warnings -std=c++11 -x cu -c /pybind11/tests/test_opaque_types.cpp -o CMakeFiles/pybind11_tests.dir/test_opaque_types.cpp.o
/pybind11/tests/test_numpy_vectorize.cpp: In lambda function:
/pybind11/tests/test_numpy_vectorize.cpp:86:6: error: 'constexpr const size_t pybind11::detail::vectorize_helper<double (*)(int, float, double), double, int, float, double>::NVectorized' is private within this context
   86 |         std::array<py::buffer_info, 3> buffers {{ arg1.request(), arg2.request(), arg3.request() }};
      |      ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/pybind11/include/pybind11/numpy.h:1490:25: note: declared private here
 1490 |     static constexpr size_t NVectorized = constexpr_sum(vectorize_arg<Args>::vectorize...);
```

tests/test_copy_move.cpp
henryiii: I added the longer note; let me know if you think it's a good idea to move it up and out of the test body, so compilers will not complain about the missing definition - then we should be able to leave it empty.

henryiii:
By the way, you can run:

```bash
docker run --rm -it -v $PWD:/pybind11 nvidia/cuda:11.0-devel-ubuntu20.04 bash
apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y cmake python3-dev python3-pytest
cmake -S pybind11 -B build -DPYBIND11_CUDA_TESTS=ON -DPYBIND11_WERROR=ON
cmake --build build -v -j2
cmake --build build --target pytest
```

To do the testing locally.

tests/CMakeLists.txt
YannickJadoul: Alright, then.

wjakob:
Fine with all of these changes, except the range check that has become very complex with the extra SFINAE.

include/pybind11/cast.h
```diff
@@ -1036,10 +1057,8 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
 
         bool py_err = py_value == (py_type) -1 && PyErr_Occurred();
 
-        // Protect std::numeric_limits::min/max with parentheses
```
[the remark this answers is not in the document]
henryiii: It's poorly documented, but yes, I think more recent NVCC versions support suppression if you can find the right code. But this will actually generate simpler/faster _runtime code_ with 2 fewer checks, if this happens fairly often.

I also think the logic may be flawed; it might be good to rework with new logic. Also I'm surprised it's showing up with some very odd types; things like pointers were `static_cast` to an integer is not allowed (which is why it's using C-style casts); I guess this is used to covert pointer addresses to Python ints?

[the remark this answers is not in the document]
henryiii: Check the new version out! Simpler logic (the tests here are pretty comprehensive, don't ask me how I know that 😊), with no more pointless comparisons.

include/pybind11/cast.h
henryiii: The reason it worked is U is always large (python's long). But I've come up with a much simpler check that is inline.

YannickJadoul (approved):
Looking at the `if (py_err || (std::is_integral<T>::value && py_value != (py_type) (T) py_value)) {`, this looks good to me :-)

henryiii:
Okay, thanks! Going in, will rebase the PGI PR on it soon, since they collide a bit on a couple of lines.

PR #2462, opened 2020-09-04
fix: use classic extension handling unless otherwise requested

henryiii:
WITH_SOABI is only defined in CMake 3.17+, and it doesn't really quite do the expected thing for PyPy (being fixed on the PyPy side). For now, let's use the classic method for adding the extension unless a user requests otherwise.


wjakob:
Regarding the test failure: maybe there is a difference between `SO` and `EXT_SUFFIX` after all? Looks great other than that!

henryiii:
It has to work with `SO`, otherwise the classic mode would be broken. The problem was that the function runs in the calling scope, so it doesn't "capture" variables when it was defined (CMake needs a C++11 capture specifier ;) ). When using add_subdirectory, you are in a scope where the variable wasn't defined, causing a completely empty extension.

PR #2463, opened 2020-09-04
pybind11_add_module(): OPT_SIZE argument for size-based optimization

wjakob:
This PR adds way of switching this specific target over to “optimize size” flags (-Os) no matter what the project itself uses (except in debug mode, where optimizations remain disabled). Performance within bindings is usually not that critical, and this can make a quite substantial difference in projects that are weighted down by heavy (10-100MB) python extension libraries. I’m seeing ~25% size reduction just from that. 

The PR also contains two other very minor changes:
1. cmake appends a dummy define -D${target_name}_EXPORTS when compiling shared libraries. These are often not needed, especially not for binding code. I've added directives that remove them.

2. pybind11 prints two redundant lines when included by a parent project (via `add_subdirectory`), I've deleted one of them.

### DIFF
--- a/tools/pybind11Common.cmake
+++ b/tools/pybind11Common.cmake
@@ -10,6 +10,7 @@ Adds the following targets::
     pybind11::python_link_helper - Adds link to Python libraries
     pybind11::python2_no_register - Avoid warning/error with Python 2 + C++14/7
     pybind11::windows_extras - MSVC bigobj and mp for building multithreaded
+    pybind11::opt_size - avoid optimizations that increase code size
 
 Adds the following functions::
 
@@ -132,6 +133,23 @@ if(MSVC)
       PROPERTY INTERFACE_COMPILE_OPTIONS $<$<NOT:$<CONFIG:Debug>>:$<$<COMPILE_LANGUAGE:CXX>:/MP>>)
   endif()
 endif()
+
+# ----------------------- Optimize binary size --------------------------
+
+add_library(pybind11::opt_size IMPORTED INTERFACE ${optional_global})
+
+if(MSVC)
+  set(PYBIND11_OPT_SIZE /Os)
+else()
+  set(PYBIND11_OPT_SIZE -Os)
+endif()
+
+set_property(
+  TARGET pybind11::opt_size
+  APPEND
+  PROPERTY INTERFACE_COMPILE_OPTIONS $<$<CONFIG:Release>:${PYBIND11_OPT_SIZE}>
+           $<$<CONFIG:MinSizeRel>:${PYBIND11_OPT_SIZE}>
+           $<$<CONFIG:RelWithDebInfo>:${PYBIND11_OPT_SIZE}>)
 
 # ----------------------- Legacy option --------------------------
 
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -138,8 +138,8 @@ endif()
 # WITHOUT_SOABI and WITH_SOABI will disable the custom extension handling used by pybind11.
 # WITH_SOABI is passed on to python_add_library.
 function(pybind11_add_module target_name)
-  cmake_parse_arguments(PARSE_ARGV 1 ARG "STATIC;SHARED;MODULE;THIN_LTO;NO_EXTRAS;WITHOUT_SOABI"
-                        "" "")
+  cmake_parse_arguments(PARSE_ARGV 1 ARG
+                        "STATIC;SHARED;MODULE;THIN_LTO;OPT_SIZE;NO_EXTRAS;WITHOUT_SOABI" "" "")
 
   if(ARG_ADD_LIBRARY_STATIC)
     set(type STATIC)
@@ -203,6 +203,10 @@ function(pybind11_add_module target_name)
 
   if(MSVC)
     target_link_libraries(${target_name} PRIVATE pybind11::windows_extras)
+  endif()
+
+  if(ARG_OPT_SIZE)
+    target_link_libraries(${target_name} PRIVATE pybind11::opt_size)
   endif()
 endfunction()
 
--- a/tools/pybind11Tools.cmake
+++ b/tools/pybind11Tools.cmake
@@ -1,6 +1,6 @@
 # tools/pybind11Tools.cmake -- Build system for the pybind11 modules
 #
-# Copyright (c) 2015 Wenzel Jakob <wenzel@inf.ethz.ch>
+# Copyright (c) 2020 Wenzel Jakob <wenzel.jakob@epfl.ch>
 #
 # All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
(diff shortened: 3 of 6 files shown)


tools/pybind11Tools.cmake
```diff
@@ -124,10 +124,10 @@ endfunction()
 
 # Build a Python extension module:
 # pybind11_add_module(<name> [MODULE | SHARED] [EXCLUDE_FROM_ALL]
-#                     [NO_EXTRAS] [THIN_LTO] source1 [source2 ...])
+#                     [NO_EXTRAS] [THIN_LTO] [OPT_SIZE] source1 [source2 ...])
 #
 function(pybind11_add_module target_name)
-  set(options MODULE SHARED EXCLUDE_FROM_ALL NO_EXTRAS SYSTEM THIN_LTO)
+  set(options "MODULE;SHARED;EXCLUDE_FROM_ALL;NO_EXTRAS;SYSTEM;THIN_LTO;OPT_SIZE")
```
henryiii, line 130: I'm guessing the change in syntax was for cmake-format wanting to wrap? It's fine, just curious. This is more symmetric with the inline def in the other module.

henryiii (approved):
Looks good to me!

tools/pybind11NewTools.cmake
```diff
@@ -181,6 +181,13 @@ function(pybind11_add_module target_name)
   if(MSVC)
     target_link_libraries(${target_name} PRIVATE pybind11::windows_extras)
   endif()
+
+  if(ARG_OPT_SIZE)
+    target_link_libraries(${target_name} PRIVATE pybind11::opt_size)
+  endif()
+
+  # Avoid the -D${target_name}_EXPORTS definition that CMake adds by default
```
henryiii, line 189: Are these harmful, though? I would assume they do nothing if never used, and someone might expect them to be there?

henryiii:
> This branch has conflicts that must be resolved

This is why changlog entries in PRs can be a bit of a pain. I've never come up with a reasonable solution, though.

Since _both_ of my merged PRs have affected this, I've rebased for you.

(If you need to update, `git fetch && git reset --hard` will sync back up - make sure you don't have local changes (if so, leave of the `--hard`).

tools/pybind11Tools.cmake
henryiii: Yep, saw this when I rebased.

tools/pybind11Common.cmake
```diff
@@ -133,6 +134,23 @@ if(MSVC)
...
+  set(PYBIND11_OPT_SIZE /Os)
+else()
+  set(PYBIND11_OPT_SIZE -Os)
+endif()
+
+set_property(
+  TARGET pybind11::opt_size
+  APPEND
+  PROPERTY INTERFACE_COMPILE_OPTIONS $<$<CONFIG:Release>:${PYBIND11_OPT_SIZE}>
+           $<$<CONFIG:MinSizeRel>:${PYBIND11_OPT_SIZE}>
```
YannickJadoul, line 151: Is this not already set in the default `MinSizeRel` configuration? I just tried, and I already got this on Linux by just adding `-DCMAKE_BUILD_TYPE=MinSizeRel`.

And how do `-O2` (or `-O3`) and `-Os` then interact? Which concern takes precedence?

I'm wondering if we shouldn't try leaving this to CMake and the user, and add a note to the docs advising about `MinSizeRel`? (Or maybe even making `MinSizeRel` the default build configuration, if nothing is set?)

tools/pybind11Tools.cmake
```diff
@@ -1,6 +1,6 @@
 # tools/pybind11Tools.cmake -- Build system for the pybind11 modules
 #
-# Copyright (c) 2015 Wenzel Jakob <wenzel@inf.ethz.ch>
+# Copyright (c) 2020 Wenzel Jakob <wenzel.jakob@epfl.ch>
```
YannickJadoul, line 3: Far from an expert, but I thought I most often see `2015-2020` in copyright messages?

E.g. https://softwareengineering.stackexchange.com/questions/227812/why-do-we-need-years-in-a-software-license or https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

tools/pybind11Common.cmake
henryiii: Maybe we could default to MinSizeRel for the tests, and add that as a suggestion to the docs? I do prefer using the CMake way of doing things rather than manually adding flags (plus this way you don't risk having both flags present).

tools/pybind11Tools.cmake
henryiii: Shouldn't we update those everywhere?

YannickJadoul: I think so, yes. But I just saw the change here.

tools/pybind11Common.cmake
YannickJadoul: It ís optional, this way, so it's not going to be unintentionally harmful to anyone. But CMake might be able to support more exotic compilers/environments?

tools/pybind11Tools.cmake
wjakob: Indeed -- pre-commit rewrote this into a bunch of lines, and I just set it to follow the convention in the other file.

tools/pybind11NewTools.cmake
wjakob: Ok, fair enough -- I'll delete them again.

tools/pybind11Common.cmake
wjakob: @YannickJadoul -- you are correct to point out that switching to `MinSizeRel` at a global level indeed gives size-based optimizations. The point of this PR is to do so selectively on a per-target level. A very common project layout is:

1. `libmy_project.so` -- all the expensive computation happens in here, optimize as much as possible with auto-vectorization, `-march=native -O3`, etc.
2. `my_project_cli`: linking to `lib_myproject.so` to expose functionality via command line interface
3. `my_project.cpython-..so`: Python bindings. These can become huge (~10-100MB) if the bindings are very comprehensive. Yet, performance here isn't all that critical, because we're just forwarding a bunch of arguments to long-running algorithms that live in the other shared library.

So the goal of this PR is to compile parts (1+2) in release mode, while using MinSizeRel just for the binding target (3).

When specifying a sequence of optimization flags, the latter one have precedence, at least for our compilers in the "officially supported" list.

tools/pybind11Tools.cmake
wjakob: Not really a big deal, the first stackoverflow link you posted says that it's kind of unnecessary. Just made an edit because I spotted an ancient email address that no longer works.

tools/pybind11Common.cmake
YannickJadoul: Right, OK, now I see where this is coming from! Thanks for the explanation.
Why not, then :-)

YannickJadoul (approved):
For what it's worth, adding my checkmark :-)

wjakob:
Merged with some documentation improvements following @YannickJadoul's question. I got rid of the originally proposed change to remove the -D..._EXPORTS flag.

PR #2465, opened 2020-09-05
Add note about specifying custom base class for Exceptions.

michalsustr:


docs/advanced/exceptions.rst
```diff
@@ -79,6 +79,20 @@ This call creates a Python exception class with the name ``PyExp`` in the given
 module and automatically converts any encountered exceptions of type ``CppExp``
 into Python exceptions of type ``PyExp``.
 
+It is possible to specify base class for the exception using the third 
+parameter, a pointer to `PyObject`:
```
YannickJadoul, line 83: I know this is currently correct, but do we have any idea why this isn't a `handle`?

docs/advanced/exceptions.rst
```diff
@@ -79,6 +79,20 @@ This call creates a Python exception class with the name ``PyExp`` in the given
...
+.. code-block:: cpp
+
+    py::register_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
+
+Then `PyExp` can be caught both as `PyExp` and `RuntimeError`.
+Common base classes from the table above are `PyExc_Exception` (default), 
+`PyExc_IndexError`, `PyExc_KeyError`, `PyExc_MemoryError`, `PyExc_OverflowError`, 
+`PyExc_RuntimeError`, `PyExc_StopIteration`, `PyExc_ValueError`. 
+More definitions can be found in Python's `pyerrors.h` file, and they follow
+the pattern `PyExc_XXX`.
```
YannickJadoul, line 94: Maybe we can just refer to the Python docs? Those are more accessible than `pyerror.h`, and then this part can be smaller (i.e. "The class objects of the built-in Python exceptions are listed in the Python documentation:" or so) : https://docs.python.org/3/c-api/exceptions.html#standard-exceptions

docs/advanced/exceptions.rst
```diff
@@ -79,6 +79,20 @@ This call creates a Python exception class with the name ``PyExp`` in the given
...
 into Python exceptions of type ``PyExp``.
 
+It is possible to specify base class for the exception using the third 
+parameter, a pointer to `PyObject`:
+
+.. code-block:: cpp
+
+    py::register_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
+
+Then `PyExp` can be caught both as `PyExp` and `RuntimeError`.
```
YannickJadoul, line 89: Does this need to be explicitly specified? This is really just the basic way Python handles an exception hierarchy. (Fine to keep it though, if you want.)

YannickJadoul (commented):
Thanks, @michalsustr! A few minor remarks, but good to have this documented :-)

docs/advanced/exceptions.rst
michalsustr: I'd like to keep it, even though it's basic (makes sure there is common understanding)

docs/advanced/exceptions.rst
michalsustr: No idea, first time contributing to pybind.

YannickJadoul: No problem. Let's see if @bstaletic or anyone else knows! :-)

docs/advanced/exceptions.rst
YannickJadoul: OK, fine with me.

docs/advanced/exceptions.rst
```diff
@@ -79,6 +79,19 @@ This call creates a Python exception class with the name ``PyExp`` in the given
 module and automatically converts any encountered exceptions of type ``CppExp``
 into Python exceptions of type ``PyExp``.
 
+It is possible to specify base class for the exception using the third
```
YannickJadoul, line 82: The trailing space on this line is causing our CI to fail it's "Format" job.

YannickJadoul (approved):
Looks good to me, once the CI gets fixed! :-)

docs/advanced/exceptions.rst
michalsustr: It is indeed formatting very well :D fixed

YannickJadoul: A little nitpicky, but now it's passing :-)

henryiii (approved):

docs/advanced/exceptions.rst
bstaletic: Nope, no clue other than "the source says so".

bstaletic (approved):
Looks good.

docs/advanced/exceptions.rst
YannickJadoul: OK, let's maybe change this in a different PR, then? (Actually, we might even change this to `py::type`, @henryiii?)

YannickJadoul:
No further comments, it seems, so let's merge. (If anyone still has remarks, afterwards, it's just docs and easy enough to change.)

YannickJadoul:
Thanks again, @michalsustr!

Issue #1310, opened 2018-03-08
Convert py::object Numpy array to py::array


kohr-h:
I'm playing around with embedding the interpreter, and I'd like to access a Numpy array stored in a `py::object` using pybind11's array handling functionality, and possibly hand it off to other code.

Here's a minimal example:
```C++
#include <pybind11/embed.h>
namespace py = pybind11;

int main() {
    py::scoped_interpreter guard{};
    py::object np = py::module::import("numpy");
    py::object arr = np.attr("ones")(3);
    // want it as py::array object now and call C++ functions on it
}
```

Here probably my ignorance of C++ shows, but how is this conversion done? Is there functionality for that in pybind11?

A follow-up on that question would then be: Can I stop the interpreter from deleting the object when the interpreter goes out of scope? If not, or if it requires too much hacking, I'd be okay with a copy.

If there's an elegant solution, I think it would be a nice addition to the docs on embedding the interpreter.

jagerman:
The built-in python wrappers like `py::array` are implicitly convertible from `py::object`s, so you can simply write:

```C++
py::array_t<double> arr = np.attr("ones")(3);
```

jagerman:
> A follow-up on that question would then be: Can I stop the interpreter from deleting the object when the interpreter goes out of scope? If not, or if it requires too much hacking, I'd be okay with a copy.

No.  When the interpreter goes out of scope there won't actually be a Python object left and so there would be no way back into the object.  It *is* possible, however, to have your data survive a interpreter shutdown and restart by having numpy operate on your own memory: pass a pointer to allocated memory (of at least the required size) into the `py::array`/`py::array_t` constructors.  Then then numpy object won't allocate anything itself, but will just directly access/modify your memory.

kohr-h:
Great, thanks @jagerman for the explanation! Would it make sense to add this kind of tiny example to the docs? I can make a PR.

jagerman:
Sure; it would fit as a new section in `docs/advanced/pycpp/object.rst`, I suppose.

It would also be worth mentioning that the implicit conversion does a type-check and possibly a conversion at runtime, raising an exception if the object type isn't right and conversion isn't possible.  (For example, if you tried: `py::dict d = np.attr("ones")(3);` you'd get an exception.  On the other hand, I believe `py::list l = np.attr("ones")(3);` *is* allowed because a python list can be constructed from anything iterable).

EricCousineau-TRI:
[...]

Issue #1201, opened 2017-12-01
pybind11 cheatsheet documentation


ezyang:
I have found the pybind11 documentation to be not so great for quickly trying to figure out "how the heck do I get from type A to type B". So I wrote my own notes. I'm interesting in getting these back into the official docs, but I am not sure what is the most appropriate form.

Here goes.

----

The important types:

* `py::handle` - equivalent to `PyObject*` (no automatic refcounting)
* `py::object` - does automatic refcounting. Subclass of py::handle.

### PyObject* ↔ py::handle

This is a trivial conversion. These two types are basically totally equivalent.

```
PyObject *x;

py::handle h = x; // implicit conversion OK
x = h.ptr();
```

### py::handle (or subclass) → py::object (or subclass)

The syntax here works for both py::object, and any of its subclasses (e.g., py::str, etc)
http://pybind11.readthedocs.io/en/stable/reference.html#_CPPv26object

All of these are **unchecked**. Do an isinstance check beforehand if necessary. (Note that if you try to do anything with the py::object subclass, pybind11 will probably throw an exception at that point, so you probably aren't completely hosed. But casts to py::foo things are not checked at the point you do the cast!)

**Common gotcha:** you MUST pass a template argument to py::cast if you're casting from a PyObject*, it will silently do the wrong thing if you don't.

```
py::handle h;

// Copying
auto o = py::reinterpret_borrow<py::object>(h); // h can be PyObject*
  // or
auto o = x.cast<py::object>();
  // or
auto o = py::cast<py::object>(h); // h can be PyObject*
// equivalent to:
//    Py_INCREF(x);
//    THPObjectPtr o(x);

// Moving
auto o = py::reinterpret_steal<py::object>(h); // h can be PyObject*
  // or
auto o = py::object(std::move(h)); // h can be PyObject*
// equivalent to:
//    THPObjectPtr o(x);
```

### py::object (or superclass) → py::handle

Stealing is pretty useful if you're trying to invoke a C API function that steals its argument.

```
py::object o;

// Borrows from o
py::handle h2 = o;
// refcount NOT increased; h2 is only live as long as o is

// Steals from o
py::handle h = o.release();
// o no longer valid
```

NB: a “copy” from object into handle doesn't make sense, because handle doesn't know how to manage its own memory.

### py::object →  C++

* Checked conversion!
* Reference: http://pybind11.readthedocs.io/en/stable/reference.html#_CPPv2NK6handle4castEv
[...]

PR #2466, opened 2020-09-06
Add py::object casting example to embedding docs

kohr-h:
Some examples for auto-casting from `py::object` to subtypes. These were the simplest I could come up with; happy to take suggestions.

Closes: #1310

### DIFF
--- a/docs/advanced/embedding.rst
+++ b/docs/advanced/embedding.rst
@@ -105,6 +105,25 @@ The two approaches can also be combined:
         std::cout << message;
     }
 
+Typically, a call to a function in a `py::module` returns a generic `py::object`,
+which can often be autmatically cast to a specialized pybind11 type:
+
+.. code-block:: cpp
+
+    py::module os = py::module::import("os");
+    py::module path = py::module::import("os.path");  // like 'import os.path as path'
+    py::module np = py::module::import("numpy");  // like 'import numpy as np'
+
+    py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));
+    py::print(py::str("Current directory: ") + curdir_abs);
+    py::dict environ = os.attr("environ");
+    py::print(environ["HOME"]);
+    py::array_t<float> arr = np.attr("ones")(3, "dtype"_a="float32");
+    py::print(py::repr(arr + py::int_(1)));
+
+(Note that this example requires ``#include <pybind11/numpy.h>`` and
+``using namespace pybind11::literals;`` to work.)
+
 Importing modules
 =================
 
--- a/docs/advanced/embedding.rst
+++ b/docs/advanced/embedding.rst
@@ -105,25 +105,6 @@ The two approaches can also be combined:
         std::cout << message;
     }
 
-Typically, a call to a function in a `py::module` returns a generic `py::object`,
-which can often be autmatically cast to a specialized pybind11 type:
-
-.. code-block:: cpp
-
-    py::module os = py::module::import("os");
-    py::module path = py::module::import("os.path");  // like 'import os.path as path'
-    py::module np = py::module::import("numpy");  // like 'import numpy as np'
-
-    py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));
-    py::print(py::str("Current directory: ") + curdir_abs);
-    py::dict environ = os.attr("environ");
-    py::print(environ["HOME"]);
-    py::array_t<float> arr = np.attr("ones")(3, "dtype"_a="float32");
-    py::print(py::repr(arr + py::int_(1)));
-
-(Note that this example requires ``#include <pybind11/numpy.h>`` and
-``using namespace pybind11::literals;`` to work.)
-
 Importing modules
 =================
 
--- a/docs/advanced/pycpp/object.rst
+++ b/docs/advanced/pycpp/object.rst
@@ -64,39 +64,6 @@ This example obtains a reference to the Python ``Decimal`` class.
     py::object scipy = py::module::import("scipy");
     return scipy.attr("__version__");
 
-.. _implicit_casting:
-
-Implicit (automatic) casting
-============================
-
-Instead of using a generic :class:`object` as return type, it is possible to
-specialize to a subtype, e.g., to use the lookup syntax of :class:`dict`.
-
-.. note::
-    Casting to the wrong type will currently lead to downstream errors, e.g.,
-    when casting a Python dict to :class:`list` and using the ``append`` method.
-    In the future, such a bad cast will likely result in a :class:`cast_error`,
-    see `PR #2349 <https://github.com/pybind/pybind11/pull/2349>`_.
-
-.. code-block:: cpp
-    #include <pybind11/numpy.h>
-    using namespace pybind11::literals;
-
-    py::module os = py::module::import("os");
-    py::module path = py::module::import("os.path");  // like 'import os.path as path'
-    py::module np = py::module::import("numpy");  // like 'import numpy as np'
-
-    py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));
-    py::print(py::str("Current directory: ") + curdir_abs);
-    py::dict environ = os.attr("environ");
-    py::print(environ["HOME"]);
-    py::array_t<float> arr = np.attr("ones")(3, "dtype"_a="float32");
-    py::print(py::repr(arr + py::int_(1)));
-
-This constructor syntax is available for subclasses of :class:`object`; there
-is no need to call ``obj.cast()`` explicitly as for custom classes, see
-:ref:`_casting_back_and_forth`.
-
 
 .. _calling_python_functions:
 
@@ -211,6 +178,47 @@ Generalized unpacking according to PEP448_ is also supported:
     Python functions from C++, including keywords arguments and unpacking.
 
 .. _PEP448: https://www.python.org/dev/peps/pep-0448/
+
+.. _implicit_casting:
+
+Implicit casting
+================
+
+When using this C++ interface for Python types or calling Python functions,
+and objects of type :class:`object` are returned, it is possible to
+specialize to a subtype like :class:`dict`. The same holds for the proxy objects
+returned by ``operator[]`` or ``obj.attr()``.
+Casting to subtypes improves code readability and allows values to be used in
+C++ functions that require a specific subtype rather than a generic :class:`object`.
+
+.. code-block:: cpp
+
+    #include <pybind11/numpy.h>
+    using namespace pybind11::literals;
+
+    py::module os = py::module::import("os");
+    py::module path = py::module::import("os.path");  // like 'import os.path as path'
+    py::module np = py::module::import("numpy");  // like 'import numpy as np'
+
+    py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));
+    py::print(py::str("Current directory: ") + curdir_abs);
+    py::dict environ = os.attr("environ");
+    py::print(environ["HOME"]);
+    py::array_t<float> arr = np.attr("ones")(3, "dtype"_a="float32");
+    py::print(py::repr(arr + py::int_(1)));
+
+These implicit conversions are available for subclasses of :class:`object`; there
+is no need to call ``obj.cast()`` explicitly as for custom classes, see
+:ref:`_casting_back_and_forth`.
+
+.. note::
+    If a trivial conversion via move constructor is not possible, both implicit and
+    explicit casting (calling ``obj.cast()``) will attempt a "rich" conversion.
+    For instance, ``py::list env = os.attr("environ");`` will succeed and is
+    equivalent to the Python code ``env = list(os.environ)`` that produces a
+    list of the dict keys.
+
+..  TODO: Adapt text once PR #2349 has landed
 
 Handling exceptions
 ===================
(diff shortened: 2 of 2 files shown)


docs/advanced/embedding.rst
```diff
@@ -105,6 +105,25 @@ The two approaches can also be combined:
         std::cout << message;
     }
 
+Typically, a call to a function in a `py::module` returns a generic `py::object`,
+which can often be autmatically cast to a specialized pybind11 type:
```
bstaletic, line 109: Calling `py::module::import` returns a `py::module`.

YannickJadoul:
Note: #2349 will probably turn this into an error if it's nót the type you say it is (but doing that now is already wrong and potentially dangerous). But maybe we should make this more clear?

Also, I'm not sure this should be added to the embedding-specific part. This can also be useful for extension modules, so I'd suggest adding it to https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html ? (Not sure if that's the perfect place either, though.)

kohr-h:
> Note: #2349 will probably turn this into an error if it's nót the type you say it is (but doing that now is already wrong and potentially dangerous). But maybe we should make this more clear?

What do you mean by "this" (... into an error)? The code in the examples? 
What's your suggestion for this PR, adding a note on using the correct type on the LHS, or `auto` if we don't care?


> Also, I'm not sure this should be added to the embedding-specific part. This can also be useful for extension modules, so I'd suggest adding it to https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html ? (Not sure if that's the perfect place either, though.)

Okay sure, would that be under "Casting back and forth", as implicit cast without calling `obj.cast()`?

docs/advanced/embedding.rst
kohr-h: That's why I wrote "typically" :wink: 
But maybe that's not true either. Is it?

YannickJadoul:
> What do you mean by "this" (... into an error)? The code in the examples?
> What's your suggestion for this PR, adding a note on using the correct type on the LHS, or `auto` if we don't care?

No, those should be fine, actually. But this kind of approach will fail if the type is not the expected type (i.e., if an attribute is a `list` and you're assigning it to a `dict`). Currently, it might not fail immediately but maybe later (e.g. when setting with `d[...] = ...;`), but after #2349, it should fail immediately (if it gets merged, ofc).

For this PR, we might want to make clear that for e.g. `py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));`, you need to be sure that `curdir_abs` is a `str`? Or do you think that's implicit?
 
> Okay sure, would that be under "Casting back and forth", as implicit cast without calling `obj.cast()`?

Hmmm, maybe, but it would make that (now very succinct) section very long. What do you think about adding another section to the bottom?
I'm also fine with putting it somewhere else, if you can find something better. But I think this is more widely applicable than just for embedding :-)

kohr-h:
> > What do you mean by "this" (... into an error)? The code in the examples?
> > What's your suggestion for this PR, adding a note on using the correct type on the LHS, or `auto` if we don't care?
> 
> No, those should be fine, actually. But this kind of approach will fail if the type is not the expected type (i.e., if an attribute is a `list` and you're assigning it to a `dict`). Currently, it might not fail immediately but maybe later (e.g. when setting with `d[...] = ...;`), but after #2349, it should fail immediately (if it gets merged, ofc).
> 
> For this PR, we might want to make clear that for e.g. `py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));`, you need to be sure that `curdir_abs` is a `str`? Or do you think that's implicit?

Alright, got ya. I was actually assuming that the cast would fail early, but if that's not the case at the moment, a warning makes sense. If PR #2349 is likely to land, I will add a note that casting to the wrong type will lead to downstream errors, and later on to immediate cast errors.



> Hmmm, maybe, but it would make that (now very succinct) section very long. What do you think about adding another section to the bottom?

Okay, maybe "Implicit casting" or something? I would put it right below the "Casting back and forth".



> I'm also fine with putting it somewhere else, if you can find something better. But I think this is more widely applicable than just for embedding :-)

Yes I agree, embedding is only where **I** stumbled on it :-). I'll see where it fits best, but I suspect it will be somewhere under "Python types".

kohr-h:
I've moved the stuff over to `object.rst`. Please see if the formulations make sense, I'm not too firm with C++ terminology.

Also a question: how should I use things like `py::object` in the documentation to make sure that a link to the API doc is created? I went for :class:\`object\` now, but it would be nicer to write something like \`py::object\` where the link is correctly generated nonetheless. Any ideas?
In other places of the doc, this isn't handled consistently either. Some links are dead, some are just code snippets (double backtick).

docs/advanced/pycpp/object.rst
```diff
@@ -62,6 +64,40 @@ This example obtains a reference to the Python ``Decimal`` class.
     py::object scipy = py::module::import("scipy");
     return scipy.attr("__version__");
 
+.. _implicit_casting:
+
+Implicit (automatic) casting
```
YannickJadoul, line 69: Let's get rid of "automatic", here, maybe? I'm not sure if it helps and it is probably implicit in "implicit".

docs/advanced/pycpp/object.rst
```diff
@@ -62,6 +64,40 @@ This example obtains a reference to the Python ``Decimal`` class.
...
+    py::module np = py::module::import("numpy");  // like 'import numpy as np'
+
+    py::str curdir_abs = path.attr("abspath")(path.attr("curdir"));
+    py::print(py::str("Current directory: ") + curdir_abs);
+    py::dict environ = os.attr("environ");
+    py::print(environ["HOME"]);
+    py::array_t<float> arr = np.attr("ones")(3, "dtype"_a="float32");
+    py::print(py::repr(arr + py::int_(1)));
+
+This constructor syntax is available for subclasses of :class:`object`; there
```
YannickJadoul, line 96: "This constructor syntax"? What about just "These implicit conversions" ?

docs/advanced/pycpp/object.rst
```diff
@@ -62,6 +64,40 @@ This example obtains a reference to the Python ``Decimal`` class.
     py::object scipy = py::module::import("scipy");
     return scipy.attr("__version__");
 
+.. _implicit_casting:
+
+Implicit (automatic) casting
+============================
+
+Instead of using a generic :class:`object` as return type, it is possible to
```
YannickJadoul, line 72: This start of the paragraph is missing the answer to "When?".
What about e.g., "When using this C++ interface for Python types or calling Python functions, and `py::object` are returned, [...]. The same holds for the proxy objects returned from `operator[]` or `attr`."?

This might actually be a reason to move this down a few more sections, such that it's after the "Calling Python functions" and "Calling Python methods" (and then maybe after the arguments sections, because they're related?). But then again, there will always be this circular dependency, so you could just put a forward reference. I'll leave this up to your judgement.

docs/advanced/pycpp/object.rst
```diff
@@ -62,6 +64,40 @@ This example obtains a reference to the Python ``Decimal`` class.
...
+specialize to a subtype, e.g., to use the lookup syntax of :class:`dict`.
+
+.. note::
+    Casting to the wrong type will currently lead to downstream errors, e.g.,
+    when casting a Python dict to :class:`list` and using the ``append`` method.
+    In the future, such a bad cast will likely result in a :class:`cast_error`,
+    see `PR #2349 <https://github.com/pybind/pybind11/pull/2349>`_.
```
YannickJadoul, line 79: Maybe this isn't necessary to add? I'm a bit hesitant to add open issues/PRs to this part of the docs. I'd say it's up to #2349 to add something here.

YannickJadoul: But I guess it's fine as well. As you please.

(Feel free to ping #2349, so we won't forget there.)

YannickJadoul (commented):
Starts looking good to me! :-)

> Also a question: how should I use things like `py::object` in the documentation to make sure that a link to the API doc is created? I went for :class:`object` now, but it would be nicer to write something like `py::object` where the link is correctly generated nonetheless. Any ideas?
> In other places of the doc, this isn't handled consistently either. Some links are dead, some are just code snippets (double backtick).

As you say: it's not handled consistently. But it's nice to have this link there, so whichever works best with respect to that; :class:\`object\` then, probably? (Just \`object\` doesn't work?)

Also, you could report an issue on this, if there's broken links. It won't be at the top of the list of things to do, but it's useful to know!

kohr-h:
> As you say: it's not handled consistently. But it's nice to have this link there, so whichever works best with respect to that; :class:`object` then, probably? (Just `object` doesn't work?)

I suppose it would. Sphinx would probably do a best guess, starting from the primary `cpp` domain. May depend on whether `:any:` domain is activated in the config, though.

> Also, you could report an issue on this, if there's broken links. It won't be at the top of the list of things to do, but it's useful to know!

It seems like only a few links actually work, so an issue would cover the majority of documentation :sweat_smile:

docs/advanced/pycpp/object.rst
kohr-h: Good suggestion, thanks

docs/advanced/pycpp/object.rst
kohr-h: Hm you're right, it fits better a bit farther below. Also because the subsequent sections use `py::object` throughout. I've moved it down.

kohr-h: And thanks for the text suggestion, I've used parts of it.

docs/advanced/pycpp/object.rst
kohr-h: I turned the PR part into an RST comment.

On the larger context: I turns out (new to me) that `py::list lst = obj;` first tries to move-construct `lst` from `obj`, and if that's not possible, does something like `list(obj)` would do in Python. So the old claim that casting a `dict` to `py::list` would fail later on was wrong: it would create a list of the dict keys.

 I've adapted the note box accordingly. Please have a look.

docs/advanced/pycpp/object.rst
```diff
@@ -176,6 +179,47 @@ Generalized unpacking according to PEP448_ is also supported:
...
+:ref:`casting_back_and_forth`.
+
+.. note::
+    If a trivial conversion via move constructor is not possible, both implicit and
+    explicit casting (calling ``obj.cast()``) will attempt a "rich" conversion.
+    For instance, ``py::list env = os.attr("environ");`` will succeed and is
+    equivalent to the Python code ``env = list(os.environ)`` that produces a
+    list of the dict keys.
+
+..  TODO: Adapt text once PR #2349 has landed
```
YannickJadoul, line 221: Great, thanks! :-)

docs/advanced/pycpp/object.rst
YannickJadoul: Ha, it does? OK, then. This is also related to #1201, so with a bit of luck, we can straighten out things a bit more in the future anyway (conversions vs. moving vs. reinterpreting, etc.) !

YannickJadoul (approved):
Thanks a lot, @kohr-h! Looks good to me :-)

YannickJadoul:
@bstaletic, @henryiii, @EricCousineau-TRI: I think you might all be interested in this? If so, a quick look would be appreciated, and then this could go in soon :-)

henryiii (approved):

docs/advanced/pycpp/object.rst
```diff
@@ -176,6 +179,47 @@ Generalized unpacking according to PEP448_ is also supported:
...
 .. _PEP448: https://www.python.org/dev/peps/pep-0448/
 
+.. _implicit_casting:
+
+Implicit casting
+================
+
+When using this C++ interface for Python types or calling Python functions,
+and objects of type :class:`object` are returned, it is possible to
+specialize to a subtype like :class:`dict`. The same holds for the proxy objects
```
bstaletic, line 189: The first sentence is too long and somewhat difficult to read. I think this would be nicer to read.

When using the C++ interface for Python types, or calling Python functions, objects of type :class:`object` are returned. It is possible to invoke implicit conversions to a subclass like :class:`dict`.

Usually, I don't pay much attention to the wording of the docs, but this time it actually took me 3 readings to figure out what is this trying to convey.

docs/advanced/embedding.rst
EricCousineau-TRI: It is true, at least from my reading:
https://github.com/pybind/pybind11/blob/37f845a1dc171d981149a53f01c13c6ba85e0f2e/include/pybind11/pybind11.h#L914

I think the "typically" can be dropped, because in C++, it is `py::module` (the C++ type itself).

EricCousineau-TRI (approved):
Looks great!

FWIW I typically try to minimize how much Python logic I write using `pybind11`s interface API.

If I find myself writing more than a few `if` statements, more than say 15 lines, or with more complex semantics, I try to pass it directly to Python code.

Example from our codebase:
https://github.com/RobotLocomotion/drake/blob/8581223802b7790504529b9d66f1f1aab6eaaee9/bindings/pydrake/pydrake_pybind.h#L134-L139

docs/advanced/pycpp/object.rst
kohr-h: Thanks! That's definitely easier to read. I've adapted the text.

bstaletic (approved):

henryiii:
Thanks!

Issue #2388, opened 2020-08-12
refactor: Replace `h.get_type()` with `py::type(h)`


henryiii:
With the addition of #2364, a very natural extension to `py::type<T>()` would be `py::type(h)`, mimicking `py::cast`. I would propose we depreciate and eventually remove `h.get_type()` in favor of `py::type(h)`. Reasons:

* Free functions are generally better than member functions in C++; they are simpler, composable, etc.
* It would simplify the mixin, which currently gets added to several different classes, and keeps the API closer to Python.
* `h.str()` is already deprecated in favor of `py:str(h)`.

It is only used internally, not publicly in any docs or tests, so to me it seems it would read better (more python like) and simplify the mixin.

For example, randomly picking one of the 10 places it's currently used:

```cpp
if (!obj.get_type().is(matrix_type))
// Becomes
if (!py::type(obj).is(matrix_type))
```
For comparison, the Python would be:

```python
if not type(obj) is matrix_type:
```

(Discussion moved from #2364).

YannickJadoul:
Jup, you already convinced me in #2364! ;-)

YannickJadoul:
We could make this similar to `py::str`, though, and solve #1692 at the same time.

YannickJadoul:
(#2364 hasn't been released yet, so we can still change that interface?)

henryiii:
Technically, I think that issue would be solved by `py::type<std::string>()()`, if we supported str/int/etc. (which we don't yet, in that PR)? I don't think there was a need for a typeobject type expressed there, but rather a way to get the type as a handle?

YannickJadoul:
Right, but I'm thinking we could create a `class type : public object {`, just like with `str`? That would allow to e.g. use it in signatures, call `py::insinstance<py::type>(...)`, etc.

henryiii:
Then we would lose `py::type<T>()`, which only works if `py::type` is a function. I'm not sure we would really gain much from it being a class. Also, you'd run in to the same issue with str, where it can't be a handle since it's based on object.

YannickJadoul:
Agh, I had naively thought/hoped you could make make a templated `template <typename T> type() {}` constructor, but apparently that's not allowed for constructors :-(

Best I can then think of is:

```c++
class type {
public:
    template <typename T>
    static type of() { return type(typeid(T)); }

    explicit type(const std::type_info &type_info) : m_type_info(type_info) {}

    void print() {
        std::cout << m_type_info.name() << std::endl;
[...]

PR #2467, opened 2020-09-06
Change base parameter type in register_exception and exception constructor from PyObject* to handle

YannickJadoul:
Follow-up of #2465. This should be the same, but is the pybind11 alternative, rather than a raw C API type.

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -80,7 +80,7 @@ module and automatically converts any encountered exceptions of type ``CppExp``
 into Python exceptions of type ``PyExp``.
 
 It is possible to specify base class for the exception using the third
-parameter, a pointer to `PyObject`:
+parameter, a `handle`:
 
 .. code-block:: cpp
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1868,7 +1868,7 @@ template <typename type>
 class exception : public object {
 public:
     exception() = default;
-    exception(handle scope, const char *name, PyObject *base = PyExc_Exception) {
+    exception(handle scope, const char *name, handle base = PyExc_Exception) {
         std::string full_name = scope.attr("__name__").cast<std::string>() +
                                 std::string(".") + name;
         m_ptr = PyErr_NewException(const_cast<char *>(full_name.c_str()), base, NULL);
@@ -1901,7 +1901,7 @@ PYBIND11_NAMESPACE_END(detail)
 template <typename CppException>
 exception<CppException> &register_exception(handle scope,
                                             const char *name,
-                                            PyObject *base = PyExc_Exception) {
+                                            handle base = PyExc_Exception) {
     auto &ex = detail::get_exception_object<CppException>();
     if (!ex) ex = exception<CppException>(scope, name, base);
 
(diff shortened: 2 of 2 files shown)


bstaletic:
>     /usr/include/python3.7m/pyerrors.h:326:24: note: candidate function not viable: no known conversion from 'pybind11::handle' to 'PyObject *' (aka '_object *') for 2nd argument

So this is a breaking change. `py::handle` can not be constructed from a `PyObject*` without also specifying whether you want to borrow or steal the reference.

On top of that, you mentioned using `py::type` instead of `py::handle`?

YannickJadoul:
> > ```
> > /usr/include/python3.7m/pyerrors.h:326:24: note: candidate function not viable: no known conversion from 'pybind11::handle' to 'PyObject *' (aka '_object *') for 2nd argument
> > ```
> 
> So this is a breaking change. `py::handle` can not be constructed from a `PyObject*` without also specifying whether you want to borrow or steal the reference.

Huh, I thought you could. `handle` is only a wrapper around `PyObject*`, not doing any refcounting (borrow/steal are for `object`, when the refcount needs to be increased or not).

There's also an implicit conversion, so it's a bit weird that this doesn't work?
https://github.com/pybind/pybind11/blob/3bd0d7a8d58afed91a4aacf5d0cdb8443e107825/include/pybind11/pytypes.h#L179

> On top of that, you mentioned using `py::type` instead of `py::handle`?

Yes, but it's not there, yet. So indeed, let's link to #2388.

YannickJadoul:
> There's also an implicit conversion, so it's a bit weird that this doesn't work?

It's the opposite way that doesn't work implicitly, `handle` to `PyObject*`. Should've given the thing a compilation run before pushing "a harmless change" ;-)

bstaletic (approved):
Right, I was mixing things up with `py::object(py::handle)`, which doesn't work.

Anyway, this looks good now.

henryiii (approved):

michalsustr:
Does this mean that now it is possible to do:
```
py::register_exception<ExceptionA>(m, "ExceptionA", PyExc_RuntimeError);
py::register_exception<ExceptionB>(m, "ExceptionB", ExceptionA);
```
i.e. make a hierarchy of exceptions, starting with python ones and continuing into the ones defined on the C++ side?

YannickJadoul:
@michalsustr Not sure, actually. I would think it has always worked, but that before, you'd need to say `py::register_exception<ExceptionB>(m, "ExceptionB", ExceptionA.ptr());`, because `handle` is not implicitly convertible to a `PyObject*`.

YannickJadoul:
@michalsustr (and others), I quickly checked, and this seems to already work:

```python

namespace py = pybind11;

class X : public std::runtime_error {};
class Y : public std::runtime_error {};

PYBIND11_MODULE(example, m)
{
        auto exceptionA = py::register_exception<X>(m, "ExceptionA", PyExc_RuntimeError);
        py::register_exception<Y>(m, "ExceptionB", exceptionA.ptr());
}
```

So this PR just removes the need to add `.ptr()`, but that's all. So this PR shouldn't break/enable anything more?
It would of course still be nice to unify with `py::class_`, but that's a whole other expedition into unknown territory.

PR #2472, opened 2020-09-08
ci: disallow some common capitalization mistakes

henryiii:
Stop a few common mistakes (using pre-commit). Pulled the Linux-only annotation fix from #2433 since that one is not approved yet. :'(

- ci: only annotate Linux for now
- style: block some common mistakes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,14 @@ repos:
 
 - repo: local
   hooks:
+  - id: disallow-caps
+    name: Disallow improper capitalization
+    language: pygrep
+    entry: PyBind|Numpy|Cmake
+    exclude: .pre-commit-config.yaml
+
+- repo: local
+  hooks:
   - id: check-style
     name: Classic check-style
     language: system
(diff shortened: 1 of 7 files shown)


bstaletic (approved):

YannickJadoul (approved):

.pre-commit-config.yaml
```diff
@@ -34,6 +34,14 @@ repos:
     types: [file]
     files: (\.cmake|CMakeLists.txt)(.in)?$
 
+- repo: local
+  hooks:
+  - id: disallow-caps
+    name: Disallow improper capitalization
+    language: pygrep
+    entry: PyBind|Numpy|Cmake
```
YannickJadoul, line 42: So this is just a list of things we don't want to see?
I guess there's no better way to implemented this? There's no way to say "this is the way these words should be capitalized, and if you do encounter a different capitalization, it's wrong" ?

henryiii: This is just a simple Python regex (it's just a custom hook in pre-commit with `pygrep` as the language); if there's a program somewhere that is smarter, we could use that instead. If the regex matches, pre-commit shows the failing line and doesn't pass.

YannickJadoul: Well, if we encounter something, why not, later. I'll get this in now :-)

Issue #2407, opened 2020-08-18
[Feature request]  Support for Nvidia (former Portland Group) compilers


andriish:
Hi all,

I've tried to compile some Python bindings with the Nvidia compiler (https://developer.nvidia.com/hpc-sdk)  and it turns out it more or less works with a tiny exception that  one should guard the occurrences of 
``std::exception_ptr``  and  ``std::current_exception()`` with something like  

```
#if !defined (__PGIC__)
....
# endif
```
I've used  6 such guards in total. Four in the details/internals.h and two in the pybind11.h

I was wondering if that would be interesting for the developers to have this modifications and support the NVidia compiler suite in this way.

Best regards,
Andrii

henryiii:
Is there a way to test builds on them (ideally on GitHub Actions) (a docker container or such?)

andriish:
Hi @henryiii ,

I have a native installation (v20), but it looks the contaners are available as well
https://ngc.nvidia.com/catalog/containers/hpc:pgi-compilers (at least v19).

Best regards,

Andrii

YannickJadoul:
> one should guard the occurrences of  `std::exception_ptr` and `std::current_exception()` with something like

Is this safe to do? I thought pybind11 does depend on usage of these two symbols (e.g., for custom exception translators). So that means that we should disable all functionality using this?

andriish:
Hi @YannickJadoul ,

 >So that means that we should disable all functionality using this?

Only for this particular stl implementation as this implementation does not have the neede classes (if i'm not mistaken).

Optionally one can provide a simplistic implementation for these classes. Or even dummies.

Best regards,
Andrii

henryiii:
Running this:

```bash
docker run --rm -it -v $PWD:/pybind11 nvcr.io/hpc/pgi-compilers:ce
apt-get update && apt-get install -y python3-dev python3-pip python3-pytest
wget -qO- "https://cmake.org/files/v3.18/cmake-3.18.1-Linux-x86_64.tar.gz" | tar --strip-components=1 -xz -C /usr/local
cmake -S pybind11/ -B build
cmake --build build
```

(Eek, one layer alone of that docker image is over 5GB!)

<details><summary>Produces this: (click to expand)</summary>

```bash
cmroot@9a2d2cfec530:/# cmake -S pybind11/ -B build
-- The CXX compiler identification is PGI 19.10.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
[...]

Issue #2214, opened 2020-05-15
PGI compiler issues


XL64:
## Issue description

When building the tests with PGI Compilers some errors are raised.
To be able to build them I needed to allow more recursion :
```
  if (CMAKE_CXX_COMPILER_ID·STREQUAL·PGI)
    target_compile_options(${target_name} PRIVATE -Wc,--pending_instantiations=0)
  endif()
```
I also needed to comment out some test because they raised more complex errors: test_smart_ptr.cpp and  test_virtual_functions.cpp

I am not enough C++ expert to fix them:
```
cd /appli_RD/LACOSTE/DNG/diva-nextgen/build/pybind11-prefix/src/pybind11-build/tests && /data_local/sw/pgi/20.04/linuxpower/20.4/bin/pgc++  -DPYBIND11_TEST_EIGEN -Dpybind11_tests_EXPORTS -I/appli_RD/LACOSTE/DNG/diva-nextgen/build/pybind11-prefix/src/pybind11/include -I/data_local/sw/anaconda3/include/python3.7m -I/appli_RD/LACOSTE/OMEGA/cmakesuperbuild/build-gnu/eigen/src/eigen_project  -O2 -s -DNDEBUG -fPIC   -std=c++14 -Wc,--pending_instantiations=0 -o CMakeFiles/pybind11_tests.dir/test_smart_ptr.cpp.o -c /appli_RD/LACOSTE/DNG/diva-nextgen/build/pybind11-prefix/src/pybind11/tests/test_smart_ptr.cpp
At end of source: warning: routine is both "inline" and "noinline"

PGC++-S-0000-Internal compiler error. getsname: illegal stype   76807  (/usr/lib/gcc/ppc64le-redhat-linux/4.8.5/../../../../include/c++/4.8.5/tuple)
PGC++-S-0000-Internal compiler error. getsname: illegal stype   76808  (/usr/lib/gcc/ppc64le-redhat-linux/4.8.5/../../../../include/c++/4.8.5/tuple: 169)
PGC++-S-0000-Internal compiler error. getsname: illegal stype   76809  (/usr/lib/gcc/ppc64le-redhat-linux/4.8.5/../../../../include/c++/4.8.5/tuple: 172)
PGC++-S-0000-Internal compiler error. getsname: illegal stype   76810  (/usr/lib/gcc/ppc64le-redhat-linux/4.8.5/../../../../include/c++/4.8.5/tuple: 142)
[...]

PR #2475, opened 2020-09-09
Support NVidia HPC SDK (PGI)

andriish:
In this MR:
  - Support NVidia HPC SDK with PG compilers

This is work in progress, please see also #2407 and discussion with @henryiii


.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ on:
       - master
       - stable
       - v*
+      - support_pgc
```
henryiii, line 11: Fine for a WIP, but just to let you know, you can also trigger a build in the GitHub GUI by going to Actions -> CI -> Trigger Workflow -> pick branch (`support_pgc`).

include/pybind11/detail/internals.h
```diff
@@ -100,7 +100,9 @@ struct internals {
     std::unordered_set<std::pair<const PyObject *, const char *>, overload_hash> inactive_overload_cache;
     type_map<std::vector<bool (*)(PyObject *, void *&)>> direct_conversions;
     std::unordered_map<const PyObject *, std::vector<PyObject *>> patients;
+#if !defined (__PGIC__)
```
henryiii, line 103:
```suggestion
#ifndef __PGIC__
```

This is maybe a little cleaner for a single check?

tests/CMakeLists.txt
```diff
@@ -264,6 +265,15 @@ foreach(target ${test_targets})
   if(MSVC)
     target_compile_options(${target} PRIVATE /utf-8)
   endif()
+  if (${CMAKE_CXX_COMPILER_ID} MATCHES "PGI")
+    if (WIN32)
+      target_compile_options(${target} PRIVATE -Wc,--pending_instantiations=0 )
+    else()
+      target_compile_options(${target} PRIVATE -Wc,--pending_instantiations=0 -D__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1)
```
henryiii, line 272: Are these always required, or just for the tests? If always, maybe they should go into pybind11Common for pybind11 or some similar target.

tests/CMakeLists.txt
```diff
@@ -82,12 +82,13 @@ set(PYBIND11_TEST_FILES
...
     test_pytypes.cpp
     test_sequences_and_iterators.cpp
-    test_smart_ptr.cpp
+    #AV  this fails with PGI test_smart_ptr.cpp
     test_stl.cpp
     test_stl_binders.cpp
     test_tagbased_polymorphic.cpp
     test_union.cpp
-    test_virtual_functions.cpp)
+    #AV  this fails with PGI test_virtual_functions.cpp
```
henryiii, line 90: These should be removed if compiler is PGI. I can add that if you want me to.

andriish:
Hi @henryiii ,

just to avoid misunderstanding, I will quote myself from the #2407 
"I'm not sure if I will have any time to actually implement the full support, so the pull request is marked as work in progress.
Feel free to close it at any time.".
I mean that some things can be picked up from this merge request and might be useful. On some I can comment, the others would require some really dedicated work and changes in the build system. Not something that I have time for (even if I would like to in some cases). So most things are up you and other developers.

Best regards,
Andrii

.github/workflows/ci.yml
andriish: Hi @henryiii , 

OK. I was trying to find it but w/o any result.

Thanks in any case!

include/pybind11/detail/internals.h
andriish: That is a question of taste.
Later, when the exceptions will be implemented, that should be something like

```
#if !defined (__PGIC__)|| (__PGIC_MAJOR__ > 21)
```

tests/CMakeLists.txt
andriish: ``-Wc,--pending_instantiations=0`` is about the recursion level in templates. If I recall that correctly.
So, well, in some codes it will not be needed. But in the others the compilation will fail without this option.
Note that this option makes compilation significantly slower.

`-D__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1` was deduced from experiments with old compilers on Linux.

tests/CMakeLists.txt
andriish: Would be great!
I mean I haven't found an easy way ot exclude the execution of the python tests.

include/pybind11/detail/internals.h
henryiii: Okay, fair enough.

.github/workflows/ci.yml
```diff
@@ -289,6 +289,56 @@ jobs:
...
       run: cmake --build build --target test_cmake_build
 
+  centos-nvhpc:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        centos:
+          - 7  # GCC 4.8
+          - 8
```
henryiii, line 299: Any reason you are building twice, with identical compilers, just with a different base system? You aren't using the host GCC.

andriish: Hi @henryiii ,

There are differences, e.g. Python and cmake. But in general, I would consider CI to be an "environment test" rather than a test of some specific compiler version.  The later has little practical value, as for me. 

Best regards,
Andrii

henryiii:
Have you tried running the tests? It seems like many of them error out, mostly with things like "SystemError: Exception escaped from default exception translator!".

henryiii:
Looked into PGI a bit more and realized we are probably marking a fundamental mistake. Will fix tomorrow.

andriish:
> Have you tried running the tests

 Not the tests with exceptions. I see no hope for simple fixes for them.

> we are probably making a fundamental mistake

I'm intrigued. Do you mean that it is not possible to support their implementation of stl? Or the approach is wrong?
Ok. Will just wait.

henryiii:
I looked into PGI and it has complete C++11 support since 2015. So I then added `#include <exception>` (we do not run include-what-you-use, obviously) and that fixed all the issues with "missing" std::exception and such; PGI's stdlib is less "leaky" and we didn't get `<exception>` for free.

I re-enabled all the tests, and now only one test fails:

```
>       assert msg(capture) == strip_comments("""
            noisy new               # allocation required to attempt first overload
            noisy delete            # have to dealloc before considering factory init overload
            noisy new               # pointer factory calling "new", part 1: allocation
            NoisyAlloc(double 1.5)  # ... part two, invoking constructor
            ---
            ~NoisyAlloc()  # Destructor
            noisy delete   # operator delete
        """)
E           noisy new
E         + noisy delete
E           noisy new
E           NoisyAlloc(double 1.5)
E           ---
E           ~NoisyAlloc()
E           noisy delete

../../tests/test_factory_constructors.py:358: AssertionError
```

henryiii:
There are still quite a few warnings - we manually silence quite a few warnings for GCC/Clang, which obviously doesn't scale very well. I'm not sure how to manually silence them for PGI in-source.

andriish:
> are still quite a few warnings

```
          statement is unreachable
          return false;
```
Yes. That is why"more different compilers" is good for debug.

> how to manually silence them for PGI in-source.

https://forums.developer.nvidia.com/t/disable-unreachable-warnings/133575

Unfortunately that is "all on nothing".

include/pybind11/detail/internals.h
```diff
@@ -11,6 +11,8 @@
 
 #include "../pytypes.h"
 
+#include <exception>
```
henryiii, line 14: I'm putting these in the "standard" location, at the top of the file they are used in, rather than bunched together and included via other files. Assuming we use include-what-you-use, the rest of the includes will show up here eventually.

tests/CMakeLists.txt
```diff
@@ -16,6 +16,29 @@ else()
...
 
+# Only needed for CMake < 3.5 support
+include(CMakeParseArguments)
+
+# Filter out items; print an optional message if any items filtered
+#
+# Usage:
+#   pybind11_filter_tests(LISTNAME file1.cpp file2.cpp ... MESSAGE "")
+#
+macro(PYBIND11_FILTER_TESTS LISTNAME)
```
henryiii, line 27: Not really related, just a minor cleanup.

henryiii:
```
          statement is unreachable
          return false;
```

That's actually not a helpful warning; the code above is not clearly unreachable; and if it is, it's only because all the functions above are noexcept, which not all compilers could detect. We could hide that one warning with an `#if`, but there are several others we so not really worth it.

It's really a shame that you can't turn off single warnings with PGI (the old link might not be correct anymore, but regardless, it does not seem to be possible using pragmas). We have several that we can't avoid, like inline + noinline, which is intentional.

But it did help discover the missing `<exception>`, so that's good. :)

andriish:
Hi @henryiii 
1) Amazing, it passes the CI! Thanks a lot!

2) Some answers:

> That's actually not a helpful warning;

Completely true. But it is one of few compilers or even the only one that issues this warning. Quite useful to catch logical errors. Not here, but elsewhere, of course. 

>a shame that you can't turn off single warnings with PGI (

Exactly.

3) Some random ideas about the CI:

a) Practically the PGI is not much used so far, so that could make sense to use it only in regular nightly or weekly builds.
b) Another approach would be to actually make the CI fast:    tar the /opt directory with the installed PGI and upload it as an artifact. Then in CI download it and mount ( not even untar) the tarball to /opt. However, I have no idea if that would be much faster. 

c)  That could make sense to use the standard repositories for the installation, e.g. for CentOS
```
centos:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        centos:
          - 7  # GCC 4.8
          - 8
        include:
          - centos: 7
            cmakeexe: cmake3
          - centos: 8
            cmakeexe: cmake

    name: "🐍 3 • CentOS ${{ matrix.centos }} • x64"
    container: "centos:${{ matrix.centos }}"

    steps:
    - uses: actions/checkout@v2

    - name: Add Python 3
      run: |
        yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-${{ matrix.centos }}.noarch.rpm
        yum update -y && yum install -y python3-devel gcc-c++ make git eigen3-devel catch-devel ${{ cmakeexe }}

    - name: Update pip
      run: python3 -m pip install --upgrade pip

    - name: Install dependencies
      run: python3 -m pip install cmake -r tests/requirements.txt --prefer-binary

    - name: Configure
      shell: bash
      run: >
        ${{ cmakeexe }} -S . -B build
        -DPYBIND11_WERROR=ON
        -DCMAKE_CXX_STANDARD=11
        -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")


```
Hopefully, I have correct actions syntax there... Numpy and others are available as well, but it looks the version is not sufficient on CentOS7, I think. With a slightly lower requirements for the tests the whole CI could be executed using only the standard repositories w/o pip. But I don't know if that would be "pythonic way".

Best regards,

Andrii

henryiii:
> But it is one of few compilers or even the only one that issues this warning

Other compilers catch that one, but we have it disabled, see https://github.com/pybind/pybind11/blob/b47efd35fb436abe5c93db36b4ff5a1b5fdf593f/include/pybind11/pybind11.h#L30
for example. I'm actually quite disappointed by the small number of warnings (and poor quality) that PGI produces. Still always good to have more implementations being tested!

> tar the /opt directory with the installed

GHA has built in support for this, with the "cache" action, but there's a limit to how much you can store - 5GB per repo, I think - which should probably be completely covered by this one directory! The correct solution is to use a docker container - that's basically just a tar of your whole file system. I think GHA caches them, too; the clang & gcc jobs often finish before the "traditional" jobs.

> stock fedora images and run yum inside

The current docker images are much faster. Good to note, though.

> w/o pip

The current trend is to try to avoid using the system packager for Python, as it's (way) too slow in getting updates and not controllable. I grab pytest from it just because it's there, though.

---

I'm not too worried about the speed; the general consensus seemed to be that it's better to catch problems before merging then to have a more complicated CI setup (though it would be easy to do a daily/weekly workflow if we wanted to go that way). PyPy already is the slowest job, beating this one by a bit (and it does do a little caching!), and 15 jobs or so run in parallel, so the time to completion is unchanged with this PR.

If NVIDIA keeps up their docker container, we could switch to that one, which would probably save 3-4 minutes. I'm a little worried that it's been 1-2 months and the container is not updated. Since PGI is _so slow_ to compile, that's really the biggest slowdown here, though. Maybe after #2445 the compile portion will be shorter, and we can work on speeding things up further.

henryiii:
> Requires non-centalized maintenance, i.e. manpower that could be used for other things.

We do not maintain Docker's official GCC images, silkeh's Clang images, or NVIDIA's HPC images. The only one I'd be a little worried about is the Clang images, but for the past several years, silkeh has been very good at maintaining them, and responsive when I brought up an issue. They are also pretty heavily used nowadays. (and of course the worry I expressed about the HPC image being old, which is why we aren't using it yet here)

If you do need a customized docker image, GHA has some very nice infrastructure for it, I've created a few custom docker actions and they take very little maintenance. See https://github.com/CLIUtils/CLI11/tree/master/.github/actions/cmake_config - the build and caching are completely abstracted away by GitHub Actions, and I can run a dozen or more jobs in under 10 seconds each. See https://github.com/CLIUtils/CLI11/blob/c1666528ef0b1b5ddeacc60fba0f01f67cc7c7e8/.github/workflows/tests.yml#L37-L119

> but system repositories is what people use

I don't think that's all that true for Python. Almost everyone uses Pip; system repos may get used for a few dependencies (NumPy, for example, though mostly due to historical reasons, wheels for NumPy used to not be that good). Anyway, pip is quite fast, so not worried there. Actually, the only pip usage above is for CMake, since the new CLI is much nicer for setting up CI builds; I am using pytest from the system repo.

> One can play with it and set it to 100 or 50 

Okay, that's worth checking out! Thanks!

include/pybind11/pybind11.h
```diff
@@ -50,6 +50,8 @@
 #  include <cxxabi.h>
 #endif
 
+#include <exception>
```
wjakob, line 53: Seems redundant, can we add the header at just one of these places?

tests/CMakeLists.txt
```diff
@@ -296,6 +316,13 @@ foreach(target ${test_targets})
     target_compile_options(${target} PRIVATE /utf-8)
   endif()
 
+  if(${CMAKE_CXX_COMPILER_ID} MATCHES "PGI")
+    target_compile_options(${target} PRIVATE -Wc,--pending_instantiations=0)
+    if(NOT WIN32)
+      target_compile_definitions(${target} PRIVATE -D__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1)
```
wjakob, line 322: I think that mysterious define deserves an explanation, same for the --pending_instantiations

tests/CMakeLists.txt
wjakob: nice!

wjakob (commented):
Looks all good to me, just some minor comments.

andriish:
Hi @henryiii ,

it does not work on CentOS7 any more.
Will try to find out why.

Andrii

henryiii:
You have a duplicate key. Duplicate keys are caught by pre-commit, if you use that. Hint, hint. :)

include/pybind11/pybind11.h
henryiii: I would like to move toward include-what-you-use, where every header used in a file is listed at the top of that file, rather than assuming a different file will include it as a side affect. It's also a better basis for when C++20 modules come along. :)

Since we don't follow that style yet, I can put this in a single location, somewhere else, though. Let me know and I'll move it.

andriish:
Thanks for the fix!

andriish:
Hi @henryiii ,

OK. The guards around exception_ptr were removed. 

The original MR had them to disable the exception_ptr completely. But well, if that does not look reasonable as a first step to support PGC, one has to play with defines to make PGCC aware about this class.



Andrii

henryiii:
> one has to play with defines to make PGCC aware about this class.

No, it's just defined in `#include <exception>`, which pybind11 _did not explicitly include_. We were relying on the other files in the standard library including `<exception>` for us; PGI's standard library doesn't leak this include, so that's why it complained about this "missing". It's not missing, it's been there since at least 2015.

The problem here is we are compiling with C++14, but it's picking up (some) includes from CentOS 7's GCC 4.8, which doesn't support C++14. C++11 should be fine. But mixing stdlibs like this is dangerous, and is not pybind11's fault.

andriish:
Hi @henryiii ,

Ok. Let us try just compile it in CI on CentOS7.
Do you know how to diable BOOST?


Andrii

henryiii:
That is odd, installing PGI on CentOS 7 actively adds the GCC standard library, but it doesn't seem to support it (even in C++11 mode).

```
NVC++-S-0000-Internal compiler error. getsname: illegal stype   87113  (/usr/include/c++/4.8.5/tuple)
```

The newer GCC standard library on CentOS 8 is fine, as is the docker container. I'm not sure this is really our fault; this does seem to happen first when it's 84% done with the test suite, so you could just turn off this test for GCC stdlib 4.8 + PGI.

It's not actually fully using stdlibc++, though, since it does't automatically include `<exception>`, and GCC does.

---

I'm going to try removing those defines, to see if they are still needed. Locally this seems to be fine now that exception is included.

henryiii:
Wow, CentOS 8 is perfectly happy. CentOS 7 breaks due to the template issue. Have no idea what we'd add to detect if `-Wc,--pending_instantiations=0` is needed...

henryiii:
Looks like this has been seen before: https://github.com/pybind/pybind11/issues/2214

tests/CMakeLists.txt
henryiii: This is not needed, except for CentOS 7 (and the GCC thing was not needed at all). Added to the CI that runs CentOS 7 and removed from the CMakeLists entirely. :)

henryiii:
Okay, this is good, CentOS 7 compiles (though requires a few workarounds due PGI's issues with the stdlib). Users shouldn't have to mess with those normally, as it's only for a few corner cases in the test suite. Unless we want to move the `<exception>` include, I think this is ready to go in.

include/pybind11/pybind11.h
wjakob: I think that runs the risk of unnecessarily repeating many include directives all over the place. There is a core set of C++11 headers that are required by pybind11 (essentially any part of it). My preferred solution would be that we just put those into one of the core ``internals`` headers and be done with it.

henryiii: That's why includes have include guards, including multiple times is explicitly allowed. Having them bundled in one file makes it necessary for a reader to hunt down the include file by file to see what is included.  Includes are a useful tool for understanding a file, like a glossary. Also, when C++20 modules start showing up, this will become a requirement, as modules don't "leak" their includes from file to file. Also in the separation PR, #2445, some includes may not be needed on all compile units - having them only on the file they are used is much simpler/cleaner.

Totally not the point of this PR, though, so I'll try to find out which file this is supposed to go in.

henryiii: Moved!

andriish:
Hi @henryiii ,

>CentOS 7 compiles 

you mean with the defines or without?

Without I still have the same errors.
But the Ci passes....

Andrii

henryiii:
For CentOS 7, you have to add:

```
  -DCMAKE_CXX_FLAGS="-Wc,--pending_instantiations=0" \
  -DPYBIND11_TEST_FILTER="test_smart_ptr.cpp;test_virtual_functions.cpp"
```

This is _only_ for pybind11's tests - normal project may not need the extra template depth, and won't be likely to trigger the internal compiler error that these two tests trigger. And if they do, the solution is to just use a newer stdlib. :)

andriish:
Hi @henryiii ,

yes, I had it.  It still does not work. I also don't understad how it works on CI.

Also, is there a way to disable pytest and just compile everything?

Thse dependencieas are really really annoying. The versions are so random and there are no docs how to disable them.


Andrii

henryiii:
PyTest should be a warning, not a fatal error. I can fix that. Boost should _normally_ be fine as long as it's a new enough version; I take it Boost + PGI is an issue?

andriish:
Hi @henryiii ,

no, pytest is fatal.

The compilation fails even before.

Best regards,

Andrii

henryiii:
This is the compile line (since I'll remove the verbose flag next time):

```
cd /__w/pybind11/pybind11/build/tests &&
  /opt/nvidia/hpc_sdk/Linux_x86_64/20.7/compilers/bin/nvc++
-Dpybind11_tests_EXPORTS -I/__w/pybind11/pybind11/include -I/usr/include/python3.6m -Wc,--pending_instantiations=0 -O2 -s -DNDEBUG -fPIC --c++11 -A -o CMakeFiles/pybind11_tests.dir/test_stl.cpp.o -c /__w/pybind11/pybind11/tests/test_stl.cpp
```

henryiii:
> no, pytest is fatal.

Yes, I said I could fix it.

> The compilation fails even before.

Can you show me what is failing? This works in CI and locally for me, so I'd like to get it merged. I expect you have an issue locally, possibly picking up other things installed in your system since it's not in a clean container.

(Building your binaries in a container and then using them outside might not be a bad idea, by the way, if you are having trouble with a clean work environment).

henryiii:
I did not know about this flag! Nice:

https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html

So `-DCMAKE_DISABLE_FIND_PACKAGE_Boost=ON` will disable the search for Boost (first run only). :)

henryiii:
By the way, I'm pretty sure you will need a very recent version of CMake, CMake's support for "odd" compilers will not be great in older versions, much less if the compiler didn't even exist when CMake came out. So for 2020.7, you'll want CMake from mid 2020 or newer. See this page for help: https://cliutils.gitlab.io/modern-cmake/chapters/intro/installing.html

henryiii:
If you want to try things out, please make a new branch and work on your own account's CI that is not shared by pybind11. You are hammering our resources to try to debug a local issue, and stopping a useful PR from being merged.

andriish:
1) The error message 

```
 -c /home/andriish/Projects/pybind11/tests/pybind11_cross_module_tests.cpp
"/home/andriish/Projects/pybind11/include/pybind11/detail/internals.h", line
          103: error: namespace "std" has no member "exception_ptr"
                                       ^
```
2)

>  a very recent version of CMake

PGI is supported since a very long time. I also use system cmake 3.17 or so.


3)

>  picking up other things installed in your system since it's not in a clean container

I've tried it on two systems. Both have installations only from repositories. No scripts, no garbage in .bashrc, no virtual environments. The chances are higher that pip installs something unexpected in CI.



Best regards,

Andrii

henryiii:
Then use `wget -qO- "https://cmake.org/files/v3.18/cmake-3.18.2-Linux-x86_64.tar.gz" | tar --strip-components=1 -xz -C /usr/local` (on CI or docker, not locally, for obvious reasons)

henryiii:
Okay, I'll merge my pytest fix, and drop pip from the install.

PS: I didn't realize EPEL had updated CMake3 to 3.17, nice!

andriish:
cmake configures everything correctly. It should not be a problem.

The interesting part is how the compiler gets the definition of exception_ptr in one case but not in the other.

henryiii:
Can you check the file `exception` locally? Maybe even just try a 3-4 line test file, compiled by hand, just with exception_ptr?

henryiii:
I left the verbose in, wasn't too long. I've reordered that test and now pytest is no longer a FATAL_ERROR, so you can build without it.

andriish:
> Can you check the file exception locally? 

been there.
Literally the code from https://en.cppreference.com/w/cpp/error/exception_ptr

```
nvc++ 1.cc 
"1.cc", line 6: error: namespace "std" has no member "exception_ptr"
   void handle_eptr(std::exception_ptr eptr) // passing by value is ok
          
```
Also, grepping the nvidia installation shows that "exception_ptr" is located only in the libstdc++.ipl

henryiii:
I think it's getting it from libstdc++ on your system (GCC 4.8), not from the install.

andriish:
I mean the definition. There is just one header with exception_ptr definition and that definition is behind very funny condition expressions.

andriish:
Could you, please try this code locally on CentOS7 with and w/o defines?

```
#if defined(__PGIC__) and defined(__GXX_EXPERIMENTAL_CXX0X__) and (__cplusplus >= 201103L)
#ifndef __GCC_ATOMIC_TEST_AND_SET_TRUEVAL
#define __GCC_ATOMIC_TEST_AND_SET_TRUEVAL 1
#endif
#if !defined(ATOMIC_INT_LOCK_FREE)
#define ATOMIC_INT_LOCK_FREE 3
#include <bits/exception_ptr.h>
#undef ATOMIC_INT_LOCK_FREE
#endif
#endif


#include <iostream>
#include <string>
#include <exception>
#include <stdexcept>

void handle_eptr(std::exception_ptr eptr) // passing by value is ok
{
    try {
        if (eptr) {
            std::rethrow_exception(eptr);
        }
    } catch(const std::exception& e) {
        std::cout << "Caught exception \"" << e.what() << "\"\n";
    }
}

int main()
{
    std::exception_ptr eptr;
    try {
        std::string().at(1); // this generates an std::out_of_range
    } catch(...) {
        eptr = std::current_exception(); // capture
    }
    handle_eptr(eptr);
} // destructor for std::out_of_range called here, when the eptr is destructed

```

andriish:
with -std=c++11

henryiii:
I was able to reproduce locally (in docker). Very weird, no idea why this is different compared to docker on GHA! Try this:

```
cmake3 -S pybind11 -B build -DDOWNLOAD_CATCH=ON \
                              -DCMAKE_CXX_STANDARD=11 \
                             -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") \
                             -DCMAKE_CXX_FLAGS="-Wc,--pending_instantiations=0 -D__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1 -D__GCC_ATOMIC_INT_LOCK_FREE=3" \
                             -DPYBIND11_TEST_FILTER="test_smart_ptr.cpp;test_virtual_functions.cpp"
```

henryiii:
You are running natively, right (not containerized)?

andriish:
>  are running natively, right 

Yes. Always!  (no, sometimes in VB).


Ok, now more instanity.  PGI 19.10 works perfectly.

henryiii:
That compiles for me. Is "3" even a valid number for that? https://en.cppreference.com/w/c/atomic/ATOMIC_LOCK_FREE_consts

andriish:
Yes. Basically the <bits/exception_ptr.h> has condition  bla-bla>2.

henryiii:
I think PGI 20.7 doesn't support GCC 4.8 (even though it probably should if they are providing a repo for it, I bet it can be filed as a bug). The compiler is supposed to define that, and I think that's been dropped in 20.7. Except when running in GHA, in which case it magically _doesn't_ drop that. That's about where I am. :S

henryiii:
In short, since there is a command line workaround, I don't think we should hack the source. Probably report it as a bug with 20.7 and CentOS 7 and move on.

henryiii:
We should probably like to this PR in somewhere, so if it breaks we can add the workaround to the CI.

andriish:
>magically     

Container magic as is.

Ok. Agree to your suggestions. Basically to refine them:

 - put these frags into the cmake file with condition specific to this compiler major anf minor version

 - fill bug report to Nvidia. I will do this and put you to cc.

 - Merge if you are ok with the changes. That is totally up to you.

What is your opinion?


Best regards,

Andrii

henryiii:
I would not add the changes to CMake. Maybe put a note in. But this is a _user_ workaround that has nothing to do with pybind11 (it's needed to compile the example code from cppref above), and it's OS (probably stdlib) + compiler specific. Just a note in the changlog, perhaps, saying this is supported but on old systems, you may need the workaround in issue #2475.

henryiii:
(I'll try this locally on CentOS 8 later tonight)

andriish:
> But this is a user workaround 

Yes,  so that should not appear in the code, but it is ok to put it in the test suite.

I also think this is a pure bug, not an "unsupported version".

henryiii:
We don't understand the details of the bug, what versions it affects, or why the workaround is needed/what it affects. Hardcoding it into the test suite hides that the workaround is needed. Let's leave it as a note. (Unless another maintainer feels otherwise).

andriish:
Ok. Note is ok. One ca also issue a message during cmake configuration in case pgi is detected.

henryiii:
> message during cmake configuration in case pgi is detected

Once we understand it better, that's a possibility.

henryiii:
When I did this by hand, I got, in CentOS 8:

```
"/usr/include/c++/8/bits/atomic_base.h", line 210: error: identifier
          "__GCC_ATOMIC_TEST_AND_SET_TRUEVAL" is undefined
      { return __i ? __GCC_ATOMIC_TEST_AND_SET_TRUEVAL : 0; }
```

(Again, this is a problem out of the box, not with pybind11).

Looking here: https://forums.developer.nvidia.com/t/support-for-atomic-in-libstdc-missing/135403/4

It looks a bit to me like that it was rather designed for an HPC admin to create a siterc that has defines like this that describe the HPC hardware that you will be targeting (since you compile on a host node, not a compute node, in general, and they may not be homogenous). That also is likely why it uses environment-modules; this is really not intended for "normal" users to drop in and go. The docker container is set up properly, but the rest is "standard PGI" configuration that you have to do.

(I could totally be wrong here. I also rather don't think they are targeting normal users yet, just HPC clusters)

henryiii:
Let me know / CC me on the issue with NVIDIA! For now, this should now be better / compilable via flags or site configuration instead of source changes.

henryiii:
The rpm repo seems to be a little unstable. If it causes build to fail occasionally in the long run, I'll move to the container for the main test suite and make the two CentOS builds weekly or something like that.

YannickJadoul:
Congrats on getting this to work, @henryiii & @andriish! :-)

PR #2476, opened 2020-09-10
Use defined for PYBIND11_HAS_OPTIONAL, PYBIND11_HAS_EXP_OPTIONAL, PYBIND11_HAS_VARIANT

cirosantilli2:
These variables are not set in certain circumstances, and if the project
using pybind11 sets -Wundef, the warnings will show.

In all other usages outside of stl.h, ifdef/defined has already been used,
this commit converts the remaining ones under stl.h.

The test build is also modified to catch the problem.

The errors looked like:

```
ext/pybind11/include/pybind11/stl.h:292:5: error: "PYBIND11_HAS_OPTIONAL" is not defined, evaluates to 0 [-Werror=undef]
  292 | #if PYBIND11_HAS_OPTIONAL
      |     ^~~~~~~~~~~~~~~~~~~~~
ext/pybind11/include/pybind11/stl.h:300:5: error: "PYBIND11_HAS_EXP_OPTIONAL" is not defined, evaluates to 0 [-Werror=undef]
  300 | #if PYBIND11_HAS_EXP_OPTIONAL
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
ext/pybind11/include/pybind11/stl.h:372:5: error: "PYBIND11_HAS_VARIANT" is not defined, evaluates to 0 [-Werror=undef]
  372 | #if PYBIND11_HAS_VARIANT
      |     ^~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -289,7 +289,7 @@ template<typename T> struct optional_caster {
     PYBIND11_TYPE_CASTER(T, _("Optional[") + value_conv::name + _("]"));
 };
 
-#if PYBIND11_HAS_OPTIONAL
+#if defined(PYBIND11_HAS_OPTIONAL)
 template<typename T> struct type_caster<std::optional<T>>
     : public optional_caster<std::optional<T>> {};
 
@@ -297,7 +297,7 @@ template<> struct type_caster<std::nullopt_t>
     : public void_caster<std::nullopt_t> {};
 #endif
 
-#if PYBIND11_HAS_EXP_OPTIONAL
+#if defined(PYBIND11_HAS_EXP_OPTIONAL)
 template<typename T> struct type_caster<std::experimental::optional<T>>
     : public optional_caster<std::experimental::optional<T>> {};
 
@@ -369,7 +369,7 @@ struct variant_caster<V<Ts...>> {
     PYBIND11_TYPE_CASTER(Type, _("Union[") + detail::concat(make_caster<Ts>::name...) + _("]"));
 };
 
-#if PYBIND11_HAS_VARIANT
+#if defined(PYBIND11_HAS_VARIANT)
 template <typename... Ts>
 struct type_caster<std::variant<Ts...>> : variant_caster<std::variant<Ts...>> { };
 #endif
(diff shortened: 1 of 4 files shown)


henryiii:
Note this changes the behavior; defining this to 0 causes it to be used now, instead of skipped. A completely backward compatible version would be `#if defined(PYBIND11_HAS_OPTIONAL) && PYBIND11_HAS_OPTIONAL`.

YannickJadoul:
> Note this changes the behavior; defining this to 0 causes it to be used now, instead of skipped. A completely backward compatible version would be `#if defined(PYBIND11_HAS_OPTIONAL) && PYBIND11_HAS_OPTIONAL`.

We're detecting and defining this ourselves, now, so... I don't think this is documented/meant to be set from the outside?

henryiii:
> We're detecting and defining this ourselves, now, so... 

Probably okay, then. We need to change several `#if (__APPLE__)`'s to `#if defined(__APPLE__)` and such.

YannickJadoul:
(Just wait for @bstaletic to pop up and shout about [Hyrum's law](https://www.hyrumslaw.com/), though.)

But another argument: pybind11's code detecting it also doesn't check if something is already defined:

https://github.com/pybind/pybind11/blob/fbc7563623247f410b119f7cee44841411059cd7/include/pybind11/stl.h#L27-L49

So if someone manually defines it, it's 1) overwrtiten (I think ?), and 2) likely gives some warning. So even considering a mild version of Hyrum's law, we should be fine? :-)

Would you mind having a look for the other macros that @henryiii mentions, and adding them to this PR, @cirosantilli2? :-)

cirosantilli2:
I've updated the PR to update `__APPLE__` and `__clang__` too, there however only one instance of those not using `defined`. Let me know if you can spot any others.

henryiii:
Your warning addition should spot them 👍

henryiii:
```
/__w/pybind11/pybind11/tests/test_stl.cpp:18:5: error: 'PYBIND11_HAS_VARIANT' is not defined, evaluates to 0 [-Werror,-Wundef]
57
#if PYBIND11_HAS_VARIANT
```

YannickJadoul:
@cirosantilli2 If I'm not mistaken, warnings are treated as errors in the CI builds. So just follow the smell of the warnings/errors ;-)

henryiii (approved):

YannickJadoul (approved):

cirosantilli:
Thanks for the review!

PR #2477, opened 2020-09-10
Add check if `str(handle)` correctly converted the object, and throw py::error_already_set if not (bis)

YannickJadoul:
Same as #2392, already reviewed and merged into `str-bytes-cleanup`.
This are just the original commits cherry-picked, plus one additional fix to the tests (since #2380 is on `str-bytes-cleanup` but not on `master` yet).

### DIFF
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -104,10 +104,19 @@ def test_str(doc):
 
     assert m.str_from_object(A()) == "this is a str"
     assert m.repr_from_object(A()) == "this is a repr"
+    assert m.str_from_handle(A()) == "this is a str"
 
     s1, s2 = m.str_format()
     assert s1 == "1 + 2 = 3"
     assert s1 == s2
+
+    malformed_utf8 = b"\x80"
+    with pytest.raises(UnicodeDecodeError) as excinfo:
+        assert m.str_from_object(malformed_utf8)
+    assert 'invalid start byte' in str(excinfo.value)
+    with pytest.raises(UnicodeDecodeError) as excinfo:
+        assert m.str_from_handle(malformed_utf8)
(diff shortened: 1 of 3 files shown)


YannickJadoul:
@rwgk, yes, you were right in #2473: it is messy to get the old commits, since they are more intertwined with the other fixes. Apologies, #2473 should not per se have been closed, because I'm now ending up with basically the same PR.

But here we are, anyway. Since this already passed review, go ahead and merge this (I think it's the same as you had, except that I referred to the chimera-PR and left the intended test in place).

tests/test_pytypes.py
```diff
@@ -104,11 +104,23 @@ def __repr__(self):
...
     assert s1 == "1 + 2 = 3"
     assert s1 == s2
 
+    malformed_utf8 = b"\x80"
+    assert m.str_from_object(malformed_utf8) is malformed_utf8  # To be fixed; see #2380
+    if env.PY2:
+        # with pytest.raises(UnicodeDecodeError):
+        #     assert m.str_from_object(malformed_utf8)
+        with pytest.raises(UnicodeDecodeError):
+            assert m.str_from_handle(malformed_utf8)
```
rwgk, line 119: The assert here and in the commented out code above isn't needed.

rwgk (approved):
Just one minor nit. Please go ahead merging this PR.

tests/test_pytypes.py
YannickJadoul: Argh, you're right. Another difference I missed :-( Sorry; fixing!

rwgk:
Awesome @YannickJadoul, thanks for merging!

YannickJadoul:
Thanks for the reminder!

PR #2478, opened 2020-09-11
style: clang-tidy

henryiii:
This adds the ability to run a clang-tidy list of checks in CI, and adds a few checks to get started. Due to the methodology used, _only checks that made changes_ (one commit per check) were added. Only extremely safe checks were added. Eventually, it would be nice to expand further; there are some very useful/powerful clang tidy checks. You can see the [groups here](https://clang.llvm.org/extra/clang-tidy/). You can do groups of checks, like `modernize-*` (that's C++11), `readability-*`, `portability-*`, `bugprone-*`, `cppcoreguidelines-*`, or `performance-*` (and remove ones that you don't support). You can also disable inline for special cases.

Checks added:
* Namespaces always have a closing comment
* `override` is always used when overriding
* `.empty()` is used to check empty containers instead of size comparisons
* `using` used instead of `typedef`
* `= default` used for trivial bodies (except where MSVC 2015 requires otherwise)
* `auto` used _when the typename would be otherwise duplicated on the same line_ (not everywhere, as that would be worse, not better!)
* `emplace` used instead of push back for containers

I've also added instructions for running this locally (where you can perform the fixes automatically).

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -59,7 +59,7 @@ public:
         Py_CLEAR(ptr);
 
         // A heuristic to reduce the stack's capacity (e.g. after long recursive calls)
-        if (stack.capacity() > 16 && stack.size() != 0 && stack.capacity() / stack.size() > 2)
+        if (stack.capacity() > 16 && !stack.empty() && stack.capacity() / stack.size() > 2)
             stack.shrink_to_fit();
     }
 
@@ -163,7 +163,7 @@ inline const std::vector<detail::type_info *> &all_type_info(PyTypeObject *type)
  */
 PYBIND11_NOINLINE inline detail::type_info* get_type_info(PyTypeObject *type) {
     auto &bases = all_type_info(type);
-    if (bases.size() == 0)
+    if (bases.empty())
         return nullptr;
     if (bases.size() > 1)
         pybind11_fail("pybind11::detail::get_type_info: type has multiple pybind11-registered bases");
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -432,7 +432,7 @@ PYBIND11_NOINLINE inline std::string error_string() {
 
 #if !defined(PYPY_VERSION)
     if (scope.trace) {
-        PyTracebackObject *trace = (PyTracebackObject *) scope.trace;
+        auto *trace = (PyTracebackObject *) scope.trace;
 
         /* Get the deepest trace possible */
         while (trace->tb_next)
@@ -1244,7 +1244,7 @@ template <typename StringType, bool IsView = false> struct string_caster {
             load_src.ptr(), UTF_N == 8 ? "utf-8" : UTF_N == 16 ? "utf-16" : "utf-32", nullptr));
         if (!utfNbytes) { PyErr_Clear(); return false; }
 
-        const CharT *buffer = reinterpret_cast<const CharT *>(PYBIND11_BYTES_AS_STRING(utfNbytes.ptr()));
+        const auto *buffer = reinterpret_cast<const CharT *>(PYBIND11_BYTES_AS_STRING(utfNbytes.ptr()));
         size_t length = (size_t) PYBIND11_BYTES_SIZE(utfNbytes.ptr()) / sizeof(CharT);
         if (UTF_N > 8) { buffer++; length--; } // Skip BOM for UTF-16/32
         value = StringType(buffer, length);
@@ -1258,7 +1258,7 @@ template <typename StringType, bool IsView = false> struct string_caster {
 
     static handle cast(const StringType &src, return_value_policy /* policy */, handle /* parent */) {
         const char *buffer = reinterpret_cast<const char *>(src.data());
-        ssize_t nbytes = ssize_t(src.size() * sizeof(CharT));
+        auto nbytes = ssize_t(src.size() * sizeof(CharT));
         handle s = decode_utfN(buffer, nbytes);
         if (!s) throw error_already_set();
         return s;
@@ -1364,7 +1364,7 @@ public:
         // errors.  We also allow want to allow unicode characters U+0080 through U+00FF, as those
         // can fit into a single char value.
         if (StringCaster::UTF_N == 8 && str_len > 1 && str_len <= 4) {
-            unsigned char v0 = static_cast<unsigned char>(value[0]);
+            auto v0 = static_cast<unsigned char>(value[0]);
             size_t char0_bytes = !(v0 & 0x80) ? 1 : // low bits only: 0-127
                 (v0 & 0xE0) == 0xC0 ? 2 : // 0b110xxxxx - start of 2-byte sequence
                 (v0 & 0xF0) == 0xE0 ? 3 : // 0b1110xxxx - start of 3-byte sequence
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -220,7 +220,7 @@ struct value_and_holder {
     {}
 
     // Default constructor (used to signal a value-and-holder not found by get_value_and_holder())
-    value_and_holder() {}
+    value_and_holder() = default;
 
     // Used for past-the-end iterator
     value_and_holder(size_t index) : index{index} {}
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1929,7 +1929,7 @@ inline namespace literals {
     String literal version of `arg`
  \endrst */
 constexpr arg operator"" _a(const char *name, size_t) { return arg(name); }
-}
+} // namespace literals
 
 PYBIND11_NAMESPACE_BEGIN(detail)
 
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -30,7 +30,7 @@ private:
     object pywrite;
     object pyflush;
 
-    int overflow(int c) {
+    int overflow(int c) override {
         if (!traits_type::eq_int_type(c, traits_type::eof())) {
             *pptr() = traits_type::to_char_type(c);
             pbump(1);
@@ -38,7 +38,7 @@ private:
         return sync() == 0 ? traits_type::not_eof(c) : traits_type::eof();
     }
 
-    int sync() {
+    int sync() override {
         if (pbase() != pptr()) {
             // This subtraction cannot be negative, so dropping the sign
             str line(pbase(), static_cast<size_t>(pptr() - pbase()));
@@ -67,7 +67,7 @@ public:
     pythonbuf(pythonbuf&&) = default;
 
     /// Sync before destroy
-    ~pythonbuf() {
+    ~pythonbuf() override {
         sync();
     }
 };
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -38,7 +38,10 @@ private:
         return sync() == 0 ? traits_type::not_eof(c) : traits_type::eof();
     }
 
-    int sync() override {
+    // This function must be non-virtual to be called in a destructor. If the
+    // rare MSVC test failure shows up with this version, then this should be
+    // simplified to a fully qualified call.
+    int _sync() {
         if (pbase() != pptr()) {
             // This subtraction cannot be negative, so dropping the sign
             str line(pbase(), static_cast<size_t>(pptr() - pbase()));
@@ -52,6 +55,10 @@ private:
             setp(pbase(), epptr());
         }
         return 0;
+    }
+
+    int sync() override {
+        return _sync();
     }
 
 public:
@@ -68,7 +75,7 @@ public:
 
     /// Sync before destroy
     ~pythonbuf() override {
-        sync();
+        _sync();
     }
 };
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -165,7 +165,7 @@ protected:
             /* Get a pointer to the capture object */
             auto data = (sizeof(capture) <= sizeof(call.func.data)
                          ? &call.func.data : call.func.data[0]);
-            capture *cap = const_cast<capture *>(reinterpret_cast<const capture *>(data));
+            auto *cap = const_cast<capture *>(reinterpret_cast<const capture *>(data));
 
             /* Override policy for rvalues -- usually to enforce rvp::move on an rvalue */
             return_value_policy policy = return_value_policy_override<Return>::policy(call.func.policy);
@@ -420,7 +420,7 @@ protected:
         }
 
         /* Install docstring */
-        PyCFunctionObject *func = (PyCFunctionObject *) m_ptr;
+        auto *func = (PyCFunctionObject *) m_ptr;
         if (func->m_ml->ml_doc)
             std::free(const_cast<char *>(func->m_ml->ml_doc));
         func->m_ml->ml_doc = strdup(signatures.c_str());
@@ -465,7 +465,7 @@ protected:
                               *it = overloads;
 
         /* Need to know how many arguments + keyword arguments there are to pick the right overload */
-        const size_t n_args_in = (size_t) PyTuple_GET_SIZE(args_in);
+        const auto n_args_in = (size_t) PyTuple_GET_SIZE(args_in);
 
         handle parent = n_args_in > 0 ? PyTuple_GET_ITEM(args_in, 0) : nullptr,
                result = PYBIND11_TRY_NEXT_OVERLOAD;
@@ -860,7 +860,7 @@ public:
     explicit module(const char *name, const char *doc = nullptr) {
         if (!options::show_user_defined_docstrings()) doc = nullptr;
 #if PY_MAJOR_VERSION >= 3
-        PyModuleDef *def = new PyModuleDef();
+        auto *def = new PyModuleDef();
         std::memset(def, 0, sizeof(PyModuleDef));
         def->m_name = name;
         def->m_doc = doc;
@@ -1020,7 +1020,7 @@ protected:
     void install_buffer_funcs(
             buffer_info *(*get_buffer)(PyObject *, void *),
             void *get_buffer_data) {
-        PyHeapTypeObject *type = (PyHeapTypeObject*) m_ptr;
+        auto *type = (PyHeapTypeObject*) m_ptr;
         auto tinfo = detail::get_type_info(&type->ht_type);
 
         if (!type->ht_type.tp_as_buffer)
@@ -1242,7 +1242,7 @@ public:
 
     template <typename Func> class_& def_buffer(Func &&func) {
         struct capture { Func func; };
-        capture *ptr = new capture { std::forward<Func>(func) };
+        auto *ptr = new capture { std::forward<Func>(func) };
         install_buffer_funcs([](PyObject *obj, void *ptr) -> buffer_info* {
             detail::make_caster<type> caster;
             if (!caster.load(obj, false))
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -55,7 +55,7 @@ PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 /// Wraps an arbitrary C++ function/method/lambda function/.. into a callable Python object
 class cpp_function : public function {
 public:
-    cpp_function() { }
+    cpp_function() = default;
     cpp_function(std::nullptr_t) { }
 
     /// Construct a cpp_function from a vanilla function pointer
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1789,7 +1789,7 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
           typename KeyType = decltype((*std::declval<Iterator>()).first),
           typename... Extra>
 iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
-    typedef detail::iterator_state<Iterator, Sentinel, true, Policy> state;
+    using state = detail::iterator_state<Iterator, Sentinel, true, Policy>;
 
     if (!detail::get_type_info(typeid(state), false)) {
         class_<state>(handle(), "iterator", pybind11::module_local())
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -626,7 +626,7 @@ protected:
                 }
 
                 // 3. Check everything was consumed (unless we have a kwargs arg)
-                if (kwargs && kwargs.size() > 0 && !func.has_kwargs)
+                if (kwargs && !kwargs.empty() && !func.has_kwargs)
                     continue; // Unconsumed kwargs, but no py::kwargs argument to accept them
 
                 // 4a. If we have a py::args argument, create a new tuple with leftovers
@@ -814,7 +814,7 @@ protected:
             }
             if (kwargs_in) {
                 auto kwargs = reinterpret_borrow<dict>(kwargs_in);
-                if (kwargs.size() > 0) {
+                if (!kwargs.empty()) {
                     if (some_args) msg += "; ";
                     msg += "kwargs: ";
                     bool first = true;
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -543,7 +543,7 @@ protected:
                         self_value_and_holder.type->dealloc(self_value_and_holder);
 
                     call.init_self = PyTuple_GET_ITEM(args_in, 0);
-                    call.args.push_back(reinterpret_cast<PyObject *>(&self_value_and_holder));
+                    call.args.emplace_back(reinterpret_cast<PyObject *>(&self_value_and_holder));
                     call.args_convert.push_back(false);
                     ++args_copied;
                 }
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -425,14 +425,14 @@ template <int N> class BreaksBase { public:
 };
 template <int N> class BreaksTramp : public BreaksBase<N> {};
 // These should all compile just fine:
-typedef py::class_<BreaksBase<1>, std::unique_ptr<BreaksBase<1>>, BreaksTramp<1>> DoesntBreak1;
-typedef py::class_<BreaksBase<2>, BreaksTramp<2>, std::unique_ptr<BreaksBase<2>>> DoesntBreak2;
-typedef py::class_<BreaksBase<3>, std::unique_ptr<BreaksBase<3>>> DoesntBreak3;
-typedef py::class_<BreaksBase<4>, BreaksTramp<4>> DoesntBreak4;
-typedef py::class_<BreaksBase<5>> DoesntBreak5;
-typedef py::class_<BreaksBase<6>, std::shared_ptr<BreaksBase<6>>, BreaksTramp<6>> DoesntBreak6;
-typedef py::class_<BreaksBase<7>, BreaksTramp<7>, std::shared_ptr<BreaksBase<7>>> DoesntBreak7;
-typedef py::class_<BreaksBase<8>, std::shared_ptr<BreaksBase<8>>> DoesntBreak8;
+using DoesntBreak1 = py::class_<BreaksBase<1>, std::unique_ptr<BreaksBase<1>>, BreaksTramp<1>>;
+using DoesntBreak2 = py::class_<BreaksBase<2>, BreaksTramp<2>, std::unique_ptr<BreaksBase<2>>>;
+using DoesntBreak3 = py::class_<BreaksBase<3>, std::unique_ptr<BreaksBase<3>>>;
+using DoesntBreak4 = py::class_<BreaksBase<4>, BreaksTramp<4>>;
+using DoesntBreak5 = py::class_<BreaksBase<5>>;
+using DoesntBreak6 = py::class_<BreaksBase<6>, std::shared_ptr<BreaksBase<6>>, BreaksTramp<6>>;
+using DoesntBreak7 = py::class_<BreaksBase<7>, BreaksTramp<7>, std::shared_ptr<BreaksBase<7>>>;
+using DoesntBreak8 = py::class_<BreaksBase<8>, std::shared_ptr<BreaksBase<8>>>;
 #define CHECK_BASE(N) static_assert(std::is_same<typename DoesntBreak##N::type, BreaksBase<N>>::value, \
         "DoesntBreak" #N " has wrong type!")
 CHECK_BASE(1); CHECK_BASE(2); CHECK_BASE(3); CHECK_BASE(4); CHECK_BASE(5); CHECK_BASE(6); CHECK_BASE(7); CHECK_BASE(8);
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -103,7 +103,7 @@ TEST_SUBMODULE(class_, m) {
         BaseClass() = default;
         BaseClass(const BaseClass &) = default;
         BaseClass(BaseClass &&) = default;
-        virtual ~BaseClass() {}
+        virtual ~BaseClass() = default;
     };
     struct DerivedClass1 : BaseClass { };
     struct DerivedClass2 : BaseClass { };
@@ -353,7 +353,7 @@ TEST_SUBMODULE(class_, m) {
     // test_reentrant_implicit_conversion_failure
     // #1035: issue with runaway reentrant implicit conversion
     struct BogusImplicitConversion {
-        BogusImplicitConversion(const BogusImplicitConversion &) { }
+        BogusImplicitConversion(const BogusImplicitConversion &) = default;
     };
 
     py::class_<BogusImplicitConversion>(m, "BogusImplicitConversion")
@@ -407,7 +407,7 @@ TEST_SUBMODULE(class_, m) {
     py::class_<IsNonFinalFinal>(m, "IsNonFinalFinal", py::is_final());
 
     struct PyPrintDestructor {
-        PyPrintDestructor() {}
+        PyPrintDestructor() = default;
         ~PyPrintDestructor() {
             py::print("Print from destructor");
         }
--- a/tests/test_factory_constructors.cpp
+++ b/tests/test_factory_constructors.cpp
@@ -58,13 +58,13 @@ class TestFactory4 : public TestFactory3 {
 public:
     TestFactory4() : TestFactory3() { print_default_created(this); }
     TestFactory4(int v) : TestFactory3(v) { print_created(this, v); }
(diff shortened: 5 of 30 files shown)


henryiii:
<details><summary>FYI, this is the full list of checks in Clang 10: (click to expand)</summary>

```
abseil-duration-addition
    abseil-duration-comparison
    abseil-duration-conversion-cast
    abseil-duration-division
    abseil-duration-factory-float
    abseil-duration-factory-scale
    abseil-duration-subtraction
    abseil-duration-unnecessary-conversion
    abseil-faster-strsplit-delimiter
    abseil-no-internal-dependencies
    abseil-no-namespace
    abseil-redundant-strcat-calls
    abseil-str-cat-append
    abseil-string-find-startswith
    abseil-time-comparison
    abseil-time-subtraction
    abseil-upgrade-duration-conversions
    android-cloexec-accept
    android-cloexec-accept4
    android-cloexec-creat
    android-cloexec-dup
    android-cloexec-epoll-create
    android-cloexec-epoll-create1
    android-cloexec-fopen
    android-cloexec-inotify-init
    android-cloexec-inotify-init1
    android-cloexec-memfd-create
    android-cloexec-open
    android-cloexec-pipe
    android-cloexec-pipe2
    android-cloexec-socket
    android-comparison-in-temp-failure-retry
    boost-use-to-string
    bugprone-argument-comment
    bugprone-assert-side-effect
    bugprone-bad-signal-to-kill-thread
    bugprone-bool-pointer-implicit-conversion
    bugprone-branch-clone
    bugprone-copy-constructor-init
    bugprone-dangling-handle
    bugprone-dynamic-static-initializers
    bugprone-exception-escape
    bugprone-fold-init-type
    bugprone-forward-declaration-namespace
    bugprone-forwarding-reference-overload
    bugprone-inaccurate-erase
    bugprone-incorrect-roundings
    bugprone-infinite-loop
    bugprone-integer-division
    bugprone-lambda-function-name
    bugprone-macro-parentheses
    bugprone-macro-repeated-side-effects
    bugprone-misplaced-operator-in-strlen-in-alloc
    bugprone-misplaced-widening-cast
    bugprone-move-forwarding-reference
    bugprone-multiple-statement-macro
    bugprone-narrowing-conversions
    bugprone-not-null-terminated-result
    bugprone-parent-virtual-call
    bugprone-posix-return
    bugprone-signed-char-misuse
    bugprone-sizeof-container
    bugprone-sizeof-expression
    bugprone-string-constructor
    bugprone-string-integer-assignment
    bugprone-string-literal-with-embedded-nul
    bugprone-suspicious-enum-usage
    bugprone-suspicious-memset-usage
    bugprone-suspicious-missing-comma
    bugprone-suspicious-semicolon
    bugprone-suspicious-string-compare
    bugprone-swapped-arguments
    bugprone-terminating-continue
    bugprone-throw-keyword-missing
    bugprone-too-small-loop-variable
    bugprone-undefined-memory-manipulation
    bugprone-undelegated-constructor
    bugprone-unhandled-self-assignment
    bugprone-unused-raii
    bugprone-unused-return-value
    bugprone-use-after-move
    bugprone-virtual-near-miss
    cert-dcl03-c
    cert-dcl16-c
    cert-dcl21-cpp
    cert-dcl50-cpp
    cert-dcl54-cpp
    cert-dcl58-cpp
    cert-dcl59-cpp
    cert-env33-c
    cert-err09-cpp
    cert-err34-c
    cert-err52-cpp
    cert-err58-cpp
    cert-err60-cpp
    cert-err61-cpp
    cert-fio38-c
    cert-flp30-c
    cert-mem57-cpp
    cert-msc30-c
    cert-msc32-c
    cert-msc50-cpp
    cert-msc51-cpp
    cert-oop11-cpp
    cert-oop54-cpp
    cert-oop58-cpp
    cert-pos44-c
    clang-analyzer-apiModeling.StdCLibraryFunctions
    clang-analyzer-apiModeling.TrustNonnull
    clang-analyzer-apiModeling.google.GTest
    clang-analyzer-apiModeling.llvm.CastValue
    clang-analyzer-apiModeling.llvm.ReturnValue
    clang-analyzer-core.CallAndMessage
    clang-analyzer-core.DivideZero
    clang-analyzer-core.DynamicTypePropagation
    clang-analyzer-core.NonNullParamChecker
    clang-analyzer-core.NonnilStringConstants
    clang-analyzer-core.NullDereference
    clang-analyzer-core.StackAddrEscapeBase
    clang-analyzer-core.StackAddressEscape
    clang-analyzer-core.UndefinedBinaryOperatorResult
    clang-analyzer-core.VLASize
    clang-analyzer-core.builtin.BuiltinFunctions
    clang-analyzer-core.builtin.NoReturnFunctions
    clang-analyzer-core.uninitialized.ArraySubscript
    clang-analyzer-core.uninitialized.Assign
    clang-analyzer-core.uninitialized.Branch
    clang-analyzer-core.uninitialized.CapturedBlockVariable
    clang-analyzer-core.uninitialized.UndefReturn
    clang-analyzer-cplusplus.InnerPointer
    clang-analyzer-cplusplus.Move
    clang-analyzer-cplusplus.NewDelete
    clang-analyzer-cplusplus.NewDeleteLeaks
    clang-analyzer-cplusplus.PureVirtualCall
    clang-analyzer-cplusplus.SelfAssignment
    clang-analyzer-cplusplus.SmartPtr
    clang-analyzer-cplusplus.VirtualCallModeling
    clang-analyzer-deadcode.DeadStores
    clang-analyzer-fuchsia.HandleChecker
    clang-analyzer-nullability.NullPassedToNonnull
    clang-analyzer-nullability.NullReturnedFromNonnull
    clang-analyzer-nullability.NullabilityBase
    clang-analyzer-nullability.NullableDereferenced
    clang-analyzer-nullability.NullablePassedToNonnull
    clang-analyzer-nullability.NullableReturnedFromNonnull
    clang-analyzer-optin.cplusplus.UninitializedObject
    clang-analyzer-optin.cplusplus.VirtualCall
    clang-analyzer-optin.mpi.MPI-Checker
    clang-analyzer-optin.osx.OSObjectCStyleCast
    clang-analyzer-optin.osx.cocoa.localizability.EmptyLocalizationContextChecker
    clang-analyzer-optin.osx.cocoa.localizability.NonLocalizedStringChecker
    clang-analyzer-optin.performance.GCDAntipattern
    clang-analyzer-optin.performance.Padding
    clang-analyzer-optin.portability.UnixAPI
    clang-analyzer-osx.API
    clang-analyzer-osx.MIG
    clang-analyzer-osx.NSOrCFErrorDerefChecker
    clang-analyzer-osx.NumberObjectConversion
    clang-analyzer-osx.OSObjectRetainCount
    clang-analyzer-osx.ObjCProperty
    clang-analyzer-osx.SecKeychainAPI
    clang-analyzer-osx.cocoa.AtSync
    clang-analyzer-osx.cocoa.AutoreleaseWrite
    clang-analyzer-osx.cocoa.ClassRelease
    clang-analyzer-osx.cocoa.Dealloc
    clang-analyzer-osx.cocoa.IncompatibleMethodTypes
    clang-analyzer-osx.cocoa.Loops
    clang-analyzer-osx.cocoa.MissingSuperCall
    clang-analyzer-osx.cocoa.NSAutoreleasePool
    clang-analyzer-osx.cocoa.NSError
    clang-analyzer-osx.cocoa.NilArg
    clang-analyzer-osx.cocoa.NonNilReturnValue
    clang-analyzer-osx.cocoa.ObjCGenerics
    clang-analyzer-osx.cocoa.RetainCount
    clang-analyzer-osx.cocoa.RetainCountBase
    clang-analyzer-osx.cocoa.RunLoopAutoreleaseLeak
    clang-analyzer-osx.cocoa.SelfInit
    clang-analyzer-osx.cocoa.SuperDealloc
    clang-analyzer-osx.cocoa.UnusedIvars
    clang-analyzer-osx.cocoa.VariadicMethodTypes
    clang-analyzer-osx.coreFoundation.CFError
    clang-analyzer-osx.coreFoundation.CFNumber
    clang-analyzer-osx.coreFoundation.CFRetainRelease
    clang-analyzer-osx.coreFoundation.containers.OutOfBounds
    clang-analyzer-osx.coreFoundation.containers.PointerSizedValues
    clang-analyzer-security.FloatLoopCounter
    clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling
    clang-analyzer-security.insecureAPI.SecuritySyntaxChecker
    clang-analyzer-security.insecureAPI.UncheckedReturn
    clang-analyzer-security.insecureAPI.bcmp
    clang-analyzer-security.insecureAPI.bcopy
    clang-analyzer-security.insecureAPI.bzero
    clang-analyzer-security.insecureAPI.decodeValueOfObjCType
    clang-analyzer-security.insecureAPI.getpw
    clang-analyzer-security.insecureAPI.gets
    clang-analyzer-security.insecureAPI.mkstemp
    clang-analyzer-security.insecureAPI.mktemp
    clang-analyzer-security.insecureAPI.rand
    clang-analyzer-security.insecureAPI.strcpy
    clang-analyzer-security.insecureAPI.vfork
    clang-analyzer-unix.API
    clang-analyzer-unix.DynamicMemoryModeling
    clang-analyzer-unix.Malloc
    clang-analyzer-unix.MallocSizeof
    clang-analyzer-unix.MismatchedDeallocator
    clang-analyzer-unix.Vfork
    clang-analyzer-unix.cstring.BadSizeArg
    clang-analyzer-unix.cstring.CStringModeling
    clang-analyzer-unix.cstring.NullArg
    clang-analyzer-valist.CopyToSelf
    clang-analyzer-valist.Uninitialized
    clang-analyzer-valist.Unterminated
    clang-analyzer-valist.ValistBase
    cppcoreguidelines-avoid-c-arrays
    cppcoreguidelines-avoid-goto
    cppcoreguidelines-avoid-magic-numbers
    cppcoreguidelines-c-copy-assignment-signature
    cppcoreguidelines-explicit-virtual-functions
    cppcoreguidelines-init-variables
    cppcoreguidelines-interfaces-global-init
    cppcoreguidelines-macro-usage
    cppcoreguidelines-narrowing-conversions
    cppcoreguidelines-no-malloc
    cppcoreguidelines-non-private-member-variables-in-classes
    cppcoreguidelines-owning-memory
    cppcoreguidelines-pro-bounds-array-to-pointer-decay
    cppcoreguidelines-pro-bounds-constant-array-index
    cppcoreguidelines-pro-bounds-pointer-arithmetic
    cppcoreguidelines-pro-type-const-cast
    cppcoreguidelines-pro-type-cstyle-cast
    cppcoreguidelines-pro-type-member-init
    cppcoreguidelines-pro-type-reinterpret-cast
    cppcoreguidelines-pro-type-static-cast-downcast
    cppcoreguidelines-pro-type-union-access
    cppcoreguidelines-pro-type-vararg
    cppcoreguidelines-slicing
    cppcoreguidelines-special-member-functions
    darwin-avoid-spinlock
    darwin-dispatch-once-nonstatic
    fuchsia-default-arguments-calls
    fuchsia-default-arguments-declarations
    fuchsia-header-anon-namespaces
    fuchsia-multiple-inheritance
    fuchsia-overloaded-operator
    fuchsia-restrict-system-includes
    fuchsia-statically-constructed-objects
    fuchsia-trailing-return
    fuchsia-virtual-inheritance
    google-build-explicit-make-pair
    google-build-namespaces
    google-build-using-namespace
    google-default-arguments
    google-explicit-constructor
    google-global-names-in-headers
    google-objc-avoid-nsobject-new
    google-objc-avoid-throwing-exception
    google-objc-function-naming
    google-objc-global-variable-declaration
    google-readability-avoid-underscore-in-googletest-name
    google-readability-braces-around-statements
    google-readability-casting
    google-readability-function-size
    google-readability-namespace-comments
    google-readability-todo
    google-runtime-int
    google-runtime-operator
    google-runtime-references
    google-upgrade-googletest-case
    hicpp-avoid-c-arrays
    hicpp-avoid-goto
    hicpp-braces-around-statements
    hicpp-deprecated-headers
    hicpp-exception-baseclass
    hicpp-explicit-conversions
    hicpp-function-size
    hicpp-invalid-access-moved
    hicpp-member-init
    hicpp-move-const-arg
    hicpp-multiway-paths-covered
    hicpp-named-parameter
    hicpp-new-delete-operators
    hicpp-no-array-decay
    hicpp-no-assembler
    hicpp-no-malloc
    hicpp-noexcept-move
    hicpp-signed-bitwise
    hicpp-special-member-functions
    hicpp-static-assert
    hicpp-undelegated-constructor
    hicpp-uppercase-literal-suffix
    hicpp-use-auto
    hicpp-use-emplace
    hicpp-use-equals-default
    hicpp-use-equals-delete
    hicpp-use-noexcept
    hicpp-use-nullptr
    hicpp-use-override
    hicpp-vararg
    linuxkernel-must-check-errs
    llvm-header-guard
    llvm-include-order
    llvm-namespace-comment
    llvm-prefer-isa-or-dyn-cast-in-conditionals
    llvm-prefer-register-over-unsigned
    llvm-qualified-auto
    llvm-twine-local
    misc-definitions-in-headers
    misc-misplaced-const
    misc-new-delete-overloads
    misc-non-copyable-objects
    misc-non-private-member-variables-in-classes
    misc-redundant-expression
    misc-static-assert
    misc-throw-by-value-catch-by-reference
    misc-unconventional-assign-operator
    misc-uniqueptr-reset-release
    misc-unused-alias-decls
    misc-unused-parameters
    misc-unused-using-decls
    modernize-avoid-bind
    modernize-avoid-c-arrays
    modernize-concat-nested-namespaces
    modernize-deprecated-headers
    modernize-deprecated-ios-base-aliases
    modernize-loop-convert
    modernize-make-shared
    modernize-make-unique
    modernize-pass-by-value
    modernize-raw-string-literal
    modernize-redundant-void-arg
    modernize-replace-auto-ptr
    modernize-replace-random-shuffle
    modernize-return-braced-init-list
    modernize-shrink-to-fit
    modernize-unary-static-assert
    modernize-use-auto
    modernize-use-bool-literals
    modernize-use-default-member-init
    modernize-use-emplace
    modernize-use-equals-default
    modernize-use-equals-delete
    modernize-use-nodiscard
    modernize-use-noexcept
    modernize-use-nullptr
    modernize-use-override
    modernize-use-trailing-return-type
    modernize-use-transparent-functors
    modernize-use-uncaught-exceptions
    modernize-use-using
    mpi-buffer-deref
    mpi-type-mismatch
    objc-avoid-nserror-init
    objc-forbidden-subclassing
    objc-missing-hash
    objc-property-declaration
    objc-super-self
    openmp-exception-escape
    openmp-use-default-none
    performance-faster-string-find
    performance-for-range-copy
    performance-implicit-conversion-in-loop
    performance-inefficient-algorithm
    performance-inefficient-string-concatenation
    performance-inefficient-vector-operation
    performance-move-const-arg
    performance-move-constructor-init
    performance-no-automatic-move
    performance-noexcept-move-constructor
    performance-trivially-destructible
    performance-type-promotion-in-math-fn
    performance-unnecessary-copy-initialization
    performance-unnecessary-value-param
    portability-simd-intrinsics
    readability-avoid-const-params-in-decls
    readability-braces-around-statements
    readability-const-return-type
    readability-container-size-empty
    readability-convert-member-functions-to-static
    readability-delete-null-pointer
    readability-deleted-default
    readability-else-after-return
    readability-function-size
    readability-identifier-naming
    readability-implicit-bool-conversion
    readability-inconsistent-declaration-parameter-name
    readability-isolate-declaration
    readability-magic-numbers
    readability-make-member-function-const
    readability-misleading-indentation
    readability-misplaced-array-index
    readability-named-parameter
    readability-non-const-parameter
    readability-qualified-auto
    readability-redundant-access-specifiers
    readability-redundant-control-flow
    readability-redundant-declaration
    readability-redundant-function-ptr-dereference
    readability-redundant-member-init
    readability-redundant-preprocessor
    readability-redundant-smartptr-get
    readability-redundant-string-cstr
    readability-redundant-string-init
    readability-simplify-boolean-expr
    readability-simplify-subscript-expr
    readability-static-accessed-through-instance
    readability-static-definition-in-anonymous-namespace
    readability-string-compare
    readability-uniqueptr-delete-release
    readability-uppercase-literal-suffix
    zircon-temporary-objects
```

</details>

The goal is _not_ to enable them all, as many of them conflict (google vs. llvm, etc), but to just enable the ones or the groups (minus a few) that are useful.

henryiii:
I think that this solves the occasional output failure on Windows CI (and I think I've seen it in GooFit on Unix) - the destructor was incorrect, as discovered by the default analyzer checks in clang-tidy.

Here's the page on possible solutions, I chose the private member one, though the fully qualified one might work too. https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors

bstaletic:
> I chose the private member one, though the fully qualified one might work too.

That does silence the warning, but it shouldn't have any other effect. When executing the destructor, every member function is treated as if it were non-virtual. All in all, this should have exactly the same assembly output and I'd be very surprised if it has any effect on the flake you're referring to.

henryiii:
The problem is that the correct destructor is guaranteed to be called, but when it is called (due to destruction order), the table for the _base_ is loaded, not the child. So `->sync()` is the base's sync, not the child's sync. By skipping the table when calling sync, then it calls the correct one. This is exactly the GooFit bug / CI flake bug we see - the destructor does _not_ force a purge; if the CI is loaded down (or pretty much anytime in GooFit that it is working), the destructor does not cause the final output to show up.

(Sync calls virtual functions too, but those are correct in the base. It's the call from the destructor that's a problem.

If it's a clear solution, I could do the fully qualified solution instead if you prefer.

bstaletic:
> the table for the base is loaded, not the child.

This is the part that I wasn't aware, but it makes sense. Thanks for the explanation.

henryiii:
I wasn't aware of it either, which is why I've been trying to solve this bug (in the context of GooFit, anyway) for over two years...

The fully qualified solution is a little cleaner/clearer, will change to that.

henryiii:
(It is possible I'm wrong here, but it's still slightly more correct this way, and we can watch for that bug in the future)

bstaletic:
A simple check says that it's not as broken as you seem to suggest.

https://godbolt.org/z/4fYojh

Notice that `b` has a static type of `Base*`, but dynamic type of `Derived*`. Two observations:

1. The correct destructor is called, just like you said it would. The destructor of `Derived`.
2. Once we're inside that destructor, a call to `f()` *isn't virtual*. We're already "inside" `Derived` and so `Derived::f` gets called.

Can we come up with a case where "some other `f()`" gets called?

henryiii:
For pybind11, this only fails on msvc (and NVCC for GooFit). I would rather expect the other compilers already handle this properly.

include/pybind11/cast.h
```diff
@@ -432,7 +432,7 @@ PYBIND11_NOINLINE inline std::string error_string() {
 
 #if !defined(PYPY_VERSION)
     if (scope.trace) {
-        PyTracebackObject *trace = (PyTracebackObject *) scope.trace;
+        auto *trace = (PyTracebackObject *) scope.trace;
```
YannickJadoul, line 435: `auto *trace` or `auto trace`? (Not sure what's most common.)

include/pybind11/iostream.h
```diff
@@ -67,8 +72,8 @@ class pythonbuf : public std::streambuf {
     pythonbuf(pythonbuf&&) = default;
 
     /// Sync before destroy
-    ~pythonbuf() {
-        sync();
+    ~pythonbuf() override {
+        _sync();
```
YannickJadoul, line 76: No, you shouldn't call virtual functions anymore in a destructor, indeed. But you should be able to just call `pythonbuf::sync` instead, no? Isn't that making things simpler ánd more clear than this workaround?

include/pybind11/pybind11.h
```diff
@@ -543,7 +543,7 @@ class cpp_function : public function {
                         self_value_and_holder.type->dealloc(self_value_and_holder);
 
                     call.init_self = PyTuple_GET_ITEM(args_in, 0);
-                    call.args.push_back(reinterpret_cast<PyObject *>(&self_value_and_holder));
+                    call.args.emplace_back(reinterpret_cast<PyObject *>(&self_value_and_holder));
```
YannickJadoul, line 546: In this case, it shouldn't matter, though? It's just a pointer...

YannickJadoul (approved):
This is great! :-)
I also really like the incremental approach, for these kinds of changes. As far as I'm concerned, we could merge small batches of these in separate PRs, even. That would keep things more easily reviewable.

The only one I'm a bit skeptical about is `modernize-use-using`. Not sure it that's so problematic. But yes, it's modernizing our code base, so I'm also not opposed.

Personal request: maybe something that flags C-style casts? (`cppcoreguidelines-pro-type-cstyle-cast`?) Then again, in the context of the Python C API, C-style casts might make sense, ofc, since it is C.

include/pybind11/iostream.h
```diff
@@ -68,7 +73,7 @@ class pythonbuf : public std::streambuf {
 
     /// Sync before destroy
     ~pythonbuf() override {
-        sync();
+        _sync();
```
YannickJadoul, line 76: Independently of @bstaletic's experiments: why this way, and not just `pythonbuf::sync();`. That should also make it a non-virtual call, no?
I think the problem [your useful link](https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors) tries to solve with a private method is meant for if the `virtual` function would call through to the base class?

include/pybind11/pybind11.h
henryiii: It either doesn't matter, or is better, so might as well consistently use the one that's better. :)

include/pybind11/cast.h
henryiii: I think the star to remind that it's a pointer? Especially with the star aligned on the right (which some people really hate). Not sure either though.

include/pybind11/pybind11.h
YannickJadoul: Yeah, true, but ... poor `push_back`, being rendered useless :-(
Also, I do expect to see the constructor arguments in `emplace_back`, and somehow that doesn't fit well in my brain with pointers (I copy pointers, I don't copy-construct pointers :-D ).

But yes, this wasn't really a complaint.

include/pybind11/cast.h
YannickJadoul: Probably. It obviously works, ofc, but you're still kind of repeating part of the type (just not the name part), so it's a weird correction to me...

No need to change it, though, as far as I'm concerned. Just curious on the conventions here.

> (which some people really hate)

I'm a (big) fan, btw ;-)

include/pybind11/iostream.h
henryiii: Because `pythonbuf::sync()` didn't fix it. I haven't proven that this fixes it either, though.

YannickJadoul: Huh? That's weird, though, no? If you take away the `virtual` nature of the call, if should be equivalent, or not?

henryiii:
C-style casts are required in a few places, I believe, but it would be good to identify those and replace the others. I tried this briefly (as part of several others - doing many at once leads to conflicts, though). I know the cast I worked on recently was not allowed to be written as a static_cast, since it was converting something like `*T` to int (or something like that, I don't remember).

I didn't add the NULL/0 -> nullptr because it looked odd to have a nullptr in a C api. :)

tests/test_class.cpp
```diff
@@ -400,13 +400,13 @@ template <int N> class BreaksBase { public:
...
-typedef py::class_<BreaksBase<6>, std::shared_ptr<BreaksBase<6>>, BreaksTramp<6>> DoesntBreak6;
-typedef py::class_<BreaksBase<7>, BreaksTramp<7>, std::shared_ptr<BreaksBase<7>>> DoesntBreak7;
-typedef py::class_<BreaksBase<8>, std::shared_ptr<BreaksBase<8>>> DoesntBreak8;
+using DoesntBreak2 = py::class_<BreaksBase<2>, BreaksTramp<2>, std::unique_ptr<BreaksBase<2> > >;
+using DoesntBreak3 = py::class_<BreaksBase<3>, std::unique_ptr<BreaksBase<3> > >;
+using DoesntBreak4 = py::class_<BreaksBase<4>, BreaksTramp<4> >;
+using DoesntBreak5 = py::class_<BreaksBase<5> >;
+using DoesntBreak6 = py::class_<BreaksBase<6>, std::shared_ptr<BreaksBase<6> >, BreaksTramp<6> >;
+using DoesntBreak7 = py::class_<BreaksBase<7>, BreaksTramp<7>, std::shared_ptr<BreaksBase<7> > >;
+using DoesntBreak8 = py::class_<BreaksBase<8>, std::shared_ptr<BreaksBase<8> > >;
```
henryiii, line 409: @YannickJadoul This reads _much_ better now, doesn't it? Having the type name on the left, like everything else in C++, is nicer and more consistent (and in newer C++, there are some further tricks to using). Just ignoring typedef as if it never existed is better than mixing them randomly :)

YannickJadoul: Yes, OK, there's something to be said for that, indeed :-) (It looks drastic to now change it everywhere, but it will be better.)

Now that you point it out, what about these changes of `>>>` -> `> > >`?

include/pybind11/iostream.h
henryiii: I'm not sure this fixes it, but is fills in so many points; the observed bug on MSVC would exactly match what this warning claims happens, and compilers like GCC optimize this by replacing the virtual call - thereby causing the correct behavior. Maybe it's not it, but I think we should keep this as a possible fix; and if we get the IO error again in the future, we can convert it to the fully qualified call (which also silences the warning). I think MSVC _might_ still treat this as an overloaded virtual function, even though you are trying to qualify it.

YannickJadoul:
> C-style casts are required in a few places, I believe, but it would be good to identify those and replace the others. I tried this briefly (as part of several others - doing many at once leads to conflicts, though). I know the cast I worked on recently was not allowed to be written as a static_cast, since it was converting something like `*T` to int (or something like that, I don't remember).

Can be in a different PR, though. I can put some effort into that, if you want; not per se expecting you to do it, ofc ;-)

Also: ugh, never realized there's C casts you can't even do with a `reinterpret_cast`...
 
> I didn't add the NULL/0 -> nullptr because it looked odd to have a nullptr in a C api. :)

Agreed. That's why I'm a bit hesitant indeed.

include/pybind11/iostream.h
YannickJadoul: > I think MSVC _might_ still treat this as an overloaded virtual function, even though you are trying to qualify it.

:roll_eyes:

But OK, I like this proposed approach! If it does fix things, we need to have it; otherwise, we can simplify :-)
You might want to add an extra line of comment underneath `// This function must be non-virtual to be called in a destructor` for future readers that this can't (currently) be simplified to the fully-qualified version, then?

henryiii:
Ahh, I think I didn't try reinterpret_cast.

tests/test_class.cpp
henryiii: I can manually fix those, I assume it is trying to be C++03 compatible. When modernizing C++11 code, hmm...

include/pybind11/cast.h
henryiii: I think there's a fully qualified auto check that adds stars and consts to auto if they apply (IIRC).

henryiii: I understand both * camps. The argument for left star is that it's really part of the type, not the name, and right star is that is (sadly, IMO) how the C syntax decided to do multiple definitions in one line. Which left star camp believes should be avoided, with one def per line always.

include/pybind11/iostream.h
henryiii: Done.

henryiii:
On that page I linked, "has fix" doesn't seem to be accurate. `cppcoreguidelines-pro-type-cstyle-cast` says it has a fix but doesn't, and `google-readability-casting` says it doesn't but does. However, (correctly) it only fixes when it's sure (so I think pretty much .cpp files), so this will take some work to convert.

henryiii:
Assuming you run it (the formula is in the PR), try running it a couple of times, it finds a few more fixes the second time. Remember to clean after you run to get it to recompile. And no multithreaded building if you are changing the source with `-fix`!

henryiii:
Yes, it's here: https://clang.llvm.org/extra/clang-tidy/checks/readability-qualified-auto.html - that recommends (based on the LLVM style guide) that auto be fully qualified. 

For example:

```
/pybind11/tests/test_callbacks.cpp:108:9: warning: 'auto result' can be declared as 'const auto *result' [readability-qualified-auto]
        auto result = f.target<fn_type>();
        ^~~~~
        const auto *
/pybind11/tests/test_callbacks.cpp:108:9: note: FIX-IT applied suggested code changes
```

I'll add that one too, along with a partial attempt for C++ style casts.

bstaletic:
> Also: ugh, never realized there's C casts you can't even do with a reinterpret_cast...

For the record, C casts are pretty wild. They try, in order:

1. `const_cast<NewType>`
2. `static_cast<NewType>`, but with relaxed rules
3. `const_cast(static_cast())`, now in combination *and* with relaxed rules
4. `reinterpret_cast<NewType>`
5. `const_cast(reinterpret_cast())`

For more info see https://en.cppreference.com/w/cpp/language/explicit_cast#Explanation

include/pybind11/cast.h
YannickJadoul: > I think there's a fully qualified auto check that adds stars and consts to auto if they apply (IIRC).

Not convinced we always want that, actually. I don't see why `*` would be treated specially. Why would `int *` become `auto *`, but `std::unique_ptr<int>` become `auto`?

YannickJadoul: > Which left star camp believes should be avoided, with one def per line always.

I actually agree, there (almost always, at least). And yet `T *x` is what I'm used to, and really not harder to read, as far as I can tell, vs.  `T* x`...

include/pybind11/iostream.h
YannickJadoul: Thanks!

tests/test_class.cpp
YannickJadoul: Not sure if it's worth your effort (and the extra annotations), though...

YannickJadoul: Oh, no, it would maybe not correct those back? Wait, let me do the effort, then ;-)

YannickJadoul: OH. I'm already too late; this view was outdated :-| Thanks, @henryiii! :-)

wjakob:
Oh wow, that is a nasty issue you've found in `iostream.h`. Although this code should probably work "as-is" (dispatching sync() to the pybind11-provided class), I can see how this is probably on the fringe of well-defined behavior in a compiler like MSVC. Your workaround with the separate non-virtual function looks good to me.

I have one objection to some of those changes here, and that's the excessive use of "fancy" C++ cast operators. They definitely do have a place in the standard, and that's when working with polymorphic classes, where some casts can lead to undefined behavior, and others will resolve into complex operations that walk through RTTI tables.

But for very basic stuff, like casting something to an `int`, or casting a `void*` pointer to a CPython API object, I find this really unreadable to have lots of `static_casts` over the place. Especially considering the later point, it's worth noting that almost all pointers in pybind11 are CPython API C-style pointers. We have virtually (ha!) no polymorphism within pybind11 itself.

henryiii:
I've dropped that cast operator commit (was always supposed to be completely non controversial), and rebased. Going in now, then!

henryiii:
> I can see how this is probably on the fringe of well-defined behavior in a compiler like MSVC. Your workaround with the separate non-virtual function looks good to me.

That's really what it looks like to me. Though I will unconvince myself of this instantly if this fails CI again at some point after the merge. ;)

rwgk:
I only got a chance to look here now: **massive** thanks @henryiii !

tests/test_factory_constructors.cpp
```diff
@@ -88,8 +88,8 @@ class PyTF6 : public TestFactory6 {
     PyTF6(PyTF6 &&f) : TestFactory6(std::move(f)) { print_move_created(this); }
     PyTF6(const PyTF6 &f) : TestFactory6(f) { print_copy_created(this); }
     PyTF6(std::string s) : TestFactory6((int) s.size()) { alias = true; print_created(this, s); }
-    virtual ~PyTF6() { print_destroyed(this); }
-    int get() override { PYBIND11_OVERRIDE(int, TestFactory6, get, /*no args*/); }
+    ~PyTF6() override { print_destroyed(this); }
+    int get() override { PYBIND11_OVERLOAD(int, TestFactory6, get, /*no args*/); }
```
henryiii, line 92: Oops, incorrect rebase, I got it backwards. Sorry, @YannickJadoul !

tests/test_factory_constructors.cpp
```diff
@@ -109,8 +109,7 @@ class PyTF7 : public TestFactory7 {
     PyTF7(int i) : TestFactory7(i) { alias = true; print_created(this, i); }
     PyTF7(PyTF7 &&f) : TestFactory7(std::move(f)) { print_move_created(this); }
     PyTF7(const PyTF7 &f) : TestFactory7(f) { print_copy_created(this); }
-    virtual ~PyTF7() { print_destroyed(this); }
-    int get() override { PYBIND11_OVERRIDE(int, TestFactory7, get, /*no args*/); }
```
henryiii, line 113: Even worse, sorry again, @YannickJadoul .

henryiii: Why does this not care about being deleted, actually?

tests/test_factory_constructors.cpp
YannickJadoul: Wait, what do you mean?
What went wrong?

tests/test_factory_constructors.cpp
YannickJadoul: Sorry, yes, I hadn't thought about conflicts, when merging that PR. I was just happy it was ready and clicked "merge" :-/

tests/test_factory_constructors.cpp
YannickJadoul: Oh, now I see. It's clearer in the full diff.

Do you want to just push a commit to master? Or do you want me to make a PR?

(Basically, `s/PYBIND11_OVERLOAD/PYBIND11_OVERRIDE/` should work)

henryiii: But not on the definition :). I think I fixed it. It was only a problem where I manually rebased, because I got it backwards.

YannickJadoul: :D Good point. But fixed by #2490!

PR #2480, opened 2020-09-11
Commands used:

rwgk:
hgd pybind11_next
iwyu --nosafe_headers third_party/pybind11:pybind11
files=$(hg status | cut -c3-)
sed -i 's:#include "third_party/pybind11/include/pybind11/:#include "pybind11/:' $files

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -8,9 +8,36 @@
     BSD-style license that can be found in the LICENSE file.
 */
 
-#pragma once
+#include <Python.h>
+#include <stdint.h>
+#include <string.h>
 
-#include "cast.h"
+#include <string>
+#include <type_traits>
+#include <typeinfo>
+#include <utility>
+#include <vector>
+
+#include "pybind11/cast.h"
+#include "pybind11/chrono.h"
+#include "pybind11/complex.h"
+#include "pybind11/detail/class.h"
+#include "pybind11/detail/common.h"
+#include "pybind11/detail/init.h"
+#include "pybind11/detail/internals.h"
+#include "pybind11/detail/typeid.h"
+#include "pybind11/embed.h"
+#include "pybind11/eval.h"
+#include "pybind11/functional.h"
+#include "pybind11/iostream.h"
+#include "pybind11/numpy.h"
+#include "pybind11/operators.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/pytypes.h"
+#include "pybind11/stl.h"
+#include "pybind11/stl_bind.h"
+
+#pragma once
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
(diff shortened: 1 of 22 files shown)


include/pybind11/attr.h
```diff
@@ -8,9 +8,36 @@
...
+#include "pybind11/functional.h"
+#include "pybind11/iostream.h"
+#include "pybind11/numpy.h"
+#include "pybind11/operators.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/pytypes.h"
+#include "pybind11/stl.h"
+#include "pybind11/stl_bind.h"
 
-#include "cast.h"
```
henryiii, line 13: I'd undo this change, I believe.

include/pybind11/attr.h
```diff
@@ -8,9 +8,36 @@
     BSD-style license that can be found in the LICENSE file.
 */
 
-#pragma once
+#include <Python.h>
```
henryiii, line 11: Nice, it's keeping Python.h at the top. :)

include/pybind11/attr.h
```diff
@@ -8,9 +8,36 @@
...
+#include <stdint.h>
+#include <string.h>
```
henryiii, line 13: I rather expect this is not needed.

include/pybind11/operators.h
```diff
@@ -7,9 +7,18 @@
...
+#include <type_traits>
+#include <utility>
 
-#include "pybind11.h"
+#include "pybind11/attr.h"
+#include "pybind11/detail/common.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/stl_bind.h"
+
+#pragma once
```
bstaletic, line 21: What's the point of moving `#pragma once` *after* all the include directives? Seems like a waste of I/O operations.

include/pybind11/pybind11.h
```diff
@@ -8,6 +8,48 @@
...
+#include <new>
+#include <string>
+#include <type_traits>
+#include <typeindex>
+#include <typeinfo>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+#include <version>
```
bstaletic, line 29: This header doesn't exist before C++20.

include/pybind11/attr.h
rwgk: It put in longer paths (`third_party/pybind11/include/pybind11`), I used `sed` to cut out the prefix but decided to keep the `pybind11` (having "explicit is better than implicit" in my head). I could cut out `pybind11` as well. BUT, I think the main question is, is the list of includes it produces useful, for cross-checking and/or occasional manual fixing up?

From this quick experiment I feel the tool is a lot too far away from what we'd need for fully automatic processing.

rwgk:
Closing this PR (as planned; it was just meant to provide an easy means to show the diffs and enable commenting).

include/pybind11/operators.h
henryiii: Doesn't really make a very big difference, but I have no idea why it moved it down.

PR #2481, opened 2020-09-12
[READY] std::chrono::time_point doesn't work for resolutions that don't match system clock's resolution

### LINKED ISSUE #2035: Alternate system clock time point resolutions
## Issue description

When translating from alternate system clock time point types, e.g. `std::chrono::time_point<std::chrono::system_clock, std::chrono::microseconds>` (on a platform where the resolution of `std::chrono::system_clock::time_point` is nanoseconds), the compiler complains in the `type_caster` code:
```
/root/.hunter/_Base/f119fdb/70a713e/0003bc6/Install/include/pybind11/chrono.h:119:19: error: no match for ‘operator=’ (operand types are ‘pybind11::detail::type_caster<std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000> > >, void>::type {aka std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000> > >}’ and ‘std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > >’)
             value = system_clock::from_time_t(std::mktime(&cal)) + microseconds(PyDateTime_DATE_GET_MICROSECOND(src.ptr()));
```
The error comes because C++ will not silently perform the assignment if it could result in loss of precision. Instead, `std::chrono::time_point_cast` must be used. In this case, the result of `system_clock::from_time_t` will be in the default resolution (which is nanoseconds on my platform), so the assignment to `value` (which has microsecond resolution) requires the explicit cast.

## Reproducible example code

[Here](https://wandbox.org/permlink/quC2KTWk0YKrGGxh) is an example of the error I'm seeing, distilled into C++.

[Here](https://wandbox.org/permlink/vW9wcqwWoUGj5cAh) is the fix.

```c++
#include <pybind11/chrono.h>

namespace py = pybind11;

using time_point_us = std::chrono::time_point<
    std::chrono::system_clock, std::chrono::microseconds>;

struct Repro
{
    time_point_us timestamp;
};

PYBIND11_MODULE(repro, m)
{
    py::class_<Repro>(m, "Repro")
        .def(py::init<>())
        .def_readwrite("timestamp", &Repro::timestamp)
        ;
}
```

```py
from datetime import datetime
from repro import Repro

r = Repro()
r.timestamp = datetime.now()
```

bstaletic:
Fixes #2035

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -140,7 +140,7 @@ public:
         }
         else return false;
 
-        value = system_clock::from_time_t(std::mktime(&cal)) + msecs;
+        value = time_point_cast<Duration>(system_clock::from_time_t(std::mktime(&cal)) + msecs);
         return true;
     }
 
--- a/tests/test_chrono.cpp
+++ b/tests/test_chrono.cpp
@@ -12,6 +12,24 @@
 #include <pybind11/chrono.h>
 #include <chrono>
 
+struct different_resolutions {
+    using time_point_h = std::chrono::time_point<
+        std::chrono::system_clock, std::chrono::hours>;
+    using time_point_m = std::chrono::time_point<
+        std::chrono::system_clock, std::chrono::minutes>;
+    using time_point_s = std::chrono::time_point<
+        std::chrono::system_clock, std::chrono::seconds>;
+    using time_point_ms = std::chrono::time_point<
+        std::chrono::system_clock, std::chrono::milliseconds>;
+    using time_point_us = std::chrono::time_point<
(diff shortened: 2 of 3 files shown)


YannickJadoul (approved):
Thanks, @bstaletic! :-) Thanks to @mhfrantz as well, for finding the issue and proposing the fix!

henryiii (approved):

Issue #807, opened 2017-04-19
tools/clang submodule needed?


jagerman:
pybind11 has a `tools/clang` submodule pointing to https://github.com/wjakob/clang-cindex-python3.

Is this actually being used or doing anything?  I've never actually done a submodule init on it, but I notice the ci builds always do.

wjakob:
It's used by the ``tools/mkdoc.py`` script which is used by all of my projects that use pybind11. It would be nice to document it at some point.

This script reads a bunch of C++ header files and invokes Clang via libclang (in parallel) to generate an AST and extract all the comments. 

Example:

```cpp
/*
  This file contains docstrings for the Python bindings.
  Do not edit! These were automatically extracted by mkdoc.py
 */

#define __EXPAND(x)                                      x
#define __COUNT(_1, _2, _3, _4, _5, _6, _7, COUNT, ...)  COUNT
#define __VA_SIZE(...)                                   __EXPAND(__COUNT(__VA_ARGS__, 7, 6, 5, 4, 3, 2, 1))
#define __CAT1(a, b)                                     a ## b
#define __CAT2(a, b)                                     __CAT1(a, b)
#define __DOC1(n1)                                       __doc_##n1
#define __DOC2(n1, n2)                                   __doc_##n1##_##n2
#define __DOC3(n1, n2, n3)                               __doc_##n1##_##n2##_##n3
#define __DOC4(n1, n2, n3, n4)                           __doc_##n1##_##n2##_##n3##_##n4
#define __DOC5(n1, n2, n3, n4, n5)                       __doc_##n1##_##n2##_##n3##_##n4##_##n5
#define __DOC6(n1, n2, n3, n4, n5, n6)                   __doc_##n1##_##n2##_##n3##_##n4##_##n5##_##n6
#define __DOC7(n1, n2, n3, n4, n5, n6, n7)               __doc_##n1##_##n2##_##n3##_##n4##_##n5##_##n6##_##n7
#define DOC(...)                                         __EXPAND(__EXPAND(__CAT2(__DOC, __VA_SIZE(__VA_ARGS__)))(__VA_ARGS__))

#if defined(__GNUG__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-variable"
#endif

static const char *__doc_mitsuba_AtomicFloat =
R"doc(Atomic floating point data type

The class implements an an atomic floating point data type (which is
not possible with the existing overloads provided by ``std::atomic``).
[...]

PR #2482, opened 2020-09-13
Set clang-cindex-python3 url to absolute

varunagrawal:
Fix for #1831.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tools/clang"]
 	path = tools/clang
-	url = ../../wjakob/clang-cindex-python3.git
+	url = git@github.com:wjakob/clang-cindex-python3.git


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tools/clang"]
 	path = tools/clang
-	url = ../../wjakob/clang-cindex-python3.git
+	url = git@github.com:wjakob/clang-cindex-python3.git
```
YannickJadoul, line 3: Shouldn't that be `https://`, for users that don't have SSH set up?

henryiii: This breaks things. It needs to stay relative. I believe this has come up before, and been refused. If you force SSH or HTTPS, then if you are on a system that doesn't have access to the other, it breaks. Relative makes them work.

I think the best fix, however, would to to add the extra tools via CMake instead of via submodule. I need to understand what these are used for before attempting anything, though.

YannickJadoul: > then if you are on a system that doesn't have access to the other

Would you ever have access to ssh but not to https?

henryiii: Depends on how you have your networking setup. I've had that happen in CI at least; Git complained that everything was not ssh. Clusters might have ssh without https, but not sure about that. But in general, if you want to use SHH, that should not also be grabbing things through https, in general.

The traditional submodule solution might be to move the repo into the pybind org, then someone cloning pybind11 could also clone this repo too, even if they were on gitlab, if they add the submodules.

If I knew what this submodule was for, that would help me provide a better recommendation.

YannickJadoul: Right, OK, yes.

It's used in https://github.com/pybind/pybind11/blob/master/tools/mkdoc.py, btw, as far as I know.

varunagrawal:
Well the irony here is hilarious. We have a submodule that's causing all sorts of issues and most people aren't even sure what it does.

bstaletic:
Git allows globally rewriting URLs based on the protocol. Something like this:

    git config --global url."ssh://".insteadOf "https://"

This way, those that have ssh, but not https, can force ssh everywhere.

Would this resolve one sides problems, so we can set the URL to whatever the other side wants?

.gitmodules
varunagrawal: @wjakob explains the use in #807.

henryiii:
I think it has to be:

```bash
git config --global url."ssh://github.com:".insteadOf "https://github.com/"
```

That might work, I've considered that. The problem is that matching the checkout style is the normally the correct thing to do, and relative URLs do that for you. Couldn't you just do:

```bash
git config url."https://github.com/wjacob".insteadOf "../wjakob"
```

If you were on gitlab? That's the rarer / more forced thing already.

Since there is only one submodule, and this is a core library that expected to be a submodule much of the time, I still think there might be a way to solve this without submodules.

.gitmodules
henryiii: Yep, I've literally written this before because I didn't know it existed. Though I wrote it in CMake instead of Python: https://github.com/GooFit/GooFit/blob/master/python/make_docs_header.cmake

henryiii: PS: Tests currently pass because a) no tests do a recursive checkout and b) this is completely untested.

YannickJadoul:
> Yep, I've literally written this before because I didn't know it existed. Though I wrote it in CMake instead of Python: https://github.com/GooFit/GooFit/blob/master/python/make_docs_header.cmake

> I still think there might be a way to solve this without submodules.

It does seem like it might be nice to advertise it as a separate tool, yes. Which could involve pushing it into a separate repo. But that's really something @wjakob needs to help thinking about

wjakob:
I'll close this PR -- the plan is to get rid of `mkdoc` and the libclang bindings in this repo and move them elsewhere.

PR #2484, opened 2020-09-13
Stop py::array_t arguments from accepting arrays that do not match the C- or F-contiguity flags

### LINKED ISSUE #2455: [QUESTION]  Enforce array_t type and ordering
I have a naive question about `array_t`. I have a c++ function that takes a vector and modify it so that I can use it in python thereafter. Reading the documentation, I understood that using `py::array_t<T, py::array::f_style | py::array::forcecast>` allows converting a numpy array from Python to the type `T` and fortran ordering in c++. 

But then, as soon as the numpy array does not match exactly what is required (so type `T` and fortran ordering) in my C++ code (with and without `py::array::forcecast`), the resulting vector in python is filled with zero.

So then, I would like to enforce the type and the ordering, is it possible ? It could throw an error in case the type or the error is not matching T and py::array::f_style.
I have seen some issues about this (#338 or #1305), but I did not find/understand a solution for my problem

YannickJadoul:
`noconvert` checks for `py::array_t` whether the passed type is compatible, but it doesn't check whether the `array::c_style` or `array::f_style` flags are satisifed. See #2455 for a demonstration.

Closes #2455

### DIFF
--- a/tests/test_numpy_array.py
+++ b/tests/test_numpy_array.py
@@ -455,23 +455,30 @@ def test_argument_conversions(forcecast, contiguity, noconvert):
 
     for dtype in [np.dtype('float32'), np.dtype('float64'), np.dtype('complex128')]:
         for order in ['C', 'F']:
-            if not noconvert:
-                # If noconvert is not passed, only complex128 needs to be truncated and
-                # "cannot be safely obtained". So without `forcecast`, it shouldn't be accepted
-                should_raise = dtype.name == 'complex128' and not forcecast
-            else:
-                # If noconvert is passed, only float64 and the matching order is accepted
-                should_raise = (
-                    dtype.name != 'float64' or
-                    (contiguity is not None and contiguity != order)
-                )
+            for shape in [(2, 2), (1, 3, 1, 1), (1, 1, 1), (0,)]:
+                if not noconvert:
+                    # If noconvert is not passed, only complex128 needs to be truncated and
+                    # "cannot be safely obtained". So without `forcecast`, the argument shouldn't
+                    # be accepted.
+                    should_raise = dtype.name == 'complex128' and not forcecast
+                else:
+                    # If noconvert is passed, only float64 and the matching order is accepted.
+                    # If at most one dimension has a size greater than 1, the array is also
+                    # trivially contiguous.
+                    trivially_contiguous = sum(1 for d in shape if d > 1) <= 1
+                    should_raise = (
+                        dtype.name != 'float64' or
+                        (contiguity is not None and
+                         contiguity != order and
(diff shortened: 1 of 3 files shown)


tests/test_numpy_array.py
```diff
@@ -435,6 +435,45 @@ def test_index_using_ellipsis():
     assert a.shape == (6,)
 
 
+@pytest.mark.parametrize("forcecast", [False, True])
+@pytest.mark.parametrize("contiguity", [None, 'C', 'F'])
```
bstaletic, line 439: In terms of integers, this would translate to `0`, `1` and `2`. Where's `c_style | f_style`? That combination still makes no sense to me, but you've shown me that it is possible in gitter.

YannickJadoul: Well, it's basically only an array with all dimensions 1. So I could add a `np.zeros(1)`, if you want? I just thought it would be even more complicated than it already is.

Should I?

bstaletic: > Well, it's basically only an array with all dimensions 1.

So that's how it works - `[[[[0]]]]`.

> So I could add a np.zeros(1), if you want? I just thought it would be even more complicated than it already is.

Why not? It's not like we need to worry about the size of the tests.

YannickJadoul: > So that's how it works - `[[[[0]]]]`.

You can't deny that that's technically contiguous in every way, right? ;-)

> Why not? It's not like we need to worry about the size of the tests.

As in: this test already exploded anyway?

YannickJadoul: Done!

bstaletic (approved):

henryiii (approved):
Looks great to me. I like the new test. 👍

YannickJadoul:
> Looks great to me. I like the new test. :+1:

Thanks! I despise the size/complexity of the test, myself :smile: But at least we know it should be quite safe to merge.

I'll wait a bit longer if anyone else has something to say, and if we get no further complaints, I'll merge :-)

wjakob:
I vaguely remember running into this annoying issue myself. Fix and test look all good to me.

YannickJadoul:
Thanks! :-)

aldanor:
That's nice! Worth releasing 2.6.0 with this feature perhaps? (it's been a while since 2.5.0 anyway)

henryiii:
There are [quite a few things](https://pybind11.readthedocs.io/en/master/changelog.html) in 2.6.0 :) I think we need a bit more time to make sure the latest changes (especially the setup.py rework) are nice and stable and useful before we release. I don't think it's too far away, though.

PR #2487, opened 2020-09-14
Avoid C style casts in the docs, at least for pointers

bstaletic:
When C++14, and thus `py::overload_cast`, isn't available the docs suggest a C style cast. A conservative `static_cast` seems a much better idea when working with function pointers.

EDIT: Let's hope I caught all C casts in the docs.

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -360,7 +360,7 @@ like so:
 .. code-block:: cpp
 
     py::class_<MyClass>("MyClass")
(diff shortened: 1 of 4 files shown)


tdegeus:
Thanks!

YannickJadoul (approved):
Thanks, @bstaletic!

henryiii (approved):

YannickJadoul:
Let's just get this in before we forget. If there are more places where we can find C-style casts, we can always create more PRs :-)

PR #2489, opened 2020-09-15
Qualify some references to `module`

### LINKED ISSUE #2488: [BUG] Compilation error on latest Visual Studio 19 because of module keyword
## Issue description
Related to https://github.com/pybind/pybind11/pull/1639 that wasn't merged.

Compilation fails because of the `module` keyword. In VS, modules are now activated when using `/std:c++latest`. 

```
1>...\include\pybind11/pybind11.h(837,5): error C2059: syntax error: 'module'
1>...\include\pybind11/pybind11.h(837,71): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
1>...\include\pybind11/pybind11.h(837,5): error C7586: a 'module' directive must end with a ';' on the same line
```

This compilation error currently happens only with the preview version of Visual Studio (VS 19 16.8.0 Preview 3.0).

[This blog post](https://devblogs.microsoft.com/cppblog/standard-c20-modules-support-with-msvc-in-visual-studio-2019-version-16-8/) talks about the recent update in Visual Studio.

VS compiler consider module as a keyword [only is it's at the beginning of the line](https://docs.microsoft.com/en-us/cpp/cpp/import-export-module?view=vs-2019#remarks). 
However, in the module class in pybind.h, there is one line with module at the beginning of the line:
```
module def_submodule(const char *name, const char *doc = nullptr)
{ ... }
```
My dirty fix to make it compile is just to add [[nodiscard]] to this function:
```
[[nodiscard]] module def_submodule(const char *name, const char *doc = nullptr)
```
Maybe a safer fix is to rename the class.

## Reproducible example code
Just including pybind11.h fails.
```
int main() { return 0; }
```

bstaletic:
Sometimes, latest MSVC gets confused about whether `module` should be
treated as a keyword or as an identifier name. In those cases, we need
to qualify `module` as `pybind11::module`.

Concerns left to address:

- Do other places need to be qualified?
- How do we test this?

Fixes #2488 

There are 10 places where `module` is the first token on the line. For reference, read #2488.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -46,7 +46,7 @@ With reference counting
 Convenience classes for specific Python types
 =============================================
 
-.. doxygenclass:: module
+.. doxygenclass:: module_
     :members:
 
 .. doxygengroup:: pytypes
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -852,12 +852,12 @@ protected:
 };
 
 /// Wrapper for Python extension modules
-class module : public object {
+class module_ : public object {
 public:
-    PYBIND11_OBJECT_DEFAULT(module, object, PyModule_Check)
+    PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    explicit module(const char *name, const char *doc = nullptr) {
+    explicit module_(const char *name, const char *doc = nullptr) {
         if (!options::show_user_defined_docstrings()) doc = nullptr;
 #if PY_MAJOR_VERSION >= 3
         PyModuleDef *def = new PyModuleDef();
@@ -871,7 +871,7 @@ public:
         m_ptr = Py_InitModule3(name, nullptr, doc);
 #endif
         if (m_ptr == nullptr)
-            pybind11_fail("Internal error in module::module()");
+            pybind11_fail("Internal error in module_::module_()");
         inc_ref();
     }
 
@@ -881,7 +881,7 @@ public:
         details on the ``Extra&& ... extra`` argument, see section :ref:`extras`.
     \endrst */
     template <typename Func, typename... Extra>
-    module &def(const char *name_, Func &&f, const Extra& ... extra) {
+    module_ &def(const char *name_, Func &&f, const Extra& ... extra) {
         cpp_function func(std::forward<Func>(f), name(name_), scope(*this),
                           sibling(getattr(*this, name_, none())), extra...);
         // NB: allow overwriting here because cpp_function sets up a chain with the intention of
@@ -900,10 +900,10 @@ public:
             py::module m2 = m.def_submodule("sub", "A submodule of 'example'");
             py::module m3 = m2.def_submodule("subsub", "A submodule of 'example.sub'");
     \endrst */
-    module def_submodule(const char *name, const char *doc = nullptr) {
+    module_ def_submodule(const char *name, const char *doc = nullptr) {
         std::string full_name = std::string(PyModule_GetName(m_ptr))
             + std::string(".") + std::string(name);
-        auto result = reinterpret_borrow<module>(PyImport_AddModule(full_name.c_str()));
+        auto result = reinterpret_borrow<module_>(PyImport_AddModule(full_name.c_str()));
         if (doc && options::show_user_defined_docstrings())
             result.attr("__doc__") = pybind11::str(doc);
         attr(name) = result;
@@ -911,11 +911,11 @@ public:
     }
 
     /// Import and return a module or throws `error_already_set`.
-    static module import(const char *name) {
+    static module_ import(const char *name) {
         PyObject *obj = PyImport_ImportModule(name);
         if (!obj)
             throw error_already_set();
-        return reinterpret_steal<module>(obj);
+        return reinterpret_steal<module_>(obj);
     }
 
     /// Reload the module or throws `error_already_set`.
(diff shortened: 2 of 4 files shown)


Predelnik:
Waiting for confirmation but most likely this error is given according to the standard. Clang allows such usage inside namespace but it probably is too lenient, gcc with modules patch [also produces an error](https://gcc.godbolt.org/z/nGnvcK). So there's a chance it will be naturally tested sometime in the future by any compiler with C++20 enabled.

bstaletic:
That's definitely a possibility. Maybe we should preemptively qualify all mentions of `module` type.

This might be a good idea for a clang-tidy check.

YannickJadoul:
Alternative solution, already proposed to @bstaletic on Gitter: what about adding a `using module_ = pybind11::module` somewhere (or even rename `module` to `module_` and have the typdef the other way around)? The underscore is also how pybind11 deals with `int`, `class`, etc.

bstaletic:
I'm fine with anything that doesn't break existing users. Just a note though.

> (or even rename `module` to `module_` and have the typdef the other way around)

That breaks ABI of `py::module`, but I'm not concerned about that, in this case.

YannickJadoul:
I'd prefer `module_` (whichever way you choose to be the "real" class vs. the alias) over arbitrarily having to fully qualify the `module` type but nothing else in pybind11, I believe.

bstaletic:
Alright, I'll wait for a while, to hear any other possible opinions before proceeding.

henryiii:
I was going to disagree but I think I agree now. As long as users continue to use `py::module`, which is still valid, I think it's fine.

include/pybind11/pybind11.h
```diff
@@ -900,30 +900,30 @@ class module : public object {
             py::module m2 = m.def_submodule("sub", "A submodule of 'example'");
             py::module m3 = m2.def_submodule("subsub", "A submodule of 'example.sub'");
     \endrst */
-    module def_submodule(const char *name, const char *doc = nullptr) {
+    module_ def_submodule(const char *name, const char *doc = nullptr) {
```
bstaletic, line 903: This is the one that created the error. It is also a member function of `py::module`. This meant that going with `using module_ = module;` doesn't help and I had to rename the whole class.

Unless we were to go with `pybind11::module`, but @YannickJadoul seems to be against that.

bstaletic (commented):
Where and how should this change be documented? Do we want to recommend `module_` from now on?

docs/reference.rst
```diff
@@ -46,7 +46,7 @@ With reference counting
 Convenience classes for specific Python types
 =============================================
 
-.. doxygenclass:: module
+.. doxygenclass:: module_
```
bstaletic, line 49: Did the change to `py::module_` break anything else in the docs?

include/pybind11/pybind11.h
henryiii: I think we should recommend `py::module`; this is just an internal change since `module` unqualified is not allowed in some cases. Users should not be exporting the pybind11 namespace. I don't think users should even know that a `module_` exists normally.

henryiii: > whichever way you choose to be the "real" class vs. the alias

He was fine with either way, just suggesting using `module_` internally instead of writing `pybind11::module` vs. unqualified names everywhere (internally only).

bstaletic: Okay, I added a note in the changelog.

henryiii (approved):

include/pybind11/pybind11.h
YannickJadoul: > but @YannickJadoul seems to be against that.

I'm not against it, for the record. I probably prefer the other one, since it's more consistent with the rest of pybind11 :-)

> I think we should recommend py::module; this is just an internal change since module unqualified is not allowed in some cases.

Also true. So normally, users shouldn't even notice, and if they do have a `using pybind11::module;` or `using namespace pybind11;`, we can't really be blamed?
But explaining to use `module_` seems a lot easier than explaining when to use `py::module`? And it's more consistent with `class_` and `enum_` and `int_` and `float_`?

henryiii: `class`, `enum`, `int`, and `float` have been C++ keywords since the dawn of time (more or less). `module` is _slighly_ special in C++20, and is not a true keyword. And the restrictions on using it are the same for pybind11 as with any other code; including user code. Someone might already write `auto module = ...; module...`, and we can't fix/help that. I don't think we should try to force pretty much every pybind11 project to change the name of `py::module` unless we really had to (and `module` is easier to remember than `module_`). And I think 99% of pybind11 follows the examples and uses `py` as the namespace.

If C++ were to make this a true keyword in 3*n years (I doubt it), then we could force a transition. But I really doubt it, it's too common a term.

YannickJadoul: No, I think I agree: no need to force anything. Just wondering if we should give the "safe" example and just start using `py::module_` in examples, etc; more from a perspective of nudging users than to tell them to transition.

If we don't want to do that, I'm also fine with `pybind11::module` in our code, btw. Thinking about it again, I don't really object to that approach either. Up to @bstaletic who took up the task of fixing this :-)

bstaletic: From what I can tell, no one has a particularly strong opinion. The current patch should work, so why not go with it, as is and, if people start having troubles in their code, we can change the docs to use `module_` more consistently?

Any objections to that idea?

To be honest, I'm just try not to end up endlessly bikeshedding mostly equivalent alternatives.

YannickJadoul (approved):

YannickJadoul:
Not to stall this PR, but this might be worth mentioning to @wjakob; it's a topic with a potential for strong feelings.

wjakob:
This change doesn't bother me at all. ABI compatibility should be fine as well -- the PR doesn't affect the memory layout of data structures that are shared between plugins, so this is all fine.

wjakob (approved):

PR #2490, opened 2020-09-15
fix: use OVERRIDE instead of OVERLOAD

henryiii:
@YannickJadoul, I undid a few changes from #2325 in #2478 due to a bad rebase (I had the change backwards for some reason). This should fix it, I think.

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -196,9 +196,9 @@ Please take a look at the :ref:`macro_notes` before using this feature.
     The :c:macro:`PYBIND11_OVERRIDE` and accompanying macros used to be called
     ``PYBIND11_OVERLOAD`` up until pybind11 v2.5.0, and :func:`get_override`
     used to be called ``get_overload``. This naming was corrected and the older
-    macro and function names have been deprecated, in order to reduce confusion
-    with overloaded functions and methods and ``py::overload_cast`` (see
-    :ref:`classes`).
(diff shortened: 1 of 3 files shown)


YannickJadoul (approved):
Looks good to me! :-)

henryiii:
Okay, one of us can merge when tests pass. ;)

YannickJadoul:
Thanks for the docs fix!

NickAlger:
FYI, the default version installed using conda still uses OVERLOAD. At present, anyone who installs pybind11 via conda will get an error when they try to follow this tutorial (this happened to me, leading to several hours of confusion before I found this thread)

YannickJadoul:
@NickAlger, what's the version you get from conda, then?

YannickJadoul:
@NickAlger, I downloaded a random package from [here](https://anaconda.org/conda-forge/pybind11/files), and it does contain the correct 2.6.0 files. Now I'm very curious where you got your version.

NickAlger:
@YannickJadoul Looks like 2.4.3, installed 3 days ago via 
conda install -c conda-forge pybind11
But looking into it further, it seems like conda decided to install version 2.4.3 for compatibility with some other package in the environment (fenics). Using a new conda environment installs 2.6.0

YannickJadoul:
@NickAlger, that's quite horrible; quite a few things have been improved since 2.4.3. Anyway, thanks for looking into that! :-)

PR #2491, opened 2020-09-16
refactor: drop mkdoc and update changelog

### LINKED ISSUE #1831: tools/clang using relative submodule breaks non-github superprojects

henryiii:
This adds a pointer to the new repo and drops the old submodule. If this is
renamed `pybind11-mkdoc` -> `pybind11_mkdoc`, this can easily be updated before
merging.

Also updates the changelog. PR authors are welcome to update / fill in / add!
Includes #2489, as it should be going in in the near future.

Follows #2482, closes #1831, closes #1872.


bstaletic (approved):
Clean up with ~400 lines removed. What's not to like?

wjakob:
This is probably slightly annoying to do, but would you mind re-basing your changes in `https://github.com/pybind/pybind11-mkdoc` in a way that keeps the history of the main `mkdocs` file in this repository? Others contributed to this over time, and it would be nice to retain the record of that. (Something like what's described here https://stackoverflow.com/questions/7375528/how-to-extract-one-file-with-commit-history-from-a-git-repo-with-index-filter).

wjakob:
Other than that, this all looks great, and thank you for adding the changelog!

wjakob (approved):

henryiii:
Yes, I didn't think about that, but I could do that. Are you interested in moving `pybind11-mkdoc` to `pybind11_mkdoc` for consistency?

henryiii:
Okay, rebased into `pybind11_mkdoc`. Putting this in, then.

PR #2492, opened 2020-09-16
refactor: replace .get_type with type::handle_of

### LINKED ISSUE #2388: refactor: Replace `h.get_type()` with `py::type(h)`

henryiii:
Closes #2388

Added a `handle_of`, which returns a handle instead, so as not to introduce a lot of reference counting where we didn't have it before.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1580,9 +1580,7 @@ template <typename D>
 str_attr_accessor object_api<D>::doc() const { return attr("__doc__"); }
 
 template <typename D>
-handle object_api<D>::get_type() const {
-    return (PyObject *) Py_TYPE(derived().ptr());
-}
+handle object_api<D>::get_type() const {return type::handle_of(*this);}
 
 template <typename D>
 bool object_api<D>::rich_compare(object_api const &other, int value) const {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -152,7 +152,8 @@ public:
 
     /// Return the object's current reference count
     int ref_count() const { return static_cast<int>(Py_REFCNT(derived().ptr())); }
-    /// Return a handle to the Python type object underlying the instance
+
+    PYBIND11_DEPRECATED("Call py::type::handle_of(h) or py::type::of(h) instead of h.get_type()")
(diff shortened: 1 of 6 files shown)


bstaletic (approved):
This looks interesting.

henryiii:
@YannickJadoul 3.9 macOS segfault example. :/

include/pybind11/pytypes.h
```diff
@@ -1568,7 +1580,7 @@ template <typename D>
 str_attr_accessor object_api<D>::doc() const { return attr("__doc__"); }
 
 template <typename D>
-handle object_api<D>::get_type() const { return (PyObject *) Py_TYPE(derived().ptr()); }
+handle object_api<D>::get_type() const {return type::handle_of(*this);}
```
henryiii, line 1583:
```suggestion
handle object_api<D>::get_type() const { return type::handle_of(*this); }
```

YannickJadoul: And add a deprecation, I thought, you were planning?

YannickJadoul (approved):
Yeah, why not :-)

include/pybind11/pytypes.h
henryiii: I pulled back because I thought our policy was to not deprecate as soon as a replacement is available; but since this was never documented, it's probably safe to do so.

PR #2494, opened 2020-09-16
docs: installing section

### LINKED ISSUE #1203: Documentation contains nothing about downloading and installation
The [current documentation](http://pybind11.readthedocs.io/en/latest/) contains absolutely nothing about where to download and how to install the software. It's not that hard to trace things back here to GitHub, but a more explicit pointer would be a good idea! (For those of us on  macOS a comment that [it's available](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pybind11.rb) via [homebrew](https://brew.sh) might also be appropriate.

henryiii:
This creates an installing section.

Closes #1203.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -1,0 +1,102 @@
+.. _installing:
+
+Installing the library
+######################
+
+There are several ways to get the pybind11 source, which lives at
+`pybind/pybind11 on GitHub <https://github.com/pybind/pybind11>`_. The pybind11
+developers recommend one of the first three ways listed here, submodule, PyPI,
+or conda-forge, for obtaining pybind11.
+
+Include as a submodule
+======================
+
+When you are working on a project in Git, you can use the pybind11 repository
+as a submodule. From your git repository, use:
+
+.. code-block:: bash
+
+    git submodule add ../../pybind/pybind11 extern/pybind11 -b v2.6.0
+    git submodule update --init
+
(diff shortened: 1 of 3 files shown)


bstaletic (approved):
Just one thought. Is there a reason to consider `brew` special enough to mention it explicitly? Linux distros all come with some sort of a package manager and at least the rolling release distros keep up with the stable releases.

henryiii:
Brew is the (by far) most popular macOS package manager, and it updates regularly (hours to days after releases). It also works (as Linuxbrew) on all flavors of linux, unlike system package managers. At least one pybind11 maintainer (me) is also a regular contributor to home-brew, and will help keep up this package if it falls behind. It's also mentioned in the original issue three years ago. :)

docs/installing.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Include as a submodule
+======================
+
+When you are working on a project in Git, you can use the pybind11 repository
+as a submodule. From your git repository, use:
+
+.. code-block:: bash
+
+    git submodule add ../../pybind/pybind11 extern/pybind11 -b v2.6.0
```
wjakob, line 19: Cool (the relative link). Can we use `-b stable` if this gets baked into the docs?

wjakob (approved):
LGTM, just a minor comment.

docs/installing.rst
henryiii: Forgot about that, yes, that's better!

PR #2496, opened 2020-09-17
docs: read version from pybind11 file

henryiii:


docs/conf.py
```diff
@@ -59,11 +59,18 @@
...
+    code = compile(f.read(), "../pybind11/_version.py", "exec")
+loc = {}
+exec(code, loc)
+
 # The full version, including alpha/beta/rc tags.
-release = '2.5.dev1'
+release = loc["__version__"]
+
+# The short X.Y version.
+version = ".".join(release.split(".")[:2])
```
bstaletic, line 73: If we can assume to always have two periods in the `release` string, then you can simplify this to `release.rsplit('.', 1)[0]`. Not sure if that's a good idea.

henryiii: I'm not really sure if we should make these different anymore, now that we use PEP 440 naming. I'll check, but I know that this shows up in the top left corner, and having the real version there would likely be better, IMO.

henryiii: (And we can always promise this will be at least length 3, so will always have two periods)

bstaletic: > (And we can always promise this will be at least length 3, so will always have two periods)

Notice that I used `rstrip()` above, so anything other than length 3 wouldn't work.

henryiii: No, I didn't, so yes, it is allowed to be length 4. `2.6.0.dev1` is the current value, actually.

bstaletic: In that case, using `split('.', 2)` would avoid splitting on the 3rd period, but as it's currently written is simpler.

bstaletic (approved):

docs/conf.py
henryiii: Got permission from @wjakob to drop the separation and just use `version`.

PR #2499, opened 2020-09-17
Include what is used in pybind11.h

Ricardicus:
Adding some includes that the file 'include/pybind11/pybind11.h' is depending on, but not including itself.

Add #include <memory> for unique_ptr<> (include/pybind11/pybind11.h:1132)
Add #include <vector> for vector<> (include/pybind11/pybind11.h:1735)
Add #include <string> for string  (include/pybind11/pybind11.h:2139)
Add #include <utility> for move  (include/pybind11/pybind11.h:2197)

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -41,14 +41,15 @@
 #  endif
 #endif
 
-#include "attr.h"
-#include "options.h"
-#include "detail/class.h"
-#include "detail/init.h"
 #include <memory>
 #include <vector>
 #include <string>
 #include <utility>
+
+#include "attr.h"
+#include "options.h"
+#include "detail/class.h"
+#include "detail/init.h"
 
 #if defined(__GNUG__) && !defined(__clang__)
 #  include <cxxabi.h>
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -41,15 +41,15 @@
 #  endif
 #endif
 
-#include <memory>
-#include <vector>
-#include <string>
-#include <utility>
-
 #include "attr.h"
 #include "options.h"
 #include "detail/class.h"
 #include "detail/init.h"
+
+#include <memory>
+#include <vector>
+#include <string>
+#include <utility>
 
 #if defined(__GNUG__) && !defined(__clang__)
 #  include <cxxabi.h>
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -45,6 +45,10 @@
 #include "options.h"
 #include "detail/class.h"
 #include "detail/init.h"
+#include <memory>
+#include <vector>
+#include <string>
+#include <utility>
 
 #if defined(__GNUG__) && !defined(__clang__)
 #  include <cxxabi.h>


include/pybind11/pybind11.h
```diff
@@ -41,6 +41,11 @@
 #  endif
 #endif
 
+#include <memory>
+#include <vector>
+#include <string>
+#include <utility>
+
```
YannickJadoul, line 48: Before or after the other pybind11 headers? I tend to include stdlib as last includes, but I'm not sure if there's a policy or habit, here. Apart from that, fine with me! :-)

YannickJadoul (approved):

include/pybind11/pybind11.h
Ricardicus: I can change the order, if you like? :)

Ricardicus: Here: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
I read: 

_Include headers in the following order: Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers._

But that is Googles rules, not necessarily yours.

henryiii: I think LLVM and therefore the `clang-*` tools have the same ordering recommendation.

henryiii: PS: There is one hard requirement for all Python modules, though: `Python.h` must precede all stdlib headers.

YannickJadoul: Also fine to keep it this way, then. Just wanted to open this for discussion, and check whether this is what we want :-)

henryiii:
Why does this break appveyor? AFAICT, it's working on other branches, but not this, twice now?

include/pybind11/pybind11.h
henryiii: Actually, I think this has to be moved. This puts stdlib includes above `<Python.h>`, which causes massive numbers of warnings on Python 2.7 on older Linux systems (at least). The rule is `Python.h` must be above any stdlib includes for any extension module, and that rule is kept by by pybind11.

Ricardicus:
Interesting. I thought standard C++ library headers were safe to include anywhere. I have no experience of your pipeline, curious to know why it breaks.

include/pybind11/pybind11.h
Ricardicus: Ok! I will do another commit soon, within the hour, moving them under the project headers!

henryiii:
> I thought standard C++ library headers were safe to include anywhere.

It's a requirement listed in the CPython docs:

https://docs.python.org/3.8/c-api/intro.html#include-files

The easiest way I know of to trigger it doing something bad is to use CentOS 7 and the built-in Python 2. You get massive numbers of mismatched redefine warnings if you don't put Python.h first.

Ricardicus:
Lets see if it fails, still.

Ricardicus:
Looks like it went well 👍

henryiii (approved):

YannickJadoul:
Thanks, @Ricardicus! :-)

Ricardicus:
Thank you!

PR #2504, opened 2020-09-17
[WIP] example of different fix for module keyword-like behavior in C++20

henryiii:
Follow up to #2489

If a user has:

```cpp
namespace pybind11 {
    class module;
```

Then the current situation breaks. This restores support for this predecleration for now.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -852,12 +852,12 @@ protected:
 };
 
 /// Wrapper for Python extension modules
-class module_ : public object {
+class module : public object {
 public:
-    PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
+    PYBIND11_OBJECT_DEFAULT(module, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    explicit module_(const char *name, const char *doc = nullptr) {
+    explicit module(const char *name, const char *doc = nullptr) {
         if (!options::show_user_defined_docstrings()) doc = nullptr;
 #if PY_MAJOR_VERSION >= 3
         auto *def = new PyModuleDef();
@@ -871,7 +871,7 @@ public:
         m_ptr = Py_InitModule3(name, nullptr, doc);
 #endif
         if (m_ptr == nullptr)
-            pybind11_fail("Internal error in module_::module_()");
+            pybind11_fail("Internal error in module::module()");
         inc_ref();
     }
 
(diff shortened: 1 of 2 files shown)


bstaletic:
You missed the one in numpy.h.

    include/pybind11/numpy.h:225:1:        module m = module::import("numpy.core.multiarray");

henryiii:
It's valid either way, but I can. I actually intended to only change the constructors, but my using declaration is after the class instead of before. I think we need an idea of the impact before picking which way to go (current vs. this way), unless everyone just loves this way.

henryiii:
I'm pretty happy with the current situation, and just adding a note that predeclarations of py::module are not supported. But I don't have a strong feeling either way. We do need to pick a path (this one, or the current one) before 2.6.0, so will add the milestone (but just closing this without merging moves toward completion, too).

YannickJadoul:
> I'm pretty happy with the current situation, and just adding a note that predeclarations of py::module are not supported. But I don't have a strong feeling either way. We do need to pick a path (this one, or the current one) before 2.6.0, so will add the milestone (but just closing this without merging moves toward completion, too).

:+1:

I'm quite happy with this as well, and as far as I know, @rwgk, who brought up the issue in the first place, also was in the end? I guess the final say here is up to @wjakob?

rwgk:
> I'm quite happy with this as well, and as far as I know, @rwgk, who brought up the issue in the first place, also was in the end? I guess the final say here is up to @wjakob?

I only reported an issue (without having an actual opinion about it). Luckily, the owners of the affected code (openspiel) accepted the most straightforward fix without comment (https://github.com/deepmind/open_spiel/commit/3244ab3f1625f0656e31c607780f57f0d4af18e8).

IMO it's best to drop this PR for now, and hopefully for good.

Related thoughts:

The current changelog.rst entry seems slightly misleading:

* Provide an additional spelling of ``py::module`` - ``py::module_`` (with a    
  trailing underscore), for C++20 compatibility. Only relevant when used        
  unqualified.                                                                  
  `#2489 <https://github.com/pybind/pybind11/pull/2489>`_                       

Would it be more accurate to state that `py::module` was renamed to `py::module_`, but an alias was added with the old name? Maybe with a note that this will break `class module;` forward declarations?

I feel the `py::module_` story will be complete only if `py::module` is deprecated and eventually removed, to completely stay out of the C++20 taken-name territory.

henryiii:
I am okay with changing the recommendation to be `module_`, though I think we should probably keep the `module` typedef around more-or-less forever, unless module becomes a full keyword. While anyone using C++ should know that `class` can't be a user-defined name, the same is not true for `module`.

PR #2510, opened 2020-09-19
fix: AppleClang 12 warnings

### LINKED ISSUE #2509: [BUG] New warning in AppleClang 12
The following warning is produced in AppleClang 12:

```
../include/pybind11/cast.h:2095:26: warning: loop variable 'a' is always a copy because the range of type 'detail::args_proxy' does not return a reference [-Wrange-loop-analysis]
        for (const auto &a : ap)
                         ^
../include/pybind11/cast.h:2095:14: note: use non-reference type 'pybind11::handle'
        for (const auto &a : ap)
             ^~~~~~~~~~~~~~~
../include/pybind11/cast.h:2127:26: warning: loop variable 'k' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
        for (const auto &k : reinterpret_borrow<dict>(kp)) {
                         ^
../include/pybind11/cast.h:2127:14: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
        for (const auto &k : reinterpret_borrow<dict>(kp)) {
             ^~~~~~~~~~~~~~~
In file included from ../tests/pybind11_cross_module_tests.cpp:10:
In file included from ../tests/pybind11_tests.h:2:
../include/pybind11/pybind11.h:1497:34: warning: loop variable 'kv' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
                for (const auto &kv : entries) {
                                 ^
../include/pybind11/pybind11.h:1497:22: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
                for (const auto &kv : entries) {
                     ^~~~~~~~~~~~~~~~
../include/pybind11/pybind11.h:1509:34: warning: loop variable 'kv' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
                for (const auto &kv : entries) {
                                 ^
../include/pybind11/pybind11.h:1509:22: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
                for (const auto &kv : entries) {
                     ^~~~~~~~~~~~~~~~
../include/pybind11/pybind11.h:1524:34: warning: loop variable 'kv' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
                for (const auto &kv : entries) {
                                 ^
../include/pybind11/pybind11.h:1524:22: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
                for (const auto &kv : entries) {
                     ^~~~~~~~~~~~~~~~
../include/pybind11/pybind11.h:1538:34: warning: loop variable 'kv' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
                for (const auto &kv : entries)
                                 ^
../include/pybind11/pybind11.h:1538:22: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
                for (const auto &kv : entries)
                     ^~~~~~~~~~~~~~~~
../include/pybind11/pybind11.h:1626:26: warning: loop variable 'kv' is always a copy because the range of type 'pybind11::dict' does not return a reference [-Wrange-loop-analysis]
        for (const auto &kv : entries)
                         ^
../include/pybind11/pybind11.h:1626:14: note: use non-reference type 'std::__1::pair<pybind11::handle, pybind11::handle>'
        for (const auto &kv : entries)
             ^~~~~~~~~~~~~~~~
7 warnings generated.
```

And two more:

```
[26/43] Building CXX object tests/CMakeFiles/pybind11_tests.dir/test_pytypes.cpp.o
../tests/test_pytypes.cpp:131:26: warning: loop variable 'item' is always a copy because the range of type 'pybind11::detail::str_attr_accessor' (aka 'accessor<accessor_policies::str_attr>') does not return a reference [-Wrange-loop-analysis]
        for (const auto &item : o.attr("begin_end")) {
                         ^
../tests/test_pytypes.cpp:131:14: note: use non-reference type 'pybind11::handle'
        for (const auto &item : o.attr("begin_end")) {
             ^~~~~~~~~~~~~~~~~~
1 warning generated.
[28/43] Building CXX object tests/CMakeFiles/pybind11_tests.dir/test_operator_overloading.cpp.o
../tests/test_operator_overloading.cpp:116:23: warning: explicitly assigning value of variable of type 'const pybind11::detail::self_t' to itself [-Wself-assign-overloaded]
        .def(py::self -= py::self)
             ~~~~~~~~ ^  ~~~~~~~~
../tests/test_operator_overloading.cpp:120:23: warning: explicitly assigning value of variable of type 'const pybind11::detail::self_t' to itself [-Wself-assign-overloaded]
        .def(py::self /= py::self)
             ~~~~~~~~ ^  ~~~~~~~~
2 warnings generated.
```

henryiii:
Fixing two sets of warnings in AppleClang 12. First is a warning that you can't really take a reference to a temporary object, so this will actually always be a copy, and should be clearly written as such.

Second is a bug where 11.0.0 or 12.0.0 or X.X.0 AppleClang will produce a warning due not passing an if statement. It's just a pragma so it should be safe even if the version doesn't match? CC @EricCousineau-TRI  (I think).

Closes #2509

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2092,7 +2092,7 @@ private:
     }
 
     void process(list &args_list, detail::args_proxy ap) {
-        for (const auto &a : ap)
+        for (auto a : ap)
             args_list.append(a);
     }
 
@@ -2124,7 +2124,7 @@ private:
     void process(list &/*args_list*/, detail::kwargs_proxy kp) {
         if (!kp)
             return;
-        for (const auto &k : reinterpret_borrow<dict>(kp)) {
+        for (auto k : reinterpret_borrow<dict>(kp)) {
             if (m_kwargs.contains(k.first)) {
 #if defined(NDEBUG)
                 multiple_values_error();
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1494,7 +1494,7 @@ struct enum_base {
                 handle type = type::handle_of(arg);
                 object type_name = type.attr("__name__");
                 dict entries = type.attr("__entries");
-                for (const auto &kv : entries) {
+                for (auto kv : entries) {
                     object other = kv.second[int_(0)];
                     if (other.equal(arg))
                         return pybind11::str("{}.{}").format(type_name, kv.first);
@@ -1506,7 +1506,7 @@ struct enum_base {
         m_base.attr("name") = property(cpp_function(
             [](handle arg) -> str {
                 dict entries = type::handle_of(arg).attr("__entries");
-                for (const auto &kv : entries) {
+                for (auto kv : entries) {
                     if (handle(kv.second[int_(0)]).equal(arg))
                         return pybind11::str(kv.first);
                 }
@@ -1521,7 +1521,7 @@ struct enum_base {
                 if (((PyTypeObject *) arg.ptr())->tp_doc)
                     docstring += std::string(((PyTypeObject *) arg.ptr())->tp_doc) + "\n\n";
                 docstring += "Members:";
-                for (const auto &kv : entries) {
+                for (auto kv : entries) {
                     auto key = std::string(pybind11::str(kv.first));
                     auto comment = kv.second[int_(1)];
                     docstring += "\n\n  " + key;
@@ -1535,7 +1535,7 @@ struct enum_base {
         m_base.attr("__members__") = static_property(cpp_function(
             [](handle arg) -> dict {
                 dict entries = arg.attr("__entries"), m;
-                for (const auto &kv : entries)
+                for (auto kv : entries)
                     m[kv.first] = kv.second[int_(0)];
                 return m;
             }, name("__members__")), none(), none(), ""
@@ -1623,7 +1623,7 @@ struct enum_base {
 
     PYBIND11_NOINLINE void export_values() {
         dict entries = m_base.attr("__entries");
-        for (const auto &kv : entries)
+        for (auto kv : entries)
             m_parent.attr(kv.first) = kv.second[int_(0)];
     }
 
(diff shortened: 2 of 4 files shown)


include/pybind11/cast.h
```diff
@@ -2092,7 +2092,7 @@ class unpacking_collector {
     }
 
     void process(list &args_list, detail::args_proxy ap) {
-        for (const auto &a : ap)
+        for (auto a : ap)
```
bstaletic, line 2095: In case anyone else wonders "why don't we fix the iterator class", here's why:

- "Iterator" is [this thing](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L836)
- [`operator*`](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L857-L863) returns a [`py::handle`](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L841) which is generally better to copy than to take by reference.
- The iterator stores and mutates `value`, so returning a reference here would be wrong as soon as you advance the iterator.

So this is complete fine.

bstaletic (approved):

YannickJadoul (approved):

include/pybind11/cast.h
EricCousineau-TRI: Meep! I guess I haven't looked what `const_iterator` looks for non-random-access containers in STL :sweat_smile:

From my understanding, your statement of "why don't we fix the iterator class" means "why don't we change `operator*` return `[const] handle&`.

I think there's still question like "what do other libs do for similar patterns?", especially for iterators implemented for something like `std::list`, etc.

EricCousineau-TRI: (No action here, to be clear, just wanted to clarify on the comment)

EricCousineau-TRI (approved):
Thanks for the heads up on the Apple `#ifdef` change!

YannickJadoul:
Thanks for [the extra clarification](https://github.com/pybind/pybind11/pull/2510#pullrequestreview-492041747), @bstaletic & @EricCousineau-TRI!

Given that this is a simple fix, I think it can safely go in.

PR #2512, opened 2020-09-19
tests: Fix test override execution

EricCousineau-TRI:
Hotfix for #2353

Ran into this when trying to add a quick test on upstream.


EricCousineau-TRI:
Er, from looking closer, seems like it should work. Perhaps I did something wrong. Will try a few more times.

EricCousineau-TRI:
K, updated (via force push).

bstaletic (approved):
I guess @henryiii should have the final word, but this makes sense.

henryiii (approved):

henryiii:
Looks good to me! Thanks!

EricCousineau-TRI:
Thanks!

PR #2514, opened 2020-09-19
Add unchecked_reference::operator() and operator[] to overload resolution of unchecked_mutable_reference

### LINKED ISSUE #1468: unchecked_mutable_reference element access methods should be const
## Issue description

The methods provided by `unchecked_mutable_reference` are not marked as `const`:
https://github.com/pybind/pybind11/blob/5ef1af138dc3ef94c05274ae554e70d64cd589bf/include/pybind11/numpy.h#L389
https://github.com/pybind/pybind11/blob/5ef1af138dc3ef94c05274ae554e70d64cd589bf/include/pybind11/numpy.h#L400
https://github.com/pybind/pybind11/blob/5ef1af138dc3ef94c05274ae554e70d64cd589bf/include/pybind11/numpy.h#L403
As `const`-ness is not transitive in C++, these three should all be marked as `const` (while they allow modifying the returned data, they themselved do not mutate the `unchecked_mutable_reference` object).

## Reproducible example code

```
#include <iostream>
#include <pybind11/numpy.h>

void foo(pybind11::detail::unchecked_mutable_reference<float, 1> const& x)
{
    // Assuming x has at least one element for brevity
    std::cout << x[0] << std::endl;
}
```

will fail to compile (even though it should) because `x` been passed as a constant reference (which is a common use case to avoid copies).

## Corrective action

Change the methods quoted above to:

`template <typename... Ix> T& operator()(Ix... index) const {`

`T &operator[](ssize_t index) const { return operator()(index); } `

`template <typename... Ix> T *mutable_data(Ix... ix) const { return &operator()(ssize_t(ix)...); }`

YannickJadoul:
Closes #1468﻿

Given that `unchecked_reference` is a public base, I think not considering these methods in the overload resolution was an oversight.

I think we could indeed also add `const` to these methods of `unchecked_mutable_reference` (similar to `std::span` or `std::reference_wrapper`), but this change is less controversial and should fix the most important use cases.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -419,6 +419,10 @@ class unchecked_mutable_reference : public unchecked_reference<T, Dims> {
     using ConstBase::ConstBase;
     using ConstBase::Dynamic;
 public:
+    // Bring in const-qualified versions from base class
+    using ConstBase::operator();
+    using ConstBase::operator[];
+
     /// Mutable, unchecked access to data at the given indices.
     template <typename... Ix> T& operator()(Ix... index) {
         static_assert(ssize_t{sizeof...(Ix)} == Dims || Dynamic,
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -318,6 +318,18 @@ TEST_SUBMODULE(numpy_array, sm) {
         return auxiliaries(r, r2);
     });
 
+    sm.def("proxy_auxiliaries1_const_ref", [](py::array_t<double> a) {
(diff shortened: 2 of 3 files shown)


bstaletic (approved):

henryiii (approved):

wjakob:
Looks all good to me as well, thanks!

wjakob (approved):

YannickJadoul:
Thanks for logging this issue and for all the arguing on why this was missing, @MoritzMaxeiner!

PR #2520, opened 2020-09-21
Moving tp_class access, and consistent fully-qualified naming for PyPy, to detail::get_tp_name

YannickJadoul:
Increasing consistency, cfr. https://github.com/pybind/pybind11/pull/2349#discussion_r465807968

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -24,7 +24,7 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 #  define PYBIND11_SET_OLDPY_QUALNAME(obj, nameobj) setattr((PyObject *) obj, "__qualname__", nameobj)
 #endif
 
-inline std::string get_tp_name(PyTypeObject *type) {
+inline std::string get_fully_qualified_tp_name(PyTypeObject *type) {
 #if !defined(PYPY_VERSION)
     return type->tp_name;
 #else
@@ -180,7 +180,7 @@ extern "C" inline PyObject *pybind11_meta_call(PyObject *type, PyObject *args, P
     for (const auto &vh : values_and_holders(instance)) {
         if (!vh.holder_constructed()) {
             PyErr_Format(PyExc_TypeError, "%.200s.__init__() must be called when overriding __init__",
-                         get_tp_name(vh.type->type).c_str());
+                         get_fully_qualified_tp_name(vh.type->type).c_str());
             Py_DECREF(self);
             return nullptr;
         }
@@ -312,7 +312,7 @@ extern "C" inline PyObject *pybind11_object_new(PyTypeObject *type, PyObject *, 
 /// following default function will be used which simply throws an exception.
 extern "C" inline int pybind11_object_init(PyObject *self, PyObject *, PyObject *) {
     PyTypeObject *type = Py_TYPE(self);
-    std::string msg = get_tp_name(type) + ": No constructor defined!";
+    std::string msg = get_fully_qualified_tp_name(type) + ": No constructor defined!";
     PyErr_SetString(PyExc_TypeError, msg.c_str());
     return -1;
 }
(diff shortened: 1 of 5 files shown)


tests/test_local_bindings.py
```diff
@@ -155,7 +155,7 @@ def test_internal_locals_differ():
     assert m.local_cpp_types_addr() != cm.local_cpp_types_addr()
 
 
-@pytest.mark.xfail("env.PYPY")
+@pytest.mark.xfail("env.PYPY and sys.pypy_version_info < (7, 3, 2)")
```
bstaletic, line 158: "Expect this to fail on PyPy < 7.3.2"

Didn't we decide to drop older versions already?

YannickJadoul: 7.3.2 is not out, yes. And it still fails on 7.3.1 :-)

YannickJadoul: If 7.3.2 is out before this gets merged, yes, let's remove it

YannickJadoul: (unrelated change, btw, but something I bumped into, when testing locally)

henryiii: I probably wouldn't drop it right away; approximately supporting 7.3.0+ would probably be ideal.

YannickJadoul: Yes, that's probably best.

include/pybind11/detail/class.h
```diff
@@ -24,6 +24,14 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 #  define PYBIND11_SET_OLDPY_QUALNAME(obj, nameobj) setattr((PyObject *) obj, "__qualname__", nameobj)
 #endif
 
+inline std::string get_tp_name(PyTypeObject *type) {
+#if !defined(PYPY_VERSION)
```
henryiii, line 28: What happens in 7.3.2? Does this line start working?

YannickJadoul: Ugh .. I'd written this yesterday, but it was still a "pending review" ...

include/pybind11/detail/class.h
```diff
@@ -24,6 +24,14 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 #  define PYBIND11_SET_OLDPY_QUALNAME(obj, nameobj) setattr((PyObject *) obj, "__qualname__", nameobj)
 #endif
 
+inline std::string get_tp_name(PyTypeObject *type) {
```
rwgk, line 27: `get_fq_type_name` or even `get_fully_qualified_type_name`?
Somehow conveying that this function is meant to produce the fully qualified name seems important to me.

rwgk (commented):
This looks like a great step in the right direction.
Are there existing unit tests for built-in types and nested classes? (If not it would be OK with me to add them later.)

include/pybind11/detail/class.h
YannickJadoul: Yes, good point!

YannickJadoul:
> Are there existing unit tests for built-in types and nested classes? (If not it would be OK with me to add them later.)

I think so, yes. Do the tests that I had to change (because PyPy is now consistent with CPython) count?

include/pybind11/detail/class.h
YannickJadoul: Done! Thanks for the suggestion.

include/pybind11/detail/class.h
```diff
@@ -476,9 +479,8 @@ extern "C" inline int pybind11_clear(PyObject *self) {
 inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type) {
     auto type = &heap_type->ht_type;
 #if defined(PYPY_VERSION) && (PYPY_VERSION_NUM < 0x06000000)
-    pybind11_fail(std::string(type->tp_name) + ": dynamic attributes are "
-                                               "currently not supported in "
-                                               "conjunction with PyPy!");
+    pybind11_fail(get_tp_type(type) + ": dynamic attributes are currently not "
```
rwgk, line 482: This still shows `get_tp_type`. Oversight?

rwgk (approved):
Thanks Yannick! I think this is good to merge, to help you finishing up #2349. I didn't have a chance to look into existing unit test coverage beyond the changes in `test_class.py` here, but we can look into that separately, later, or as the need arises.

include/pybind11/detail/class.h
YannickJadoul: Oh. That was even wrong :-| It used to be `get_tp_name`... This is part of the code is going away in #2456, though. That's why it's not tested.

YannickJadoul:
OK, sounds good to me. Thanks, @rwgk!

henryiii (approved):

henryiii:
This seems safe to put in; would like to slowly keep a little momentum heading into 2.6.0.

PR #2521, opened 2020-09-22
feat: setup_helpers parallel compile

henryiii:
One of the most common needs for compiling a setuptools extensions is a parallel compile step. For tools like Cython that require NumPy to be present, NumPy provided a parallel compiler function, `distutils.ccompiler.CCompiler.compile = numpy.distutils.ccompiler.CCompiler_compile`; just adding that makes your build compile in parallel, and is controllable via the `NPY_NUM_BUILD_JOBS ` environment variable. However, pybind11 goes through a great deal of trouble to avoid forcing NumPy to be present when compiling, so this is not available to pybind11 extensions. This provides a lightweight replacement for it, also inspired by cppimport and some stack overflow discussions.

For boost-histogram, a parallel compile is about 4-7 times faster than sequential on my MacBook. About 2x faster on GHA.

---

Make a parallel compile function. Inspired by `numpy.distutils.ccompiler.CCompiler_compile` and `cppimport`. This takes several arguments that allow you to customize the compile function created:

* `envvar`: Set an environment variable to control the compilation threads, like `NPY_NUM_BUILD_JOBS` or something customized to your extension.
* `default`: 0 will automatically multithread, or 1 will only multithread if the envvar is set.
* `max`: The limit for automatic multithreading if non-zero.

To use:

```python
ParallelCompile("NPY_NUM_BUILD_JOBS").install()
```

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -68,6 +68,23 @@ that is supported via a ``build_ext`` command override; it will only affect
         ext_modules=ext_modules
     )
 
+Since pybind11 does not require NumPy when building, a light-weight replacement
+for NumPy's parallel compilation distutils tool is included. Use it like this:
+
+    from pybind11.setup_helpers import ParallelCompile
+
+    # Optional multithreaded build
+    ParallelCompile("NPY_NUM_BUILD_JOBS").install()
+
+    setup(...
+
+The argument is the name of an environment variable to control the number of
+threads, such as ``NPY_NUM_BUILD_JOBS`` (as used by NumPy), though you can set
+something different if you want. You can also pass ``default=N`` to set the
+default number of threads (0 will take the number of threads available) and
+``max=N``, the maximum number of threads; if you have a large extension you may
+want set this to a memory dependent number.
+
 .. _setup_helpers-pep518:
 
 PEP 518 requirements (Pip 10+ required)
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -49,6 +49,7 @@ except ImportError:
     from distutils.extension import Extension as _Extension
 
 import distutils.errors
+import distutils.ccompiler
 
 
 WIN = sys.platform.startswith("win32")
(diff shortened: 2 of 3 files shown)


docs/compiling.rst
```diff
@@ -68,6 +68,23 @@ that is supported via a ``build_ext`` command override; it will only affect
         ext_modules=ext_modules
     )
 
+Since pybind11 does not require NumPy when building, a light-weight replacement
+for NumPy's parallel compilation distutils tool is included. Use it like this:
+
+    from pybind11.setup_helpers import ParallelCompile
+
+    # Optional multithreaded build
+    ParallelCompile("NPY_NUM_BUILD_JOBS").install()
```
YannickJadoul, line 77: Isn't it then still a bit weird to refer to a `NPY` environment variable?

YannickJadoul (commented):
Wow, cool! That's going to be useful!

Before going into the actual details of the implementation:

> However, pybind11 goes through a great deal of trouble to avoid forcing NumPy to be present when compiling, so this is not available to pybind11 extensions.

Just a thought here: pybind11 doesn't make NumPy mandatory, but some functionality is only available with NumPy. Is it worth maintaining this ourselves, vs. saying "if you want parallel builds, NumPy provides this functionality"?
I'm fine adding it, but you míght be the only one with intricate knowledge, so also the first against whom we'll start complaining if things break, @henryiii ;-)

docs/compiling.rst
henryiii: It's best to show a variable here (it is optional, but if you don't supply it, there's no way to control the number of threads manually, which might be nice to have). Since NumPy uses this, I just chose it as a standard (and it makes it a true drop in replacement).

henryiii:
> pybind11 doesn't make NumPy mandatory, but some functionality is only available with NumPy

No, pybind11 _never_ requires NumPy to compile; for example, Cython does require NumPy _to compile_ code that interacts with NumPy. At runtime, it's no issue - just add NumPy to your install_requires and you are good to go. But making it available at compile time is harder, and that's where you need it to use this part.

docs/compiling.rst
henryiii: (I'd be open to alternative suggestions for this name, though; it's just an example)

YannickJadoul: > Since NumPy uses this, I just chose it as a standard (and it makes it a true drop in replacement).

I see where you're coming from, yes. It's just a bit confusing at first sight.

YannickJadoul:
> No, pybind11 _never_ requires NumPy to compile; [...]. But making it available at compile time is harder, and that's where you need it to use this part.

Right, you were talking about compile-time dependencies (I thought just in general). That's true, though, yes. I don't know the difficulties of providing it at compile time, but that would be a good argument, then, yes.

I'm wondering if this wouldn't be worth a small project on its own, then, for the benefit of all Python users (and not just pybind11 users)! But I'm fine adding this to pybind11 :-)

henryiii:
Even though, with pyproject.toml, requiring NumPy is not _that_ bad and something that Cython users have to get used to, pybind11 skips several problems:

* You must specify the earliest version of NumPy supported to compile against; you can't run a wheel compiled with a newer NumPy, only older. But it has to be supported by the Python you are working with, forcing pyproject.toml's that look like this:
```
requires = [ ...
    "numpy==1.13.3; python_version<='3.6'",
    "numpy==1.14.5; python_version=='3.7'",
    "numpy==1.17.3; python_version>='3.8'",
```

(It's messier for each new Python version, and if you support < 1.13, then you need even more lines)

* Wheels sometimes are missing, especially for CPython-dev/PyPy/Arch64, etc. Then means you have to compile NumPy, and you have NumPy wheels in your wheelhouse. This adds a significant amount of time to the step, as well as sometimes not working.

boost-histogram used to require NumPy just for this one feature, but was hitting the second problem (not the first, due to pybind11's design, but I have to work with that for Cython.)

YannickJadoul (approved):
> pybind11 skips several problems

OK, that does convince me! Thank for spelling that out.

Haven't looked in maximum detail, but seems good to me :-)

(From what you've just told, I still believe this might be a nice little project to publish independently from pybind11, but that might be too much work for you to add to your list of things to maintain :-) )

henryiii:
> publish independently from pybind11

Maybe PEP 518 will change things, but in general, helper packages for setuptools/distutils are not common. That's why this helper file is standalone and supports direct copying into a project; for a no-way-to-fail project, that is the only truly foolproof way to go. And that's why people would chose setuptools builds in the first place over CMake. So I don't think a stand-alone version would be very helpful, it it was, it would already exist.

FYI, boost-histogram 0.11 (coming tomorrow) has this update in it, fully dropping NumPy as a compile requirement.

wjakob (approved):
This looks great, thank you!

PR #2522, opened 2020-09-22
fix: warning on latest AppleClang

henryiii:
Fixed in #2510 but reintroduced on one line by #2126


YannickJadoul:
Huh. I thought I'd fixed this, when rebasing, but apparently not :-/ Thanks for catching this, @henryiii!

PR #2526, opened 2020-09-24
refactor: import check as a common function

henryiii:
Pulling the import check out of the tests/CMakeLists into a reusable, generic function that works on both new/old Python CMake's. Nicer controls and output display.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -86,8 +86,7 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
     hidden)``.
 
   * Helper functions ``pybind11_strip``, ``pybind11_extension``,
-    ``pybind11_find_import`` added (see ``tools/pybind11Config.cmake.in`` for
-    docs).
+    ``pybind11_find_import`` added, see :doc:`cmake/index`.
 
 * Optional :ref:`find-python-mode` and :ref:`nopython-mode` with CMake.
   `#2370 <https://github.com/pybind/pybind11/pull/2370>`_
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -85,6 +85,10 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
     ``CMAKE_INTERPROCEDURAL_OPTIMIZATION``, ``set(CMAKE_CXX_VISIBILITY_PRESET
     hidden)``.
 
+  * Helper functions ``pybind11_strip``, ``pybind11_extension``,
+    ``pybind11_find_import`` added (see ``tools/pybind11Config.cmake.in`` for
+    docs).
+
 * Optional :ref:`find-python-mode` and :ref:`nopython-mode` with CMake.
   `#2370 <https://github.com/pybind/pybind11/pull/2370>`_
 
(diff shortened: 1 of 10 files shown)


tests/CMakeLists.txt
```diff
@@ -347,24 +347,7 @@ foreach(target ${test_targets})
...
-    COMMAND ${PYTHON_EXECUTABLE} -c "import pytest; print(pytest.__version__)"
-    RESULT_VARIABLE pytest_not_found
-    OUTPUT_VARIABLE pytest_version
-    ERROR_QUIET)
-  if(pytest_not_found)
-    message(WARNING "Running the tests requires pytest. Please install it manually"
-                    " (try: ${PYTHON_EXECUTABLE} -m pip install pytest)")
-  elseif(pytest_version VERSION_LESS 3.1)
-    message(WARNING "Running the tests requires pytest >= 3.1. Found: ${pytest_version}"
-                    "Please update it (try: ${PYTHON_EXECUTABLE} -m pip install -U pytest)")
```
YannickJadoul, line 361: So we will get more generic, less helpful messages here?

(I think I'm fine with that; just want to point this out.)

YannickJadoul, line 361: So we will get a more generic message here?

(I think I'm fine with that; just want to point this out.)

YannickJadoul (approved):
Nice and useful! :-)

Is there an easy/effortless way of adding the great documentation in `pybind11Config.cmake.in ` to the RTD docs?

tests/CMakeLists.txt
henryiii: Yes, a little bit. Though following the old message can break your system (see https://github.com/scikit-hep/particle/issues/269 for an example of a mismatch this can create), so I don't consider this a downside. :)

henryiii:
> Is there an easy/effortless way of adding the great documentation in pybind11Config.cmake.in to the RTD docs?

That's not something I've ever tried to do...

docs/changelog.rst
```diff
@@ -85,6 +85,10 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
     ``CMAKE_INTERPROCEDURAL_OPTIMIZATION``, ``set(CMAKE_CXX_VISIBILITY_PRESET
     hidden)``.
 
+  * Helper functions ``pybind11_strip``, ``pybind11_extension``,
+    ``pybind11_find_import`` added (see ``tools/pybind11Config.cmake.in`` for
+    docs).
```
wjakob, line 90: Looks fantastic, but as the CMake integration gets fancier, can we have coverage in RTD? You could just copy the text from `tools/pybind11Config.cmake.in` and expand it very slightly. As is, it  is not going to be very easy to find for users.

wjakob (changes requested):

docs/changelog.rst
henryiii: Docs pull the page in from the CMake files now, and reformatted this to look more like FindPython's docs. CMake's docs are built by Sphinx, by the way.

henryiii:
Going to merge so we can see if RtD works correctly. Works locally. We can touch up afterward if it's not perfect. :)

YannickJadoul:
The extra integration with RTD is amazing! Thanks, @henryiii :-)

PR #2530, opened 2020-09-25
ci: releases

henryiii:
Update: Includes a fix were `python -m build` might use a setuptools version that is buggy under pep517.

This updates the CI and release procedure to do the releases to PyPI automatically. The instructions look longer, but mostly because they have a bit more detail now.

@wjakob Here is my recommended procedure for setting up PyPI tokens:

1. Create a PyPI token: go to account settings, scroll down to "Create an API token", give it a name (like pybind11-GHA), and give it permissions to upload packages to "entire account".
2. You will get a token; you must copy it now (it will not be accessible again). You need to recreate it if you need to get one again.
3. Open pybind/pybind11's GitHub page, go to settings -> secrets -> new secret, and use `pypi_password` as the name, paste the token in as the value. You will also not be able to access this after setting it.

Done! `pybind/pybind11` will now have access to upload `pybind11` and create/upload `pybind11-global`.

Once we upload even a beta/rc of `pybind11-global`, go back and make two scoped tokens, one for `pybind11`, one for `pybind11-global`, and I'll change this this upload in two steps, one token each. They need different names (just tell me what you chose, and I'll match in the steps).

PS: GitHub now has organizational secrets too. I don't think that's useful here - the general rule is one token per usage. If Travis and GHA upload wheels to the same package, you can make a token for each, there's no need to share.


docs/release.rst
```diff
@@ -1,21 +1,75 @@
...
+  - ``git add`` and ``git commit``, ``git push``. **Ensure CI passes**. (If it
+    fails due to a known flake issue, either ignore or restart CI.)
+- Add a release branch if this is a new minor version
+  - ``git checkout -b vX.Y``, ``git push -u origin vX.Y``
+- Update tags
+  - ``git tag -a vX.Y.Z -m 'vX.Y.Z release'``.
+  - ``git push --tags``.
+- Update stable
+    - ``git checkout stable``
+    - ``git merge --ff-only master``
```
wjakob, line 39: The (`--ff-only`) seems unnecessarily strict. We may need to cherry-pick specific things onto stable, so there is no expectation that it matches the linearized commit sequence of `master`.

wjakob (commented):
This looks great, just one comment regarding the text in `release.rst`. FWIW generating the wheels/sdists locally fails on my machine, with the following message:
```
Writing pybind11-2.6.0.dev1/setup.cfg
Creating tar archive
removing 'pybind11-2.6.0.dev1' (and everything under it)
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py", line 280, in <module>
    main()
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py", line 263, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py", line 236, in build_sdist
    return backend.build_sdist(sdist_directory, config_settings)
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/setuptools/build_meta.py", line 188, in build_sdist
    return _file_with_extension(sdist_directory, '.tar.gz')
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/setuptools/build_meta.py", line 94, in _file_with_extension
    file, = matching
ValueError: too many values to unpack (expected 1)
  File "/Users/wjakob/.anaconda/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/bin/python3', '/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py', 'build_sdist', '/var/folders/93/bgc8wd2s46g1z2wnlypsg6kw0000gp/T/tmpl_h50i92']' returned non-zero exit status 1.
During handling of the above exception, another exception occurred:
  File "/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/build/__init__.py", line 215, in build
    raise BuildBackendException('Backend operation failed: {}'.format(e))
build.BuildBackendException: Backend operation failed: Command '['/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/bin/python3', '/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py', 'build_sdist', '/var/folders/93/bgc8wd2s46g1z2wnlypsg6kw0000gp/T/tmpl_h50i92']' returned non-zero exit status 1.
ERROR Backend operation failed: Command '['/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/bin/python3', '/Users/wjakob/RGL Dropbox/Wenzel Jakob/projects/pybind11/venv/lib/python3.7/site-packages/pep517/_in_process.py', 'build_sdist', '/var/folders/93/bgc8wd2s46g1z2wnlypsg6kw0000gp/T/tmpl_h50i92']' returned non-zero exit status 1.
```
I'll set up the various tokens as you suggested.

wjakob:
> (Optional) Create a new PyPA account called "pybind" (or similar). 
You mean PyPI, right?

wjakob:
I've created the token and set it up as a GitHub secret. Note: I didn't create a separate account because it was possible to scope the token to the pybind11 project. The secret on GitHub includes the "pypi-" prefix -- is that compatible with your tooling? (Some tools IIRC want you to remove that.)

henryiii:
We can't scope it to _two_ projects, `pybind11` and `pybind11-global` (to the best of my knowledge). Allowing it to upload to all your projects is probably not _that_ bad; it's still safer than having a username and password, which also has that ability, and the ability to delete/move/change things too. Before tokens, that's how it always worked - I'm sure I have several packages still uploading that way. Up to you on how much security you want.

I'll look into why you are getting an error in a little later; python-build is quite new so bugs may still be pretty easy to hit. If you have a recent CMake (3.15+), you could try the "classic" method, `python setup.py sdist` and `pip wheel .`? Pip or conda install (from conda-forge, probably) cmake if needed.

.github/workflows/pip.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    - uses: actions/download_artifacts@v2
+      with:
+        path: dist
+        name: artifact
+
+    - uses: pypa/gh-action-pypi-publish@v1.3.1
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
```
henryiii, line 89: @wjakob the name here needs to match the name in the secrets. It can be anything, though.

henryiii:
Reported to https://github.com/FFY00/python-build/issues/108

henryiii:
Found the problem with @FFY00's kind help. I forgot to force setuptools >=42, which I usually do (because it's required for `setuptools_scm`, which I didn't use here). But setuptools 40 has two bad bugs in it. If you have an old download of setuptools for Python 3.7, `build` uses that instead of getting a new one (because it can), and that is broken. For some reason, pip/pep517.build go ahead and get a newer setuptools instead of reusing the cache. I'll revert the WIP testing commits and require a reasonable setuptools version.

PR #2538, opened 2020-09-29
CMake: check for distutils presence when querying python extension module suffix

wjakob:
On very incomplete python installations (e.g. within Docker), it's
possible that distutils is not installed. In that case, the
``execute_command`` statement that queries distutils for the Python
module extension fails, and pybind11 uses the empty string. This commit
adds an extra check that causes a CMake failure with more actionable
information (just spent a lot of time trying to track down this problem :))

### DIFF
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -91,7 +91,13 @@ execute_process(
   COMMAND "${${_Python}_EXECUTABLE}" "-c"
           "from distutils import sysconfig; print(sysconfig.get_config_var('SO'))"
   OUTPUT_VARIABLE _PYTHON_MODULE_EXTENSION
-  ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
+  ERROR_VARIABLE _PYTHON_MODULE_EXTENSION_ERR
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+if (_PYTHON_MODULE_EXTENSION STREQUAL "")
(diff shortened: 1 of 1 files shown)


henryiii:
Sort-of related, on Windows, can we just set pyd? It's done that way here: https://gitlab.kitware.com/cmake/cmake/-/blob/39677de5e209445c8cbc5957c1e79088d5d2a03a/Modules/FindPython/Support.cmake#L3136-3138 

I think we should print the error string if it fails. I thought I had a PR that added that, actually, but maybe not.

wjakob:
Are you saying that we don't set the .pyd extension on Windows at the moment? Not sure that makes sense to me because extensions appear correctly named when building on Windows.

henryiii:
No, I meant I don't think we need to query for the extension on Windows; it is always `.pyd`. So even if distutils was missing/incomplete, it would still work on Windows. Since CMake's FindPython seems to assume .pyd on Windows, I think it might be safe.

(or that could just be a fallback on Windows if this fails).

wjakob:
But `_PYTHON_MODULE_EXTENSION_ERR` contains more than just the filename extension, no? (e.g. `.cpython-37m-darwin.so`)

henryiii (approved):
I don't think there's any reason to hold this up on not skipping Windows; and Windows might be so likely to always work here that it's not a worthwhile optimization to make.

henryiii:
Yes, in the CMake code, that's added as part of the "WITH_SOABI" option below. Okay, this is ideal, then. :)

PR #2544, opened 2020-10-02
refactor: module -> module_ with typedef

henryiii:
This moves all docs (except changelog and upgrade guide) and code to using `module_` instead of `module`, due to C++20 semi-keyword status, which breaks if `module` is the first word on a line. Also changes a couple of test variables named `module`, just in case they would break under some compilers in C++20 mode.

First commit sets up without `using module = module_` to make sure everything compiles. Second commit restores `py::module`, which can be kept indefinitely unless C++23 or later makes `module` a full keyword. Alternate to, and closes #2504 .

This was not my first idea, but my mind was changed by @rwgk's comment here: https://github.com/pybind/pybind11/pull/2504#issuecomment-699592229 . Basically, this is more transparent to users and easier to explain, and more future-proof.

I've highlighted key changes below, since otherwise it's mostly just a lot of `module`->`module_`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,9 +11,9 @@ v2.6.0 (IN PROGRESS)
 
 See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
 
-* Provide an additional spelling of ``py::module`` - ``py::module_`` (with a
-  trailing underscore), for C++20 compatibility. Only relevant when used
-  unqualified.
+* ``py::module`` was renamed ``py::module_`` to avoid issues with C++20 when
+  used unqualified, but an alias ``py::module`` is provided for backward
+  compatibility.
   `#2489 <https://github.com/pybind/pybind11/pull/2489>`_
 
 * ``pybind11_add_module()`` now accepts an optional ``OPT_SIZE`` flag that
@@ -524,7 +524,7 @@ v2.2.2 (February 7, 2018)
 v2.2.1 (September 14, 2017)
 -----------------------------------------------------
 
-* Added ``py::module::reload()`` member function for reloading a module.
+* Added ``py::module_::reload()`` member function for reloading a module.
   `#1040 <https://github.com/pybind/pybind11/pull/1040>`_.
 
 * Fixed a reference leak in the number converter.
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -904,9 +904,9 @@ public:
 
         .. code-block:: cpp
 
-            py::module m("example", "pybind11 example plugin");
-            py::module m2 = m.def_submodule("sub", "A submodule of 'example'");
-            py::module m3 = m2.def_submodule("subsub", "A submodule of 'example.sub'");
+            py::module_ m("example", "pybind11 example plugin");
+            py::module_ m2 = m.def_submodule("sub", "A submodule of 'example'");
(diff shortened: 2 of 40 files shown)


tests/test_embed/test_interpreter.cpp
```diff
@@ -278,7 +278,7 @@ TEST_CASE("Reload module from file") {
     test_module.close();
 
     // Reload the module
-    module.reload();
-    result = module.attr("test")().cast<int>();
+    module_.reload();
+    result = module_.attr("test")().cast<int>();
```
henryiii, line 282: This is an example of a test variable renamed, just in case. It is the first item on a logical line, so by definition it might break on some compilers in C++20 mode.

docs/changelog.rst
```diff
@@ -11,9 +11,9 @@ v2.6.0 (IN PROGRESS)
 
 See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
 
-* Provide an additional spelling of ``py::module`` - ``py::module_`` (with a
-  trailing underscore), for C++20 compatibility. Only relevant when used
-  unqualified.
+* ``py::module`` was renamed ``py::module_`` to avoid issues with C++20 when
+  used unqualified, but an alias ``py::module`` is provided for backward
+  compatibility.
```
henryiii, line 16: Reworded changlog entry.

tests/test_embed/test_interpreter.cpp
YannickJadoul: Interesting that we didn't get any VC++ failures, yet, then? Or are we not yet testing C++20?

include/pybind11/pybind11.h
```diff
@@ -965,12 +965,15 @@ class module_ : public object {
...
-            pybind11_fail("Internal error in module::module()");
+            pybind11_fail("Internal error in module_::module_()");
         inc_ref();
     }
 #endif
 };
 
+// When inside a namespace (or anywhere as long as it's not the first item on a line),
+// C++20 allows "module" to be used. This is provided for backward compatibility, and for
+// simplicity, if someone wants to use py::module for example, that is perfectly safe.
```
henryiii, line 976: Added comment here.

tests/test_modules.cpp
```diff
@@ -13,6 +13,7 @@
 
 TEST_SUBMODULE(modules, m) {
     // test_nested_modules
+    // This is intentionally "py::module" to verify it still can be used in place of "py::module_"
     py::module m_sub = m.def_submodule("subsubmodule");
```
henryiii, line 17: Left one test with `py::module`

tests/test_embed/test_interpreter.cpp
henryiii: We are not testing C++20 anywhere, it's always 11+17, or 14.

rwgk: FWIW, using g++ (Debian 9.3.0-14) 9.3.0 or clang version 9.0.1-12 with `-std=c++2a` I don't get any errors.
Irrespective, I like this variable change 100%.

YannickJadoul (approved):
Wow, what a huge job, @henryiii! Thanks for tackling this chore!

rwgk (approved):
Thanks Henry!

wjakob (approved):
This looks excellent, thank you Henry!

PR #2545, opened 2020-10-02
Ensure the GIL is held when copying a function.

hwright:


include/pybind11/functional.h
```diff
@@ -58,7 +58,10 @@ struct type_caster<std::function<Return(Args...)>> {
         struct func_handle {
             function f;
             func_handle(function&& f_) : f(std::move(f_)) {}
-            func_handle(const func_handle&) = default;
+            func_handle(const func_handle& f_) {
+                gil_scoped_acquire acq;
+                f = f_.f;
+            }
             ~func_handle() {
                 gil_scoped_acquire acq;
```
henryiii, line 66: Given we already do a `scoped_aquire` here, I don't think it could hurt.

henryiii (approved):

YannickJadoul (approved):
What @henryiii said. That `function f` is an instance of a `py::object` subclass, so it shouldn't be touched without holding the GIL. Thanks for catching and fixing this, @hwright!

I also don't think this seemingly unique and small bugfix needs a dedicated test, especially because the test would be a lot more complex than the fix itself.

PR #2551, opened 2020-10-05
Use PyModule_New in module_ public constructor

YannickJadoul:
Alternative to the plan in #2534 and #2413: rather than making the `py::module` constructor private, we can turn it into a "normal" `py::object` constructor, and use `PyModule_New` to create a new `module` object (in the same way as `types.ModuleType("abc", "def")` would do so).

Modules create with `PyModule_New` *cannot* be used as extension modules (unfortunately). Python (3) complains if a module object is returned that doesn't have a `def`, so this is very unfortunately not a solution for the `static PyModuleDef` :-(

Things to note/judge/discuss:
- The old `PYBIND11_PLUGIN` example will nót work anymore (because the `py::module_` constructor has changed). This would also be the case if the constructor would be made private (though things would then just stop compiling). But, things don't segfault, and `PYBIND11_PLUGIN` has been deprecated for ages.
- For all other uses, except for `PYBIND11_PLUGIN`, things should work exactly the same. `PYBIND11_MODULE` now uses `pybind11::module_::create_extension_module`, so anything that's returned as top-level extension module is correctly constructed. For all other uses, the `module_` constructor should work.
- I've gotten rid of `detail::create_top_level_module` and just moved it to a public `static module_::create_extension_module`. Downside: Python 2 and 3 still differ in signature, but maybe I'm getting over that. Could be reverted, as well.
- One test uses the `py::module_` constructor. That's why I started thinking about keeping this constructor. That test still works, as the new constructor still creates a valid module object.

I'm adding this to v2.6.0, as a way of making sure we come up with a plan on what to include now and what to postpone for the future. E.g., right now, we could judge to just deprecate the public `module_` constructors, have users be warned in 2.6.0, and add this change in 2.7.0?

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -887,19 +887,28 @@ protected:
     }
 };
 
-/// Wrapper for Python extension modules
+/// Wrapper for Python modules
 class module_ : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    PYBIND11_DEPRECATED("Use PYBIND11_MODULE or module_::create_extension_module instead")
     explicit module_(const char *name, const char *doc = nullptr) {
-#if PY_MAJOR_VERSION >= 3
-        *this = create_extension_module(name, doc, new PyModuleDef());
+#if defined(PYPY_VERSION)
+        m_ptr = PyModule_New(const_cast<char *>(name));
 #else
-        *this = create_extension_module(name, doc, nullptr);
+        m_ptr = PyModule_New(name);
 #endif
+        if (m_ptr == nullptr)
+            pybind11_fail("Internal error in module_::module_()");
+        if (doc && options::show_user_defined_docstrings()) {
+#if PY_MAJOR_VERSION >= 3 && !defined(PYPY_VERSION)
+            if (PyModule_SetDocString(m_ptr, doc) != 0)
+                throw error_already_set();
+#else
+            setattr(m_ptr, "__doc__", PYBIND11_STR_TYPE(doc));
+#endif
+        }
     }
 
     /** \rst
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -272,8 +272,10 @@ extern "C" {
 
     .. code-block:: cpp
 
+        static pybind11::module_::module_def example_module_def;
         PYBIND11_PLUGIN(example) {
-            pybind11::module_ m("example", "pybind11 example plugin");
+            auto m = pybind11::module_::create_extension_module(
+                "example", "pybind11 example plugin", &example_module_def);
             /// Set up bindings here
             return m.ptr();
         }


henryiii:
FWIW, I think we should probably drop `PYBIND11_PLUGIN` in 3.0? I've still seen code with `PYBIND11_PLUGIN` in it. GitHub shows over 9K results for `PYBIND11_PLUGIN` (many are copies of readmes, old tutorials, talks, etc. But still, probably should be a 3.0 thing, IMO)

YannickJadoul:
> FWIW, I think we should probably drop `PYBIND11_PLUGIN` in 3.0?

Right. So in that case, we cannot add this PR without breaking existing usage of `PYBIND11_PLUGIN`?

rwgk:
To be totally honest, without having a full appreciation of all the details in this PR, I feel like this is getting way too complicated for very little gain.
I tend to think futuristic, in this case: there won't be Python 2 anymore.
Then we're left with one constructor that we don't really want to expose. From the global testing we know that (most likely) nobody ever wanted to use it anyway. Simply adding a comment: "please don't use this directly, only use our macros", would seem sufficient for all practical purposes.
More generally, to me the `private` feature has essential practical value for guarding against accidents that are likely to eat a lot of human time debugging, but trying to achieve purity when there isn't much of a potential for accidents anyway is unlikely to have positive value, measured in terms of human time won.

YannickJadoul:
@rwgk Thanks for the review!

I'm not sure I fully understand what you mean, but let me try to answer, anyway:

This PR's diff looks horrible, but actually isn't that huge or complex, I believe. The main thing this PR tries to make sure we can treat `module_` like any other `py::object` subclass.

But yes, I think this PR is too much to still include in 2.6.0. I've created #2552, which basically just deprecates the public constructor. As far as I'm concerned, this is basically just a formalization of your "please don't use this directly, only use our macros" comment, no? (One that's still ignorable, but less likely to be accidentally overlooked; after all, who reads the docs, let alone the comments?)

Would you mind giving #2552 a look? I've summarized the actual changes (vs. the perceived amount of diff): https://github.com/pybind/pybind11/pull/2552#issuecomment-705248126

>  I tend to think futuristic, in this case: there won't be Python 2 anymore.

As far as I know, Python 2 doesn't make things a lot more complex, in this case. But yes, completely agree that making changes becáuse of Python 2 is not a good idea anymore!

include/pybind11/pybind11.h
```diff
@@ -883,19 +883,28 @@ class cpp_function : public function {
...
 class module_ : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    PYBIND11_DEPRECATED("Use PYBIND11_MODULE or module_::create_extension_module instead")
     explicit module_(const char *name, const char *doc = nullptr) {
-#if PY_MAJOR_VERSION >= 3
-        *this = create_extension_module(name, doc, new PyModuleDef());
+#if defined(PYPY_VERSION)
```
YannickJadoul, line 893: Sumitted https://foss.heptapod.net/pypy/pypy/-/merge_requests/769 to fix this

YannickJadoul: Which got merged. So once we bump PyPy, this can be fixed.

include/pybind11/pybind11.h
```diff
@@ -887,19 +887,28 @@ class cpp_function : public function {
...
 };
 
-/// Wrapper for Python extension modules
+/// Wrapper for Python modules
 class module_ : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    PYBIND11_DEPRECATED("Use PYBIND11_MODULE or module_::create_extension_module instead")
```
rwgk, line 896: I think it's important to add a long comment here, to explain that v2.7 changes what `module_::module` does. It also needs a long-ish section in the changelog.

rwgk (approved):
Google-global testing came back green for this PR, too.
Approving, but please see my other comment.

include/pybind11/pybind11.h
```diff
@@ -887,19 +887,28 @@ class cpp_function : public function {
...
-#if PY_MAJOR_VERSION >= 3
-        *this = create_extension_module(name, doc, new PyModuleDef());
+#if defined(PYPY_VERSION)
+        m_ptr = PyModule_New(const_cast<char *>(name));
 #else
-        *this = create_extension_module(name, doc, nullptr);
+        m_ptr = PyModule_New(name);
 #endif
+        if (m_ptr == nullptr)
+            pybind11_fail("Internal error in module_::module_()");
```
bstaletic, line 903: "Internal error" sounds pretty vague, but I see pybind uses it all over the place and nobody complained.

include/pybind11/pybind11.h
bstaletic: Should we wait with this PR until we bump PyPy?

bstaletic (approved):
Two minor comments, but approved.

include/pybind11/pybind11.h
YannickJadoul: I don't think we should. This is really just a wrong signature (see PyPy PR) and PyPy isn't doing anything non-const, potentially dangerous with the `const_cast`ed string. We also don't know when PyPy will release 7.4.0, and whether we'll immediately drop support for 7.3.x, then. Meanwhile, the core of this PR is good.

I did add a more specific `(PYPY_VERSION_NUM < 0x07030400)` check, though. This should help when we do bump the minimum PyPy version supported.

include/pybind11/pybind11.h
YannickJadoul: Fixed to the almost-as-vague-but-slightly-better-and-more-consistent `"Could not allocate module object!"`. I don't know where I got `"Internal error"`. I suppose it was already there?

include/pybind11/pybind11.h
YannickJadoul: Better than the changelog, even: I'll add something to the upgrade guide as well, mentioning the error you get when you would return this as extension module.

YannickJadoul: Done!

Let me quickly still add a changelog entry. Apparently this PR is so old that it didn't even ask me on creating this PR!

YannickJadoul:
@henryiii, since you haven been involved in the discussion too much: if you have time for a quick read of the upgrade guide and consider that to be clear enough, I think we're good :-)

PR #2552, opened 2020-10-05
Deprecate public constructors of module_ class

YannickJadoul:
OK, so, this does a few things. A summary:

Main goal
- Deprecate current public `module_` constructors (see discussion in #2534 and #2413 on how they actually are not/should not be used). The useful public constructor for Python 2 is moved to `detail::create_top_level_module`.
- Use the new way of creating modules (i.e, with a `static PyModuleDef`) also for `PYBIND11_EMBEDDED_MODULE`. (This is kind of a consequence of the previous bullet, since it the deprecation warnings make it clear this wasn't changed yet in #2413.)

Minor extra cleanup
- Fully qualify the `pybind11` namespace by using `::pybind11` in these module-related macros.
- Turn the comment above `module_::add_object` into a valid doxygen-documentation of the function (and teach Doxygen that  `PYBIND11_NOINLINE` isn't really a relevant part of the signature to generate docs).

Two more questions to be answered:
1. Should this still be part of 2.6.0? I'd argue "yes", since
    a) I don't see a lot of uses for a plain `py::module_` constructor (@rwgk also found out that `module_`'s constructor is hardly used in Google's code base, which strengthens the belief that there's not a lot of use for that constructor), and
    b) this will allow us to maybe reuse this constructor later? (cfr. #2551), and
    c) what's the point of postponing this, and delaying the pain (if any). That's what deprecation is for, no?

2. Should we have `pybind11::detail::create_top_level_module` or `static pybind11::module_::create_extension_module`? I.e., is this a public interface or an internal one? I know I was the one arguing for the latter, before, but I think this being a public, but  "named" constructor would work for me, as well.

EDIT: `pybind11::detail::create_top_level_module` was moved to `static pybind11::module_::create_extension_module`, with a unified signature, though @rwkg's suggestion of a typedef!

### DIFF
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -46,10 +46,13 @@
         }
  \endrst */
 #define PYBIND11_EMBEDDED_MODULE(name, variable)                                \
-    PYBIND11_DETAIL_MODULE_STATIC_DEF(name)                                     \
+    static ::pybind11::module_::module_def                                      \
+        PYBIND11_CONCAT(pybind11_module_def_, name);                            \
     static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);   \
     static PyObject PYBIND11_CONCAT(*pybind11_init_wrapper_, name)() {          \
-        PYBIND11_DETAIL_MODULE_CREATE(name)                                     \
+        auto m = ::pybind11::module_::create_extension_module(                  \
+            PYBIND11_TOSTRING(name), nullptr,                                   \
+            &PYBIND11_CONCAT(pybind11_module_def_, name));                      \
         try {                                                                   \
             PYBIND11_CONCAT(pybind11_init_, name)(m);                           \
             return m.ptr();                                                     \
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -45,26 +45,27 @@
             });
         }
  \endrst */
-#define PYBIND11_EMBEDDED_MODULE(name, variable)                              \
-    static void PYBIND11_CONCAT(pybind11_init_, name)(pybind11::module_ &);    \
-    static PyObject PYBIND11_CONCAT(*pybind11_init_wrapper_, name)() {        \
-        auto m = pybind11::module_(PYBIND11_TOSTRING(name));                   \
-        try {                                                                 \
-            PYBIND11_CONCAT(pybind11_init_, name)(m);                         \
-            return m.ptr();                                                   \
-        } catch (pybind11::error_already_set &e) {                            \
-            PyErr_SetString(PyExc_ImportError, e.what());                     \
-            return nullptr;                                                   \
-        } catch (const std::exception &e) {                                   \
-            PyErr_SetString(PyExc_ImportError, e.what());                     \
-            return nullptr;                                                   \
-        }                                                                     \
-    }                                                                         \
-    PYBIND11_EMBEDDED_MODULE_IMPL(name)                                       \
-    pybind11::detail::embedded_module PYBIND11_CONCAT(pybind11_module_, name) \
-                              (PYBIND11_TOSTRING(name),             \
-                               PYBIND11_CONCAT(pybind11_init_impl_, name));   \
-    void PYBIND11_CONCAT(pybind11_init_, name)(pybind11::module_ &variable)
+#define PYBIND11_EMBEDDED_MODULE(name, variable)                                \
+    PYBIND11_DETAIL_MODULE_STATIC_DEF(name)                                     \
+    static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);   \
+    static PyObject PYBIND11_CONCAT(*pybind11_init_wrapper_, name)() {          \
+        PYBIND11_DETAIL_MODULE_CREATE(name)                                     \
+        try {                                                                   \
+            PYBIND11_CONCAT(pybind11_init_, name)(m);                           \
+            return m.ptr();                                                     \
+        } catch (::pybind11::error_already_set &e) {                            \
+            PyErr_SetString(PyExc_ImportError, e.what());                       \
+            return nullptr;                                                     \
+        } catch (const std::exception &e) {                                     \
+            PyErr_SetString(PyExc_ImportError, e.what());                       \
+            return nullptr;                                                     \
+        }                                                                       \
+    }                                                                           \
+    PYBIND11_EMBEDDED_MODULE_IMPL(name)                                         \
+    ::pybind11::detail::embedded_module PYBIND11_CONCAT(pybind11_module_, name) \
+                              (PYBIND11_TOSTRING(name),                         \
+                               PYBIND11_CONCAT(pybind11_init_impl_, name));     \
+    void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &variable)
 
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -874,7 +874,7 @@ public:
     PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)
 
     /// Create a new top-level Python module with the given name and docstring
-    PYBIND11_DEPRECATED("Use PYBIND11_MODULE, module_::def_submodule, or module_::import instead")
+    PYBIND11_DEPRECATED("Use PYBIND11_MODULE or module_::create_extension_module instead")
     explicit module_(const char *name, const char *doc = nullptr) {
 #if PY_MAJOR_VERSION >= 3
         *this = create_extension_module(name, doc, new PyModuleDef());
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -976,18 +976,19 @@ public:
             /* m_free */     nullptr
         };
         auto m = PyModule_Create(def);
-        if (m == nullptr)
-            pybind11_fail("Internal error in module_::create_extension_module()");
-        // TODO: Should be reinterpret_steal, but Python also steals it again when returned from PyInit_...
-        return reinterpret_borrow<module_>(m);
 #else
         // Ignore module_def *def; only necessary for Python 3
         (void) def;
         auto m = Py_InitModule3(name, nullptr, options::show_user_defined_docstrings() ? doc : nullptr);
-        if (m == nullptr)
-            pybind11_fail("Internal error in module_::create_extension_module()");
-        return reinterpret_borrow<module_>(m);
 #endif
+        if (m == nullptr) {
+            if (PyErr_Occurred())
+                throw error_already_set();
+            pybind11_fail("Internal error in module_::create_extension_module()");
+        }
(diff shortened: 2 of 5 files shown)


YannickJadoul:
OK, so, this does a few things. A summary:

Main goal
- Deprecate current public `module_` constructors (see discussion in #2534 and #2413 on how they actually are not/should not be used). The useful public constructor for Python 2 is moved to `detail::create_top_level_module`.
- Use the new way of creating modules (i.e, with a `static PyModuleDef`) also for `PYBIND11_EMBEDDED_MODULE`. (This is kind of a consequence of the previous bullet, since it the deprecation warnings make it clear this wasn't changed yet in #2413.)

Minor extra cleanup
- Fully qualify the `pybind11` namespace by using `::pybind11` in these module-related macros.
- Turn the comment above `module_::add_object` into a valid doxygen-documentation of the function (and teach Doxygen that  `PYBIND11_NOINLINE` isn't really a relevant part of the signature to generate docs).


Two more questions to be answered:
1. Should this still be part of 2.6.0. I'd argue "yes", since
    a) I don't see a lot of uses for a plain `py::module_` constructor (@rwgk also found out that `module_`'s constructor is hardly used in Google's code base, which strengthens the belief that there's not a lot of use for that constructor), and
    b) this will allow us to maybe reuse this constructor later? (cfr. #2551), and
    c) what's the point of postponing this, and delaying the pain (if any). That's what deprecation is for, no?

2. Should we have `pybind11::detail::create_top_level_module` or `static pybind11::module_::create_extension_module`? I.e., is this a public interface or an internal one? I know I was the one arguing for the latter, before, but I think this being a public, but  "named" constructor would work for me, as well.

henryiii (approved):
I think I mildly prefer the public name - macros like `PYBIND11_MODULE` tend to invariably have a few power users that need to avoid them for some special reason, so then this would be a requirement in that case. But we could wait and see, changing it to a public method in 1.6.1 if requested?

I believe the future direction is that we can better align with CPython's tools (like `PyModule_New`) in the future, possibly supporting multi-phase initialization eventually.

YannickJadoul:
> I think I mildly prefer the public name - macros like `PYBIND11_MODULE` tend to invariably have a few power users that need to avoid them for some special reason, so then this would be a requirement in that case. But we could wait and see, changing it to a public method in 1.6.1 if requested?

:+1: That's kind of what I was aiming at, yes. I had/have actually changed it in #2551 (which has become a lot smaller, after rebasing onto this PR :-) ), since it makes sense in that context (i.e., vanilla `module` constructor vs. a special "named constructor" for extension modules). The only thing that's bothering me is that the interface differs between Python 2 and 3 (but ... yeah, Python 2 should die anyway, so ...).
Just wondered if we should already use that new name here; but doing that in a later release also works for me!

> I believe the future direction is that we can better align with CPython's tools (like `PyModule_New`) in the future, possibly supporting multi-phase initialization eventually.

Yessssss!

include/pybind11/pybind11.h
```diff
@@ -946,55 +944,56 @@ class module_ : public object {
...
+        /* m_size */     -1,
+        /* m_methods */  nullptr,
+        /* m_slots */    nullptr,
+        /* m_traverse */ nullptr,
+        /* m_clear */    nullptr,
+        /* m_free */     nullptr
+    };
+    auto m = PyModule_Create(def);
+    if (m == nullptr)
+        pybind11_fail("Internal error in detail::create_top_level_module()");
```
rwgk, line 984: Idea 1: does `throw error_already_set()` work?
Idea 2: somehow include `name` in the error message, to be helpful?

rwgk (approved):
Hi @YannickJadoul, I agree this is better, I really like the deprecation. Please feel free to merge.

Optional, only if you feel it has merit: there is a lot of PY2 vs PY3 `#if` branching. Maybe it could be reduced via something like:
```
#if PY_MAJOR_VERSION >= 3
typedef PyModuleDef module_def;
#else 
struct module_def {};
#endif
```
Then the code for PY2 code could be mostly identical to PY3 code.
If this is feasible, there will be a universal `create_top_level_module()`. I'd definitely make it public in that case. I'm less certain about making it public if there are different versions for PY2 and PY3, because it will force downstream users to also have `#if`, unless they are certain only PY3 is needed.

YannickJadoul:
> If this is feasible, there will be a universal `create_top_level_module()`. I'd definitely make it public in that case. 

Great idea, yes! That's exactly what's holding me back. I had been thinking about a typedef with `void`(so we'd get `void*`), but you can't define `void` variables. I guess that extra word for an empty static struct instance won't make a difference?

rwgk:
> I guess that extra word for an empty static struct instance won't make a difference?

It's only one per module, which is almost certainly below the noise level, but also only for the un-dead Python 2: IMO nothing to worry about at all.

henryiii:
>  but also only for the un-dead Python 2

We are definitely at the point where we should design for Python 3, and if it makes Python 2 a little ugly, I don't think that's something to worry about. Python 3 will be around a lot longer. :)

YannickJadoul:
OK, perfect! I agree, but wanted to have it mentioned :-)

I'm now just looking at the error message, to finish this PR, but I can't find if a `nullptr` mean the error indicator is already set?
https://docs.python.org/3/c-api/module.html#c.PyModule_Create

rwgk (approved):
Awesome, I just have some minor polishing suggestions.

include/pybind11/pybind11.h
```diff
@@ -959,11 +949,21 @@ class module_ : public object {
...
+    using module_def = PyModuleDef;
+#else
+    struct module_def {};
+#endif
+
+    /** \rst
+        Create a new top-level module that can be used as the main module of a C extension.
 
-    explicit module_(const char *name, const char *doc, PyModuleDef *def) {
+        For Python 3, ``def`` should point to a staticly allocated PyModuleDef.
```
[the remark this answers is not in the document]
YannickJadoul: Also fine, yes; will change.

tests/test_modules.cpp
```diff
@@ -62,7 +62,8 @@ TEST_SUBMODULE(modules, m) {
         class Dupe3 { };
         class DupeException { };
 
-        auto dm = py::module_("dummy");
+        // Go ahead and leak, until we have a non-leaking py::module_ constructor
+        auto dm = py::module_::create_extension_module("dummy", nullptr, new py::module_::module_def);
```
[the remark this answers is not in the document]
YannickJadoul: The test function couldn't be called twice, then. Not sure if that's needed, but I thought I'd prefer a leak over a weird future segfault (and it was already leaking in Python 3 before)? Can change if you want to though.

#2551 would fix this by adding a non-extension module constructor that doesn't require a `PyModuleDef`, btw :-)

include/pybind11/pybind11.h
```diff
@@ -975,27 +975,27 @@ class module_ : public object {
...
+        // TODO: Should be reinterpret_steal, but Python also steals it again when returned from PyInit_...
+        return reinterpret_borrow<module_>(m);
+#else
+        // Ignore module_def *def; only necessary for Python 3
+        (void) def;
+        auto m = Py_InitModule3(name, nullptr, options::show_user_defined_docstrings() ? doc : nullptr);
+        if (m == nullptr)
+            pybind11_fail("Internal error in module_::create_extension_module()");
+        return reinterpret_borrow<module_>(m);
 #endif
```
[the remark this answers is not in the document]
YannickJadoul: I thought about that, but decided not to because of the `// TODO: Should be reinterpret_steal, but Python also steals it again when returned from PyInit_...` comment (which only holds for Python 3; in Python 2, it should be `reinterpret_borrow`).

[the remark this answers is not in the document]
YannickJadoul: > Did you see my question about `pybind11_fail` vs `throw error_already_set()` from yesterday?

Yes. Wait, let me reply in that thread.

include/pybind11/pybind11.h
YannickJadoul: Answered in the main thread, but that was maybe not a great idea:

> I'm now just looking at the error message, to finish this PR, but I can't find if a nullptr mean the error indicator is already set?
> https://docs.python.org/3/c-api/module.html#c.PyModule_Create
>
> EDIT: If it definitely does, we should throw py::error_already_set to be fully correct!

tests/test_modules.cpp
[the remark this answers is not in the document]
rwgk: I'm OK with the leak. Easy enough to fix later as needed.

include/pybind11/pybind11.h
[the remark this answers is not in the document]
YannickJadoul: Done

include/pybind11/pybind11.h
rwgk: Looking at the 3.7 sources, I see some code paths set an error for sure, but it's difficult to know if that's always the case. The bomb-proof solution would be something like `if (PyErr_Occurred()) throw error_already_set();` and `pybind11_fail()` as a fallback, with `name` in the error message.
I kind of feel it's worth it: in case this fires it's most likely a difficult to debug situation,  and showing the original error (root cause) or the module `name` could save a lot of time troubleshooting.
Unless we can convince the Python developers to document that a NULL return guarantees the error is set ...

henryiii: You might not need to add name to the error message, if it usually or hopefully always works with `error_already_set()`, as that makes `pybind11_fail()` a last-ditch fallback.

henryiii: (Assuming error_already_set would show the name?)

YannickJadoul: Myeah, or we just trust that this part of the API works in the same way? An `error_already_set` won't cause things to be messed up too badly.

I can't both throw `error_already_set` ánd have the `name` in there, though. So that's only as a fall-back?

YannickJadoul: > (Assuming error_already_set would show the name?)

I don't know; we don't control that, since that the message is already set by Python. It might be something as cryptic as "didn't manage to create a `str` from this `const char*`"

YannickJadoul: OK, didn't add the name, as the `pybind11_fail` is only our fallback option. We can add it, if we find out that (and when!) the control flow can reach that statement?

include/pybind11/embed.h
```diff
@@ -45,26 +45,30 @@
...
+        PYBIND11_CONCAT(pybind11_module_def_, name);                            \
+    static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);   \
+    static PyObject PYBIND11_CONCAT(*pybind11_init_wrapper_, name)() {          \
+        auto m = ::pybind11::module_::create_extension_module(                  \
+            PYBIND11_TOSTRING(name), nullptr,                                   \
+            &PYBIND11_CONCAT(pybind11_module_def_, name));                      \
+        try {                                                                   \
+            PYBIND11_CONCAT(pybind11_init_, name)(m);                           \
+            return m.ptr();                                                     \
+        } catch (::pybind11::error_already_set &e) {                            \
```
wjakob, line 59: Can we use PYBIND11_CATCH_INIT_EXCEPTIONS here?

wjakob:
Looks all good to me, just a minor note about the repeated exception handling code in the macro declaration.

include/pybind11/embed.h
YannickJadoul: I'll pour that into another PR, yes.

PR #2553, opened 2020-10-06
fix: PYBIND11_OBJECT required pybind11 namespace (regression)

henryiii:
Fixes a regression in #2349.


YannickJadoul (approved):
Yup, overlooked that! Thanks, @henryiii 

No clue why suddenly Python 3.9 is consistently segfaulting, though :-( I restarted 2 or 3 times, and it's still there...

PR #2555, opened 2020-10-06
docs: minor cleanup

henryiii:
Biggest fix is missing union type. A few other cleanups.

Also, at the recent HSF packaging meeting, it was pointed out that setting `CMAKE_CXX_STANDARD` as a normal variable makes it impossible to override from the command line or a super project; while I would argue that you should never set it on anything that is included elsewhere, that is a good point and probably best not to show something that could easily be wrongly copied in.

This should fix all the warnings from the recent change to GitHub Actions to patch a security venerability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,10 +91,11 @@ jobs:
         architecture: ${{ matrix.arch }}
 
     - name: Setup Boost (Windows / Linux latest)
-      run: echo "::set-env name=BOOST_ROOT::$BOOST_ROOT_1_72_0"
+      shell: bash
+      run: echo "BOOST_ROOT=$BOOST_ROOT_1_72_0" >> $GITHUB_ENV
 
     - name: Update CMake
-      uses: jwlawson/actions-setup-cmake@v1.3
+      uses: jwlawson/actions-setup-cmake@v1.4
 
     - name: Cache wheels
       if: runner.os == 'macOS'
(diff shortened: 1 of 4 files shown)


henryiii:
Final warnings should be fixed if a new version of setup-cmake is released, see https://github.com/jwlawson/actions-setup-cmake/issues/11.

YannickJadoul (approved):
Woohoo! Thanks again, Henry! :-)

.github/workflows/ci.yml
```diff
@@ -166,16 +167,17 @@ jobs:
     # MSVC, but for now, this action works:
     - name: Prepare compiler environment for Windows 🐍 2.7
       if: matrix.python == 2.7 && runner.os == 'Windows'
-      uses: ilammy/msvc-dev-cmd@v1
+      uses: ilammy/msvc-dev-cmd@v1.4.1
```
henryiii, line 170:
```suggestion
      uses: ilammy/msvc-dev-cmd@v1
```

The v1 tag was [fixed upstream](https://github.com/ilammy/msvc-dev-cmd/issues/16). Thanks @ilammy!

.github/workflows/ci.yml
```diff
@@ -91,10 +91,11 @@ jobs:
         architecture: ${{ matrix.arch }}
 
     - name: Setup Boost (Windows / Linux latest)
-      run: echo "::set-env name=BOOST_ROOT::$BOOST_ROOT_1_72_0"
+      shell: bash
+      run: echo "BOOST_ROOT=$BOOST_ROOT_1_72_0" >> $GITHUB_ENV
 
     - name: Update CMake
-      uses: jwlawson/actions-setup-cmake@v1.3
+      uses: jwlawson/actions-setup-cmake@v1.4
```
henryiii, line 98: Thanks @jwlawson for the new release!

docs/upgrade.rst
```diff
@@ -88,7 +89,8 @@ In addition, the following changes may be of interest:
 * Using ``find_package(Python COMPONENTS Interpreter Development)`` before
   pybind11 will cause pybind11 to use the new Python mechanisms instead of its
   own custom search, based on a patched version of classic ``FindPythonInterp``
-  / ``FindPythonLibs``. In the future, this may become the default.
+  / ``FindPythonLibs``. In the future, this may become the default. A recent
+  (3.15+ or 3.18.2+) version of CMake recommended.
```
rwgk, line 93: **is** recommended
?

henryiii: 😊

Issue #2558, opened 2020-10-08
[bug] Python 3.9 macOS segfault


henryiii:
The following code segfaults on Python 3.9 and macOS:

```python
from pybind11_tests import gil_scoped as m

def _python_to_cpp_to_python():
    """Calls different C++ functions that come back to Python."""
    class ExtendedVirtClass(m.VirtClass):
        def virtual_func(self):
            pass

        def pure_virtual_func(self):
            pass

    extended = ExtendedVirtClass()
    m.test_callback_py_obj(lambda: None)
    m.test_callback_std_func(lambda: None)
    m.test_callback_virtual_func(extended)
    m.test_callback_pure_virtual_func(extended)


for x in range(1000):
    _python_to_cpp_to_python()
```

Debug output:

```
* thread #2, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x74736e6f43)
  * frame #0: 0x000000010017434a Python`meth_dealloc + 120
    frame #1: 0x000000010014645a Python`instancemethod_dealloc + 65
    frame #2: 0x0000000100168feb Python`PyDict_Clear + 409
    frame #3: 0x00000001001866ec Python`type_clear + 59
    frame #4: 0x0000000100233f00 Python`collect + 2558
    frame #5: 0x00000001002334f1 Python`_PyGC_CollectNoFail + 67
    frame #6: 0x000000010020619c Python`_PyImport_Cleanup + 1545
    frame #7: 0x0000000100216680 Python`Py_FinalizeEx + 160
    frame #8: 0x000000010023279e Python`Py_RunMain + 1408
    frame #9: 0x0000000100232c86 Python`pymain_main + 306
    frame #10: 0x0000000100232cd4 Python`Py_BytesMain + 42
    frame #11: 0x00007fff6910dcc9 libdyld.dylib`start + 1
```

_Originally posted by @henryiii in https://github.com/pybind/pybind11/issues/2391#issuecomment-705289620_

rwgk:
> ```python
> from pybind11_tests import gil_scoped as m
> ```

IIUC `gil_scoped` is a submodule, which is a little tricker for me to get going in the Google environment. Is there a chance you could try isolating the 4 callback functions in a minimal main module? — I could do that myself, but then I will not be sure that it still fails in your environment.

henryiii:
That'll take a little time, since I have to pull the cross module utils out to build separately.

rwgk:
[...]

PR #2559, opened 2020-10-08
test: hide segfault

henryiii:
Trying to trigger segfault in main process. See #2558 .

- tests: try to isolate segfault into one test
- docs: minor typo caught by @rwgk

### DIFF
--- a/tests/test_gil_scoped.py
+++ b/tests/test_gil_scoped.py
@@ -94,6 +94,7 @@ def test_python_to_cpp_to_python_from_process():
     assert _run_in_process(_python_to_cpp_to_python) == 0
 
 
+@pytest.mark.xfail("env.PY >= (3,9) and env.MACOS", run=False)
 def test_cross_module_gil():
     """Makes sure that the GIL can be acquired by another module from a GIL-released state."""
     m.test_cross_module_gil()  # Should not raise a SIGSEGV
--- a/tests/test_gil_scoped.py
+++ b/tests/test_gil_scoped.py
@@ -2,15 +2,6 @@
 import multiprocessing
 import threading
 
-import pytest
-
-import env  # noqa: F401
-
-
-if env.PY >= (3,9) and env.MACOS:
-    pytest.skip("skipping segfaulting macOS Python 3.9 tests", allow_module_level=True)
-
-
 from pybind11_tests import gil_scoped as m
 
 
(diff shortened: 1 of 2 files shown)


henryiii:
I can get this to trigger standalone, but it doesn't seem to trigger in pytest (locally too).

tests/test_gil_scoped.py
```diff
@@ -97,3 +97,9 @@ def test_python_to_cpp_to_python_from_process():
 def test_cross_module_gil():
     """Makes sure that the GIL can be acquired by another module from a GIL-released state."""
     m.test_cross_module_gil()  # Should not raise a SIGSEGV
+
+
+# Checks for segfault
```
rwgk, line 102: # Attempts to generate non-deterministic Python 3.9 MacOS segfault.
?

tests/test_gil_scoped.py
```diff
@@ -97,3 +97,9 @@ def test_python_to_cpp_to_python_from_process():
...
+def test_potential_segfault():
```
rwgk, line 103: I could port just this test to `unittest`, so that I can run it with our sanitizers (ASAN, MSAN, TSAN).
Does that sound like a useful idea?

henryiii: Yes yes yes! Just running it in plain Python does cause a segfault ~50% of the time (possibly regardless of the number of times it runs). There are no actual asserts here, so should be trivial to run.

henryiii: (The single file I posted in the issue fails by itself if run directly; you just need to put it in the right place to load the cpp output module).

tests/test_gil_scoped.py
henryiii: Yes, though I probably will just pull this one experimental test; having our tests fail due to this, now that 3.9 is out, is probably not a good idea as long as we are actively trying to solve the issue. And this isn't breaking the tests anyway...

henryiii:
Made this just not run the segfaulting tests, so we can see more green checkmarks. The bug is being investigated separately in #2559.

henryiii:
I think the problem is more serious than first expected. The segfault happens when the process is cleaning up. For these examples, since it runs in a separate process, the return code of "-11" means the process segfaulted. But the thing causing the segfault can be other places, too, since this still segfaults at the end. I can try disabling the whole test, and hope that it's just that final test that's causing the issue.

PR #2561, opened 2020-10-08
2.6.0: Use sorted(glob()) in example setup.py

jbarlow83:
I suggest changing the example setup.py as follows below, to use a sorted glob instead of naming one source file.

This strategy eliminates errors from forgetting to enumerate source files one by one when a new .cpp is added. Using `sorted(glob())` also ensures the build will be reproducible on file systems that do not enumerate files in sorted order (most Linux filesystems).

I also added a missing code-block directive.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -31,13 +31,14 @@ An example of a ``setup.py`` using pybind11's helpers:
 
 .. code-block:: python
 
+    from glob import glob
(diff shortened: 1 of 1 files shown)


henryiii (approved):
I think this looks like a good idea. Was trying to keep simple, but this doesn't add too much, and keeps the sorted detail present.

Issue #856, opened 2017-05-16
py::keep_alive: nurse destructor can run before patient's


bmerry:
#### Issue description

The py::keep_alive weakref trick doesn't always work. In some cases (at least in Python 2.7, haven't tried 3), the weakref callback can run before the referenced object is cleared. This happens in the cyclic garbage collector: first all weakref callbacks are handled (`handle_weakrefs` function), then the garbage is cleared. If the nurse is being destroyed on this path, then it can lead to the patient destructor running before the nurse's (leading to undefined behaviour if the destructor of the nurse references the patient). Note that the nurse doesn't itself have to be part of a reference cycle, it just has to be reachable from one. However, it needs to have `Py_TPFLAGS_HAVE_GC`. In the example code I've ensured that by specifying `py::dynamic_attr`, but I think when I originally encountered the problem it was set due to having a declared base.

I've had similar issues with Boost.Python before. I suspect the weakref approach is fundamentally untenable; it may be necessary to store a list of handles (with strong references) in the PyObject and decref them when the object is cleared.

#### Reproducible example code

keepalive.cpp:
```c++
#include <pybind11/pybind11.h>

namespace py = pybind11;

class B;

class A {
private:
    B &b;
public:
    A(B &b) : b(b) {}
    ~A() { py::print("In A::~A()"); }
};

class B {
public:
    ~B() { py::print("In B::~B()"); }
};

PYBIND11_PLUGIN(keepalive) {
    py::module m("keepalive");

    py::class_<A>(m, "A", py::dynamic_attr())
        .def(py::init<B &>(), py::keep_alive<1, 2>());

    py::class_<B>(m, "B").def(py::init<>());
    return m.ptr();
}
```

test.py:
```python
from keepalive import A, B

lst = [A(B())]
lst.append(lst)
del lst
```

Output is
```
In B::~B()
In A::~A()
```
Expected output is the other way around.

bmerry:
Or maybe it would work for the callback function to call the C++ destructor if it hasn't already been called, and deal with that gracefully when the object is cleared - but I don't know nearly enough about the internals to know if that's sane or not.

jagerman:
[...]

PR #2564, opened 2020-10-08
fix: test_factory_constructors.py failure triggered by test_register_duplicate_class

YannickJadoul:
To reproduce, this fails on the `centos:8` docker image (with development tools installed - see `ci.yml` - but without per se installing nvidia compilers), by running `PYTEST_ADDOPTS="-k 'test_register_duplicate_class or test_init_factory_alias'" cmake --build build_centos8 --target pytest`

Cross-reference #2335

EDIT: Figured out what's wrong. Short description of the bug:
- `test_class.py::test_register_duplicate_class` registers `py::class_` instances in different scopes, a module and a class (to test whether duplicate class names or types are caught; but that's not really the issue here).
- In this test, the actual scopes passed to the `py::class_` constructor are not referenced anymore after the scope of the test, and they go out of scope and get garbage collected. Since the `py::class_`es are only referenced by these scopes that are getting garbage collected, they also get garbage collected. However, they _remain in the internals' `registered_types_py`_.
- When Python is allocating a new type, it reuses the old memory location of the garbage collected type object from before.
- So, the same `PyTypeObject` pointer in `registered_types_py` now points to a new Python type, but _still associates this with the old `type_info`_ (it is likely important that the new type allocated in "recycled" memory is a non-pybind11 type, in the case I debugged; if not, the associated `type_info` in `registered_types_py` would be overwritten).
- pybind11 doesn't understand things anymore, because it's allocating the new type, but there's now a mismatch between the Python and C++ type, because of the stale `PyTypeObject*` in `registered_types_py`.

Current solution:
- I copied the `weakref` from `detail::all_type_info_get_cache`. I still want to check if there's a better way of solving this, but this solves the problem.

Possible follow-up:
- Now that stale classes get removed from the internals, we could think about not checking whether we're overwriting an already existing attribute in `class_`. I'll give this some more thought and I'll make a PR to further discuss this.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1021,6 +1021,30 @@ inline dict globals() {
 }
 
 PYBIND11_NAMESPACE_BEGIN(detail)
+/// Cleanup the type-info for a pybind11-registered type.
+PYBIND11_NOINLINE inline void cleanup_type_info(detail::type_info *tinfo) {
+    auto &internals = get_internals();
+    auto tindex = std::type_index(*tinfo->cpptype);
+    internals.direct_conversions.erase(tindex);
+
+    if (tinfo->module_local)
+        registered_local_types_cpp().erase(tindex);
+    else
+        internals.registered_types_cpp.erase(tindex);
+    internals.registered_types_py.erase(tinfo->type);
+
+    // Actually just `std::erase_if`, but that's only available in C++20
+    auto &cache = internals.inactive_override_cache;
+    for (auto it = cache.begin(), last = cache.end(); it != last; ) {
+        if (it->first == (PyObject *) tinfo->type)
+            it = cache.erase(it);
+        else
+            ++it;
+    }
+
+    delete tinfo;
+}
+
 /// Generic support for creating new Python heap types
 class generic_type : public object {
     template <typename...> friend class class_;
@@ -1037,11 +1061,10 @@ protected:
                           "\" is already registered!");
 
         m_ptr = make_new_python_type(rec);
-        auto type = (PyTypeObject *) m_ptr;
 
         /* Register supplemental type information in C++ dict */
         auto *tinfo = new detail::type_info();
-        tinfo->type = type;
+        tinfo->type = (PyTypeObject *) m_ptr;
         tinfo->cpptype = rec.type;
         tinfo->type_size = rec.type_size;
         tinfo->type_align = rec.type_align;
@@ -1056,31 +1079,16 @@ protected:
 
         auto &internals = get_internals();
         auto tindex = std::type_index(*rec.type);
-        auto module_local = rec.module_local;
         tinfo->direct_conversions = &internals.direct_conversions[tindex];
-        if (module_local)
+        if (rec.module_local)
             registered_local_types_cpp()[tindex] = tinfo;
         else
             internals.registered_types_cpp[tindex] = tinfo;
-        internals.registered_types_py[type] = { tinfo };
+        internals.registered_types_py[(PyTypeObject *) m_ptr] = { tinfo };
 
         // Clean up our internals after the Python type object gets garbage collected
-        weakref(m_ptr, cpp_function([type, tindex, module_local](handle wr) {
-            auto &internals = get_internals();
-            internals.direct_conversions.erase(tindex);
-            if (module_local)
-                registered_local_types_cpp().erase(tindex);
-            else
-                internals.registered_types_cpp.erase(tindex);
-            internals.registered_types_py.erase(type);
-            // Actually just `std::erase_if`, but that's only available in C++20
-            auto &cache = internals.inactive_override_cache;
-            for (auto it = cache.begin(), last = cache.end(); it != last; ) {
-                if (it->first == (PyObject *) type)
-                    it = cache.erase(it);
-                else
-                    ++it;
-            }
+        weakref(m_ptr, cpp_function([tinfo](handle wr) {
+            cleanup_type_info(tinfo);
             wr.dec_ref();
         })).release();
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1073,10 +1073,7 @@ protected:
             else
                 internals.registered_types_cpp.erase(tindex);
             internals.registered_types_py.erase(type);
-            // C++20: std::erase_if(internals.inactive_override_cache,
-            //                      [type](const auto &entry) {
-            //                          return entry.first == (PyObject *) type;
-            //                      });
+            // Actually just `std::erase_if`, but that's only available in C++20
             auto &cache = internals.inactive_override_cache;
             for (auto it = cache.begin(), last = cache.end(); it != last; ) {
                 if (it->first == (PyObject *) type)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1028,7 +1028,7 @@ public:
     PYBIND11_OBJECT_DEFAULT(generic_type, object, PyType_Check)
 protected:
     void initialize(const type_record &rec) {
-        if (rec.scope && hasattr(rec.scope, "__dict__") && rec.scope.attr("__dict__").contains(rec.name))
+        if (rec.scope && hasattr(rec.scope, rec.name))
             pybind11_fail("generic_type: cannot initialize type \"" + std::string(rec.name) +
                           "\": an object with that name is already defined");
 
@@ -1946,7 +1946,7 @@ public:
         std::string full_name = scope.attr("__name__").cast<std::string>() +
                                 std::string(".") + name;
         m_ptr = PyErr_NewException(const_cast<char *>(full_name.c_str()), base.ptr(), NULL);
-        if (hasattr(scope, "__dict__") && scope.attr("__dict__").contains(name))
+        if (hasattr(scope, name))
             pybind11_fail("Error during initialization: multiple incompatible "
                           "definitions with name \"" + std::string(name) + "\"");
         scope.attr(name) = *this;
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1037,10 +1037,11 @@ protected:
                           "\" is already registered!");
 
         m_ptr = make_new_python_type(rec);
+        auto type = (PyTypeObject *) m_ptr;
 
         /* Register supplemental type information in C++ dict */
         auto *tinfo = new detail::type_info();
-        tinfo->type = (PyTypeObject *) m_ptr;
+        tinfo->type = type;
         tinfo->cpptype = rec.type;
         tinfo->type_size = rec.type_size;
         tinfo->type_align = rec.type_align;
@@ -1060,7 +1061,11 @@ protected:
             registered_local_types_cpp()[tindex] = tinfo;
         else
             internals.registered_types_cpp[tindex] = tinfo;
-        internals.registered_types_py[(PyTypeObject *) m_ptr] = { tinfo };
+        internals.registered_types_py[type] = { tinfo };
+        weakref(m_ptr, cpp_function([type](handle wr) {
+            get_internals().registered_types_py.erase(type);
+            wr.dec_ref();
+        })).release();
 
         if (rec.bases.size() > 1 || rec.multiple_inheritance) {
             mark_parents_nonsimple(tinfo->type);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1028,7 +1028,7 @@ public:
     PYBIND11_OBJECT_DEFAULT(generic_type, object, PyType_Check)
 protected:
     void initialize(const type_record &rec) {
-        if (rec.scope && hasattr(rec.scope, rec.name))
+        if (rec.scope && hasattr(rec.scope, "__dict__") && rec.scope.attr("__dict__").contains(rec.name))
             pybind11_fail("generic_type: cannot initialize type \"" + std::string(rec.name) +
                           "\": an object with that name is already defined");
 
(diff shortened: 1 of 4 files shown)


henryiii:
I wonder if this is related to #2558 ?

YannickJadoul:
> I wonder if this is related to #2558 ?

It would be amazing if it is (and we only need to solve 1 issue instead of 2), but I wouldn't really know why it would be?

henryiii:
https://github.com/pybind/pybind11/pull/2564/checks?check_run_id=1227843279 looks suspicious to me. Notice the second test that failed (though not a segfault).

wjakob:
That's a nice catch. One word of caution about `py::weakref`: the garbage collector is a complex beast, and although the weakref callback will eventually be invoked, it's very difficult to guarantee that this happens at any particular point in time, or with a particular ordering compared to other stuff that is happening. Although it was a really neat trick, we had to get rid of weak references in the implementation of `py::keep_alive` (https://github.com/pybind/pybind11/issues/856) for similar reasons. Based on these experiences, it seems to me that you might just be replacing one kind of undefined behavior by another (albeit likely much more rare).

YannickJadoul:
Hmmm, interesting. I'll still give that a look!

I did think of this, but it seemed reasonably deterministic, the way it's described in the Python docs (both normal as well as C API):

- The Python docs for weakref (https://docs.python.org/3/library/weakref.html#weakref.ref):
   > If callback is provided and not None, and the returned weakref object is still alive, the callback will be called when the object is about to be finalized; the weak reference object will be passed as the only parameter to the callback; the referent will no longer be available.
- The Python docs for https://docs.python.org/3/c-api/weakref.html#c.PyWeakref_NewRef is even more clear
9:14

    The second parameter, callback, can be a callable object that receives notification when ob is garbage collected; it should accept a single parameter, which will be the weak reference object itself.

YannickJadoul:
> I did think of this, but it seemed reasonably deterministic, the way it's described in the Python docs (both normal as well as C API):

Scratch all that. Should've read the issue first; this is not really related.

@wjakob, @henryiii, @EricCousineau-TRI, I still think we should be safe, because we're not too concerned about order, here? Worst case that could happen is some type gets removed from the internals, a bit before that type is actually destructed, but áfter the last reference to this type is gone? (Except if there's some weird cycle, perhaps, and the type is still used in the destructor? But I'm not sure we could fix that anyway, given the arbitrary destruction order of a cycle?)

We could try a similar approach to #856; I thought about that before seeing the `weakref` approach, but it would involve the metaclass, so we'd again bump into potential issues when using custom metaclasses? :-/

henryiii:
I think this is fine, as it fixes the issue we are seeing, though if something else comes up, we can revisit?

YannickJadoul:
> I think this is fine, as it fixes the issue we are seeing, though if something else comes up, we can revisit?

Let's give @EricCousineau-TRI still a chance to review? He said he might have some time tomorrow. And I have a feeling he's been playing in these parts of pybind11 before.

YannickJadoul:
This is in `internals`:

```
    std::unordered_set<std::pair<const PyObject *, const char *>, override_hash> inactive_override_cache;
    type_map<std::vector<bool (*)(PyObject *, void *&)>> direct_conversions;
    std::vector<PyObject *> loader_patient_stack; // Used by `loader_life_support`
    std::forward_list<std::string> static_strings; // Stores the std::strings backing detail::c_str()
    PyTypeObject *static_property_type;
    PyTypeObject *default_metaclass;
    PyObject *instance_base;
```

What else should we clear out? `registered_types_cpp` as well, right?

YannickJadoul:
Thinking a bit further: is that the behavior we actually want?

I.e., in this case, once the Python type object (the one of the class_) gets garbage collected, it's unregistered. So it would also mean you can't go from C++ to Python, anymore. So you couldn't register an "anonymous" type with `class_` that you're only returning from C++ to Python (unless you explicitly keep it alive on the C++ side).

Or, you register a C++ type `Foo` as `my_lovely_module.Bar` Python type, and a user manage to clear all references to `my_lovely_module.Bar`, and suddenly, a function returning `Foo` will not be able to convert its return type. Are we happy with that?

(To be clear, the above scenario currently results in a segfault, so at least it's better. But the other option would be to keep C++-registered type alive forever.)

henryiii:
I think this is good for now, as it's at least better, and I don't think deleting class objects is common (except in some tests). And if you go to the trouble of deleting all references to a class, would you expect it to magically reappear if you tried to do a conversion afterwords?

include/pybind11/pybind11.h
```diff
@@ -1040,12 +1041,36 @@ class generic_type : public object {
...
+        // Clean up our internals after the Python type object gets garbage collected
+        weakref(m_ptr, cpp_function([type, tindex, module_local](handle wr) {
+            auto &internals = get_internals();
+            internals.direct_conversions.erase(tindex);
+            if (module_local)
+                registered_local_types_cpp().erase(tindex);
+            else
+                internals.registered_types_cpp.erase(tindex);
+            internals.registered_types_py.erase(type);
+            // C++20: std::erase_if(internals.inactive_override_cache,
```
EricCousineau-TRI, line 1061: nit While this is cool, I think it's a bit o' clutter. Would you be able to delete this comment?

(You could make a permalink, and then reference it here? e.g.
```
// TODO: Use C++20 code (see commit X on PR Y).
```

include/pybind11/pybind11.h
```diff
@@ -1040,12 +1041,36 @@ class generic_type : public object {
...
+            //                          return entry.first == (PyObject *) type;
+            //                      });
+            auto &cache = internals.inactive_override_cache;
+            for (auto it = cache.begin(), last = cache.end(); it != last; ) {
+                if (it->first == (PyObject *) type)
+                    it = cache.erase(it);
+                else
+                    ++it;
+            }
+            wr.dec_ref();
```
EricCousineau-TRI, line 1072: Not sure if it'd be easily addressable, but there may be lingering references from other types here:
https://github.com/pybind/pybind11/blob/7c71dd386161703dc25860870f579cca0ff3a9a8/include/pybind11/detail/internals.h#L135-L137

e.g. you're freeing type A, but type B has a conversion related to that type

EricCousineau-TRI: FWIW This PR seems like a strict improvement as-is, so I'd totes be fine with a TODO / follow-up issue.

EricCousineau-TRI: - `implicit_casts` looks like it's only used for MI, so by construction I don't think you'd have a dangling parent base.
- `direct_conversions` looks like it's only used for NumPy dtypes
- `implicit_conversions` looks like it requires that both types be alive, so also should be fine. However, you may have unbounded memory growth.

So yeah, looks like everything is fine. Just may want to make a minor note on `implicit_conversiosn` being unbounded; however, it's unlikely that unbounded growth will matter much? (unless someone likes running the GC every 30sec for 24hrs?)

EricCousineau-TRI (approved):
Looks good! Just some minor comments.

include/pybind11/pybind11.h
henryiii: Honestly, I don't mind it (and I hate commented out code). It shows the intended logic in a more concise way especially readable to someone used to C++20. But commented out code becomes "stale", and is untested, so I'd also be okay to see it go.

YannickJadoul: As you wish. I was just frustrated that C++11 didn't have `erase_if` (and `remove_if` is not really applicable to `unordered_set`, I think).

I don't think the permalink would make it so much better, then. I can just remove it and put just something like `// Actually just `std::erase_if`, but only available in C++20`. By the time we move to C++20, I'm sure it'll be easy enough to clean up?

YannickJadoul:
> And if you go to the trouble of deleting all references to a class, would you expect it to magically reappear if you tried to do a conversion afterwords?

> Additionally, a user could prevent that behavior by stashing a reference to the module (or type), preventing garbage colllection:

True, OK. I was just afraid of unexpected surprises where suddenly a functions return type can't be returned anymore. In Python you'd never have that, because there'd still be a reference to it, when creating the object.

But I'm fine with waiting for complaints :-)

include/pybind11/pybind11.h
YannickJadoul: Let me quickly check this again. I think you have a point! I checked something, but ...

YannickJadoul: Follow-up issue in #2595, in case we don't get to this anymore before the 2.6.0 release. I'm not sure how much time I'll have over the next few days, and you're right that it is not worth it blocking the release over this! At least it's better than before :-)

EricCousineau-TRI: Sounds good!

YannickJadoul: Thanks for catching these, btw; even if we don't need to fix anything, it might be nice to add a comment on why we can (currently) ignore them! :-)

include/pybind11/pybind11.h
YannickJadoul: Forgot about this. Let me still quickly fix this!

wjakob:
After taking another look, I am now convinced that the weak reference will work. That said, it seems far too heavy to create `cpp_function` for each type to set up a unique destruction callback. How about creating this callback method once in `internals` and then just initializing a single weak reference in `cpp_type`?

This would imply that the implementation has to be sufficiently generic -- it must e.g. map from the type to `std::type_index` and then figure stuff out from there.

wjakob:
Tangentially related, it seems that there is a function 

```inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_type_info_get_cache(PyTypeObject *type) {
```

that is not used anywhere in pybind11, and which does some suspiciously related things.

YannickJadoul:
@wjakob We do need access to the specific Python type object, though. And as far as I tried, you can't add arbitrary content to a `weakref`?

https://docs.python.org/3/library/weakref.html#weakref.ref says that
> If callback is provided and not None, and the returned weakref object is still alive, the callback will be called when the object is about to be finalized; the weak reference object will be passed as the only parameter to the callback; the referent will no longer be available.

But maybe [`PyWeakref_GET_OBJECT`](https://docs.python.org/3/c-api/weakref.html#c.PyWeakref_GET_OBJECT) would still work?

YannickJadoul:
> Tangentially related, it seems that there is a function
> 
> ```
> inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_type_info_get_cache(PyTypeObject *type) {
> ```
> 
> that is not used anywhere in pybind11, and which does some suspiciously related things.

That's where I got this approach, yes. I got reasonably confused for a while on whý things weren't getting cleaned up.

It is used here, though, I think:

```
inline const std::vector<detail::type_info *> &all_type_info(PyTypeObject *type) {
    auto ins = all_type_info_get_cache(type);
    if (ins.second)
        // New cache entry: populate it
        all_type_info_populate(type, ins.first->second);

    return ins.first->second;
}
 ```

wjakob:
> @wjakob We do need access to the specific Python type object, though. And as far as I tried, you can't add arbitrary attributes to a `weakref`?
> 
> https://docs.python.org/3/library/weakref.html#weakref.ref says that
> 
> > If callback is provided and not None, and the returned weakref object is still alive, the callback will be called when the object is about to be finalized; the weak reference object will be passed as the only parameter to the callback; the referent will no longer be available.

That is an excellent point, I had forgotten about that. Then how about catching things at a lower level: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_finalize, in this case in the metaclass used to construct pybind11 types?

YannickJadoul:
> That is an excellent point, I had forgotten about that. Then how about catching things at a lower level: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_finalize, in this case in the metaclass used to construct pybind11 types

That is the other option, yes. The detail here is that using a custom metaclass through `py::metaclass(...)` again will result in issues? And I preferred the non-invasive method of `weakref`, but maybe it's fine like this, for now?

YannickJadoul:
> it's fine like this, for now?

By which I mean: going with `tp_finalize`, and checking later how we can safely enable custom metaclasses?

YannickJadoul:
> But maybe [`PyWeakref_GET_OBJECT`](https://docs.python.org/3/c-api/weakref.html#c.PyWeakref_GET_OBJECT) would still work?

This won't: https://github.com/python/cpython/blob/cf693e537dc8aaa14315a7f59baec4a31d1167d3/Include/weakrefobject.h#L70-L80

YannickJadoul:
@wjakob, @rwgk, @EricCousineau-TRI, @henryiii, I moved the cleanup function to `tp_dealloc` in the metaclass.

Please have a look and compare this PR [with](https://github.com/pybind/pybind11/pull/2564/files/ae96b37866bdb7a2e7f314681707d8407690a639) and [without](https://github.com/pybind/pybind11/pull/2564/files/51978a845edbcc15d1758463de5d796848e31a09) this last commit ae96b37, to compare the two approaches.

YannickJadoul:
@rwgk I fixed the `detail::type_info *tinfo` leak you found, in both versions. I've done it in a slightly different way as originally suggested, but I think this is more elegant.

henryiii:
I take it the second approach breaks custom meta classes that don't also call this?

YannickJadoul:
> I take it the second approach breaks custom meta classes that don't also call this?

Yes. Then again, they'd only be as broken as the current situation (which has been there for a long time, if not forever, as far as I know?).

bstaletic:
I was going to ask if custom metaclasses ever worked in combination with pybind11.

YannickJadoul:
> I was going to ask if custom metaclasses ever worked in combination with pybind11.

Kind of (there is at least the `py::metaclass` tag), but anything that's defined in our pybind11-metaclass is of course not there by default (as far as I know, these are things like static properties, the extra check of `__init__` having been called, ...)

rwgk (approved):
This PR passed the Google-global testing.

henryiii:
> Then again, they'd only be as broken as the current situation

Okay, either approach is fine with me then.

wjakob:
Beautiful, thank you for the great work @YannickJadoul. I think that we can likely install a similar callback also for custom metaclasses, but this a much less important point that is not a blocker for v.2.6.0 (this kind issue revolving around  garbage collection of pybind11-created types is rare enough in practice that we are just bothered by it now, and custom metaclasses are a super-rarely used feature of pybind11). Pr{A ∩ B}≈ Pr{A}·Pr{B}=tiny number, I hope 🙂.

PR #2565, opened 2020-10-08
docs: known issues

### LINKED ISSUE #2563: [DOCS] Add known issues
I think we should add a known issues section to the docs somewhere. Ideas:

* It could be on the readme, just below compilers.
   - Gets baked into the documentation, so "stable" would list the known issues in stable. But would follow master elsewhere, like in the GitHub Page.
* It could be a new page in the docs.
    - Only in the documentation, basically
* It could be just above the changes on the changelog, or part of the changelog.
    - Could be kept with each version, but long term, might be a little harder to see what is current, if integrated into the changelog?
* It could go in one of the pages at the bottom.
    - A little less visible, but hopefully released known issues remain small.
    - I just noticed limitations, it might fit well there?

Other ideas welcome, too.

henryiii:
Adding known issues. Closes #2563 

Also updating the CMake FAQ.

The auto labeler is a little smarter and labels with more labels, sooner.

Updates the CHANGELOG with a a few more PRs (including the last one waiting for 2.6.0), and adds a bit more structure.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,28 @@
+needs changelog:
+  - all:
+    - '!docs/changelog.rst'
+    any:
+    - 'include/pybind11/**/*'
+    - 'pybind11/**/*'
+    - 'tools/**/*'
+    - 'CMakeLists.txt'
+
+docs:
+  - any:
+    - 'docs/basics.rst'
+    - 'docs/benchmark.rst'
+    - 'docs/classes.rst'
+    - 'docs/compiling.rst'
+    - 'docs/faq.rst'
+    - 'docs/index.rst'
+    - 'docs/installing.rst'
+    - 'docs/limitations.rst'
+    - 'docs/reference.rst'
+    - 'docs/release.rst'
+    - 'docs/advanced/**/*.rst'
+    - 'docs/cmake/**/*.rst'
+
+
+ci:
+  - any:
+    - '.github/workflows/*.yaml'
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -41,7 +41,7 @@ New features:
 
 * ``py::class_<union_type>`` is now supported. Note that writing to one data
   member of the union and reading another (type punning) is UB in C++. Thus
-  pybind11-bound enums should never be used for such conversion.
+  pybind11-bound enums should never be used for such conversions.
   `#2320 <https://github.com/pybind/pybind11/pull/2320>`_.
 
 * Classes now check local scope when registering members, allowing a subclass
@@ -133,9 +133,10 @@ Packaging / building improvements:
   `#2346 <https://github.com/pybind/pybind11/pull/2346>`_
 
 * ``pybind11_add_module()`` now accepts an optional ``OPT_SIZE`` flag that
-  switches the binding target to size-based optimization regardless global
-  CMake build type (except in debug mode, where optimizations remain disabled).
-  This reduces binary size quite substantially (~25%).
+  switches the binding target to size-based optimization if the global build
+  type can not always be fixed to ``MinSizeRel`` (except in debug mode, where
+  optimizations remain disabled).  ``MinSizeRel`` or this flag reduces binary
+  size quite substantially (~25% on some platforms).
   `#2463 <https://github.com/pybind/pybind11/pull/2463>`_
 
 Smaller or developer focused features:
@@ -143,8 +144,8 @@ Smaller or developer focused features:
 * Moved ``mkdoc.py`` to a new repo, `pybind11-mkdoc`_. There are no longer
   submodules in the main repo.
 
-* ``py::memoryview``  update, with ``py::memoryview::from_memory`` in Python 3,
-  and documentation.
+* ``py::memoryview`` segfault fix and update, with new
+  ``py::memoryview::from_memory`` in Python 3, and documentation.
   `#2223 <https://github.com/pybind/pybind11/pull/2223>`_
 
 * Fix for ``buffer_info`` on Python 2.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,16 +11,7 @@ v2.6.0 (IN PROGRESS)
 
 See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
 
-* ``py::module`` was renamed ``py::module_`` to avoid issues with C++20 when
-  used unqualified, but an alias ``py::module`` is provided for backward
-  compatibility.
-  `#2489 <https://github.com/pybind/pybind11/pull/2489>`_
-
-* ``pybind11_add_module()`` now accepts an optional ``OPT_SIZE`` flag that
-  switches the binding target to size-based optimization regardless global
-  CMake build type (except in debug mode, where optimizations remain disabled).
-  This reduces binary size quite substantially (~25%).
-  `#2463 <https://github.com/pybind/pybind11/pull/2463>`_
+New features:
 
 * Keyword-only arguments supported in Python 2 or 3 with ``py::kw_only()``.
   `#2100 <https://github.com/pybind/pybind11/pull/2100>`_
@@ -28,13 +19,16 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
 * Positional-only arguments supported in Python 2 or 3 with ``py::pos_only()``.
   `#2459 <https://github.com/pybind/pybind11/pull/2459>`_
 
-* Access to the type object now provided with ``py::type::of<T>()`` and
-  ``py::type::of(h)``.
-  `#2364 <https://github.com/pybind/pybind11/pull/2364>`_
+* ``py::is_final()`` class modifier to block subclassing (CPython only).
+  `#2151 <https://github.com/pybind/pybind11/pull/2151>`_
 
 * Added ``py::prepend()``, allowing a function to be placed at the beginning of
   the overload chain.
   `#1131 <https://github.com/pybind/pybind11/pull/1131>`_
+
+* Access to the type object now provided with ``py::type::of<T>()`` and
+  ``py::type::of(h)``.
+  `#2364 <https://github.com/pybind/pybind11/pull/2364>`_
 
 * Perfect forwarding support for methods.
   `#2048 <https://github.com/pybind/pybind11/pull/2048>`_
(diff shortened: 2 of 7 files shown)


docs/limitations.rst
```diff
@@ -14,6 +17,38 @@ certain limitations:
...
+^^^^^^^^^^^^^^^^^
+
+These are issues that are probably solvable, but have not been fixed yet. A
+clean, well written patch would likely be accepted to solve them.
+
+- Containers of ``char *`` are currently not supported.
+  `#2245 <https://github.com/pybind/pybind11/issues/2245>`_
+
+- Type casters are not kept alive recursively
+  `#2527 <https://github.com/pybind/pybind11/issues/2527>`_
```
YannickJadoul, line 54: This is basically the same issue.

YannickJadoul (approved):

docs/limitations.rst
henryiii: Should we merge them? Any suggestions on how we should merge them?

YannickJadoul: Something along the lines that "Containers of ``char *`` are currently not supported." is a consequence of "Type casters are not kept alive recursively"? But it's not really urgent though; maybe it's more clear to keep things separate, as well. I don't really mind.

docs/changelog.rst
```diff
@@ -45,19 +39,56 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
 * ``py::hash`` is now public.
   `#2217 <https://github.com/pybind/pybind11/pull/2217>`_
 
-* ``py::is_final()`` class modifier to block subclassing (CPython only).
-  `#2151 <https://github.com/pybind/pybind11/pull/2151>`_
+* ``py::class_<union_type>`` is now supported. Note that writing to one data
+  member of the union and reading another (type punning) is UB in C++. Thus
+  pybind11-bound enums should never be used for such conversion.
```
rwgk, line 44: such a conversion
or
such conversions
?

docs/changelog.rst
```diff
@@ -88,33 +120,35 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
...
+  `#2265 <https://github.com/pybind/pybind11/pull/2265>`_ and
+  `#2346 <https://github.com/pybind/pybind11/pull/2346>`_
 
-* ``PYBIND11_OVERLOAD*`` macros and ``get_overload`` function replaced by
-  correctly-named ``PYBIND11_OVERRIDE*`` and ``get_override``, fixing
-  inconsistencies in the presene of a closing ``;`` in these macros.
-  ``get_type_overload`` is deprecated.
-  `#2325 <https://github.com/pybind/pybind11/pull/2325>`_
+* ``pybind11_add_module()`` now accepts an optional ``OPT_SIZE`` flag that
+  switches the binding target to size-based optimization regardless global
```
rwgk, line 136: regardless of the global
?

docs/changelog.rst
```diff
@@ -88,33 +120,35 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
...
+  CMake build type (except in debug mode, where optimizations remain disabled).
+  This reduces binary size quite substantially (~25%).
```
rwgk, line 138: On all platforms, some platforms, certain platforms?

docs/changelog.rst
```diff
@@ -88,33 +120,35 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
...
-* ``py::class_<union_type>`` is now supported. Note that writing to one data
-  member of the union and reading another (type punning) is UB in C++. Thus
-  pybind11-bound enums should never be used for such conversion.
-  `#2320 <https://github.com/pybind/pybind11/pull/2320>`_.
+* Moved ``mkdoc.py`` to a new repo, `pybind11-mkdoc`_. There are no longer
+  submodules in the main repo.
 
-Smaller or developer focused features:
+* ``py::memoryview``  update, with ``py::memoryview::from_memory`` in Python 3,
+  and documentation.
```
rwgk, line 147: This was actually a bug fix, too: segfault because of incorrect use of std::string::c_str() past the lifetime of the string.

docs/faq.rst
```diff
@@ -303,10 +308,24 @@ In contrast this code:
...
-It is advised to avoid using ``find_package(PythonInterp)`` and ``find_package(PythonLibs)`` from CMake and rely
-on pybind11 in detecting Python version. If this is not possible CMake machinery should be called *before* including pybind11.
+will detect Python 3.x for pybind11 and may crash on
+``find_package(PythonLibs)`` afterwards.
+
+There are three possible solutions:
+
+1. Avoid using ``find_package(PythonInterp)`` and ``find_package(PythonLibs)``
+   from CMake and rely on pybind11 in detecting Python version. If this is not
+   possible CMake machinery should be called *before* including pybind11.
```
rwgk, line 318: possible the Cmake
?

docs/limitations.rst
```diff
@@ -14,6 +17,38 @@ certain limitations:
   class like ``Eigen::Array`` or ``boost.multi_array``.
 
 These features could be implemented but would lead to a significant increase in
-complexity. I've decided to draw the line here to keep this project simple and
-compact. Users who absolutely require these features are encouraged to fork
-pybind11.
+complexity. Pybind11 strives to be simple and compact. Users who absolutely
+require these features are encouraged to fork pybind11.
```
rwgk, line 21: Reading this hurts a little. Instead, I'd encourage them to suggest plug-in-kind-of mechanisms (like the "If you know how" just below), so they can layer their functionality on top. This must be possible somehow, without forking, in the worst case with `#ifdef`s, no?

docs/limitations.rst
```diff
@@ -14,6 +17,38 @@ certain limitations:
...
+unsolved. If you know how to help with one of these issues, contributions
+are welcome!
+
+- The test suite currently segfaults on macOS and Python 3.9.0 when exiting the
+  interpreter.  This was suspected to be related to the cross module GIL code,
+  but could be a broader Python 3.9.0 issue.
+  `#2558 <https://github.com/pybind/pybind11/issues/2558>`_
+
+- The ``cpptest`` does not run on Windows with Python 3.8 or newer, due to DLL
+  loader changes. Normally packages user code should not be affected.
```
rwgk, line 37: Normally packaged
?
Is that correct? I'd still be a bit wondering about the meaning of this sentence TBH? Could it be reworded?

docs/limitations.rst
rwgk: I'd combine these the way Yannick suggested, I think it's a lot clearer.

docs/upgrade.rst
```diff
@@ -37,9 +36,13 @@ If ``__eq__`` defined but not ``__hash__``, ``__hash__`` is now set to
 ``None``, as in normal CPython. You should add ``__hash__`` if you intended the
 class to be hashable, possibly using the new ``py::hash`` shortcut.
 
-Usage of the ``PYBIND11_OVERLOAD*`` macros and ``get_overload`` function should
-be replaced by ``PYBIND11_OVERRIDE*`` and ``get_override``. In the future, the
-old macros may be deprecated and removed.
+The constructors for ``py::array`` now always take signed integers for size,
+for consistency. Can cause some compiler warnings to appear. Cast to
```
rwgk, line 40: This may lead to compiler warning on some systems.
?

rwgk (approved):
Thanks for the awesome amount of great work! I have mostly very minor suggestions.

docs/changelog.rst
henryiii: Rewording this a bit, mentioning MinSizeRel (which is the correct, non-hack solution for most people, I think), and using "some platforms", since I don't think this works on absolutely every compiler and certainly was not tested everywhere.

docs/limitations.rst
henryiii: I've reworded this quite a bit now, I originally left it (almost) alone. Includes a pointer to an example now too.

docs/changelog.rst
henryiii: Authors are encouraged to help with the changelog. :)

Fixed (hopefully).

docs/upgrade.rst
henryiii: Tried to reword a bit. It's a bit tricky to say properly. For me, this is the biggest thing I notice when upgrading 2.5 -> 2.6, this starts spewing warnings. It's about 50/50 on how many `size_t` casts I have to remove and how many `py::ssize_t` casts I have to add, I think.

docs/limitations.rst
henryiii: Reworded. Thanks for these!

docs/changelog.rst
```diff
@@ -45,19 +39,56 @@ See :ref:`upgrade-guide-2.6` for help upgrading to the new version.
...
+  `#2489 <https://github.com/pybind/pybind11/pull/2489>`_
+
+* Public constructors for ``py::module_`` have been deprecated; please use
+  ``pybind11::module_::create_extension_module`` if you were using the public
+  constructor (fairly rare after ``PYBIND11_MODULE`` was introduced).
+  `#2552 <https://github.com/pybind/pybind11/pull/2552>`_
+
+* ``PYBIND11_OVERLOAD*`` macros and ``get_overload`` function replaced by
+  correctly-named ``PYBIND11_OVERRIDE*`` and ``get_override``, fixing
+  inconsistencies in the presene of a closing ``;`` in these macros.
```
YannickJadoul, line 79: presene -> presence

docs/faq.rst
```diff
@@ -275,17 +263,34 @@ been received, you must either explicitly interrupt execution by throwing
...
+version of Python and link against that. When this fails, or when there are
+multiple versions of Python and it finds the wrong one, delete
+``CMakeCache.txt`` and then add ``-DPYTHON_EXECUTABLE=$(which python)`` to your
+CMake configure line. (Replace ``$(which python)`` with a path to python if
+your prefer.)
+
+You can alternatively try ``-DPYBIND11_FINDPYTHON=ON``, which will activate the
+new CMake FindPython support instead of pybind11's custom search. Requires
+CMake 3.12+, and 3.15+ or 3.18.2+ are even better. You can set this in your
+``CMakeLists.txt`` before using pybind11, as well.
```
YannickJadoul, line 279: "before using" should maybe be before including the pybind11 subdirectory or installation?

docs/limitations.rst
```diff
@@ -11,9 +14,43 @@ certain limitations:
 
 - The NumPy interface ``pybind11::array`` greatly simplifies accessing
   numerical data from C++ (and vice versa), but it's not a full-blown array
-  class like ``Eigen::Array`` or ``boost.multi_array``.
+  class like ``Eigen::Array`` or ``boost.multi_array``. ``Eigen`` objects are
+  directly supported, however, with ``pybind11/eigen.h``.
+
+Large but useful features could be implemented in pybind11 but would lead to a
+significant increase in complexity. Pybind11 strives to be simple and compact.
```
YannickJadoul, line 21: Nitpicking: does "pybind11" take a capital at the start of a sentence?

YannickJadoul (approved):

docs/limitations.rst
rwgk: That's awesome! — There is another one like this in the pipeline, much closer to home, currently https://github.com/pybind/pybind11_abseil/tree/experimental. Just mentioning FYI, I don't think we should mention it here at the moment.

docs/limitations.rst
henryiii: Yes, at least it does in 

```
docs/advanced/cast/eigen.rst:binding functions that change their (Matrix) arguments.  Pybind11 allows you to
docs/advanced/cast/functional.rst:    case, there is no overhead. Pybind11 will extract the underlying C++
docs/advanced/classes.rst:type via a holder type argument to ``class_``. Pybind11 provides a helper class
docs/advanced/functions.rst:Pybind11 provides a ``py::kw_only`` object that allows you to implement
docs/cmake/index.rst:Pybind11 can be used with ``add_subdirectory(extern/pybind11)``, or from an
docs/compiling.rst:    from pybind11.setup_helpers import Pybind11Extension
docs/compiling.rst:Pybind11 supports modern CMake usage patterns with a set of interface targets,
docs/faq.rst:   3.18.2+ best). Pybind11 in these cases uses the new CMake FindPython instead
docs/faq.rst:3. Set ``PYBIND11_NOPYTHON`` to ``TRUE``. Pybind11 will not search for Python.
docs/limitations.rst:significant increase in complexity. Pybind11 strives to be simple and compact.
tools/pybind11NewTools.cmake:      "${_Python} was found without the Interpreter component. Pybind11 requires this component.")
```

I didn't write all of those (though I did write some of them).

.github/labeler.yml
```diff
@@ -0,0 +1,28 @@
...
+    - 'docs/classes.rst'
+    - 'docs/compiling.rst'
+    - 'docs/faq.rst'
+    - 'docs/index.rst'
+    - 'docs/installing.rst'
+    - 'docs/limitations.rst'
+    - 'docs/reference.rst'
+    - 'docs/release.rst'
+    - 'docs/advanced/**/*.rst'
+    - 'docs/cmake/**/*.rst'
```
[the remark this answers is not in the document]
henryiii: No, because then it triggers on `changelog.rst` and `upgrade.rst`. This doesn't change all that often, so assumed it was safe.

docs/limitations.rst
YannickJadoul: OK, fine, then

.github/labeler.yml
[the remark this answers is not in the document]
YannickJadoul: Ah, yes. And there's not a way to exclude those (like `- '!docs/changelog.rst'` above), rather than include these?

docs/limitations.rst
henryiii: I noticed that, and figured you'd know something about it. ;)

.github/labeler.yml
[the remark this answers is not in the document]
henryiii: We can try that later, but I didn't quickly find a way.

PR #2566, opened 2020-10-09
Update PGI to 20.9

andriish:
Update PGI to 20.9

Tag @henryiii

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,7 +258,7 @@ jobs:
   # Testing CentOS 8 + PGI compilers
   centos-nvhpc8:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


henryiii:
The windows error is a known flake. Ignore. :)

henryiii:
The readme number can be bumped to 20.9, as well, if you'd like.

henryiii:
How about we leave one test (CentOS 8) with 20.7, and only bump the other? That way, the readme is truthful in saying we check both.

henryiii:
(I can make that change if you want)

andriish:
Hi @henryiii ,

Yes, why not.
> I can make that change if you want) 

Yes, please, I have to leave now for some hours.

Best regards,
Andrii

PR #2569, opened 2020-10-09
Do not use abi::__forced_unwind with libc++, even with gcc instead of clang

### LINKED ISSUE #2568: __forced_unwind causing failure for systems using LLVM libc++ and not using clang and not using stdlibc++
https://github.com/pybind/pybind11/blob/e4e5c49a26dbf478ff93e3d6aacd747f19627c76/include/pybind11/pybind11.h#L744-L748

asobhy-qnx:
Fixes #2568

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -741,7 +741,7 @@ protected:
         } catch (error_already_set &e) {
             e.restore();
             return nullptr;
-#if defined(__GNUG__) && !defined(__clang__)
+#if defined(__GNUG__) && !defined(__clang__) && !defined(_LIBCPP_VERSION)
         } catch ( abi::__forced_unwind& ) {
             throw;
 #endif
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -741,7 +741,7 @@ protected:
         } catch (error_already_set &e) {
             e.restore();
             return nullptr;
-#if defined(__GNUG__) && !defined(__clang__) && !defined(_LIBCPP_VERSION)
+#if defined(__GNUG__) && !defined(_LIBCPP_VERSION)
         } catch ( abi::__forced_unwind& ) {
             throw;
 #endif


include/pybind11/pybind11.h
```diff
@@ -741,7 +741,7 @@ class cpp_function : public function {
         } catch (error_already_set &e) {
             e.restore();
             return nullptr;
-#if defined(__GNUG__) && !defined(__clang__)
+#if defined(__GNUG__) && !defined(__clang__) && !defined(_LIBCPP_VERSION)
```
YannickJadoul, line 744: If this is purely `libc++` vs `libstdc++` Shouldn't this then just replace the line? I.e., be fully `#if !defined(_LIBCPP_VERSION)`? Or actually, the other way around: `#if defined(__GLIBCXX__)`.

YannickJadoul (commented):
I'm going to ping @henryiii on this, though.

include/pybind11/pybind11.h
asobhy-qnx: yes this would work as well

henryiii: We already test quite heavlily for LLVM, both on linux and macOS, so I assume the problem fixed here is we don't support LLVM + libstdc++?

Agree with @YannickJadoul , it looks simpler to just check for `__GLIBCXX__`.

henryiii:
```suggestion
#ifdef __GLIBCXX__
```

asobhy-qnx: clang is a compiler wrapper for C language family. libc++ is the library used by clang for c++. GNU compilers like GCC are more flexible and can choose whether to use libc++ or libstdc++ using the compiler option "-stdlib=". The choice of library usually depends on the system being developed for and for QNX OS libc++ is the default C++ library. On linus the native library is libstdc++ that does not mean that you cannot link agains libc++ on linux if someone wants to.

The problem here is that I am trying to Cross compile for QNX which uses gcc and libc++. In the code above when running the g++ compiler with the option -stdlib=libc++ __GNUG__ is defined but __clang__ is not thus the compiler tries to compile __forced_unwind which does not exist in libc++ and causes the build to fail.

From my understanding the definition __GLIBCXX__ is no longer being used. I tested and it did not work.

asobhy-qnx: Here is a link to a discussion about detecting LLVM libc++ https://stackoverflow.com/questions/31657499/how-to-detect-stdlib-libc-in-the-preprocessor

henryiii: Looks like that is not always reliably defined before gcc 6.1, using the link posted in the issue https://stackoverflow.com/a/31658120/2402816? Maybe `#ifndef _LIBCPP_VERSION` is safer?

asobhy-qnx: I tested _LIBCPP_VERSION and it worked

henryiii: > GNU compilers like GCC are more flexible

PS: Clang can compile with libstdc++ too. In fact, that was the default some time ago; even on macOS it didn't change till 10.9. I would say Clang by design is more flexible than GCC. This is why Cling, clang-* tools, etc. all exist.

henryiii:
```suggestion
#ifndef _LIBCPP_VERSION
```
?

henryiii: Though we'd need to make sure that's not triggering on msvc, too, I assume?

henryiii:
```suggestion
#if defined(__GNUG__) && !defined(_LIBCPP_VERSION)
```

This will not trigger on MSVC, but will trigger on GCC or Clang when using libstdc++. Would that work?

asobhy-qnx: > > GNU compilers like GCC are more flexible
> 
> PS: Clang can compile with libstdc++ too. In fact, that was the default some time ago; even on macOS it didn't change till 10.9. I would say Clang by design is more flexible than GCC. This is why Cling, clang-* tools, etc. all exist.

@henryiii thank you for this clarification.

asobhy-qnx:
Can we also push this commit to 2.5.0?

henryiii:
2.5.0 is a tag. How do you push to an existing version that been released for months? We can try to get it into 2.6.0rc1, though, which I'm planning to release today.

henryiii (approved):

asobhy-qnx:
That would be good

include/pybind11/pybind11.h
YannickJadoul: Why are we still checking the compiler, though? We should just check the standard lib?

Oh, or wait, is `__GNUG__` defined both in libstdc++ as well as libc++? Independently of the compiler?
So `defined(__GNUG__)` selects all combinations of `libstdc++`/`libc++` and `gcc`/`clang`, and `!defined(_LIBCPP_VERSION)` deselects `libc++`?

YannickJadoul: Still, isn't there a "positive" way of checking for libstdc++?

henryiii: `__GNUG__` is defined for both gcc and clang. Possibly other "gcc-like" compilers. Otherwise, MSVC will also be not defining `_LIBCPP_VERSION`.

> Still, isn't there a "positive" way of checking for libstdc++

This would be ideal; not sure, the obvious one seems to not have been added until 6.1?

henryiii: As far as I can tell, `__GLIBCXX__` didn't go in the header listed until 6.1, but it still should be defined in other headers, I think.

henryiii: Yes, as far as I can tell, any public header (except the "do-nothing-but-define" header linked, should include this macro. @asobhy-qnx, what happened when you tried using `__GLIBCXX__`? Unless we are unlucky and libc++ _also_ defined it, that seems like it should work.

henryiii:
Let's merge this, then immediately try a PR with `__GLIBCXX__` which @asobhy-qnx can test, just to make sure, and we can see how our CI behaves.

PR #2572, opened 2020-10-09
refactor: followup to #2569

henryiii:
Followup to #2569

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -741,7 +741,7 @@ protected:
         } catch (error_already_set &e) {
             e.restore();
             return nullptr;
-#if defined(__GNUG__) && !defined(_LIBCPP_VERSION)
+#ifdef __GLIBCXX__
         } catch ( abi::__forced_unwind& ) {
             throw;
 #endif


YannickJadoul:
Would you mind giving this a test, @asobhy-qnx?

asobhy-qnx:
I do not mind testing it again. I did already test it before creating the first pull request. double checking does not hurt. I will let you know the results.

henryiii:
Fantastic, @asobhy-qnx, thank you!

YannickJadoul:
Thanks! :-)

YannickJadoul:
@henryiii, one of these tests is running GCC 4.8, right?

Ah, but we don't actually test if that works, ofc :-/ We don't have a test that the right branch actually compiles for libstdc++?

henryiii:
We don't do any cross-compiling, that is, no GCC with libc++ or Clang with libstdc++ (I'm slightly misusing the term there), so we don't know if it gets detected correctly - and yes, we don't know if this branch gets ignored, unless there's a test throwing this (not sure what triggers it?).

@asobhy-qnx You are compiling from GCC with libc++, right? So the only way this could "fail" for you is if the branch is enabled; that is, if `__GLIBCXX__` is defined, I would assume by GCC itself.

asobhy-qnx:
I thought __GLIBCXX__ == libc++ I was completely wrong __GLIBCXX__ == libstdc++ and this is why it was not working for me.

It did work and I was able to build the examples.

henryiii:
That's why double checking is useful. :) Merging then!

YannickJadoul:
Ah, great! Thanks, @asobhy-qnx :-)

asobhy-qnx:
test for __GBLICXX__ done and was successful

YannickJadoul:
Thanks for checking, @asobhy-qnx!

PR #2573, opened 2020-10-10
CI: Intel icc/icpc via oneAPI

ax3l:
Add testing for Intel icc/icpc via the oneAPI images.
Intel oneAPI is in a late beta stage, currently shipping oneAPI beta09 with ICC 20.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,14 +409,14 @@ jobs:
         echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python cmake
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip libeigen3-dev
 
     - name: Update pip & pytest
       shell: bash
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        python -m pip install --upgrade pip
-        python -m pip install --upgrade pytest
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade pytest
 
     - name: Configure
       shell: bash
@@ -428,7 +428,7 @@ jobs:
         -DDOWNLOAD_EIGEN=OFF    \
         -DCMAKE_CXX_STANDARD=11             \
         -DCMAKE_CXX_COMPILER=$(which icpc)  \
-        -DPYTHON_EXECUTABLE=$(python -c "import sys; print(sys.executable)")
+        -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
 
     - name: Build
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,21 +396,20 @@ jobs:
       fail-fast: false
 
     name: "🐍 3 • ICC latest • x64"
-    container: "gcc"
 
     steps:
     - uses: actions/checkout@v1
 
     - name: Add apt repo
       run: |
-        apt-get update
-        apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg
+        sudo apt-get update
+        sudo apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg
         wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
-        apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
-        echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
+        sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python libc6-dev
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python
 
     - name: Update pip & pytest
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -430,7 +430,7 @@ jobs:
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         cmake -S . -B build     \
         -DPYBIND11_WERROR=ON    \
-        -DDOWNLOAD_CATCH=ON     \
+        -DDOWNLOAD_CATCH=OFF    \
         -DDOWNLOAD_EIGEN=ON     \
         -DCMAKE_CXX_STANDARD=11             \
         -DCMAKE_CXX_COMPILER=$(which icpc)  \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -442,8 +442,7 @@ jobs:
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         sudo service apport stop
-        cmake --build build --target pytest; ls; ls /home/runner/work/pybind11/pybind11/build/tests
-        exit 1
+        cmake --build build --target check
 
     - name: C++ tests
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -429,8 +429,8 @@ jobs:
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         cmake -S . -B build     \
         -DPYBIND11_WERROR=ON    \
-        -DDOWNLOAD_CATCH=OFF    \
-        -DDOWNLOAD_EIGEN=ON     \
+        -DDOWNLOAD_CATCH=ON     \
+        -DDOWNLOAD_EIGEN=OFF    \
         -DCMAKE_CXX_STANDARD=11             \
         -DCMAKE_CXX_COMPILER=$(which icpc)  \
         -DPYTHON_EXECUTABLE=$(python -c "import sys; print(sys.executable)")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -334,6 +334,7 @@ jobs:
     - name: Interface test
       run: cmake3 --build build --target test_cmake_build
 
+
   # Testing on GCC using the GCC docker images (only recent images supported)
   gcc:
     runs-on: ubuntu-latest
@@ -386,6 +387,78 @@ jobs:
 
     - name: Interface test
       run: cmake --build build --target test_cmake_build
+
+
+  # Testing on ICC using the oneAPI apt repo
+  icc:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    name: "🐍 3 • ICC latest • x64"
+    container: "gcc"
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Add apt repo
+      run: |
+        apt-get update
+        apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg
+        wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
+
+    - name: Add ICC & Python 3
+      run: apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python
+
+    - name: Update pip & pytest
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        python -m pip install --upgrade pip
+        python -m pip install --upgrade pytest
+
+    - name: Setup CMake 3.18
+      uses: jwlawson/actions-setup-cmake@v1.4
+      with:
+        cmake-version: 3.18
+
+    - name: Configure
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake -S . -B build     \
+        -DPYBIND11_WERROR=ON    \
+        -DDOWNLOAD_CATCH=ON     \
+        -DDOWNLOAD_EIGEN=ON     \
+        -DCMAKE_CXX_STANDARD=11             \
+        -DCMAKE_CXX_COMPILER=$(which icpc)  \
+        -DPYTHON_EXECUTABLE=$(python -c "import sys; print(sys.executable)")
+
+    - name: Build
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build -j 2
+
+    - name: Python tests
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build --target pytest
+
+    - name: C++ tests
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build --target cpptest
+
+    - name: Interface test
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build --target test_cmake_build
 
 
   # Testing on CentOS (manylinux uses a centos base, and this is an easy way
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,7 +409,7 @@ jobs:
         echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python cmake
 
     - name: Update pip & pytest
       shell: bash
@@ -417,11 +417,6 @@ jobs:
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python -m pip install --upgrade pip
         python -m pip install --upgrade pytest
-
-    - name: Setup CMake 3.18
-      uses: jwlawson/actions-setup-cmake@v1.4
-      with:
-        cmake-version: 3.18
 
     - name: Configure
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,7 +409,7 @@ jobs:
         echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic cmake python3-dev python3-numpy python3-pytest python3-pip
 
     - name: Update pip & pytest
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -441,7 +441,9 @@ jobs:
       shell: bash
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        cmake --build build --target pytest
+        sudo service apport stop
+        cmake --build build --target pytest; ls; ls /home/runner/work/pybind11/pybind11/build/tests
+        exit 1
 
     - name: C++ tests
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,7 +409,7 @@ jobs:
         echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip libeigen3-dev
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip
 
     - name: Update pip & pytest
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -391,7 +391,7 @@ jobs:
 
   # Testing on ICC using the oneAPI apt repo
   icc:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -411,12 +411,15 @@ jobs:
     - name: Add ICC & Python 3
       run: sudo apt-get update; sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic cmake python3-dev python3-numpy python3-pytest python3-pip
 
-    - name: Update pip & pytest
+    - name: Update pip
       shell: bash
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python3 -m pip install --upgrade pip
-        python3 -m pip install --upgrade pytest==5.4.3
+
+    - name: Install dependencies
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        python3 -m pip install -r tests/requirements.txt --prefer-binary
 
     - name: Configure
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -391,7 +391,7 @@ jobs:
 
   # Testing on ICC using the oneAPI apt repo
   icc:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -428,6 +428,7 @@ jobs:
         -DDOWNLOAD_EIGEN=OFF    \
         -DCMAKE_CXX_STANDARD=11             \
         -DCMAKE_CXX_COMPILER=$(which icpc)  \
+        -DCMAKE_VERBOSE_MAKEFILE=ON         \
         -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
 
     - name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -418,6 +418,7 @@ jobs:
         python3 -m pip install --upgrade pip
 
     - name: Install dependencies
+      run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python3 -m pip install -r tests/requirements.txt --prefer-binary
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -410,7 +410,7 @@ jobs:
         echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
 
     - name: Add ICC & Python 3
-      run: apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python
+      run: apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-python libc6-dev
 
     - name: Update pip & pytest
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -416,7 +416,7 @@ jobs:
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python3 -m pip install --upgrade pip
-        python3 -m pip install --upgrade pytest
+        python3 -m pip install --upgrade pytest==5.4.3
 
     - name: Configure
       shell: bash
--- a/tests/test_callbacks.cpp
+++ b/tests/test_callbacks.cpp
@@ -119,8 +119,8 @@ TEST_SUBMODULE(callbacks, m) {
 
     class AbstractBase {
     public:
-        // ICC can't handle this being defaulted
-        virtual ~AbstractBase() {}; // NOLINT
+        // [workaround(intel)] = default does not work here
+        virtual ~AbstractBase() {};  // NOLINT(modernize-use-equals-default)
         virtual unsigned int func() = 0;
     };
     m.def("func_accepting_func_accepting_base", [](std::function<double(AbstractBase&)>) { });
--- a/tests/test_callbacks.cpp
+++ b/tests/test_callbacks.cpp
@@ -119,7 +119,7 @@ TEST_SUBMODULE(callbacks, m) {
 
     class AbstractBase {
     public:
-        virtual ~AbstractBase() = default;
+        virtual ~AbstractBase() {};
         virtual unsigned int func() = 0;
     };
     m.def("func_accepting_func_accepting_base", [](std::function<double(AbstractBase&)>) { });
--- a/tests/test_callbacks.cpp
+++ b/tests/test_callbacks.cpp
@@ -119,7 +119,8 @@ TEST_SUBMODULE(callbacks, m) {
 
     class AbstractBase {
     public:
-        virtual ~AbstractBase() {};
+        // ICC can't handle this being defaulted
+        virtual ~AbstractBase() {}; // NOLINT
         virtual unsigned int func() = 0;
     };
     m.def("func_accepting_func_accepting_base", [](std::function<double(AbstractBase&)>) { });
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -323,6 +323,7 @@ TEST_SUBMODULE(class_, m) {
 
     class PublicistB : public ProtectedB {
     public:
+    ~PublicistB() override {};
         using ProtectedB::foo;
     };
 
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -323,7 +323,8 @@ TEST_SUBMODULE(class_, m) {
 
     class PublicistB : public ProtectedB {
     public:
-        ~PublicistB() override {};  // NOLINT(modernize-use-equals-default) breaks ICPC
+        // [workaround(intel)] = default does not work here
+        ~PublicistB() override {};  // NOLINT(modernize-use-equals-default)
         using ProtectedB::foo;
     };
 
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -323,7 +323,7 @@ TEST_SUBMODULE(class_, m) {
 
     class PublicistB : public ProtectedB {
     public:
-    ~PublicistB() override {};
+        ~PublicistB() override {};  // NOLINT(modernize-use-equals-default) breaks ICPC
         using ProtectedB::foo;
     };
 
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -185,7 +185,7 @@ TEST_SUBMODULE(builtin_casters, m) {
 
     // test_bool_caster
     m.def("bool_passthrough", [](bool arg) { return arg; });
-    m.def("bool_passthrough_noconvert", [](bool arg) { return arg; }, py::arg().noconvert());
+    m.def("bool_passthrough_noconvert", [](bool arg) { return arg; }, py::arg{}.noconvert());
 
     // test_reference_wrapper
     m.def("refwrap_builtin", [](std::reference_wrapper<int> p) { return 10 * p.get(); });
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -183,6 +183,8 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("load_nullptr_t", [](std::nullptr_t) {}); // not useful, but it should still compile
     m.def("cast_nullptr_t", []() { return std::nullptr_t{}; });
 
+    // [workaround(intel)] ICC 20/21 breaks with py::arg().stuff, using py::arg{}.stuff works.
+
     // test_bool_caster
     m.def("bool_passthrough", [](bool arg) { return arg; });
     m.def("bool_passthrough_noconvert", [](bool arg) { return arg; }, py::arg{}.noconvert());
--- a/tests/test_custom_type_casters.cpp
+++ b/tests/test_custom_type_casters.cpp
@@ -101,16 +101,16 @@ TEST_SUBMODULE(custom_type_casters, m) {
     };
     py::class_<ArgInspector>(m, "ArgInspector")
         .def(py::init<>())
-        .def("f", &ArgInspector::f, py::arg{}, py::arg{} = ArgAlwaysConverts())
-        .def("g", &ArgInspector::g, "a"_a.noconvert(), "b"_a, "c"_a.noconvert()=13, "d"_a=ArgInspector2(), py::arg{} = ArgAlwaysConverts())
-        .def_static("h", &ArgInspector::h, py::arg{}.noconvert(), py::arg{} = ArgAlwaysConverts())
+        .def("f", &ArgInspector::f, py::arg(), py::arg() = ArgAlwaysConverts())
+        .def("g", &ArgInspector::g, "a"_a.noconvert(), "b"_a, "c"_a.noconvert()=13, "d"_a=ArgInspector2(), py::arg() = ArgAlwaysConverts())
+        .def_static("h", &ArgInspector::h, py::arg{}.noconvert(), py::arg() = ArgAlwaysConverts())
         ;
     m.def("arg_inspect_func", [](ArgInspector2 a, ArgInspector1 b, ArgAlwaysConverts) { return a.arg + "\n" + b.arg; },
-            py::arg{}.noconvert(false), py::arg_v{nullptr, ArgInspector1()}.noconvert(true), py::arg{} = ArgAlwaysConverts());
+            py::arg{}.noconvert(false), py::arg_v(nullptr, ArgInspector1()).noconvert(true), py::arg() = ArgAlwaysConverts());
 
-    m.def("floats_preferred", [](double f) { return 0.5 * f; }, py::arg{"f"});
+    m.def("floats_preferred", [](double f) { return 0.5 * f; }, py::arg("f"));
     m.def("floats_only", [](double f) { return 0.5 * f; }, py::arg{"f"}.noconvert());
-    m.def("ints_preferred", [](int i) { return i / 2; }, py::arg{"i"});
+    m.def("ints_preferred", [](int i) { return i / 2; }, py::arg("i"));
     m.def("ints_only", [](int i) { return i / 2; }, py::arg{"i"}.noconvert());
 
     // test_custom_caster_destruction
--- a/tests/test_custom_type_casters.cpp
+++ b/tests/test_custom_type_casters.cpp
@@ -101,17 +101,17 @@ TEST_SUBMODULE(custom_type_casters, m) {
     };
     py::class_<ArgInspector>(m, "ArgInspector")
         .def(py::init<>())
-        .def("f", &ArgInspector::f, py::arg(), py::arg() = ArgAlwaysConverts())
-        .def("g", &ArgInspector::g, "a"_a.noconvert(), "b"_a, "c"_a.noconvert()=13, "d"_a=ArgInspector2(), py::arg() = ArgAlwaysConverts())
-        .def_static("h", &ArgInspector::h, py::arg().noconvert(), py::arg() = ArgAlwaysConverts())
+        .def("f", &ArgInspector::f, py::arg{}, py::arg{} = ArgAlwaysConverts())
+        .def("g", &ArgInspector::g, "a"_a.noconvert(), "b"_a, "c"_a.noconvert()=13, "d"_a=ArgInspector2(), py::arg{} = ArgAlwaysConverts())
+        .def_static("h", &ArgInspector::h, py::arg{}.noconvert(), py::arg{} = ArgAlwaysConverts())
         ;
     m.def("arg_inspect_func", [](ArgInspector2 a, ArgInspector1 b, ArgAlwaysConverts) { return a.arg + "\n" + b.arg; },
-            py::arg().noconvert(false), py::arg_v(nullptr, ArgInspector1()).noconvert(true), py::arg() = ArgAlwaysConverts());
+            py::arg{}.noconvert(false), py::arg_v{nullptr, ArgInspector1()}.noconvert(true), py::arg{} = ArgAlwaysConverts());
 
-    m.def("floats_preferred", [](double f) { return 0.5 * f; }, py::arg("f"));
-    m.def("floats_only", [](double f) { return 0.5 * f; }, py::arg("f").noconvert());
-    m.def("ints_preferred", [](int i) { return i / 2; }, py::arg("i"));
-    m.def("ints_only", [](int i) { return i / 2; }, py::arg("i").noconvert());
+    m.def("floats_preferred", [](double f) { return 0.5 * f; }, py::arg{"f"});
+    m.def("floats_only", [](double f) { return 0.5 * f; }, py::arg{"f"}.noconvert());
+    m.def("ints_preferred", [](int i) { return i / 2; }, py::arg{"i"});
+    m.def("ints_only", [](int i) { return i / 2; }, py::arg{"i"}.noconvert());
 
     // test_custom_caster_destruction
     // Test that `take_ownership` works on types with a custom type caster when given a pointer
--- a/tests/test_custom_type_casters.cpp
+++ b/tests/test_custom_type_casters.cpp
@@ -99,6 +99,7 @@ TEST_SUBMODULE(custom_type_casters, m) {
         }
         static ArgInspector2 h(ArgInspector2 a, ArgAlwaysConverts) { return a; }
     };
+    // [workaround(intel)] ICC 20/21 breaks with py::arg().stuff, using py::arg{}.stuff works.
     py::class_<ArgInspector>(m, "ArgInspector")
         .def(py::init<>())
         .def("f", &ArgInspector::f, py::arg(), py::arg() = ArgAlwaysConverts())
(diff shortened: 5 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -392,6 +393,78 @@ jobs:
...
+    - name: Add apt repo
+      run: |
+        apt-get update
+        apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg
+        wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
+
+    - name: Add ICC & Python 3
+      run: apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro intel-oneapi-mkl-devel intel-oneapi-python
```
ax3l, line 417: mkl: let's see if this fixes
```
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - not found
CMake Error at /__t/cmake/3.18.4/x64/cmake-3.18.4-Linux-x86_64/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find Threads (missing: Threads_FOUND)
Call Stack (most recent call first):
  /__t/cmake/3.18.4/x64/cmake-3.18.4-Linux-x86_64/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  /__t/cmake/3.18.4/x64/cmake-3.18.4-Linux-x86_64/share/cmake-3.18/Modules/FindThreads.cmake:234 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  tests/test_embed/CMakeLists.txt:17 (find_package)
```

henryiii:
Have you tried running the same docker commands locally?

ax3l:
Yep, can reproduce on `ubuntu:18.04`.

henryiii:
Storing a recipe for docker for my use:

``` bash
apt-get update && apt install -y wget build-essential pkg-config cmake ca-certificates gnupg && wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB && apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB && echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list && apt-get update && apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip

```

henryiii:
I am getting a few remarks:
```
remark #11074: Inlining inhibited by limit max-size
remark #11076: To get full report use -qopt-report=4 -qopt-report-phase ipo
...
remark #11074: Inlining inhibited by limit max-size
remark #11074: Inlining inhibited by limit max-total-size
remark #11076: To get full report use -qopt-report=4 -qopt-report-phase ipo
```
And then I do get the segfault. (this is testing on the fix/intel branch, and using Ubuntu 18 docker)

wjakob:
Not having a virtual destructor in `ProtectedA` is fine: there is no polymorphism, only 1 POD member. TBH I wouldn't be surprised if this is just a miscompilation.

@ax3l : When you make the change to avoid linker errors, is the behavior identical in Debug/Release mode?

ax3l:
Interestingly, I don't get any linker errors in release builds, but the same segfault.

Inlining limits: yes, I saw them as well... We can increase the limit and check if that mitigates it, but that would be curious.

Maybe the problem is stemming from IPO?

tobiasleibner:
This was mentioned in #2679, so I had a look at the segfault. Apparently, it is a specific problem with the ``noconvert`` and ``none`` constructs. Calling
```c++
sm.def("accept_double_noconvert",
            [](py::array_t<double, 0>) {},
            py::arg("a").noconvert());
```
segfaults. When replacing this by
```c++
auto py_arg = py::arg("a");
sm.def("accept_double_noconvert",
       [](py::array_t<double, 0>) {},
       py_arg.noconvert());
```
the segfault is gone. Same for ``py::arg().none()``. After commenting out all occurrences of ``py::arg().none()`` and ``py::arg().noconvert()`` in the tests,
```
python3 -c "import pybind11_tests"
```
does not segfault anymore and only gives the output
```
/opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.factory_constructors.TestFactory3' is using an old-style placement-new '__init__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
  return f(*args, **kwds)
/opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.factory_constructors.NoisyAlloc' is using an old-style placement-new '__init__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
  return f(*args, **kwds)
/opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.pickling.Pickleable' is using an old-style placement-new '__setstate__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
  return f(*args, **kwds)
/opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.pickling.PickleableWithDict' is using an old-style placement-new '__setstate__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
  return f(*args, **kwds)
```
The ``py::arg().noconvert()`` construct seems dangerous, ``py::arg()`` is creating a temporary object, and the ``noconvert()`` call returns a mutable reference to this temporary. I don't know what the ``def`` function does with that reference, but I guess the Intel compiler discards the temporary before it is accessed.

henryiii:
Given that this is a common idiom, shouldn't we just make copies with `.noconvert()` and relatives? Returning a mutable reference seems scary and undefined to me, and there's no real performance issue to worry about with these. @bstaletic, @YannickJadoul?

henryiii:
Why does `py::arg{}.noconvert()` work but `py::arg().noconvert()` not? That's odd. Maybe that's a good "patch" solution and a change could be made in 2.7? If this mutates and returns a copy of the mutated object, I think 99% of possible workflows would be covered, however.

bstaletic:
That was my first idea as well. There's just one catch. `py::arg::operator=` returns `py::arg_v`, which inherits from `py::arg`. In case a user already has an `arg_v` object on which the user calls `noconvert()` , returning by value would slice the object. However, idiomatic uses aren't affected.

`py::arg("a").noconvert()` does return a reference to a temporary, but it is fine in idiomatic use case as well, because the temporary object is supposed to live until the end of the full expression. I do agree that this might be walking on thin ice.

bstaletic:
> Why does py::arg{}.noconvert() work but py::arg().noconvert() not?

That's a very good question. Both should call `noconvert()` on a default constructed `py::arg`...

YannickJadoul:
I'm a bit hesitant to make big changes for a misbehaving compiler, though. Can't we use the `{}`-instead-of-`()` workaround in our tests, while keeping the current design?

> In case a user already has an `arg_v` object on which the user calls `noconvert()` , returning by value would slice the object. However, idiomatic uses aren't affected.

You could duplicate the same method on `arg_v`, though? Or follow some fancy CTRP-pattern (similar to `object_api<T>`) ?

Good catch, though, @tobiasleibner! Kudos for pinning down this issue! :-)

henryiii:
> a misbehaving compiler

But does the standard promise that the temporary is kept around till the end of the line, or is that just a compiler choice we happen to be depending on?

> Can't we use the {}-instead-of-() workaround in our tests

I've tried replacing `arg(...).<...>` with `arg{...}.<...>`, but it still seems to segfault. Does `"x".a.<...>` need to be changed too?  I think there were 1-3 of those. Also, users who want to use Intel will also need to use this workaround, or not use the idiom we prescribe when writing their code. Logically fixing it would cause user code to work, too, without depending on a compiler-specific quirk that is not part of the standard (unless the answer is yes above).

YannickJadoul:
> But does the standard promise that the temporary is kept around till the end of the line, or is that just a compiler choice we happen to be depending on?

Hmmm, interesting. This might [only be for const lvalue references](https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime)? In that case, I fully agree we should fix it (and yes, no worries about overhead; the compiler ought to be able to optimize that, anyway?).

henryiii:
Const references are guarantied, yes. But not mutable ones AFAIK. We _already_ duplicate noconvert and none in arg_v, to make them return arg_v instead of arg, so it seems safe to fix.

YannickJadoul:
Wouldn't this fall under [binding a reference to a prvalue](https://en.cppreference.com/w/cpp/language/reference_initialization), though? But @bstaletic tends to be better at understanding the standard than I am.

bstaletic:
> But does the standard promise that the temporary is kept around till the end of the line

[A temporary object is materialized when binding a reference to an r-value](https://eel.is/c++draft/class.temporary#2.1).

[The lifetime of a temporary object, bound to a reference in a function call lasts until the end of a "full-expression"](https://eel.is/c++draft/class.temporary#6.9).

[A full-expression is](https://eel.is/c++draft/intro.execution#def:full-expression), when you eliminate the cases that don't apply to what we're discussing [any expression that is not a subexpression](https://eel.is/c++draft/intro.execution#5.6). I.e. until the semicolon in most cases.

This means that things like `f(T().ref());` should be perfectly fine. The whole thing is a single full expression, so `T()` lives until the semicolon.

henryiii:
Excellent! So this is a compiler bug. (@ax3l could you report it to Intel, perhaps?). If we can work around it for the tests using a weird quirk like `arg{}` behaving differently than `arg()`, that's fine with me. @tobiasleibner did you have to change the `"..."_a` lines too? (or do you see anything I should change in my commit that would cause this to work?)

We could also change a few of these to pre-define, then have test or two that runs on everything but Intel (mayne even just this version) with it written the "normal" way.

YannickJadoul:
(It's a much more tricky construction, though, so if there's a way to fix this without changing the standard usage, that's also fine, ofc ;-) )

tobiasleibner (commented):
After fixing the two remaining occurrences (and the linker errors in debug mode), this branch works for me :+1:

tobiasleibner:
> Excellent! So this is a compiler bug. (@ax3l could you report it to Intel, perhaps?). If we can work around it for the tests using a weird quirk like `arg{}` behaving differently than `arg()`, that's fine with me. @tobiasleibner did you have to change the `"..."_a` lines too? (or do you see anything I should change in my commit that would cause this to work?)

You can find the minimum changes I needed for fixing the segfaults and the linking errors in debug mode [in this branch](https://github.com/tobiasleibner/pybind11/tree/intel_segfault_workaround). Your branch contains some additional changes that are not needed (but also do not hurt).

I reported the bug to Intel via their button "Feedback for oneAPI", but if someone can file an actual bug report that would probably be better.

Though it is a compiler bug, the fact that we had to think about this for quite some time before we were sure that it is indeed correct C++ might be reason enough to change ``noconvert`` and ``none`` to return a copy, if this is possible.

henryiii:
It is still segfaulting, I've tried making those two changes, then tried to match the branch exactly. Maybe the `{}` syntax does not reliably compile with it? Or I missed something?

tobiasleibner:
Unfortunately it works for me locally. Running
```
python3 -m pytest /home/tobias/pybind11/tests/test_async.py /home/tobias/pybind11/tests/test_buffers.py /home/tobias/pybind11/tests/test_builtin_casters.py /home/tobias/pybind11/tests/test_call_policies.py /home/tobias/pybind11/tests/test_callbacks.py /home/tobias/pybind11/tests/test_chrono.py /home/tobias/pybind11/tests/test_class.py /home/tobias/pybind11/tests/test_constants_and_functions.py /home/tobias/pybind11/tests/test_copy_move.py /home/tobias/pybind11/tests/test_custom_type_casters.py /home/tobias/pybind11/tests/test_docstring_options.py /home/tobias/pybind11/tests/test_eigen.py /home/tobias/pybind11/tests/test_enum.py /home/tobias/pybind11/tests/test_eval.py /home/tobias/pybind11/tests/test_exceptions.py /home/tobias/pybind11/tests/test_factory_constructors.py /home/tobias/pybind11/tests/test_gil_scoped.py /home/tobias/pybind11/tests/test_iostream.py /home/tobias/pybind11/tests/test_kwargs_and_defaults.py /home/tobias/pybind11/tests/test_local_bindings.py /home/tobias/pybind11/tests/test_methods_and_attributes.py /home/tobias/pybind11/tests/test_modules.py /home/tobias/pybind11/tests/test_multiple_inheritance.py /home/tobias/pybind11/tests/test_numpy_array.py /home/tobias/pybind11/tests/test_numpy_dtypes.py /home/tobias/pybind11/tests/test_numpy_vectorize.py /home/tobias/pybind11/tests/test_opaque_types.py /home/tobias/pybind11/tests/test_operator_overloading.py /home/tobias/pybind11/tests/test_pickling.py /home/tobias/pybind11/tests/test_pytypes.py /home/tobias/pybind11/tests/test_sequences_and_iterators.py /home/tobias/pybind11/tests/test_smart_ptr.py /home/tobias/pybind11/tests/test_stl.py /home/tobias/pybind11/tests/test_stl_binders.py /home/tobias/pybind11/tests/test_tagbased_polymorphic.py /home/tobias/pybind11/tests/test_union.py /home/tobias/pybind11/tests/test_virtual_functions.py
```
I am getting the output
```
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.7.7, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: /home/tobias/pybind11/tests, configfile: pytest.ini
collected 434 items                                                                                                                                                                                       

../../tests/test_async.py ..                                                                                                                                                                        [  0%]
../../tests/test_buffers.py .........                                                                                                                                                               [  2%]
../../tests/test_builtin_casters.py ....s...........                                                                                                                                                [  6%]
../../tests/test_call_policies.py ........                                                                                                                                                          [  8%]
../../tests/test_callbacks.py .........                                                                                                                                                             [ 10%]
../../tests/test_chrono.py ...........................................                                                                                                                              [ 20%]
../../tests/test_class.py .............................                                                                                                                                             [ 26%]
../../tests/test_constants_and_functions.py ....                                                                                                                                                    [ 27%]
../../tests/test_copy_move.py ....s..                                                                                                                                                               [ 29%]
../../tests/test_custom_type_casters.py ..                                                                                                                                                          [ 29%]
../../tests/test_docstring_options.py .                                                                                                                                                             [ 29%]
../../tests/test_eigen.py .........................                                                                                                                                                 [ 35%]
../../tests/test_enum.py .......                                                                                                                                                                    [ 37%]
../../tests/test_eval.py ...                                                                                                                                                                        [ 38%]
../../tests/test_exceptions.py .........                                                                                                                                                            [ 40%]
../../tests/test_factory_constructors.py ...............                                                                                                                                            [ 43%]
../../tests/test_gil_scoped.py .....                                                                                                                                                                [ 44%]
../../tests/test_iostream.py .............                                                                                                                                                          [ 47%]
../../tests/test_kwargs_and_defaults.py ........                                                                                                                                                    [ 49%]
../../tests/test_local_bindings.py ..........                                                                                                                                                       [ 51%]
../../tests/test_methods_and_attributes.py ....................                                                                                                                                     [ 56%]
../../tests/test_modules.py .....                                                                                                                                                                   [ 57%]
../../tests/test_multiple_inheritance.py ...........                                                                                                                                                [ 60%]
../../tests/test_numpy_array.py ................................................                                                                                                                    [ 71%]
../../tests/test_numpy_dtypes.py ...............                                                                                                                                                    [ 74%]
../../tests/test_numpy_vectorize.py ........                                                                                                                                                        [ 76%]
../../tests/test_opaque_types.py ...                                                                                                                                                                [ 77%]
../../tests/test_operator_overloading.py ....                                                                                                                                                       [ 78%]
../../tests/test_pickling.py .....                                                                                                                                                                  [ 79%]
../../tests/test_pytypes.py ................................                                                                                                                                        [ 86%]
../../tests/test_sequences_and_iterators.py ........                                                                                                                                                [ 88%]
../../tests/test_smart_ptr.py ........CHKP: Bounds check error ptr=0x55ff5107ccc0 sz=4 lb=0x55ff5107ccb8 ub=0x55ff5107ccb7 loc=0x7f94e1c041d3
Traceback:
    at address 0x7f94e1c041d3 in function 
    in file unknown line 0
    at address 0x7f94e1bd6629 in function 
    in file unknown line 0
    at address 0x7f94e1e11a52 in function 
    in file unknown line 0
    at address 0x7f94e1e114f9 in function 
    in file unknown line 0
    at address 0x7f94e1bd6dd9 in function 
    in file unknown line 0
    at address 0x7f94e1bf12fd in function 
    in file unknown line 0
    at address 0x7f94e1d90bd7 in function 
    in file unknown line 0
    at address 0x55ff503d7244 in function _PyMethodDef_RawFastCallKeywords
    in file unknown line 0
    at address 0x55ff503d7384 in function _PyCFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff504418a5 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503823ba in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff50382484 in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff5043e12b in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381308 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503d66da in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
CHKP: Bounds check error ptr=0x55ff5107ccc0 sz=4 lb=0x55ff5107ccb8 ub=0x55ff5107ccb7 loc=0x7f94e1c04243
Traceback:
    at address 0x7f94e1c04243 in function 
    in file unknown line 0
    at address 0x7f94e1bd6629 in function 
    in file unknown line 0
    at address 0x7f94e1e11a52 in function 
    in file unknown line 0
    at address 0x7f94e1e114f9 in function 
    in file unknown line 0
    at address 0x7f94e1bd6dd9 in function 
    in file unknown line 0
    at address 0x7f94e1bf12fd in function 
    in file unknown line 0
    at address 0x7f94e1d90bd7 in function 
    in file unknown line 0
    at address 0x55ff503d7244 in function _PyMethodDef_RawFastCallKeywords
    in file unknown line 0
    at address 0x55ff503d7384 in function _PyCFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff504418a5 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503823ba in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff50382484 in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff5043e12b in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381308 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503d66da in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
CHKP: Bounds check error ptr=0x55ff5107ccc0 sz=4 lb=0x55ff5107ccb8 ub=0x55ff5107ccb7 loc=0x7f94e1c041d3
Traceback:
    at address 0x7f94e1c041d3 in function 
    in file unknown line 0
    at address 0x7f94e1bd6629 in function 
    in file unknown line 0
    at address 0x7f94e1e11a52 in function 
    in file unknown line 0
    at address 0x7f94e1e114f9 in function 
    in file unknown line 0
    at address 0x7f94e1bd6dd9 in function 
    in file unknown line 0
    at address 0x7f94e1bf12fd in function 
    in file unknown line 0
    at address 0x7f94e1d90bd7 in function 
    in file unknown line 0
    at address 0x55ff503d7244 in function _PyMethodDef_RawFastCallKeywords
    in file unknown line 0
    at address 0x55ff503d7384 in function _PyCFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff504418a5 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503823ba in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff50382484 in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff5043e12b in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381308 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503d66da in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
CHKP: Bounds check error ptr=0x55ff5107ccc0 sz=4 lb=0x55ff5107ccb8 ub=0x55ff5107ccb7 loc=0x7f94e1c041d3
Traceback:
    at address 0x7f94e1c041d3 in function 
    in file unknown line 0
    at address 0x7f94e1bd6629 in function 
    in file unknown line 0
    at address 0x7f94e1e11a52 in function 
    in file unknown line 0
    at address 0x7f94e1e114f9 in function 
    in file unknown line 0
    at address 0x7f94e1bd6dd9 in function 
    in file unknown line 0
    at address 0x7f94e1bf12fd in function 
    in file unknown line 0
    at address 0x7f94e1d90bd7 in function 
    in file unknown line 0
    at address 0x55ff503d7244 in function _PyMethodDef_RawFastCallKeywords
    in file unknown line 0
    at address 0x55ff503d7384 in function _PyCFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff504418a5 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503823ba in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff50382484 in function _PyFunction_FastCallDict
    in file unknown line 0
    at address 0x55ff5043eb10 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381c68 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff5043e12b in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff50381308 in function _PyEval_EvalCodeWithName
    in file unknown line 0
    at address 0x55ff503d6976 in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
    at address 0x55ff503d66da in function _PyFunction_FastCallKeywords
    in file unknown line 0
    at address 0x55ff50441958 in function _PyEval_EvalFrameDefault
    in file unknown line 0
.....                                                                                                                                                         [ 91%]
../../tests/test_stl.py ........ss........                                                                                                                                                          [ 95%]
../../tests/test_stl_binders.py .........                                                                                                                                                           [ 97%]
../../tests/test_tagbased_polymorphic.py .                                                                                                                                                          [ 97%]
../../tests/test_union.py .                                                                                                                                                                         [ 98%]
../../tests/test_virtual_functions.py ...s....                                                                                                                                                      [100%]

============================================================================================ warnings summary =============================================================================================
../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
  /opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.factory_constructors.TestFactory3' is using an old-style placement-new '__init__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
    return f(*args, **kwds)

../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
  /opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.factory_constructors.NoisyAlloc' is using an old-style placement-new '__init__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
    return f(*args, **kwds)

../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
  /opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.pickling.Pickleable' is using an old-style placement-new '__setstate__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
    return f(*args, **kwds)

../../../../../opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219
  /opt/intel/oneapi/intelpython/latest/lib/python3.7/importlib/_bootstrap.py:219: FutureWarning: pybind11-bound class 'pybind11_tests.pickling.PickleableWithDict' is using an old-style placement-new '__setstate__' which has been deprecated. See the upgrade guide in pybind11's docs. This message is only visible when compiled in debug mode.
    return f(*args, **kwds)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
========================================================================================= short test summary info =========================================================================================
SKIPPED [1] ../../tests/test_builtin_casters.py:141: no <string_view>
SKIPPED [1] ../../tests/test_copy_move.py:77: no <optional>
SKIPPED [1] ../../tests/test_stl.py:109: no <optional>
SKIPPED [1] ../../tests/test_stl.py:137: no <experimental/optional>
SKIPPED [1] ../../tests/test_virtual_functions.py:188: NCVirt does not work on Intel/PGI/NVCC compilers
=============================================================================== 429 passed, 5 skipped, 6 warnings in 6.42s ================================================================================
CHKP Total number of bounds violations: 4
```
The bounds check errors are probably because I compiled in Debug mode with ``-Wcheck -check-pointers=rw -check-pointers-undimensioned``, they are gone in ``RelWithDebInfo`` mode. 

So it looks like the ``{}`` workaround is not reliable, unfortunately. 

However, I used my pre-existing Intel compiler Docker image, I can test in an image that more closely resembles the CI environment later.

tobiasleibner:
It seems the (new) problem occurs only on Ubuntu 18.04. Trying to import any of the pybind11 libraries (not only ``pybind11_tests``, it immediately segfaults with a backtrace that seems unrelated to pybind11:
```
#0  __GI___libc_free (mem=0x7f4b77aaf010) at malloc.c:3113
#1  0x00007f4b7a56c08f in open_verify (name=name@entry=0x7ffc4687c3b0 "/opt/intel/oneapi/intelpython/python3.7/bin/../lib/libsvml.so", fbp=fbp@entry=0x7ffc4687c600, loader=loader@entry=0x5592f68579a0, 
    whatcode=whatcode@entry=4, mode=mode@entry=-2147483648, found_other_class=found_other_class@entry=0x7ffc4687c5ef, free_name=false, fd=3) at dl-load.c:1971
#2  0x00007f4b7a56c3d8 in open_path (name=name@entry=0x7f4b78510fa2 "libsvml.so", namelen=namelen@entry=11, mode=mode@entry=-2147483648, sps=0x7f4b7a791488, realname=realname@entry=0x7ffc4687c5f0, 
    fbp=fbp@entry=0x7ffc4687c600, loader=0x5592f68579a0, whatcode=4, found_other_class=0x7ffc4687c5ef) at dl-load.c:2052
#3  0x00007f4b7a56f6d8 in _dl_map_object (loader=0x5592f68579a0, name=0x7f4b78510fa2 "libsvml.so", type=2, trace_mode=0, mode=<optimized out>, nsid=<optimized out>) at dl-load.c:2284
#4  0x00007f4b7a5744e2 in openaux (a=a@entry=0x7ffc4687cbd0) at dl-deps.c:63
#5  0x00007f4b797101ef in __GI__dl_catch_exception (exception=0x7ffc4687cbb0, operate=0x7f4b7a5744b0 <openaux>, args=0x7ffc4687cbd0) at dl-error-skeleton.c:196
#6  0x00007f4b7a57485b in _dl_map_object_deps (map=map@entry=0x5592f68579a0, preloads=preloads@entry=0x0, npreloads=npreloads@entry=0, trace_mode=trace_mode@entry=0, 
    open_mode=open_mode@entry=-2147483648) at dl-deps.c:249
#7  0x00007f4b7a57b109 in dl_open_worker (a=a@entry=0x7ffc4687ce20) at dl-open.c:278
#8  0x00007f4b797101ef in __GI__dl_catch_exception (exception=0x7ffc4687ce00, operate=0x7f4b7a57af80 <dl_open_worker>, args=0x7ffc4687ce20) at dl-error-skeleton.c:196
#9  0x00007f4b7a57a98a in _dl_open (file=0x7f4b79555de0 "/root/pybind11/build/tests/pybind11_tests.cpython-37m-x86_64-linux-gnu.so", mode=-2147483646, 
    caller_dlopen=0x5592f495e2ea <_PyImport_FindSharedFuncptr+138>, nsid=<optimized out>, argc=3, argv=<optimized out>, env=0x5592f67a9300) at dl-open.c:605
#10 0x00007f4b7a143f96 in dlopen_doit (a=a@entry=0x7ffc4687d050) at dlopen.c:66
#11 0x00007f4b797101ef in __GI__dl_catch_exception (exception=exception@entry=0x7ffc4687cff0, operate=0x7f4b7a143f40 <dlopen_doit>, args=0x7ffc4687d050) at dl-error-skeleton.c:196
#12 0x00007f4b7971027f in __GI__dl_catch_error (objname=0x5592f6830890, errstring=0x5592f6830898, mallocedp=0x5592f6830888, operate=<optimized out>, args=<optimized out>) at dl-error-skeleton.c:215
#13 0x00007f4b7a144745 in _dlerror_run (operate=operate@entry=0x7f4b7a143f40 <dlopen_doit>, args=args@entry=0x7ffc4687d050) at dlerror.c:162
#14 0x00007f4b7a144051 in __dlopen (file=<optimized out>, mode=<optimized out>) at dlopen.c:87
#15 0x00005592f495e2ea in _PyImport_FindSharedFuncptr () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/dynload_shlib.c:96
#16 0x00005592f49865f8 in _PyImport_LoadDynamicModuleWithSpec () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/importdl.c:129
#17 0x00005592f49868c3 in _imp_create_dynamic_impl.isra.15 (file=0x0, spec=0x7f4b79541210) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/import.c:2171
#18 _imp_create_dynamic () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/clinic/import.c.h:289
#19 0x00005592f487e2e2 in _PyMethodDef_RawFastCallDict () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:530
#20 0x00005592f487e414 in _PyCFunction_FastCallDict () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:586
#21 0x00005592f487e4c5 in PyCFunction_Call () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:791
#22 0x00005592f491b8b4 in do_call_core (kwdict=0x7f4b7a67fbe0, callargs=0x7f4b7a60c350, func=0x7f4b7a6fb7d0) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4642
#23 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3192
#24 0x00005592f485a308 in _PyEval_EvalCodeWithName () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3931
#25 0x00005592f48af976 in _PyFunction_FastCallKeywords () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:433
#26 0x00005592f491a958 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4617
#27 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3094
#28 0x00005592f48af6da in function_code_fastcall (globals=<optimized out>, nargs=2, args=<optimized out>, co=<optimized out>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:283
#29 _PyFunction_FastCallKeywords () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:408
#30 0x00005592f491636f in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4617
#31 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3111
#32 0x00005592f48af6da in function_code_fastcall (globals=<optimized out>, nargs=1, args=<optimized out>, co=<optimized out>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:283
#33 _PyFunction_FastCallKeywords () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:408
#34 0x00005592f4916165 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4617
#35 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3125
#36 0x00005592f48af6da in function_code_fastcall (globals=<optimized out>, nargs=1, args=<optimized out>, co=<optimized out>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:283
#37 _PyFunction_FastCallKeywords () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:408
#38 0x00005592f4916165 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4617
#39 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3125
#40 0x00005592f48af6da in function_code_fastcall (globals=<optimized out>, nargs=2, args=<optimized out>, co=<optimized out>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:283
#41 _PyFunction_FastCallKeywords () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:408
#42 0x00005592f4916165 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:4617
#43 _PyEval_EvalFrameDefault () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Python/ceval.c:3125
#44 0x00005592f485b3ba in function_code_fastcall (globals=<optimized out>, nargs=2, args=<optimized out>, co=0x7f4b7a6f0930) at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:283
#45 _PyFunction_FastCallDict () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:322
#46 0x00005592f485b832 in _PyObject_FastCallDict () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:98
#47 0x00005592f487c3ad in object_vacall () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:1202
#48 0x00005592f48dbb2c in _PyObject_CallMethodIdObjArgs () at /home/sat_bot/base/conda-bld/python_1602182515362/work/Objects/call.c:1252
```
Using Ubuntu 20.04, everything works for me (see comment above).
For reference, this is the Dockerfile I used:
```
FROM ubuntu:20.04
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update; \
    apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg git bash grep sed; \
    wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB; \
    apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB; \
    echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
RUN apt-get update; apt-get install -y intel-oneapi-dpcpp-cpp-compiler-pro cmake python3-dev python3-numpy python3-pytest python3-pip
RUN cd; git clone https://github.com/ax3l/pybind11.git; cd pybind11; git checkout topic-icc
RUN apt-get update; apt-get install -y gdb
```
Then I manually run
```
source /opt/intel/oneapi/setvars.sh
python3 -m pip install --upgrade pip
python3 -m pip install --upgrade pytest
cmake -S . -B build     \
        -DPYBIND11_WERROR=ON    \
        -DDOWNLOAD_CATCH=ON     \
        -DDOWNLOAD_EIGEN=OFF    \
        -DCMAKE_CXX_STANDARD=11             \
        -DCMAKE_CXX_COMPILER=$(which icpc)  \
        -DCMAKE_VERBOSE_MAKEFILE=ON         \
        -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
```
inside the Docker container.

wjakob:
FWIW I am not a big fan of changing all `arg{}` by `arg()` in the test suite (which is also a kind of reference on how to use this library), given how severely buggy the Intel compiler is. Let's  report those issues to Intel and let them fix their software.

wjakob:
I would also strongly recommend to them that they include pybind11 in their internal regression suite. We've had issue after issue with this compiler, and it's always different and weird things.

henryiii:
The point of this is that we can now run Intel's test suite in our CI, since they provide it publicly for the first time now as part of OneAPI. This way we can be aware of and catch issues as quickly as possible, instead of letting it slip by us for several releases. That's one of the main reasons, IMO, that they have so many issues - no public compiler meant no CI's testing major libraries with it, like pybind11 and fmtlib.

I would say that most users should use `"arg"_a`, and if not, `py::arg{"a"}` is valid and `{}` initialization is better (more strict), so if it works around an issue, I'd be fine with the change (though, as I mentioned before, I'd probably be for having at least one example of `py::arg()` protected by a compiler specific version check so we can detect if they fix this in the next release).  If it was a bad example, then I wouldn't be for it.

bstaletic:
@YannickJadoul once came across a segfault on Ubuntu 18.04, that reached into `libapt.so`, from the default signal handler, if my memory serves. And yes, that's `libapt` as in aptitude the package manager. Canonical scares me...

YannickJadoul:
> @YannickJadoul once came across a segfault on Ubuntu 18.04, that reached into `libapt.so`, from the default signal handler, if my memory serves. And yes, that's `libapt` as in aptitude the package manager. Canonical scares me...

I think that's related to a mismatch between the versions of this "apport" hook. By default, my Python has this:

```pycon
>>> import sys
>>> sys.excepthook
<function apport_excepthook at 0x7fa9df058730>
>>> sys.excepthook.__module__
'apport_python_hook'
```

... which I believe is meant to link into `apt-get` to tell you "if you can't find this module, do apt-get install python3-whatever".

A simple solution is to reset this to Python's default by executing `sys.excepthook = sys.__excepthook__`.

I can't really see anything in the stack traces that this is what's happening, though?

tobiasleibner:
Unfortunately, I don't really know how to debug that error on Ubuntu 18.04, the backtrace doesn't have much information. Any ideas? 

Can we simply switch the CI image to use Ubuntu 20.04? At least as a first step to ensure that the ``{}`` workaround works in principle?

henryiii:
Looks like normal tests pass, and it dies on test_embed, which is a step forward. :)

.github/workflows/ci.yml
```diff
@@ -392,6 +393,75 @@ jobs:
...
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build -j 2
+
+    - name: Python tests
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        sudo service apport stop
+        cmake --build build --target pytest; ls; ls /home/runner/work/pybind11/pybind11/build/tests
+        exit 1
```
tobiasleibner, line 450: This looks like the cause of the test failure, or should this line somehow be skipped if everything works?

henryiii: Oh, or it could be an `exit 1`... Probably should actually try to run the full test suite instead of the Python checks here.

tests/test_callbacks.cpp
```diff
@@ -119,7 +119,7 @@ TEST_SUBMODULE(callbacks, m) {
 
     class AbstractBase {
     public:
-        virtual ~AbstractBase() = default;
+        virtual ~AbstractBase() {};
```
bstaletic, line 122: This change is now causing clang-tidy errors.

tobiasleibner: Nice :(. When the destructor is defaulted, the Intel compiler can't find it when linking (no clue why). In Debug builds only, RelWithDebInfo or Release builds are fine.

bstaletic: We can add `// NOLINT` to silence the clang-tidy errors.

henryiii: I feel like I already had to do this for a poor compiler in a couple of places (Intel or PGI). Adding NOLINT is fine, and make it clear that it is needed for some reason (though I'd add a comment to explain why too).

tobiasleibner, line 122:
```suggestion
        // Defaulting this destructor leads to linking errors with the Intel compiler
        // (in Debug builds only, tested with icpc (ICC) 2021.1 Beta 20200827)
        virtual ~AbstractBase() {}; // NOLINT
```

tests/test_class.cpp
```diff
@@ -322,6 +322,7 @@ TEST_SUBMODULE(class_, m) {
 
     class PublicistB : public ProtectedB {
     public:
+    ~PublicistB() override {};
```
tobiasleibner, line 325:
```suggestion
    // Removing this destructor leads to linking errors with the Intel compiler
    // (in Debug builds only, tested with icpc (ICC) 2021.1 Beta 20200827)
    ~PublicistB() override {};
```

tests/test_callbacks.cpp
tobiasleibner: I added suggestions adding NOLINT and comments.

tobiasleibner:
What are the next steps? Can this be merged once the clang-tidy error is fixed, or is there something left to do?

henryiii:
I would like one known "broken" `py::arg("")` usage that is protected with an if not intel. Then we can disable it on the next compiler release to see if Intel fixes this.

@wjakob probably will need to sign off on this, since he expressed a lack of interest in changing this for Intel. But personally, I think that a) using `{}` is better anyway, and b) having our tests in CI FINALLY running on Intel will be a big help. I'm also poking around to see if I can get Intel to run our test suite.

tobiasleibner:
> I would like one known "broken" `py::arg("")` usage that is protected with an if not intel. Then we can disable it on the next compiler release to see if Intel fixes this.

I added back a `py::arg().noconvert()` call guarded by a `!defined(__INTEL_COMPILER)` and verified that it segfaults without the guard (see my [topic-icc branch](https://github.com/tobiasleibner/pybind11/tree/topic-icc)).

Also, this should now (mostly) work with ``std=c++17`` and thus fix #2707. Note that I basically disabled one test for the Intel compiler to fix compilation (it has a simple workaround on the user side, using ``noexcept(true)`` instead of ``noexcept``). However, one Python tests is failing with C++17 that is fine with C++11/14:
```
    def test_aligned():
        if hasattr(m, "Aligned"):
            p = m.Aligned().ptr()
>           assert p % 1024 == 0
E           assert (18375712 % 1024) == 0

../../tests/test_class.py:383: AssertionError
```

Since there were some new workarounds needed to make the C++17 mode work, maybe we could add a second Intel CI run using C++17?

To summarize the current status:
- Everything seems to work with Ubuntu 20.04 and C++11/14
- For C++17, everything compiles but the Python test above fails
- On Ubuntu 18.04, we immediately get a segfault for unknown reasons (see https://github.com/pybind/pybind11/pull/2573#issuecomment-732967161).

henryiii:
I've spoken with the Intel team - can I get a list of issue numbers so they can be checked and possibly prioritized? I think 1 or 2 issues have been submitted (@tobiasleibner, you submitted at least one, right? @ax3l maybe too?).

tests/test_class.cpp
```diff
@@ -322,6 +322,9 @@ TEST_SUBMODULE(class_, m) {
 
     class PublicistB : public ProtectedB {
     public:
+    // Removing this destructor results in linking errors with the Intel compiler
+    // (in Debug builds only, tested with icpc (ICC) 2021.1 Beta 20200827)
+    ~PublicistB() override {};
```
henryiii, line 327:
```suggestion
    ~PublicistB() override {}; // NOLINT
```

tobiasleibner:
> I've spoken with the Intel team - can I get a list of issue numbers so they can be checked and possibly prioritized? I think 1 or 2 issues have been submitted (@tobiasleibner, you submitted at least one, right? @ax3l maybe too?).

I used a feedback button on the Intel side to report the bug, so I don't have an issue number. I just realized that I probably should have used [the Intel forum](https://community.intel.com/t5/Intel-C-Compiler/bd-p/c-compiler). I can create issues there, if that helps.

tests/test_class.cpp
```diff
@@ -324,7 +324,7 @@ TEST_SUBMODULE(class_, m) {
     public:
     // Removing this destructor results in linking errors with the Intel compiler
     // (in Debug builds only, tested with icpc (ICC) 2021.1 Beta 20200827)
-    ~PublicistB() override {};
+    ~PublicistB() override {}; // NOLINT
```
tobiasleibner, line 327: I added the same change in [my branch](https://github.com/tobiasleibner/pybind11/tree/topic-icc). That branch also contains some more workarounds to fix https://github.com/pybind/pybind11/issues/2707.

henryiii: Once we get this in could you PR the fixes needed? We could add a matrix with C++17 in that PR.

tobiasleibner: > Once we get this in could you PR the fixes needed? We could add a matrix with C++17 in that PR.

@mkuron created a follow-up PR which also adds a C++17 CI run. We probably should merge #2573 first, though. It seems as if there is one ``// NOLINT`` missing and one Python test failing here. Any ideas about the Python test?

bstaletic:
`test_python_alreadyset_in_destructor` is the only test that fails in this PR.

ax3l:
`arg()`/`arg{}` transformation lifetime issue submitted as intel bug 04916223 on https://supporttickets.intel.com

ax3l:
Checking the last failing test, I think this is a problem with the `CaptureFixture` in pytest, because I do see
```
PytestUnraisableExceptionWarning: Exception ignored in: 'already_set demo'
```
on the command line when running it - just `capsys.readouterr()` is empty...

ax3l:
Yep, this is an issue with `pytest` 6, downgrading to `pytest==5.4.3` solves the capture issue.

.github/workflows/ci.yml
```diff
@@ -388,6 +389,74 @@ jobs:
...
+
+    - name: Add ICC & Python 3
+      run: sudo apt-get update; sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic cmake python3-dev python3-numpy python3-pytest python3-pip
+
+    - name: Update pip & pytest
+      shell: bash
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade pytest==5.4.3
```
bstaletic, line 419: This looks odd. Why is pytest 6 not causing problems on other CI builds? Are we relying on a breaking change of pytest 6?

bstaletic: First bullet point [here](https://docs.pytest.org/en/stable/changelog.html#features) seems relevant.

YannickJadoul:
> Yep, this is an issue with `pytest` 6, downgrading to `pytest==5.4.3` solves the capture issue.

@ax3l, is this a general issue with `pytest`, or only the combination with ICC?

.github/workflows/ci.yml
henryiii: We pin pytest; most CI jobs get the same one or an older one. So it's quite possible that we don't check on PyTest 6. That first bullet point looks relevant indeed, maybe we should try turning off that feature via the method mentioned in the log. Conditionally for Pytest 6, I expect 🤦.

YannickJadoul: So, this should probably `pip install test/requirements.txt`?

YannickJadoul: I'd suggest going for this. I just confirmed pytest 6 also fails for all other tests, when updating `tests/requirements.txt`: #2741

ax3l: Looks good, I will just rebase this PR and add the `pip install -r tests/requirements.txt` line

henryiii:
I'm not totally sure that 20e467d142ba3ac7e3d6a687482851b7f9c62765 is better; hoping it will get cached, and that might save some setup time. I'd probably recommend saving the commit and reverting it if it's not better (unless we like it better?).

YannickJadoul:
> I'm not totally sure that [20e467d](https://github.com/pybind/pybind11/commit/20e467d142ba3ac7e3d6a687482851b7f9c62765) is better; hoping it will get cached, and that might save some setup time. I'd probably recommend saving the commit and reverting it if it's not better (unless we like it better?).

It took 4 minutes now, the "Building docker image". How can we check if it gets cached?

I don't think I like it better, though . Couldn't we do something like https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer, to run this job inside the container with ICC set up, but still have all the different steps individually (similar to the other jobs) ?

henryiii:
Sure, but we'd need a container, and I don't want to set something up just for this. If there's a way to add a job that builds a container and uploads it to the GH registry if it doesn't exist or needs updating, then we pull from the local registry, that could work, but it would be harder to setup. I think the local action gets cached automatically if unchanged.

I'm also thinking about how we expand this to C++17, but we could just do that as more steps in one job.

It's not all that bad, so we can just go with what we had (though if this works, I'd want to do it for PGI too, and that might make it more reliable since pulling it fails once in a while).

I can pull this commit out and make a new PR with it, probably better.

henryiii:
8 mins 27 seconds, currently.

henryiii:
6m 55s for the non-docker version, 2+ mins in setup. Might investigate later, but not critical (and the job is not as long as I thought it was).

PR #2574, opened 2020-10-10
Add support for custom smart pointers that don't have default constructors

drmoose:
Trying to use [`PYBIND11_DECLARE_HOLDER_TYPE`](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#custom-smart-pointers) to register a smart pointer type that has no default constructor (such as [`gsl::not_null<T>`](https://github.com/microsoft/GSL/blob/master/include/gsl/pointers#L48)) results in a long, difficult-to-read compiler error ultimately caused by [the `holder` field in `struct copyable_holder_caster`](https://github.com/pybind/pybind11/blob/f537093/include/pybind11/cast.h#L1570) trying to default-construct an instance of `holder_type`.

We're working around this in our code with an explicit specialization of `copyable_holder_caster` that uses `unique_ptr<holder_type>` instead of using `holder_type` directly. But, this requires us to copy & paste the entire body of `copyable_holder_caster` (which is rather long) except for the last line. 

This PR attempts to create a more elegant solution by factoring out the storage from `copyable_holder_caster` to a separate `copyable_holder_base` that contains the `holder` and has a SFINAE partial-specialization that uses `unique_ptr<holder_type>` instead of just `holder_type` when `!std::is_default_constructible<holder_type>::value`.

Thanks!

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -16,6 +16,7 @@
 #include "detail/internals.h"
 #include <array>
 #include <limits>
+#include <memory>
 #include <tuple>
 #include <type_traits>
 
@@ -1491,9 +1492,50 @@ struct holder_helper {
     static auto get(const T &p) -> decltype(p.get()) { return p.get(); }
 };
 
+/// Helper class used by copyable_holder_caster to store its underlying holder based on
+/// whether the holder_type is default-constructable. If it is, it's stored as a field (this
+/// template), otherwise as a unique_ptr<holder_base> (the partial specialization below).
+template <typename holder_type, typename SFINAE = void>
+class copyable_holder_base {
+protected:
+    template <typename... Args>
+    void set_holder(Args&&... value) {
+        holder = holder_type(value...);
+    }
+    template <typename T>
+    void set_holder(T&& value) {
+        holder = value;
+    }
+    holder_type& get_holder() { return holder; }
+private:
+    holder_type holder;
+};
+
+template <typename holder_type>
+class copyable_holder_base<holder_type, typename std::enable_if<!std::is_default_constructible<holder_type>::value>::type> {
+protected:
+    template <typename... Args>
+    void set_holder(Args&&... value) {
+        if (holder)
+            *holder = holder_type(value...);
+        else
+            holder = std::unique_ptr<holder_type>(new holder_type(value...));
+    }
+    template <typename T>
+    void set_holder(T&& value) {
+        if (holder)
+            *holder = value;
+        else
+            holder = std::unique_ptr<holder_type>(new holder_type(value));
+    }
+    holder_type& get_holder() { return *holder; }
+private:
+    std::unique_ptr<holder_type> holder;
+};
+
 /// Type caster for holder types like std::shared_ptr, etc.
 template <typename type, typename holder_type>
-struct copyable_holder_caster : public type_caster_base<type> {
+struct copyable_holder_caster : public type_caster_base<type>, copyable_holder_base<holder_type> {
 public:
     using base = type_caster_base<type>;
     static_assert(std::is_base_of<base, type_caster<type>>::value,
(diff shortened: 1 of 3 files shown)


h00shi:
Cannot you just create and use a type wrapping `unique_ptr<unique_ptr<non_nullable_pointer>>` as your holder instead of `non_nullable_pointer`?

drmoose:
@h00shi I'm not sure I understand your suggestion. Replacing my cast.h changes with a nested holder such as this:

```cpp
PYBIND11_DECLARE_HOLDER_TYPE(T, std::unique_ptr<non_null_ptr<T>>);

template <typename T>
struct holder_helper<std::unique_ptr<non_null_ptr<T>>> {
	static const T *get(const std::unique_ptr<non_null_ptr<T>> &p) { return p.get().get(); }
};
}}
```

results in a runtime TypeError

```
____________________________________________________________________________________ test_non_null_ptr _____________________________________________________________________________________

    def test_non_null_ptr():
        ib = m.IntBox(3)
        nnp = m.NonNullPointers()
>       ib2 = nnp.get()
E       TypeError: Unable to convert function return value to a Python type! The signature was
E       	(self: pybind11_tests.smart_ptr.NonNullPointers) -> non_null_ptr<test_submodule_smart_ptr(module_&)::IntBox>

../../tests/test_smart_ptr.py:297: TypeError
```

It may be that I don't properly understand how `PYBIND11_DECLARE_HOLDER_TYPE` is meant to work. I've pushed up this version of the code [here](https://github.com/pybind/pybind11/compare/master...drmoose:non_null_ptr_h00shi). Could you explain what I've done wrong?

h00shi:
This won't work I think since when the type_caster sees unique_ptr it thinks the type it is receiving is what is directly in it. But maybe this would:
```cpp

template<typename T>
class MyPtr {
   // You have to define the constructor here too.
    std::unique_ptr<non_null_ptr<T>> value;
}

PYBIND11_DECLARE_HOLDER_TYPE(T, MyPtr<T>);

namespace pybind11::detail {
template <typename T>
struct holder_helper<MyPtr<T>> {
	static const T *get(const MyPtr &p) { return p.value.get().get(); }
};
}
```

drmoose:
Updating [my `non_null_ptr_h00shi`](https://github.com/pybind/pybind11/compare/master...drmoose:non_null_ptr_h00shi) branch with your `MyPtr` class and filling in the missing constructors & operators results in something that compiles but gets completely ignored because there's no way for the compiler to know that `non_null_ptr<T>` should be cast to a `MyPtr<T>` as it works its way through pybind, since the issue isn't `py::class_` (where I'm still using the normal `std::shared_ptr<T>` as the holder), but rather the `.def`s for member functions that use `gsl::not_null<T>` as parameters or return values.

To try to address the situation, I tried adding an explicit specialization of `type_caster`:

```cpp
template <typename T>
struct type_caster<non_null_ptr<T>> : type_caster<MyPtr<T>> {
    operator non_null_ptr<T>&() const { return this->holder; }
};
```

With that, the `non_null_ptr_h00shi` branch compiles and the caster is correctly detected by the template engine, but unfortunately the resulting test program segfaults. 

I assume that's a problem with me getting pointer lifetimes wrong somewhere rather than a problem with the overall approach. It's something I'll investigate further in the morning, but I'm now pretty confident the approach can work. 

Given a sane workaround exists to the problem this PR tries to solve, I'm not sure whether I should leave the PR up or just close it.

@h00shi Thanks for the idea & for taking the time to explain.

h00shi:
Ok I think you have confused two things. The holder is what you pass to `class_`. If you pass std::shared_ptr, then your holder is `std::shared_ptr`, and `non_null_ptr` will have nothing to do with the holder.

It sounds to me that you only want automatic converstion to `shared_ptr` for functions that return `non_null_ptr` or want `non_null_ptr` as input. That can just be done by defining the right caster. Similar to the ones in [`pybind11/stl.h`](https://github.com/pybind/pybind11/blob/master/include/pybind11/stl.h).

I made an attempt to make one. 
```cpp
template <typename T>
struct type_caster<non_null_ptr<T>> {
    static handle cast(non_null_ptr<T> &&src, return_value_policy p, handle h) {
        std::shared_ptr<T> shared_ptr = std::move(src);
        return type_caster<std::shared_ptr<T>>::cast(shared_ptr, p, h);
    }

        if(!subcaster.load(src, convert)) return false;
        return true;
    }

    // Implementation
    type_caster<std::shared_ptr<T>> subcaster;

    // Each type caster must have these
    template <typename _> using cast_op_type = non_null_ptr<T>;
    static constexpr auto name = type_caster<std::shared_ptr<T>>::name;
    operator non_null_ptr<T>() { return non_null_ptr<T>(static_cast<std::shared_ptr<T>>(subcaster)); }
};
}}
```

It seems to be working. Here is my [branch](https://github.com/h00shi/pybind11/tree/non_null_ptr).

drmoose:
Yep, that's exactly what I had wrong. Thanks!

PR #2575, opened 2020-10-11
Throw error_already_set in py::len on failing PyObject_Length

YannickJadoul:
Found by @Stehsegler and reported [on Gitter﻿](https://gitter.im/pybind/Lobby?at=5f82f20a99e1ab4dd1f18adb).

Previously, `pybind11_fail` would throw an error when `PyObject_Length` indicates an error is thrown, but the error indicator is not cleared. A better approach seems to be to just pass on the Python error.

This should even be backwards compatible, since `py::error_already_set` is a subclass of `std::runtime_error` and `pybind11_fail` would just throw an instance of `std::runtime_error`. So try-catch statements in C++ should continue to work?

Can this still go into 2.6.0, as it's kind of a bugfix? I'll add it to the milestone, but if it's too late, we can just remove it again.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1530,13 +1530,17 @@ inline memoryview memoryview::from_buffer(
 
 /// \addtogroup python_builtins
 /// @{
+
+/// Get the length of a Python object.
 inline size_t len(handle h) {
     ssize_t result = PyObject_Length(h.ptr());
     if (result < 0)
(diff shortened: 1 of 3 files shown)


bstaletic (approved):
I was sure we've fixed this recently...

henryiii (approved):
Bugfixes are fine during the RC phase, and I'd call this a Bugfix. :)

PR #2576, opened 2020-10-11
fix: valgrind-detected after-freeing access of PyMethodDef (macOS Python 3.9.0 segfaults)

### LINKED ISSUE #2558: [bug] Python 3.9 macOS segfault

YannickJadoul:
Python 3.9 added [this line to `meth_dealloc`](https://github.com/python/cpython/blob/c230fde8475e4e5581e74a4235654d17ccf4cff8/Objects/methodobject.c#L169), unfortunately for us, 2 lines too low.

Short summary of what goes wrong:
- pybind11 stores the `function_record` in a `capsule`, and passes this as the `self` argument to `PyCFunction_NewEx`
- The `PyCFunctionObject` object nicely keeps track of that, but decrefs `m->m_self` 2 lines before decrefing `PyCFunction_GET_CLASS(m)`. The `capsule` with the `function_record` with the `PyModuleDef` gets freed.
- `PyCFunction_GET_CLASS` accesses `m_ml -> ml_flags`, where `m_ml` now points to something that was destructed.
- _BOOM_

I guess this being so close explains why we don't often get a crash here. Maybe macOS protects its memory better?

- 61e40f6 demonstrates that leaking the `PyModuleDef` resolves the Python 3.9 crash on macOS.
- d10f154 would be a (the?) solution, if it weren't calling `cpp_function` during the construction of a `cpp_function` during the construction of a `cpp_function` during the construction of a `cpp_function` during ... and thus blowing up the stack.

Closes #2558

Related issue on bugs.python.org: https://bugs.python.org/issue41237
PR on CPython adding the line to `meth_dealloc` that makes things crash: https://github.com/python/cpython/pull/19936

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -466,7 +466,7 @@ protected:
             }
             if (rec->def) {
                 std::free(const_cast<char *>(rec->def->ml_doc));
-                delete rec->def;
+                //delete rec->def;
             }
             delete rec;
             rec = next;
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -452,8 +452,10 @@ protected:
 
     /// When a cpp_function is GCed, release any memory allocated by pybind11
     static void destruct(detail::function_record *rec) {
+        // If on Python 3.9, check the interpreter "MICRO" (patch) version.
+        // If this is running on 3.9.0, we have to work around a bug.
         #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
-            static bool is_zero = Py_GetVersion()[4] != '0';
+            static bool is_zero = Py_GetVersion()[4] == '0';
         #endif
 
         while (rec) {
@@ -471,10 +473,10 @@ protected:
             if (rec->def) {
                 std::free(const_cast<char *>(rec->def->ml_doc));
                 // Python 3.9.0 decref's these in the wrong order; rec->def
-                // If loaded on 3.9.0, let these leak (use Python 3.9.1 to fix)
+                // If loaded on 3.9.0, let these leak (use Python 3.9.1 at runtime to fix)
                 // See https://github.com/python/cpython/pull/22670
                 #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
-                    if (is_zero)
+                    if (!is_zero)
                         delete rec->def;
                 #else
                     delete rec->def;
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -452,6 +452,10 @@ protected:
 
     /// When a cpp_function is GCed, release any memory allocated by pybind11
     static void destruct(detail::function_record *rec) {
+        #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
+            static bool is_zero = Py_GetVersion()[4] != '0';
+        #endif
+
         while (rec) {
             detail::function_record *next = rec->next;
             if (rec->free_data)
@@ -470,7 +474,7 @@ protected:
                 // If loaded on 3.9.0, let these leak (use Python 3.9.1 to fix)
                 // See https://github.com/python/cpython/pull/22670
                 #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
-                    if (Py_GetVersion()[4] != '0')
+                    if (is_zero)
                         delete rec->def;
                 #else
                     delete rec->def;
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -466,7 +466,15 @@ protected:
             }
             if (rec->def) {
                 std::free(const_cast<char *>(rec->def->ml_doc));
-                //delete rec->def;
+                // Python 3.9.0 decref's these in the wrong order; rec->def
+                // If loaded on 3.9.0, let these leak (use Python 3.9.1 to fix)
+                // See https://github.com/python/cpython/pull/22670
+                #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
+                    if (Py_GetVersion()[4] != '0')
+                        delete rec->def;
+                #else
+                    delete rec->def;
+                #endif
             }
             delete rec;
             rec = next;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,12 +70,6 @@ jobs:
             python: pypy3
             arch: x64
 
-            # TODO: renable
-            # Currently segfaults on macOS Python 3.9
-          - runs-on: macos-latest
-            python: 3.9
-            arch: x64
-
     name: "🐍 ${{ matrix.python }} • ${{ matrix.runs-on }} • ${{ matrix.arch }} ${{ matrix.args }}"
     runs-on: ${{ matrix.runs-on }}
 


include/pybind11/pybind11.h
```diff
@@ -466,7 +466,15 @@ class cpp_function : public function {
             }
             if (rec->def) {
                 std::free(const_cast<char *>(rec->def->ml_doc));
```
henryiii, line 468: I left this as-is without touching this line, feel free to change if needed.

YannickJadoul: Nah, python/cpython#22670 just got merged, and backported into python/cpython#22674, so we should be fine excluding 3.9.0 and having something specificially works with this exact patch version :-)

include/pybind11/pybind11.h
```diff
@@ -466,7 +466,15 @@ class cpp_function : public function {
...
-                delete rec->def;
+                // Python 3.9.0 decref's these in the wrong order; rec->def
+                // If loaded on 3.9.0, let these leak (use Python 3.9.1 to fix)
+                // See https://github.com/python/cpython/pull/22670
+                #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
+                    if (Py_GetVersion()[4] != '0')
```
YannickJadoul, line 473: https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Python/getversion.c is not extremely cheap, though, as it pastes together a bunch of strings, each time: https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Python/getversion.c

Is that fine, or should we add this to `details()` or so?

It's probably fine, but just notifying, because this overhead will be there for all 3.9 versions, so also after 3.9.1 is out?

henryiii: I thought this was static, but I see that it isn't. We could avoid running it each time then.

YannickJadoul: The question is if it's worth polluting the rest of the code with, though.

Oh, but it could be made static and shared with all other extension modules, ofc, since they would all be loaded into the same interpreter. We just need to make sure it's safe wrt ABI, I guess?

rwgk: Did you consider using `PY_MICRO_VERSION`? (I double-checked, it exists in Python 2.7 already, although it doesn't actually matter here, as long as you test `PY_MAJOR_VERSION` and `PY_MINOR_VERSION` first.)

henryiii: That's the compile time version, not the runtime version. Py_GetVersion grabs the version from when the running interpreter. So wheels work on any Python version.

rwgk: Oh, got it, thanks! Might be nice to add a comment after line 472, with that explanation.

include/pybind11/pybind11.h
```diff
@@ -452,6 +452,10 @@ class cpp_function : public function {
 
     /// When a cpp_function is GCed, release any memory allocated by pybind11
     static void destruct(detail::function_record *rec) {
+        #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
```
YannickJadoul, line 455: We could even move this to the location it's used, or not?

henryiii: That's where it was originally, but thought it was better not in the loop. Defining a static variable inside an if seems rather odd, but if the compiler handles it correctly, it's probably fine?

include/pybind11/pybind11.h
```diff
@@ -452,6 +452,10 @@ class cpp_function : public function {
...
+            static bool is_zero = Py_GetVersion()[4] != '0';
```
bstaletic, line 456: The variable is called `is_zero`, but the value it holds is the opposite.

include/pybind11/pybind11.h
YannickJadoul: I don't really mind, but yeah, that was my thinking

include/pybind11/pybind11.h
YannickJadoul: Whoops. And then there's Boris, actually bringing up issues with the code...

henryiii: 😳

include/pybind11/pybind11.h
henryiii: The assembly when including a static variable definition is a bit more complex, using a guard variable and a lock, so I think it's better to just do it once, outside the loop. But I'm happy either way. https://gcc.godbolt.org/z/cMcsW9

YannickJadoul: Right. Just keep it like this then, indeed! :-)

YannickJadoul (commented):
Replacing `delete rec->def;` with
```
                static int counter = 0;
                std::cout << ++counter << std::endl;
```

and running pybind11's tests, gives me a final count of 1808, of which 49 before cleaning up (so things that would be cleaned up during, by pybind11, that are the true leaks, here):

```
======================================== test session starts ========================================
platform linux -- Python 3.8.6, pytest-5.3.2, py-1.8.0, pluggy-0.13.1
rootdir: /home/yannick/pybind11/tests, inifile: pytest.ini
plugins: lazy-fixture-0.6.2
collected 430 items                                                                                 

../../tests/test_async.py ..                                                                  [  0%]
../../tests/test_buffers.py .........                                                         [  2%]
../../tests/test_builtin_casters.py ....s...........                                          [  6%]
../../tests/test_call_policies.py ....1
.2
...                                                    [  8%]
../../tests/test_callbacks.py 3
4
...5
.6
.....                                                       [ 10%]
../../tests/test_chrono.py ...........................................                        [ 20%]
../../tests/test_class.py .............7
8
......9
10
.........11
12
13
.                                       [ 26%]
../../tests/test_constants_and_functions.py ....                                              [ 27%]
../../tests/test_copy_move.py ..14
15
16
17
..s..                                                         [ 29%]
../../tests/test_custom_type_casters.py ..                                                    [ 30%]
../../tests/test_docstring_options.py .                                                       [ 30%]
../../tests/test_eigen.py .........................                                           [ 36%]
../../tests/test_enum.py ......                                                               [ 37%]
../../tests/test_eval.py 18
19
.20
.                                                                   [ 37%]
../../tests/test_exceptions.py .........                                                      [ 40%]
../../tests/test_factory_constructors.py ....21
.22
..23
........                                      [ 43%]
../../tests/test_gil_scoped.py 24
25
26
27
.....                                                          [ 44%]
../../tests/test_iostream.py ............                                                     [ 47%]
../../tests/test_kwargs_and_defaults.py ........                                              [ 49%]
../../tests/test_local_bindings.py ..........                                                 [ 51%]
../../tests/test_methods_and_attributes.py .....28
........29
.......                               [ 56%]
../../tests/test_modules.py .....                                                             [ 57%]
../../tests/test_multiple_inheritance.py ........30
31
32
33
34
35
36
37
38
39
40
41
42
43
44
45
...                                          [ 60%]
../../tests/test_numpy_array.py ................................................              [ 71%]
../../tests/test_numpy_dtypes.py ...............                                              [ 74%]
../../tests/test_numpy_vectorize.py ........                                                  [ 76%]
../../tests/test_opaque_types.py ...                                                          [ 77%]
../../tests/test_operator_overloading.py ....                                                 [ 78%]
../../tests/test_pickling.py .....                                                            [ 79%]
../../tests/test_pytypes.py ...............................                                   [ 86%]
../../tests/test_sequences_and_iterators.py ........                                          [ 88%]
../../tests/test_smart_ptr.py .............                                                   [ 91%]
../../tests/test_stl.py ........ss........                                                    [ 95%]
../../tests/test_stl_binders.py .........                                                     [ 97%]
../../tests/test_tagbased_polymorphic.py .                                                    [ 97%]
../../tests/test_union.py .                                                                   [ 98%]
../../tests/test_virtual_functions.py .46
47
.48
.49
.....                                                [100%]

====================================== short test summary info ======================================
SKIPPED [1] test_builtin_casters.py:135: no <string_view>
SKIPPED [1] test_copy_move.py:69: no <optional>
SKIPPED [1] test_stl.py:109: no <optional>
SKIPPED [1] test_stl.py:137: no <experimental/optional>
================================== 426 passed, 4 skipped in 5.81s ===================================
```

Issue #2479, opened 2020-09-11
[BUG] pybind11_add_module forcing -fvisiblity=hidden is a bit too aggressive?


EricCousineau-TRI:
## Issue description

In https://github.com/pybind/pybind11/commit/97aa54fefa, `-fvisibility=hidden` was introduced to `pybind11_add_module` as a means to effectively avoid warnings (by means of setting `CXX_VISIBILITY_PRESET` to `"hidden"`).

I ran into this on our downstream project (which uses a `pybind11` fork) here:
https://github.com/RobotLocomotion/drake/issues/14047
In the fork, I had previously stripped out that CMake code, then restored it when merging in the changes from more active development. (In this issue, I've provided the workaround, which is to re-set the visibility to "default".)

The long and short is, we (unfortunately) have very template-heavy public API and bindings, and setting `-fvisibility=hidden` in downstream units mean that some of the bound symbols may get new RTTI information, which we *don't* want in our project.

From reading the comments in the commit (and the surround commit on current master), it looks like this preset isn't really necessary; it's really only to suppress errors in user's code who aren't sufficiently careful to curate their symbol visibility.

That's fine, but I would still love to not have to reverse that in code and try to track changes.

## Reproducible example code

See downstream issue.

## Possible solutions

From most conservative (keep non-expert users happy) to least conservative (require users be cognizant):

1. No change. Maybe add a docstring to `pybind11_add_module` talking about templates, RTTI, and visibility with bindings, and mention the workaround (which is now coupled to internal implementation details, albeit publicly documented).
2. Add a flag to `pybind11_add_module` to not foist hidden visibility on downstream code (instantiations).
3. Remove the flag. Update the docs to tell users to add it themselves (not super suggested) or add more explicit visibility annotations (heavily suggested).

@rwgk @YannickJadoul @bstaletic @wjakob Thoughts?

EricCousineau-TRI:
Hm... Thinking about this more, option (3) may be a bit hypocritical on my part given that we don't explicitly annotate public visibility on those templates :sweat_smile: 

So perhaps (2) is the best for us (though I guess it's really to enable laziness), and (1) is still the simplest.

wjakob:
[...]

Issue #94, opened 2016-02-04
Failing tests and segfaults with Intel compiler


bennybp:
I fixed a few minor issues when compiling pybind11 with the Intel compiler (v15). After fixing, there are two failing tests.

In example5, Intel seems to use an additional move construction of the lambda payload test at the end. This is interesting, but probably harmless.

    Payload constructor
    Payload copy constructor
    Payload move constructor
    Payload move constructor
    Payload destructor
    Payload destructor
    Payload destructor
    Payload destructor

example8, however, segfaults around cast.h:451 or shortly thereafter. It could be a bug or an an idiosyncrasy  somewhere in the Intel compiler, since gcc seems to be fine.

I'll see if I can track it down in the next few days but if anyone wants to chime in feel free.

PS: The one fix I had to do to get it to compile under Intel was to change the signature of `ignore_unused` in attr.h. Intel rejects this for some reason...

    -template<class T> void ignore_unused(const T&) { }
    +template<class T> void ignore_unused(const T []) { }

wjakob:
Can you provide the error message for the original ignore_unused? The latter version may compile on Intel but doesn't look right to me (why an array?!)-- perhaps there is another workaround that makes all compilers happy.

wjakob:
for reference, what OS is this?

bennybp:
I thought that was weird also. Here's the error:

    <path>/include/pybind11/attr.h(268): error: argument of type "int *" is incompatible with parameter of type "const int (&)[2]"
            ignore_unused(unused);

I'm willing to bet that is actually a bug in Intel - I'll see what they say. It only seems to happen when expanding the parameter pack like that.

I can't believe it's this hard to ignore a variable! I don't really have any other portable workaround. I have seen something like 

    static void init(const Args&... args, function_record *r) {
        using ignore = int[];
        ignore{ 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
    }

It seems to work on GCC 4.9.2 and Intel 15, but clang still reports an unused variable. I have never used that setup before, so I have no idea about its portability.

My OS in Fedora Linux.

wjakob:
Yes, sounds like an Intel compiler bug. I'm actually surprised it can compile this codebase with minor modifications ;)

bennybp:
Ok minor update. I finally installed Intel 2016, and the compilation issue with ignore_unused seems to be fixed.

[...]

PR #2577, opened 2020-10-12
fix: Intel 18+ required

henryiii:
Drops an old workaround for #94, and adds a workaround developed for PGI in #2479.

I've tried to build 2.5.0, 2.4.0, 2.3.0, 2.2.0, and 2.0.0 on Intel 17, and none of them can build the py::vectorize code in the tests. Intel 18 seems to be the first one that can. So I think we should simply say we support Intel 18+, and drop the list of older versions that support Intel 16 and 15, since the support wasn’t even complete and these versions are quite old now anyway.

- fix: Intel 18+ fully supported
- fix: Intel compiler workaround no longer needed

Also gathering a bit of readme cleanup I'd been collecting.

Tested on 18.0 and 19.1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@ grown beyond Boost.Python in many ways, leading to dramatically simpler
 binding code in many common situations.
 
 Tutorial and reference documentation is provided at
-`pybind11.readthedocs.io <https://pybind11.readthedocs.io/en/latest>`_).
+`pybind11.readthedocs.io <https://pybind11.readthedocs.io/en/latest>`_.
 A PDF version of the manual is available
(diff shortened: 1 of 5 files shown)


YannickJadoul (approved):
So this is you testing locally, then? @ax3l is still working on the CI part of this in #2573?

henryiii:
Yes. We can't test anything before 20.2 in CI AFAIK due to Intel licensing.

YannickJadoul:
> Yes. We can't test anything before 20.2 in CI AFAIK due to Intel licensing.

Booooooooo! Well, not our problem, then, if Intel won't help us support their compiler...

wjakob:
LGTM!

henryiii:
Sadly this doesn't solve the instant segfault for 20.2, though.

PR #2580, opened 2020-10-13
docs: PR template

henryiii:
I've restored the classic changelog needed job, since it was not adding labels correctly, and if it did, syncing would make it hard to manage. As an alternative, this is a PR template that includes a spot for a changelog entry that we can manually collect.


rwgk (approved):
Looks good to me.

PR #2582, opened 2020-10-13
docs: some final 2.6.0 updates

henryiii:
Polishing a bit, and highlighting the 3.9.0 issue (is this too high up? Should it be lower down?)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,9 +8,16 @@ pybind11 — Seamless operability between C++11 and Python
 
 .. warning::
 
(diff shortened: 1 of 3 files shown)


YannickJadoul:
> (is this too high up? Should it be lower down?)

Maybe, yes? I would put it by the end of that section (but maybe I'm just too sad about this bug ;-) )

henryiii:
I think "my library is segfaulting on 3.9!" is going to start popping up a lot (the conda-forge migration, for example...), so making the warning pretty visible is important. But yes, I might have overdone it.

PR #2584, opened 2020-10-13
fix: install to DATAROOTDIR instead of hardcoding share

### LINKED ISSUE #2571: [QUESTION] Why is GNUInstallDirs not used in CMakeLists.txt?
Regarding the line below:

https://github.com/pybind/pybind11/blob/59323a975ade742a5904fff23dbd3aecf72ad091/CMakeLists.txt#L181-L184

Why can't it be:

`set(PYBIND11_CMAKECONFIG_INSTALL_DIR "${CMAKE_INSTALL_DATAROOTDIR}/cmake/${PROJECT_NAME}" CACHE STRING "install path for pybind11Config.cmake")`

henryiii:
Closes #2571; this is defined to share, and unlike DATADIR, is not sometimes overwritten incorrectly for storing CMake files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,9 +178,8 @@ target_compile_features(pybind11_headers INTERFACE cxx_inheriting_constructors c
 
 if(PYBIND11_INSTALL)
   install(DIRECTORY ${PYBIND11_INCLUDE_DIR}/pybind11 DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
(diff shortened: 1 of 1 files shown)


henryiii:
I don't see any reason this can't go into 2.6.0, as it's just defined to `share`. This is useful for cross-compiling, as it can be set manually by toolchains. It also should correctly support installing to `/` if someone wanted to do that, as all GNU paths get `usr` added.

Issue #2032, opened 2019-12-16
support for nested classes


pavlis:
## Issue description
I have read through the pybind11 documentation at least a dozen times, looked through all the examples, and did every permutation I could think of searching with google but cannot crack the problem of nested classes in pybind11.   There are hints from some examples that this might be possible, but with no documentation on the topic and no easily found examples methinks this totally standard part of the C++ language may not be supported?   The actual project I'm working on that needs this is much more complex, but the example below captures the idea I'm trying to implement.

## Reproducible example code
#include <string>
#include <list>
#include <vector>
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
#include <pybind11/stl_bind.h>
using namespace std;
namespace py=pybind11;
/* This is a simple error logger class*/
class logger
{
public:
	logger(){};  //probably unnecessary
	void log_error(std::string mess)
	{
		messages.push_back(mess);
	};
	list<string> get_log()
	{
		return messages;
	};
private:
	list<string> messages;
};
/* This is a simple data object that includes an error logger as a 
 * nested class - type example of "has a" for a nested class */
class data_object
{
public:
	data_object(vector<double> d) : v(d){};
	vector<double> get_data(){return v;};
	logger elog;
private:
	vector<double> v;
};

/* Now the pybind11 code */
PYBIND11_MODULE(test_nested,m)
{
	/* Probably not essential for this test, but would be needed*/
	py::bind_vector<std::vector<double>>(m,"Vector");
	py::class_<logger>(m,"logger","Simple error logger class")
		.def(py::init<>())
		.def("log_error",&logger::log_error)
		.def("get_log",&logger::get_log)
	;
	py::class_<data_object>(m,"data_object","Trivial data object for test")
		.def(py::init<vector<double>>())
		.def("get_data",&data_object::get_data)
	       /* Forms to try for logger*/
		//simple reference to symbol
		//.def("elog",&data_object::elog::log_error)
		//variant
		//.def("elog",&data_object::logger::log_error)
		//lambda function modeled after pickle example
		//.def("log_error",[](const std::string mess)
			//{elog.log_error(mess);})  // variation 1	
			//{this->elog.log_error(mess);}) //variation 2	
	;
}

## Explanations
The issue is that data_object has a nested class of logger.  The idea here is to allow processing of data objects in a massively parallel system and allow errors to be posted with the data to allow after the fact sorting out problem results.   This is a classic example of a nested class where data_object "has a" logger.   

[...]

PR #2585, opened 2020-10-13
docs: add note regarding nested type binding to "Instances and static fields" section

xaedes:
Add note regarding nested type binding to "Instances and static fields" section.

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -166,6 +166,10 @@ input for the read function.
     static variables and properties. Please also see the section on
     :ref:`static_properties` in the advanced part of the documentation.
 
+.. note::
+    
+    This can be used to bind nested types.
+
 Dynamic attributes
 ==================
 


YannickJadoul:
How, why? I don't think this is the recommended way to bind nested classes. `py::class_` takes a `scope` argument, if you want nested classes.

YannickJadoul:
Now seeing this is related to #2032. I still don't fully understand the issue there, though, especially given the lack of formatting. Could you elaborate in which case you'd need this?

xaedes:
Yes, good catch, I understood it wrong!
What I meant is nested type definitions like `class Foo { struct Bar { }; };`.

YannickJadoul:
That's what I understood as well. Still not sure about #2032, though, but ... for this, the scope seems to result in the correct Python equivalent.

xaedes:
I tried scope but that wasn't consumed by py::class_. 
I finally found how to do it (in the docs under section "Enumerations and **internal types**").
In hindsight it is obvious, instead of giving the module as first parameter to py::class_ you give it the parent class python binding:
```cpp
class Foo { struct Bar { }; };
auto foo = py::class_<Foo>(m, "Foo");
auto bar = py::class_<Foo>(foo, "Bar");
```

I still think it is too hard to find, given the amount of time I spent looking for it, but maybe I am just stupid ^^

YannickJadoul:
@xaedes, I guess the hope is just that the signature of the constructor
```
    template <typename... Extra>
    class_(handle scope, const char *name, const Extra &... extra) {
```
would put forward the suggestion that the first argument doesn't need to be a module. But you're right that this is not really documented. The docs can always use some polishing, so if you would find a place where one or two sentences can be added (or maybe an example), that should definitely be possible to include! :-)

xaedes:
I will draft something =)

YannickJadoul:
Thanks! :-) Don't overdo it, I'd suggest. Just one or a few well placed sentences might give the right suggestion! :-)

PR #2586, opened 2020-10-13
setup_helpers: fix bug in _add_lflags

jaysoffian:
## Description

Fix minor bug in setup_helpers that allows a link flag to be added more than once.

## Suggested changelog entry:

```rst
Fix minor bug in setup_helpers that allows a link flag to be added more than once.
```


henryiii:
githubstatus.com claims GHA is back online, so going to retrigger.

henryiii:
Thanks!

PR #2588, opened 2020-10-13
feat: typing support for helpers

henryiii:
## Description

The goal here is to enable _external_ typing support, so users can type check code containing pybind11 calls (just the `import pybind11`, not something magical with compiled extensions)! It is also checked _internally_ with mypy, as well, integrated into the pre-commit checks.

However, it did expose a small logical issue when checked internally; `sysconfig.get_path` logically is not required to return a string, so we could be adding None's to the path. Fixed by only adding truthy strings.

If this is too large to go into 2.6.0, we can go with #2589 instead.

## Suggested changelog entry:

```rst
- Added external typing support to the helper module, code from ``import pybind11`` can now be type checked.
```

### DIFF
--- a/pybind11/setup_helpers.pyi
+++ b/pybind11/setup_helpers.pyi
@@ -1,0 +1,47 @@
+from typing import Any, Iterator, Optional, Type, TypeVar, Union
+from types import TracebackType
+
+from distutils.command.build_ext import build_ext as _build_ext  # type: ignore
+from distutils.extension import Extension as _Extension
+import distutils.ccompiler
+import contextlib
+
+WIN: bool
+PY2: bool
+MACOS: bool
+STD_TMPL: str
+
+class Pybind11Extension(_Extension):
+    def _add_cflags(self, *flags: str) -> None: ...
+    def _add_lflags(self, *flags: str) -> None: ...
+    def __init__(
+        self, *args: Any, cxx_std: int = 0, language: str = "c++", **kwargs: Any
+    ) -> None: ...
+    @property
+    def cxx_std(self) -> int: ...
+    @cxx_std.setter
+    def cxx_std(self, level: int) -> None: ...
+
+@contextlib.contextmanager
+def tmp_chdir() -> Iterator[str]: ...
+def has_flag(compiler: distutils.ccompiler.CCompiler, flag: str) -> bool: ...
+def auto_cpp_level(compiler: distutils.ccompiler.CCompiler) -> Union[int, str]: ...
+
+class build_ext(_build_ext):  # type: ignore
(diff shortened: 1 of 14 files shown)


YannickJadoul (approved):
As far as I'm concerned this is not too big or too extensive to still add to 2.6.0. But neither would I expect it's very urgent (do people actually add typing to their `setup.py`?).

henryiii:
It doesn't really touch real code, just adds supporting files mostly, so I rather think it's fine too.

> do people actually add typing to their

Given how poor support for typing is in typeshed for distutils (`build_ext` was missing), I wouldn't expect so. But that could/should change? And, besides, many people use PyCharm (myself included once in a while), and this will just start helping them work with our tools immediately. :)

> neither would I expect it's very urgent

Nope, which is why I didn't get around to it until now. (Plus it's best to add a pyi file after the py file is mostly stable).

YannickJadoul:
> And, besides, many people use PyCharm (myself included once in a while), and this will just start helping them work with our tools immediately. :)

Ah yes, if it helps, why not. This one CI job keeps the `.pyi` in sync with the actual code, right? Will it complain if there's a new function added, for example?

henryiii:
It should complain if you add a function that has no type info (the `--strict` parameter). If you lie about types(*), this is also caught by the internal type checking (external entities only see the .pyi when type checking code against us, if there is one with the same name). The reason to use `.pyi` over type comments is to avoid importing typing and added a dependency; if it was possible to not add a `.pyi`, I did avoid it.

*: Lying defined as making a type statement that does not satisfy the code type check. You may lie about inner details, inheritance, skip non public methods, etc, and that's not lying in this sense.

pybind11/setup_helpers.pyi
```diff
@@ -0,0 +1,47 @@
+from typing import Any, Iterator, Optional, Type, TypeVar, Union
```
rwgk, line 1: Might be best to mark this file and setup_helpers.py prominently with a comment, something like:
IMPORTANT: If you change this file, also review \<the other file\> for matching changes.
That will probably preempt many time-consuming trips through CI to catch mismatches.

rwgk (approved):
I think this is fine to include in 2.6.0.

pybind11/setup_helpers.pyi
henryiii: I would assume if you know enough to edit a `pyi` file, you'd know to edit the `py` file too - but the other direction should have a note for sure. :) I'll add both just to cover all bases.

rwgk: That's a good point and it's fine with me without the comment in the .pyi file, but I really like your wording there!

PR #2589, opened 2020-10-13
fix: minor correctness check (smaller alternative to #2588)

henryiii:
## Description

Backporting a correctness fix from #2588 for 2.6.0. If #2588 is too large to go into 2.6.0, we can do this instead, and save #2588  for later.

## Suggested changelog entry:

None needed.


henryiii:
Looks like #2588 is okay to merge to 2.6.0, so closing this.

PR #2590, opened 2020-10-14
fix: ipo should be off for debug or relwithdebinfo

### LINKED ISSUE #2587: [BUG] Warnings on windows when using `pybind11_add_module` and build type `RelWithDebInfo`
## Issue description

When using `-DCMAKE_BUILD_TYPE=RelWithDebInfo` on Windows, I see some warnings being logged.

<details>
<summary>Example error output</summary>

```
14:39:25   Building Custom Rule C:/ci/ws/src/ros2/rosbag2/rosbag2_py/CMakeLists.txt

14:39:25   _reader.cpp

14:39:25 LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification [C:\ci\ws\build\rosbag2_py\_reader.vcxproj]

14:39:25      Creating library C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_reader.lib and object C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_reader.exp
```
</details>

Other relevant details:
- I checked what are the default link flags that cmake uses on Windows when build type is `RelWithDebInfo`, `/INCREMENTAL` is one of the flags (see [here](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.18.4/Modules/Platform/Windows-MSVC.cmake#L273)).
- `/LTCG` prevents windows `/INCREMENTAL` linkage, thus the warning ([incremental docs](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=vs-2019), [ltcg docs](https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=vs-2019)).
- `/LTCG` is implied by `/GL` compile flag ([ltcg docs](https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=vs-2019)), so `/GL` shouldn't be passed either.

Based on that, it seems that the generator expressions [here](https://github.com/paamand/pybind11/blob/9513b0daeff546568a4143d3c78563be84063528/tools/pybind11Tools.cmake#L103-L108) should be extended so that `LTO` flags are ignored when build type is `RelWithDebInfo` (at least on Windows, it wasn't a problem in other platforms).

## Reproducible example code

I have tried this in an example with many dependencies (code [here](https://github.com/ros2/rosbag2/tree/master/rosbag2_py)), but it should be reproducible in any pybind11 example using cmake and `-DCMAKE_BUILD_TYPE=RelWithDebInfo`.
If you need a confirmation with a simpler example (e.g. https://github.com/pybind/cmake_example/), please let me know and I will add steps on how to reproduce the issue.

cmake version: 3.18.4 (but I think it shouldn't matter)

henryiii:
## Description

Closes #2587 by counting "RelWithDebInfo" like "Debug" for turning off our IPO flags.

If someone wants to make IPO + RelWithDebInfo, they should use the built-in CMake feature and not our old flag method. As a reminder, if `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is set (even to OFF), our IPO flags disable and the built-in CMake feature is respected. You can even set `CMAKE_INTERPROCEDURAL_OPTIMIZATION_<CONFIG>`.

## Suggested changelog entry:

None needed.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -264,8 +264,15 @@ As stated above, LTO is enabled by default. Some newer compilers also support
 different flavors of LTO such as `ThinLTO`_. Setting ``THIN_LTO`` will cause
 the function to prefer this flavor if available. The function falls back to
 regular LTO if ``-flto=thin`` is not available. If
-``CMAKE_INTERPROCEDURAL_OPTIMIZATION`` is set (either ON or OFF), then that
-will be respected instead of the built-in flag search.
+``CMAKE_INTERPROCEDURAL_OPTIMIZATION`` is set (either ``ON`` or ``OFF``), then
+that will be respected instead of the built-in flag search.
+
+.. note::
+
+   If you want to set the property form on targets or the
+   ``CMAKE_INTERPROCEDURAL_OPTIMIZATION_<CONFIG>`` versions of this, you should
+   still use ``set(CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF)`` (otherwise a
+   no-op) to disable pybind11's ipo flags.
 
 The ``OPT_SIZE`` flag enables size-based optimization equivalent to the
 standard ``/Os`` or ``-Os`` compiler flags and the ``MinSizeRel`` build type,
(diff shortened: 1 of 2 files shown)


YannickJadoul (approved):

bstaletic (approved):

henryiii:
Added some notes to the the docs about disabling our IPO to use CMake's.

rwgk (approved):

Issue #2596, opened 2020-10-15
[BUG] Windows PyPy2 (32 bit) support


henryiii:
## Issue description

There are infinite hangs and/or popups about memory corruption on pypy (2) on Windows (32 bit only currently). 

## Reproducible example code

Activate the `- pypy2` job on Windows in CI, or build locally.

henryiii:
Ahah! It's not 3.9, it's the start of the 32 bit wheels! We are broken on 32 bit windows.

bstaletic:
Isn't this resolved by #2599?

henryiii:
No, 32 bit pypy2 Windows builds, but hangs the test suite on CI or pops up error messages (windows version of segfault, I think) when run locally. The PR resolved build errors, and pypy3 passes the test suite happily.

mattip:
I think this can be closed as a won't-fix. PyPy has dropped support for 32-bit windows, and anyone using pypy + pybind11 + windows32-bit + python2 should speak up and request support.

henryiii:
We won't even be keeping Python 2 support around that much more (less than two months), so sounds fine to me.

PR #2598, opened 2020-10-15
ci: Add more win32 testing

henryiii:
## Description

Add Win32, C++20 MSVC, PyPy Windows testing. Showing issue in #2596 . Better way to activate 2017 and (NEW) 2015 support (we could drop Appveyor).

## Suggested changelog entry:

None needed (maybe if fixes are found).

WIP! Disables the normal test suite for easier testing.


bstaletic:
Looking at the output:

###  🐍 3.9 • windows-latest • x86 -DCMAKE_CXX_STANDARD=20

1. A ton of warning about `operator+` and `operator&` (and probably other binary operators) being deprecated between enums of different types. These are all coming from Eigen headers, not pybind11 headers.
2. `std::is_pod` is deprecated in C++20. C++20 removed the concept of "POD", because it's actually not that useful, because the requirements are overly constraining. "POD" used to mean `std::is_standard_layout_v<T> && std::is_trivial_v<T>`, though we might want to look at `numpy.h:1451` and see what our actual requirements are.

###  🐍 pypy2 • windows-latest • x86 and 🐍 pypy3 • windows-latest • x86

1. `warning C4505: 'PySlice_GetIndicesEx': unreferenced local function has been removed`. How does that make sense? `PySlice_GetIndicesEx` is C API and not a "local function". Does `PySlice_GetIndicesEx` not exist in PyPy? The warning *is* originating in `pypy_decl.h`... We're using it in `py::slice::compute`.
2. In `pybind11/numpy.h:586`, `shape->data()` and `strides->data()` return `pybind11::ssize_t*`, but the function takes `Py_intptr_t*` and MSVC wants to see a `reinterpret_cast`.
3. MSVC apparently can not convert from `ssize_t` to `pybind11::ssize_t&`... What does MSVC think `pybind11::ssize_t` is? Seen in `pybind11/tests/test_numpy_vectorize.cpp:89`.
4. Same as 3, but `ssize_t*` to `pybind11::ssize_t*`. Seen in `pybind11/numpy.h:758`.

henryiii:
Taking out Eigen and adding GCC & Clang, both MSVC & GCC complain about the POD, but Clang 10 happily compiles.

henryiii:
"Functions" are a complicated thing on PyPy, generally they are all macros. Sometimes they are on CPython too. See https://bugs.python.org/issue29943. (Don't know how this relates - it would helpful to have access to the includes instead, but it is interesting, nevertheless)

PR #2599, opened 2020-10-15
fix: win32 PyPy3 and C++20

henryiii:
## Description

This adds the fixes from #2598 and additions to the CI:
* Eigen is now 3.3.8 (still no C++20 support)
* Using `py::ssize_t` in more places instead of POSIX `ssize_t`
* Avoid `std::is_pod` (deprecated in C++20)
* Fix warning on PyPy + MSVC (could be fixed upstream?)
* Use reinterpret cast as required by PyPy on Windows
* Fix the extension produced by the classic tooling on PyPy + Windows (at least for 7.3.1, not sure if 7.3.2 would have been better)
* Drop extra new line in PYPY_VERSION
* Add lots of new tests to the CI: C++20 Clang, GCC, and MSVC; 32-bit Windows (including PyPy3 for the first time); and MSVC 2015 and improved MSVC 2017. We can now drop AppVeyor if we want to.

## Suggested changelog entry:

Done.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -34,7 +34,9 @@
    whole npy_intp / ssize_t / Py_intptr_t business down to just ssize_t for all size
    and dimension types (e.g. shape, strides, indexing), instead of inflicting this
    upon the library user. */
-static_assert(sizeof(ssize_t) == sizeof(Py_intptr_t), "ssize_t != Py_intptr_t");
+static_assert(sizeof(::pybind11::ssize_t) == sizeof(Py_intptr_t), "ssize_t != Py_intptr_t");
+static_assert(std::is_signed<Py_intptr_t>::value, "Py_intptr_t must be signed");
+// We now can reinterpret_cast between py::ssize_t and Py_intptr_t (MSVC + PyPy cares)
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
@@ -590,7 +592,10 @@ public:
 
         auto &api = detail::npy_api::get();
         auto tmp = reinterpret_steal<object>(api.PyArray_NewFromDescr_(
-            api.PyArray_Type_, descr.release().ptr(), (int) ndim, shape->data(), strides->data(),
+            api.PyArray_Type_, descr.release().ptr(), (int) ndim,
+            // Use reinterpret_cast for PyPy on Windows (remove if fixed, checked on 7.3.1)
+            reinterpret_cast<Py_intptr_t*>(shape->data()),
+            reinterpret_cast<Py_intptr_t*>(strides->data()),
             const_cast<void *>(ptr), flags, nullptr));
         if (!tmp)
             throw error_already_set();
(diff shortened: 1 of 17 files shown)


include/pybind11/numpy.h
```diff
@@ -583,7 +597,9 @@ class array : public buffer {
 
         auto &api = detail::npy_api::get();
         auto tmp = reinterpret_steal<object>(api.PyArray_NewFromDescr_(
-            api.PyArray_Type_, descr.release().ptr(), (int) ndim, shape->data(), strides->data(),
+            api.PyArray_Type_, descr.release().ptr(), (int) ndim,
+            reinterpret_cast<Py_intptr_t*>(shape->data()),
```
YannickJadoul, line 601: Could you add a comment somewhere that we know to remove it, once PyPy does the same thing a CPyhon?

include/pybind11/pybind11.h
```diff
@@ -29,6 +29,7 @@
 #  pragma warning(disable: 4996) // warning C4996: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name
 #  pragma warning(disable: 4702) // warning C4702: unreachable code
 #  pragma warning(disable: 4522) // warning C4522: multiple assignment operators specified
+#  pragma warning(disable: 4505) // warning C4505: 'PySlice_GetIndicesEx': unreferenced local function has been removed (PyPy only)
```
YannickJadoul, line 32: :+1: I like this better than the `PyPySlice_...` variant!

YannickJadoul (approved):

include/pybind11/pybind11.h
henryiii: That variant didn't work - it still complained about the unused function. :/

include/pybind11/numpy.h
```diff
@@ -33,7 +34,9 @@
    whole npy_intp / ssize_t / Py_intptr_t business down to just ssize_t for all size
    and dimension types (e.g. shape, strides, indexing), instead of inflicting this
    upon the library user. */
-static_assert(sizeof(ssize_t) == sizeof(Py_intptr_t), "ssize_t != Py_intptr_t");
+static_assert(sizeof(::pybind11::ssize_t) == sizeof(Py_intptr_t), "ssize_t != Py_intptr_t");
+static_assert(std::is_signed<Py_intptr_t>::value, "Py_intptr_t must be signed");
+// We now can reinterpret_cast between py::ssize_t and Py_intptr_t (MSVC + PyPy cares)
```
bstaletic, line 39: Can we just use `Py_intptr_t` as a scalar type, instead of `py::ssize_t`?

It's uglier, but it should be possible to avoid casting that way.

include/pybind11/numpy.h
```diff
@@ -328,6 +331,17 @@ template <typename T> using is_pod_struct = all_of<
...
 
+// Replacement for std::is_pod (deprecated in C++20)
+#if !defined(__GNUG__) || defined(_LIBCPP_VERSION) || defined(_GLIBCXX_USE_CXX11_ABI)
+template <typename T> using is_pod = all_of<
+    std::is_standard_layout<T>,
+    std::is_trivially_constructible<T>
+>;
+#else
+template <typename T>
+using is_pod = std::is_pod<T>;
```
bstaletic, line 342: It's simpler to just never use `std::is_pod`.

henryiii: GCC 4.8/4.9 are missing alternative.

include/pybind11/numpy.h
henryiii: Wouldn't the cast still need to happen somewhere, though? Seems like that would just move the problem?

henryiii: And I think the hope is to do whatever CPython does eventually in PyPy, then remove this.

bstaletic: You might be right. I didn't really pay attention how far would we have

include/pybind11/numpy.h
bstaletic: Not according to compiler explorer. Even gcc 4.5 had both `is_trivial` and `is_standard_layout`.

https://godbolt.org/z/5jKx9Y

henryiii: Ah, I was using `is_trivially_constructable`

PR #2602, opened 2020-10-16
fix: allow the ABI string to be forced

henryiii:
## Description

In version 2.4, commit https://github.com/pybind/pybind11/commit/c9f5a added extra protections to the ABI string, which affect some power users (see commits there) as well as PyTorch. This is a small patch that simply allows the various components to be pre-defined by users who know what they are doing (or a user on an odd compiler could set the string for the new compiler!). This is not a "fix", per se, which would take carful study to check ABI compatibility, but would allow users stuck on 2.3 to upgrade or work in special cases.

## Suggested changelog entry:

Should it be only for power users, or have a small changelog entry, and maybe even be listed in the docs?

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -154,49 +154,60 @@ struct type_info {
 
 /// On MSVC, debug and release builds are not ABI-compatible!
 #if defined(_MSC_VER) && defined(_DEBUG)
-#   define PYBIND11_BUILD_TYPE "_debug"
+#  define PYBIND11_BUILD_TYPE "_debug"
 #else
-#   define PYBIND11_BUILD_TYPE ""
+#  define PYBIND11_BUILD_TYPE ""
 #endif
 
 /// Let's assume that different compilers are ABI-incompatible.
-#if defined(_MSC_VER)
-#   define PYBIND11_COMPILER_TYPE "_msvc"
-#elif defined(__INTEL_COMPILER)
-#   define PYBIND11_COMPILER_TYPE "_icc"
-#elif defined(__clang__)
-#   define PYBIND11_COMPILER_TYPE "_clang"
-#elif defined(__PGI)
-#   define PYBIND11_COMPILER_TYPE "_pgi"
-#elif defined(__MINGW32__)
-#   define PYBIND11_COMPILER_TYPE "_mingw"
-#elif defined(__CYGWIN__)
-#   define PYBIND11_COMPILER_TYPE "_gcc_cygwin"
-#elif defined(__GNUC__)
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/internals.h
```diff
@@ -154,49 +154,60 @@ struct type_info {
 
 /// On MSVC, debug and release builds are not ABI-compatible!
 #if defined(_MSC_VER) && defined(_DEBUG)
-#   define PYBIND11_BUILD_TYPE "_debug"
```
rwgk, line 157: Very minor & optional: I'd undo the whitespace change here, to reduce the potential for confusion.

rwgk (approved):

include/pybind11/detail/internals.h
henryiii: https://github.com/pybind/pybind11/pull/2602/files?diff=unified&w=1

henryiii: (I can still undo if you think it should be undone - obviously I didn't search out and fix all uneven whitespace ;) )

wjakob:
(Already merged, but for the record this look all good to me.)

YannickJadoul:
What's the immediate motivation behind this, actually? Is there any way 2.3 _can_ work with 2.6?

This does feel like it could cause weird segfaults that we'll have to debug for users? :-(

henryiii:
The immediate need is for PyTorch. They have been stuck trying to upgrade to 2.4 [since the beginning of the year](https://github.com/pytorch/pytorch/pull/31920) in order to fix bugs, but haven't been able to. The problem is that they compile a bunch of modules, then at run time they also have a JIT that compiles and needs to use the existing module. If the JIT compiler ID doesn't exactly match, 2.4 and later fail due to this addition. They have been happily using it without this for now, and this would allow the upgrade to move forward separately of this change. We could not even debug the issue without the ability to force this.

> we'll have to debug for users

So if a user is intentionally forcing the ABI string with these defs, then coming to us and complaining that the ABI is not actually compatible? And, it would be nice to be able to eventually either support a compiler-independent ABI (might already at least support gcc vs. clang), and have only truly incompatible modules separated by different ABI strings. To test that, we have to be able to force the ABI strings to match.

See https://github.com/pytorch/pytorch/pull/46415 and https://github.com/pytorch/pytorch/issues/46367

YannickJadoul:
Right. Thanks, @henryiii!



> So if a user is intentionally forcing the ABI string with these defs, then coming to us and complaining that the ABI is not actually compatible?

Well, the danger is not knowing it when the user forgets to report it. Guess we'll just have to be extra careful?

Issue #2603, opened 2020-10-17
PDF link returns 404


dsuch:
Hello,

I am just reporting that the link to the PDF version of the documentation currently returns HTTP 404.

https://media.readthedocs.org/pdf/pybind11/master/pybind11.pdf

Regards.

bstaletic:
Wrong link, from a wrong host/organization.

https://pybind11.readthedocs.io/en/stable/
https://pybind11.readthedocs.io/_/downloads/en/stable/pdf/

dsuch:
That wrong link is provided in the documentation and on GitHub. 

https://github.com/pybind/pybind11/blob/01453b05c49a5796fa87d7c1d90931b33beb42a4/README.rst#L52

dsuch:
Interestingly enough, the PDF document that you linked to - for which I thank you - also has a wrong link to itself, i.e. https://buildmedia.readthedocs.org/media/pdf/pybind11/master/pybind11.pdf

YannickJadoul:
Thanks for noticing and taking the time to report, @dsuch. Do you have time to make a PR?

YannickJadoul:
I guess RTD updated their API/links at some point? From that perspective, having this manual link to the PDF might not be the smartest move, but then again, it is quite handy.

dsuch:
Hi @YannickJadoul - I can do it, but is README.rst the only place to update?

bstaletic:
As far as I can tell, it is.

```
bstaletic@Gallifrey pybind11  (git)-[tags/v2.6.0rc3]-% rg media.readthedocs
README.rst
52:`here <https://media.readthedocs.org/pdf/pybind11/master/pybind11.pdf>`_.

docs/conf.py
138:            "//media.readthedocs.org/css/sphinx_rtd_theme.css",
139:            "//media.readthedocs.org/css/readthedocs-doc-embed.css",
```

YannickJadoul:
@dsuch, great!

Yes, since recently, `README.rst` gets included into the docs, such that the same information isn't repeated multiple times :-)

henryiii:
Wow, I didn't realize the PDF build was working, I had that filed away for fixing later. :)

henryiii:
We need to work on the table of contents, though...

```
1 Core features 3
2 Goodies 5
3 Supported compilers 7
4 About 9
Bibliography 157
Index 159
```

henryiii:
Are you going to make a PR @dsuch ?

[...]

PR #2604, opened 2020-10-18
fix: much better pdf

henryiii:
## Description

Fixes the PDF generation, some chars, and table of contents. Even better once #2603 is fixed.

## Suggested changelog entry:

None needed.


henryiii:
ToC is still not quite right...

PR #2606, opened 2020-10-18
fix: chapters in PDF again

henryiii:
## Description

Follow up to #2604, fixing chapters listed in #2603.

## Suggested changelog entry:

None needed.


README.rst
```diff
@@ -1,8 +1,7 @@
 .. figure:: https://github.com/pybind/pybind11/raw/master/docs/pybind11-logo.png
    :alt: pybind11 logo
 
-pybind11 — Seamless operability between C++11 and Python
-========================================================
+**pybind11 — Seamless operability between C++11 and Python**
```
bstaletic, line 4: This change looks wrong. You're changing it from "a section" to "emphasised text". Are you sure this is correct?

henryiii: It's currently the only way I can find to get the table of contents working in the PDF. (And, for this one change, the pybind11 title, it happens to look better too, since the image is the actual "heading" - maybe not the other ones).  The problem is that sections in the index.rst are used instead of the table of contents for PDFs. I've played with several other ideas. `.. only::` cannot affect the section structure. A Transformer happens after the structure is parsed.

Since the readme is short and has not TOC of its own, I don't think making these "real" section headers is all that important.

I'm open to other ideas!

bstaletic: I expected as much. It's good enough for me.

bstaletic (approved):

README.rst
henryiii: I found a way to keep these using a bit of a hack. :)

henryiii: PS: The LaTeX ToC is known to have issues, see https://groups.google.com/g/sphinx-users/c/_zIypOv2d4g

docs/conf.py
```diff
@@ -17,6 +17,10 @@
 import os
 import shlex
 import subprocess
+from pathlib import Path
```
henryiii, line 20: This means Python 3 is required to build the docs. It's 2020.

PR #2611, opened 2020-10-20
fix: Reject keyword argument `None` with `.none(false)`

potpath:
## Description

Demo a failing test case where passing keyword as `None` to `py::arg("a").none(false)` is still accepted.

```cpp
m.def("no_none", [](T*){}, py::arg("a").none(false));
```

```python
m.no_none(None)    # ok raises `TypeError` 
m.no_none(a=None)  # not ok, this is accepted
```

## Suggested changelog entry:

N/A.
This PR is to show a failing test case that can be used as a starting point.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -606,15 +606,15 @@ protected:
                 // 1.5. Fill in any missing pos_only args from defaults if they exist
                 if (args_copied < func.nargs_pos_only) {
                     for (; args_copied < func.nargs_pos_only; ++args_copied) {
-                        const auto &arg = func.args[args_copied];
+                        const auto &arg_rec = func.args[args_copied];
                         handle value;
 
-                        if (arg.value) {
-                            value = arg.value;
+                        if (arg_rec.value) {
+                            value = arg_rec.value;
                         }
                         if (value) {
                             call.args.push_back(value);
-                            call.args_convert.push_back(arg.convert);
+                            call.args_convert.push_back(arg_rec.convert);
                         } else
                             break;
                     }
@@ -628,11 +628,11 @@ protected:
                     bool copied_kwargs = false;
 
                     for (; args_copied < num_args; ++args_copied) {
-                        const auto &arg = func.args[args_copied];
+                        const auto &arg_rec = func.args[args_copied];
 
                         handle value;
-                        if (kwargs_in && arg.name)
-                            value = PyDict_GetItemString(kwargs.ptr(), arg.name);
+                        if (kwargs_in && arg_rec.name)
+                            value = PyDict_GetItemString(kwargs.ptr(), arg_rec.name);
 
                         if (value) {
                             // Consume a kwargs value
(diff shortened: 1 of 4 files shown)


YannickJadoul:
Thanks for reporting, @potpath! Weird that this was never encountered before.

I've pushed a fix to your branch that hopefully fixes this! :-)

include/pybind11/pybind11.h
```diff
@@ -628,26 +628,30 @@ class cpp_function : public function {
                     bool copied_kwargs = false;
 
                     for (; args_copied < num_args; ++args_copied) {
-                        const auto &arg = func.args[args_copied];
+                        const auto &arg_rec = func.args[args_copied];
```
YannickJadoul, line 631: This rename is, strictly speaking, unrelated, but to increase conformity with the naming in step `// 1.` above, and to more clearly indicate the type, I renamed it.

tests/test_methods_and_attributes.py
```diff
@@ -404,6 +404,19 @@ def test_accepts_none(msg):
     assert m.ok_none4(None) == -1
     assert m.ok_none5(None) == -1
 
+    with pytest.raises(TypeError) as excinfo:
+        m.no_none_kw(None)
```
bstaletic, line 408: `no_none_kw` seems to suggest that one keyword arguments with `None` as the value are forbidden. At least that's how I'm reading it.

YannickJadoul: Also after the others, `no_none{1..5}` and `ok_none{1..5}`?
Any suggestions? `no_none_as_kw`?

henryiii:
Let's add a change log entry too, unless this worked in 2.5. Might also Greg and make sure this didn't have some other use or attempt to use elsewhere, and also make sure it's mentioned in the docs.

tests/test_methods_and_attributes.py
bstaletic: `no_none` sounds good.

YannickJadoul:
> Let's add a change log entry too, unless this worked in 2.5.

Right, I overlooked that by building further on this existing PR.

> Might also grep and make sure this didn't have some other use or attempt to use elsewhere, and also make sure it's mentioned in the docs.

Where would it? Tests pass, so I assume this wasn't tested or considered before?

YannickJadoul:
> Right, I overlooked that by building further on this existing PR.

Fine to log under "Smaller or developer focused features and fixes:"?

bstaletic (approved):

rwgk (approved):
If this can still be included in 2.6.0 that would be great, but only if it doesn't cause a lot of extra work.

YannickJadoul:
> but only if it doesn't cause a lot of extra work.

Good point, but I don't think so; I think it's ready. To me, this is almost purely a bug fix (since I don't think we mention in the docs that this only works for positional arguments?).

henryiii:
I think it's fine with me.

henryiii (approved):

potpath:
Thank you everyone for this prompt fix!

PR #2613, opened 2020-10-21
Remove friend declaration of py::class_ in py::detail::generic_type

bstaletic:
## Description

This line had two bugs:

1. It declares `py::detail::class_` as `friend`.
2. After fixing that, we would have to change it to
   `template <typename, typename...>`

The first one was introduced ~5 years ago, when a large refactoring was
made, probably as an intermediate step during refactoring.

The second was made when `generic_type` was made to be agnostic with
respect to the order of `py::class_` template parameters.

&nbsp;

We're removing the declaration altogether, because it was never relied
on. This is what makes me think that it was an intermediate step in
refactoring that shouldn't have ended up in commit history.

## Suggested changelog entry:

None needed. This is just changing the internals.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1027,7 +1027,6 @@ inline dict globals() {
 PYBIND11_NAMESPACE_BEGIN(detail)
 /// Generic support for creating new Python heap types
 class generic_type : public object {
-    template <typename...> friend class class_;
 public:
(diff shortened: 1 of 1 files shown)


YannickJadoul (approved):
This builds without, and passes all tests. I can't see any complaints against this, so I'm merging this, to still get into 2.6.0 :-)

YannickJadoul:
Thanks for fixing this, after the conversation on Gitter, @bstaletic!
Thanks for reporting it on Gitter, @wanchaol!

PR #2614, opened 2020-10-21
Fix mispelling in basics.rst

winksaville:
## Description

Replaced adviced with advised as adviced is not a word.


YannickJadoul (approved):

YannickJadoul:
Thanks! :-)

PR #2616, opened 2020-10-23
Add __builtins__ to globals argument of `py::exec` and `py::eval` if not present

### LINKED ISSUE #1091: builtins not available by default in py::eval
## Issue description

builtins are not available by default when using py::eval, i.e. when using an empty dict as scope.  Instead, it is necessary to pass e.g. a module's `__dict__`, which is actually semantically weird because a module's `__dict__` does NOT, in fact, contain the builtins (as can be checked manually).

Instead, py::eval should just use the same semantics as the builtin eval, i.e. `builtins.__dict__` should always be available in the execution scope, regardless of what gets passed in as globals and locals.

The same remark applies to py::exec, of course.

## Reproducible example code

Starting from the python_example
```
#include <pybind11/eval.h>

namespace py = pybind11;

    // auto scope = py::module::import("__main__").attr("__dict__");  // this works
    auto scope = py::dict{};  // this crashes Py3.6 at import time with a SystemError
    py::eval("print(123)", scope);
}
```
build, and try to `import python_example`.

---

#887 appears to be a similar, but different issue.

### LINKED ISSUE #1654: missing __builtins__ in the scope with py::exec
## Issue description

When creating a module using the constructor directly instead of using a macro from `pybind11`, and when defining multiple functions in it using `py::exec`, there seem to be an issue. I'm not sure yet if the issue comes from `py::exec`, but the following code works fine when `py::exec` is replaced by `py::module::import("builtins").attr("exec")`.

## Reproducible example code

```c++
py::module my_module("my_module");
std::string my_code = "\n"
       "def foo():\n"
       "    return 4\n"
       "\n"
       "def bar():\n"
       "    return foo()\n";
py::exec(my_code, py::globals(), my_module.attr("__dict__"));

py::print(my_module.attr("foo")()); // Prints "4" as expected
py::print(my_module.attr("bar")()); // Oops, "NameError: global name 'foo' is not defined"
```

### LINKED ISSUE #2557: [BUG] ImportError: ImportError: __import__ not found Python 3.7
I used python_example (https://github.com/pybind/python_example) to create new package.
After installation I have

```
>>> import mymodule
ImportError: ImportError: __import__ not found

At:
  <string>(3): <module>
  <frozen importlib._bootstrap>(222): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(914): create_module
  <frozen importlib._bootstrap>(577): module_from_spec
  <frozen importlib._bootstrap>(666): _load_unlocked
  <frozen importlib._bootstrap>(958): _find_and_load_unlocked
  <frozen importlib._bootstrap>(969): _find_and_load
  <stdin>(1): <module>
```

On python3.8 works perfectly

## Reproducible example code

Debian10. Python 3.7

Clone and install python_example

Add this code to main.cpp

```
auto global = m.attr("__dict__");
py::exec(R"(
    import contextlib
)", global);
```

EDIT (@YannickJadoul): "*After reading, remove this checklist and the template text in parentheses below.*"

YannickJadoul:
## Description

Python (even 2.7!) adds `__builtins__` when calling `exec` or `eval`:

```
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
>>> d = {}
>>> d.keys()
dict_keys([])
>>> exec("print(42)", d)
42
>>> d.keys()
dict_keys(['__builtins__'])
```

```
[GCC 7.5.0] on linux2
>>> d = {}
>>> exec("print 42", d)
42
>>> d.keys()
['__builtins__']
```

For the C API's `PyRun_String`, this only happens since Python 3.8 (python/cpython#13362; see https://github.com/pybind/pybind11/issues/1091#issuecomment-668594173). It would make sense to me to backport/downport/sideport/<whatever>port this to pybind11 on all versions of Python, for the sake of consistency. No?

Closes #1091
Closes #2557
Closes #1654 (thanks, @bstaletic!)

## Suggested changelog entry:

```rst
`py::exec`, `py::eval`, and `py::eval_file` now add the builtins module as ``"__builtins__"`` to their ``globals`` argument, better matching `exec` and `eval` in pure Python.
```

### DIFF
--- a/include/pybind11/eval.h
+++ b/include/pybind11/eval.h
@@ -14,6 +14,22 @@
 #include "pybind11.h"
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+inline void ensure_builtins_in_globals(object &global) {
+    #if PY_VERSION_HEX < 0x03080000
+        // Running exec and eval on Python 2 and 3 adds `builtins` module under
+        // `__builtins__` key to globals if not yet present.
+        // Python 3.8 made PyRun_String behave similarly. Let's also do that for
+        // older versions, for consistency.
+        if (!global.contains("__builtins__"))
+            global["__builtins__"] = module_::import(PYBIND11_BUILTINS_MODULE);
+    #else
+        (void) global;
+    #endif
+}
+
+PYBIND11_NAMESPACE_END(detail)
 
 enum eval_mode {
     /// Evaluate a string containing an isolated expression
@@ -31,14 +47,7 @@ object eval(str expr, object global = globals(), object local = object()) {
     if (!local)
         local = global;
 
-#if PY_VERSION_HEX < 0x03080000
-    // Running exec and eval on Python 2 and 3 adds `builtins` module under
-    // `__builtins__` key to globals if not yet present.
-    // Python 3.8 made PyRun_String behave similarly. Let's also do that for
-    // older versions, for consistency.
-    if (!global.contains("__builtins__"))
-        global["__builtins__"] = module_::import(PYBIND11_BUILTINS_MODULE);
-#endif
+    detail::ensure_builtins_in_globals(global);
 
     /* PyRun_String does not accept a PyObject / encoding specifier,
        this seems to be the only alternative */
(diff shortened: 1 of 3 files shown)


bstaletic (approved):

henryiii (approved):

bstaletic:
Fixes #1654 as well.

martinRenou:
Thanks a lot @YannickJadoul

wjakob (approved):
Looks great!

rwgk (approved):
This looks good to me as-is, but what do you think about moving the new code into `inline void detail::ensure_global_contains_builtins()`? That way you'd only need one `#if` (inside that function).

YannickJadoul:
> what do you think about moving the new code into `inline void detail::ensure_global_contains_builtins()`

I'm not entirely convinced on which side to pick here. On a high level, definitely, avoid code duplication. In practice ... I'm a bit in doubt whether it looks better. But I guess it also doesn't look worse.

YannickJadoul:
One or two reviews of this last commit, please?

bstaletic (approved):

YannickJadoul:
Thanks, all!

PR #2621, opened 2020-10-26
Rework function signatures in docstrings of overloaded functions

### LINKED ISSUE #2619: [FEAT] Generate docstrings that are compatible with Sphinx and stubgen
This issue follows on from a discussion at https://github.com/python/mypy/issues/9070

## Current Behaviour

Currently, when a function is overloaded in pybind, the docstring generated includes a single, generic type signature at the top (`name(*args, **kwargs) -> Any).
Sphinx autodoc will document the signature of function using this generic type signature (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_docstring_signature). However it recently added support for reading multiple type signatures from the docstring (https://github.com/sphinx-doc/sphinx/issues/2106). This is apparently an already established convention and it's used in CPython source code (https://github.com/python/cpython/blob/c0f22fb8b3006936757cebb959cee94e285bc503/Objects/rangeobject.c#L156).
Furthermore, stubgen (https://mypy.readthedocs.io/en/stable/stubgen.html) uses this generic function signature, which makes the type stubs that it generates useless (this was the original issue discussed in https://github.com/python/mypy/issues/9070).

## New Behaviour

This issue is a feature request to prepend docstrings with all of the overload signatures of the function (unless `options.disable_function_signatures()` has been called).

To use an example from the pybind11 documentation:

```cpp
py::class_<Pet>(m, "Pet")
   .def(py::init<const std::string &, int>())
   .def("set", (void (Pet::*)(int)) &Pet::set, "Set the pet's age")
   .def("set", (void (Pet::*)(const std::string &)) &Pet::set, "Set the pet's name");
```

This produces docstrings that looks like the following:

```python
>>> help(example.Pet)

class Pet(__builtin__.object)
 |  Methods defined here:
 |
 |  __init__(...)
 |      __init__(self: Pet, arg0: str, arg1: int) -> None
 |
 |  set(...)
 |      set(*args, **kwargs) -> Any
 |      Overloaded function.
 |
 |      1. set(self: Pet, arg0: int) -> None
 |
 |      Set the pet's age
 |
 |      2. set(self: Pet, arg0: str) -> None
 |
 |      Set the pet's name
```

This request is to have the docstrings instead look like the following:

```python
>>> help(example.Pet)

class Pet(__builtin__.object)
 |  Methods defined here:
 |
 |  __init__(...)
 |      __init__(self: Pet, arg0: str, arg1: int) -> None
 |
 |  set(...)
 |      set(self: Pet, arg0: int) -> None
 |      set(self: Pet, arg0: str) -> None
 |      Overloaded function.
 |
 |      1. set(self: Pet, arg0: int) -> None
 |
 |      Set the pet's age
 |
 |      2. set(self: Pet, arg0: str) -> None
 |
 |      Set the pet's name
```

Furthermore, I think it needs to be possible to have all of the function signatures at the start of the docstring generated but without the section headings. This would allow a user to write the whole docstring as one, but still have the function signatures generated for Sphinx and/or stubgen.

So it should be possible to generate a docstring like this:

```python
>>> help(example.Pet.set)

set(...)
 |      set(self: Pet, arg0: int) -> None
 |      set(self: Pet, arg0: str) -> None
 |      Set attributes on the pet.
 |
 |      When arg0 is an integer, the pet's age is set.
 |      When arg0 is a string, the pet's name is set.
```

## Implementation Details

I think the second change, to generate the signatures without signature headings, would require the addition of a new option. To maintain current behaviour, the `options.disable_function_signatures()` function could still turn all signatures off. A new option `disable_signature_headings()` would turn only the section headings off.

So the following code would produce the docstring seen above:

```cpp
py::options options;
options.disable_signature_headings();

py::class_<Pet>(m, "Pet")
   .def(py::init<const std::string &, int>())
   .def("set", (void (Pet::*)(int)) &Pet::set)
   .def("set", (void (Pet::*)(const std::string &)) &Pet::set, "Set attributes on the pet.\n\nWhen arg0 is an integer, the pet's age is set.\nWhen arg0 is a string, the pet's name is set.");
```

AWhetter:
## Description

This pull request changes the docstring generation so that all function signatures are prepended to the docstring of an overloaded function, instead of the generic signature that is prepended currently.
It also includes an additional change that allows the section headings in the docstring to be disabled independently of whether or not function signatures are shown. To keep backwards compatibility, calling `disable_function_signatures()` will still turn off both the prepended signatures and the section headings. Therefore section headings will only be shown if both function signatures and section headings are enabled. To rephrase, section headings cannot be enabled independently of disabling function signatures.

Closes #2619 

## Suggested changelog entry:

```rst
* All function signatures of an overloaded function are now prepended to the function's docstring.
  To support this change, a new ``disable_section_headings()`` function has been added to the ``options`` class to allow
  the section headings in the docstring of an overloaded function to be disabled.
  `#2621 <https://github.com/pybind/pybind11/pull/2621>
```


YannickJadoul:
Messing around with this myself, it seems Sphinx is expecting a backslash at the end of the line, if the following is still supposed to be part of the overloaded signature. Apart from that, things seem to work.

AWhetter:
Nice find! I'm not sure how I missed that before. But a small end-to-end test is working as it should for me now.

> which is quite ugly, in the raw docstring, but...

You're allowed to have a space before the backslash. I haven't done that here but if you think it looks better with the space there then let me know and I can change it.

YannickJadoul:
@AWhetter, ah, it does allow that space? Didn't know, but that might be slightly less ugly, yes. I think it still ugly, though (and I don't fully understand why it's necessary; Sphinx could just try parsing the next line and check if it's a signature, just like it does for the first line?), but yes, I've also made a similar ad-hoc change to the pybind11 subtree in my project, now, so thanks for pointing me to this! :-)

Is this ready for review? Guess this is only scheduled for 2.7, anyway, but I could go over this after having played around with it myself :-)

AWhetter:
I've just squashed the commits back up, so this is now ready for review.

YannickJadoul:
> To rephrase, section headings cannot be enabled independently of disabling function signatures.

Just a thought: as you say, this PR is making two changes: 1) it is adding the overloaded signatures to be recognized by e.g. Sphinx, and 2) it is removing these "section headings".

What about splitting this into two? I would expect that the different sections are still useful if people have different docstrings for different overloads, but that having the overloaded signatures at the start is almost always better (and so we could turn that to on-by-default, maybe)?

It's just an idea. I'm not sure I like the option-class to become too fine-grained either, actually, but...

We might want @wjakob to weight in on this as well, actually.

tests/test_factory_constructors.py
```diff
@@ -101,6 +104,17 @@ def test_init_factory_signature(msg):
     """  # noqa: E501 line too long
     )
 
+    assert (
+        msg(m.TestFactory2.__init__.__doc__)
+        == """
+        __init__(self: m.factory_constructors.TestFactory2, arg0: m.factory_constructors.tag.pointer_tag, arg1: int) -> None \\
+        __init__(self: m.factory_constructors.TestFactory2, arg0: m.factory_constructors.tag.unique_ptr_tag, arg1: str) -> None \\
+        __init__(self: m.factory_constructors.TestFactory2, arg0: m.factory_constructors.tag.move_tag) -> None
+        This is one part of the docstring.
```
YannickJadoul, line 113: So, as said above, I think I'd like to see a single white line above this. Wouldn't that look better, splitting off the signatures from the body?

Anyway, just pointing out in the tests, which seems easier to judge.

include/pybind11/pybind11.h
```diff
@@ -427,13 +434,16 @@ class cpp_function : public function {
...
+                     first_user_def = false;
+                }
+                else {
+                    signatures += "\n";
                 }
-                if (options::show_function_signatures()) signatures += "\n";
                 signatures += it->doc;
-                if (options::show_function_signatures()) signatures += "\n";
+                if (show_signature_headings) {
+                    signatures += "\n";
```
YannickJadoul, line 445: If you're really feeling adventurous: it seems there's always a newline attached to the end of the docstring if `show_signature_headings`? Not worth making the code more complicate for, but maybe this helps restructuring the logic?

If you want, I can have a look, as well.

YannickJadoul (commented):
As far as I'm concerned, this is an improvement :-) I'd really like to use it, so, yeah, let's add it to the PRs to be merged for 2.7?

YannickJadoul:
Related to your comment in #2619: it's a bit of a pity that CPython's docstrings don't have the trailing `\` :-/
Did you check that stubgen handles this new version with `\` correctly as well?

AWhetter:
By chance, stubgen works, but only when the signature headings are included. stubgen isn't capable of parsing the backslash separated signatures at the start of the docstring, so it ignores them, but it can parse the section headings. So the original issue I described of stubgen outputting `myfunc(*args, **kwargs) ->Any` no longer happens.

I still think there's value in making the section headings in a docstring optional, but I'm more worried about resolving the issue of stubgen outputting the `*args, **kwargs` signature.
If we do decide that making the section headings optional is worth it, I've submitted a pull request to stubgen to support that: https://github.com/python/mypy/pull/10019

include/pybind11/pybind11.h
```diff
@@ -408,15 +408,22 @@ class cpp_function : public function {
...
-            // First a generic signature
-            signatures += rec->name;
-            signatures += "(*args, **kwargs)\n";
-            signatures += "Overloaded function.\n\n";
+            for (auto it = chain_start; it != nullptr; it = it->next) {
+                signatures += rec->name;
+                signatures += it->signature;
+                if (it->next != nullptr) signatures += " \\";
+                signatures += "\n";
+            }
```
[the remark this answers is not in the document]
AWhetter: It does look better. Both Sphinx and stubgen support it so I've added it!

include/pybind11/pybind11.h
AWhetter: The original code always added a new line. But if we wanted to get rid of it then I think it would make the code more complicated. I'm thinking we would have to do something like

```cpp
        for (auto it = chain_start; it != nullptr; it = it->next) {
            bool appending_docstring = (it->doc && strlen(it->doc) > 0 && options::show_user_defined_docstrings());
            bool last_section = (it->next == nullptr);
            if (show_signature_headings) {
                if (index > 0) signatures += "\n";
                if (chain)
                    signatures += std::to_string(++index) + ". ";
                signatures += rec->name;
                signatures += it->signature;
                if (appending_docstring || !last_section)) signatures += "\n";
            }
            if (appending_docstring) {
                // If we're appending another docstring, and aren't printing function signatures, we
                // need to append a newline first:
                if (!show_signature_headings && first_user_def)
                     first_user_def = false;
                else
                    signatures += "\n";
                signatures += it->doc;
                if (show_signature_headings && !last_section) signatures += "\n";
            }
        }
```

YannickJadoul:
> By chance, stubgen works, but only when the signature headings are included. stubgen isn't capable of parsing the backslash separated signatures at the start of the docstring, so it ignores them, but it can parse the section headings. So the original issue I described of stubgen outputting `myfunc(*args, **kwargs) ->Any` no longer happens.

Ugh. I had just thrown out the headings :-(

> I still think there's value in making the section headings in a docstring optional, but I'm more worried about resolving the issue of stubgen outputting the `*args, **kwargs` signature.

Agreed. Does it make sense to submit an issue to stubgen or Sphinx (or both, such that they communicate!) to make sure that they act the same? :-/

include/pybind11/pybind11.h
YannickJadoul: Your call on whether it's worth it. I know the original code did it, but it seems rather silly. I had hoped it would make the code easier/more intuitive. :-(

YannickJadoul:
> Does it make sense to submit an issue to stubgen or Sphinx (or both, such that they communicate!) to make sure that they act the same? :-/ (i.e., with `\` or without `\`; I don't really care either way, though without `\` does look nicer to me)

Oh. python/mypy#10019. I should learn how to read. Sorry! (Though asking Sphinx to not require the `\` also still could make sense to me, though. But it's going to be harder to implement?)

sizmailov:
@AWhetter Would it be less intrusive to make sphinx and stubgen support pybind11 overloading syntax? 
FYI: A set of changes to stubgen is on the way: https://github.com/python/mypy/pull/9903

AWhetter:
It would be less intrusive for pybind, yes. However I'm trying to make everything agree on a convention that is also used outside of pybind, rather than have everything support pybind's way of doing it.

You're right that we should be looking for a solution that minimises intrusion though, and I think that @YannickJadoul has identified that solution; get Sphinx to support finding signatures without a backslash, stubgen would require no changes, and get pybind to put function signatures at the beginning of the docstring.

sizmailov:
Ok, that seems reasonable. 

What about signatures that span multiple lines? Presumably pybind11 would always produce exactly one signature per line, but we are talking about 3rd party tools too. How should it be handled? One can argue for a like-python approach: keep parsing new lines until signature is syntactically complete. 

Example: 
```python
foo(int) -> Callable[
              Optional[Tuple[int, int]]
         ]
foo(str) -> Callable[
              Optional[Tuple[str, str]]
         ]
foo() -> None
```

Alternatively, multiline signatures can be disallowed/unsupported, but this seems to be unnecessary restriction. What do you think?

AWhetter:
For that to work I think we would have to change the signature parsing of Sphinx so that it matches what stubgen does where it tokenises that string and properly parses the signatures. At the moment Sphinx is doing a basic regex search on each line of the docstring. It's definitely possible.
The change I'm about to make to Sphinx is quite small, but you could submit an issue to Sphinx to see if they would consider a bigger change to the signature parsing?

sizmailov:
The sphinx choice of trailing backslash as indication for multiple signatures become fairly unfortunate one. It might result into a cumbersome signature parsing implementation which supports both "legacy" and "new" versions... but is doable for sure. I may have a too few clock cycles for that in next months, can't promise, sorry :disappointed: 

Probably it's not the issue that requires immediate solution, but still it's worth to think about ahead of time.

AWhetter:
Good news! Sphinx v4 will support the function signatures separated without backslashes.
I've updated this pull request to not include separating backslashes. I've done another small end to end test using an install of Sphinx from its "master" branch and using stubgen from the latest mypy, and both work with the docstrings that pybind outputs from this pull request.
If `disable_section_headings()` is not set, stubgen outputs duplicate type signatures but that's a separate issue.

ka-bo:
Since this may be related: Does this also fix #3037 or is it possible to include a fix?

JeanChristopheMorinPerso:
Hello, just dropping a little message to see if there is still some interest from maintainers into merging this? I'd personally be interested in getting that to use it on a project I contribute to (https://github.com/PixarAnimationStudios/OpenTimelineIO).

Thanks!

rwgk:
This looks useful to me, the change is relatively small, and the added documentation is well done.

Caveat: I don't know much about sphinx and stubgen (don't use it in my own work) and haven't worked directly in or around the changed code in pybind11.h.

My main concern is backward compatibility. Specifically:

> instead of the generic signature that is prepended currently.

How much trouble might this cause? — I honestly don't know, but from general experience, I'd bet some tool chains will break.

What's the extra cost for adding another option, to get back the generic signature? (Seems small.)

What are the arguments for opt-in (old behavior is the default) vs opt-out (new behavior is the default)?

Gut feeling: assuming there will only be a fairly small group of affected users, opt-out would seem like a good choice.

---------------------------

@AWhetter is there a chance that you could rebase this PR (as is)? Then I could run it through our (Google) global testing to see if there are any surprises.

rwgk:
Thanks for updating this PR!

It turns out, this PR breaks one test (out of millions). The breaking test involves this unexpected diff:
```
@@ -6,4 +6,8 @@
     def __init__(self, arg0: int) -> None: ...
     @overload
     def __init__(self, arg0: Span[str]) -> None: ...
+    @overload
+    def __init__(self, arg0: int) -> None: ...
+    @overload
+    def __init__(self, arg0: Span[str]) -> None: ...
     def set_current_measurement(self, value: Span[float], timestamp, sensor_index: int = ...) -> None: ...
```

Apparently the docstrings are mined to build .pyi files, in this case picking up the function signatures twice. By patching `bool show_section_headings = false;` in pybind11/options.h this particular issue is fixed, but there is still another type of failure:

```
-    def get_current(self, arg0: int) -> float: ...
+    def get_current(self, *args, **kwargs) -> Any: ...
```

I'd need to dig in to understand this one, but I'm convinced already, we will force users into difficult situations by changing the docstrings abruptly. I feel this change needs to be introduced gradually, e.g. by adding the new feature as opt-in (old behavior still the default) in the next pybind11 release, then flipping the default to the new behavior a few months later. Maybe the code for the old behavior can be removed completely after a year or two.

I overlooked before: It would be much cleaner to add new tests to test_docstring_options, designed to cover all combinations of options and corner cases, and not to change test_factory_constructors (at least until the new behavior is made the default).

AWhetter:
> Apparently the docstrings are mined to build .pyi files, in this case picking up the function signatures twice.

Is stubgen being used to do this and does each overload have its own docstring? It currently pulls the signature from both the signature list at the top of the docstring and from each of the section headings, but keeps the duplicates (mentioned here originally: https://github.com/pybind/pybind11/pull/2621#issuecomment-813069464).

I don't know what's going on with `get_current` without more information so let me know when you find something.

AWhetter:
As for disabling this behaviour by default, it will require adding another option and I don't love that this pull request would then increase the number of options from two to four. It'll also make what's already a complex part of pybind code even more difficult to understand. To have to do that to protect such a small percentage of users doesn't seem worth it to me.
I'm happy accept whatever you as maintainers think is best but I felt that it was worth raising.

rwgk:
> it will require adding another option and I don't love that this pull request would then increase the number of options from two to four.

I feel the same, but based on past experience working with a huge codebase, and mixed use internal vs OSS, the alternative is far worse. That's the price we pay for success: seemingly simple changes are not actually simple anymore. I think we'll break a fair number of people. The Google global testing is a bit like a canary in the coal mine in this respect. Just for me, navigating a pybind11 upgrade for Google with that one broken test suite will cost me more time than the extra time needed to add the option. If some other breakage pops up later, for example in a OSS rollout a month later, rollbacks will reach nightmare level.

> It'll also make what's already a complex part of pybind code even more difficult to understand.

I don't think it's not that complex, luckily. At least when compared to 10, or 100, or more (nobody knows) people having to deal with complicated rollbacks and roll forwards.

AWhetter:
I've made the requested changes. The test failures look unrelated to my changes.
I did try out using a single new option to switch between the current docstrings, and a docstring with prepended overload signatures and no section heading, but this was harder to explain in the documentation and I can see legitimate reasons for wanting to keep section headings but still prepend the overload signatures to the docstring.
Maybe we can consolidate and clean up the options in the next major version.

rwgk:
I'll close and reopen after a few seconds to trigger a CI rerun, to see if that shakes out the pip errors (I saw them in another PR yesterday but then it worked again a few hours ago).

rwgk:
I glanced back and forth a few times and it looks great to me. (Unfortunately the force pushes make it difficult to re-review. For the future, please just add commits, force-push only after rebase on master. The commits will be squashed when merging.)

I'll run this through the Google global testing again tonight.

AWhetter:
I can separate out my latest changes from the original commits if you like? I wasn't sure if you would squash them or if that was up to me.

rwgk:
No it's fine, no need to spend extra time.
I just submitted the PR for the global testing. If there are no hiccups (rare), results tomorrow morning (Pacific).

rwgk:
Global test results look good! (Sorry I had to spend more time than usual deflaking).

I'm thinking, for completeness it would be ideal to add a 4th new test with just `disable_section_headings()`. What do you think? (We'll also be sure that the line breaks work out as desired.)

AWhetter:
Yes good idea. I'll get this added.

rwgk:
The force push was for `git rebase master`.

AWhetter:
Apologies for taking so long to get to this. It's been a busy week.
I've added the additional test. The CI fail looks unrelated. Maybe there's a flaky test because I think it's the same error message as last time.

rwgk:
Yes, that's a known flake.

I just pushed https://github.com/pybind/pybind11/commit/7e7c558530430615d889b5ed41954a2fd7015a31 directly (not worth a PR), see commit message.

I'll rebase this PR in a sec, which will re-trigger the CI. Chances are the flake will go away.

tests/test_docstring_options.py
```diff
@@ -40,3 +40,51 @@ def test_docstring_options():
...
+        "A function which adds two numbers.\n"
+        "\n"
+        "Internally, a simple addition is performed.\n"
+        "Both numbers can be None, and None will be returned."
+    )
+
+    assert m.test_overloaded7.__doc__ == (
+        "test_overloaded7(*args, **kwargs)\n"
+        "A function which adds two numbers.\n"
+        "\n"
```
rwgk, line 87: A complete nit, only affecting this test: I'd find it slightly easier to reach the "aha" point like this:
```
"A function which adds two numbers.\n\n"
```

rwgk (approved):
Thank you! I hope we can make this the default a few months down the road, but we can leave that question for later.

rwgk:
The force push was for a fresh git rebase master. Trying to get the attention of at least one other maintainer.

rwgk:
The one CI failure is an infrastructure issue (job not even starting up).

henryiii:
Function signatures are processed by mypy; does it handle these (yet)? I'm pretty sure it handles the list at the top, but not so sure about the ones with mixed signatures and descriptions.

henryiii:
The signature stuff has mostly been dormant, since we want to try to keep the size down and avoid a massive list of options or making arbitrary “it looks better” changes. Ideally, in my opinion, there’s a clear forward direction - signatures should match what mypy expects. Type checking is well defined by now, and very useful. And mypy now tests pybind11, we should probably do the same, and run mypy’s stubgen on our development branch.

rwgk:
Henry pointed this out elsewhere: https://github.com/python/mypy/tree/master/test-data/stubgen/pybind11_mypy_demo

Pasting here thinking it may be useful for us exploring how to evolve the docstring generation.

AWhetter:
> Function signatures are processed by mypy; does it handle these

mypy itself does not parse the docstrings. They have to be turned into a stub file first by stubgen. These stub files then get read by mypy.

The current behaviour of pybind and stubgen is as follows.

```cpp
py::class_<Pet>(m, "Pet")
   .def(py::init<const std::string &, int>())
   .def("set", (void (Pet::*)(int)) &Pet::set, "Set the pet's age")
   .def("set", (void (Pet::*)(const std::string &)) &Pet::set, "Set the pet's name");
```

The above pybind class would have the following docstring:

```python
>>> help(example.Pet)

class Pet(__builtin__.object)
 |  Methods defined here:
 |
 |  __init__(...)
 |      __init__(self: Pet, arg0: str, arg1: int) -> None
 |
 |  set(...)
 |      set(*args, **kwargs) -> Any
 |      Overloaded function.
 |
 |      1. set(self: Pet, arg0: int) -> None
 |
 |      Set the pet's age
 |
 |      2. set(self: Pet, arg0: str) -> None
 |
 |      Set the pet's name
```

Stubgen would produce a stub file that looks like the following:

```python
from typing import overload

class Pet:
    def __init__(self: Pet, arg0: str, arg1: int) -> None: ...
    @overload
    def set(self: Pet, arg0: int) -> None: ...
    @overload
    def set(self: Pet, arg0: str) -> None: ...
    @overload
    def set(*args, **kwargs) -> Any: ...
```

As the examples show, the signature at the top of the docstring and the sginatures that are section headings are both picked up.

With this change, pybind will instead produce the following docstring (when configured to do so):

```python
>>> help(example.Pet)

class Pet(__builtin__.object)
 |  Methods defined here:
 |
 |  __init__(...)
 |      __init__(self: Pet, arg0: str, arg1: int) -> None
 |
 |  set(...)
 |      set(self: Pet, arg0: int) -> None
 |      set(self: Pet, arg0: str) -> None
 |      Overloaded function.
 |
 |      1. set(self: Pet, arg0: int) -> None
 |
 |      Set the pet's age
 |
 |      2. set(self: Pet, arg0: str) -> None
 |
 |      Set the pet's name
```

Which stubgen would turn into:

```python
from typing import overload

class Pet:
    def __init__(self: Pet, arg0: str, arg1: int) -> None: ...
    @overload
    def set(self: Pet, arg0: int) -> None: ...
    @overload
    def set(self: Pet, arg0: str) -> None: ...
    @overload
    def set(self: Pet, arg0: int) -> None: ...
    @overload
    def set(self: Pet, arg0: str) -> None: ...
```

So unfortunately it produces duplicate overload signatures and a change would need to be made to stubgen to either have it remove duplicates or have a way for it to read the docstrings from the head of the docstring only. Having an option to stubgen to have it read only header docstrings would probably be preferred because that's then inline with Sphinx's behaviour as well, but we will need to discuss with the mypy maintainers as well to reach a conclusion.

AWhetter:
I've resolved all conflicts and removed additional options.
To cater for the removal of the options, I've made it such that the overload signature are always prepended to the docstring, even if `disable_function_signatures()` has been called. This is closer to the behaviour I'm looking for personally where I don't want the function signatures, partly because they still produce duplicate stubs with stubgen and partly because I want to be able to format the docstring myself, but I still want Sphinx and stubgen to have access to the overload signatures.
The downside of doing it this way is that users no longer have the option to write the signatures themselves.

AWhetter:
Is there anything that I can do to help progress this change?

AWhetter:
Should I address the conflicts? Is there anything else that I should do with this change to help move it forwards?

rwgk:
> Should I address the conflicts?

Yes please.

> Is there anything else that I should do with this change to help move it forwards?

I'll look again later tonight (assuming the conflicts are resolved).

AWhetter:
Thanks for continuing to spend time on this. I've resolved the conflicts. Given that it's been so long since this was submitted I've rebased the commit as well.

rwgk:
@henryiii is there a chance you could take a look?

I believe we should merge this. mypy is only one consumer, and there are issues they haven't responded to for a long time (#4888). I don't think this PR makes anything worse in that respect.

The other consumer are humans. Clearly it's useful to see the signature for all overloads.

rwgk:
@AWhetter @henryiii 

I ran this PR through Google's global testing. It passes after rerunning stubgen for ~20 .pyi files (the generated files are checked in, unfortunately).

A typical .pyi diff is:

```diff
     @overload
     def __setitem__(self, arg0: int, arg1: Program) -> None: ...
     @overload
     def __setitem__(self, arg0: slice, arg1: ProgramList) -> None: ...
+    @overload
+    def __setitem__(self, arg0: int, arg1: Program) -> None: ...
+    @overload
+    def __setitem__(self, arg0: slice, arg1: ProgramList) -> None: ...
```

I.e. stubgen generates the same signature twice; not surprisingly.

I'll have to rerun the global testing with the updated .pyi files. Might take a day to get the results.

rwgk:
> I.e. stubgen generates the same signature twice; not surprisingly.
> 
> I'll have to rerun the global testing with the updated .pyi files. Might take a day to get the results.

Good news, nothing breaks, but seeing that duplication in the stubgen-generated .pyi files got me to change my mind. I no longer think the duplication is useful.

The "generic" signature that is replaced here is far more confusing than helpful.

So I did this, based on this PR: https://github.com/pybind/pybind11/pull/5329

The core diff there is simply:

```diff
diff --git a/include/pybind11/pybind11.h b/include/pybind11/pybind11.h
index 9f356828..27581348 100644
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -613,10 +613,7 @@ protected:
         /* Create a nice pydoc rec including all signatures and
            docstrings of the functions in the overload chain */
         if (chain && options::show_function_signatures()) {
-            // First a generic signature
-            signatures += rec->name;
-            signatures += "(*args, **kwargs)\n";
-            signatures += "Overloaded function.\n\n";
+            signatures += "Overloaded function:\n\n";
         }
         // Then specific overload signatures
         bool first_user_def = true;
```

If have to put that alternative through global testing again.

rwgk:
Global testing for #5329 passed, *without* having to make any changes to existing stubgen-generated .pyi files.

@AWhetter What do you think?

#5329 still is mostly your work. Do you want to update this PR with my changes on top?

AWhetter:
#5329 does not address the original issue on its own. The lack of prepended signatures means that Sphinx is unable to parse and document the signatures. The readability is improved though, and the situation for stubgen is also improved.

> I no longer think the duplication is useful.

Do you mean the duplication in the docstring or the duplication in the stubs generated by stubgen?
If it's the duplication in the docstring then, to be honest, I'm less invested in the "Overloaded function" type of docstring. I'm more interested in getting this type of docstring out:

```
    add(...)
     |      add(arg0: int, arg1: int) -> int
     |      add(arg0: float, arg1: float) -> float
     |      add(arg0: None, arg1: None) -> None
     |      A function which adds two numbers.
     |
     |      Internally, a simple addition is performed.
     |      Both numbers can be None, and None will be returned.
```

So if we need to find a compromise, I'd be happy with the following being the default:

```
    add(...)
     |      Overloaded function:
     |
     |      1. add(arg0: int, arg1: int) -> int
     |
     |      Add two integers together.
     |
     |      2. add(arg0: float, arg1: float) -> float
     |
     |      Add two floating point numbers together.
```

and then there being a pybind option to get the first docstring that I mentioned.
Or you previously mentioned some kind of slow transition to the new docstring format (https://github.com/pybind/pybind11/pull/2621#issuecomment-1003240359).

If the duplication in the stubs is the issue, then I can look at addressing this in stubgen.

rwgk:
Regarding duplication, what I had in mind is this (docs/advanced/misc.rst in this PR):

```
    add(...)
     |      add(arg0: int, arg1: int) -> int
     |      add(arg0: float, arg1: float) -> float
     |      Overloaded function.
     |
     |      1. add(arg0: int, arg1: int) -> int
     |
     |      Add two integers together.
     |
     |      2. add(arg0: float, arg1: float) -> float
     |
     |      Add two floating point numbers together.
```

The signatures appear twice.

For the human readability aspect:

Sometimes, quite commonly, signatures are long and unwieldy. I often give up trying to parse them visually. If they appear twice, I give up even quicker, because it just looks even more like a wall of incomprehensible stuff.

For stubgen: I'm worried tooling may behave weirdly if the signatures are duplicated.

For sphinx: I don't know.

Can we find a way to make sphinx happy without duplicating the signatures?

If not, improving sphinx would seem the way to go.

> If the duplication in the stubs is the issue, then I can look at addressing this in stubgen.

Getting into sphinx seems better in that case.

But, **tangentially**, if you could help making progress with the ideas under #4888, that would be great.

PR #2622, opened 2020-10-26
fix: min macos

### LINKED ISSUE #2620: [BUG] Setup helpers: don't set minimum macOS higher than current macOS
## Issue description

Currently the minimum macOS version is set to 10.14, even if you run on 10.13 or 10.12. While users should set the correct environment variable instead, the default should be smarter and not set a number lower than the current macOS version.

Also see https://github.com/pybind/python_example/issues/44 

## Reproducible example code

Seen in https://github.com/pybind/python_example/pull/64.

henryiii:
## Description

Closes #2620. On macOS 10.13, with automatic discovery on, the C++17 flag would work, and the min macOS version would be set to 10.14!

## Suggested changelog entry:

```rst
* setup_helpers will no longer set a minimum macOS version lower than the current version.
```

### DIFF
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -45,6 +45,7 @@ import shutil
 import sys
 import tempfile
 import threading
+import platform
 import warnings
 
 try:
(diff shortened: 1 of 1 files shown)


bstaletic (approved):

rwgk (approved):
Looks good to me. I tried out the commands interactively and it works fine on my 10.15 system.

PR #2625, opened 2020-10-27
[common.h] Mark another entry point as "unused".

tkoeppe:
For rationale, see #2241, eeb1044; this annotates a newly added variable as unused.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -311,7 +311,7 @@ extern "C" {
 \endrst */
 #define PYBIND11_MODULE(name, variable)                                        \
     static ::pybind11::module_::module_def                                     \
-        PYBIND11_CONCAT(pybind11_module_def_, name);                           \
+        PYBIND11_CONCAT(pybind11_module_def_, name) PYBIND11_MAYBE_UNUSED;     \
     PYBIND11_MAYBE_UNUSED                                                      \
     static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);  \
     PYBIND11_PLUGIN_IMPL(name) {                                               \


henryiii (approved):

YannickJadoul:
When is this necessary/detected as unused? Isn't it always, unconditionally used here? https://github.com/pybind/pybind11/blob/ace4deb4f09f172e3f0cd622774a85de3874463f/include/pybind11/detail/common.h#L320-L322

tkoeppe:
@YannickJadoul: please see the linked previous rationale: The annotation assists dead code analysis tools. Since these variables and functions are FFI entry points, there is no naive C++ code that references them, which makes some dead-code detectors consider the entire set of Python binding code unused. By annotating the code as "maybe unused", these symbols become "root nodes" or "entry points" in the analysis. The attribute really means "assume that this might be used"; the words "maybe unused" are short for "may look like it's unused".

YannickJadoul:
@tkoeppe Yes, of course, I understood that rationale for `static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);`.
But the newly annotated `PYBIND11_CONCAT(pybind11_module_def_, name)` *is* used in other code that's been annotated as maybe unused (and thus becomes '"root nodes" or "entry points" in the analysis', in your words). So my question was: isn't that good enough, that the variable is used by some code that's "not dead"?

Or in other words: I'd be curious to know whether this is preemptive, or whether you saw a static analyzer get confused/falsely flag this as "unused".

YannickJadoul:
Also, just to double-check: there are very similar macros in `embed.h`, but they are referenced at compile time (and thus don't need this), right?

tkoeppe:
@YannickJadoul: ah, I see, sorry! That's a good question, it does indeed appear that the `PLUGIN_IMPL` expands to something that is itself marked maybe-unused.

This is not preemtive, I started seeing new dead-code findings after a recent update of the pybind11 code. But I can tell you whether this change is effective once I have a new analysis result, so we can definitely wait for that!

I hadn't noticed embed.h before: it's perfectly possible that that code should also be annotated, it just so happens that we don't seem to be using it. I can't say for sure! Generally, anything that has external linkage, is exported, and isn't called from other C++ code should probably receive a maybe-unused attibute; I haven't checked whether that code is has internal references.

Thanks!

tkoeppe:
Actually, let me double-check this: the `pybind11_module_def_*` variables are `static`, so the problem is not that they're not roots. Something else is amiss. I'll get back to you!

tkoeppe:
Ah, but no, the `pybind11_init_*` function is also `static` and also needed to be annotated. So, I think (for now) that the change will work, but the explanation needs a change. We seem to be marking these static variables as a kind of workaround instead of marking an actual entry point. I'm not quite sure now why I did it that way; I was basically going directly for the symbol that was reported dead, rather than look for a higher-up function that should instead be annotated. So the same is true of this new variable in this PR.

YannickJadoul:
@tkoeppe OK, thanks for the explanation! If there's a problem, then I'm fine just merging this! Never meant to block this PR in anyway, but it got me riddled :-)

If you think there's a better "root" to annotate (that might enable us to remove some of the other `PYBIND11_MAYBE_UNUSED` uses?), that's also fine, but let's not spend too much energy on this either. (I would/could give this a try myself, but I don't really know which tool I should use to reproduce.)

YannickJadoul (approved):

YannickJadoul:
Also, these macros are a mess for syntax highlighting and making sense of the structure, and then some identifiers are constructed by concatenation, so I'm always really confused as well, when reading this...

tkoeppe:
Thanks a lot! Do let's wait till tomorrow so I can confirm, but I appreciate the review! I can try again and see if I can find a better place for the annotation, but perhaps we can indeed leave that for a separate change. Thanks!

tkoeppe:
I can confirm that this change is effective; our new round of dead-code analysis following my local patch of this change no longer considers any pybind11-bindings dead.

YannickJadoul:
@tkoeppe, OK, great :-) So, as far as you're concerned, we can just merge this PR as is?

> (I haven't yet found a better root, though.)

The actual root should be `PyObject *PyInit_##name()` (for Python 3; `PYBIND11_EXPORT void init##name()` for Python 2); see [the Python docs](https://docs.python.org/3/extending/extending.html#the-module-s-method-table-and-initialization-function) (or [the Python 2 docs](https://docs.python.org/2/extending/extending.html#the-module-s-method-table-and-initialization-function)). This is marked as `MAYBE_UNUSED`, but only the forward declaration is, and not the actual implementation of that function in the macro. So maybe something more is going on in the static code analyzer?

tkoeppe:
> @tkoeppe, OK, great :-) So, as far as you're concerned, we can just merge this PR as is?

Yes please, I'd welcome that :-)

> > (I haven't yet found a better root, though.)
> 
> The actual root should be `PyObject *PyInit_##name()` (for Python 3; `PYBIND11_EXPORT void init##name()` for Python 2); see [the Python docs](https://docs.python.org/3/extending/extending.html#the-module-s-method-table-and-initialization-function) (or [the Python 2 docs](https://docs.python.org/2/extending/extending.html#the-module-s-method-table-and-initialization-function)). This is marked as `MAYBE_UNUSED`, but only the forward declaration is, and not the actual implementation of that function in the macro. So maybe something more is going on in the static code analyzer?

So, yes, I noticed that, I think I installed that annotation last time. And I think function attributes are only needed on one declaration, not on all of them. I've added unused-attributes to extern-C entry points in our codebase in many places, often only on the non-defining declaration in the header, and that worked. I'm not sure why in this case it isn't enough.

I'm trying to see if I can run the analysis locally, quickly, so that I can experiment with this and discover the better root.

YannickJadoul:
> So, yes, I noticed that, I think I installed that annotation last time. And I think function attributes are only needed on one declaration, not on all of them. I've added unused-attributes to extern-C entry points in our codebase in many places, often only on the non-defining declaration in the header, and that worked. I'm not sure why in this case it isn't enough.
> 
> I'm trying to see if I can run the analysis locally, quickly, so that I can experiment with this and discover the better root.

OK. Well, if you do find something out, a new PR cleaning this up is welcome. But don't waste too much time on it. After all, this is already pretty clean!

YannickJadoul:
Thanks for your effort, btw, @tkoeppe!

tkoeppe:
It's my pleasure, thank you!

PR #2626, opened 2020-10-28
clang -Wnon-virtual-dtor compatibility

rwgk:
## Description

Adding missing virtual destructors to silence `clang -Wnon-virtual-dtor` warnings.

Tested with clang version 9.0.1-12 under an Ubuntu-like OS.
Originally discovered in the Google-internal environment.

## Suggested changelog entry:

```rst
`clang -Wnon-virtual-dtor` compatibility.
```

### DIFF
--- a/tests/test_callbacks.cpp
+++ b/tests/test_callbacks.cpp
@@ -117,7 +117,11 @@ TEST_SUBMODULE(callbacks, m) {
         }
     });
 
-    class AbstractBase { public: virtual unsigned int func() = 0; };
+    class AbstractBase {
+    public:
+        virtual ~AbstractBase() = default;
+        virtual unsigned int func() = 0;
+    };
     m.def("func_accepting_func_accepting_base", [](std::function<double(AbstractBase&)>) { });
 
     struct MovableObject {
--- a/tests/test_exceptions.cpp
+++ b/tests/test_exceptions.cpp
@@ -32,6 +32,13 @@ class MyException3 {
 public:
     explicit MyException3(const char * m) : message{m} {}
     virtual const char * what() const noexcept {return message.c_str();}
+    // Rule of 5 BEGIN: to preempt compiler warnings.
+    MyException3(const MyException3&) = default;
+    MyException3(MyException3&&) = default;
+    MyException3& operator=(const MyException3&) = default;
+    MyException3& operator=(MyException3&&) = default;
+    virtual ~MyException3() = default;
+    // Rule of 5 END.
 private:
     std::string message = "";
 };
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -243,8 +243,15 @@ function(pybind11_enable_warnings target_name)
   if(MSVC)
     target_compile_options(${target_name} PRIVATE /W4)
   elseif(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Intel|Clang)" AND NOT PYBIND11_CUDA_TESTS)
-    target_compile_options(${target_name} PRIVATE -Wall -Wextra -Wconversion -Wcast-qual
-                                                  -Wdeprecated -Wundef)
+    target_compile_options(
+      ${target_name}
+      PRIVATE -Wall
+              -Wextra
+              -Wconversion
+              -Wcast-qual
+              -Wdeprecated
+              -Wundef
+              -Wnon-virtual-dtor)
   endif()
 
   if(PYBIND11_WERROR)


rwgk:
@henryiii Is there a good way to add CI testing with `-Wnon-virtual-dtor`?

henryiii:
See #2617 for an example (or an all-at-once attempt).

Looks like you need to follow the rule of 5 if you add the virtual destructor.

rwgk:
> See #2617 for an example (or an all-at-once attempt).
> 
> Looks like you need to follow the rule of 5 if you add the virtual destructor.

Thanks Henry! I don't have a way to reproduce the failures interactively. I just force-pushed adding the copy constructors only. Let's see ...

rwgk:
Just adding the copy constructors seems to do the trick, the CI is green :-)
But I will mark this PR as Draft for the minute, I will try to add in the `-Wnon-virtual-dtor` option for at least one build tomorrow.

tests/test_callbacks.cpp
```diff
@@ -117,7 +117,12 @@ TEST_SUBMODULE(callbacks, m) {
         }
     });
 
-    class AbstractBase { public: virtual unsigned int func() = 0; };
+    class AbstractBase {
+    public:
+        AbstractBase(const AbstractBase&) = default;
```
YannickJadoul, line 122: Wait, you can't construct and abstract base class. Can you copy-construct it? And why would you need to?

YannickJadoul (approved):
Apart from figuring out why a copy constructor is needed for `AbstractBase` (`MyException` is fine; that's not an abstract base class), and checking why the just the copy-constructor and not the others, this looks good to me :-)

tests/test_callbacks.cpp
bstaletic: Slice a non-abstract "leaf" class and you get an instance of an abstract base class. You just can't construct it directly, but copying definitely works.

However, I also wonder why would a copy constructor be needed.

bstaletic (commented):
I would have moaned about actually following the rule of 5, but this is only a test class, so I guess it's fine

tests/test_callbacks.cpp
henryiii: What about `AbstractBase(const AbstractBase&) = delete;`?

rwgk:
> However, I also wonder why would a copy constructor be needed.

Good catch, thanks guys, I will try again without now. Possibly I got confused between the two cases, maybe there was an error only for the other one.

rwgk:
> > However, I also wonder why would a copy constructor be needed.
> 
> Good catch, thanks guys, I will try again without now. Possibly I got confused between the two cases, maybe there was an error only for the other one.

Indeed, the copy constructor isn't needed for the abstract class: CI is green without.

I'm testing another change, fully implementing the "Rule of 5" for `MyException3`, with surrounding `BEGIN` & `END` comments to explain that this is just to preempt compiler warnings.

tests/test_exceptions.cpp
```diff
@@ -32,6 +32,13 @@ class MyException3 {
 public:
     explicit MyException3(const char * m) : message{m} {}
     virtual const char * what() const noexcept {return message.c_str();}
+    // Rule of 5 BEGIN: to preempt compiler warnings.
+    MyException3(const MyException3&) = default;
+    MyException3(MyException3&& other) = delete;
+    virtual MyException3& operator=(const MyException3& other) = delete;
```
YannickJadoul, line 38: Should this be virtual?

YannickJadoul: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-assignment says "no"

YannickJadoul: Also, why delete copy-assignment, while you can copy-construct?

But as @bstaletic says, it's just tests. So I'm fine with just adding the copy-ctr to make the compiler happy, and forget about everything else. (Same with the comments; this is just good practice, so the comments feel a bit like overkill, personally.)

YannickJadoul:
Sorry, @rwgk. This very quickly slided into bikeshedding something that doesn't matter. Already approved, though, so as you see fit :-)

tests/test_exceptions.cpp
rwgk: > Should this be virtual?

In the face of ambiguity ... dunno! :-)
I'm using trial-and-error here to see what works across our ~60 CI actions.
Currently I'm experimenting with `default` vs `delete`, aiming to minimize `default`.
Recommendations appreciated, I'll try them out.

rwgk: > https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-assignment says "no"

Thanks!
For completeness:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c63-make-move-assignment-non-virtual-take-the-parameter-by--and-return-by-non-const-

YannickJadoul:
Also fine. So they need to be movable?

rwgk:
> Also fine. So they need to be movable?

Yes, at least for this run: https://github.com/pybind/pybind11/runs/1321231557

```
/Users/runner/work/pybind11/pybind11/tests/test_exceptions.cpp:140:35: error: call to deleted constructor of 'MyException3'
    m.def("throws3", []() { throw MyException3("this error cannot be translated"); });
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/pybind11/pybind11/tests/test_exceptions.cpp:37:5: note: 'MyException3' has been explicitly marked deleted here
    MyException3(MyException3&& other) = delete;
    ^
1 error generated.
```

YannickJadoul:
> Yes, at least for this run: https://github.com/pybind/pybind11/runs/1321231557

OK, nice to know! :-)

YannickJadoul:
MSVC is not a fan of move-only exceptions :-( https://quuxplusone.github.io/blog/2019/05/11/msvc-what-are-you-doing/
Better make them all `= default` and get this over with?

rwgk:
> MSVC is not a fan of move-only exceptions :-( https://quuxplusone.github.io/blog/2019/05/11/msvc-what-are-you-doing/
> Better make them all `= default` and get this over with?

Thanks for the link, awesome! I just made them all `default`. Re-testing.

Next up: inserting `-Wnon-virtual-dtor` in `tests/CMakeLists.txt` somehow.

YannickJadoul:
Something went wrong during these runs. I'm cancelling and restarting.

YannickJadoul:
(for the record: it seems to have been some kind of upstream fluke with setting up CMake, btw; nothing to be worried about in the context of this PR, I think ;-) )

rwgk:
One Windows failure:
https://github.com/pybind/pybind11/pull/2626/checks?check_run_id=1321724701
I'm assuming that's a fluke, too.

henryiii:
It's a known fluke. Would love to solve it, have tried before. But for now, ignore it.

rwgk:
@henryiii A little bit to my surprise, adding the option for all GNU|Intel|Clang works!

bstaletic (approved):

henryiii (approved):

PR #2629, opened 2020-10-29
Allow NULL value in pybind11_meta_setattro

### LINKED ISSUE #2009: Calling del Class.field from python, when its a def_readonly_static segfaults
## Issue description

def_readonly_static defined fields segfault when used with python delattr, Python as a safe environment it would be expected to instead see a TypeError

## Reproducible example code
```cpp
class Test {
  public:
     static const int field;
};

const int Test::field = 10;

PYBIND11_MODULE(example, m) {
   py::class_<Test>(m, "Test")
       .def_readonly_static("field", &Test::field);
}

```

```python
from example import Test
del Test.field  # BOOM
```

The relevant segfault st

```
     null-deref
     #0 0x7fe1836b55f4 in PyObject_IsInstance /home/engshare/third-party2/python/3.7/src/cpython-3.7/Objects/abstract.c:2379
     #1 0x7fe172d0634a in pybind11_meta_setattro third-party-buck/platform007/build/pybind11/f3b1ccc/include/pybind11/detail/class.h:110
```

bstaletic:
## Description

Python says we have to handle `NULL` values in `pybind11_meta_setattro` as "delete property".

https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_setattro

However, pybind11 currently doesn't allow users to `del`ete non-static properties. I'm not really sure what's the right course of action, this PR is just the first thing that worked, in the sense of not segfaulting as in #2009.

Fixes #2009

## Suggested changelog entry:

```rst
Allows deleting of static properties.

```

### DIFF
--- a/tests/test_methods_and_attributes.py
+++ b/tests/test_methods_and_attributes.py
@@ -171,6 +171,19 @@ def test_static_properties():
     assert m.TestPropertiesOverride().def_readonly == 99
     assert m.TestPropertiesOverride.def_readonly_static == 99
 
+    # Only static attributes can be deleted
+    del m.TestPropertiesOverride.def_readonly_static
+    assert (
+        hasattr(m.TestPropertiesOverride, "def_readonly_static")
+        and m.TestPropertiesOverride.def_readonly_static
+        is m.TestProperties.def_readonly_static
+    )
+    assert "def_readonly_static" not in m.TestPropertiesOverride.__dict__
+    properties_override = m.TestPropertiesOverride()
+    with pytest.raises(AttributeError) as excinfo:
+        del properties_override.def_readonly
+    assert "can't delete attribute" in str(excinfo.value)
+
 
 def test_static_cls():
     """Static property getter and setters expect the type object as the their only argument"""
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -129,7 +129,7 @@ extern "C" inline int pybind11_meta_setattro(PyObject* obj, PyObject* name, PyOb
     //   2. `Type.static_prop = other_static_prop` --> setattro:  replace existing `static_prop`
     //   3. `Type.regular_attribute = value`       --> setattro:  regular attribute assignment
     const auto static_prop = (PyObject *) get_internals().static_property_type;
-    const auto call_descr_set = descr && PyObject_IsInstance(descr, static_prop)
+    const auto call_descr_set = descr && value && PyObject_IsInstance(descr, static_prop)
                                 && !PyObject_IsInstance(value, static_prop);
     if (call_descr_set) {
         // Call `static_property.__set__()` instead of replacing the `static_property`.


tests/test_methods_and_attributes.py
```diff
@@ -171,6 +171,13 @@ def test_static_properties():
     assert m.TestPropertiesOverride().def_readonly == 99
     assert m.TestPropertiesOverride.def_readonly_static == 99
 
+    # Static attributes can be deleted
+    del m.TestPropertiesOverride.def_readonly_static
+    assert !hasattr( m.TestPropertiesOverride, 'def_readonly_static' )
```
henryiii, line 176:
```suggestion
    assert not hasattr(m.TestPropertiesOverride, "def_readonly_static")
```

tests/test_methods_and_attributes.py
```diff
@@ -171,6 +171,13 @@ def test_static_properties():
...
+    properties_override = m.TestPropertiesOverride()
+    del properties_override.def_readonly
+    assert !hasattr( properties_override, 'def_readonly' )
```
henryiii, line 179:
```suggestion
    assert not hasattr(properties_override, "def_readonly")
```

tests/test_methods_and_attributes.py
henryiii: Wrong language. ;)

bstaletic: Not my fault python is weird! Done.

tests/test_methods_and_attributes.py
```diff
@@ -171,6 +171,13 @@ def test_static_properties():
     assert m.TestPropertiesOverride().def_readonly == 99
     assert m.TestPropertiesOverride.def_readonly_static == 99
 
+    # Static attributes can be deleted
+    del m.TestPropertiesOverride.def_readonly_static
+    assert not hasattr(m.TestPropertiesOverride, "def_readonly_static")
```
bstaletic, line 176: So this check fails because:

```
>>> import pybind11_tests
>>> m = pybind11_tests.methods_and_attributes
>>> m.TestPropertiesOverride.def_readonly_static
99
>>> del m.TestPropertiesOverride.def_readonly_static
>>> hasattr(m.TestPropertiesOverride, 'def_readonly_static')
True
>>> m.TestPropertiesOverride.def_readonly_static
1
>>> del m.TestPropertiesOverride.def_readonly_static
AttributeError: def_readonly_static
```

Can anyone make sense of it?

YannickJadoul: The latter value, 1, seems to come from `m.TestProperties`, a base class of `m.TestPropertiesOverride`?

bstaletic: That's right.

```
>>> m = pybind11_tests.methods_and_attributes
>>> del m.TestPropertiesOverride.def_readonly_static
>>> m.TestPropertiesOverride.def_readonly_static
1
>>> del m.TestProperties.def_readonly_static
>>> m.TestPropertiesOverride.def_readonly_static
AttributeError: type object 'pybind11_tests.methods_and_attributes.TestProperti' has no attribute 'def_readonly_static'
```

This makes me even less sure if this is the right course of action.

YannickJadoul: Can we see/mimic how this would react in Python?

YannickJadoul: Just tried this out:

```pycon
>>> class Base:
...     answer = 6 * 9
... 
>>> class Derived(Base):
...     answer = 42
... 
>>> Derived.answer
42
>>> del Derived.answer
>>> Derived.answer
54
>>> del Derived.answer
AttributeError: answer
>>> del Base.answer
>>> Derived.answer
AttributeError: type object 'Derived' has no attribute 'answer'
>>> 
```

So even without any properties, static attributes in pure Python would match our behavior, no?

So ... not a bug, just counter-intuitive Python behaviour?

YannickJadoul, line 176:
```
assert hasattr(m.TestPropertiesOverride, "def_readonly_static") and m.TestPropertiesOverride.def_readonly_static is  m.TestProperties.def_readonly_static
assert "def_readonly_static" not in m.TestPropertiesOverride.__dict__
```

Or something like this?

YannickJadoul (approved):
Unless there's a reason to match non-static properties (or something that needs to be kept alive or so), I'm fine with this. It matches pure Python behavior, and as far as I know, that's what we most of the time aim for.

henryiii:
2.6.1 or 2.7?

henryiii:
That was fast... :)

YannickJadoul:
> That was fast... :)

Hahahaha, I was just thinking I missed a PR or a few somewhere, and was already looking around ;-)
Do we need all PRs to be added to a milestone? I still haven't heard how we're going to manage versions.

tests/test_methods_and_attributes.py
```diff
@@ -171,6 +171,19 @@ def test_static_properties():
     assert m.TestPropertiesOverride().def_readonly == 99
     assert m.TestPropertiesOverride.def_readonly_static == 99
 
+    # Static attributes can be deleted
```
rwgk, line 174: `# Only static attributes can be deleted`

(When I look at this a "but" or "only" is clearly missing, somewhere.)

rwgk (approved):
My vote is to include this one in 2.6.1.

henryiii (approved):

YannickJadoul:
Merged! Good catch, @fried, and thanks for fixing, @bstaletic!

Issue #2578, opened 2020-10-12
[CI] Add valgrind run to CI test suite


YannickJadoul:
Reminder to work on adding a (or multiple?) valgrind run to the CI, to catch memory issues sooner. Potentially, we could think about some sanitizers as well?

bstaletic:
`python -c 'import pybind11_tests'` leaks this thing: https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1320

PR #2634, opened 2020-11-01
Use weakref to clean up captured function object in def_buffer

YannickJadoul:
## Description

Leak discovered independently by @bstaletic (see https://github.com/pybind/pybind11/issues/2578#issuecomment-718913643) and @rwgk. This is one solution that cleans up after itself.

A few things to be considered:

- I would think this is safe, and the order of destruction with the `class_` object shouldn't matter? (cfr. #856) But I wouldn't mind a few people sanity-checking this, please :-)

- Another solution, suggested by @rwgk, would be to delete the `capture` as part of destructing `detail::type_info`. In order to do so, we would need some extra type info (since the `class_<T, ...>::get_buffer<F>::capture *` gets type-erased to `void *`); one solution would be to use a `shared_ptr<void>` (which remembers its deleter), or similar. Is this a better solution?

- If we do consider the solution above as better: this does require increasing the ABI version counter. Do we want to do this now, or go with the previous solution until we have a few ABI-breaking changes together and do them all at once? (Meanwhile, we could create a list of things we'd like to do that would break the ABI.)

## Suggested changelog entry:

```rst
Seal a leak in `def_buffer`, cleaning up the `capture` object after the `class_` object goes out of scope.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1315,7 +1315,8 @@ public:
         return *this;
     }
 
-    template <typename Func> class_& def_buffer(Func &&func) {
+    template <typename Func>
+    class_& def_buffer(Func &&func) {
         struct capture { Func func; };
         auto *ptr = new capture { std::forward<Func>(func) };
         install_buffer_funcs([](PyObject *obj, void *ptr) -> buffer_info* {
@@ -1324,6 +1325,10 @@ public:
                 return nullptr;
             return new buffer_info(((capture *) ptr)->func(caster));
         }, ptr);
+        weakref(m_ptr, cpp_function([ptr](handle wr) {
+            delete ptr;
+            wr.dec_ref();
+        })).release();
         return *this;
     }
 


rwgk (approved):
I ran this through the Google-internal global testing and didn't find any issues.
As discussed on Slack, this PR fixes ASAN errors.
My vote is to merge this PR as-is as soon as possible, and include it in 2.6.1.

henryiii (approved):
I agree with @rwgk

YannickJadoul:
Well, there's no reason to not include it in 2.6.1, so since this was a fix, that was already implicit, to me.

I would want to get those two other questions discussed/answered, though, before merging this.

henryiii:
We will not bump the ABI counter until at least 2.7, and maybe a future version (especially if 2.7 focuses on #2445). We should probably combine several changes then - I have seen _at least_ one PR with "let's add this when we bump the ABI counter" (which should probably be a label).

YannickJadoul:
> We will not bump the ABI counter until at least 2.7

But would we want to, in this case? I.e., should we add a label to this PR/add this to a list somewhere?

bstaletic:
I think this would be a case where C++ smart pointers would get in the way, because you're passing a `void*` to python. The weakref solution just seems more elegant.

rwgk:
> I think this would be a case where C++ smart pointers would get in the way, because you're passing a `void*` to python. The weakref solution just seems more elegant.

Just for completeness:

https://www.nextptr.com/tutorial/ta1227747841/the-stdshared_ptrvoid-as-arbitrary-userdata-pointer

seems like a perfect fit for this situation, except it means we'd need to bump ABI version counter.

My thinking: Yannick's solution is "fine forever". It solves the problem without the need to mess with the ABI version counter. IIUC the solution relies on more moving parts than the `shared_ptr<void>` alternative, but in the unlikely event that we have something like https://github.com/python/cpython/pull/22670, we can just push for pre-release fixes in new Python versions. I wouldn't revisit this leak unless we really run out of other important things to do.

YannickJadoul:
> I think this would be a case where C++ smart pointers would get in the way

We wouldn't per se need `shared_ptr`. We could also store a `void (*deleter)(void *);` in `detail::type_info` and use it when cleaning up.

But yes, I agree it's probably OK to keep it like this.

> but in the unlikely event that we have something like [python/cpython#22670](https://github.com/python/cpython/pull/22670), we can just push for pre-release fixes in new Python versions.

For the record: this is not exactly the ordering I was referencing (but the thing that when multiple things get garbage-collected at the same time, the order of destruction is undefined; I guess we should be fine, though). But yes, agreed.

include/pybind11/pybind11.h
```diff
@@ -1324,6 +1325,10 @@ class class_ : public detail::generic_type {
                 return nullptr;
             return new buffer_info(((capture *) ptr)->func(caster));
         }, ptr);
+        weakref(m_ptr, cpp_function([ptr](handle wr) {
+            delete ptr;
+            wr.dec_ref();
+        })).release();
```
YannickJadoul, line 1331: I'm thinking we could still refactor this to something like

```
template <typename F>
void cleanup_through_weakref(handle parent, F &&f) {
    weakref(parent, cpp_function([f=std::forward<F>(f)](handle wr) {
        f();
        wr.dec_ref();
    })).release();
};
```

But we'd also need some workaround for C++11's inability to capture-by-move. So let's maybe leave this for a next PR, if this pattern keeps coming back.

YannickJadoul:
Right, two approvals, a non-complaining @bstaletic, and a satisfied valgrind/ASAN/MSAN. Should be good enough to merge :-)

YannickJadoul:
Thanks, all!

PR #2636, opened 2020-11-02
fix: Python include directory was missing from DIRS

### LINKED ISSUE #2632: [BUG] find_package(pybind11) does not set up python header path properly
## Issue description

Pybind11 2.6.0 (build from source by us)

We hit this issue when building [onnx 1.7.0](https://github.com/onnx/onnx/tree/v1.7.0):

```
FAILED: CMakeFiles/onnx_cpp2py_export.dir/onnx/cpp2py_export.cc.o
ccache /usr/bin/g++-8 -DONNX_ML=1 -DONNX_NAMESPACE=onnx -Donnx_cpp2py_export_EXPORTS -I../ -I. -fdebug-prefix-map='/tmp/scratch'='/usr/local/src' -g -Wnon-virtual-dtor -O3 -DNDEBUG -fPIC -fvisibility=hidden -std=gnu++11 -MD -MT CMakeFiles/onnx_cpp2py_export.dir/onnx/cpp2py_export.cc.o -MF CMakeFiles/onnx_$pp2py_export.dir/onnx/cpp2py_export.cc.o.d -o CMakeFiles/onnx_cpp2py_export.dir/onnx/cpp2py_export.cc.o -c ../onnx/cpp2py_export.cc
In file included from /usr/local/include/pybind11/pytypes.h:12,
                 from /usr/local/include/pybind11/cast.h:13,
                 from /usr/local/include/pybind11/attr.h:13,
                 from /usr/local/include/pybind11/pybind11.h:45,
                 from ../onnx/cpp2py_export.cc:1:
/usr/local/include/pybind11/detail/common.h:122:10: fatal error: Python.h: No such file or directory
 #include <Python.h>
          ^~~~~~~~~~
compilation terminated.
```

image

Pybind11 is introduced by the following cmake code:
https://github.com/onnx/onnx/blob/0c070abb0c40fec649f81a73a75b0098662ec486/CMakeLists.txt#L401-L413

The same script worked before and there's no recent release from onnx, so I think it should be a regression related to pybind11 repo.

henryiii:
## Description

The `PYBIND11_INCLUDE_DIRS` variable was not being set up correctly. Closes #2632.

## Suggested changelog entry:

```rst
* ``pybind11_INCLUDE_DIRS`` was incorrect, potentially causing a regression if
  it was expected to include ``PYTHON_INCLUDE_DIRS`` (please use targets instead)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,10 +142,17 @@ endif()
 string(REPLACE "include/" "${CMAKE_CURRENT_SOURCE_DIR}/include/" PYBIND11_HEADERS
                "${PYBIND11_HEADERS}")
 
-# Cache variables so pybind11_add_module can be used in parent projects
-set(PYBIND11_INCLUDE_DIR
+# Cache variable so this can be used in parent projects
+set(pybind11_INCLUDE_DIR
     "${CMAKE_CURRENT_LIST_DIR}/include"
-    CACHE INTERNAL "")
+    CACHE INTERNAL "Directory where pybind11 headers are located")
+
+# Backward compatible variable for add_subdirectory mode
+if(NOT PYBIND11_MASTER_PROJECT)
+  set(PYBIND11_INCLUDE_DIR
+      "${pybind11_INCLUDE_DIR}"
+      CACHE INTERNAL "")
+endif()
 
 # Note: when creating targets, you cannot use if statements at configure time -
 # you need generator expressions, because those will be placed in the target file.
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -142,9 +142,14 @@ endif()
 string(REPLACE "include/" "${CMAKE_CURRENT_SOURCE_DIR}/include/" PYBIND11_HEADERS
                "${PYBIND11_HEADERS}")
 
-# Cache variables so pybind11_add_module can be used in parent projects
-set(PYBIND11_INCLUDE_DIR
+# Cache variable so this can be used in parent projects
+set(pybind11_INCLUDE_DIR
```
YannickJadoul, line 146: Oh, is the convention of the new CMake system from 2.6.0 to use `pybind11` instead of `PYBIND11`? I hadn't noticed, really.
Or is the an ad-hoc workaround to have 2 versions of the variable?

henryiii: Config mode requires `pybind11_` versions. If you write `find_package(pybind11`, then you have to have `pybind11_*` variables like `pybind11_FOUND`. Otherwise, you'd have to have `find_package(PYBIND11`, which we don't provide (and you can't provide both on Windows, even if you wanted to). The old system had `PYBIND11_*` variables for subdirectory mode only - we have standardized on using the Config mode variable names when they overlap rather than having to change this based on which mode you used. This is provided for backward compatibility only.

Internally we were still using `PYBIND11_INCLUDE_DIR` and then setting `pybind11_INCLUDE_DIR` at the end - it made more sense to use the proper variable everywhere and just provide `PYBIND11_INCLUDE_DIR` in subdirectory mode.

YannickJadoul: Right, OK, that makes sense, thanks! :-) I was just thinking that I was now using `PYBIND11_FINDPYTON=ON` etc...

YannickJadoul (approved):

CMakeLists.txt
henryiii: Yes, variables you set are `PYBIND11_*`. Variables you read are `pybind11_*`.

YannickJadoul: OK, another convention I just learned about. Thanks! :-)

PR #2637, opened 2020-11-03
Add argument names to enum_ methods

### LINKED ISSUE #2623: [QUESTION] For enums, how to change the docstring for __init__()?
Yesterday I had chance to discuss my question with @YannickJadoul and @bstaletic in gitter, starting here: https://gitter.im/pybind/Lobby?at=5f97172661007f7d1b979a2b. It was big fun and very fruitful! nevertheless I want to report the issue here, and summarize the workaround we found.

I have a simple enumeration with a standard binding:

```c++
  enum class Hand {LEFT, RIGHT};

  py::enum_<Hand> hand{parent, "Hand", "Hand enumeration."};
  hand.value("LEFT", Hand::LEFT, "Left hand enumeration value.");
  hand.value("RIGHT", Hand::RIGHT, "Right hand enumeration value.");
```

The docstring looks like this now:

```
Docstring:     
Hand enumeration.

Members:
  LEFT : Left hand enumeration value.
  RIGHT : Right hand enumeration value.

Init docstring: __init__(self: example.Hand, arg0: int) -> None
```

I don't like the `arg0: int` part, therefore I want to change this, to be consistent with the rest of my bindings, where I give every arguiment a name. My first, naive try:

```c++
  hand.def(py::init<long>(), py::arg("value"), "Default initializer.");
```

This doesn't compile. I started a chat on gitter with this, and we came to the following workaround:

```c++
  py::delattr(hand, "__init__");
  hand.def(
    py::init([] (std::underlying_type_t<Hand> value) { return static_cast<Hand>(value); }),
    py::arg("value"), 
    "Default Initializer."
  );
```

Now, everything looks like I wanted. And hopefully, no negative side-effects. But this is more a hack than a clean solution, right?

The question is now: would it be possible to implement a way for docstring customization for enums (and maybe other predefined components, methods, functions, etc)?

And as @YannickJadoul mentioned in the chat, maybe the default names for some method arguments should be changes to be more readable (`value` for enums init, `other` for `__eq__` and so on)?

I know this is not very important, more the icing on the cake, but it would be nice to have a comlpete, clean, and documented way to accomplish good names in all docstrings.

YannickJadoul:
## Description

Add generic names to the methods added by `enum_`, to avoid autogenerated `arg0`, etc. names.

This is not a full solution to #2623, but I think it's the best we can easily do? Taking extra parameters seems reasonably over the top, and I can't immediately come up with a better solution. Suggestions welcome, though.

Closes #2623

## Suggested changelog entry:

```rst
Added parameter names to the ``py::enum_`` constructor and methods, avoiding ``arg0`` in the generated docstrings.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1628,7 +1628,7 @@ struct enum_base {
                         strict_behavior;                                               \
                     return expr;                                                       \
                 },                                                                     \
-                name(op), is_method(m_base))
+                name(op), is_method(m_base), arg("other"))
 
         #define PYBIND11_ENUM_OP_CONV(op, expr)                                        \
             m_base.attr(op) = cpp_function(                                            \
@@ -1636,7 +1636,7 @@ struct enum_base {
                     int_ a(a_), b(b_);                                                 \
                     return expr;                                                       \
                 },                                                                     \
-                name(op), is_method(m_base))
+                name(op), is_method(m_base), arg("other"))
 
         #define PYBIND11_ENUM_OP_CONV_LHS(op, expr)                                    \
             m_base.attr(op) = cpp_function(                                            \
@@ -1644,7 +1644,7 @@ struct enum_base {
                     int_ a(a_);                                                        \
                     return expr;                                                       \
                 },                                                                     \
-                name(op), is_method(m_base))
+                name(op), is_method(m_base), arg("other"))
 
         if (is_convertible) {
             PYBIND11_ENUM_OP_CONV_LHS("__eq__", !b.is_none() &&  a.equal(b));
@@ -1730,7 +1730,7 @@ public:
         constexpr bool is_convertible = std::is_convertible<Type, Scalar>::value;
         m_base.init(is_arithmetic, is_convertible);
 
-        def(init([](Scalar i) { return static_cast<Type>(i); }));
+        def(init([](Scalar i) { return static_cast<Type>(i); }), arg("value"));
         def("__int__", [](Type value) { return (Scalar) value; });
         #if PY_MAJOR_VERSION < 3
             def("__long__", [](Type value) { return (Scalar) value; });
@@ -1744,7 +1744,7 @@ public:
                 detail::initimpl::setstate<Base>(v_h, static_cast<Type>(arg),
                         Py_TYPE(v_h.inst) != v_h.type->type); },
             detail::is_new_style_constructor(),
-            pybind11::name("__setstate__"), is_method(*this));
+            pybind11::name("__setstate__"), is_method(*this), arg("state"));
     }
 
     /// Export enumeration entries into the parent scope
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -218,3 +218,10 @@ def test_duplicate_enum_name():
     with pytest.raises(ValueError) as excinfo:
         m.register_bad_enum()
     assert str(excinfo.value) == 'SimpleEnum: element "ONE" already exists!'
+
+
+def test_docstring_signatures():
+    for enum_type in [m.ScopedEnum, m.UnscopedEnum]:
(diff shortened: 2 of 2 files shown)


YannickJadoul:
@andioz, if you would want to have a look, and double-check if I didn't miss anything, please go ahead! :-)

andioz:
@YannickJadoul, thanks for the quick implementation! I tried the new code, it looks good. If I understand it right, this pull request changes the hard-coded argument names from `arg0` to `value` for enums,  to `state` for `__setstate__` methods, respectively 3 more argument names to `other` (I cannot find out for which methods right now).

You say this is no full solution, you mean there is still no method to customize the name, right? Well thats fine right now, good to have good default names.

I would suggest to add all changed names to the changelog, and maybe mention it in the documentation?

And finally, I would suggest to add piece of code to the tests, just to show the changed behavior.

Good job :+1:

YannickJadoul:
Thanks, @andioz!

> respectively 3 more argument names to `other` (I cannot find out for which methods right now).

Those are the macros for all the comparison operators (`__eq__`, `__lt__`, etc).

> You say this is no full solution, you mean there is still no method to customize the name, right? Well thats fine right now, good to have good default names.

Yes, that's what I meant. After all, I agree `arg0` looks rather ugly, but `value` should work for most people in 99% of the cases? And if not, there are workarounds anyway (i.e., manually turning a `class_` into an enumeration type).

> I would suggest to add all changed names to the changelog, and maybe mention it in the documentation?

Is it that big/important? The changelog (see above) will point here anyway, and for the docs I'm not convinced this isn't just going to be "noise" (users would just notice in their docstring signatures; I'm not sure there's anything to explain about how this works?).

andioz:
Regarding documentation, I agree, it is not important enough. But a changelog entry should be there. And a short test ;) I've seen there are test for docstrings available.

YannickJadoul:
Yep, the changelog gets added afterwards; it's in the original message at the top of this PR.

Tests could be possible. I also don't think it's that urgent for this kind of case, but it can't really hurt either, I suppose :-)

andioz:
OMG - you should write tests before changing code :laughing: Well, just kidding, but seriously, I learned to find it really useful to write a test before changing code. Anyway, your choice.

YannickJadoul:
Hahaha, myeah, I know. But I'm too focused on fixing it, and especially in this case, it feels so immediate that ... yeah. When it gets complex, I do sometimes stash my own changes to see if tests failed before, though ;-)

andioz:
I'm fine with it :) So, do you think about closing the issue at all, even functions for customization are not available yet? With some rejection note?

YannickJadoul:
> So, do you think about closing the issue at all, even functions for customization are not available yet? With some rejection note?

Well, yeah. It's kind of solved, no? I thought this was at least good enough, unless someone can come up with some great idea. Or don't you agree?

andioz:
Agreed and approved 👍 I close the issue, thanks Yannick

YannickJadoul:
> Agreed and approved +1 I close the issue, thanks Yannick

No worries. This will automatically close that issue ;-)

andioz:
haha I'm not used to this github stuff, but I will learn

YannickJadoul:
Given this "Closes #2623" (or "Fixes") in the main description above, merging this PR will close that issue automatically.

andioz:
Oh a test! Great!! 👏

YannickJadoul:
Not sure what else to test, so I'm going to call this done ;-)

andioz:
Sure, good test are often pretty simple, if the code base is clean. Thanks for all, great work!

henryiii (approved):

rwgk (approved):
This PR passed Google-global testing.
I strongly support merging now, so it ships with 2.6.1.

henryiii:
That's good enough for me.

PR #2638, opened 2020-11-03
Fix so is deprecated, use ext suffix

MatthiasAbtIfm:
## Suggested changelog entry:

```rst

```

PR #2640, opened 2020-11-03
fix: match new extension discovery with changes to classic discovery

henryiii:
## Description

Followup to #2638 - this was fixed in 2.6.0, but only for classic Python

## Suggested changelog entry:

None needed, too small.

### DIFF
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -97,7 +97,7 @@ if(NOT DEFINED PYTHON_MODULE_EXTENSION)
   execute_process(
     COMMAND
       "${${_Python}_EXECUTABLE}" "-c"
-      "from distutils import sysconfig as s;print(s.get_config_var('SO') or s.get_config_var('EXT_SUFFIX'))"
+      "from distutils import sysconfig as s;print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
     OUTPUT_VARIABLE _PYTHON_MODULE_EXTENSION
     ERROR_VARIABLE _PYTHON_MODULE_EXTENSION_ERR
     OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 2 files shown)


tools/pybind11NewTools.cmake
```diff
@@ -95,8 +95,9 @@ endif()
 # required for PyPy3 (as of 7.3.1)
 if(NOT DEFINED PYTHON_MODULE_EXTENSION)
   execute_process(
-    COMMAND "${${_Python}_EXECUTABLE}" "-c"
-            "from distutils import sysconfig; print(sysconfig.get_config_var('SO'))"
+    COMMAND
+      "${${_Python}_EXECUTABLE}" "-c"
+      "from distutils import sysconfig as s;print(s.get_config_var('SO') or s.get_config_var('EXT_SUFFIX'))"
```
bstaletic, line 100: If accessing `s.get_config_var('SO')` raises a deprecation warning, we should check `EXT_SUFFIX` first to avoid raising the warning unnecessarily.

henryiii: This breaks PyPy, at least 7.3.1 and older, as it either returns the wrong thing or cannot accept a wheel with the correct EXT_SUFFIX. (I could be remembering incorrectly). This is mostly to protect if it no longer is a warning and instead goes away.

YannickJadoul: We could ignore the warnings with a context manager, and make a note to update this once PyPy 7.4 is released and we drop 7.3 support?

henryiii: Good news! I tried this on 3.7.1, and "EXT_SUFFIX" is empty on PyPy2 (which is okay if we check both), and it matches "SO" on PyPy3. So I think it's safe to do this. I'm going to try it and see if it seems to work, tests should catch 7.3.1 issues. 7.3.2 hopefully is not worse, at least.

henryiii: I think the one that was causing issues in CMake was the SOABI variable.

bstaletic (approved):

rwgk (approved):
Based on @bstaletic approval I support merging this PR into 2.6.1.

PR #2643, opened 2020-11-05
feat: way to only recompile changed files

henryiii:
## Description

Adding a "cache" feature that assists in development. If a source has not changed since the file has been created, you can just reuse the old compile. Opt-in, and really only affects development builds (`pip install -e .` or similar), since otherwise everything is built in a clean, new directory. Requested by @HDembinski for the HEP/Astro project iMinuit. I've based it on the existing code there, and verified that it works on boost-histogram (the `python_example` only has one file, which makes it rather uninformative for testing this). Added notes in the docs about CCache, too.

```python
from pybind11.setup_helpers import ParallelCompile, simple_recompile

# Or, custom function:
# def simple_recompile(obj, src):
#     return os.stat(obj).st_mtime < os.stat(src).st_mtime

ParallelCompile("NPY_NUM_BUILD_JOBS", needs_recompile = simple_recompile).install()
```

## Suggested changelog entry:

```rst
* Added ``needs_recompile`` optional function to the ParallelCompiler helper, to allow a recompile to be skipped based on a user-defined function.
```


henryiii:
PS: this is a very minor feat, so I'd be okay with it going into 1.6.1, what do you think?

YannickJadoul:
Why don't we just always advise ccache? That seems like the obvious way to do this, taking all kinds of subtleties into account, rather than adding unrelated code to pybind11. (I know that's putting it rather harshly; intentionally so, since I expect you have a good answer ;-) )

Same as with the `ParallelCompile` thing: I'm a bit weary on why pybind11 should offer this, and why it's not a separate package/project that can be pip installed/included as CMake submodule or subtree/...

pybind11/setup_helpers.py
```diff
@@ -309,6 +309,9 @@ class ParallelCompile(object):
     envvar: Set an environment variable to control the compilation threads, like NPY_NUM_BUILD_JOBS
     default: 0 will automatically multithread, or 1 will only multithread if the envvar is set.
     max: The limit for automatic multithreading if non-zero
+    lazy: Do not try to recompile up-to-date outputs if True
```
YannickJadoul, line 312: Wondering if "lazy" is the right thing here. When I hear lazy in computational context, I think of postponing work until it's really necessary, rather than not doing work that has already be done. But maybe I'm confusing the general "lazy" and specific "lazy evaluation"?

henryiii:
CCache is a separate dependency, and not very officially supported on Windows. This is a simpler (and off-by-default) workaround that works everywhere, and enabling it doesn't affect "normal" users, as a non-editable install is in a new build folder anyway. It's also easy for us to include here, but hard for a user to inject.

It's not trivial to include dependencies at setuptime, each method of doing so (PEP 518, setup_requires, submodules, copy-and-paste) has drawbacks and doesn't work everywhere (conda disables most of them, for example). IMO, pybind11 should offer reasonable assistance with users wanting to use setuptools, because that's a key part of writing a good pybind11 module.

If, for example, we added `ParallelCompile` to a package, then it would not be available for the non-setuptools users, because we can't ask for dependencies like a normal Python package can. If we add it as a submodule, it breaks users on GitLab. Either way, it wouldn't work for add_submodule usage. If we require uses to also use that package to get the feature, then they would have to solve the problem of adding the new package again, after adding pybind11.

pybind11/setup_helpers.py
henryiii: Feel free to suggest a better term. I has `cached` originally.

YannickJadoul:
> This is a simpler (and off-by-default) workaround that works everywhere, and enabling it doesn't affect "normal" users, as a non-editable install is in a new build folder anyway.

> IMO, pybind11 should offer reasonable assistance with users wanting to use setuptools

That's true, so I'm not really opposed to it; you managed to convince me of `ParallelCompile` as well.

> It's not trivial to include dependencies at setuptime

But ... pybind11 is already a dependency at setuptime? One that users either require with `pip` or `conda`, or include as submodule/subtree, or vendor, or ... Why would this be so much different? If users can install pybind11, why couldn't they install in a very similar way "henryiii/setuptoolstools" (or something like that; I'm not pinning you to this horrible name :-p ). You already have this whole explanation of `setup_helpers`, and we're not using this `ParallelCompile` ourselves in pybind11, are we?

But, as I said, I don't oppose including it. To some degree this is just an additional feature/fix to the already existing `ParallelCompile`, so...

pybind11/setup_helpers.py
YannickJadoul: Any reason you discarded `cached`? Otherwise, `use_cache`, perhaps?

But no change needed for just me. Just mentioning and seeing if you or anyone else would feel the same about `lazy`.

henryiii: `cached=True` doesn't really explain what it's doing that well, I though. How about `if/as_needed=True`?

YannickJadoul: True, yes. Or `always_recompile=False`?

YannickJadoul: > True, yes. Or `always_recompile=False`?

Nah, that's not fully clear either.
So, OK, from that perspective, `lazy` is a great term, it just has an unfortunate connotation.

YannickJadoul: `incremental=True`?

henryiii:
My point was any external solution will significantly complicate usage from pybind11, where I expect most users will be. This makes life much easier for simple to intermediate pybind11 users (advanced users probably will set everything up for themselves anyway).

pybind11/setup_helpers.py
henryiii: I actually liked `always_recompile=False` better than `incremental=True`, which sounds like incremental linking in Windows.

How about `only_changed=True`?

YannickJadoul: (I was thinking of incremental builds, but OK)

Could also work, yes. None of these all fully convinced me, so far. But none of them are awful either, so ...

HDembinski:
Hi, just to chime in, @YannickJadoul you are right, of course, that this functionality and parallel builds should ideally come from setuptools, but that's unfortunately not the case. @henryiii made a good case against ccache and it is a rather small patch with a lot of convenience value, so not really adding bloat to pybind11. I think everyone here would be happy to rip this out again once this functionality is available somewhere upstream.

I missed why `lazy` was renamed to `only_changed`, I would prefer the shorter `lazy` and it is pretty much the usual term for this sort of thing, if it is opt-in, because that means anyone who actually wants to use this has to type it a lot.

henryiii:
`lazy` sounds like it's delaying compilation or changing the parallel method somehow; it's not quite as clear as `only_changed`, which describes exactly what it's going to do if you turn it on (only recompile changed files). It's not something you type often (assuming even with the most rapid prototyping you do not make a new setup.py every day? And copy-paste is rather common in setup code), and it's rather specialized, so I'd rather have it read like documentation than save a few characters. I'm not strongly against `lazy` either, though.

I have tested this locally and verified it works with `pip install -e . -v`; it's a nice speed boost even in a parallel environment if you edit a quicker-to-compile file!

Remember, it will look like this:

```python
# Use the environment variable NPY_NUM_BUILD_JOBS
ParallelCompile("NPY_NUM_BUILD_JOBS", only_changed=True).install()
# or
ParallelCompile("NPY_NUM_BUILD_JOBS", lazy=True).install()
```

YannickJadoul:
@HDembinski

> this functionality and parallel builds should ideally come from setuptools, but that's unfortunately not the case.

True, or some independent plugin. E.g., I don't see why this couldn't be interesting for Cython project. And since our docs suggest vendoring the file or using `PEP 518`, there's not a lot of extra complexity if a user has to add this `ParallelCompile` in the same way?

> @henryiii made a good case against ccache and it is a rather small patch with a lot of convenience value, so not really adding bloat to pybind11. I think everyone here would be happy to rip this out again once this functionality is available somewhere upstream.

Agreed; I see the ccache argument, yes (though ccache still seems to be the correct way to set something up on the long term, to me?), and indeed, it's just tweaking `ParallelCompile`, so I'm fine to see this merged. I'm more concerned about re-usability and separation of concerns than about this not being maintainable :-)

And yes, `lazy` was nice and short, but Henry correctly summarized my doubts about that name. If it's decide that that connotation to `lazy` is not a problem, I'm fine with that as well, though.

henryiii:
> I don't see why this couldn't be interesting for Cython project

ParallelCompile is already available (in a less configurable form) in NumPy, which Cython users have to have at compile time anyway, while pybind11 users do not. Agree that eventually we might also make then available in a separate package, but that will complicate usage here. We probably would want to follow pip's method of including dependencies - it has dozens of dependencies, but they [vendor them in](https://github.com/pypa/pip/tree/master/src/pip/_vendor) - probably with git subtree or something similar (submodules have drawbacks for something so integral as pybind11).

HDembinski:
> It's not something you type often (assuming even with the most rapid prototyping you do not make a new setup.py every day?

Sorry, I misunderstood how this works, I thought I have to add `only_changed` as a command-line option to `pip install`.

henryiii:
So I think we are around 2-3 votes for `only_changed`?

henryiii:
I've noticed that NumPy does this by default: https://github.com/numpy/numpy/blob/cfee2df31bdc127c345894aee594cbcbb8cddd0e/numpy/distutils/ccompiler.py#L303 

Should we consider changing the default from `only_changed=False` to `only_changed=True` in pybind11 2.7.0? Or here?

henryiii:
I've restructured this quite a bit. With the observation that this is not recursive (it can't tell if you changed a header file), I've pulled the check function out and instead provided a way for users to add their own, with the "simple" one as a suggestion. This a) doesn't assume anything about a user's project structure, b) keeps the code that we provide simple, and c) keeps us from falling down a rabbit hole if users start complaining about changing a header and not having it show up when they rebuild.

Many projects are probably fine and mostly .cpp files, but not all. As a counter-example, boost-histogram has more pybind11 code in headers than it does in .cpp files.

New usage:

```python
from pybind11.setup_helpers import ParallelCompile

class SmartCompile(ParallelCompile):
    @staticmethod
    def needs_recompile(obj, src):
        return os.stat(obj).st_mtime < os.stat(src).st_mtime

SmartCompile("NPY_NUM_BUILD_JOBS").install()
```

Note that in my original reasoning I mentioned that one of the key reasons was user's couldn't modify this. Now they can. :)

henryiii:
Hopefully this will make @YannickJadoul happier, and @HDembinski if you could "review and approve" if you like it, that would be helpful too. :)

rwgk (approved):
Looks awesome to me :-)

YannickJadoul:
> Hopefully this will make @YannickJadoul happier

I wasn't per se unhappy about this PR ;-) Especially not about this part, because I didn't actually notice the recursive/header issue.
How does `numpy` handle this recursive case, then, btw?

Two suggestions:
- Rather than just in the docs, we could also provide this extra simple class, to avoid different project reimplementing this? `NaiveIncrementalCompile`/`SimpleIncrementalCompile` or so?
- Instead of subclassing, you could also accept a callable `should_recompile`? (Not convinced this is better, but throwing this out there, if anyone would prefer this to subclassing.)

Apart from that, sure, go ahead and merge, as far as I'm concerned :-)

henryiii:
> we could also provide this extra simple class

Sure, that would be useful, I think.

> Instead of subclassing, you could also accept a callable

I thought about it when designing this - either way works, I slightly prefer the extra structure provided by subclassing (plus, that way we could provide a single import "SimpleIncrementalCompile" instead of having to get and use a second import for the function and pass it in. But no strong preference here. @HDembinski do you have an opinion?

YannickJadoul:
> The "import" version would have looked like this:

Yep, something like that is what I was thinking! No strong opinion either way, though (this is slightly more cumbersome, but also avoids subclassing, so potentially a slight increase in future combinatoriality?). `naive` might be better than `simple`, though, as a warning to users, or do you think that's too negative?

HDembinski:
> I thought about it when designing this - either way works, I slightly prefer the extra structure provided by subclassing (plus, that way we could provide a single import "SimpleIncrementalCompile" instead of having to get and use a second import for the function to pass it in. But no strong preference here. @HDembinski do you have an opinion?

I favor the "pass a callable" approach, because of minimal coupling and minimal design. The function is independent of the class. Requiring me to subclass something to add a staticmethod is not a minimal elegant design. Subclassing the preferred solution only if it is conceivable that the callable may need to access privates of `ParallelCompile`, but it is designed anyway to not access `self`, so...

HDembinski:
> naive might be better than simple, though, as a warning to users, or do you think that's too negative?

I also like `naive`, gives a fair warning.

henryiii:
Okay, you've convinced me - I've pushed a free function version.

docs/compiling.rst
```diff
@@ -89,6 +89,27 @@ default number of threads (0 will take the number of threads available) and
...
+with the following:
+
+.. code-block:: python
+
+    from pybind11.setup_helpers import ParallelCompile, naive_recompile
+
+    SmartCompile("NPY_NUM_BUILD_JOBS", needs_recompile=naive_recompile).install()
+
+
+If you have a more complex build, you can implement a smarter function, or you
```
rwgk, line 105: ..., you can pass a smarter function to `needs_recompile`, ...
?

rwgk (approved):
Nice!

docs/compiling.rst
henryiii: Thanks, updated text slightly.

pybind11/setup_helpers.py
```diff
@@ -306,24 +324,42 @@ class ParallelCompile(object):
...
+    function can be provided via needs_recompile.  If the output has not yet
+    been generated, the compile will always run, and this function is not
+    called.
     """
 
-    __slots__ = ("envvar", "default", "max", "old")
+    __slots__ = ("envvar", "default", "max", "_old", "needs_recompile")
 
-    def __init__(self, envvar=None, default=0, max=0):
+    def __init__(self, envvar=None, default=0, max=0, needs_recompile=no_recompile):
```
YannickJadoul, line 357: Oh, interesting. I was just expecting `needs_recompile=None` as default.
But since this is nothing performance-critical, why not like this, yes.

pybind11/setup_helpers.py
```diff
@@ -306,24 +324,42 @@ class ParallelCompile(object):
...
+        Set an environment variable to control the compilation threads, like
+        NPY_NUM_BUILD_JOBS
+    default:
+        0 will automatically multithread, or 1 will only multithread if the
+        envvar is set.
+    max:
+        The limit for automatic multithreading if non-zero
+    needs_recompile:
+        A function of (obj, src) that returns True when recompile is needed.  No
+        effect in isolated mode; use ccache instead, see
```
YannickJadoul, line 337: If you want another nitpicky remark on capitals: "CCache"?

YannickJadoul: Actually, apparently it's "Ccache" or "ccache", according to their own website: https://ccache.dev/ So maybe the other ones should be changed?

(Or just ignore this; I'm sure there's better ways to spend your time.)

pybind11/setup_helpers.pyi
```diff
@@ -1,7 +1,7 @@
 # IMPORTANT: Should stay in sync with setup_helpers.py (mostly checked by CI /
 # pre-commit).
 
-from typing import Any, Iterator, Optional, Type, TypeVar, Union
+from typing import Any, Iterator, Optional, Type, TypeVar, Union, Callable
```
YannickJadoul, line 4: Another nitpicky one:

```suggestion
from typing import Any, Callable, Iterator, Optional, Type, TypeVar, Union
```

YannickJadoul (approved):
Looks pretty clean!

pybind11/setup_helpers.py
henryiii: Dissalowing CCache, and fixed.

pybind11/setup_helpers.pyi
henryiii: Fixed, thanks!

pybind11/setup_helpers.py
henryiii: This provided slightly simpler typing, being just `Callable[[str, str],bool ` instead of `Optional[Callable[[str, str],bool]`, which is why I did it this way.

docs/compiling.rst
```diff
@@ -89,6 +89,36 @@ default number of threads (0 will take the number of threads available) and
...
+pip install -e .`` would be the way to use it with GCC, for example. Unlike the
+simple solution, this even works even when not compiling in editable mode, but
+it does require Ccache to be installed.
+
+Keep in mind that pip will not even attempt to rebuild if it things it has
+already built a copy of your code, which it deduces from the version number.
+One way to avoid this is to use [setuptools_scm]_, which will generate a
+version number that includes the number of commits since your last tag and a
+hash for a dirty directory. Another way to force a rebuild is purge your cache
+or use pip's ``--no-cache-dir`` option.
```
henryiii, line 116: FYI, I've just added this section, as this is likely a common issue related to caching. @YannickJadoul

docs/compiling.rst
```diff
@@ -89,6 +89,36 @@ default number of threads (0 will take the number of threads available) and
...
+pip install -e .`` would be the way to use it with GCC, for example. Unlike the
+simple solution, this even works even when not compiling in editable mode, but
+it does require Ccache to be installed.
+
+Keep in mind that Pip will not even attempt to rebuild if it thinks it has
+already built a copy of your code, which it deduces from the version number.
+One way to avoid this is to use [setuptools_scm]_, which will generate a
+version number that includes the number of commits since your last tag and a
+hash for a dirty directory. Another way to force a rebuild is purge your cache
+or use Pip's ``--no-cache-dir`` option.
```
henryiii, line 116: FYI, I've just added this section, as this is likely a common issue related to caching. @YannickJadoul

HDembinski:
I am not an official reviewer, but for the record, it looks all good to me, thank you for the great work, @henryiii

henryiii:
I think we have enough approvals, then. Thanks everyone!

PR #2645, opened 2020-11-08
style: clang-tidy: modernize-use-using

rhaschke:
I stumbled across some clang-tidy warnings, which obviously are left over from the cleanup in b342c37388177067db9e8ba609aacc119c24b1a4. @henryiii

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -35,7 +35,12 @@ jobs:
       run: apt-get update && apt-get install -y python3-dev python3-pytest
 
     - name: Configure
-      run: cmake -S . -B build -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--warnings-as-errors=*"
+      run: >
+        cmake -S . -B build
+        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--warnings-as-errors=*"
+        -DDOWNLOAD_EIGEN=ON
+        -DDOWNLOAD_CATCH=ON
+        -DCMAKE_CXX_STANDARD=17
 
     - name: Build
       run: cmake --build build -j 2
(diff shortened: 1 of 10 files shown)


henryiii:
Great, but why are we not seeing these in CI?

rhaschke:
> Great, but why are we not seeing these in CI?

Which version of clang-tidy are you using in CI? I was using clang-tidy-10 locally. 
And I noticed in other contexts that clang-tidy isn't 100% reliable. Nevertheless, it's a great tool!

henryiii:
Same: https://github.com/pybind/pybind11/blob/06b673a0daef1db4f921a19676a51abec6fb13e8/.github/workflows/format.yml#L30-L41

Any idea if there is something different in how we are running it? C++ version, Eigen, Catch, or anything?

henryiii:
I can get a couple of those warnings with Eigen enabled. Still missing most of them.

henryiii (approved):

rhaschke:
> Any idea if there is something different in how we are running it? C++ version, Eigen, Catch, or anything?

No, sorry. I think, I had enabled Eigen, which triggered a few of those warnings. Then, I explicitly searched for typedefs in the codebase and fixed them as well.

rwgk (commented):
This looks good to me. I'll include this PR in a google-global-testing run I'm about to trigger. I'll approve when I get the results (will take ~7 hours).

YannickJadoul:
If all our compilers are happy with these changes, there's no way a testing suite is going to trip over this, though.
So merge if you want this in, I'd say, @henryiii.

rwgk:
> If all our compilers are happy with these changes, there's no way a testing suite is going to trip over this, though.
> So merge if you want this in, I'd say, @henryiii.

I agree, good to merge!
My approval was mainly meant as a flag for "yes I ran those tests".

bstaletic (approved):

rwgk (approved):
Our "smoke test" finished already (random sample of 1000 tests). That's only possible if our compiler is happy with the changes in this PR. There is absolutely no need to wait for the full run.

PR #2647, opened 2020-11-09
[WIP] sandbox for #2047

rhaschke:
This is just a sandbox branch to test move/copy semantics of native C++ for #2047.


rhaschke:
I don't understand the CI compile failures, which seem to be unrelated to the suggested code changes. Locally, the source builds fine, but fails with a new unit tests (because the move-enabling commit of #2047 is missing here).

PR #2650, opened 2020-11-10
Feature/local exception translator

jesse-sony:
## Description

Allow exception translators to be optionally registered local to a module instead of applying globally across all pybind11 modules.

This is useful in a situation where there are more than one compiled modules (e.g., clients for different micro-services) that share a lot of the same core code on the c++ side, including exceptions. Shared classes can be handled via the py::module_local() tag on the classes, but there is currently no way to deal with exception remapping.

So in code like:
```
    py::register_exception_translator([](std::exception_ptr p) {
      try {
        if (p) {
          std::rethrow_exception(p);
        }
      }
      catch (const ::vasg::exceptions::VasgException& e) {
        auto err = py::cast(e);
        auto errType = err.get_type().ptr();
        PyErr_SetObject(errType, err.ptr());
      }
```
The exception translator is global, so which module's exception you get back is determined by import order.

As an example, if you did something like
```
import tempo # compiled lib 

tempo.resolveKey(badPath)
```
you get back an exception from the tempo module.

However, if you were to do
```
import rhapsody # compiled lib
import tempo # compiled lib

tempo.resolveKey(badPath)
```
you'd get back an exception from the rhapsody module. Since those are different types in python (but the same type in c++), it means catch statements which are expecting the exception to come from the same module as the call are failing.

## Suggested changelog entry:

```rst

Allow exception translators to be optionally registered local to a module instead of applying globally across all pybind11 modules.
Use 
register_local_exception_translator(ExceptionTranslator&& translator)
instead of 
register_exception_translator(ExceptionTranslator&& translator)
to keep your exception remapping code local to the module.
```

### Why does this not require an Internals Version bump?
The internals struct in pybind11 is shared between all the modules. Any changes to that definitely requires a lot of careful thought about whether backwards compatibility is being broken. The local_internals that are added here should, by design, be segregated to a single module. Whether a module has local internals or not should not impact any other modules, because the only things accessing the local internals is the module that contains them.

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -75,9 +75,10 @@ Registering custom translators
 
 If the default exception conversion policy described above is insufficient,
 pybind11 also provides support for registering custom exception translators.
-To register a simple exception conversion that translates a C++ exception into
-a new Python exception using the C++ exception's ``what()`` method, a helper
-function is available:
+Similar to pybind11 classes, exception translators can be local to the module
+they are defined in or global to the entire python session.  To register a simple
+exception conversion that translates a C++ exception into a new Python exception
+using the C++ exception's ``what()`` method, a helper function is available:
 
 .. code-block:: cpp
 
@@ -87,12 +88,20 @@ This call creates a Python exception class with the name ``PyExp`` in the given
 module and automatically converts any encountered exceptions of type ``CppExp``
 into Python exceptions of type ``PyExp``.
 
+A matching function is available for registering a local exception translator:
+
+.. code-block:: cpp
+
+    py::register_local_exception<CppExp>(module, "PyExp");
+
+
 It is possible to specify base class for the exception using the third
 parameter, a `handle`:
 
 .. code-block:: cpp
 
     py::register_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
+    py::register_local_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
 
 Then `PyExp` can be caught both as `PyExp` and `RuntimeError`.
 
@@ -100,16 +109,18 @@ The class objects of the built-in Python exceptions are listed in the Python
 documentation on `Standard Exceptions <https://docs.python.org/3/c-api/exceptions.html#standard-exceptions>`_.
 The default base class is `PyExc_Exception`.
 
-When more advanced exception translation is needed, the function
-``py::register_exception_translator(translator)`` can be used to register
+When more advanced exception translation is needed, the functions
+``py::register_exception_translator(translator)`` and
+``py::register_local_exception_translator(translator)`` can be used to register
 functions that can translate arbitrary exception types (and which may include
-additional logic to do so).  The function takes a stateless callable (e.g.  a
+additional logic to do so).  The functions takes a stateless callable (e.g. a
 function pointer or a lambda function without captured variables) with the call
 signature ``void(std::exception_ptr)``.
 
 When a C++ exception is thrown, the registered exception translators are tried
 in reverse order of registration (i.e. the last registered translator gets the
-first shot at handling the exception).
+first shot at handling the exception). All local translators will be tried
+before a global translator is tried.
 
 Inside the translator, ``std::rethrow_exception`` should be used within
 a try block to re-throw the exception.  One or more catch clauses to catch
@@ -167,6 +178,56 @@ section.
     Note that ``libc++`` and ``libstdc++`` `behave differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
     with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI boundaries need to be explicitly exported, as exercised in ``tests/test_exceptions.h``.
     See also: "Problems with C++ exceptions" under `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
+
+
+Local vs Global Exception Translators
+=====================================
+
+Similar to how the ``py::module_local`` flag allows uesrs to limit a bound class
+to a single compiled module to prevent name collisions. When a global exception
+translator is registered, it will be applied across all modules in the reverse
+order of registration. This can create behavior where the order of module import
+influences how exceptions are translated.
+
+If module1 has the following translator:
+
+.. code-block:: cpp
+
+      py::register_exception_translator([](std::exception_ptr p) {
+        try {
+            if (p) std::rethrow_exception(p);
+        } catch (const std::invalid_argument &e) {
+            PyErr_SetString("module1 handled this")
+        }
+      }
+
+and module2 has the following similar translator:
+
+.. code-block:: cpp
+
+      py::register_exception_translator([](std::exception_ptr p) {
+        try {
+            if (p) std::rethrow_exception(p);
+        } catch (const std::invalid_argument &e) {
+            PyErr_SetString("module2 handled this")
+        }
+      }
+
+then which translator handles the invalid_argument will be determined by the
+order that module1 and module2 are imported. Since exception translators are
+applied in the reverse order of registration, which ever module was imported
+last will "win" and that translator will be applied.
+
+If there are multiple pybind11 modules that share exception types (either
+standard built-in or custom) loaded into a single python instance and
+consistent error handling behavior is needed, then local translators should be
+used.
+
+Changing the previous example to use register_local_exception_translator would
+mean that when invalid_argument is thrown in the module2 code, the module2
+translator will always handle it, while in module1, the module1 translator will
+do the same.
+>>>>>>> 854aa95a (Update documentation for new local exception feature)
 
 .. _handling_python_exceptions_cpp:
 
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -183,11 +183,9 @@ section.
 Local vs Global Exception Translators
 =====================================
 
-Similar to how the ``py::module_local`` flag allows uesrs to limit a bound class
-to a single compiled module to prevent name collisions. When a global exception
-translator is registered, it will be applied across all modules in the reverse
-order of registration. This can create behavior where the order of module import
-influences how exceptions are translated.
+When a global exception translator is registered, it will be applied across all
+modules in the reverse order of registration. This can create behavior where the
+order of module import influences how exceptions are translated.
 
 If module1 has the following translator:
 
@@ -223,11 +221,10 @@ standard built-in or custom) loaded into a single python instance and
 consistent error handling behavior is needed, then local translators should be
 used.
 
-Changing the previous example to use register_local_exception_translator would
-mean that when invalid_argument is thrown in the module2 code, the module2
-translator will always handle it, while in module1, the module1 translator will
-do the same.
->>>>>>> 854aa95a (Update documentation for new local exception feature)
+Changing the previous example to use ``register_local_exception_translator``
+would mean that when invalid_argument is thrown in the module2 code, the
+module2 translator will always handle it, while in module1, the module1
+translator will do the same.
 
 .. _handling_python_exceptions_cpp:
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -325,7 +325,7 @@ inline module_internals &get_module_internals() {
 }
 
 /// Works like `internals.registered_types_cpp`, but for module-local registered types:
-inline type_map<type_info *> &registered_local_types_cpp() {
+inline type_map<type_info *> &get_registered_local_types_cpp() {
     return get_module_internals().registered_local_types_cpp;
 }
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -320,7 +320,7 @@ PYBIND11_NOINLINE inline internals &get_internals() {
 
 // the internals struct (above) is shared between all the modules. local_internals are only
 // for a single module. Any changes made to internals may require an update to
-// PYBIND11_INTERNALS_VERSION, breaking backwards compatiblity. local_internals is, by design,
+// PYBIND11_INTERNALS_VERSION, breaking backwards compatibility. local_internals is, by design,
 // restricted to a single module. Whether a module has local internals or not should not
 // impact any other modules, because the only things accessing the local internals is the
 // module that contains them.
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -12,7 +12,11 @@
 #include "../pytypes.h"
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+
+using ExceptionTranslator = void (*)(std::exception_ptr);
+
 PYBIND11_NAMESPACE_BEGIN(detail)
+
 // Forward declarations
 inline PyTypeObject *make_static_property_type();
 inline PyTypeObject *make_default_metaclass();
@@ -100,7 +104,7 @@ struct internals {
     std::unordered_set<std::pair<const PyObject *, const char *>, override_hash> inactive_override_cache;
     type_map<std::vector<bool (*)(PyObject *, void *&)>> direct_conversions;
     std::unordered_map<const PyObject *, std::vector<PyObject *>> patients;
-    std::forward_list<void (*) (std::exception_ptr)> registered_exception_translators;
+    std::forward_list<ExceptionTranslator> registered_exception_translators;
     std::unordered_map<std::string, void *> shared_data; // Custom data to be shared across extensions
     std::vector<PyObject *> loader_patient_stack; // Used by `loader_life_support`
     std::forward_list<std::string> static_strings; // Stores the std::strings backing detail::c_str()
@@ -316,7 +320,7 @@ PYBIND11_NOINLINE inline internals &get_internals() {
 
 struct local_internals {
   type_map<type_info *> registered_local_types_cpp;
-  std::forward_list<void (*) (std::exception_ptr)> registered_local_exception_translators;
+  std::forward_list<ExceptionTranslator> registered_local_exception_translators;
 };
 
 /// Works like `get_internals`, but for things which are locally registered.
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -314,19 +314,15 @@ PYBIND11_NOINLINE inline internals &get_internals() {
 }
 
 
-struct module_internals {
+struct local_internals {
   type_map<type_info *> registered_local_types_cpp;
   std::forward_list<void (*) (std::exception_ptr)> registered_local_exception_translators;
 };
 
-inline module_internals &get_module_internals() {
-  static module_internals locals;
+/// Works like `get_internals`, but for things which are locally registered.
+inline local_internals &get_local_internals() {
+  static local_internals locals;
   return locals;
-}
-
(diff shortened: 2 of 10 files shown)


rhaschke (approved):
I think this is an important feature and the implementation looks reasonable to me.

include/pybind11/detail/internals.h
```diff
@@ -311,12 +311,23 @@ PYBIND11_NOINLINE inline internals &get_internals() {
...
 
+
+struct module_internals {
+  type_map<type_info *> registered_local_types_cpp;
+  std::forward_list<void (*) (std::exception_ptr)> registered_local_exception_translators;
+};
+
+inline module_internals &get_module_internals() {
+  static module_internals locals;
+  return locals;
```
YannickJadoul, line 322: This should probably not be called `locals` anymore, since it's not just the local variables anymore?

include/pybind11/detail/internals.h
```diff
@@ -311,12 +311,23 @@ PYBIND11_NOINLINE inline internals &get_internals() {
...
+}
+
 /// Works like `internals.registered_types_cpp`, but for module-local registered types:
 inline type_map<type_info *> &registered_local_types_cpp() {
-    static type_map<type_info *> locals{};
-    return locals;
+    return get_module_internals().registered_local_types_cpp;
```
YannickJadoul, line 327: Not sure how other parts of pybind11 do it, but you do have a duplicate use of `registered_local_types_cpp` here (field as well as the function name). Can you check what conventions other parts of `internals.h` follow?

YannickJadoul: Well, I overcame my laziness and looked up myself. Can we go for `get_registered_local_types_cpp` as function name, for example? I don't think this should break ABI things.

include/pybind11/detail/internals.h
```diff
@@ -311,12 +311,23 @@ PYBIND11_NOINLINE inline internals &get_internals() {
     return **internals_pp;
 }
 
+
+struct module_internals {
```
YannickJadoul, line 315: Can/should we make this `module_local_internals`?

YannickJadoul: (same for `get_module_local_internals`, then, if we do)

include/pybind11/pybind11.h
```diff
@@ -777,6 +777,18 @@ class cpp_function : public function {
                 - delegate translation to the next translator by throwing a new type of exception. */
 
             auto last_exception = std::current_exception();
+
```
YannickJadoul, line 780: Can you update the comment above? I.e., the "give each registered exception translator a chance to translate it to a Python exception in reverse order of registration."

include/pybind11/pybind11.h
```diff
@@ -1939,6 +1951,18 @@ void register_exception_translator(ExceptionTranslator&& translator) {
         std::forward<ExceptionTranslator>(translator));
 }
 
+
+/**
+  * Add a new module-local exception translator. This function will be applied
```
YannickJadoul, line 1956: > "This function will be applied"

"These will be tried ..." or so? "applied" seems weird

YannickJadoul (commented):
I general, I'm happy with this (but I was also involved in the Gitter discussion, so I'm not entirely well-placed to judge this).
This is also one of the things @wjakob should probably have a look at? Does it make sense to always check module-local exception translators before the global, shared ones?

We're still lacking changes to the actual documentation, by the way. A note about the existence of module-local exception/exception translators, and stating the order in which they are resolved is important.

Also, what about the auxiliary function `register_exception`? Should it get an extra argument, or somehow use the `py::module_local()` tag, or an extra template argument, or a sibling function `register_module_local_exception`, or ... ?

include/pybind11/detail/internals.h
jesse-sony: it was all stuff that was local to the module, so seemed appropriate. I'm happy to rename it though. Since it never escapes the context of the function, not sure how much it really matters

include/pybind11/detail/internals.h
jesse-sony: would you prefer me to rename this function or just replace usages of it with `get_module_internals().registered_local_types_cpp;` so it works more like get_internals?

jesse-sony: or would that break ABI? I don't think this function should be ever used in user code

include/pybind11/detail/internals.h
jesse-sony: I'd do either `local_internals` or `module_internals` to contrast with the normal `internals`. I don't see a reason to give it all the names. but if that is what people want, I'll do it

include/pybind11/detail/internals.h
YannickJadoul: Doesn't really matter that much, indeed. Just something that caught my eye, since I thought the old name was linked to Python's `locals()`; maybe it isn't, though.

include/pybind11/detail/internals.h
YannickJadoul: Matching `get_internals()` closer might be nice, indeed, yes, if you ask me :-)
And no, this shouldn't break user code (since it's in the `detail` namespace), and ... yeah, no, the whole point is that this isn't shared between modules, so there's no shared ABI :-)

include/pybind11/detail/internals.h
YannickJadoul: I was thinking of "module local" being a concept, but it's true it also wasn't done for `registered_local_types_cpp`. Maybe matching `local` is more important than `module`, though?

Let's see what others say.

include/pybind11/pybind11.h
jesse-sony: done

include/pybind11/pybind11.h
jesse-sony: verbiage has been adjusted

jesse-sony: You can tell I've been doing a lot of functional programming lately, because I'm out here applying functions to things

include/pybind11/detail/internals.h
jesse-sony: I added a `register_local_exception` function, as well.  Someone who is better at c++ template wrangling than me could probably remove the amount of copy-pasta between register_exception and register_local_exception.

include/pybind11/detail/internals.h
jesse-sony: I have updated this to be called `get_local_internals()`, deleted `get_registered_local_types_cpp()` completely, and changed the usage to be more like `get_internals()`. Do we want to leave the double `local` in the names or match the member names to the `internals` member names?
currently:
```
get_local_internals().registered_local_types_cpp;
```
proposed change to remove the local in the member name
```
get_local_internals().registered_types_cpp;
```

include/pybind11/detail/internals.h
jesse-sony: currently went with `local_internals` and `get_local_internals`. obviously easy to change depending on what everyone wants

jesse-sony:
Documentation has also been updated

include/pybind11/pybind11.h
YannickJadoul: Hah :-D Thanks!

YannickJadoul (approved):
Looks good to me. If this gets one or two more reviews, I think it can be merged :-)

include/pybind11/pybind11.h
```diff
@@ -777,6 +781,18 @@ class cpp_function : public function {
...
+            auto & registered_local_exception_translators = get_local_internals().registered_local_exception_translators;
+            for (auto& translator : registered_local_exception_translators) {
+                try {
+                    translator(last_exception);
+                } catch (...) {
+                    last_exception = std::current_exception();
+                    continue;
+                }
+                return nullptr;
+            }
```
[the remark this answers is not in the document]
jesse-sony: The sort of thing that `chain` from python is perfect for but doesn't seem to be std lib version of it

docs/advanced/exceptions.rst
```diff
@@ -67,9 +67,10 @@ Registering custom translators
 
 If the default exception conversion policy described above is insufficient,
 pybind11 also provides support for registering custom exception translators.
-To register a simple exception conversion that translates a C++ exception into
-a new Python exception using the C++ exception's ``what()`` method, a helper
-function is available:
+Similar to pybind11 classes, exception translators can be local to the module
+they are defined in or global to the entire python session.  To register a simple
```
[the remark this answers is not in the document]
jesse-sony: done

docs/advanced/exceptions.rst
```diff
@@ -79,29 +80,39 @@ This call creates a Python exception class with the name ``PyExp`` in the given
...
+    py::register_local_exception<CppExp>(module, "PyExp");
+
+
 It is possible to specify base class for the exception using the third
 parameter, a `handle`:
 
 .. code-block:: cpp
 
     py::register_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
+    py::register_local_exception<CppExp>(module, "PyExp", PyExc_RuntimeError);
```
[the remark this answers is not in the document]
jesse-sony: removed

docs/advanced/exceptions.rst
```diff
@@ -156,6 +167,54 @@ section.
     may be explicitly (re-)thrown to delegate it to the other,
     previously-declared existing exception translators.
 
+Local vs Global Exception Translators
+=====================================
+
+Similar to how the ``py::module_local`` flag allows uesrs to limit a bound class
+to a single compiled module to prevent name collisions. When a global exception
```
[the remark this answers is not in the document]
jesse-sony: Redid that whole section to be less duplicative.

docs/advanced/exceptions.rst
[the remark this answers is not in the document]
YannickJadoul: Thanks :-)

include/pybind11/cast.h
```diff
@@ -171,7 +171,7 @@ PYBIND11_NOINLINE inline detail::type_info* get_type_info(PyTypeObject *type) {
 }
 
 inline detail::type_info *get_local_type_info(const std::type_index &tp) {
-    auto &locals = registered_local_types_cpp();
+    auto &locals = get_local_internals().registered_local_types_cpp;
```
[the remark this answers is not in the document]
jesse-sony: okay. My preference is to remove the second local too. Hopefully other people agree

tests/test_exceptions.cpp
```diff
@@ -6,7 +6,7 @@
     All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file.
 */
-
```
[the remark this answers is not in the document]
jesse-sony: done

rwgk:
Thanks @Skylion007 for pointing out this PR, I missed it before. I didn't look at all details, but I see others have already. My thinking: @jesse-sony if you rebase and resolve merge conflicts, I'll run the refreshed version through the Google-global testing (includes testing a large number of open source packages) asap, if that checks out I'll merge.

jesse-sony:
@rwgk Branch has been rebased and pushed. Should be in a good state to merge

rwgk:
Thanks! My testing will take a day or so.

rwgk (changes requested):
The Google-global testing came back green, feature-wise this PR looks great to me, but the code duplication is a concern. I think it's important to go the extra mile for best practices straightaway. Usually such things don't get fixed later but rather tend to get duplicated even more (slippery slope).

include/pybind11/detail/internals.h
```diff
@@ -313,12 +313,19 @@ PYBIND11_NOINLINE inline internals &get_internals() {
...
-    return locals;
+
+struct local_internals {
+  type_map<type_info *> registered_local_types_cpp;
+  std::forward_list<void (*) (std::exception_ptr)> registered_local_exception_translators;
+};
+
+/// Works like `get_internals`, but for things which are locally registered.
+inline local_internals &get_local_internals() {
+  static local_internals locals;
```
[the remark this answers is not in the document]
jesse-sony: My thought process on this (and I'm definitely not an expert in all the intricacies of pybind11 ABI compatibility) is that the internals struct in pybind11 is shared between all the modules. Any changes to that definitely requires a lot of careful thought about whether backwards compatibility is being broken. The `local_internals` that I added here should, by design, be segregated to a single module. Whether a module has local internals or not should not impact any other modules, because the only things accessing the local internals is the module that contains them.
If this makes sense to you as well, I'll happily add a comment to the end of this PR explaining that thought process

docs/advanced/exceptions.rst
```diff
@@ -168,6 +179,56 @@ section.
...
+If there are multiple pybind11 modules that share exception types (either
+standard built-in or custom) loaded into a single python instance and
+consistent error handling behavior is needed, then local translators should be
+used.
+
+Changing the previous example to use register_local_exception_translator would
+mean that when invalid_argument is thrown in the module2 code, the module2
+translator will always handle it, while in module1, the module1 translator will
+do the same.
+>>>>>>> 854aa95a (Update documentation for new local exception feature)
```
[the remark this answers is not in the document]
jesse-sony: whoops

include/pybind11/cast.h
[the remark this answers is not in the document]
jesse-sony: @rwgk any thoughts on the naming in the `local_internals` struct?
```
struct local_internals {
  type_map<type_info *> registered_local_types_cpp;
  std::forward_list<void (*) (std::exception_ptr)> registered_local_exception_translators;
};
```
vs
```
struct local_internals {
  type_map<type_info *> registered_types_cpp;
};
```

docs/advanced/exceptions.rst
```diff
@@ -168,6 +179,56 @@ section.
     with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI boundaries need to be explicitly exported, as exercised in ``tests/test_exceptions.h``.
     See also: "Problems with C++ exceptions" under `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
 
+
+Local vs Global Exception Translators
+=====================================
+
+Similar to how the ``py::module_local`` flag allows uesrs to limit a bound class
```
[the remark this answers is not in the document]
jesse-sony: removed

docs/advanced/exceptions.rst
```diff
@@ -168,6 +179,56 @@ section.
...
+order that module1 and module2 are imported. Since exception translators are
+applied in the reverse order of registration, which ever module was imported
+last will "win" and that translator will be applied.
+
+If there are multiple pybind11 modules that share exception types (either
+standard built-in or custom) loaded into a single python instance and
+consistent error handling behavior is needed, then local translators should be
+used.
+
+Changing the previous example to use register_local_exception_translator would
```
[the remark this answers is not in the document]
jesse-sony: added

include/pybind11/pybind11.h
```diff
@@ -2026,6 +2042,19 @@ void register_exception_translator(ExceptionTranslator&& translator) {
         std::forward<ExceptionTranslator>(translator));
 }
 
+
+/**
+  * Add a new module-local exception translator. Locally registered functions
+  * will be tried before any globally registered exception translators, which
+  * will only be invoked if the moduke-local handlers do not deal with
```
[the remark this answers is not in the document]
jesse-sony: fixed

include/pybind11/pybind11.h
```diff
@@ -2085,6 +2114,32 @@ exception<CppException> &register_exception(handle scope,
     return ex;
 }
 
+/**
+ * Registers a Python exception in `m` of the given `name` and installs an exception translator to
+ * translate the C++ exception to the created Python exception using the exceptions what() method.
```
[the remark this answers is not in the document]
jesse-sony: done.

include/pybind11/pybind11.h
```diff
@@ -2026,6 +2042,19 @@ void register_exception_translator(ExceptionTranslator&& translator) {
...
 
+
+/**
+  * Add a new module-local exception translator. Locally registered functions
+  * will be tried before any globally registered exception translators, which
+  * will only be invoked if the moduke-local handlers do not deal with
+  * the exception.
+  */
+template <typename ExceptionTranslator>
+void register_local_exception_translator(ExceptionTranslator&& translator) {
```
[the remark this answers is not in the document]
jesse-sony: Okay, I did this. I did not change any of the type information, because I'm pretty sure that would PYBIND11_INTERNALS_VERSION ABI breaking change (so internals still has
```
std::forward_list<ExceptionTranslator> registered_exception_translators;
```
not
```
std::forward_list<const ExceptionTranslator> registered_exception_translators;
```
as it would in your PR.

But I removed all the templating from the register functions. I also turned `ExceptionTranslator` into a `using` alias, which makes the typing a little bit nicer to read.

include/pybind11/pybind11.h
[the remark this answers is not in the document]
jesse-sony: this duplication has been removed and made into a helper. Let me know if you don't like something about my implementation.

rwgk:
Hi @jesse-sony, today I only got a chance to glance through your comments really quickly, not even the code. The comments I saw look good, thanks!. I'll come back here asap.

include/pybind11/detail/internals.h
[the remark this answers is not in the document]
rwgk: Make total sense to me, but I don't consider myself an authority on this, too. Adding this to the PR description is to expose it, in case someone is able to poke a hole into the rationale, we want to know asap.

include/pybind11/pybind11.h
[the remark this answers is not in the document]
rwgk: Yes, perfect, thanks. I did not mean to suggest adding the const. My PR was just to record that I thought about it carefully and then discarded the idea because I think it's too disruptive for too little gain.

include/pybind11/detail/internals.h
```diff
@@ -313,12 +313,19 @@ PYBIND11_NOINLINE inline internals &get_internals() {
     return **internals_pp;
 }
 
-/// Works like `internals.registered_types_cpp`, but for module-local registered types:
-inline type_map<type_info *> &registered_local_types_cpp() {
-    static type_map<type_info *> locals{};
-    return locals;
+
+struct local_internals {
+  type_map<type_info *> registered_local_types_cpp;
```
rwgk, line 322: Answering the question you asked in a different comment:
I'd prefer the 2nd alternative you pointed out there, with `local` **not** repeated inside the struct.
But I'd approve either.

rwgk (approved):
Looks awesome, thanks!

I'm still struggling a bit navigating this review system. If I overlooked a question you asked, please tag me again. (I tried to tie all my responses to the Review.)

I'm holding off merging until you make you final decision about the local_internals naming. Just let me know.

include/pybind11/detail/internals.h
jesse-sony: Names have been changed, since everyone seems to prefer them without the `local`

jesse-sony:
@rwgk Cleaned up the naming, updated the PR description with a note about why it does not break ABI compatability with the internals struct, and also add some notes in the internals.h file about the same ABI stuff. Happy to make other changes if there is more to be done, but I think I've hit everything mentioned so far.

rwgk:
The one failing CI job is a flake:
```
Run jwlawson/actions-setup-cmake@v1.9
Error: getaddrinfo ENOTFOUND api.github.com
```
Safe to ignore.

rwgk:
Thanks @jesse-sony for the change, @rhaschke, @YannickJadoul for the reviews, and @Skylion007 for pointing out this PR as one we should look at again. I'll merge this now.

PR #2651, opened 2020-11-11
Add full_type_id() to yield fully-qualified type name

rhaschke:
I found it handy for debugging to print the fully qualified type name, including reference specifiers.
This code originates from https://stackoverflow.com/questions/81870/is-it-possible-to-print-a-variables-type-in-standard-c

### DIFF
--- a/include/pybind11/detail/typeid.h
+++ b/include/pybind11/detail/typeid.h
@@ -52,4 +52,19 @@ template <typename T> static std::string type_id() {
     return name;
 }
 
+template <typename T> static std::string full_type_id() {
+    using TR = typename std::remove_reference<T>::type;
+    std::string name(type_id<TR>());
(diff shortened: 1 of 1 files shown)


YannickJadoul:
I'm not sure adding a dead code, for debugging purposes only, is a great idea from a maintenance point of view.

rhaschke:
Understandable. Nevertheless, using `full_type_id()` instead of `type_id()` might become handy in some existing unit tests or debug outputs of pybind11 as well, particularly if the reference specifiers are relevant.
But don't mind to reject this PR. I just wanted to propose it, because I have it lying around.

YannickJadoul:
Just my two cents, though. Others can chime in.

If this is would actually be used somewhere, that's a different story.

PR #2656, opened 2020-11-12
docs: better badges

henryiii:
## Description

Adding a few more badges and example links to the main page.

Some small fixes to the CI too; it seems the actions filter was filtering the out-of-date usage of the cmake action.

## Suggested changelog entry:

None needed, docs related.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,12 @@
 
 |Latest Documentation Status| |Stable Documentation Status| |Gitter chat| |CI| |Build status|
 
+|Repology| |PyPI package| |Conda-forge| |Python Versions|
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -5,6 +5,12 @@
 
 |Latest Documentation Status| |Stable Documentation Status| |Gitter chat| |CI| |Build status|
 
+|Repology| |PyPI package| |Conda-forge| |Python Versions|
```
YannickJadoul, line 8: A bit too late, but shouldn't we first put our official PyPI package, rather than all the things we don't control?

Issue #2215, opened 2020-05-15
Add noexcept(false) to destructors for gil_scoped_release


ezyang:
When a thread attempts to acquire the GIL but the Python interpreter has already destructed, Python will attempt to terminate the thread using `pthread_exit`. In many implementations of `pthread_exit`, this will trigger a stack unwinding, which will immediately call `std::terminate` if you are inside a destructor with `noexcept(true)`. Which is the case for the destructor of `gil_scoped_release`, which will attempt to acquire the GIL on reentry. The net effect of this any code which uses `gil_scoped_release` and is called from a daemon thread from Python is likely to cause your process to unceremoniously exit.

The fix seems to be quite simple:

```
diff --git a/include/pybind11/pybind11.h b/include/pybind11/pybind11.h
index 7fa0f0e..eb501a7 100644
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1954,7 +1954,7 @@ public:
             PYBIND11_TLS_DELETE_VALUE(key);
         }
     }
-    ~gil_scoped_release() {
+    ~gil_scoped_release() noexcept(false) {
         if (!tstate)
             return;
         PyEval_RestoreThread(tstate);
```

Do you agree with this change? If so I can submit a PR for it. More context: https://github.com/pytorch/pytorch/issues/38228

It would be reasonably simple to produce a repro that doesn't involve PyTorch, please let me know if that would be helpful.

bstaletic:
> When a thread attempts to acquire the GIL but the Python interpreter has already destructed

This is always a bug, which makes me wonder if `noexcept(false)` is a good idea here.

wjakob:
This is kind of a ugly situation but sometimes unavoidable in larger projects. The `noexcept` modifier would be okay for me, but perhaps add an extra comment (`/* enable stack unwinding when the interpreter is shutting down */`).

bstaletic:
I'm just thinking that what is undoubtedly a bug *should* make a big and loud crash, so that detecting such a bug is easier.

ezyang:
> This is always a bug, which makes me wonder if noexcept(false) is a good idea here.

[...]

PR #2657, opened 2020-11-13
Avoid thread termination in scoped_released

malfet:
## Description

Do not call `PyEval_RestoreThread()` from `~gil_scoped_release()` if python runtime is finalizing, as it will result in thread termination in Python runtime newer than 3.6, as documented in https://docs.python.org/3/c-api/init.html#c.PyEval_RestoreThread
Similarly do not call `PyThreadState_DeleteCurrent` from `~gil_scoped_acquire()` if runtime is finalizing.

Discovered while debugging PyTorch crash using Python-3.9 described in  https://github.com/pytorch/pytorch/issues/47776

## Suggested changelog entry:

```rst
Avoid unwanted termination if `gil_scoped_release()` is destroyed while Python runtime is finalizing 
```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -695,6 +695,19 @@ using expand_side_effects = bool[];
 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) (void)pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
 #endif
 
+// Thread state manipulation C API should not be called while Python runtime is finalizing.
+// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
+// `is_finalizing()` provides a version agnostic way to check if runtime is finalizing.
+inline bool is_finalizing() {
+#if PY_MAJOR_VERSION < 3
+    return false;
+#elif PY_VERSION_HEX >= 0x03070000
+    return _Py_IsFinalizing();
+#else
+    return _Py_Finalizing;
+#endif
+}
+
 PYBIND11_NAMESPACE_END(detail)
 
 /// C++ bindings of builtin Python exceptions
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -695,19 +695,6 @@ using expand_side_effects = bool[];
 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) (void)pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
 #endif
 
-// Thread state manipulation C API should not be called while Python runtime is finalizing.
-// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
-// `is_finalizing()` provides a version agnostic way to check if runtime is finalizing.
-inline bool is_finalizing() {
-#if PY_MAJOR_VERSION < 3
-    return false;
-#elif PY_VERSION_HEX >= 0x03070000
-    return _Py_IsFinalizing();
-#else
-    return _Py_Finalizing;
-#endif
-}
-
 PYBIND11_NAMESPACE_END(detail)
 
 /// C++ bindings of builtin Python exceptions
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -328,6 +328,19 @@ const char *c_str(Args &&...args) {
     return strings.front().c_str();
 }
 
+// Thread state manipulation C API should not be called while Python runtime is finalizing.
+// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
+// `finalization_guard()` provides a version agnostic way to check if runtime is finalizing.
+inline bool finalization_guard() {
+#if PY_MAJOR_VERSION < 3
+    return false;
+#elif PY_VERSION_HEX >= 0x03070000
+    return _Py_IsFinalizing();
+#else
+    return _Py_Finalizing;
+#endif
+}
+
 PYBIND11_NAMESPACE_END(detail)
 
 /// Returns a named pointer that is shared among all extension modules (using the same
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -328,19 +328,6 @@ const char *c_str(Args &&...args) {
     return strings.front().c_str();
 }
 
-// Thread state manipulation C API should not be called while Python runtime is finalizing.
-// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
-// `finalization_guard()` provides a version agnostic way to check if runtime is finalizing.
-inline bool finalization_guard() {
-#if PY_MAJOR_VERSION < 3
-    return false;
-#elif PY_VERSION_HEX >= 0x03070000
-    return _Py_IsFinalizing();
-#else
-    return _Py_Finalizing;
-#endif
-}
-
 PYBIND11_NAMESPACE_END(detail)
 
 /// Returns a named pointer that is shared among all extension modules (using the same
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2128,10 +2128,11 @@ public:
         }
     }
 
-    /// This method will disable the PyThreadState_DeleteCurrent call. This
-    /// should be called if the interpreter is shutting down, as the thread
-    /// deletion is not allowed during shutdown.
-    /// (_Py_IsFinalizing() on Python 3.7+)
+    /// This method will disable the PyThreadState_DeleteCurrent call and the
+    /// GIL won't be acquired. This method should be used if the interpreter
+    /// could be shutting down when this is called, as thread deletion is not
+    /// allowed during shutdown. Check _Py_IsFinalizing() on Python 3.7+, and
+    /// protect subsequent code.
     PYBIND11_NOINLINE void disarm() {
         active = false;
     }
@@ -2160,11 +2161,12 @@ public:
             PYBIND11_TLS_DELETE_VALUE(key);
         }
     }
-    //
-    /// This method will disable the PyThreadState_DeleteCurrent call. This
-    /// should be called if the interpreter is shutting down, as the thread
-    /// deletion is not allowed during shutdown.
-    /// (_Py_IsFinalizing() on Python 3.7+)
+
+    /// This method will disable the PyThreadState_DeleteCurrent call and the
+    /// GIL won't be acquired. This method should be used if the interpreter
+    /// could be shutting down when this is called, as thread deletion is not
+    /// allowed during shutdown. Check _Py_IsFinalizing() on Python 3.7+, and
+    /// protect subsequent code.
     PYBIND11_NOINLINE void disarm() {
         active = false;
     }
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2121,12 +2121,8 @@ public:
                     pybind11_fail("scoped_acquire::dec_ref(): internal error!");
             #endif
             PyThreadState_Clear(tstate);
-#if PY_VERSION_HEX >= 0x03070000
-            if (!_Py_IsFinalizing())
+            if (!detail::finalization_guard())
                 PyThreadState_DeleteCurrent();
-#else
-            PyThreadState_DeleteCurrent();
-#endif
             PYBIND11_TLS_DELETE_VALUE(detail::get_internals().tstate);
             release = false;
         }
@@ -2158,14 +2154,9 @@ public:
     ~gil_scoped_release() {
         if (!tstate)
             return;
-#if PY_VERSION_HEX >= 0x03070000
-        // PyEval_RestoreThread() should not be called if runtime is finalizing
-        // See https://docs.python.org/3/c-api/init.html#c.PyEval_RestoreThread
-        if (!_Py_IsFinalizing())
+        // `PyEval_RestoreThread()` should not be called if runtime is finalizing
+        if (!detail::finalization_guard())
             PyEval_RestoreThread(tstate);
-#else
-        PyEval_RestoreThread(tstate);
-#endif
         if (disassoc) {
             auto key = detail::get_internals().tstate;
             PYBIND11_TLS_REPLACE_VALUE(key, tstate);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1172,12 +1172,10 @@ inline void add_class_method(object& cls, const char *name_, const cpp_function 
 // For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
 // `is_finalizing()` provides a version agnostic way to check if runtime is finalizing.
 inline bool is_finalizing() {
-#if PY_MAJOR_VERSION < 3
-    return false;
-#elif PY_VERSION_HEX >= 0x03070000
+#if PY_VERSION_HEX >= 0x03070000
     return _Py_IsFinalizing();
 #else
-    return _Py_Finalizing;
+    return false;
 #endif
 }
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2121,7 +2121,12 @@ public:
                     pybind11_fail("scoped_acquire::dec_ref(): internal error!");
             #endif
             PyThreadState_Clear(tstate);
+#if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 6) && !defined(Py_LIMITED_API)
+            if (!_Py_IsFinalizing())
+                PyThreadState_DeleteCurrent();
+#else
             PyThreadState_DeleteCurrent();
+#endif
             PYBIND11_TLS_DELETE_VALUE(detail::get_internals().tstate);
             release = false;
         }
@@ -2153,7 +2158,14 @@ public:
     ~gil_scoped_release() {
         if (!tstate)
             return;
+#if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 6) && !defined(Py_LIMITED_API)
+        // PyEval_RestoreThread() should not be called if runtime is finilizing
+        // See https://docs.python.org/3/c-api/init.html#c.PyEval_RestoreThread
+        if (!_Py_IsFinalizing())
+            PyEval_RestoreThread(tstate);
+#else
         PyEval_RestoreThread(tstate);
+#endif
         if (disassoc) {
             auto key = detail::get_internals().tstate;
             PYBIND11_TLS_REPLACE_VALUE(key, tstate);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2121,7 +2121,7 @@ public:
                     pybind11_fail("scoped_acquire::dec_ref(): internal error!");
             #endif
             PyThreadState_Clear(tstate);
-#if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 6) && !defined(Py_LIMITED_API)
+#if PY_VERSION_HEX >= 0x03070000
             if (!_Py_IsFinalizing())
                 PyThreadState_DeleteCurrent();
 #else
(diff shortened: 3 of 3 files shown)


malfet:
@henryiii please let me know if change makes sense to you (and suggest who could review it)

include/pybind11/pybind11.h
```diff
@@ -2121,7 +2121,12 @@ class gil_scoped_acquire {
                     pybind11_fail("scoped_acquire::dec_ref(): internal error!");
             #endif
             PyThreadState_Clear(tstate);
+#if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 6) && !defined(Py_LIMITED_API)
```
bstaletic, line 2124: No point in checking `Py_LIMITED_API`, since pybind11 has never cared about that, but does pypy have the same problem, or does it behave better?

Also, `#if PY_VERSION_HEX >= 0x03060000` seems like a better check.

henryiii (approved):
Looks good to me. @rwgk & @YannickJadoul might want to check/verify.

YannickJadoul (approved):
If that's what the docs say, definitely! 

2 small things:
- What about 3.6 and before? Seems the behavior was just not documented? Do we need to always call it, or could we use the post-3.6 code for 3.6 and earlier as well?
- I guess there's no point in mixing compile time and runtime checks to collapse this into `if (PY_VERSION_HEX < 0x0307000 || !_Py_IsFinalizing())`? I guess it just changes two times one line of code duplication, but that's probably not worth the potential static analysis warnings?

YannickJadoul:
> If that's what the docs say, definitely!

Thanks for catching/fixing this, btw, @malfet! :-)

include/pybind11/pybind11.h
```diff
@@ -2153,7 +2158,14 @@ class gil_scoped_release {
     ~gil_scoped_release() {
         if (!tstate)
             return;
+#if PY_VERSION_HEX >= 0x03070000
+        // PyEval_RestoreThread() should not be called if runtime is finilizing
+        // See https://docs.python.org/3/c-api/init.html#c.PyEval_RestoreThread
+        if (!_Py_IsFinalizing())
+            PyEval_RestoreThread(tstate);
+#else
```
rwgk, line 2166: Minor: typo: finilizing

More significantly: I think the code could be made more readable and maintainable by centralizing the `#if #else #endif` in a, e.g., `detail::finalization_guard()` inline helper function, so that there is a formal link to the comment from both places, and less preprocessor clutter.

rwgk (commented):
How much trouble is it to add a minimal unit test that is expected to work reliably only with this change? (I'm OK without adding a tests if the setup is too much trouble.)

include/pybind11/pybind11.h
YannickJadoul: Sure that that's not just going to add more confusion, to have another indirection? As long as it's just these two (very related) spots, to me, this seem reasonably fine? (We currently have these kinds of things inline in other places as well, and this should make it easy to remove once we drop <= 3.6?)

henryiii:
> How much trouble is it to add a minimal unit test that is expected to work reliably only with this change

If it only happens on interpreter finalization, it would have to be a test that spawns a Python instance. Probably not trivial, but not awful either (we have some already like that).

include/pybind11/pybind11.h
henryiii: The practical version of DRY/SPOT that I've seen states that exactly one duplication is allowed, if it's clearly more complex to combine it. Trying to take DRY/SPOT to the extreme of truly one point of truth for all code can impose an unmanageable level of complexity, so allowing it to be relaxed by one at programmer discretion balances the ideal with practicality. (and if there's more than one duplication, the extra complexity is always worth it).

In short, I'm fine either way.

rwgk: Thanks Henry! I hadn't heard of DRY/SPOT before, good to know there is something to refer to.

In this particular case I'm not so much concerned about the code duplication, but mostly about the comment only appearing in one place. For someone arriving at the second place from some completely different context, they will not even know the comment exists in the other place, and may miss it. But if they get curious about `detail::finalization_guard()`, they are guaranteed to find the comment.

YannickJadoul: The comment is true, though! I completely agree there. There should at least be a "see a few lines up/down, at ...".

Apart from that, I don't feel too strongly, so up to @malfet AFAIC, but I just though the `#if SOME_PYTHON_VERSION` pattern exists inline in lots of places of code.
And it's kind of hard to capture what it's doing (i.e., `finalization_guard` doesn't really describe what's happening) ?

malfet: Thank you for the feedback. Fixed typo and moved and extended runtime finalization check to `pybind11/detail/internals.h`

rwgk (approved):
Looks good to me, thanks!
Super minor, in two place you have "runtime is finalizing". It feels like a "the" is missing: "the runtime is finalizing".

malfet:
> * What about 3.6 and before? Seems the behavior was just not documented? Do we need to always call it, or could we use the post-3.6 code for 3.6 and earlier as well?

I don't know to tell the truth, but there is a `_Py_Finalizing` global variable in 3.5 and 3.6, which probably should be used to detect if runtime is winding down.

bstaletic:
Suppose we have two threads. One is holding the GIL and started to finalize the interpreter. The other wants to use a CPython API and so first calls `py::gil_scoped_acquire guard;`. However, `py::gil_scoped_acquire` does exactly nothing because the interpreter is shutting down, thus the thread "thinks" it acquired the GIL, when it actually didn't. Now that thread accessing some CPython API without holding the GIL. We all know that's no good.

Any thoughts regarding that?

include/pybind11/detail/internals.h
```diff
@@ -328,6 +328,19 @@ const char *c_str(Args &&...args) {
     return strings.front().c_str();
 }
 
+// Thread state manipulation C API should not be called while Python runtime is finalizing.
+// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
+// `finalization_guard()` provides a version agnostic way to check if runtime is finalizing.
+inline bool finalization_guard() {
```
YannickJadoul, line 334: This should definitely not be in `internals.h`, though. This is not related to the pybind11 internals (which have a specific meaning, being linked to the internal information pybind11 stores about types and instances).

Please move it back to somewhere related to the threads. Or potentially `common.h`, if this could be used in other places; but I don't think so, since the comment you wrote above is very specific to the threading functionality.

I would also want `finalization_guard` to get a better name, showing that it's linked to these threads, such that it doesn't get reused for other things, since it doesn't report "finalizing" correctly on Python 2? Or is there no concept of "finalizing" on Python 2?

YannickJadoul (changes requested):

include/pybind11/detail/internals.h
malfet: @YannickJadoul , how about `internal::is_finalizing()`? And I don't know much about Python-2 runtime, but quick grep thru the headers suggests that there isn't any.

YannickJadoul: No, we have no `internal` namespace, so `detail`, you mean?

Then, yes, perhaps, but then the documentation you have on top should partially move back to where you use it? "Thread state manipulation C API should not be called while Python runtime is finalizing." (and accompanying link, etc.) does not belong to `is_finalizing`, then, but to where you use it, I assume?

YannickJadoul: Also, now that I think about it: if we don't know how to detect whether Python 2 is finalizing, this is a reasonably dangerous API to create, no?

malfet: Sorry, that was a typo, I've meant `detail::is_finalizing()`. 
As for Python-2: as there are no concept of runtime finalization in Python 2, therefore `is_finalizing()` should always return false to indicate that all C APIs are safe to call all the time.

YannickJadoul: Here's `Py_Finalize` in 2.7: https://github.com/python/cpython/blob/2.7/Python/pythonrun.c#L411

I indeed can't immediately spot some finalizing flag being set.

YannickJadoul: (While I can, on `master`: https://github.com/python/cpython/blob/296a796951032f678d063008f588ccc6958d0df1/Python/pylifecycle.c#L1636)

include/pybind11/detail/common.h
```diff
@@ -695,6 +695,19 @@ using expand_side_effects = bool[];
 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) (void)pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
 #endif
 
+// Thread state manipulation C API should not be called while Python runtime is finalizing.
+// For more detail see https://docs.python.org/3.7/c-api/init.html#c.PyEval_RestoreThread
```
YannickJadoul, line 699: I'm not entirely convinced this should be here (instead of at the point where it's used), if you're re-purposing this `is_finalizing` to be more generic?

YannickJadoul (commented):
Yup, this already makes me happier to see `detail::is_finalizing` and not in `internals.h` :-)

I do still want to see @bstaletic's comment addressed though. I don't know about all the GIL/threading internals of Python, but @bstaletic's points seems to be a valid concern, where we might be breaking other code?

include/pybind11/detail/internals.h
malfet: > Here's `Py_Finalize` in 2.7: https://github.com/python/cpython/blob/2.7/Python/pythonrun.c#L411
> I indeed can't immediately spot some finalizing flag being set.

Sorry, I've meant there are no finalization flag/API that Python-2 exposes and there are doesn't seem to be any checks in `PyGILState_Ensure()` that prohibit calling this function while/after `Py_Finalize()` has been called.

YannickJadoul: Perfect! Thanks for making sure :-)

YannickJadoul: (That's what I also meant, btw, that I couldn't find any evidence of Python 2 having a "finalizing" state)

include/pybind11/detail/common.h
malfet: I don't have a strong preference on the method location, moved it back to `pybind11.h` (haven't realized that you guys are OK with mixing several namespaces in the same header file)

YannickJadoul: Sorry if I didn't make myself understood correctly. I was talking about the two lines of documentation (as mentioned before: https://github.com/pybind/pybind11/pull/2657#discussion_r525238495). If you preferred, the function could stay in `common.h`.

malfet:
> Suppose we have two threads. One is holding the GIL and started to finalize the interpreter. The other wants to use a CPython API and so first calls `py::gil_scoped_acquire guard;`. However, `py::gil_scoped_acquire` does exactly nothing because the interpreter is shutting down, thus the thread "thinks" it acquired the GIL, when it actually didn't.

This scenario is unrelated to the change, as it does not change the behavior of `py::gil_scoped_acquire` constructor.  But in the separate PR (or may be even in the same one) I would argue, that pybind11_fail() should be raised from constructor if runtime is being finalized to match the termination behavior documented at https://docs.python.org/3/c-api/init.html#c.PyEval_AcquireThread

bstaletic:
> the change [...] does not change the behavior of py::gil_scoped_acquire constructor. 

Right, my bad. What about a thread that's doing somethting like this?

```c++
auto the_other_thread() {
	{
		pybind11::gil_scoped_release g;
		while(!_Py_IsFinalizing());
	}
	pybind11::list l;
	for(auto i = 0; i < 1000000; ++i) l.append(pybind11::str("something long"));
}

PYBIND11_MODULE(foo, m) {
	m.def("the_other_thread", the_other_thread);
}
```

Obviously, this snippet shouldn't pass any kind of code review, but the point is that the thread might attempt to do anything with the CPython API. Before this PR, it would try to acquire the gil and would get killed. This makes the problem manifest as close as possible to the cause of the same problem. How do things change with this PR? If I understand things correctly, since the destructor of `gil_scoped_release` doesn't acquire the GIL, "the other thread" ends up playing with the python API, also, without holding the GIL. Again, if I'm correct, that would make the cause (attempting to acquire the GIL at a horrible moment) "far away" from the effect (hopefully a SEGFAULT).

malfet:
> Obviously, this snippet shouldn't pass any kind of code review, but the point is that the thread might attempt to do anything with the CPython API. Before this PR, it would try to acquire the gil and would get killed. This makes the problem manifest as close as possible to the cause of the same problem. How do things change with this PR? If I understand things correctly, since the destructor of `gil_scoped_release` doesn't acquire the GIL, "the other thread" ends up playing with the python API, also, without holding the GIL. 
Hmm. In that case, would throwing a runtime_error exception results in a better behaviour?

bstaletic:
We're talking about `~gil_scoped_release()`. In C++ destructors are `noexcept` by default, so throwing from one would actually call `std::terminate()`. Before anyone proposes adding `noexcept(false)` to the destructor, the problem with that is that you can't have two exceptions in flight at the same time, else `std::terminate()`. So, if `~gil_scoped_release()` was called because something already "exploded" and threw an exception then throwing a `runtime_error` from `~gil_scoped_release()` gets us back to square 1 - the thread gets terminated.

Now my first thought was "bad things are already happening, it's no worse than right now", but that might not be true. If we allow `~git_scoped_release()` to throw, then, as far as we're aware, there would be to "sources" of exceptions from inside this destructor.

1. The proposed runtime error.
2. https://github.com/pybind/pybind11/issues/2215

To be clear, throwing destructors still make me very anxious. However, this *might* be a non-evil exception. If we are absolutely 100% positive that only the code that was already horribly broken (such as my snippet above) would stay horribly broken, this throwing destructor might be okay. To whomever is still reading, when thinking about this, consider a really bad situation:

1. One thread is finalizing the interpreter.
2. The other thread is detached and actually outlives the interpreter shutting down, as in #2215 
3. That same other thread also throws a C++ exception, causing the stack to unwind, instead of return normally.
4. The `~gil_scoped_release()` also tries to throw an exception.

#2215 left one thing unanswered. In the bad scenario described in #2215, we know the behaviour of POSIX systems, but we have no idea how Windows would behave, as that question was left unanswered. Which, to be honest, makes me even more wary of throwing an exception from `~gil_scoped_release()`.

___

Anyway, that was me trying to convince myself that a throwing destructor might be a viable option. A completely different solution, though not great, could be a sort of an escape hatch. Something like `gil_scoped_acquire::release()` and `gil_scoped_release::release()` that would "disarm" the appropriate member function to make it do nothing. This would pass the ball back to the users, making them responsible for disarming these RAII guards if they are sure that disarming is safe.

And yes, I'm fully aware of the downsides that are involved in actually using such a mechanism. I'm just thinking out loud. None of the ideas I've thought of are perfect and I'm absolutely open for ideas. I'm also open to pybind11 adopting a least bad solution as long as we understand what the implications are.

henryiii:
How about `gil_scoped_X::deactivate()`? Then you'd call:

```cpp
auto the_other_thread() {
	{
		pybind11::gil_scoped_release g;
		while(!_Py_IsFinalizing());
                if(_Py_IsFinalizing())
                    g.deactivate();
	}
        // Now it's obvious that you may not have the gil here, and you could add an if _Py_IsFinalizing()
	pybind11::list l;
	for(auto i = 0; i < 1000000; ++i) l.append(pybind11::str("something long"));
}
```

henryiii:
And, of course, for pytorch, you could then subclass these two guards, and add a conditional if that calls this in the destructor.

henryiii:
Should the milestone get moved to 2.7? I think we decided on `disarm`?

include/pybind11/pybind11.h
```diff
@@ -2132,13 +2121,21 @@ class gil_scoped_acquire {
...
                 PyThreadState_DeleteCurrent();
             PYBIND11_TLS_DELETE_VALUE(detail::get_internals().tstate);
             release = false;
         }
     }
 
+    /// This method will disable the PyThreadState_DeleteCurrent call. This
+    /// should be called if the interpreter is shutting down, as the thread
+    /// deletion is not allowed during shutdown.
+    /// (_Py_IsFinalizing() on Python 3.7+)
```
bstaletic, line 2134: I feel like this comment is too friendly and thus doesn't really convey the dangers of disarming this RAII object.

include/pybind11/pybind11.h
```diff
@@ -2176,24 +2183,31 @@ class gil_scoped_release {
...
 };
 #else
-class gil_scoped_acquire { };
-class gil_scoped_release { };
+class gil_scoped_acquire {
+    void disarm() {}
+};
+class gil_scoped_release {
+    void disarm() {}
+};
```
bstaletic, line 2210: Could have kept this as two lines, but no big deal.

henryiii: I don't like nesting just to save space. It makes it harder to read - functions should always be on their own line.

include/pybind11/pybind11.h
henryiii:
```suggestion
    /// This method will disable the PyThreadState_DeleteCurrent call. This
    /// should be called if the interpreter is shutting down, as the thread
    /// deletion is not allowed during shutdown. Be sure you do not try to
    /// rely on this guard if you have disarmed it!
    /// (Use _Py_IsFinalizing() on Python 3.7+ to detect shutdown)
```

henryiii: ?

include/pybind11/pybind11.h
bstaletic: That's fine with me.

include/pybind11/pybind11.h
YannickJadoul: Sound good, yes.
The one thing that's missing is that it's not entirely clear why we don't do the check ourselves (cfr. this original PR). But if @bstaletic is happy, this is great to me! :-)

YannickJadoul (approved):
Cool, this is pretty slim/clean!

2 more things:
- Do we need a test? As far as I'm concerned, we're following the Python docs and a test would probably be more hazardous than the actual straightforwards code changes we have here. Also the GIL RAII classes are already tested.
- Can we confirm this solution also works for @malfet? In principle, you could now still wrap this in a PyTorch-specific RAII wrapper, right? It's just that at least `disarm` will show the danger and not do dangerous things by default, but functionally, the same should be possible.

include/pybind11/pybind11.h
bstaletic: How... How did I not reply to this one??

I was thinking more of an explicit warning that, if disarmed, the RAII guard won't acquire the GIL. Also the `should be called` can lead to confusion whether a user *should* actually call `disarm()` anyway. Here's my proposal that I apparently forgot to post:

```
This method will disable the PyThreadState_DeleteCurrent call and the GIL won't be acquired. This
may be called if the interpreter is shutting down, as the thread
deletion is not allowed during shutdown.
```

To summarize:

1. Mention the GIL explicitly.
2. Use `may` instead of `should`.
3. Optionally restate the warning at the end of the comment, like you already proposed.
4. Optionally mention this issue discussion, like @YannickJadoul proposed, though we do have `git blame` already.
5. Don't forget the other similar comment below.

Anyway, I'm not going to nag too much about the exact wording. I just want to drive home the point that this `disarm()` is dangerous so that any future user knows what he's getting into. All that said, I'll approve

YannickJadoul: >     4\. Optionally mention this issue discussion, like @YannickJadoul proposed, though we do have `git blame` already.

For the record, I didn't really mean this; I'm not a huge fan of (too many) links to GitHub in the code itself. I mostly meant the current explanation potentially leaves open the question on "why did you not implement this in the `gil_scoped_acquire` destructor?", since it doesn't show doing it during shutdown is still reasonably dangerous, depending on the situation;

henryiii: I'll modify just a bit, but that's about right. I do think it is "Should" - you really should do this or you will get a segfault. The reason we don't do it by default is it is _also_ dangerous to assume this is always armed; if you have a `if ... .disarm()`, then hopefully a user will also check the same condition instead assuming the guard is always armed.

In short: not calling disarm if you know the interpreter may be shutting down when this is called is also dangerous.

How about this:

> This method will disable the PyThreadState_DeleteCurrent call and the GIL won't be acquired. This
  method should be used if the interpreter could be shutting down when this is called, as the thread
  deletion is not allowed during shutdown. Check _Py_IsFinalizing() on Python 3.7+, and protect subsequent code.

henryiii:
No test, unless we want to just call it once to make sure it doesn't vanish into thin air someday. Trying to run a test on interpreter shutdown is likely very tricky.

include/pybind11/pybind11.h
bstaletic: That sounds good to me.

YannickJadoul: Great! :-)

YannickJadoul:
> No test, unless we want to just call it once to make sure it doesn't vanish into thin air someday. Trying to run a test on interpreter shutdown is likely very tricky.

Agreed!

bstaletic:
> Trying to run a test on interpreter shutdown is likely very tricky.

I tried and that's an understatement.

PR #2662, opened 2020-11-15
fix: make FindPython2 and FindPython3 work

### LINKED ISSUE #2664: [BUG] cmake lead to missng symbols when linking against `pybind11::embed` while preloading python with `find_package(Python3)`
Pre-loading python3 in cmake leads to the target not being linked against python library.
```
find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
find_package(pybind11 2.3.0 REQUIRED)
add_library(my_target [...])
target_link_libraries(my_target pybind11::embed)
```

I believe this is due to the following line linking against `Python3::Python3` in that case instead of `Python3::Python`provided by the cmake module
https://github.com/pybind/pybind11/blob/02746cb69f202579efb0004e55ac7984a5c466bb/tools/pybind11NewTools.cmake#L147
https://github.com/pybind/pybind11/blob/02746cb69f202579efb0004e55ac7984a5c466bb/tools/pybind11NewTools.cmake#L151

knarfS:
## Description

When using FindPython3 (or FindPython2) from outside for a project that embeds the Python interpreter (with `pybind11:embed`), pybind11 isn't detecting Python correctly. This results in a working cmake execution, but with linking errors after building. FindPython is working though.
This is due to wrong target names (e.g. `Python3::Python3` instead of `Python3::Python`) used in `pybind11NewTools.cmake`.

This was tested under Debian testing with an actual pybind11 git clone, cmake 3.16.3 and Python 3.8.5

## Suggested changelog entry:

```rst
fix: make FindPython2 and FindPython3 work
```

Close #2664

### DIFF
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -144,11 +144,11 @@ if(DEFINED ${_Python}_VERSION AND ${_Python}_VERSION VERSION_LESS 3)
 endif()
 
 # In CMake 3.18+, you can find these separately, so include an if
-if(TARGET ${_Python}::${_Python})
+if(TARGET ${_Python}::Python)
   set_property(
     TARGET pybind11::embed
     APPEND
-    PROPERTY INTERFACE_LINK_LIBRARIES ${_Python}::${_Python})
+    PROPERTY INTERFACE_LINK_LIBRARIES ${_Python}::Python)
 endif()
 
 # CMake 3.15+ has this


henryiii (approved):
This looks correct. We only test regular FindPython in CI.

YannickJadoul:
> This looks correct. We only test regular FindPython in CI.

I guess you're the CI expert, so if you feel like merging this, go ahead, @henryiii!

YannickJadoul:
This also closes #2664, btw, as remarked there, already.

knarfS:
Thanks for merging so quickly! :)

Issue #2658, opened 2020-11-13
[FEAT] def_async


hiaselhans:
i made a simple Awaitable class in c++ to support pythons async/await syntax with c++ deferred functions:

```
template<typename ResultType>
class Awaitable : public std::enable_shared_from_this<Awaitable<ResultType>>{
    public:
        Awaitable();
        Awaitable(std::future<ResultType>&);
        std::shared_ptr<Awaitable<ResultType>> __iter__();
        std::shared_ptr<Awaitable<ResultType>> __await__();
        
        void __next__();

    private:
        std::future<ResultType> future;
};
```

is this something that would have a chance to land in pybind11?
So, supporting an interface like:

`.def_async("name", &X:function_returning_a_future)`

What seems to be missing for this is a StopIteration exception type with templated "return" type.

YannickJadoul:
What would need to happen in pybind11 for this? Could this e.g. work as some kind of extension in a `py::async_class` in a separate `<pybind11/async.h>` header? (or project, even; depends a bit on how much code/maintenance this would be)

hiaselhans:
thx for the quick answer @YannickJadoul!

I wanted to see first how you would agree with it, but i can get my head around it and create a PR for it.
It would definitely take me some time but my wish would be to have .def_async for both: instance methods and python methods. They would require a function to return an std::future and python would receive an awaitable.

It's not a big deal to implement it in a few lines of c++, but with python (and other languages) moving more and more towards using async/await it could be a good feature to provide.

From your comment i read that you prefer to have it separate from py::class and module.def?

YannickJadoul:
> I wanted to see first how you would agree with it, but i can get my head around it and create a PR for it.

> It's not a big deal to implement it in a few lines of c++, but with python (and other languages) moving more and more towards using async/await it could be a good feature to provide.

Not the only one going over this, btw, but if there are use cases for this, and it can be done cleanly & without overhead (but runtime as well as binary size), then I don't see why this wouldn't be welcome! :-)

> instance methods and python methods

How do you mean, exactly? I'm not sure I see the distinction.

> From your comment i read that you prefer to have it separate from py::class and module.def?

Possibly. Just curious how easy this would be. Not per se for this specifically, but in general, it would be strive towards making things as extensible as possible, such that `py::class_` doesn't end up with dozens of methods, and it would probably help to make things "opt-in"?

hiaselhans:
after some tries, i think, for now, there is one main pitfall for extensibility:

[...]

PR #2663, opened 2020-11-15
add async_class and async_class::def_async

hiaselhans:
## Description

As indicated in #2658 this is a working draft to implement def_async on a py::class_async.
It basically starts a new thread and returns a Waitable object which conforms with pythons coroutine API.

## TODO

- [ ] add docs (including a warning that c++ methods must be thread safe!)
- [ ] write some tests
- [ ] add return type to StopIteration
- [ ] unfortionately cpp_function's initialize can not be virtual because it is templated. Maybe there is a more elegant approach than copying all constructors.
- [ ] class_::def returns a downcasted class_ 'this' reference. We'd need some kind of covariant functions here. for now my approach is to put all .def_async before any .def calls.

Any comments welcome


hiaselhans:
ping @YannickJadoul :)

this is how i used it:

```c++
PYBIND11_MODULE(async_module, m) {

    py::class_async<Solver>(m, "Solver")
        .def_async("run_steps_async", &Solver::run_steps)
        .def(py::init<>());

    py::class_<py::Awaitable<void>, std::shared_ptr<py::Awaitable<void>>>(m, "Awaitable")
        .def(py::init<>())
        .def("__iter__", &py::Awaitable<void>::__iter__)
        .def("__await__", &py::Awaitable<void>::__await__)
        .def("__next__", &py::Awaitable<void>::__next__);
}
```

then use it in python:

```python
async def run():
    solver = async_module.Solver()
    await solver.run_steps_async()
```

another TODO is to have the Awaitable as py::class somewhere in pybind11's headers...

include/pybind11/async.h
```diff
@@ -0,0 +1,142 @@
...
+    public:
+        Awaitable() {
+            this->future = std::future<ResultType>();
+        };
+
+        Awaitable(std::future<ResultType>& _future) {
+            this->future = std::move(_future);
+        };
+
+        std::shared_ptr<Awaitable<ResultType>> __iter__() {
```
YannickJadoul, line 31: This is bad practice in C++, no? Aren't double underscores reserved for the language/compiler or so?

include/pybind11/async.h
```diff
@@ -0,0 +1,142 @@
+#pragma once
+
+#include <memory>
+#include <future>
+#include <map>
+#include <algorithm>
+#include <sstream>
+
+#include <chrono>
+#include <iostream>
```
YannickJadoul, line 10: Do we need all these? At least `<iostream>` and `<sstream>` are definitely not used, right?

include/pybind11/async.h
```diff
@@ -0,0 +1,142 @@
...
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+#include "pybind11/embed.h"
+#include "pybind11/operators.h"
+
+
+namespace pybind11 {
+
+template<typename ResultType>
+class Awaitable : public std::enable_shared_from_this<Awaitable<ResultType>>{
```
YannickJadoul, line 21: How does pybind11 know how to map this to Python?

YannickJadoul: But OK, yes. Maybe this could become some kind of custom caster or convenience macro or so.

include/pybind11/async.h
hiaselhans: right, i will change it and just leave the py interface with double underscores!

include/pybind11/async.h
hiaselhans: you are right, i just picked everything from where i had it before :) still have to clean up...

include/pybind11/async.h
hiaselhans: thx @YannickJadoul !

so basically you are suggesting to have some sort of a .def(_, _, py::async()) argument?
and some sort of py::enable_async() macro for the module definition?

YannickJadoul:
> so basically you are suggesting to have some sort of a .def(_, _, py::async()) argument?

Yes. Something similar to e.g. `is_operator()` or so?

> and some sort of py::enable_async() for the module definition?

Not really. Why would this be necessary? What would it do?

hiaselhans:
> > so basically you are suggesting to have some sort of a .def(_, _, py::async()) argument?
> 
> Yes. Something similar to e.g. `is_operator()` or so?
> 
I will checks, thx for pointing me there... :)

> > and some sort of py::enable_async() for the module definition?
> 
> Not really. Why would this be necessary? What would it do?

I meant to simplify mapping to the "Awaitable" class...
it would rather be py::async_ or something similar.

Of course there are other options, but that seems to be the "most" opt-in

YannickJadoul:
> Just to simplify mapping to the "Awaitable" class...
> And I guess you are right, it could rather be a py::async_ class or something similar.
> 
> Of course there are other options, but that seems to be the "most" opt-in to me?

Hmmm, well, the problem is that `Awaitable` is a class template, right? So you need to expose each return type separately?

At any rate, I'd just go for some kind of utility function call like `bind_awaitable<void>()` or so?

YannickJadoul:
Also, Python has `asyncio.Future`, [apparently](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future). Do we need a specific `py::class_` ourselves?

hiaselhans:
Thx for your comments @YannickJadoul !

Thanks for the hint to use asyncio.Future. I was playing around a bit with and got to this test case:

```cpp
#include <thread>

PYBIND11_MODULE(example, m) {

    m.def("async_function", [](int time){
                    auto py_async = py::module::import("asyncio");
                    auto py_future = py_async.attr("Future")();
                    
                    auto long_running_function = [&py_future, &time]() {
                        std::this_thread::sleep_for(std::chrono::milliseconds(time));
    
                        py::gil_scoped_acquire gil;
                        py_future.attr("set_result")(py::cast(time));
                    };
    
                    std::thread my_thread (long_running_function);
                    my_thread.detach();
    
                    return py_future;
    
                });
}

```

however it does not really work this way. Segfaults in plain python and hangs in jupyter.
using std::async()  doesn't work either because the future would be destroyed and thus the function would be blocking.

Do you have any idea about how to get it working with asyncio.Future? Or do you think it could work if this PR proposed a more generic Awaitable<py::object> ?

hiaselhans:
i did some more investigation and found that the main problem is that async's event loop relies on asyncio.Future's callback being called which isn't the case from the detached thread...

i really don't know how and why, but i think a genergic Awaitable<py::object> class might work...

hiaselhans:
so, this is what i came up with:

```cpp
class StopIteration : public py::stop_iteration {
    public:
        StopIteration(py::object result) : stop_iteration("--"), result(result) {};
        //using py::stop_iteration::stop_iteration;
        /// Set the error using the Python C API
        void set_result(py::object result) {
            this->result = result;
        }
        void set_error() const override {
            PyErr_SetObject(PyExc_StopIteration, this->result.ptr()); 
        }
    private:
        py::object result;
};


class Awaitable : public std::enable_shared_from_this<Awaitable>{
    public:
        Awaitable() {
            this->future = std::future<py::object>();
        };

        Awaitable(std::future<py::object>& _future) {
            this->future = std::move(_future);
        };

        std::shared_ptr<Awaitable> __iter__() {
            return this->shared_from_this();
        };

        std::shared_ptr<Awaitable> __await__() {
            return this->shared_from_this();
        };

        void __next__() {
            // check future status (zero timeout)
            auto status = this->future.wait_for(std::chrono::milliseconds(0));

            if (status == std::future_status::ready) {
                // future is ready -> raise StopInteration with the future result set
                auto exception = StopIteration(this->future.get());
                //exception.set_result(this->future.get());

                throw exception;
            }
        };

    private:
        std::future<py::object> future;
};


PYBIND11_MODULE(async_cpp, m) {
    m.doc() = "pybind11 async test";

    py::class_<Awaitable, std::shared_ptr<Awaitable>>(m, "Awaitable")
        .def(py::init<>())
        .def("__iter__", &Awaitable::__iter__)
        .def("__await__", &Awaitable::__await__)
        .def("__next__", &Awaitable::__next__);
    

    m.def("async_function", [](int time){              
        auto long_running_function = [](int time) {
            std::this_thread::sleep_for(std::chrono::milliseconds(time));
            return time+1;
        };
        auto function_wrapper = [long_running_function, time]() {
            auto result = long_running_function(time);

            py::gil_scoped_acquire gil;
            
            auto py_result = py::cast(time+1);
            return py_result;
        };

        auto future = std::async(std::launch::async, function_wrapper);
        auto awaitable = Awaitable(future);

        return awaitable;

    });

}
```

```python
import async_cpp
import time

t1=time.time()
awaitables = []
for _ in range(10):
    awaitables.append(async_cpp.async_function(2000))

await asyncio.gather(*awaitables)
print(time.time() - t1)


>>> 2.0256118488311768
```


What do you think about this approach? would it make sense to allow for all py::exceptions to be initialized with py::object in general?

0lru:
> however it does not really work this way. Segfaults in plain python and hangs in jupyter.

When using futures, I think you have to fulfill the future in the loop of the caller.
You need to get the current async loop and then you invoke  "set_result" with "call_soon_threadsafe" of that loop.
Pseudocode:
```
loop = py::whatever(asyncio.get_running_loop())
future = py::whatever(create_future);
_thead = std::thread([loop, future]() {
    <do something>
    py::whatever(loop.call_soon_threadsafe(future, set_result, result));
})
```

If you detach the thread or keep it's instance is of course an implementation detail.

hiaselhans:
Thanks for your comment @0lru 

I am using whats in this pr in production.
I found it easier to use a plain c++/pybind11 object which goes to resolved state after the detached thread has finished.

However this is still a bit messy and i think it could be improved by making it a 'policy'/call_guard.
I just didn't find the time for that yet, but i think this could be a good addon then.

0lru:
For completeness (working example):
```cpp
            auto py_async = py::module::import("asyncio");
            auto py_loop = py_async.attr("get_event_loop")();
            auto py_future = py_loop.attr("create_future")();
            std::thread([
                py_future,
                py_loop,
            ]() {
                try
                {
                    std::this_thread::sleep_for(std::chrono::seconds(10));
                    {
                        py_loop.attr("call_soon_threadsafe")(py_future.attr("set_result"), py::none());
                    }
                }
                catch (...)
                {
                    auto call_soon_threadsafe = py_loop.attr("call_soon_threadsafe");
                    call_soon_threadsafe(py_future.attr("set_exception"), py::cast(std::current_exception()));
                }
            }).detach();
...
            <do something with the future (await)>

```

include/pybind11/async.h
```diff
@@ -0,0 +1,208 @@
...
+            auto proxy = [f](Args... args) -> Awaitable* {
+                auto thread_func = [f](Args... args) {
+                    f(std::forward<Args>(args) ...);
+
+                    py::gil_scoped_acquire gil;
+
+                    auto py_result = py::cast(Py_None);
+                    return py_result;
+                };
+                auto bound_thread_func = std::bind(thread_func, std::forward<Args>(args)...);
```
Skylion007, line 170: std::bind is now forbidden in the code base (merge in the new clang-tidy rules). Should be a lambda instead.

include/pybind11/async.h
```diff
@@ -0,0 +1,208 @@
...
+PYBIND11_NAMESPACE_BEGIN(async)
+
+
+class StopIteration : public py::stop_iteration {
+    public:
+        StopIteration(py::object result) : stop_iteration("--"), result(result) {};
+        //using py::stop_iteration::stop_iteration;
+        /// Set the error using the Python C API
+        void set_result(py::object result) {
+            this->result = result;
```
Skylion007, line 23:
```suggestion
            this->result = std::move(result);
```
or make it take a const ref instead.

include/pybind11/async.h
```diff
@@ -0,0 +1,208 @@
...
+        Awaitable() {
+            this->future = std::future<py::object>();
+        };
+
+        Awaitable(std::future<py::object>& _future) {
+            this->future = std::move(_future);
+        };
+
+        std::shared_ptr<Awaitable> iter() {
+            return this->shared_from_this();
```
rwgk, line 44: pybind11 maintains a registry of instances. I think you could simply `return this;`, it will reuse the existing Python instance. Then you also don't need to inherit from enable_shared_from_this and you can just use the default holder.

include/pybind11/async.h
```diff
@@ -0,0 +1,208 @@
...
+
+template <typename type_, typename... options>
+class class_async : public class_<type_, options...> {
+
+    using type = type_;
+    public:
+        // using parent constructor
+        using class_<type_, options...>::class_;
+
+        template <typename Func, typename... Extra>
```
rwgk, line 196: Could we add it into `class_`? That would seem nicer.

include/pybind11/async.h
hiaselhans: Thanks for the hint! i updated the pr accordingly

include/pybind11/async.h
hiaselhans: thanks for the hint! it wasn't used so i removed it altogether...

include/pybind11/async.h
hiaselhans: that would be a lot shorter, yes.

I think the cleanest would be to have something similar to a call_guard, but thats out of scope for me.

include/pybind11/async.h
hiaselhans: fixed, thanks!

hiaselhans: NOT!

i tried the two following but it's causing segfaults:

``` c++
auto bound_thread_func = [&args..., &thread_func]{
    return thread_func(std::forward<Args>(args)...);
};
```

``` c++
auto bound_thread_func = [&args..., &thread_func]{
    return thread_func(args...);
};
```

i added a minimal reproducer down below if you have an idea..

hiaselhans:
thanks for the comments!

here is a minimal reproducer:

```c++
#include <chrono>
#include <thread>

#include <pybind11/async.h>

namespace py = pybind11;
using namespace py::literals;
using namespace std::chrono_literals;

class Solver {
    public:
        unsigned int run_steps_async(const unsigned int n) {
            unsigned int i;
            for (i=0; i<n; i++) {
                std::this_thread::sleep_for(100ms);
                this->count++;
            }

            return this->count;
        }
    
    private:
        unsigned int count;
};


PYBIND11_MODULE(test, m) {
    m.doc() = "test module";

    py::async::class_async<Solver>(m, "Solver")
        .def_async("run_steps_async", &Solver::run_steps_async)
        .def(py::init());
    
    py::async::enable_async(m);
}
```

```python
import asyncio
import test

solver = test.Solver()

loop = asyncio.get_event_loop()

print("start")
result = loop.run_until_complete(solver.run_steps_async(30))

print(f"did {result} steps")

```

include/pybind11/async.h
```diff
@@ -0,0 +1,205 @@
...
+            auto proxy = [f](Args... args) -> Awaitable* {
+                auto thread_func = [f](Args... args) {
+                    auto result = f(std::forward<Args>(args) ...);
+
+                    py::gil_scoped_acquire gil;
+
+                    auto py_result = py::cast(result);
+                    return py_result;
+                };
+                auto bound_thread_func = std::bind(thread_func, std::forward<Args>(args)...);
```
Skylion007, line 136: Another std::bind

include/pybind11/async.h
```diff
@@ -0,0 +1,205 @@
...
+};
+
+
+class Awaitable {
+    public:
+        Awaitable() {
+            this->future = std::future<py::object>();
+        };
+
+        Awaitable(std::future<py::object>& _future) {
```
Skylion007, line 35: Shouldn't this just be a const ref copy constructor which would negate the need for std::move?

include/pybind11/async.h
```diff
@@ -0,0 +1,205 @@
...
+        }
+    private:
+        py::object result;
+};
+
+
+class Awaitable {
+    public:
+        Awaitable() {
+            this->future = std::future<py::object>();
```
Skylion007, line 32: You can just have this be default member initialized and user a member initializer list in the the other constructor.

include/pybind11/async.h
```diff
@@ -0,0 +1,205 @@
...
+namespace py = pybind11;
+
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(async)
+
+
+class StopIteration : public py::stop_iteration {
+    public:
+        StopIteration(py::object result) : stop_iteration("--"), result(result) {};
```
Skylion007, line 19:
```suggestion
        StopIteration(py::object result) : stop_iteration("--"), result(std::move(result)) {};
```

Skylion007 (changes requested):
None of this is actually being checked by Clang-Tidy or other static analyzers  and won't be unless you add a simple unit test.

include/pybind11/async.h
rwgk: Why is it out of scope? Too much work? Or something fundamental, technically?

hiaselhans: I'd say out of scope for my skills and I really need some time to check pybinds internals which i don't seem to have for the near future.

I think there are three things left to address:

- use the approach from @0lru
- get rid of std::bind (i tried to use lambdas but it seems to break things//segfaults//needs more work)
- make a cleaner interface (like an attribute)

0lru: I'm not using any of these (two) approaches defined here. Maybe it's also possible to do something
with c++-coroutines(?)... However, I'm currently using an abstract Promise class. 
This has the advantage, that I don't need to start a new thread with std::async or std::thread inside the pybind11-definition:
```cpp
template<typename T>
    class Promise
    {
    public:
        class Implementation
        {
        public:
            virtual ~Implementation() = default;
            virtual void set_value(T) = 0;
            virtual void set_exception(std::exception_ptr) = 0;
        };

        Promise(std::shared_ptr<Implementation> implementation)
            : _implementation(std::move(implementation))
        {
        }

        void set_value(T t)
        {
            _implementation->set_value(std::move(t));
        }

        void set_exception(std::exception_ptr e)
        {
            _implementation->set_exception(std::move(e));
        }

    private:
        std::shared_ptr<Implementation> _implementation;
    };

... later ... :
    
    void Foo::get_bar(Promise<int>&&) const;

```

Another approach was to define a "TaskQueue" (a threaded event-loop), where "Driven" means, that
the object is assigned to such a "TaskQueue":
```cpp
 template<typename Driven, typename Param>
    auto async(py::module& asyncio, void(Driven::* member)(Param))
    {
        return [=](Driven& object, Param param) mutable
        {
            auto promise = std::make_shared<Promise<void>>(asyncio);
            auto future = promise->get_future();
            object.task_queue()->run([object = &object, member, promise{ std::move(promise) }, param{std::move(param)}]() mutable {
                try
                {
                    (*object.*member)(std::move(param));
                    promise->set_value();
                }
                catch (...)
                {
                    promise->set_exception(std::current_exception());
                }
            });
            return future;
        };
    }
```
What I want to note/say: This topic is quite complex and I want to question if std::future is
a good match for doing interaction with asyncio :-).

rwgk: > I'd say out of scope for my skills and I really need some time to check pybinds internals which i don't seem to have for the near future.

Caveat: I'm not familiar enough with threading, futures, async to have a meaningful opinion about the high-level design or implementation details.

From a high level perspective, this still looks like a lot of work to me (realistically weeks worth of time) before it could be merged. I'd be very reluctant to consider adding `class_async` as an interface, we need full comprehensive, well-written test coverage, documentation, and agreement between people with relevant experience. This is a fairly major new feature, until it's out there for a while (a few months to a couple years), we'd need some commitment to supporting it with fixes or even bigger adjustments in case early design decisions don't pan out well for general use. Do you think you are prepared for that?

hiaselhans: As i said, it is hard for me to find consecutive weeks of time now - during summer season, but i am motivated to push this topic forward after that.
of course the PR in it's current state is quite far from enough and as you pointed out, adding class_async doesn't seem right. I'd like to reduce the amount of code to a minimum and get rid of all copy & paste parts (class_ > class_async).

Until then, should i close this draft-pr or do you shall we leave it as a good starting-point for anyone interested / a place for discussions/ideas?

rwgk: > Until then, should i close this draft-pr or do you shall we leave it as a good starting-point for anyone interested / a place for discussions/ideas?

Up to you, keeping this open as a draft sounds good to me.

Random / general / something to think about for the maintainers:
* a `contrib/` directory might be useful for things like this, things that take a while to mature and may need the involvement of multiple people.
* having some kind of hook in the `class_` code would seem fine to me if that's helpful (as an alternative to the `call_guard` approach, or in addition to, idk), so that you could maintain your work as an add-on without being forced to introduce `class_async`.

To the second point, that's the approach I'm [currently pursuing on the smart_holder branch](https://github.com/pybind/pybind11/blob/ee945e705891cc0a5abaf8dd8174caf7cc90fb21/include/pybind11/detail/smart_holder_sfinae_hooks_only.h). (To avoid misunderstanding, the hooks are currently NOT on master, but that's where the idea comes from.)

include/pybind11/async.h
Skylion007: Clang-Tidy can automatically rewrite this function for you.
```
auto bound_thread_func = [&thread_func](auto&& args){
     return thread_func(std::forward<Args>(args)...);
};

**Comment by green4free** [comment]
Hello. I don't know if this is the correct place to ask, but what is the status of this?
What is currently the "correct" way of spawning a C++ thread, which completion is awaitable in python?

**Comment by eirrgang** [comment]
> Hello. I don't know if this is the correct place to ask, but what is the status of this?

I think the issue would be more appropriate than the pull request.

> What is currently the "correct" way of spawning a C++ thread, which completion is awaitable in python?

The simplest thing is probably to do launch a regular function (exported with pybind `def()`) from a Python ThreadPoolExecutor using either [`concurrent.futures.ThreadPoolExecutor.submit()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) or [`asyncio.AbstractEventLoop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)

**Comment by hiaselhans** [comment]
I ended up using this snippet in my own codebase:

```
# pragma once

#include <pybind11/stl.h>
#include <pybind11/embed.h>
#include <pybind11/operators.h>

namespace py=pybind11;
using namespace py::literals;


class StopIteration : public py::stop_iteration {
    public:
        StopIteration(py::object result) : stop_iteration("--"), result(std::move(result)) {};

        void set_error() const override {
            PyErr_SetObject(PyExc_StopIteration, this->result.ptr());
        }
    private:
        py::object result;
};


class Awaitable {
    public:
        Awaitable() : future() {};

        Awaitable(std::future<py::object>& _future) : future(std::move(_future)) {};

        Awaitable* iter() {
            return this;
        };

        Awaitable* await() {
            return this;
        };

        void next() {
            // check if the future is resolved (with zero timeout)
            auto status = this->future.wait_for(std::chrono::milliseconds(0));

            if (status == std::future_status::ready) {
                // job done -> throw
                auto exception = StopIteration(this->future.get());

                throw exception;
            }
        };

    private:
        std::future<py::object> future;
};


py::class_<Awaitable> enable_async(py::module m) {
    return py::class_<Awaitable>(m, "Awaitable")
        .def(py::init<>())
        .def("__iter__", &Awaitable::iter)
        .def("__await__", &Awaitable::await)
        .def("__next__", &Awaitable::next);
};
```

then used it like that:


```
PYBIND11_MODULE(my_module, m) {
    m.doc() = "async python module";
    py::class_<AsyncClass>(m, "AsyncClass")
        .def("fn_async", [](AsyncClass& cls) {
            auto bound_thread_func = [&cls](){
                auto result = cls.do_something();
                py::gil_scoped_acquire gil;
                return py::cast(result);
            };

            auto future = std::async(std::launch::async, bound_thread_func);
            
            return Awaitable(future);
        })
```

**Comment by eirrgang** [comment]
Seems reasonable.

>         void next() {
>             // check if the future is resolved (with zero timeout)
>             auto status = this->future.wait_for(std::chrono::milliseconds(0));
> 
>             if (status == std::future_status::ready) {
>                 // job done -> throw
>                 auto exception = StopIteration(this->future.get());
> 
>                 throw exception;
>             }
>         };

This has the potential to make the event loop thread busier than it needs to be. If it is a moderately busy event loop, I wouldn't think there would be much of a problem. But for lighter workloads, it would be nice to actually yield while waiting for results, and for large numbers of tasks, it would be nice not to have to keep calling `next()` on each task.

Note that when `asyncio.Task` wraps a coroutine, the `__await__` iterator only yields once, then uses '_asyncio_future_blocking' to give a lot of flexibility to the event loop and the coroutine implementation itself to potentially sleep or (even to block) until the coroutine is done or until the chained future done_callback gets triggered.

**Comment by rroohhh** [comment]
Note, the wrapper for `StopIteration` presented here does not work for tuples (you only get the first element of the tuple on the python side). A quick workaround is always wrapping the result object in another tuple:
```
class StopIteration : public py::stop_iteration {
    public:
        StopIteration(py::object result) : stop_iteration("--"), result(py::make_tuple(std::move(result))) {};

        void set_error() const override {
            PyErr_SetObject(PyExc_StopIteration, this->result.ptr());
        }
    private:
        py::object result;
};
```

PR #2668, opened 2020-11-16
Compose literals for argument values in docstring

### LINKED ISSUE #2667: [FEAT] Syntactically valid default argument docstring
Currently, enum and class default arguments in functions generate the following docstrings:

```c++
enum class Color {Red};
struct ColorContainer {
  Color color;
};

PYBIND11_MODULE(foo, m) {
  pybind11::enum_<Color> (m, "Color").value("Red", Color::Red);
  m.def("a", [](Color) {}, pybind11::arg("a") = Color::Red);
  pybind11::class_<ColorContainer> color_container(m, "ColorContainer");
  m.def("b", [](ColorContainer) {}, pybind11::arg("a") = ColorContainer {Color::Red});
}
```

```python
>>> help(foo)
[...]
FUNCTIONS
    a(...) method of builtins.PyCapsule instance
        a(a: foo.Color = <Color.Red: 0>) -> None
    b(...) method of builtins.PyCapsule instance
        b(a: foo.ColorContainer = <foo.ColorContainer object at 0x7f7234cb1f70>) -> None
```

This is somewhat unhelpful for the automatic generation of stubfiles. Representations with these brackets are not syntactically valid Python, making parsing of these expressions subject to whatever representation pybind is currently sticking with. Importantly, bound enum values could be fully represented. Additionally, I would argue bound classes should not be represented with strings containing pointer values in docstrings since that representation is random and non-reproducible. Bound class values are more or less impossible to represent as literals (unless some special machinery is added to `arg` and `arg_v` to handle some exceptions, like inline construction through literals, e.g. as `py::arg("a") = py::literal<BoundClassType>(ctor_args...)`, although I think that is likely to be controversial.

## Suggested resolution

I would suggest altering the docstring generation to fully represent bound enum values, replace bound class instances with an ellipsis `...`, and leaving the call to `__repr__` for any non-bound types, something like this (please forgive any potential no-nos, I'm admittedly not really familiar with the codebase):

```diff
diff --git a/include/pybind11/pybind11.h b/include/pybind11/pybind11.h
index e2ddda0..ded5020 100644
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -235,8 +235,24 @@ protected:
                 a.name = strdup(a.name);
             if (a.descr)
                 a.descr = strdup(a.descr);
-            else if (a.value)
-                a.descr = strdup(repr(a.value).cast<std::string>().c_str());
+            else if (a.value) {
+                auto typehandle = type::handle_of(a.value);
+                if (detail::get_internals().registered_types_py.count(Py_TYPE(a.value.ptr())) > 0) {
+                  if (hasattr(typehandle, "__members__")) {
+                    // Bound enum type, can be fully represented
+                    auto descr = typehandle.attr("__module__").cast<std::string>();
+                    descr += "." + typehandle.attr("__qualname__").cast<std::string>();
+                    descr += "." + a.value.attr("name").cast<std::string>();
+                    a.descr = strdup(descr.c_str());
+                  } else {
+                    // Use ellipsis instead of repr to keep syntactic validity
+                    a.descr = strdup("...");
+                  }
+                } else {
+                  // Types not bound by pybind are well-represented by repr
+                  a.descr = strdup(repr(a.value).cast<std::string>().c_str());
+                }
+            }
         }
 
         rec->is_constructor = !strcmp(rec->name, "__init__") || !strcmp(rec->name, "__setstate__");
```

This causes the following:

```c++
enum class Color {Red};
struct ColorContainer {
  Color color;
};

PYBIND11_MODULE(foo, m) {
  pybind11::enum_<Color> (m, "Color").value("Red", Color::Red);
  m.def("a", [](Color) {}, pybind11::arg("a") = Color::Red);
  m.def("b", [](int) {}, pybind11::arg("a") = 1);
  m.def("c", [](std::string) {}, pybind11::arg("a") = "Hello");
  m.def("d", [](std::vector<int>) {}, pybind11::arg("a") = std::vector<int> {{1, 2, 3, 4}});

  pybind11::class_<ColorContainer> color_container(m, "ColorContainer");
  m.def("e", [](ColorContainer) {}, pybind11::arg("a") = ColorContainer {Color::Red});
}
```

```python
>>> help(foo)
[...]
FUNCTIONS
    a(...) method of builtins.PyCapsule instance
        a(a: foo.Color = foo.Color.Red) -> None

    b(...) method of builtins.PyCapsule instance
        b(a: int = 1) -> None

    c(...) method of builtins.PyCapsule instance
        c(a: str = 'Hello') -> None

    d(...) method of builtins.PyCapsule instance
        d(a: List[int] = [1, 2, 3, 4]) -> None

    e(...) method of builtins.PyCapsule instance
        e(a: foo.ColorContainer = ...) -> None
```

jargonzombies:
## Description
Closes #2667 by addition of a `compose_literal` function that attempts to handle the generation of valid Python literals for default arguments in docstrings. Can handle built-in data structure nesting by recursion. Generates literals for bound enums, replaces non-enum bound class instances with an ellipsis.

## Suggested changelog entry:

```rst
Attempt to generate syntactically valid literals for default function arguments in docstrings
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -119,6 +119,73 @@ protected:
         return new detail::function_record();
     }
 
+    template<typename ContainerHandle, typename StringifierUnary>
+    static std::string join(ContainerHandle container, StringifierUnary&& f, std::string sep = ", ") {
+        std::string joined;
+        for (auto element : container) {
+            joined += f(element) + sep;
+        }
+        if (!joined.empty()) {
+            joined.erase(joined.size() - sep.size());
+        }
+        return joined;
+    }
+
+    // Generate a literal expression for default function argument values
+    static std::string compose_literal(pybind11::handle h) {
+        auto typehandle = type::handle_of(h);
+        if (detail::get_internals().registered_types_py.count(Py_TYPE(h.ptr())) > 0) {
+            if (hasattr(typehandle, "__members__") && hasattr(h, "name")) {
+                // Bound enum type, can be fully represented
+                auto descr = typehandle.attr("__module__").cast<std::string>();
+                descr += "." + typehandle.attr("__qualname__").cast<std::string>();
+                descr += "." + h.attr("name").cast<std::string>();
+                return descr;
+            }
+
+            // Use ellipsis expression instead of repr to ensure syntactic validity
+            return "...";
+        }
+
+        if (isinstance<dict>(h)) {
+            std::string literal = "{";
+            literal += join(
+                reinterpret_borrow<dict>(h),
+                [](const std::pair<handle, handle>& v) { return compose_literal(v.first) + ": " + compose_literal(v.second); }
+            );
+            literal += "}";
+            return literal;
+        }
+
+        if (isinstance<list>(h)) {
+            std::string literal = "[";
+            literal += join(reinterpret_borrow<list>(h), &compose_literal);
+            literal += "]";
+            return literal;
+        }
+
+        if (isinstance<tuple>(h)) {
+            std::string literal = "(";
+            literal += join(reinterpret_borrow<tuple>(h), &compose_literal);
+            literal += ")";
+            return literal;
+        }
+
+        if (isinstance<set>(h)) {
+            auto v = reinterpret_borrow<set>(h);
+            if (v.empty()) {
+                return "set()";
+            }
+            std::string literal = "{";
+            literal += join(v, &compose_literal);
+            literal += "}";
+            return literal;
+        }
+
+        // All other types should be terminal and well-represented by repr
+        return repr(h).cast<std::string>();
+    }
+
     /// Special internal constructor for functors, lambda functions, etc.
     template <typename Func, typename Return, typename... Args, typename... Extra>
     void initialize(Func &&f, Return (*)(Args...), const Extra&... extra) {
@@ -235,8 +302,10 @@ protected:
                 a.name = strdup(a.name);
             if (a.descr)
                 a.descr = strdup(a.descr);
-            else if (a.value)
-                a.descr = strdup(repr(a.value).cast<std::string>().c_str());
+            else if (a.value) {
+                std::string literal = compose_literal(a.value);
+                a.descr = strdup(literal.c_str());
+            }
         }
 
         rec->is_constructor = !strcmp(rec->name, "__init__") || !strcmp(rec->name, "__setstate__");
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -97,6 +97,7 @@ set(PYBIND11_TEST_FILES
     test_copy_move.cpp
     test_custom_type_casters.cpp
     test_docstring_options.cpp
+    test_docstring_function_signature.cpp
     test_eigen.cpp
     test_enum.cpp
     test_eval.cpp
--- a/tests/test_docstring_function_signature.cpp
+++ b/tests/test_docstring_function_signature.cpp
@@ -1,0 +1,23 @@
+/*
+    tests/test_docstring_options.cpp -- generation of docstrings function signatures
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#include "pybind11_tests.h"
+#include "pybind11/stl.h"
+
+enum class Color {Red};
+
+TEST_SUBMODULE(docstring_function_signature, m) {
+    // test_docstring_function_signatures
+    pybind11::enum_<Color> (m, "Color").value("Red", Color::Red);
+    m.def("a", [](Color) {}, pybind11::arg("a") = Color::Red);
+    m.def("b", [](int) {}, pybind11::arg("a") = 1);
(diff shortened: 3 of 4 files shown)


jargonzombies:
Seeking review @henryiii

YannickJadoul:
It's interesting to see this, but I'm afraid it has practically no chance of getting merged. A few reasons:
- @wjakob has repeatedly expressed that he's not willing to increase the binary size of all pybind11-generated extension modules, just to make the docstrings perfect.
- This solution seems reasonably specific and long, while there's no way for users to opt-in/out to this, or extend this.
- Isn't that what `__repr__` is for, anyway, to have a full, detailed specification of a value?
- `py::argv` [already allows doing this](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#default-arguments-revisited), so in principle, it's not critical for this to be added to pybind11's core functionality. I suppose you could even make your own function/subclass/custom literal/... that creates the `argv` instance/docstring values you want?

Do you have a specification/example/... of what e.g. mypy or Sphinx expect?
If so, there were some ideas to figure out a way to have users opt-in or opt-out  of changes that increase the binary size. Maybe we can add this to the list of docstring things we'd like to do, once this is possible?

jargonzombies:
Thank you for the detailed explanation @YannickJadoul!

> not willing to increase the binary size of all pybind11-generated extension modules, just to make the docstrings perfect

Perfectly understandable, especially considering I offered no configuration to the modifications I made.

> Isn't that what __repr__ is for, anyway, to have a full, detailed specification of a value?

Quoting the [docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__):

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

There's an escape hatch in there. The method does not _require_ you to generate a valid Python literal that would evaluate back to the same value. Choosing `__repr__` as the closest approximation to getting valid literals into docstrings is admittedly a good and economic choice in that for the built-in Python types it does generate valid literals, but for everything else, placing this string representation into the function signature is - strictly speaking - incorrect. Conversely, implying that every user-defined class _should_ implement literal-generating `__repr__` methods is, as the documentation suggests, not always possible: Besides classes that are too complex to express in terms of a constructor and built-in literals, for instance, as far as I can tell, I cannot change the `__repr__` method of bound enums. I would argue replacement by ellipsis in docstrings is the better alternative to angle-bracket `__repr__`s (with the exception of bound enums, where all the necessary information is always available).

> py::argv already allows doing this

I can see why you're suggesting this customization point as an alternative to my blanket solution suppressing user-defined `__repr__` functions. I must admit I wasn't aware of it and I agree that it's a viable path, but I suppose I was hoping to minimize my maintenance effort. I think the point remains, and I think you agree, that in principle the docstrings could be better.

> Do you have a specification/example/... of what e.g. mypy or Sphinx expect?

I admittedly don't. I tried `mypy`'s `stubgen` on a pybind module of mine, saw that the generated results are clearly syntactically wrong or not satisfactory - `f(*args and **kwargs) -> Any` everywhere - and immediately moved on to `pybind11-stubgen`, a standalone project trying to process the generated docstrings into valid typechecking stubs. There too, the angle-bracket representations of enum values trip up the parser and generate incorrect stubs. I suppose I was trying to fix the encountered problem at the source.

> If so, there were some ideas to figure out a way to have users opt-in or opt-out of changes that increase the binary size. Maybe we can add this to the list of docstring things we'd like to do, once this is possible?

I would appreciate that. Do you see any way forward for this PR?

jargonzombies:
It occurred to me that the only reason this PR is as large as it is, is because I'm somewhat insistent on handling bound enums. If I gave that up, the minimum solution to getting syntactically valid docstrings might be as simple as string-replacing anything in the result of `__repr__` in angle brackets with an ellipsis. Might that be viable?

sizmailov:
@jargonzombies I had similar intention in #2243, but then I came to same conclusion: `repr` is not the right function for default value representation. I think a viable PR should add customization of this point:

https://github.com/pybind/pybind11/blob/17c22b9e0f56276f212b44f6cf0e6846a01623b2/include/pybind11/pybind11.h#L239

sizmailov:
Sorry, it did a short glance at your PR without thinking too much. I didn't realize that at the moment of docstring construction we deal with python objects which not necessarily have  bare C++ counterpart, which disables one from introducing template-based customization point.

`py::arg`/`py::arg_v` are another candidates to do the trick. In fact I just realized that to some degree this customization point already exists since we can overload `py::arg::operator=(T&& value)`:

```c++ 
namespace pybind11 {
    template<>
    py::arg_v arg::operator=(int &&value) const {
        return {std::move(*this), std::forward<int>(value), "my_custom_int_repr"};
    }
}

    m.def("add", [](int a, int b) { return a + b; }, py::arg("a"), py::arg("b") = 1);
   // renders `add(a: int, b: int = my_custom_int_repr) -> int`
}
```

This seems to be enough to customize behavior for bound C++ types and python builtin types (using `py::handle` as a template parameter and implementation from this PR).

jargonzombies:
I'm going to close this since the customization point suggested by @sizmailov is suitable for my use case.

PR #2669, opened 2020-11-16
fix typo in pickle example

bjodah:
## Description
Pretty sure it meant to say `self`?


YannickJadoul (approved):
Right! Thanks for catching that, @bjodah!

bstaletic (approved):

PR #2671, opened 2020-11-17
List all pybind11 exceptions and refer to code.

jblespiau:
I wanted to know excacly what existed, because I wanted to raise a
`RuntimeError`. In practice, it took me quite some time to find where
(a) the pybind11 exceptions are defined, (b) how the C++ exceptions
are converted to Python ones.

At the end, I discovered everything else is converted to RuntimeError,
so I do not need to do anything.

The link is a result to github search, which may be more robust to
moving code around.

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -9,11 +9,15 @@ that will trap C++ exceptions, translate them to the corresponding Python except
 and raise them so that Python code can handle them.
 
 pybind11 defines translations for ``std::exception`` and its standard
-subclasses, and several special exception classes that translate to specific
-Python exceptions. Note that these are not actually Python exceptions, so they
+subclasses (see [translate_exception]_) and several special exception classes
+that translate to specific Python exceptions (see [PYBIND11_RUNTIME_EXCEPTION]_).
+Note that these are not actually Python exceptions, so they
 cannot be examined using the Python C API. Instead, they are pure C++ objects
 that pybind11 will translate the corresponding Python exception when they arrive
 at its exception handler.
+
+.. [translate_exception] https://github.com/pybind/pybind11/search?q=translate_exception
+.. [PYBIND11_RUNTIME_EXCEPTION] https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h
 
 .. tabularcolumns:: |p{0.5\textwidth}|p{0.45\textwidth}|
 
@@ -43,14 +47,22 @@ at its exception handler.
 |                                      | of bounds access in ``__getitem__``, |
 |                                      | ``__setitem__``, etc.)               |
 +--------------------------------------+--------------------------------------+
-| :class:`pybind11::value_error`       | ``ValueError`` (used to indicate     |
-|                                      | wrong value passed in                |
-|                                      | ``container.remove(...)``)           |
-+--------------------------------------+--------------------------------------+
 | :class:`pybind11::key_error`         | ``KeyError`` (used to indicate out   |
 |                                      | of bounds access in ``__getitem__``, |
 |                                      | ``__setitem__`` in dict-like         |
 |                                      | objects, etc.)                       |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::value_error`       | ``ValueError`` (used to indicate     |
+|                                      | wrong value passed in                |
+|                                      | ``container.remove(...)``)           |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::type_error`        | ``TypeError``                        |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::buffer_error`      | ``BufferError``                      |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::import_error`      | ``import_error`                      |
++--------------------------------------+--------------------------------------+
+| Any other exception                  | ``RuntimeError``                     |
 +--------------------------------------+--------------------------------------+
 
 Exception translation is not bidirectional. That is, *catching* the C++
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -9,15 +9,11 @@ that will trap C++ exceptions, translate them to the corresponding Python except
 and raise them so that Python code can handle them.
 
 pybind11 defines translations for ``std::exception`` and its standard
-subclasses (see [translate_exception]_) and several special exception classes
-that translate to specific Python exceptions (see [PYBIND11_RUNTIME_EXCEPTION]_).
-Note that these are not actually Python exceptions, so they
+subclasses, and several special exception classes that translate to specific
+Python exceptions. Note that these are not actually Python exceptions, so they
 cannot be examined using the Python C API. Instead, they are pure C++ objects
 that pybind11 will translate the corresponding Python exception when they arrive
 at its exception handler.
-
-.. [translate_exception] https://github.com/pybind/pybind11/search?q=translate_exception
-.. [PYBIND11_RUNTIME_EXCEPTION] https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h
 
 .. tabularcolumns:: |p{0.5\textwidth}|p{0.45\textwidth}|
 
@@ -60,7 +56,7 @@ at its exception handler.
 +--------------------------------------+--------------------------------------+
 | :class:`pybind11::buffer_error`      | ``BufferError``                      |
 +--------------------------------------+--------------------------------------+
-| :class:`pybind11::import_error`      | ``import_error`                      |
+| :class:`pybind11::import_error`      | ``import_error``                     |
 +--------------------------------------+--------------------------------------+
 | Any other exception                  | ``RuntimeError``                     |
 +--------------------------------------+--------------------------------------+


docs/advanced/exceptions.rst
```diff
@@ -43,15 +47,23 @@ at its exception handler.
...
+|                                      | wrong value passed in                |
+|                                      | ``container.remove(...)``)           |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::type_error`        | ``TypeError``                        |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::buffer_error`      | ``BufferError``                      |
++--------------------------------------+--------------------------------------+
+| :class:`pybind11::import_error`      | ``import_error``                     |
++--------------------------------------+--------------------------------------+
+| Any other exception                  | ``RuntimeError``                     |
```
bstaletic, line 65: I don't think this last line is correct. Especially considering that the table starts with `std::exception`. Really, it's "any other exception inheriting from `std::exception`". I don't think pybind11 would handle `throw 5;` or `throw boost::exception{whateveer};`.

docs/advanced/exceptions.rst
```diff
@@ -9,12 +9,16 @@ that will trap C++ exceptions, translate them to the corresponding Python except
...
-Python exceptions. Note that these are not actually Python exceptions, so they
+subclasses (see [translate_exception]_) and several special exception classes
+that translate to specific Python exceptions (see [PYBIND11_RUNTIME_EXCEPTION]_).
+Note that these are not actually Python exceptions, so they
 cannot be examined using the Python C API. Instead, they are pure C++ objects
 that pybind11 will translate the corresponding Python exception when they arrive
 at its exception handler.
 
+.. [translate_exception] https://github.com/pybind/pybind11/search?q=translate_exception
+.. [PYBIND11_RUNTIME_EXCEPTION] https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h
```
YannickJadoul, line 20: I don't think we have other places in our docs where we link to the code. So I'm very wary of adding this. No regular user should immediately go see the code; power-users will be familiar with the code anyway. I would suggest linking to the API reference.

Even more so about the search: everyone can do a search on GitHub, so I see no reason for that link (but do make it a Sphinx reference to the API, if you want).

docs/advanced/exceptions.rst
```diff
@@ -43,15 +47,23 @@ at its exception handler.
...
-|                                      | ``container.remove(...)``)           |
-+--------------------------------------+--------------------------------------+
 | :class:`pybind11::key_error`         | ``KeyError`` (used to indicate out   |
 |                                      | of bounds access in ``__getitem__``, |
 |                                      | ``__setitem__`` in dict-like         |
 |                                      | objects, etc.)                       |
 +--------------------------------------+--------------------------------------+
+| :class:`pybind11::value_error`       | ``ValueError`` (used to indicate     |
+|                                      | wrong value passed in                |
+|                                      | ``container.remove(...)``)           |
```
YannickJadoul, line 57: Why was this moved? This adds to the diff (and future `blame`) without actually being a difference.

docs/advanced/exceptions.rst
YannickJadoul: It does though, I think, @bstaletic:

https://github.com/pybind/pybind11/blob/17c22b9e0f56276f212b44f6cf0e6846a01623b2/include/pybind11/detail/internals.h#L239-L242

YannickJadoul (commented):
> At the end, I discovered everything else is converted to RuntimeError,
> so I do not need to do anything.

You probably want to register a nice exception translator anyway, to not have "unknown error" shouted at the user?

YannickJadoul:
> In practice, it took me quite some time to find where
> (a) the pybind11 exceptions are defined, (b) how the C++ exceptions
> are converted to Python ones.

Also, I think neither of these two are user concerns (but only for developers), so I think they should not be in the documentation for the users.

docs/advanced/exceptions.rst
bstaletic: Fair enough. Forget what I said.

jblespiau: (This comment is for me an argument in favour of being able to find the code for an advanced reader, as the ground truth is the code).

docs/advanced/exceptions.rst
jblespiau: > > In practice, it took me quite some time to find where
> > (a) the pybind11 exceptions are defined, (b) how the C++ exceptions
> > are converted to Python ones.
> 
> Also, I think neither of these two are user concerns (but only for developers), so I think they should not be in the documentation for the users.

I know this may seem controversial, but I really believe it's useful here for the following reasons:
- In practice, I have been working a lot with pybind11, so I am way more than a "random" user, but more of an advanced user. And still,even thought I look at the code often, I am not familiar with the code.
- In practice, the documentation was stale, and I did not got the information I wanted (which was "if I raise an std::runtime_error" what happens?), so I had to try it out and look at the code, and it took me easily 20 minutes. Documentation become obsolete, it's the way it is, and it's way more user-friendly to give some hint where to look, instead of having documentation not in sync in the code, and no way to check. This is particularly true for like enums, or this which was a sequence of `if/else if`.

I was not able to do a search easily, because I did not know what to search for (i.e. `translate_exception` and `PYBIND11_RUNTIME_EXCEPTION`). If there had been one sentence like the one I am adding, it would have taken me very little time.

I am very flexible to change this a way you prefer, but I really would like to have "some way" to hint the user, so next time it's stale, or some user wants to understand the behavior/check the doc is up to date, they can.
I can remove the link, and just say, below the paragraph or below the table "For definitions in the code, see `translate_exception` and `PYBIND11_RUNTIME_EXCEPTION`", and let the user do a Github search (but I am wondering why we want to give *more* work for a curious reader than less.

I can also say "See PYBIND11_RUNTIME_EXCEPTION in common.h" without any link if you prefer (but I don't know why not linking it^^).

Concerning your last suggestion "do make it a Sphinx reference to the API, if you want", maybe it solves the issue, but I do not understand exactly what you mean. For me, these are internal, no need to document them on the API reference, so adding them + linking to them would be even more work, and harder to maintain.

Having read this comment, what change would you be fine to have? (I am very open to any way to make this less visible if you want, like a footnote, moving it, saying "For advanced users, see [...]", but I think removing it will lead to more people spend time like I did.

> > At the end, I discovered everything else is converted to RuntimeError,
> > so I do not need to do anything.
> 
> You probably want to register a nice exception translator anyway, to not have "unknown error" shouted at the user?

For my use-case, I will be fine with this (it's not something people should catch). I was just explaining the use-case I had that lead me to update the doc.

docs/advanced/exceptions.rst
jblespiau: I did this because it's the same order as in the code now, so it ease the work for maintainers, as they can easily check everything is correct, as it's in the same order etc.

The work is really simple here: open the code, copy/paste the names of the exceptions into this table^^

docs/advanced/exceptions.rst
YannickJadoul: To me, that's an argument for adding correct documentation. You don't want to dive into the code every time, right?
This is implementation. And of course the code is the ground truth, but you don't just say "let's give up on docs and always read the code", right?

YannickJadoul: Anyway, that's what I think. If you can convince other reviewers, why not, but I'm definitely not convinced of those links. Very happy to see the table completed, though!

docs/advanced/exceptions.rst
jblespiau: (Also note that we discuss over 2 pairs of 2 words in parenthesis. If it's specifically the links, I can remove them, but it's just easier for the reader to click on it. These 4 words can save time to advanced readers, I do not think they will reduce the experience of a beginner reader. Maybe putting these into a footnote would help reduce their visibility?)

docs/advanced/exceptions.rst
YannickJadoul: OK, that's a good reason indeed. That just wasn't immediately clear :-)

docs/advanced/exceptions.rst
jblespiau: (I am finding such footnote useful, but I applied the suggestion of removing any reference to the code, as It would not be merged with them).

YannickJadoul: Sorry! Didn't mean to not reply, and definitely not to shut down the conversation over this!

I can definitely see where you're coming from, by the way. I have a similar experience, that at some point, I needed to look into the code to find details! But then that's why docs should be improved, I suppose.

But as you said, that only happens at a more advanced stage, so I would expect these advanced users to start feeling more accustomed to the pybind11 code, anyway.

> In practice, the documentation was stale, and I did not got the information I wanted (which was "if I raise an std::runtime_error" what happens?), so I had to try it out and look at the code, and it took me easily 20 minutes. Documentation become obsolete, it's the way it is, and it's way more user-friendly to give some hint where to look, instead of having documentation not in sync in the code, and no way to check.

Yep, that's an issue, definitely. But the only way to fix the documentation is by looking at the code? Any link to the code that's precise enough would probably soon become obsolete as well? And otherwise, the kind of links you had now, will probably stay accurate enough for some time, but are also not very precise, so a "find all"-search in any IDE would provide as much information?
(By the way, these are the things for which we're very happy to point you in the right direction on Gitter. But I can see that sometimes you don't have time to wait for that.)

> I was not able to do a search easily, because I did not know what to search for (i.e. `translate_exception` and `PYBIND11_RUNTIME_EXCEPTION`). If there had been one sentence like the one I am adding, it would have taken me very little time.

We also really don't want to document these things, I believe. We already had users complain internal convenience macros like `NAMESPACE_BEGIN` were renamed. Especially in this header-only type of project, it's slightly dangerous to encourage users to dive into the internals :-/
If `PYBIND11_RUNTIME_EXCEPTION` is meant to be used by users, it should be documented in the API reference and/or elsewhere. I'm not sure, actually, but since it's not documented, it should be considered an implementation detail.


So in conclusion, those are the main things that worry me: mixing user and (non-existing) developer documentation (and this way documenting internals that are currently allowed to change between releases), and adding more opportunities for inconsistencies/outdated documentation (if hidden in links in different sections).

But I do completely agree that the developer documentation is minimal (or non-existent), and that a "quick tour of what each file +- contains" would be great (as long as it's clearly marked on a separate page, under "Development" or so, if you ask me).

YannickJadoul: But, let's see what others think, as far as I'm concerned :-)

YannickJadoul (approved):
Up to other reviewers (@henryiii, for example) whether they want to bring back some of the extra changes from 1cd996550317649f49d79cddcb8b0db83b6f6e57.

docs/advanced/exceptions.rst
henryiii: I agree, at least since we don't already link to code, and keeping links to code up-to-date would be practically impossible. For things that are "private", that is, only developers should care, that should be documented in the source but probably not outside of that.

Issue #2583, opened 2020-10-13
[QUESTION] How to pass std::unique_ptr ownership to class member?


rwgk:
Given a C++ library that passes `std::unique_ptr` ownership to a class member, how could it be wrapped with pybind11?

Minimal example, distilled from production code: [unique_ptr_member.h](https://github.com/google/clif/blob/main/examples/unique_ptr_member/unique_ptr_member.h)

PyCLIF supports this pattern: [bindings](https://github.com/google/clif/blob/main/examples/unique_ptr_member/python/unique_ptr_member.clif), [tests](https://github.com/google/clif/blob/main/examples/unique_ptr_member/python/unique_ptr_member_test.py).

We are working on integrating pybind11 into PyCLIF and need to find some way to keep supporting the existing wrappers, like the minimal example, or provide an alternative/workaround. What are our options?

Some background:

Quoting from the [relevant pybind11 documentation](https://pybind11.readthedocs.io/en/latest/advanced/smart_ptrs.html#std-unique-ptr):

> `void do_something_with_example(std::unique_ptr<Example> ex) { ... }`
>  
> The above signature would imply that Python needs to give up ownership of an object that is passed to this function, which is generally not possible (for instance, the object might be referenced elsewhere). 

The PyCLIF approach to [giving up ownership is implemented here](https://github.com/google/clif/blob/e3f76102177f5c21813304369f04408602be50a3/clif/python/instance.h#L118).

Quoting from the source code comment there:

    // Will give up ownership of the pointee if the pointee is a) owned by this
    // Instance, and b) if this Instance is the only pointer to the pointee.

Questions about the "Missing value for wrapped C++ type" error message are very rare. While having its dangers, on balance this feature appears to be very useful.

EricCousineau-TRI:
Here's our fork's version of ownership transfer, similar to what you mentioned for PyCLIF:
https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/include/pybind11/pytypes.h#L1517-L1614

It's a tad more complicated since it's also trying to address slicing (#1145).

gentlegiantJGC:
Is this resolved now that the smart holder branch has been merged?

rwgk:
Yes:

> Minimal example, distilled from production code: [unique_ptr_member.h](https://github.com/google/clif/blob/main/examples/unique_ptr_member/unique_ptr_member.h)

That's now:

[...]

Issue #2646, opened 2020-11-09
[FEAT] Rework casting


rhaschke:
Recently @rwgk revived the discussion on supporting movable custom types (and holders), see #2583, #2040, #1132.
I want to summarize the discussion so far and structure the problem into several subproblems, in the hope that we can focus our future discussion on the corresponding subproblems.

1. Fix segfaults due to mismatching holder types, i.e. fixing #1138 and #1215.
pybind11 assumes that holder types are used consistently, i.e. once registered, a custom type needs to stick with its holder type, let it be the default `unique_ptr` or any custom smart ptr. However, this basic assumption is not validated (yet) and thus can cause severe segfaults if users don't obey this rule.
I think #2644 addresses this issue in a clean and efficient manner (building on the great work #1161 from @jagerman) by validating holder compatibility at function/class definition time (runtime).

2. Support moving (rvalue-reference arguments) for custom types, allowing to (easily) wrap functions like this:
    ```c++
    void consume(CustomType&& object) { CustomType sink(std::move(object)); }
    ```
    This doesn't work out of the box (yet), but is in principle possible already right now, employing a small wrapper function as pointed out by @YannickJadoul in https://github.com/pybind/pybind11/pull/2047#issuecomment-705229376.
My PR #2047 is an initial attempt to enable this feature. However, I recently detected an open issue with this approach (https://github.com/pybind/pybind11/pull/2047#issuecomment-713168819), but I am working on it...

3. Support moving of holder types, i.e. ownership transfer from python to C++, allowing to wrap functions like this:
    ```c++
    void consume(unique_ptr<CustomType>&& object) { CustomType sink(std::move(*holder.release()); }
    ```
    There are two possible implementations for this: #1146 (rather huge) and #2046. However, to quote @wjakob:
    > Transferring ownership from Python to C++ is a super-dangerous and rather unnatural (non-pythonic) operation and intentionally not supported in pybind11.

    If implementing this, we need to ensure that all python object references of the moved holder are validated before `loading`. For an open issues on this, see https://github.com/pybind/pybind11/pull/2046#issuecomment-570081915.

[...]

PR #2672, opened 2020-11-17
Adding py::smart_holder (for smart-pointer interoperability).

rwgk:
Background:

* We are working on changing [PyCLIF](http://github.com/google/clif) to be based on pybind11.

* pybind11 as-is (Nov 2020) does NOT support passing `unique_ptr`, by choice:
https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-unique-ptr

* PyCLIF does: https://github.com/pybind/pybind11/issues/2583

* Removing the feature from PyCLIF would go against modern C++ best practices: https://github.com/pybind/pybind11/pull/2047#issuecomment-712385940  (It would also take weeks or even months to make the required changes in the Google codebase.)

* Note also that pybind11 has this long-standing bug related to `shared_ptr`/`unique_ptr` handling: https://github.com/pybind/pybind11/issues/1138

* The purpose of this draft PR is to track experiments and observations, figuring out how to add support for modern C++ best practices involving `shared_ptr`/`unique_ptr`, safe & bug free.


rwgk:
@eacousineau QQ: did you consider building a custom holder type (using `PYBIND11_DECLARE_HOLDER_TYPE`) as a way to enable passing `std::unique_ptr<T>`?

Wondering, could this work:
* custom holder is e.g. `disownable_shared<T>`
* plus caster to/from `std::unique_ptr<T>`
* plus caster to/from `std::shared_ptr<T>`

?

rwgk:
Quick update, I tested with this change by @rhaschke (rebased #2046):

https://github.com/pybind/pybind11/compare/master...rhaschke:rvalue-argument-fix2

With this change and a small tweak (expecting `TypeError` instead of `ValueError`) `test_unique_ptr_member.py` runs successfully!

rwgk:
To track progress working with the rebased #2046 by @rhaschke (https://github.com/pybind/pybind11/compare/master...rhaschke:rvalue-argument-fix2):

* `test_unique_ptr_member` also works with `std::shared_ptr<pointee>` as holder.

* After reverting Robert's change to `common.h` (diff below), all but two `pytest` scripts run successfully (using current upstream master):

    Before reverting (using `TypeError`): 20 failures in 9 `test_*.py` files
    After reverting (using `RuntimeError`): 9 failures in 2 `test_*.py` files

    Using `RuntimeError`, there are:
    * 7 failures in `test_factory_constructors.py`
    * 2 failures in `test_pickling.py`

* When running the Google-global testing with Robert's patch included (and the two `pytest` scripts with failures disabled) there were no issues related to the change! It looks like we just need to fix up the two `pytest` scripts.

    The Google-global testing covers at least ~270 pybind11 extensions, with millions of indirect dependents.

```
diff --git a/include/pybind11/detail/common.h b/include/pybind11/detail/common.h
index af871ca..56385f5 100644
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -719,8 +719,8 @@ PYBIND11_RUNTIME_EXCEPTION(value_error, PyExc_ValueError)
 PYBIND11_RUNTIME_EXCEPTION(type_error, PyExc_TypeError)
 PYBIND11_RUNTIME_EXCEPTION(buffer_error, PyExc_BufferError)
 PYBIND11_RUNTIME_EXCEPTION(import_error, PyExc_ImportError)
-PYBIND11_RUNTIME_EXCEPTION(cast_error, PyExc_TypeError) /// Thrown when pybind11::cast or handle::call fail due to a type casting error
-PYBIND11_RUNTIME_EXCEPTION(reference_cast_error, PyExc_TypeError) /// Used internally
+PYBIND11_RUNTIME_EXCEPTION(cast_error, PyExc_RuntimeError) /// Thrown when pybind11::cast or handle::call fail due to a type casting error
+PYBIND11_RUNTIME_EXCEPTION(reference_cast_error, PyExc_RuntimeError) /// Used internally
 
 [[noreturn]] PYBIND11_NOINLINE inline void pybind11_fail(const char *reason) { throw std::runtime_error(reason); }
 [[noreturn]] PYBIND11_NOINLINE inline void pybind11_fail(const std::string &reason) { throw std::runtime_error(reason); }
```

rwgk:
Current state:

upstream master
* does not build (but builds if the `ptr_owner` `init` (passing `unique_ptr`) is commented out).

https://github.com/pybind/pybind11/pull/2046
* closest to desired behavior, but returning `unique_ptr` or `shared_ptr` deletes the `pointee` prematurely (the returned object is invalidated).

https://github.com/RobotLocomotion/pybind11
* does not invalidate the disowned Python object (potentially dangling pointer).
* passing `unique_ptr` not possible with `shared_ptr` holder.

rwgk:
I update this PR with **highly** experimental `vptr_holder` code (currently needs C++17, for `std::variant`). The idea is to cleanly model C++ behavior: a `unique_ptr` can be promoted to a `shared_ptr` but not the other way around. I don't know if such a behavior can be achieved with `class_`. I fear the existing `shared_ptr` handling may be in the way, but I have to find out. I may try to go back to the idea of building an alternative, maybe named `vclass`, using the `vptr_holder`, with the desired caster behavior around `vclass`.

Before I worked on the `vptr_holder` I ran into trouble with `unique_ptr` and `shared_ptr` returns, using @rhaschke's  #2046 code (https://github.com/pybind/pybind11/compare/master...rhaschke:rvalue-argument-fix2). On return, the `pointee` smart pointer is reset prematurely. I developed doubts if the clean C++-like behavior can be shoe-horned into `class_`, because it would need to have a `shared_ptr` but also remember somehow if it's actually only a `unique_ptr`. That thought led me to the `std::variant` idea.

rhaschke:
To be honest, I think that you pursue the wrong approach to realize the automatic conversion of holder types. The `std::variant` based holder is limited to two holder types: unique and shared ptr, while pybind11, in principle, supports any holder type. However, I agree that these holder types already cover most standard use cases.

Your approach essentially works around a major design flaw of pybind11, namely that it doesn't check for compatibility of holder types when passing them between python and C++: a holder type is simply reinterpret-casted to another one. #2644 (which nobody looked at yet) is an attempt to fix this issue, rejecting attempts to use incompatible holder types.

However, you are striving for _implicit_ holder conversion, the 4th and most complicated goal in my summary #2646. I think, if you want to tackle implicit holder conversion, you should mimic the existing code for implicit type conversions, namely implementing a runtime lookup to find the correct conversion mechanism.

As outlined in #2646, I envision an _explicit_ mechanism to handle holder type conversions.

rwgk:
Hi @rhaschke, I looked through the changes, and to be totally honest, this PR is currently totally over my head. (I need to learn a lot more than I know at the moment to meaningfully review.)

But I tried to use this PR Google-internal. I didn't get past the initial manual testing stage, first running into this error:

```
ImportError: Cannot register function using not yet registered type 'proto2::Message'
```

I changed the corresponding code to move the `class_` definition for `proto2::Message` up. Trying again I got:

```
ImportError: Detected mismatching holder types when declaring function '__copy__': attempting to use holder type std::__u::unique_ptr<proto2::Message, std::__u::default_delete<proto2::Message> >, but proto2::Message was declared using holder type std::__u::shared_ptr<proto2::Message>
```

I know these bindings are used in ~100 extensions and pass thousands if not 100s of thousands of tests. Is it plausible that so many tests run successfully with mismatched holders? (Or could there be a problem with the mismatch detection?)

I could dig in deeper, but wanted to check with you first.

(The affected code is open source in theory, but the open source repo (pybind11_protobuf) is behind and doesn't have a working build system.)

include/pybind11/vptr_holder.h
```diff
@@ -0,0 +1,72 @@
...
+
+    std::shared_ptr<T> get_shared() {
+        auto s = std::get_if<1>(&vptr_);
+        if (s) {
+            return *s;
+        }
+        auto u = std::get_if<0>(&vptr_);
+        if (u) {
+            auto result = std::shared_ptr<T>(std::move(*u));
+            vptr_ = result;
```
rwgk, line 56: @rwgk: what happens if the unique_ptr was disowned before?

rwgk: 1f4f5e2 adds `test_promotion_of_disowned_to_shared`, which works as desired.

include/pybind11/vptr_holder.h
```diff
@@ -0,0 +1,77 @@
...
+  public:
+    explicit vptr(T *ptr = nullptr) : vptr_{std::unique_ptr<T>(ptr)} {
+        std::cout << std::endl << "explicit vptr(T *ptr = nullptr)" << std::endl;
+        //TRIGGER_SEGSEV
+    }
+    explicit vptr(std::unique_ptr<T> u) : vptr_{std::move(u)} { std::cout << std::endl << "explicit vptr(std::unique_ptr<T> u)" << std::endl; }
+    explicit vptr(std::shared_ptr<T> s) : vptr_{s} { std::cout << std::endl << "explicit vptr(std::shared_ptr<T> s)" << std::endl; }
+
+    int ownership_type() const {
+        if (std::get_if<0>(&vptr_)) {
```
laramiel, line 23: return vptr_.index()?

rwgk:
**Pure analysis** of the current inner workings of pybind11.

@rhaschke, @EricCousineau-TRI, @YannickJadoul: please let me know corrections.

`test_holder_unique_ptr` & `test_holder_shared_ptr` were added to systematically exercise returning and passing `unique_ptr<T>`, `shared_ptr<T>` with `unique_ptr`, `shared_ptr` holder.

Observations:
```
test_holder_unique_ptr:
  make_unique_pointee  OK
  pass_unique_pointee  BUILD_FAIL (as documented)
  make_shared_pointee  Abort free(): double free detected
  pass_shared_pointee  RuntimeError: Unable to load a custom holder type from a default-holder instance

test_holder_shared_ptr:
  make_unique_pointee  Segmentation fault (#1138)
  pass_unique_pointee  BUILD_FAIL (as documented)
  make_shared_pointee  OK
  pass_shared_pointee  OK
```

`test_smart_ptr_base_derived` was added to systematically exercise casts between `shared_ptr<base>` (short `s<B>`) and `shared_ptr<derived>` (`s<D>`). Concrete and polymorphic (`virtual`) types are handled slightly differently, the polymorphic case is most general:

When converting a `s<B>` C++ RETURN value with a virtual `B`:
* `polymorphic_type_hook` uses `typeid(B)` which will yield the `std::type_info` RTTI for `D` if `B` was upcast previously in C++: https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/cast.h#L851
* the RTTI is used to lookup the Python type for `D`, essentially by `typeid().name()`: https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/cast.h#L896
* a new `D` instance is created, using an **explicit cast from `s<B>` to `s<D>`**: https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/pybind11.h#L1505

This means the `s<B>` C++ return value appears as a `D` instance in Python.

When passing a Python `D` instance to a C++ function with a `s<B>` ARGUMENT:
* pybind11 compares the RTTI for all Python bases of `D` to find a match for `B`: https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/cast.h#L683
*  a `s<B>` reference is created using a **`reinterpret_cast` from `s<D>` to `s<B>`**: https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/cast.h#L235

For the case of a `unique_ptr` return value with a `shared_ptr` holder, the explicit cast from `s<B>` to `s<D>` above becomes an explicit cast from `u<B>` to `s<D>`, typically leading to a segfault (#1138), although it is a classical case of UB (Undefined Behavior).

For the case of a `shared_ptr` return value with a `unique_ptr` holder, the explicit cast is from `s<B>` to `u<D>`, which is also UB and typically leads to a double free. (Probably, it doesn't crash earlier because the raw pointers inside `s<B>` and `u<D>` are at the same relative memory locations, as suggested earlier by @rhaschke (not sure where).)

rwgk:
e0207d6 adds a minimal demonstration of UB in the handling of shared_ptr holder. The test is based on https://godbolt.org/z/4fdjaW by jorgbrown@ (thanks Jorg!).  See below for the test result (segfault).

```
$ clang++ --version
clang version 9.0.1-14
Target: x86_64-pc-linux-gnu

$ /usr/bin/python3 $HOME/clone/pybind11_scons/run_tests.py ../pybind11 -vv test_smart_ptr_private_first_base
================================================================================= test session starts =================================================================================
platform linux -- Python 3.8.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /usr/local/google/home/rwgk/forked/pybind11/tests, inifile: pytest.ini
collected 1 item                                                                                                                                                                      

test_smart_ptr_private_first_base.py::test_make_pass FAILED                                                                                                                     [100%]

====================================================================================== FAILURES =======================================================================================
___________________________________________________________________________________ test_make_pass ____________________________________________________________________________________

    def test_make_pass():
        d = m.make_shared_drvd()
        i = m.pass_shared_base(d)
>       assert i == 200
E       assert 0 == 200
E         +0
E         -200

test_smart_ptr_private_first_base.py:9: AssertionError
================================================================================== 1 failed in 0.08s ==================================================================================
Fatal Python error: Segmentation fault

Current thread 0x00007f0a65ada740 (most recent call first):
<no Python frame>
```

rwgk:
I ported `test_smart_ptr_private_first_base` to Boost.Python: https://github.com/rwgk/rwgk_tbx/commit/debf3a5ec37cf9b1ccc93877a86010455b7125f8

It runs without a problem.

rwgk:
47b81af adds an additional test designed to demonstrate UB in the shared_ptr holder handling, with root cause in the explicit cast here (this link is also in my comment from 4 days ago):

https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/pybind11.h#L1505

The test segfaults in this line: `i = m.pass_shared_drvd(b)`

The Boost.Python port of this additional test passes without a problem: https://github.com/rwgk/rwgk_tbx/commit/7d7494ae33a687adbff40385af0bb5f5bd04ea0e

EricCousineau-TRI:
Haven't yet fully paged into all the work you've done here, but awesome!
I saw some work on virtual inheritance and your mention of segfaulting - can I ask if it's related to slicing of derived objects? (#1145 and other similar issues?)

FWIW These are the tests in our fork that I had written to address them:
https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/tests/test_ownership_transfer.py

EricCousineau-TRI:
> @eacousineau QQ: did you consider building a custom holder type (using `PYBIND11_DECLARE_HOLDER_TYPE`) as a way to enable passing `std::unique_ptr<T>`?

Nope, hadn't considered that! Maybe that'd be better; however, it may still require additional expectations on the holder API, at least for my use cases. For example, to address #1145 (the slicing from inheritance), I wanted to add some form of `.release()` functionality. (Can explain more if you'd like.)

---

Also, I read the above summary. That seems about accurate!
I had trouble disentangling this each time, so I ended up basically only supporting single-inheritance when it came down to holder conversion and/or ownership transfer:
https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/include/pybind11/cast.h#L492-L851

rwgk:
Tracking a line of thought (resulting in an UNPROVEN idea):

This was really useful for me to learn about the caster mechanisms:

https://gist.github.com/rwgk/4fe600fad45442a5d790b53195a4da51

It's an annotated/edited gdb stack trace, for the situation reported by @eacousineau under #1138.

You can see in frames `#8`, `#7`, `#6` the `unique_ptr` holder type information is available, but in frames `#5`, `#4` it's lost (all you have there is a `const void*`), then, after a runtime search for the Python type, in frames `#3`, `#2`, `#1` the holder type is `shared_ptr`.

In frame `#0` the `void*` is `reinterpret_casted` to `shared_ptr`. That's exactly the root cause of the UB reported in previous comments, even if the holder types match. Looking only at that simpler case: what's needed to achieve well-defined behavior is a `std::dynamic_pointer_cast`. How can that be achieved with the current high-level code organization? In frame `#0` we can at best only have RTTI information for the source type, but not compile-time type information.

This lead me to a new idea:
```
  std::shared_ptr<void> ivp = std::dynamic_pointer_cast<void>(shared_drvd_uc);
  std::shared_ptr<drvd> shared_drvd_dc = std::reinterpret_pointer_cast<drvd>(ivp);
```

Complete code: https://godbolt.org/z/sPTae3

It runs to completion. I'm asking our C++ experts if this is well-defined.

rwgk:
8dba00b adds an additional test designed to demonstrate UB in the handling of polymorphic pointers, with root cause in the explicit cast here:

https://github.com/pybind/pybind11/blob/30eb39ed79d1e2eeff15219ac00773034300a5e6/include/pybind11/cast.h#L229

The new test segfaults as shown below.
Note that this new demo does NOT involve smart pointers at all, unlike the otherwise similar `test_smart_ptr_private_first_base`.

The Boost.Python port of this additional test passes without a problem: https://github.com/rwgk/rwgk_tbx/commit/97919e39f319c82a6f88d38e4cd8c9d8b348bbab


```
================================================================================= test session starts =================================================================================
platform linux -- Python 3.8.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /usr/local/google/home/rwgk/forked/pybind11/tests, inifile: pytest.ini
collected 2 items / 1 deselected / 1 selected                                                                                                                                         

test_private_first_base.py::test_make_drvd_pass_base FAILED                                                                                                                     [100%]

====================================================================================== FAILURES =======================================================================================
______________________________________________________________________________ test_make_drvd_pass_base _______________________________________________________________________________

    def test_make_drvd_pass_base():
        d = m.make_drvd()
        i = m.pass_base(d)
>       assert i == 200
E       assert 0 == 200
E         +0
E         -200

test_private_first_base.py:10: AssertionError
=========================================================================== 1 failed, 1 deselected in 0.06s ===========================================================================
Fatal Python error: Segmentation fault
```

rwgk:
@EricCousineau-TRI wrote:
> I saw some work on virtual inheritance and your mention of segfaulting - can I ask if it's related to slicing of derived objects? (#1145 and other similar issues?)

I looked briefly at #1145 but I'm not clear, in particular what exactly is slicing? But maybe you can see faster if there is a connection, based on the even more reduced test under https://github.com/pybind/pybind11/pull/2672#issuecomment-750883611? The real root cause of what I'm observing is that a pointer obtained through a `dynamic_cast<void*>` can only safely be cast back directly to the most derived type. Does that sound like it's related to what you call "slicing"?

Here is a godbolt demo for the safe vs unsafe casting back: https://godbolt.org/z/8873Wd

rwgk:
This comment is related to #1333, #2757, #2646.
Keyword: inheritance slicing

9eea90702d97a75182f2c8d905152a7c9b77fdb3 adds a test similar to the reproducer under #1333.
An equivalent Boost.Python test is here:
https://github.com/rwgk/rwgk_tbx/blob/6c9a6d6bc72d5c1b8609724433259c5b47178680/cpp_base_py_derived_ext.cpp
https://github.com/rwgk/rwgk_tbx/blob/6c9a6d6bc72d5c1b8609724433259c5b47178680/tst_cpp_base_py_derived.py

Observations:
* Boost.Python fails the simpler `test_base`, apparently because it's missing a to-python converter for `std::shared_ptr<base>` when the holder is `std::shared_ptr<base_callback>` (`base_callback` is the Boost.Python equivalent of pybind's trampoline feature).
* Boost.Python passes the more involved `test_drvd`, where `drvd` is a Python subclass of the wrapped C++ `base`.
* Conversely, pybind11 passes `test_base` but fails `test_drvd`, see below. Interestingly, the unexpected `203107` return value means that there are two confounded issues: 1. the cloned `shared_ptr<base>` is not a pointer for the trampoline class, 2. it is a default-constructed `base` (`_base_num` `100`) when it should be constructed from the `clone` function (`_base_num` `150`).
* The new pybind11 test runs ASAN and MSAN clean, and it is reliably reproducible (in other words Not flakey).

Speculative: I believe Boost.Python passes `test_drvd` because the `shared_ptr<base>` in `clone_get_num` has a custom deleter that keeps the wrapper PyObject alive: https://github.com/boostorg/python/blob/5e4f6278e01f018c32cebc4ffc3b75a743fb1042/include/boost/python/converter/shared_ptr_from_python.hpp#L52

```
platform linux -- Python 3.8.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /usr/local/google/home/rwgk/forked/pybind11/tests, inifile: pytest.ini
collected 2 items                                                                                                                                                                     

test_cpp_base_py_derived.py::test_base PASSED                                                                                                                                   [ 50%]
test_cpp_base_py_derived.py::test_drvd FAILED                                                                                                                                   [100%]

====================================================================================== FAILURES =======================================================================================
______________________________________________________________________________________ test_drvd ______________________________________________________________________________________

    def test_drvd():
      d = drvd()
      assert d.get_num() == 200
      assert m.get_num(d) == 200
      dc = d.clone()
      assert dc.get_num() == 250
>     assert m.clone_get_num(d) == 203257
E     assert 203107 == 203257
E       +203107
E       -203257

test_cpp_base_py_derived.py:35: AssertionError
============================================================================= 1 failed, 1 passed in 0.05s =============================================================================
```

rwgk:
For some reason the github interface won't let me add @rhaschke as reviewer. Tagging here.

rwgk:
This PR is ready for a first round of feedback. It implements a `py::smart_holder` that can be used in combination with `py::class_`. I hope the user-facing interface is in it's final form and all smart-pointer interactions I could think of are implemented and fully tested, ASAN, MSAN, UBSAN clean. The internals could use some more work. — And apparently workarounds for various platforms. I just see MSVC is very unhappy.

rwgk:
Hi @EricCousineau-TRI and @YannickJadoul, the CI is green now. :-)

Suggested starting points for review are:

1. `include/pybind11/detail/smart_holder_poc.h` & `tests/pure_cpp/smart_holder_poc_test.cpp`
2. `tests/test_type_caster_bare_interface.*`
3. `include/pybind11/detail/smart_holder_type_casters.h`

Notes:
Re 1.: the cmake file for `tests/pure_cpp` is missing, but I did ASAN, MSAN, UBSAN test the code internally.
Re 2.: that was the starting point for 3. It works with pure `master`. In addition to showing the API virtually clutter-free, I think it's a valuable test in itself, systematically exercising the `type_caster` API. 
Re. 3: it glues 1. & 2. together, based on bits and pieces taken mainly from `cast.h` and `pybind11.h`. You can still see the duct tape. Intentionally I didn't refactor more than necessary at this stage, so that it is easier to match back to the bits and pieces I copied in (not a whole lot!).

All the rest are just tests and misc trivial adjustments.
Everything together, old and new, is also ASAN, MSAN, UBSAN tested. It's also been through our global testing.
All new code is `clang-format`ed using our new `.clang-config`.

rwgk:
Small update: I added the cmake file for `tests/pure_cpp`, CI is still green (I spot-checked that `smart_holder_poc_test` actually runs).

rwgk:
For the record, I'm currently continuing the development here:

https://github.com/rwgk/pybind11/tree/pr2672_sideline_cast_shtc_interleaved

That sideline was started from this branch, but with the `smart_holder_type_caster.h` code copied into the middle of `cast.h`. That's the only reasonably straightforward way I could think of, to enable using `smart_holder` as default holder, with a define to switch between default holders, so that all existing unit test can be run either with `unique_ptr` or `smart_holder` as the default holder.

It's too time consuming back-porting changes from the sideline to here while things are still in flux. I'm holding out hope that someone approves PR #2841, which would allow me to move development back here.

I'm holding off creating a PR for the sideline branch until I need CI runs.

rwgk:
For the record, tracking a learning experience:

Re: Lifetime management of `value_and_holder` when this line runs:

https://github.com/pybind/pybind11/blob/c78dfb69f23639148870bf7d97587b5a7e7053ab/tests/test_factory_constructors.cpp#L191

The memory for the `TestFactory3` value constructed there is allocated here before the main [`dispatcher`](https://github.com/pybind/pybind11/blob/c78dfb69f23639148870bf7d97587b5a7e7053ab/include/pybind11/pybind11.h#L542) calls the code above:

https://github.com/pybind/pybind11/blob/c78dfb69f23639148870bf7d97587b5a7e7053ab/include/pybind11/cast.h#L608

After the call the control flow passes through this line:

https://github.com/pybind/pybind11/blob/c78dfb69f23639148870bf7d97587b5a7e7053ab/include/pybind11/pybind11.h#L933

`init_instance` calls `init_holder` to wind up here, constructing a `std::shared_ptr`:

https://github.com/pybind/pybind11/blob/c78dfb69f23639148870bf7d97587b5a7e7053ab/include/pybind11/pybind11.h#L1539

The deallocation is then handled by the `shared_ptr` using the default deleter.

Known issue (but easy to work around using new-style `init<>`): this mechanism doesn't work for classes with custom operator new: https://github.com/pybind/pybind11/issues/948

rwgk:
I merged the `pr2672_sideline_cast_shtc_interleaved` into this PR to run the CI.

Current state: all unit tests run ASAN clean with unique_ptr or smart_holder as the default.

**Caveat**: commit https://github.com/pybind/pybind11/pull/2672/commits/249df7cbdb09817fed0ddf80f01ba5af12466820 needs to be reverted after adopting the existing move-only-if-refcount-is-1
logic used by type_caster_base. Currently the `smart_holder_type_caster` moves all movable values unconditionally.

PR #2679, opened 2020-11-19
fix: define PYBIND11_CPP14 for recent intel compilers

### LINKED ISSUE #1649: overload_cast failure with recent Intel compilers
## Issue description

As the title says, I can't make `overload_cast` to work with recent Intel compilers:
1. icpc version 18.0.3 (gcc version 7.3.0 compatibility)
2. icpc version 18.0.1 (gcc version 6.4.0 compatibility) 
3. icpc version 17.0.4 (gcc version 6.4.0 compatibility)
4. icpc version 17.0.1 (gcc version 6.3.0 compatibility)

## Reproducible example code

As a MWE, I've modified the `cmake_example` to use `overload_cast` and the latest release of pybind11:
```

int add(int i, int j) {
    return i + j;
}

double add(double a, double b) {
    return a + b;
}

namespace py = pybind11;

PYBIND11_MODULE(cmake_example, m) {
    m.doc() = R"pbdoc(
        Pybind11 example plugin
        -----------------------

        .. currentmodule:: cmake_example

        .. autosummary::
           :toctree: _generate

           add
           subtract
    )pbdoc";

    m.def("add", py::overload_cast<int, int>(&add), R"pbdoc(
        Add two numbers

        Some other explanation about the add function.
    )pbdoc");

    m.def("add", py::overload_cast<double, double>(&add), R"pbdoc(
        Add two numbers

        Some other explanation about the add function.
    )pbdoc");

    m.def("subtract", [](int i, int j) { return i - j; }, R"pbdoc(
        Subtract two numbers

        Some other explanation about the subtract function.
    )pbdoc");

#ifdef VERSION_INFO
    m.attr("__version__") = VERSION_INFO;
#else
    m.attr("__version__") = "dev";
#endif
}
```
I configure with ` cmake -H. -Bbuild -DCMAKE_CXX_COMPILER=icpc -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_STANDARD=14` and I invariably get:
```
Scanning dependencies of target cmake_example
[ 50%] Building CXX object CMakeFiles/cmake_example.dir/src/main.cpp.o
/cluster/home/roberto/Workspace/robertodr/cmake_example/pybind11/include/pybind11/detail/common.h(755): error: static assertion failed with "pybind11::overload_cast<...> requires compiling in C++14 mode"
      static_assert(detail::deferred_t<std::false_type, Args...>::value,
      ^
          detected during instantiation of class "pybind11::overload_cast<Args...> [with Args=<int, int>]" at line 27 of "/cluster/home/roberto/Workspace/robertodr/cmake_example/src/main.cpp"
```

This seems to be due to the fact that `PYBIND11_CPP14` is never defined for the Intel compilers in [`detail/common.h`](https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h) I added the following lines to the `CMakeLists.txt` to remedy for that:
```
target_compile_definitions(cmake_example
  PRIVATE
    $<$<CXX_COMPILER_ID:Intel>:PYBIND11_CPP14>
  )
```
Now I get the following error for toolchains 1 and 2 above:
```
Scanning dependencies of target cmake_example
[ 50%] Building CXX object CMakeFiles/cmake_example.dir/src/main.cpp.o
/cluster/home/roberto/Workspace/robertodr/cmake_example/pybind11/include/pybind11/detail/descr.h(74): error: expression must have a constant value
      static constexpr auto digits = descr<sizeof...(Digits), 0>({ ('0' + Digits)..., '\0' }, { nullptr });
```
this error for toolchain 4:
```
In file included from /cluster/home/roberto/Workspace/robertodr/cmake_example/pybind11/include/pybind11/attr.h(13),
                 from /cluster/home/roberto/Workspace/robertodr/cmake_example/pybind11/include/pybind11/pybind11.h(43),
                 from /cluster/home/roberto/Workspace/robertodr/cmake_example/src/main.cpp(1):
/cluster/home/roberto/Workspace/robertodr/cmake_example/pybind11/include/pybind11/cast.h(838): error: alias template "pybind11::detail::type_caster_base<type>::cast_op_type" may not be used in the type-id of the alias-declaration
      template <typename T> using cast_op_type = cast_op_type<T>;
```
**but** no error for toolchain 3.

To summarize:
- The `PYBIND11_CPP14` preprocessor variable should be defined for the Intel compilers.
- The 2018 releases of the Intel compilers I have access to do not work with C++14.

For the latter, I can't suggest a solution, but I'd be glad to test anything you might suggest.

tobiasleibner:
I just encountered #1649 with the Intel compiler 2021 (beta). Recent Intel compilers seem to support C++14 just fine, so just defining ``PYBIND11_CPP14`` for recent Intel compilers should fix the problem.

EDIT (@YannickJadoul): Closes #1649

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -27,7 +27,7 @@
 #  endif
 #endif
 
-#if !(defined(_MSC_VER) && __cplusplus == 199711L) && !defined(__INTEL_COMPILER)
+#if !(defined(_MSC_VER) && __cplusplus == 199711L)
 #  if __cplusplus >= 201402L
 #    define PYBIND11_CPP14
 #    if __cplusplus >= 201703L
@@ -49,6 +49,8 @@
 #if defined(__INTEL_COMPILER)
 #  if __INTEL_COMPILER < 1800
 #    error pybind11 requires Intel C++ compiler v18 or newer
+#  elif __INTEL_COMPILER < 1900 && defined(PYBIND11_CPP14)
+#    error pybind11 supports only C++11 with Intel C++ compiler v18. Use v19 or newer for C++14.
 #  endif
 #elif defined(__clang__) && !defined(__apple_build_version__)
 #  if __clang_major__ < 3 || (__clang_major__ == 3 && __clang_minor__ < 3)


include/pybind11/detail/common.h
```diff
@@ -27,7 +27,11 @@
 #  endif
 #endif
 
-#if !(defined(_MSC_VER) && __cplusplus == 199711L) && !defined(__INTEL_COMPILER)
+#if defined(__INTEL_COMPILER) && __INTEL_COMPILER >= 1900
```
henryiii, line 30: Is it really only added on 19? We only support 18+, so it's much simpler if it was added in 18. :'( I don't see anything specific in the release notes for 19 that indicate that they just started defining `__cplusplus`.

tobiasleibner: I guess ``__cplusplus`` is already defined for 18, but the first comment in #1649 mentions that the C++14 support is not sufficient in that version. I will check the value of ``__cplusplus`` in version 18.

tobiasleibner: Apparently, ``__cplusplus`` is already defined to ``201402L`` in version 18. So it should work in theory, but apparently it does not work in practice, see #1649. To support version 18, we thus would have to add specializations to work around the intel compiler bugs, I guess.

henryiii: This would be better than forcing C++ mode to 11 if 14 is requested. It would even be better, IMO, to throw an error if 14 is requested instead of using 11 when 14 is set, with a statement like "only C++11 is supported with Intel 18". I could make a quick attempt to get this working on Intel 18 before doing something like that.

It would be so much nicer if we could test this in CI, but only the most recent version is available and it doesn't work (#2573 )

YannickJadoul: I'm not a fan of adding workarounds to support a compiler lying about supporting C++14, though :-/

henryiii: At the minimum, let's kill the Intel specific limiting here. Since we already have a list of `#error` printouts for minimum supported compilers, we could add on that forces Intel 18 to C++11 mode.

tobiasleibner: I removed the first check and added an error message.

henryiii (approved):

ax3l (approved):

Issue #2682, opened 2020-11-20
[BUG] test_methods_and_attributes.py crash in cpython 3.9 debug build


albanD:
## Issue description
The test test_methods_and_attributes crashes with the following stack trace:
```
../../tests/test_methods_and_attributes.py ..............python: Objects/typeobject.c:3228: _PyType_Lookup: Assertion `!PyErr_Occurred()' failed.
--Type <RET> for more, q to quit, c to continue without paging--

Thread 1 "python" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c6b859 in __GI_abort () at abort.c:79
#2  0x00007ffff7c6b729 in __assert_fail_base (fmt=0x7ffff7e01588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x55555582e2a3 "!PyErr_Occurred()", file=0x55555583c5c3 "Objects/typeobject.c", line=3228, function=<optimized out>)
    at assert.c:92
#3  0x00007ffff7c7cf36 in __GI___assert_fail (assertion=assertion@entry=0x55555582e2a3 "!PyErr_Occurred()", 
    file=file@entry=0x55555583c5c3 "Objects/typeobject.c", line=line@entry=3228, 
    function=function@entry=0x55555583fd58 <__PRETTY_FUNCTION__.16944> "_PyType_Lookup") at assert.c:101
#4  0x000055555561e30a in _PyType_Lookup (type=type@entry=0x5555559565c0 <PyModule_Type>, name=name@entry='should_fail')
    at Objects/typeobject.c:3222
#5  0x00005555556005ea in _PyObject_GenericGetAttrWithDict (obj=obj@entry=<module at remote 0x7ffff5967dd0>, 
    name=name@entry='should_fail', dict=dict@entry=0x0, suppress=suppress@entry=0) at Objects/object.c:1194
#6  0x0000555555600e2d in PyObject_GenericGetAttr (obj=obj@entry=<module at remote 0x7ffff5967dd0>, name=name@entry='should_fail')
    at Objects/object.c:1280
[...]

Issue #2681, opened 2020-11-19
[BUG] gil_scoped_acquire fails with cpython 3.9 debug build


albanD:
## Issue description

Trying to use gil_scoped_acquire with a cpython 3.9 debug build fails with the following stack trace:
```
[New Thread 0x7ffff6910700 (LWP 28320)]
python: Python/ceval.c:135: is_tstate_valid: Assertion `!_PyMem_IsPtrFreed(tstate->interp)' failed.

Thread 2 "python" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff6910700 (LWP 28320)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c6b859 in __GI_abort () at abort.c:79
#2  0x00007ffff7c6b729 in __assert_fail_base (fmt=0x7ffff7e01588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x555555864f00 "!_PyMem_IsPtrFreed(tstate->interp)", file=0x5555558648e3 "Python/ceval.c", line=135, 
    function=<optimized out>) at assert.c:92
#3  0x00007ffff7c7cf36 in __GI___assert_fail (assertion=assertion@entry=0x555555864f00 "!_PyMem_IsPtrFreed(tstate->interp)", 
    file=file@entry=0x5555558648e3 "Python/ceval.c", line=line@entry=135, 
    function=function@entry=0x555555866760 <__PRETTY_FUNCTION__.16391> "is_tstate_valid") at assert.c:101
#4  0x0000555555679fdc in is_tstate_valid (tstate=tstate@entry=0x7ffff0000b70) at ./Include/internal/pycore_pymem.h:49
#5  0x000055555567eaa1 in take_gil (tstate=tstate@entry=0x7ffff0000b70) at Python/ceval_gil.h:227
#6  0x000055555567f053 in PyEval_AcquireThread (tstate=0x7ffff0000b70) at Python/ceval.c:385
#7  0x00007ffff6b31087 in pybind11::gil_scoped_acquire::gil_scoped_acquire (this=0x7ffff690fe90)
[...]

PR #2683, opened 2020-11-20
Remove workaround code in gil_scoped_acquire that is not needed since #1211

albanD:
## Description

Remove code that is not needed since #1211 that causes CPython 3.9 debug build to fail with #2681

I tested this with the following python debug builds:
- python 3.6:
  - master: no crash
  - with this fix: no crash
- python 3.8: 
  - master: fails with https://github.com/pybind/pybind11/issues/2682
  - with this fix: fails with https://github.com/pybind/pybind11/issues/2682
  - with this fix and proposed fix for #2682: no crash
- python 3.9:
  - master: fails in gil_scoped_acquire as described in #2681
  - with this fix: fails with https://github.com/pybind/pybind11/issues/2682
  - with this fix and proposed fix for #2682: no crash

Let me know if you want me to do other tests.

## Suggested changelog entry:

Not sure if this should be added or not. Let me know if it needs updating.

```rst
Fix py::gil_scoped_acquire assert with CPython 3.9 debug build
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2089,15 +2089,7 @@ public:
         }
 
         if (release) {
-            /* Work around an annoying assertion in PyThreadState_Swap */
-            #if defined(Py_DEBUG)
-                PyInterpreterState *interp = tstate->interp;
-                tstate->interp = nullptr;
-            #endif
             PyEval_AcquireThread(tstate);
-            #if defined(Py_DEBUG)
-                tstate->interp = interp;
-            #endif
         }
 
         inc_ref();


YannickJadoul (approved):
> Let me know if you want me to do other tests.

Is it easy for you to quickly test out 2.7? It's not the main focus, but we still support it officially.

> Suggested changelog entry:
> 
> Not sure if this should be added or not. Let me know if it needs updating.

Not sure either, but I guess it can't hurt? It's not 3.9-specific, though, or is it?
And maybe `issue` should be more specific and say "assert"?

albanD:
> It's not 3.9-specific, though, or is it?

From the test I did, this issue does not happen on 3.6 and 3.8. So I guess it is only for 3.9.

> Is it easy for you to quickly test out 2.7?

It takes a bit longer because old cpython debug build make your rebuild every single package you need from source.
But tested with latest 2.7 tag on cpython. No crash before or after this change.

YannickJadoul:
> From the test I did, this issue does not happen on 3.6 and 3.8. So I guess it is only for 3.9.

> It takes a bit longer because old cpython debug build make your rebuild every single package you need from source.
> But tested with latest 2.7 tag on cpython. No crash before or after this change.

OK, AFAIC, this is more than good enough, then :-) Let's build towards the future of Python; not the past. And maybe we can add some debug Python builds in CI.

henryiii (approved):
We really should add some debug tests to the CI, but that shouldn't hold up this PR. LGTM!

Issue #2336, opened 2020-07-28
Casting from an unregistered type does not throw.


nyibbang:
Hello,

## Issue description

I have this issue on pybind11 v2.5.0.

In the documentation, in [object.rst](../blob/master/docs/advanced/pycpp/object.rst), section `Casting back and forth`, the following is stated:

**`When conversion fails, both directions throw the exception cast_error.`**

But when trying to cast an unregistered custom type, it does not. Instead it seems to be setting an error (`PyErr_Occurred` returns a `TypeError`) and it returns a `None` object.

## Reproducible example code

```cxx
#include <pybind11/pybind11.h>
#include <pybind11/embed.h>
#include <cassert>

struct F {};

int main() {
  pybind11::scoped_interpreter interp;
  F f;
  try {
    auto obj = pybind11::cast(f);
  }
  catch(...) {
    return 0;
  }
  assert(((void)"cast did not throw an exception", false));
  return -1;
}
```
According to the documentation, we should not get an assertion failure with this code.

YannickJadoul:
I can reproduce and confirm the bug. Minor correction: I seems like `obj` is a `nullptr` and not `None`?

YannickJadoul:
Aaaaaaah, that's what this old patch I had lying around fixes! I forgot what it actually does, but it seems this fixes things?

```diff
diff --git a/pybind11/include/pybind11/cast.h b/pybind11/include/pybind11/cast.h
index fab985873..d94005328 100644
--- a/pybind11/include/pybind11/cast.h
+++ b/pybind11/include/pybind11/cast.h
@@ -719,7 +719,8 @@ public:
         detail::clean_type_id(tname);
         std::string msg = "Unregistered type : " + tname;
         PyErr_SetString(PyExc_TypeError, msg.c_str());
-        return {nullptr, nullptr};
+        throw error_already_set();
+        // return {nullptr, nullptr};
     }
 
     const type_info *typeinfo = nullptr;
```

I need to check whether this doesn't break anything else that is expecting a `{nullptr, nullptr}` to be returned, though.

nyibbang:
Thank you, so you're confirming that this should be the expected behavior ?

One workaround for me is to do the following:

```cxx
    auto obj = pybind11::cast(f);
    if (PyErr_Occurred())
       throw pybind11::error_already_set();
```

If too tedious, it's a fairly simple thing to put in a function.

YannickJadoul:
[...]

PR #2684, opened 2020-11-20
Fix erronous python error setting on failed cast

albanD:
## Description

Proposed fix for #2682 if maintainers agree that this is the right thing to do.

Fix crash in test suite for both debug cpython 3.8 and 3.9

## Suggested changelog entry:

Not sure how detailed the description below should be. Let me know if this needs updating.

```rst
Prevent setting python error twice on failed argument cast
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -739,18 +739,12 @@ public:
 
 
     // Called to do type lookup and wrap the pointer and type in a pair when a dynamic_cast
-    // isn't needed or can't be used.  If the type is unknown, sets the error and returns a pair
-    // with .second = nullptr.  (p.first = nullptr is not an error: it becomes None).
+    // isn't needed or can't be used.  If the type is unknown, returns a pair of nullptr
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Cfr. https://github.com/pybind/pybind11/issues/2682#issuecomment-731046664, this PR is more complex than the proposed workaround in #2422 (https://github.com/pybind/pybind11/issues/2422#issuecomment-724136788), but doesn't seem to handle all the subtleties regarding #2336. In order to fix #2682, I propose to stick to @bstaletic's minimal workaround, and fix #2336 after a more thorough investigation of the issue and its consequences.

albanD:
I do agree that it should be throwing. I opened this PR mostly so that you could see the concrete code I was talking about.
Closing as not the right fix.

YannickJadoul:
Perfect. Thanks for the effort and investigation! And again apologies that I didn't make the link yesterday evening (evening/night for me, at least) :-)

albanD:
No worries!

PR #2691, opened 2020-11-24
Update installing.rst

jgfoster:
`git submodule add` needs the branch before the repository or else it is ignored. The previous code checked out the `master` branch, not the `stable` branch.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -16,7 +16,7 @@ as a submodule. From your git repository, use:
 
 .. code-block:: bash
 
-    git submodule add ../../pybind/pybind11 extern/pybind11 -b stable
+    git submodule add -b stable ../../pybind/pybind11 extern/pybind11
     git submodule update --init
 
 This assumes you are placing your dependencies in ``extern/``, and that you are


YannickJadoul:
Do you mind bumping `jwlawson/actions-setup-cmake@v1.3` to `1.4` in `.github/workflows/configure` (cfr. [this line](https://github.com/pybind/pybind11/pull/2656/files#diff-23f9290814044c606e570ce24178da9d597a61f321c2c9510c64852a8622f1d1R58) in #2656, which is not part of `stable` yet). Alternatively, target `master`, I suppose.

jgfoster:
@YannickJadoul, I'm new to `pybind`, I found a problem with the docs, did an on-line edit, and submitted a PR. I don't really know anything about the GitHub workflows for this project and it seems like you are requesting a change that has nothing to do with fixing the documentation. Is there a reason you think I should make this change and do so as part of this PR? I think it would make more sense for someone who understands your request to open an issue or submit a new PR. Is that reasonable? Is there something I'm missing? If your suggestion is to help get tests to pass, that is fine, but my change didn't break them and my knowledge is sufficiently thin that I don't think I should be making other changes.

YannickJadoul:
> Is there a reason you think I should make this change and do so as part of this PR?

Well, yeah, the idea was to make the CI pass, indeed. There's absolutely no way your change caused this, but by targeting `stable`, you have apparently shown the CI on `stable` is outdated. Preferably, we don't merge a PR before it turns green (if only because our stable will suddenly be red).

I can make that change quickly as well, but let's ask @henryiii if this PR should be targeted to `stable` or `master`.

henryiii:
It's really up to @wjakob, but I think we should always target master, and then update stable for new releases. For docs, it is potentially a bit more complicated, since stable is what shows up by default in readthedocs, so a docs-only PR could in theory target stable. But if that's the case, it should still target master, and should be backported to stable with some sort of note or label.

henryiii (approved):
Windows jobs randomly have an error due to a known issue - that's fine. LGTM.

henryiii:
This won't show up in the "stable" branch on RtD until we have a release that includes it, but it will change on the "latest" docs immediately. Thanks!

Issue #2690, opened 2020-11-23
[BUG] bytes constructor takes size as size_t, not ssize_t


anntzer:
## Issue description

The `py::bytes(char const*, size_t)` constructor takes its input size as unsigned size_t, even though the underlying CPython API (https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromStringAndSize) uses signed ssize_t.  See also https://github.com/pybind/pybind11/issues/1599 for a similar case.

## Reproducible example code

Use the following snippet into the setup provided in the pybind11/python_example repo:
```cpp
#include <pybind11/pybind11.h>

namespace py = pybind11;

PYBIND11_MODULE(python_example, m) {
    ssize_t n = 3;
    m.attr("foo") = py::bytes{"foo", n};
}
```
Compilation (pybind11 2.6.0, gcc 10.2.0) yields
```
src/main.cpp: In function ‘void pybind11_init_python_example(pybind11::module_&)’:
src/main.cpp:7:38: warning: narrowing conversion of ‘n’ from ‘ssize_t’ {aka ‘long int’} to ‘pybind11::size_t’ {aka ‘long unsigned int’} [-Wnarrowing]
    7 |     m.attr("foo") = py::bytes{"foo", n};
```

YannickJadoul:
But you're passing the wrong type to pybind11? I don't see why 1) a length of a string should be possibly negative, and 2) why you would ant to pass a `ssize_t` here? As long as there is no warning internally (there isn't, `pybind11` casts from `size_t` to `ssize_t`), wouldn't this cause only confusion?

anntzer:
pybind11 is just forwarding the arguments to PyBytes_FromStringAndSize, which takes a ssize_t (indeed, calling `PyBytes_FromStringAndSize("foo", n);` in the same example emits no warning).  There are many places where you naturally get a ssize_t that you may want to forward to the `bytes` constructor (e.g. from the shape of a `py::array_t`).
See also https://www.python.org/dev/peps/pep-0353/#why-not-size-t, perhaps.

YannickJadoul:
Yes, but pybind11 wraps the C API. Saying that `PyBytes_FromStringAndSize` is completely irrelevant and does not constitute an argument on why the pybind11 API should accept `ssize_t`?

>  There are many places where you naturally get a ssize_t that you may want to forward to the `bytes` constructor (e.g. from the shape of a `py::array_t`).

[...]

PR #2692, opened 2020-11-24
Add sequence constructors taking ssize_t parameter.

anntzer:
This patch makes `ssize_t s = 3; py::bytes{"foo", s};` (and likewise for
str, tuple, and list) compile without a -Wnarrowing (ssize_t to size_t)
warning.  Allowing signed sizes is consistent with the CPython C-API
(PyBytes_FromStringAndSize) which also takes a Py_ssize_t parameter, and
also allows e.g. directly passing in a size that comes from the shape of
an array_t (which is likewise signed).

In order to achieve this, I had to add a templated overload for *all*
integral types and then specialize specifically for `ssize_t` (the
base case) and `size_t` (to remove the -Wnarrowing warning for it,
for convenience/backcompat).  Adding only overloads for `ssize_t`
and `size_t` would not work as `py::bytes{"foo", 3}` would result in
ambiguous overload resolution.

## Description

See more discussion at https://github.com/pybind/pybind11/issues/2690.

## Suggested changelog entry:

```rst
Constructors of sequences (tuple, list, str, bytes) no longer emit narrowing warnings when given a signed size as parameter.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -936,10 +936,16 @@ class str : public object {
 public:
     PYBIND11_OBJECT_CVT(str, object, detail::PyUnicode_Check_Permissive, raw_str)
 
-    str(const char *c, size_t n)
-        : object(PyUnicode_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+    str(const char *c, ssize_t n)
+        : object(PyUnicode_FromStringAndSize(c, n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate string object!");
     }
+
+    template <typename T, typename = detail::enable_if_t<std::is_integral<T>::value>>
+    str(const char *c, T n) : str(c, ssize_t{n}) { }
+
+    // For convenience, do not emit -Wnarrowing when passing in an unsigned size_t:
+    str(const char *c, size_t n) : str(c, ssize_t(n)) { }
 
     // 'explicit' is explicitly omitted from the following constructors to allow implicit conversion to py::str from C++ string-like objects
     str(const char *c = "")
@@ -1009,10 +1015,16 @@ public:
         if (!m_ptr) pybind11_fail("Could not allocate bytes object!");
     }
 
-    bytes(const char *c, size_t n)
-        : object(PYBIND11_BYTES_FROM_STRING_AND_SIZE(c, (ssize_t) n), stolen_t{}) {
+    bytes(const char *c, ssize_t n)
+        : object(PYBIND11_BYTES_FROM_STRING_AND_SIZE(c, n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate bytes object!");
     }
+
+    template <typename T, typename = detail::enable_if_t<std::is_integral<T>::value>>
+    bytes(const char *c, T n) : bytes(c, ssize_t{n}) { }
+
+    // For convenience, do not emit -Wnarrowing when passing in an unsigned size_t:
+    bytes(const char *c, size_t n) : bytes(c, ssize_t(n)) { }
 
     // Allow implicit conversion:
     bytes(const std::string &s) : bytes(s.data(), s.size()) { }
@@ -1260,9 +1272,13 @@ public:
 class tuple : public object {
 public:
     PYBIND11_OBJECT_CVT(tuple, object, PyTuple_Check, PySequence_Tuple)
-    explicit tuple(size_t size = 0) : object(PyTuple_New((ssize_t) size), stolen_t{}) {
+    explicit tuple(ssize_t size = 0) : object(PyTuple_New(size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate tuple object!");
     }
+    template <typename T, typename = detail::enable_if_t<std::is_integral<T>::value>>
+    tuple(T size) : tuple(ssize_t{size}) { }
+    // For convenience, do not emit -Wnarrowing when passing in an unsigned size_t:
+    tuple(size_t size) : tuple(ssize_t(size)) { }
     size_t size() const { return (size_t) PyTuple_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
@@ -1320,9 +1336,13 @@ public:
 class list : public object {
 public:
     PYBIND11_OBJECT_CVT(list, object, PyList_Check, PySequence_List)
-    explicit list(size_t size = 0) : object(PyList_New((ssize_t) size), stolen_t{}) {
+    explicit list(ssize_t size = 0) : object(PyList_New(size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate list object!");
     }
+    template <typename T, typename = detail::enable_if_t<std::is_integral<T>::value>>
+    list(T size) : list(ssize_t{size}) { }
+    // For convenience, do not emit -Wnarrowing when passing in an unsigned size_t:
+    list(size_t size) : list(ssize_t(size)) { }
     size_t size() const { return (size_t) PyList_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }


anntzer:
Kindly bumping?

include/pybind11/pytypes.h
```diff
@@ -936,11 +936,18 @@ class str : public object {
...
 
-    str(const char *c, size_t n)
-        : object(PyUnicode_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+    str(const char *c, ssize_t n)
+        : object(PyUnicode_FromStringAndSize(c, n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate string object!");
     }
 
+    template <typename T,
+              typename = typename std::enable_if<std::is_integral<T>::value>::type>
```
Skylion007, line 945:
```suggestion
              typename = detail::enable_if_t<std::is_integral<T>::value>>
```

include/pybind11/pytypes.h
```diff
@@ -1009,11 +1016,18 @@ class bytes : public object {
...
 
-    bytes(const char *c, size_t n)
-        : object(PYBIND11_BYTES_FROM_STRING_AND_SIZE(c, (ssize_t) n), stolen_t{}) {
+    bytes(const char *c, ssize_t n)
+        : object(PYBIND11_BYTES_FROM_STRING_AND_SIZE(c, n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate bytes object!");
     }
 
+    template <typename T,
+              typename = typename std::enable_if<std::is_integral<T>::value>::type>
```
Skylion007, line 1025:
```suggestion
              typename = detail::enable_if_t<std::is_integral<T>::value>>
```

include/pybind11/pytypes.h
```diff
@@ -1260,9 +1274,14 @@ class capsule : public object {
 class tuple : public object {
 public:
     PYBIND11_OBJECT_CVT(tuple, object, PyTuple_Check, PySequence_Tuple)
-    explicit tuple(size_t size = 0) : object(PyTuple_New((ssize_t) size), stolen_t{}) {
+    explicit tuple(ssize_t size = 0) : object(PyTuple_New(size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate tuple object!");
     }
+    template <typename T,
+              typename = typename std::enable_if<std::is_integral<T>::value>::type>
```
Skylion007, line 1281:
```suggestion
              typename = detail::enable_if_t<std::is_integral<T>::value>>
```

include/pybind11/pytypes.h
```diff
@@ -1320,9 +1339,14 @@ class sequence : public object {
 class list : public object {
 public:
     PYBIND11_OBJECT_CVT(list, object, PyList_Check, PySequence_List)
-    explicit list(size_t size = 0) : object(PyList_New((ssize_t) size), stolen_t{}) {
+    explicit list(ssize_t size = 0) : object(PyList_New(size), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate list object!");
     }
+    template <typename T,
+              typename = typename std::enable_if<std::is_integral<T>::value>::type>
```
Skylion007, line 1346:
```suggestion
              typename = detail::enable_if_t<std::is_integral<T>::value>>
```

Skylion007 (changes requested):

include/pybind11/pytypes.h
```diff
@@ -1320,9 +1339,14 @@ class sequence : public object {
 class list : public object {
 public:
     PYBIND11_OBJECT_CVT(list, object, PyList_Check, PySequence_List)
-    explicit list(size_t size = 0) : object(PyList_New((ssize_t) size), stolen_t{}) {
+    explicit list(ssize_t size = 0) : object(PyList_New(size), stolen_t{}) {
```
Skylion007, line 1342: Should we have asserts for a negative ssize?

rwgk: What does `PyList_New` to in that case?
Do we already have a unit test for the max value of `size_t`?
I'd answer the question from that angle.

Skylion007: Good point.

include/pybind11/pytypes.h
```diff
@@ -936,11 +936,18 @@ class str : public object {
...
-    str(const char *c, size_t n)
-        : object(PyUnicode_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+    str(const char *c, ssize_t n)
+        : object(PyUnicode_FromStringAndSize(c, n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate string object!");
     }
 
+    template <typename T,
+              typename = typename std::enable_if<std::is_integral<T>::value>::type>
+    str(const char *c, T n) : str(c, ssize_t{n}) { }
```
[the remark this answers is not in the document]
Skylion007: > I had to add a templated overload for all
integral types and then specialize specifically for ssize_t (the
base case) and size_t (to remove the -Wnarrowing warning for it,
for convenience/backcompat). Adding only overloads for ssize_t
and size_t would not work as py::bytes{"foo", 3} would result in
ambiguous overload resolution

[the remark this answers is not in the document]
rwgk: ah, saw that but didn't fully appreciate before, sorry, thanks.
hm, it seems very heavy-handed.
I don't feel too strongly in favor of this PR. I almost feel it's best to keep as-is, and let people decide if they want to simply cast themselves to avoid the warning, or use a helper that we could provide, e.g. `py::save_ssize_t()` which throws a `ValueError`, maybe we have that already?

include/pybind11/pytypes.h
anntzer: sure

include/pybind11/pytypes.h
anntzer: ditto

include/pybind11/pytypes.h
anntzer: ditto

include/pybind11/pytypes.h
anntzer: ditto

include/pybind11/pytypes.h
anntzer: PyList_New explicitly checks for negative sizes and errors on them: https://github.com/python/cpython/blob/4ceec495598e78f0776dd46d511dcc612a434dc3/Objects/listobject.c#L142
So in that sense (I didn't actually realize this before) I'd argue that supporting size_t inputs is actually less correct than supporting ssize_t, because the former suggests that e.g. (SIZE_T_MAX-1) would be a valid input whereas it is not.

include/pybind11/pytypes.h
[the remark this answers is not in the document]
anntzer: I'm not really sure what you suggest that the helper would do?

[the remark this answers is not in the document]
rwgk: Make the conversion safe, at runtime.
Maybe that's not what you had in mind? Is this PR just meant to suppress the `-Wnarrowing` warning? Even if there is actually a `ssize_t` overflow?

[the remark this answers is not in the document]
anntzer: My point was indeed just to suppress the warning: as noted at https://github.com/pybind/pybind11/pull/2692#discussion_r693384234, CPython itself will check the input and return NULL (which pybind11 already checks for) and set an error if there's an overflow.

include/pybind11/pytypes.h
rwgk: I agree with that. I think it's best to simply replace all `size_t` with `ssize_t`, without adding new overloads, and noting the change in the upgrade guide.

Just that will leave people out in the cold when they get the `-Wnarrowing` diagnostic, but if we also add something like a `py::safe_as_ssize_t(T)` helper in this PR we could tell them to use that. I think (very old) boost code has safe conversions from any integral type to a narrower integral type. Probably there are many solutions out there, but the boost one is likely to get all corner cases right. Or maybe there is even something in the C++ standard library? I don't know and didn't look.

In summary:
* pybind11 will be safer.
* users will have a choice, hard cast (they need to take responsibility explicitly), or the safe helper with a runtime check.

anntzer: :tada:

I can add `py::safe_as_ssize_t` if you prefer (well, as you say, it may be worth finding a good implementation somewhere, which I don't have right now), but once again, note that if the input was between SSIZE_T_MAX and SIZE_T_MAX, then we were *already* going to get a C++ exception thrown at that point (because PyList_New would have returned NULL, which pybind11 already catches and converts to a C++ exception), so I don't think another helper just so that the exception gets thrown *before* the call to PyList_New rather than after adds much?

rwgk:
> For easier bookkeeping, I converted this PR to Draft mode until I'm done with #3219. I'll close here after that one is done.

All done!

anntzer:
Great, thanks!

PR #2698, opened 2020-11-26
Only allow integer type_caster to call __int__ method when conversion is allowed; always call __index__

YannickJadoul:
## Description

As the title says and the patch shows: currently, `PyLong_AsLong`/`PyLong_AsLongLong` gets called independently of `convert` in the `type_caster` for integral types. This seems inconsistent with other type casters.

## Suggested changelog entry:

```rst
The ``type_caster`` for integers does not convert Python objects with ``__int__`` anymore with ``noconvert`` or during the first round of trying overloads.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1020,6 +1020,14 @@ public:
         if (!src)
             return false;
 
+#if !defined(PYPY_VERSION)
+        auto index_check = [](PyObject *o) { return PyIndex_Check(o); };
+#else
+        // In PyPy 7.3.3, `PyIndex_Check` is implemented by calling `__index__`,
+        // while CPython only considers the existence of `nb_index`/`__index__`.
+        auto index_check = [](PyObject *o) { return hasattr(o, "__index__"); };
+#endif
+
         if (std::is_floating_point<T>::value) {
             if (convert || PyFloat_Check(src.ptr()))
                 py_value = (py_type) PyFloat_AsDouble(src.ptr());
@@ -1027,7 +1035,7 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
-        } else if (!convert && !PyIndex_Check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
+        } else if (!convert && !index_check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
         } else if (std::is_unsigned<py_type>::value) {
             py_value = as_unsigned<py_type>(src.ptr());
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1027,6 +1027,8 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
+        } else if (!convert && !PyLong_Check(src.ptr())) {
+            return false;
         } else if (std::is_unsigned<py_type>::value) {
             py_value = as_unsigned<py_type>(src.ptr());
         } else { // signed integer:
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1027,7 +1027,7 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
-        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr()) && !PyIndex_Check(src.ptr())) {
+        } else if (!convert && !PyIndex_Check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
         } else if (std::is_unsigned<py_type>::value) {
             py_value = as_unsigned<py_type>(src.ptr());
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1027,7 +1027,7 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
-        } else if (!convert && !PyLong_Check(src.ptr())) {
+        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
         } else if (std::is_unsigned<py_type>::value) {
             py_value = as_unsigned<py_type>(src.ptr());
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1027,7 +1027,7 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
-        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr())) {
+        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr()) && !PyIndex_Check(src.ptr())) {
             return false;
         } else if (std::is_unsigned<py_type>::value) {
             py_value = as_unsigned<py_type>(src.ptr());
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -98,6 +98,10 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("i64_str", [](std::int64_t v) { return std::to_string(v); });
     m.def("u64_str", [](std::uint64_t v) { return std::to_string(v); });
 
+    // test_int_convert
+    m.def("int_passthrough", [](int arg) { return arg; });
+    m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg().noconvert());
+
     // test_tuple
     m.def("pair_passthrough", [](std::pair<bool, std::string> input) {
         return std::make_pair(input.second, input.first);
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -100,7 +100,7 @@ TEST_SUBMODULE(builtin_casters, m) {
 
     // test_int_convert
     m.def("int_passthrough", [](int arg) { return arg; });
-    m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg().noconvert());
+    m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg{}.noconvert());
 
     // test_tuple
     m.def("pair_passthrough", [](std::pair<bool, std::string> input) {
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -263,6 +263,17 @@ def test_int_convert():
         def __float__(self):
             return 41.99999
 
+    class IndexedThought(object):
+        def __index__(self):
+            return 42
+
+    class RaisingThought(object):
+        def __index__(self):
+            raise ValueError
+
+        def __int__(self):
+            return 42
+
     convert, noconvert = m.int_passthrough, m.int_passthrough_noconvert
 
     def require_implicit(v):
@@ -278,6 +289,11 @@ def test_int_convert():
     require_implicit(DeepThought())
     cant_convert(ShallowThought())
     cant_convert(FuzzyThought())
+    if not env.PY2:
+        # I have no clue why __index__ is not picked up by Python 2's PyIndex_check
+        assert convert(IndexedThought()) == 42
+        assert noconvert(IndexedThought()) == 42
+        cant_convert(RaisingThought())  # no fall-back to `__int__`if `__index__` raises
 
 
 def test_numpy_int_convert():
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -252,14 +252,14 @@ def test_integer_casting():
 
 
 def test_int_convert():
-    class DeepThought:
+    class DeepThought(object):
         def __int__(self):
             return 42
 
-    class ShallowThought:
+    class ShallowThought(object):
         pass
 
-    class FuzzyThought:
+    class FuzzyThought(object):
         def __float__(self):
             return 41.99999
 
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -289,8 +289,8 @@ def test_int_convert():
     require_implicit(DeepThought())
     cant_convert(ShallowThought())
     cant_convert(FuzzyThought())
-    if not env.PY2:
-        # I have no clue why __index__ is not picked up by Python 2's PyIndex_check
+    if env.PY >= (3, 8):
+        # Before Python 3.8, `int(obj)` does not pick up on `obj.__index__`
         assert convert(IndexedThought()) == 42
         assert noconvert(IndexedThought()) == 42
         cant_convert(RaisingThought())  # no fall-back to `__int__`if `__index__` raises
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -308,7 +308,8 @@ def test_numpy_int_convert():
     assert convert(np.intc(42)) == 42
     assert noconvert(np.intc(42)) == 42
 
-    assert convert(np.float32(3.14159)) == 3  # This might be wrong/unwanted?
+    # The implicit conversion from np.float32 is undesirable but currently accepted.
+    assert convert(np.float32(3.14159)) == 3
     require_implicit(np.float32(3.14159))
 
 
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -259,10 +259,6 @@ def test_int_convert():
     class ShallowThought:
         pass
 
-    class AlternativeThought:
-        def __index__(self):
-            return 54
-
     class FuzzyThought:
         def __float__(self):
             return 41.99999
@@ -281,8 +277,6 @@ def test_int_convert():
     assert convert(DeepThought()) == 42
     require_implicit(DeepThought())
     cant_convert(ShallowThought())
-    assert convert(AlternativeThought()) == 54
-    require_implicit(AlternativeThought())
     cant_convert(FuzzyThought())
 
 
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -251,6 +251,41 @@ def test_integer_casting():
         assert "incompatible function arguments" in str(excinfo.value)
 
 
+def test_int_convert():
+    class DeepThought:
+        def __int__(self):
+            return 42
+
+    class ShallowThought:
+        pass
+
+    class AlternativeThought:
+        def __index__(self):
+            return 54
+
+    class FuzzyThought:
+        def __float__(self):
+            return 41.99999
+
+    convert, noconvert = m.int_passthrough, m.int_passthrough_noconvert
+
+    def require_implicit(v):
+        pytest.raises(TypeError, noconvert, v)
+
+    def cant_convert(v):
+        pytest.raises(TypeError, convert, v)
+
+    assert convert(7) == 7
+    assert noconvert(7) == 7
+    cant_convert(3.14159)
+    assert convert(DeepThought()) == 42
+    require_implicit(DeepThought())
+    cant_convert(ShallowThought())
+    assert convert(AlternativeThought()) == 54
+    require_implicit(AlternativeThought())
+    cant_convert(FuzzyThought())
+
+
 def test_tuple(doc):
     """std::pair <-> tuple & std::tuple <-> tuple"""
     assert m.pair_passthrough((True, "test")) == ("test", True)
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -280,6 +280,22 @@ def test_int_convert():
     cant_convert(FuzzyThought())
 
 
+def test_numpy_int_convert():
+    np = pytest.importorskip("numpy")
+
+    convert, noconvert = m.int_passthrough, m.int_passthrough_noconvert
+
+    def require_implicit(v):
+        pytest.raises(TypeError, noconvert, v)
+
+    # `np.intc` is an alias that corresponds to a C++ `int`
+    assert convert(np.intc(42)) == 42
+    assert noconvert(np.intc(42)) == 42
+
+    assert convert(np.float32(3.14159)) == 3  # This might be wrong/unwanted?
+    require_implicit(np.float32(3.14159))
+
+
 def test_tuple(doc):
     """std::pair <-> tuple & std::tuple <-> tuple"""
     assert m.pair_passthrough((True, "test")) == ("test", True)


henryiii:
Milestone?

YannickJadoul:
I'd consider this a fix; do you agree, @henryiii? If not, feel free to change to 2.7.0

rwgk:
Observation only, no analysis yet:

This PR breaks https://github.com/google/mediapipe/blob/master/mediapipe/python/packet_test.py
The test output is below.
The test runs fine without this PR, after patching in this PR it fails.

```
[ RUN      ] PacketTest.test_boolean_packet
[       OK ] PacketTest.test_boolean_packet
[ RUN      ] PacketTest.test_bytes_packet
[       OK ] PacketTest.test_bytes_packet
[ RUN      ] PacketTest.test_detection_proto_packet
[       OK ] PacketTest.test_detection_proto_packet
[ RUN      ] PacketTest.test_double_packet
[       OK ] PacketTest.test_double_packet
[ RUN      ] PacketTest.test_empty_packet
[       OK ] PacketTest.test_empty_packet
[ RUN      ] PacketTest.test_float_array_packet
[       OK ] PacketTest.test_float_array_packet
[ RUN      ] PacketTest.test_float_image_frame_packet
[       OK ] PacketTest.test_float_image_frame_packet
[ RUN      ] PacketTest.test_float_packet
[       OK ] PacketTest.test_float_packet
[ RUN      ] PacketTest.test_float_vector_packet
[       OK ] PacketTest.test_float_vector_packet
[ RUN      ] PacketTest.test_image_frame_packet_copy_creation_with_cropping
[       OK ] PacketTest.test_image_frame_packet_copy_creation_with_cropping
[ RUN      ] PacketTest.test_image_frame_packet_creation_copy_mode
[       OK ] PacketTest.test_image_frame_packet_creation_copy_mode
[ RUN      ] PacketTest.test_image_frame_packet_creation_reference_mode
[       OK ] PacketTest.test_image_frame_packet_creation_reference_mode
[ RUN      ] PacketTest.test_int16_packet
[  FAILED  ] PacketTest.test_int16_packet
[ RUN      ] PacketTest.test_int32_packet
[  FAILED  ] PacketTest.test_int32_packet
[ RUN      ] PacketTest.test_int64_packet
[  FAILED  ] PacketTest.test_int64_packet
[ RUN      ] PacketTest.test_int8_packet
[  FAILED  ] PacketTest.test_int8_packet
[ RUN      ] PacketTest.test_int_array_packet
[       OK ] PacketTest.test_int_array_packet
[ RUN      ] PacketTest.test_int_packet
[  FAILED  ] PacketTest.test_int_packet
[ RUN      ] PacketTest.test_int_vector_packet
[       OK ] PacketTest.test_int_vector_packet
[ RUN      ] PacketTest.test_matrix_packet
[       OK ] PacketTest.test_matrix_packet
[ RUN      ] PacketTest.test_matrix_packet_with_non_c_contiguous_data
[       OK ] PacketTest.test_matrix_packet_with_non_c_contiguous_data
[ RUN      ] PacketTest.test_packet_vector_packet
[       OK ] PacketTest.test_packet_vector_packet
[ RUN      ] PacketTest.test_string_packet
[       OK ] PacketTest.test_string_packet
[ RUN      ] PacketTest.test_string_to_packet_map_packet
[       OK ] PacketTest.test_string_to_packet_map_packet
[ RUN      ] PacketTest.test_string_vector_packet
[       OK ] PacketTest.test_string_vector_packet
[ RUN      ] PacketTest.test_uint16_image_frame_packet
[       OK ] PacketTest.test_uint16_image_frame_packet
[ RUN      ] PacketTest.test_uint16_packet
[  FAILED  ] PacketTest.test_uint16_packet
[ RUN      ] PacketTest.test_uint32_packet
[  FAILED  ] PacketTest.test_uint32_packet
[ RUN      ] PacketTest.test_uint64_packet
[       OK ] PacketTest.test_uint64_packet
[ RUN      ] PacketTest.test_uint8_image_frame_packet
[       OK ] PacketTest.test_uint8_image_frame_packet
[ RUN      ] PacketTest.test_uint8_packet
[  FAILED  ] PacketTest.test_uint8_packet
======================================================================
ERROR: test_int16_packet (__main__.PacketTest)
PacketTest.test_int16_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 73, in test_int16_packet
    p2 = mp.packet_creator.create_int16(np.int16(1))
TypeError: create_int16(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_int32_packet (__main__.PacketTest)
PacketTest.test_int32_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 86, in test_int32_packet
    p2 = mp.packet_creator.create_int32(np.int32(1))
TypeError: create_int32(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_int64_packet (__main__.PacketTest)
PacketTest.test_int64_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 96, in test_int64_packet
    p2 = mp.packet_creator.create_int64(np.int64(1))
TypeError: create_int64(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_int8_packet (__main__.PacketTest)
PacketTest.test_int8_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 61, in test_int8_packet
    p2 = mp.packet_creator.create_int8(np.int8(1))
TypeError: create_int8(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_int_packet (__main__.PacketTest)
PacketTest.test_int_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 49, in test_int_packet
    p2 = mp.packet_creator.create_int(np.intc(1))
TypeError: create_int(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_uint16_packet (__main__.PacketTest)
PacketTest.test_uint16_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 120, in test_uint16_packet
    p2 = mp.packet_creator.create_uint16(np.uint16(1))
TypeError: create_uint16(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_uint32_packet (__main__.PacketTest)
PacketTest.test_uint32_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 132, in test_uint32_packet
    p2 = mp.packet_creator.create_uint32(np.uint32(1))
TypeError: create_uint32(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

======================================================================
ERROR: test_uint8_packet (__main__.PacketTest)
PacketTest.test_uint8_packet
----------------------------------------------------------------------
  File "mediapipe/python/packet_test.py", line 108, in test_uint8_packet
    p2 = mp.packet_creator.create_uint8(np.uint8(1))
TypeError: create_uint8(): incompatible function arguments. The following argument types are supported:
    1. (arg0: int) -> mediapipe.python._framework_bindings.packet.Packet

Invoked with: 1

----------------------------------------------------------------------
Ran 31 tests in 0.523s

FAILED (errors=8)
```

henryiii:
`__index__` is somewhat special. Would it work if `__int__` was not called automatically, but `__index__` was?

> `object.__index__(self)`
>
> Called to implement `operator.index()`, and whenever Python needs to losslessly convert the numeric object to an integer object (such as in slicing, or in the built-in `bin()`, `hex()` and `oct()` functions). Presence of this method indicates that the numeric object is an integer type. Must return an integer.

Therefore, having an `__index__` means that this object _is_ an integer type, with a lossless representation as an integer. `__int__` means you can convert it to an integer, possibly/probably lossy, and the object is possibly not an integer.

rwgk:
`test_int_convert_numpy` is a minimal example of the mediapipe issue.

Main conclusion: this PR is working as intended, the mediapipe code needs to remove `py::arg().noconvert()`.

My feeling about milestone: given that we now know this PR breaks someone, maybe 2.7 is better?

Might be nice to include the extra test below in this PR. It's similar to `test_numpy_bool` but uses `int_passthrough`, i.e. adds a little bit of coverage.

Tangential: `test_numpy_bool` `cant_convert(np.zeros(2, dtype="int"))` seems to convolute two things: an array cannot be converted to a scalar, `int` vs `bool`. Note that `cant_convert(np.zeros(2, dtype="bool"))` also passes the test. How about this instead (to keep the two things separate)?
```
    pytest.raises(TypeError, convert, np.zeros(2, dtype="bool"))
    pytest.raises(TypeError, noconvert, np.int32(0))
```

```
diff --git a/tests/test_builtin_casters.py b/tests/test_builtin_casters.py
index 3942482..f130dd2 100644
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -280,6 +280,16 @@ def test_int_convert():
     cant_convert(FuzzyThought())
 
 
+def test_int_convert_numpy():
+    np = pytest.importorskip("numpy")
+
+    convert, noconvert = m.int_passthrough, m.int_passthrough_noconvert
+
+    v = np.int32(84)
+    assert convert(v) == 84
+    pytest.raises(TypeError, noconvert, v)
+
+
 def test_tuple(doc):
     """std::pair <-> tuple & std::tuple <-> tuple"""
     assert m.pair_passthrough((True, "test")) == ("test", True)
```

YannickJadoul:
> `__index__` is somewhat special. Would it work if `__int__` was not called automatically, but `__index__` was?

The problem is that both are treated equally by `PYBIND11_LONG_CHECK`, I believe?
But yes, that would make sense.

I feel stupid for not documenting this better, but I believe this was tackling a complaint from Gitter, where the conversion from `float` to `int` is considered `noconvert` (since `float` has `__int__`). I'll check tomorrow if I can figure out the context!

So this was also why I considered this as a bugfix.

> My feeling about milestone: given that we now know this PR breaks someone, maybe 2.7 is better?

But @rwgk's comment also makes sense, from that respect, yes. Though ... well, it's very tricky behavior to depend on, I'd say?

> Might be nice to include the extra test below in this PR. It's similar to `test_numpy_bool` but uses `int_passthrough`, i.e. adds a little bit of coverage.

I'll have a look at the tests as well, but in principle, these NumPy types are just as well just types with or without `__int__`/`__index__` methods, no? So this should already be covered by my explicit, custom test classes `DeepThought` and/or `ShallowThought`, I believe? I'll check again, though.

rwgk:
> I'll have a look at the tests as well, but in principle, these NumPy types are just as well just types with or without `__int__`/`__index__` methods, no? So this should already be covered by my explicit, custom test classes `DeepThought` and/or `ShallowThought`, I believe? I'll check again, though.

Yes, that's true, if you assume numpy behaves that way (seems like a very safe assumption).
Clearly, the test isn't essential, but I'd add those few lines there anyway, mainly to point people (like the mediapipe folks) to, which will imply the message: here, we thought about it, we know, that's the behavior we want, you need to adjust.

bstaletic:
Those numpy types do simply implement `__int__`. I was going to argue in favour of sticking to milestone 2.6.2, but after thinking about it a bit more, I think @rwgk is right. We're essentially changing what pybind11 considers "an int".

henryiii:
`PYBIND11_LONG_CHECK`

Does that need to be updated or a different check used? There should be a difference between a class with only `__int__` (possibly lossy conversion, like a float), and `__index__` ("I am an int").

henryiii:
I think this is a 2.7 fix then, and I also still think `__index__` should not be "a conversion", but `__int__` should be - that's what those methods mean.

YannickJadoul:
I fixed the `__index__` thing, demonstrated by a new test (which failed, before). This change should now be way less intrusive (not saying I need it into 2.6.2, but there's a chance it would fit, now).

[`PyLong_AsLongLong` (and others)](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong) only call `__index__` from 3.8 onwards, but always call `__int__`, so I think we're fine in this respect? Only if `__int__` and `__index__` do different things, we're in trouble, but if that's the case, I'm not blaming pybind11. (Also, `__int__` being deprecated is now deprecated and gives a warning in Python >= 3.8, so I assume this is part of the same change, where `PyLong_AsLongLong` switches to ónly take `__index__` into account?)

Let's see why Python 2.7 is unhappy. *sigh*

tests/test_builtin_casters.py
```diff
@@ -251,6 +251,51 @@ def test_integer_casting():
...
+    convert, noconvert = m.int_passthrough, m.int_passthrough_noconvert
+
+    def require_implicit(v):
+        pytest.raises(TypeError, noconvert, v)
+
+    # `np.intc` is an alias that corresponds to a C++ `int`
+    assert convert(np.intc(42)) == 42
+    assert noconvert(np.intc(42)) == 42
+
+    assert convert(np.float32(3.14159)) == 3  # This might be wrong/unwanted?
```
YannickJadoul, line 295: This is also kind of weird. It's because we only explicitly test for `float` and not for "float-like" types. I'm not sure if there's a general way to recognize "float-like" types, though. We could just add a special case for NumPy float types?

Not really related to this PR, though, so definitely another PR.

YannickJadoul:
> Let's see why Python 2.7 is unhappy. _sigh_

*even deeper sigh* Old-style classes...

tests/test_builtin_casters.py
```diff
@@ -251,6 +251,67 @@ def test_integer_casting():
...
+
+    assert convert(7) == 7
+    assert noconvert(7) == 7
+    cant_convert(3.14159)
+    assert convert(DeepThought()) == 42
+    require_implicit(DeepThought())
+    cant_convert(ShallowThought())
+    cant_convert(FuzzyThought())
+    if not env.PY2:
+        # I have no clue why __index__ is not picked up by Python 2's PyIndex_check
```
YannickJadoul, line 293: Too bad for Python 2? I don't know what went wrong, here...

This does work in Python 2.7, though:

```pycon
[GCC 7.5.0] on linux2
>>> x = list(range(50))
>>> class IndexedThought(object):
...     def __index__(self):
...         return 42
... 
>>> x[IndexedThought()]
42
```

YannickJadoul:
Ready for another review

include/pybind11/cast.h
```diff
@@ -1027,6 +1027,8 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
+        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr()) && !PyIndex_Check(src.ptr())) {
```
rwgk, line 1030: Maybe reverse the order of `PYBIND11_LONG_CHECK` and `PyIndex_Check`?
`PyIndex_Check` is a very simple lookup of struct members:
https://github.com/python/cpython/blob/2edfc86f69d8a74f4821974678f664ff94a9dc22/Include/internal/pycore_abstract.h#L13
`PyLong_Check` seems more involved:
https://github.com/python/cpython/blob/2edfc86f69d8a74f4821974678f664ff94a9dc22/Include/longobject.h#L14

tests/test_builtin_casters.py
rwgk: I agree, although it looks tricky to implement, therefore I wonder about cost/benefit of working on this.
Pragmatic idea: report the situation in a comment and wait for an opportune moment in the future (e.g. when there is an actual problem that needs to be resolved). My first stab:
```
# The implicit conversion from np.float is undesirable but difficult to detect, without making the pybind11
# int converter dependent on numpy. Currently Python does not have a generalized concept of float-like types.
```

include/pybind11/cast.h
YannickJadoul: Makes sense, yes. Will do.

tests/test_builtin_casters.py
YannickJadoul: We do already have something like this (a special case for NumPy) in the `type_caster` for bool, though.

But yes, this might be something we'll have to live with? At the very least, this should now be considered a "conversion".

YannickJadoul:
I need to look into these new failures, but, @rwgk, this might break a lot fewer tests, on your end?

include/pybind11/cast.h
YannickJadoul: In the end, `PyLong_Check` is also pretty trivial, though: https://github.com/python/cpython/blob/2edfc86f69d8a74f4821974678f664ff94a9dc22/Include/object.h#L671-L684

YannickJadoul:
I think this should fix the tests (it does so on my machine anyway, where I had failed to run a _3 <= Python version < 3.8_, before).

Remaining issue: [`PyLong_AsLongLong`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong) does not call `__index__` before 3.8. This is consistent with calling `int(...)` but not with `some_list[...]`.

Python 3.8:
```pycon
>>> int(IndexedThought())
42
>>> list(range(100))[IndexedThought()]
42
```

Python 3.6:
```pycon
>>> int(IndexedThought())
TypeError: int() argument must be a string, a bytes-like object or a number, not 'IndexedThought'
>>> list(range(100))[IndexedThought()]
42
```

Python 2.7:
```pycon
>>> int(IndexedThought())
TypeError: int() argument must be a string or a number, not 'IndexedThought'
>>> list(range(100))[IndexedThought()]
42
```

So should we make this consistent, and always call `PyNumber_Index` in versions < 3.8? I'm not sure, but we could. We could also say: "this is just how Python versions < 3.8 handle int conversions, so pybind11 is not going to change that (and you should update anyway)".

Minor extra inconsistency: in versions < 3.6, we now check `PyIndex_Check` (whether it has an `__index__` and is thus an integer numeric type!) to determine if the conversion is `noconvert`, but later, `PyLong_AsLongLong` still calls `__int__`. That's ... weird, at least. And maybe an argument to say we should manually call `PyNumber_Index` anyway?

rwgk:
I quick observation: https://github.com/google/mediapipe/blob/master/mediapipe/python/packet_test.py passes with the current version of this PR (915423de979a44a27be4f52c5378733a6c40e26f).
I verified that the mediapipe bindings still have the `.noconvert()` (that code was last changed in on Nov 17).

tests/test_builtin_casters.py
rwgk: The Google-global testing of this PR had no issues.
I see two behavior changes here: 1. the one currently implemented, and 2. potentially also raising a TypeError for np.float types (with ifdefs I guess to use the numpy api if available).
Given that this PR is exhaustively tested and doesn't disturb anything anymore, I feel it's best to merge now (for 2.6.2), and try the 2. behavior change in a separate PR. The main rationale for this two-stage approach is to use our time efficiently. Putting off this well-tested change will cost extra time.
For the comment, this is my 2nd suggested version:
```
# The implicit conversion from np.float is undesirable but currently accepted.
```
Just a hint that we thought about it, leaving all options open for the future.

rwgk (approved):

tests/test_builtin_casters.py
YannickJadoul: Agreed. Never meant to lump this in this PR. I really thought it would just be a straightforward one, but once again, Python had some surprises.

YannickJadoul: Btw, to make things more complicated, `np.float` to `int` isn't allowed, because `np.float` is just a Python `float` (I'm not joking; this is really the case). Anyway, I'll add `32` and commit this suggestion.

YannickJadoul:
Added back to 2.6.2, as @rwgk suggested; we can always switch back to 2.7 (or 2.6.3), if there's no time to get this fully reviewed and in.

There's still the issue from https://github.com/pybind/pybind11/pull/2698#issuecomment-752218748 that doesn't really feel nice to me. I'm wondering if we shouldn't strive for more consistency in pybind11, and adopt 3.8+ behavior.

YannickJadoul:
I've moved the last question to be decided to a more concrete PR, #2801, demonstrating the implementation is straightforward. All that's left is arguing why we do or don't want this.

YannickJadoul:
#2801 still discovered an inconsistency with PyPy. @rwgk, do have a look at the latest 0e89b14e92324c0257dd2a82d101c77efe3c09d9, but I'm assuming you're fine with this since it doesn't change CPython behavior?

rwgk (approved):
Yes, the PyPy fix looks great to me. Thank @YannickJadoul!

YannickJadoul:
Except that ICC isn't happy with this. It's hardly been a day... :-|

henryiii:
Aren't you glad we added that check? 😉

YannickJadoul:
> Aren't you glad we added that check? wink

I'm thrilled. Anyway, yeah, otherwise that other PR had to do it.

I'm trying to get rid of that lambda, by means of a macro, I guess. Wouldn't know what else it could be the issue.

Sorry for the mess. Things worked smoothly, locally.

tests/test_builtin_casters.cpp
```diff
@@ -98,6 +98,10 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("i64_str", [](std::int64_t v) { return std::to_string(v); });
     m.def("u64_str", [](std::uint64_t v) { return std::to_string(v); });
 
+    // test_int_convert
+    m.def("int_passthrough", [](int arg) { return arg; });
+    m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg().noconvert());
```
henryiii, line 103: This kills ICC.

henryiii: This is also why I'm continuing to work with the compiler team to fix this.
```suggestion
    m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg{}.noconvert());
```

YannickJadoul: Aaaaaah, of course!

YannickJadoul: So it's just the new push, and not the last changes. Thanks; I'll push a fix!

YannickJadoul:
Submitted as https://foss.heptapod.net/pypy/pypy/-/issues/3383

YannickJadoul:
Oof, all green again 😅
That stress really wasn't necessary, ICC.

YannickJadoul:
@henryiii, #2801 is also green, now, so as far as I'm concerned, this can go in, as the one loose end is handled in #2801.

Issue #2332, opened 2020-07-27
Why pybind11 generated enum type is not PEP 435 compatible. say, it doesn't have value attribute and it's not iterable.


frank-xlj:
Make sure you've completed the following steps before submitting your issue -- thank you!

1. Check if your question has already been answered in the [FAQ](http://pybind11.readthedocs.io/en/latest/faq.html) section.
2. Make sure you've read the [documentation](http://pybind11.readthedocs.io/en/latest/). Your issue may be addressed there.
3. If those resources didn't help and you only have a short question (not a bug report), consider asking in the [Gitter chat room](https://gitter.im/pybind/Lobby).
4. If you have a genuine bug report or a more complex question which is not answered in the previous items (or not suitable for chat), please fill in the details below.
5. Include a self-contained and minimal piece of code that reproduces the problem. If that's not possible, try to make the description as clear as possible.

*After reading, remove this checklist and the template text in parentheses below.*

## Issue description

(Provide a short description, state the expected behavior and what actually happens.)

## Reproducible example code

(The code should be minimal, have no external dependencies, isolate the function(s) that cause breakage. Submit matched and complete C++ and Python snippets that can be easily compiled and run to diagnose the issue.)

bstaletic:
Because it predates `enum.Enum` by a lot.

You can use `.def()` on `py::enum_`s and give them `values()` and `__iter__()` to get what you want.

frank-xlj:
Thanks @bstaletic for the comments. So pybind11 doesn't have the plan to make enum type PEP-435 compatible to keep backward compatibility, right?

bstaletic:
There are no backwards compatibility concerns here. For `value`, it's just a simple

    .def("value", [](enum_type e) { return static_cast<int>(e); })

Or you can do it from python with `int(e)`. This is quite a small addition, so maybe we should add it.

As for `__iter__`, python implements it by storing a dictionary of `name`s to `enum_values` inside the enum. I don't know how to implement that efficiently (without allocating), which doesn't mean it's impossible. Inefficient implementation would only hurt those who do not need iteration over an enum. Again, users can implement these on their own.

frank-xlj:
got it. Thanks.

YannickJadoul:
Seems solved/answered.

EricCousineau-TRI:
Coming back to this, I would love to have this in `pybind11` proper!

[...]

Issue #253, opened 2016-06-27
Iterate over enums


romainthomas:
Actually we can't iterate over a enum. It could be nice to enable this kind of code:

``` python
for e in myEnum:
    print e
```

wjakob:
Are you proposing to implement such a feature?

aldanor:
A side thought, would it make sense to consider using standard library's `enum` inside of rolling our own? (only in Python 3, but there's an official backport to Python 2)

https://docs.python.org/3/library/enum.html

wjakob:
It sounds tricky -- ``Enum`` is a kind of magic class in Python3, and it requires installing an extra library on python2. I think duck typing is the way to go here.

romainthomas:
> Are you proposing to implement such a feature? A PR with a custom __iter__ handler would be welcomed.

@wjakob I will try.

Actually I use this trick:
```python
import re
def enum_to_dict(enum):
    return {k: v for k, v in enum.__dict__.iteritems() if not re.match("__(.*)__", str(k))}
```

wjakob:
I'll close this ticket since there isn't an issue with pybind11 -- if you'd like to contribute this as a feature, please open a pull request.

Thanks,
Wenzel

PR #2704, opened 2020-11-30
Standard library enum support via an optional include

AWhetter:
## Description

This change is a proof of concept implementation that allows C++ enums to be exposed as `enum.Enum` objects on the Python side. I've decided to make this change partly to spark a design discussion about what would make the most sense when trying to expose C++ enums as standard library enums.

* Should this functionality exist using a new class (as done here) or should we try to make it work with `enum_`? This implementation uses a new object type `stdlib_enum`. Personally the backwards compatibility problems seem complex enough that making a new enum type is worthwhile, but there's a tradeoff of increased binary size. I think we would want to phase out the old enum class in favour of this new one.
* Should you be able to add methods to enums? This one does not, but it's inflexible as a result.
* Should you be allowed to select the type of enum to create (`Enum`, `IntEnum`, `IntFlag`)? If so, how? An optional argument to the constructor seems like the obvious choice.
* What should the interface for creating these classes look like? Does using a destructor make sense as implemented here, does a `finalize()` method make more sense? How should that interact with being able to add methods (if at all)?
* Does an optional include (as done here) make sense or should this be integrated into pybind properly?

Currently there are no docs or tests because the design might change a lot.

Kudos to @aldanor who's design I started from.

Implemented #530

## Suggested changelog entry:

```rst
TODO upon design decisions
```

### DIFF
--- a/include/pybind11/stdlib_enum.h
+++ b/include/pybind11/stdlib_enum.h
@@ -1,0 +1,161 @@
+/*
+    pybind11/stdlib_enum.h: Declaration and conversion enums as Enum objects.
+
+    Copyright (c) 2020 Ashley Whetter <ashley@awhetter.co.uk>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include "detail/common.h"
+#include "pybind11.h"
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+template<typename U>
+struct enum_mapper {
+    handle type = {};
+    std::unordered_map<U, handle> values = {};
+
+    enum_mapper(handle type, const dict& values) : type(type) {
+        for (auto item : values) {
+            this->values[item.second.cast<U>()] = type.attr(item.first);
+        }
+    }
+};
+
+template<typename T>
+struct type_caster<T, enable_if_t<std::is_enum<T>::value>> {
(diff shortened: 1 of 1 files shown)


diiigle:
Just bounced over this coming from #253.

anntzer:
> Should you be allowed to select the type of enum to create (Enum, IntEnum, IntFlag)? If so, how? An optional argument to the constructor seems like the obvious choice.

Being able to pick the base stdlib class is nice, at least for using `Flag`/`IntFlag` which have highly readable reprs.  In https://gist.github.com/anntzer/96f27c0b88634dbc61862d08bff46d10 (my own implementation of binding-to-stdlib enums), I simply pass the base class as a string, which in theory would even allow using third-party base enum classes with compatible signatures (e.g. for backporting to Py<3.4).

EricCousineau-TRI:
Ooh, saw this per @bstaletic's post in #2739 (issue: #2332). Will compare notes.

EricCousineau-TRI:
I spewed some text here: https://github.com/pybind/pybind11/pull/2739#issuecomment-750496252

If y'all have a chance, might you be able to weigh in on my opinions there? (i.e. rolling all the functionality into `py::enum_` proper, rather than relying on macros / `type_caster` specialization)

james-ro-williams:
Why didn't this go anywhere? This is still an issue in 2025!!

rwgk:
> Why didn't this go anywhere? This is still an issue in 2025!!

Did you see #5555, coincidentally merged yesterday.

This is a volunteer-driven community project.

AWhetter:
I'm going to close this now that we have `native_enum` :tada:

PR #2705, opened 2020-11-30
Adjusting `type_caster<std::reference_wrapper<T>>` to support const/non-const propagation in `cast_op`.

laramiel:
## Description

Before `std::reference_wrapper<T>` would invoke `cast_op<type>` rather than `cast_op<T&>` to construct the conversion operator type. This introduces different behavior for reference wrapped types when the underlying `type_caster<>::cast_op<T>&` differs from `type_caster<>::cast_op<T&>`.

There is additional enforcement from a `static_assert` that requires the type resulting from `type_caster<>::cast_op<T>` to be equivalent to `intrinsic_t<T>` (essentially `std:decay<T>`).

After, the `type_caster<>` specialization invokes `cast_op<type&>`, which allows `std::reference_wrapper<>` to behave in the same way as a native reference type.

Also, the `static_assert` is changed to allow the resulting type  to be either `T&` or `const T&`, which gives slightly more flexibility to `type_caster` overloads.

This change is expected to be low impact for existing code and tests. Existing tests use a use the default implementation of cast_op which applies intrinsic_t to the result, effectively ignoring const, and making 

The added test case, test_freezable_type_caster, demonstrates custom casting to either mutable or immutable types, and is a cleanroom test based somewhat on ideas from https://github.com/protocolbuffers/protobuf/tree/master/python/google/protobuf/pyext


rwgk (commented):
Hi Laramie, I tried out your change locally, with small changes applied on top:

https://github.com/rwgk/pybind11/commit/86299e3751270d356c72c8bc11743fd2b6f57553

That commit adds back the original code with `#if #else #endif`, and uncomments one of your additional tests.

I found that your additional tests work with and without your change to `type_caster<std::reference_wrapper<type>>`. Is that expected? Would it be possible to add tests that fail without your change? 

For completeness: I was using clang version 9.0.1 (gLinux rodette) with `-std=c++17`.

laramiel:
... except you changed the test. The behavior below fails:

    assert m.refwrap_lvalue_const().value == 1
    m.refwrap_lvalue_const().value =2
    assert m.refwrap_lvalue_const().value == 1   # not 2

The problem with your uncommented test is that pybind11 doesn't really have a mechanism to note mutable vs. immutable value types that correspond to c++ references vs c++ const references. It's possible to construct them by implementing a custom type_caster that manages a lot of the complexity itself. A skeleton might look like this:

```
template <>
struct type_caster<UserType> {

  static constexpr auto name = _<UserType>();

  // C++ -> Python
  static handle cast(const UserType *src, return_value_policy policy,
                     handle parent) {
     // ... construct an immutable python type referencing src.  This isn't really available in pybind11
    // and would have to be built by hand.
  }
  static handle cast(const UserType &src, return_value_policy policy, handle parent) {
   return cast(&src, policy, parent);
  }

  static handle cast(UserType *src, return_value_policy policy,
                     handle parent) {
    // ... construct a mutable python type referencing src. This is the default pybind11 path.
  }
  static handle cast(UserType &src, return_value_policy policy, handle) {
   return cast(&src, policy, parent);
  }

  static handle cast(UserType &&src, return_value_policy policy, handle) {
    // construct a mutable python type owning src.
  }

  // Convert Python->C++.
  	// ... either reference a wrapped c++ type or construct a new one. Good luck. :)
  }

  // cast_op_type determines which operator overload to call for a given c++
  // input parameter type. In this case we want to propagate const, etc.
  template <typename T_>
  using cast_op_type =
      std::conditional_t<
          std::is_same_v<std::remove_reference_t<T_>, const UserType *>,
              const UserType *,
      std::conditional_t<
          std::is_same_v<std::remove_reference_t<T_>, UserType *>, UserType *,
      std::conditional_t<
          std::is_same_v<T_, const UserType &>, const UserType &,
      std::conditional_t<std::is_same_v<T_, UserType &>, UserType &,
      /*default is T&&*/ T_>>>>;


  // PYBIND11_TYPE_CASTER
  operator const UserType *() { return value; }
  operator const UserType &() {
    if (!value) throw reference_cast_error();
    return *value;
  }
  operator ProtoType *() { return value; }
  operator ProtoType &() {
    if (!value) throw reference_cast_error();
    return *const_cast<UserType*>(value);
  }
  operator UserType &&() && {
    if (!value) throw reference_cast_error();
    owned = *value;
    return std::move(owned);
  }

 protected:
  const UserType *value;
  UserType owned;
};
```

But, well, putting that into the test is a bit much.

laramiel:
... so, the tests are expected to pass in both versions, but the custom type caster might instead choose to copy const&, for example.

rwgk:
My take:

* The commented test is best uncommented and changed so that it passes, even if we don't like the current behavior. In a terse comment we can acknowledge and explain. That way people know exactly what to expect. — Note that the loss of `const` isn't just limited to this situation. It existed in Boost.Python already, and is very difficult to avoid. (I wouldn't try.)

* Adding a test for the situation you're making the change for is important, even if the test involves many more lines of code than the behavior change itself. The test is the only protection you'll have against someone later breaking the new behavior you want to rely on.

I'll add two other maintainers as reviewers, maybe they have different takes. Two approvals from maintainers are needed to merge.

YannickJadoul:
> Two approvals from maintainers are needed to merge.

No, as far as I know, we have never agreed on any explicit rules. Depends on the available expertise and the intrusiveness of a change.

laramiel:
Ok, this test_chimera demonstrates the requirement by creating a custom type and using pybind11 type_caster to convert between it and the c++ type.  The custom type holds a pointer to an c++ object and a flag that indicates mutability.

laramiel:
... And I've pushed the other change that demonstrates that const-references are made non-const by pybind11.

tests/test_chimera.cpp
```diff
@@ -0,0 +1,408 @@
+/*
+    tests/test_chimera.cpp -- This demonstrates a hybrid usage of pybind11, where the
```
rwgk, line 2: Casters for types not wrapped with `py::class_` are perfectly normal: `chrono.h`, `complex.h`, `eigen.h`, `functional.h`, `numpy.h`, `stl.h` and probably many more in client code.

I think it's nice to have this test, even though it's more an end-to-end test than a unit test, but it's important to explain here, for example starting with:

This original purpose of this test is to exercise that `type_caster<std::reference_wrapper<T>>` works for sub-casters that only have a `const T&` operator, but not a `T&` operator. ... continue with explanation why this is not a more focused unit test ...

tests/test_chimera.cpp
```diff
@@ -0,0 +1,408 @@
...
+#endif
+};
+
+
+#if defined(PY38_WARNING_WORKAROUND_ENABLED)
+#undef PY38_WARNING_WORKAROUND_ENABLED
+#pragma GCC diagnostic pop
+#endif
+
+static std::unordered_map<Chimera*, void*>* mapping =
```
rwgk, line 170: The shorter
`static std::unordered_map<Chimera*, void*> mapping;`
(with `mapping->` below replaced by `mapping`) also works. Is there a strong reason for allocating `mapping` on the heap? (Its size is only 56 bytes in a 64-bit build.)

Similarly for the `shared` and `shared_const` variables below.

tests/test_chimera.cpp
```diff
@@ -0,0 +1,408 @@
...
+#include <map>
+#include <type_traits>
+#include <unordered_map>
+#include <cstring>
+#include <memory>
+
+#include "pybind11_tests.h"
+
+/// C++ type
+class Chimera {
```
rwgk, line 21: `class FreezableInt`
?
"Chimara" fits for the situation as a whole, but not for this type by itself.
Meaningful naming has a huge impact on readability, especially if this is meant to be a demo.

rwgk (commented):
The suggestions are purely for the tests and mostly even just about naming and wording of comments. The core change is definitely fine. I ran it through the Google-global testing and found no issues.

tests/test_chimera.cpp
laramiel: Done.

include/pybind11/cast.h
```diff
@@ -960,9 +960,15 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
...
-            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
+    using reference_t = typename std::add_lvalue_reference<type>::type;
+    using subcaster_cast_op_type =
+        typename caster_t::template cast_op_type<reference_t>;
+
+    static_assert((std::is_same<typename std::remove_const<type>::type &,
+                             subcaster_cast_op_type>::value ||
+                std::is_same<reference_t, subcaster_cast_op_type>::value),
+               "std::reference_wrapper<T> caster requires T to have a caster "
+               "with an `T &` operator or `const T&` operator");
```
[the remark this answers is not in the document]
laramiel: Done.

tests/CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ set(PYBIND11_TEST_FILES
     test_constants_and_functions.cpp
     test_copy_move.cpp
     test_custom_type_casters.cpp
+    test_chimera.cpp
```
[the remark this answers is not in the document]
laramiel: Done.

tests/test_builtin_casters.py
```diff
@@ -324,6 +325,15 @@ def test_reference_wrapper():
         m.refwrap_usertype(None)
     assert "incompatible function arguments" in str(excinfo.value)
 
+    assert m.refwrap_lvalue().value == 1
+    m.refwrap_lvalue().value = 4
+    assert m.refwrap_lvalue().value == 4
+
+    # const-ness is not propagated.
```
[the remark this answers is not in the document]
laramiel: Done.

tests/test_chimera.cpp
laramiel: Done.

tests/test_chimera.cpp
laramiel: Done.

tests/test_chimera.cpp
laramiel: Done.

tests/test_freezable_type_caster.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_roundtrip_const():
+    m.reset(1)
+    # by value, so the const ref is converted.
+    assert m.roundtrip(m.get_const_ref()).x == 3
+
+    # by reference, so it's not converted.
+    with pytest.raises(TypeError):
+        assert m.roundtrip_ref(m.get_const_ref()).x == 4
```
rwgk, line 78: The `assert` isn't actually needed, this by itself raises the `TypeError`:
```
        m.roundtrip_ref(m.get_const_ref())
```

tests/test_freezable_type_caster.cpp
```diff
@@ -0,0 +1,413 @@
+/*
+    tests/test_freezable_type_caster.cpp
+
+    This test verifies that const-ness is propagated through the type_caster framework.
```
rwgk, line 4: Thank you for this line specifically!

rwgk (approved):
@henryiii or @bstaletic, is there a chance you could help giving this a second look? The core change (in `cast.h`) is very small. I double-checked that the new test fails without the core change, i.e. we certainly have full coverage for the change.

tests/test_freezable_type_caster.py
laramiel: Done.

rwgk:
I saw the CI / 🐍 pypy2 • ubuntu-latest • x64 failure we're getting here also on current pure master. It is definitely unrelated to this PR.

include/pybind11/cast.h
```diff
@@ -960,9 +960,15 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
 private:
     using caster_t = make_caster<type>;
     caster_t subcaster;
-    using subcaster_cast_op_type = typename caster_t::template cast_op_type<type>;
-    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value,
-            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
+    using reference_t = typename std::add_lvalue_reference<type>::type;
+    using subcaster_cast_op_type =
```
YannickJadoul, line 964: If this is changed, why does a few lines later

```c++
operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type&(); }
```

not become

```c++
operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type(); }
```
?

YannickJadoul (commented):
If I revert the change in `cast.h`, all tests pass (except for the newly added `test_freezable_type_caster.cpp` that I haven't yet looked at).

So it seems the extra tests added do not really test anything?

tests/test_freezable_type_caster.cpp
```diff
@@ -0,0 +1,418 @@
...
+#include "pybind11_tests.h"
+
+/// C++ type
+class FreezableInt {
+ public:
+  int64_t x = 1;
+};
+
+/// Python wrapper for C++ type which supports mutable/immutable variants.
+typedef struct PyFreezableInt {
```
bstaletic, line 32: Does this need to be implemented with the raw CPython API?

bstaletic (commented):
Sorry for the late reply. I can't see anything wrong with the changes to the headers, but the ~400 lines of `FreezableInt` look kinda annoying. I get that the whole point is the `is_immutable` flag.

YannickJadoul:
> (except for the newly added `test_freezable_type_caster.cpp` that I haven't yet looked at)

Additionally, the only thing stopping `test_freezable_type_caster` from compiling/passing is the `static_assert`, which seems to indicate to me that something's the matter with the `type_caster`?

So either the tests aren't thorough enough, or your patch includes unnecessary changes?

I'm also not a huge fan of the extra, long `test_freezable_type_caster` test. It seems to be testing something pybind11 explicitly states isn't currently supported (i.e., `const` correctness while casting)? I'm happy to see barriers removed to solve this problem, but including a test on this kind of unsupported workaround ... I'm not really convinced about that (since we'll also be the ones that need to maintain this).

tests/test_freezable_type_caster.cpp
```diff
@@ -0,0 +1,418 @@
...
+}
+
+/// pybind11 type_caster which returns custom PyFreezableInt python wrapper instead
+/// of a pybind11 generated type; this is used to make const* and const& immutable
+/// in python.
+namespace pybind11 {
+namespace detail {
+
+template <>
+struct type_caster<FreezableInt> {
```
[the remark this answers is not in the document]
laramiel: While we might, I think that is less clear and a worse example. so much of the base caster functionality is unused that any overlap with the base caster is incidental and basically limited to 2 cast overloads.

include/pybind11/cast.h
laramiel: Done. Indeed I missed that and it is no longer required.  From a C++ perspective they are equivalent, though.

tests/test_freezable_type_caster.cpp
```diff
@@ -0,0 +1,418 @@
...
+  m.def("get_const_ref", []() -> const FreezableInt& {
+    shared_const.x++;
+    return shared_const;
+  });
+  m.def("get_const_wrap", []() -> std::reference_wrapper<const FreezableInt> {
+    shared_const.x++;
+    return std::cref(shared_const);
+  });
+
+  m.def("roundtrip", [](FreezableInt c) -> FreezableInt {
```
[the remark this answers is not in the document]
laramiel: They validate the roundtrip yields the same object; basically just ensuring that the underlying reference map call works and are not specific to the type_caster itself.

tests/test_freezable_type_caster.cpp
laramiel: I want a freezable representation of c++ object available to python. I don't see a better option to achieve what I'm trying to achieve here.

laramiel:
> > (except for the newly added `test_freezable_type_caster.cpp` that I haven't yet looked at)
> 
> Additionally, the only thing stopping `test_freezable_type_caster` from compiling/passing is the `static_assert`, which seems to indicate to me that something's the matter with the `type_caster`?

Which `static_assert`, precisely? The one in the `type_caster<std::reference<T>>` specialization? That's the specific change that I made without which my test fails to build.

include/pybind11/cast.h
```diff
@@ -960,9 +960,15 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
 private:
     using caster_t = make_caster<type>;
     caster_t subcaster;
-    using subcaster_cast_op_type = typename caster_t::template cast_op_type<type>;
-    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value,
-            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
```
YannickJadoul, line 965: > > > (except for the newly added `test_freezable_type_caster.cpp` that I haven't yet looked at)
> > 
> > 
> > Additionally, the only thing stopping `test_freezable_type_caster` from compiling/passing is the `static_assert`, which seems to indicate to me that something's the matter with the `type_caster`?
> 
> Which `static_assert`, precisely? The one in the `type_caster<std::reference<T>>` specialization? That's the specific change that I made without which my test fails to build.

These 2 lines. If I'm not dreaming, all tests pass when I comment these out (in the original code, without your changes). At least on my Linux, Python 3.8, but this part of the casters should be independent from Python versions, I believe.

Moreover, all tests (except for `test_freezable_type_caster`) pass with the original code.

tests/test_freezable_type_caster.cpp
YannickJadoul: You could make some kind of a wrapper `FreezableIntWrapper`, expose that as `class_<FreezableIntWrapper>`, and have a type-caster specialization for `FreezableInt`, which uses the (auto-generated by pybind11) `FreezableIntWrapper`.

I think it will save you a lot of overhead in code (at the cost of being maybe slightly less efficient). I currently don't have the time to dig into it, but maybe over the next days.

include/pybind11/cast.h
laramiel: Exactly the point of the change.


1/ For std::reference<T>, the original implementation determines operator to call via the equivalent of `cast_op_type = make_type_caster<T>::cast_op<std::decay<T>>` and then invokes `operator cast_op_type&()`. Basically removing the reference and the const before getting the cast op, then adding the reference back in.

The original static_assert ensures that there is such an operator as `cast_op_type&` on `make_type_caster<T>`.


2/ The updated implementation determines the operator to call via the equivalent of `cast_op_type = make_type_caster<T>::cast_op<std::add_reference<T>>` and then invokes `operator cast_op_type()`. 
It adds the reference to the desired type before invoking cast_op, and then uses that operator without rewriting.

The static assert is the equivalent, requiring that the cast operator is one of either `T&` or `const T&`. The new behavior allows cast_op<> to select which operator overload to call instead of manipulating it afterwards, and the static_assert requires that the behavior is either the same as before (T&), or allows it to be const (const T&).

It's good that all the original tests pass with the original code; the fact that the added test fails also indicates that the code works as desired.

tests/test_freezable_type_caster.cpp
rwgk: > You could make some kind of a wrapper `FreezableIntWrapper`, expose that as `class_<FreezableIntWrapper>`, and have a type-caster specialization for `FreezableInt`, which uses the (auto-generated by pybind11) `FreezableIntWrapper`.
> 
> I think it will save you a lot of overhead in code (at the cost of being maybe slightly less efficient). I currently don't have the time to dig into it, but maybe over the next days.

The new test as-is has many lines of code, but it is as straightforward as it can possibly be and is 100% representative of how real-world non-pybind11-wrapped types interact with pybind11. Your suggested alternative will be more compact, but lose those two virtues. My suggestion is to respect Laramie's decision to go this route and not spend any more time working on the test. It seems very unlikely that the extra investment will ever pay off, via reduced maintenance cost. Adjustments for future changes in the Python C API are likely to be tiny. Maintaining some fabricated wrapper code could well have a higher maintenance cost, in part simply because it's less straightforward for anyone to get their head around.

YannickJadoul: @rwgk, you said yourself we don't support this:

> Note that the loss of `const` isn't just limited to this situation. It existed in Boost.Python already, and is very difficult to avoid. (I wouldn't try.)

As I said, if there's a straightforward way to change pybind11 such that this weird scenario is supported, we can do it. But I strongly oppose adding and maintaining more raw C API than necessary, especially in a test (I believe that's what @bstaletic complained about as well?).
There ought to be a better way to test this, without the introduction of a whole new test file and 200 lines of raw C API?

include/pybind11/cast.h
YannickJadoul: > It's good that all the original tests pass with the original code; the fact that the added test fails also indicates that the code works as desired.

Nono, what I'm saying is that the new tests you added (except for `test_freezable_type_caster`) pass with the old code.
And `test_freezable_type_caster` also passes with the old code (ignoring the static_assert).

So either your tests are not demonstrating things well, or cannot see a good reason for your changes to `cast.h`. According to this experiment, this only thing that needs to be changed is a more relaxed `static_assert` check?

Apart from that, I need another fresh look tomorrow. I can see your 1/ and 2/, what's happening, but the way other casters are written, I'm far from convinced that that is the right way to go. `const` is just something that's currently not supported by pybind11 and would probably take a lot more work to support.

Meanwhile, do you have a single, simple example, that show that `std::reference_wrapper<T>` currently (on `master`) behaves differently from `T&`? (as stated in your title, "Allow type_caster of std::reference_wrapper<T> to behave like a native reference.")

tests/test_freezable_type_caster.cpp
rwgk: > @rwgk, you said yourself we don't support this:
> 
> > Note that the loss of `const` isn't just limited to this situation. It existed in Boost.Python already, and is very difficult to avoid. (I wouldn't try.)

What I meant: loss of `const` is a fundamental shortcoming, but also deep, therefore I wouldn't try to fix it at that fundamental level.

> As I said, if there's a straightforward way to change pybind11 such that this weird scenario is supported, we can do it.

The scenario may be unusual, but looking at this with a birds-eye view, it's weird to have e.g. `set` and `frozenset`. Laramie's test is reflective of the real-world situation he wrote this test for (related to Google's OSS protobuf library), which will implement the freezing logic at runtime. I think it's totally worth supporting that, with the tiny tweak he needs here.

Since the core change doesn't break any existing test, and it is tiny, would you accept it without additional test?

The only thing lost from the protobuf library viewpoint is a protection against regression, but that's a small loss (we'd have to upstream breakage reports, probably rare).

include/pybind11/cast.h
laramiel: No, you're right. The `static_assert` is the most significant change. The `cast_op<add_reference<T>>` part of the change only matters when you do different things with cast_op<T>& and cast_op<T&>. And making a reasonable test that does that is challenging.

As for my other claim regarding const, it appears that on re-reading the it was not completely correct. I was originally using `cast_op_type`, which is where `intrinsic_t<T>`, which removes const, was applied, and the const-removal was not native to reference_wrapper.

The test now demonstrates the failure case better because the before behavior invokes the cast_op specialization in a way that discards const when not passed as a reference or a pointer, and the roundtrip tests now fail with the older variant beyond just the static_assert failure.

laramiel:
Since there is concern about the test complexity, I've pulled it out.

rwgk (approved):
Hi @bstaletic & @YannickJadoul, could you please take another look?

This PR is now tiny, does not upset any existing unit tests, and passed the google-global testing. The change in this PR is a subtle desirable behavior change that just happens to stay under the radar of everything previously existing; the proof is in the removed new test.

We are maintaining the removed test internally, but will also copy it to a new draft PR after this PR is merged, in case someone is interested in working on condensing it. If we see breakages internally (seems unlikely), we will rebase the draft PR and trigger a new CI run, to surface the breakages publicly.

include/pybind11/cast.h
```diff
@@ -960,9 +960,15 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
...
+    using reference_t = typename std::add_lvalue_reference<type>::type;
+    using subcaster_cast_op_type =
+        typename caster_t::template cast_op_type<reference_t>;
+
+    static_assert((std::is_same<typename std::remove_const<type>::type &,
+                             subcaster_cast_op_type>::value ||
+                std::is_same<reference_t, subcaster_cast_op_type>::value),
+               "std::reference_wrapper<T> caster requires T to have a caster "
+               "with an operator `T &` or `const T&`");
```
YannickJadoul, line 971: Since this line isn't as long as this one below:

```
static handle cast(const std::reference_wrapper<type> &src, return_value_policy policy, handle parent) {
```

It's probably more clear if we put it like this? (also with correct indentation and with `const T &` instead of `const T&`, like the rest of pybind11's codebase)

```suggestion
    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value ||
                  std::is_same<reference_t, subcaster_cast_op_type>::value,
                  "std::reference_wrapper<T> caster requires T to have a caster with an "
                  "`operator T &()` or `operator const T &()`");
```

tests/test_builtin_casters.cpp
```diff
@@ -147,6 +147,17 @@ TEST_SUBMODULE(builtin_casters, m) {
...
+    m.def("refwrap_usertype_const", [](std::reference_wrapper<const UserType> p) { return p.get().value(); });
+
+    m.def("refwrap_lvalue", []() -> std::reference_wrapper<UserType> {
+        static UserType x(1);
+        return std::ref(x);
+    });
+    m.def("refwrap_lvalue_const", []() -> std::reference_wrapper<const UserType> {
+        static UserType x(1);
+        return std::cref(x);
+    });
```
YannickJadoul, line 159: This is related to `type_caster<std::reference_wrapper<T>>::cast`, which isn't touched by this PR? Why add these tests?

include/pybind11/cast.h
```diff
@@ -960,9 +960,15 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
 private:
     using caster_t = make_caster<type>;
     caster_t subcaster;
-    using subcaster_cast_op_type = typename caster_t::template cast_op_type<type>;
-    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value,
-            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
+    using reference_t = typename std::add_lvalue_reference<type>::type;
```
YannickJadoul, line 963: Why not just `type&`? Apparently the only difference is `void`, but I don't think `std::reference_wrapper<void>` is supported, or is it?

YannickJadoul (commented):
OK, it took me a few hours to dig through the whole caster construction (together with @bstaletic), but here I am with another review.

I think one of the things that was confusing to me is that these are 3 independent changes in one PR:

1. Changing the subcaster's `cast_op_type<type>` to `cast_op_type<type&>`

  OK, so you're asking the subcaster how to cast to a `type&` instead of to a `type`. The reason this is the case is that in almost all cases in pybind11, these two types would get the same conversion operator.
  Do you have a good example of when these would be different? This might be good to add in a test (see later remark about tests).
  I think it's OK to change though, since indeed, you actually want a `type&` from the subcaster.

2. Loosening the `static_cast` that does currently not allow for `subcaster::operator const type&()`.

   Fair enough, yes. The repetition of explicitly listing `const` seems a bit ugly, but the `std::remove_const` was already there, so ...
   An alternative would be something like `std::is_convertible<subcaster_cast_op_type, std::reference_wrapper<type>>::value`? But fine to keep as it is now, I think.

3. The `&` in `subcaster.operator subcaster_cast_op_type&()` was completely superfluous already before this PR, I believe (remember we just asserted that it was already an lvalue reference!). So it's a correct change, but independent from this PR.


Now, as I said before: what bothers me a bit is that tests are added, but these tests already didn't fail before the actual changes of this PR. Normally, tests added in a PR would fail before the function changes of the PR?
So should we still add a minimal (!) test with a  `type_caster` that has an `operator const Foo &()`  that's different from `operator Foo &()`? And a `type_caster` that differentiates its `cast_op_type<Bar&>` from `cast_op_type<Bar>`? (do you have a good example of such a case, perhaps; in a smaller version than the full `FreezableInt` file?)

I'd also like to hear from @bstaletic if this summary and assessment makes sense, before officially approving.

include/pybind11/cast.h
YannickJadoul: To be more specific, `type_caster<std::reference_wrapper<void>>` would already result in a compilation error a few lines further, to evaluate `std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value`: `typename std::remove_const<void>::type` is still `void`, and would also result in `void &`.

include/pybind11/cast.h
```diff
@@ -973,7 +979,7 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
         return caster_t::cast(&src.get(), policy, parent);
     }
     template <typename T> using cast_op_type = std::reference_wrapper<type>;
-    operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type&(); }
+    operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type(); }
```
YannickJadoul, line 982:
```suggestion
    operator std::reference_wrapper<type>() { return cast_op<reference_t>(subcaster); }
```

or even (depending on the other discussion)

```suggestion
    operator std::reference_wrapper<type>() { return cast_op<type&>(subcaster); }
```

YannickJadoul:
> So should we still add a minimal (!) test with a `type_caster` that has an `operator const Foo &()` that's different from `operator Foo &()`? And a `type_caster` that differentiates its `cast_op_type<Bar&>` from `cast_op_type<Bar>`?

I'm playing around with this kind of minimal test, currently, but it's getting too late. I'll get back to it tomorrow or later this week.

```c++
struct ConstRefCasted {
    bool is_const;
    bool is_ref;
};

PYBIND11_NAMESPACE_BEGIN(pybind11)
template<>
class type_caster<ConstRefCasted> {
public:
    static constexpr auto name = "ConstRefCasted";

    bool load(handle, bool) { return true; }

    operator ConstRefCasted() { value = {false, false}; return value; }
    operator const ConstRefCasted() { value = {true, false}; return value; }
    operator ConstRefCasted &() { value = {false, true}; return value; }
    operator const ConstRefCasted &() { value = {true, true}; return value; }

    template <typename T_> using cast_op_type = T_;

private:
    ConstRefCasted value = {false, false};
};
PYBIND11_NAMESPACE_END(pybind11)
```

include/pybind11/cast.h
laramiel: So you already listed point 3/ where I removed a & based on review feedback as a "separate change"; how is this not the same?

tests/test_builtin_casters.cpp
laramiel: Because I saw no test returning a `std::reference_wrapper` at all and I didn't want things to break. I think that it's reasonable to add.

YannickJadoul: True, but it's unclear that it's not related to the functional changes in this PR.

YannickJadoul: (or at least, it was to me)

include/pybind11/cast.h
YannickJadoul: What do you mean? I don't mean this should be taken out, but if you go and change `std::add_lvalue_reference<type>` to `type &`, this part can also be simplified?

YannickJadoul: If you don't agree with the suggestion in https://github.com/pybind/pybind11/pull/2705#discussion_r542041783, that's also fine with me, then just argue why not and do not do it!

tests/test_const_ref_caster.cpp
```diff
@@ -0,0 +1,88 @@
+/*
+    tests/test_const_ref_caster.cpp
```
bstaletic, line 2: These tests seem more appropriate in `test_builtin_casters.cpp`/`test_builtin_casters.py`.

tests/test_builtin_casters.cpp
bstaletic: I'd say keep the tests. They aren't hurting anything.

include/pybind11/cast.h
laramiel: I mean `operator std::reference_wrapper<type>() { return cast_op<type &>(subcaster); }` is clearly a more substantive change than removing `&` from `operator std::reference_wrapper<type>() { return subcaster.operator subcaster_cast_op_type(); }`

tests/test_const_ref_caster.cpp
laramiel: Perhaps, but I think that since they are intended to exercise different aspects, putting it into a separate file makes the intent a bit more clear.

include/pybind11/cast.h
laramiel: Done.

include/pybind11/cast.h
laramiel: Done.

include/pybind11/cast.h
YannickJadoul: True, yes. I just saw this being used in other casters.

YannickJadoul: I'm also fine with not applying this change; it's supposed to do exactly the same, though.

tests/test_const_ref_caster.cpp
YannickJadoul: Yes, pybind11's tests are a mess. But I think the idea it to keep a minimal amount of files, because they all take quite some time to compile. I also don't think this one test is enough to create a separate file for.

Meanwhile, `type_caster<std::reference_wrapper<T>>` is already tested in `test_builtin_casters.cpp` and other test files already have custom casters as well (`test_custom_type_casters.cpp`, `test_copy_move.cpp`, and even one in `pybind11_tests.h`). So if we can keep the scope of the caster focused enough, I think it's fine.

laramiel: The other counter point is that this requires a custom type caster; so maybe move them to custom_type_casters.cc?  I think that the current test is about as minimal as it can be while exercising the various paths we want to test.

rwgk: Just logging my view, not a request: it makes me sad to see this test lumped in with other things. I know the existing tests do that a lot, but I think it really isn't a good approach, because it convolutes unrelated things, makes them harder to understand/follow, and obviously is counter the the idea of testing small self-contained units. Sticking with the established pattern only has so much value. Nudging things in a better direction has value, too. My preference is definitely to keep this test separate, if Boris or Yannick can agree.

include/pybind11/cast.h
laramiel: Done.

tests/test_const_ref_caster.cpp
laramiel: Anyway, Done.

laramiel: I agree with @rwgk however I will change it to the reviewers consensus.

rwgk:
The new test is an awesome demo for how `cast_op` works! It's really helping me understanding the caster system better. Another argument for not burring it in a bigger thing, because it'll be harder to work/experiment with.

For myself I changed it around like this:

```
struct ConstRefCasted {                                                         
  int unique_tag;                                                               
};                                                                              
```

```
  operator ConstRefCasted&&() { value = {1}; return std::move(value); }         
  operator ConstRefCasted&() { value = {2}; return value; }                     
  operator ConstRefCasted*() { value = {3}; return &value; }                    
                                                                                
  operator const ConstRefCasted&() { value = {4}; return value; }               
  operator const ConstRefCasted*() { value = {5}; return &value; }              
```

```
  m.def("takes", [](ConstRefCasted x) {                                         
    return x.unique_tag;                                                        
  });
  ...etc...                                                                           
```

```
    utw = m.ConstRefCasted()                                                    
                                                                                
    assert m.takes(utw) == 1                                                    
                                                                                
    assert m.takes_ptr(utw) == 3                                                
    assert m.takes_ref(utw) == 2                                                
    assert m.takes_ref_wrap(utw) == 2                                           
                                                                                
    assert m.takes_const_ptr(utw) == 5                                          
    assert m.takes_const_ref(utw) == 4                                          
    assert m.takes_const_ref_wrap(utw) == 4                                     
```

I think that makes it even easier to understand.
The name of the type could be `UniqueTagWrapper`.

tests/test_const_ref_caster.cpp
YannickJadoul: > The other counter point is that this requires a custom type caster; so maybe move them to custom_type_casters.cc? I think that the current test is about as minimal as it can be while exercising the various paths we want to test.

Well, coming from this PR, it's testing the built-in caster for `std::reference_wrapper`, so I thought it would be best to keep this as close to the other tests for `reference_wrapper`, and to not create yet another file is "something with casters" (which takes time to compile, each time!).

> move them to custom_type_casters.cc

It's not really about the custom caster, though. It's about testing `reference_wrapper`, currently, I though? If this maybe gets expanded further in the future, the caster could be shared in a header, though, when it's necessary.

> it makes me sad to see this test lumped in with other things.

Why exactly? Because I had explicitly tried to design/shrink this to be as minimal, unit-testy as possible, such that it could fit in with the rest of the `std::reference_wrapper` tests (that were even extended) in `builtin_casters.cpp`.

And to me it's more confusing to break convention than to add a few lines to the existing tests.

I do agree the tests are messy, btw, but I don't see the urgency of splitting things up. Build times are (partially) to blame, I think. So maybe once #2445 gets merged, it will be more feasible to split up tests, in a separate PR that brings order to the tests in a less ad-hoc fashion?

> I agree with @rwgk however I will change it to the reviewers consensus.

Not sure there's consensus, though. If you want, we can give @henryiii a decisive vote, but ... well, yeah, I've explained my reasoning :-)

YannickJadoul:
> I think that makes it even easier to understand.

Huh, OK. This was my first idea, but then I thought "let's make things more descriptive". Wasn't `is_const` and `is_ref` more descriptive of what's being tested, as you usually prefer? But I'm fine with the integer tags, if you prefer.

> The name of the type could be `UniqueTagWrapper`.

I'm just not convinced by this. That name describes the internals, not what's being tested by it?
Also, the class is not wrapping anything or any unique tag; only the caster knows about tags.

laramiel:
I think that it's easier to see that each cast path has a unique result, and the numeric is easier to map back than a pair would be.

YannickJadoul:
> I think that it's easier to see that each cast path has a unique result, and the numeric is easier to map back than a pair would be.

That's true, yes, I like that argument!
I just found it funny, because I had opted for the longer, more expressive way of doing things over the first idea of , among other reasons, because I though @rwgk would prefer it and make a good point on how that's better than integers :-) Bad intuition, it seems.
But this is definitely fine, as well! I think there's also other tests doing this, like `test_stl.py::test_function_with_string_and_vector_string_arg`.

include/pybind11/cast.h
```diff
@@ -960,9 +960,14 @@ template <typename type> class type_caster<std::reference_wrapper<type>> {
 private:
     using caster_t = make_caster<type>;
     caster_t subcaster;
-    using subcaster_cast_op_type = typename caster_t::template cast_op_type<type>;
-    static_assert(std::is_same<typename std::remove_const<type>::type &, subcaster_cast_op_type>::value,
-            "std::reference_wrapper<T> caster requires T to have a caster with an `T &` operator");
+    using reference_t = type&;
```
YannickJadoul, line 963: This could still be collapsed where it's used, to be shorter and more transparent (which was why I suggested `type &` instead of `add_lvalue_reference<type>::type`). But if that's the only thing left to change, I'm fine with this as well, and you don't need to add yet another commits, AFAIC.

YannickJadoul (approved):
Oh, I forgot to approve, it seems.
Thanks for the PR, for holding on during the review, and for fixing all the bikeshedding comments, @laramiel!

YannickJadoul:
Regarding CI: I restarted a few times, but that didn't help. No clue why https://developer.download.nvidia.com/hpc-sdk/nvhpc-20-7-20.7-1.x86_64.rpm is unreachable, and PyPy 2 has been failing for a week or so now (also ignore; #2724 is dealing with disabling PyPy 2 in CI for now).

henryiii:
Please assign a milestone, either 2.6.2 or 2.7.0. Thanks. :)

YannickJadoul:
Is this a fix? I guess it could be considered one. I'm happy to add this into 2.6.2; this shouldn't really be affecting existing code, I would hope.

rwgk:
I'll take care of the change log.

YannickJadoul:
> I'll take care of the change log.

Do we need a changelog? In principle these should be internal changes, that don't immediately concern users. @henryiii?

PR #2716, opened 2020-12-08
fix: regression with installed pybind11 overriding local one

### LINKED ISSUE #2709: [BUG] regression when mixing installed pybind11 with local copy
## Issue description

Hi all, since the recent refactor of CMake files (pull request #2370), when setting up include paths, `pybind11_add_module` puts priority on the already installed version of `pybind11 (PYTHON_INCLUDE_DIRS)` instead of the local copy (`pybind11_INCLUDE_DIR`). As an effect, wrong header files are included.

I understand that the documentation of pybind11_add_module says that pybind11 must be properly installed for it to work, but this nevertheless looks like a bug -- why add `pybind11_INCLUDE_DIR` if you expect it to be ignored?

Any idea how to get the old behavior? I found a hacky solution by modifying pybind11_add_module to do

```python
target_link_libraries(${target_name} PRIVATE pybind11::headers pybind11::module)
```
instead of

```python
target_link_libraries(${target_name} PRIVATE pybind11::module)
```

but I hope to find something better. Thanks!

From @ikicic on Gitter.

henryiii:
## Description

This was not always including the discovered pybind11 first. Closes #2709.

## Suggested changelog entry:

```rst
* CMake: mixing local and installed pybind11's would prioritize the installed one over the local one (regression in 2.6.0)
```

### DIFF
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -127,10 +127,20 @@ endif()
 # Check on every access - since Python2 and Python3 could have been used - do nothing in that case.
 
 if(DEFINED ${_Python}_INCLUDE_DIRS)
+  # Only add Python for build - must be added during the import for config
+  # since it has to be re-discovered.
+  #
+  # This needs to be an target to it is included after the local pybind11
+  # directory, just in case there are multiple versions of pybind11, we want
+  # the one we expect.
+  add_library(pybind11::python_headers INTERFACE IMPORTED)
+  set_property(
+    TARGET pybind11::python_headers PROPERTY INTERFACE_INCLUDE_DIRECTORIES
+                                             "$<BUILD_INTERFACE:${${_Python}_INCLUDE_DIRS}>")
   set_property(
     TARGET pybind11::pybind11
     APPEND
-    PROPERTY INTERFACE_INCLUDE_DIRECTORIES $<BUILD_INTERFACE:${${_Python}_INCLUDE_DIRS}>)
+    PROPERTY INTERFACE_LINK_LIBRARIES pybind11::python_headers)
   set(pybind11_INCLUDE_DIRS
       "${pybind11_INCLUDE_DIR}" "${${_Python}_INCLUDE_DIRS}"
       CACHE INTERNAL "Directories where pybind11 and possibly Python headers are located")
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -130,9 +130,9 @@ if(DEFINED ${_Python}_INCLUDE_DIRS)
   # Only add Python for build - must be added during the import for config
   # since it has to be re-discovered.
   #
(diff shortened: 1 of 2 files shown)


tools/pybind11NewTools.cmake
```diff
@@ -127,10 +127,20 @@ endif()
...
+  # directory, just in case there are multiple versions of pybind11, we want
+  # the one we expect.
+  add_library(pybind11::python_headers INTERFACE IMPORTED)
+  set_property(
+    TARGET pybind11::python_headers PROPERTY INTERFACE_INCLUDE_DIRECTORIES
+                                             "$<BUILD_INTERFACE:${${_Python}_INCLUDE_DIRS}>")
   set_property(
     TARGET pybind11::pybind11
     APPEND
-    PROPERTY INTERFACE_INCLUDE_DIRECTORIES $<BUILD_INTERFACE:${${_Python}_INCLUDE_DIRS}>)
```
henryiii, line 133: May have fixed a possible issue if using new FindPython and if Python's install dir had a space in it.

ikicic:
@henryiii The patch works for me. Thanks!

henryiii:
Quick description of what this fixes and why:

Say I install pybind11 2.5.0. Knowing me, I probably did a `brew install pybind11`, but how it got installed doesn't really matter, as long as it sits in the same place that Python.h will be coming from. Then I grab a project that has a built-in pybind11 submodule using a recent master. It suddenly matters if I include Python's include directory before pybind11's, because there is a sneaky installed copy of pybind11 next to Python, so if I put pybind11's local directory after, I might as well not put it in at all, the compiler will always get the installed copy even though I never ask for it.

We always add the local include directory first in the CMakeLists, though, so what's the problem? Well, when CMake builds the targets, it is just combining properties. The way it ended up in 2.6.{0,1}, `pybind11::pybind11` had pybind11's local include included via a target, `pybind11::headers`, but Python's include via a the includes property - which always comes first over the linked targets property!

By also making `pybind11::python_headers` and then including it that way, it's now on even footing with `pybind11::headers`,  in the exact same linked targets property, and it comes after it, so the include order is correct again.

YannickJadoul (approved):
Ooof, that that order of includes matters is quite ... ugly, but nothing we can do about fundamentally.

Thanks a lot for the extra explanation, @henryiii! Surely this will also be useful towards the future.

tools/pybind11NewTools.cmake
```diff
@@ -127,10 +127,20 @@ endif()
 # Check on every access - since Python2 and Python3 could have been used - do nothing in that case.
 
 if(DEFINED ${_Python}_INCLUDE_DIRS)
+  # Only add Python for build - must be added during the import for config
+  # since it has to be re-discovered.
+  #
+  # This needs to be an target to it is included after the local pybind11
```
YannickJadoul, line 133: Something must have gone wrong rewriting, here

Issue #1908, opened 2019-09-04
Did you know that py::buffer_info::format has a different meaning on Windows?


jpivarski:
On Windows (both 32-bit and 64-bit Python),

```c++
std::string f(py::array a) {
    py::buffer_info info = a.request();
    return info.format;
}
```

returns `"l"` for a `numpy.int32` array and `"q"` for a `numpy.int64` array, whereas MacOS and Linux return `"i"` and `"l"`, respectively. (That is, `"l"` is ambiguous.)

To be safe against misinterpretations, I'm using

```c++
auto b = a.cast<py::array_t<int64_t, py::array::c_style | py::array::forcecast>>();
```

to convert whatever I have into `int64_t` (and similarly for `int32_t`), if necessary. (Then, choosing `int32_t` vs `int64_t` based on platform is just an optimization, not needed for correctness.)

jpivarski:
The same is true on a 32-bit Docker image (`i386/ubuntu`):

   * `numpy.int32` → `format == "l"`
   * `numpy.int64` → `format == "q"`

whereas on 64-bit Linux:

   * `numpy.int32` → `format == "i"`
   * `numpy.int64` → `format == "l"`

Windows, regardless of whether it's 32-bit or 64-bit, take the same meanings as 32-bit Linux.

jpivarski:
Idioms like the following are _successful_ on 32-bit Linux, 64-bit Linux, 32-bit Windows, and 64-bit Windows (and 64-bit MacOS; haven't tested 32-bit MacOS). It's possible that the `__i386__` macro is only defined on gcc, so that's another thing I'll have to consider.

```c++
    if (format_.compare("d") == 0) {
      out = std::make_shared<PrimitiveType>(parameters_, PrimitiveType::float64);
    }
    else if (format_.compare("f") == 0) {
      out = std::make_shared<PrimitiveType>(parameters_, PrimitiveType::float32);
    }
#if defined _MSC_VER || defined __i386__
    else if (format_.compare("q") == 0) {
#else
    else if (format_.compare("l") == 0) {
#endif
      out = std::make_shared<PrimitiveType>(parameters_, PrimitiveType::int64);
    }
#if defined _MSC_VER || defined __i386__
    else if (format_.compare("Q") == 0) {
#else
    else if (format_.compare("L") == 0) {
#endif
      out = std::make_shared<PrimitiveType>(parameters_, PrimitiveType::uint64);
    }
#if defined _MSC_VER || defined __i386__
[...]

PR #2720, opened 2020-12-08
pybind11/numpy.h does not require numpy at build time.

anntzer:
## Description

The lack of build-time dependency on numpy is nice enough to be mentioned explicitly in the docs.  (See https://github.com/pybind/pybind11/issues/1908#issuecomment-666854391 for why.)

### DIFF
--- a/docs/advanced/pycpp/numpy.rst
+++ b/docs/advanced/pycpp/numpy.rst
@@ -150,8 +150,10 @@ NumPy array containing double precision values.
 
 When it is invoked with a different type (e.g. an integer or a list of
 integers), the binding code will attempt to cast the input into a NumPy array
-of the requested type. Note that this feature requires the
-:file:`pybind11/numpy.h` header to be included.
+of the requested type. This feature requires the :file:`pybind11/numpy.h`
+header to be included. Note that :file:`pybind11/numpy.h` does not depend on
+the NumPy headers, and thus can be used without declaring a build-time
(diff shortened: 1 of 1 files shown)


bstaletic (approved):
Thanks for the pull request.

YannickJadoul (approved):
Nice! I think there's a minimum version of NumPy, though (1.7.0 or so?; I forgot, but maybe someone knows; it's based on `numpy.core.multiarray._ARRAY_API`). Could be worth mentioning that, or not? It's a pretty old version, though, so maybe not useful.

Nitpicky, I'd suggest to remove the parenthesis around a full sentence, but I'm also find to keep them.

YannickJadoul:
> Could be worth mentioning that, or not?

It's 1.7.0:
https://github.com/pybind/pybind11/blob/028812ae7eee307dca5f8f69d467af7b92cc41c8/include/pybind11/numpy.h#L238-L239

https://pypi.org/project/numpy/1.7.0/

> Released: Feb 12, 2013

We can safely ignore that minimum version, it seems :-)

anntzer:
reworded

YannickJadoul:
That's great, AFAIC! Thanks, @anntzer! :-)

henryiii (approved):
CentOS 8 looks rather like it ran out of memory. And we have a known issue with PyPy2 since the 7.3.3 update. Looks fine to me. :)

PR #2724, opened 2020-12-11
ci: drop pypy2 linux, PGI 20.7, add Python 10 dev

henryiii:
## Description

Drop the pypy2 build on Linux until we solve the issue with 7.3.3. Adds Python 3.10 development testing, drops a few of the middle Python tests (3.7 and 3.8) assuming the we cover things well enough with 2.7, 3.5, 3.6 (with different settings) and 3.9. (Prompted by an environmental discussion on Gitter, we could keep them if preferred).

Dropping NVIDIA PGI 20.7 since it's currently missing. Tried to add 20.11, but it has an internal compiler error.

## Suggested changelog entry:

None needed, CI only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,44 +242,48 @@ jobs:
       run: cmake --build build --target pytest
 
 
-  # Testing CentOS 8 + PGI compilers
-  centos-nvhpc8:
-    runs-on: ubuntu-latest
-    name: "🐍 3 • CentOS8 / PGI 20.7 • x64"
-    container: centos:8
-
-    steps:
-    - uses: actions/checkout@v2
-
-    - name: Add Python 3 and a few requirements
-      run: yum update -y && yum install -y git python3-devel python3-numpy python3-pytest make environment-modules
-
-    - name: Install CMake with pip
-      run: |
-        python3 -m pip install --upgrade pip
-        python3 -m pip install cmake --prefer-binary
-
-    - name: Install NVidia HPC SDK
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -250,44 +242,48 @@ jobs:
...
-    - name: Add Python 3 and a few requirements
-      run: yum update -y && yum install -y git python3-devel python3-numpy python3-pytest make environment-modules
-
-    - name: Install CMake with pip
-      run: |
-        python3 -m pip install --upgrade pip
-        python3 -m pip install cmake --prefer-binary
-
-    - name: Install NVidia HPC SDK
-      run: yum -y install https://developer.download.nvidia.com/hpc-sdk/nvhpc-20-7-20.7-1.x86_64.rpm https://developer.download.nvidia.com/hpc-sdk/nvhpc-2020-20.7-1.x86_64.rpm
```
YannickJadoul, line 271: Seems the new links are these two:

- https://developer.download.nvidia.com/hpc-sdk/20.7/nvhpc-20-7-20.7-1.x86_64.rpm
- https://developer.download.nvidia.com/hpc-sdk/20.7/nvhpc-2020-20.7-1.x86_64.rpm

But we can drop them as well, if you want

henryiii: Ah, so it did move. I have a feeling, if they release every two months, that older images aren't all that important. It was more important to have CentOS 7 and 8, since they had different workarounds. I'm hoping we can get 20.11 in in the not too distant future, and if not, we can restore 20.7 with the corrected links.

YannickJadoul: Yes, sounds good to me!

PR #2725, opened 2020-12-13
Implementing keep_alive with placement

lnstadrum:
## Description

I am possibly doing things wrong or come up with an unnecessarily complex approach. If there is an existing simple solution to my problem, please point me out.

### Problem

Consider the following Python code using a pybind11 module `module` having `Processor` and `Source` classes in it.

```py
processor = module.Processor()

# load first source to process
processor.input = module.Source("large_file1.avi")
processor.process()

# load second source to process
processor.input = module.Source("large_file2.avi")
processor.process()
```

`processor.input` is a property with a usual setter and getter bound to it. A `Processor` instance makes use of `input` property value (a `Source` instance) to do some sophisticated processing in `process()` method (so the value needs to be protected against being garbage-collected). Then the property value changed with another `Source` instance for a second processing iteration.

The only easy solution I found to make this code running without segfaults is to decorate the property setter with `keep_alive<1, 2>()`, making sure the `Source` instance is not destroyed while the `Processor` instance is alive and might be using it.

The problem with this solution arises when the processing is done in a loop, and there are many more `Source` instances potentially taking large memory. As far as only one instance of `Source` is actually used at a time, it makes sense to free the previously hold `input` value as soon as it becomes unused.

### Suggested solution

I tried different things like implementing some wrappers or changing the backend code. I run into different troubles that I will not detail right away. Then I ended up with implementing the keep_alive with placement feature in pybind11 code  which worked and required minimal overall changes.

This feature consists in assigning a "placement index" parameter to a Nurse-Patient link in `keep_alive` attribute. When a bound method with this attribute is called, a new Patient is not added unconditionnally as with regular `keep_alive`, but it replaces a previously hold Patient at the same Placement (if there is such). The placement index is an arbitrary user-defined non-negative integer number.

An explanation is added in the docs. A unit test is included.

Changes overview.
 * `keep_alive<Nurse, Patient, Placement>` is introduced.
 * A map holding patients with placements is added internally. `add_patient` and `clear_patients` methods upgraded accordingly.

Other notes.
  - The original `keep_alive` without placement is indeed maintained.
  - I am not a big fan of having three integer indices passed as template arguments. I tried to implement something like `keep_alive<Nurse, Patient>(Placement)`, but this seems to require more changes of other internal entities.
  - Placement can in principle be inferred automatically for object-valued properties as in the example. However, I found useful the ability to have the same placement in different methods, for example, for overloaded setters that operate on the same backend property.

## Suggested changelog entry:

```rst
Introducing ``keep_alive<Nurse, Patient, Placement>`` attribute
```

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -212,6 +212,28 @@ ties the lifetime of the constructor element to the constructed object:
     Patient != 0) and ``with_custodian_and_ward_postcall`` (if Nurse/Patient ==
     0) policies from Boost.Python.
 
+Keep alive with placement
+-------------------------
+
+When ``keep_alive<Nurse, Patient>`` is supplied in a method binding, the Patient
+object is kept alive unconditionally until the Nurse is destroyed. This implies
+that every call of the method adds a possibly new Patient. If the method is called
+in a loop, the number of Patients keeps growing and their underlying resources are
+not freed until the Nurse object is garbage-collected.
+
+``keep_alive<Nurse, Patient, Placement>`` adds a Patient and attributes a positive
+integer index to it (Placement). Whenever a method is called for the same Nurse and
+Placement, the reference to a previously hold Patient is removed (which causes its
+destruction if it is not referred elsewhere), and newly passed Patient takes its
+place.
+
+The same Placement index may be used across multiple methods. It is allowed to pass
+``None`` as patients to explicitly clear the Placement.
+
+This technique is convenient for binding regular setter methods that replace a
+referred resource with another resource and allows to keep the reference records
+up-to-date.
+
 Call guard
 ----------
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -66,7 +66,7 @@ struct base {
 };
 
 /// Keep patient alive while nurse lives
-template <size_t Nurse, size_t Patient>
+template <size_t Nurse, size_t Patient, size_t Placement = detail::KEEP_ALIVE_NO_PLACEMENT>
 struct keep_alive {};
 
 /// Annotation indicating that a class is involved in a multiple inheritance relationship
@@ -169,7 +169,7 @@ enum op_type : int;
 struct undefined_t;
 template <op_id id, op_type ot, typename L = undefined_t, typename R = undefined_t>
 struct op_;
-void keep_alive_impl(size_t Nurse, size_t Patient, function_call &call, handle ret);
+void keep_alive_impl(size_t Nurse, size_t Patient, size_t Placement, function_call &call, handle ret);
 
 /// Internal data structure which holds metadata about a keyword argument
 struct argument_record {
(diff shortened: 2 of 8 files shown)


bstaletic:
Hi. This looks like an oddly designed API, which might be the cause of your troubles. Why not something like this:

    processor.process(module.Source("large_file1.avi"))

Then it would be the responsibility of `process()` to ensure lifetime of input.

lnstadrum:
> Hi. This looks like an oddly designed API, which might be the cause of your troubles. Why not something like this:
> 
> ```
> processor.process(module.Source("large_file1.avi"))
> ```
> 
> Then it would be the responsibility of `process()` to ensure lifetime of input.

Hi,

If the source is only used within `process()` and can be freed right after, this would work indeed.

There are many other situations where it would not, as well as there are good reasons why the Source is scoped as a property of Processor and is not just a value required within a single call. There might be other API entries in Processor that require a Source, or `process()` may submit the source for an asynchronous processing inside the C++ code and quit right after, etc. In addition, the ability to convert the C++ API which has its setters and getters (and there are reasons for) into a similar API in Python without changing signatures of mapped functions and scopes of passed values is an ultimately nice thing to have.

I believe the problem is more general: object-valued writable properties in pybind11 are unusable without putting the hard keep_alive link or making a copy, and these two options are too constraining. (Again, if I miss a simple solution, please provide a pointer.)

YannickJadoul:
I'm not really a huge fan of the interaction between `keep_alive<Nurse, Patient, Placement>` and `keep_alive<Nurse, Patient>`. If done in the wrong order, one would/could override the other, no?

I agree this is kind of an issue, but aren't trampoline classes (where you add another `py::object` field to store "fixed" keep_alives) a solution?

lnstadrum:
> I'm not really a huge fan of the interaction between `keep_alive<Nurse, Patient, Placement>` and `keep_alive<Nurse, Patient>`. If done in the wrong order, one would/could override the other, no?
> 
> I agree this is kind of an issue, but aren't trampoline classes (where you add another `py::object` field to store "fixed" keep_alives) a solution?

Hi @YannickJadoul, 

In the current implementation, the regular `keep_alive` and the suggested `keep_alive` with placement use different containers to store the patient-nurse relationships and are thus uncorrelated. Therefore, `keep_alive<Nurse, Patient, Placement>` has practically no effect if `keep_alive<Nurse, Patient>` is present: it keeps incrementing / decrementing the refcount of the patient object when it changes, but the regular keep_alive ensures at least one existing reference until the nurse is deleted. This behaviour appears the most natural for me, making the combo of two different `keep_alive` kind of useless, but acting harmlessly and in the expected way.

As for the trampoline class, in the example given in the description it may solve the problem indeed. However, if I remember correctly, the blocking point with a trampoline-based option in my actual code was related to the fact that `input` property is carried by an abstract base of `Processor`: there are different processors naturally sharing a common interface containing the input source. The trick is then needed to be implemented in this abstract base class. I had an impression that the trampoline for the abstract base class is not instantiated. 

Generally, as far as I understand, for pybind11-registered classes `A`, `B` derived from `A`, and their respective trampoline classes `PyA` and `PyB`, when an object of `B` class is created, `PyA` is not instantiated. Therefore, if `input` is in `A`, the trampoline does not help. Please correct if I am wrong.

lnstadrum:
Hi,

Sorry for being stubborn and raising this old thing again, but I believe this feature is not only useful to me. From the discussion above I conclude there is no other solution to the situation I am facing, and yet it is not that uncommon or the API design is odd. Please let me know if there is another way of solving this.

Rebased on top of v2.8.1, fixed a few formatting issues.

lnstadrum

rwgk:
I looked at this for a good 30 minutes. I'm having trouble getting comfortable with this PR, similar to the concerns raised by @bstaletic and @YannickJadoul. My admittedly rough thoughts:

* The use case is an anti-pattern in my opinion. "side effect" (== "BAD") comes to mind.

* The changes are involved and introduce two systems for a similar purpose, quote: `keep_alive` and the suggested `keep_alive` with placement use different containers to store the patient-nurse relationships and are thus uncorrelated. "tacked on" comes to mind. Such an approach could easily become an obstacle to adopting a well thought out solution later.

I do think that keeping alive disassociated patients is not great. Could `shared_ptr` be leveraged here? That would also be a better C++ interface.

PR #2727, opened 2020-12-15
lower requirement for list_caster from sequence to iterable

HDembinski:
## Description

The list_caster requires at minimum a Python `Sequence`, while a Python `Iterable` also does nicely. I lowered the requirement, so now also generators can be converted and other collection types which do not implement the full Sequence protocol, notably `torch.Tensor`. If the iterable has a `__len__` method, it is used reserve space in a vector, so this implementation is as efficient as the old one.

I also simplified the TMP of reserve_maybe with modern TMP techniques.

## Suggested changelog entry:

```rst
The list_caster in ``pybind11/stl.h`` now accepts a Python ``Iterable`` (before it required a ``Sequence``). This allows casting Python generators and collection types that do not implement the Sequence protocol, like ``torch.Tensor``.
```

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -144,11 +144,11 @@ template <typename Type, typename Value> struct list_caster {
     using value_conv = make_caster<Value>;
 
     bool load(handle src, bool convert) {
-        if (!isinstance<sequence>(src) || isinstance<str>(src))
+        if (!isinstance<iterable>(src) || isinstance<str>(src))
             return false;
-        auto s = reinterpret_borrow<sequence>(src);
+        auto s = reinterpret_borrow<iterable>(src);
         value.clear();
-        reserve_maybe(s, &value);
+        reserve_maybe(s);
         for (auto it : s) {
             value_conv conv;
             if (!conv.load(it, convert))
@@ -159,10 +159,12 @@ template <typename Type, typename Value> struct list_caster {
     }
 
 private:
-    template <typename T = Type,
-              enable_if_t<std::is_same<decltype(std::declval<T>().reserve(0)), void>::value, int> = 0>
-    void reserve_maybe(sequence s, Type *) { value.reserve(s.size()); }
-    void reserve_maybe(sequence, void *) { }
+    template <typename T = Type>
+    void reserve_maybe(decltype(std::declval<T>().reserve(0), iterable{}) it) {
+      if (hasattr(it, "__len__"))
+          value.reserve(len(it));
+    }
+    void reserve_maybe(handle) { } // always a worse match, no ambiguity
 
 public:
     template <typename T>
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -13,6 +13,7 @@ def test_vector(doc):
     lst.append(2)
     assert m.load_vector(lst)
     assert m.load_vector(tuple(lst))
+    assert m.load_vector(x for x in (1, 2))  # not a sequence, but iterable
 
     assert m.cast_bool_vector() == [True, False]
     assert m.load_bool_vector([True, False])


include/pybind11/stl.h
```diff
@@ -159,10 +159,12 @@ template <typename Type, typename Value> struct list_caster {
...
 
 private:
-    template <typename T = Type,
-              enable_if_t<std::is_same<decltype(std::declval<T>().reserve(0)), void>::value, int> = 0>
-    void reserve_maybe(sequence s, Type *) { value.reserve(s.size()); }
-    void reserve_maybe(sequence, void *) { }
+    template <typename T = Type>
+    void reserve_maybe(decltype(std::declval<T>().reserve(0), iterable{}) it) {
+      if (hasattr(it, "__len__"))
+          value.reserve(len(it));
```
bstaletic, line 165: `hasattr(it, "__len__")` is very inefficient, considering that we know that we're dealing with a sequence. For this use case, CPython provides `PyObject_LengthHint()` since python 3.4. To save you the trouble of error checking, pybind11 provides `len_hint()`.

YannickJadoul:
Could you argue a little bit why this is necessary, and why any Python iterable should convert to an `std::vector` by default? I'm not saying I'm radically against this, but I'm also far from convinced this is a good idea. At the very least, it's not such a trivial change as this PR suggests.

For example, calling an overloaded function `foo(std::vector<int>)` and `foo(std::map<int, int>)` with a `dict` `{6 * 9: 42}` will change behavior, through this PR, I believe, as a `dict` is `iterable` in Python?

So at the very least, this should only be allowed when `convert == true`, but even then, an overload of `foo(std::vector<float>)` and `foo(std::map<float, float>)` will change behavior, if you pass `{6 * 9: 42}` (since `int` only gets converted to `float` when `convert`).

YannickJadoul:
> other collection types which do not implement the full Sequence protocol, notably `torch.Tensor`

Oh, I missed this part of your explanation. You would expect `torch.Tensor` to convert, I suppose? I don't really know. Though tensors are multi-dimensional, right? So to me it's not obvious they would convert to a 1D `std::vector`. I'd rather expect them to convert to `py::array` or so?

HDembinski:
> Could you argue a little bit why this is necessary, and why any Python iterable should convert to an `std::vector` by default? I'm not saying I'm radically against this, but I'm also far from convinced this is a good idea. At the very least, it's not such a trivial change as this PR suggests.

It is not by default, this caster is called only when you explicitly cast some Python object to `std::vector`, i.e. this code
```c++
auto vec = py::cast<std::vector<double>>(my_python_object);
```
You as the programmer decide to cast to a `std::vector<double>` here. If you have `stl.h` included, this then triggers the use of the list_caster. I did not change any of that, this is how pybind11 works.

This cast should succeed, when the cast is reasonable. In other words, if the Python object is a collection with an element type that matches `std::vector<double>`.

In case of `torch.Tensor`, it works only if the other tensor is 1D and has the right scalar type. Likewise for a `numpy.ndarray`. If these types have the wrong dtypes or the wrong dimension, this cast will fail with a runtime error message, as it did before. This is just more general than the previous implementation.

bstaletic:
> It is not by default, this caster is called only when you explicitly

Not true. It gets invoked implicitly. That's the whole point of `stl.h`.

https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html

YannickJadoul:
> It is not by default, this caster is called only when you explicitly cast some Python object to `std::vector`

So, as @bstaletic pointed out, that's not true, since all casting code is shared between `py::cast` and the casting for bound function signatures.

> You as the programmer decide to cast to a `std::vector<double>` here. If you have `stl.h` included, this then triggers the use of the list_caster. I did not change any of that, this is how pybind11 works.

Moreover, the issue with `stl.h` (and actually casters in general) is that you need to be consistent. You can't choose one specialization of `type_caster<T>` for binding one function/method, and another one for another. So `stl.h` needs to be reasonably conservative, as can't selectively opt-in.

> In case of `torch.Tensor`, it works only if the other tensor is 1D and has the right scalar type. Likewise for a `numpy.ndarray`.

So, that actually makes perfect sense to me? `std::vector` is a 1D list in C++. I'd be very surprised if an (N>1)D structure automatically gets casted by `list_caster` (which shows `List[T]` in the signature!) to a 1D version. For multiple dimensions, we have `py::array`/`py::array_t`.

Note that we do have `py::iterable` that you use in a signature to accept any iterable. And if you need to make the example that you gave work, I just suggest explicitly opting-in to this behavior. I.e., change

> ```c++
> auto vec = py::cast<std::vector<double>>(my_python_object);
> ```

to

> ```c++
> auto vec = py::cast<std::vector<double>>(py::list(my_python_object));
> ```

So given that this works (and similar for a function accepting `py::iterable`) and that the consequences of relaxing the requirement from `sequence` to `iterable` are pretty big (see https://github.com/pybind/pybind11/pull/2727#issuecomment-745513923), I'm getting more and more convinced I don't like the idea of this PR.

henryiii:
> `auto vec = py::cast<std::vector<double>>(py::list(my_python_object));`

This creates a new Python list in memory, and then copies it over, does it not? Probably using far more memory than `my_python_object` (if a tuple/array) and `std::vector<double>` combined?

There really are a lot of places in Python where any iterable is accepted, even if it doesn't make sense:

```python
a = []
a += {1:2}
assert a == [1]
```

What many libraries like NumPy do in Python is have a `from_iter()` function for arbitrary iterables, and restrict what the normal constructor takes.

The downside here is that if you have two overloads,

```c
f(std::vector<double> x); // .def(...)
f(py::array_t<double> x); // .def(...)
```

Before this change, the second function is matched if you pass an Iterable, while after this change the first one is matched. Are people doing this though? And, not considering backward compatibly, I really think I would expect an iterator of doubles to match `std::vector<double>()`, rather than relying on it being a Sequence, which seems a little arbitrary. A `std::vector  is itself a Sequence, not an Iterator, though, so I see that argument too.

My question is without this change, is there a simple way to make a function that accepts an Iterable but does not do extra needless copies?

bstaletic:
> This creates a new Python list in memory, and then copies it over, does it not?

I had the same question, but @YannickJadoul intended that example to be just about python generators. Not as a generic solution.

> The downside here is that if you have two overloads,

It's even worse for this overload set:

    m.def("f", [](std::vector<int>){}); // 1
    m.def("f", [](std::map<int, int>){}); // 2

and calling `f({1:2})` from python. It would be very surprising if this suddenly started to match `//1`.

> Are people doing this though?

They do. Not often, but we've seen these "ambiguous" overloads in bug reports, where people were confused about how pybind11 exactly does overload resolution.

> My question is without this change, is there a simple way to make a function that accepts an Iterable but does not do extra needless copies?

1. Take `py:iterable` as the argument.
2. If the argument is already a sequence, use `cast<T>` directly.
3. Otherwise, the iterable is some sort of generator. You can take @YannickJadoul's suggestion or write a loop and cast elements of the generator one by one.

HDembinski:
> and calling `f({1:2})` from python. It would be very surprising if this suddenly started to match `//1`.

Ok, fair enough, that kills it then. Will close.

YannickJadoul:
> What many libraries like NumPy do in Python is have a `from_iter()` function for arbitrary iterables, and restrict what the normal constructor takes.

Is that a no-go, in this case? I'm definitely not claiming you might not have a good use for it, but I'm rather terrified to add this to `stl.h`'s default behavior :-/

> This creates a new Python list in memory, and then copies it over, does it not? Probably using far more memory than `my_python_object` (if a Tensor/array) and `std::vector<double>` combined?

It does, yes. I hadn't thought of memory, but when it comes to runtime efficiency, you're already calling in and out of Python for every time, so you probably shouldn't do this on large data, anyway, and crossing the boundary from/to Python for every single element?

More efficiently but also more verbosely, doesn't `for (auto element : my_python_object) { my_vector.emplace_back(py::cast<T>(element)); }` work? The only disadvantage (apart from being more verbose) is that this won't check the next overload during overload resolution (because you accept `py::iterable`, and it's part of the implementation.

YannickJadoul:
I'm currently blaming the very rough `bool convert` argument to the caster, btw, rather than a more detailed order of conversions. Maybe casters should take different levels/precedence of conversions or so. But that will of course make the implementation a less efficient :-/

PR #2729, opened 2020-12-18
Intel ICC C++17 compatibility

### LINKED ISSUE #2707: [BUG] ICC cannot compile master and current release due to ICC's bug
## Issue description

Due to an ICC [bug](https://community.intel.com/t5/Intel-C-Compiler/Bug-report-icpc-fe-can-t-handle-fold-expressions-with-using/m-p/1233016#M38025), current usage of `pybind11::detail::all_of` in SFINAE-context will be rejected by ICC.

When compiling the code below using linux64 icc, the following error message will be omitted:

```
$ icpc -o test -std=c++17 -isystem $PYBIND11_ROOT/include -isystem /usr/include/python3.8 test.cpp
test.cpp(17): error: no instance of function template "foo" matches the argument list
            argument types are: (const char *)
        foo(x);
        ^
test.cpp(11): note: this candidate was rejected because at least one template argument could not be deduced
  void foo(Args &&...){
       ^

compilation aborted for test.cpp (code 2)
```

Tested with current master and latest release(2.6.0) and the problem exists.

This problem blocks usage of `object_api<Derived>::operator()` in `cast.h`, which eventually makes it unable to use expressions such as `.def(pybind11::init<T...>(pybind11::arg("foo")...)`.

## Reproducible example code

```c++
#include <iostream>
#include <type_traits>

#include <pybind11/pytypes.h>

using namespace pybind11;
using namespace pybind11::detail;

template <typename ...Args,
        typename = std::enable_if_t<all_of<is_positional<Args>...>::value>>
void foo(Args &&...){
        std::cout << "bbb" << std::endl;
}

int main(){
        const char *x = "aaa";
        foo(x);
        return 0;
}
```

mkuron:
Adds C++17 support.

* ICC can’t handle complex enable if’s inline, like `enable_if_t<all_of<is_positional<Args>...>::value>>`, but if it is refactored into a separate constexpr function, it works.
    - But MSVC doesn't like it if the function is in the return type.
* Fold expressions cause an internal compiler error, so we have to avoid turning them on even in C++17 mode
* Plain `noexcept` can cause weird errors with following syntax, `noexcept(true)` works fine, workaround in tests

This also fixes the NVIDIA HPC SDK compiler in C++17 mode.

## Description

The Intel compiler has trouble with pybind11 when compiling in C++17 mode. The first two examples in the code below result in errors like "no instance of function template [...] matches the argument list", while the third one triggers an internal compiler error.
```c++

void test1() {
  pybind11::print("Hello World");
}

void test2(pybind11::object obj) {
   obj();
}

class C {
};

void test3(pybind11::module_ &m) {
	pybind11::class_<C>(m, "C");
}
```
Details can be found in #2707. @tobiasleibner implemented the fix for the template mismatch, while the workaround for the internal compiler error is mine.  Part of @tobiasleibner's work went into #2573 already, which hasn't been merged yet. Since he currently doesn't have time, he [suggested](https://github.com/pybind/pybind11/issues/2707#issuecomment-748009778) I post a pull request.

Related to #2573
Fixes #2714
Fixes #2707

## Suggested changelog entry:

```rst
* Support ICC and NVIDIA HPC SDK in C++17 mode
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -2172,16 +2172,26 @@ private:
     dict m_kwargs;
 };
 
+// [workaround(intel)] Separate function required here
+// We need to put this into a separate function because the Intel compiler
+// fails to compile enable_if_t<!all_of<is_positional<Args>...>::value>
+// (tested with ICC 2021.1 Beta 20200827).
+template <typename... Args>
+constexpr bool args_are_all_positional()
+{
+  return all_of<is_positional<Args>...>::value;
+}
+
 /// Collect only positional arguments for a Python function call
 template <return_value_policy policy, typename... Args,
-          typename = enable_if_t<all_of<is_positional<Args>...>::value>>
+          typename = enable_if_t<args_are_all_positional<Args...>()>>
 simple_collector<policy> collect_arguments(Args &&...args) {
     return simple_collector<policy>(std::forward<Args>(args)...);
 }
 
 /// Collect all arguments, including keywords and unpacking (only instantiated when needed)
 template <return_value_policy policy, typename... Args,
-          typename = enable_if_t<!all_of<is_positional<Args>...>::value>>
+          typename = enable_if_t<!args_are_all_positional<Args...>()>>
 unpacking_collector<policy> collect_arguments(Args &&...args) {
     // Following argument order rules for generalized unpacking according to PEP 448
     static_assert(
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -383,6 +383,8 @@ constexpr bool args_any_are_buffer() {
     return detail::any_of<std::is_same<Args, buffer_protocol>...>::value;
 }
 
+// [workaround(intel)] Separate function required here
+// [workaround(msvc)] Can't use constexpr bool in return type
 
 // Add the buffer interface to a vector
 template <typename Vector, typename Class_, typename... Args>
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -377,8 +377,16 @@ struct vector_has_data_and_format<Vector, enable_if_t<std::is_same<decltype(form
 
 // Add the buffer interface to a vector
 template <typename Vector, typename Class_, typename... Args>
+#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
+// [workaround(intel)] Separate function required here
+// Workaround as the Intel compiler does not compile the enable_if_t part below
+// (tested with icc (ICC) 2021.1 Beta 20200827)
+void vector_buffer(Class_& cl) {
+  if constexpr (detail::any_of<std::is_same<Args, buffer_protocol>...>::value) {
+#else
 enable_if_t<detail::any_of<std::is_same<Args, buffer_protocol>...>::value>
 vector_buffer(Class_& cl) {
+#endif
     using T = typename Vector::value_type;
 
     static_assert(vector_has_data_and_format<Vector>::value, "There is not an appropriate format descriptor for this vector");
@@ -413,10 +421,16 @@ vector_buffer(Class_& cl) {
     }));
 
     return;
+#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
+  }
+}
+#else
 }
 
 template <typename Vector, typename Class_, typename... Args>
 enable_if_t<!detail::any_of<std::is_same<Args, buffer_protocol>...>::value> vector_buffer(Class_&) {}
+
+#endif
 
 PYBIND11_NAMESPACE_END(detail)
 
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -375,18 +375,18 @@ struct vector_has_data_and_format : std::false_type {};
 template <typename Vector>
 struct vector_has_data_and_format<Vector, enable_if_t<std::is_same<decltype(format_descriptor<typename Vector::value_type>::format(), std::declval<Vector>().data()), typename Vector::value_type*>::value>> : std::true_type {};
 
-// Add the buffer interface to a vector
-template <typename Vector, typename Class_, typename... Args>
-#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
 // [workaround(intel)] Separate function required here
 // Workaround as the Intel compiler does not compile the enable_if_t part below
 // (tested with icc (ICC) 2021.1 Beta 20200827)
-void vector_buffer(Class_& cl) {
-  if constexpr (detail::any_of<std::is_same<Args, buffer_protocol>...>::value) {
-#else
-enable_if_t<detail::any_of<std::is_same<Args, buffer_protocol>...>::value>
+template <typename... Args>
+constexpr bool args_any_are_buffer() {
+    return detail::any_of<std::is_same<Args, buffer_protocol>...>::value;
+}
+
+// Add the buffer interface to a vector
+template <typename Vector, typename Class_, typename... Args>
+enable_if_t<args_any_are_buffer<Args...>()>
 vector_buffer(Class_& cl) {
-#endif
     using T = typename Vector::value_type;
 
     static_assert(vector_has_data_and_format<Vector>::value, "There is not an appropriate format descriptor for this vector");
@@ -421,16 +421,11 @@ vector_buffer(Class_& cl) {
     }));
 
     return;
-#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
-  }
-}
-#else
 }
 
 template <typename Vector, typename Class_, typename... Args>
-enable_if_t<!detail::any_of<std::is_same<Args, buffer_protocol>...>::value> vector_buffer(Class_&) {}
+enable_if_t<!args_any_are_buffer<Args...>()> vector_buffer(Class_&) {}
 
-#endif
 
 PYBIND11_NAMESPACE_END(detail)
 
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -383,10 +383,10 @@ constexpr bool args_any_are_buffer() {
     return detail::any_of<std::is_same<Args, buffer_protocol>...>::value;
 }
 
+
 // Add the buffer interface to a vector
 template <typename Vector, typename Class_, typename... Args>
-enable_if_t<args_any_are_buffer<Args...>()>
-vector_buffer(Class_& cl) {
+void vector_buffer_impl(Class_& cl, std::true_type) {
     using T = typename Vector::value_type;
 
     static_assert(vector_has_data_and_format<Vector>::value, "There is not an appropriate format descriptor for this vector");
@@ -424,8 +424,12 @@ vector_buffer(Class_& cl) {
 }
 
 template <typename Vector, typename Class_, typename... Args>
-enable_if_t<!args_any_are_buffer<Args...>()> vector_buffer(Class_&) {}
+void vector_buffer_impl(Class_&, std::false_type) {}
 
+template <typename Vector, typename Class_, typename... Args>
+void vector_buffer(Class_& cl) {
+    vector_buffer_impl<Vector, Class_, Args...>(cl, detail::any_of<std::is_same<Args, buffer_protocol>...>{});
+}
 
 PYBIND11_NAMESPACE_END(detail)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -474,7 +474,7 @@ jobs:
     name: "🐍 3 • ICC latest • x64"
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Add apt repo
       run: |
@@ -488,7 +488,6 @@ jobs:
       run: sudo apt-get update; sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic cmake python3-dev python3-numpy python3-pytest python3-pip
 
     - name: Update pip
-      shell: bash
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python3 -m pip install --upgrade pip
@@ -498,43 +497,69 @@ jobs:
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         python3 -m pip install -r tests/requirements.txt --prefer-binary
 
-    - name: Configure
-      shell: bash
+    - name: Configure C++11
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        cmake -S . -B build     \
+        cmake -S . -B build-11     \
         -DPYBIND11_WERROR=ON    \
         -DDOWNLOAD_CATCH=ON     \
         -DDOWNLOAD_EIGEN=OFF    \
         -DCMAKE_CXX_STANDARD=11             \
         -DCMAKE_CXX_COMPILER=$(which icpc)  \
-        -DCMAKE_VERBOSE_MAKEFILE=ON         \
         -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
 
-    - name: Build
-      shell: bash
+    - name: Build C++11
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        cmake --build build -j 2
+        cmake --build build-11 -j 2 -v
 
-    - name: Python tests
-      shell: bash
+    - name: Python tests C++11
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
         sudo service apport stop
-        cmake --build build --target check
+        cmake --build build-11 --target check
 
-    - name: C++ tests
-      shell: bash
+    - name: C++ tests C++11
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        cmake --build build --target cpptest
+        cmake --build build-11 --target cpptest
 
-    - name: Interface test
-      shell: bash
+    - name: Interface test C++11
       run: |
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        cmake --build build --target test_cmake_build
+        cmake --build build-11 --target test_cmake_build
+
+    - name: Configure C++17
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake -S . -B build-17     \
+        -DPYBIND11_WERROR=ON    \
+        -DDOWNLOAD_CATCH=ON     \
+        -DDOWNLOAD_EIGEN=OFF    \
+        -DCMAKE_CXX_STANDARD=17             \
+        -DCMAKE_CXX_COMPILER=$(which icpc)  \
+        -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
+
+    - name: Build C++17
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build-17 -j 2 -v
+
+    - name: Python tests C++17
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        sudo service apport stop
+        cmake --build build-17 --target check
+
+    - name: C++ tests C++17
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build-17 --target cpptest
+
+    - name: Interface test C++17
+      run: |
+        set +e; source /opt/intel/oneapi/setvars.sh; set -e
+        cmake --build build-17 --target test_cmake_build
 
 
   # Testing on CentOS (manylinux uses a centos base, and this is an easy way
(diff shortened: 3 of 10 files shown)


tobiasleibner:
Thanks for creating the PR! 

Maybe you can add a comment to your fix explaining why it is needed? Just a short sentence similar to the comments I added to my changes.

You should probably rebase on/merge in the master branch, maybe that fixes some of the non-ICC failures in CI. Not sure what the test failure in the ICC branches is about, I think I haven't seen that error before.

mkuron:
Thanks for the comment. I added a note about the internal compiler error and merged master (still getting unrelated CI errors though). It might make sense to wait for #2573 to be merged first and then squash/rebase the present pull request to eliminate the overlap.

include/pybind11/cast.h
```diff
@@ -2172,16 +2172,35 @@ class unpacking_collector {
...
 };
 
+#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
+// We need this Intel compiler workaround because it fails to compile
+// the enable_if_t<!all_of<is_positional<Args>...>::value> part below
+// (tested with ICC 2021.1 Beta 20200827).
+template <typename... Args>
+constexpr bool args_are_all_positional()
+{
+  return all_of<is_positional<Args>...>::value;
```
bstaletic, line 2182: Should we bite the bullet and use this unconditionally? The comment can still stay.

mkuron: Sounds reasonable, so I changed it. The two variants are absolutely identical.

bstaletic:
Rebase is needed. Plus there's that extra empty line in the yaml file.

henryiii:
CI always bases "PR" builds on the latest master when it changes (including when it reopens). I think the empty line was in a different PR and I fixed it already.

henryiii:
Oh, you mean it needs to be rebased after #2573? Yes, apparently so.

bstaletic:
> Oh, you mean it needs to be rebased after #2573? Yes, apparently so.

No, no. I thought it needed a rebase after fixing the CI.

___

The `test_aligned()` fails and it could be argued that it's pybind's fault. This test wants to make sure that an overaligned type is correctly overaligned by specifying `alignas(1024)`. Here's why this doesn't work:

1. The definition of "extended alignment"/"overaligned type" is [here](https://eel.is/c++draft/basic.align#3). It's essentially any type for which `alignof(T) > alignof(std::max_align_t)`.
2. The same clause of the standard says that exact support for overaligned types/extended alignment is implementation defined.
3. [This clause](https://eel.is/c++draft/dcl.align#2) explains what are the requirements of `alignas(1024)`.
3.1. The first requirement just says that our `1024` has to be a compile time constant integer.
3.2. [The second requirement](https://eel.is/c++draft/dcl.align#2.2) says that if our `1024` is not supported by the C++ implementation, the program is ill-formed.

Therefore, ICC is well within its rights to disregard the extended alignment request. Granted, the standard says **ill-formed**, not **ill-formed, no diagnostic required**. At least according to [cppreference](https://en.cppreference.com/w/cpp/language/ub), ICC is required to display a warning.

___

For what it's worth, we asked for a 1024-byte alignment and we got 16-byte alignment.

bstaletic:
On the other hand, compiler explorer says that 1024-byte alignment works fine in ICC 21.1.

https://godbolt.org/z/f8oqPq

mkuron:
Okay, so what do we do about the alignment test? I played around with it a bit yesterday and was unable to fix it, but also unable to reproduce outside pybind11. Disable it on Intel?

bstaletic:
https://github.com/pybind/pybind11/blob/master/tests/test_class.cpp#L392-L399

Maybe just add `&& !defined(__INTEL__)` or whatever is the ICC-identifying macro? Though I'm curious if ICC would respect alignment in an extension module if we remove pybind11:

```c++
#include <Python.h>

struct alignas(1024) Aligned {
	uintptr_t ptr() { return (uintptr_t)this; }
};

typedef struct {
    PyObject_HEAD
    Aligned* aligned;
} PyAligned;

static void PyAligned_dealloc(PyAligned *self) {
    delete self->aligned;
    Py_TYPE(self)->tp_free((PyObject *) self);
}

static PyObject * PyAligned_ptr(PyAligned *self, PyObject *Py_UNUSED(ignored)) {
    if (self->aligned == NULL) {
        PyErr_SetString(PyExc_AttributeError, "last");
        return NULL;
    }
    return PyUnicode_FromFormat("%x", self->aligned->ptr());
}
static PyObject * PyAligned_new(PyTypeObject *type, PyObject *args, PyObject *kwds) {
    PyAligned *self = (PyAligned *) type->tp_alloc(type, 0);
    if (self != NULL) self->aligned = new Aligned;
    return (PyObject *) self;
}


static PyMethodDef PyAligned_methods[] = {
    {"ptr", (PyCFunction) PyAligned_ptr, METH_NOARGS},
    {NULL}  /* Sentinel */
};

static PyTypeObject AlignedType = {
    PyVarObject_HEAD_INIT(NULL, 0)
    .tp_name = "t.Aligned",
    .tp_basicsize = sizeof(PyAligned),
    .tp_itemsize = 0,
    .tp_dealloc = (destructor) PyAligned_dealloc,
    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
    .tp_methods = PyAligned_methods,
    .tp_new = PyAligned_new,
};

static PyModuleDef tmodule = {
    PyModuleDef_HEAD_INIT,
    .m_name = "t",
    .m_size = -1,
};

PyMODINIT_FUNC PyInit_t(void) {
    if (PyType_Ready(&AlignedType) < 0)
        return NULL;

    PyObject* m = PyModule_Create(&tmodule);
    if (m == NULL) return NULL;

    Py_INCREF(&AlignedType);
    if (PyModule_AddObject(m, "Aligned", (PyObject *) &AlignedType) < 0) {
        Py_DECREF(&AlignedType);
        Py_DECREF(m);
        return NULL;
    }

    return m;
}
```

Test with:

    python -c 'print(__import__("t").Aligned().ptr() % 1024)'

One difference between what pybind11 is doing and what the above (almost) C is doing is how the bindings are made. Pybind11 adds an empty type object to the module object and then adds python methods to modules or types, as it encounters `def` calls. The above C statically prepares the entire type object and adds the complete thing to the `t` module.

mkuron:
```
$ icpc -std=c++17 -shared -fPIC -o t.so t.cpp $(python3-config --cflags --ldflags)
$ python3 -c 'print(__import__("t").Aligned().ptr() % 1024)'
736
```
So it's not aligned either. I'll add the `#ifdef __INTEL_COMPILER` to disable the test.

bstaletic:
The remaining test failure, is also present in the C++11 ICC run. I'm not sure what's going on there.

As for the alignment problem, here's a tiny example that repro's the problem: https://godbolt.org/z/MYj5r5
Considering that lbistdc++ does have aligned new and the ICC uses whatever standard library comes with the system, this has to be a result of ICC not defining some feature test macro!

Sure enough, libstdc++ protects the aligned `new` operators with `__cpp_aligned_new`. So, what happens if we lie a bit?
https://godbolt.org/z/sbM4vK

cppreference says that ICC supports aligned `new` since version 19, but Intel seems to be lazy about feature test macros...
https://en.cppreference.com/w/cpp/compiler_support

mkuron:
> Sure enough, libstdc++ protects the aligned new operators with __cpp_aligned_new. So, what happens if we lie a bit?

Excellent find! If I add the lie to your pybind11-less example from earlier today, it aligns correctly. What would you think about adding
```c++
#if defined(__INTEL_COMPILER) && __INTEL_COMPILER >= 1900 && defined(PYBIND11_CPP17)
  // Intel compiler doesn't advertise that it supports alignas, so libstdc++ doesn't
  // enable the aligned new operators by default.
  #define __cpp_aligned_new 1
#endif
```
somewhere near https://github.com/pybind/pybind11/blob/7bd4b39754297ca4435810868520251820cc3091/include/pybind11/detail/common.h#L104?

bstaletic:
I was wary about suggesting that. I'm not sure what would happen if one translation contained that lie but another did not and both tried to use overaligned types? We would still be in ill-formed land, with a bit of ODR violations sprinkled around? Would it be possible to construct a case where one TU would think the pointer is overaligned when it really isn't?

That's why I don't want to make the decision and would rather let others chime in with their thoughts.

mkuron:
I've found an alternative and less hacky solution: adding `#include <aligned_new>`. That's an Intel-provied header file that defines global `operator new`s and `operator delete`s and is documented as the official solution to getting aligned new support in ICC: https://software.intel.com/content/www/us/en/develop/articles/aligned-operator-new-support-in-intel-c-compiler.html, https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/optimization-and-programming-guide/automatically-aligned-dynamic-allocation/automatically-aligned-dynamic-allocation-1.html . The documentation says nothing about that one has to include this header in all translation units, so I assume it's safe to only do so in some.

Since this is Intel's official (and not standard-compliant) way of requesting aligned new support, I guess the correct thing for us to do would be to not add `#include <aligned_new>` to the pybind11 headers, but only do so in test_class.cpp. If someone wants to use `alignas` in the code they are interfacing with pybind11, they should already be including that file if they want to build with Intel.

bstaletic:
That header, at least on linux, overrides the aligned new operators and makes them a thin wrapper around `posix_memalign()`.

> I guess the correct thing for us to do would be to not add #include <aligned_new> to the pybind11 headers, but only do so in test_class.cpp.

Does ICC support `-include` flag, like gcc and clang do? If so, we could add `-include aligned_new` to CXXFLAGS in cmake, when we detect C++17 and ICC.

mkuron:
> Does ICC support `-include` flag, like gcc and clang do?

It does, but anyone using `alignas` in their code with an Intel compiler should already be including aligned_new per Intel's documentation. Since pybind11 doesn't do any aligning by itself, it's not really our job to inject it into the CXXFLAGS. We would only include it in the test case.

henryiii:
I'd be okay just to disable that test for now.

mkuron:
Okay, skipping the test now. I don't really understand how the exception handling works, so I don't think I would have been able to fix it. Perhaps the original author of that test (@jbarlow83, #2372) has an idea and could fix it at some later point.

Thanks for milestoning, looking forward to seeing full Intel support in a release soon.

henryiii:
Looks like aligned is acting up again?

jbarlow83:
@mkuron 

1. What's the error message for the `test_python_alreadyset_in_destructor` failure? Or does it core dump?
2. If you comment out the line in test_exceptions.cpp, `ex.discard_as_unraisable(s);`, how does the test fail?

mkuron:
@jbarlow83: I cannot reproduce this error locally, so I can only answer your first question from the CI build log:
```
../../tests/test_exceptions.py:77: AssertionError
=============================== warnings summary ===============================
test_exceptions.py::test_python_alreadyset_in_destructor
  /home/runner/.local/lib/python3.8/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: 'already_set demo'
  
    File "/home/runner/work/pybind11/pybind11/tests/test_exceptions.py", line 71, in test_python_alreadyset_in_destructor
      assert m.python_alreadyset_in_destructor("already_set demo") is True
  KeyError: 'bar'
  
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
```

mkuron:
> Looks like aligned is acting up again?

Sorry about that, I put the `#ifdef` in the wrong place. We need to `#include <aligned_new>` before pybind11 so the new operator gets picked up everywhere, but I accidentally moved it to a place where `PYBIND11_CPP17` is not defined yet. Should be fixed now.

jbarlow83:
What failed was our test for whether an appropriate error message was sent to stderr. I suspect it's related to the version of pytest or somehow CI-related. It should be compiler independent.

(The traceback is not a concern. That part is supposed to happen.)

mkuron:
> I suspect it's related to the version of pytest or somehow CI-related. It should be compiler independent.

@jbarlow83, I could not reproduce this on Ubuntu 18.04 after upgrading pytest to the same version as provided by Ubuntu 20.04. I also could not reproduce it on Ubuntu 20.04 with GCC instead of Intel ICC (I don't have Intel available on that machine). The NVCC job in CI is also running on Ubuntu 20.04 though, so it would have the exact same Python environment as the ICC job, yet it does not exhibit the problem -- so there might be some compiler dependency. No idea why the choice of compiler should affect the capturing of stderr inside pytest though.

henryiii:
Is there a way to pin a version? This seems to just pull "latest", which could change and break our tests.

henryiii:
#### Changes in this PR:

* Add Intel to the CI.

Tests only:

* `py::arg() to py::arg{}`. This is a weird one. I expect user code to use `""_a` anyway. One test still runs with the old syntax, protected from the Intel compiler.
* Drop python_alreadyset_in_destructor for intel, because it seems to fail in some cases (specifically, in our CI, but hard to reproduce locally)
* Swap noexcept for `noexcept(true)` in a test only on Intel.
* Add `<aligned_new>` to a test since Intel don’t support alignment by default (probably a backward compatibility thing)
* Change `= default` to `{}` in a few places in tests. Other compilers tend to be randomly picky about this too, causing linking errors.

Changes to pybind11:

* stl.h `vector_buffer` on C++17 uses `if constexpr` instead of `enable_if_t`
* Factored out `enable_if_t` into two simpler parts for two places. Also helps NVIDIA.
* Don’t use `__cpp_fold_expressions` even if defined on Intel.

mkuron:
> Is there a way to pin a version?

Do you mean Ubuntu version? GitHub‘s `ubuntu-latest` as specified in our CI config is currently 18.04 and will soon be 20.04, but you can also [specify a specific version](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on). Some of our CI jobs run inside Docker containers, where the Ubuntu version is determined by the base image used when the image is built.

henryiii:
No, I mean Intel compiler version. Ubuntu version doesn't bother me as much, though I certainly have seen it break/fix things. :/

mkuron:
Ah. No, as far as I can tell they only provide a single package for the latest version via their apt repo.

YannickJadoul:
> Changes in this PR:

Can we somehow indicate these changes in the code, with some kind of standard comment that a) shows Intel where to look if they want to fix things and provide a C++-compatible compiler, and b) allows us to maybe revert these in time?

henryiii:
We already almost have that. `tested with ICC 2021.1 Beta 20200827`. But we should probably formalize it, though. Including workarounds for MSVC, etc. Any proposals? It should include something searchable, the compiler (ICPC), the type of workaround, and the version tested, possibly a range or list of versions. Maybe we could even collect all of them via regex and display them somewhere in the docs?

The type of workaround would be universal or ifdef. A universal workaround is one like the `enable_if_t`, all compilers can use it. An ifdef change is one that has to be ifdef'd around based on the compiler. And finally, if we do automatically collect these, we would track if they are in tests or in the code base.

Might be a little overboard, and certainly beyond the scope of this PR, though if we come up with something, the changes in the PR could be the first to follow it?

A few of the changes aren't harmful, like the factoring out of the too-complex `enable_if_t`; those can have a comment for why (and Intel should try to fix it), but I think having the factored out, simpler form is easier to compile and easier to understand as well.

YannickJadoul:
> It should include something searchable

That's what I meant, yes, just something standardized you can look for like "ICC workaround" or "Intel workaround", or something. Apart from that, GitHub and blame should help pinning down where this came from, so I'm not too worried (unless it's something that's very important/dangerous when making changes).

> Maybe we could even collect all of them via regex and display them somewhere in the docs?

Not sure this is very relevant to users, though.

> A few of the changes aren't harmful, like the factoring out of the too-complex `enable_if_t`

Nope, true. And if we're perfectly fine with the changes, then let's just go with that and forget about it, now that we have CI. But I was thinking about cases where we grudgingly make changes that we would want to reverse if possible (such as perhaps `py::arg()` to `py::arg{}`, since that doesn't match the style in the rest of the code base; or when `detail::enable_if_t<detail::all_of<detail::is_keyword_or_ds<Args>...>::value>` gets expanded into 5 extra lines + comments).

Alternatively, we make some kind of a `compiler_workarounds.md` file or so, that contains a list. I'm not a big fan of overly formalizing this, or adding this lits to the docs themselves, but just as a "clean up if possible" list.

henryiii:
I don't think it would hurt in https://pybind11.readthedocs.io/en/latest/limitations.html, or maybe we could add an un-linked page. A compiler_workarounds.md page would be okay, but it would be harder to maintain (by a little bit) than piggy-backing on our docs generation.

mkuron:
> certainly beyond the scope of this PR, though if we come up with something, the changes in the PR could be the first to follow it?

Is there anything you still need me to do in this pull request before it can be merged, @henryiii?

henryiii:
We need a sign off from @wjakob for this one. He’s worried about worried about working around bugs in the Intel compiler instead of getting Intel to fix their bugs. If we provide a clear path forward, that will help. Tomorrow I’m going to see if the old ifdef workarounds are still needed. If we can drop a few workarounds while adding more, that will help. Yannicks idea for cataloging workarounds could help too. I’ll also try to see if I can get the bugs reported to Intel if I don’t have an issue number for them (I have very few currently, maybe one).

ax3l:
@jeffhammond would your compiler folks be interested in triaging those icc bugs?

ax3l:
C++17 compiler bug request submitted as intel bug 04916230 on https://supporttickets.intel.com

henryiii:
> C++17 compiler bug request submitted

Which bug is this? The `enable_if_t` problems on C++17? Can we collect a list (@ax3l please edit below if needed)?


* arg()/arg{} transformation lifetime issue submitted as intel bug 04916223
* C++17 compiler bug request submitted as intel bug 04916230

(This would be a great use for GitHub Discussions, actually ;) )

ax3l:
Yes, the `enable_if_t` part in this PR. I pointed to both PRs in the bug reports with a brief summary, so that the compiler devs can see the work-arounds needed for ICC. It's probably a good approach to rebase this PR once #2573 is merged so the work-arounds here are easier to spot.

mkuron:
> It's probably a good approach to rebase this PR once #2573 is merged so the work-arounds here are easier to spot.

Will do. I‘ll squash the commits into one per bug report and add the bug numbers to the commit messages. Could you please also file a bug report for the internal compiler error in `PYBIND11_EXPAND_SIDE_EFFECTS`?

include/pybind11/stl_bind.h
```diff
@@ -377,8 +377,15 @@ struct vector_has_data_and_format<Vector, enable_if_t<std::is_same<decltype(form
 
 // Add the buffer interface to a vector
 template <typename Vector, typename Class_, typename... Args>
+#if defined(__INTEL_COMPILER) && defined(PYBIND11_CPP17)
+// Workaround as the Intel compiler does not compile the enable_if_t part below
```
laramiel, line 381: Why not make the same constexpr workaround that you have elsewhere?

henryiii:
@bstaletic @YannickJadoul Any idea why MSVC hates my final commit, but everybody else is fine with it, and it looks pretty much like what we are already doing?

YannickJadoul:
@henryiii, what about using `satisfies_all_of`? Does that help?

henryiii:
Interesting, so MSVC gets a different `any_of`/`all_of`. I could try `satisfies_any_of`, just in case, but it aught to be similar. Don't see how `<expression>` and `!<expression>` can be different.

YannickJadoul:
To state the obvious: the difference with the working example above is that `enable_if` is here the return value, rather than an optional template argument. So maybe it gets SFINAEd too late? The return value is not part of overloading, so ... maybe?

Is there a way we can add this as an argument? Or maybe in two steps, with an `_impl`, or so?

bstaletic:
MSVC not liking valid templates in C++17 usually smells of a missing `/permissive-` flag. Does that help?

henryiii:
It should be there in some places, and it's also failing in 14. It's failing for all MSVC, 2015, 2017, and 2019.

henryiii:
I think I came up with an `impl` that works. If it passes all the way, I'll push a comment and README update.

henryiii:
Okay, ready for a review. I've cleaned up the one ugly workaround, and added greppable comments and such, and listed the three workarounds implemented at the top of the PR.

include/pybind11/stl_bind.h
```diff
@@ -375,10 +375,20 @@ struct vector_has_data_and_format : std::false_type {};
...
+// [workaround(intel)] Separate function required here
+// Workaround as the Intel compiler does not compile the enable_if_t part below
+// (tested with icc (ICC) 2021.1 Beta 20200827)
+template <typename... Args>
+constexpr bool args_any_are_buffer() {
+    return detail::any_of<std::is_same<Args, buffer_protocol>...>::value;
+}
+
+// [workaround(intel)] Separate function required here
+// [workaround(msvc)] Can't use constexpr bool in return type
```
YannickJadoul, line 387: Is MSVC failing on the `constexpr` or on the combination with `std::enable_if`?

henryiii: The original all-in-one was too complicated for ICC to compile. Refactoring it to match the others, with a constexpr function, confused MSVC if it was in the return type. This is simpler to compile and avoids a constexpr function in a return type. As I don't think thing this function is an external interface, and it's only used here, I could avoid the intermediate function and call the "impl" style function directly, if that's seen as cleaner.

YannickJadoul (commented):
After #2573, I think this one can go in as well. Workarounds are never pretty, but let's hope we can get rid of them soon enough, if Intel is working on it :-)

PR #2730, opened 2020-12-18
Add a std::filesystem::path <-> os.PathLike caster.

anntzer:
## Description

On C++17/Py3.6+, convert std::filesystem::path to pathlib.Path and os.PathLike to std::filesystem::path.

I don't know whether std::filesystem::path gets much usage, but pathlib.Path is certainly getting quite popular, so having a way to auto-handle it would be nice.

## Suggested changelog entry:

```rst
`pybind11/stl.h` now registers a type caster that, on C++17/Python 3.6+, converts `std::filesystem::path` to `pathlib.Path` and any `os.PathLike` to `std::filesystem::path`.
```

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -41,11 +41,21 @@
 #    include <variant>
 #    define PYBIND11_HAS_VARIANT 1
 #  endif
+// std::filesystem::path
+#  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
+      PY_VERSION_HEX >= 0x03060000
+#    include <filesystem>
+#    define PYBIND11_HAS_FILESYSTEM 1
+#  endif
 #elif defined(_MSC_VER) && defined(PYBIND11_CPP17)
 #  include <optional>
 #  include <variant>
 #  define PYBIND11_HAS_OPTIONAL 1
 #  define PYBIND11_HAS_VARIANT 1
+#  if PY_VERSION_HEX >= 0x03060000
+#    include <filesystem>
+#    define PYBIND11_HAS_FILESYSTEM 1
+#  endif
 #endif
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
@@ -375,6 +385,77 @@ struct variant_caster<V<Ts...>> {
 #if defined(PYBIND11_HAS_VARIANT)
 template <typename... Ts>
 struct type_caster<std::variant<Ts...>> : variant_caster<std::variant<Ts...>> { };
+#endif
+
+#if defined(PYBIND11_HAS_FILESYSTEM)
+template<typename T> struct path_caster {
+
+private:
+    static PyObject* unicode_from_fs_native(const std::string& w) {
+#if !defined(PYPY_VERSION)
+        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
+#else
+        // PyPy mistakenly declares the first parameter as non-const.
+        return PyUnicode_DecodeFSDefaultAndSize(
+            const_cast<char*>(w.c_str()), ssize_t(w.size()));
+#endif
+    }
+
+    static PyObject* unicode_from_fs_native(const std::wstring& w) {
+        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
+    }
+
+public:
+    static handle cast(const T& path, return_value_policy, handle) {
+        if (auto py_str = unicode_from_fs_native(path.native())) {
+            return module::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
+                   .release();
+        }
+        return nullptr;
+    }
+
+    bool load(handle handle, bool) {
+        // PyUnicode_FSConverter and PyUnicode_FSDecoder normally take care of
+        // calling PyOS_FSPath themselves, but that's broken on PyPy (PyPy
+        // issue #3168) so we do it ourselves instead.
+        PyObject* buf = PyOS_FSPath(handle.ptr());
+        if (!buf) {
+            PyErr_Clear();
+            return false;
+        }
+        PyObject* native = nullptr;
+        if constexpr (std::is_same_v<typename T::value_type, char>) {
+            if (PyUnicode_FSConverter(buf, &native)) {
+                if (auto c_str = PyBytes_AsString(native)) {
+                    // AsString returns a pointer to the internal buffer, which
+                    // must not be free'd.
+                    value = c_str;
+                }
+            }
+        } else if constexpr (std::is_same_v<typename T::value_type, wchar_t>) {
+            if (PyUnicode_FSDecoder(buf, &native)) {
+                if (auto c_str = PyUnicode_AsWideCharString(native, nullptr)) {
+                    // AsWideCharString returns a new string that must be free'd.
+                    value = c_str;  // Copies the string.
+                    PyMem_Free(c_str);
+                }
+            }
+        }
+        Py_XDECREF(native);
+        Py_DECREF(buf);
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+            return false;
+        } else {
+            return true;
+        }
+    }
+
+    PYBIND11_TYPE_CASTER(T, _("os.PathLike"));
+};
+
+template<> struct type_caster<std::filesystem::path>
+    : public path_caster<std::filesystem::path> {};
 #endif
 
 PYBIND11_NAMESPACE_END(detail)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -247,6 +247,41 @@ if(Boost_FOUND)
   endif()
 endif()
 
+# Check if we need to add -lstdc++fs or -lc++fs or nothing
+if(MSVC)
+  set(STD_FS_NO_LIB_NEEDED TRUE)
+else()
+  file(
+    WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+    "#include <filesystem>\nint main(int argc, char ** argv) {\n  std::filesystem::path p(argv[0]);\n  return p.string().length();\n}"
+  )
+  try_compile(
+    STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR}
+    SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+    COMPILE_DEFINITIONS -std=c++17)
+  try_compile(
+    STD_FS_NEEDS_STDCXXFS ${CMAKE_CURRENT_BINARY_DIR}
+    SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+    COMPILE_DEFINITIONS -std=c++17
+    LINK_LIBRARIES stdc++fs)
+  try_compile(
+    STD_FS_NEEDS_CXXFS ${CMAKE_CURRENT_BINARY_DIR}
+    SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+    COMPILE_DEFINITIONS -std=c++17
+    LINK_LIBRARIES c++fs)
+endif()
+
+if(${STD_FS_NEEDS_STDCXXFS})
+  set(STD_FS_LIB stdc++fs)
+elseif(${STD_FS_NEEDS_CXXFS})
+  set(STD_FS_LIB c++fs)
+elseif(${STD_FS_NO_LIB_NEEDED})
+  set(STD_FS_LIB "")
+else()
+  message(WARNING "Unknown compiler - not passing -lstdc++fs")
+  set(STD_FS_LIB "")
+endif()
+
 # Compile with compiler warnings turned on
 function(pybind11_enable_warnings target_name)
   if(MSVC)
@@ -356,6 +391,8 @@ foreach(target ${test_targets})
     target_link_libraries(${target} PRIVATE Boost::headers)
     target_compile_definitions(${target} PRIVATE -DPYBIND11_TEST_BOOST)
   endif()
+
+  target_link_libraries(${target} PRIVATE ${STD_FS_LIB})
 
   # Always write the output file directly into the 'tests' directory (even on MSVC)
   if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -238,6 +238,12 @@ TEST_SUBMODULE(stl, m) {
         .def("member_initialized", &opt_exp_holder::member_initialized);
 #endif
 
+#ifdef PYBIND11_HAS_FILESYSTEM
+    // test_fs_path
+    m.attr("has_filesystem") = true;
+    m.def("parent_path", [](const std::filesystem::path& p) { return p.parent_path(); });
+#endif
+
 #ifdef PYBIND11_HAS_VARIANT
     static_assert(std::is_same<py::detail::variant_caster_visitor::result_type, py::handle>::value,
                   "visitor::result_type is required by boost::variant in C++11 mode");
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -162,6 +162,25 @@ def test_exp_optional():
     assert holder.member_initialized()
 
 
+@pytest.mark.skipif(not hasattr(m, "has_filesystem"), reason="no <filesystem>")
+def test_fs_path():
+    from pathlib import Path
+
+    class PseudoStrPath:
+        def __fspath__(self):
+            return "foo/bar"
+
+    class PseudoBytesPath:
+        def __fspath__(self):
+            return b"foo/bar"
+
+    assert m.parent_path(Path("foo/bar")) == Path("foo")
+    assert m.parent_path("foo/bar") == Path("foo")
+    assert m.parent_path(b"foo/bar") == Path("foo")
+    assert m.parent_path(PseudoStrPath()) == Path("foo")
+    assert m.parent_path(PseudoBytesPath()) == Path("foo")
+
+
 @pytest.mark.skipif(not hasattr(m, "load_variant"), reason="no <variant>")
 def test_variant(doc):
     assert m.load_variant(1) == "int"
--- a/docs/advanced/cast/overview.rst
+++ b/docs/advanced/cast/overview.rst
@@ -151,6 +151,8 @@ as arguments and return values, refer to the section on binding :ref:`classes`.
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::variant<...>``              | Type-safe union (C++17)   | :file:`pybind11/stl.h`        |
 +------------------------------------+---------------------------+-------------------------------+
+| ``std::filesystem::path<T>``       | STL path (C++17) [#]_     | :file:`pybind11/stl.h`        |
++------------------------------------+---------------------------+-------------------------------+
 | ``std::function<...>``             | STL polymorphic function  | :file:`pybind11/functional.h` |
 +------------------------------------+---------------------------+-------------------------------+
 | ``std::chrono::duration<...>``     | STL time duration         | :file:`pybind11/chrono.h`     |
@@ -163,3 +165,7 @@ as arguments and return values, refer to the section on binding :ref:`classes`.
 +------------------------------------+---------------------------+-------------------------------+
 | ``Eigen::SparseMatrix<...>``       | Eigen: sparse matrix      | :file:`pybind11/eigen.h`      |
 +------------------------------------+---------------------------+-------------------------------+
+
+.. [#] ``std::filesystem::path`` is converted to ``pathlib.Path`` and
+   ``os.PathLike`` is converted to ``std::filesystem::path``, but this requires
+   Python 3.6 (for ``__fspath__`` support).


skoslowski:
I have been using such a caster for quite a while now and find it very useful. 
Haven't had a chance to make a proper PR yet, here is my version https://github.com/skoslowski/pybind11/commit/767ac4dd79d7865fb418501fb64fc72a69a5c535

Looking at this PR 
* I really like that a pathlib.Path is returned (instead of str)
* using the os-native string types is probably more efficient compared to the utf8 API
* I am pretty sure the error state must be cleared before returning.

anntzer:
> I am pretty sure the error state must be cleared before returning.

Thanks for pointing this out, I was not aware of that.  I can confirm this is needed e.g. to make variant casters work.  OTOH it looks like the docs at https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html#custom-type-casters are wrong?  The example `load` clearly returns `!PyErr_Occurred()`...

---

AFAICT the only relevant failing tests occur with versions of clang for which is it still necessary to link `-lc++fs`.  I am a bit confused as to why, as the tests use clang 10 but https://libcxx.llvm.org/docs/UsingLibcxx.html#using-filesystem says it's only needed until clang 9.  Any help on that side would be most appreciated (https://gitlab.kitware.com/cmake/cmake/-/issues/17834 is not very helpful...).

include/pybind11/stl.h
```diff
@@ -305,6 +314,59 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+    template<typename T> static PyObject* unicode_from_fs_native(T w) {
+        if constexpr (std::is_same_v<T, std::string>) {
+            return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
+        } else if constexpr (std::is_same_v<T, std::wstring>) {
+            return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
+        }
+    }
+
+    static handle cast(std::filesystem::path path, return_value_policy, handle) {
+        PyObject* py_str = unicode_from_fs_native(path.native());
```
bstaletic, line 329: You took `path` by value. Now that you've paid the price of the copy, use `std::move(path).native()`. (Hopefully in C++23) it will let you move the string out.

Maybe we need two overloads - `const&` and `&&`?

anntzer: good point, but I think just taking `const&` is enough here?  changed accordingly.

include/pybind11/stl.h
```diff
@@ -305,6 +314,59 @@ template<> struct type_caster<std::experimental::nullopt_t>
     : public void_caster<std::experimental::nullopt_t> {};
 #endif
 
+#if defined(PYBIND11_HAS_FILESYSTEM)
+template<> struct type_caster<std::filesystem::path> {
+
+    static PyObject* unicode_from_fs_native(const std::string& w) {
+        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
```
bstaletic, line 321: Why this decode thing and not `PyUnicode_FromStringAndSize()`?

include/pybind11/stl.h
```diff
@@ -305,6 +314,59 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+    }
+
+    static PyObject* unicode_from_fs_native(const std::wstring& w) {
+        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
+    }
```
bstaletic, line 326: We could merge these two. `if constexpr` would be fine, because this caster already depends on C++17. My first attempt was:

```c++
static PyObject* unicode_from_fs_native(const std::filesystem::path::string_type& w) {
    if constexpr (std::is_same_v<std::filesystem::path::string_type, std::string>) {
        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
    } else if (!std::is_same_v<std::filesystem::path::string_type, std::wstring>) {
        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
    }
}
```

That doesn't work, because the compiler can still see that `w.c_str()` doesn't match both `char*` and `wchar_t*`. So we have to make it a template...

```c+
template<typename T = std::filesystem::path::string_type>
PyObject* unicode_from_fs_native(const T& w) {
    if constexpr (std::is_same_v<T, std::string>) {
        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
    } else if (std::is_same_v<T, std::wstring>) {
        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
    }
}
```

The problem with the above is that we really don't want `T` to be deduced - we want the default to always apply. The language calls this a [non-deduced context](https://eel.is/c++draft/temp.deduct.type#5) and `std::type_identity_t<T>` is exactly a non-deduced `T`.

```c++
template<typename T = std::filesystem::path::string_type>
PyObject* unicode_from_fs_native(const std::type_identity_t<T>& w) {
    if constexpr (std::is_same_v<T, std::string>) {
        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
    } else if (std::is_same_v<T, std::wstring>) {
        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
    }
}
```

This version finally works, but `std::type_identity<T>` is C++20... Final result, with type identity reimplemented, can be found here: https://godbolt.org/z/hPYdq7

include/pybind11/stl.h
skoslowski: IMO this is the correct call for decoding a "native path name as a character string" as obtained from `path.native()` below. Assuming UTF-8 here will probably work anyway - to be sure, however, `w` should be obtained from `path.u8string()`.

anntzer: Assuming UTF8 is wrong (as the Python devs learned painfully during the transition to Py3).  See e.g. the series of posts starting at https://vstinner.github.io/python30-listdir-undecodable-filenames.html, or https://www.python.org/dev/peps/pep-0538/.  In particular, undecodable filenames must be supported, which is done in Python using the surrogateescape handler (which is what DecodeFSDefault does).

include/pybind11/stl.h
anntzer: I basically went through the same motions until realizing the need for a non-deduced context, and not being aware of `type_identity` (thanks for pointing it out to me) I decided a good old overload was good enough.  I'd say in this case the overload is actually more legible than the `if constexpr` + backport `type_identity` version?
Still, if you really prefer that, let me know and I'll make the changes.

bstaletic: Yeah, I'm also not convinced that template + `if constexpr` + backport of `type_identity` is worth it. Keep it as is.

include/pybind11/stl.h
bstaletic: In my defense, it was almost 04:30 when I made the review. Those articles will be an interesting read, thanks.

include/pybind11/stl.h
anntzer: Actually I found a nicer way to achieve this, using a lambda taking auto as parameter (which effectively gives us something like a non-deduced context).

include/pybind11/stl.h
```diff
@@ -305,6 +314,58 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+            return nullptr;
+        }
+        return module::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
+               .release();
+    }
+
+    bool load(handle handle, bool) {
+        using char_type = std::filesystem::path::value_type;
+        if constexpr (std::is_same_v<char_type, char>) {
+            PyBytesObject* py_bytes = nullptr;
```
bstaletic, line 339: You're supposed to use `PyObject*` even when you know you need a bytes object. It would help with the `reinterpret_cast` below.

include/pybind11/stl.h
```diff
@@ -305,6 +314,58 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+            char* c_str = nullptr;
+            if (PyUnicode_FSConverter(handle.ptr(), &py_bytes)
+                // Silence MSVC warning.
+                && ((c_str = PyBytes_AsString(reinterpret_cast<PyObject*>(py_bytes)))) != 0) {
+                value = c_str;
+            }
+            Py_XDECREF(py_bytes);
+        } else if constexpr (std::is_same_v<char_type, wchar_t>) {
+            PyUnicodeObject* py_unicode = nullptr;
```
bstaletic, line 348: Same here - just use `PyObject*`.

include/pybind11/stl.h
```diff
@@ -305,6 +314,58 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+            if constexpr (std::is_same_v<char_type, char>) {
+                return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
+            } else if constexpr (std::is_same_v<char_type, wchar_t>) {
+                return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
+            }
+        } (path.native());
+        if (!py_str) {
+            return nullptr;
+        }
+        return module::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
```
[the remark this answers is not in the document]
anntzer: I don't think there's much to be gained from that, and this feels just annoying for end users who may want to actually perform filesystem ops using the Python object (then they'd need to cast the thing to a Path)?  Also std::filesystem::path does come with filesystem access (non-member) functions, so that makes it closer to Path than to PurePath anyways.
I think PurePath is really only relevant for the case of creating Windows paths on Linux and vice-versa, but std::filesystem::path doesn't allow that AFAICT...

include/pybind11/stl.h
anntzer: good point, this allows factoring out a bit more common code.

include/pybind11/stl.h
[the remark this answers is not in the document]
bstaletic: Good point.

include/pybind11/stl.h
```diff
@@ -305,6 +314,55 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+            if (PyUnicode_FSConverter(handle.ptr(), &tmp)) {
+                if (auto c_str = PyBytes_AsString(tmp)) {
+                    value = c_str;
+                }
+            }
+        } else if constexpr (std::is_same_v<char_type, wchar_t>) {
+            if (PyUnicode_FSDecoder(handle.ptr(), &tmp)) {
+                if (auto c_str = PyUnicode_AsWideCharString(tmp, nullptr)) {
+                    value = c_str;
+                    PyMem_Free(c_str);
```
bstaletic, line 349: Why is this `PyMem_Free` here? Looks to me like it's leaving `value` in an invalid state.

tests/test_stl.cpp
```diff
@@ -237,6 +237,12 @@ TEST_SUBMODULE(stl, m) {
         .def("member_initialized", &opt_exp_holder::member_initialized);
 #endif
 
+#ifdef PYBIND11_HAS_FILESYSTEM
+    // test_fs_path
+    m.attr("has_filesystem") = true;
+    m.def("parent_path", [](std::filesystem::path p) { return p.parent_path(); });
```
bstaletic, line 243: `const&`? It's a test, so it's not like performance matters...

include/pybind11/stl.h
anntzer: AFAICT the assignment will *copy* the contents of the string; this can be checked by inserting e.g. `c_str[0] = 'z';` immediately after `value = c_str;`.  (I have checked that this is indeed the case on Linux; I haven't checked on Windows but that seems most likely to be the case too...)

In any case we need to call PyMem_Free somewhere to avoid a memleak, and it's certainly not the std::filesystem::path destructor that will do that for us...

tests/test_stl.cpp
anntzer: sure

include/pybind11/stl.h
bstaletic: Aha! That ends up calling `value.operator=(c_str)` which will create a temporary `std::wstring`. That makes sense.

For anyone else wondering "Doesn't this mean that we need to free `c_str` in the other branch", no. The other branch with `char*` doesn't allocate.

anntzer:
Looks like PyPy is actually broken per https://foss.heptapod.net/pypy/pypy/-/issues/3168.  I just ifdef'd it out for now...

henryiii:
Without looking into it in detail, would it be best to put this into `stlfs.h` or something like that instead? I'm worried about it if it's not clearly opt-in, because most compilers made `std::filesystem` optional via an extra library, such as with `-lstdc++fs`.

I'm a fan, just trying to avoid any issues with disruption; users should not be required to link to the filesystem lib if they don't use it. If it only requires linking with the filesystem if really is used, then ignore this and it can stay here.

bstaletic:
> because most compilers made std::filesystem optional via an extra library, such as with -lstdc++fs, until pretty recently.

Is clang 8 considered "pretty recent"? I guess it is.

> I'm a fan, just trying to avoid any issues with disruption; users should not be required to link to the filesystem lib if they don't use it. If it only requires linking with the filesystem if really is used, then ignore this and it can stay here.

Templates aren't instantiated if they aren't actually used. This should PR should be safe, though I can't guarantee.

> would it be best to put this into stlfs.h

It would definitely make `stl.h` not larger and people needing `stlfs.h` wouldn't need to include `list_caster` and similar. I was thinking about proposing moving the caster to a new header, but that's because I'm not a big fan of `stl.h`.

include/pybind11/stl.h
YannickJadoul: If this tripped @bstaletic, it might need a comment?

YannickJadoul:
> Without looking into it in detail, would it be best to put this into `stlfs.h` or something like that instead?

> I was thinking about proposing moving the caster to a new header, but that's because I'm not a big fan of `stl.h`.

Yes, for exactly the reason @bstaletic mentions; I think that's a good one. This might also be the moment to think about another PR where `optional` is moved out of `stl.h` (and e.g. have `stl_utils.h` vs. `stl_containers.h` and a (soon-to-be-deprecated?) `stl.h` that just includes both?), as @bstaletic has been requesting for a while? :-)

include/pybind11/stl.h
YannickJadoul: Can I be annoying and say that this lambda too me a lot longer to parse than selecting the right overload?

tests/test_stl.py
```diff
@@ -162,6 +162,13 @@ def test_exp_optional():
     assert holder.member_initialized()
 
 
+@pytest.mark.skipif(not hasattr(m, "has_filesystem"), reason="no <filesystem>")
+def test_fs_path():
+    from pathlib import Path
+
+    assert m.parent_path(Path("foo/bar")) == Path("foo")
```
YannickJadoul, line 169: There should probably be a test with a custom `PathLike` class, here?

bstaletic:
> think about another PR where optional is moved out of stl.h

First attempt of that was 3 years ago in #1210.

YannickJadoul:
> > think about another PR where optional is moved out of stl.h
> 
> First attempt of that was 3 years ago in #1210.

OK, so if extensions not using this don't need these special flags (as @henryiii said), this part could be ignored until later?

include/pybind11/stl.h
```diff
@@ -305,6 +316,55 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+                }
+            }
+        }
+        Py_XDECREF(tmp);
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+            return false;
```
YannickJadoul, line 358: It might be a good idea to move these clears and returns closer to the location where you could return. E.g., after `value = c_str;`, you could `return true;`.

Reason for this: first of all, it makes the decision clearer at the spot where it's tested. Second reason to also move up `PyErr_Clear`: you cannot call other C API functions with this error set.

Both would follow the style from the other casters more closely.

YannickJadoul: Oh, but what about `Py_XDECREF(tmp);`? :-/

Can we somehow turn `tmp` into a `py::object`? Or otherwise, to clarify the flow in `load`, it might be nice to move the `PyUnicode_FSDecoder` to a private (maybe static) function. Other casters also often do this.

YannickJadoul: Just a nudge/suggestion, though. Maybe you tried this, and there are good reasons for the current way things look!

include/pybind11/stl.h
```diff
@@ -305,6 +316,55 @@ template<> struct type_caster<std::experimental::nullopt_t>
...
+        } else {
+            return true;
+        }
+    }
+
+    PYBIND11_TYPE_CASTER(std::filesystem::path, _("PathLike"));
```
YannickJadoul, line 364: `_("PathLike")` or `_("os.PathLike")`? This is shown in the signatures, I believe (is it used for anything else?), and I would expect the module there. But maybe that expectation is wrong?

YannickJadoul (commented):
An annoying question: is this something that should be in pybind11, rather than say ... an external plugin `pybind11_filesystem`? I was first skeptical about this, but the actual implementation is rather clean/small, so I guess I'm OK with this.
I'd like to see pybind11 evolve to a more "plugin-based" architecture, but I guess this specific feature isn't a huge burden to maintain?

I wanted to suggest thinking about `pathlib2`, for Python 2 (and PyPy 2, which isn't deprecated), but PyPy doesn't work, currently, ofc. And in Python 2 we still lack `os.PathLike` and `__fspath__`?

Next question: does this also accept strings? If not, should strings be accepted, maybe if `convert == true` in `load`? If they already are, a test on this would be nice.

Finally, a request: could this be turned into an abstraction like `optional_caster`? Basically, this would rename the current caster, and just have `template <> struct type_caster<std::filesystem::path> : public optional_caster<std::filesystem::path> {}`.

This would allow for other C++ path-like type to quickly opt-in to this behavior, if they support a minimal interface (be constructible/assignable from `std::string` and having `.native()` or so).

henryiii:
> think about another PR where optional is moved out of stl.h

This is just about the first thing I thought of. I'd do it boost-style, where you have:

```cpp
stl.h
stl/optional.h
stl/filesystem.h
stl/...
```
Then `stl.h` just includes all the most common items in `stl/`.

YannickJadoul:
> think about another PR where optional is moved out of stl.h

That's indeed exactly the proposal of #1210 (long outdated and full of conflicts by now, ofc), and yes, that seems perfect to me :-) But so maybe we can keep this PR in `stl.h` (if no extra flags/linking are/is needed to build stuff that's not using the new casters) and do this in in a follow-up PR?

include/pybind11/stl.h
anntzer: sure, done.

tests/test_stl.py
anntzer: done

include/pybind11/stl.h
anntzer: os.PathLike seems fine (consistently e.g. with what datetime does).

include/pybind11/stl.h
anntzer: I'm not really a fan of using `py::object` here given that the whole point is to interact with the CPython C-API (via PyUnicode_\* and PyBytes_\*), so we'd have to keep wrapping and unwrapping things...  In "classic C" style having all the cleanup at the end (perhaps even using `goto exit;` on errors) is pretty standard, so at least I'm more comfortable using that.  It's not even clear to me how exactly py::object would handle refcounts wrt. being set a new value in `PyUnicode_FSConverter(..., &tmp)`.

To be clear I'm not opposed to using py::object here... I'm just writing in the style I'm comfortable with, but if you want to propose your own patch here I'm fine too.

anntzer:
Thanks for the reviews.

- I personally don't like the splitting over multiple headers (if it was up to me everything in stdlib {Python,C++} would go into stl.h...) especially as (IIRC?) you get no warning until runtime if you forgot to include a header.  Perhaps at least `module.def` could check at compile time that all relevant type casters are registered, and warn if not?  In any case, that's a separate issue, and the agreement here seems to be to keep this in `stl.h` for this PR and possibly split it out later.  At least I believe (in agreement with @bstaletic) that there's no need to link stdc++fs unless you actually use the functionality, so that specific point is not a problem.

- I switched to having a template path_caster, similar to optional_caster.

- str/bytes and PathLikes are fine, I added tests.  On the other hand pathlib backports won't really work here given that the whole PR is based on using the newer fspath-based CPython APIs.  If you want to support older versions, then you'd need to add fully separated codepaths for them, which I guess can be done but could be, well, a separate PR.

include/pybind11/stl.h
anntzer: I don't really care about which approach to use, but I'll let you and @bstaletic decide together which one you want once and for all instead of switching back and forth ;)

YannickJadoul: Sure. Up to @bstaletic. I am/was not saying it should be changed, but just surprised that this is considered cleaner than the overload :-)

include/pybind11/stl.h
YannickJadoul: Thanks! :-)

tests/test_stl.py
YannickJadoul: Great!

include/pybind11/stl.h
YannickJadoul: Fair enough. Just an idea to see if this could work/be more homogeneous with other casters. As long as the code doesn't get more complex this is fine. I guess I'm just more comfortable with C++, where you try to make sure you don't forget things with RAII and be able to return in the middle.

>  It's not even clear to me how exactly py::object would handle refcounts wrt. being set a new value in `PyUnicode_FSConverter(..., &tmp)`.

True, yes. That's an unusual interface for the Python C API.

YannickJadoul: So, to summarize, unless @bstaletic/@henryiii/other reviewers think this can be clean up/made more similar to other casters, I'm fine with the current situation, and we can just see how this piece of code evolves.

YannickJadoul:
>    * I personally don't like the splitting over multiple headers (if it was up to me everything in stdlib {Python,C++} would go into stl.h...) especially as (IIRC?) you get no warning until runtime if you forgot to include a header.  Perhaps at least `module.def` could check at compile time that all relevant type casters are registered, and warn if not?  In any case, that's a separate issue, and the agreement here seems to be to keep this in `stl.h` for this PR and possibly split it out later.  At least I believe (in agreement with @bstaletic) that there's no need to link stdc++fs unless you actually use the functionality, so that specific point is not a problem.

Indeed, let's keep this for another PR.
You're welcome to argue against it, then (I'll ping you, if I don't forget). But the current proposal would be to still keep a `stl.h` anyway (even if it'd just be for backwards compatibility). The thing is that the container casters are pretty invasive/dangerous, so I can see why people would not like to include those.

>    * I switched to having a template path_caster, similar to optional_caster.

Thanks, I think that looks good!

Potentially, `T::value_type` could be passed as a second template parameter rather than as dependent type, but let's see what @bstaletic & @henryiii think about that.
 
>    * str/bytes and PathLikes are fine, I added tests.  On the other hand pathlib backports won't really work here given that the whole PR is based on using the newer fspath-based CPython APIs.  If you want to support older versions, then you'd need to add fully separated codepaths for them, which I guess can be done but could be, well, a separate PR.

OK, fair enough; given that this is new functionality, I can live with not having this for Python 2 and 3.5, I think. Thanks for clarifying this further!

include/pybind11/stl.h
bstaletic: With a bunch of `.ptr()` sprinkled around, I'm not sure `py::object` would be cleaner.

include/pybind11/stl.h
bstaletic: I'm fine either way. If @YannickJadoul finds the overload version easier to read, I guess we should do that.

@anntzer apologies for nagging.

bstaletic:
> Potentially, T::value_type could be passed as a second template parameter

Personally, I don't see a need for that.

include/pybind11/stl.h
anntzer: back to good old overloading.

include/pybind11/stl.h
anntzer: By the way, best would probably be to have os.PathLike for parameter types (as that's what we accept) but pathlib.Path for return types (as that's what we return), but I don't know if that's something there's machinery for right now?

include/pybind11/stl.h
YannickJadoul: Sorry for nagging as well, and making you revert this, @anntzer!

include/pybind11/stl.h
```diff
@@ -305,6 +316,60 @@ template<> struct type_caster<std::experimental::nullopt_t>
     : public void_caster<std::experimental::nullopt_t> {};
 #endif
 
+#if defined(PYBIND11_HAS_FILESYSTEM)
+template<typename T> struct path_caster {
+
+    static PyObject* unicode_from_fs_native(const std::string& w) {
```
YannickJadoul, line 322: These could still become `private`, but I don't per se need this, if there are no other commits being done.

anntzer:
I think the only relevant failure remaining is with clang10, which is missing an explicit linking to -lc++fs.  Any suggestions on how to achieve that? (my cmake-fu is pretty weak...)

include/pybind11/stl.h
anntzer: private it is

include/pybind11/stl.h
YannickJadoul: I thought about that too; I don't think pybind11 currently allows for this. It should be possible, though, but probably for another PR.

At any rate, there are other places where signatures aren't perfect, when it would cost too much code or performance/space overhead, so it's not a huge worry.

YannickJadoul:
> I think the only relevant failure remaining is with clang10, which is missing an explicit linking to -lc++fs. Any suggestions on how to achieve that? (my cmake-fu is pretty weak...)

`target_link_libraries`? I can't immediately find anything better/built-in into CMake, than https://gist.github.com/berkus/aafb15b06f753f025e009d5dd2ae7e5b. But @henryiii will probably know better.

bstaletic:
> with clang10, which is missing an explicit linking to -lc++fs

Uhmmm... `-lc++fs` has nothing to do with clang, but rather the version of the C++ standard library. So the question is which standard library is being picked up.

Either way, [this](https://github.com/bstaletic/ycmd/blob/cxx17/cpp/ycm/CMakeLists.txt#L317-L345) is how I've solved the problem in my project.

include/pybind11/stl.h
YannickJadoul: Thanks! And apologies for all the nitpicking ;-)

anntzer:
Thanks, looks like everything is working now :)

tests/CMakeLists.txt
```diff
@@ -238,6 +238,41 @@ if(Boost_FOUND)
   endif()
 endif()
 
+# Check if we need to add -lstdc++fs or -lc++fs or nothing
+if(MSVC)
```
bstaletic, line 242: There's no need to special-case MSVC if you specify the standard version the way `try_compile` specifies in the docs: https://cmake.org/cmake/help/latest/command/try_compile.html

tests/CMakeLists.txt
```diff
@@ -238,6 +238,41 @@ if(Boost_FOUND)
...
+endif()
+
+if(${STD_FS_NEEDS_STDCXXFS})
+  set(STD_FS_LIB stdc++fs)
+elseif(${STD_FS_NEEDS_CXXFS})
+  set(STD_FS_LIB c++fs)
+elseif(${STD_FS_NO_LIB_NEEDED})
+  set(STD_FS_LIB "")
+else()
+  message(WARNING "Unknown compiler - not passing -lstdc++fs")
```
bstaletic, line 272: This is likely wrong. The `try_compile` failed in all three cases and we're going \*shrug\* "let's try it anyway!", so should probably be a fatal error.

tests/CMakeLists.txt
```diff
@@ -238,6 +238,41 @@ if(Boost_FOUND)
...
+  set(STD_FS_LIB "")
+endif()
+
```
bstaletic, line 275: For the record, these lines have been taken from my other project that is using the GPLv3 license. I believe this calls for a public statement.

As the only copyrighter and author of the copied snippet, I permit usage under the MIT license and without any requirement of GPLv3.

anntzer: Just to be clear, do you want attribution here?  I can add a comment to that effect.

bstaletic: It's a cmake snippet! I don't care about attribution for such a small thing.

tests/CMakeLists.txt
anntzer: thanks.

tests/CMakeLists.txt
anntzer: I don't think so?  We could be on a compiler that doesn't support C++17 at all, in which case

bstaletic: Good point. My project received complaints from users hitting this branch, but I unconditionally depend on `<filesystem>`. Maybe still wrap the whole block in an `if ( standard is 17 )` or whatever the spelling is for cmake.

anntzer: Does this have to do as part of this PR?  TBH I'm not so comfortable with cmake...

henryiii: This can’t be done in CMake in general, too many ways someone could set the std level, though since it’s our tests we can do it. I can look at this later if it’s needed.

anntzer:
Kindly bumping?

bstaletic (approved):
I'm fine with this PR. @anntzer Sorry for the delay.

anntzer:
No worries.

anntzer:
Actually, I realized it would be easy enough to also support PyPy and force-pushed a new version, but now CI seems broken due to something unrelated?  (cmake doesn't get installed properly by CI) :/

YannickJadoul:
CI should be fixed again ;-)

anntzer:
PyPy is supported now as well.

include/pybind11/stl.h
```diff
@@ -36,6 +36,12 @@
 #    include <experimental/optional>
 #    define PYBIND11_HAS_EXP_OPTIONAL 1
 #  endif
+// std::filesystem::path
+#  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
+      PY_VERSION_HEX >= 0x03060000
```
YannickJadoul, line 41: Nitpicking: Slightly concerned about this, where the other macros just mention the C++ feature (i.e., `PYBIND11_HAS_OPTIONAL` just means "does C++ has `std::optional`?"). For the future, this has a potential for confusion.

Could/should we split this into two parts, where `PYBIND11_HAS_FILESYSTEM` is here, and the Python check is further down at the definition of the caster? Just a thought; I'm not completely convinced. @bstaletic?

YannickJadoul: Ah, also, could you move this block down 5 lines, please (I said I was nitpicking, didn't I?). I'd like to keep the order `HAS_OPTIONAL`, `HAS_VARIANT`, `HAS_FILESYSTEM` consistent with the next branch of the main `#if`.

YannickJadoul (approved):
I think I'm fine with this. Be aware, though, that this will probably only go into 2.7.0 (if we follow semantic versioning closely enough).

include/pybind11/stl.h
anntzer: Done for the second point (and also moved the implementation down to match the order).  I don't really agree with the first point; I think `PYBIND11_HAS_FOO` means "pybind11 supports foo" rather than "C++ has foo" (well, before this PR both matched :)).  In particular changing this would mean that test_stl.cpp would have  to duplicate the version gating logic (which may change if someone decides at some point to also support older pythons).  But I can change that if you really want...

anntzer:
No worries re: which release.  (Although do you have any ETA for 2.7?)

include/pybind11/stl.h
YannickJadoul: > But I can change that if you really want...

I'm not sure if I really want that, either.

I've just been looking at this set of macros like that before (and wonder if in the future, if there's going to be something else depending on `std::filesystem` that's not per se the caster, the same macro might be reused). In principle, we're not advertising that macro, so users shouldn't depend on what exactly it means; in practice, well, ...

Your interpretation does indeed also match the current state, yes, so I don't know. That's why I'm asking other reviewers. :-)

>  In particular changing this would mean that test_stl.cpp would have to duplicate the version gating logic (which may change if someone decides at some point to also support older pythons).

Yes, or another macro that more specifically mentions the caster?

It would even be nicer if there would be an easy way to support 2.7 and 3.5, ofc. But out of scope for this finished PR :-)

YannickJadoul:
> No worries re: which release. (Although do you have any ETA for 2.7?)

Euhm, not really, yes. 2.6.2 is pending (matter of a couple of days/weeks, hopefully), and I would expect 2.7 is scheduled after that and we start merging a bunch of PRs.
The good news is that you should be able to just add this caster to your own code. It's annoying, but it perfectly works with older versions of pybind11.

anntzer:
> Euhm, not really, yes. 2.6.2 is pending (matter of a couple of days/weeks, hopefully), and I would expect 2.7 is scheduled after that and we start merging a bunch of PRs.

No worries.

> The good news is that you should be able to just add this caster to your own code. It's annoying, but it perfectly works with older versions of pybind11.

I know, I've been doing that for a few months already :)

include/pybind11/stl.h
anntzer: I'll just leave things as they are for now and see what other reviewers propose.

bstaletic: I'd stick with the current definition of the `PYBIND11_HAS_FILESYSTEM` for now. Once pybind11 gets `$NEW_THING` that depends on `<filesystem>`, but not on `os.Path`, the macro can be revised.

YannickJadoul: OK. I was just expecting you to pop up and shout "Hyrum's law", but let's leave it, then ;-)

EricCousineau-TRI: At present, looks good to me! Will mark this convo as resolved.

tests/CMakeLists.txt
EricCousineau-TRI: I think this is fine as-is for the purposes of this PR; if an MSVC user encounters an issue, I'm sure we can handle it more acutely.

tests/CMakeLists.txt
EricCousineau-TRI: I'll throw in peanut-galler 2 cents: Looks good for a starting point! Can address acute defects after merge?

tests/test_stl.cpp
```diff
@@ -237,6 +237,12 @@ TEST_SUBMODULE(stl, m) {
         .def("member_initialized", &opt_exp_holder::member_initialized);
 #endif
 
+#ifdef PYBIND11_HAS_FILESYSTEM
+    // test_fs_path
+    m.attr("has_filesystem") = true;
```
EricCousineau-TRI, line 242: nit (minor) Consider setting `m.attr("has_filesytem") = false;` in alternative branch.
Better to be explicit, then you simplify do `skipif(not m.has_filesystem)` and make the local contract clearer; otherwise, you could have `has_filesystem = False` and still admit the unittest.

EricCousineau-TRI (approved):
Looks great! Made some minor asides here 'n there.

I think this would be ready to merge, pending resolution on CMake `try_compile` bits?

tests/test_stl.cpp
anntzer: I was being consistent with all the other similar checks (has_optional, has_exp_optional, load_variant) which all use hasattr().  I agree just setting it to True/False looks cleaner, but that should rather be done globally in a separate cleanup PR?

EricCousineau-TRI: Great point! Will resolve this and make minor cleanup PR.

rwgk:
Hi @anntzer sorry this has been sitting here with 3 approvals for such a long time.
Is there a chance you could rebase this PR? That will trigger the CI again and run new clang-tidy checks that we added last week.

I already integrated this PR for Google-internal global testing, just to be sure. As part of that it also ran through clang sanitizers, without issues. Global test results will take ~12 hours.

anntzer:
rebased

Skylion007 (approved):

rwgk:
This also passed Google-internal global testing. Merging!

henryiii:
I believe this broke support for building wheels that set anything less than macOS 10.15 (10.9 is the default, 10.12 is "safe", but 10.15 is painfully new). I thought I asked for this to be placed in a separate file?

henryiii:
I have tried to add things like this before to CLI11, and it's really hard to get right, _especially_ for features that depend on compiled libraries being linked (like libfs), since you can't detect that at header time. Putting this in a separate file would have avoided this and made it opt in.

henryiii:
Example of a broken build: https://github.com/deepmind/open_spiel/runs/2999582108

rwgk:
> I believe this broke support for building wheels that set anything less than macOS 10.15 (10.9 is the default, 10.12 is "safe", but 10.15 is painfully new). I thought I asked for this to be placed in a separate file?
> 
> See [#2730 (comment)](https://github.com/pybind/pybind11/pull/2730#issuecomment-750507575)

Sorry I missed that, with so many approval I was thinking "What could possibly go wrong?"
Let's fix this asap.

henryiii:
I clearly said exactly what could go wrong, it was just ignored. :P

rwgk:
I told the open_spiel folks about this mishap.
Let me try a fix.

henryiii:
Just pull it out into a file and make it opt-in by including the file, say `stl/path.h`. That's how all the contents of stl should be, with "safe" ones included in a "convenience" `stl.h`. Just because past organization wasn't ideal doesn't mean we shouldn't follow best practices for new code. Especially non-header-only additions to the STL.

PR #2731, opened 2020-12-18
minor cleanup: fixing or silencing flake8 errors

rwgk:
Fixing errors that appeared after a flake8 version bump.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,8 @@ jobs:
 
     - name: Update CMake
       uses: jwlawson/actions-setup-cmake@v1.4
+      with:
+        cmake-version: 3.19.1
 
     - name: Cache wheels
       if: runner.os == 'macOS'
@@ -550,6 +552,8 @@ jobs:
 
     - name: Update CMake
       uses: jwlawson/actions-setup-cmake@v1.4
+      with:
+        cmake-version: 3.19.1
 
     - name: Prepare MSVC
       uses: ilammy/msvc-dev-cmd@v1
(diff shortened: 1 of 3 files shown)


bstaletic (approved):
Maybe we should just disable B015? Either way, approved.

rwgk:
Thanks @bstaletic ! I'll wait a bit to see if we can get the CI green (waiting for @henryiii's cmake pinning).

I think in general B015 is useful. I feel disabling it completely would be a loss, I'd go there only if we had to put in a whole bunch more of the noqa comments.

.github/workflows/ci.yml
```diff
@@ -596,6 +600,9 @@ jobs:
 
     - name: Update CMake
       uses: jwlawson/actions-setup-cmake@v1.4
+      with:
+        cmake-version: 3.19.1
+
```
bstaletic, line 605: CI is complaining about an extra empty line here.

henryiii: Are we interested in enabling pre-commit.ci? It would automatically fix things like this for us.

henryiii:
```suggestion
```

tests/test_local_bindings.py
```diff
@@ -193,7 +193,7 @@ def test_stl_caster_vs_stl_bind(msg):
     v2 = [1, 2, 3]
     assert m.load_vector_via_caster(v2) == 6
     with pytest.raises(TypeError) as excinfo:
-        cm.load_vector_via_binding(v2) == 6
+        cm.load_vector_via_binding(v2)
```
henryiii, line 196: I think `B015` helps in cases like this. It clarifies the test is really about the call, and not the comparison, which never happens.

PR #2734, opened 2020-12-21
ci: update cmake action

henryiii:
## Description

Manually updating this instead of waiting for dependabot because it fixes the issue with cmake 3.19.2 on macOS.

## Suggested changelog entry:

None needed. CI.


YannickJadoul (approved):
Woohoo! Merge that thing :-)

PR #2736, opened 2020-12-22
docs: fix pdf build

henryiii:
## Description

Docs are currently broken when generating the PDF. This fixes it and updates
the doc build a bit.

This is fixed by removing the header of the README (the svg to PDF on the second line of badges was broken - didn't seem to matter what badges were there, just having the second line broke it. Badges aren't useful in a PDF anyway, so dropped them.

## Suggested changelog entry:

None needed, docs fix for breaking after 2.6.1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,6 +11,8 @@
 • `Scikit-build example <https://github.com/pybind/scikit_build_example>`_
 • `CMake example <https://github.com/pybind/cmake_example>`_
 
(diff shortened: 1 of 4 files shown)


rwgk (approved):
Looks good, thanks! (Took me a while to see that "removing the header of the README" isn't literal; nice trick with the `.. start`.)

PR #2738, opened 2020-12-23
CONTRIBUTING: Add suggestion about passing pytest flags

EricCousineau-TRI:
## Description

I usually inject things like `-x` or `-s` into `tests/CMakeLists.txt`, so this is a simple addition to allow it to be used via config.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -126,13 +126,25 @@ cmake --build build --target check
 `--target` can be spelled `-t` in CMake 3.15+. You can also run individual
 tests with these targets:
 
-* `pytest`: Python tests only
+* `pytest`: Python tests only, using the
+[pytest](https://docs.pytest.org/en/stable/) framework
 * `cpptest`: C++ tests only
 * `test_cmake_build`: Install / subdirectory tests
 
 If you want to build just a subset of tests, use
 `-DPYBIND11_TEST_OVERRIDE="test_callbacks.cpp;test_pickling.cpp"`. If this is
 empty, all tests will be built.
+
+You may also pass flags to the `pytest` target by editing `tests/pytest.ini` or
+by using the `PYTEST_ADDOPTS` environment variable
+(see [`pytest` docs](https://docs.pytest.org/en/2.7.3/customize.html#adding-default-options)). As an example:
+
+```bash
+env PYTEST_ADDOPTS="--capture=no --exitfirst" \
+    cmake --build build --target pytest
+# Or using abbreviated flags
+env PYTEST_ADDOPTS="-s -x" cmake --build build --target pytest
+```
 
 ### Formatting
 


EricCousineau-TRI:
@henryiii Might you be able to review this one when you have a chance?

henryiii (approved):
Don't see any issue with this, the only other way to get flags in would be to edit the pytest config file, AFAIK. By the way, though, a space separated list should be fine here too, since you don't put quotes around the variable when you use it, it works with spaces.

tests/CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ set(PYBIND11_TEST_OVERRIDE
 set(PYBIND11_TEST_FILTER
     ""
     CACHE STRING "Tests from ;-separated list of *.cpp files will be removed from all tests")
+set(PYBIND11_PYTEST_FLAGS
+    ""
+    CACHE STRING "Flags from ;-separatedlist to pass to pytest")
```
henryiii, line 65:
```suggestion
    CACHE STRING "Flags from ;-separated list to pass to pytest")
```

Check with space separated list, I believe that works too.

YannickJadoul:
> Don't see any issue with this, the only other way to get flags in would be to edit the pytest config file, AFAIK.

I tend to just use `PYTEST_ADDOPTS='--foo' make pytest` (or ``PYTEST_ADDOPTS='--foo' cmake --build build --target pytest`, to make @henryiii happy).

Not against this PR, but any idea how this differs? I guess this makes extra flags persist/settable by default in your CMake config?

henryiii:
Ahh, I was thinking there was a way to pass in flags via variable, but couldn't find it in a lightning search. @EricCousineau-TRI , would this work for you? We can add a note in the contributing docs instead of adding a new option?

If you want a persistent setting, you should probably edit pytest.ini (unless that gets overridden when you regenerated the config, which is possible).

EricCousineau-TRI:
Ooh, yeah, both of those (envvar or ini settings) sounds like a better option to me!

Will update this PR to just be a doc suggestion tracing to those.

EricCousineau-TRI:
Updated!

henryiii:
Dropped the changelog entry since those will show up via our collection script. :)

EricCousineau-TRI:
Thank y'all!

YannickJadoul:
OK, as you wish :-)

I could perhaps see some value for a CMake var, as well, but if you're happy with this, even better!

henryiii:
I’d rather be frugal with custom vars. Something that works everywhere is better. :)

YannickJadoul:
OK, great! Just didn't mean to shut down @EricCousineau-TRI's PR :-)

EricCousineau-TRI:
Agreed! And no worries - PR wasn't shutdown, just optimized ;) Thanks!

PR #2739, opened 2020-12-23
enum: Add Enum.value property

EricCousineau-TRI:
## Description

Towards #2332 - adds a `.value` attribute

## Suggested changelog entry:

```rst
enum: add missing Enum.value property
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -46,6 +46,7 @@
 #include "options.h"
 #include "detail/class.h"
 #include "detail/init.h"
+#include "detail/typeid.h"
 
 #include <memory>
 #include <vector>
@@ -1569,6 +1570,23 @@ inline str enum_name(handle arg) {
     return "???";
 }
 
+template <typename Type, typename Scalar, bool is_convertible = true>
+struct enum_value {
+    static Scalar run(handle arg) {
+        Type value = pybind11::cast<Type>(arg);
+        return static_cast<Scalar>(value);
+    }
+};
+
+template <typename Type, typename Scalar>
+struct enum_value<Type, Scalar, false> {
+    static Scalar run(handle) {
+      throw pybind11::cast_error(
+            "Enum for " + type_id<Type>() + " is not convertible to " +
+            type_id<Scalar>());
+    }
+};
+
 struct enum_base {
     enum_base(handle base, handle parent) : m_base(base), m_parent(parent) { }
 
@@ -1728,7 +1746,13 @@ public:
       : class_<Type>(scope, name, extra...), m_base(*this, scope) {
         constexpr bool is_arithmetic = detail::any_of<std::is_same<arithmetic, Extra>...>::value;
         constexpr bool is_convertible = std::is_convertible<Type, Scalar>::value;
+        auto property = handle((PyObject *) &PyProperty_Type);
         m_base.init(is_arithmetic, is_convertible);
+
+        attr("value") = property(
+            cpp_function(
+                &detail::enum_value<Type, Scalar, is_convertible>::run,
+                pybind11::name("value"), is_method(*this)));
 
         def(init([](Scalar i) { return static_cast<Type>(i); }), arg("value"));
         def("__int__", [](Type value) { return (Scalar) value; });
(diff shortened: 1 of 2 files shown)


include/pybind11/pybind11.h
```diff
@@ -1731,6 +1732,7 @@ template <typename Type> class enum_ : public class_<Type> {
         m_base.init(is_arithmetic, is_convertible);
 
         def(init([](Scalar i) { return static_cast<Type>(i); }), arg("value"));
+        def_property("value", [](Type value) { return (Scalar) value; }, nullptr);
```
bstaletic, line 1735: Just use `def_property_readonly`. No need to introduce `def_property`.

bstaletic (commented):
This looks fine in general. However, shouldn't we be looking at python3.4 enums, as in https://github.com/pybind/pybind11/pull/2704 ? Yes, "what about python2" is a good argument, but even `pip` is dropping support for python2 in January.

EricCousineau-TRI:
> However, shouldn't we be looking at python3.4 enums, as in #2704 ?

Fancy! Didn't see that PR (should've searched lol)
The main concern I have is ensuring we maintain the same semantics as what we currently have - i.e. strict increase in API / capability. The solution space from @AWhetter / @aldnanor / @anntzer entail doing a sort-of `stl_bind.h` / macro-`type_caster` approach. That's totes an option, but I'm concerned about how much it deviates from the existing RTTI-based registration based in `py::class_` (more concretely, I dunno how to do that well, esp. when having a pure Python base *and* meta-class that we may not explicitly control).
(If y'all think I misread the solution, please let me know!)

I do believe I have solution to accommodate the intent of #2704, but while still maintaining the current setup for `py::enum_`. I'll be filing that shortly (read: within the next week).

> Yes, "what about python2" is a good argument, but even pip is dropping support for python2 in January.

I'd be totes down for ditching Python 2!
However, per the internal pybind11 Slack discussions, I'm unaware of explicit plans to cut Python 2 out in the future. I think the consensus was, "once CI becomes a P.I.T.A", which may not happen if we pin an older version of `pip` in that CI config?

Also, FWIW, I don't think it's a huge deal to conditionally support certain aspects in Python 2. I think it'd be easy to transparently  make the Python 3 flavor behave like `Enum`, and possibly extend to the `Flag` flavors. I think, as the current `py::enum_` implementation is defined, `IntFlag` will be implicitly possibly (i.e. ensuring closure of the types across bitwise operations).

include/pybind11/pybind11.h
EricCousineau-TRI: Done.

EricCousineau-TRI:
Er, also - I shoulda put that design comment in the issue.

Moving it there, so further discussion doesn't block this relatively simple PR.

EricCousineau-TRI:
@bstaletic Any chance you got a brief few min to approve the latest changes here?

bstaletic (approved):
Sorry for the delay. This looks perfectly fine.

YannickJadoul (approved):
Looks good! Thanks, @EricCousineau-TRI!

Are you also planning to tackle the `__iter__` part of #2332?

YannickJadoul:
Let's see if this still makes it into 2.6.2? I guess it could conflict with users' code when they already defined `name()`, though, but ... shouldn't break stuff?

YannickJadoul:
> Are you also planning to tackle the `__iter__` part of #2332?

I should read my emails better, and notice #2744...

EricCousineau-TRI:
> Let's see if this still makes it into 2.6.2? I guess it could conflict with users' code when they already defined name(), though, but ... shouldn't break stuff?

Hopefully it's sufficiently benign! It does modify the class defined by `py::enum_<CppEnum>`, so it could interrupt downstream code. But hopefully,that's not too much of an issue (given that `Enum.name` already existed).

My vote is, onwards! It cool if I merge?! :grimacing:

YannickJadoul:
As far as I'm concerned, yes :-) 

Should we quickly check with @henryiii on the 2.6.2 thing; he's been unofficially managing the last few releases?

EricCousineau-TRI:
Sounds good! Henry, you cool with pushing the merge button on this one?

henryiii:
I've edited the changelog a bit to make it sound a bit more like a patch-worthy addition. ;)

YannickJadoul:
Thanks for picking this up, @EricCousineau-TRI!

EricCousineau-TRI:
Sweet, thank y'all for reviewing 'n merging!

PR #2741, opened 2020-12-23
Update pinned pytest version to 6.2.1 and fix error in test_exceptions.py::test_python_alreadyset_in_destructor

YannickJadoul:
## Description

As noted in https://github.com/pybind/pybind11/pull/2573#issuecomment-749916541, pytest 6 seems to make one of our tests fail. Let's check this, and then figure out if/how to fix that.

## Suggested changelog entry:

```rst
* Fix issue with a test failing on PyTest 6.2.
```


henryiii:
I think we just need to add a marker to turn this extra pytest feature off for this test. We do want this behavior.

I noticed the dropping of 3.5 in the changelog.

YannickJadoul:
It would be nice if we could explicitly turn it off for just that test, though, and not for everything, with the blunt `-p no:unraisableexception` flags?

`@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")` ignores the error, but still fails our test. Presumably because pytest holds `sys.unraisablehook` hostage, and doesn't propagate the behavior to the default implemenation, on which we depend in the test.

bstaletic:
Can we use the `monkeypatch` fixture to force our hook and then restore pytest's hook?

YannickJadoul:
Got something. The original built-in `sys.unraisablehook` is kept as `sys.__unraisablehook__`, so we can just use that?

henryiii:
Why make it a string but still error out of unfound... needs an if to see if that exception exists.

tests/test_exceptions.py
```diff
@@ -50,13 +50,23 @@ def test_python_call_in_catch():
     assert d["good"] is True
 
 
+if hasattr(pytest, "PytestUnraisableExceptionWarning"):  # Python >= 3.8 and pytest >= 6
+    ignore_pytest_unraisable_warning = pytest.mark.filterwarnings(
+        "ignore::pytest.PytestUnraisableExceptionWarning"
+    )
+else:
+    ignore_pytest_unraisable_warning = lambda f: f  # NOQA: E731
```
YannickJadoul, line 58: Yes, I could not put a lambda, and put 2 lines `def ignore_pytest_unraisable_warning(f):` and `    return f`, but then `black` suddenly adds an empty line after `else:`, so that's just as ugly.

YannickJadoul:
> Why make it a string but still error out if unfound, PyTest... needs an if to see if that exception exists.

Yeah, indeed ... :-/ Guess they're just following the `warnings` module?

tests/test_exceptions.py
```diff
@@ -50,13 +50,23 @@ def test_python_call_in_catch():
     assert d["good"] is True
 
 
+if hasattr(pytest, "PytestUnraisableExceptionWarning"):  # Python >= 3.8 and pytest >= 6
```
YannickJadoul, line 53: Alternatively, we move this into the function, and directly use the `warnings` module. But I'm not sure whether such a filter would get reset after the test or not, this way.

YannickJadoul:
Green! :-) Ready to be reviewed.

tests/test_exceptions.py
```diff
@@ -50,13 +50,23 @@ def test_python_call_in_catch():
...
+    ignore_pytest_unraisable_warning = pytest.mark.filterwarnings(
+        "ignore::pytest.PytestUnraisableExceptionWarning"
+    )
+else:
+    ignore_pytest_unraisable_warning = lambda f: f  # NOQA: E731
+
+
+@ignore_pytest_unraisable_warning
 def test_python_alreadyset_in_destructor(monkeypatch, capsys):
```
bstaletic, line 62:
```python
def ignore_pytest_unraisable_warning():
    if hasattr(pytest, "PytestUnraisableExceptionWarning"):  # Python >= 3.8 and pytest >= 6
        return pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
    else:
        return lambda f: f

@ignore_pytest_unraisable_warning()
def test_python_alreadyset_in_destructor(monkeypatch, capsys):
```

That's if you want to be fancy.

bstaletic (commented):
Like I said "a decorator factory"  could be a fancy way of avoiding code in global scope, but not a blocker.

tests/test_exceptions.py
YannickJadoul: Does this have an advantage, moving things into a function?

bstaletic: The only advantage, besides personal dislike for procedural global code, that I can think of is delayed evaluation of those 4 lines, which should speed up importing of the `test_exceptions` module. I'm not saying the difference would be significant.

henryiii: I'm _mildly_ in favor of this, but would be happy either way. Though the parens seem odd. Why not:

```python
def ignore_pytest_unraisable_warning(f):
    unraisable = "PytestUnraisableExceptionWarning"
    if hasattr(pytest, unraisable):  # Python >= 3.8 and pytest >= 6
        dec = pytest.mark.filterwarnings("ignore::pytest.{}".format(unraisable))
        return dec(f)
    else:
        return f
```

That's the blacked code.

YannickJadoul: OK, yes, that looks better!

> Why not

Because I had decided the line wasn't too long, but black didn't agree and "fixed" it to this abomination.

YannickJadoul: Thanks, @henryiii! That looks much better :-)

tests/test_exceptions.py
```diff
@@ -50,13 +50,24 @@ def test_python_call_in_catch():
...
     hooked = False
     triggered = [False]  # mutable, so Python 2.7 closure can modify it
 
     if hasattr(sys, "unraisablehook"):  # Python 3.8+
         hooked = True
-        default_hook = sys.unraisablehook
+        # Don't take `sys.unraisablehook`, as that's overwritten by pytest
+        default_hook = sys.__unraisablehook__
```
YannickJadoul, line 70: I take it we're happy with this?

If some case where this would be overwritten (by something else than pytest), we're ignoring it as well, ofc, but ... myeah, best I could immediately think of.

henryiii: Yes, it seems fine to me. We don't want pytest's version here, and it's a test, so nothing else should be fiddling with it.

tests/test_exceptions.py
henryiii: If you have code like this:
```python
my_big_function_that_is_so_long("my_big_string_that_is_so_long")
```
And then black "fixes" it because it's too long, then write it like this:
```python
msg = "my_big_string_that_is_so_long"
my_big_function_that_is_so_long(msg)
```

Now black is happy, and it's more readable too. If you depend on the line length, say by tiling your monitor with windows, you don't want a random line poking off. And if you don't really mean the line length limit, change it - that's one of the (~2?) settings black allows. But I think this promotes better, more readable code (and I like 0 configuration - a so-so standard is better than perfect but non-standard).

YannickJadoul: That's true. Let's just say I have not a lot of experience with forcibly keeping lines short. But this works, indeed.

YannickJadoul:
Does this need to be marked 2.6.2, still? I guess it's kind of independent from the releases, but ...

Guess it's mostly #2573 that needs to make sure it also installs `tests/requirements.txt`, as I noted there.

henryiii (approved):

henryiii:
Great! We don't really control the PyTest version if someone wants to run our tests (a requirements.txt is an internal pinning, not an external one), and a few of our tests use system PyTests, but it's good not to be (too) version specific, tests failing due to PyTest is not ideal.

YannickJadoul:
Do you want a changelog for this, @henryiii?

henryiii:
Yes, I added it to the description.

Here it is in setuptools changelog, for example: https://setuptools.readthedocs.io/en/latest/history.html#misc

YannickJadoul:
> Yes, I added it to the description.

Great! Hadn't spotted that; thanks! :-)

Issue #1742, opened 2019-03-26
Make it possible to provide scope to/store scope of Python function


ingomueller-net:
## Issue description

A lambda function constructed with a certain scope (`local` and `global`) from within `py::eval` cannot be called, as the scope is unavailable when the function is called.

## Reproducible example code

```c++
#include <pybind11/embed.h>

namespace py = pybind11;

int main(int, char**) {
    py::scoped_interpreter guard{};
    auto global = py::dict(py::module::import("__main__").attr("__dict__"));
    auto local = py::dict();
    local["x"] = 42;
    py::function f = py::eval("lambda: x", global, local);
    py::print("defined f");
    f();
    py::print("called f");
}
```

Output:

```
defined f
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  NameError: name 'x' is not defined
```

## Initial discussion

The problem is that the local variable `a` is not available when `f` is called; in fact, `f` does not seem to have a scope.

Note that the following works in Python:

```Python
def make_f():
    a=42
    return eval('lambda: a')
a=123
print(make_f()())   # outputs 123
```

I have no idea how, but it would be great if there was a way to provide a scope to `py::function`s to make above example work.

ingomueller-net:
I think I found a work-around:

```c++
namespace py = pybind11;
auto global = py::dict(py::module::import("__main__").attr("__dict__"));
auto local = py::dict();
local["x"] = 42;
local["f"] = py::eval("lambda: x");
py::print("defined f");
py::eval("f", global, local);
py::print("called f");
```

This produces the following output as desired:

```
defined f
called f
```

Unless you have a more elegant solution or think that there should be one, you can close this issue.

ingomueller-net:
I tried many things and nothing worked. I must be missing some fundamental understanding of what locals and globals are.

One thing worked, though: making `x` global, i.e., saying `globals["x"] = 42`, or, more generally:

```c++
global.attr("update")(local);
```

I'd still be glad to here why this is happening, though.

eacousineau:
I think the core of this is due to the, uh, interesting nuance of how Python's scoping for `globals` `locals` works with `eval + lambda`:
https://stackoverflow.com/questions/50579541/why-lambda-in-eval-function-cant-closure-the-variables-in-the-user-defined-loc

Modifying your above code corroborates this:
https://github.com/eacousineau/repro/blob/3852e054da9489b132852e34898958e7b51706c0/python/lambda_eval_scope.py#L19

[...]

PR #2743, opened 2020-12-24
test_eval: Show example of working closure

EricCousineau-TRI:
## Description

Shows an example unittest with using `py::exec` with a closure.
I had forgotten what was necessary, then came across #1742 again. I'm adding an example here just to show it.


tests/test_eval.cpp
```diff
@@ -96,4 +97,15 @@ TEST_SUBMODULE(eval_, m) {
...
     });
+
+    m.def("test_eval_closure", [global, copy]() {
+        py::dict local = copy(global);
+        py::exec(R"(
+            closure_value = 10
+
+            def func():
+                return closure_value
+            )", local, local);
```
YannickJadoul, line 108: Why is `local` being passed twice here? And from reading #1742, shouldn't this be called `global` as it's the globals dict?

YannickJadoul (commented):
Ah, cool! I knew things were annoying and complicated, but not _this_ annoying and complicated. Good to have a test on this to remind ourselves!

YannickJadoul:
Suggestion to expand this test, to show differences between what's captured and what's not capture with locals vs. globals:

```c++
    // test_eval_closure
    m.def("test_eval_closure", []() {
        py::dict global;
        global["closure_value"] = 42;
        py::dict local;
        local["closure_value"] = 0;
        py::exec(R"(
            local_value = closure_value

            def func_global():
                return closure_value

            def func_local():
                return local_value
            )", global, local);
        return std::make_pair(global, local);
    });
```

and

```py
def test_eval_closure():
    global_, local = m.test_eval_closure()

    assert global_["closure_value"] == 42
    assert local["closure_value"] == 0

    assert "local_value" not in global_
    assert local["local_value"] == 0

    assert "func_global" not in global_
    assert local["func_global"]() == 42

    assert "func_local" not in global_
    with pytest.raises(NameError):
        local["func_local"]()
```

Maybe I'm misunderstanding/butchering the original aim of this test & PR, though? Please do tell me, then, if this extension is irrelevant or obfuscating to the closure construction you're trying to test, @EricCousineau-TRI!

tests/test_eval.cpp
```diff
@@ -96,4 +97,15 @@ TEST_SUBMODULE(eval_, m) {
         auto int_class = py::eval("isinstance(42, int)", global);
         return global;
     });
+
+    m.def("test_eval_closure", [global, copy]() {
```
YannickJadoul, line 101: Maybe also add the `// test_eval_closure` comment above. In this case, it's a bit over the top since your function is already called that, but I like the existing convention of indicating the Python-side tests that C++-side bindings belong to.

YannickJadoul:
> expand this test

If you like the idea, I have this on my git's stash, btw, so I can just commit and push, if that saves trouble.

EricCousineau-TRI:
> If you like the idea, I have this on my git's stash, btw, so I can just commit and push, if that saves trouble.

Ooh, yeah, that'd be great! Though understandable if it's gotten lost in your stash after a month :sweat_smile: 
Just lemme know!

YannickJadoul:
Luckily I hardly ever clear out that thing, so it was still somewhere down ;-)
(Also lucky you don't need to pop the stash like a true stack :-P )

Skylion007 (approved):

henryiii (approved):

rwgk (approved):
Just documenting that I looked, too :-)

PR #2744, opened 2020-12-24
enum: Add iteration and simple type-based check, towards PEP 435

### LINKED ISSUE #2332: Why pybind11 generated enum type is not PEP 435 compatible. say, it doesn't have value attribute and it's not iterable.

EricCousineau-TRI:
## Description

(Currently draft)
Resolves #2332

Duck-type, but does not support `isinstance(pybind11_enum_value, enum.Enum)`.

## Suggested changelog entry:

```rst
enum: Add iteration and simple type-based check, towards PEP 435
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1569,6 +1569,59 @@ inline str enum_name(handle arg) {
     return "???";
 }
 
+class enum_meta_info {
+public:
+    static pybind11::object enum_meta_cls() {
+        return get().enum_meta_cls_;
+    }
+
+    static pybind11::object enum_base_cls() {
+        return get().enum_base_cls_;
+    }
+
+private:
+    template <typename T>
+    friend T& pybind11::get_or_create_shared_data(const std::string&);
+
+    static const enum_meta_info& get() {
+        return pybind11::get_or_create_shared_data<enum_meta_info>(
+            "_pybind11_enum_meta_info");
+    }
+
+    enum_meta_info() {
+        handle copy = pybind11::module::import("copy").attr("copy");
+        locals_ = copy(pybind11::globals());
+        locals_["pybind11_meta_cls"] = reinterpret_borrow<object>(
+            reinterpret_cast<PyObject*>(get_internals().default_metaclass));
+        locals_["pybind11_base_cls"] = reinterpret_borrow<object>(
+            get_internals().instance_base);
+        // TODO: Make the base class work.
+        const char code[] = R"""(
+pybind11_enum_base_cls = None
+
+class pybind11_enum_meta_cls(pybind11_meta_cls):
+    is_pybind11_enum = True
+
+    def __iter__(cls):
+        return iter(cls.__members__.values())
+
+    def __len__(cls):
+        return len(cls.__members__)
+)""";
+        PyObject *result = PyRun_String(
+            code, Py_file_input, locals_.ptr(), locals_.ptr());
+        if (result == nullptr) {
+            throw error_already_set();
+        }
+        enum_meta_cls_ = locals_["pybind11_enum_meta_cls"];
+        enum_base_cls_ = locals_["pybind11_enum_base_cls"];
+    }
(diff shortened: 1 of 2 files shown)


tests/test_enum.py
```diff
@@ -1,7 +1,44 @@
...
+    assert list(cls) == list(values)
+    assert is_enum(cls)
+    if enum:
+        assert not issubclass(cls, enum.Enum)
+    for name, value, raw_value in zip(names, values, raw_values):
+        assert isinstance(value, cls)
+        if enum:
+            assert not isinstance(value, enum.Enum)
+        assert value.name == name
+        assert value.value == raw_value
```
EricCousineau-TRI, line 40: Self-note: Should try to implement `Flag` based interface. (i.e. closure under arithmetic operations)

include/pybind11/pybind11.h
```diff
@@ -1569,6 +1569,59 @@ inline str enum_name(handle arg) {
...
+pybind11_enum_base_cls = None
+
+class pybind11_enum_meta_cls(pybind11_meta_cls):
+    is_pybind11_enum = True
+
+    def __iter__(cls):
+        return iter(cls.__members__.values())
+
+    def __len__(cls):
+        return len(cls.__members__)
```
sizmailov, line 1609: > does not support isinstance(pybind11_enum_value, enum.Enum)

Is there a particular reason to not implement `__instancecheck__` here?

sizmailov: Sorry, didn't realize this should be done in `enum.EnumMeta`, not here. :facepalm: 

Is there a way to do it without altering builtin classes?

YannickJadoul:
Urgh, I didn't realize this would be so much harder than it seems, when reading the original comments in #2332 :-|

include/pybind11/pybind11.h
EricCousineau-TRI: Not that I'm aware of :( I had gone the route of modifying builtin classes, but *yuck*:
https://github.com/pybind/pybind11/compare/master...a41944c57509fa3845ef190a84972811a87eac80

EricCousineau-TRI:
> Urgh, I didn't realize this would be so much harder than it seems [...]

Yeah lol, gotta get them meta classes in :(

include/pybind11/pybind11.h
sizmailov: I agree, it does look too hacky and doesn't sound like right thing to do. 

Actually pybind-enum is NOT `enum.Enum`, so maybe we shouldn't try to make this check return true? I think it would be more appropriate to introduce `collections.abc.Enum` (e.g. in python 3.10) and make `isinstance(pybind_enum_value, colletions.abc.Enum)` recognize all PEP-complaining enum implementations? In meantime one probably have no choice other than stick to self-written check method (like one in tests in this PR). 

I think such doubtful hacks would greatly reduce the PR merge chances, so I would leave attempts to introduce them, at least for now.

Issue #2722, opened 2020-12-10
[BUG] Error printing objects with pprint in python3


beatmax:
## Issue description

Objects of classes created with pybind11 can't be pretty printed with the pprint module if the result is larger than the given width.
This only happens with python3.

It used to happen with cython too, but cython stopped using 'instancemethod', which solved the issue. The pprint bug was discarded because "Functions/methods should be immutable, so yes, I think it is a safe assumption that they should be hashable":

https://bugs.python.org/issue33395

## Reproducible example code

example.cpp:
```
#include <pybind11/pybind11.h>
#include <string>

namespace py = pybind11;

struct Dummy
{
  std::string text;
};

PYBIND11_MODULE(example, m) {
  using namespace pybind11::literals;

  py::class_<Dummy>(m, "Dummy")
      .def(py::init<std::string>(), "text"_a)
      .def_readonly("text", &Dummy::text)
      .def("__str__", [](const Dummy& w) { return w.text; })
      .def("__repr__", [](const Dummy& w) { return w.text; });
}
```

bug.py:
```
import example
import pprint

pprint.pprint(example.Dummy('abc'), width=3)
```

Result:
```
Traceback (most recent call last):
  File "bug.py", line 4, in <module>
    pprint.pprint(example.Dummy('abc'), width=2)
  File "/usr/lib/python3.8/pprint.py", line 53, in pprint
    printer.pprint(object)
  File "/usr/lib/python3.8/pprint.py", line 148, in pprint
    self._format(object, self._stream, 0, 0, {}, 0)
  File "/usr/lib/python3.8/pprint.py", line 173, in _format
    p = self._dispatch.get(type(object).__repr__, None)
TypeError: unhashable type: 'instancemethod'
```

YannickJadoul:
That's very unfortunate. Seems that the [`__hash__` for `instancemethod`](https://github.com/python/cpython/blob/7cf0aad96d1d20f07d7f0e374885f327c2d5ff27/Objects/classobject.c#L585) is still commented out. Not sure why; [bpo-33395](https://bugs.python.org/issue33395) doesn't seem to give a good answer?

I'm not sure how pybind11 is to blame for this bug, and not even how pybind11 could fix this, though. What did you have in mind?

beatmax:
Hi Yannick, since they stopped using instancemethod in cython, which solved they issue, maybe the same can be done in pybind11? (See https://github.com/cython/cython/pull/2105 which is mentioned in the pprint bug).

[...]

PR #2745, opened 2020-12-25
Do not set docstring for function when it's empty

lqf96:
## Description

In pybind11, automatic docstring generated from function signature and user-provided documents can be suppressed by disabling all docstring-related options. However, this currently produces an empty docstring, which is unnecessary and interferes with situations where users want to manually add docstring to the bound function later directly through the Python C API (for example, [PyTorch currently does this to its manual C bindings](https://github.com/pytorch/pytorch/blob/master/torch/csrc/Module.cpp#L221) and I find it difficult to extend this to work with pybind11-bound functions without touching pybind11's implementation details). This pull request modifies this behavior, such that when all docstring-related options are disabled and the resulting docstring is empty, no docstring is set on the function.

## Suggested changelog entry

```rst
When all docstring-related options are disabled and the resulting docstring is empty, no docstring is now set on the function. Previously, pybind11 sets an empty docstring.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -439,9 +439,14 @@ protected:
 
         /* Install docstring */
         auto *func = (PyCFunctionObject *) m_ptr;
-        if (func->m_ml->ml_doc)
+        // Free and reset previous docstring if it exists
+        if (func->m_ml->ml_doc) {
             std::free(const_cast<char *>(func->m_ml->ml_doc));
-        func->m_ml->ml_doc = strdup(signatures.c_str());
+            func->m_ml->ml_doc = nullptr;
+        }
+        // Then install new one if it's non-empty (when at least one option is enabled)
+        if (!signatures.empty())
+            func->m_ml->ml_doc = strdup(signatures.c_str());
 
         if (rec->is_method) {
             m_ptr = PYBIND11_INSTANCE_METHOD_NEW(m_ptr, rec->scope.ptr());
@@ -472,7 +477,9 @@ protected:
                 arg.value.dec_ref();
             }
             if (rec->def) {
-                std::free(const_cast<char *>(rec->def->ml_doc));
+                if (rec->def->ml_doc) {
+                    std::free(const_cast<char *>(rec->def->ml_doc));
+                }
                 // Python 3.9.0 decref's these in the wrong order; rec->def
                 // If loaded on 3.9.0, let these leak (use Python 3.9.1 at runtime to fix)
                 // See https://github.com/python/cpython/pull/22670
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -440,10 +440,8 @@ protected:
         /* Install docstring */
         auto *func = (PyCFunctionObject *) m_ptr;
         std::free(const_cast<char *>(func->m_ml->ml_doc));
-        func->m_ml->ml_doc = nullptr;
         // Install docstring if it's non-empty (when at least one option is enabled)
-        if (!signatures.empty())
-            func->m_ml->ml_doc = strdup(signatures.c_str());
+        func->m_ml->ml_doc = signatures.empty() ? nullptr : strdup(signatures.c_str());
 
         if (rec->is_method) {
             m_ptr = PYBIND11_INSTANCE_METHOD_NEW(m_ptr, rec->scope.ptr());
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -439,12 +439,9 @@ protected:
 
         /* Install docstring */
         auto *func = (PyCFunctionObject *) m_ptr;
-        // Free and reset previous docstring if it exists
-        if (func->m_ml->ml_doc) {
-            std::free(const_cast<char *>(func->m_ml->ml_doc));
-            func->m_ml->ml_doc = nullptr;
-        }
-        // Then install new one if it's non-empty (when at least one option is enabled)
+        std::free(const_cast<char *>(func->m_ml->ml_doc));
+        func->m_ml->ml_doc = nullptr;
+        // Install docstring if it's non-empty (when at least one option is enabled)
         if (!signatures.empty())
             func->m_ml->ml_doc = strdup(signatures.c_str());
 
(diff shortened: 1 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -472,7 +477,9 @@ class cpp_function : public function {
                 arg.value.dec_ref();
             }
             if (rec->def) {
-                std::free(const_cast<char *>(rec->def->ml_doc));
+                if (rec->def->ml_doc) {
```
YannickJadoul, line 480: In principle, no braces necessary (cfr. other parts of this code); fine to keep them, though.

YannickJadoul: Actually, `std::free` can be called on a `nullptr`: https://en.cppreference.com/w/cpp/memory/c/free says

> If ptr is a null pointer, the function does nothing.

So if anything, I'd put a comment that it could be `nullptr`, but I don't think that's necessary.

YannickJadoul (approved):
> interferes with situations where users want to manually add docstring to the bound function later directly through the Python C API

Could you elaborate on this? How exactly does it interfere?

Just curious to know what you mean. Apart from that, obviously, I'm not against not doing work when it's not necessary, so this looks good to me :-)

> I find it difficult to extend this to work with pybind11-bound functions without touching pybind11's implementation details

Agreed. I hope that on the medium/long-term, pybind11 would allow a more fine-grained control over customizing this.

include/pybind11/pybind11.h
YannickJadoul: Extra: potentially, you could reset to `nullptr` after freeing, to easy debugging when things go wrong. But it wasn't there, so I don't think this PR per se _needs_ to do that.

lqf96:
> > interferes with situations where users want to manually add docstring to the bound function later directly through the Python C API
> 
> Could you elaborate on this? How exactly does it interfere?
> 
> Just curious to know what you mean. Apart from that, obviously, I'm not against not doing work when it's not necessary, so this looks good to me :-)

I have been working on a series of pull requests for PyTorch and one of them (pytorch/pytorch#49771) is replacing legacy manual bindings of `torch.Generator` (PyTorch's equivalent of `random.Random` and `numpy.random.RandomState`) with pybind11 bindings. The legacy bindings does not have any docstrings associated with them when they are created using the Python C API. Rather, it is added later through an internal function (`torch._C._add_docstr`), which [sets the docstring for binding functions or properties from the C side](https://github.com/pytorch/pytorch/blob/master/torch/csrc/Module.cpp#L221). In this way all docstrings can be [kept in a separated Python file](https://github.com/pytorch/pytorch/blob/master/torch/_torch_docs.py) and added to the APIs during initialization.

Previously, `torch._C._add_docstr` only handles manual bindings. When I'm trying to extend this function to work with pybind11 bindings, I find out that even if all docstring options are disabled, the resulting `instancemethod` would have an empty docstring. To replace these docstrings with PyTorch ones, I need to first free the existing `ml_doc` are then assign it with a copy of the PyTorch docstring created by `strdup`, otherwise there will be memory leaks. However, the way `ml_doc` is allocated and released by pybind11 is clearly implementation detail, and by doing so I'm utilizing this implementation detail which is not a good thing.

With this pull request, the `ml_doc` will be empty and I no longer need to free the existing docstring. However, to set the `ml_doc` field I still need to `strdup` PyTorch-provided docstrings, so the problem of using implementation detail is not totally resolved. I plan to come up with another PR to acknowledge this. I also feel that even without all the context, this is still a natural thing to do because in pure Python if you don't set the docstring of a function, the `__doc__` field will be `None`, and this PR matches that behavior exactly.

YannickJadoul:
That's very elaborate, thanks! :-)

> I need to first free the existing `ml_doc` are then assign it with a copy of the PyTorch docstring created by `strdup`, otherwise there will be memory leaks.

And yep, I had completely forgotten about this! Ofc, you could set the `__doc__` Python attribute (which is what I thought you were doing), but that would probably be a lot less efficient :-)

> However, the way `ml_doc` is allocated and released by pybind11 is clearly implementation detail, and by doing so I'm utilizing this implementation detail which is not a good thing.

Yep, though expecting that pybind11 will create `PyCFunctions` might also be considered an implementation detail? One of the roads to fixing #2722 might involve steering away from those; but nothing's been decided or even seriously tried, though.

lqf96:
> And yep, I had completely forgotten about this! Ofc, you could set the `__doc__` Python attribute (which is what I thought you were doing), but that would probably be a lot less efficient :-)

Will this is actually impossible :( Trying to set the `__doc__` attribute will result in an error saying that `__doc__` is readonly for `instancemethod`s.

BTW I just pushed another commit to remove the redundant pointer check logic. Now it looks cleaner :)

YannickJadoul:
> > And yep, I had completely forgotten about this! Ofc, you could set the `__doc__` Python attribute (which is what I thought you were doing), but that would probably be a lot less efficient :-)
> 
> Will this is actually impossible :( Trying to set the `__doc__` attribute will result in an error saying that `__doc__` is readonly for `instancemethod`s.

Argh, yes, sorry. I should have known that. Those silly C-function wrappers.

> BTW I just pushed another commit to remove the redundant pointer check logic. Now it looks cleaner :)

I saw; thanks! Looks good, now :-) I had already approved, but I'll invite a few others, to see if we can get one or a few more sanity checks :-)

include/pybind11/pybind11.h
```diff
@@ -439,9 +439,11 @@ class cpp_function : public function {
 
         /* Install docstring */
         auto *func = (PyCFunctionObject *) m_ptr;
-        if (func->m_ml->ml_doc)
-            std::free(const_cast<char *>(func->m_ml->ml_doc));
-        func->m_ml->ml_doc = strdup(signatures.c_str());
+        std::free(const_cast<char *>(func->m_ml->ml_doc));
+        func->m_ml->ml_doc = nullptr;
```
henryiii, line 443: Couldn’t this be in an else branch below the next line? Or maybe oven better, make it a ternary operator?

rwgk (approved):
Looks good to me, as-is or with the ternary operator as suggested by Henry. Thanks Qifan!

lqf96:
Just pushed another commit to address @henryiii 's comment. If there's no more change needed I can squash the commits.

include/pybind11/pybind11.h
lqf96: Yeah, I just made it a ternary operator... and I almost forget that ternary operator is lazily evaluated like the if statement in C++. Haven't been writing C++ for some time...

YannickJadoul:
> If there's no more change needed I can squash the commits.

We'll automatically squash when merging ;-)

@henryiii, if there's nothing left, go ahead and push that button.

henryiii:
We only need nice history if there's a reason to keep more than one commit for a PR, like a large one or one where a file changed names and also was edited. Thanks!

PR #2753, opened 2020-12-29
Add shortcut type::name()

lqf96:
## Description
Add `type::name()` as a shortcut for getting the name of a type, like `cpp_function::name()`. I also feel that maybe we can move `cpp_function::name()` to `function::name()`, as all Python functions have names?

## Suggested changelog entry:
```rst
Add shortcut `type::name()` for getting the name of a type
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -923,6 +923,9 @@ public:
     /// See https://github.com/pybind/pybind11/issues/2486
     template<typename T>
     static type of() {return type(type::handle_of<T>(), borrowed_t{}); }
+
+    /// Return the type name
+    object name() const { return attr("__name__"); }
 };
 
 class iterable : public object {
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -160,6 +160,10 @@ TEST_SUBMODULE(class_, m) {
         return py::type(ob);
     });
 
+    m.def("get_instance_type_name", [](py::object ob) {
+        return py::type::of(ob).name();
+    });
+
     // test_mismatched_holder
     struct MismatchBase1 { };
     struct MismatchDerived1 : MismatchBase1 { };
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -68,6 +68,15 @@ def test_as_type_py():
         assert m.as_type(m.DerivedClass1()) == m.DerivedClass1
 
 
+def test_type_name():
+    class PurePythonClass(object):
+        pass
+
+    assert m.get_instance_type_name(1) == "int"
+    assert m.get_instance_type_name(PurePythonClass()) == "PurePythonClass"
+    assert m.get_instance_type_name(m.DerivedClass1()) == "DerivedClass1"
+
+
 def test_docstrings(doc):
     assert doc(UserType) == "A `py::class_` type for testing"
     assert UserType.__name__ == "UserType"


lqf96:
Question: should we make the return type `object` or `str_attr_accessor`, since for some Python classes their name is writable? I don't have a use case for the latter, and I feel the demand of changing type names might be too niche...

include/pybind11/pytypes.h
```diff
@@ -923,6 +923,9 @@ class type : public object {
     /// See https://github.com/pybind/pybind11/issues/2486
     template<typename T>
     static type of() {return type(type::handle_of<T>(), borrowed_t{}); }
+
+    /// Return the type name
+    object name() const { return attr("__name__"); }
```
YannickJadoul, line 928: I don't see the point of returning an `object`. Then you could just as well use `attr("__name__")`, no?

If you want to add something to the C++ API, why not make it return `std::string`?

YannickJadoul (commented):
> Question: should we make the return type `object` or `str_attr_accessor`, since for some Python classes their name is writable?

If you do that, it should be right above `doc()` in `object_api<T>`.

> Add `type::name()` as a shortcut for getting the name of a type, like `cpp_function::name()`

To be fair, I don't see the point of the `cpp_function::name()` shortcut either. It seems rather silly to return an `object`, rather than a C++ type. Why would you need it, and why exactly is it better than `attr("__name__")`? (It not that often needed, is it?)

include/pybind11/pytypes.h
lqf96: Yeah returning `std::string` is better, since for both types and functions their `__name__` must be string, otherwise Python will throw type errors...

YannickJadoul: Well, I'm not entirely convinced of that either, actually, because it would be inconsistent with `cpp_function::name()`.

lqf96:
TBH the primary motivation for this PR is that I find it really inconvenient to get the type name of a Python object... With Python C APIs I just need to write:

```c++
Py_TYPE(object)->tp_name  // object is a PyObject*
```

But if I'm using pybind11 it becomes:

```c++
py::type::of(object).attr("__name__").cast<std::string>()  // object is a py::object
```

So I agree with @YannickJadoul 's comments that probably `type::name()` should return a `std::string` rather than an object from ergonomic perspective.

YannickJadoul:
Yeah, sorry to be annoying about this. The issue is, once we reserve that name, we're bound to a promised API, so we can't just change our mind afterwards. Next to that, I'm not entirely sure why `cpp_function::name()` is there; I think just for convenience in `def`? (This might be useful historical information?)

So I agree it should probably return `std::string`, but then the other one already returns a `py::object` :-/ So that's where I'm then wondering if this convenience function won't just be confusing? (To play the devil's advocate: it's a few characters more, and you could define easily `type_name = [](py::handle o) { return py::cast<std::string>(py::type::of(o).attr("__name__")); }`) But I agree it could perhaps be useful and would make a nice API.

Also, if we go the `py::object` route, it should probably be in `object_api<T>`.

I'll call in @bstaletic on this. Any ideas, @bstaletic?

lqf96:
Just push another commit that returns type name as a string. I also changed the implementation to directly use Python C API because this can be more efficient. The implementation is copied from CPython's source code.

Personally, I prefer to only have `name()` on `type` and `function`... `type` and `function` are the two most common use cases where we want to access `__name__`, and they restrict `__name__` to strings only (setting `__name__` to anything other than a string results in a type error). This will make it handy to just return `std::string` instead of `py::object`. Also, for both of them there are convenient ways to get `__name__` from the C side, as opposed to doing attribute lookup.

include/pybind11/pytypes.h
lqf96: I think we can remove that one later and replace it with a `function::type()` that also returns `std::string`. It will be an incompatible change though...

YannickJadoul: No, we cannot. That's a breaking change and cannot be done before pybind11 3.0, without getting error reports.

lqf96: How about just leave it as-is for now, and change it once pybind11 reaches 3.0?

YannickJadoul:
I saw, but I'm even less of a fan of copying CPython's internal code; it'll only allow for inconsistencies (also with PyPy), I'd say.

You also haven't addressed the inconsistency with `cpp_function::name()`. Removing or changing that is not an option.

> Also, for both of them there are convenient ways to get `__name__` from the C side, as opposed to doing attribute lookup.

How are you going to do that for `function`? `function` can be anything callable, so doesn't even need to have a `__name__`.

Which is finally a good argument against putting it in `object_api` (rather than "I prefer"): I did fail to notice   `__name__` isn't always set; so forget about that.

include/pybind11/pytypes.h
YannickJadoul: Where it's used, if it's not a `const char *`, it's converted back to a Python object. So it's not as simple as that.

bstaletic:
I'm also not convinced by this whole PR.

> TBH the primary motivation for this PR is that I find it really inconvenient to get the type name of a Python object... With Python C APIs I just need to write:
> 
> ```c++
> Py_TYPE(object)->tp_name  // object is a PyObject*
> ```
> 
> But if I'm using pybind11 it becomes:
> 
> ```c++
> py::type::of(object).attr("__name__").cast<std::string>()  // object is a py::object
> ```

Those two lines don't do nearly the same thing. Further, I'm also not sure whether you're looking for a fully qualified name. The direct "translation of the C API would be:

    reinterpret_cast<PyTypeObject*>(py::type::of(object).ptr())->tp_name;

Conversely, the C API version of what you've written with pybind11 would be (don't do this):

    py::cast<std::string>(PyObject_GetAttrString(Py_TYPE(object), "__name__"));

Note that both `const char*` and `py::str` can be implicitly converted to `std::string`. (The latter might not be the best idea, though...) That way the user has a choice whether the call to this API will allocate. (Yes, `std::string` has small buffer optimization implemented, but that's besides the point.)

I'm also not convinced a few saved are a good motivation.

To make my point clear, I think we need a better motivation for this pull request. On top of that, returning `py::str`, in case we do want this PR seems a better idea.

lqf96:
Hey guys, sorry for the long silence on this PR as I'm dealing with some personal affairs and PyTorch code review. Yeah there's indeed some problems on the API signature and what it should return (`__name__`, `__qualname__` or full name) which I don't think I can come up with a one-size-fit-all solution. I'm thinking about closing this right now but if you do have ideas, just let me know and I'll consider about proceeding with this.

YannickJadoul:
@lqf96 No, I don't immediately know. Looking back now, your original proposal (just an ordinary `attr("__name__")`) might still be the best, but then we still have the issue about assigning to it? The question is also if `object name()` really is so much better/shorter than `attr("__name__")`, but OK, `cpp_function` already has it that way as well, ofc.

I do now agree about the `object_api<T>`, btw: not every object has a `__name__`, so you're right that that doesn't make sense (apologies for overlooking that at first).

PR #2746, opened 2020-12-26
Add a Valgrind build on debug Python 3.9 

YannickJadoul:
## Description

Finally, together with @bstaletic, we've finally taken the time and effort to add a build with Valgrind; currently still failing, because of #2742.
We might also need to suppress some Python errors reported by Valgrind, but let's take it step by step.

Things to fix:
- [x] Plumb or ignore leaks until things pass.
- [x] Shut up warnings coming from the test that are now visible, since we have a debug version of Python.

Things to decide:
- [x] Do we just want to run this on Python 3.9 (probably; according to @bstaletic, earlier versions had issues) and on Linux?
- [x] Do we want to keep the tests run on debug Python first, before running them again with Valgrind?

Things to tweak in CMake/CI config:
- [x] @henryiii, `valgrind cmake --build build --target pytest` doesn't work (we don't want to run Valgrind on CMake). Do we want to include this into the CMake config? ~The fact that `valgrind python` also doesn't work with `pyenv` (it's bash script wrapper/launcher), and we need to use this `$(pyenv which python) ` doesn't make things simpler either.~
- [x] We can probably cache the built version of pyenv's Python build? It just takes 2 minutes, but still, this would make it faster.
- [x] Same for Valgrind, maybe; can we cache this?
- [x] There have to be ways to get a cleaner CI config?
- [x] We should also check out https://pypi.org/project/pytest-valgrind/

Oh, yes, and:
- [x] Undo ed0c494033f0b6999b4df5fa015fccb4200f0b03, before merging, ofc. That's just to make the test runs cleaner :-)

Things for follow-up PRs:
- This should also be run on the embedded tests.
- UBSan, ASan, MSan, TSan, ...?
- Revert `test_dispatch_issue` workarounds and debug and fix the underlying issue.

Closes #2578

## Suggested changelog entry:

```rst
Added a Valgrind build, checking for leaks and memory-related UB, to CI.
```

(@henryiii, is this something we actually want to add to add to the changelog?)

---
This change is [Reviewable](https://reviewable.io/reviews/pybind/pybind11/2746)

PR #2755, opened 2020-12-30
Fix invalid access when reinterpret_casting a non-pybind11 PyObject* to instance* (found by Valgrind in #2746)

YannickJadoul:
## Description

Carved out from #2746 to ease reviewing process.

## Suggested changelog entry:

```rst
Fix invalid access when calling a pybind11 ``__init__`` on a non-pybind11 class instance.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -504,14 +504,14 @@ protected:
 
         auto self_value_and_holder = value_and_holder();
         if (overloads->is_constructor) {
-            const auto tinfo = get_type_info((PyTypeObject *) overloads->scope.ptr());
-            const auto pi = reinterpret_cast<instance *>(parent.ptr());
-            self_value_and_holder = pi->get_value_and_holder(tinfo, false);
-
-            if (!self_value_and_holder.type || !self_value_and_holder.inst) {
+            if (!PyObject_TypeCheck(parent.ptr(), (PyTypeObject *) overloads->scope.ptr())) {
                 PyErr_SetString(PyExc_TypeError, "__init__(self, ...) called with invalid `self` argument");
                 return nullptr;
             }
+
+            const auto tinfo = get_type_info((PyTypeObject *) overloads->scope.ptr());
+            const auto pi = reinterpret_cast<instance *>(parent.ptr());
+            self_value_and_holder = pi->get_value_and_holder(tinfo, true);
 
             // If this value is already registered it must mean __init__ is invoked multiple times;
             // we really can't support that in C++, so just ignore the second __init__.


YannickJadoul:
@rwgk, this PR might benefit from a wider test, but since you already tested #2746, this should have already passed?

Ready for review. Given the subtlety of the actual problem, it would be nice to get a thorough review, here.

include/pybind11/pybind11.h
```diff
@@ -504,15 +504,15 @@ class cpp_function : public function {
 
         auto self_value_and_holder = value_and_holder();
         if (overloads->is_constructor) {
-            const auto tinfo = get_type_info((PyTypeObject *) overloads->scope.ptr());
-            const auto pi = reinterpret_cast<instance *>(parent.ptr());
```
YannickJadoul, line 508: Comments from #2746:

- @bstaletic: Without checking the type first, we were reinterpreting `NotPybindDerived()` as `instance*`.
- @YannickJadoul: See also the code below that gets simplified; `cfr. if (!self_value_and_holder.type || !self_value_and_holder.inst) -> pi->get_value_and_holder(tinfo, true);`

bstaletic (approved):
This one looks good to me. Is this the same commit just cherry-picked from #2746? That would make rebasing #2746 easier.

YannickJadoul:
> Is this the same commit just cherry-picked from #2746?

I believe so. At any rate, I'll take responsibility for rebasing :-)

EricCousineau-TRI (approved):
Looks great! I'm assuming there may be a bit of performance penalty, but nothing that can't be resolved later.

Mayhaps there's a way to fold this fail-fast more directly into `get_type_info`, but this is def. a strict improvement!

henryiii (approved):

YannickJadoul:
> I'm assuming there may be a bit of performance penalty, but nothing that can't be resolved later.

Yes. But as fa as I know, the same penalty as for a normal cast, if it would not have been `self`, but one of the other arguments?

YannickJadoul:
Thanks all. Since it was already tested in #2746, I'll merge this! :-)

YannickJadoul:
@EricCousineau-TRI, there's a bit of optimization in `type_caster_generic`: https://github.com/pybind/pybind11/blob/e612043d43e5382848466a5a6fe59e58edcf74fb/include/pybind11/cast.h#L675-L713

Worth checking out, indeed. Do we create an issue for this?

EricCousineau-TRI:
Nah, just speculative at this point. I think it only merits an issue if there's measurable / traceable performance penalties. (I'll keep in mind to try benchmarking older versions, say 2.3.x, to get some semblance of history, per #2760)

Also, not sure if I understand how `type_caster_generic` relates here as an optimization... Can you expand on that, if you think it's worth the time?

(FWIW, I don't think this is too actionable yet, so we can let sleeping dogs lie :dog: :zzz:)

YannickJadoul:
> Also, not sure if I understand how `type_caster_generic` relates here as an optimization... Can you expand on that, if you think it's worth the time?

Maybe it's not, though; I saw this `if (srctype == typeinfo->type)` vs. `else if (PyType_IsSubtype(srctype, typeinfo->type))`, but `PyType_IsSubtype` should do that itself? I think we might just be able to ignore this, actually. Let's leave it, for now, indeed :-)

Issue #2754, opened 2020-12-29
[BUG] TSAN error logs


rwgk:
TSAN logs as requested by @YannickJadoul here:

https://github.com/pybind/pybind11/pull/2746#issuecomment-751906066

Using `master` b7dfe5cc84ca1891b50e728e83c9b5b393cf5272
`+` #2409
`+` #2746 (https://github.com/YannickJadoul/pybind11/commit/727b3aeda951f40d7019d2b189edc3b1184764e0)
`+` 2 additional minor local patches.

`test -c dbg --config=tsan` (Google-internal toolchain)

Failing tests:
* `test_gil_scoped.py`
* `test_iostream.py` [**EDIT**: see PR #2995]: https://github.com/pybind/pybind11/blob/d587a2fd17dc89c0b2d4b6dc7715594bed45fb58/tests/test_iostream.cpp#L43

rwgk:
[test_gil_scoped_tsan_log.txt](https://github.com/pybind/pybind11/files/5749818/test_gil_scoped_tsan_log.txt)

rwgk:
[test_iostream_tsan_log.txt](https://github.com/pybind/pybind11/files/5749819/test_iostream_tsan_log.txt)

YannickJadoul:
To link up things:
- The `test_iostream` failure seems to be more info about https://github.com/pybind/pybind11/pull/2675#issuecomment-730014386

YannickJadoul:
I knew I'd seen the other one as well:
- The `test_gil_scoped` race condition is also shown in #2391

bstaletic:
I know this is the TSAN thread, but TSAN seems to be a major PITA to set up. I'm hoping no one would mind me submitting helgrind's analysis.

https://gist.github.com/bstaletic/98441d24b5e0010118051c31036ef2f8

Note: debug symbols for all dependency are necessary to get useful tracebacks.

Observation: `test_gil_scoped.py` doesn't seem to touch pybind11 as far as threading errors are concerned, but `test_iostream.py` does.

YannickJadoul:
The main issue with Helgrind (I've been playing around with it, a few days ago), is that it doesn't recognize C++'s `atomic`, etc?

Btw, the main issue (or one of the main issues) seems to lie in this line:

https://github.com/pybind/pybind11/blob/e612043d43e5382848466a5a6fe59e58edcf74fb/tests/test_iostream.cpp#L40

[...]

PR #2756, opened 2020-12-30
Plug leaking function_records in cpp_function initialization in case of exceptions (found by Valgrind in #2746)

YannickJadoul:
## Description

Another leak fix taken out of #2746.

~~I am not entirely happy with this solution yet, mostly because we are not reusing `destruct(function_record *)`. I tried adding this as the `unique_ptr`'s deleter, but the issue is that the `char *`s only get copied along the way, and are not owned before those copies.~~ (kind of fixed that)
~~Also, `rec->data` can still leak as `rec->free_data` is not called after an exception.~~ (fixed that)

## Suggested changelog entry:

```rst
Resolved memory leak in cpp_function initialization when exceptions occured.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -256,6 +256,9 @@ protected:
     /// Register a function call with Python (generic non-templated code goes here)
     void initialize_generic(unique_function_record &&unique_rec, const char *text,
                             const std::type_info *const *types, size_t args) {
+        // Do NOT receive `unique_rec` by value. If this function fails to move out the unique_ptr,
+        // we do not want this to destuct the pointer. `initialize` (the caller) still relies on the
+        // pointee being alive after this call. Only move out if a `capsule` is going to keep it alive.
         auto rec = unique_rec.get();
 
         // Keep track of strdup'ed strings, and clean them up as long as the function's capsule
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -114,14 +114,16 @@ public:
     object name() const { return attr("__name__"); }
 
 protected:
-    using unique_function_record = std::unique_ptr<detail::function_record, void (*)(detail::function_record *)>;
+    struct InitializingFunctionRecordDeleter {
+        // `destruct(function_record, false)`: `initialize_generic` copies strings and
+        // takes care of cleaning up in case of exceptions. So pass `false` to `free_strings`.
+        void operator()(detail::function_record * rec) { destruct(rec, false); }
+    };
+    using unique_function_record = std::unique_ptr<detail::function_record, InitializingFunctionRecordDeleter>;
 
     /// Space optimization: don't inline this frequently instantiated fragment
     PYBIND11_NOINLINE unique_function_record make_function_record() {
-        // `destruct(function_record, false)`: `initialize_generic` copies strings and
-        // takes care of cleaning up in case of exceptions. So set `free_srings` to `false`.
-        auto destruct_no_free_strings = [](detail::function_record * rec) { destruct(rec, false); };
-        return unique_function_record(new detail::function_record(), destruct_no_free_strings);
+        return unique_function_record(new detail::function_record());
     }
 
     /// Special internal constructor for functors, lambda functions, etc.
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -224,21 +224,41 @@ protected:
         }
     }
 
+    class strdup_guard {
+    public:
+        ~strdup_guard() {
+            for (auto s : strings)
+                std::free(s);
+        }
+        char *operator()(const char *s) {
+            auto t = strdup(s);
+            strings.emplace_back(t);
+            return t;
+        }
+        void release() {
+            strings.clear();
+        }
+    private:
+        std::vector<char *> strings;
+    };
+
     /// Register a function call with Python (generic non-templated code goes here)
     void initialize_generic(std::unique_ptr<detail::function_record> &&unique_rec, const char *text,
                             const std::type_info *const *types, size_t args) {
         auto rec = unique_rec.get();
 
+        strdup_guard guarded_strdup;
+
         /* Create copies of all referenced C-style strings */
-        rec->name = strdup(rec->name ? rec->name : "");
-        if (rec->doc) rec->doc = strdup(rec->doc);
+        rec->name = guarded_strdup(rec->name ? rec->name : "");
+        if (rec->doc) rec->doc = guarded_strdup(rec->doc);
         for (auto &a: rec->args) {
             if (a.name)
-                a.name = strdup(a.name);
+                a.name = guarded_strdup(a.name);
             if (a.descr)
-                a.descr = strdup(a.descr);
+                a.descr = guarded_strdup(a.descr);
             else if (a.value)
-                a.descr = strdup(repr(a.value).cast<std::string>().c_str());
+                a.descr = guarded_strdup(repr(a.value).cast<std::string>().c_str());
         }
 
         rec->is_constructor = !strcmp(rec->name, "__init__") || !strcmp(rec->name, "__setstate__");
@@ -321,13 +341,13 @@ protected:
 #if PY_MAJOR_VERSION < 3
         if (strcmp(rec->name, "__next__") == 0) {
             std::free(rec->name);
-            rec->name = strdup("next");
+            rec->name = guarded_strdup("next");
         } else if (strcmp(rec->name, "__bool__") == 0) {
             std::free(rec->name);
-            rec->name = strdup("__nonzero__");
+            rec->name = guarded_strdup("__nonzero__");
         }
 #endif
-        rec->signature = strdup(signature.c_str());
+        rec->signature = guarded_strdup(signature.c_str());
         rec->args.shrink_to_fit();
         rec->nargs = (std::uint16_t) args;
 
@@ -361,6 +381,7 @@ protected:
             capsule rec_capsule(unique_rec.release(), [](void *ptr) {
                 destruct((detail::function_record *) ptr);
             });
+            guarded_strdup.release();
 
             object scope_module;
             if (rec->scope) {
@@ -396,12 +417,14 @@ protected:
                 rec->next = chain;
                 auto rec_capsule = reinterpret_borrow<capsule>(((PyCFunctionObject *) m_ptr)->m_self);
                 rec_capsule.set_pointer(unique_rec.release());
+                guarded_strdup.release();
             } else {
                 // Or end of chain (normal behavior)
                 chain_start = chain;
                 while (chain->next)
                     chain = chain->next;
                 chain->next = unique_rec.release();
+                guarded_strdup.release();
             }
         }
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -243,7 +243,7 @@ protected:
         }
         char *operator()(const char *s) {
             auto t = strdup(s);
-            strings.emplace_back(t);
+            strings.push_back(t);
             return t;
         }
         void release() {
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -126,6 +126,7 @@ protected:
         struct capture { remove_reference_t<Func> f; };
 
         /* Store the function including any extra state it might have (e.g. a lambda capture object) */
+        // The unique_ptr makes sure nothing is leaked in case of an exception.
         auto unique_rec = make_function_record();
         auto rec = unique_rec.get();
 
@@ -208,6 +209,7 @@ protected:
         PYBIND11_DESCR_CONSTEXPR auto types = decltype(signature)::types();
 
         /* Register the function with Python from generic (non-templated) code */
+        // Pass on the ownership over the `unique_rec` to `initialize_generic`. `rec` stays valid.
         initialize_generic(std::move(unique_rec), signature.text, types.data(), sizeof...(Args));
 
         if (cast_in::has_args) rec->has_args = true;
@@ -224,6 +226,8 @@ protected:
         }
     }
 
+    // Utility class that keeps track of all duplicated strings, and cleans them up in its destructor,
+    // unless they are released. Basically a RAII-solution to deal with exceptions along the way.
     class strdup_guard {
     public:
         ~strdup_guard() {
@@ -247,6 +251,11 @@ protected:
                             const std::type_info *const *types, size_t args) {
         auto rec = unique_rec.get();
 
+        // Keep track of strdup'ed strings, and clean them up as long as the function's capsule
+        // has not taken ownership yet (when `unique_rec.release()` is called).
+        // Note: This cannot easily be fixed by a `unique_ptr` with custom deleter, because the strings
+        // are only referenced before strdup'ing. So only *after* the following block could `destruct`
+        // safely be called, but even then, `repr` could still throw in the middle of copying all strings.
         strdup_guard guarded_strdup;
 
         /* Create copies of all referenced C-style strings */
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -115,8 +115,8 @@ public:
 
 protected:
     /// Space optimization: don't inline this frequently instantiated fragment
-    PYBIND11_NOINLINE detail::function_record *make_function_record() {
-        return new detail::function_record();
+    PYBIND11_NOINLINE std::unique_ptr<detail::function_record> make_function_record() {
+        return std::unique_ptr<detail::function_record>(new detail::function_record());
     }
 
     /// Special internal constructor for functors, lambda functions, etc.
@@ -126,7 +126,8 @@ protected:
         struct capture { remove_reference_t<Func> f; };
 
         /* Store the function including any extra state it might have (e.g. a lambda capture object) */
-        auto rec = make_function_record();
+        auto unique_rec = make_function_record();
+        auto rec = unique_rec.get();
 
         /* Store the capture object directly in the function record if there is enough space */
         if (sizeof(capture) <= sizeof(rec->data)) {
@@ -207,7 +208,7 @@ protected:
         PYBIND11_DESCR_CONSTEXPR auto types = decltype(signature)::types();
 
         /* Register the function with Python from generic (non-templated) code */
-        initialize_generic(rec, signature.text, types.data(), sizeof...(Args));
+        initialize_generic(std::move(unique_rec), signature.text, types.data(), sizeof...(Args));
 
         if (cast_in::has_args) rec->has_args = true;
         if (cast_in::has_kwargs) rec->has_kwargs = true;
@@ -224,8 +225,9 @@ protected:
     }
 
     /// Register a function call with Python (generic non-templated code goes here)
-    void initialize_generic(detail::function_record *rec, const char *text,
+    void initialize_generic(std::unique_ptr<detail::function_record> &&unique_rec, const char *text,
                             const std::type_info *const *types, size_t args) {
+        auto rec = unique_rec.get();
 
         /* Create copies of all referenced C-style strings */
         rec->name = strdup(rec->name ? rec->name : "");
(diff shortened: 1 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -126,7 +126,8 @@ class cpp_function : public function {
         struct capture { remove_reference_t<Func> f; };
 
         /* Store the function including any extra state it might have (e.g. a lambda capture object) */
-        auto rec = make_function_record();
+        auto unique_rec = make_function_record();
```
YannickJadoul, line 129: Comment from #2746:

- @YannickJadoul (regarding `function_record::data`/`function_record::free_data`): This can still leak, I have just realized.

  It's very annoying that we cannot just have a `std::unique_ptr<detail::function_record, void (*)(detail::function_record *)>`, with destruct as deleter. The issue is that we cannot do this before all the `char *`s are `strdup`ed, and that 1) is only done in `initialize_generic`, and 2) the `repr` calls during `strdup` can also throw.

bstaletic: Making `make_function_record()` return a `py::capsule` already wouldn't help, right?

YannickJadoul: That would have the same issues as `std::unique_ptr<detail::function_record, void (*)(detail::function_record *)>`, I think? I.e., when do you know for sure that those strings have been `strdup`ed?

YannickJadoul: I think I fixed it; trying to demonstrate these leaks in #2746's 8e8fc6a2cfd5e8e8ce92f4a07217435490e6649a

YannickJadoul:
So, this should be ready to review:

- 1d56f738dc085197184e2cb722a4f3f992257295 in #2746 demonstrates some extra leaks that weren't fixed yet
- 6c8e2f16da10517d5cb035e6034fa2b645699a7e shows that the last commit does fix it, now

The one question I now have, is whether this fix is worth it. As far as I know, all of these leaks are actually errors by the pybind11 user, and can't/shouldn't happen during actual use, if the library with pybind11 is correctly designed and programmed. So do we want to incur this overhead?
Then again, this overhead is only during `cpp_function` initialization, so it shouldn't be on any critical path, either. So I lean towards merging this PR. But open to alternative suggestions!

rwgk:
Hi @YannickJadoul,

Is this sentence in the PR top comment still true?
> Also, rec->data can still leak as rec->free_data is not called after an exception.

What do you think about the diff below, to make it easier for the compiler/optimizer to see that a runtime bool works here?

```
index 114f72b..e57dccc 100644
@@ -120,7 +120,7 @@ protected:
     PYBIND11_NOINLINE unique_function_record make_function_record() {
         // `destruct<true>(function_record)`: `initialize_generic` copies strings and
         // takes care of cleaning up in case of exceptions.
-        return unique_function_record(new detail::function_record(), destruct<false>);
+        return unique_function_record(new detail::function_record(), destruct_no_delete_strings);
     }
 
     /// Special internal constructor for functors, lambda functions, etc.
@@ -392,7 +392,7 @@ protected:
             rec->def->ml_flags = METH_VARARGS | METH_KEYWORDS;
 
             capsule rec_capsule(unique_rec.release(), [](void *ptr) {
-                destruct((detail::function_record *) ptr);
+                destruct_with_delete_strings((detail::function_record *) ptr);
             });
             guarded_strdup.release();
 
@@ -489,9 +489,16 @@ protected:
         }
     }
 
+    static void destruct_with_delete_strings(detail::function_record *rec) {
+        destruct_impl(rec, true);
+    }
+
+    static void destruct_no_delete_strings(detail::function_record *rec) {
+        destruct_impl(rec, false);
+    }
+
     /// When a cpp_function is GCed, release any memory allocated by pybind11
-    template <bool DeleteStrings = true>
-    static void destruct(detail::function_record *rec) {
+    static void destruct_impl(detail::function_record *rec, bool delete_strings) {
         // If on Python 3.9, check the interpreter "MICRO" (patch) version.
         // If this is running on 3.9.0, we have to work around a bug.
         #if !defined(PYPY_VERSION) && PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 9
@@ -504,7 +511,7 @@ protected:
                 rec->free_data(rec);
             // During initialization, these strings might not have been copied yet,
             // so they cannot be freed. Once the function has been created, they can.
-            if (DeleteStrings) {
+            if (delete_strings) {
                 std::free((char *) rec->name);
                 std::free((char *) rec->doc);
                 std::free((char *) rec->signature);
```

YannickJadoul:
> Hi @YannickJadoul,
> 
> Is this sentence in the PR top comment still true?
> 
> > Also, rec->data can still leak as rec->free_data is not called after an exception.

Nope that's what I fixed, just today :-) Thanks, I fixed the original message!

> What do you think about the diff below, to make it easier for the compiler/optimizer to see that a runtime bool works here?

Hmm, not sure what you mean. In my version it is a compile-time bool, which should be easier to optimize, no?

That being said, your version should result in a smaller size overhead (at the cost of that extra runtime check), if we make sure that `destruct_impl` is not inlined (there's some pybind11 macro for that). So this is a valid trade-off to consider, that I hadn't yet thought of. Any thoughts?

rwgk:
> That being said, your version should result in a smaller size overhead (at the cost of that extra runtime check)

The overhead for the runtime check is almost certainly not measurable.
But the compiler or linker will have to work pretty hard to figure out how to avoid duplicating a sizable chunk of generated machine code.
Seems like a classical case of "template bloating" that's very easy to avoid, as suggested, and it makes the code clearer, too, at least for me.

YannickJadoul:
Alright, yes. Given other decisions in pybind11, prioritizing space over a tiny bit of performance makes sense to me! I've pulled this to runtime. (I have kept the default argument `= true`, though, indicating that that should be the default/normal case.)

YannickJadoul:
Alright, that commit passes Valgrind in #2746 (da8859d231f5ea78a6d10478ac12216fbf51be80) :-)

include/pybind11/pybind11.h
```diff
@@ -114,9 +114,14 @@ class cpp_function : public function {
     object name() const { return attr("__name__"); }
 
 protected:
+    using unique_function_record = std::unique_ptr<detail::function_record, void (*)(detail::function_record *)>;
```
bstaletic, line 117: A function pointer has state. A non-capturing function object that just calls the deleter function has no state. The standard library has a way of not storing a stateless deleter which makes `sizeof(T*) == sizeof(std::unique_ptr<T>)`. Can we make the deleter stateless?

Reading a bit more into this, I guess storing the function pointer was just convenient, because lambdas in non-evaluated context are a C++20 thing. We'll have to fall back to C++98 function objects.

include/pybind11/pybind11.h
```diff
@@ -207,7 +214,8 @@ class cpp_function : public function {
         PYBIND11_DESCR_CONSTEXPR auto types = decltype(signature)::types();
 
         /* Register the function with Python from generic (non-templated) code */
-        initialize_generic(rec, signature.text, types.data(), sizeof...(Args));
+        // Pass on the ownership over the `unique_rec` to `initialize_generic`. `rec` stays valid.
```
bstaletic, line 217: I'm not too happy about the sort-of-implicit contract of "you take ownership, but can't yet destroy the object". I understand why this is the case, it's just not what "pass on the ownership" really means.

include/pybind11/pybind11.h
```diff
@@ -223,20 +231,48 @@ class cpp_function : public function {
...
+    // unless they are released. Basically a RAII-solution to deal with exceptions along the way.
+    class strdup_guard {
+    public:
+        ~strdup_guard() {
+            for (auto s : strings)
+                std::free(s);
+        }
+        char *operator()(const char *s) {
+            auto t = strdup(s);
+            strings.emplace_back(t);
```
bstaletic, line 244: Any reason to use `emplace_back` instead of `push_back` here?

include/pybind11/pybind11.h
```diff
@@ -223,20 +231,48 @@ class cpp_function : public function {
...
+        void release() {
+            strings.clear();
+        }
+    private:
+        std::vector<char *> strings;
+    };
+
     /// Register a function call with Python (generic non-templated code goes here)
-    void initialize_generic(detail::function_record *rec, const char *text,
+    void initialize_generic(unique_function_record &&unique_rec, const char *text,
```
bstaletic, line 255: Why are you taking `std::unique_ptr` by reference instead of by-value? Remember that r-value references mean "maybe move".

Considering that we don't really want the unique_ptr here, rather just the pointer and a way to tell the unique_ptr it isn't in charge, maybe take by const-ref? A const-ref-unique-pointer has weird semantic meaning, transferring ownership from unique pointer to a capsule is probably equally weird.

include/pybind11/pybind11.h
```diff
@@ -223,20 +231,48 @@ class cpp_function : public function {
...
+            for (auto s : strings)
+                std::free(s);
+        }
+        char *operator()(const char *s) {
+            auto t = strdup(s);
+            strings.emplace_back(t);
+            return t;
+        }
+        void release() {
+            strings.clear();
```
bstaletic, line 248: `strings.clear()` doesn't call `free()` on the cleared pointers. Should it?

include/pybind11/pybind11.h
YannickJadoul: No, I agree, but 1) it's already what was happening before (it's just syntactically explicit now), and 2) I wouldn't immediately know what to do otherwise? Return the pointer from `initialize_generic`? That also feels reasonably ugly/non-intuitive.

include/pybind11/pybind11.h
YannickJadoul: Nope (except for maybe `clang-tidy`'s `modernize-use-emplace`?) Any reason for `push_back`?

include/pybind11/pybind11.h
YannickJadoul: > Considering that we don't really want the unique_ptr here, rather just the pointer and a way to tell the unique_ptr it isn't in charge, maybe take by const-ref?

That not going to work, because it needs to give up ownership to the `capsule`, later. I really want to pass ownership, not just `initialize_generic` use it.

> Remember that r-value references mean "maybe move".

True. So, ordinarily by value?

include/pybind11/pybind11.h
YannickJadoul: Nope, or it wouldn't be releasing! (This name is meant to mirror `unique_ptr::release()`.)

include/pybind11/pybind11.h
YannickJadoul: Done, I think.

include/pybind11/pybind11.h
bstaletic: Right. I was just thinking out loud.

include/pybind11/pybind11.h
YannickJadoul: Combined with https://github.com/pybind/pybind11/pull/2756#discussion_r551519918, this is even worse, though, the accept by value :-/

At least with `unique_function_record &&`, you have the contract "we only move in case of success, and otherwise the caller keeps ownership"

include/pybind11/pybind11.h
bstaletic: It's a simpler mechanism when you already have a `T` and it also doesn't need to instantiate a variadic member function template. On the other hand `emplace()` can't provide its main advantage of constructing in place, because you're already passing a `T`.

Not a big deal either way. Some people prefer to use `push_back(t)` while others prefer the simpler rule of "always use `emplace`".

include/pybind11/pybind11.h
bstaletic: > > Considering that we don't really want the unique_ptr here, rather just the pointer and a way to tell the unique_ptr it isn't in charge, maybe take by const-ref?
> 
> That not going to work, because it needs to give up ownership to the `capsule`, later. I really want to pass ownership, not just `initialize_generic` use it.
> 

Right, but passing by mutable reference would work. Though you know what I think about mutable references.

> > Remember that r-value references mean "maybe move".
> 
> True. So, ordinarily by value?

I guess so.

include/pybind11/pybind11.h
YannickJadoul: Yeah, OK. I do agree, actually, if it's a pointer, but something tells me clang-tidy bit me, last time I tried it...

@henryiii, any preferences? You're the one that set up clang-tidy :-P

include/pybind11/pybind11.h
YannickJadoul: > Right, but passing by mutable reference would work. Though you know what I think about mutable references.

Is that any cleaner? At least now we know that we're "maybe moving"?

If you have a nice combo with https://github.com/pybind/pybind11/pull/2756#discussion_r551519918, I'll implemented it (or feel free to push something), but otherwise...

bstaletic: > a nice combo with #2756 (comment)

Unfortunately not. Let's just go with pass by value.

include/pybind11/pybind11.h
bstaletic: I think clang-tidy would be fine with `push_back()` here, though I'm not really against leaving it as-is.

include/pybind11/pybind11.h
YannickJadoul: Added a comment on why we should probably not do so: https://github.com/pybind/pybind11/pull/2756/commits/9942614495cada974791019b1644d9bb07d21a8d

YannickJadoul:
@henryiii, @rwgk, @EricCousineau-TRI, can I gently ping? This is the last thing standing between us, and rebasing #2746 and soon having Valgrind checks run on all PRs :-)

rwgk:
Hi @YannickJadoul, a few days ago I offered to run this PR through Google's global testing system, but you discouraged it. I'd happily approve this PR after convincing myself that the tests come back clean.

Fully absorbing all the nuances of this very hairy change would cost me more time than I can give it. I trust that you and @bstaletic have done a great job. Given that we have comprehensive testing including sanitizers in place, getting more heads to look at the details doesn't seem like a productive division of labor.

Please let me know when it this is a good time to globally test this PR again.

YannickJadoul:
> Hi @YannickJadoul, a few days ago I offered to run this PR through Google's global testing system, but you discouraged it. I'd happily approve this PR after convincing myself that the tests come back clean.

Well, you asked that for #2746, and I informed you that this code was already tested (except for the last few additional commits that were not made because tests fail, but because reviewing the code found a logical corner case for leaks that wasn't covered by tests; not at Google either, since you said ASAN/LSAN came back clean?). Also, #2746 isn't supposed to merge this functionality; I'll rebase it on top of this, and #2746 will be a bare-bones change to CI and CMake.

It's the central part of our code, so the successful route is tested by all our tests. I don't believe more tests would tell us much more, since the nature of this PR is fixing leaks in exceptional cases (most often programming mistakes, actually). Though I'm of course not stopping you to run tests! (They're just not the right tool to fully assess this PR, I think.)

I was mainly asking for one or two more reviews and approval, so we can merge this and get full Valgrind runs in CI (like @bstaletic's, but he was involved in creating this, so I guess that only counts as half an approval?). I'm sure that before 2.6.2 is released, you'll still run Google's tests anyway?

rwgk:
> so I guess that only counts as half an approval?).

I look at it as a full approval: you still have two heads looking at the same thing.
Of course, more is always better, but also more expensive: diminishing returns.

> I'm sure that before 2.6.2 is released, you'll still run Google's tests anyway?

Running tests for one thing at a time has a lot of value: easier to find root causes.
My company is spending a lot of resources to make retesting easy, for humans. Let the machines to the work! That frees up human energy.

YannickJadoul:
Also: what the hell is going on with our CI/GitHub Actions?

```
/home/runner/work/_temp/9fcba6cd-dc0e-4f5e-a169-2da04a13086d.sh: line 1: /opt/hostedtoolcache/cmake/3.19.3/x64/cmake-3.19.3-Linux-aarch64/bin/cmake: cannot execute binary file: Exec format error
```

We're not running `aarch64`, you silly thing.

YannickJadoul:
> Running tests for one thing at a time has a lot of value: easier to find root causes.
> My company is spending a lot of resources to make retesting easy, for humans. Let the machines to the work! That frees up human energy.

That's a good argument, yes. But as argued, in this case, I'm not sure the machines can tell us anything more than we already know (unless there's specialized tests that try to stress test pybind11's exceptions?), and won't say much more interesting :-/

YannickJadoul:
Anyway, go head, @rwgk. But given that you already tested the logic (all that was added since is an extra cleanup), I'm not expecting anything there. But if it convinces you we can merge this, then why not.

include/pybind11/pybind11.h
henryiii: If clang-tidy wants emplace_back, then that's fine with me, though I think it should be happy with push_back for a pointer like this.

rwgk:
> Also, #2746 isn't supposed to merge this functionality; I'll rebase it on top of this, and #2746 will be a bare-bones change to CI and CMake.

Thanks, good to know. I'll test this PR specifically. Our global testing doesn't use CMake at all.

include/pybind11/pybind11.h
YannickJadoul: Let me try. I don't think I actually tried this out.

YannickJadoul: OK, you're right; clang-tidy passes! I must have misremembered/misinterpreted an old clang-tidy failure. This is better, indeed :-)

YannickJadoul:
> Also: what the hell is going on with our CI/GitHub Actions?
> 
> ```
> /home/runner/work/_temp/9fcba6cd-dc0e-4f5e-a169-2da04a13086d.sh: line 1: /opt/hostedtoolcache/cmake/3.19.3/x64/cmake-3.19.3-Linux-aarch64/bin/cmake: cannot execute binary file: Exec format error
> ```
> 
> We're not running `aarch64`, you silly thing.

OK, fixed by rebasing onto #2790. The remaining failures are still #2774, so unrelated.

rwgk:
Quick confirmation: this PR ~by itself*** on top of `master` makes the tests under pybind11/tests run ASAN & MSAN clean. The only sanitizer errors I'm still seeing are from TSAN, as reported under #2754.

*** I'm also using #2409 and two small local patches I've been carrying for a while.

YannickJadoul:
> Quick confirmation: this PR ~by itself*** on top of `master` makes the tests under pybind11/tests run ASAN & MSAN clean.

Huh, that's funny. So it's not flagging the one issue we couldn't really track down in #2746, and had to work around? (https://github.com/pybind/pybind11/pull/2746/commits/c1831201c047bce4c72ff469c2a9b81a5b9f41f8) Maybe/hopefully it's a Valgrind fluke, then, but it was quite consistent, so not sure I'm believing that :-/

henryiii (approved):
I'm happy to get this in.

rwgk:
> Huh, that's funny. So it's not flagging the one issue we couldn't really track down in #2746, and had to work around? ([c183120](https://github.com/pybind/pybind11/commit/c1831201c047bce4c72ff469c2a9b81a5b9f41f8))

Nope. I ran ASAN with pretty strict options, but MSAN only with defaults. I could play more with the options ... later! Let's get this in as-is asap IMO, pending only the results of the global testing run.

YannickJadoul:
Thanks, @henryiii :-)

> Nope. I ran ASAN with pretty strict options, but MSAN only with defaults. I could play more with the options ... later!

No worries. It's the one thing @bstaletic and I couldn't figure out. So the plan is to get Valgrind in, with the workaround, then undo the workaround in another PR that can serve as bug report. So I propose we keep discussion of this one thing for that one undoing-the-workaround PR?

rwgk:
The global testing came back clean. @yannick, please merge!

This PR was already extremely useful for my work on #2672, to sanitize my new code (I found a couple bugs already).

rwgk (approved):

rwgk:
FYI: After rolling out this PR Google-internal, our testing system discovered this (very minor) leak:
https://github.com/tensorflow/tensorflow/blob/10c06df9492c44062f3a6b0c04bec325c7738426/tensorflow/python/eager/pywrap_tfe_src.cc#L4335
I don't know how this PR unmasked that leak or how the test even succeeded before this PR. I mailed a fix. Let's see if the review sheds some light on this.

YannickJadoul:
Interesting; if you do find out anything relevant or need to link back to these changes, I'm very interested.

Issue #2742, opened 2020-12-24
[BUG] Memory leaks when casting raw pointers to python types


bstaletic:
## Issue description

Once again I've been playing with my toys - valgrind, pybind11@ #2741 and CPython 3.9.1

Good news: We're almost there. Almost a clean run.
Bad news: Some casters leak like a broken faucet.

More specifically, `py::cast(new int{})` leaks and so do all the other casters that copy the data - `list_caster` from `stl.h` is another example. Another thing that leaks is `new CustomClass{}` with `py::return_value_policy::move`.

While `py::cast(new int{});` could be argued that it's a user error, that argument doesn't take into account the following:

    m.def("f", []{ return new int{};});
    m.def("g", []{ return Class{};}, py::return_value_plicy::move);

Even our docs have [`return new Example();`](https://pybind11.readthedocs.io/en/latest/advanced/classes.html) and "don't mix things that don't play well together" in this case is very subtle of a contract to introduce after so many years of people assuming that it just works (tm).

## Reproducible example code

```c++
#include <pybind11/pybind11.h>
struct s{};
PYBIND11_MODULE(test, m) {
    pybind11::class_<s>(m, "s");
    m.def("leak1", []{ return new int{}; });
    m.def("no_leak", []{ return new s{}; });
    m.def("leak2", []{ return new s{}; }, pybind11::return_value_policy::move);
}
```

Calling all three functions results in these valgrind errors:

```
==7672== HEAP SUMMARY:
==7672==     in use at exit: 811,198 bytes in 5,907 blocks
==7672==   total heap usage: 32,402 allocs, 26,495 frees, 4,805,059 bytes allocated
==7672== 
==7672== 1 bytes in 1 blocks are definitely lost in loss record 2 of 1,971
==7672==    at 0x483ADEF: operator new(unsigned long) (vg_replace_malloc.c:342)
==7672==    by 0x5B58637: pybind11_init_test(pybind11::module_&)::{lambda()#3}::operator()() const (foo.cpp:7)
==7672==    by 0x5B59BD5: s* pybind11::detail::argument_loader<>::call_impl<s*, pybind11_init_test(pybind11::module_&)::{lambda()#3}&, , pybind11::detail::void_type>(pybind11_init_test(pybind11::module_&)::{lambda()#3}&, std::integer_sequence<unsigned long>, pybind11::detail::void_type&&) && (cast.h:2022)
[...]

PR #490, opened 2016-11-11
Fix stl_bind to support movable, non-copyable value types

jagerman:
Fixes #487 (but not ready for merging).

We require copyable types for pretty much all vector modifiers, so this basically moves them all into a method that is only enabled for copyable types.  (Technically we could allow `pop` and `__delitem__` without copying, but it seems a little weird to expose deletion but no other modifiers).

In practice, this means that if the type is non-copyable, it will be, for all intents and purposes, readonly from the Python side--but at least it's something (currently it fails to compile).

This is still incomplete: it needs similar support for bind_map, and needs a better solution for the current dirty hack in `cast.h`, on which I'd appreciate input on how to deal with it.  I definitely don't think it should remain as is: this was just a hack to get it working as a starting point.

The issue is that `make_copy_constructor` fails for `std::vector<vt>` where `vt` is non-copyable.  The `decltype(new std::vector<vt>(...))` succeeds (i.e. is a type), but an actual attempt to invoke the constructor results in a compilation failure.  (`std::is_copy_constructible` also fails in the same way, i.e. by returning true even when actually invoking the copy constructor results in a compilation failure).  For now I'm just hacking around it (and the hack is only on the non-MSVC path: the MSVC build of this will fail), but a better solution is needed.  @alzie (in #487) points to a workaround in http://stackoverflow.com/questions/18404108/issue-with-is-copy-constructible; any thoughts on something better?

PR #2758, opened 2020-12-30
Fix various minor memory leaks in the tests (found by Valgrind in #2746)

YannickJadoul:
## Description

Five reasonably trivial fixes for leaks in the tests, commits carved out of #2746:

- Fix leak in the test_copy_move::test_move_fallback
- Fix leaking PyMethodDef in test_class::test_implicit_conversion_life_support
- Plumb leak in test_buffer, occuring when a mutable buffer is requested for a read-only object, and enable test_buffer.py
- Fix weird return_value_policy::reference in test_stl_binders, and enable those tests
- Cleanup nodelete holder objects in test_smart_ptr, and enable those tests

## Suggested changelog entry:

```rst
Fixed various minor memory leaks in pybind11's test suite.
```

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -550,6 +550,12 @@ extern "C" inline int pybind11_getbuffer(PyObject *obj, Py_buffer *view, int fla
     }
     std::memset(view, 0, sizeof(Py_buffer));
     buffer_info *info = tinfo->get_buffer(obj, tinfo->get_buffer_data);
+    if ((flags & PyBUF_WRITABLE) == PyBUF_WRITABLE && info->readonly) {
+        delete info;
+        // view->obj = nullptr;  // Was just memset to 0, so not necessary
+        PyErr_SetString(PyExc_BufferError, "Writable buffer requested for readonly storage");
+        return -1;
+    }
     view->obj = obj;
     view->ndim = 1;
     view->internal = info;
@@ -559,12 +565,6 @@ extern "C" inline int pybind11_getbuffer(PyObject *obj, Py_buffer *view, int fla
     for (auto s : info->shape)
         view->len *= s;
     view->readonly = info->readonly;
-    if ((flags & PyBUF_WRITABLE) == PyBUF_WRITABLE && info->readonly) {
-        if (view)
-            view->obj = nullptr;
-        PyErr_SetString(PyExc_BufferError, "Writable buffer requested for readonly storage");
-        return -1;
-    }
     if ((flags & PyBUF_FORMAT) == PyBUF_FORMAT)
         view->format = const_cast<char *>(info->format.c_str());
     if ((flags & PyBUF_STRIDES) == PyBUF_STRIDES) {
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -231,7 +231,8 @@ TEST_SUBMODULE(class_, m) {
         };
 
         auto def = new PyMethodDef{"f", f, METH_VARARGS, nullptr};
-        return py::reinterpret_steal<py::object>(PyCFunction_NewEx(def, nullptr, m.ptr()));
+        py::capsule def_capsule(def, [](void *ptr) { delete reinterpret_cast<PyMethodDef *>(ptr); });
+        return py::reinterpret_steal<py::object>(PyCFunction_NewEx(def, def_capsule.ptr(), m.ptr()));
     }());
 
     // test_operator_new_delete
--- a/tests/test_copy_move.cpp
+++ b/tests/test_copy_move.cpp
@@ -214,6 +214,7 @@ TEST_SUBMODULE(copy_move_policies, m) {
     };
     py::class_<MoveIssue2>(m, "MoveIssue2").def(py::init<int>()).def_readwrite("value", &MoveIssue2::v);
 
-    m.def("get_moveissue1", [](int i) { return new MoveIssue1(i); }, py::return_value_policy::move);
+    // #2742: Don't expect ownership of raw pointer to `new`ed object to be transferred with `py::return_value_policy::move`
+    m.def("get_moveissue1", [](int i) { return std::unique_ptr<MoveIssue1>(new MoveIssue1(i)); }, py::return_value_policy::move);
     m.def("get_moveissue2", [](int i) { return MoveIssue2(i); }, py::return_value_policy::move);
 }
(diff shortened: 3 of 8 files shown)


include/pybind11/detail/class.h
```diff
@@ -550,6 +550,12 @@ extern "C" inline int pybind11_getbuffer(PyObject *obj, Py_buffer *view, int fla
     }
     std::memset(view, 0, sizeof(Py_buffer));
     buffer_info *info = tinfo->get_buffer(obj, tinfo->get_buffer_data);
+    if ((flags & PyBUF_WRITABLE) == PyBUF_WRITABLE && info->readonly) {
+        delete info;
```
YannickJadoul, line 554: Comment from #2746:

- @YannickJadoul: This was the actual leak; The fragment was just moved up to not do futile work and such that view->obj = nullptr; isn't necessary anymore.

tests/test_stl_binders.cpp
```diff
@@ -86,13 +86,13 @@ TEST_SUBMODULE(stl_binders, m) {
 
     // test_noncopyable_containers
     py::bind_vector<std::vector<E_nc>>(m, "VectorENC");
-    m.def("get_vnc", &one_to_n<std::vector<E_nc>>, py::return_value_policy::reference);
```
YannickJadoul, line 89: Comments from #2746:

- @bstaletic: These rvps look bogus. The return is by value.
- @YannickJadoul: Huh? I thought it returned a pointer to the container?
- @bstaletic: I was just entirely unable to figure out why/how the reference rvp is here. This test was added in #490 (commit 617fbcf). The PR involves returning elements by reference, but I don't see why these factory methods return the container by reference.
- @YannickJadoul: Huh? I thought it returned a pointer to the container?
- @bstaletic: That's what happens when you (read: I) do a review while really really tired. Listen to @YannickJadoul, everyone!
- @YannickJadoul: Hah 😄 No worries, we've all been there.
- @bstaletic: I'm not resolving this discussion, just so it doesn't get collapsed before others take a look.

tests/test_class.cpp
```diff
@@ -231,7 +231,8 @@ TEST_SUBMODULE(class_, m) {
         };
 
         auto def = new PyMethodDef{"f", f, METH_VARARGS, nullptr};
-        return py::reinterpret_steal<py::object>(PyCFunction_NewEx(def, nullptr, m.ptr()));
+        py::capsule def_capsule(def, [](void *ptr) { delete reinterpret_cast<PyMethodDef *>(ptr); });
+        return py::reinterpret_steal<py::object>(PyCFunction_NewEx(def, def_capsule.ptr(), m.ptr()));
```
bstaletic, line 235: `PyCFunction_NewEx()` just calls [`PyCMethod_new()`](https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Objects/methodobject.c#L44) with the last parameter set to NULL.

What this changes is [`m_self`](https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Objects/methodobject.c#L110) from NULL to `def_capsule.ptr()`.

___

Beyond that, I have no idea how this works. `PyCFunction*` aren't documented, which is discussed in [bpo-16776](https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Objects/methodobject.c#L110).

tests/test_copy_move.cpp
```diff
@@ -214,6 +214,7 @@ TEST_SUBMODULE(copy_move_policies, m) {
     };
     py::class_<MoveIssue2>(m, "MoveIssue2").def(py::init<int>()).def_readwrite("value", &MoveIssue2::v);
 
-    m.def("get_moveissue1", [](int i) { return new MoveIssue1(i); }, py::return_value_policy::move);
+    // #2742: Don't expect ownership of raw pointer to `new`ed object to be transferred with `py::return_value_policy::move`
+    m.def("get_moveissue1", [](int i) { return std::unique_ptr<MoveIssue1>(new MoveIssue1(i)); }, py::return_value_policy::move);
```
bstaletic, line 218: Just to make things clear, this is avoiding the issue reported in #2742, but not proposing a definite solution.

Moreover, the issue that was a precursor to adding this tests doesn't seem to depend on `get_moveissue1()` returning a raw owning pointer.

tests/test_class.cpp
YannickJadoul: > What this changes is [`m_self`](https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Objects/methodobject.c#L110) from NULL to `def_capsule.ptr()`.

We already do this in the actual pybind11 code, btw.

Alternatively, if you prefer, we could play the "`weakref` trick", here? But I thought this would be easier, since AFAICT `m_self` being `NULL` wasn't the point of this test.

bstaletic: I'm just trying to understand how does setting `m_self` work to our benefit in this case. Does it steal the reference or borrow it? Is our reference counting right in this case? If it appeased valgrind, I guess it's correct, I'd just like a better understanding of how it works.

YannickJadoul: It [does borrow](https://github.com/python/cpython/blob/04b8631d84a870dda456ef86039c1baf34d08500/Objects/methodobject.c#L108-L109), yes, and keeps it alive as some kind of CPython-style `void *data`, passed as first argument to our `cpp_function::dispatcher`: https://github.com/pybind/pybind11/blob/d587a2fd17dc89c0b2d4b6dc7715594bed45fb58/include/pybind11/pybind11.h#L492

See here for the other case where CPython uses it:
https://github.com/pybind/pybind11/blob/d587a2fd17dc89c0b2d4b6dc7715594bed45fb58/include/pybind11/pybind11.h#L372

This mechanism was also our problem with Python 3.9.0, `m_self` being decref'ed a few lines too early while pybind11 was still accessing it, in #2576.

tests/test_copy_move.cpp
YannickJadoul: No, but if we decide to have a different approach to #2742, we'll add new tests for those, anyway? So this change could be semi-permanent?

bstaletic (commented):
This looks good to me.

henryiii (approved):
Small fix then mostly test changes to be tidy, looks good to me.

YannickJadoul:
@henryiii, you can't think why these `test_stl_binders` tests were returning with `return_value_policy::reference`, in #490 / 378debbc02e070f98af2b9dbd7297fa403845350?
If not, I think we can merge; the other fixes should be uncontroversial enough :-)

henryiii:
Looks like a (natural) mistake to me.

YannickJadoul:
OK, then I'll merge and rebase! Thanks for checking!

PR #2759, opened 2020-12-30
Ignore deprecation warnings about old-style __init__/__setstate__ constructors in the tests (originally done in #2746)

YannickJadoul:
## Description

As the title says. The test suite contains a few tests to make sure the old-style `__init__` (with placement `new`s) still works and interacts correctly with the new, non-deprecated way of doing things. Using a debug build of Python (in #2746), pybind11 shows these deprecation warnings, so let's filter these out to remove some output clutter and make sure we have a higher chance of catching useful warnings.

## Suggested changelog entry:

```rst
Suppressed some deprecation warnings about old-style ``__init__``/``__setstate__`` in the tests.
```

### DIFF
--- a/tests/pybind11_tests.h
+++ b/tests/pybind11_tests.h
@@ -69,3 +69,17 @@ public:
 };
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(pybind11)
+
+/// Simplified ``with warnigns.catch_warnings()`` wrapper
+template <typename F>
+void ignoreOldStyleInitWarnings(F &&body) {
+    auto message = "pybind11-bound class '.+' is using an old-style placement-new '(?:__init__|__setstate__)' which has been deprecated";
+    auto category = py::reinterpret_borrow<py::object>(PyExc_FutureWarning);
+    auto warnings = py::module_::import("warnings");
+    auto context_mgr = warnings.attr("catch_warnings")();
+    context_mgr.attr("__enter__")();
+    warnings.attr("filterwarnings")("ignore", py::arg("message")=message, py::arg("category")=category);
+    body();
+    // Exceptions in `body` not handled; see PEP 343 when these would need to be added
+    context_mgr.attr("__exit__")(py::none(), py::none(), py::none());
+}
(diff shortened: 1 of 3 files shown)


tests/pybind11_tests.h
```diff
@@ -69,3 +70,15 @@ template<> class type_caster<RValueCaster> {
 };
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(pybind11)
+
+template <typename F>
+void ignoreOldStyleInitWarnings(F &&body) {
+    py::exec(R"(
```
YannickJadoul, line 76: Comment from #2746:

- @YannickJadoul: For reference, this was the old code, without `py::exec` (and thus without requiring `<pybind11/eval.h>` to be included in `pybind11_tests.h` and thus in all tests).

  @bstaletic slightly prefers this [new version], and I think I do to. However, curious to learn what other reviewers think.

  ```c++
  /// Simplified ``with warnigns.catch_warnings()`` wrapper
  template <typename F>
  void ignoreOldStyleInitWarnings(F &&body) {
      auto message = "pybind11-bound class '.+' is using an old-style placement-new '(?:__init__|__setstate__)' which has been deprecated";
      auto category = py::reinterpret_borrow<py::object>(PyExc_FutureWarning);
      auto warnings = py::module_::import("warnings");
      auto context_mgr = warnings.attr("catch_warnings")();
      context_mgr.attr("__enter__")();
      warnings.attr("filterwarnings")("ignore", py::arg("message")=message, py::arg("category")=category);
      body();
      // Exceptions in `body` not handled; see PEP 343 when these would need to be added
      context_mgr.attr("__exit__")(py::none(), py::none(), py::none());
  }
  ```

henryiii (approved):
All test changes to avoid warnings because we are intentionally checking things that warn, looks good to me.

bstaletic (approved):
Since I helped with the PR, am I allowed to approve?

YannickJadoul:
Let's merge this, then? If anyone objects against `#include <pybind11/eval.h>` in `pybind11_tests.h`, we can still change it!

Issue #2761, opened 2020-12-30
[FEAT] keep_alive_impl should admit reference cycles, and ideally be released by GC?


EricCousineau-TRI:
Upstream motivation:
https://github.com/RobotLocomotion/drake/issues/14387

Basically, because of our usage of `unique_ptr` and how we've design our C++ public API and bindings (https://github.com/RobotLocomotion/drake/issues/13058), there are some awkward edge cases when we have "opaque" memory transfers.

This arose because we wanted to fix a segfault: https://github.com/RobotLocomotion/drake/pull/14356
However, this reference cycle appears to grow without bounds.

TODO(me): Explicitly try `gc.collect()` with aggressive collection arguments.

I'm filing this now because @YannickJadoul mentioned that `keep_alive` may be useful for #1941 (`py::dynamic_attr` coordinating with "mismatched" lifetime between `py::detail::instance` and the actual `shared_ptr<T>`). A `keep_alive` may fix that issue in a more general way, but will admit reference cycles

*Note*: I am speculating that the current approach (based on Boost.Python, per @bmerry's comment in #880) doesn't admit GC collection when cycles are introduced. I should confirm this.

\cc @rwgk @rhaschke about this sharp edge case with `unique_ptr<>` (#2646)

EricCousineau-TRI:
Related references about CPython GC can be seen here:
https://github.com/pybind/pybind11/issues/2757

Note: I will only be considering CPython as we do not use PyPy for our usage at present (Drake: https://github.com/RobotLocomotion/drake)

EricCousineau-TRI:
Some thoughts after reading the CPython dev guide (https://devguide.python.org/garbage_collector/)

- Could be possible if keep-alive tracking is exposed more directly to garbage collector; e.g. make them be proper references, but somehow add a `gc` hook to rescan the keep alive list.
- Could also make an explicit admission of reference-compatible keep-alive; i.e. `py::keep_alive_cycle<A, B>()`. However, making this work could be a bit weird.

Additional steps: 
- [ ] 1. Write code to confirm that a `keep_alive` cycle is not discovered by `gc.collect()` on pybind11 mater
- [ ] 2. Review Boost.Python's setup for similar mechanisms (see if they have notes there)
- [ ] 3. Review other binding setups (e.g. cppyy, SWIG)

On point (3) - @wlav, can I ask if y'all ever had to deal with this kind of thing in cppyy?

EricCousineau-TRI:
[...]

PR #2762, opened 2020-12-30
[WIP] test_keep_alive: Show that keep_alive cycle escapes GC

EricCousineau-TRI:
This is a PR to track progress for #2761


YannickJadoul:
I haven't even taken a shallow look, so apologies if I completely missed something, but what about enabling the GC to look into C types, using https://docs.python.org/3/c-api/gcsupport.html ?

EricCousineau-TRI:
Nice! I don't think it'll fit in the current `keep_alive` infrastructure as-is, but I believe it could fit in with some sort of indirection (simplest would be cycle-only keep-alive, better would be for any time)

bmerry:
> I haven't even taken a shallow look, so apologies if I completely missed something, but what about enabling the GC to look into C types, using https://docs.python.org/3/c-api/gcsupport.html ?

I think that would be the only sensible way to do it.

YannickJadoul:
Not sure how, but there's bound to be some kind of way of opting into this, similar to the current `py::dynamic_attributes()` (which is also translated into a flag set on the actual `PyHeapObject`, I believe).

Issue #2275, opened 2020-06-30
Allow __doc__ to be settable on enums


AWhetter:
## Issue description

Currently when declaring an enum, the `__doc__` attribute of the enum is not settable. However the `__doc__` on `object()` is settable so it makes sense for the `__doc__` of an enum to be settable also.

The use case where this is coming up:
I'm generating type stubs of a pybind module using stubgen. When running mypy with the generated stubs I see the following error:

```
Signature of "__doc__" incompatible with supertype of "object".
```

## Potential Fix
I'm not sure what the right fix is here. I can of course make the `__doc__` property writable on all enum objects, but implementing that properly would mean adding a new attribute to enum objects and that seems wasteful.

## Reproducible example code

```cpp
# include <pybind11/pybind11.h>

namespace py = pybind11;

enum Numbers {
    zero,
    one
};

PYBIND11_MODULE(numbers, m) {
    py::enum_<Numbers>(m, "Numbers")
        .value("zero", zero)
        .value("one", one)
        .export_values()
    ;
}
```

bstaletic:
This is an interesting issue. I tried a few things and I'll quote myself.

> `py::enum_`s are weird...
Even this doesn't help:

```
PYBIND11_MODULE(numbers, m) {
    py::enum_<Numbers>e(m, "Numbers");
    e
        .value("zero", zero)
        .value("one", one)
        .export_values()
    ;
    e.attr("__doc__") = "test";
}
```

Result:

```
>>> import numbers
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: AttributeError: can't set attribute
```

sizmailov:
At the moment pybind doesn't allow to overwrite docstrings of functions, methods or class/module attributes: 
```
attribute '__doc__' of 'instancemethod' objects is not writable
attribute '__doc__' of 'builtin_function_or_method' objects is not writable
'Foo' object attribute '__doc__' is read-only
```

Although class and module `__doc__` is writable.

bstaletic:
Except that `enum_` has a `def()`, which turns it into a `class_` and `__doc__` is still not assignable.

sizmailov:
All prohibiting cases boils down to assign-to-property problem. This can be hacked using `def_property_readonly_static`:

```c++
e.def_property_readonly_static("__doc__", [](py::object &) { return "test"; });
```

AWhetter:
[...]

PR #2768, opened 2021-01-05
Add option for enable/disable enum members in docstring.

knarfS:
## Description

This adds the option to disable/enable the enum members listed in the docstring (default is enabled). This is useful, when the documentation tool is able to generate documentation for class variables, so the enum members / class variable aren't listed twice in the documentation. An other benefit of using the features of the documentation generator is, that the members are linkable in the documentation.

I'm using pdoc3 to generate the documentation of an embedded python module, but the above also applies partially to pydoc. No idea about sphinx, I wasn't able to generate the documentation programmatically with sphinx...

I'd be happy to add that option to the documentation, but need a hint to which section this would fit best. There is a chapter about sphinx, but that doesn't exactly match.

## Suggested changelog entry:

```
Add option for enable/disable enum members in docstring.
```

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -324,12 +324,12 @@ The class ``options`` allows you to selectively suppress auto-generated signatur
         m.def("add", [](int a, int b) { return a + b; }, "A function which adds two numbers");
     }
 
-pybind11 also appends all members of an enum to the resulting enum docstring,
-this default behaviour can be disabled by using the ``disable_enum_members_docstring()``
+pybind11 also appends all members of an enum to the resulting enum docstring.
+This default behavior can be disabled by using the ``disable_enum_members_docstring()``
 function of the ``options`` class.
 
 With ``disable_user_defined_docstrings()`` all user defined docstrings of
-``module_::def()``, ``class_::def()`` and ``enum_()`` are discarded, but the
+``module_::def()``, ``class_::def()`` and ``enum_()`` are disabled, but the
 function signatures and enum members are included in the docstring, unless they
 are disabled separately.
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -326,12 +326,12 @@ The class ``options`` allows you to selectively suppress auto-generated signatur
 
 pybind11 also appends all members of an enum to the resulting enum docstring,
 this default behaviour can be disabled by using the ``disable_enum_members_docstring()``
-function of the ``options`` classe.
+function of the ``options`` class.
 
 With ``disable_user_defined_docstrings()`` all user defined docstrings of
 ``module_::def()``, ``class_::def()`` and ``enum_()`` are discarded, but the
 function signatures and enum members are included in the docstring, unless they
-are disabled seperately.
+are disabled separately.
 
 Note that changes to the settings affect only function bindings created during the
 lifetime of the ``options`` instance. When it goes out of scope at the end of the module's init function,
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -324,6 +324,15 @@ The class ``options`` allows you to selectively suppress auto-generated signatur
         m.def("add", [](int a, int b) { return a + b; }, "A function which adds two numbers");
     }
 
+pybind11 also appends all members of an enum to the resulting enum docstring,
+this default behaviour can be disabled by using the ``disable_enum_members_docstring()``
+function of the ``options`` classe.
+
+With ``disable_user_defined_docstrings()`` all user defined docstrings of
+``module_::def()``, ``class_::def()`` and ``enum_()`` are discarded, but the
+function signatures and enum members are included in the docstring, unless they
+are disabled seperately.
+
 Note that changes to the settings affect only function bindings created during the
 lifetime of the ``options`` instance. When it goes out of scope at the end of the module's init function,
 the default settings are restored to prevent unwanted side effects.
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1972,29 +1972,32 @@ struct enum_base {
             name("name"),
             is_method(m_base));
 
-        m_base.attr("__doc__") = static_property(
-            cpp_function(
-                [](handle arg) -> std::string {
-                    std::string docstring;
-                    dict entries = arg.attr("__entries");
-                    if (((PyTypeObject *) arg.ptr())->tp_doc) {
-                        docstring += std::string(((PyTypeObject *) arg.ptr())->tp_doc) + "\n\n";
-                    }
-                    docstring += "Members:";
-                    for (auto kv : entries) {
-                        auto key = std::string(pybind11::str(kv.first));
-                        auto comment = kv.second[int_(1)];
-                        docstring += "\n\n  " + key;
-                        if (!comment.is_none()) {
-                            docstring += " : " + (std::string) pybind11::str(comment);
+        if (options::show_enum_members_docstring()) {
+            m_base.attr("__doc__") = static_property(
+                cpp_function(
+                    [](handle arg) -> std::string {
+                        std::string docstring;
+                        dict entries = arg.attr("__entries");
+                        if (((PyTypeObject *) arg.ptr())->tp_doc) {
+                            docstring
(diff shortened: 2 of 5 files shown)


knarfS:
>     1. Should this be a global setting, or just a local flag, passed to the enum constructor? I'm not sure; other reviewers, what do you think? At any rate, I'd like some arguments pro or contra this approach.

I wanted to put this in the same place as the other docstring related options (`function_signatures`and `user_defined_docstrings`), but yeah, maybe the enum ctor would be a better place?

>     2. `m_base.attr("__doc__") = static_property(...)` isn't doing anything more than adding something to the docstring. The `if` statement can/should just go around this whole creation of the `static_property`, I think.

There is still the docstring of the enum class itself (`std::string(((PyTypeObject *) arg.ptr())->tp_doc)`), which shouldn't be affected by this option.

YannickJadoul:
> I wanted to put this in the same place as the other docstring related options (`function_signatures`and `user_defined_docstrings`), but yeah, maybe the enum ctor would be a better place?

Yeah, I'm not 100% convinced of either approach, but I thought it's worth bringing up).
The advantage of a global thing is that it's consistent across all `enums_`. The thing that makes me doubt this approach is whether this isn't too fine-grained to make a full global option out of it.
The disadvantage of a constructor argument is that C++ doesn't have keyword arguments or even keyword-only arguments... :-/

If there are no further complaints, I'm happy to keep the global option, though.

> There is still the docstring of the enum class itself (`std::string(((PyTypeObject *) arg.ptr())->tp_doc)`), which shouldn't be affected by this option.

I believe that if you don't set `__doc__`, asking for `__doc__` will/should give you the contents of `tp_doc`. I might of course be completely wrong. Could you please give this a try?

knarfS:
> I believe that if you don't set `__doc__`, asking for `__doc__` will/should give you the contents of `tp_doc`. I might of course be completely wrong. Could you please give this a try?

You are right, I've just tested this and it works!

YannickJadoul:
Thanks, that's already better, I think :-)

>  The disadvantage of a constructor argument is that C++ doesn't have keyword arguments or even keyword-only arguments... :-/

Just to add to future discussion (I'm still not saying this is preferred): if we want to get around this, another option is to come up with some custom tag like `py::enum_members_docstring(false)` or so.

knarfS:
This will also fix https://github.com/pybind/pybind11/issues/2275 when disabling the enum members docstring (aka not setting `__doc__`):

```c++
# include <pybind11/pybind11.h>

namespace py = pybind11;

enum Numbers {
    zero,
    one
}

PYBIND11_MODULE(numbers, m) {
    py::options options;
    options.disable_enum_members_docstring();

    py::enum_<Numbers>e(m, "Numbers");
    e
        .value("zero", zero)
        .value("one", one)
        .export_values();
    e.attr("__doc__") = "test";
}
```

YannickJadoul:
Thanks for checking that as well, @knarfS!

rwgk:
Looks good to me, based on the understanding that this PR adds an option with a default that leaves the existing behavior unchanged, unless the option is actually used. Is that correct?

The only thing missing are tests and possibly a couple lines in the documentation (I'd look for documentation for existing options as a starting point).

knarfS:
> Looks good to me, based on the understanding that this PR adds an option with a default that leaves the existing behavior unchanged, unless the option is actually used. Is that correct?

Yes, the current docstring behavior is untouched. I've added a test to secure that.

I've also added a section about the undocumented option `disable_user_defined_docstrings()`.

docs/advanced/misc.rst
```diff
@@ -324,6 +324,15 @@ The class ``options`` allows you to selectively suppress auto-generated signatur
         m.def("add", [](int a, int b) { return a + b; }, "A function which adds two numbers");
     }
 
+pybind11 also appends all members of an enum to the resulting enum docstring,
+this default behaviour can be disabled by using the ``disable_enum_members_docstring()``
```
rwgk, line 328: I think this will read slightly nicer as a new sentence:
```
... resulting enum docstring.
This default behavior ...
```

docs/advanced/misc.rst
```diff
@@ -324,6 +324,15 @@ The class ``options`` allows you to selectively suppress auto-generated signatur
...
+function of the ``options`` class.
+
+With ``disable_user_defined_docstrings()`` all user defined docstrings of
+``module_::def()``, ``class_::def()`` and ``enum_()`` are discarded, but the
```
rwgk, line 332: I'd use "disabled" instead of "discarded" here.

rwgk (approved):
Nice, thank you!

knarfS:
Thank you @rwgk for reviewing and approving. I updated the PR with your suggestions.

rwgk:
Great, thanks! I'll wait a day or so to give @Skylion007 I chance to review, too.

Skylion007 (approved):

include/pybind11/pybind11.h
```diff
@@ -1972,29 +1972,32 @@ struct enum_base {
...
-                            docstring += " : " + (std::string) pybind11::str(comment);
+        if (options::show_enum_members_docstring()) {
+            m_base.attr("__doc__") = static_property(
+                cpp_function(
+                    [](handle arg) -> std::string {
+                        std::string docstring;
+                        dict entries = arg.attr("__entries");
+                        if (((PyTypeObject *) arg.ptr())->tp_doc) {
+                            docstring
+                                += std::string(((PyTypeObject *) arg.ptr())->tp_doc) + "\n\n";
```
Skylion007, line 1983: nit: reinterpret_cast ?

include/pybind11/pybind11.h
```diff
@@ -1972,29 +1972,32 @@ struct enum_base {
...
-            none(),
-            none(),
-            "");
+                        docstring += "Members:";
+                        for (auto kv : entries) {
+                            auto key = std::string(pybind11::str(kv.first));
+                            auto comment = kv.second[int_(1)];
+                            docstring += "\n\n  " + key;
+                            if (!comment.is_none()) {
+                                docstring += " : " + (std::string) pybind11::str(comment);
```
Skylion007, line 1991: nit: static_cast or `.cast<std::string>`

include/pybind11/pybind11.h
```diff
@@ -1972,29 +1972,32 @@ struct enum_base {
...
-                },
-                name("__doc__")),
-            none(),
-            none(),
-            "");
+                        docstring += "Members:";
+                        for (auto kv : entries) {
+                            auto key = std::string(pybind11::str(kv.first));
+                            auto comment = kv.second[int_(1)];
+                            docstring += "\n\n  " + key;
```
Skylion007, line 1989:
```suggestion
                            docstring += "\n\n  ";
                            docstring += key;
```

Skylion007:
> Great, thanks! I'll wait a day or so to give @Skylion007 I chance to review, too.

Sorry, missed this in my Github notifications.

include/pybind11/pybind11.h
rwgk: Applied in commit c2456d9d3b8b154035f77e0895f7d81fc1cb0ccd

include/pybind11/pybind11.h
rwgk: Also applied in Applied in commit c2456d9d3b8b154035f77e0895f7d81fc1cb0ccd

I also changed the code to systematically use `+=`.

rwgk:
I went ahead with a rebase & applying the review suggestions, to get this into the next smart_holder update.

Waiting for the CI.

rwgk:
The only one CI failure is a known flake (test_gil_scoped DEADLOCK), definitely unrelated to this RP.

knarfS:
Thanks @rwgk  for merging!

Not sure, but maybe https://github.com/pybind/pybind11/issues/2275 can be also closed now ([see test above](https://github.com/pybind/pybind11/pull/2768#issuecomment-755266595)).

rwgk:
> Thanks @rwgk for merging!
> 
> Not sure, but maybe #2275 can be also closed now ([see test above](https://github.com/pybind/pybind11/pull/2768#issuecomment-755266595)).

@AWhetter Is there a chance you could confirm & update or close #2275?

AWhetter:
This does indeed fix that issue!

rwgk:
Thanks @AWhetter!

PR #2772, opened 2021-01-07
Update 'inactive_override_cache' key

### LINKED ISSUE #1922: get_type_overload may encounter false cache hits if derived instances are GC'd

EricCousineau-TRI:
## Description

Avoid false cache hits for virtual overrides by using `__qualname__` as the hashing key. Change storage to `std::string` to avoid memory management mishaps.

Supersedes #2092 - just merge conflict resolution; I'll address any further changes.
Resolves #1922

## Suggested changelog entry:

```rst
Avoid false cache hits for virtual overrides by using ``__qualname__`` as the hashing key. Change storage of key to use ``std::string`` to avoid memory management mishaps.
```

\cc @BetsyMcPhail

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -83,9 +83,9 @@ template <typename value_type>
 using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
 
 struct override_hash {
-    inline size_t operator()(const std::pair<const PyObject *, const char *>& v) const {
+    inline size_t operator()(const std::pair<const PyObject *, std::string>& v) const {
         size_t value = std::hash<const void *>()(v.first);
-        value ^= std::hash<const void *>()(v.second) + 0x9e3779b9 + (value<<6) + (value>>2);
+        value ^= std::hash<std::string>()(v.second);
         return value;
     }
 };
@@ -97,7 +97,7 @@ struct internals {
     type_map<type_info *> registered_types_cpp; // std::type_index -> pybind11's type information
     std::unordered_map<PyTypeObject *, std::vector<type_info *>> registered_types_py; // PyTypeObject* -> base type_info(s)
     std::unordered_multimap<const void *, instance*> registered_instances; // void * -> instance*
-    std::unordered_set<std::pair<const PyObject *, const char *>, override_hash> inactive_override_cache;
+    std::unordered_set<std::pair<const PyObject *, std::string>, override_hash> inactive_override_cache;
     type_map<std::vector<bool (*)(PyObject *, void *&)>> direct_conversions;
     std::unordered_map<const PyObject *, std::vector<PyObject *>> patients;
     std::forward_list<void (*) (std::exception_ptr)> registered_exception_translators;
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -82,10 +82,16 @@ struct type_equal_to {
 template <typename value_type>
 using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
 
+// Adapted from boost::hash_combine(). See also:
+// https://stackoverflow.com/a/27952689/7829525.
+inline void hash_combine(size_t& lhs, size_t rhs ) {
+    lhs ^= rhs + 0x9e3779b9 + (lhs << 6) + (lhs >> 2);
+}
+
 struct override_hash {
     inline size_t operator()(const std::pair<const PyObject *, std::string>& v) const {
         size_t value = std::hash<const void *>()(v.first);
-        value ^= std::hash<std::string>()(v.second);
+        hash_combine(value, std::hash<std::string>()(v.second));
         return value;
     }
 };
@@ -150,7 +156,7 @@ struct type_info {
 };
 
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version
-#define PYBIND11_INTERNALS_VERSION 4
+#define PYBIND11_INTERNALS_VERSION 5
 
 /// On MSVC, debug and release builds are not ABI-compatible!
 #if defined(_MSC_VER) && defined(_DEBUG)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2221,7 +2221,8 @@ inline function get_type_override(const void *this_ptr, const type_info *this_ty
     if (!self)
         return function();
     handle type = type::handle_of(self);
-    auto key = std::make_pair(type.ptr(), name);
+    std::string full_name = type.attr("__qualname__").cast<std::string>() + "." + name;
+    auto key = std::make_pair(type.ptr(), full_name);
 
     /* Cache functions that aren't overridden in Python to avoid
        many costly Python dictionary lookups below */
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2221,7 +2221,7 @@ inline function get_type_override(const void *this_ptr, const type_info *this_ty
     if (!self)
         return function();
     handle type = type::handle_of(self);
-    std::string full_name = type.attr("__qualname__").cast<std::string>() + "." + name;
+    std::string full_name = get_fully_qualified_tp_name((PyTypeObject*) type.ptr()) + "." + name;
     auto key = std::make_pair(type.ptr(), full_name);
 
     /* Cache functions that aren't overridden in Python to avoid


EricCousineau-TRI:
@YannickJadoul / @rgwk Would one of y'all be up for reviewing this?

include/pybind11/detail/internals.h
```diff
@@ -97,7 +97,7 @@ struct internals {
     type_map<type_info *> registered_types_cpp; // std::type_index -> pybind11's type information
     std::unordered_map<PyTypeObject *, std::vector<type_info *>> registered_types_py; // PyTypeObject* -> base type_info(s)
     std::unordered_multimap<const void *, instance*> registered_instances; // void * -> instance*
-    std::unordered_set<std::pair<const PyObject *, const char *>, override_hash> inactive_override_cache;
+    std::unordered_set<std::pair<const PyObject *, std::string>, override_hash> inactive_override_cache;
```
EricCousineau-TRI, line 100: Er... mayhaps this change requires a bump on the internal version.

I can't think of an easy to way to keep `const char*` around, as I'm unaware of what happens with Python strings 'n such...
Ideas?

YannickJadoul:
> @YannickJadoul / @rwgk Would one of y'all be up for reviewing this?

Yes! Tomorrow or during the weekend good enough?

include/pybind11/detail/internals.h
YannickJadoul: Definitely needs to bump that, yes :-/ That does complicate things slightly, release-wise?

YannickJadoul:
Note to self and others: also look at review comments in #2092, and in particular @bstaletic's https://github.com/pybind/pybind11/pull/2092#issuecomment-709562096 (so @bstaletic should probably also review this again).

include/pybind11/detail/internals.h
henryiii: Ouch, not going to do that until the next release, at least. We should bundle all such changes together and then do a bump, I think. Also, I thought standard library templated classes were not valid in an ABI (@jpivarski)? I take it this is why we have so many issues with the ABI? Could we redesign this to just be simple classes and do our own memory management here?

EricCousineau-TRI:
> Note to self and others: also look at review comments in #2092 [...]

:facepalm: Good point! I'll take a gander there!

> Yes! Tomorrow or during the weekend good enough?

Yup, sounds awesome!

include/pybind11/detail/internals.h
EricCousineau-TRI: Yeah, I think that could be done! I'll check out Boris's suggestion in Betsy's original PR, tho, and see if that gives us `const char*` :crossed_fingers:

include/pybind11/detail/internals.h
```diff
@@ -83,9 +83,9 @@ template <typename value_type>
 using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
 
 struct override_hash {
-    inline size_t operator()(const std::pair<const PyObject *, const char *>& v) const {
+    inline size_t operator()(const std::pair<const PyObject *, std::string>& v) const {
         size_t value = std::hash<const void *>()(v.first);
-        value ^= std::hash<const void *>()(v.second) + 0x9e3779b9 + (value<<6) + (value>>2);
+        value ^= std::hash<std::string>()(v.second);
```
EricCousineau-TRI, line 88: From Boris:

> Also, also, the changes to the hash function are really wrong. You don't want to combine hashes with just an xor.

EricCousineau-TRI: Hm... Should I keep the original pattern then?
```
value ^= name_hash + 0x9e3779b9 + (value<<6) + (value>>2)
```
(is there a good reference you'd suggest for this?)

include/pybind11/detail/internals.h
EricCousineau-TRI: Dern, looking at #2520, it's `std::string`. But yeah, I can see about doing some naive memory management.

EricCousineau-TRI: Er, just realized that a few lines below, we quite a few usages of `std::string` hehe
I'll file a separate issue for that.

include/pybind11/detail/internals.h
EricCousineau-TRI: Ah: https://www.variadic.xyz/2018/01/15/hashing-stdpair-and-stdtuple/

EricCousineau-TRI: Done in 7e991aa

include/pybind11/detail/internals.h
YannickJadoul: I believe all of these macros should save us, no? https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/include/pybind11/detail/internals.h#L153-L217

In particular `PYBIND11_STDLIB` and `PYBIND11_BUILD_ABI` being part of `PYBIND11_INTERNALS_ID` should make sure that the STL classes are really the same.

Or is it a different issue from different STL implementations?

EricCousineau-TRI: Done in 7e991aa.
Deferred ABI to #2773 hehe

EricCousineau-TRI: Ah, for soothe. I'll close out #2773. Thanks!

YannickJadoul: I don't know for sure, though! Might be worth hearing out if this is what @henryiii and @jpivarski meant!

henryiii: These macros are evil; they force us to match extensions with exact compilers; while it would be much nicer to be able to interact with any extension that has a matching ABI flag. I think it's because we have lots of STL in our ABI, which causes it to be very picky; if we only interacted through standard types, and internally/externally handled the memory and such, then we could get rid of these macros entirely. That would be the direction to move, rather than adding more (like a string). But it's not "new" to a string.

PyTorch ignores these macros, because the internal JIT compiler may not match the host compiler (the JIT compiler just stores the host's definition of these macros and uses them, IIRC). At least one other user needed this to talk between different compilers, as well, for CUDA I think.

If we bump the ABI version, we'll need to coordinate with conda-forge, they pin on ABI versions and have to trigger a migration (I think) if it gets bumped.

henryiii: TLDR: for now, if we can keep the ABI the same, that's best. Eventually we will need a bump. In the future, maybe we can design internals that work cross-compiler?

EricCousineau-TRI: Ah, I see. Cool, I'll re-open and re-scope #2773 to capture this.

EricCousineau-TRI:
Hm... mystified by this error: https://github.com/pybind/pybind11/pull/2772/checks?check_run_id=1665921586#step:11:70

Gonna re-run to see if it's a flake, as the failing test (`test_iostream.py, test_err` + `test_iostream.cpp, raw_err`) don't seem like they'd be related to this change.

rwgk:
Yes, it's one of the two known thread-related issues: #2754

include/pybind11/detail/internals.h
```diff
@@ -82,10 +82,16 @@ struct type_equal_to {
...
+
 struct override_hash {
-    inline size_t operator()(const std::pair<const PyObject *, const char *>& v) const {
+    inline size_t operator()(const std::pair<const PyObject *, std::string>& v) const {
         size_t value = std::hash<const void *>()(v.first);
-        value ^= std::hash<const void *>()(v.second) + 0x9e3779b9 + (value<<6) + (value>>2);
+        hash_combine(value, std::hash<std::string>()(v.second));
         return value;
     }
 };
```
bstaletic, line 97: Okay, this at least doesn't regress the hash like #2092 does. However this is still not what Boost's `hash_combine` does. Boost's equivalent would be:

```c++
size_t seed = 0;
size_t lhs = std::hash<const void*>()(v.first); // Same as here
size_t rhs = std::hash<std::string>()(v.second); // Same as here
seed ^= lhs + 0x9e3779b9 + (seed << 6) + (seed >> 2); // This line is missing!
seed ^= rhs + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```

Right now, instead of the first `seed ^= lhs...` we're just using `seed = lhs`. With initial seed being 0, boost's `hash_combine` would really do just `lhs += 0x9e3779b9;` before the `lhs ^= rhs + ...` line.

Full disclaimer, I do not claim to be an expert on good hashing functions. I'm just noticing the slight difference. I don't actually know if this should be addressed. I would however trust Boost's implementation on the grounds of "`boost::hash_combine` was written by someone who cares about hashes more than I do". Also, the `boost::hash_combine` has had this magic number since the beginning, [in 2005](https://www.boost.org/doc/libs/1_34_0/doc/html/boost/hash_combine.html).

rwgk:
@henryiii wrote: These macros are evil;

I agree, and want to add a meta comment regarding the version bump: We are paying an enormous tax for this feature: Stifling progress.
Using `std::string` here is clearly the right thing to do, but we're stuck, on something so simple.
What are we gaining for that tax?
What's worth more?
* Ability to improve/fix our product?
* Enable interoperability of extensions compiled with different pybind11 versions?
Clearly, both have a value, but do we have the right balance?
Small attempt to reason out (the best I understand the situation, please correct me if needed):
* The version isolation is awesome, an absolute keeper.
* Users need to worry about version mismatches only if they need interoperability.
* How much of a hardship is it to require that extensions needing interoperability have to be built with similar-enough pybind11 versions?

If the consensus or de-facto conclusion is that the hardship is worse than stifling progress, maybe we're at a stage where we need to branch, to maintain and develop concurrently? Like Python is already developing 3.10 while still supporting 3.6.

include/pybind11/detail/internals.h
YannickJadoul: I made an "abi break" - very-brighly colored - label (https://github.com/pybind/pybind11/labels/abi%20break). Let's use this to indicate PRs that break the ABI and should be pooled together?

YannickJadoul: @henryiii Agreed that it would be amazing if we can get rid of these and use something that's basically shared between all compilers (guess that's POD types and `extern "C"`?). Just curious if ríght now, we have a problem when using STL in the internals, or whether that's at least +- safely worked around.

YannickJadoul:
> If the consensus or de-facto conclusion is that the hardship is worse than stifling progress, maybe we're at a stage where we need to branch, to maintain and develop concurrently? Like Python is already developing 3.10 while still supporting 3.6.

A few things: Python has a much larger maintenance than pybind11. I think we're current happy if the number of issues and PRs stops going up.
Though, apart from that, currently, I don't think there's anything stopping us from merging things into `master` and still releasing a cherry-picked/backported 2.6.2. There's just not really been any urgent things, I think, that need to get merged into `master` immediately? So it's been easier to say "let's wait until we know 2.7 is next". (Maybe I'm wrong here, but that's been my impression.)

At any rate, any ABI change is at least a minor bump, and any ABI breakage needs to be rare (I know from Gitter that people do make different modules that are supposed to interact, and see @henryiii's comment about conda-forge), so it's a good plan to label and "collect" ABI-breaking PRs, and not end up with an exploding ABI version counter.

henryiii:
Breaking the ABI will hit quite a few projects. PyTorch might not be able to load previously JITed code, for example - they would probably need to update PyBind11 past an ABI break exactly on a specific release. And pybind11's distribution is really scattered; some projects pin via submodule, some use PyPI, some use Conda, some use brew - and there’s no way to override it.

include/pybind11/detail/internals.h
henryiii: The current workaround is fine, but something we'd like to possibly redesign to enable wider compatibility in the future, perhaps. In the current design, std::string should be fine, I think it's not worse than unordered map or anything else. I know std::vector is not allowed, so guessing pretty much everything else is out. Somewhere I think @bstaletic just listed the options, which I think is what I was thinking; either we do everything ourselves, or we pay a cost. (Though I feel like we can probably do pretty well). But also haven't really looked at this part of the code much at all, mostly pretty general observations ATM.

include/pybind11/detail/internals.h
```diff
@@ -82,10 +82,16 @@ struct type_equal_to {
 template <typename value_type>
 using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
 
+// Adapted from boost::hash_combine(). See also:
+// https://stackoverflow.com/a/27952689/7829525.
+inline void hash_combine(size_t& lhs, size_t rhs ) {
+    lhs ^= rhs + 0x9e3779b9 + (lhs << 6) + (lhs >> 2);
+}
+
 struct override_hash {
```
YannickJadoul, line 91: Note: the change to this hash is probably also breaking ABI?

YannickJadoul (commented):
Reading the original bug report at https://github.com/pybind/pybind11/issues/1922, this sounded very familiar. I remembered why and found this PR: #2564

So I _think_  (not sure), that this might already be fixed. See https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/include/pybind11/detail/class.h#L220-L229 (introduced by #2564)

Did you have a way of reproducing the original issue, @EricCousineau-TRI? If this already fixed it, maybe we can forget about the whole breaking ABI discussion here (until the next time ;-) ), or at least discuss it from a different perspective (maybe the full type name would still be useful? The `combine_hash` should definitely be kept, though it wouldn't be a change that breaks the ABI anymore, I think?).

EricCousineau-TRI:
> Did you have a way of reproducing the original issue, @EricCousineau-TRI?

Yup, though I think it's a bit picky about which machine, etc. Here's my plan to try the repro:

- [ ] (Result: ?) From the issue, I will try this issue as-is to show that it fails: https://github.com/EricCousineau-TRI/pybind11/commit/97260868f5fb0f1a0ba81c38b173bee99b6c71c4
- [ ] Before #2564 (6364b732~)
- [ ] After #2564, before this PR patch.
- [ ] After this patch.

EricCousineau-TRI:
Oooh, nice! Yup, @YannickJadoul, your PR does fix this! So yeah, I'll close this PR, and carve out the `combine_hash` mebbe (and I can ensure it has the same functionality to preserve ABI).

Thanks!

YannickJadoul:
Ah, nice! Thanks for testing, @EricCousineau-TRI! :-) And sorry for not seeing this earlier, or finding these existing PRs/issues together with that fix; I didn't know you and @BetsyMcPhail had already put in so much effort :-/

> carve out the `combine_hash` mebbe (and I can ensure it has the same functionality to preserve ABI).

Sounds good to me, yes!

YannickJadoul:
Also, it's worth checking if (when we would break ABI), this solution would be better than the current solution? If so, we could replace the current fix, next time we are break ABI anyway?

Issue #2774, opened 2021-01-08
[BUG]  🐍 3.10-dev (ubuntu, mac): TypeError: 'str' object is not callable


EricCousineau-TRI:
Seen this crop up in my PR and another:
```
/home/runner/work/pybind11/pybind11/tests/test_embed/test_interpreter.cpp:244: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  TypeError: 'str' object is not callable
```

Example PRs:
- #2768 | https://github.com/pybind/pybind11/pull/2768/checks?check_run_id=1656501003#step:12:36
- #2772 | https://github.com/pybind/pybind11/pull/2772/checks?check_run_id=1665608739#step:12:35

Guessing it's just a minor thing that'll blow over. I guess we can close this if we don't see it reoccur?

FYI @henryiii @YannickJadoul

YannickJadoul:
Already spotted this in #2768 as well, but haven't had time to look closer. We might now want to wait until 3.10 is released to report this (_ahem_ we'd never let such a thing happen _cough_) ;-)

henryiii:
Before looking at the code, this looked like something is being passed through as a type annotation and being called, but now type annotations are all strings, so it's now calling a string instead of a callable. But have no clue where anything like that would be in our code. It seems to possibly be happening when a module is being loaded? That could be an issue somewhere in Python's stdlib code.

I think 3.10a4 is the first with "from __future__ import annotations" on by default, which is also suspicious.

henryiii:
We don't run these tests on Windows after 3.7 due to the DLL loading changes which we've never worked around, IIRC. So it's likely a problem everywhere.

EricCousineau-TRI:
> We don't run these tests on Windows after 3.7 due to the DLL loading changes which we've never worked around, IIRC. So it's likely a problem everywhere.

Er, I understood your prior post, but I'm not sure how Windows 3.7 w/ DLL relates to 3.10-dev - can you expand on that?

EricCousineau-TRI:
Oh - meaning I should remove the `(ubuntu, mac)` specifier. Derp.

bstaletic:
> We don't run these tests on Windows after 3.7 due to the DLL loading changes which we've never worked around, IIRC. So it's likely a problem everywhere.

In 3.8+ loading DLLs on Winblows changes in this way (`os.add_dll_directory`):

https://docs.python.org/3/whatsnew/3.8.html#changes-in-the-python-api
[...]

PR #2779, opened 2021-01-09
Add `dict::get` that mirrors `dict.get` in Python

lqf96:
## Description
Add shortcut method `dict::get` that mirrors `dict.get` in Python. `dict::get` has four versions: two of them take `py::handle` as key and call `PyDict_GetItem`, while two others take `const char*` as key and call `PyDict_GetItemString`. Like `dict.get` if the default value is not provided it will be `None`.

## Suggested changelog entry:

```rst
Add shortcut method `dict::get` that mirrors `dict.get` in Python
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1292,6 +1292,25 @@ public:
         return PyDict_Contains(m_ptr, detail::object_or_cast(std::forward<T>(key)).ptr()) == 1;
     }
 
+    object get(handle key, handle default_) const {
+        if (PyObject *result = PyDict_GetItem(m_ptr, key.ptr())) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            return reinterpret_borrow<object>(default_);
+        }
+    }
+
+    object get(const char *key, handle default_) const {
+        if (PyObject *result = PyDict_GetItemString(m_ptr, key)) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            return reinterpret_borrow<object>(default_);
+        }
+    }
+
+    object get(handle key) const { return get(key, none()); }
+    object get(const char *key) const { return get(key, none()); }
+
 private:
     /// Call the `dict` Python type -- always returns a new reference
     static PyObject *raw_dict(PyObject *op) {
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -74,6 +74,18 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("dict_contains", [](py::dict dict, const char* val) {
         return dict.contains(val);
     });
+    m.def("dict_get", [](py::dict dict, py::object key, py::object default_) {
+        return dict.get(key, default_);
+    });
+    m.def("dict_get", [](py::dict dict, const char* key, py::object default_) {
+        return dict.get(key, default_);
+    });
+    m.def("dict_get", [](py::dict dict, py::object key) {
+        return dict.get(key);
+    });
+    m.def("dict_get", [](py::dict dict, const char* key) {
+        return dict.get(key);
+    });
 
     // test_str
     m.def("str_from_string", []() { return py::str(std::string("baz")); });
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -92,6 +92,16 @@ def test_dict(capture, doc):
     assert m.dict_contains({42: None}, 42)
     assert m.dict_contains({"foo": None}, "foo")
 
+    d2 = {True: 42, 3: "abc", "abc": 3}
+    assert m.dict_get(d2, True) == 42
+    assert m.dict_get(d2, False) is None
+    assert m.dict_get(d2, 3, "def") == "abc"
+    assert m.dict_get(d2, 5, "def") == "def"
+    assert m.dict_get(d2, "abc") == 3
+    assert m.dict_get(d2, "def") is None
+    assert m.dict_get(d2, "abc", 5) == 3
+    assert m.dict_get(d2, "def", 5) == 5
+
     assert doc(m.get_dict) == "get_dict() -> dict"
     assert doc(m.print_dict) == "print_dict(arg0: dict) -> None"
 


include/pybind11/pytypes.h
```diff
@@ -1292,6 +1292,25 @@ class dict : public object {
...
+    object get(const char *key, handle default_) const {
+        if (PyObject *result = PyDict_GetItemString(m_ptr, key)) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            return reinterpret_borrow<object>(default_);
+        }
+    }
+
+    object get(handle key) const { return get(key, none()); }
+    object get(const char *key) const { return get(key, none()); }
```
bstaletic, line 1312: Why not just make the first two overloads' second parameter defaulted?

lqf96: Done.

YannickJadoul:
First reaction: we already have `item_accessor object_api<T>::operator [](handle key) const` and `item_accessor object_api<T>::operator [](const char *key) const`. But yes, this doesn't contain `get`, so OK, maybe.

Second reaction: don't use that raw C API, as we already have that `operator []` for all `py::objects`. But ah, that's `getattr`, so probably (a lot?) less efficient than `PyDict_GetItem`. So, right. But why don't we have a fast `operator []` for `dict`s, then?

Is this really the case? Is `PyDict_GetItem` a lot faster than `PyObject_GetAttribute`?

Discovery: `list` and `tuple` have this `detail::list_accessor` and `detail::tuple_accessor`, which "specialize" `operator []` for `list` and `tuple`.

So, should we 1) extend this PR to also have an efficient `operator []`, and 2) reuse the `accessor` functionality as it has been used for other classes?

YannickJadoul:
Note, [`PyDict_GetItemString`](https://github.com/python/cpython/blob/d16f6176abdecbb7ab231dc78beccfaa095beff6/Objects/dictobject.c#L3605) is a shallow wrapper (the string is converted to a `PyObject` anyway), and can probably just be implemented similarly to the [`item_accessor object_api<D>::operator[](const char *key)` overload](https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/include/pybind11/pytypes.h#L1583).

YannickJadoul:
Just to be clear, given the existing interfaces on other `py::object` subtypes, I think this PR is a good idea, but I'd like to make it slightly more complete/consistent with the other classes :-)

lqf96:
> First reaction: we already have `item_accessor object_api<T>::operator [](handle key) const` and `item_accessor object_api<T>::operator [](const char *key) const`. But yes, this doesn't contain `get`, so OK, maybe.

A `get` function is necessary if a fallback value is desired when the key is not found. Yes you can do it by calling `operator []` and catching `KeyError` but that seems awkward.

> Second reaction: don't use that raw C API, as we already have that `operator []` for all `py::objects`. But ah, that's `getattr`, so probably (a lot?) less efficient than `PyDict_GetItem`. So, right. But why don't we have a fast `operator []` for `dict`s, then?
> 
> Is this really the case? Is `PyDict_GetItem` a lot faster than `PyObject_GetAttribute`?
>
> Discovery: `list` and `tuple` have this `detail::list_accessor` and `detail::tuple_accessor`, which "specialize" `operator []` for `list` and `tuple`.

It should be, although I'm not absolutely certain... Yes I think `operator []` should be specialized for `dict` so that they enjoy the same benefits of `get`.

> So, should we 1) extend this PR to also have an efficient `operator []`, and 2) reuse the `accessor` functionality as it has been used for other classes?

Personally, yes for 1) and no for 2), because `get` is just about getting some from the dictionary. The set side should be instead handled by `setdefault` (or `PyDict_SetDefault`). If you want I can also address it in this PR.

YannickJadoul:
> A `get` function is necessary if a fallback value is desired when the key is not found. Yes you can do it by calling `operator []` and catching `KeyError` but that seems awkward.

`operator []` does not throw a key error until trying to access it. Have a look at the `accessor<Policy>` template. But `operator bool` seems to unfortunately be deprecated.

> 2), because `get` is just about getting some from the dictionary.

Huh? What do you mean? I'm talking about implementing `dict::operator []` in the same fashion as e.g. `list::operator []`. Why would that not be a good idea?

lqf96:
> Huh? What do you mean? I'm talking about implementing `dict::operator []` in the same fashion as e.g. `list::operator []`. Why would that not be a good idea?

Sorry I misunderstood your suggestion... I just pushed a prototype that specializes `dict::operator []` for faster item access. Due to some problems around exception handling I have to use `PyDict_GetItemWithError` for Python 3 and fall back to `generic_item::get` for Python 2. I also found out I can't reuse the implementation of `dict::operator []` for `get` again because of exception handling.

YannickJadoul:
Argh, apologies, @lqf96. I thought this would be a lot more straightforward, after I looked how small `list::operator []` is :-(
Let's see. If this becomes to big, we revert to your original PR, but now that you've done the work, it would be nice to get this in :-)

Let me give you another quick review.

include/pybind11/pytypes.h
```diff
@@ -613,6 +615,31 @@ struct tuple_item {
...
+struct dict_item {
+    using key_type = object;
+
+    static object get(handle obj, handle key) {
+#if PY_MAJOR_VERSION >= 3
+        if (PyObject *result = PyDict_GetItemWithError(obj.ptr(), key.ptr())) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            if (!PyErr_Occurred())
+                if (PyObject* key_repr = PyObject_Repr(key.ptr()))
```
YannickJadoul, line 628: Why take the `repr`? According to this, that's not done by Python?

```
>>> try:
...     {}[42]
... except Exception as e:
...     exc = e
... 
>>> exc
KeyError(42,)
>>> exc.args
(42,)
>>> exc.args[0]
42
```

So you can just `PyErr_SetObject(PyExc_KeyError, key)` and save a level of indentation, or not?

include/pybind11/pytypes.h
```diff
@@ -613,6 +615,31 @@ struct tuple_item {
...
+
+struct dict_item {
+    using key_type = object;
+
+    static object get(handle obj, handle key) {
+#if PY_MAJOR_VERSION >= 3
+        if (PyObject *result = PyDict_GetItemWithError(obj.ptr(), key.ptr())) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            if (!PyErr_Occurred())
```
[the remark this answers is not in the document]
YannickJadoul: (I had to look this up, since most often, when Python's C API returns `NULL`, the error is already set)

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1312,42 @@ class dict : public object {
 
     size_t size() const { return (size_t) PyDict_Size(m_ptr); }
     bool empty() const { return size() == 0; }
+    detail::dict_accessor operator[](const char *key) const { return {*this, pybind11::str(key)}; }
```
YannickJadoul, line 1315: Nice! I must say this does look good :-)

include/pybind11/pytypes.h
```diff
@@ -613,6 +615,31 @@ struct tuple_item {
         }
     }
 };
+
+struct dict_item {
+    using key_type = object;
+
+    static object get(handle obj, handle key) {
+#if PY_MAJOR_VERSION >= 3
+        if (PyObject *result = PyDict_GetItemWithError(obj.ptr(), key.ptr())) {
```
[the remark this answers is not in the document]
YannickJadoul: We could probably add a `PYBIND11_DICT_GET_ITEM_WITH_ERROR` to `common.h` to the already long list of Python 2 vs. Python 3.

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1312,42 @@ class dict : public object {
...
     detail::dict_iterator begin() const { return {*this, 0}; }
     detail::dict_iterator end() const { return {}; }
     void clear() const { PyDict_Clear(ptr()); }
     template <typename T> bool contains(T &&key) const {
         return PyDict_Contains(m_ptr, detail::object_or_cast(std::forward<T>(key)).ptr()) == 1;
     }
 
+    object get(handle key, handle default_ = none()) const {
+#if PY_MAJOR_VERSION >= 3
+        if (PyObject *result = PyDict_GetItemWithError(m_ptr, key.ptr())) {
```
YannickJadoul, line 1326: Same as above here. I'd be up for using `_PyDict_GetItemWithError` and halve this amount of code.

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1312,42 @@ class dict : public object {
...
+            return reinterpret_borrow<object>(result);
+        } else {
+            if (PyErr_Occurred())
```
YannickJadoul, line 1329: This is probably faster and not much less clean than reusing `operator []` from above? I think I'm OK with this.

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1312,42 @@ class dict : public object {
...
+            if (e.type().ptr() == PyExc_KeyError) {
+                return reinterpret_borrow<object>(default_);
+            } else {
+                throw;
+            }
+        }
+#endif
+    }
+
+    object get(const char *key, handle default_ = none()) const {
```
YannickJadoul, line 1347: To come back to the `setdefault` you mentioned: Python 3 has `PyDict_SetDefault`, but Python 2 doesn't. Since it's also less used than `get`, I'd be fine for leaving this out (for now) and just advising to use `attr("setdefault")(key, default)`.

YannickJadoul (commented):
Nice, thanks for doing the extra effort and adding `operator []`! Almost there, I think :-)

include/pybind11/pytypes.h
lqf96: If `PyDict_SetDefault` is not present in Python 2 then I'd also want to leave it out now...

YannickJadoul: Maybe it's again hidden in one way or the other, but ... eh, not worth it. So I agree.

YannickJadoul: > Maybe it's again hidden in one way or the other

Note: there isn't. It's a function with internal linkage, added to the custom methods: https://github.com/python/cpython/blob/2.7/Objects/dictobject.c#L2048 and https://github.com/python/cpython/blob/2.7/Objects/dictobject.c#L2327

include/pybind11/pytypes.h
lqf96: My bad 🤦‍♂️ no need to call `repr` on the key...

include/pybind11/pytypes.h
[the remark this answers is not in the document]
lqf96: I'm fine with that so long as you think using `_PyDict_GetItemWithError` for Python 2... Yeah this will greatly simplify the code since we don't need to special-case Python 2 now.

include/pybind11/pytypes.h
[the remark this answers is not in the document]
lqf96: The document for [`PyDict_GetItemWithError`](https://docs.python.org/3/c-api/dict.html#c.PyDict_GetItemWithError) is:

> Variant of PyDict_GetItem() that does not suppress exceptions. Return NULL with an exception set if an exception occurred. Return NULL without an exception set if the key wasn’t present.

I'll also add some comments to this PR...

lqf96:
@YannickJadoul Pushed an update with your suggestions... I also have one question on whether I should `inc_ref` the key before calling `PyErr_SetObject`...

include/pybind11/pytypes.h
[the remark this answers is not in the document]
YannickJadoul: Discussed privately with @bstaletic, and he's also fine with this. His extra argument (besides Cython using it and PyPy also providing it): Python 2 is dead, so it won't change this (semi-private) API anymore, anyway :-)

So if you're happy with it, I think we should go for this, adding a macro in `common.h` :-)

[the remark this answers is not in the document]
YannickJadoul: Oops, you already did! Great, thanks! :-)

include/pybind11/pytypes.h
YannickJadoul: At least my review is worth something, then ;-)

YannickJadoul:
> I also have one question on whether I should `inc_ref` the key before calling `PyErr_SetObject`...

I don't think so. (C)Python [increases the reference count in `_PyErr_SetObject`](https://github.com/python/cpython/blob/ace008c531dd685a30c1dd68f9b5ba35f20171cf/Python/errors.c#L116) and other callers in that file decrease the refcount after calling `_PyErr_SetObject`, so we're good like this.

lqf96:
> I don't think so. (C)Python [increases the reference count in `_PyErr_SetObject`](https://github.com/python/cpython/blob/ace008c531dd685a30c1dd68f9b5ba35f20171cf/Python/errors.c#L116) and other callers in that file decrease the refcount after calling `_PyErr_SetObject`, so we're good like this.

Ok i'll then remove it...

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1311,33 @@ class dict : public object {
...
+    detail::dict_accessor operator[](const char *key) const { return {*this, pybind11::str(key)}; }
+    detail::dict_accessor operator[](handle h) const { return {*this, reinterpret_borrow<object>(h)}; }
     detail::dict_iterator begin() const { return {*this, 0}; }
     detail::dict_iterator end() const { return {}; }
     void clear() const { PyDict_Clear(ptr()); }
     template <typename T> bool contains(T &&key) const {
         return PyDict_Contains(m_ptr, detail::object_or_cast(std::forward<T>(key)).ptr()) == 1;
     }
 
+    object get(handle key, handle default_ = none()) const {
```
YannickJadoul, line 1323: Other reviewers: should this be moved out of the class definition? I think it's on the edge of being too long, but good enough and maybe nicer to keep it just as-is than to split it up into two parts.

YannickJadoul (approved):
Thanks, @lqf96. Nothing more from me. I think this can perfectly go in with the next minor release, so I'll add it to that milestone!

The test failure is fine (at least if it's only 3.10-dev); see #2774.

include/pybind11/pytypes.h
```diff
@@ -613,6 +615,30 @@ struct tuple_item {
...
+    static object get(handle obj, handle key) {
+        if (PyObject *result = PYBIND11_DICT_GET_ITEM_WITH_ERROR(obj.ptr(), key.ptr())) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            // NULL with an exception means exception occurred when calling
+            // "__hash__" or "__eq__" on the key
+            // NULL without an exception means the key wasn’t present
+            if (!PyErr_Occurred())
+                // Synthesize a KeyError with the key
+                PyErr_SetObject(PyExc_KeyError, key.ptr());
```
bstaletic, line 631: Is this the right error to set? Should we talk about `__hash__` or `__eq__`?

Should we even care about throwing from these two? In C++, a throw from `std::hash<T>` is just plain insane. What's the state of `std::unordered_map` after a throw from `std::hash` or `std::equal`?

include/pybind11/pytypes.h
```diff
@@ -1285,13 +1311,33 @@ class dict : public object {
...
     }
 
+    object get(handle key, handle default_ = none()) const {
+        if (PyObject *result = PYBIND11_DICT_GET_ITEM_WITH_ERROR(m_ptr, key.ptr())) {
+            return reinterpret_borrow<object>(result);
+        } else {
+            // NULL with an exception means exception occurred when calling
+            // "__hash__" or "__eq__" on the key
+            if (PyErr_Occurred())
+                throw error_already_set();
```
bstaletic, line 1330: Throw or return default?

include/pybind11/pytypes.h
YannickJadoul: As noted, there's two kinds of errors:
1) `NULL` returned but not `PyErr_Occurred()`: key is not present (we have to set an error ourselves)
2) `NULL` returned and `PyErr_Occurred()`: exception occurred during `__hash__` or `__eq__` (and the error is already set).

include/pybind11/pytypes.h
YannickJadoul: See https://github.com/pybind/pybind11/pull/2779#discussion_r555720277

include/pybind11/pytypes.h
bstaletic: Right. My bad. Though the second part is important too.

> Should we even care about throwing from these two? In C++, a throw from std::hash<T> is just plain insane. What's the state of std::unordered_map after a throw from std::hash or std::equal?

The problem is not in inserting a single element and throwing. The problem is rehashing the dictionary and throwing half-way during the operation. At that point, basically anything goes. Thus, should we care about

> 1. `NULL` returned but not `PyErr_Occurred()`: key is not present

YannickJadoul: Rgiht, I kind of missed that. But anyway, yes, Python throws an exception so we need to propagate this as C++ exception. We're not going to change how Python deals with this? I think the current implementation follows accessing `dict`s in pure Python.

rwgk:
I'm NOT opposed to this PR, but optimizing in this situation makes no sense to me. Is this really worth adding code?

If runtime performance matters: use a C++ type (e.g. std::unordered_map), push the operations on the type to C++ in its entirety, wrap that. — That's exactly why pybind11 is so important: it's super easy to wrap another truly optimized function.

If you are dealing with Python objects: you implicitly admit that performance doesn't really matter. Optimizations are expensive in terms of human effort and added code complexity (also human effort, really, for maintenance), with probably insignificant benefit in the big picture.

Skylion007:
@rwgk After looking through the codebase, I think it's definitely worth optimizing since we often call the Python attributes throughout the c++ code which is backed by a py::dict. I think this functionality is worth including and optimizing and this is not the first time this PR has been requested.

rwgk:
OK, sounds good, I support this PR.

rwgk:
Two quick high-level observations:
* clang-tidy failures need to be fixed.
* I see two `throw error_already_set();` in newly added code but those don't seem to be exercised in the added tests. Are they are covered somewhere else already?

lqf96:
@rwgk I haven't been looking at this PR for a long time... Will try to follow up in a few days.

Skylion007:
@lqf96 Yeah. apologies for that. We are cleaning up old PRs. I am fine with the current complexity of this PR, we can always reduce it to a shim that calls "attr()" at worse case.

PR #2782, opened 2021-01-10
Refactor and unify creation of heap types

lqf96:
## Description
Previously, common code for creating heap types are repeated multiple times in `detail/class.h`. For better readability, this PR moves these code into a `make_heap_type` function. `make_heap_type` accepts the name, qualifying name and module name of a class, as well as the base class and metaclass. It also accepts a closure that can be used to configure the class before calling `PyType_Ready`. This is an internal change and does not change the behavior of any public APIs. This will also be helpful for https://github.com/pybind/pybind11/issues/2722#issuecomment-757542235 if we are going to adopt that idea.

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -36,8 +36,45 @@ inline std::string get_fully_qualified_tp_name(PyTypeObject *type) {
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


bstaletic:
Just a thought. Can we use `PyType_FromSpec()`? It should be possible to simplify code with it. I just don't know if it's available on Python 2.

PR #2783, opened 2021-01-10
Update attr.h: fix a warning found by static code analyzer

snnn:
## Description

Update attr.h: fix a warning found by Visual Studio static code analyzer

Severity:Warning
Code:C6323
Description: Use of arithmetic operator on Boolean type(s).
Location:	C:\src\onnxruntime\debug\pybind11\src\pybind11\include\pybind11\attr.h:547

## Suggested changelog entry:

(nothing)

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -544,7 +544,7 @@ template <typename... Extra,
           size_t named = constexpr_sum(std::is_base_of<arg, Extra>::value...),
           size_t self  = constexpr_sum(std::is_same<is_method, Extra>::value...)>
 constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
-    return named == 0 || (self + named + has_args + has_kwargs) == nargs;
+    return named == 0 || (self + named + (has_args ? 1 : 0) + (has_kwargs ? 1 : 0)) == nargs;
 }
 
 PYBIND11_NAMESPACE_END(detail)
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -544,7 +544,7 @@ template <typename... Extra,
           size_t named = constexpr_sum(std::is_base_of<arg, Extra>::value...),
           size_t self  = constexpr_sum(std::is_same<is_method, Extra>::value...)>
 constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
-    return named == 0 || (self + named + (has_args ? 1 : 0) + (has_kwargs ? 1 : 0)) == nargs;
+    return named == 0 || (self + named + std::size_t(has_args) + std::size_t(has_kwargs)) == nargs;
 }
 
 PYBIND11_NAMESPACE_END(detail)
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -544,7 +544,7 @@ template <typename... Extra,
           size_t named = constexpr_sum(std::is_base_of<arg, Extra>::value...),
           size_t self  = constexpr_sum(std::is_same<is_method, Extra>::value...)>
 constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
-    return named == 0 || (self + named + std::size_t(has_args) + std::size_t(has_kwargs)) == nargs;
+    return named == 0 || (self + named + size_t(has_args) + size_t(has_kwargs)) == nargs;
 }
 
 PYBIND11_NAMESPACE_END(detail)


snnn:
Just to make the compiler happy

YannickJadoul:
To be fair, I don't see the issue with adding `bool`s, so I find that quite a bogus/superfluous warning by MSVC. @bstaletic, @henryiii?

henryiii:
This is technically correct, I'd be fine with it. Really, true + true = 2 is a little weird, we are just rather used to it. I'd use it in a pinch, but writing properly for a library is better, IMO.

YannickJadoul:
OK, then; why not. I don't consider this such an anti-pattern (but maybe I've been using Python for too long, where they even make this official with `issubclass(bool, int)` ;-) - which is, to be fair, at the least very ugly).

henryiii:
In Python this is specially ugly, since a bool-like may not be a bool. `or` and `and` both return the object rather than a true bool, so `my_bool = a or b` might not be a bool. MyPy would catch this, but only if you specify `my_bool: bool`, not if you just assume it's 1 or 0-like.

bstaletic:
Here are a few arguments against this:

1. The current master is concise and, in my horrible opinion, more readable.
2. Branches put pressure on branch predictors. What guarantees do we have that this redundant "if true return true else false" will be optimized away?
3. The weakest of arguments - current master is correct.

YannickJadoul:
Amazing how we can bikeshed even these 18 additional bytes :-D Apologies, @snnn, and thanks for making a PR, at any rate! :-)

@bstaletic has a point (and I feel the same), but I'm willing to concede it might be nicer to users to just work around MSVC's (bad?) advice and make things compile with as little warnings as possible. Any self-respecting compiler should optimize this, no? (https://godbolt.org/z/4GK7K6 says it's the same in release for GCC, Clang, and MSVC; not in debug, though)

henryiii:
You could explicitly cast to int to be make the intent clear, that would also avoid the error?

YannickJadoul:
That seems to always be producing the same (even without any optimizations): https://godbolt.org/z/MGvnP4

However, I can't reproduce the warning. What kind of dark magic do I need to perform to see that warning on godbolt?

snnn:
> That seems to always be producing the same (even without any optimizations): https://godbolt.org/z/MGvnP4
> 
> However, I can't reproduce the warning. What kind of dark magic do I need to perform to see that warning on godbolt?

Append " /analyze" to the flags. :-)

snnn:
> You could explicitly cast to int to be make the intent clear, that would also avoid the error?

Yes.

YannickJadoul:
> Append " /analyze" to the flags. :-)

Ah, wow, that's a very strict flag though. So @bstaletic is quite right that this is kind of over the top from MSVC.

But I guess we should look at this from the positive side. If this is the only issue that's found by this flag ... that's not too bad.

henryiii:
Could we add that flag to our testing?

henryiii:
Such as here: https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/tests/CMakeLists.txt#L244

If we don't fix this, we have to add the error skip to https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/include/pybind11/pybind11.h#L23-L32, which I'd rather not do!

include/pybind11/attr.h
```diff
@@ -544,7 +544,7 @@ template <typename... Extra,
           size_t named = constexpr_sum(std::is_base_of<arg, Extra>::value...),
           size_t self  = constexpr_sum(std::is_same<is_method, Extra>::value...)>
 constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
-    return named == 0 || (self + named + has_args + has_kwargs) == nargs;
+    return named == 0 || (self + named + (has_args ? 1 : 0) + (has_kwargs ? 1 : 0)) == nargs;
```
henryiii, line 547:
```suggestion
    return named == 0 || (self + named + int(has_args) + int(has_kwargs)) == nargs;
```
This would be the casting version.

YannickJadoul: `size_t`, before the next compiler comes whining about signed/unsigned operations ;-) (which is actually a valid concern, if you ask me). Is there a difference between `size_t(has_args)` and `static_cast<size_t>(has_args)`? I guess not.

henryiii: I think we allow c-style casts. The latter would be a C++ one, but less readable. I'm not sure I like this better - the if is logically correct, add 1 if true else 0 is logically what we are doing. This and the other assumes that true is 1, which is specified by the standard but still a arbitrary choice, this would not work in bash, for example. ;)

snnn:
> > Append " /analyze" to the flags. :-)
> 
> Ah, wow, that's a very strict flag though. So @bstaletic is quite right that this is kind of over the top from MSVC.
> 
> But I guess we should look at this from the positive side. If this is the only issue that's found by this flag ... that's not too bad.

Yes, this is the only one. :-)

include/pybind11/attr.h
YannickJadoul: OK, agreed :-) I think converting `bool`s to integer types is pretty fine, though. I'd distrust anything that doesn't do this after you ask it to.

Final thing: should this be `size_t` instead of `std::size_t`? I know that's correct, but in the signature, it's also `size_t` without `std::`.

snnn: Updated.

YannickJadoul (approved):
Eh, I'm done bikeshedding this. Guess there's no harm in this fix.

Thanks for holding on, @snnn!

Issue #2775, opened 2021-01-08
[QUESTION] Why passing a float64 list to c++, c++ get vector<float> but not double?


liym27:
<!--

If you have a question, please check the following first:

1. Check if your question has already been answered in the [FAQ][] section.
2. Make sure you've read the [documentation][]. Your issue may be addressed there.
3. If those resources didn't help and you only have a short question (not a bug report), consider asking in the [Gitter chat room][]
4. Search the [issue tracker][], including the closed issues, to see if your question has already been asked/answered. +1 or comment if it has been asked but has no answer.
5. If you have a more complex question which is not answered in the previous items (or not suitable for chat), please fill in the details below.
6. Include a self-contained and minimal piece of code that illustrates your question. If that's not possible, try to make the description as clear as possible.

[FAQ]: http://pybind11.readthedocs.io/en/latest/faq.html
[documentation]: https://pybind11.readthedocs.io
[issue tracker]: https://github.com/pybind/pybind11/issues
[Gitter chat room]: https://gitter.im/pybind/Lobby

*After reading, remove this checklist.*
--!>

I want to pass a float64 list `[2.0**128]` to c++, but c++ only received `vector<float>`, precision is lost. I print the vector<float>, only get 'inf'

bstaletic:
Probably because you're using a vector of floats? Show your code, as we still haven't got our crystal balls to work.

liym27:
For example:

pybind_sum will call `Func(vector<float> nums)` but not `Func(vector<double> nums)`, 
```python
import numpy as np
a = [2.0**129, 2.0**129] # Note float32:[-2^128, 2^128)

out = pybind_func(a)
```

```c++
m.def("pybind_func", Func);

double Func(vector<double> nums){
  // ...
  // do something
  // ...
}

double Func(vector<float> nums){
  
   // Note: nums[0] and nums[1] are 'inf';
  // do something
  // ...
}

```

bstaletic:
>     m.def("pybind_func", Func);

That line wouldn't compile if `Func` is an overload set. So what did you actually do in your code?

YannickJadoul:
Agreeing with @bstaletic. pybind11 doesn't magically call an overload. You specified somewhere that you want to bind the overload with `vector<float>` instead of `vector<double>`.

liym27:
Sorry, the code is as follows:

```python
import numpy as np
a = [2.0**129, 2.0**129] # Note float32:[-2^128, 2^128)

[...]

Issue #1625, opened 2018-11-27
std::variant of integral and bool populates wrong alternative


johnshaw:
Populating a variant containing an integral type and a `bool` from a Python bool may select the wrong alternative, depending on the ordering of alternatives.

Consider the function `void foo(std::variant<int, bool> v)`. When called from Python with the expression `foo(True)`, the value `v` in C++ will end up with the `int` alternative populated with a value of `1`. If the order of the alternatives are swapped such that `bool` occurs before `int`, the boolean alternative will be populated, as expected.

The reason this occurs is that the `type_caster` for integral types is defined for all types satisfying `is_arithmetic`, which `bool` satisfies, combined with the fact that loading a `bool` into an integer type is not considered a "conversion" in the context of the `load` method, so it is selected on the first pass.

I think this should be improved so that bool doesn't match the arithmetic type caster on the first pass.

johnshaw:
Turns out the arithmetic conversion is even more permissive than this. It will consider any conversion to the arithmetic type as valid. For example, my code has a custom decimal class in c++ with a `type_caster` defined for conversion to and from `decimal.Decimal`. Assigning a Python `Decimal` value into a `std::variant<long, MyDecimalType>` will result in assignment to the `long` alternative, truncating the fractional part. I would not expect this conversion to be accepted on the first pass, especially since it's lossy.

Monomix:
Also came across this issue - the variant population also doesn't appear to be completely deterministic. When calling a function with a `std::variant<bool, int32_t, int64_t, uint32_t, uint64_t>` as a parameter and passing in an `int` from the Python side, the selected alternative is sometimes a `bool`, and sometimes some flavour of integer - it appears to change randomly.

tarcisiofischer:
I can confirm that this is still happening, but doesn't seem random.
What I did to "fix" this issue is to change the order of the `std::variant`, and set the `bool` type to be the first (I was assuming that this list's order was being used to try the type conversion). - It worked (And I'm amazed).

So, if someone is trying to do this, my workaround suggestion is this: Change the order of the std::variant and set bool, before ints.

Example: Instead of
```
std::map<std::string, std::variant<
        int,
        bool>> value_map;
```

Do this:

```
std::map<std::string, std::variant<
        bool,
        int>> value_map;
```

m4ce:
Just bumped into this - is this going to be fixed any time soon?

wjakob:
[...]

PR #2784, opened 2021-01-11
Documenting that the order of alternatives for variant types matters, and follows the same rules as overload resolution

YannickJadoul:
## Description

This issue keeps coming back (#2775), even though it has been answered before (#1625, and who knows where else).

## Suggested changelog entry:

Just docs. No need for a changelog entry, I think.


YannickJadoul:
Note: should we disable the 10-dev builds for now, while we wait for resolution of #2774?

bstaletic:
I've asked multiple times before, but here goes...

Should we close all currently open issues that come down to "specified overload order matters"?

YannickJadoul:
> Should we close all currently open issues that come down to "specified overload order matters"?

Are there still any open?

Also, this PR is just for `variant`. Function overloads have been documented for ages, now.

PR #2791, opened 2021-01-13
ci: use fixed action

henryiii:
Trying out https://github.com/jwlawson/actions-setup-cmake/issues/18. Followup to #2790.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,9 +71,7 @@ jobs:
       run: echo "BOOST_ROOT=$BOOST_ROOT_1_72_0" >> $GITHUB_ENV
 
     - name: Update CMake
-      uses: jwlawson/actions-setup-cmake@v1.5
-      with:
-        cmake-version: 3.19.2
+      uses: jwlawson/actions-setup-cmake@v1.6
 
     - name: Cache wheels
       if: runner.os == 'macOS'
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Wow, that's amazingly fast!

YannickJadoul:
Failing tests are #2774. Let's get this in!

.github/workflows/ci.yml
```diff
@@ -363,10 +361,8 @@ jobs:
     - name: Update pip
       run: python3 -m pip install --upgrade pip
 
-    - name: Setup CMake 3.18
-      uses: jwlawson/actions-setup-cmake@v1.5
-      with:
-        cmake-version: 3.18
```
YannickJadoul, line 369: 😳 Whoops, bit late, after I just merged, but this is 3.18.

Was there a reason for it to be 3.18? I guess not, because you also removed the name change, @henryiii?

henryiii: If it's passing, I think it's okay, since we test each version elsewhere. But I didn't see that it was 3.18, I thought it was an inconsistently named 3.19.

henryiii: There might have been a bug in an early version of 3.19, but hopefully if so, it is fixed.

YannickJadoul: Tests are passing, so we should be fine?

PR #2792, opened 2021-01-13
Disable builds for 3.10.0a4, and enable a nightly 3.10-dev build

YannickJadoul:
## Description

Temporary fix for #2774. Meanwhile, this also adds 3.10-dev nightly builds from [deadsnakes](https://github.com/deadsnakes/action).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,9 +42,6 @@ jobs:
             python: 3.6
             args: >
               -DPYBIND11_FINDPYTHON=ON
-          - runs-on: ubuntu-latest
-            python: 3.10-dev
-            nightly: true
 
         # These items will be removed from the build matrix, keys must match.
         exclude:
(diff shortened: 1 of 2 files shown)


YannickJadoul:
@henryiii, this seems to be working. But I'm not sure the nightly Python versions are really worth the extra complexity (and additional almost-never-executed "Setup Python X.Y (nightly)" step). Maybe it should be a separate matrix of jobs, where we also test some debug versions? I don't know.

At any rate, feel free to push to this branch to propose changes :-)

YannickJadoul:
> @henryiii, this seems to be working.

Neeeeeeeeevermind that. Why can't it find the correct Python? :-(

henryiii:
I'd add a new job at least, I think.

YannickJadoul:
Heh...

```
Run deadsnakes/action@v2.1.0
  with:
    python-version: 3.10-dev
    debug: false
```

And yet `10-dev` is still acting as if `debug` was specified. I might need to fix my own PR deadsnakes/action#18

YannickJadoul:
Submitted a fix: deadsnakes/action#20

@henryiii, if you want to merge this before deadsnakes/action#20 would be merged, that's possible. Either as-is (with my branch, temporarily), or without the last commit (which will install the 3.10-dev debug version, but eh, should be fine).

YannickJadoul:
> @henryiii, if you want to merge this before [deadsnakes/action#20](https://github.com/deadsnakes/action/pull/20) would be merged, that's possible. Either as-is (with my branch, temporarily), or without the last commit (which will install the 3.10-dev debug version, but eh, should be fine).

That went fast. Scratch that, the PR already got merged, and a new release released :-)

Python 2.7 doesn't work with `deadsnakes/action` (because there's no `python2.7-venv` package, and no `venv`), but ... well let's cross that bridge (and make that PR) when we need to?

PR #2793, opened 2021-01-14
fix: Don't override global settings for VISIBILITY if set

### LINKED ISSUE #2479: [BUG] pybind11_add_module forcing -fvisiblity=hidden is a bit too aggressive?

henryiii:
## Description

Sometimes programmers want to control this, and while it can be changed after the fact, it's commonly set via a CMAKE_ variable; if that variable is set, we should respect that (like the CMAKE_INTERPROCEDURAL_OPTIMIZATION setting).

Closes #2479

## Suggested changelog entry:

```rst
* Respect the setting for ``CMAKE_CXX_VISIBILITY_PRESET`` if defined.
```

### DIFF
--- a/tools/pybind11Config.cmake.in
+++ b/tools/pybind11Config.cmake.in
@@ -87,7 +87,7 @@ you can either use the basic targets, or use the FindPython tools:
   target_link_libraries(MyModule2 pybind11::headers)
   set_target_properties(MyModule2 PROPERTIES
                                   INTERPROCEDURAL_OPTIMIZATION ON
-                                  CXX__VISIBILITY_PRESET ON
+                                  CXX_VISIBILITY_PRESET ON
                                   VISIBLITY_INLINES_HIDDEN ON)
 
 If you build targets yourself, you may be interested in stripping the output
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -214,8 +214,18 @@ function(pybind11_add_module target_name)
     target_link_libraries(${target_name} PRIVATE pybind11::python2_no_register)
   endif()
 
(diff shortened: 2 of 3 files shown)


YannickJadoul (approved):
Yeah, this seems like a fair enough fix.

PR #2796, opened 2021-01-14
fix: STATIC and SHARED flags not being detected

heyer2:
I noticed SHARED and STATIC flags no longer being respected after updating pybind11. This fixes it.

## Suggested changelog entry:

```rst
* Fix ``STATIC`` / ``SHARED`` being ignored in FindPython mode.
```


henryiii:
This only fixes the new FindPython, assuming there is nothing wrong with the classic version?

heyer2:
Classic one seems fine to me.

henryiii (approved):

YannickJadoul (approved):

YannickJadoul:
Thanks, @heyer2!

Issue #1946, opened 2019-10-04
Test failure with Python 3.8.0


susilehtola:
pybind11 version 2.4.2 fails to build from source in Fedora rawhide with Python 3.8.0rc1:
```
_____________________________ test_class_refcount ______________________________
    @pytest.unsupported_on_pypy
    def test_class_refcount():
        """Instances must correctly increase/decrease the reference count of their types (#1029)"""
        from sys import getrefcount
    
        class PyDog(m.Dog):
            pass
    
        for cls in m.Dog, PyDog:
            refcount_1 = getrefcount(cls)
            molly = [cls("Molly") for _ in range(10)]
            refcount_2 = getrefcount(cls)
    
            del molly
            pytest.gc_collect()
            refcount_3 = getrefcount(cls)
    
>           assert refcount_1 == refcount_3
E           assert 6 == 16
test_class.py:255: AssertionError
```
as reported in Fedora ticket https://bugzilla.redhat.com/show_bug.cgi?id=1758474

wjakob:
Oh oh, that's not good. It sounds very much like a regression in the Python beta version though. (Something is leaking references in Python, which is tricky to track down because it just wastes memory instead of e.g. causing a crash.)

wjakob:
I'm super-busy with a grant deadline in the next 2 weeks and unfortunately won't have free clock cycles to get to this-- it would be great if somebody could help out here. The task would be to `git bisect` Python master to the commit that caused this test to fail and then post a ticket on the Python bugtracker. (Assuming that my hunch about this being a Python regression is correct.)

sizmailov:
@wjakob It's not a bug, it's change in python3.8 C API

> Instances of heap-allocated types (such as those created with PyType_FromSpec()) hold a reference to their type object. Increasing the reference count of these type objects has been moved from PyType_GenericAlloc() to the more low-level functions, PyObject_Init() and PyObject_INIT(). This makes types created through PyType_FromSpec() behave like other classes in managed code.
Statically allocated types are not affected.
For the vast majority of cases, there should be no side effect. However, types that manually increase the reference count after allocating an instance (perhaps to work around the bug) may now become immortal. To avoid this, these classes need to call Py_DECREF on the type object during instance deallocation.

https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api

wjakob:
[...]

PR #2797, opened 2021-01-15
Revert workaround of test_virtual_functions.py from #2746

YannickJadoul:
## Description

We made workaround in #2746 for an issue flagged by Valgrind we didn't understand/track down. So far it's unclear whether this is a real issue (a very tricky one, then, because this small change triggers it), or an issue with Valgrind. In any case, we'd like to revert this :-)

### DIFF
--- a/tests/test_virtual_functions.py
+++ b/tests/test_virtual_functions.py
@@ -251,7 +251,8 @@ def test_dispatch_issue(msg):
                 == 'Tried to call pure virtual function "Base::dispatch"'
             )
 
-            return m.dispatch_issue_go(PyClass1())
+            p = PyClass1()
+            return m.dispatch_issue_go(p)
 
     b = PyClass2()
     assert m.dispatch_issue_go(b) == "Yay.."


bstaletic:
While everyone can take a look at the valgrind report, here are the things to notice:

1. The block really was freed in `pybind11_meta_dealloc`.
2. Then the interpreter starts to shut down and we get another free in `free_keys_object`.
3. Both frees get passed the `PyObject*` corresponding to `p` (see this pull request's diff).
4. If you as much as look at this test funny, the double free doesn't happen.
4.1. I've forgotten which of the two frees doesn't happen.

bstaletic (commented):
This was fun. First of all, here's how to reliably repro this issue with a self-contained example:

```c++
#include <Python.h>

static void pybind11_object_dealloc(PyObject *self) {
	auto type = Py_TYPE(self);
	type->tp_free(self);
	Py_DECREF(type);
}

static PyModuleDef pybind11_module_def_m{
	.m_base = PyModuleDef_HEAD_INIT,
	.m_name = "m",
	.m_size = -1,
};

static PyObject *pybind11_init_impl_m() {
	auto m = PyModule_Create(&pybind11_module_def_m);
	PyType_Slot slots[] = {
		{Py_tp_dealloc, (void*)pybind11_object_dealloc},
		{0, nullptr}
	};
	PyType_Spec spec{"m.B", sizeof(PyObject), 0, Py_TPFLAGS_BASETYPE | Py_TPFLAGS_HEAPTYPE, slots};
	auto type = PyType_FromSpec(&spec);
	PyObject_SetAttrString(m, "B", (PyObject*)type);
	return m;
}

int main() {
	PyImport_AppendInittab("m", pybind11_init_impl_m);
	Py_InitializeEx(1);
	PyRun_SimpleString("import m\n"
			   "def t():\n"
			   "  class A:\n"
			   "    class B(m.B):pass\n"
			   "    b=B()\n"
			   "t()\n");
	Py_Finalize();
}
```

A few interesting things one can point out:

1. The above snippet doesn't depend on pybind11. The only thing slightly resembling pybind11 is `pybind11_object_dealloc()`. Pay attention to it.
2. The snippet is slightly evil - the heap type flags omit `Py_TPFLAGS_DEFAULT` and thus don't play well with stackless python and also have bad performance. Also, no `tp_repr` or `tp_str` means you're not allowed to print `m.B()`.
2.1. I was trying to be terse, so the code does just enough to "anger" valgrind.

This same snippet had ~4 of these use-after-free errors in python 3.8. In python 3.7 there were hundreds of similar errors, but back then things were different. Also, python 3.10 has this use-after-free error as well.

References for further reading:

1. https://github.com/pybind/pybind11/issues/1946
2. https://bugs.python.org/issue35810
3. https://github.com/sizmailov/pybind11/commit/3beea65dcb6559d4bfb114785fb9c6552a05eaaa
4. https://github.com/pybind/pybind11/commit/6cb584e9de6e8d54f5576c299a308f89bfdcb519
5. https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api

That should explain the "things were different in days of python3.7". The suggested `tp_dealloc` from link number 5 does the following:

```
static void
foo_dealloc(foo_struct *instance) {
    PyObject *type = Py_TYPE(instance);
    PyObject_GC_Del(instance);
#if PY_VERSION_HEX >= 0x03080000
    // This was not needed before Python 3.8 (Python issue 35810)
    Py_DECREF(type);
#endif
}
```

If we take a look at [`pybind11_object_dealloc` as it is in pybind11](https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/class.h#L419-L438), we see the same approach. It really looks like we're doing everything by the book!

If anyone is wondering what happens if we just skip the `Py_DECREF` in python3.8+, well, leaks happen. We start leaking references, as in https://github.com/pybind/pybind11/issues/1946.

Considering that before python3.8 these use-after-free errors were plenty and that the above snippet is irreducible without hiding the error, I'm pretty sure this is a CPython bug. An obscure one, but still.

Seems like we, yet again, have to choose between leaking and (obscure) memory errors. Now who wants to bet that the actual fix for this issue will, yet again, break some part of CPython API?

___

As for this specific pull request, I think we shouldn't merge it until CPython actually fixes the error. Before anyone asks, I haven't submitted a report on BPO.

YannickJadoul:
Issue submitted to bugs.python.org by @bstaletick, btw: https://bugs.python.org/issue42961

Issue #2523, opened 2020-09-22
[FEAT] Wrap bytearray with pybind11::bytearray


fritzr:
It would be nice to have a `pybind11::bytearray` type for directly manipulating `bytearray` objects, similar to `pybind11::bytes`. This way a wrapper function could allocate and return a native mutable Python buffer. A toy example:

```c++
#include <pybind11/pybind11.h>
namespace py = pybind11;

extern size_t decompressed_size(const void *in, size_t in_size);
extern int decompress(const void *in, size_t in_size, void *out, size_t out_size);

PYBIND11_MODULE(mycompression, m) {
    m.def("decompress_mutable", [](py::buffer compressed_data) {
        py::buffer_info in_inf = compressed_data.request();
        py::bytearray out(decompressed_size(in_inf.ptr, in_inf.len));
        (void) decompress(in_inf.ptr, in_inf.size, out.data(), out.size());
        return out;
    });
}
```

This example borrows the idea of exposing `data()` and `size()` methods on the `py::bytearray` type which was suggested for `py::bytes` in #2517. These could use `PyByteArray_AsString` and `PyByteArray_GetSize` with appropriate error checking under the hood.

If deemed worthwhile, this should be simple to implement using the [`PyByteArray_*` family of functions](https://docs.python.org/c-api/bytearray.html). These functions mirror the `PyBytes_*` family currently in use for `pybind11::bytes`, the definition of which is only some 30 lines of code.

YannickJadoul:
This seems viable, indeed :-) Do you know how similar Python 2 and 3 when it comes to `bytearray`?

in42:
Reading this answer and the following documentation:
https://stackoverflow.com/questions/1740696/bytes-vs-bytearray-in-python-2-6-and-3
https://docs.python.org/2.7/library/functions.html?highlight=bytearray#bytearray
https://docs.python.org/3/library/stdtypes.html?highlight=bytearray#bytearray
it seems to me that the python 2 and 3 `bytearray` are exactly similar.

YannickJadoul:
@in42 Great! Then this is definitely something that's possible to add to pybind11 :-)

YannickJadoul:
Added by @in42 in #2799!

PR #2799, opened 2021-01-16
Add pybind11::bytearray

in42:
## Description

Implement #2523 

## Suggested changelog entry:

```rst
Add pybind11::bytearray
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1061,6 +1061,21 @@ inline str::str(const bytes& b) {
 
 /// \addtogroup pytypes
 /// @{
+class bytearray : public object {
+public:
+    PYBIND11_OBJECT(bytearray, object, PyByteArray_Check)
+
+    bytearray(const char *c, size_t n)
+        : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytes object!");
+    }
+};
+// Note: breathe >= 4.17.0 will fail to build docs if the below two constructors
+// are included in the doxygen group; close here and reopen after as a workaround
+/// @} pytypes
+
+/// \addtogroup pytypes
+/// @{
 class none : public object {
 public:
     PYBIND11_OBJECT(none, object, detail::PyNone_Check)
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1080,8 +1080,8 @@ public:
 
     operator std::string() const {
         char *buffer = PyByteArray_AS_STRING(m_ptr);
-        Py_ssize_t size = PyByteArray_GET_SIZE(m_ptr);
-        return std::string(buffer, size);
+        ssize_t size = PyByteArray_GET_SIZE(m_ptr);
+        return std::string(buffer, static_cast<size_t>(size));
     }
 };
 // Note: breathe >= 4.17.0 will fail to build docs if the below two constructors
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1079,8 +1079,9 @@ public:
     bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
 
     operator std::string() const {
-        char *buffer = PyByteArray_AsString(m_ptr);
-        return std::string(buffer);
+        char *buffer = PyByteArray_AS_STRING(m_ptr);
+        Py_ssize_t size = PyByteArray_GET_SIZE(m_ptr);
+        return std::string(buffer, size);
     }
 };
 // Note: breathe >= 4.17.0 will fail to build docs if the below two constructors
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1067,7 +1067,15 @@ public:
 
     bytearray(const char *c, size_t n)
         : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
-        if (!m_ptr) pybind11_fail("Could not allocate bytes object!");
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
+    // Allow implicit conversion:
+    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+
+    operator std::string() const {
+        char *buffer = PyByteArray_AsString(m_ptr);
+        return std::string(buffer);
     }
 };
 // Note: breathe >= 4.17.0 will fail to build docs if the below two constructors
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1065,6 +1065,11 @@ class bytearray : public object {
 public:
     PYBIND11_OBJECT(bytearray, object, PyByteArray_Check)
 
+    bytearray()
+        : object(PyByteArray_FromStringAndSize("", (ssize_t) 0), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
     bytearray(const char *c, size_t n)
         : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1078,8 +1078,6 @@ public:
     size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
 
     explicit operator std::string() const {
-        if (!m_ptr) pybind11_fail("Bytearray object is null!");
-
         char *buffer = PyByteArray_AS_STRING(m_ptr);
         ssize_t size = PyByteArray_GET_SIZE(m_ptr);
         return std::string(buffer, static_cast<size_t>(size));
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1078,6 +1078,8 @@ public:
     size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
 
     explicit operator std::string() const {
+        if (!m_ptr) pybind11_fail("Bytearray object is null!");
+
         char *buffer = PyByteArray_AS_STRING(m_ptr);
         ssize_t size = PyByteArray_GET_SIZE(m_ptr);
         return std::string(buffer, static_cast<size_t>(size));
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1078,6 +1078,8 @@ public:
     // Allow implicit conversion:
     bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
 
+    size_t size() const { return PyByteArray_Size(m_ptr); }
+
     operator std::string() const {
         char *buffer = PyByteArray_AS_STRING(m_ptr);
         ssize_t size = PyByteArray_GET_SIZE(m_ptr);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1078,7 +1078,7 @@ public:
     // Allow implicit conversion:
     bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
 
-    size_t size() const { return PyByteArray_Size(m_ptr); }
+    size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
 
     operator std::string() const {
         char *buffer = PyByteArray_AS_STRING(m_ptr);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1063,24 +1063,22 @@ inline str::str(const bytes& b) {
 /// @{
 class bytearray : public object {
 public:
-    PYBIND11_OBJECT(bytearray, object, PyByteArray_Check)
-
-    bytearray()
-        : object(PyByteArray_FromStringAndSize("", (ssize_t) 0), stolen_t{}) {
-        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
-    }
+    PYBIND11_OBJECT_CVT(bytearray, object, PyByteArray_Check, PyByteArray_FromObject)
 
     bytearray(const char *c, size_t n)
         : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
     }
 
+    bytearray()
+        : bytearray("", 0) {}
+
     // Allow implicit conversion:
     bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
 
     size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
 
-    operator std::string() const {
+    explicit operator std::string() const {
         char *buffer = PyByteArray_AS_STRING(m_ptr);
         ssize_t size = PyByteArray_GET_SIZE(m_ptr);
         return std::string(buffer, static_cast<size_t>(size));
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1073,8 +1073,7 @@ public:
     bytearray()
         : bytearray("", 0) {}
 
-    // Allow implicit conversion:
-    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+    explicit bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
 
     size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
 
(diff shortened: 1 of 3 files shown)


YannickJadoul:
Nice; thanks! This is still "[WIP]", I see, but just ping me whenever you're ready for review :-)

I guess we'd need a few more useful methods, and of course some tests (`test_pytypes.{cpp,py}` is probably where you want to add a few things ;-) )

in42:
> Nice; thanks! This is still "[WIP]", I see, but just ping me whenever you're ready for review :-)
> 
> I guess we'd need a few more useful methods, and of course some tests (`test_pytypes.{cpp,py}` is probably where you want to add a few things ;-) )

Hi, I have added tests and few more methods. Can you review the patch? I am especially not sure if I have added the tests correctly or not.

in42:
> Nice; thanks! This is still "[WIP]", I see, but just ping me whenever you're ready for review :-)
> 
> I guess we'd need a few more useful methods, and of course some tests (`test_pytypes.{cpp,py}` is probably where you want to add a few things ;-) )

Also do I need to add to the documentation?

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,34 @@ inline str::str(const bytes& b) {
...
+        : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
+    // Allow implicit conversion:
+    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+
+    operator std::string() const {
+        char *buffer = PyByteArray_AsString(m_ptr);
+        return std::string(buffer);
```
bstaletic, line 1083: 1. Get the size as well, with `PyByteArray_GET_SIZE`, because `std::string{buffer, size}` is more efficient.
2. Do we need to `free(buffer)` after passing it off to the string?
3. Can `PyByteArray_AsString()` ever return `NULL`? Perhaps we need to guard against it.

in42: 1. Done
2. I don't think we need to free the buffer since PyByteArray_AsString is not allocating anything it is just returning a pointer to already allocated bytearray as far as I understand.
3. PyByteArray_AsString() will return NULL only if m_ptr which it will never be as it is checked in the constructor. So in fact I am now using macro versions of the methods as documented here: https://docs.python.org/3/c-api/bytearray.html which do not check if m_ptr is NULL.

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,35 @@ inline str::str(const bytes& b) {
     m_ptr = obj.release().ptr();
 }
 
+/// \addtogroup pytypes
+/// @{
+class bytearray : public object {
+public:
+    PYBIND11_OBJECT(bytearray, object, PyByteArray_Check)
```
YannickJadoul, line 1066: What about `PYBIND11_OBJECT_CVT(bytearray, object, PyByteArray_Check, PyByteArray_FromObject)`? This would allow to call `py::bytearray(some_other_object_with_buffer_protocol)`.

YannickJadoul (approved):
Nice! Thanks for this :-)

Possibly useful functions (mostly stolen from https://docs.python.org/3/c-api/bytearray.html):
- `char *get()` and/or `operator char*()`? I think this wouldn't copy, like string does, and would allow to change the contents? Maybe that's a reason to not want this, though.
- `size_t size()`; this seems potentially useful.
- `bytesarray concat(bytesarray)`: `PyByteArray_Concat` exists. Not sure how useful it would be, though? Not crucial to have, afaic.
- `resize`: `PyByteArray_Resize` exists.
- Something to assign and change the contents of the `bytearray`? I thought that was the whole point of a `bytearray`, being a mutable `bytes`? @bstaletic, any thoughts?

YannickJadoul (commented):
(Whoops, sorry; didn't mean to approve yet. I definitely will, but there's a few more things to resolve.)

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,35 @@ inline str::str(const bytes& b) {
...
+    bytearray(const char *c, size_t n)
+        : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
+    // Allow implicit conversion:
+    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+
+    operator std::string() const {
+        char *buffer = PyByteArray_AS_STRING(m_ptr);
```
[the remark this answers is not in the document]
bstaletic: A NULL pointer would likely segfault. A pointer to, say, `PyLongObject`, reinterpreted as a pointer to `PyByteArrayObject` would be more dangerous, because that one might segfault, but also might just evaluate to corrupt data. I do not know how well should we try to protect the users from reckless casting...

Regarding the "macros don't check for NULL", despite what the documentation implies, the "regular" `PyByteArray_AsString()` also doesn't check anything. It has some `assert()`s, but those are turned off in regular use.

bstaletic (commented):
> char *get() and/or operator char*()? I think this wouldn't copy, like string does, and would allow to change the contents? Maybe that's a reason to not want this, though

`data()`? That would be in line with STL.

> size_t size(); this seems potentially useful.

Agreed.

> bytesarray concat(bytesarray): PyByteArray_Concat exists. Not sure how useful it would be, though? Not crucial to have, afaic.

`operator+=()`?

> resize: PyByteArray_Resize exists.

Agreed.

> Something to assign and change the contents of the bytearray? I thought that was the whole point of a bytearray, being a mutable bytes? @bstaletic, any thoughts?

`operator[](ssize_t)`?

With all the overloaded operators, it would actually feel like a C++ type.

include/pybind11/pytypes.h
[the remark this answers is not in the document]
YannickJadoul: > A NULL pointer would likely segfault.

Yes. The issue is that for some `py::object` subclasses, we allow `NULL` as valid, and for some not. And it's not very clear which is which, or what's the correct thing to do.

> A pointer to, say, `PyLongObject`, reinterpreted as a pointer to `PyByteArrayObject` would be more dangerous, because that one might segfault, but also might just evaluate to corrupt data. I do not know how well should we try to protect the users from reckless casting...

Yes, but that's why it's called `reinterpret_steal`/`reinterpret_borrow`? We've already fixed the constructors, but I do think that `reinterpret_*` shouts to the user _"you're reponsible"_. So I wouldn't worry about this; if you tell pybind11 that somethings a `bytearray`, who are we to question that?

> Regarding the "macros don't check for NULL", despite what the documentation implies, the "regular" `PyByteArray_AsString()` also doesn't check anything. It has some `assert()`s, but those are turned off in regular use.

Ugh; that's very ugly. But yes, now you say so, I do remember seeing that.

YannickJadoul:
> With all the overloaded operators, it would actually feel like a C++ type.

That's true, yes! Of course, that's also up to @in42. I'm not saying this needs to per se go into this PR in order for this PR to be useful. So up to @in42 to judge what could still go in?

> `operator[](ssize_t)`?

Hmm, there's no dedicated C API for this, is there? So this might already be supported (though slow, since there's no C API) through inheritance from `object_api<T>`?

bstaletic:
> Hmm, there's no dedicated C API for this, is there? So this might already be supported (though slow, since there's no C API) through inheritance from object_api<T>?

I'd actually be surprised that the generic `PyObject_GetAttrString()` would work here. Also, `operator[](ssize_t)` could be accompanied with `at(ssize_t)`, matching STL containers.

> That's true, yes! Of course, that's also up to @in42. I'm not saying this needs to per se go into this PR in order for this PR to be useful. So up to @in42 to judge what could still go in?

Fair enough.

in42:
> Nice! Thanks for this :-)
> 
> Possibly useful functions (mostly stolen from https://docs.python.org/3/c-api/bytearray.html):
> 
> * `char *get()` and/or `operator char*()`? I think this wouldn't copy, like string does, and would allow to change the contents? Maybe that's a reason to not want this, though.
> * `size_t size()`; this seems potentially useful.
> * `bytesarray concat(bytesarray)`: `PyByteArray_Concat` exists. Not sure how useful it would be, though? Not crucial to have, afaic.
> * `resize`: `PyByteArray_Resize` exists.
> * Something to assign and change the contents of the `bytearray`? I thought that was the whole point of a `bytearray`, being a mutable `bytes`? @bstaletic, any thoughts?

I will add the additional methods - but I remember seeing for `pybind11::bytes`, these methods were not defined. Is there any reason why they were not implemented for `pybind11::bytes`?

bstaletic:
A few possible reasons:

- Nobody requested it before. Simple omission.
- It was postponed for "later".
- `bytes` being immutable, it's worth taking a second look at whether `CharT& operator[](size_t);` is a good idea.

YannickJadoul:
@in42, if you don't see a use for it, you don't have to ofc, as I suggested. But as far as I can tell, it could be useful. Especially so for `bytearray`, as it's a mutable structure.

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,35 @@ inline str::str(const bytes& b) {
...
+
+    bytearray(const char *c, size_t n)
+        : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
+    // Allow implicit conversion:
+    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+
+    operator std::string() const {
```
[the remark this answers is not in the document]
in42: Done.

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,35 @@ inline str::str(const bytes& b) {
     m_ptr = obj.release().ptr();
 }
 
+/// \addtogroup pytypes
+/// @{
+class bytearray : public object {
+public:
+    PYBIND11_OBJECT(bytearray, object, PyByteArray_Check)
+
+    bytearray()
```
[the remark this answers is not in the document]
in42: Done.

include/pybind11/pytypes.h
in42: Done.

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,35 @@ inline str::str(const bytes& b) {
...
+    bytearray(const char *c, size_t n)
+        : object(PyByteArray_FromStringAndSize(c, (ssize_t) n), stolen_t{}) {
+        if (!m_ptr) pybind11_fail("Could not allocate bytearray object!");
+    }
+
+    bytearray()
+        : bytearray("", 0) {}
+
+    // Allow implicit conversion:
+    bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
```
YannickJadoul, line 1077: Do we want implicit conversion here? I know `bytes` does it, but I'm not a huge fan of unnecessarily implicit conversions.

YannickJadoul (approved):
One more implicit conversion remark. I'm reasonably confident @bstaletic will feel the same about this?

include/pybind11/pytypes.h
in42: I will make it explicit.

YannickJadoul: Thanks!

include/pybind11/pytypes.h
```diff
@@ -1059,6 +1059,34 @@ inline str::str(const bytes& b) {
...
+
+    bytearray()
+        : bytearray("", 0) {}
+
+    explicit bytearray(const std::string &s) : bytearray(s.data(), s.size()) { }
+
+    size_t size() const { return static_cast<size_t>(PyByteArray_Size(m_ptr)); }
+
+    explicit operator std::string() const {
+        char *buffer = PyByteArray_AS_STRING(m_ptr);
```
bstaletic, line 1081: One last time. Are we sure we don't want to check for NULL and throw?

bstaletic (approved):
Looks good.

include/pybind11/pytypes.h
in42: ok I will add that otherwise the code may become risky.

in42: As I can see [here](https://github.com/python/cpython/blob/a6192635f1e62af2bb8a435487ebb51800edd671/Objects/bytearrayobject.c#L159), `PyByteArray_AsString` does not check for `NULL` at all, except for some `assert`s which you said maybe disabled at runtime. So, we have check for `nullptr` on our side.
If I find `m_ptr` is `NULL`, since I can't throw `error_already_set`, which exception should I throw?

bstaletic: Right... It just does this:

https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Include/cpython/bytearrayobject.h#L15-L17

It still might dereference `self`. I believe that in these cases pybind11 uses `pybind11_fail(what_str)`.

in42: Oh right

in42: Done.

bstaletic:
Final thought: commit history is a mess here. There are at least three empty commits and none of the 14 commits suggests that it has anything to do with `py::bytearray`.

include/pybind11/pytypes.h
YannickJadoul: Actually, I was about to argue that this was fine. You shouldn't be using a `NULL` `py::object` (and you can check for it with `is_none()`), and the constructors don't allow to construct one, so the only way of getting a `NULL` `bytearray` is through `reinterpret_borrow`/`reinterpret_steal`, which is an "advanced" operation anyway?

in42:
> Final thought: commit history is a mess here. There are at least three empty commits and none of the 14 commits suggests that it has anything to do with `py::bytearray`.

I am little unfamiliar with how changes are merged using a pull request. Aren't you supposed to squash the commits and merge them in a single commit?

I am more familiar with the gerrit model.

YannickJadoul:
@bstaletic, the pybind11 GitHub project will squash anyway, and use the title of this PR, I believe. So we should be fine?

include/pybind11/pytypes.h
bstaletic: Are you arguing that we should revert?

bstaletic:
Alright, if @YannickJadoul says not to worry about history, fine by me.

include/pybind11/pytypes.h
YannickJadoul: I was wondering if this was necessary, yes. But unless you strongly agree, I'm not arguing to revert, since this thing has been going on for long enough.

in42: Isn't it better to keep the check anyways for future if someday the invariant that it will never be NULL is not true anymore?

YannickJadoul: Myeah, maybe, but I think there's lots of other places where non-`NULL` is already assumed.

And then `PyByteArray_Size()` isn't much safer either?

https://github.com/python/cpython/blob/3cf0833f42ebde24f6435b838785ca4f946b988f/Objects/bytearrayobject.c#L149-L156

YannickJadoul: But yes, it was just a suggestion. If you or @bstaletic aren't convinced, let's just go for this.

bstaletic: > And then PyByteArray_Size() isn't much safer either?

Oh... in that case either we should check for NULL there as well, or not check for NULL in either place. @YannickJadoul is right when he says

> the only way of getting a NULL bytearray is through reinterpret_borrow/reinterpret_steal, which is an "advanced" operation anyway?

So I'm not against dropping the NULL check, but I would like to see consistent behaviour with respect to `m_ptr == NULL`.

in42: Ok, to keep it consistent with the other places, I will remove the NULL check.

in42: Should I go ahead and remove?

bstaletic: Go ahead.

YannickJadoul: Yeah, to be fair, pybind11 is already not consistent I think. Sometimes safe methods are used, but ... yeah, I don't know :-) Nothing to be fixed in this PR, anyway.

YannickJadoul: If users come with a good complaint about why we should check for `NULL` instead of them, we can still add a check.

in42: Done.

YannickJadoul:
It's been long enough, and both @bstaletic and I approved. Thanks for holding on and making the PR, @in42! :-)

crimsoncor:
great feature. I'm actually making use of it today. But it isn't documented anywhere. I was able to go through the PR and the source to figure out how to use it, but should probably be added to the documentation

gatopeich:
Late to the party here, but it would be great to have an implicit conversion to string_view like bytes and str do.

Skylion007:
@gatopeich PR welcome on this front.

Issue #2786, opened 2021-01-11
[BUG] Missing error checking for overflowing integer casts


hawkinsp:
## Issue description

There appears to be some missing error checking around Python int to C++ long casts that may overflow. In the example below, since the Python -> C++ cast fails, I expect a C++ exception to be thrown. However, it appears the Python interpreter's error state ends up set but pybind11 fails to notice.

I suspect that this method needs to check the interpreter's error state (since `PyLong_AsLong` may fail if the integer is larger than a C `long`)
https://github.com/pybind/pybind11/blob/98f1bbb8004f654ba9e26717bdf5912fb899b05a/include/pybind11/pytypes.h#L1143

## Reproducible example code

```
#include <pybind11/pybind11.h>
#include <iostream>

int docast(pybind11::object a) {
  int b = pybind11::cast<pybind11::int_>(a);
  std::cerr << "Should not reach here" << std::endl;
  return b;
}

PYBIND11_MODULE(t, m) {
    m.def("docast", &docast, "Perform a cast");
}
```

```
In [1]: import t
In [2]: t.docast(3123412423423423234234234)
Should not reach here
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
OverflowError: Python int too large to convert to C long

The above exception was the direct cause of the following exception:

SystemError                               Traceback (most recent call last)
<ipython-input-3-688cef4b7a48> in <module>
----> 1 t.docast(3123412423423423234234234)

SystemError: <built-in method docast of PyCapsule object at 0x102917720> returned a result with an error set
```

YannickJadoul:
Huh, that's funny. What version of pybind11 are you on?

This seems very similar to #2336, but not really.

hawkinsp:
I just verified this with 2.6.1, which appears to be current on Pypi.

YannickJadoul:
Urgh. Let me check, then. Thanks for reporting, already!

YannickJadoul:
The error seems to be in the implicit conversion from `py::int_` to `int`:

```
    template <typename T,
              detail::enable_if_t<std::is_integral<T>::value, int> = 0>
    operator T() const {
        return std::is_unsigned<T>::value
            ? detail::as_unsigned<T>(m_ptr)
            : sizeof(T) <= sizeof(long)
              ? (T) PyLong_AsLong(m_ptr)
              : (T) PYBIND11_LONG_AS_LONGLONG(m_ptr);
    }
```

[...]

PR #2801, opened 2021-01-16
Always call PyNumber_Index when casting from Python to a C++ integral type, also pre-3.8

YannickJadoul:
## Description

See the remaining decision to be made in #2698: https://github.com/pybind/pybind11/pull/2698#issuecomment-752218748

In one sentence, this makes the casting from Python objects to C++ integer types consistent across Python versions, "backporting" Python 3.8 behavior on handling `__index__` to pre-3.8 Pythons.

As @bstaletic pointed out to me, we have done this before, most often backporting Python 3 behavior to Python 2 (unicode, for example), but also in #2616, "backporting" 3.8 behavior.

Given the original issue fixed by #2698, and especially the discussion following that ("everything with `__index__` is an integer type"), I think this is the logical to minimize surprises when developing pybind11 libraries. See also the changed test demonstrating how things are more consistent across versions.

I'm cheekily adding this to 2.6.2, because in my eyes, it's still part of #2698, and I'd like to see a decision made ("no, because ..." is also fine, btw; then we close this PR! But it doesn't make sense to me to delay this simple decision on policy.). This PR mainly makes it easier to judge the changes (after soon merging #2698, at least).

## Suggested changelog entry:

```rst
When casting to a C++ integer, `__index__` is always called and not considered as conversion, consistent with Python 3.8+.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1043,30 +1043,25 @@ public:
         } else if (!convert && !index_check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
         } else {
-            handle obj = src;
+            handle src_or_index = src;
 #if PY_VERSION_HEX < 0x03080000
-            bool do_decref = false;
+            object index;
             if (index_check(src.ptr())) {
-                PyObject *tmp = PyNumber_Index(src.ptr());
-                if (!tmp) {
+                index = reinterpret_steal<object>(PyNumber_Index(src.ptr()));
+                if (!index) {
                     PyErr_Clear();
                     return false;
                 }
-                do_decref = true;
-                obj = tmp;
+                src_or_index = index;
             }
 #endif
             if (std::is_unsigned<py_type>::value) {
-                py_value = as_unsigned<py_type>(obj.ptr());
+                py_value = as_unsigned<py_type>(src_or_index.ptr());
             } else { // signed integer:
                 py_value = sizeof(T) <= sizeof(long)
-                    ? (py_type) PyLong_AsLong(obj.ptr())
-                    : (py_type) PYBIND11_LONG_AS_LONGLONG(obj.ptr());
+                    ? (py_type) PyLong_AsLong(src_or_index.ptr())
+                    : (py_type) PYBIND11_LONG_AS_LONGLONG(src_or_index.ptr());
             }
-#if PY_VERSION_HEX < 0x03080000
-            if (do_decref)
-                obj.dec_ref();
-#endif
         }
 
         // Python API reported an error
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1042,12 +1042,30 @@ public:
             return false;
         } else if (!convert && !index_check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
-        } else if (std::is_unsigned<py_type>::value) {
-            py_value = as_unsigned<py_type>(src.ptr());
-        } else { // signed integer:
-            py_value = sizeof(T) <= sizeof(long)
-                ? (py_type) PyLong_AsLong(src.ptr())
-                : (py_type) PYBIND11_LONG_AS_LONGLONG(src.ptr());
+        } else {
+            handle obj = src;
+#if PY_VERSION_HEX < 0x03080000
+            bool do_decref = false;
+            if (PyIndex_Check(src.ptr())) {
+                PyObject *tmp = PyNumber_Index(src.ptr());
+                if (!tmp) {
+                    py_value = (py_type) -1;
+                }
+                do_decref = true;
+                obj = tmp;
+            }
+#endif
+            if (std::is_unsigned<py_type>::value) {
+                py_value = as_unsigned<py_type>(obj.ptr());
+            } else { // signed integer:
+                py_value = sizeof(T) <= sizeof(long)
+                    ? (py_type) PyLong_AsLong(obj.ptr())
+                    : (py_type) PYBIND11_LONG_AS_LONGLONG(obj.ptr());
+            }
+#if PY_VERSION_HEX < 0x03080000
+            if (do_decref)
+                obj.dec_ref();
+#endif
         }
 
         // Python API reported an error
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1044,17 +1044,12 @@ public:
             return false;
         } else {
             handle src_or_index = src;
+            py_value = (py_type) -1;
 #if PY_VERSION_HEX < 0x03080000
             object index;
-            if (index_check(src.ptr())) {
+            if (!PYBIND11_LONG_CHECK(src.ptr()) && index_check(src.ptr())) {
                 index = reinterpret_steal<object>(PyNumber_Index(src.ptr()));
-                if (!index) {
-                    src_or_index = handle();
-                    py_value = (py_type) -1;
-                }
-                else {
-                    src_or_index = index;
-                }
+                src_or_index = index ? index : handle();
             }
 #endif
             if (src_or_index) {
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1046,10 +1046,11 @@ public:
             handle obj = src;
 #if PY_VERSION_HEX < 0x03080000
             bool do_decref = false;
-            if (PyIndex_Check(src.ptr())) {
+            if (index_check(src.ptr())) {
                 PyObject *tmp = PyNumber_Index(src.ptr());
                 if (!tmp) {
-                    py_value = (py_type) -1;
+                    PyErr_Clear();
+                    return false;
                 }
                 do_decref = true;
                 obj = tmp;
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1049,18 +1049,22 @@ public:
             if (index_check(src.ptr())) {
                 index = reinterpret_steal<object>(PyNumber_Index(src.ptr()));
                 if (!index) {
-                    PyErr_Clear();
-                    return false;
+                    src_or_index = handle();
+                    py_value = (py_type) -1;
                 }
-                src_or_index = index;
+                else {
+                    src_or_index = index;
+                }
             }
 #endif
-            if (std::is_unsigned<py_type>::value) {
-                py_value = as_unsigned<py_type>(src_or_index.ptr());
-            } else { // signed integer:
-                py_value = sizeof(T) <= sizeof(long)
-                    ? (py_type) PyLong_AsLong(src_or_index.ptr())
-                    : (py_type) PYBIND11_LONG_AS_LONGLONG(src_or_index.ptr());
+            if (src_or_index) {
+                if (std::is_unsigned<py_type>::value) {
+                    py_value = as_unsigned<py_type>(src_or_index.ptr());
+                } else { // signed integer:
+                    py_value = sizeof(T) <= sizeof(long)
+                        ? (py_type) PyLong_AsLong(src_or_index.ptr())
+                        : (py_type) PYBIND11_LONG_AS_LONGLONG(src_or_index.ptr());
+                }
             }
         }
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1040,26 +1040,30 @@ public:
                 return false;
         } else if (PyFloat_Check(src.ptr())) {
             return false;
-        } else if (!convert && !index_check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
+        } else if (!convert && !PYBIND11_LONG_CHECK(src.ptr()) && !index_check(src.ptr())) {
             return false;
         } else {
             handle src_or_index = src;
-            py_value = (py_type) -1;
 #if PY_VERSION_HEX < 0x03080000
             object index;
-            if (!PYBIND11_LONG_CHECK(src.ptr()) && index_check(src.ptr())) {
+            if (!PYBIND11_LONG_CHECK(src.ptr())) {  // So: index_check(src.ptr())
                 index = reinterpret_steal<object>(PyNumber_Index(src.ptr()));
-                src_or_index = index ? index : handle();
+                if (!index) {
+                    PyErr_Clear();
+                    if (!convert)
+                        return false;
+                }
+                else {
+                    src_or_index = index;
+                }
             }
 #endif
-            if (src_or_index) {
-                if (std::is_unsigned<py_type>::value) {
-                    py_value = as_unsigned<py_type>(src_or_index.ptr());
-                } else { // signed integer:
-                    py_value = sizeof(T) <= sizeof(long)
-                        ? (py_type) PyLong_AsLong(src_or_index.ptr())
-                        : (py_type) PYBIND11_LONG_AS_LONGLONG(src_or_index.ptr());
-                }
+            if (std::is_unsigned<py_type>::value) {
+                py_value = as_unsigned<py_type>(src_or_index.ptr());
+            } else { // signed integer:
+                py_value = sizeof(T) <= sizeof(long)
+                    ? (py_type) PyLong_AsLong(src_or_index.ptr())
+                    : (py_type) PYBIND11_LONG_AS_LONGLONG(src_or_index.ptr());
             }
         }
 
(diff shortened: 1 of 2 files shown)


YannickJadoul:
Aaaaand ICC segfaults. Why!?
It's been slightly over a day since we merged that PR...

Valgrind seems perfectly happy, btw.

include/pybind11/cast.h
```diff
@@ -1020,19 +1020,48 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
+                if (!tmp) {
+                    PyErr_Clear();
+                    return false;
+                }
+                do_decref = true;
+                obj = tmp;
+            }
+#endif
+            if (std::is_unsigned<py_type>::value) {
+                py_value = as_unsigned<py_type>(obj.ptr());
```
YannickJadoul, line 1055: This might seems like a reasonably big change, but after this PR, I want to fix #2786, which involves a minor refactoring of casting to C++ integer types (to ensure future consistency with `py::int_::operator int()`), so keep that in mind when reviewing, please ;-)

If we think that consistency between Python < 3.8 and >= 3.8 versions is a nice thing to have, then I personally don't really think this is a too high implementation price to pay.

YannickJadoul: (Also, #2786's fix shouldn't be complex either, so if you're able to wait 1 or 2 more days, it can also still be a fix to go into 2.6.2. But we need to draw a line somewhere, ofc.)

henryiii:
There was an issue in setuptools 51.3.0 fixed in 51.3.1, I've restarted the build. https://github.com/pypa/setuptools/issues/2535

YannickJadoul:
> There was an issue in setuptools 51.3.0 fixed in 51.3.1, I've restarted the build. [pypa/setuptools#2535](https://github.com/pypa/setuptools/issues/2535)

Thanks :-) I'd noticed setuptools had release suspiciously recently, but I didn't manage to figure out what was going in anymore, at 2 am.

henryiii:
I'm mildly in favor. We don't want to rush a release out; after I get the PRs that were blocked by the GitHub API issue & CMake, I'd like @rwgk to run a global test before we pull the trigger on a release. So there's a bit more time.

YannickJadoul:
In principle, this is ready. Surely things can still be cleaned up, but let's first make a decision on what we want?

#2802 (fixing #2786) is a bit more of a mess, though. I had hoped I could still get it in, together with this, but ... well, I don't know, right now.

rwgk:
@henryiii I ran this PR through the global testing already, but I'm happy to repeat with the latest of everything. Could you please ping me when you think it's a good time to retest? (now-ish?)

I think this PR as-is is a good improvement by itself, and a great baseline for potential follow-up work as @yannick pointed out.

rwgk:
OOPS, this is #2801, I mistakingly thought it's #2698. I just deleted my previous comment. Sorry!

rwgk:
> I'd like @rwgk to run a global test before we pull the trigger on a release

Will do. (I'm pretty neutral on this PR yes/no for v2.6.2.)

include/pybind11/cast.h
```diff
@@ -1042,12 +1042,31 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
         } else if (!convert && !index_check(src.ptr()) && !PYBIND11_LONG_CHECK(src.ptr())) {
             return false;
-        } else if (std::is_unsigned<py_type>::value) {
-            py_value = as_unsigned<py_type>(src.ptr());
-        } else { // signed integer:
-            py_value = sizeof(T) <= sizeof(long)
-                ? (py_type) PyLong_AsLong(src.ptr())
-                : (py_type) PYBIND11_LONG_AS_LONGLONG(src.ptr());
+        } else {
+            handle obj = src;
```
rwgk, line 1046: `handle src_or_index = src;`
to be more descriptive.

include/pybind11/cast.h
```diff
@@ -1042,12 +1042,31 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
+#if PY_VERSION_HEX < 0x03080000
+            bool do_decref = false;
```
rwgk, line 1048: `object index_owner;`

include/pybind11/cast.h
```diff
@@ -1042,12 +1042,31 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
+            if (index_check(src.ptr())) {
+                PyObject *tmp = PyNumber_Index(src.ptr());
+                if (!tmp) {
+                    PyErr_Clear();
+                    return false;
+                }
+                do_decref = true;
```
rwgk, line 1055: `index_owner = reinterpret_steal<object>(tmp);`

That way you don't need the second `#if PY_VERSION_HEX < 0x03080000` below and this code become exception safe.

I'd also use `idx` (or similar) instead of `tmp`, to be more descriptive.

YannickJadoul: I thought of/tried that, but didn't want to incur an overhead refcounting on Python >= 3.8, and this is also what CPython does. 

But wait, maybe you mean something else, that doesn't need this! I'll give this a shot :-)

YannickJadoul: Yesss, that does work out beautifully! Thanks!

include/pybind11/cast.h
YannickJadoul: Done.

include/pybind11/cast.h
YannickJadoul: Still seeing if this could easily be refactored out.

YannickJadoul: OK, it's quite hard to refactor into a separate private function without incurring an additional `inc_ref`/`dec_ref`, it seems. It's already cleaner than before, though, so is it fine to leave like this for now?

rwgk:
Thanks @YannickJadoul, that looks great! I'll run this through our big testing system asap (probably tonight).

YannickJadoul:
> Thanks @YannickJadoul, that looks great! I'll run this through our big testing system asap (probably tonight).

Good, thanks! I'm pretty confident this PR does as it says (given our own tests and how they caught some corner cases), but it's good to have an idea how this "backported behavior" interacts in a larger context.

bstaletic (approved):
This looks fine to me.

rwgk:
I'm seeing 2 test failures that look related to this PR.
We're still on Python 3.6, so the `#if PY_VERSION_HEX < 0x03080000` branch kicks in.
Debugging.

rwgk:
In both failing tests a NumPy array with one `float` was passed for an `int` arg. I mailed fixes, boiling down to `int(arr[0])`, with comment:

> A NumPy array was passed instead of an integer. The implicit conversion is disabled by pybind11 PR #2801, which backports a behavior change introduced by Python 3.8, for consistency.

@YannickJadoul, is that description accurate?

henryiii:
I think the current (in this PR) behavior is the correct one; a NumPy float should not be automatically converted to an int. If the they want to support floats, there should be a float function/method/constructor that does the conversion. NumPy no longer allows floats in indexing, as well.

YannickJadoul:
Huh, this is weird though. I thought this PR would only make things more permissive! Let me try out a few things.

rwgk:
I agree with @henryiii and I'm OK if you want to merge for 2.6.2, although strictly speaking it's bending the rules to introduce this behavior change in a minor release, now that we know there are things that will break.

henryiii:
Are you sure that current master without this PR doesn't also trigger that? Assuming it's the other PR that caused this?

henryiii:
It's breaking something that shouldn't have worked, though, so it's a bit of a grey area.

rwgk:
Yes, sure. Before attempting a fix, I commented out the `return false;`
after `PyNumberIndex`, and the `src_or_index = index;` assignment, which
made the test work.

YannickJadoul:
```
$ python3.6
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
>>> import numpy as np
>>> x = np.array(42)
>>> int(x)
42
>>> list(range(100))[x]
42
>>> y = np.array(3.14159)
>>> int(y)
3
>>> list(range(100))[y]
TypeError: only integer scalar arrays can be converted to a scalar index
```

Yeah, so `int(some_float_array)` works (also in 3.8), so I would think that we still expect to convert this on `noconvert(false)`, right?

YannickJadoul:
@rwgk: what happens on 3.8 before and after this PR?

rwgk:
I cannot easily test that, because internally we're still on 3.6. I'd have
to ask around if and how I could test with 3.8.
Or I'd have to make up a minimal reproducer to add to the standard pybind11
unit tests.

YannickJadoul:
Judging from CPython's code, I would expect this to already fail on Python 3.8+?

https://github.com/python/cpython/blob/b745a6143ae79efe00aa46affe5ea31a06b0b532/Objects/longobject.c#L386-L394

>  cannot easily test that, because internally we're still on 3.6. I'd have to ask around if and how I could test with 3.8. Or I'd have to make up a minimal reproducer to add to the standard pybind11 unit tests.

OK, let me quickly try something, then!

YannickJadoul:
> OK, let me quickly try something, then!

```c++

namespace py = pybind11;

PYBIND11_MODULE(example, m)
{
    m.def("foo", [](int bar) { return bar; });
}
```

This works with Python 3.9; but I don't get what this PR is doing differently from CPython's `PyLong_AsLong` :-( Time to start some debugging.

YannickJadoul:
Think I've got it. This test should reproduce your error, @rwgk. `TypeError` is handled differently from `ValueError`. So that's why ...

rwgk:
Thanks @YannickJadoul, I'll run this PR through our big old testing mill again, later tonight.

FYI: In the meantime one team already merged my fix for them. (I'm still waiting for feedback on the second fix.)

YannickJadoul:
> Thanks @YannickJadoul, I'll run this PR through our big old testing mill again, later tonight.
> 
> FYI: In the meantime one team already merged my fix for them. (I'm still waiting for feedback on the second fix.)

Oh, OK. I don't think they're even to blame, actually, since things did just work on 3.8. Thanks for finding this!

I still need to figure out how to get 2.7 to at least pass tests. (Maybe not worth putting too much effort into 2.7.)

rwgk:
The failures were actually useful. I don't think the code was working as intended by the original authors, only scraping by.
But I agree with your approach to revising this PR, to fully match Python 3.8 behavior.
Python 2.7 is on it's last leg as far as I see, not worth a significant effort.

YannickJadoul:
Ah, good to know it was at least worth the effort, then!
For the record, I believe that some of these kinds of conversions start producing deprecation warnings from 3.8+ onwards (I had to silence them). Python just has a really weird approach to deprecation warnings and tends to hide them (though it's already better in more recent versions, I believe).

YannickJadoul:
OK, got it. Python 2's `long` vs. `int` :-D

YannickJadoul:
I'll get back to this/fix this tomorrow, btw.
I'm still confused, since it seems that both `PyLong_AsLong` as well as `PyNumber_Long` call `__int__`? The docs are slightly confusing, so I'll check out CPython's source, but I need some sleep and a fresh perspective on this.

tests/test_builtin_casters.py
```diff
@@ -286,14 +300,20 @@ def cant_convert(v):
     assert noconvert(7) == 7
     cant_convert(3.14159)
     assert convert(DeepThought()) == 42
-    require_implicit(DeepThought())
+    requires_conversion(DeepThought())
+    assert convert(DoubleThought()) == 0  # Fishy; `int(DoubleThought)` == 42
```
YannickJadoul, line 304: This is not great, but kind of a consequence of us saying "everything with `__index__` is already an `int`, so don't try converting".

YannickJadoul (commented):
OK, this ~fixes~ patches things, adding some more duct tape here and there.

Seems like refactoring/restructuring the `float`/`int` type caster is overdue, as well, but I propose to not cram that into this PR anymore.

Also, yes, this could use another test round, @rwgk. Who knows what I missed, this time.
Next to that test round, it's also worth checking: is this series of tests the behavior we want?

include/pybind11/cast.h
```diff
@@ -1040,14 +1040,31 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
...
+                else {
+                    src_or_index = index;
+                }
+            }
+#endif
+            if (std::is_unsigned<py_type>::value) {
+                py_value = as_unsigned<py_type>(src_or_index.ptr());
+            } else { // signed integer:
+                py_value = sizeof(T) <= sizeof(long)
+                    ? (py_type) PyLong_AsLong(src_or_index.ptr())
```
YannickJadoul, line 1065: Oh, another note: this results in warnings, and I think that's correct. Because it's not just getting out the `long` from the `PyLong` object, but it's **also** trying to do the conversion.

Yes, the C API is quite messy here. And it's only made worse by the structure of this caster.

rwgk:
FYI: I tried running this through our global testing last night but something went wrong, the tests only ran with a previous version of this PR. I'll try again.

rwgk:
To keep track of an observation, below is the error I saw in the first global testing run. With the current version of this PR the same test passes.

For completeness: the dreamplace code used here is ~1 year behind github.
The non-matching args were: `tensor(3.), tensor(3.)`

```
  File "dreamplace/ops/electric_potential/electric_overflow.py", line 240, in forward
    self.num_threads
TypeError: fixed_density_map(): incompatible function arguments. The following argument types are supported:
    1. (arg0: at::Tensor, arg1: at::Tensor, arg2: at::Tensor, arg3: at::Tensor, arg4: at::Tensor, arg5: float, arg6: float, arg7: float, arg8: float, arg9: float, arg10: float, arg11: int, arg12: int, arg13: int, arg14: int, arg15: int, arg16: int, arg17: int) -> at::Tensor

Invoked with: tensor([ 124.0602,    0.0000,  499.0000,    0.0000,  315.0000,   85.0000,
         124.0602,  100.0000,  499.0000,  101.0000,   65.0000,  105.0000]), tensor([   1.8796,    0.0000,    0.0000,    0.0000,  120.0000,   80.0000]), tensor([   1.8796,    0.0000,    0.0000,    0.0000,  120.0000,   40.0000]), tensor([  31.2500,   93.7500,  156.2500,  218.7500,  281.2500,  343.7500,
         406.2500,  468.7500]), tensor([  31.2500,   93.7500,  156.2500,  218.7500,  281.2500,  343.7500,
         406.2500,  468.7500]), 0.0, 0.0, 500.0, 500.0, tensor(62.5000), tensor(62.5000), 1, 5, 8, 8, tensor(3.), tensor(3.), 8
```

YannickJadoul:
> To keep track of an observation, below is the error I saw in the first global testing run. With the current version of this PR the same test passes.

This should now also be covered by our own tests. I added the `TypeErrorThought` (yes, this naming joke got out of hand quite quickly) in caa53820704eb640d41fdc866b936896b79f1c91, which demonstrated the observed failure. And of course, the current version of the passes this new test.

rwgk (approved):
This PR passed the Google global testing now.
Thanks @YannickJadoul!

tests/test_builtin_casters.py
```diff
@@ -256,6 +256,13 @@ class DeepThought(object):
         def __int__(self):
             return 42
 
+    class DoubleThought(object):
```
henryiii, line 259: Double meaning it has index and int?

henryiii (approved):

tests/test_builtin_casters.py
YannickJadoul: Yes, that's what I meant :-) As admitted yesterday:

> yes, this naming joke got out of hand quite quickly

Should I still pick better names?

henryiii: Yes. IntIndexThought, at least. :)

YannickJadoul: Done so. This should make things more clear? More boring as well, but ...

7fd2db5a1105aa073edc46f23e20b063ed7ec04f has _only_ changes to our own tests, so if these pass, look good/better to you, and you're happy with the way `DoubleThought`/`IntAndIndex` is handled, then we can merge this, @henryiii

YannickJadoul: > Yes. IntIndexThought, at least. :)

I liked the mix of HHGG and 1984, but yes ;-)

henryiii:
LGTM!

YannickJadoul:
Thanks, all! Another tiny bit of progress :-)

PR #2804, opened 2021-01-18
fix: support basic dual includes

### LINKED ISSUE #2800: [BUG] Possible issue loading pybind11 as submodule twice
This came up on the CMake issue tracker; pybind11 might not support being loaded twice. For example, if you have two submodules, and they both load pybind11, then they interfere (and, glancing at the code, I think this is will not work currently because we create the first target or two without checking for existence). To properly test, we ideally should put loads into two directories, since directory-wide variables need to be reloaded (targets should be global and should not be recreated).

Reported by @dmikushin in https://gitlab.kitware.com/cmake/cmake/-/issues/16414#note_886491

I _think_ we support it in Config mode, but not in submodule mode.

henryiii:
## Description

This allows pybind11 to be included twice in CMake. It is just a basic check; it is possible some variables will need to be set again that are being skipped, but I think it mostly should work.

(Some variables are directory scoped, and IMPORT targets are as well).

Fixes #2800.

## Suggested changelog entry:

```rst
* Avoid error if included from two submodule directories.
```

### DIFF
--- a/tests/test_cmake_build/CMakeLists.txt
+++ b/tests/test_cmake_build/CMakeLists.txt
@@ -25,7 +25,7 @@ function(pybind11_add_build_test name)
   endif()
 
   if(NOT ARG_INSTALL)
-    list(APPEND build_options "-DPYBIND11_PROJECT_DIR=${pybind11_SOURCE_DIR}")
+    list(APPEND build_options "-Dpybind11_SOURCE_DIR=${pybind11_SOURCE_DIR}")
   else()
     list(APPEND build_options "-DCMAKE_PREFIX_PATH=${pybind11_BINARY_DIR}/mock_install")
   endif()
@@ -77,3 +77,6 @@ if(PYBIND11_INSTALL)
 endif()
 
 add_dependencies(check test_cmake_build)
+
+add_subdirectory(subdirectory_target EXCLUDE_FROM_ALL)
+add_subdirectory(subdirectory_embed EXCLUDE_FROM_ALL)
--- a/tests/test_cmake_build/subdirectory_target/CMakeLists.txt
+++ b/tests/test_cmake_build/subdirectory_target/CMakeLists.txt
@@ -11,7 +11,7 @@ endif()
 
 project(test_subdirectory_target CXX)
 
-add_subdirectory(${PYBIND11_PROJECT_DIR} pybind11)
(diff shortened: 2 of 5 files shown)


YannickJadoul:
Wait, that's all? I was expecting this to be a horrible PR!

YannickJadoul:
("horrible" in the amount of code/hacks necessary to get this done, I mean; not horrible in it's implementation by you, ofc, @henryiii ;-) )

henryiii:
It's not horrible if one writes good CMake to begin with. ;)

henryiii:
And, this is just to make it possible, it's not really testing to see if all the needed directory-level variables get defined per-directory. But that should hopefully only be minor tweaks on top. Hopefully @dmikushin can test. :)

tests/test_cmake_build/subdirectory_target/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 
 project(test_subdirectory_target CXX)
 
-add_subdirectory(${PYBIND11_PROJECT_DIR} pybind11)
+add_subdirectory("${pybind11_SOURCE_DIR}" pybind11)
```
bstaletic, line 14: How are these two variables different?

henryiii: This is now the same name as the original variable, rather than being renamed, so it works when added from the main file as well.

tests/test_cmake_build/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ function(pybind11_add_build_test name)
   endif()
 
   if(NOT ARG_INSTALL)
-    list(APPEND build_options "-DPYBIND11_PROJECT_DIR=${pybind11_SOURCE_DIR}")
+    list(APPEND build_options "-Dpybind11_SOURCE_DIR=${pybind11_SOURCE_DIR}")
```
henryiii, line 28: @bstaletic this is the important bit.

bstaletic: Ah, that makes sense!

bstaletic (approved):
Obligatory moaning about cmake: for once, reading cmake source didn't hurt... too much.

PR #2805, opened 2021-01-18
ci: update to setup-cmake v1.6

henryiii:
## Description

This has stopped finding older CMakes (3.18 is fine, but 3.4, 3.7, and 3.8 are printing out "Unable to find version matching"), first trying a version bump before bugging @jwlawson. :)

## Suggested changelog entry:

None needed, CI.


henryiii:
Will update after this gets worked out in https://github.com/jwlawson/actions-setup-cmake/pull/23 - the problem is the GitHub API suddenly dropping `Link:` from the releases header, in contrast to what the docs say about pagination. 😠

henryiii:
Github fixed their issue.

PR #2806, opened 2021-01-18
style: avoid using unintialized variables

henryiii:
Tested with cmake --warn-unintialized -S . -B build

## Description

Avoiding use of uninitialized variables.

## Suggested changelog entry:

```rst
* Support running with --warn-uninitialized active in CMake
```


tests/test_cmake_build/CMakeLists.txt
```diff
@@ -55,6 +55,14 @@ function(pybind11_add_build_test name)
   add_dependencies(test_cmake_build test_build_${name})
 endfunction()
 
+# Avoid uninitialized variables
+if(NOT DEFINED PYTHON_MODULE_EXTENSION)
+  set(PYTHON_MODULE_EXTENSION "")
+endif()
```
YannickJadoul, line 61: Worth turning this into a function, and just write single lines like `set_uninitialized(PYTHON_MODULE_EXTENSION "")`?

YannickJadoul: (because 1. CMake if-statments are quite some clutter, and 2. it's slightly clearer in intention and wouldn't need the comment anymore, but I can see there's a tradeoff with maintenance of that function)

YannickJadoul (approved):
Merge as you see fit, I'd say.

If anything, it would even be nice to group stuff at the start and just say "all these settings need to be initialized to `""`", but ... eh.

tests/test_cmake_build/CMakeLists.txt
henryiii: Ehh, we could add it, as long the places we need it are not standalone.  How about `possibly_uninitialized` or `touch`?

YannickJadoul: Those two could work as well, yes, if you always want to set them to `""`?

Also, just a suggestion. Feel free to ignore it if you're not convinced; I'm not that advanced in CMake to foresee the consequences. I just know enough to know it's possible ;-)

henryiii: Okay, added `possibly_uninitialized(PYTHON_MODULE_EXTENSION Python_INTERPRETER_ID)` here.

You'll always want to define them to "" since if they are uninitialized, that's what they act like. :)

tools/FindEigen3.cmake
```diff
@@ -64,6 +64,9 @@ if(EIGEN3_INCLUDE_DIR)
   set(EIGEN3_FOUND ${EIGEN3_VERSION_OK})
 
 else(EIGEN3_INCLUDE_DIR)
+  if(NOT DEFINED KDE4_INCLUDE_DIR)
```
YannickJadoul, line 67: Ah, right, we can't use `possibly_uninitialized` here? Ah, too bad. So the solution is already less useful :-)

tests/test_cmake_build/CMakeLists.txt
YannickJadoul: Great, thanks for the explanation.

YannickJadoul:
Thanks! Merge as you see fit, I'd say.

PR #2807, opened 2021-01-19
Move `cpp_function` to its own header file for further refactoring

lqf96:
## Description
Move `cpp_function` to its own header file `pybind11/detail/function.h`, because I find it hard to refactor `cpp_function` as `pybind11/pybind11.h` is too long. I specifically choose the `detail` directory as I'd like to split `cpp_function` into two parts in the long term: one part (in `pybind11` namespace) being an `object` wrapper around a C++ function and another part (in `detail` namespace) dispatching the actual function call. I'll provide more details about these refactoring in upcoming PRs.

This PR does not change the behavior of `pybind11` so no changelog entry is needed.


include/pybind11/detail/function.h
```diff
@@ -0,0 +1,913 @@
...
+                msg += "\n";
+            }
+            msg += "\nInvoked with: ";
+            auto args_ = reinterpret_borrow<tuple>(args_in);
+            bool some_args = false;
+            for (size_t ti = overloads->is_constructor ? 1 : 0; ti < args_.size(); ++ti) {
+                if (!some_args) some_args = true;
+                else msg += ", ";
+                try {
+                    msg += repr(args_[ti]);
```
bstaletic, line 870: Any reason this was changed from `pybind11::repr` to `repr`? It should be perfectly fine, since `args_[ti]` has a type from the `pybind11` namespace, so even ADL would only find `pybind11::repr`, just curious if the change was intentional.

YannickJadoul:
I'd like to first get some context on this refactoring, before moving things around. Can you please create an issue with what you're planning to do and, more importantly, why, so that we can discuss? Anything that's refactoring something this big should be discussed with maintainers, before just throwing a series of PRs at us, since it has implications for existing PRs and forks, and shouldn't just be done "for fun".

Take the interaction with #2445 (first proposed/discussed in #2322) as an example of why this needs to be coordinated.

lqf96:
@YannickJadoul Yeah I agree we should probably discuss about this in advance... Let me open an issue later about the motivations and intentions...

lqf96:
@bstaletic @YannickJadoul I just opened #2810 to describe the problem I ran into, the refactors I want to do and the potential problems. It is more like a RFC so feel free to give me suggestions there.

YannickJadoul:
@lqf96, thanks, I saw! When I (soon) pull together my courage to think about this, I'll have a closer look and comment in that PR :-)

PR #2808, opened 2021-01-19
Prepend Pybind11Extension flags rather than appending them.

anntzer:
## Description

This allows users to set e.g. `extra_compile_args=["-g"]` (temporarily,
for debugging purposes) and not have that get overridden by
Pybind11Extension's default `-g0`.

In order to minimize "order inversion" (not that it really matters,
though) I chose to collect all flags and add them at once (hence the
dropping of `*args`).  I also dropped flag deduplication, which seems
unneeded.

## Suggested changelog entry:

```rst
``extra_compile_args`` and ``extra_link_args`` automatically set by
Pybind11Extension are now prepended, which allows them to be overridden
by user-set ``extra_compile_args`` and ``extra_link_args``.
```

### DIFF
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -99,15 +99,14 @@ class Pybind11Extension(_Extension):
     this is an ugly old-style class due to Distutils.
     """
 
-    def _add_cflags(self, *flags):
-        for flag in flags:
-            if flag not in self.extra_compile_args:
-                self.extra_compile_args.append(flag)
+    # flags are prepended, so that they can be further overridden, e.g. by
+    # ``extra_compile_args=["-g"]``.
 
-    def _add_lflags(self, *flags):
-        for flag in flags:
-            if flag not in self.extra_link_args:
-                self.extra_link_args.append(flag)
+    def _add_cflags(self, flags):
+        self.extra_compile_args[:0] = flags
+
+    def _add_ldflags(self, flags):
+        self.extra_link_args[:0] = flags
 
     def __init__(self, *args, **kwargs):
 
@@ -139,13 +138,17 @@ class Pybind11Extension(_Extension):
         # Have to use the accessor manually to support Python 2 distutils
         Pybind11Extension.cxx_std.__set__(self, cxx_std)
 
+        cflags = []
+        ldflags = []
         if WIN:
-            self._add_cflags("/EHsc", "/bigobj")
+            cflags += ["/EHsc", "/bigobj"]
         else:
-            self._add_cflags("-fvisibility=hidden", "-g0")
+            cflags += ["-fvisibility=hidden", "-g0"]
             if MACOS:
-                self._add_cflags("-stdlib=libc++")
-                self._add_lflags("-stdlib=libc++")
+                cflags += ["-stdlib=libc++"]
+                ldflags += ["-stdlib=libc++"]
+        self._add_cflags(cflags)
+        self._add_ldflags(ldflags)
 
     @property
     def cxx_std(self):
@@ -174,7 +177,8 @@ class Pybind11Extension(_Extension):
         if not level:
             return
 
-        self.extra_compile_args.append(STD_TMPL.format(level))
+        cflags = [STD_TMPL.format(level)]
+        ldflags = []
 
         if MACOS and "MACOSX_DEPLOYMENT_TARGET" not in os.environ:
             # C++17 requires a higher min version of macOS. An earlier version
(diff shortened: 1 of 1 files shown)


pybind11/setup_helpers.py
```diff
@@ -99,15 +99,14 @@ class Pybind11Extension(_Extension):
...
-                self.extra_compile_args.append(flag)
+    # flags are prepended, so that they can be further overridden, e.g. by
+    # ``extra_compile_args=["-g"]``.
 
-    def _add_lflags(self, *flags):
-        for flag in flags:
-            if flag not in self.extra_link_args:
-                self.extra_link_args.append(flag)
+    def _add_cflags(self, flags):
+        self.extra_compile_args[:0] = flags
```
henryiii, line 106: Wow, I did not know this was valid! But it is, even back in 2.7!

```python
WARNING: Python 2.7 is not recommended.
This version is included in macOS for compatibility with legacy software.
Future versions of macOS will not include Python 2.7.
Instead, it is recommended that you transition to using 'python3' from within Terminal.

Python 2.7.16 (default, Nov 23 2020, 08:01:20)
[GCC Apple LLVM 12.0.0 (clang-1200.0.30.4) [+internal-os, ptrauth-isa=sign+stri on darwin
>>> a = [1,2,3]
>>> a[:0] = [4,5,6]
>>> a
[4, 5, 6, 1, 2, 3]
```

Even can be used in the middle:

```python
>>> a[5:5] = [10]
>>> a
[4, 5, 6, 1, 2, 10, 3]
```

Maybe this is common knowledge, but it surprised me. :)

pybind11/setup_helpers.py
```diff
@@ -186,18 +190,21 @@ def cxx_std(self, level):
...
+                cflags += ["/wd5033"]
             elif level >= 17:
-                self.extra_compile_args.append("-Wno-register")
+                cflags += ["-Wno-register"]
             elif level >= 14:
-                self.extra_compile_args.append("-Wno-deprecated-register")
+                cflags += ["-Wno-deprecated-register"]
+
+        self._add_cflags(cflags)
+        self._add_ldflags(ldflags)
```
henryiii, line 207: Is there a reason these are prepended? This may happen at the beginning (if it is manually specified), at an arbitrary point (if assigned to, in Python 3), or at the end (if left to "automatic"). The order of these particular flags shouldn't really matter, in theory, I'd expect?

anntzer: I chose to just prepend everything to simplify the explanation ("pybind11 puts all its flags at the beginning") (also, perhaps someone really wants to override one of the -Wflags...), but admittedly the only thing I personally really care about is `-g`, so I can revert the changes on this specific part.

henryiii:
Question you might be able to answer: do you know how to get setuptools to put the user includes before the system includes? Setuptools adds the main directory in front of the user includes (specifically, the pybind11 one), so if you have a system install of pybind11, it might override the local one. I'm not sure of this, where I've seen the bug is for boost - my system install of boost was overriding the local one I was trying to provide in Pybind11Extension. I fixed this in CMake, but don't know how to fix it in setuptools at the moment.

pybind11/setup_helpers.py
henryiii: I think this is fine, it's the most consistent (I think) if this gets triggered at different times.

henryiii (approved):

anntzer:
> Question you might be able to answer: do you know how to get setuptools to put the user includes before the system includes? Setuptools adds the main directory in front of the user includes (specifically, the pybind11 one), so if you have a system install of pybind11, it might override the local one if it's in the same places as the Python include.

Just to be clear, this is unrelated to this PR?  I can have a look, but perhaps we can move the discussion to a separate issue?

In any case I'm not sure what's the behavior you're mentioning, as trying e.g. to compile `python_example` with `PYTHONPATH` set to `/path/to/pybind11/checkout` results for me in
```
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fno-semantic-interposition -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fPIC -DVERSION_INFO=0.0.1 -I/path/to/pybind11/checkout/pybind11/include -I/usr/include/python3.9 -c src/main.cpp -o build/temp.linux-x86_64-3.9/src/main.o -fvisibility=hidden -g0 -std=c++17
```
so here it's indeed the local install that comes first.

henryiii:
Yes, it's a separate issue. I'll try to test it a bit more in the near future, and I'll refer back here if I can show it again. Thanks!

anntzer:
feel free to ping me as needed.

PR #2814, opened 2021-01-22
Force the builtin module key to be the correct type.

crimsoncor:
Previously it was always going to be a std::string which converted into
unicode. Python 2 appears to want module keys to be normal str types, so
this was breaking code that expected plain string types in the
builtins.keys() data structure

## Suggested changelog entry:

```rst
Change the type of the module key in builtins that pybind11 uses to store its internals from std::string to a python str type. This will have no effect in python 3, but will change the key type from unicode to str on python 2
```

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -75,3 +75,11 @@ def test_duplicate_registration():
     """Registering two things with the same name"""
 
     assert m.duplicate_registration() == []
+
+
+def test_builtinsKeyType():
+    """Test that all the keys in the builtin modules have type str.
+
+    Previous versions of pybind11 would add a unicode key in python 2.
+    """
+    assert {type(k) for k in __builtins__.keys()} == {str}
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -77,14 +77,14 @@ def test_duplicate_registration():
     assert m.duplicate_registration() == []
 
 
-def test_builtinsKeyType():
+def test_builtin_key_type():
     """Test that all the keys in the builtin modules have type str.
 
     Previous versions of pybind11 would add a unicode key in python 2.
     """
-    if hasattr(__builtins__, 'keys'):
+    if hasattr(__builtins__, "keys"):
         keys = __builtins__.keys()
-    else: # this is to make pypy happy since builtins is different there.
+    else:  # this is to make pypy happy since builtins is different there.
         keys = __builtins__.__dict__.keys()
 
     assert {type(k) for k in keys} == {str}
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -82,4 +82,9 @@ def test_builtinsKeyType():
 
     Previous versions of pybind11 would add a unicode key in python 2.
     """
-    assert {type(k) for k in __builtins__.keys()} == {str}
+    if hasattr(__builtins__, 'keys'):
+        keys = __builtins__.keys()
+    else: # this is to make pypy happy since builtins is different there.
+        keys = __builtins__.__dict__.keys()
+
+    assert {type(k) for k in keys} == {str}
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -266,7 +266,7 @@ PYBIND11_NOINLINE inline internals &get_internals() {
         const PyGILState_STATE state;
     } gil;
 
-    constexpr auto *id = PYBIND11_INTERNALS_ID;
+    PYBIND11_STR_TYPE id(PYBIND11_INTERNALS_ID);
     auto builtins = handle(PyEval_GetBuiltins());
     if (builtins.contains(id) && isinstance<capsule>(builtins[id])) {
         internals_pp = static_cast<internals **>(capsule(builtins[id]));


YannickJadoul:
We discussed this on Gitter. I agree that it makes sense that a module's `__dict__`'s keys would be the Python version's `str` (so `py::bytes` for Python 2) rather than always `py::str` (so `unicode` in Python 2).

Fun thing, this is actually not breaking the ABI, and (probably) not even stopping different pybind11 libraries from interacting, since:

```pycon
[GCC 7.5.0] on linux2
>>> d = {}
>>> d[u'answer'] = 42
>>> d[u'answer']
42
>>> d['answer']
42
>>> type(u'answer')
<type 'unicode'>
>>> type('answer')
<type 'str'>
>>> 
```

Do you mind checking this, @crimsoncor, whether a pybind11 library from our current `master` manages to interact with another library that's built with this PR (on Python 2) ? If so, I don't really see a reason why this couldn't go into pybind11 2.7. (Potentially even 2.6.2, but that's maybe too close to release to still add this.)

crimsoncor:
Did some very basic testing of this (source code in the attached zip). I built one module with a pre-patch pybind11 and the other with post-patch (and then also flipped the order). In all cases, both modules were able to access the same internals block. What I did notice was that which one was loaded first determined the type of the argument in the builtin dict:
```
/crypt/code/other/pythonTest> python
Python 2.7.18 (default, Apr 23 2020, 09:27:04) [GCC] on linux2
>>> import foo
>>> import bar
>>> print( {type(k) for k in __builtins__.__dict__.keys()})
set([<type 'unicode'>, <type 'str'>])
>>> 
/crypt/code/other/pythonTest> python
Python 2.7.18 (default, Apr 23 2020, 09:27:04) [GCC] on linux2
>>> import bar
>>> import foo
>>> print( {type(k) for k in __builtins__.__dict__.keys()})
set([<type 'str'>])
```
But even with that said, regardless of the type of the key, it still worked for both modules. Also double checked to make sure there was ever only a single entry in the map (despite it being str vs unicode)
```
/crypt/code/other/pythonTest> python
Python 2.7.18 (default, Apr 23 2020, 09:27:04) [GCC] on linux2
>>> import foo
>>> import bar
>>> for k in __builtins__.__dict__.keys():
...   if "pybind" in k:
...     print(k)
... 
__pybind11_internals_v4_gcc_libstdcpp_cxxabi1011__
>>> 
/crypt/code/other/pythonTest> python
Python 2.7.18 (default, Apr 23 2020, 09:27:04) [GCC] on linux2
>>> import bar
>>> import foo
>>> for k in __builtins__.__dict__.keys():
...   if "pybind" in k:
...     print(k)
... 
__pybind11_internals_v4_gcc_libstdcpp_cxxabi1011__
```
And that looks like it also was correct

[pythonTest.zip](https://github.com/pybind/pybind11/files/5858524/pythonTest.zip)

YannickJadoul:
@henryiii, I'm going to say it's your call on whether we've already stopped accepting PRs for 2.6.2 :-P

YannickJadoul (approved):

henryiii:
If we can get another global testing run (which I think we will, @rwgk ?), maybe over the weekend, and release Monday, perhaps?

rwgk:
> If we can get another global testing run

OK, np, I'll include this PR in the next run.

rwgk:
@crimsoncor if you could add a minimal test that would be great.

crimsoncor:
> @crimsoncor if you could add a minimal test that would be great.

What sort of test do you think would be applicable? The existing tests that exercise internals being shared across multiple modules continue to function fine, which shows that changing the type of the key in python 2.7 has not caused any regression.

The only other test that seems like it might be applicable would be testing a module built with an older version and a module built with the new code (like I did manually) and I'm not sure how to add that in any sane fashion in the pybind11 test suite.

YannickJadoul:
@crimsoncor, what about just adding a single test that checks `{type(k) for k in __builtins__.__dict__.keys()} == {str}` ?

crimsoncor:
done

rwgk:
Awesome, thanks!
Minor issue: no newline at the end of the modified test file (I'm guessing some CI component will fail because of it).

tests/test_modules.py
```diff
@@ -75,3 +75,16 @@ def test_duplicate_registration():
...
+
+def test_builtin_key_type():
+    """Test that all the keys in the builtin modules have type str.
+
+    Previous versions of pybind11 would add a unicode key in python 2.
+    """
+    if hasattr(__builtins__, "keys"):
+        keys = __builtins__.keys()
+    else:  # this is to make pypy happy since builtins is different there.
+        keys = __builtins__.__dict__.keys()
```
bstaletic, line 88: Why not simply always go through `__builtins__.__dict__.keys()`? It works in CPython and PyPy.

crimsoncor: answered in gitter, but I'll add here as well. Python 2 has the following amazing feature
```
    In module __main__:
        __builtins__ is a reference to module __builtin__.
        __builtin__ only exists if you import it.
    In any other module:
        __builtins__ is a reference to module __builtin__'s __dict__.
        __builtin__ only exists if you import it.
```
so we need to handle both cases.

YannickJadoul: _sigh_

rwgk (approved):
I ran this through our global testing and found no issues.

YannickJadoul:
Thanks for creating this PR, @crimsoncor/@jesse-sony!

PR #2816, opened 2021-01-23
Enable -Wstrict-aliasing warning

bstaletic:
## Description

Even though our cmake uses `-fno-strict-aliasing`, not all extensions do. Since C++17, we can actually fix `-Wstrict-aliasing` errors. For now, let's just see how things fail in CI.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -10,14 +10,6 @@
 
 #pragma once
 
-#if defined(__cpp_lib_launder) && !(defined(_MSC_VER) && (_MSC_VER < 1914))
-#  define PYBIND11_STD_LAUNDER std::launder
-#  define PYBIND11_HAS_STD_LAUNDER 1
-#else
-#  define PYBIND11_STD_LAUNDER
-#  define PYBIND11_HAS_STD_LAUNDER 0
-#endif
-
 #if defined(__INTEL_COMPILER)
 #  pragma warning push
 #  pragma warning disable 68    // integer conversion resulted in a change of sign
@@ -43,9 +35,6 @@
 #  pragma GCC diagnostic ignored "-Wunused-but-set-parameter"
 #  pragma GCC diagnostic ignored "-Wunused-but-set-variable"
 #  pragma GCC diagnostic ignored "-Wmissing-field-initializers"
-#  if !PYBIND11_HAS_STD_LAUNDER
-#    pragma GCC diagnostic ignored "-Wstrict-aliasing"
-#  endif
 #  pragma GCC diagnostic ignored "-Wattributes"
 #  if __GNUC__ >= 7
 #    pragma GCC diagnostic ignored "-Wnoexcept-type"
@@ -63,6 +52,13 @@
 #include <string>
 #include <utility>
 
+#if defined(__cpp_lib_launder) && !(defined(_MSC_VER) && (_MSC_VER < 1914))
+#  define PYBIND11_STD_LAUNDER std::launder
+#  define PYBIND11_HAS_STD_LAUNDER 1
+#else
+#  define PYBIND11_STD_LAUNDER
+#  define PYBIND11_HAS_STD_LAUNDER 0
+#endif
 #if defined(__GNUG__) && !defined(__clang__)
 #  include <cxxabi.h>
 #endif
@@ -161,8 +157,20 @@ protected:
 #if defined(__GNUG__) && !defined(__clang__) && __GNUC__ >= 6
 #  pragma GCC diagnostic pop
 #endif
+#if defined(__GNUG__) && !PYBIND11_HAS_STD_LAUNDER
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#endif
+            // UB without std::launder, but without breaking ABI and/or
+            // a significant refactoring it's "impossible" to solve'
             if (!std::is_trivially_destructible<Func>::value)
-                rec->free_data = [](function_record *r) { PYBIND11_STD_LAUNDER((capture *) &r->data)->~capture(); };
+                rec->free_data = [](function_record *r) {
+                    auto cap = PYBIND11_STD_LAUNDER((capture *) &r->data);
+                    cap->~capture();
+                };
+#if defined(__GNUG__) && !PYBIND11_HAS_STD_LAUNDER
+#  pragma GCC diagnostic pop
+#endif
         } else {
             rec->data[0] = new capture { std::forward<Func>(f) };
             rec->free_data = [](function_record *r) { delete ((capture *) r->data[0]); };
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -162,11 +162,12 @@ protected:
 #  pragma GCC diagnostic ignored "-Wstrict-aliasing"
 #endif
             // UB without std::launder, but without breaking ABI and/or
-            // a significant refactoring it's "impossible" to solve'
+            // a significant refactoring it's "impossible" to solve.
             if (!std::is_trivially_destructible<Func>::value)
                 rec->free_data = [](function_record *r) {
-                    auto cap = PYBIND11_STD_LAUNDER((capture *) &r->data);
-                    cap->~capture();
+                    auto data = PYBIND11_STD_LAUNDER((capture *) &r->data);
+                    (void) data;
+                    data->~capture();
                 };
 #if defined(__GNUG__) && !PYBIND11_HAS_STD_LAUNDER
 #  pragma GCC diagnostic pop
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -10,6 +10,14 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/common.h
```diff
@@ -162,6 +162,9 @@
 #include <memory>
 #include <typeindex>
 #include <type_traits>
+#if defined(__has_include) && __has_include(<version>)
+#    include <version>
+#endif
```
bstaletic, line 167: Since C++20, `__cpp_lib_foo` feature test macros are in `<version>` header and at least clang respects that.

include/pybind11/pybind11.h
```diff
@@ -10,6 +10,14 @@
...
 #pragma once
 
+#if defined(__cpp_lib_launder) && !(defined(_MSC_VER) && (_MSC_VER < 1914))
+#  define PYBIND11_STD_LAUNDER std::launder
+#  define PYBIND11_HAS_STD_LAUNDER 1
+#else
+#  define PYBIND11_STD_LAUNDER
+#  define PYBIND11_HAS_STD_LAUNDER 0
+#endif
+
```
bstaletic, line 20: This is the best check that @YannickJadoul and I could come up with.

include/pybind11/pybind11.h
```diff
@@ -35,7 +43,9 @@
 #  pragma GCC diagnostic ignored "-Wunused-but-set-parameter"
 #  pragma GCC diagnostic ignored "-Wunused-but-set-variable"
 #  pragma GCC diagnostic ignored "-Wmissing-field-initializers"
-#  pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#  if !PYBIND11_HAS_STD_LAUNDER
+#    pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#  endif
```
bstaletic, line 48: In case we don't have `std::launder`, we can't tackle `-Wstrict-aliasin`. At least not easily, without breaking or a major refactoring.

include/pybind11/pybind11.h
```diff
@@ -151,7 +162,7 @@ class cpp_function : public function {
 #  pragma GCC diagnostic pop
 #endif
             if (!std::is_trivially_destructible<Func>::value)
-                rec->free_data = [](function_record *r) { ((capture *) &r->data)->~capture(); };
+                rec->free_data = [](function_record *r) { PYBIND11_STD_LAUNDER((capture *) &r->data)->~capture(); };
```
bstaletic, line 165: As far as I know, this is the only place where we need to launder.

bstaletic (commented):
Previous run, which errored: https://github.com/pybind/pybind11/pull/2816/checks?check_run_id=1755015877

YannickJadoul (approved):
I wish there was a way to avoid `std::launder` (and UB on C++<17), but ... myeah, it's better than before, at least.

include/pybind11/pybind11.h
bstaletic: Like I mentioned in the other comment, `__cpp_lib_launder` will be defined in `<version>` in C++20. If I define these macros here, I won't get that header from `detail/common.h`.

Should I include `<version>` here? Move these macro definitions to `common.h`?

Also `defined(__cpp_lib_launder)` should be `defined(__cpp_lib_launder) && __cpp_lib_launder >= 201606`.

include/pybind11/pybind11.h
```diff
@@ -151,7 +151,20 @@ class cpp_function : public function {
 #  pragma GCC diagnostic pop
 #endif
             if (!std::is_trivially_destructible<Func>::value)
-                rec->free_data = [](function_record *r) { ((capture *) &r->data)->~capture(); };
+#if defined(__cpp_lib_launder) && !(defined(_MSC_VER) && (_MSC_VER < 1914))
+#  define PYBIND11_STD_LAUNDER std::launder
+#  define PYBIND11_HAS_STD_LAUNDER 1
+#else
+#  define PYBIND11_STD_LAUNDER
+#  define PYBIND11_HAS_STD_LAUNDER 0
```
YannickJadoul, line 159: Oh, I'd still move this up (underneath `detail/common.h`) having been included?

include/pybind11/pybind11.h
```diff
@@ -150,8 +157,17 @@ class cpp_function : public function {
 #if defined(__GNUG__) && !defined(__clang__) && __GNUC__ >= 6
 #  pragma GCC diagnostic pop
 #endif
+#if !PYBIND11_HAS_STD_LAUNDER
```
YannickJadoul, line 160:
```suggestion
#if defined(__GNUG__) && !PYBIND11_HAS_STD_LAUNDER
```
or
```suggestion
#if defined(__GNUG__) && !defined(__clang__) && !PYBIND11_HAS_STD_LAUNDER
```

include/pybind11/pybind11.h
```diff
@@ -150,8 +157,20 @@ class cpp_function : public function {
...
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#endif
+            // UB without std::launder, but without breaking ABI and/or
+            // a significant refactoring it's "impossible" to solve'
             if (!std::is_trivially_destructible<Func>::value)
-                rec->free_data = [](function_record *r) { ((capture *) &r->data)->~capture(); };
+                rec->free_data = [](function_record *r) {
+                    auto cap = PYBIND11_STD_LAUNDER((capture *) &r->data);
+                    cap->~capture();
```
bstaletic, line 169: MSVC tells me that I'm not using `cap`. If I put it all on one line MSVC says that I am discarding the return value of `launder`, which MSVC decided to mark as `[[nodiscard]]`. Note that this affects MSVC 2015 and MSVC latest, but not MSVC 2017...

YannickJadoul:
I was not expecting that to work, but that message seems to have been received...
Well, it's Saturday evening. Let's not judge MSVC for having had one too many?

bstaletic:
I think I got this to work on icc as well.

henryiii (approved):
Always happy to see these ignores go away (even if only conditionally)! And become more specific to the code they are there for. 

I'm wondering if things like missing-field-initializers can go away after the last few PRs fixing clang-tidy?

Skylion007:
> Always happy to see these ignores go away (even if only conditionally)! And become more specific to the code they are there for.
> 
> I'm wondering if things like missing-field-initializers can go away after the last few PRs fixing clang-tidy?

Not quite, I can enable that clang-tidy check too.

bstaletic:
> I'm wondering if things like missing-field-initializers can go away after the last few PRs fixing clang-tidy?

I tried and gcc 11 did not complain. However, this is surprising:

```
In function ‘cast’,
    inlined from ‘operator()’ at /home/bstaletic/work/pybind11/include/pybind11/pybind11.h:213:43,
    inlined from ‘_FUN’ at /home/bstaletic/work/pybind11/include/pybind11/pybind11.h:191:21:
/home/bstaletic/work/pybind11/include/pybind11/cast.h:567:13: warning: ‘operator delete’ called on unallocated object ‘int_string_pair’ [-Wfree-nonheap-object]
  567 |             delete src;
      |             ^
/home/bstaletic/work/pybind11/include/pybind11/cast.h: In function ‘_FUN’:
/home/bstaletic/work/pybind11/tests/test_builtin_casters.cpp:174:40: note: declared here
  174 |     static std::pair<int, std::string> int_string_pair{2, "items"};
      |                                        ^
In function ‘cast’,
    inlined from ‘operator()’ at /home/bstaletic/work/pybind11/include/pybind11/pybind11.h:213:43,
    inlined from ‘_FUN’ at /home/bstaletic/work/pybind11/include/pybind11/pybind11.h:191:21:
/home/bstaletic/work/pybind11/include/pybind11/stl.h:186:5: warning: ‘operator delete’ called on unallocated object ‘lvv’ [-Wfree-nonheap-object]
  186 |     PYBIND11_TYPE_CASTER(Type, _("List[") + value_conv::name + _("]"));
      |     ^
/home/bstaletic/work/pybind11/include/pybind11/stl.h: In function ‘_FUN’:
/home/bstaletic/work/pybind11/tests/test_stl.cpp:79:38: note: declared here
   79 |     static std::vector<RValueCaster> lvv{2};
      |                                      ^
```

rwgk (approved):
The pragma block at the top of pybind11.h is evil (I lost a few hours already because of this), nice to see at least one item go.

Side remark: The repetition of a complex `#if` for the `pragma ... pop` seems unfortunate (DRY) but is common practice. Might be nice to go through at some point to do something like this:
```
#if something complicated
#  pragma ...
#  define PYBIND11_PRAGMA_ACTIVE 1
#endif
...
#if defined(PYBIND11_PRAGMA_ACTIVE)
#  pragma ...
#  undef PYBIND11_PRAGMA_ACTIVE
#endif 
```

rwgk:
> > I'm wondering if things like missing-field-initializers can go away after the last few PRs fixing clang-tidy?
> 
> I tried and gcc 11 did not complain. However, this is surprising:

None of our sanitizers picks up on this. I think they would if it was real.

Skylion007 (approved):

PR #2819, opened 2021-01-25
When determining if a shared_ptr already exists, use a test on the we…

elkhrt:
…ak_ptr instead of a try/catch block.

## Description

EDIT (@YannickJadoul): Quoting https://github.com/pybind/pybind11/pull/2819#discussion_r567281649

> My motivation for changing this is primarily a Google-internal one. Our C++ code is generally compiled with exceptions disabled - throwing an exception results in `SIGABRT`. For most uses of exceptions in `pybind11`, this is not an issue because exceptions are localized to Python interface code, and so we can compile that single module with exceptions enabled. But this particular `bad_weak_ptr` exception would be thrown by code in `libc` and so we can't rely on that having been built with exceptions enabled. Hence our need for avoiding the exception in 'normal' code paths.

## Suggested changelog entry:

```rst
Avoid relying on exceptions in C++17 when getting a `shared_ptr` holder from a `shared_from_this` class.
```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -162,6 +162,11 @@
 #include <memory>
 #include <typeindex>
 #include <type_traits>
+#if defined(__has_include)
+#  if __has_include(<version>)
+#    include <version>
+#  endif
+#endif
 
 #if PY_MAJOR_VERSION >= 3 /// Compatibility macros for various Python versions
 #define PYBIND11_INSTANCE_METHOD_NEW(ptr, class_) PyInstanceMethod_New(ptr)
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -869,10 +869,7 @@ inline static std::shared_ptr<T> try_get_shared_from_this(std::enable_shared_fro
 // See also https://en.cppreference.com/w/cpp/memory/enable_shared_from_this, and in particular
 // the `std::shared_ptr<Good> gp1 = not_so_good.getptr();` and `try`-`catch` parts of the example.
 #if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
-    if (!holder_value_ptr->weak_from_this().expired())
-        return holder_value_ptr->shared_from_this();
-    else
-        return nullptr;
+    return holder_value_ptr->weak_from_this().lock();
 #else
     try {
         return holder_value_ptr->shared_from_this();
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -862,5 +862,26 @@ public:
 // Forward-declaration; see detail/class.h
 std::string get_fully_qualified_tp_name(PyTypeObject*);
 
+template <typename T>
+inline static std::shared_ptr<T> try_get_shared_from_this(std::enable_shared_from_this<T> *holder_value_ptr) {
+// Pre C++17, this code path exploits undefined behavior, but is known to work on many platforms.
+// Use at your own risk!
+// See also https://en.cppreference.com/w/cpp/memory/enable_shared_from_this, and in particular
+// the `std::shared_ptr<Good> gp1 = not_so_good.getptr();` and `try`-`catch` parts of the example.
+#if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
+    if (!holder_value_ptr->weak_from_this().expired())
+        return holder_value_ptr->shared_from_this();
+    else
+        return nullptr;
+#else
+    try {
+        return holder_value_ptr->shared_from_this();
+    }
+    catch (const std::bad_weak_ptr &) {
+        return nullptr;
+    }
+#endif
+}
+
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1501,30 +1501,13 @@ public:
     }
 
 private:
-    template <typename T>
-    inline static std::shared_ptr<T> get_shared_from_this(std::enable_shared_from_this<T> *holder_value_ptr) {
-#if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
-        if (!holder_value_ptr->weak_from_this().expired())
-            return holder_value_ptr->shared_from_this();
-        else
-            return nullptr;
-#else
-        try {
-            return holder_value_ptr->shared_from_this();
-        }
-        catch (const std::bad_weak_ptr &) {
-            return nullptr;
-        }
-#endif
-    }
-
     /// Initialize holder object, variant 1: object derives from enable_shared_from_this
     template <typename T>
     static void init_holder(detail::instance *inst, detail::value_and_holder &v_h,
             const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
 
         auto sh = std::dynamic_pointer_cast<typename holder_type::element_type>(
-                get_shared_from_this(v_h.value_ptr<type>()));
+                detail::try_get_shared_from_this(v_h.value_ptr<type>()));
         if (sh) {
             new (std::addressof(v_h.holder<holder_type>())) holder_type(std::move(sh));
             v_h.set_holder_constructed();
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1501,26 +1501,34 @@ public:
     }
 
 private:
+    template <typename T>
+    inline static std::shared_ptr<T> get_shared_from_this(std::enable_shared_from_this<T> *holder_value_ptr) {
+#if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
+        if (!holder_value_ptr->weak_from_this().expired())
+            return holder_value_ptr->shared_from_this();
+        else
+            return nullptr;
+#else
+        try {
+            return holder_value_ptr->shared_from_this();
+        }
+        catch (const std::bad_weak_ptr &) {
+            return nullptr;
+        }
+#endif
+    }
+
     /// Initialize holder object, variant 1: object derives from enable_shared_from_this
     template <typename T>
     static void init_holder(detail::instance *inst, detail::value_and_holder &v_h,
             const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
-#if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
-        if (!v_h.value_ptr<type>()->weak_from_this().expired()) {
-#else
-        try {
-#endif
-            auto sh = std::dynamic_pointer_cast<typename holder_type::element_type>(
-                    v_h.value_ptr<type>()->shared_from_this());
-            if (sh) {
-                new (std::addressof(v_h.holder<holder_type>())) holder_type(std::move(sh));
-                v_h.set_holder_constructed();
-            }
-#if defined(__cpp_lib_enable_shared_from_this)
+
+        auto sh = std::dynamic_pointer_cast<typename holder_type::element_type>(
+                get_shared_from_this(v_h.value_ptr<type>()));
+        if (sh) {
+            new (std::addressof(v_h.holder<holder_type>())) holder_type(std::move(sh));
+            v_h.set_holder_constructed();
         }
-#else
-        } catch (const std::bad_weak_ptr &) {}
-#endif
 
         if (!v_h.holder_constructed() && inst->owned) {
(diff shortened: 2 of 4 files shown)


YannickJadoul (dismissed):
Thanks! Looks good to me!

Seems like this isn't even an improvement to avoid `try-catch`, but even a fix that avoids UB on C++<17?

Quoting https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/shared_from_this

> It is permitted to call `shared_from_this` only on a previously shared object, i.e. on an object managed by `std::shared_ptr` (in particular, `shared_from_this` cannot be called during construction of `*this`).
> 
> Otherwise the behavior is undefined (until C++17)`std::bad_weak_ptr` is thrown (by the `shared_ptr` constructor from a default-constructed `weak_this`) (since C++17).

There's no such note for `weak_from_this`: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/weak_from_this

Do you happen to know if this is now completely safe?

elkhrt:
Sadly `weak_from_this` only exists from C++ 17 onwards, so should probably revert to old behaviour for older versions of C++, undefined behaviour and all. I'll amend the PR.

include/pybind11/pybind11.h
```diff
@@ -1505,14 +1505,22 @@ class class_ : public detail::generic_type {
     template <typename T>
     static void init_holder(detail::instance *inst, detail::value_and_holder &v_h,
             const holder_type * /* unused */, const std::enable_shared_from_this<T> * /* dummy */) {
+#if defined(PYBIND11_CPP17)
+        if (!v_h.value_ptr<type>()->weak_from_this().expired()) {
```
bstaletic, line 1509: This looks good, but do we need to also guard it with the feature test macro? MSVC, being sloppy with feature test macros until ~recently, would need a version check instead.

- For MSVC, it's supported since 19.12.
- For normal compilers, that's `__cpp_lib_enable_shared_from_this >= 201603`.

rwgk: To not clutter code too much I usually wait until there's an actual problem. Our very comprehensive CI is green. How likely is it that those macros are actually useful?

YannickJadoul: Well, @bstaletic's solution is the correct thing to do, if we want to support all compiler ranges we say we support.

YannickJadoul: We also already have this kind of things. It's not really more clutter: https://github.com/pybind/pybind11/blob/587d5f840a234fc11475952c81ee68525875239f/include/pybind11/pybind11.h#L1197

bstaletic: Here's a simple use case that would break. Start by using Visual Studio 15.3, which comes with MSVC 19.10. Pass `/std:c++latest` to the compiler and watch it error on `weak_from_this()` not being a thing.

rwgk: The problem for us is that we don't have a practical way of testing the extra `#if defined`. It would be a shot in the blue. If you have access to "Visual Studio 15.3, which comes with MSVC 19.10", could you please help out pushing a commit with the required change to this PR?

YannickJadoul: I'm assuming @bstaletic found some official reference saying when MSVC added this feature?

These kinds of things are always tricky anyway. But I don't see how an informed, reasoned-about decision is worse/more of a stab in the dark than the current blunt `PYBIND11_CPP17`? Even if we're still wrong, we won't be more wrong than saying "C++17". And that's even more the case for when we can use the official `__cpp_lib_enable_shared_from_this`.

And pybind11 is already doing this in various other places, anyway. So if both these pass tests, let's pick the one we can reason about to be more correct than the other.
Unless there's a case where @bstaletic's solution can lead to dangerous/unpredictable situtations, ofc!

bstaletic: I mean, everyone can double check my sources:

1. [MSVC 19.12](https://devblogs.microsoft.com/cppblog/c17-progress-in-vs-2017-15-5-and-15-6/)
2. [MSVC 19.11](https://blogs.msdn.microsoft.com/vcblog/2017/08/11/c17-features-and-stl-fixes-in-vs-2017-15-3/)
3. [MSVC 19.10](https://devblogs.microsoft.com/cppblog/vs-2015-update-2s-stl-is-c17-so-far-feature-complete/)
4. [MSVC versioning](https://en.wikipedia.org/wiki/MSVC#Internal_version_numbering)
5. [`__cpp_lib_enable_shared_from_this` in the standard](https://eel.is/c++draft/version.syn#lib:__cpp_lib_enable_shared_from_this)
6. [Relevant committee paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0033r1.html)
7. [cppreference on feature test macros](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17)

And this is the error we can currently expect: https://godbolt.org/z/KKdTnE
Yes, that is with MSVC 19.10, not 19.11, but we can expect the same error on both.

elkhrt:
My attempts to add Microsoft-specific checks seem to have failed, for reasons which are obscure to me.

include/pybind11/pybind11.h
bstaletic: @elkhrt You missed the `&& (!defined(_MSC) || _MSC >=1912)` part.

Also, since this is a standard library feature test macro, since C++20, it has been moved to the `<version>` header. So we also need something like:

```
#if defined(__has_include)
#  if __has_include(<version>)
#    include <version>
# endif
#endif
```

Probably somewhere in `detail/common.h`.

And no, `defined(__has_include) && __has_include(<version>)` doesn't work, because old and crappy compilers get scared when they see that kind of thing.

Before anyone asks, yes, `<version>` is a must. I've had troubles with clang with `-std=c++20`.

YannickJadoul:
The one failed test is not to worry about; it's one we sometimes see and hasn't been fixed yet. Unrelated to this PR.

YannickJadoul:
Refactored out the function (so we only need to write this condition once) and added the `<version>` that @bstaletic mentioned. I believe this is what you meant, @bstaletic?

One more thing, @elkhrt: what was your reason for this change? According to [cppreference](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/shared_from_this), we still have UB in C++ versions < C++17, and the try-catch was already non-UB in C++17 and later. So I'm not sure what's the advantage that justifies partially duplicating this code?
Then again, it's only a handful of lines extra, so I'm fine including it. It's just that the refactoring shows we're actually not getting anything out of this extra complexity?

include/pybind11/pybind11.h
```diff
@@ -1501,18 +1501,34 @@ class class_ : public detail::generic_type {
...
 private:
+    template <typename T>
+    inline static std::shared_ptr<T> get_shared_from_this(std::enable_shared_from_this<T> *holder_value_ptr) {
+#if defined(__cpp_lib_enable_shared_from_this) && (!defined(_MSC_VER) || _MSC_VER >= 1912)
+        if (!holder_value_ptr->weak_from_this().expired())
+            return holder_value_ptr->shared_from_this();
+        else
+            return nullptr;
+#else
+        try {
```
rwgk, line 1512: We should get rid of this old code path as soon as we drop pre-c++17 support (not soon I guess, but still true). In the (long) meantime it would be prudent to warn people about UB with a comment, here is my first stab:

```
// Pre C++17, this code path exploits undefined behavior, but is known to work on many platforms.
// Use at your own risk.
// See also: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this,
// in particular the `try`-`catch` code around `std::shared_ptr<Good> gp1 = not_so_good.getptr();`
```

include/pybind11/pybind11.h
```diff
@@ -1501,18 +1501,34 @@ class class_ : public detail::generic_type {
     }
 
 private:
+    template <typename T>
```
rwgk, line 1504: While we're at it, what do you think about moving the code out into pybind11::detail (best IMO), or at least mark it as public. There is a chance I may need it for the `smart_holder` code, too. I don't see a reason for hiding/protecting it in an any way.

rwgk (approved):
Awesome, thanks @YannickJadoul!

include/pybind11/pybind11.h
rwgk: Oh, I forgot to write down one more suggestion: I'd name this `try_get_shared_from_this` or `get_shared_from_this_if_possible`, `optional_get_shared_from_this`, `get_shared_from_this_or_nullptr`, ...

YannickJadoul: Hmmm, public, I don't know, indeed, since this is not an API we really want to offer, as pybind11? `detail` makes sense, if you think you can use it in #2672. (Normally, I'd argue refactoring should be part of the PR that needs it, but this is a bit of a special case, probably ;-) )

> Oh, I forgot to write down one more suggestion: I'd name this `try_get_shared_from_this` or `get_shared_from_this_if_possible`, `optional_get_shared_from_this`, `get_shared_from_this_or_nullptr`, ...

Yes! When it's public, out of the scope of this class, definitely! I was first thinking of `safely_get_shared_from_this`, when refactoring, but that's not really true (as it's not officially "safe" pre-C++17).

`maybe_get_shared_from_this` or `try_get_shared_from_this` sounds good to me, yes.

include/pybind11/pybind11.h
YannickJadoul: Yes, this comment is needed, for later reference.

include/pybind11/detail/common.h
```diff
@@ -857,5 +862,26 @@ class any_container {
...
+    else
+        return nullptr;
+#else
+    try {
+        return holder_value_ptr->shared_from_this();
+    }
+    catch (const std::bad_weak_ptr &) {
+        return nullptr;
+    }
+#endif
```
YannickJadoul, line 883: In principle, this is just as safe/unsafe as the more complex code we have now. @elkhrt, do you have a reason to prefer `if (!holder_value_ptr->weak_from_this().expired())`?

```suggestion
    try {
        return holder_value_ptr->shared_from_this();
    }
    catch (const std::bad_weak_ptr &) {
        return nullptr;
    }
```

YannickJadoul (commented):
Thanks, @rwgk. Pushed another commit with this fixed. One more question to resolve. If it's up to me, I'd still simplify the code (and ironically +- end up where we started).

include/pybind11/detail/common.h
YannickJadoul: (note: I don't like exception too much either, but pybind11 is already permeated by exceptions anyway, so one more isn't going to make a perceivable difference?)

bstaletic: The "new" function, in C++20 could just be

    return holder_value_ptr->weak_from_this().lock();

No need for ifs and elses and...

But if I'm not making sense, give me ~20h to recover my laptop.

elkhrt: Thanks for the refinements, everyone. My motivation for changing this is primarily a Google-internal one. Our C++ code is generally compiled with exceptions disabled - throwing an exception results in `SIGABRT`. For most uses of exceptions in `pybind11`, this is not an issue because exceptions are localized to Python interface code, and so we can compile that single module with exceptions enabled. But this particular `bad_weak_ptr` exception would be thrown by code in `libc` and so we can't rely on that having been built with exceptions enabled. Hence the need for avoiding the exception in 'normal' code paths.

YannickJadoul: > The "new" function, in C++20 could just be
> 
> ```
> return holder_value_ptr->weak_from_this().lock();
> ```

I'm not going to make an exception for C++20. But https://en.cppreference.com/w/cpp/memory/weak_ptr/lock says C++11 (and `noexcept`), so let me give that a try.

YannickJadoul: @elkhrt, thanks! That's useful information for the motiviation of this PR. Let me quickly add that to the main message at the start.

Right, it's a pity that we can't really do anything for pre-C++17, but avoiding exceptions this way makes sense. Let me quickly try @bstaletic's suggestion, but then we can just merge :-)

bstaletic: Right, C++17, not C++20.

YannickJadoul: Works locally. Let's see if CI agrees. @rwgk, do you still want to test, or just merge, if things turn green?

YannickJadoul: @elkhrt, any idea if there are more parts of pybind11 like this? Not to be included in this PR, ofc, but if we're making some kind of contract that pybind11 only relies on exceptions that pybind11 throws, this would be interesting to know.

rwgk: > Works locally. Let's see if CI agrees. @rwgk, do you still want to test, or just merge, if things turn green?

I'm gonna submit this to our big testing mill now. If all goes well results will be complete in ~4-5 hours.
Thanks a lot for the thoroughness!

elkhrt: @YannickJadoul - I'm pretty sure this is the only case, based both on searching the code and on a lack of other crashes.

YannickJadoul:
Green, green, green, green, ...

Thanks, @elkhrt!

include/pybind11/detail/common.h
YannickJadoul: Great. Good to know; thanks, @elkhrt.

PR #2821, opened 2021-01-26
chore: prepare for the 2.6.2 release

henryiii:
Minor missing functionality added:

* enum: add missing Enum.value property. [#2739](https://github.com/pybind/pybind11/pull/2739)
* Allow thread termination to be avoided during shutdown for CPython 3.7+ via `.disarm` for `gil_scoped_acquire`/`gil_scoped_release`. [#2657](https://github.com/pybind/pybind11/pull/2657)

Fixed or improved behavior in a few special cases:

* Fix bug where the constructor of `object` subclasses would not throw on being passed a Python object of the wrong type. [#2701](https://github.com/pybind/pybind11/pull/2701)
* The `type_caster` for integers does not convert Python objects with  `__int__` anymore with `noconvert` or during the first round of trying overloads. [#2698](https://github.com/pybind/pybind11/pull/2698)
* When casting to a C++ integer, `__index__` is always called and not considered as conversion, consistent with Python 3.8+. [#2801](https://github.com/pybind/pybind11/pull/2801)

Build improvements:

* Setup helpers: `extra_compile_args` and `extra_link_args` automatically set by Pybind11Extension are now prepended, which allows them to be overridden by user-set `extra_compile_args` and `extra_link_args`.  [#2808](https://github.com/pybind/pybind11/pull/2808)
* Setup helpers: Don\'t trigger unused parameter warning.  [#2735](https://github.com/pybind/pybind11/pull/2735)
* CMake: Support running with `--warn-uninitialized` active.  [#2806](https://github.com/pybind/pybind11/pull/2806)
* CMake: Avoid error if included from two submodule directories.  [#2804](https://github.com/pybind/pybind11/pull/2804)
* CMake: Fix `STATIC` / `SHARED` being ignored in FindPython mode.  [#2796](https://github.com/pybind/pybind11/pull/2796)
* CMake: Respect the setting for `CMAKE_CXX_VISIBILITY_PRESET` if defined. [#2793](https://github.com/pybind/pybind11/pull/2793)
* CMake: Fix issue with FindPython2/FindPython3 not working with `pybind11::embed`.  [#2662](https://github.com/pybind/pybind11/pull/2662)
* CMake: mixing local and installed pybind11\'s would prioritize the installed one over the local one (regression in 2.6.0).  [#2716](https://github.com/pybind/pybind11/pull/2716)

Bug fixes:

* Fixed segfault in multithreaded environments when using `scoped_ostream_redirect`.  [#2675](https://github.com/pybind/pybind11/pull/2675)
* Leave docstring unset when all docstring-related options are disabled, rather than set an empty string.  [#2745](https://github.com/pybind/pybind11/pull/2745)
* The module key in builtins that pybind11 uses to store its internals changed from std::string to a python str type (more natural on Python 2, no change on Python 3).  [#2814](https://github.com/pybind/pybind11/pull/2814)
* Fixed assertion error related to unhandled (later overwritten) exception in CPython 3.8 and 3.9 debug builds.  [#2685](https://github.com/pybind/pybind11/pull/2685)
* Fix `py::gil_scoped_acquire` assert with CPython 3.9 debug build.  [#2683](https://github.com/pybind/pybind11/pull/2683)
* Fix issue with a test failing on PyTest 6.2.  [#2741](https://github.com/pybind/pybind11/pull/2741)

Warning fixes:

* Fix warning modifying constructor parameter \'flag\' that shadows a field of \'set_flag\' `[-Wshadow-field-in-constructor-modified]`.  [#2780](https://github.com/pybind/pybind11/pull/2780)
* Suppressed some deprecation warnings about old-style `__init__`/`__setstate__` in the tests.  [#2759](https://github.com/pybind/pybind11/pull/2759)

Valgrind work:

* Fix invalid access when calling a pybind11 `__init__` on a non-pybind11 class instance.  [#2755](https://github.com/pybind/pybind11/pull/2755)
* Fixed various minor memory leaks in pybind11\'s test suite.  [#2758](https://github.com/pybind/pybind11/pull/2758)
* Resolved memory leak in cpp_function initialization when exceptions occurred. [#2756](https://github.com/pybind/pybind11/pull/2756)
* Added a Valgrind build, checking for leaks and memory-related UB, to CI. [#2746](https://github.com/pybind/pybind11/pull/2746)

Compiler support:

* Intel compiler was not activating C++14 support due to a broken define. [#2679](https://github.com/pybind/pybind11/pull/2679)
* Support ICC and NVIDIA HPC SDK in C++17 mode. [#2729](https://github.com/pybind/pybind11/pull/2729)
* Support Intel OneAPI compiler (ICC 20.2) and add to CI. [#2573](https://github.com/pybind/pybind11/pull/2573)


YannickJadoul:
Wow, this was autogenerated, I hope? This is just the changelog as it will be?

setup.cfg
```diff
@@ -63,4 +63,20 @@ ignore =
 
 [mypy]
 files = pybind11
-strict = True
+python_version = 2.7
+warn_unused_configs = True
+
+# Currently (0.800) identical to --strict
```
YannickJadoul, line 69: Is this something to urgently add now?

henryiii: The old version was broken, there isn't a `strict=True` flag. This is just writing out exactly what `--strict` is on the command line (it passes).

YannickJadoul: But well, if it's the same as before, it can't really hurt either, I guess. Still, it's a bit of an odd change in just a changelog + version update.

henryiii:
Ehh, partially. I want to make it fully automated, but gave up on messing with the regex to pick out the first block, will work on it later. This is the result of pandoc + a few quick manips in VI.

setup.cfg
henryiii: In the release procedure "check the setup.cfg" is listed, and I noticed this when I was checking the setup.cfg.

henryiii: Since mypy gets checked by CI, and these settings do not affect downstream projects at all, I think it's fine.

henryiii:
And, yes, this is markdown I can copy and paste into release notes. :)

setup.cfg
YannickJadoul: Aaaah, I see, OK. Shouldn't our builds have been broken, then? :-)

YannickJadoul:
Ah, that's great! :-)

setup.cfg
henryiii: We didn't have `warn_unused_configs = True`, so it just ignored the fact we had an invalid config line here.

YannickJadoul (approved):
Not an approval you urgently need, but ... here you go

henryiii:
Thinking about dropping AppVeyor, now that we have Github actions testing MSVC 2015... One of the AppVeyor tests hung up. If it's Appveyor's fault, I'm interested in dropping it. If it's our fault, then not so much...

henryiii:
It seems to have died on `test_gil_scoped`, which is worrisome.

henryiii:
It's fine now. I've added pytest-timeout (just for AppVeyor) just in case it happens again, we might get some debugging info then (if a job times out, you get absolutely nothing. If pytest-timeout does, you often get a nice traceback to see where it died. Might not work for a compiled piece, but better than no chance at all!)

YannickJadoul:
> `test_gil_scoped`

Probably still linked to #2754?

henryiii:
A GHA MSVC job stalled this time.

henryiii:
It's out! Bumped brew too. https://github.com/Homebrew/homebrew-core/pull/69840

PR #2822, opened 2021-01-26
Factoring out find_registered_python_instance() from type_caster_generic::cast.

rwgk:
This PR is simply moving a complex nested loop into its own function. The primary motivation is to avoid code duplication in PR #2672, but this PR is also a small improvement towards more modular code in itself.

Changelog not needed.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -207,7 +207,7 @@ PYBIND11_NOINLINE inline handle get_type_handle(const std::type_info &tp, bool t
     return handle(type_info ? ((PyObject *) type_info->type) : nullptr);
 }
 
-// Searches all_type_info for a registered instance, i.e. moving up the inheritance hierarchy.
+// Searches the inheritance graph for a registered Python instance, using all_type_info().
 PYBIND11_NOINLINE inline handle find_registered_python_instance(void *src,
                                                                 const detail::type_info *tinfo) {
     auto it_instances = get_internals().registered_instances.equal_range(src);
@@ -521,8 +521,7 @@ public:
         if (src == nullptr)
             return none().release();
 
-        handle registered_inst = find_registered_python_instance(src, tinfo);
-        if (registered_inst)
+        if (handle registered_inst = find_registered_python_instance(src, tinfo))
             return registered_inst;
 
         auto inst = reinterpret_steal<object>(make_new_instance(tinfo->type));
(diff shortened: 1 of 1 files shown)


include/pybind11/cast.h
```diff
@@ -207,6 +207,19 @@ PYBIND11_NOINLINE inline handle get_type_handle(const std::type_info &tp, bool t
     return handle(type_info ? ((PyObject *) type_info->type) : nullptr);
 }
 
+// Searches all_type_info for a registered instance, i.e. moving up the inheritance hierarchy.
```
YannickJadoul, line 210: > i.e. moving up the inheritance hierarchy.

Is that comment correct? I don't believe so. Where is this thing climbing the hierarchy?

According to `all_type_info`, it's doing something like that, but not really the full hierarchy:

```
/**
 * Extracts vector of type_info pointers of pybind-registered roots of the given Python type.  Will
 * be just 1 pybind type for the Python type of a pybind-registered class, or for any Python-side
 * derived class that uses single inheritance.  Will contain as many types as required for a Python
 * class that uses multiple inheritance to inherit (directly or indirectly) from multiple
 * pybind-registered classes.  Will be empty if neither the type nor any base classes are
 * pybind-registered.
 *
 * The value is cached for the lifetime of the Python type.
 */
```

So `all_type_info` seems to just be finding the _first_ pybind11 class(es) up the hierarchy, but I don't see the entries in `all_type_info` ordered by inheritance?

YannickJadoul (approved):
Makes sense to me, as a refactoring!

include/pybind11/cast.h
rwgk: I reworded the comment for clarity.
I didn't mean to claim it's searching the entire hierarchy, I just wanted to point to `all_type_info`. I added `()` to make that more obvious, in addition to shuffling the words.
I also removed the word `up`, although I think it was correct, but it doesn't really need to be mentioned here, because `all_type_info` has a pretty detailed explanation.

include/pybind11/cast.h
```diff
@@ -508,13 +521,9 @@ class type_caster_generic {
...
 
-        auto it_instances = get_internals().registered_instances.equal_range(src);
-        for (auto it_i = it_instances.first; it_i != it_instances.second; ++it_i) {
-            for (auto instance_type : detail::all_type_info(Py_TYPE(it_i->second))) {
-                if (instance_type && same_type(*instance_type->cpptype, *tinfo->cpptype))
-                    return handle((PyObject *) it_i->second).inc_ref();
-            }
-        }
+        handle registered_inst = find_registered_python_instance(src, tinfo);
+        if (registered_inst)
```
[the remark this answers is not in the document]
rwgk: Change applied.

rwgk:
Thanks @YannickJadoul!
I'll merge this when I see that the CI is still all green.

PR #2823, opened 2021-01-27
fix: MSVC issue

henryiii:
Attempt to fix issue seen in #2821. Pulled out the non-changelog part of that PR too (adding pytest-timeout, corrected mypy config).

CC @YannickJadoul @rwgk @bstaletic

### DIFF
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -52,9 +52,11 @@ private:
 
                 pywrite(line);
                 pyflush();
+
+                // Placed inside gil_scoped_aquire as a mutex to avoid a race
+                setp(pbase(), epptr());
             }
 
-            setp(pbase(), epptr());
         }
         return 0;
     }
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,7 +19,7 @@ install:
     if ($env:PLATFORM -eq "x64") { $env:PYTHON = "$env:PYTHON-x64" }
     $env:PATH = "C:\Python$env:PYTHON\;C:\Python$env:PYTHON\Scripts\;$env:PATH"
     python -W ignore -m pip install --upgrade pip wheel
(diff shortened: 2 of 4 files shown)


YannickJadoul:
Cross-linking to #2754

YannickJadoul (approved):

include/pybind11/iostream.h
```diff
@@ -52,9 +52,12 @@ class pythonbuf : public std::streambuf {
 
                 pywrite(line);
                 pyflush();
+
+                // Placed inside gil_scoped_aquire as a makeshift mutex
```
bstaletic, line 56: Is "makeshift" the right word? Doesn't it just use the OS primitives, somewhere in the depths of CPython?

bstaletic (approved):

include/pybind11/iostream.h
henryiii: I almost dropped it... Probably should have. If it builds, I'll drop it and we give it a second chance to fail. ;)

rwgk:
I didn't follow the conversations enough to know if this PR was expected to also fix the TSAN error reported under #2754, but I just tried, and I still see this with current `master`:
```
  WARNING: ThreadSanitizer: data race llvm/llvm-project/libcxx/include/streambuf:462:13 in std::__tsan::basic_streambuf<char, std::__tsan::char_traits<char> >::xsputn(char const*, long) (pybind11/tests/test_iostream)
  WARNING: ThreadSanitizer: data race llvm/llvm-project/libcxx/include/__string:388:56 in std::__tsan::char_traits<char>::copy(char*, char const*, unsigned long) (pybind11/tests/test_iostream)
```

henryiii:
I saw the output issue again, but at least the hangup stopped, and the issue is rarer than it used to be. Issue was in one job out of 5-6 sets of builds.

YannickJadoul:
Hmmm, so we're still missing something? :-(

PR #2824, opened 2021-01-27
MSVC but not Clang: /MP

ax3l:
## Description

On Windows, clang-cl does not understand `/MP`.
```
clang-cl: warning: argument unused during compilation: '/MP' [-Wunused-command-line-argument]
```
with Clang 10.0.0 and Ninja.

## Suggested changelog entry:

```rst
Clang on Windows: do not pass /MP (ignored flag)
```

### DIFF
--- a/tools/pybind11Common.cmake
+++ b/tools/pybind11Common.cmake
@@ -115,28 +115,32 @@ endif()
 
 add_library(pybind11::windows_extras IMPORTED INTERFACE ${optional_global})
 
-if(MSVC)
-  # /MP enables multithreaded builds (relevant when there are many files), /bigobj is
-  # needed for bigger binding projects due to the limit to 64k addressable sections
+if(MSVC) # That's also clang-cl
+  # /bigobj is needed for bigger binding projects due to the limit to 64k
+  # addressable sections
   set_property(
     TARGET pybind11::windows_extras
     APPEND
     PROPERTY INTERFACE_COMPILE_OPTIONS /bigobj)
 
-  if(CMAKE_VERSION VERSION_LESS 3.11)
-    set_property(
-      TARGET pybind11::windows_extras
-      APPEND
-      PROPERTY INTERFACE_COMPILE_OPTIONS $<$<NOT:$<CONFIG:Debug>>:/MP>)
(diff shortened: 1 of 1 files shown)


henryiii (approved):

henryiii:
Needs a `pre-commit run -a` :)

tools/pybind11Common.cmake
```diff
@@ -115,28 +115,31 @@ endif()
...
-    # Only set these options for C++ files.  This is important so that, for
-    # instance, projects that include other types of source files like CUDA
-    # .cu files don't get these options propagated to nvcc since that would
-    # cause the build to fail.
-    set_property(
-      TARGET pybind11::windows_extras
-      APPEND
-      PROPERTY INTERFACE_COMPILE_OPTIONS $<$<NOT:$<CONFIG:Debug>>:$<$<COMPILE_LANGUAGE:CXX>:/MP>>)
+  # /MP enables multithreaded builds (relevant when there are many files) for MSVC
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC") # no Clang no Intel
```
zhihaoy, line 127: Or to check whether `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` is empty https://github.com/Kitware/CMake/blob/master/Modules/CMakeDetermineCXXCompiler.cmake#L129-L143

henryiii: Should be checked to make sure it is valid for all versions we support - CMake has a bad habit of not adding those the first round. If so, then it seems more elegant. :)

henryiii: `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` wasn't there in 3.11, which is where this block of code starts taking effect.

zhihaoy (approved):

henryiii:
Thats odd, `test_recarray` has a random nan in it.

henryiii:
Quite sure that has nothing to do with this PR.

bstaletic:
@henryiii Same error surprised @YannickJadoul in #2846 .

PR #2831, opened 2021-01-28
Add helper to build in-tree extensions.

anntzer:
## Description

For single-file extensions, a convenient pattern offered by cython
is to place the source files directly in the python source tree
(`foo/__init__.py`, `foo/ext.pyx`), deriving the package names from
their filesystem location.  Adapt this pattern for pybind11, using an
`intree_extensions` helper, which should be thought of as the moral
equivalent to `cythonize`.

Differences with cythonize: I chose not to include globbing support
(`intree_extensions(glob.glob("**/*.cpp"))` seems sufficient), nor to
provide extension-customization kwargs (directly setting the attributes
on the resulting Pybind11Extension objects seems sufficient).

We could choose to have `intree_extension` (singular instead) and make
users write `[*map(intree_extension, glob.glob("**/*.cpp"))]`; no strong
opinion here.

(I am not wedded to the specific design here, let me know if you have
alternative designs in mind.)

## Suggested changelog entry:

```rst
`pybind11.setup_helpers.intree_extensions` can be used to generate
`Pybind11Extension` instances from cpp files placed in the Python package
source tree.
```

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -70,6 +70,19 @@ that is supported via a ``build_ext`` command override; it will only affect
         ext_modules=ext_modules
     )
 
+If you have single-file extension modules that are directly stored in the
+Python source tree (``foo.cpp`` in the same directory as where a ``foo.py``
+would be located), you can also generate ``Pybind11Extensions`` using
+``setup_helpers.intree_extensions``: ``intree_extensions(["path/to/foo.cpp",
+...])`` returns a list of ``Pybind11Extensions`` which can be passed to
+``ext_modules``, possibly after further customizing their attributes
+(``libraries``, ``include_dirs``, etc.).  By doing so, a ``foo.*.so`` extension
+module will be generated and made available upon installation.
+
+``intree_extension`` will automatically detect if you are using a ``src``-style
+layout (as long as no namespace packages are involved), but you can also
+explicitly pass ``package_dir`` to it (as in ``setuptools.setup``).
+
 Since pybind11 does not require NumPy when building, a light-weight replacement
 for NumPy's parallel compilation distutils tool is included. Use it like this:
 
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -302,6 +302,42 @@ class build_ext(_build_ext):  # noqa: N801
         _build_ext.build_extensions(self)
 
 
+def intree_extensions(paths, package_dir=None):
+    """
+    Generate Pybind11Extensions from source files directly located in a Python
+    source tree.
+
+    ``package_dir`` behaves as in ``setuptools.setup``.  If unset, the Python
+    package root parent is determined as the first parent directory that does
+    not contain an ``__init__.py`` file.
+    """
+    exts = []
+    for path in paths:
+        if package_dir is None:
+            parent, _ = os.path.split(path)
+            while os.path.exists(os.path.join(parent, "__init__.py")):
+                parent, _ = os.path.split(parent)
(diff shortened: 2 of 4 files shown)


YannickJadoul:
@henryiii, I think this is up to you, to help determining what the scope of `setup_helpers` is, and whether this is in scope or should be an external addition.

henryiii:
I'm mildly in favor, I think; it's tested and not too hard to maintain, and adds a feature that is provided with Cython (though personally I'm not really in favor of structuring code this way). My thoughts here are we can do what we can to make pure setuptools work as smoothly as possible, and leave cmake to scikit-build or manual extensions.

EricCousineau-TRI:
I also would consider this slightly defective, but possible a desirable workflow.

FWIW We ([Drake](https://drake.mit.edu/) peeps) don't build our extension modules in source tree; however, we use [Bazel](https://docs.bazel.build/versions/4.0.0/bazel-overview.html) (open-source variant of what @rwgk uses, I believe), where dependencies (`data` and semantic `deps`) are declared. This then admits itself to appearing "inline" by means of how you construct your "runfiles" directory.

anntzer:
Having thought a bit more about this, I am still in favor of the feature, but would even additionally direcly include globbing support (i.e. allow one to write `intree_extensions("**/*.cpp")` rather than `intree_extensions(glob.glob("**/*.cpp"))`), as that seems likely to be a common use case for which we can save an import on the user side.  Thoughts?

docs/compiling.rst
```diff
@@ -70,6 +70,17 @@ that is supported via a ``build_ext`` command override; it will only affect
         ext_modules=ext_modules
     )
 
+If you have single-file extension modules that are directly stored in the
```
rwgk, line 73: One feedback datapoint: after skimming through this paragraph I'm asking myself "uhm, what is this doing?"

Guessing: is foo.cpp compiled and linked as foo.so when running pip install?
If that's it, saying that first could make the rest easier to digest.

Similar to @EricCousineau-TRI, I don't have a lot of experience with setuptools.
But if this feature is well explained and tested I'd support adding it into pybind11.
Maybe with some small addition: For questions regarding this feature please tag @anntzer.

anntzer: I clarified the point as suggested.
I'm happy to be tagged if there's any question, but would rather not add a big pointer to myself in the docs :)

rwgk: Thanks! That's clear now.

What do you think about adding a `# source code comment` in setup_helpers.py, above or below the docstring, to tell people about the tagging?

There is a bigger general question about pybind11 maintenance here, which is beyond this PR. I'm basically poking around finding some easy way to point people in the right direction when they run into issues. Currently we have 300+ open issues but no process to even just triage them. Hence my suggestion for some sort of feature-specific pointer. Ideas welcome!

rwgk (approved):

docs/compiling.rst
anntzer: `git blame` seems the way to go?  https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners may be worth leveraging too.

rwgk: That's a significant barrier, X% of people will not jump it. Code tends to get moved around over time, then even of those X%, X+Y% will find the wrong person to "blame", or give up when they see the code was moved.
This PR is not the right place to solve the more general problem. I'm OK with this PR to go in without a pointer.

henryiii (approved):

Skylion007 (approved):

PR #2832, opened 2021-01-29
Fix confusing weakref constructor overload

### LINKED ISSUE #2536: [BUG] Confusing weakref constructor copy the object instead of creating a weakref
My colleague and I stumbled on a very confusing part of the API. When calling pybind11::weakref with a single argument that is of type py::object as such:
```

py::object WRONGcreateWeakRef(py::object object)
{
        return py::weakref(object);
}

```
Here is a correct way to call it:

```
py::object GOODcreateWeakRef(py::object object)
{
        return py::weakref(static_cast<py::handle>(object));
}
```

This is caused by the fact that the weakref constructor is explicit and the copy constructor is called instead. The result is that no weakref is created. This is very confusing. The copy constructor should allow only to copy weakref. May be by using a private copy constructor that takes an object and an public copy constructor that takes a weakref.

After some discussion on gitter, it seems that https://github.com/pybind/pybind11/pull/2349 will solve this issue by raising a TypeError. However, the API would be still awkward.

YannickJadoul:
## Description

Fixes #2536, where the two `weakref` constructors from `handle` and `object` do different things.
Demonstrated by failing tests in the first commit, fixed in the second one.

## Suggested changelog entry:

```rst
Fix the ``weakref`` constructor from ``py::object`` to create a new ``weakref`` on conversion.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -819,6 +819,10 @@ PYBIND11_NAMESPACE_END(detail)
     : Parent(check_(o) ? o.release().ptr() : ConvertFun(o.ptr()), stolen_t{}) \
     { if (!m_ptr) throw error_already_set(); }
 
+#define PYBIND11_OBJECT_CVT_DEFAULT(Name, Parent, CheckFun, ConvertFun) \
+    PYBIND11_OBJECT_CVT(Name, Parent, CheckFun, ConvertFun) \
+    Name() : Parent() { }
+
 #define PYBIND11_OBJECT_CHECK_FAILED(Name, o_ptr) \
     ::pybind11::type_error("Object of type '" + \
                            ::pybind11::detail::get_fully_qualified_tp_name(Py_TYPE(o_ptr)) + \
@@ -1168,7 +1172,7 @@ public:
 
 class weakref : public object {
 public:
-    PYBIND11_OBJECT_CVT(weakref, object, PyWeakref_Check, raw_weakref)
+    PYBIND11_OBJECT_CVT_DEFAULT(weakref, object, PyWeakref_Check, raw_weakref)
     explicit weakref(handle obj, handle callback = {})
         : object(PyWeakref_NewRef(obj.ptr(), callback.ptr()), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate weak reference!");
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1168,10 +1168,15 @@ public:
 
 class weakref : public object {
 public:
-    PYBIND11_OBJECT_DEFAULT(weakref, object, PyWeakref_Check)
+    PYBIND11_OBJECT_CVT(weakref, object, PyWeakref_Check, raw_weakref)
     explicit weakref(handle obj, handle callback = {})
         : object(PyWeakref_NewRef(obj.ptr(), callback.ptr()), stolen_t{}) {
         if (!m_ptr) pybind11_fail("Could not allocate weak reference!");
+    }
+
+private:
+    static PyObject *raw_weakref(PyObject *o) {
+        return PyWeakref_NewRef(o, nullptr);
     }
 };
 
--- a/tests/test_pytypes.py
+++ b/tests/test_pytypes.py
@@ -543,7 +543,14 @@ def test_pass_bytes_or_unicode_to_string_types():
             m.pass_to_pybind11_str(malformed_utf8)
 
 
-def test_weakref():
+@pytest.mark.parametrize(
+    "create_weakref, create_weakref_with_callback",
+    [
+        (m.weakref_from_handle, m.weakref_from_handle_and_function),
+        (m.weakref_from_object, m.weakref_from_object_and_function),
+    ],
+)
+def test_weakref(create_weakref, create_weakref_with_callback):
     from weakref import getweakrefcount
 
     # Apparently, you cannot weakly reference an object()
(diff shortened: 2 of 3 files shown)


include/pybind11/pytypes.h
```diff
@@ -1163,11 +1163,16 @@ class float_ : public object {
 
 class weakref : public object {
 public:
-    PYBIND11_OBJECT_DEFAULT(weakref, object, PyWeakref_Check)
+    PYBIND11_OBJECT_CVT(weakref, object, PyWeakref_Check, raw_weakref)
```
bstaletic, line 1166: Technically an API break. Do we care about users digging up `py::weakref::weakref()`?

tests/test_pytypes.py
```diff
@@ -488,3 +488,45 @@ def test_builtin_functions():
...
+
+    # Apparently, you cannot weakly reference an object()
+    class WeaklyReferenced(object):
+        pass
+
+    called = [0]
+
+    def callback(wr):
+        # No `nonlocal` in Python 2
+        called[0] += 1
```
bstaletic, line 504: Why not just use `global`?

tests/test_pytypes.py
```diff
@@ -488,3 +488,45 @@ def test_builtin_functions():
...
+
+    obj = WeaklyReferenced()
+    assert getweakrefcount(obj) == 0
+    wr = m.weakref_from_handle(obj)  # noqa: F841
+    assert getweakrefcount(obj) == 1
+
+    obj = WeaklyReferenced()
+    assert getweakrefcount(obj) == 0
+    wr = m.weakref_from_handle_and_function(obj, callback)  # noqa: F841
```
bstaletic, line 513: Can't you just not assign the weak reference?

tests/test_pytypes.py
YannickJadoul: Because it's not `global`? It's `nonlocal`, I think (?), but that's not available in Python 2?

tests/test_pytypes.py
YannickJadoul: Well, it will be gone. Will the destructor still be called then? I don't think so. I think I/we want to check this callback being called on `del obj`, no?

include/pybind11/pytypes.h
YannickJadoul: @bstaletic and I discussed this privately, btw.

I don't think `weakref()` is that useful: it's constructing a `weakref` object with a `(PyObject *) NULL` inside. pybind11 is reasonably inconsistent about which types allow this and which ones don't. I'd say it's probably a discussion for another time on the sense/nonsense of default constructor of `object`-subclasses.

I could add either `PYBIND11_OBJECT_CVT_DEFAULT` or just add a single line `weakref() = default;` here, though?

tests/test_pytypes.py
bstaletic: I thought `global` meant "first" variable of `name` going from the global scope towards the current scope. Carry on.

tests/test_pytypes.py
YannickJadoul: This is in the interactive repl, though?

Have a look at this:

```
$ python3 - << EOF
> import weakref
> class c: pass
> o = c()
> weakref.ref(o, lambda x: print(1))
> print('bump')
> del o
> 
> EOF
bump
```
vs.
```
$ python3 - << EOF
> import weakref
> class c: pass
> o = c()
> wr = weakref.ref(o, lambda x: print(1))
> print('bump')
> del o
> EOF
bump
1
```

tests/test_pytypes.py
YannickJadoul: I wasn't completely sure either, but this test says I'm right (beware, this is ugly!):

```
>>> def f():
...     i = 0
...     def g():
...             global i
...             i += 1
...     g()
...     g()
...     return i
... 
>>> f()
  File "<stdin>", line 6, in f
  File "<stdin>", line 5, in g
NameError: name 'i' is not defined
>>> i = 0
>>> f()
0
>>> i
2
```


But wait, I can still pick something better!

YannickJadoul: Better this way? It's just as weird, but looks better and slightly more descriptive.

include/pybind11/pytypes.h
rwgk: `weakref() = default;` looks good to me, maybe with a comment
```
// Provided mainly for uniformity. Constructs a weakref object with a (PyObject *) NULL inside.
weakref() = default;
```
I almost feel the `PYBIND11_OBJECT_DEFAULT` macro isn't useful ("what default?") and getting rid of it and explicitly writing out `something() = default;` is more readable code ... but that's a question for another day, maybe.

rwgk:
Looks good to me. I'll globally test this (without the default constructor) to see if anyone is relying on it, mostly out of interest. Preserving the default constructor seems best irrespectively.

include/pybind11/pytypes.h
YannickJadoul: Agreed. This macros make things very confusing and inconsistent (some `py::object` subclasses are nullable, others aren't; some try to be safe when calling methods on them, others don't). I actually would like to doubt the validity of default constructors here in general. But maybe that's a challenge for another day.

tests/test_pytypes.py
```diff
@@ -541,3 +541,45 @@ def test_pass_bytes_or_unicode_to_string_types():
...
+
+    # Apparently, you cannot weakly reference an object()
+    class WeaklyReferenced(object):
+        pass
+
+    def callback(wr):
+        # No `nonlocal` in Python 2
+        callback.called += 1
+
+    callback.called = 0
```
rwgk, line 557: Resetting `callback.called` only once makes the tests below order dependent. If someone has a reason to shuffle the test, they may get surprised. I'd move this down do just above the the `weakref_from_*_and_function` calls.

But I think I'd go one step further, using `@pytest.mark.parametrize` to reduce copy-pasting, that would automatically avoid the order dependence.

rwgk (approved):
Google-global testing this as-is (without the default constructor) came back green. Might be nice to add something along the lines of "very unlikely to be useful but supplied for backward compatibility" in a comment when you put back the default constructor.

YannickJadoul:
> Might be nice to add something along the lines of "very unlikely to be useful but supplied for backward compatibility" in a comment when you put back the default constructor.

If we want to do that, I think we should go with an official deprecation, but I'm fine just being fully backward compatible and pushing the "let's make all `py::object` subclasses consistent" into the future.

tests/test_pytypes.py
YannickJadoul: Hmm, maybe. I tried something. I hope this is still readable enough to track down when one of the two tests fails, but it's shorter, at least.

YannickJadoul:
> you put back the default constructor.

I added `PYBIND11_OBJECT_CVT_DEFAULT` to make clear that this is part of the whole `PYBIND11_OBJECT*` macro zoo. This groups things if we ever were to e.g. deprecate the default contructors of `py::object` subclasses, in some hypothetical future.

bstaletic (approved):

YannickJadoul:
I think we went over all bits and pieces, then :-) Let's close another PR and issue.

Thanks, all!

PR #2833, opened 2021-01-29
Adding holder_caster `typename SFINAE = void` hooks to help work around the current lack of smart-pointer interoperability

rwgk:
The two new SFINAE hooks are provided to help work around the current lack of support for smart-pointer interoperability. Please consider it an implementation detail that may change in the future, as formal support for smart-pointer interoperability is added into pybind11.                                    

Changelog not needed.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1600,7 +1600,7 @@ protected:
 template <typename T>
 class type_caster<std::shared_ptr<T>> : public copyable_holder_caster<T, std::shared_ptr<T>> { };
 
-template <typename type, typename holder_type>
+template <typename type, typename holder_type, typename SFINAE = void>
 struct move_only_holder_caster {
     static_assert(std::is_base_of<type_caster_base<type>, type_caster<type>>::value,
             "Holder classes are only supported for custom types");
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1525,7 +1525,11 @@ struct holder_helper {
 };
 
 /// Type caster for holder types like std::shared_ptr, etc.
-template <typename type, typename holder_type>
+/// The SFINAE hook is provided to help work around the current lack of support
+/// for smart-pointer interoperability. Please consider it an implementation
+/// detail that may change in the future, as formal support for smart-pointer
+/// interoperability is added into pybind11.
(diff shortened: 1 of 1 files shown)


YannickJadoul:
For the (literal) record: @rwgk pointed me to these two use cases: https://github.com/pybind/pybind11_protobuf/blob/a451639e0ac6ca66f1ea1bdddb401e264ac27fc8/fast_cpp_proto_casters.h#L321 & https://github.com/pybind/pybind11_protobuf/blob/a451639e0ac6ca66f1ea1bdddb401e264ac27fc8/proto_utils.h#L38:

E.g.:

```cpp
// All protos use shared pointers as their holder, so automatically convert a
// unique_ptr to a shared pointer. Note that this requires a patch to the core
// pybind11 to add a SFINAE template parameter to move_only_holder_caster.
template <typename ProtoType, typename HolderType>
struct move_only_holder_caster<
    ProtoType, HolderType, std::enable_if_t<google::is_proto_v<ProtoType>>> {
  static handle cast(HolderType&& src, return_value_policy policy,
                     handle handle) {
    return copyable_holder_caster<ProtoType, std::shared_ptr<ProtoType>>::cast(
        std::shared_ptr(std::move(src)), policy, handle);
  }
  static constexpr auto name = type_caster_base<type>::name;
};
```

And yes, specializing `type_caster` should still be the only way, but pybind11 already uses the SFINAE of `type_caster` in the `PYBIND11_DECLARE_HOLDER_TYPE` macro, so any straightforward attack on this issue would result in an ambiguous specialization, I believe:

https://github.com/pybind/pybind11/blob/23c3edcf214086d9759e64e7b5ea5eb28e17a7d7/include/pybind11/cast.h#L1627-L1634

So once again, our interface is fuzzier than it seems. But the addition of this big fat warning should give us the freedom to blame users if they misuse this part of the internal API.

And I do mean that. Please don't blame pybind11 if you rely on this and things change.

YannickJadoul (approved):
I'm still not entirely happy, but it _does_ seem to be the simplest way of making these use cases work (and the change is totally innocuous to the good users that keep their hands off of pybind11's internal).

rwgk:
Thanks @YannickJadoul!

PR #2834, opened 2021-01-29
Test type_caster bare interface

rwgk:
Systematically exercises the `detail::type_caster<> interface` (which is a de-facto public interface). This is going a step in the direction of an integration test, to ensure multiple components of pybind11 work together correctly. It is also useful to show the `type_caster<>` interface virtually clutter-free.

This new test was copied from PR #2672, with small modifications.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -129,6 +129,7 @@ set(PYBIND11_TEST_FILES
     test_stl.cpp
     test_stl_binders.cpp
     test_tagbased_polymorphic.cpp
+    test_type_caster_bare_interface.cpp
     test_union.cpp
     test_virtual_functions.cpp)
 
--- a/tests/test_type_caster_bare_interface.cpp
+++ b/tests/test_type_caster_bare_interface.cpp
@@ -1,0 +1,211 @@
+#include "pybind11_tests.h"
+
+#include <memory>
+
+namespace pybind11_tests {
+namespace type_caster_bare_interface {
+
+struct mpty {};
+
+// clang-format off
+
+mpty        rtrn_mpty_valu() { mpty obj; return obj; }
+mpty&&      rtrn_mpty_rref() { static mpty obj; return std::move(obj); }
+mpty const& rtrn_mpty_cref() { static mpty obj; return obj; }
+mpty&       rtrn_mpty_mref() { static mpty obj; return obj; }
+mpty const* rtrn_mpty_cptr() { return new mpty; }
+mpty*       rtrn_mpty_mptr() { return new mpty; }
+
+const char* pass_mpty_valu(mpty)        { return "load_valu"; }
+const char* pass_mpty_rref(mpty&&)      { return "load_rref"; }
+const char* pass_mpty_cref(mpty const&) { return "load_cref"; }
+const char* pass_mpty_mref(mpty&)       { return "load_mref"; }
+const char* pass_mpty_cptr(mpty const*) { return "load_cptr"; }
+const char* pass_mpty_mptr(mpty*)       { return "load_mptr"; }
+
+std::shared_ptr<mpty>       rtrn_mpty_shmp() { return std::shared_ptr<mpty      >(new mpty); }
+std::shared_ptr<mpty const> rtrn_mpty_shcp() { return std::shared_ptr<mpty const>(new mpty); }
+
+const char* pass_mpty_shmp(std::shared_ptr<mpty>)       { return "load_shmp"; }
+const char* pass_mpty_shcp(std::shared_ptr<mpty const>) { return "load_shcp"; }
+
+std::unique_ptr<mpty>       rtrn_mpty_uqmp() { return std::unique_ptr<mpty      >(new mpty); }
+std::unique_ptr<mpty const> rtrn_mpty_uqcp() { return std::unique_ptr<mpty const>(new mpty); }
+
+const char* pass_mpty_uqmp(std::unique_ptr<mpty>)       { return "load_uqmp"; }
+const char* pass_mpty_uqcp(std::unique_ptr<mpty const>) { return "load_uqcp"; }
+
+// clang-format on
+
+} // namespace type_caster_bare_interface
+} // namespace pybind11_tests
+
+namespace pybind11 {
+namespace detail {
+
+using namespace pybind11_tests::type_caster_bare_interface;
+
+template <>
+struct type_caster<mpty> {
+    static constexpr auto name = _<mpty>();
+
+    // static handle cast(mpty, ...)
+    // is redundant (leads to ambiguous overloads).
+
+    static handle cast(mpty && /*src*/, return_value_policy /*policy*/, handle /*parent*/) {
+        return str("cast_rref").release();
+    }
+
+    static handle cast(mpty const & /*src*/, return_value_policy /*policy*/, handle /*parent*/) {
+        return str("cast_cref").release();
+    }
+
+    static handle cast(mpty & /*src*/, return_value_policy /*policy*/, handle /*parent*/) {
+        return str("cast_mref").release();
+    }
+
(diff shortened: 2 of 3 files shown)


tests/test_type_caster_bare_interface.cpp
```diff
@@ -0,0 +1,215 @@
...
+// correctly. It is also useful to show the type_caster<> interface virtually clutter-free.
+
+#include "pybind11_tests.h"
+
+#include <memory>
+
+namespace pybind11_tests {
+namespace type_caster_bare_interface {
+
+struct mpty {};
```
YannickJadoul, line 12: I have a hard time reading everything since 1) I have no idea how to pronounce/read this, 2) my brain's expecting capitals for "custom"/user-side classes in the test. Do you mind just turning (or I can also do it, if you don't have time) this into `Foo` or `CustomClass` or `Value` or so?

tests/test_type_caster_bare_interface.cpp
```diff
@@ -0,0 +1,215 @@
...
+std::shared_ptr<mpty>       rtrn_shmp() { return std::shared_ptr<mpty      >(new mpty); }
+std::shared_ptr<mpty const> rtrn_shcp() { return std::shared_ptr<mpty const>(new mpty); }
+
+const char* pass_shmp(std::shared_ptr<mpty>)       { return "load_shmp"; }
+const char* pass_shcp(std::shared_ptr<mpty const>) { return "load_shcp"; }
+
+std::unique_ptr<mpty>       rtrn_uqmp() { return std::unique_ptr<mpty      >(new mpty); }
+std::unique_ptr<mpty const> rtrn_uqcp() { return std::unique_ptr<mpty const>(new mpty); }
+
+const char* pass_uqmp(std::unique_ptr<mpty>)       { return "load_uqmp"; }
```
YannickJadoul, line 39: Uuumpf, this takes me 2 second per function to parse to know what it's saying. Do you feel strongly about keeping 4 letters? Personally, I find `pass_unique_ptr`, `pass_const_ref`, `pass_rvalue_ref`, `return_...`, ... easier to directly read, even though things then don't nicely line out anymore.

But your effort on writing tests, so your final say.

tests/test_type_caster_bare_interface.cpp
```diff
@@ -0,0 +1,215 @@
...
+        mpty const *,
+        conditional_t<
+            std::is_same<remove_reference_t<T_>, mpty *>::value,
+            mpty *,
+            conditional_t<
+                std::is_same<T_, mpty const &>::value,
+                mpty const &,
+                conditional_t<std::is_same<T_, mpty &>::value,
+                              mpty &,
+                              conditional_t<std::is_same<T_, mpty &&>::value, mpty &&, mpty>>>>>;
```
YannickJadoul, line 93: What about going `else if` style on this, to avoid indentation? (similar to [test_builtin_casters.cpp](https://github.com/pybind/pybind11/blob/0432ae7c5222ecb5d6e81babf996d3777bd94bf3/tests/test_builtin_casters.cpp#L42))

```suggestion
    using cast_op_type =
        conditional_t<std::is_same<remove_reference_t<T_>, mpty const *>::value,
            mpty const *,
        conditional_t<std::is_same<remove_reference_t<T_>, mpty *>::value,
            mpty *,
        conditional_t<std::is_same<T_, mpty const &>::value,
            mpty const &,
        conditional_t<std::is_same<T_, mpty &>::value,
            mpty &,
        conditional_t<std::is_same<T_, mpty &&>::value,
            mpty &&,
            /* else */ mpty>>>>>;
```

tests/CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ set(PYBIND11_TEST_FILES
     test_stl.cpp
     test_stl_binders.cpp
     test_tagbased_polymorphic.cpp
+    test_type_caster_bare_interface.cpp
```
YannickJadoul, line 132: Final nitpick: `test_type_caster_interface` also works, if we want to have this file show off the interface?

YannickJadoul (commented):
Mostly some style nitpicking, in order to be consistent with the style in other pybind11 tests, and be more intuitively readable. Apart from that, this seems useful to have tested, yes, so thanks a lot for adding this long overdue test!

One content-wise thing: this seems to be a superset of `type_caster<ConstRefCasted>` in `test_const_ref_caster`, except for the `reference_wrapper` part. Should we remove that test, and just copy the `return_reference_wrapper`/`pass_reference_wrapper` function?

Bonus nitpick: pybind11 (reasonably) consistently uses west-side `const`,  `const Foo &bar` style. I realize there's a whole discussion on which one is best and I won't take sides (though I happen to agree with pybind11), but mixing styles might be even worse?

rwgk:
Converting this to a Draft PR for now. While developing PR #2672 I find myself tweaking the systematic naming scheme. I will update this PR with matching names later.

tests/test_type_caster_bare_interface.py
```diff
@@ -0,0 +1,41 @@
...
+    assert m.rtrn_mptr() == "cast_mptr"
+
+
+def test_load():
+    assert m.pass_valu(None) == "load_valu"
+    assert m.pass_rref(None) == "load_rref"
+    assert m.pass_cref(None) == "load_cref"
+    assert m.pass_mref(None) == "load_mref"
+    assert m.pass_cptr(None) == "load_cptr"
+    assert m.pass_mptr(None) == "load_mptr"
```
rhaschke, line 21: What is actually tested here? This is just about calling different functions and comparing their return values, isn't it?

tests/test_type_caster_bare_interface.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_cast_shared_ptr():
+    assert m.rtrn_shmp() == "cast_shmp"
+    assert m.rtrn_shcp() == "cast_shcp"
+
+
+def test_load_shared_ptr():
+    assert m.pass_shmp(None) == "load_shmp"
+    assert m.pass_shcp(None) == "load_shcp"
```
rhaschke, line 31: Same here.

tests/test_type_caster_bare_interface.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_cast_unique_ptr():
+    assert m.rtrn_uqmp() == "cast_uqmp"
+    assert m.rtrn_uqcp() == "cast_uqcp"
+
+
+def test_load_unique_ptr():
+    assert m.pass_uqmp(None) == "load_uqmp"
+    assert m.pass_uqcp(None) == "load_uqcp"
```
rhaschke, line 41: and here.

tests/test_type_caster_bare_interface.cpp
rhaschke, line 93: Isn't all this just returning the same type as detected?

rhaschke (commented):
I think, testing that `rtrn_*()` calls the correct casting operation is fine.
However, changing the function signature of `pass_*()` will still call the "old" functions. 
I don't see a value in testing this. Rather, it should be tested whether the values are correctly copied/moved when passing from python to C++.

rwgk:
Hi Robert, I reworked this PR quite a bit, to be less the stepping stone it was for me back in December, and a more meaningful test / demo (and be more similar to what's now test_class_sh_basic on the smart_holder branch). Could you please take another look? I'm now tracking `CpCtor` and `MvCtor`. I also removed all the smart-pointer code, because it's not relevant to the copy/move logic we're concerned about here.

The valgrind CI failure seems unrelated, btw; something todo with pypocketfft.

rwgk:
Closing this stale PR to reduce distractions.

Files archived here as-is:

https://github.com/rwgk/rwgk_tbx/commit/faec6cbbdfdf9f253911838e29f05585a79867ef

PR #2838, opened 2021-01-31
Make sure all warnings in pytest get turned into errors

YannickJadoul:
## Description

Follow-up on #2837.

There's a weird thing going on where `filterwarnings=error` starts ignoring the `DeprecationWarning` from https://github.com/python/cpython/blob/3c8d6934436e20163be802f5239c5b4e4925eeec/Objects/longobject.c#L226.

I have no clue why. Apart from that, the `-Wa` flag seemed to take precedence, so no warning was actually an error.

## Suggested changelog entry:

None, just some tests.

### DIFF
--- a/tests/pytest.ini
+++ b/tests/pytest.ini
@@ -7,11 +7,11 @@ addopts =
     -rs
     # capture only Python print and C++ py::print, but not C output (low-level Python errors)
     --capture=sys
-    # enable all warnings
-    -Wa
 filterwarnings =
     # make warnings into errors but ignore certain third-party extension issues
     error
+    # somehow, some DeprecationWarnings do not get turned into errors
+    always::DeprecationWarning
     # importing scipy submodules on some version of Python
     ignore::ImportWarning
     # bogus numpy ABI warning (see numpy/#432)
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -434,8 +434,7 @@ TEST_SUBMODULE(class_, m) {
     struct SamePointer {};
     static SamePointer samePointer;
     py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer")
-        .def(py::init([]() { return &samePointer; }))
-        .def("__del__", [](SamePointer&) { py::print("__del__ called"); });
+        .def(py::init([]() { return &samePointer; }));
 
     struct Empty {};
     py::class_<Empty>(m, "Empty")
--- a/tests/test_builtin_casters.py
+++ b/tests/test_builtin_casters.py
@@ -299,7 +299,9 @@ def test_int_convert():
     assert convert(7) == 7
     assert noconvert(7) == 7
     cant_convert(3.14159)
-    assert convert(Int()) == 42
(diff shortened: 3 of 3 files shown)


tests/pytest.ini
```diff
@@ -7,11 +7,11 @@ addopts =
     -rs
     # capture only Python print and C++ py::print, but not C output (low-level Python errors)
     --capture=sys
-    # enable all warnings
-    -Wa
 filterwarnings =
     # make warnings into errors but ignore certain third-party extension issues
     error
+    # somehow, some DeprecationWarnings do not get turned into errors
+    always::DeprecationWarning
```
bstaletic, line 14: Why, oh why, does python try so hard to hide deprecation warnings?!

bstaletic (commented):
Reasons for CI being red:

- `conftest.py:19:RuntimeWarning: a __del__ method added to an existing type will not be called` on pypy3
- `test_int_convert` did not warn on the valgrind run.

tests/pytest.ini
YannickJadoul: I have absolutely no clue. It gets funnier, btw: I added `warnings.warn("abc", DeprecationWarning)` and that _did_ get turned into an error!

But I didn't feel like wasting even more time, and digging further than https://github.com/python/cpython/blob/3c8d6934436e20163be802f5239c5b4e4925eeec/Objects/longobject.c#L226

YannickJadoul:
> * `test_int_convert` did not warn on 3.10-dev

Because

https://github.com/python/cpython/blob/3c8d6934436e20163be802f5239c5b4e4925eeec/Objects/longobject.c#L514

got changed into

https://github.com/python/cpython/blob/a1e9a1e120a11c563e166c15721169184c802f8b/Objects/longobject.c#L390

So the deprecation period ended, but ... our tests don't fail because we still fall back to `PyNumber_Long` anyway? Reason the more we need to not loose track of #2802, it seems.

tests/test_class.cpp
```diff
@@ -434,8 +434,7 @@ TEST_SUBMODULE(class_, m) {
     struct SamePointer {};
     static SamePointer samePointer;
     py::class_<SamePointer, std::unique_ptr<SamePointer, py::nodelete>>(m, "SamePointer")
-        .def(py::init([]() { return &samePointer; }))
-        .def("__del__", [](SamePointer&) { py::print("__del__ called"); });
```
YannickJadoul, line 438: If this looks like a "sweep it under the carpet" kind of thing, you wouldn't be wrong. But we were never really using the fact that `__del__` prints something anyway (possibly because it already didn't work on PyPy?).

tests/pytest.ini
henryiii: FYI, FutureWarning _is_ (as clarified by [PEP 565](https://www.python.org/dev/peps/pep-0565/#additional-use-case-for-futurewarning) the "user facing" deprecation warning.

bstaletic (approved):

YannickJadoul:
Just our own tests. This should be safe enough to merge :-) Thanks, @henryiii & @bstaletic!

Issue #2757, opened 2020-12-30
Should investigate usage of `py::wraper<>` trampoline shim with GC garbage collection


EricCousineau-TRI:
## Motivation

Towards #1333 (as part of the larger umbrella of #2646), this is meant to be my recording of investigating a `py::wrapper<>` shim trampoline with "resurrection" via the Python garbage collection.

## References

- [CPython Object Model - `object.__del__`](https://docs.python.org/3.9/reference/datamodel.html#object.__del__)
- [CPython Developer Guide - Additional Resources](https://devguide.python.org/#additional-resources)
  - [Garbage Collector](https://devguide.python.org/garbage_collector/)
- CPython issue tracker
  - [Issue 40240](https://bugs.python.org/issue40240) - Expose (read-only) GC info
  - [Issue 32377](https://bugs.python.org/issue32377) - CPython 3.x GC ressurection got more picky

## Proof of Concept

I have a working proof-of-concept that is used for in a fork of pybind11 - see [`RobotLocomtion/pybind11:README_DRAKE.md`](https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/README_DRAKE.md).

The primary points to see are:

- Resurrection via GC: [`py::wrapper<>`](https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/include/pybind11/pytypes.h#L1517-L1614)
- Additional mechanisms for holder ownership transfer to\/from C++: [`py::class_<> methods`](https://github.com/RobotLocomotion/pybind11/blob/58a368ea8e89638e01a7385cad9ce4345d70003d/include/pybind11/pybind11.h#L1323-L1473) (note: some of this is coupled to other issues mentioned in 
- Tests: [`test_ownership_transfer.py`](https://github.com/RobotLocomotion/pybind11/blob/drake/tests/test_ownership_transfer.py), [`test_ownership_transfer.cpp`](https://github.com/RobotLocomotion/pybind11/blob/drake/tests/test_ownership_transfer.cpp)

## Drawbacks

- This approach is a bit messy, as it's from me "just making" things work for our library ([Drake](https://drake.mit.edu)).
- As shown above, CPython 3.8 (and possibly 3.7.5+), we are unable to resurrect an object more than once.

## Additional Notes

This may not be the best mechanism to prevent slicing. Willi be listing other mechanisms (i.e. PRs / suggestions from others).

Issue #1937, opened 2019-09-28
Possible memory issues when using `shared_ptr` as holder type


lyskov:
Dear Pybind11 developers, i would like to report a strange case which on a surface looks like a situation when C++ fail to retain ownership of classes defined in Python and derived from C++ base classes. 

Below is a minimal example to reproduce this. Note that Python script have `--bug` command line argument which allow to trigger a bug. And if option is not supplied then code work as expected without any errors.

In short: issues seems to arise when derived class is created in C++ by a 'clone' method. It looks like after creation 'Python' portion of `Snake` class will get immediately deleted unless it explicitly stored in one of Python variables.

When `--bug` is supplied execution terminated with error message:
```
Snake.clone!
Traceback (most recent call last):
  File "inheritance-clone.py", line 34, in <module>
    example.clone_and_call_go(a)
RuntimeError: Tried to call pure virtual function "Animal::go"
```

```C++
#include <pybind11/pybind11.h>


class Animal {
public:
    virtual ~Animal() { }
    virtual std::string go(int n_times) = 0;

    virtual std::shared_ptr<Animal> clone() = 0;
};


std::string clone_and_call_go(std::shared_ptr<Animal> animal) {
    auto new_animal = animal->clone();
    return new_animal->go(3);
}


class PyAnimal : public Animal {
public:
    /* Inherit the constructors */
    using Animal::Animal;

    /* Trampoline (need one for each virtual function) */
    std::string go(int n_times) override {
        PYBIND11_OVERLOAD_PURE(
            std::string, /* Return type */
            Animal,      /* Parent class */
            go,          /* Name of function in C++ (must match Python name) */
            n_times      /* Argument(s) */
        );
    }

	std::shared_ptr<Animal> clone() override {
        PYBIND11_OVERLOAD_PURE(
            std::shared_ptr<Animal>,
            Animal,
            clone
        );
    }

};


namespace py = pybind11;

PYBIND11_MODULE(example, m) {

	py::class_<Animal, std::shared_ptr<Animal>, PyAnimal>(m, "Animal")
        .def(py::init<>())
        .def("go",    &Animal::go)
        .def("clone", &Animal::clone)
	;

	m.def("clone_and_call_go", &clone_and_call_go);

}
```

```Python
import argparse

import example

class Snake(example.Animal):
    def __init__(self, bug):
        #super().__init__()
        example.Animal.__init__(self)
        self.copies_ = []
        self.bug_ = bug

    def go(self, n_times):
[...]

Issue #1902, opened 2019-09-01
How to let C++ handle python instances of pybind11 derived types


mmodenesi:
## Issue description

My use case is as follows:

Main project written in C++ lets users extend the codebase by subclassing a virtual `CppClass` and providing custom implementation of a virtual `method`. After registering a `user_defined_factory_function`, the application starts creating objects using a pointer to the polymorphic `CppClass` type, calling their `method` and finally deleting those objects in a way that is completely transparent to users. The whole point of the application is to load the user implementation of `CppClass` derived class.

I want to use pybind11 to allow users to subclass `CppClass` and override this `method` in python.

Supose the intepreter is initialized and finalized properly.

Supose the user factory function takes care of finding the python module and creating instances of the Python class deriving from `CppClass`. Then casts the `py::object` and reuturns a `CppClass` pointer.

The problem that I find is that the python objects are deleted as soon as the factory function returns. I need C++ to take ownership of the object. I repeat: the application takes care of deleting those objects when time comes.

I've read the documentation thoroughly. I've investigated all the questions tagged with `pybind11` on SO. I've searched through the issues and I know this overlaps with:

- #1389 (this is probably the closest)
- #1774 and all the issues mentioned in [this comment](https://github.com/pybind/pybind11/issues/1774#issuecomment-487321016)

Why do I add another issue, then?

I've explored the solutions proposed on other issues and couldn't fully understand them or tweak them to meet my needs. I am more of a python guy than a C++ guy, I'm trying to improve, though :smiley: . In such vein, I think it would be useful to add a section to the documentation exploring the posibilities of using python objects on C++ in more involved ways, and explaining python vs. C++ objects lifecycle.

I think my issue adds value by providing a clear description of the use case as well as a complete example, exploring several posibilities.

## Reproducible example code

Working [example](https://gist.github.com/mmodenesi/021eb1ecf74257557b68311c7f130011), just `make` and then `./program`.

This is what I am pursuing:

```
[Python] __init__ called
[C++   ] end of user_defined_function
[C++   ] calling method on cpp pointer
[Python] method called
[Python] __del__ called
[C++   ]: destructor called
```

This is what I get:

```
[Python] __init__ called
[C++   ] end of user_defined_function
[Python] __del__ called
[C++   ]: destructor called
[C++   ] calling method on cpp pointer
Segmentation fault
```

If I manually inc_ref the underlying python object:

```
[Python] __init__ called
# inc_ref object here
[C++   ] end of user_defined_function
[...]

PR #2839, opened 2021-02-01
Attach python lifetime to shared_ptr passed to C++

### LINKED ISSUE #1333: [BUG] Problem when creating derived Python objects from C++ (inheritance slicing)

virtuald:
## Description

Modifies the base type caster to ensure that python instances don't die when passed to C++. 

The shared_ptr is always given to C++ code, so we construct a new shared_ptr that is given a custom deleter. The custom deleter increments the python reference count to bind the python instance lifetime with the lifetime of the shared_ptr. **This behavior is only applied to instances that have an associated trampoline class**

This enables things like passing the last python reference of a subclass to a C++ function without the python reference dying -- which means custom virtual function overrides will still work.

Reference cycles will cause a leak, but this is a limitation of shared_ptr

Thanks to @YannickJadoul for the inspiration putting me on this path. Fixes #1333 and #1120, #1145, #1389, #1552, #1546, #1566, #1774, #1902, #1937 

Other inspiration from Boost::Python. Quoting from @rwgk at https://github.com/pybind/pybind11/issues/2646:

> When passing to shared_ptr arguments, Boost.Python always constructs a new shared_ptr, using a properly cast raw pointer, and a custom deleter keeping the PyObject owning the raw pointer alive:
https://github.com/boostorg/python/blob/5e4f6278e01f018c32cebc4ffc3b75a743fb1042/include/boost/python/converter/shared_ptr_from_python.hpp#L52
Note that this code uses the shared_ptr aliasing constructor.
In this way shared_ptr upcasts and downcasts work correctly, although the argument shared_ptr::use_count results tend to surprise users. Note that this mechanism is independent of the HeldType, in other words, no matter what the HeldType, it is possible to pass to shared_ptr arguments.

Downsides:

* Reference cycles are possible as a result, but shared_ptr is already susceptible to this in C++. #2757 discusses this topic a bit
* Might have a small negative effect on performance?
* Probably things I haven't thought of yet

Potential improvements:

* ~Right now this does this for all shared_ptr, might make sense to only do this for things that are python instances of python classes?~ Now only does this for things that have trampolines
* Should we document how to do this for other holders? Not sure if the value_and_holder class is something we want to expose...

## Suggested changelog entry:

```rst
* Objects that are wrapped with a shared_ptr holder and a trampoline class (or alias class) are now kept alive until all shared_ptr references are deleted
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -664,6 +664,14 @@ struct holder_retriever<std::shared_ptr<T>> {
     };
 
     static auto get_derivative_holder(const value_and_holder &v_h) -> std::shared_ptr<T> {
+        // If there's no trampoline class, nothing special needed
+        if (!v_h.inst->has_alias) {
+            return v_h.template holder<std::shared_ptr<T>>();
+        }
+
+        // If there's a trampoline class, ensure the python side of the object doesn't
+        // die until the C++ portion also dies
+        //
         // The shared_ptr is always given to C++ code, so construct a new shared_ptr
         // that is given a custom deleter. The custom deleter increments the python
         // reference count to bind the python instance lifetime with the lifetime
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#include "gil.h"
 #include "pytypes.h"
 #include "detail/common.h"
 #include "detail/descr.h"
@@ -641,6 +642,42 @@ struct holder_helper {
     static auto get(const T &p) -> decltype(p.get()) { return p.get(); }
 };
 
+/// Another helper class for holders that helps construct derivative holders from
+/// the original holder
+template <typename T>
+struct holder_retriever {
+    static auto get_derivative_holder(const value_and_holder &v_h) -> decltype(v_h.template holder<T>()) {
+        return v_h.template holder<T>();
+    }
+};
+
+template <typename T>
+struct holder_retriever<std::shared_ptr<T>> {
+    struct shared_ptr_deleter {
+        // Note: deleter destructor fails on MSVC 2015 and GCC 4.8, so we manually
+        // call dec_ref here instead
+        handle ref;
+        void operator()(T *) {
+            gil_scoped_acquire gil;
+            ref.dec_ref();
+        }
+    };
+
+    static auto get_derivative_holder(const value_and_holder &v_h) -> std::shared_ptr<T> {
+        // The shared_ptr is always given to C++ code, so construct a new shared_ptr
+        // that is given a custom deleter. The custom deleter increments the python
+        // reference count to bind the python instance lifetime with the lifetime
+        // of the shared_ptr.
+        //
+        // This enables things like passing the last python reference of a subclass to a
+        // C++ function without the python reference dying.
+        //
+        // Reference cycles will cause a leak, but this is a limitation of shared_ptr
+        return std::shared_ptr<T>((T*)v_h.value_ptr(),
+            shared_ptr_deleter{handle((PyObject*)v_h.inst).inc_ref()});
+    }
+};
+
 /// Type caster for holder types like std::shared_ptr, etc.
 /// The SFINAE hook is provided to help work around the current lack of support
 /// for smart-pointer interoperability. Please consider it an implementation
@@ -683,7 +720,7 @@ protected:
     bool load_value(value_and_holder &&v_h) {
         if (v_h.holder_constructed()) {
             value = v_h.value_ptr();
-            holder = v_h.template holder<holder_type>();
+            holder = holder_retriever<holder_type>::get_derivative_holder(v_h);
             return true;
         }
         throw cast_error("Unable to cast from non-held to held instance (T& to Holder<T>) "
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -665,7 +665,7 @@ struct holder_retriever<std::shared_ptr<T>> {
 
     static auto get_derivative_holder(const value_and_holder &v_h) -> std::shared_ptr<T> {
         // If there's no trampoline class, nothing special needed
-        if (!v_h.inst->has_alias) {
+        if (!v_h.inst->is_alias) {
             return v_h.template holder<std::shared_ptr<T>>();
         }
 
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -178,6 +178,8 @@ extern "C" inline PyObject *pybind11_meta_call(PyObject *type, PyObject *args, P
 
     // This must be a pybind11 instance
     auto instance = reinterpret_cast<detail::instance *>(self);
+    // Mark this instance as an instance of an alias class
+    instance->is_alias = instance->has_alias;
 
     // Ensure that the base __init__ function(s) were called
     for (const auto &vh : values_and_holders(instance)) {
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -580,8 +580,10 @@ struct instance {
     bool simple_instance_registered : 1;
     /// If true, get_internals().patients has an entry for this object
     bool has_patients : 1;
-    /// If true, created with an associated alias class (set via `init_instance`)
+    /// If true, the type of this instance has an associated alias class (set via `init_instance`)
     bool has_alias : 1;
+    /// If true, this instance has an associated alias class and was constructed by Python
+    bool is_alias : 1;
 
     /// Initializes all of the above type/values/holders data (but not the instance values themselves)
     void allocate_layout();
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -580,6 +580,8 @@ struct instance {
     bool simple_instance_registered : 1;
     /// If true, get_internals().patients has an entry for this object
     bool has_patients : 1;
+    /// If true, created with an associated alias class (set via `init_instance`)
+    bool has_alias : 1;
 
     /// Initializes all of the above type/values/holders data (but not the instance values themselves)
     void allocate_layout();
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1630,6 +1630,7 @@ private:
     /// optional pointer to an existing holder to use; if not specified and the instance is
     /// `.owned`, a new holder will be constructed to manage the value pointer.
     static void init_instance(detail::instance *inst, const void *holder_ptr) {
+        inst->has_alias = has_alias;
         auto v_h = inst->get_value_and_holder(detail::get_type_info(typeid(type)));
         if (!v_h.instance_registered()) {
             register_instance(inst, v_h.value_ptr(), v_h.type);
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -132,6 +132,7 @@ set(PYBIND11_TEST_FILES
     test_stl_binders.cpp
     test_tagbased_polymorphic.cpp
     test_thread.cpp
+    test_trampoline_shared_ptr_cpp_arg.cpp
     test_union.cpp
     test_virtual_functions.cpp)
 
--- a/tests/test_trampoline_shared_ptr_cpp_arg.cpp
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.cpp
@@ -52,4 +52,13 @@ TEST_SUBMODULE(trampoline_shared_ptr_cpp_arg, m) {
         .def("set_object", &SpBaseTester::set_object)
         .def("is_base_used", &SpBaseTester::is_base_used)
         .def_readwrite("obj", &SpBaseTester::m_obj);
+
+    // For testing that a C++ class without an alias does not retain the python
+    // portion of the object
+
+    py::class_<SpGoAway, std::shared_ptr<SpGoAway>>(m, "SpGoAway").def(py::init<>());
+
+    py::class_<SpGoAwayTester>(m, "SpGoAwayTester")
+        .def(py::init<>())
+        .def_readwrite("obj", &SpGoAwayTester::m_obj);
 }
--- a/tests/test_trampoline_shared_ptr_cpp_arg.cpp
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.cpp
@@ -1,0 +1,55 @@
+// Copyright (c) 2021 The Pybind Development Team.
+// All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include "pybind11_tests.h"
+
+namespace {
+
+// For testing whether a python subclass of a C++ object dies when the
+// last python reference is lost
+struct SpBase {
+    // returns true if the base virtual function is called
+    virtual bool is_base_used() { return true; }
+
+    SpBase()               = default;
+    SpBase(const SpBase &) = delete;
+    virtual ~SpBase()      = default;
+};
+
+struct PySpBase : SpBase {
+    bool is_base_used() override { PYBIND11_OVERRIDE(bool, SpBase, is_base_used); }
+};
+
+struct SpBaseTester {
+    std::shared_ptr<SpBase> get_object() const { return m_obj; }
+    void set_object(std::shared_ptr<SpBase> obj) { m_obj = std::move(obj); }
+    bool is_base_used() { return m_obj->is_base_used(); }
+    std::shared_ptr<SpBase> m_obj;
+};
+
+// For testing that a C++ class without an alias does not retain the python
+// portion of the object
+struct SpGoAway {};
+
+struct SpGoAwayTester {
+    std::shared_ptr<SpGoAway> m_obj;
+};
+
+} // namespace
+
+TEST_SUBMODULE(trampoline_shared_ptr_cpp_arg, m) {
+    // For testing whether a python subclass of a C++ object dies when the
+    // last python reference is lost
+
+    py::class_<SpBase, std::shared_ptr<SpBase>, PySpBase>(m, "SpBase")
+        .def(py::init<>())
+        .def("is_base_used", &SpBase::is_base_used);
+
+    py::class_<SpBaseTester>(m, "SpBaseTester")
+        .def(py::init<>())
+        .def("get_object", &SpBaseTester::get_object)
+        .def("set_object", &SpBaseTester::set_object)
+        .def("is_base_used", &SpBaseTester::is_base_used)
+        .def_readwrite("obj", &SpBaseTester::m_obj);
+}
--- a/tests/test_trampoline_shared_ptr_cpp_arg.cpp
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.cpp
@@ -12,6 +12,12 @@ struct SpBase {
     // returns true if the base virtual function is called
     virtual bool is_base_used() { return true; }
 
+    // returns true if there's an associated python instance
+    bool has_python_instance() {
+        auto tinfo = py::detail::get_type_info(typeid(SpBase));
+        return (bool)py::detail::get_object_handle(this, tinfo);
+    }
+
     SpBase()               = default;
     SpBase(const SpBase &) = delete;
     virtual ~SpBase()      = default;
@@ -25,6 +31,11 @@ struct SpBaseTester {
     std::shared_ptr<SpBase> get_object() const { return m_obj; }
     void set_object(std::shared_ptr<SpBase> obj) { m_obj = std::move(obj); }
     bool is_base_used() { return m_obj->is_base_used(); }
+    bool has_instance() { return (bool)m_obj; }
+    bool has_python_instance() { return m_obj && m_obj->has_python_instance(); }
+    void set_nonpython_instance() {
+        m_obj = std::make_shared<SpBase>();
+    }
     std::shared_ptr<SpBase> m_obj;
 };
 
@@ -44,13 +55,17 @@ TEST_SUBMODULE(trampoline_shared_ptr_cpp_arg, m) {
 
     py::class_<SpBase, std::shared_ptr<SpBase>, PySpBase>(m, "SpBase")
         .def(py::init<>())
-        .def("is_base_used", &SpBase::is_base_used);
+        .def("is_base_used", &SpBase::is_base_used)
+        .def("has_python_instance", &SpBase::has_python_instance);
 
     py::class_<SpBaseTester>(m, "SpBaseTester")
         .def(py::init<>())
         .def("get_object", &SpBaseTester::get_object)
         .def("set_object", &SpBaseTester::set_object)
         .def("is_base_used", &SpBaseTester::is_base_used)
+        .def("has_instance", &SpBaseTester::has_instance)
+        .def("has_python_instance", &SpBaseTester::has_python_instance)
+        .def("set_nonpython_instance", &SpBaseTester::set_nonpython_instance)
         .def_readwrite("obj", &SpBaseTester::m_obj);
 
     // For testing that a C++ class without an alias does not retain the python
--- a/tests/test_trampoline_shared_ptr_cpp_arg.py
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.py
@@ -41,7 +41,9 @@ def test_shared_ptr_cpp_prop():
 
     # python reference is still around since C++ has it now
     assert tester.is_base_used() is False
+    assert tester.has_python_instance() is True
     assert tester.obj.is_base_used() is False
+    assert tester.obj.has_python_instance() is True
 
 
 def test_shared_ptr_arg_identity():
@@ -59,11 +61,54 @@ def test_shared_ptr_arg_identity():
     # python reference is still around since C++ has it
     assert objref() is not None
     assert tester.get_object() is objref()
+    assert tester.has_python_instance() is True
 
     # python reference disappears once the C++ object releases it
     tester.set_object(None)
     pytest.gc_collect()
     assert objref() is None
+
+
+def test_shared_ptr_alias_nonpython():
+    tester = m.SpBaseTester()
+
+    # C++ creates the object, a python instance shouldn't exist
+    tester.set_nonpython_instance()
+    assert tester.is_base_used() is True
+    assert tester.has_instance() is True
+    assert tester.has_python_instance() is False
+
+    # Now a python instance exists
+    cobj = tester.get_object()
+    assert cobj.has_python_instance()
+    assert tester.has_instance() is True
+    assert tester.has_python_instance() is True
+
+    # Now it's gone
+    del cobj
+    pytest.gc_collect()
+    assert tester.has_instance() is True
+    assert tester.has_python_instance() is False
+
+    # When we pass it as an arg to a new tester the python instance should
+    # disappear because it wasn't created with an alias
+    new_tester = m.SpBaseTester()
+
+    cobj = tester.get_object()
+    assert cobj.has_python_instance()
+
+    new_tester.set_object(cobj)
+    assert tester.has_python_instance() is True
+    assert new_tester.has_python_instance() is True
+
+    del cobj
+    pytest.gc_collect()
+
+    # Gone!
+    assert tester.has_instance() is True
+    assert tester.has_python_instance() is False
+    assert new_tester.has_instance() is True
+    assert new_tester.has_python_instance() is False
 
 
 def test_shared_ptr_goaway():
--- a/tests/test_trampoline_shared_ptr_cpp_arg.py
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.py
@@ -64,3 +64,23 @@ def test_shared_ptr_arg_identity():
     tester.set_object(None)
     pytest.gc_collect()
     assert objref() is None
+
+
+def test_shared_ptr_goaway():
+    import weakref
+
+    tester = m.SpGoAwayTester()
+
+    obj = m.SpGoAway()
+    objref = weakref.ref(obj)
+
+    assert tester.obj is None
+
+    tester.obj = obj
+    del obj
+    pytest.gc_collect()
+
+    # python reference is no longer around
+    assert objref() is None
+    # C++ reference is still around
+    assert tester.obj is not None
--- a/tests/test_trampoline_shared_ptr_cpp_arg.py
+++ b/tests/test_trampoline_shared_ptr_cpp_arg.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+import pybind11_tests.trampoline_shared_ptr_cpp_arg as m
+
+
+def test_shared_ptr_cpp_arg():
+    import weakref
+
+    class PyChild(m.SpBase):
+        def is_base_used(self):
+            return False
+
+    tester = m.SpBaseTester()
+
+    obj = PyChild()
+    objref = weakref.ref(obj)
+
+    # Pass the last python reference to the C++ function
+    tester.set_object(obj)
+    del obj
+    pytest.gc_collect()
+
+    # python reference is still around since C++ has it now
+    assert objref() is not None
+    assert tester.is_base_used() is False
+    assert tester.obj.is_base_used() is False
+    assert tester.get_object() is objref()
+
+
+def test_shared_ptr_cpp_prop():
+    class PyChild(m.SpBase):
+        def is_base_used(self):
+            return False
+
+    tester = m.SpBaseTester()
+
+    # Set the last python reference as a property of the C++ object
+    tester.obj = PyChild()
+    pytest.gc_collect()
+
+    # python reference is still around since C++ has it now
+    assert tester.is_base_used() is False
+    assert tester.obj.is_base_used() is False
+
+
+def test_shared_ptr_arg_identity():
+    import weakref
+
+    tester = m.SpBaseTester()
+
+    obj = m.SpBase()
+    objref = weakref.ref(obj)
+
+    tester.set_object(obj)
+    del obj
+    pytest.gc_collect()
+
+    # python reference is still around since C++ has it
+    assert objref() is not None
+    assert tester.get_object() is objref()
+
+    # python reference disappears once the C++ object releases it
+    tester.set_object(None)
+    pytest.gc_collect()
+    assert objref() is None


include/pybind11/cast.h
```diff
@@ -1524,6 +1524,37 @@ struct holder_helper {
...
+template <typename T>
+struct holder_retriever {
+    static auto get_derivative_holder(const value_and_holder &v_h) -> decltype(v_h.template holder<T>()) {
+        return v_h.template holder<T>();
+    }
+};
+
+template <typename T>
+struct holder_retriever<std::shared_ptr<T>> {
+    struct shared_ptr_deleter {
```
virtuald, line 1538: > The delete expression must be well-formed, have well-defined behavior and not throw any exceptions. 

... I feel like this could throw? Alternatives?

YannickJadoul:
Hi, @virtuald. Thanks for trying this out in more detail than just the slightly crazy "hack" suggested on Gitter! :-)

I'm not sure I meant for this hack to be the default way `shared_ptr` is always handled, though. In it's current state, I can see a few reasons why this solution could cause massive havoc:
- Decreasing the reference count needs to hold the GIL, as it can call back into the Python interpreter (e.g. when deallocating). An straightforward fix would be to acquire the GIL in that deleter, of course.
  Doing so might still be tricky, as it's completely hiding a call into Python and could cause hard-to-fix deadlocks, while it _seems_ like it's a pure C++ construct. On the other hand, arguably `PYBIND11_OVERRIDE` already does this as well?
  There's also only a point in doing this if there's a trampoline class, right? Checking that might be another part of making this safer.
  We could add a test to demonstrate this issue?
- This will only work for `shared_ptr`. That's already better than it was of course, but I wonder how much confusion and bug reports this might cause ("This (unrelated) thing works with `shared_ptr` but not with `unique_ptr`.").

Starting from these two concerns and thinking further, I see a number of alternative approaches we could think about:
1. Document this hack/workaround, and maybe provide some kind of utility function.
2. Add some kind of a tag (similar to e.g. `py::keep_alive`) which applies this to selected arguments.
3. Come up with a small wrapper type (e.g. `py::shared_python_object<T>` or something like that), which uses this caster and has some kind of `std::shared_ptr<T> get()` method. This would make things explicit in the signature, and still make things easy by wrapping with a lambda.
4. Who knows what else there is. But given the issues from above, I'd think we need to be careful about doing things implicitly.

> Reference cycles are possible as a result, but shared_ptr is already susceptible to this in C++. #2757 discusses this topic a bit

This PR doesn't really suffer from the reference-cycles I mentioned to you on Gitter. The reference cycles are an issue when the Python object keeps a C++ object alive (which it does) while the C++ would keep the Python object alive. In this case, that's not really an issue, because there's no reference from the C++ object to the Python object; the `shared_ptr` is basically serving as an alias to a `py::object`, so we should be good.

I also know that @bstaletic was interested in pushing pybind11 towards using full-on garbage-collected heap types (as CPython [is apparently pushing people away from non-GC heap types](https://mail.python.org/archives/list/python-dev@python.org/thread/C4ILXGPKBJQYUN5YDMTJOEOX7RHOD4S3/)). A radically different solution, which would also work more generally, would be to do something similar to @EricCousineau-TRI's `wrapper` type, where the C++ trampoline class keeps a reference to the Python object and _make this cycle visible to Python's GC_ (with `tp_traverse`).

I'll also tag @EricCousineau-TRI, since he will be the one who has been thinking about this by far the longest of all of us.

YannickJadoul:
Also, I had forgotten about this, but I think I found where I got the original idea. This issue is a very interesting read: https://github.com/pybind/pybind11/issues/1049#issuecomment-326688270

virtuald:
> This will only work for shared_ptr. That's already better than it was of course, but I wonder how much confusion and bug reports this might cause ("This (unrelated) thing works with shared_ptr but not with unique_ptr.").

I believe you get an error if you pass something with a unique_ptr holder to something with a shared_ptr argument? Or at least, I tried it and I got an error, will have to check again tonight. If there was an automatic unique -> shared conversion going on somewhere, we could add this there as well.

I agree that it would be ideal if this could only be done for instances of trampoline classes. Any idea how to determine that?

YannickJadoul:
> I believe you get an error if you pass something with a unique_ptr holder to something with a shared_ptr argument? Or at least, I tried it and I got an error, will have to check again tonight. If there was an automatic unique -> shared conversion going on somewhere, we could add this there as well.

Yep, you're right, that's true; currently, you can't take away ownership from Python (though there's some work happening on that).

> I agree that it would be ideal if this could only be done for instances of trampoline classes. Any idea how to determine that?

Good question. I don't have time for a thorough search, currently, but this might be harder than I thought. It might be that all this information is lost after declaring a class and its `__init__` :-/

rwgk:
Currently I don't have the free bandwidth to fully page in here, but some quick remarks:

> I believe you get an error if you pass something with a unique_ptr holder to something with a shared_ptr argument? Or at least, I tried it and I got an error, will have to check again tonight. If there was an automatic unique -> shared conversion going on somewhere, we could add this there as well.

https://github.com/pybind/pybind11/pull/2672#issuecomment-748392993
```
pass_shared_pointee  RuntimeError: Unable to load a custom holder type from a default-holder instance
```

> ("This (unrelated) thing works with shared_ptr but not with unique_ptr.").

* `unique_ptr` with custom deleter: possible, if the deleter owns the `py::object`. This has only limited use because the `unique_ptr` deleter is part of the type. Additionally, the optimization compared to working with `shared_ptr` is probably negligible: for the extra trouble working with the `unique_ptr` custom deleter you only get a very small benefit in return.

* `unique_ptr` with default deleter: strongly on my do-not-even-try list. It's breaking the "you fully own the resources" contract that comes with `unique_ptr`: the new owner has no way of taking ownership of the `py::object`. For that to be safe, you'd need access to the `std::default_delete` internals, which is practically impossible as far as I know.

YannickJadoul:
> * `unique_ptr` with default deleter: strongly on my do-not-even-try list. It's breaking the "you fully own the resources" contract that comes with `unique_ptr`: the new owner has no way of taking ownership of the `py::object`. For that to be safe, you'd need access to the `std::default_delete` internals, which is practically impossible as far as I know.

Nonono, let's not go that way. Completely agreed!

The GC-type solution (similar to @EricCousineau-TRI's current solution) would solve this independently of holders, though (so also for non-`shared_ptr` copyable holders).
But I understand if @virtuald doesn't want to dive in that deeply :-) (It's also not a guaranteed success, and comes with its own caveats on non-deterministic GC.) So that's why my other suggestion is making things more explicit, here.

virtuald:
Unfortunately, gil_scoped_acquire is in pybind11.h, so I can't use it here. There's also a note there about how the GIL shouldn't be acquired if the interpreter is shutting down, but it's not clear to me how to avoid that problem here.

virtuald:
Regarding grabbing the GIL in the deleter, it appears that the std::function wrapper does this already so it's not without precedent.

virtuald:
I think this is worth doing by default, and I'm -1 on requiring a user to enable it via a tag or other hack. The current behavior of potentially losing the python half of an object when passing it to a C++ function is (while rare) clearly incorrect, unexpected, and very difficult to root cause without digging into the C++ code. 

Evidence that users agree with me: #1120, #1145, #1333, #1389, #1552, #1546 (which has a similar implementation as this as a workaround!), #1566, #1774, #1902, #1937, and probably others.

I've looked for a bit, and as far as I can tell there is not currently a mechanism to determine whether a class was constructed by python or not. To record that information:

* Add a bool `has_alias` to `type_record`, set that from `class_::has_alias`
* Add a bool `has_alias` to `type_info`
  * Requires bumping `PYBIND11_INTERNALS_VERSION`
* Set `type_info.has_alias` in `make_new_python_type` 

Even then, that just tells you there's an alias class associated with the python type, not whether the constructed python type was created from an alias. You might be able to add a check for that in `pybind11_meta_call` and set a flag in the instance. Not sure it's worth the hassle.

virtuald:
> This will only work for shared_ptr. That's already better than it was of course, but I wonder how much confusion and bug reports this might cause ("This (unrelated) thing works with shared_ptr but not with unique_ptr.").

Looking at this more, you could enable upgrading a unique holder to shared_ptr by providing an implementation of `holder_retriever` for `std::unique_ptr`, and removing the `check_holder_compat`.

include/pybind11/cast.h
virtuald: This is done in the std::function destructor too, so not going to deal with it for now.

virtuald:
Added the gil release, and a test inspired by #2844

virtuald:
This has been updated to only use the special deleter if an alias class is used.

I found a place to store the alias indicator that isn't currently covered by the `PYBIND11_INTERNALS_VERSION` guarantee -- in py::instance! ... which upon reflection, probably should be covered by it? I'm guessing that if objects from two different versions of pybind11 that did and didn't support this feature, you'd sometimes get the broken behavior and sometimes wouldn't. But that would probably be the case anyways unless we bumped `PYBIND11_INTERNALS_VERSION`?

Anyways, the last commit here changes init_instance to vary based on whether an alias class was used. If an alias class is used, a version of init_instance is used which sets a flag on the py::instance. This flag is used to determine whether the special deleter that keeps the python portion alive should be used or not.

cdyson37:
Sorry I'm a bit late to the party and haven't read the above fully, but I think this does a similar thing to my (now quite stale) PR https://github.com/pybind/pybind11/pull/1566 . I never worked out how to detect if it was an alias though, and in my version I did the shared_ptr dance a bit differently, haven't had a close look so don't know which is better, but I suspect yours. Would be great to get this problem fixed once and for all!

virtuald:
@cdyson37 yep, this solves the same problem yours does, but takes care of only doing this for aliased shared_ptrs.

rwgk:
> I found a place to store the alias indicator that isn't currently covered by the `PYBIND11_INTERNALS_VERSION` guarantee -- in py::instance! ... which upon reflection, probably should be covered by it? I'm guessing that if objects from two different versions of pybind11 that did and didn't support this feature, you'd sometimes get the broken behavior and sometimes wouldn't. But that would probably be the case anyways unless we bumped `PYBIND11_INTERNALS_VERSION`?

This is an awesome PR! My only significant concern is exactly this question.
@wjakob what's the right thing to do?

I tested this PR in the Google environment with all sanitizers (ASAN, MSAN, TSAN, UBSAN) and didn't find any issues (except a couple known TSAN, UBSAN failures that already exist on master).

I think the suggested changelog is a bit unclear. It doesn't mention alias classes. Is the following true? "For classes wrapped with (A) shared_ptr as the holder, and (B) having a trampoline (aka alias class), shared_ptrs passed from Python to C++ keep the Python object alive."

virtuald:
> I think the suggested changelog is a bit unclear. It doesn't mention alias classes. 

Thanks for catching that, I'll update the changelog entry. The alias class thing was added later.

rwgk:
Hi Dustin, I will merge #2841 in a minute, which will generate a merge conflict here I think. Sorry, but it's probably easy to resolve.

If it wasn't for the pesky `PYBIND11_INTERNALS_VERSION` question I think this PR could get merged very soon, but we needs Wenzel's input about it, and depending on what he recommends, we may need to wait. I'm not sure.

In the meantime, what do you think about splitting out a separate PR, just for factoring out `gil.h`? That would make this PR easier to review (for Wenzel) and rebase.

virtuald:
Already did that :) https://github.com/pybind/pybind11/pull/2845

virtuald:
Rebased 2845 and this on master to make it easier.

bstaletic:
> I think this PR could get merged very soon

In that case I believe I should restate @YannickJadoul's concern.

> Decreasing the reference count needs to hold the GIL, as it can call back into the Python interpreter (e.g. when deallocating). An straightforward fix would be to acquire the GIL in that deleter, of course.

I still think this is a very valid concern and I'll try to explain why I don't think the following statement from @virtuald, while true, isn't really a precedent.

> Regarding grabbing the GIL in the deleter, it appears that the std::function wrapper does this already so it's not without precedent.

That *is* true, however, a `std::function` owning a python callable also needs GIL for `operator()`. If I'm not mistaken, it automatically acquires the GIL there too. Still, it seems obvious that, if a call is reaching into the python interpreter, you'll need GIL for one reason or another. On the other hand, if we silently start shipping a reference to `self` in every `std::shared_ptr`, what used to be a pure C++ block, suddenly might need to acquire GIL to destruct a shared pointer. If this happens in a loop... oh, boy... While `std::shared_ptr`'s ref count is atomic and thus isn't the fastest thing in the world, it's still not GIL. The atomic ref count is local to the control block. GIL is global to the whole process and could effectively turn someone's multithreaded code into effectively single threaded. Granted, this means someone is destroying shared pointers in a tight loop, which is bad.

If my reasoning is, for whatever reason, invalid, feel free to correct me. Or call me an idiot.

To conclude, I'm not saying this isn't worth doing, but an opt-in mechanism might be a good idea.

virtuald:
> GIL is global to the whole process and could effectively turn someone's multithreaded code into effectively single threaded.

You could use this argument to justify not having trampoline classes. A python object that inherits from a C++ class and overrides a virtual function is effectively forcing C++ code that calls the virtual function to be single threaded because of the GIL. If you don't want single-threaded behavior, don't mix C++ and python, or be careful when you do.

I'll repeat what I said a few comments before and add emphasis:

> The current behavior of potentially losing the python half of an object when passing it to a C++ function is (while rare) **clearly incorrect, unexpected, and very difficult to root cause** without digging into the C++ code.

The current behavior is just wrong. Consider this C++ class:

```c++
struct C {
  virtual std::string getLanguage() { return "C++"; }
}; 
```
And this python class that overrides some trampoline:
```py
class PyC(mod.C):
  def getLanguage(self):
    return "python"
```
Now consider this C++ function:
```
std::string getLanguage(std::shared_ptr<C> c) {
  return c->getLanguage();
}
```
Without this patch, `mod.getLanguage(PyC())` returns `C++`. Who would ever want that behavior?

bstaletic:
> You could use this argument to justify not having trampoline classes.

Trampolines are very explicit and boilerplate heavy, even with all the macros. You do have to opt-in. It's the implicitness of this solution that doesn't sit right with me.

> The current behavior is just wrong.

Agreed, but not everyone who uses `std::shared_ptr` or trampolines holders is affected. I did say that a solution is worth pursuing.

virtuald:
> Trampolines are very explicit and boilerplate heavy, even with all the macros. You do have to opt-in. It's the implicitness of this solution that doesn't sit right with me.

This solution only affects users that have already opted into trampolines, and it does not affect any other users. Why should there be an additional step to opt-in?

virtuald:
I've tweaked it slightly and added an additional test for the new conditions. Now, the keep alive feature will only be applied IFF:

* Object type has an associated alias class
* Object was constructed by python (so objects constructed by C++ will never have this applied to them)

... I'm not totally sold on the necessity of this, but perhaps this will satisfy your objections for opting in @bstaletic and @YannickJadoul

virtuald:
One of my users has reported that this fix isn't quite there yet, they ran into an issue today where the python object disappeared again -- I wonder if the shared_ptr transfers the deleter as it's being moved around? I'll see if I can write up a duplicating test this weekend.

rwgk:
> One of my users has reported that this fix isn't quite there yet, they ran into an issue today where the python object disappeared again -- I wonder if the shared_ptr transfers the deleter as it's being moved around? I'll see if I can write up a duplicating test this weekend.

Thanks for letting us now. That sounds odd though, shared_ptr definitely remembers the deleter (I think in the control block shared by all copies of the same shared_ptr).

On another note, in the meantime Wenzel confirmed (via a chat today) that `PYBIND11_INTERNALS_VERSION` needs to be incremented for changes like you're making to `detail::instance`. I asked questions about handling the version increment, but haven't gotten feedback yet.

Yet another item: my PR #2875, which conflicts with your changes to test_smart_ptr.cpp. The main trouble is that it is messy for me to apply your PR here to the smart_holder branch, which I'd like to do, to run it through the Google global testing. Is there a chance you could move your `SP` structs out of `TEST_MODULE`, into the anonymous namespace? Or maybe even better, simply create a new `test_alias_shared_ptr`? I think that's best even if we didn't have the conflict between our PRs. For compatibility with the `smart_holder` branch it's still best to have the structs before `TEST_MODULE` (I'm not sure if your change also works with smart_holder as holder for you test `class_`es, it's something I'd like to find out, too).

virtuald:
If `PYBIND11_INTERNALS_VERSION` needs to be incremented, I might want to tweak this then to push has_alias into the per-type information instead of the per-instance information. There was another approach that I was contemplating also that might be better if I can remember what it was.

... I'm happy to make the test changes you suggested, but I don't quite understand what you want? I have $DAYJOB things to do, so I won't be able to get to this before the weekend. Feel free to make the changes you want and overwrite this branch?

rwgk:
> If `PYBIND11_INTERNALS_VERSION` needs to be incremented, I might want to tweak this then to push has_alias into the per-type information instead of the per-instance information. There was another approach that I was contemplating also that might be better if I can remember what it was.

FWIW: I floated a suggestion to introduce something like `-DPYBIND11_INTERNALS_VERSION_FUTURE` that let's you jump to an `#ifdef`ed future in situations where new features are more important than ABI stability. Being pinned down by compatibility concerns Is an issue that's popping up regularly. I think in general it's very undesirable that implementations are compromised long-term, to not have to increment the internals version (shorter-term). That may or may not apply to your case here, I don't know.
 
> ... I'm happy to make the test changes you suggested, but I don't quite understand what you want? I have $DAYJOB things to do, so I won't be able to get to this before the weekend. Feel free to make the changes you want and overwrite this branch?

Thanks!

rwgk:
Just a quick remark: I just merged #2875. I'll rebase this branch asap (it may take me a few hours).

rwgk:
Me not knowing better right now, I created https://github.com/pybind/pybind11/pull/2877, purely for running the CI.
In the past I saw people pushing changes to my own PRs, but I don't know how they did it.
@virtuald, could you please help me doing this right?

rwgk:
I played with this PR a bit. Some random things:

* Your new unit tests works with this simpler (and I believe preferable) change to `pybind.h` (I also pushed it to my temporary PR #2877, which I hope you'll adopt here):
```diff
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1547,6 +1547,7 @@ private:
     /// optional pointer to an existing holder to use; if not specified and the instance is
     /// `.owned`, a new holder will be constructed to manage the value pointer.
     static void init_instance(detail::instance *inst, const void *holder_ptr) {
+        inst->has_alias = has_alias;
         auto v_h = inst->get_value_and_holder(detail::get_type_info(typeid(type)));
         if (!v_h.instance_registered()) {
             register_instance(inst, v_h.value_ptr(), v_h.type);
```

* I think `inst->has_alias` could be leveraged to preempt UB like so (big caveat: I didn't actually test the fail case here):
```diff
diff --git a/include/pybind11/cast.h b/include/pybind11/cast.h
index 947cd356..e4b04d17 100644
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -626,6 +626,9 @@ struct holder_helper {
 template <typename T>
 struct holder_retriever {
     static auto get_derivative_holder(const value_and_holder &v_h) -> decltype(v_h.template holder<T>()) {
+        if (v_h.inst->has_alias) {
+          pybind11_fail("Missing life support for Python object owning pointee.");
+        }
         return v_h.template holder<T>();
     }
 };
```

* I experimented with holder mismatches, mostly just to understand what happens: UB, as on master, except when your new feature kicks in: I think you found an elegant way to return a correct `shared_ptr` even if the `class_` holder is something else (because you only use `value_ptr`)! — This experimental diff is only for future reference:
```diff
diff --git a/tests/test_trampoline_shared_ptr_return.cpp b/tests/test_trampoline_shared_ptr_return.cpp
index 505a9180..5b088691 100644
--- a/tests/test_trampoline_shared_ptr_return.cpp
+++ b/tests/test_trampoline_shared_ptr_return.cpp
@@ -36,13 +36,34 @@ struct SpGoAwayTester {
     std::shared_ptr<SpGoAway> m_obj;
 };
 
+template <typename T>
+class custom_unique_ptr {
+    std::unique_ptr<T> impl;
+public:
+    custom_unique_ptr(T* p) : impl(p) { }
+    T* get() const { return impl.get(); }
+};
+
+template <typename T>
+class padded_shared_ptr {
+    uint64_t pre_padding[10];
+    std::shared_ptr<T> impl;
+    uint64_t post_padding[10];
+public:
+    padded_shared_ptr(T* p) : impl(p) { }
+    T* get() const { return impl.get(); }
+};
+
 } // namespace
 
+PYBIND11_DECLARE_HOLDER_TYPE(T, custom_unique_ptr<T>);
+PYBIND11_DECLARE_HOLDER_TYPE(T, padded_shared_ptr<T>);
+
 TEST_SUBMODULE(trampoline_shared_ptr_return, m) {
     // For testing whether a python subclass of a C++ object dies when the
     // last python reference is lost
 
-    py::class_<SpBase, std::shared_ptr<SpBase>, PySpBase>(m, "SpBase")
+    py::class_<SpBase, custom_unique_ptr<SpBase>, PySpBase>(m, "SpBase")
         .def(py::init<>())
         .def("is_base_used", &SpBase::is_base_used);
 
@@ -56,7 +77,7 @@ TEST_SUBMODULE(trampoline_shared_ptr_return, m) {
     // For testing that a C++ class without an alias does not retain the python
     // portion of the object
 
-    py::class_<SpGoAway, std::shared_ptr<SpGoAway>>(m, "SpGoAway").def(py::init<>());
+    py::class_<SpGoAway, custom_unique_ptr<SpGoAway>>(m, "SpGoAway").def(py::init<>());
 
     py::class_<SpGoAwayTester>(m, "SpGoAwayTester")
         .def(py::init<>())
```

* Finally, your work doesn't affect the behavior of `class_` with `smart_holder` in any way, but if you get this PR merged, I could probably leverage `inst->has_alias` (or wherever the extra `bool` ends up) to also have the C++->Python life support with `smart_holder`.

* I'm very curious to learn how the feature failed for your user.

rwgk:
If your user gave you a reproducer, and you make it accessible somewhere, I could run it with our sanitizers. (I don't know if you have access to sanitizers.)

rwgk:
A couple more random things:

* I ran Google's global testing with #2877 patched in: no issues. — I didn't really expect any, but wanted to be sure.
* Somehow I missed your 3rd commit (adding in `is_alias`), my #2877 clone only has the first two. — I only realized too late, I can pull that in later, if we decide it's actually worth the trouble having both `has_alias` and `is_alias`.
* Just to preempt confusion: Last night I renamed (under #2877) the new files for your additional unit tests. The first name I picked was unfortunate.

virtuald:
Thanks for the feedback @rwgk, sorry I've been AWOL. I'm hoping to get to this tonight. 

My user did give me a reproducer, but I need to distill it down to a smaller test case. I think I understand the 'how' but not the 'why'. Once I get a reproducing test case I'll push it here -- but I'll make sure to pull in your changes first.

rwgk:
> but I'll make sure to pull in your changes first.

That could be messy because I missed your 3rd commit. I think it'll be better if you just manually apply the small simplification suggested under the [1st bullet point here](https://github.com/pybind/pybind11/pull/2839#issuecomment-786195704). When you're done I could move the new tests to a separate file again, un-interleaved (i.e. all structs first, then all bindings).

virtuald:
Rebased with your changes, looking at making a reproducer now.

virtuald:
The issue that my user reported ended up not to be an issue with this PR, though it is has yielded an interesting edge case that this doesn't cover. The class was inheriting from `enable_shared_from_this`, and it somewhere this was happening:

```c++
  return std::make_shared<OtherClass>(obj->shared_from_this());
```

Since the shared_ptr returned by `shared_from_this` was the original shared_ptr, it didn't hold a reference to the python object, and thus the python part died (while the C++ part lived on). 

Will have to think about this.

rwgk:
> The issue that my user reported ended up not to be an issue with this PR, though it is has yielded an interesting edge case that this doesn't cover. The class was inheriting from `enable_shared_from_this`, and it somewhere this was happening:
> 
> ```c++
>   return std::make_shared<OtherClass>(obj->shared_from_this());
> ```
> 
> Since the shared_ptr returned by `shared_from_this` was the original shared_ptr, it didn't hold a reference to the python object, and thus the python part died (while the C++ part lived on).
> 
> Will have to think about this.

Interesting. Just throwing in some thoughts, not being certain if/how this will help:

* We have special handling for enable_shared_from_this here: https://github.com/pybind/pybind11/blob/9c0aa69937dbb9d7023123d5d98914855432982c/include/pybind11/pybind11.h#L1505
* I didn't transfer this to the smart_holder adaption of the `init_instance` code. I just left it out when I wasn't sure what it's needed for, and all existing unit tests worked with smart_holder as default anyway. Now wondering if your user's case could be handled better via special handling of enable_shared_from_this.
* I'm turning over this vague idea in my mind, for the smart_holder branch: if `init_instance` is called for an alias, should the holder unconditionally create a shared_ptr with the custom deleter in this PR? Idea: that would automatically 1. ensure that the Python part outlives the C++ part, 2. trying to pass a unique_ptr (unsafe) will raise a runtime error. — This is highly speculative, potential detail devils left and right.

BTW, the smart_holder I'm talking about is [this](https://github.com/pybind/pybind11/blob/6285177afe80a65a3ef10762b4290cd53ad38ddc/include/pybind11/detail/smart_holder_poc.h).

YannickJadoul:
> I've tweaked it slightly and added an additional test for the new conditions. Now, the keep alive feature will only be applied IFF:
> 
>    * Object type has an associated alias class
> 
>    * Object was constructed by python (so objects constructed by C++ will never have this applied to them)
> 
> 
> ... I'm not totally sold on the necessity of this, but perhaps this will satisfy your objections for opting in @bstaletic and @YannickJadoul

It is the only way of not ending up with completely unexpected behavior for existing bindings.

Even this solution is still a high price to pay, IMO, as really the only time you need to keep the Python object alive is when you store it (as long as you stay in the same called function, the Python object will be alive and all of this is overhead). So I do still believe the kind of `def` tag I suggested at the start, similar to how we now need to annotate functions with `py::keep_alive` would be safer, making things explicit. Python and C++ just have different memory models and sometimes need some explicit patching, and we already need to do this anyway.

Again, I originally suggested this solution to your specific problem on Gitter, but I never remotely meant to apply it to every single `shared_ptr`.

But at least, after this change, the PR is just adding some inefficiency now, and (probably?) not being actively dangerous to existing code.

rwgk:
Hi @virtuald, FYI, there is now a solution for `unique_ptr` as well: #2902

Luckily the original PyCLIF authors mostly figured it out for us. I just had to nudge it a little bit more.

virtuald:
That's awesome @rwgk . I'm probably paged out until the end of the month, but the smart holder stuff sounds promising.

rwgk:
The smart_holder branch mentioned here previously just go a brand-new README:

https://github.com/pybind/pybind11/blob/smart_holder/README_smart_holder.rst

The branch includes the equivalent functionality of this PR (in smart_holder_type_casters.h), and a port of the new tests here.

virtuald:
So.. what's the deal here?

It seems that smart_holder does everything I (and pybind11 users!) want and more, so maybe this is OBE? If that's the case, what's the timeline for merging smart_holder?

rwgk:
> So.. what's the deal here?
> 
> It seems that smart_holder does everything I (and pybind11 users!) want and more, so maybe this is OBE? If that's the case, what's the timeline for merging smart_holder?

Hi Dustin, see here: https://github.com/pybind/pybind11/discussions/3576#discussioncomment-1889143

(Especially the cmake help bullet.)

iwanders:
I subscribed to this PR way back when #1566 was active, so I'm super excited to hear progress has been made on this @rwgk :tada: 

In regards to your CMake bullet point, what did you actually have in mind in regards to the cleanup / changes? As far as I can see the usual situation is: One python file, which is associated to one c++ file of the same name.

In case of the new unit test, the situation is different and there's one Python file and three c++ files, namely `class_sh_module_local_#` three times. Looking a bit further there's already like two other sets of unit tests where arbitrary test targets need to be added (`PYBIND11_CROSS_MODULE_TESTS`, `PYBIND11_CROSS_MODULE_GIL_TESTS`).

So, generalizing this I presume you want something generic that allows you to specify for a set of unit tests any list of additional targets? I tried making that in 7fe46ac93f564ef6f5222e878c182cfedecfb015, it's still not that pretty though, but I wouldn't know how to make it any prettier than this while preserving the arbitrary `n tests -> m additions` mapping.

virtuald:
@rwgk I tried using the smart_holder branch, and it almost worked -- but it couldn't handle my usage of enable_shared_from_this, so this PR still is useful for me.

The enable_shared_from_this error was the:

> smart_holder_type_casters loaded_as_shared_ptr failure: not implemented: trampoline-self-life-support for external shared_ptr to type inheriting from std::enable_shared_from_this.

I'm not exactly sure what's going on with it as it seems like pretty normal code to me (well.. its a trampoline of the child of a class that uses enable_shared_from_this), I'll have to create a small reproducer. If I were to guess from your comments in #3023, it probably stems from your reluctance to create reference cycles (whereas this PR is fine with them -- that's what you sign up for when you use shared_ptr!).

rwgk:
> I'll have to create a small reproducer.

I'll definitely take a close look at it.

> If I were to guess from your comments in #3023, it probably stems from your reluctance to create reference cycles

Hm ... not sure, yes in general, but no if it's made reasonably obvious by the API or some other means that a cycle is created. I wouldn't want that to be a trap, but creating cycles by choice is up to the user.

rwgk:
> In regards to your CMake bullet point, what did you actually have in mind in regards to the cleanup / changes?

@iwanders thanks a lot for helping with this! It would be great if you could move your work to a PR, so we can work on it together more easily.

From very rusty memory, my main worry is about running only selected tests. I have it in my mind that my change is breaking that feature. I'll look again asap to make that more concrete.

virtuald:
I've been using `smart_holder` for the last year or so, and it works great. I hope it hits master soon.

kliegeois:
@rwgk thanks for `smart_holder`. Is there a plan to merge the branch into master in the near/mid future?

rwgk:
> Is there a plan to merge the branch into master in the near/mid future?

Not the smart_holder branch. Long story. There is a plan but it's a bit too early to publicly commit to it.

I'm committed though to keeping the smart_holder branch up-to-date, for as long as it's useful, with a target of one or two weeks at the very most behind master. When we have pybind11 releases from master, usually I update the smart_holder branch the same day.

Is it difficult for you to deploy from the smart_holder branch? Would tagging help?

kliegeois:
> > Is there a plan to merge the branch into master in the near/mid future?
> 
> Not the smart_holder branch. Long story. There is a plan but it's a bit too early to publicly commit to it.
> 
> I'm committed though to keeping the smart_holder branch up-to-date, for as long as it's useful, with a target of one or two weeks at the very most behind master. When we have pybind11 releases from master, usually I update the smart_holder branch the same day.
> 
> Is it difficult for you to deploy from the smart_holder branch? Would tagging help?

Thanks for your message @rwgk. Currently, we can develop using the smart_holder branch (and we really appreciate the fact that you support this branch!). However, once we will want to release our work, the fact that we do not use the master branch could be an issue.
A second potential issue is that we consider to propose an update to binder to generate pybind11 codes compliant with the smart_holder. The fact that this is not part of the master branch might delay the acceptation of those changes up to the point the branch is merged.

rwgk:
> Thanks for your message @rwgk. Currently, we can develop using the smart_holder branch (and we really appreciate the fact that you support this branch!). However, once we will want to release our work, the fact that we do not use the master branch could be an issue. A second potential issue is that we consider to propose an update to binder to generate pybind11 codes compliant with the smart_holder. The fact that this is not part of the master branch might delay the acceptation of those changes up to the point the branch is merged.

There is a lot of complexity here, therefore I'll respond in two comments below: "Background", "More immediately practical maybe"

rwgk:
**Background**

One big issue with the smart_holder branch is that it has two modes, "Conservative" and "Progressive". That doubles the testing load: the CI workflow is run twice (CI = Conservative, CI-SH-DEF = Progressive). (I would be very uncomfortable not running the "Progressive" testing.)

Another issue is that we added in changes that go beyond the smart_holder functionality: `type_caster_odr_guard`, `return_value_policy::_return_as_bytes`, `return_value_policy::_clif_automatic`.

Those changes are critical to my primary work that basically funds all my pybind11 activity, integrating pybind11 into PyCLIF. That work is very far along: out of the many million Google-internal tests we have to satisfy, only a few thousand still don't work. But those are exactly the really tough ones to solve, and we need more intrusive changes in pybind11: #4349, #4499

Those two PRs are so far beyond the smart_holder functionality, we decided we need to branch further. This brings us to a "Construction Zone": https://github.com/google/pywrapcc

pywrapcc is really just another branch of pybind11, in a different repo only for practical-technical reasons (I really didn't like doing this myself, but the practical reasons are strong: automatic integration with the Google-internal repo & tooling).

pywrapcc will be the new destination for #4349, #4499 (soon, I hope).

The current plan from here (this is pretty firm):

* Continue with smart_holder as we have for the past two years (git merge master every week or so).
* On pywrapcc: git merge smart_holder as needed, probably in tandem with smart_holder updates.
* Google production will continue to use smart_holder for now.
* PyCLIF-pybind11 integration work will use pywrapcc.
* When we have the number of PyCLIF-pybind11 test failures down to zero: let's look at master, smart_holder, pywrapcc to decide how we want to organize the future.

Longer-term thinking (significantly less firm and not completely certain to work out exactly like planned):

Before you read on, please take a quick look at the top section of the [pywrapcc README](https://github.com/google/pywrapcc/blob/a73498a8e6c8a0d791d73ab0d31e70a973a30cdd/README.rst), mostly for the idea to do away with abrupt ABI breaks.

* Make smart_holder the default holder:
  * Completely remove the option to have `std:unique_ptr` or `std::shared_ptr` as holders: specifying those will select `smart_holder` anyway,
  * but retaining support for custom holders, complete with associated UB, for backward compatibility,
  * all without breaking compatibility with existing `internals` v4 wheels.

That will get rid of the "Conservative" / "Progressive" modes, but also be clearly a major new pybind11 version.

rwgk:
**More immediately practical maybe**

We could cherry-pick most of the smart_holder branch into master:

* Leaving behind `type_caster_odr_guard` and the two added `return_value_policy` enum members (easy).
* Also leaving behind the "Progressive" mode completely: testing & `#ifdef`s (a little hairy).  —  The original purpose was to be sure all corner cases are exercised with smart_holder as the holder. Maybe that's not so important anymore, given where we are with the PyCLIF-pybind11 integration work: that gives us a ton of in-the-wild testing. I'd be willing to give up the "Progressive mode", to not double the testing load on master.
* After smart_holder "Conservative" is on master: retire the smart_holder branch and continue only with master and pywrapcc.

I don't think this is very difficult or time-consuming. What I would need to get started are strong signals of support by other maintainers, and of course @wjakob, and a committed reviewer or two (fast turnaround, because having this change hanging in the air for more than a week or two would seriously hamper my primary project).

Tagging some potentially interested people for comment: @ax3l @eacousineau @henryiii  @Lalaland @Skylion007  @wangxf123456

Lalaland:
@rwgk I'd be happy to help review your smart_holder branch (especially because it would fix so many long standing bugs / issues). As long as it's ABI and API compatible it seems like a good idea to integrate within pybind11.

I think it would actually be good to have both progressive and conservative mode available, with an eventual goal of switching everything to smart_holder by default (which can be done at the next Python version release where everyone is forced to recompile the world anyways).

rwgk:
> @rwgk I'd be happy to help review your smart_holder branch (especially because it would fix so many long standing bugs / issues). As long as it's ABI and API compatible it seems like a good idea to integrate within pybind11.

A quick big thank you, that's awesome!

> I think it would actually be good to have both progressive and conservative mode available, with an eventual goal of switching everything to progressive by default (which can be done at the next Python version release when everyone is forced to recompile the world anyways).

I'm turning this over in my mind still. Will get back soon.

rwgk:
> I think it would actually be good to have both progressive and conservative mode available

I'd really like to remove the Progressive mode. The main reason: I think if we have to support it, it could be more difficult to move to smart_holder as the default, in case we need subtle behavior tweaks. I don't know if that's actually the case, but I would be more surprised than not if there aren't any.

But, is there anyone already relying on the Progressive mode?  —  I don't know about any use case. @virtuald @rhaschke?

(I know of one or two people that tried but wound up using the Conservative mode.)

How could I find out more systematically?

rhaschke:
> But, is there anyone already relying on the Progressive mode?

I guess "Progressive mode" refers to `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`, right? If so, I confirm that we don't use that mode anymore.

petersteneteg:
Hi, we compile with PYBIND11_USE_SMART_HOLDER_AS_DEFAULT and it works great! But I have no problem dropping that for classic mode, if it means getting it in master!

virtuald:
I also use `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`. I don't see why one would bother with the other holders (except for custom holder support) since they're broken.

But either way, I would like to also see smart_holder be merged to master, so I don't care how it gets there. I use my own fork of pybind11, so I can tweak my generator code as necessary.

rwgk:
> > But, is there anyone already relying on the Progressive mode?
> 
> I guess "Progressive mode" refers to `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`, right?

Yes.

> If so, I confirm that we don't use that mode anymore.

Thanks!

wjakob:
Regarding the merging `smart_holder` into the master branch: for a balanced discussion, I think it is useful to also mention potential downsides:

With the latest version of both branches, I am observing a _1.31x_ compilation time increase and a _1.48x_ binary size increase in a release build of a class binding benchmark (this compares `master` to `smart_holder` in progressive mode). See https://nanobind.readthedocs.io/en/latest/benchmark.html. That is a rather significant overhead on top of baseline numbers that are arguably already concerning.

rwgk:
Thanks for sharing those results. The measurements are larger than I would have guessed, especially the binary size increase is really surprising.

For my particular situation:
1. I don't have a choice, I absolutely need those features.
2. In a typical real-world situation, hopefully the size of the bindings compared to the size of what's being bound is relatively small.

People have a choice, that's good.

kliegeois:
@rwgk do you have an update regarding the potential merging of `smart_holder` ? Is it still considered and if yes do you know when it might happen?

Thanks!

rwgk:
> @rwgk do you have an update regarding the potential merging of `smart_holder` ? Is it still considered and if yes do you know when it might happen?

Sorry not sure still. — The project I wrote the smart_holder for is at just a little over 1 failure in 100,000 tests, but I still cannot know for sure when we're all the way through, and that's the highest priority.

I'll continue merging master into smart_holder with just a few days delay at most.

If someone wants to help that would be great:

* Test are needed for the fix under #4850.
* Two `return_value_policy` additions need to be removed (the main work there is merging the resulting smart_holder into pywrapcc, where we want to keep them).
* A pass through the documentation.

galenbwill:
I have also run into this problem. Is there any ETA yet for merging in the [`smart_holder`](https://github.com/pybind/pybind11/tree/smart_holder) branch?

rwgk:
> I have also run into this problem. Is there any ETA yet for merging in the [`smart_holder`](https://github.com/pybind/pybind11/tree/smart_holder) branch?

See #5339.

I requested feedback five days ago (2024-08-26) but haven't gotten any. I want to encourage everyone to chime in.

rwgk:
Closing, after PR #5542 was merged. The smart_holder feature was influenced by this PR.

PR #2840, opened 2021-02-01
Add make_namespace function and tests

joukewitteveen:
## Description

This  feature request has not been discussed (by me, at least) elsewhere.

Namespaces can be convenient as arguments to Python functions called from C++. In simple cases, they are easier to construct than that classes are instantiated.

On the Python side, these simple namespaces are very simple. They are basically objects on which attributes can be set and removed.

Next to the addition of `make_namespace`, I did some minor expansion of documentation and tests.

## Suggested changelog entry:

```rst
Add `make_namespace` function for instantiating simple namespaces.
```


joukewitteveen:
Indeed, simple namespaces are only available from Python 3.3 onward. That is why some checks are not successful. Please let me know if this MR has any chance of being accepted before I try and think of a fallback for older Python versions.

YannickJadoul:
Hi, @joukewitteveen! Thanks for this!

A few questions, if you don't mind:
- I personally never had any use for these, but I'm assuming you have something in mind?
- Is there any Python documentation on this? I'm wondering if we should link to it in the docs, to avoid confusion with C++ namespaces. At any rate, maybe you could still add a note that the equivalent of C++ namespaces would normally just be `def_submodule`?
- Is there a point in having a `py::namespace` object? Since you're making the comparison to `py::tuple`/`py::make_tuple`, that would make sense. But at the same time, there doesn't seem to be a lot of API related to Python's namespaces (and hardly any docs, it seems).

Other than that: _shrug_, why not, if it can be useful :-)

tests/test_pytypes.cpp
```diff
@@ -75,6 +75,19 @@ TEST_SUBMODULE(pytypes, m) {
         return dict.contains(val);
     });
 
+    // test_tuple
+    m.def("make_tuple", []() { return py::make_tuple(42, py::none(), "spam"); });
```
YannickJadoul, line 79: Isn't this tested anywhere else!? If so, wow, good catch! If not: do we want to repeat this here?

joukewitteveen: There are other tests that rely on `make_tuple` to work, so if it wouldn't work that would not go unnoticed. However, nowhere are specific tests for `make_tuple` grouped. I thought it would be good to be a bit more explicit in that regard and, if nothing else, provide a space for such tests to be collected.

joukewitteveen:
>     * I personally never had any use for these, but I'm assuming you have something in mind?

I needed to pass an instantiated `Logger` class in a Python call. I wanted logging to be taken care of by C++ functions, and using a namespace with a few attributes (`cppfunction`s) is easier than instantiating and configuring a class.

>     * Is there any Python documentation on this? I'm wondering if we should link to it in the docs, to avoid confusion with C++ namespaces. At any rate, maybe you could still add a note that the equivalent of C++ namespaces would normally just be `def_submodule`?

I added a link to it, but decided against a note on C++ namespaces versus (sub)modules. I'm not sure it is true that those are conceptually linked so strongly. As I understand it, a namespace is pretty fundamental in Python and modules (and classes and everything else) are mostly namespaces with "extra logic". Thus, a `SimpleNamespace` (from `make_namespace`) can in fact be thought of as an (maybe not "the") equivalent of a C++ namespace.

>     * Is there a point in having a `py::namespace` object? Since you're making the comparison to `py::tuple`/`py::make_tuple`, that would make sense. But at the same time, there doesn't seem to be a lot of API related to Python's namespaces (and hardly any docs, it seems).

The core API relevant to a `SimpleNamespace`, I think, consists of `hasattr`, `delattr`, `getattr`, `setattr`. Those are present on objects, so we don't need a `py::namespace`. In fact, `make_namespace()` is one way to get an "empty" object (not just what effectively amounts to a `nullptr`).

include/pybind11/cast.h
```diff
@@ -1890,6 +1890,14 @@ template <return_value_policy policy = return_value_policy::automatic_reference,
     return result;
 }
 
+#if PY_MAJOR_VERSION >= 3
+template <typename... Args,
+          typename = detail::enable_if_t<args_are_all_keyword_or_ds<Args...>()>>
+object make_namespace(Args&&... args_) {
+    return reinterpret_steal<object>(_PyNamespace_New(dict(std::forward<Args>(args_)...).ptr()));
```
YannickJadoul, line 1897: Can `_PyNamespace_New` return `NULL` and set an error inside CPython?

bstaletic:
Does `py::isinstance` work with these things? Usually there's `PyThing_Check(PyThingObject*)`, but there isn't a `PyNamespace_Check`. However, besides the `_PyNamespace_New()`, we also have access to `_PyNamespace_Type` type object. So... if `PyObject_IsInstance(namespace_ptr, &_PyNamespace_Type);` works, `py::isinstance` should work as well. At that point, it would be inconsistent to allow `py::isinstance(object, type)` to work, but not `py::isinstance<pybind_type>(object)`.

YannickJadoul:
> but decided against a note on C++ namespaces versus (sub)modules. I'm not sure it is true that those are conceptually linked so strongly.

Isn't this the closest you can practically get to C++ namespaces?

To me, it seems Python namespaces are basically the lambdas of objects; some kind of anonymous object?
C++ namespaces and Python submodules are +- meant to organize functions, classes, ... hierarchically within a project, so are to me (practially speaking) much closer.

Also, we really need to take care of people coming from both C++ as well as Python. The link to the Python docs already helps a bit, though.

include/pybind11/cast.h
bstaletic: https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Objects/namespaceobject.c#L33

It can return NULL, but not sure about setting the error. Error part is on `PyDict_New`.

joukewitteveen: It can indeed return NULL. I'm not sure this will actually happen, but I revised it anyway so that it is more explicit.

joukewitteveen:
> Does `py::isinstance` work with these things? Usually there's `PyThing_Check(PyThingObject*)`, but there isn't a `PyNamespace_Check`. However, besides the `_PyNamespace_New()`, we also have access to `_PyNamespace_Type` type object. So... if `PyObject_IsInstance(namespace_ptr, &_PyNamespace_Type);` works, `py::isinstance` should work as well. At that point, it would be inconsistent to allow `py::isinstance(object, type)` to work, but not `py::isinstance<pybind_type>(object)`.

I am not sure what you mean here. Can you give a specific code example? Without a `py::namespace` type, it is unclear to me how this is relevant.

> > but decided against a note on C++ namespaces versus (sub)modules. I'm not sure it is true that those are conceptually linked so strongly.
> 
> Isn't this the closest you can practically get to C++ namespaces?
> 
> To me, it seems Python namespaces are basically the lambdas of objects; some kind of anonymous object?
> C++ namespaces and Python submodules are +- meant to organize functions, classes, ... hierarchically within a project, so are to me (practially speaking) much closer.
> 
> Also, we really need to take care of people coming from both C++ as well as Python. The link to the Python docs already helps a bit, though.

I agree that we should avoid confusion. Do you have any specific addition to the documentation in mind?

The way I think of it, a namespace is nothing more than a mapping from names to, in the case of Python, objects. A `SimpleNamespace` does this in a very bare bones manner. I'm not sure what you mean by 'anonymous object', but indeed, modules usually have [more attributes](https://docs.python.org/3/reference/import.html#import-related-module-attributes). Whether this makes them more akin to C++ namespaces I don't know. Both can provide a hierarchical organization of functions, classes, etc..

joukewitteveen:
@YannickJadoul @bstaletic Is there anything expected from me to advance this pull request, or is the request simply awaiting its turn in the 2.7 cycle for further decisions?

bstaletic:
> I am not sure what you mean here. Can you give a specific code example? Without a py::namespace type, it is unclear to me how this is relevant.

Does `py::isinstance(namespace_object, &_PyNamespace_Type);` return `true`?

joukewitteveen:
> Does `py::isinstance(namespace_object, &_PyNamespace_Type);` return `true`?

`py::isinstance(obj, (PyObject *) &_PyNamespace_Type);` returns `true` on namespace object and `false` on others (e.g. tuples). Shall I add this observation to the tests?

bstaletic:
I'm saying that if `py::isinstance` works, and you've just said that it does, this warrants a proper `py::namespace` type. Why should `_PyNamespace_Type` be different from all the other python types that pybind11 supports?

joukewitteveen:
@bstaletic: That makes sense. I tried my hand at making `namespace_` a proper type (do we want `using namespace = namespace_` to enable `py::namespace`?). It fails on PyPy, where `SimpleNamespace`s are implemented as Python classes. I would like to do
```diff
diff --git a/include/pybind11/pytypes.h b/include/pybind11/pytypes.h
index e6cfc70..e070026 100644
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -764,7 +764,14 @@ inline bool PyUnicode_Check_Permissive(PyObject *o) { return PyUnicode_Check(o)
 #endif
 
 #if PY_MAJOR_VERSION >= 3
-inline bool PyNamespace_Check(PyObject *o) { return isinstance(o, (PyObject *) &_PyNamespace_Type); }
+inline bool PyNamespace_Check(PyObject *o) {
+#if !defined(PYPY_VERSION)
+    auto type = (PyObject *) &_PyNamespace_Type;
+#else
+    auto type = module_::import("types").attr("SimpleNamespace");
+#endif
+    return isinstance(o, type);
+}
 #endif
 
 inline bool PyStaticMethod_Check(PyObject *o) { return o->ob_type == &PyStaticMethod_Type; }
```
but I guess `module_` is not available in `pytypes.h`. Any suggestions?

bstaletic:
> Do we want using namespace = namespace_ to enable py::namespace?

No. `using module = module_;` is there for backwards compatibility. You'll notice that there isn't `py::class`, only `py::class_`.

> module_::import("types").attr("SimpleNamespace");


    auto types_module = py::reinterpret_steal<py::object>(PyImport_ImportModule("types"));

YannickJadoul:
> > module_::import("types").attr("SimpleNamespace");
> 
> ```
> auto types_module = py::reinterpret_steal<py::object>(PyImport_ImportModule("types")); 
> ```

Why? That's what `module_::import("types")` is also doing, no?

YannickJadoul:
Given this complication with PyPy, I'm wondering, btw: is there any point in doing `isinstance(obj, py::namespace_)`? These objects are completely meant as duck-typed things, anyway, right? Is there then any point in making this a full `py::object` subtype?

joukewitteveen:
Another possibility is to introduce a `detail::namespace_type`, which is set to `(PyObject *) &_PyNamespace_Type` with Python and to `module_::import("types").attr("SimpleNamespace")` with PyPy. The latter definition would be placed below the definition of `class module_` in `pybind11.h`.
This does mean we import `types` whenever we run with PyPy.

> Given this complication with PyPy, I'm wondering, btw: is there any point in doing `isinstance(obj, py::namespace_)`? These objects are completely meant as duck-typed things, anyway, right? Is there then any point in making this a full `py::object` subtype?

Personally I cannot think of a use case where it makes a lot of sense to check for namespace instances. As I argued before, these simple namespaces are as close to a bare object as I think you can get. I don't think it hurts to think of them as just `py::object`s.

Shall I revert back to my original `make_namespace` proposal?

YannickJadoul:
> Shall I revert back to my original `make_namespace` proposal?

If it's up to me, and you don't mind, sure. I think I was fine with the PR, but hadn't gotten the chance to click "approve".

> Personally I cannot think of a use case where it makes a lot of sense to check for namespace instances. As I argued before, these simple namespaces are as close to a bare object as I think you can get. I don't think it hurts to think of them as just `py::object`s.

Not saying this PR isn't useful or good, but funny you say this: why not just literally create a `py::object` and assign with `obj.attr("foo") = ...`? Just out of curiosity; not saying this PR needs a big overhaul!
I recently also had a situation where this would have been useful, btw :-)

joukewitteveen:
> > Shall I revert back to my original `make_namespace` proposal?
> 
> If it's up to me, and you don't mind, sure. I think I was fine with the PR, but hadn't gotten the chance to click "approve".

Done.

> > Personally I cannot think of a use case where it makes a lot of sense to check for namespace instances. As I argued before, these simple namespaces are as close to a bare object as I think you can get. I don't think it hurts to think of them as just `py::object`s.
> 
> Not saying this PR isn't useful or good, but funny you say this: why not just literally create a `py::object` and assign with `obj.attr("foo") = ...`? Just out of curiosity; not saying this PR needs a big overhaul!
> I recently also had a situation where this would have been useful, btw :-)

Maybe I am missing something completely, but this would not work because constructing a `py::object` does not instantiate a Python object (it defaults to a `nullptr`), right? At least `.attr` segfaults on a freshly created pure `py::object` for me.

YannickJadoul (approved):
> Done.

Thanks a lot! Apologies for the back and forth on this one; guess it's part of figuring out the desired API. But this pretty looks good, to me :-)

If you still feel adventurous, allowing `auto d = py::dict("foo"_a = 42, "bar"_a = py::none(); py::make_namespace(**d);` could be fun, but I'm perfectly happy adding that when there's any need for it.

> Maybe I am missing something completely, but this would not work because constructing a `py::object` does not instantiate a Python object (it defaults to a `nullptr`), right? At least `.attr` segfaults on a freshly created pure `py::object` for me.

Nope, you're not wrong. I was just loosely thinking that maybe there should be a way of constructing an empty `py::object`. But let's not dive into that in this PR anymore :-)

joukewitteveen:
> > Done.
> 
> Thanks a lot! Apologies for the back and forth on this one; guess it's part of figuring out the desired API. But this pretty looks good, to me :-)

No problem. It was a good opportunity to get to know some of the internals of pybind11.

> If you still feel adventurous, allowing `auto d = py::dict("foo"_a = 42, "bar"_a = py::none(); py::make_namespace(**d);` could be fun, but I'm perfectly happy adding that when there's any need for it.

I'm not sure this would be a good idea, but I agree that it looks like a fun addition. I would not encourage people to think of namespaces as dictionaries too much, since namespaces have different key requirement.

> > Maybe I am missing something completely, but this would not work because constructing a `py::object` does not instantiate a Python object (it defaults to a `nullptr`), right? At least `.attr` segfaults on a freshly created pure `py::object` for me.
> 
> Nope, you're not wrong. I was just loosely thinking that maybe there should be a way of constructing an empty `py::object`. But let's not dive into that in this PR anymore :-)

Haha, we've come full circle. You can't really have an empty `py::object`, because there is no such thing as a bare object type. One thing that comes real close to it will be `py::make_namespace()` ;-).

YannickJadoul:
> Haha, we've come full circle. You can't really have an empty `py::object`, because there is no such thing as a bare object type. One thing that comes real close to it will be `py::make_namespace()` ;-).

Right!
Well, also kind of wrong: you can make an `object()` (I've seen it being used as `sentinel = object()` when `None` is a valid value). But you can't assign attributes to it, so I did miss that, indeed!
```pycon
>>> x = object()
>>> x.foo = 3
AttributeError: 'object' object has no attribute 'foo'
```

docs/advanced/pycpp/object.rst
```diff
@@ -20,6 +20,46 @@ Available types include :class:`handle`, :class:`object`, :class:`bool_`,
...
+A `simple namespace`_ can be instantiated using :func:`py::make_namespace`:
+
+.. code-block:: cpp
+
+    using namespace pybind11::literals; // to bring in the `_a` literal
+    py::object ns = py::make_namespace("spam"_a=py::none(), "eggs"_a=42);
+
+Attributes on a namespace can be modified with the :func:`py::delattr`,
+:func:`py::getattr`, and :func:`py::setattr` functions. Namespaces can be useful
+as stand-ins for class instances.
```
rwgk, line 52: ... as very light-weight stand-ins ...

include/pybind11/cast.h
```diff
@@ -1018,6 +1018,16 @@ template <return_value_policy policy = return_value_policy::automatic_reference,
     return result;
 }
 
+#if PY_MAJOR_VERSION >= 3
```
rwgk, line 1021: In the documentation I see "New in version 3.3."
Mainly to be slightly more informative, at ~no cost, could you make this `PY_VERSION_HEX >= 0x03030000`?

docs/advanced/pycpp/object.rst
```diff
@@ -20,6 +20,46 @@ Available types include :class:`handle`, :class:`object`, :class:`bool_`,
...
+
+.. note::
+
+    ``make_namespace`` is not available in Python 2.
+
+.. versionchanged:: 2.7
```
rwgk, line 58: 2.8

tests/test_pytypes.cpp
```diff
@@ -70,6 +70,19 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("dict_contains",
           [](const py::dict &dict, const char *val) { return dict.contains(val); });
 
+    // test_tuple
+    m.def("get_tuple", []() { return py::make_tuple(42, py::none(), "spam"); });
+
+#if PY_MAJOR_VERSION >= 3
```
rwgk, line 76: Here too, if you find my other suggestion useful.

rwgk (approved):
Really nice, thank you @joukewitteveen!
Caveat: I didn't read the many PR comments. But I really like the end result!

joukewitteveen:
Thanks, @rwgk, for your suggestions. I implemented them.

Since 2.7 is out, I think the corresponding milestone should be closed. It would be cool if that happened automatically when tagging a release. There's probably a way to do that.

Note that #2451 introduces functionality that should also be mentioned in the documentation added in this PR. I think it is best to first merge this one and only merge #2451 after it is rebased and amended with some documentation updates.

Skylion007:
My only other qualm is make_namespace to generic / not specific of a name? Especially with C++ namespaces.

rwgk:
> My only other qualm is make_namespace to generic / not specific of a name? Especially with C++ namespaces.

That's a good point. Systematically inserting "simple" would be great.

joukewitteveen:
I don't think there is a real risk of confusion. At the end of the day, both C++ and Python have namespaces and the objects returned from `make_namespace` fit the Python definition of a namespace.

That being said, I too think adding `simple` is a good idea (although the name gets a bit long). Shall I go with `make_simplenamespace` or `make_simple_namespace`?

rwgk:
> That being said, I too think adding `simple` is a good idea (although the name gets a bit long). Shall I go with `make_simplenamespace` or `make_simple_namespace`?

The 2nd version, with the underscore, looks much better to me.
(I think code will be easier to read with the longer name. Without the "simple" people would first go huh? until they go look up what it is.)

tests/test_pytypes.cpp
```diff
@@ -70,6 +70,19 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("dict_contains",
           [](const py::dict &dict, const char *val) { return dict.contains(val); });
 
+    // test_tuple
+    m.def("get_tuple", []() { return py::make_tuple(42, py::none(), "spam"); });
+
+#if PY_VERSION_HEX >= 0x03030000
+    // test_namespace
```
rwgk, line 77: Could you please insert the "simple" here as well (`test_simple_namespace`, `get_simple_namespace`)?

That will make it easier to pin-point everything to do with simple_namespace (useful for maintenance).

joukewitteveen: Done.

rwgk (approved):
Awesome, thanks!
Just waiting for the CI to finish, and @Skylion007 to take another look.

PR #2841, opened 2021-02-02
Splitting out detail/type_caster_base.h from cast.h, with iwyu cleanup.

rwgk:
The first part of `cast.h` is moved to `detail/type_caster_base.h`.
The break point is the end of `class type_caster_base`.

I believe breaking up the very large `cast.h` is a good step in its own right, but the immediate motivation is to help moving PR #2672 forward.

Cleanup of includes was guided by `include-what-you-use 0.12 based on clang version 9.0.1-11`.
The approach was ad-hoc but simple:
```
echo '#include "cast.h"' > include/pybind11/cast.cc
echo '#include "type_caster_base.h"' > include/pybind11/detail/type_caster_base.cc
iwyu -std=c++17 -DPYBIND11_TEST_BOOST -Ipybind11/include -I/usr/include/python3.8 -I/usr/include/eigen3 include/pybind11/cast.cc
iwyu -std=c++17 -DPYBIND11_TEST_BOOST -Ipybind11/include -I/usr/include/python3.8 -I/usr/include/eigen3 include/pybind11/detail/type_caster_base.cc
```

* The suggested fixes were applied manually, with pybind11 includes first, system includes second.
* A few suggested forward declarations and warnings about missing `inline` functions were ignored. Resolving them probably needs refactoring of other files.
* Suggested Python includes (such as `patchlevel.h`) were ignored.
* A few missing `std::` prefixes were added to avoid C-style includes.

Changelog not needed.


rwgk:
This is what I currently have to do: https://github.com/rwgk/pybind11/blob/0e5d2d3a21e2607f0307bbe88e3603f263ebba27/include/pybind11/cast.h#L958

include/pybind11/cast.h
```diff
@@ -10,947 +10,11 @@
...
-#  endif
-#endif
-#ifdef PYBIND11_HAS_STRING_VIEW
-#include <string_view>
-#endif
-
-#if defined(__cpp_lib_char8_t) && __cpp_lib_char8_t >= 201811L
-#  define PYBIND11_HAS_U8STRING
-#endif
+#include "detail/type_caster_base.h"
```
YannickJadoul, line 13: Not a full review, but a first drive-by comment: this is not really good include-what-you-use practice (referring to the principle, not the tool!).

Any chance of a quick scan on what part of the original includes is still used in the leftover file? This doesn't need to be exactly perfect (the rest of pybind11 isn't either), but depending on the includes of an included file is not amazing (as far as I know).

YannickJadoul (commented):
I see @wjakob was requested for review. That sounds like a good idea, indeed.

In general, this PR doesn't seem like a bad idea to me, though I'm slightly confused how it helps you on the other PR? Before, you would include `pybind11/cast.h`, now you include `pybind11/detail/type_caster_base.h`?

henryiii:
I'm generally in favor of more, smaller files with better separation. Makes it easier to refactor, find what you are looking for, include just the right part, etc.

The correct includes should be places in each file, though. Ideally include-what-you-use style.

rwgk:
> The correct includes should be places in each file, though. Ideally include-what-you-use style.

Thanks @henryiii, will do. In the meantime I made more progress running all existing unit tests with the new `smart_holder` as the default holder (in my "sideline" branch). I'll get back here either as soon as I get all the way through, or reach some stable intermediate point.

include/pybind11/cast.h
rwgk: See updated PR description (I actually used the `iwyu` tool).

wjakob (approved):
LGTM!

rwgk:
Thanks! :-)

PR #2845, opened 2021-02-03
Extract gil management functions to separate header

virtuald:
## Description

Move gil_scoped_* to gil.h so that it can be used from cast.h, required for #2839 

## Suggested changelog entry:

N/A


henryiii (approved):
I'm generally in favor of more, smaller files with better separation. Makes it easier to refactor, find what you are looking for, include just the right part, etc.

rwgk:
CI all green! I'll merge this now. Thanks @henryiii for your approval!

PR #2846, opened 2021-02-04
fix: add null pointer check with std::localtime

knzivid:
## Description

Related conversation https://gitter.im/pybind/Lobby?at=601be2a61ed88c58d81d719a

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -161,9 +161,16 @@ public:
         // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
         // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
+
+        // std::localtime returns a pointer to a static internal std::tm object on success,
+        // or null pointer otherwise
+        std::tm *localtime_ptr = std::localtime(&tt);
+        if (!localtime_ptr)
+            throw cast_error("Unable to represent system_clock in local time");
+
         // this function uses static memory so it's best to copy it out asap just in case
         // otherwise other code that is using localtime may break this (not just python code)
-        std::tm localtime = *std::localtime(&tt);
+        std::tm localtime = *localtime_ptr;
 
         return PyDateTime_FromDateAndTime(localtime.tm_year + 1900,
                                           localtime.tm_mon + 1,


include/pybind11/chrono.h
```diff
@@ -161,9 +161,17 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
         // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
         // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
+
+        // std::localtime returns a pointer to a static internal std::tm object on success,
+        // or null pointer otherwise
+        // (https://en.cppreference.com/w/cpp/chrono/c/localtime)
+        std::tm *localtime_ptr = std::localtime(&tt);
+        if (!localtime_ptr)
+            pybind11_fail("pybind11::chrono: Unable to represent system_clock in local time");
```
YannickJadoul, line 170: `pybind11_fail` is meant for internal errors, so this is not correct. Maybe throw `py::cast_error`? Maybe have a look at what other casters do, please.

include/pybind11/chrono.h
```diff
@@ -161,9 +161,17 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
         // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
         // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
+
+        // std::localtime returns a pointer to a static internal std::tm object on success,
+        // or null pointer otherwise
+        // (https://en.cppreference.com/w/cpp/chrono/c/localtime)
+        std::tm *localtime_ptr = std::localtime(&tt);
```
YannickJadoul, line 168: When can this fail? It would be nice to know when "otherwise" is the case. AFAIK, there haven't been issues with this, yet, but it's good to check, I suppose.

include/pybind11/chrono.h
```diff
@@ -161,9 +161,17 @@ template <typename Duration> class type_caster<std::chrono::time_point<std::chro
         // > If std::time_t has lower precision, it is implementation-defined whether the value is rounded or truncated.
         // (https://en.cppreference.com/w/cpp/chrono/system_clock/to_time_t)
         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));
+
+        // std::localtime returns a pointer to a static internal std::tm object on success,
+        // or null pointer otherwise
+        // (https://en.cppreference.com/w/cpp/chrono/c/localtime)
```
YannickJadoul, line 167: IMO, the link is not really needed, since we can find it back, when necessary. But it can stay if you prefer.

include/pybind11/chrono.h
knzivid: found this on SO https://stackoverflow.com/a/58432714/3951920

include/pybind11/chrono.h
knzivid: I can remove it

YannickJadoul: You don't per se have to; it's not wrong either! (I just thought cppreference.com is the go-to reference anyway :-) )

include/pybind11/chrono.h
YannickJadoul: Thanks!

include/pybind11/chrono.h
knzivid: How about now?

YannickJadoul (approved):

include/pybind11/chrono.h
YannickJadoul: If the CI wants to turn green, this seems better, indeed. Thanks! :-)

knzivid: I am not sure what broke in `test_recarray` in test_numpy_dtypes.py on CentOS 7. I will try debugging it, now that it has failed twice.

YannickJadoul: Very weird that it would be related, though

henryiii: It's not, it's started breaking everywhere.

PR #2848, opened 2021-02-07
test: Test against Python 3.10

kumaraditya303:
## Description

Test against Python 3.10a5


YannickJadoul (approved):
We already are building against nightly `deadsnakes` builds, but yeah, why not reenable this one :-)

Issue #2862, opened 2021-02-17
[BUG] Unchecked errors can potentially swallow signals/other exceptions


virtuald:
## Issue description

I'm trying to figure out why CTRL-C isn't working in my Linux application.

I'm periodically checking PyErr_CheckSignals in my inner loop in the main thread and it never seems to be firing -- but python is clearly seeing it at some point, as when I close the UI window the python code throws the KeyboardInterrupt and everything crashes because a thread tries to take the GIL during shutdown. That is not what this issue is about -- I'm sure I'm not doing something right, just need to figure out the right way to do it.

However, in my search the following two links come up:

* https://bugs.python.org/issue35459 (PR @ https://github.com/python/cpython/pull/11112)
* https://stackoverflow.com/a/25877795/5483904

Basically, `PyDict_Get/Set/Del` can potentially swallow errors (including signals!) if the WithError variants aren't used. There are a few places in pybind11 where the less safe versions of these functions are used, which could be potentially contributing to my problem. It seems the WithError variants were introduced in Python 3, so to switch to these would require a Python 2.7 shim.

Has anyone else been contemplating this? There aren't that many places where errors aren't checked in pybind11, so it probably wouldn't be terrible to find them all and squash em. Ideally it would be nice if we could add some kind of linter to make sure nobody introduces such things in the future, but that seems hard.

## Reproducible example code

N/A

PR #2863, opened 2021-02-17
Check dict item accesses where it isn't already checked

virtuald:
## Description

* Initial stab at https://github.com/pybind/pybind11/issues/2862

## Suggested changelog entry:

N/A

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -485,6 +485,43 @@ inline handle get_function(handle value) {
     return value;
 }
 
+// Reimplementation of python's dict helper functions to ensure that exceptions
+// aren't swallowed (see #2862)
+
+// copied from cpython _PyDict_GetItemStringWithError
+inline PyObject * dict_getitemstring(PyObject *v, const char *key)
+{
+#if PY_MAJOR_VERSION >= 3
+    PyObject *kv, *rv;
+    kv = PyUnicode_FromString(key);
+    if (kv == NULL) {
+        throw error_already_set();
+    }
+
+    rv = PyDict_GetItemWithError(v, kv);
+    Py_DECREF(kv);
+    if (rv == NULL && PyErr_Occurred()) {
+        throw error_already_set();
+    }
+    return rv;
+#else
+    return PyDict_GetItemString(v, key);
+#endif
+}
+
+inline PyObject * dict_getitem(PyObject *v, PyObject *key)
+{
+#if PY_MAJOR_VERSION >= 3
+    PyObject *rv = PyDict_GetItemWithError(v, key);
+    if (rv == NULL && PyErr_Occurred()) {
+        throw error_already_set();
+    }
+    return rv;
+#else
+    return PyDict_GetItem(v, key);
+#endif
+}
+
 // Helper aliases/functions to support implicit casting of values given to python accessors/methods.
 // When given a pyobject, this simply returns the pyobject as-is; for other C++ type, the value goes
 // through pybind11::cast(obj) to convert it to an `object`.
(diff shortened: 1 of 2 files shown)


include/pybind11/pytypes.h
```diff
@@ -478,6 +478,42 @@ inline handle get_function(handle value) {
...
+    if (rv == NULL && PyErr_Occurred()) {
+        throw error_already_set();
+    }
+    return rv;
+#else
+    return PyDict_GetItemString(v, key);
+#endif
+}
+
+inline PyObject * dict_getitem(PyObject *v, PyObject *key)
```
bstaletic, line 504: Why isn't this a member of `py::dict`?

virtuald: Because the code that uses it uses the raw CPython API, presumably for performance reasons?

virtuald:
Also, if someone could retrigger the checks for this, it's likely that the tests will pass.

include/pybind11/pytypes.h
bstaletic: Fair enough.

include/pybind11/pytypes.h
```diff
@@ -478,6 +478,42 @@ inline handle get_function(handle value) {
...
+inline PyObject * dict_getitemstring(PyObject *v, const char *key)
+{
+#if PY_MAJOR_VERSION >= 3
+    PyObject *kv, *rv;
+    kv = PyUnicode_FromString(key);
+    if (kv == NULL) {
+        throw error_already_set();
+    }
+
+    rv = PyDict_GetItemWithError(v, kv);
```
bstaletic, line 493: At this point, can't you just call `dict_getitemstring`?

virtuald: Nope, still has potential to raise an exception. See https://bugs.python.org/issue35459

As copied from cpython @ https://github.com/python/cpython/blob/d5fc99873769f0d0d5c5d5d99059177a75a4e46e/Objects/dictobject.c#L1545

bstaletic: My bad, I was thinking about calling your `dict_getitem`. From this line on, `dict_getitemstring` and `dict_getitem` are the same.

virtuald: Not quite the same, Py_DECREF needs to be called on the key before the error is thrown.

include/pybind11/pytypes.h
```diff
@@ -478,6 +478,42 @@ inline handle get_function(handle value) {
     return value;
 }
 
+// Reimplementation of python's dict helper functions
+
+// copied from cpython _PyDict_GetItemStringWithError
+inline PyObject * dict_getitemstring(PyObject *v, const char *key)
```
EricCousineau-TRI, line 484: nit Can you summarize rationale for this? (and also for traceability?)
e.g. "We should use `*WithError` flavors to ensure we don't accidentally shadow other signals like KeyboardInterrupt (#2862)" or sthg like that? (either ref'ing your issue, or https://bugs.python.org/issue35459)?

EricCousineau-TRI (approved):
Looks great from my side! Just a minor nit on rationale comment. (Yeah, discoverable via `git blame` + clicks, but mayhaps a terse statement could help!)

rwgk:
Hi Dustin, sorry for the long delay! This PR escaped me before. It looks very important.
Could you please rebase this PR to trigger the CI? That will run new clang-tidy checks we added last week.

include/pybind11/pytypes.h
virtuald: Done.

rwgk:
Nice, thanks! Good to see that the new clang-tidy is happy.

rwgk:
I forgot to mention, I already integrated this PR for Google-internal global testing. I'll hold off merging until I get the results, later tonight if things go well.

rwgk:
The Google-internal testing was successful. Merging!

Issue #2860, opened 2021-02-12
Unable to do the difference between kind() and type() for dtype in numpy.h


bemichel:
Hi,
We are using pybind11 to give access to Python/C++ arrays in on a numerical project. For this, we are using the pybind11:array_t<T>.

But, we would like to have access to __type__ attribute in the __PyArray_Descr__ in API C or the __dtype.char__ on Python interface, but only the __kind()__ method is available in the class __pybin11::dtype__ in numpy.h.

The kind method gives only the "general" kind of the dtype, i.e. 
```c++
    auto float32 = py::dtype::of<float>();
    std::cout << "float32.kind() = " << float32.kind() << std::endl; /// return 'f'
    auto float64 = py::dtype::of<double>();
    std::cout << "float64.kind() = " << float64.kind() << std::endl; /// return 'f'
```
So it is not possible to distinguish __float__ from __double__, or __int__ from __long int__.

Only by adding a new method in the __pybind11::dtype::type()__ for exemple which returns the __type__ attribute from __PyArray_Descr__, like the following
```c++
class dtype : public object {
  ....
    /// Single-character for dtype's kind (ex: float and double are 'f' or int and long int are 'i')
    char kind() const {
        return detail::array_descriptor_proxy(m_ptr)->kind;
    }

    /// Single-character for dtype's type (ex: float is 'f' and double 'd')
    char type() const {
        return detail::array_descriptor_proxy(m_ptr)->type;
    }
private:
  ....
};
```
It will be now possible to distinguish the effective type 
```c++
    auto float32 = py::dtype::of<float>();
    std::cout << "float32.kind() = " << float32.kind() << std::endl; /// return 'f'
    std::cout << "float32.type() = " << float32.type() << std::endl; /// return 'f'
    auto float64 = py::dtype::of<double>();
    std::cout << "float64.kind() = " << float64.kind() << std::endl; /// return 'f'
    std::cout << "float64.type() = " << float64.type() << std::endl; /// return 'd'
```
Sorry, I am not sure if it is the good way to asking to add this little piece of code and/or if it is just possible ?
Should I create a merge request instead ?

Best regards,
Bertrand M.

YannickJadoul:
Hi, @bemichel. If you don't mind, a PR would indeed be easier. If not, let us know, and someone can do it.

By the way, in Python itself, it's `dtype.char`. So maybe we should adopt the same interface for `py::dtype`?

```py
>>> np.dtype(np.float32).char
'f'
[...]

PR #2864, opened 2021-02-17
Dtype kind vs char 

bemichel:
#2860  Description

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -507,21 +507,18 @@ public:
         return detail::array_descriptor_proxy(m_ptr)->names != nullptr;
     }
 
-    /**
-     * Single-character for dtype's kind
-     * ex: float and double are 'f' or int and long int are 'i',
-     *     i.e. "float and double" could be "floating point types",
-     *     and "int and long" could be "integral types")
-     */
+    /// Single-character code for dtype's kind.
+    /// For example, floating point types are 'f' and integral types are 'i'.
     char kind() const {
         return detail::array_descriptor_proxy(m_ptr)->kind;
     }
 
-    /** Single-character for dtype's type (ex: float is 'f' and double 'd').
-     *  NB: Following the Python API (i.e., dtype.char) is preferred here
-     *      rather than NumPy's internal C API (PyArray_Descr::type)
-     */
+    /// Single-character for dtype's type.
+    /// For example, ``float`` is 'f', ``double`` 'd', ``int`` 'i', and ``long`` 'd'.
     char char_() const {
+        // Note: The signature, `dtype::char_` follows the naming of NumPy's
+        // public Python API (i.e., ``dtype.char``), rather than its internal
+        // C API (``PyArray_Descr::type``).
         return detail::array_descriptor_proxy(m_ptr)->type;
     }
 
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -507,9 +507,14 @@ public:
         return detail::array_descriptor_proxy(m_ptr)->names != nullptr;
     }
 
-    /// Single-character type code.
+    /// Single-character for dtype's kind (ex: float and double are 'f' or int and long int are 'i')
     char kind() const {
         return detail::array_descriptor_proxy(m_ptr)->kind;
+    }
+
+    /// Single-character for dtype's type (ex: float is 'f' and double 'd')
+    char type() const {
+        return detail::array_descriptor_proxy(m_ptr)->type;
     }
 
 private:
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -513,7 +513,7 @@ public:
     }
 
     /// Single-character for dtype's type (ex: float is 'f' and double 'd')
-    char type() const {
+    char char_() const {
         return detail::array_descriptor_proxy(m_ptr)->type;
     }
 
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -507,12 +507,20 @@ public:
         return detail::array_descriptor_proxy(m_ptr)->names != nullptr;
     }
 
-    /// Single-character for dtype's kind (ex: float and double are 'f' or int and long int are 'i')
+    /**
+     * Single-character for dtype's kind
+     * ex: float and double are 'f' or int and long int are 'i',
+     *     i.e. "float and double" could be "floating point types",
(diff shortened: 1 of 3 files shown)


include/pybind11/numpy.h
```diff
@@ -507,11 +507,16 @@ class dtype : public object {
         return detail::array_descriptor_proxy(m_ptr)->names != nullptr;
     }
 
-    /// Single-character type code.
+    /// Single-character for dtype's kind (ex: float and double are 'f' or int and long int are 'i')
```
YannickJadoul, line 510: "float and double" could be "floating point types", and "int and long" could be "integral types". But this works as well.

include/pybind11/numpy.h
```diff
@@ -507,11 +507,16 @@ class dtype : public object {
...
     char kind() const {
         return detail::array_descriptor_proxy(m_ptr)->kind;
     }
 
+    /// Single-character for dtype's type (ex: float is 'f' and double 'd')
+    char char_() const {
+        return detail::array_descriptor_proxy(m_ptr)->type;
```
YannickJadoul, line 517: Is it worth adding a comment that we're following the Python API of `dtype` (i.e., `dtype.char`), rather than NumPy's internal C API (`PyArray_Descr::type`), for the name?

See also https://numpy.org/devdocs/reference/c-api/types-and-structures.html#c.PyArray_Descr for the root cause of the `type` vs. `char` confusion.

tests/test_numpy_dtypes.py
```diff
@@ -169,6 +169,9 @@ def test_dtype(simple_dtype):
         np.zeros(1, m.trailing_padding_dtype())
     )
 
+    assert m.test_dtype_kind() == ["b"] + ["i"] * 3 + ["f"] * 3
```
YannickJadoul, line 172: If it's up to me, I'd say to just spell it out, as it nicely matches with the list below.

If you really want to make things easy to write, then go for `list("biiifff")` and `list("?hilfdg")`. Just a suggestion. I'm not claiming that's better readable; just claiming it doesn't require opening and closing quotes and adding commas.

YannickJadoul (commented):
Nice, thanks, @bemichel! :-)

include/pybind11/numpy.h
bemichel: okay !

include/pybind11/numpy.h
bemichel: okay !

tests/test_numpy_dtypes.py
bemichel: okay !

bemichel:
Hi,
I do not understand why tests are not passed : it seems a cmake error in the report ???

YannickJadoul:
> Hi,
> I do not understand why tests are not passed : it seems a cmake error in the report ???

No worries. Our CMake is broken, but is getting fixed in #2865. I'll rerun the tests once that passes

bemichel:
Okay !
Let me known if this last commit satisfy your demand ?
Bertrand.

YannickJadoul:
Let's restart those tests :-)

Also, I had a few more proposed changes, but let me push them.

YannickJadoul (dismissed):
@bemichel, I still proposed a few changes in 400ffeaee0fe75d1e726093066ea6fab49fcc190, but if those are fine with you, I'm perfectly happy with this PR.

tests/test_numpy_dtypes.cpp
```diff
@@ -251,6 +251,38 @@ py::list test_dtype_ctors() {
...
+py::list test_dtype_char_() {
+    py::list list;
+    for (auto& dt : {
+                     py::dtype("bool8"),   // bool
+                     py::dtype("int16"),   // short
+                     py::dtype("int32"),   // int
+                     py::dtype("int64"),   // long int
+                     py::dtype("float32"), // float
+                     py::dtype("float64"), // double
+                     py::dtype("float128") // long double
```
bstaletic, line 279: On Winblows `sizeof(long double) == sizeof(double)`, which might explain the error.

bstaletic (commented):
@YannickJadoul I know you're not a fan of Windoze, but this PR breaks that OS.

tests/test_numpy_dtypes.cpp
YannickJadoul: I was about to say: MSVC doesn't really like this, @bemichel.

YannickJadoul: Also, those comments are not completely right: sometimes `np.int32 == np.int`, but sometimes `np.int64 == np.int`. It really depends on the platform, I think?

But then, as far as I remember (I once contributed to this part of NumPy's docstrings in a sprint at a conference), the letters are actually linked to `np.bool`, `np.int`, np.float`, etc, and not to the sized variants!

~Scratch that. Seems that either they changed, or I misremember.~

No, I was right. But `np.int` is an alias for a Python `int`, not for a C int, that's `np.intc`.

YannickJadoul: At any rate, here are the docs:
https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types

`intN` or `floatN` are aliases, so I'd suggest using the non-alias versions.

YannickJadoul: Actually. It's easier if I just try fixing this. Let's see.

YannickJadoul: Done. Let's see what this says and does.

YannickJadoul:
OK, things are working. Want to check if you agree with the changes I made, @bemichel?

YannickJadoul (approved):

bemichel:
> @bemichel, I still proposed a few changes in [400ffea](https://github.com/pybind/pybind11/commit/400ffeaee0fe75d1e726093066ea6fab49fcc190), but if those are fine with you, I'm perfectly happy with this PR.

>>> Perfect !

bemichel:
I agree with everything !!! Great idea to factorize the list of "C/C++" types.
Bertrand.

YannickJadoul:
Great; let's see who else will review then, before this is done :-)

bstaletic (approved):

YannickJadoul:
> Great; let's see who else will review then, before this is done :-)

Seems that was @bstaletic :-)

Thanks, @bemichel!

PR #2870, opened 2021-02-23
Use correct duration representation when casting from datetime.timdelta to std::chrono::duration

### LINKED ISSUE #2861: [QUESTION] Chrono bindings and negative datetime.timedelta
Hi,

I was looking at the type caster for datetime.timedelta --> steady_clock::duration and thought that the choice of uint_fast32_t as `rep` for `days` was a bit odd.

https://github.com/pybind/pybind11/blob/417067eeb8d93e8ed49f3ff580712ae2b9a85259/include/pybind11/chrono.h#L38

In the python3 docs (https://docs.python.org/3/library/datetime.html#datetime.timedelta.min), it says that timedelta.min is -999999999 days and timedelta.max is 999999999 days. Neither of these limits are representible in a 32bit unsigned integer.

Looking at the source of CPython they store days in an `int` (which also seems odd).

https://github.com/python/cpython/blob/cc54001c2eb3b14320c1667b22602d69c90d5865/Include/datetime.h#L38

This looks to me like if I pass `timedelta(days=-1)` or `timedelta(seconds=-1)` (which then turns into  `timedelta(days=-1, seconds=86399`) to a C++ function that expects chrono::nanoseconds, pybind11 will do some signed --> unsigned --> signed math under the hood? Is this UB or fine?

So questions:

- Is it fine to use uint_fast32_t here? Should int64_t be used instead?
- Am I misunderstanding the use of std::ratio? i.e. does that struct have an impact on the correctness of the choice for `days`?
- Should there be more tests with negative time deltas? (probably yes :) ).

bstaletic:
## Description

this is an attempt to fix #2861 

## Suggested changelog entry:

```rst
Allow negative timedelta values to roundtrip.
```

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -35,7 +35,7 @@ public:
     using rep = typename type::rep;
     using period = typename type::period;
 
-    using days = std::chrono::duration<uint_fast32_t, std::ratio<86400>>;
+    using days = std::chrono::duration<int_least32_t, std::ratio<86400>>; // signed 25 bits required by the standard.
 
     bool load(handle src, bool) {
         using namespace std::chrono;
--- a/tests/test_chrono.py
+++ b/tests/test_chrono.py
@@ -39,9 +39,7 @@ def test_chrono_system_clock_roundtrip():
 
     # They should be identical (no information lost on roundtrip)
     diff = abs(date1 - date2)
-    assert diff.days == 0
-    assert diff.seconds == 0
-    assert diff.microseconds == 0
+    assert diff == datetime.timedelta(0)
 
 
 def test_chrono_system_clock_roundtrip_date():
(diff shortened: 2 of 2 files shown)


YannickJadoul (approved):

YannickJadoul:
For the record and future reference: #2861 discusses all of this in great detail, why the current approach is an issue and how this simple change fixes things (and should cause any further issues?).

rwgk:
LGTM and I ran this through our Google-global testing (just be cause it's easy and highly conclusive). Thanks @bstaletic!

knzivid:
@YannickJadoul Which next release version is this PR going into?

rwgk:
This was merged into master. We're aiming for a 2.7 release (sorry the dev tag is currently misleading). We don't have a fixed date for the next release. I do not think the internals version will increase. You should be able to use current `master` in combination with other pybind11 extensions built against `stable`. `master` is extremely well tested these days.

PR #2878, opened 2021-02-27
Enabling use of smart_holder for types with non-public destructors.

rwgk:
This change is needed for https://github.com/google/clif/blob/7df2a5784c414be1c0f242470ce86506b7efe1d3/clif/pybind11/staging/smart_ptrs.cc#L81.


rwgk:
Merging for easier Google-internal testing. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2880, opened 2021-02-27
Changing cast_error("... Python instance was disowned.") to value_error.

rwgk:
Changes `RuntimeError` to `ValueError` in the Python interpreter.

`ValueError` is deemed more fitting and intuitive, and is compatible with established PyCLIF behavior.


rwgk:
Merging for easier Google-internal testing. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2879, opened 2021-02-27
define PYBIND11_USE_SMART_HOLDER_AS_DEFAULT

rwgk:
IGNORE this PR please.

This is purely for running the CI against branch `smart_holder` (or other branches based on branch `smart_holder), with the only difference that PYBIND11_USE_SMART_HOLDER_AS_DEFAULT is defined.

PR #2882, opened 2021-03-03
shared_ptr<bool> vptr_deleter_armed_flag_ptr (instead of unique_ptr)

rwgk:
Fixes heap-use-after-free bug.

This fix comes at the cost of using `shared_ptr<bool>` instead of `unique_ptr`.

AddressSanitizer errors for the two newly added tests, _before_ changes to `smart_holder_poc.h`:
```
SUMMARY: AddressSanitizer: heap-use-after-free include/pybind11/detail/smart_holder_poc.h:68:13 in void pybindit::memory::guarded_builtin_delete<int>::operator()<int, 0>(int*)
SUMMARY: AddressSanitizer: heap-use-after-free include/pybind11/detail/smart_holder_poc.h:86:13 in pybindit::memory::guarded_custom_deleter<int, helpers::functor_builtin_delete<int> >::operator()(int*)
```

With the changes to `smart_holder_poc.h` the AddressSanitizer errors disappear.

For the record, commands used for ASAN testing (64-bit Debian-like system):
```
clang++ -fsanitize=address -std=c++11 -O0 -g -Wall -Wextra -Wconversion -Wcast-qual -Wdeprecated -Wnon-virtual-dtor -I./include -I../../clone/Catch2/single_include/catch2 -I/usr/include/python3.8 ./tests/pure_cpp/smart_holder_poc_test.cpp
./a.out
```

Additionally ASAN-tested in the Google build environment.

### DIFF
--- a/include/pybind11/detail/smart_holder_poc.h
+++ b/include/pybind11/detail/smart_holder_poc.h
@@ -104,10 +104,10 @@ struct smart_holder {
     bool is_populated : 1;
 
     // Design choice: smart_holder is movable but not copyable.
-    smart_holder(smart_holder &&other) = default;
+    smart_holder(smart_holder &&) = default;
     smart_holder(const smart_holder &) = delete;
-    smart_holder &operator=(smart_holder &&other) = default;
-    smart_holder &operator=(const smart_holder &other) = delete;
+    smart_holder &operator=(smart_holder &&) = default;
+    smart_holder &operator=(const smart_holder &) = delete;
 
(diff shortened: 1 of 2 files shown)


rwgk:
The CI is green, also with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` (#2879): Merging for easier Google-internal testing. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2883, opened 2021-03-03
Changing all but one std::runtime_error to std::invalid_argument.

rwgk:
`std::invalid_argument` appears as `ValueError` in the Python interpreter, which is more intuitive than `RuntimeError`, and compatible with existing PyCLIF behavior.


rwgk:
The CI is green, also with PYBIND11_USE_SMART_HOLDER_AS_DEFAULT (#2879): Merging for easier Google-internal testing. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2895, opened 2021-03-09
Updating version number to reflect current plans.

rwgk:


rwgk:
This PR is very obsolete. — We're on track for 2.7.0 but we will update the version number a different way.

PR #2902, opened 2021-03-14
Adding virtual_overrider_self_life_support.

rwgk:
Enables safely passing `unique_ptr` from Python to C++, even for trampolines.

A disowned trampoline Python object is kept it alive until the  `unique_ptr` pointee is destroyed. The critical small trick needed was found in PyCLIF (probably invented by @mrovner):

https://github.com/google/clif/commit/acacae6728810dd67df57157c337963297a2e1eb

The critical part of the trick is: the `Py_INCREF` is triggered only when the trampoline Python object is in the process of being disowned.

This PR goes beyond PyCLIF, in that the disowned Python object still maintains a pointer that is only reset and deregistered when the `unique_ptr` pointee is deleted, which means methods of the pointee can still be safely accessed from Python. This could be viewed as "lazy disowning".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ set(PYBIND11_HEADERS
     include/pybind11/detail/smart_holder_type_casters.h
     include/pybind11/detail/type_caster_base.h
     include/pybind11/detail/typeid.h
+    include/pybind11/detail/virtual_overrider_self_life_support.h
     include/pybind11/attr.h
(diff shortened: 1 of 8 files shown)


rwgk:
The CI is green, also with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` (#2879). Google-internal testing is also green. Merging here to unblock follow-on PyCLIF work. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

Issue #2901, opened 2021-03-14
smart_holder: open issues


rhaschke:
When using @rwgk's `smart_holder` branch and providing a custom holder type like this:
```cpp
py::class_<MyClass, std::shared_ptr<MyClass>, MyBaseClass>(m, "MyClass")
		.def(py::init<>());
```
I've got a misleading static assert:
```
pybind11/include/pybind11/pybind11.h:1297:13: error: static assertion failed: py::class_ holder vs type_caster mismatch: missing PYBIND11_SMART_POINTER_HOLDER_TYPE_CASTERS(T, std::shared_ptr<T>)?
 1297 |             !(detail::is_instantiation<std::shared_ptr, holder_type>::value
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1298 |               && wrapped_type_uses_smart_holder_type_caster),
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I would assume: 
1. That it is possible to define custom holders
2. If multiple holder types are defined, a more informative assert would be thrown

rwgk:
Hi Robert, are you using `-DPYBIND11_USE_SMART_HOLDER_AS_DEFAULT`?
Or are you using the `PYBIND11_SMART_HOLDER_TYPE_CASTERS(MyClass)`?
If that's not it, could you please share the code that fails building?

rhaschke:
I'm using `-DPYBIND11_USE_SMART_HOLDER_AS_DEFAULT`. I successfully migrated my code to the `smart_holder` branch - at least everything compiles. See [here](https://github.com/ubi-agni/moveit_task_constructor/commit/7d50093a5ecb261bd1619b7576311c0358840deb) for required changes:
- explicitly specifying a custom holder type doesn't work anymore
- (reference) arguments of `std::shared_ptr<T>&` don't work anymore
- `make_iterator` doesn't work for a container (`std::list`) of `std::unique_ptr<T>`

rwgk:
> * explicitly specifying a custom holder type doesn't work anymore

It does :-) I tried to be really careful about. What you need is actually in the `static_cast` message (just with `T` instead of your typename):
```
PYBIND11_SMART_POINTER_HOLDER_TYPE_CASTERS(MyClass, std::shared_ptr<MyClass>)
```
[...]

PR #2904, opened 2021-03-16
Missing cast from const unique_ptr&

rhaschke:
The issue mentioned in #2901 (the compiler complaining about impossible casting from `const unique_ptr&` to `unique_ptr&&`) originates from a missing cast implementation. I added a dummy implementation (69769a9) and the compiler errors are gone.

IMO, casting from a `const unique_ptr&` should distinguish the following two cases:
1. If the instance is already registered, it should be returned with an increased refcount.
2. Otherwise, a new instance should be created (which holder?), which must not be deleted or moved. Is that possible at all?

### DIFF
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -716,6 +716,15 @@ struct smart_holder_type_caster<std::unique_ptr<T, D>> : smart_holder_type_caste
 
         return inst.release();
     }
+    static handle cast(const std::unique_ptr<T, D> &src, return_value_policy policy, handle parent) {
+        if (!src)
+            return none().release();
+        if (policy == return_value_policy::automatic)
+            policy = return_value_policy::reference_internal;
+        if (policy != return_value_policy::reference_internal)
+            throw cast_error("Invalid return_value_policy for unique_ptr&");
+        return smart_holder_type_caster<T>::cast(src.get(), policy, parent);
+    }
 
     template <typename>
     using cast_op_type = std::unique_ptr<T, D>;
--- a/tests/test_class_sh_basic.cpp
+++ b/tests/test_class_sh_basic.cpp
@@ -17,6 +17,17 @@ struct atyp { // Short for "any type".
     atyp(atyp &&other) { mtxt = other.mtxt + "_MvCtor"; }
 };
 
+struct uconsumer {  // unique_ptr consumer
+    std::unique_ptr<atyp> held;
+    bool valid() const { return static_cast<bool>(held); }
+
+    void pass_valu(std::unique_ptr<atyp> obj) { held = std::move(obj); }
+    void pass_rref(std::unique_ptr<atyp> &&obj) { held = std::move(obj); }
+    std::unique_ptr<atyp> rtrn_valu() { return std::move(held); }
+    std::unique_ptr<atyp>& rtrn_lref() { return held; }
+    const std::unique_ptr<atyp> &rtrn_cref() { return held; }
+};
+
 // clang-format off
 
 atyp        rtrn_valu() { atyp obj{"rtrn_valu"}; return obj; }
@@ -57,7 +68,11 @@ std::string pass_udcp(std::unique_ptr<atyp const, sddc> obj) { return "pass_udcp
 
 // Helpers for testing.
 std::string get_mtxt(atyp const &obj) { return obj.mtxt; }
+std::ptrdiff_t get_ptr(atyp const &obj) { return reinterpret_cast<std::ptrdiff_t>(&obj); }
+
 std::unique_ptr<atyp> unique_ptr_roundtrip(std::unique_ptr<atyp> obj) { return obj; }
+const std::unique_ptr<atyp>& unique_ptr_cref_roundtrip(const std::unique_ptr<atyp>& obj) { return obj; }
+
 struct SharedPtrStash {
     std::vector<std::shared_ptr<const atyp>> stash;
     void Add(std::shared_ptr<const atyp> obj) { stash.push_back(obj); }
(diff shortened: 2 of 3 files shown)


include/pybind11/detail/smart_holder_type_casters.h
```diff
@@ -698,6 +698,9 @@ struct smart_holder_type_caster<std::unique_ptr<T, D>> : smart_holder_type_caste
 
         return inst.release();
     }
+    static handle cast(const std::unique_ptr<T, D> &src, return_value_policy policy, handle parent) {
```
rwgk, line 701: Aha :-)
That makes everything clear immediately, thanks!
Initial thought: This is very similar to `static handle cast(const T *src, ...)`. We could mimic that implementation, or maybe even reuse? The easy case is that an instance exists already and we just incref as you suggest. Or there needs to be `return_value_policy::reference_internal` or an explicit `keep_alive`.
Is there a chance you could help with that implementation? It's unlikely that I can prioritize working on this myself anytime soon, unless I hit on a use case in PyCLIF (difficult to predict). But I will review & merge very quickly if you take the lead.
My revised suggestion for a separate new unit test name *iff* we cannot reuse `static handle cast(const T *src, ...)`: test_class_sh_const_unique_return
Keeping the tests separate helps minimizing merge issues (between master, drake, smart_holder) and is best practice IMO regardless.

rhaschke: Done. Thanks for the hint to `cast(const T*)`. Regarding the unit test: I think we should have a unit test in any case.
Do you suggest to move the test into a new file, named `test_class_sh_const_unique_return.cpp` or to include a test named `test_class_sh_const_unique_return` into some existing file?

rhaschke: Hm. I don't understand why CI fails, while locally it works fine. Need to investigate later...

rwgk: Oh, sorry to be ambiguous: I was thinking of `test_class_sh_const_unique_return.cpp` and `.py`.
I think I also want to add a few lines to `test_class_sh_basic.cpp` and `.py`, but that's a small enough effort I could squeeze that in myself.

include/pybind11/detail/smart_holder_type_casters.h
```diff
@@ -698,6 +698,15 @@ struct smart_holder_type_caster<std::unique_ptr<T, D>> : smart_holder_type_caste
...
+        if (!src)
+            return none().release();
+        if (policy == return_value_policy::automatic)
+            policy = return_value_policy::reference_internal;
+        if (policy != return_value_policy::reference_internal)
+            throw cast_error("Invalid return_value_policy for unique_ptr&");
+        return type_caster_base<T>::cast(src.get(), policy, parent);
```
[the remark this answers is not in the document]
rhaschke: Sure. I still do need to familiarize myself with your new code...

rwgk:
I looked at one of the CI failures: https://github.com/pybind/pybind11/pull/2904/checks?check_run_id=2120821136
It looks like it's somehow falling back to `move_only_holder_caster`, which shouldn't be involved at all for types using the `smart_holder_type_caster` family.

There is a lot going on in test_smart_ptr.cpp, especially with the TYPE_CASTERS macros, which could be easily confusing. If you move the new test to a separate pair of files we have a simpler context to worry about.

About `test_smart_ptr` in general: it was the most special test for me to deal with. From memory, I think none of the `py::class_` in there uses `smart_holder`, even if you build with `ENABLE_SMART_HOLDER_AS_DEFAULT`.

My general approach for the existing tests was: if a holder is specified explicitly, keep it that way, with or without `ENABLE_SMART_HOLDER_AS_DEFAULT`. If a derived `py::class_` uses a custom holder, make the `std::unique_ptr` holder for the base explicit. — That way I could be sure that I'm not trying to run tests designed specifically for the old holder handling for the new `smart_holder` code. The `smart_holder` code has independent dedicated tests in the `test_class_sh_*` files.

rhaschke:
Thanks for these hints. I will move my tests into `test_class_sh_*` files. I also noticed that I should use `py::classh` instead of `py::class_` in order to get smart holders. The CI failures are probably due to the fact that CI doesn't use `ENABLE_SMART_HOLDER_AS_DEFAULT` but requires manual `smart_holder` triggering via `py::classh`.

rwgk:
> The CI failures are probably due to the fact that CI doesn't use ENABLE_SMART_HOLDER_AS_DEFAULT but requires manual smart_holder triggering via py::classh.

Oh yes, that's another stumbling stone. I have PR #2879 for that purpose. I keep reusing it, by rebasing it on whatever I want to test at the moment.

rwgk:
Hi Robert, I just merged #2902 into the smart_holder branch. You may want to rebase this PR.

rhaschke:
This should be ready for a fresh review. I added several new test cases, [one of them still failing](
https://github.com/pybind/pybind11/pull/2904/commits/90561ad9daf52a460daa8bd2d6455877440ad6a9#diff-e614dc3a68da8c26363419d3c06816825b1b4344e35103c5dd004f7dcf3a6f84R125). Particularly, this roundtrip test attempts to pass a `unique_ptr` from python to C++ as a `const reference` and returning the very same reference to Python again. IMO, this should leave the original python instance alive and just increase the reference count. However, the instance is moved to C++ and thus disowned in Python.

rwgk:
> IMO, this should leave the original python instance alive and just increase the reference count. However, the instance is moved to C++ and thus disowned in Python.

I think (!) you could leverage the [`dynamic_cast`](https://github.com/pybind/pybind11/blob/469792032a779119e2771d914ba01f828a355110/include/pybind11/detail/smart_holder_type_casters.h#L394) that I just merged (#2902) to achieve that for types that inherit from `virtual_overrider_self_life_support`, but without something like that I think it's not possible: when you get the same pointer value back, you don't know if it's still the same object, or a different newly allocated one that just happens to have the same address.

rhaschke:
CI still failing, because windows moves more often... Will fix.

rhaschke:
CI passes! :tada:

rwgk:
Thanks Robert! I'll merge this now.

PR #2906, opened 2021-03-17
Issue: object ref-count ignored when moving std::unique_ptr

rhaschke:
Inspecting the unit tests of #2904 in more detail, I noticed that the smart holder code is actually ignoring an object's ref-count, although there is this test: 
https://github.com/pybind/pybind11/blob/784092dfd29d861af6fa611c6920c172389d97c4/include/pybind11/detail/smart_holder_type_casters.h#L380

Nevertheless, as shown in [this example](https://github.com/pybind/pybind11/compare/smart_holder...rhaschke:debug-smart-holder?expand=1#diff-e614dc3a68da8c26363419d3c06816825b1b4344e35103c5dd004f7dcf3a6f84R153-R156), the refcount of the object before `moving` it, is 3! Something weird is going on there. Moving should fail in this case, shouldn't it? @rwgk 

Having that said, I actually want to have the possibility to move an object with a refcount > 1. Otherwise, I couldn't move a variable, but only "rvalues". There are extra checks in place to ensure that I don't access a disowned object again...

### DIFF
--- a/tests/test_class_sh_basic.py
+++ b/tests/test_class_sh_basic.py
@@ -3,6 +3,7 @@
 import re
 
 import pytest
+import sys
 
 from pybind11_tests import class_sh_basic as m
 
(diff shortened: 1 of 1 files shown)


rhaschke:
Ah. Ok. Sorry for the noise.

PR #2908, opened 2021-03-17
Adding clang-format to pre-commit checks.

rwgk:
Adding `clang-format-lint` to Format workflow, and updating clang-format section in .github/CONTRIBUTING.md.

This PR has two commits, PLEASE DO NOT SQUASH when merging.

The first commit was generated automatically to systematically insert `//clang-format off` in all 75 files that currently generate clang-format errors. Please review only one file. Requests for changes will be systematically reapplied.

The second commit is for manual changes to .pre-commit-config.yaml and .github/CONTRIBUTING.md.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -98,3 +98,15 @@ repos:
     types:
     - c++
     entry: ./tools/check-style.sh
+
+- repo: local
+  hooks:
+  - id: docker-clang-format
+    name: Docker Clang Format
+    language: docker_image
+    types:
+    - c++
+    entry: unibeautify/clang-format:latest
+    args:
+    - -style=file
+    - -i
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,12 @@
 #
 # See https://github.com/pre-commit/pre-commit
 
+# If we enable pre-commit.ci, this will be useful:
+ci:
+  skip:
+  - docker-clang-format
+  # check-style maybe too?
+
 repos:
 # Standard hooks
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1,3 +1,7 @@
+// clang-format off
+// Please consider `clang-format`ting this file before starting with significant
+// changes, ideally in a separate commit. See also: .github/CONTRIBUTING.md
+
 /*
     pybind11/cast.h: Partial template specializations to cast between
     C++ and Python types
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -180,9 +180,13 @@ pre-commit install
 
 As of v2.6.2, pybind11 ships with a [`clang-format`][clang-format]
 configuration file at the top level of the repo (the filename is
-`.clang-format`). Currently, formatting is NOT applied automatically, but
-manually using `clang-format` for newly developed files is highly encouraged.
-To check if a file needs formatting:
+`.clang-format`). Formatting is checked automatically as part of
+`pre-commit run`, but many source files were grandfathered in by adding
+`// clang-format off` at the top. Please consider `clang-format`ing these
+files before starting with major changes, ideally in a separate commit, or even
+a separate PR, to keep development changes separate from `clang-format` cleanup.
+
+To quickly check if a certain file needs formatting:
 
 ```bash
 clang-format -style=file --dry-run some.cpp
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -1,3 +1,7 @@
+// clang-format off
+// Please consider `clang-format`ting this file before starting with significant
+// changes, ideally in a separate commit. See also: .github/CONTRIBUTING.md
+
 /*
     pybind11/attr.h: Infrastructure for processing custom
     type and function attributes
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -1,3 +1,7 @@
+// clang-format off
+// Please consider `clang-format`ting this file before starting with significant
+// changes, ideally in a separate commit. See also: .github/CONTRIBUTING.md
+
 /*
     pybind11/buffer_info.h: Python buffer object interface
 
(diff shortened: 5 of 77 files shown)


henryiii:
Why is this not a pre-commit? I thought we were unifying all style checks, this is fragmenting them again. Developers now have to know how to run clang-format and pre-commit. https://github.com/scikit-hep/boost-histogram/blob/3dc87dd53267ca7cc18cdaf7dc570ee362607275/.pre-commit-config.yaml#L92-L102

rwgk:
> Why is this not a pre-commit? I thought we were unifying all style checks, this is fragmenting them again. Developers now have to know how to run clang-format and pre-commit. https://github.com/scikit-hep/boost-histogram/blob/3dc87dd53267ca7cc18cdaf7dc570ee362607275/.pre-commit-config.yaml#L92-L102

Oh, that's much better! Sorry I didn't even know that .pre-commit-config.yaml exists. Thanks for the link. I copy-pasted the config block and it seems to work as-is. I updated CONTRIBUTING.md accordingly. Please take another look.

.pre-commit-config.yaml
```diff
@@ -98,3 +104,15 @@ repos:
...
+  hooks:
+  - id: docker-clang-format
+    name: Docker Clang Format
+    language: docker_image
+    types:
+    - c++
+    entry: unibeautify/clang-format:latest
+    args:
+    - -style=file
+    - -i
```
rhaschke, line 118: It is not necessary to run clang-format within a docker container. In [MoveIt we are using it directly](https://github.com/ros-planning/moveit/blob/master/.pre-commit-config.yaml#L35-L43) like this:
```suggestion

- repo: local
  hooks:
      - id: clang-format
        name: clang-format
        description: Format files with clang-format.
        entry: clang-format-10
        language: system
        files: \.(c|cc|cxx|cpp|frag|glsl|h|hpp|hxx|ih|ispc|ipp|java|js|m|proto|vert)$
        args: ['-fallback-style=none', '-i']
```
This just requires installing `clang-format` before running the `pre-commit` action:
https://github.com/ros-planning/moveit/blob/master/.github/workflows/format.yml#L18-L19

Note, that the different versions of `clang-format` are not interoperable. You need to choose (and stick to) a particular version.

.pre-commit-config.yaml
```diff
@@ -12,6 +12,12 @@
 #
 # See https://github.com/pre-commit/pre-commit
 
+# If we enable pre-commit.ci, this will be useful:
+ci:
+  skip:
+  - docker-clang-format
+  # check-style maybe too?
+
```
rhaschke, line 20: Why would you want to skip clang-format for pre-commit.ci? 
If I understand correctly, you are not yet using `pre-commit.ci`, so it doesn't yet matter.

rhaschke (commented):
I would prefer running clang-format directly, not via docker. Is there a reason, you proposed the docker solution, @henryiii?

.pre-commit-config.yaml
henryiii: The docker solution a) works everywhere Docker does, b) doesn't require special pre-installs locally (which is one of the main advantages of pre-commit), and c) ensures the same version is used everywhere, rather than depending on what a user happened to pre-install.

Now, what you _can_ do (and I do have for https://github.com/cliutils/cli11, for example) is two pre-commit files. If you can't run docker locally but don't mind installing the correct version of clang-format, you can specifically ask for the "no-docker" version. But it's a bit of a pain to keep them in sync, and most people just use the docker one because it's simpler.

.pre-commit-config.yaml
henryiii: Because it runs in a docker container, so can't run docker. It doesn't have the correct version (or any version) of pre-commit preinstalled, the local one wouldn't help either.

No, we don't have it yet, but maybe someday. It would auto-update the pins, and apply Black changes automatically to PRs.

.pre-commit-config.yaml
rhaschke: So, you then manually configure your local pre-commit git-hook to use the `no-docker` version?

henryiii: Yes. Well, I don't, I use docker, but I had one person request this. (I think they have since moved to docker)

EricCousineau-TRI:
This PR seems that it's in a draft state, with some checkboxes being implicit.

@rwgk When you have time, can you add those checkboxes? (e.g. ``[ ] Review infrastructure, [ ] Preliminary reformatting review (`pybind11.h`), [ ] Apply incrementally to remaining files``)

include/pybind11/cast.h
```diff
@@ -1,3 +1,7 @@
+// clang-format off
+// Please consider `clang-format`ting this file before starting with significant
```
EricCousineau-TRI, line 2: Objective is this statement isn't completely clear (could be made more concise).

Is this a "DNM" (Do Not Merge) for this PR, or is this something that would get merged in mainline?
If it's draft for this PR, consider reflecting that?
```
// DNM: This will be reformatted as part of this PR.
```

If it's intended to be merged to mainline, consider:
```
// Please consider `clang-format`ting any new lines added when submitting a PR.
```

However, TBH, I'm not sure how smooth of a process that will be?

rwgk: What I had in mind is explained in CONTRIBUTING.md. I was hoping people look there as suggested by the `See also` in the next line here. This text from CONTRIBUTING.md seemed too long to duplicate at the top of all files:

> Please consider `clang-format`ing these files before starting with major changes, ideally in a separate commit, or even a separate PR, to keep development changes separate from `clang-format` cleanup.

Formatting only new or modified lines is another option, but I didn't want to make the explanations too complicated, and I'm also ambivalent about the resulting patchwork of `// clang-format on/off` directives.

I think the healthiest option is to globally clang-format everything in one big PR that does nothing else, but I created this PR as a minimalistic alternative because Wenzel didn't seem supportive. Wenzel doesn't even seem to support this one.

rwgk:
I'm closing this PR to avoid confusion going forward: based on discussions with @EricCousineau-TRI, @henryiii, @Skylion007, I'll work on a different one that applies clang-format universally and automatically as part of the pre-commit. (Merging to be deferred until after the 2.7 release.)

Skylion007:
Yeah, happy to help with that once the next version releases.

PR #2910, opened 2021-03-18
Using `dynamic_cast<AliasType>` to determine `pointee_depends_on_holder_owner`.

rwgk:
This PR is a follow-on to PR #2902, implementing a refinement of a safety guard that previously prevented not only unsafe behavior, but also safe behavior. All relevant situations are exercised by the newly added test_class_sh_virtual_py_cpp_mix.py. Specifically, e.g. `m.get_from_cpp_unique_ptr(m.Base())` fails before this PR.

The critical change in this PR is in smart_holder_type_casters.h:
```diff
-        v_h.holder<holder_type>().pointee_depends_on_holder_owner = has_alias;
+        v_h.holder<holder_type>().pointee_depends_on_holder_owner
+        = dynamic_raw_ptr_cast_if_possible<AliasType>(v_h.value_ptr<WrappedType>()) != nullptr;
```

`has_alias` is true for any instance of a `py::class_` with an alias class (aka trampoline). E.g. it is true for `m.Base()` and `PyBase()` in test_class_sh_virtual_py_cpp_mix.py. The newly used `dynamic_cast`-based test is true only for `PyBase()`.

The idea for using a `dynamic_cast` here goes back to: https://github.com/pybind/pybind11/blob/0e01c243c7ffae3a2e52f998bacfe82f56aa96d9/include/pybind11/detail/init.h#L49

The exception message for the safety guard (also in smart_holder_type_casters.h) was made more precise and helpful:
```diff
         if (self_life_support == nullptr && holder().pointee_depends_on_holder_owner) {
-            throw value_error("Ownership of instance with virtual overrides in Python cannot be "
-                             "transferred to C++.");
+            throw value_error("Alias class (also known as trampoline) does not inherit from "
+                              "py::virtual_overrider_self_life_support, therefore the "
+                              "ownership of this instance cannot safely be transferred to C++.");
         }
```

Not directly related to the behavior change, `py::detail::virtual_overrider_self_life_support` was moved to `py::virtual_overrider_self_life_support`, to reflect that it is part of the public interface.

The behavior change under this PR has an effect on test_class_sh_trampoline_shared_ptr_cpp_arg.py, which was adopted from PR #2839 previously. `test_shared_ptr_alias_nonpython` is now in line with the behavior under PR #2839, but `test_shared_ptr_arg_identity` behaves differently here. This may need another look, but it appears that the `has_alias/is_alias` logic under PR #2839 may need a refinement similar to what is implemented in this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ set(PYBIND11_HEADERS
     include/pybind11/detail/class.h
     include/pybind11/detail/common.h
     include/pybind11/detail/descr.h
+    include/pybind11/detail/dynamic_raw_ptr_cast_if_possible.h
     include/pybind11/detail/init.h
     include/pybind11/detail/internals.h
     include/pybind11/detail/smart_holder_poc.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,6 @@ set(PYBIND11_HEADERS
     include/pybind11/detail/smart_holder_type_casters.h
     include/pybind11/detail/type_caster_base.h
     include/pybind11/detail/typeid.h
-    include/pybind11/detail/virtual_overrider_self_life_support.h
     include/pybind11/attr.h
     include/pybind11/buffer_info.h
     include/pybind11/cast.h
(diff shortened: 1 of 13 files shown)


rwgk:
The CI is green, also with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` (#2879). Google-internal testing is also green. Merging here to unblock follow-on PyCLIF work. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2911, opened 2021-03-20
Fix argument passing from trampoline methods

rhaschke:
This adds some more `smart_holder` tests that mimic some of my (currently failing) application use cases. 
Particularly, they consider _roundtrip_ situations where an object is passed around as a reference: 
- Python -> C++ -> Python
- C++ -> Python -> C++

The single failing test in [`test_unique_ptr_cref_consumer_roundtrip`](https://github.com/rhaschke/pybind11/blob/315625340a9dfcf25363faf1b4cac48dfc7222e5/tests/test_class_sh_basic.py#L165-L173) and [`test_unique_ptr_cref_roundtrip`](https://github.com/rhaschke/pybind11/blob/315625340a9dfcf25363faf1b4cac48dfc7222e5/tests/test_class_sh_basic.py#L184-L185) aren't important. I added them just for completeness. However, [`test_unique_ptr_consumer_roundtrip`](https://github.com/rhaschke/pybind11/blob/bbe45d15bfd1e2446a1f6efcf7be28982c0476e6/tests/test_class_sh_with_alias.py#L86-L88) is. For this latter test, I didn't look into the reason yet. Will continue tomorrow...

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -434,7 +434,11 @@ enum class return_value_policy : uint8_t {
         collected while Python is still using the child. More advanced
         variations of this scheme are also possible using combinations of
         return_value_policy::reference and the keep_alive call policy */
-    reference_internal
+    reference_internal,
+
+    /*  This internally-only used policy applies to C++ arguments passed
+        to virtual methods overridden in Python to allow reference passing. */
+    automatic_override
 };
 
 PYBIND11_NAMESPACE_BEGIN(detail)
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -731,6 +731,9 @@ struct smart_holder_type_caster<std::unique_ptr<T, D>> : smart_holder_type_caste
     template <typename>
     using cast_op_type = std::unique_ptr<T, D>;
 
+    // TODO: This always returns a new, moving unique_ptr instance to the raw pointer,
+    // even if argument should be passed as reference.
+    // See test_class_sh_basic.py::test_unique_ptr_cref_roundtrip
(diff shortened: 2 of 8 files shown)


rhaschke:
The culprit was (as coarsely expected) a wrong `return_value_policy` for passing arguments from C++ to python in calls to overridden class methods. I have implemented two alternative implementations for this:
1. Introduce a new return policy `reference_override` that yields `reference` behavior in most cases, but switches to copy in case of `const reference` arguments. Otherwise, behavior (hopefully) isn't changed.
2. I think we should actually enforce constness of C++ references/pointers via copying, even if users explicitly provide another return policy. Doing so, we can drop the introduction of the new return policy.

rhaschke:
I think this is ready for review. However, I'm not sure about the [failing valgrind test](https://github.com/pybind/pybind11/runs/2159050867?check_suite_focus=true). Looks like it fails in https://github.com/pybind/pybind11/blob/c50f90eca6cee9d665913f56a8f3d00591b262ff/tests/test_gil_scoped.py#L22
But, I don't understand why it fails for the valgrind test only. Does it run into the 10s timeout? Is it just flaky behavior?

rhaschke:
To track down the CI failure, I will perform a git bisect using github actions for validation. Don't wonder...

rhaschke:
This PR is failing the unit tests, but only with the valgrind config. I have no idea why. The first failing commit is https://github.com/pybind/pybind11/pull/2911/commits/4b0b8ae559232f6ccfd13b8bf7a6f4ac12bc19dd.
@YannickJadoul, did you originally introduce the valgrind config? Do you have any idea, why `test_gil_scoped.py` fails in this situation? Is valgrind too slow?

rhaschke:
Superseeded by #2916.

PR #2912, opened 2021-03-21
Adding tests to exercise corner cases involving disowning.

rwgk:
The most interesting corner case is in test_class_sh_disowning.py `test_mixed`. See comments there. A "maybe later" idea:

* For each overloaded function considered, in the first pass keep track if the call will raise `ValueError` "disowned" for one or more arguments. If this is true for the winning overload, raise a special exception before actually converting any arguments, to make the behavior predictable/portable (no arguments will be disowned).

This PR also fixing minor namespace naming inconsistencies across test_class_sh_*.cpp files.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -102,6 +102,7 @@ set(PYBIND11_TEST_FILES
     test_chrono.cpp
     test_class.cpp
     test_class_sh_basic.cpp
+    test_class_sh_disowning.cpp
     test_class_sh_factory_constructors.cpp
     test_class_sh_inheritance.cpp
     test_class_sh_trampoline_shared_ptr_cpp_arg.cpp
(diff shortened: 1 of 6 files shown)


rwgk:
CI run produces 34 failing, 30 successful, with the failure below (seems to be all of the 34).

Question: Is the order in which arguments are converted from Python to C++ not deterministic (across multiple platforms)?

```
    def test_mixed():
        while True:
            obj1a = m.Atype1(90)
            obj2a = m.Atype2(25)
            assert m.mixed(obj1a, obj2a) == (90 * 10 + 1) * 200 + (25 * 10 + 2) * 20
            obj1b = m.Atype1(0)
            with pytest.raises(ValueError):
                m.mixed(obj1b, obj2a)
            with pytest.raises(ValueError):
>               obj1b.get()  # obj1b was disowned even though m.mixed(obj1b, obj2a) failed.
E               Failed: DID NOT RAISE <class 'ValueError'>
```

rwgk:
The CI is green, also with PYBIND11_USE_SMART_HOLDER_AS_DEFAULT (#2879). This PR just affects tests, merging now to stay organized. This change will get reviewed internally. I'll open a new PR if there are requests for changes.

PR #2914, opened 2021-03-21
Catch missing self argument in overloads constructor

rhaschke:
The following unittest code was segfaulting due to missing self argument in call to `__init__`
```python
class BrokenTF6(m.TestFactory6):
    def __init__(self):
        m.TestFactory6.__init__()
```
Fixed now, by catching this edge case and throwing an appropriate exception.

## Suggested changelog entry:

```rst
Catch a missing ``self`` argument in calls to ``__init__()``.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -555,8 +555,8 @@ protected:
 
         auto self_value_and_holder = value_and_holder();
         if (overloads->is_constructor) {
-            if (!PyObject_TypeCheck(parent.ptr(), (PyTypeObject *) overloads->scope.ptr())) {
-                PyErr_SetString(PyExc_TypeError, "__init__(self, ...) called with invalid `self` argument");
+            if (!parent || !PyObject_TypeCheck(parent.ptr(), (PyTypeObject *) overloads->scope.ptr())) {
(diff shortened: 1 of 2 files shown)


henryiii (approved):
This looks like a good idea to me.

henryiii:
Could you please re-add the changlog entry to the issue description? Those get auto-processed into the changelog draft.

rhaschke:
@henryiii, I added a changelog suggestion. Hope this is in line with the desired verbosity and formatting.

henryiii:
Formatting gets auto applied (bullets and periods if missing), and I can tweak it, but it helps to have a starting point. Thanks! (I will change it to be RST though ;) )

YannickJadoul (approved):
Huh, how has this never been seen/caught before!?

Good catch, @rhaschke!

henryiii:
Probably because the normal way to call this is with `super`, and there you have to supply self (or it gets automatically picked up in Python 3).

YannickJadoul:
> Probably because the normal way to call this is with `super`, and there you have to supply self (or it gets automatically picked up in Python 3).

Hmmm, though pybind11 actually explicitly warns not to use `super`, since it won't fit with the way multiple inheritance currently works.

PR #2915, opened 2021-03-23
Passing reference arguments to trampoline methods

rhaschke:
This is a set of new unittests revealing that _reference_ arguments passed to trampoline methods,
i.e. Python-overridden virtual methods of C++ classes, are not actually passed by reference, but by copy.
This is due to the fact that `copy` is the default return_value_policy to pass from C++ to Python.

However, being able to pass by reference is essential in this case to allow object modifications
performed in Python code to become visible in C++.


rhaschke:
There is a simple fix to the illustrated issue: Just enforce `return_value_policy::reference` when calling from a trampoline.

rhaschke:
The last commit introduces the new `return_value_policy::trampoline_reference` to explicitly indicate a call from a trampoline dispatcher method. As the object transfer from C++ to Python is qualitatively different from returning from C++ function calls,
I thought it might be sensible to make these cases explicitly distinguishable for the casters.
In practice, this new policy just boils down to `return_value_policy::reference`. @rwgk @wjakob any thoughts?

rhaschke:
I think [`CI / 🐍 3.10-dev • windows-latest • x64`](https://github.com/pybind/pybind11/pull/2915/checks?check_run_id=2173597677) is failing for an unrelated reason:
```
Run actions/setup-python@v2
Version >= 3.10.0-a0 was not found in the local cache
Error: getaddrinfo ENOTFOUND api.github.com
```

PR #2920, opened 2021-03-26
More precise return_value_policy::automatic documentation.

rwgk:
I wrote the test under this PR to convince myself that the existing documentation is not actually precise. I think it's nice to keep the test as a reference.

### DIFF
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -290,4 +290,4 @@ def test_return_vector_bool_raw_ptr():
         v = m.return_vector_bool_raw_ptr()
         assert isinstance(v, list)
         assert len(v) == 4513
-        return  # Comment out for manual leak checking (use `top` command).
+        break  # Comment out for manual leak checking (use `top` command).
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -283,3 +283,11 @@ def test_issue_1561():
     bar.list = [m.Issue1561Inner("bar")]
     bar.list
     assert bar.list[0].data == "bar"
+
+
+def test_return_vector_bool_raw_ptr():
+    while True:
+        v = m.return_vector_bool_raw_ptr()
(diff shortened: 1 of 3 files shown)


tests/test_stl.py
```diff
@@ -283,3 +283,11 @@ def test_issue_1561():
...
     bar.list
     assert bar.list[0].data == "bar"
+
+
+def test_return_vector_bool_raw_ptr():
+    while True:
+        v = m.return_vector_bool_raw_ptr()
+        assert isinstance(v, list)
+        assert len(v) == 4513
+        break  # Comment out for manual leak checking (use `top` command).
```
henryiii, line 293: Kind of odd to have in a test? Maybe 
```suggestion
    # Add while True: for manual leak checking
    v = m.return_vector_bool_raw_ptr()
    assert isinstance(v, list)
    assert len(v) == 4513
```

henryiii (approved):

tests/test_stl.py
rwgk: Done. Thanks Henry!

Side remark & question:
* I have a lot of those on the smart_holder branch, and need to turn them on-off quite a bit when going into the smart_holder code.
* Do you happen to be aware of a pytest feature or plugin that would automate the fool-proof top-like leak check with adding a certain flag?

I've been thinking of this as a potential Google Summer of Code project if it doesn't exist already. Roughly: run decorated test functions in a loop while monitoring the equivalent of top's RES from within the same Python process, use some simple heuristics to decide "this could be a leak" or "no this doesn't look like a leak after running for N seconds".

rwgk:
I decided to remove the needs changelog label. It's an important detail to have right for someone looking carefully at the docs, but I feel that highlighting this PR in the changelog gives it too much weight (more distracting than helpful).

PR #2921, opened 2021-03-27
Fix Pybind11Extension on Mingw64

jeromerobert:
## Description

* `Pybind11Extension` add the `/EHsc /bigobj /std:c++14` flags on Windows.
  This is good for Visual C++ but not for Mingw.
* According
  https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python2/0410-MINGW-build-extensions-with-GCC.patch
  sysconfig.get_platform() is the way to check for a Mingw64

## Suggested changelog entry:

```rst
-  Fix the default ``Pybind11Extension`` compilation flags with a Mingw64 python
```

### DIFF
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -47,6 +47,7 @@ import tempfile
 import threading
 import platform
 import warnings
+import sysconfig
 
 try:
     from setuptools.command.build_ext import build_ext as _build_ext
@@ -59,7 +60,7 @@ import distutils.errors
 import distutils.ccompiler
 
 
-WIN = sys.platform.startswith("win32")
+WIN = sys.platform.startswith("win32") and sysconfig.get_platform() != "mingw"
 PY2 = sys.version_info[0] < 3
 MACOS = sys.platform.startswith("darwin")
 STD_TMPL = "/std:c++{}" if WIN else "-std=c++{}"


henryiii:
Great to see fixes for MinGW! Would you have suggestions for how to run minGW in GHA? I'd like to test for it, too, but don't know anything about using MinGW.

rwgk (approved):
I feel it's fine to merge this change even if we don't have a MINGW GitHub Action at the moment.

henryiii (approved):
I'm perfectly fine to merge this change as is, but I don't find someone with MinGW knowledge that often, so I'm hoping I can get some advice and/or maybe actually finally add a MinGW run to the tests. :) (And I thought I already approved, but I guess not)

rwgk:
Thanks @jeromerobert and @henryiii ! I'll merge this now. If you can help us with GHA that'll be great, but optional.

rwgk:
Just logging my thinking here about GHA:
* GHA MinGW would be a great to have.
* But it is an effort to get going, and will have a cost maintaining, too.
* How many people care/are affected? Maybe just a few for MinGW?
* Assuming breakages are rare, and we merge such tiny fixes like this one quickly, that's actually an efficient use of everybody's time.

henryiii:
These things pop up every so often, especially the the example repos. It's very much worth having if we can do it easily. I'm guessing that it's possible and pretty easy if you know how to set it up. We promise MinGW support in the readme, so it should be tested (I think now it is the only one promised that is not tested, actually). I'm hoping someone who cares about MinGW will help set it up one day.

jeromerobert:
@henryiii Using https://github.com/msys2/setup-msys2 it should be easy enough to add a test for Mingw64. What existing PyBind11 Github Actions job could we take inspiration from?

henryiii:
Well, any of the checks, maybe this one is close: 

https://github.com/pybind/pybind11/blob/f676782becc5bcd30c56b4da1ff1026db61bcce3/.github/workflows/ci.yml#L751-L795

We don't really test the one thing that you fixed here, but just that pybind11 compiles correctly; IIRC the python extensions are tested only in pybind/python_example, https://github.com/pybind/python_example/blob/master/.github/workflows. If we had a test in pybind11, it could be duplicated there (it uses pip instead of CMake).

henryiii:
FYI, this doesn't seem to work (like all untested code 🤣 ). I believe sysconfig only tells you what Python was built with, not what you are targeting. That requires a custom build_ext (which I believe I literally wrote in a comment somewhere in the docs).

A fix will need to patch the build_ext command and change these arguments if mingw64 is being used.

PR #2925, opened 2021-03-29
enum docstrings for sphinx

loriab:
## Description

I get errors like:
```
  File "/home/psilocaluser/toolchainconda/envs/ghadocs3/lib/python3.8/site-packages/sphinx/util/docstrings.py", line 62, in prepare_docstring
    lines = s.expandtabs(tabsize).splitlines()
AttributeError: 'pybind11_static_property' object has no attribute 'expandtabs'
```
for bindings like:
```
    py::enum_<Molecule::GeometryUnits>(m, "GeometryUnits", "The units used to define the geometry")
        .value("Angstrom", Molecule::Angstrom)
        .value("Bohr", Molecule::Bohr)
        .export_values();
```
unless this patch is applied in pybind11 or the result is stringified, `str(s).expandtabs(tabsize).splitlines()`, upstream in Sphinx https://github.com/sphinx-doc/sphinx/pull/9037

These bindings were fine for years with Sphinx <2. The proposed solution was modeled on https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L986 , but I don't know how you want to handle the issue. Or even if I perhaps missed an upgrade step.

I'd be glad to hear where you think the change should happen: my project, pybind11, or sphinx.

## Suggested changelog entry:

```rst
Fixed bug where enum docstrings errored in processing by Sphinx.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1644,7 +1644,7 @@ struct enum_base {
             }, name("name"), is_method(m_base)
         );
 
-        m_base.attr("__doc__") = static_property(cpp_function(
+        m_base.attr("__doc__") = pybind11::str(static_property(cpp_function(
             [](handle arg) -> std::string {
                 std::string docstring;
                 dict entries = arg.attr("__entries");
@@ -1660,7 +1660,7 @@ struct enum_base {
                 }
                 return docstring;
             }, name("__doc__")
-        ), none(), none(), "");
+        ), none(), none(), ""));
 
         m_base.attr("__members__") = static_property(cpp_function(
             [](handle arg) -> dict {


EricCousineau-TRI:
Hm... at first blush, I'm not sure if you're doing what you want (here, or in Sphinx PR).

My understanding of your goal is to make sure the binding docstrings make it through.

However, pybind11 uses a `static_property` to plumb it's docstring; instead of calling that method, you're instead stringifying the function itself, hence the CI errors at present (9521fe4) like
```
AssertionError: assert 'An unscoped enumeration' in '<pybind11_builtins.pybind11_static_property object at 0x7f324f9258b0>'
```

EricCousineau-TRI:
Might you have time to try to reproduce the error in pure Python 3.x for Sphinx 2.x?
It may be worth to file a Sphinx issue, as they may not be processing properties / descriptors correctly? (or maybe `pybind11` isn't doing something right?)

Lemme know if you have questions about this!

loriab:
Thanks for the hints, @eacousineau. I've taken another look, and I think I understand better what's going on. The key is the difference between how the `__doc__` attribute is accessed. Below, `psi4.core` is a pb11 module and [`GeometryUnits` is an enum](https://github.com/psi4/psi4/blob/master/psi4/src/export_mints.cc#L426-L429).

```
>>> import psi4
>>> psi4.core.GeometryUnits.__doc__
'The units used to define the geometry\n\nMembers:\n\n  Angstrom\n\n  Bohr'
>>> getattr(psi4.core.GeometryUnits, "__doc__", None)
'The units used to define the geometry\n\nMembers:\n\n  Angstrom\n\n  Bohr'
>>> psi4.core.GeometryUnits.__dict__["__doc__"]
<pybind11_builtins.pybind11_static_property object at 0x7f0c2d031d10>
```

The pb11 setup is clearly fine with builtin `getattr`, but sphinx extracts the docstring with its [own `get_attr`](https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/ext/autodoc/__init__.py#L1673) that allows sphinx extensions to register their own attribute extraction, https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/ext/autodoc/__init__.py#L2624-L2626 . I've been using the [Astropy automodapi](https://sphinx-automodapi.readthedocs.io/en/latest/) extension, and it registers a [function that catches type `property`](https://github.com/astropy/sphinx-automodapi/blob/main/sphinx_automodapi/autodoc_enhancements.py#L54) and extracts the `obj.__dict__[attr]` that returns the static property repr rather than the straightforward `getattr(obj, attr)` that returns the desired docstring.

I don't see pb11 or sphinx itself doing anything wrong here - the trouble comes from the automodapi extension's handling. I'll see about filing an issue over there. But if you've any ideas on how to distinguish the pb11 properties from the properties they're targeting or any other comments, I'd be glad to hear them. I'll change the title of this PR to narrow the scope.

PR #2930, opened 2021-04-01
CI for smart holders as default

rhaschke:
This adds another CI workflow enabling `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`

### DIFF
--- a/.github/workflows/ci.sh.yml
+++ b/.github/workflows/ci.sh.yml
@@ -1,0 +1,860 @@
+name: "CI SH"
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable
+      - smart_holder
+      - v*
+
+jobs:
+  # This is the "main" test suite, which tests a large number of different
+  # versions of default compilers and Python versions in GitHub Actions.
+  standard:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


rhaschke:
cmake [implicitly sets `CMAKE_CXX_FLAGS="/GR /EHsc"` for MSVC](https://gitlab.kitware.com/cmake/cmake/-/issues/20610). Manually setting `CMAKE_CXX_FLAGS` overrides this default and thus requires adding those flags explicitly.

rwgk (commented):
I really like that it's so easy, but I'm also really worried about introducing two copies of something so substantial. @henryiii, is there an obvious way to avoid the duplication?

Naive idea: have a small patch file somewhere else, and a bash script that applies that patch to generate this file, with a big comment at the top: "This is an auto-generated file, do not edit." And another comment with instructions how to run the bash script. — But I hope Henry knows of something more organized.

rhaschke:
I tried to merge the new SH builds into the existing `CI` workflow before, but that ended up in a mess, because the workflow defines its jobs using matrix + includes. Doing all jobs with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` enabled required way more changes to the `CI` workflow. With the present approach, it should be easy to keep the `CI` and `CI SH` workflow in sync.

rwgk:
The work here was included in #2963. Closing. Thanks Robert!

PR #2939, opened 2021-04-06
Connecting PYBIND11_INTERNALS_VERSION to PYBIND11_USE_SMART_HOLDER_AS_DEFAULT.

rwgk:
Extensions built with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` defined are incompatible with extensions built against master, or `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` undefined, therefore `PYBIND11_INTERNALS_VERSION` needs to be different. The approach chosen here is to add `1000` to the `PYBIND11_INTERNALS_VERSION` on master.

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -155,7 +155,7 @@ struct type_info {
 #define PYBIND11_INTERNALS_VERSION 4
 #else
 // See README_smart_holder.rst:
-// Classic / conservative / progressive cross-module compatibility
+// Classic / Conservative / Progressive cross-module compatibility
 #define PYBIND11_INTERNALS_VERSION 1004
 #endif
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -154,6 +154,8 @@ struct type_info {
 #ifndef PYBIND11_USE_SMART_HOLDER_AS_DEFAULT
 #define PYBIND11_INTERNALS_VERSION 4
 #else
+// See README_smart_holder.rst:
+// Classic / conservative / progressive cross-module compatibility
 #define PYBIND11_INTERNALS_VERSION 1004
 #endif
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "../pytypes.h"
+#include "smart_holder_sfinae_hooks_only.h"
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
@@ -150,7 +151,11 @@ struct type_info {
 };
 
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version
+#ifndef PYBIND11_USE_SMART_HOLDER_AS_DEFAULT
 #define PYBIND11_INTERNALS_VERSION 4
+#else
+#define PYBIND11_INTERNALS_VERSION 1004
+#endif
 
 /// On MSVC, debug and release builds are not ABI-compatible!
 #if defined(_MSC_VER) && defined(_DEBUG)
--- a/README_smart_holder.rst
+++ b/README_smart_holder.rst
@@ -76,8 +76,8 @@ Conservative or Progressive mode?
 =================================
 
 It depends. To a first approximation, for a stand-alone, new project, the
-progressive mode will be easiest to use. For larger projects or projects
-that integrate with third-party pybind11-based projects, the conservative
+Progressive mode will be easiest to use. For larger projects or projects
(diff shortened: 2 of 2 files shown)


rwgk:
@rhaschke the github web UI won't let me add you as a reviewer, I have no idea why. Tagging you here instead.

rwgk:
https://github.com/pybind/pybind11/pull/2879 is used for testing with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` defined.

rhaschke (commented):
I would expect a user to enable `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` via a compiler define on the cmdline. I followed this approach in #2930, which enables both variants as two different workflows: `CI` and `CI SH`.

rhaschke:
> The github web UI won't let me add you as a reviewer, I have no idea why.

You can only tag people as a reviewer that are part of the maintainer team.

include/pybind11/detail/internals.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "../pytypes.h"
+#include "smart_holder_sfinae_hooks_only.h"
```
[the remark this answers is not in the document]
rwgk: Initially mainly to have an easy way to run the CI with smart_holder as default. But then I thought it's good to keep as an option, to give people an easy way to experiment using smart_holder as default without having to go into their cmake/CI/IDE/whatever they may have. As-is, it works both ways. I'm reluctant to take one option away, especially because the cost is just a few extra lines in smart_holder_sfinae_hooks_only.h. With pytypes.h included first here, everything else is included already anyway.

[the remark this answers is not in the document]
rhaschke: I wasn't asking to remove that option. I was just wondering. For CI, I suggest to use the cmake argument option to allow both checks to run on the same code base. I'm still struggling with some issues in #2930 though... But, I'm optimistic :smile:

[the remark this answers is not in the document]
rwgk: Oh, good, thanks! And yes, it'll be awesome to trigger all testing with just one PR.

include/pybind11/detail/internals.h
```diff
@@ -150,7 +151,11 @@ struct type_info {
 };
 
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version
+#ifndef PYBIND11_USE_SMART_HOLDER_AS_DEFAULT
 #define PYBIND11_INTERNALS_VERSION 4
+#else
+#define PYBIND11_INTERNALS_VERSION 1004
```
EricCousineau-TRI, line 157: Not sure if I 100% understand what "incompatible" means from PR overview.

Is it just that `pybind11` doesn't store `typeid` info for the holder type, so code using `py::class_<>` across diff. modalities (non-SH vs SH) may get messed up, so a new flavor of #1138?

If so, kewl for now! Just wanna confirm; perhaps worth adding a checkbox / issue / TODO next to this?
Ideally, the TODO would be removed when SH gets merged as a mainline option.

rwgk: I'm not 100% certain myself, but my understanding is that modules built with differing `PYBIND11_INTERNALS_VERSION` are fully isolated from each other. If that's true:

* _Without_ this PR, you'd indeed see a new flavor of #1138 (holder mismatch, leading to a `reinterpret_cast` with UB).
* But with this PR, types `py::class_`-wrapped in module_compiled_with_smart_holder_as_default are fully isolated from `py::class_` types wrapped in module_compiled_with_std_unique_ptr_as_default.

Does that sound plausible?

The upshot is that two modules that both use a certain `py::class_<T>`-wrapped type need to be built with matching `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`, or the casters for `T` will only be available in the module that has the `py::class_<T>`.

EricCousineau-TRI: Yup, totes sounds plausible! Would it be possible to encode this in a 1-sentence TODO or something?

Perhaps also mentioning obvious downshot: Because you're isolated, you can't share instances.

rwgk: Thanks for confirming! I'll rebase this PR and will add a small section to README_smart_holder.rst.

rwgk: @EricCousineau-TRI The "1-sentence TODO or something" became the longest section in README_smart_holder.rst :-)

**Classic / Conservative / Progressive cross-module compatibility**

I'll merge this PR now because we agree on the code change. I'll update the documentation as needed in separate PRs. As before, feedback for the documentation is highly welcome, chat, comment on this PR, new PR, whatever fits best.

EricCousineau-TRI: Sounds great, and sorry to motivate a large section haha! Thanks!

PR #2944, opened 2021-04-10
Allow function pointer extraction from overloaded functions

otamachan:
## Description

When I try to avoid  C++ -> Python -> C++ roundtrip for overloaded functions, it only works for the first overloaded function.
This PR allows function pointer extraction from all overloaded functions.

I added a [failure test case](https://github.com/pybind/pybind11/commit/10a1c967fa6d28f286831d2dd58724e39a172658). Without the fix in this PR, it fails as follows:
```python
>       assert (
            m.test_dummy_function(m.dummy_function_overloaded)
            == "matches dummy_function: eval(1) = 2"
        )
E       assert "can't conver...: eval(1) = 2" == 'matches dumm...: eval(1) = 2'
E         - matches dummy_function: eval(1) = 2
E         + can't convert to function pointer: eval(1) = 2
```

I applied clang-format to only the related code block so the style was changed a bit.

## Suggested changelog entry:

```rst
Allow function pointer extraction from overloaded functions
```

### DIFF
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -46,11 +46,17 @@ public:
             auto c = reinterpret_borrow<capsule>(PyCFunction_GET_SELF(cfunc.ptr()));
             auto rec = (function_record *) c;
 
-            if (rec && rec->is_stateless &&
-                    same_type(typeid(function_type), *reinterpret_cast<const std::type_info *>(rec->data[1]))) {
-                struct capture { function_type f; };
-                value = ((capture *) &rec->data)->f;
-                return true;
+            while (rec != nullptr) {
+                if (rec->is_stateless
+                    && same_type(typeid(function_type),
+                                 *reinterpret_cast<const std::type_info *>(rec->data[1]))) {
+                    struct capture {
+                        function_type f;
+                    };
+                    value = ((capture *) &rec->data)->f;
+                    return true;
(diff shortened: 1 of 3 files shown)


include/pybind11/functional.h
```diff
@@ -46,11 +46,17 @@ struct type_caster<std::function<Return(Args...)>> {
...
+                if (rec->is_stateless
+                    && same_type(typeid(function_type),
+                                 *reinterpret_cast<const std::type_info *>(rec->data[1]))) {
+                    struct capture {
+                        function_type f;
+                    };
+                    value = ((capture *) &rec->data)->f;
+                    return true;
+                }
+                rec = rec->next;
```
henryiii, line 59: This is the only actual change, I think. The rest is formatting. Note that if we used clang-format, this would not randomly change. :)

henryiii (approved):

include/pybind11/functional.h
rwgk: Yep, I feel the same.

rwgk (approved):
Looks great to me, thanks for the fix Tamaki!
(I'm a bit amazed that this oversight survived for so long.)

I already ran this PR through our Google-internal global testing system: all green.
I'll go ahead with merging this PR.

PR #2950, opened 2021-04-14
pytypes: Add iterable_t<>

### LINKED ISSUE #1773: std::vector from py::iterable
At the moment `list_caster` in `stl.h` explicitly checks whether the input argument is a `sequence`, but I would like to argue that that's slightly too restrictive and `iterable` should be accepted.

The only potential downside I can think of is that generators should definitely not be accepted as a failed conversion of an element halfway through the generator would leave things in a terrible state, but a well-placed `PyGen_Check` could prevent that.

EricCousineau-TRI:
## Description

Resolves #1773 as an alternative / less disruptive apporach

## Suggested changelog entry:

```rst
iterable_t<T> added as a simple pytype
```


include/pybind11/pytypes.h
```diff
@@ -753,6 +763,39 @@ inline bool PyIterable_Check(PyObject *obj) {
...
+            // cast error.
+            Py_DECREF(iter);
+            throw cast_error(
+                "iterable_t<T> cannot be used with exhaustible iterables "
+                "(e.g., iterators, generators).");
+        }
+        bool good = true;
+        // Now that we know that the iterable `obj` will not be exhausted,
+        // let's check the contained types.
+        for (handle h : handle(iter)) {
```
EricCousineau-TRI, line 785: Working: C++11 is struggling with forward reference here. Should see how hard it is to just use pure CPython API, or shift this def. to after `py::handle` is declared.

tests/test_pytypes.py
```diff
@@ -21,6 +21,45 @@ def test_iterable(doc):
...
+    # Negative: Exhaustible iterables (e.g. iterators, generators).
+    gen_empty = (x for x in set_empty)
+    with pytest.raises(RuntimeError):
+        m.accept_iterable_t(gen_empty)
+    iter_empty = iter(list_empty)
+    with pytest.raises(RuntimeError):
+        m.accept_iterable_t(iter_empty)
+
+    # Str.
+    list_of_str = ["hey", "you"]
```
rwgk, line 43: I think it'll work with Python 2.7 if you make this `u"hey"` etc.

tests/test_pytypes.py
```diff
@@ -21,6 +21,45 @@ def test_iterable(doc):
...
+    set_of_str = {"hey", "you"}
+    assert m.accept_iterable_t(list_of_str) == "str"
+    assert m.accept_iterable_t(set_of_str) == "str"
+    # - Negative: Not fully `str`.
+    list_of_str_and_then_some = ["hey", 0]
```
rwgk, line 48: `u"hey"` here, too, although I'm guessing you get a `TypeError` either way.

include/pybind11/cast.h
```diff
@@ -50,6 +50,18 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 template <typename type, typename SFINAE = void> class type_caster : public type_caster_base<type> { };
 template <typename type> using make_caster = type_caster<intrinsic_t<type>>;
 
+template <typename T>
+struct is_generic_type<
```
rwgk, line 54: Two questions:

* Will this work if I add another specialization pair for `is_base_of<smart_holder_type_caster_base_tag, ...>`?

* What do you think about choosing a different name, to connect to the intended purpose instead of pegging the name on `type_caster_generic`, e.g. `is_iterable_t_compatible`? (Otherwise it will get really confusing that `generic` is connected to `smart_holder_type_caster`.)

rwgk (commented):
Cool! Just questions regarding compatibility with the smart_holder branch.

include/pybind11/pytypes.h
```diff
@@ -753,6 +763,39 @@ inline bool PyIterable_Check(PyObject *obj) {
...
+            if (!isinstance<T>(h)) {
+                good = false;
+                break;
+            }
+        }
+        Py_DECREF(iter);
+        return good;
+    } else {
```
Skylion007, line 793: else after return is bad style.

Skylion007 (changes requested):

rwgk:
> Just commenting to say I'm working on something atm for which this would be useful :-)

This needs to be resurrected after the sweeping changes for clang-format and Python 2 drop. You could transfer the diffs to a new PR and work on it there.

include/pybind11/cast.h
EricCousineau-TRI: Done w/ db7ed8d - I removed `is_generic_type<>` as it's inclusion here may be a bit confusing

Will make a small separate PR

include/pybind11/pytypes.h
EricCousineau-TRI: Done (I think) w/ 34de38d

tests/test_pytypes.py
EricCousineau-TRI: I think not applicable anymore - no more Python 2 :partying_face:

tests/test_pytypes.py
EricCousineau-TRI: I think not applicable anymore - no more Python 2 :partying_face:

EricCousineau-TRI:
> This needs to be resurrected after the sweeping changes for clang-format and Python 2 drop. You could transfer the diffs to a new PR and work on it there.

It was luckily small enough w/ just a merge + conflict resolution. I think the review is still fine w/ merge. Let me know if you'd still like a new PR.

EricCousineau-TRI:
`is_generic_type` split off into #3857

include/pybind11/pytypes.h
EricCousineau-TRI: Done 19e8992e - I think

rwgk (commented):
High-level:

> At the moment list_caster in stl.h explicitly checks whether the input argument is a sequence, but I would like to argue that that's slightly too restrictive and iterable should be accepted.

The code seems to be dicy ("bad! For now,") and the gain seems really minute ("slightly too restrictive").

The current implementation doesn't seem exception safe (leaks b/o missed Py_DECREF if exceptions are thrown before the control flow reaches those).

But really I feel it's best not to get into this scraping-the-bottom-of-the-barrel territory at all.

EricCousineau-TRI:
Aye - so you think it's best to not introduce the feature? Fine by me, just want to clarify some points.

> The code seems to be dicy ("bad! For now,") and the gain seems really minute ("slightly too restrictive").

Too restrictive in what sense? Not an iterator, or shouldn't check types, or something else?

rwgk:
> Aye - so you think it's best to not introduce the feature? Fine by me, just want to clarify some points.
> 
> > The code seems to be dicy ("bad! For now,") and the gain seems really minute ("slightly too restrictive").
> 
> Too restrictive in what sense? Not an iterator, or shouldn't check types, or something else?

Sorry I wasn't clear, the "slightly too restrictive" quote is from here: https://github.com/pybind/pybind11/issues/1773

I.e. my interpretation was that even the original requester was thinking the current behavior is only "slightly too restrictive". (I agree.)

My cost/benefit assessment is just one vote! Yes, I'd say let's not add this, but if there are other opinions I'll reconsider. I'm just offering my current opinion.

EricCousineau-TRI:
Sounds good! I'm not too wed to this, but will await opinions, and close if we don't hear back for a bit (~month?)

@alexdewar If you're available, is this still useful as-is?

PR #2956, opened 2021-04-16
Fix compilation with gcc < 5

mvoelkle-cern:
## Description

When the user defines _GLIBCXX_USE_CXX11_ABI=0 to force the pre-c++11 ABI, numpy.h assumes that is_trivially_copyable is available.
It is not necessarily the case. This patch uses clang's feature detection instead.

## Suggested changelog entry:

```rst
Fixed compilation with GCC < 5 when the user defines _GLIBCXX_USE_CXX11_ABI.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -319,12 +319,13 @@ template <typename T> using remove_all_extents_t = typename array_info<T>::type;
 
 template <typename T> using is_pod_struct = all_of<
     std::is_standard_layout<T>,     // since we're accessing directly in memory we need a standard layout type
-#if !defined(__GNUC__) || __GNUC__ >= 5 || (defined(__clang__) && __has_feature(is_trivially_copyable))
-    std::is_trivially_copyable<T>,
-#else
-    // libstdc++ < 5 doesn't implement is_trivially_copyable, so approximate it
+#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
+    // libstdc++ < 5 (including versions 4.8.5, 4.9.3 and 4.9.4 which were released after 5)
+    // don't implement is_trivially_copyable, so approximate it
     std::is_trivially_destructible<T>,
     satisfies_any_of<T, std::has_trivial_copy_constructor, std::has_trivial_copy_assign>,
+#else
+    std::is_trivially_copyable<T>,
 #endif
     satisfies_none_of<T, std::is_reference, std::is_array, is_std_array, std::is_arithmetic, is_complex, std::is_enum>
 >;
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -319,12 +319,10 @@ template <typename T> using remove_all_extents_t = typename array_info<T>::type;
 
 template <typename T> using is_pod_struct = all_of<
     std::is_standard_layout<T>,     // since we're accessing directly in memory we need a standard layout type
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Howdy @mvoelkle-cern! Do you have a theory as to why our present CI coverage doesn't already check this case?
From a naive read at present `master`, it seems like we should have?
https://github.com/pybind/pybind11/blob/114be7f4ade0ad798cd4c7f5d65ebe4ba8bd892d/.github/workflows/ci.yml#L565-L574

Also, I believe these CI errors are caused by your PR - can you fix those?

mvoelkle-cern:
Hi, sorry my fix didn't compile outside of clang, I had wrongly assumed that preprocessor boolean expressions would short-circuit. Here is an other approach that targets libstdc++ specifically.

mvoelkle-cern:
> Howdy @mvoelkle-cern! Do you have a theory as to why our present CI coverage doesn't already check this case?
> From a naive read at present `master`, it seems like we should have?
> 
> https://github.com/pybind/pybind11/blob/114be7f4ade0ad798cd4c7f5d65ebe4ba8bd892d/.github/workflows/ci.yml#L565-L574

The issue is when using 4.8 AND the user defines _GLIBCXX_USE_CXX11_ABI=0. The CI coverage doesn't define it.

include/pybind11/numpy.h
```diff
@@ -319,14 +319,13 @@ template <typename T> using remove_all_extents_t = typename array_info<T>::type;
...
 template <typename T> using is_pod_struct = all_of<
     std::is_standard_layout<T>,     // since we're accessing directly in memory we need a standard layout type
-#if !defined(__GNUG__) || defined(_LIBCPP_VERSION) || defined(_GLIBCXX_USE_CXX11_ABI)
-    // _GLIBCXX_USE_CXX11_ABI indicates that we're using libstdc++ from GCC 5 or newer, independent
-    // of the actual compiler (Clang can also use libstdc++, but it always defines __GNUC__ == 4).
-    std::is_trivially_copyable<T>,
-#else
-    // GCC 4 doesn't implement is_trivially_copyable, so approximate it
+#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
+    // libstdc++ < 5 (including versions 4.8.5, 4.9.3 and 4.9.4 which were released after 5)
```
henryiii, line 323: 🤦 Why can't they just have provided a nice way to get the major/minor numbers? 😢  Looks like a more exact solution than what we had, sounds fine to me.

henryiii (approved):

include/pybind11/numpy.h
EricCousineau-TRI: Agreed - thanks!

EricCousineau-TRI (approved):
I can't state w/ 100% confidence that I can verify, but will trust testing on your side as well as our CI (and of course, Henry's review as well). Thanks!

damian-cell:
> ## Description
> When the user defines _GLIBCXX_USE_CXX11_ABI=0 to force the pre-c++11 ABI, numpy.h assumes that is_trivially_copyable is available.
> It is not necessarily the case. This patch uses clang's feature detection instead.
> 
> ## Suggested changelog entry:
> ```
> Fixed compilation with GCC < 5 when the user defines _GLIBCXX_USE_CXX11_ABI.
> ```

damian-cell:
> > ## Description
> > When the user defines _GLIBCXX_USE_CXX11_ABI=0 to force the pre-c++11 ABI, numpy.h assumes that is_trivially_copyable is available.
> > It is not necessarily the case. This patch uses clang's feature detection instead.
> > ## Suggested changelog entry:
> > ```
> > Fixed compilation with GCC < 5 when the user defines _GLIBCXX_USE_CXX11_ABI.
> > ```

PR #2958, opened 2021-04-18
Bug fix: adding back `!is_alias<Class>(ptr)` that were accidentally omitted.

rwgk:
This PR started as a small bug fix for include/pybind11/detail/init.h.

The init.h bug escaped detection because of missing test coverage in test_factory_constructors. To systematically plug such holes, the `PYBIND11_SH_DEF` macro was introduced and applied to test_factory_constructors, test_multiple_inheritance, and test_methods_and_attributes.

To have a complete toolset for Classic / Conservative / Progressive transitions, the `PYBIND11_SH_AVL` macro was added. test_mock_cpp and README_smart_holder.rst were revised accordingly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,8 @@ jobs:
         - 3.5
         - 3.6
         - 3.9
-        - 3.10-dev
+        # Broken b/o https://github.com/pytest-dev/pytest/issues/8539
+        # - 3.10-dev
         - pypy2
         - pypy3
 
--- a/README_smart_holder.rst
+++ b/README_smart_holder.rst
@@ -101,7 +101,7 @@ holder:
        py::classh<Foo>(m, "Foo");
    }
 
-There are three small differences compared to classic pybind11:
+There are three small differences compared to Classic pybind11:
 
 - ``#include <pybind11/smart_holder.h>`` is used instead of
   ``#include <pybind11/pybind11.h>``.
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -23,7 +23,8 @@ jobs:
         - 3.5
         - 3.6
         - 3.9
-        - 3.10-dev
+        # Broken b/o https://github.com/pytest-dev/pytest/issues/8539
+        # - 3.10-dev
```
henryiii, line 27: Sounds fine to me until this gets resolved.

henryiii: By the way, this would be much better as it's own PR, not mixed into an unrelated PR. Also makes it easier to revert.

henryiii: If you can cherry-pick b3fff905bed994a9aada72565cc29305991c03b5 I think we can accept it immediately and then restart the CI (or just ignore these jobs) for the pending PRs.

rwgk: Hi @henryiii, thanks, I'm not sure how obvious this is in your view, this is only the smart_holder branch. (Pointing out to make sure you're not surprised.)

rwgk:
The CI is green, also with `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` (#2879). This change will get reviewed Google-internally. Merging. I'll open a new PR if there are requests for changes.

.github/workflows/ci.yml
henryiii: Not super obvious, that explains it; worth cherry-picking and applying to master via a PR, I think.

henryiii: That does help explain the smart-holder PRs. Maybe a labeler or prefix (like `[smart holder]`) or such would help. Or I can just watch for the target branch. :)

PR #2959, opened 2021-04-18
tests: Skip/enable tests more explicitly (in lieu of `hasattr`)

EricCousineau-TRI:
## Description

Minor test cleanup per @anntzer's suggestion in https://github.com/pybind/pybind11/pull/2730#discussion_r615405935

## Suggested changelog entry:

N/A. Testing only.

### DIFF
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -82,6 +82,7 @@ TEST_SUBMODULE(builtin_casters, m) {
     // Under Python 2.7, invalid unicode UTF-32 characters don't appear to trigger UnicodeDecodeError
     if (PY_MAJOR_VERSION >= 3)
         m.def("bad_utf32_string", [=]() { return std::u32string({ a32, char32_t(0xd800), z32 }); });
+    m.attr("wchar_t_size") = (sizeof(wchar_t) == 2);
     if (PY_MAJOR_VERSION >= 3 || sizeof(wchar_t) == 2)
         m.def("bad_wchar_string", [=]() { return std::wstring({ wchar_t(0x61), wchar_t(0xd800) }); });
     m.def("u8_Z", []() -> char { return 'Z'; });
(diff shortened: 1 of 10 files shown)


tests/test_builtin_casters.py
```diff
@@ -27,13 +27,13 @@ def test_unicode_conversion():
         m.bad_utf16_string()
 
     # These are provided only if they actually fail (they don't when 32-bit and under Python 2.7)
-    if hasattr(m, "bad_utf32_string"):
+    if not env.PY2:
         with pytest.raises(UnicodeDecodeError):
             m.bad_utf32_string()
-    if hasattr(m, "bad_wchar_string"):
+    if not env.PY2 and m.wchar_t_size == 2:
```
EricCousineau-TRI, line 33: FWIW This may be a dumb hill to die on. Lemme know if I should revert this.

tests/test_builtin_casters.cpp
```diff
@@ -113,6 +114,8 @@ TEST_SUBMODULE(builtin_casters, m) {
     // test_single_char_arguments
     m.def("ord_char8", [](char8_t c) -> int { return static_cast<unsigned char>(c); });
     m.def("ord_char8_lv", [](char8_t &c) -> int { return static_cast<unsigned char>(c); });
+#else  // PYBIND11_HAS_U8STRING
```
rwgk, line 117: I really like this PR, but find it unfortunate that the true/false assignments are separated in space. How about:
```
#ifndef PYBIND11_HAS_U8STRING
    m.attr("has_u8string") = false;
#else
    m.attr("has_u8string") = true;
    ...  
#endif
```

rwgk (approved):

henryiii (approved):

tests/test_builtin_casters.cpp
henryiii: Starting with negative and then putting the bulk in a double negative is not really ideal, either? How about:

```cpp
#ifdef PYBIND11_HAS_U8STRING
    ...
    m.attr("has_u8string") = true;
#else
    m.attr("has_u8string") = false;
#endif
```

rwgk: Looks good. Either way is fine with me.

Skylion007 (approved):

henryiii:
I'm not sure it's better to add a bunch of manual true/falses that have to be maintained instead of just checking to see if something is there, but okay if you want to go this way - needs a rebase/merge, though, there's a conflict. (Some of theme are clear improvements when they check something _other_ than just injecting a true/false replacement for hasattr)

PR #2967, opened 2021-04-21
tests: use master commit for pytest on 3.10

henryiii:
## Description

Alternative to, and closes #2962. I just learned this has landed in master.


rwgk (approved):
Awesome!
Independently from the immediate CI problem this solving, testing pytest master along with the latest Python seems like the best thing to do.

PR #2972, opened 2021-04-22
pickle setstate: setattr __dict__ only if not empty

rwgk:
## Description

The motivation for this PR is to not force the use of `py::dynamic_attr()` unnecessarily. Concrete use case:

https://github.com/deepmind/open_spiel/blob/643dfd2a5ef023e796e20d06758800ec34a1204b/open_spiel/python/pybind11/pyspiel.cc#L309

`py::dynamic_attr()` comes with the disadvantage that it enables adding arbitrary attributes, which opens the door to accidents that tend to be subtle and time-consuming to debug (example situation below). The cost for avoiding this disadvantage is minuscule: just one additional `if` with 2 simple conditions in pybind11/detail/init.h.

The newly added unit tests reflect a typical situation in which this PR helps: a wrapped C++ virtual base class with a trampoline, combined with the use of base class pointers to C++ derived classes.

Example situation:
* You need to pickle a C++ derived class (as in the new test), but without this PR that is only possible if you use `py::dynamic_attr()`.
* Therefore you decide to add `py::dynamic_attr()`.
* Time goes by ...
* Someone accidentally assigns a misspelled attribute, but because of `py::dynamic_attr()` the code runs and produces nonsensical results. It takes you N hours to trace that out.

## Suggested changelog entry:

```rst
pickle support enhancement: ``setstate`` implementation will attempt to ``setattr`` ``__dict__`` only if the unpickled ``dict`` object is not empty, to not force use of ``py::dynamic_attr()`` unnecessarily.
```


EricCousineau-TRI:
And for given motivation, is it possible to give precise numbers (or rough) in terms of performance gain (if it was speed / memory)?

rwgk:
Hi @EricCousineau-TRI and @elkhrt, this PR is complete now. Could you please review?

rwgk:
The one CI failure is just our most common flake (see #2995).

elkhrt (approved):
Thanks Ralf - this is great.

include/pybind11/detail/init.h
```diff
@@ -293,7 +293,13 @@ template <typename Class, typename T, typename O,
...
 void setstate(value_and_holder &v_h, std::pair<T, O> &&result, bool need_alias) {
     construct<Class>(v_h, std::move(result.first), need_alias);
-    setattr((PyObject *) v_h.inst, "__dict__", result.second);
+    auto d = handle(result.second);
+    if (PyDict_Check(d.ptr()) && PyDict_Size(d.ptr()) == 0) {
+        // Skipping setattr below, to not force use of py::dynamic_attr() for Class unnecessarily.
+        // See PR #2972 for details.
+        return;
+    }
+    setattr((PyObject *) v_h.inst, "__dict__", d);
```
EricCousineau-TRI, line 302: nit For clarity, can you negate above condition, and place `setattr` inside negated conditional?
(short-circuit has weird readability IMO)

tests/CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ set(PYBIND11_TEST_FILES
     test_opaque_types.cpp
     test_operator_overloading.cpp
     test_pickling.cpp
+    test_pickling_trampoline.cpp
```
EricCousineau-TRI, line 126: nit Is it possible to merge this w/ existing testing file?

tests/test_pickling_trampoline.py
```diff
@@ -0,0 +1,37 @@
...
+    pass
+
+
+def test_roundtrip_simple_py_derived():
+    p = SimplePyDerived()
+    p.num = 202
+    p.stored_in_dict = 303
+    data = pickle.dumps(p, pickle.HIGHEST_PROTOCOL)
+    p2 = pickle.loads(data)
+    assert p2.num == 202
```
EricCousineau-TRI, line 24: This could pass w/o the type being the same.

Could you assert that the type is the same? `assert isinstance(p2, SimplePyDerived)`?

tests/test_pickling_trampoline.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+import env  # noqa: F401
+
+from pybind11_tests import pickling_trampoline as m
+
+try:
+    import cPickle as pickle  # Use cPickle on Python 2.7
```
EricCousineau-TRI, line 9: nit This is an indirect conditional; instead of `try/catch`, can you use `if env.PY2` to make it explicit? (then no comment necessary)

tests/test_pickling_trampoline.py
```diff
@@ -0,0 +1,37 @@
...
+    assert p2.stored_in_dict == 303
+
+
+def test_roundtrip_simple_cpp_derived():
+    p = m.make_SimpleCppDerivedAsBase()
+    p.num = 404
+    if not env.PYPY:
+        with pytest.raises(AttributeError):
+            # To ensure that future changes do not accidentally invalidate this unit test.
+            p.__dict__
```
EricCousineau-TRI, line 34: nit Can this `try / catch` be replaced with `assert not hasattr(p, '__dict__')`?

tests/test_pickling_trampoline.py
```diff
@@ -0,0 +1,37 @@
...
+    assert p2.num == 202
+    assert p2.stored_in_dict == 303
+
+
+def test_roundtrip_simple_cpp_derived():
+    p = m.make_SimpleCppDerivedAsBase()
+    p.num = 404
+    if not env.PYPY:
+        with pytest.raises(AttributeError):
+            # To ensure that future changes do not accidentally invalidate this unit test.
```
EricCousineau-TRI, line 33: nit "future changes" is a tad ambiguous. What should we be looking out for?

tests/test_pickling_trampoline.py
```diff
@@ -0,0 +1,37 @@
...
+            p.__dict__
+    data = pickle.dumps(p, pickle.HIGHEST_PROTOCOL)
+    p2 = pickle.loads(data)
```
EricCousineau-TRI, line 36: Same as above - possible failure mode is if type is somehow dropped.

Can you assert the type is preserved round-trip?

tests/test_pickling_trampoline.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "pybind11_tests.h"
+
+#include <memory>
+#include <stdexcept>
+#include <utility>
+
+namespace {
+
+struct SimpleBase {
+    int num               = 0;
```
EricCousineau-TRI, line 14: nit Whitespace alignment is inconsistent with other files.

Can you remove visual indents for now? e.g. `sed -E 's#(\w) +#\1 #g'`?

EricCousineau-TRI (changes requested):
Looks good, thank you for providing context (and soz for delay!)

Requesting more concrete testing, with nits on consistent styling w/ existing and usage of more explicitly conditionals in lieu of `try/catch`. (Feel free to punt on nits!)

For the overview, I can see two possible "debugging" issue:
(1) you accidentally assign a non-existent attribute but `py::dynamic_attr()` allows it (thus "ouch", you gotta trace that out)
(2) is that you cannot unpickle a C++ derived class, motivating `py::dynamic_attr()`, forcing you to run into (1).

Is it possible for you to rewrite the overview / commit stmt to expliclty focus on (2)?
(then (1) becomes an obvious side effect?)

Lemme know if you have any Q's or would like push back on this!

rwgk (commented):
Thanks Eric! I updated the code and the PR description starting from your suggestion.

rwgk:
Apparently the Centos 8 build broke for external reasons on 6/3. Trying again (PR Close-Reopen) to see if it fixed itself.

include/pybind11/detail/init.h
EricCousineau-TRI: OK Sounds good!

tests/test_pickling.py
```diff
@@ -45,3 +45,35 @@ def test_enum_pickle():
...
+    p = m.make_SimpleCppDerivedAsBase()
+    p.num = 404
+    if not env.PYPY:
+        # To ensure that this unit test is not accidentally invalidated.
+        with pytest.raises(AttributeError):
+            # Mimics the `setstate` C++ implementation.
+            setattr(p, "__dict__", {})  # noqa: B010
+    data = pickle.dumps(p, pickle.HIGHEST_PROTOCOL)
+    p2 = pickle.loads(data)
+    assert isinstance(p2, m.SimpleBase)
```
EricCousineau-TRI, line 78: This test now makes me concerned about object slicing: https://en.wikipedia.org/wiki/Object_slicing
As I read this, this object is now *sliced*; it was originally `SimpleCppDerived`, but Python only knows it as `SimpleBase`, therefore the object (and its vtable) will get sliced down to `SimpleBase`.

Ideally, there's a fix, either:
(a) If you pickle a `typeid` that isn't explicitly registered with pybind11, fail fast (otherwise, pybind11 doesn't know how to reconsistitute)
(b) Somehow, magically, teach pybind11 how to pickle+unpickle lower type hierarchies (I dunno how to do this without dark magic)

That being said, I'm totally fine with that fix being delegated to an issue+TODO (if it doesn't already exist), as this PR is *definitely* a strict improvement.

(Is this already a documented defect in `pybind11`?)

EricCousineau-TRI (changes requested):
Looks great! However, I think you possibly uncovered a new bug :crying_cat_face: 
I don't think we need to block this PR on it, but we should at least leave paper trail (issue+TODO) for it!

tests/test_pickling.py
EricCousineau-TRI: Filed #3062

Can you refer to that, and perhaps steal some of the testing expectations I added in my branch?

rwgk:
Closing-Reopening to trigger CI.

tests/test_pickling.py
rwgk: > Filed #3062
> 
> Can you refer to that, and perhaps steal some of the testing expectations I added in my branch?

Done, although I used a `dynamic_cast` instead (before I realized there was test code under #3062).

Based on @elkhrt's [comment](https://github.com/pybind/pybind11/issues/3062#issuecomment-871499868) I believe the situation that is helped by the change here is a valid use case.

EricCousineau-TRI (approved):

PR #2988, opened 2021-05-01
Splitting out include/pybind11/detail/pragma_warning_block.h

rwgk:
Alternative to moving the pragma block to common.h. This way the code organization on master can stay exactly the same.


rwgk:
Merging to cut down on number of PRs to manage. This change is still awaiting review under #2873.

Issue #1272, opened 2018-02-08
pybind breaks exception handling


wichert:
I've spent many hours today trying to figure out why my exception handling does not work. I have a C++ library which raises an exception in its own namespace: `stylist::unsuitable_error`. Catching this error works perfectly within that library. In my wrapper code it fails. The exception is defined like this:

```c++
class unsuitable_error : public std::logic_error {
public:
	explicit unsuitable_error(const std::string& __arg) : std::logic_error(__arg), reasons(), causes() { }
	virtual ~unsuitable_error() throw() { }

	std::list<const char*> reasons;
	struct cause_type { … };
	std::list<cause_type> causes;
};
```

I have some code which tries to handle that exception being thrown:

```c++
	AdviceWrapper* doit(const Article& article, const Body& body) {
		auto result = new AdviceWrapper();

		try {
			result->advise = advise(article, body);
			cerr << "All OK" << endl;
		} catch (const unsuitable_error &e) {
			cerr << "Not suitable" << endl;
		} catch (...) {
			cerr << "Bad exception" << endl;
		}
		return result;
	}

void GO() {
  article = … ;
  body = … ;
  doit(article, body);
}
```

This works perfectly in a standalone program:

```
$ cat src/standalone.cc 
void GO();

int main() {
	GO();
	return 0;
}
$ ./standalone 
Not suitable
```

However, the *exact* same code in pybind11 breaks. The wrapper code is simple enough:

```
#include <pybind11/pybind11.h>

namespace py = pybind11;


void GO();

PYBIND11_MODULE(_stylist, m) {
	m.doc() = "CurveTips stylist";

	m.def("GO", GO);
}
```

and running that results in:

```
$ python3 -c 'import _stylist ; _stylist.GO()'
Bad exception
```

The standalone tool and Python wrapper are compiled from the same CMake config:

```
project(_stylist)
cmake_minimum_required(VERSION 2.8.12)

set(PACKAGE stylist)

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
set(Boost_USE_MULTITHREADED ON)

find_package(Stylist REQUIRED)
include_directories(SYSTEM ${Stylist_INCLUDE_DIRS})
link_directories(${Stylist_LIBRARY_DIRS})

add_subdirectory(pybind11)
pybind11_add_module(_stylist SHARED src/stylist.cc src/tst.cc)
[...]

PR #2999, opened 2021-05-11
Explicitly export exception types.

### LINKED ISSUE #2847: [BUG] register_exception_translator have different behaviour when building with different STL.
## Issue description

If I have a module `a` who have an exception translator and module `b` who throws an exception,
and import `a`, `b` in order in another python module.

With libstdc++, when `b` throws an error, it will be catched by the registered translator,
while with libc++, it won't.

Output1:
```
+ python -c 'import foo; foo.bar.thr()'
RuntimeError
```
Output2:
```
+ python -c 'import foo; foo.bar.thr()'
KeyError: ''
```

See below for detailed information.

## Reproducible example code

This repro requires additional dependencies. I recommend you run it from a docker container (`debian:buster` has been verified to work).

<details>
<summary>Install dependencies</summary>
<p>

```shell
#!/bin/sh
set -ex

apt-get update

apt-get install -y -qq python-dev libc++1 libc++-dev pybind11-dev g++
ln -s /usr/lib/llvm-7/lib/libc++abi.so.1 /usr/lib/llvm-7/lib/libc++abi.so || true
```

</p>
</details>

<details>
<summary>Generate test files</summary>
<p>

```shell
#!/bin/sh
set -ex

cat > foo.h <<EOF
#include <stdexcept>

struct e : public std::runtime_error {
    explicit e(const char* what) :
        std::runtime_error(what) {
    }
};
EOF

cat > foo.cc <<EOF

#include "foo.h"

namespace py = pybind11;

PYBIND11_MODULE(_foo, m) {
    py::register_exception_translator([](std::exception_ptr p) {
        try {
            if (p) std::rethrow_exception(p);
        } catch (const e &_e) {
            PyErr_SetString(PyExc_KeyError, _e.what());
        }
    });
};
EOF

cat > bar.cc <<EOF

#include "foo.h"

void thr() {
    throw e{""};
}

namespace py = pybind11;

PYBIND11_MODULE(bar, m)
{
    m.def("thr", &thr);
}
EOF

cat > foo.py <<EOF
import _foo

import bar
EOF
```

</p>
</details>

<details>
<summary>Build and run</summary>
<p>

```shell
#!/bin/sh
set -ex

# https://libcxx.llvm.org/docs/UsingLibcxx.html

g++ -I/usr/include/pybind11/ -I/usr/include/python2.7 \
-nostdinc++ -I/usr/lib/llvm-7/include/c++/v1 \
-fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -std=gnu++11 \
-o bar.o -c bar.cc

g++ -fPIC \
-nostdinc++ -I/usr/lib/llvm-7/include/c++/v1 \
-flto -shared \
-nodefaultlibs -L/usr/lib/llvm-7/lib -lc++ -lc++abi -lm -lc -lgcc_s -lgcc \
-o bar.so bar.o

g++ -I/usr/include/pybind11/ -I/usr/include/python2.7 \
-nostdinc++ -I/usr/lib/llvm-7/include/c++/v1 \
-fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -std=gnu++11 \
-o _foo.o -c foo.cc

g++ -fPIC \
-nostdinc++ -I/usr/lib/llvm-7/include/c++/v1 \
-flto -shared \
-nodefaultlibs -L/usr/lib/llvm-7/lib -lc++ -lc++abi -lm -lc -lgcc_s -lgcc \
-o _foo.so _foo.o

python -c "import foo; foo.bar.thr()" || true


g++ -I/usr/include/pybind11/ -I/usr/include/python2.7 \
-fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -std=gnu++11 \
-o bar.o -c bar.cc

g++ -fPIC \
-flto -shared \
-o bar.so bar.o

g++ -I/usr/include/pybind11/ -I/usr/include/python2.7 \
-fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -std=gnu++11 \
-o _foo.o -c foo.cc

g++ -fPIC \
-flto -shared \
-o _foo.so _foo.o

python -c "import foo; foo.bar.thr()" || true

rm -rf *.o *.so
```

</p>
</details>

oraluben:
## Description

closes #2847 by setting visibility of exceptions to default. cc @XZiar
closes #3016

## Suggested changelog entry:

```rst
``pybind11::builtin_exception`` is now explicitly exported, which means the types included/defined in different modules are identical, and exceptions raised in different modules can be caught correctly. The documentation was updated to explain that custom exceptions that are used across module boundaries need to be explicitly exported as well.
```

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -164,9 +164,9 @@ section.
     may be explicitly (re-)thrown to delegate it to the other,
     previously-declared existing exception translators.
 
-    Libc++ `behaves differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
-    with ``-fvisibility=hidden``. So you'll need to explicitly export exceptions who'll be used across ABI boundaries as ``tests/test_exceptions.h``.
-    See also section "Problems with C++ exceptions" in `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
+    Note that ``libc++`` and ``libstdc++`` `behave differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
+    with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI boundaries need to be explicitly exported, as exercised in ``tests/test_exceptions.h``.
+    See also: "Problems with C++ exceptions" under `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
 
 .. _handling_python_exceptions_cpp:
 
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -164,6 +164,9 @@ section.
     may be explicitly (re-)thrown to delegate it to the other,
     previously-declared existing exception translators.
 
+    Libc++ `behaves differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
+    with ``-fvisibility=hidden``. So you'll need to explicitly export exceptions who'll be used across ABI boundaries as ``tests/test_exceptions.h``.
+
 .. _handling_python_exceptions_cpp:
 
 Handling exceptions from Python in C++
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -166,6 +166,7 @@ section.
 
     Libc++ `behaves differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
     with ``-fvisibility=hidden``. So you'll need to explicitly export exceptions who'll be used across ABI boundaries as ``tests/test_exceptions.h``.
+    See also section "Problems with C++ exceptions" in `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
 
 .. _handling_python_exceptions_cpp:
 
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -3,6 +3,8 @@ import sys
 
 import pytest
 
+import env  # noqa: F401
+
 from pybind11_tests import exceptions as m
 import pybind11_cross_module_tests as cm
 
@@ -43,7 +45,14 @@ def test_cross_module_exceptions():
     with pytest.raises(StopIteration) as excinfo:
         cm.throw_stop_iteration()
 
-    with pytest.raises(KeyError) as excinfo:
+
+# TODO: FIXME
+@pytest.mark.skipif(
+    "env.PYPY and env.MACOS",
+    reason="Known failure with PyPy and libc++ (Issue #2847 & PR #2999)",
+)
+def test_cross_module_exception_translator():
+    with pytest.raises(KeyError) as _:
         # translator registered in cross_module_tests
         m.throw_should_be_translated_to_key_error()
 
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -50,10 +50,10 @@ def test_cross_module_exceptions():
 @pytest.mark.xfail(
     "env.PYPY and env.MACOS",
     raises=RuntimeError,
-    reason="Known failure with PyPy and libc++ (Issue #2847 & PR #2999)",
+    reason="Expected failure with PyPy and libc++ (Issue #2847 & PR #2999)",
 )
 def test_cross_module_exception_translator():
-    with pytest.raises(KeyError) as _:
+    with pytest.raises(KeyError):
         # translator registered in cross_module_tests
         m.throw_should_be_translated_to_key_error()
 
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -43,6 +43,9 @@ def test_cross_module_exceptions():
     with pytest.raises(StopIteration) as excinfo:
         cm.throw_stop_iteration()
 
+    with pytest.raises(KeyError) as excinfo:
+        m.throw_()
+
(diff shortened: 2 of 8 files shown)


rwgk:
The CI looks good.
My only concern is that we don't have a new test that failed before but works with this PR, so that we don't accidentally undo your fix in the future. How much trouble would it be to add?

oraluben:
Hi @rwgk , the idea sounds good. Currently, I'm only aware of the `test_<test>.{py,cpp}` mechanism, but this test requires a custom build/include logic, what's the proper way to do this?

rwgk:
> Hi @rwgk , the idea sounds good. Currently, I'm only aware of the `test_<test>.{py,cpp}` mechanism, but this test requires a custom build/include logic, what's the proper way to do this?

I'm not very familiar with the exception handling code in pybind11 myself, but at first sight there is hope it could be straightforward adding to existing tests:

* https://github.com/pybind/pybind11/blob/master/tests/test_exceptions.cpp
* https://github.com/pybind/pybind11/blob/master/tests/test_exceptions.py
* https://github.com/pybind/pybind11/blob/f61855b9d8821e2576960109a2a67379a6c2366f/tests/pybind11_cross_module_tests.cpp#L27

Do those look like a useful starting point?
What makes your situation a little more tricky is that you need to test interactions between two modules IIUC, but hopefully you can just add to those files above without having to go into the cmake files.

JIC you haven't see this already, this includes instructions for running the tests: https://github.com/pybind/pybind11/blob/master/.github/CONTRIBUTING.md

oraluben:
Hi @rwgk, I've added a test and updated some docs, does that look good to you? (I'll refine some word/name, e.g. `tmp_e`.)

rwgk:
Hi @oraluben, the macos PyPy tests are failing. Did you see that already?
Note that we already skip several tests in the PyPy tests, in case there is no obvious or reasonable fix. But making it work is best.

oraluben:
I take a look at it but can't understand at all. I haven't used Pypy, but I doubt that Pypy has special library handling/importing behaviour, because after setting the visibility of exceptions to default, the hashcodes of thrown/caught exceptions are still different, which cause the `catch` statement to malfunction.

I'll just skip the test when pypy+libc++ for now and see if I can find more later.

rwgk:
Almost! :-)
Just Format failing now.
I always blindly run `pre-commit run --all-files` before `git commit` (see .github/CONTRIBUTING.md).

oraluben:
format fixed and passed `pre-commit run --all-files` now.

oraluben:
Found related #1272, trying to remove unused workaround code.

oraluben:
The latest commit removes some old workaround for this issue, which passed tests locally. If there's any failure in the CI (most likely in PyPy) I'll revert it.

rwgk:
> The latest commit removes some old workaround for this issue, which passed tests locally. If there's any failure in the CI (most likely in PyPy) I'll revert it.

Yep, PyPy failing again.

BTW: I still need to find time to verify that the new test fails without the core change, and run this PR through our Google-internal testing (no issues expected, but standard practice). Hopefully later today.

oraluben:
> Yep, PyPy failing again.

reverted & doc updated.

> BTW: I still need to find time to verify that the new test fails without the core change, and run this PR through our Google-internal testing (no issues expected, but standard practice). Hopefully later today.

test-only version can be found at #3016, failed as expected.

rwgk:
> test-only version can be found at #3016, failed as expected.

Awesome, thanks! Verified by triggering the CI.

docs/advanced/exceptions.rst
```diff
@@ -164,6 +164,10 @@ section.
     may be explicitly (re-)thrown to delegate it to the other,
     previously-declared existing exception translators.
 
+    Libc++ `behaves differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
```
rwgk, line 167: Suggested edit:
```
    Note that ``libc++`` and ``libstdc++`` `behave differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
    with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI boundaries need to be explicitly exported, as exercised in ``tests/test_exceptions.h``.
    See also: "Problems with C++ exceptions" under `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
```

tests/test_exceptions.py
```diff
@@ -44,6 +46,18 @@ def test_cross_module_exceptions():
         cm.throw_stop_iteration()
 
 
+# TODO: FIXME
+@pytest.mark.xfail(
+    "env.PYPY and env.MACOS",
+    raises=RuntimeError,
+    reason="Known failure with PyPy and libc++ (Issue #2847 & PR #2999)",
```
rwgk, line 53: Maybe "Expected" instead of "Known"?

tests/test_exceptions.py
```diff
@@ -44,6 +46,18 @@ def test_cross_module_exceptions():
...
+)
+def test_cross_module_exception_translator():
+    with pytest.raises(KeyError) as _:
```
rwgk, line 56: You don't need the `as _` (please remove).

rwgk (commented):
Looks great, just some minor suggestions.

rwgk:
The Google-internal global testing came back successful, too. We're ready to merge this PR, but could you please review the "Suggested changelog entry" I added to the PR description? Please modify as needed. I'll merge this PR after you confirm the Description and Suggested change log entry are final.

oraluben:
Thanks for your help! "setting visibility" is not concrete on the second thought, since the statement is Linux-only but this PR actually affects MSVC (see failed Windows CIs in #3016). I've updated the title and some parts of the Suggested changelog entry, the rest looks perfect! The messages are final if they look good to you.

PR #3005, opened 2021-05-14
use CMAKE_INSTALL_FULL_INCLUDEDIR

dotlambda:
## Description
see https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path

## Suggested changelog entry:

```rst
fix pybind11_INCLUDE_DIR in case CMAKE_INSTALL_INCLUDEDIR is absolute
```


henryiii (approved):
This looks correct to me, and it available at least since 3.4 (only one I checked). Might need to look over that list and update https://cliutils.gitlab.io/modern-cmake/ .

Issue #3011, opened 2021-05-18
[smart_holder] RuntimeError: Invalid return_value_policy for shared_ptr


jakobandersen:
While trying to convert from Boost.Python I ran into the lifetime issue raised, e.g., in https://github.com/pybind/pybind11/pull/2839, so I am using the smart_holder branch.
A minimal test case can be found at https://github.com/jakobandersen/pyTest/tree/pybind, with the central part at https://github.com/jakobandersen/pyTest/blob/pybind/libpytest.cpp#L42.

Essentially I have a custom function class which can be derived in Python. The problem is occurs when the function takes a ``std::shared_ptr`` as argument and is called from C++. When using ``PYBIND11_OVERRIDE_PURE_NAME`` to implement the call back to the Python implementation it will call the override with the default ``return_value_policy``, which seems to be ``automatic_reference``.
This gives the exception ``RuntimeError: Invalid return_value_policy for shared_ptr.``.
Expanding the macro and giving ``automatic`` as policy makes it work as intended (as done in https://github.com/jakobandersen/pyTest/blob/pybind/libpytest.cpp#L58).

Am I missing something in the setup? Is this manual hax to make it work correct?

jakobandersen:
A simpler minimal case than above:
```c++
class C {
	C(int data);
public:
	int data;
public:
	static std::shared_ptr<C> make(int data) {
		return std::shared_ptr<C>(new C(data));
	}
};

void trigger() {
	auto l = py::list();
	auto c = C::make(42);
	l.append(c); // throws RuntimeError: Invalid return_value_policy for shared_ptr.
}
```
I'm not sure if this intersects the same code as the original example.

jakobandersen:
As noted in #3012, the original issue seems fixed in the smart_holder branch, but the second issue is still present.

rwgk:
Hi @jakobandersen, echoing @rhaschke 's request: https://github.com/pybind/pybind11/pull/3012#issuecomment-861610697

That would help us working on the issue together.

BTW: I will respond to PRs much faster, I just don't have the free bandwidth to pay much attention to the pybind11 github issues.

jakobandersen:
> Hi @jakobandersen, echoing @rhaschke 's request: [#3012 (comment)](https://github.com/pybind/pybind11/pull/3012#issuecomment-861610697)
> 
> That would help us working on the issue together.

Sure, I started working in #3039, at least for adding tests for the cases I ran into and see if a simple change works out.

[...]

PR #3012, opened 2021-05-20
smart_holder fixups

rhaschke:
These are some fixups for the smart_holder branch. See individual commit messages for details. @rwgk

### DIFF
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -625,8 +625,8 @@ struct smart_holder_type_caster<std::shared_ptr<T>> : smart_holder_type_caster_l
     static handle cast(const std::shared_ptr<T> &src, return_value_policy policy, handle parent) {
         auto src_raw_ptr = src.get();
         auto st          = type_caster_base<T>::src_and_type(src_raw_ptr);
-        if (st.first == nullptr)
-            return none().release(); // PyErr was set already.
+        if (st.second == nullptr)
+            return handle(); // no type info: error will be set already
 
         void *src_raw_void_ptr         = static_cast<void *>(src_raw_ptr);
         const detail::type_info *tinfo = st.second;
@@ -690,8 +690,8 @@ struct smart_holder_type_caster<std::unique_ptr<T, D>> : smart_holder_type_caste
 
         auto src_raw_ptr = src.get();
         auto st          = type_caster_base<T>::src_and_type(src_raw_ptr);
-        if (st.first == nullptr)
-            return none().release(); // PyErr was set already.
+        if (st.second == nullptr)
+            return handle(); // no type info: error will be set already
 
         void *src_raw_void_ptr         = static_cast<void *>(src_raw_ptr);
         const detail::type_info *tinfo = st.second;
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -623,6 +623,9 @@ struct smart_holder_type_caster<std::shared_ptr<T>> : smart_holder_type_caster_l
     static constexpr auto name = _<std::shared_ptr<T>>();
 
     static handle cast(const std::shared_ptr<T> &src, return_value_policy policy, handle parent) {
+        if (!src)
+            return none().release();
+
         auto src_raw_ptr = src.get();
         auto st          = type_caster_base<T>::src_and_type(src_raw_ptr);
         if (st.second == nullptr)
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/type_caster_base.h
```diff
@@ -739,16 +739,9 @@ class type_caster_generic {
...
-
-        // Not found, set error:
-        std::string tname = rtti_type ? rtti_type->name() : cast_type.name();
-        detail::clean_type_id(tname);
-        std::string msg = "Unregistered type : " + tname;
-        PyErr_SetString(PyExc_TypeError, msg.c_str());
-        return {nullptr, nullptr};
+            const void *src, const std::type_info &cast_type) {
+        auto *tpi = get_type_info(cast_type, true);
+        return {src, const_cast<const type_info *>(tpi)};
```
rhaschke, line 744: I just noticed that this change should be applied to master. I will file a separate PR for that.

rhaschke:
@rwgk, given that you have forbidden passing of a `std::shared_ptr` from C++ to Python before, I guess you also don't yet have roundtrip tests for that use case, i.e. C++ -> Python -> C++  and  Python -> C++ -> Python.

rhaschke:
Sorry, obviously it was too late yesterday evening when I wrote this: I neglected that you did handle the cases of `automatic` and `reference_internal`. I was particularly missing `automatic_reference`, which is the default for (C++) arguments passed to python functions from C++.  
Regarding `copy` and `move`: these are the actual policies `automatic` would resolve to for non-holder values. For this reason, these two should be definitely supported. Also, if you support `reference_internal`, you should support `reference` as well. They don't differ in the handling of the argument itself, but only in the life-time handling of its parent.
I agree that `take_ownership` doesn't really make sense for a shared pointer. On the other side, it doesn't harm to allow this policy and handle it just like all others. It would be the only one throwing an exception. Allowing it too, we can skip the test completely.

rwgk:
I just re-read the documentation for the return value policies. The documentation for `reference` explicitly mentions "do not take ownership", which in the context of `shared_ptr` is unclear/weird. For `copy` and `move` I still feel strongly that those are also best not allowed to avoid guessing and confusion about what they may do. I could be convinced otherwise if those restrictions generate burdensome situations elsewhere, but at the moment I'm thinking simply guiding people to not specify a return value policy that doesn't make a difference anyway is the right thing to do.

If someone really wants to copy or move a value pointed at by a `shared_ptr`, they can do that easily via a lambda, dereferencing the `shared_ptr`. Then it's clear and obvious in the bindings code that something unusual is happening, and that it's not different from the behavior for a raw pointer return.

The other changes look good. If you put back the return value policy test I'll merge this PR (after also testing google-internally). We could continue the return value policy discussion separately if needed.

rwgk:
Thanks for the fixes! I'll merge this now.

> So, why would you restrict the allowed values?

The main reason is to not lead people into "what does this do?" guessing. Some will just always harbor doubts but never look. That's just wasting a few cognitive cycles. Others (most?) will just copy-paste what they see without understanding. At large scale (our code base has 100s of millions of lines) that copy-paste clutter has a tendency to turn into life-consuming cleanup jobs, when at some point later, it is found that e.g. `copy` or `move` actually can have useful semantics. Then those previously meaningless policies scattered around the code base need to be identified and removed. — Coincidentally, I just spent more than a week on something like that. I want to do everything I can to weed out the potential for such a waste of time right at the root.

jakobandersen:
This fixed my original issue in #3011. However, the second issue (https://github.com/pybind/pybind11/issues/3011#issuecomment-846080517) still gives the error ``RuntimeError: Invalid return_value_policy for shared_ptr.``.
I think I traced it to https://github.com/pybind/pybind11/blob/smart_holder/include/pybind11/cast.h#L905 where the ``automatic`` and ``automatic_reference`` policies are prune before going further with the casting.

rhaschke:
I guess the `automatic` return value policy is concretized to `copy` or `move` in the referenced code snippet and then rejected by the [`smart_holder_type_caster` code](https://github.com/pybind/pybind11/pull/3012/files#diff-f58376a4b8a5dbcf9feb803452bf2212035ad2922510ad05689c4cd610db8ffbR627-R628). Thus, I think we should allow these policies as well. 
@jakobandersen: do you mind preparing a corresponding PR to validate this assumption?

rwgk:
In the meantime, while working on #3023, I found a combination of `std::shared_ptr` & `return_value_policy::copy` here:

https://github.com/pybind/pybind11/blob/4c7697dbe973ed01ae6fbec37d186ebd05982e1f/tests/test_smart_ptr.cpp#L376-L380

We need to ensure we're compatible with existing behavior, or document the differences if we decide it's better to deviate. — I didn't look enough to be sure what exactly we need to do.

PR #3013, opened 2021-05-20
Unify cast_error message thrown by [simple|unpacking]_collector

rhaschke:
`simple_collector` and `unpacking_collector` throw different error messages when the casting of an argument failed: 
While the former mentioned `make_tuple()`, the latter emphasizes the call argument (and its name/position):
https://github.com/pybind/pybind11/blob/ad6bf5cd39ca64b4a9bf846b84b11c4c8df1c8e1/include/pybind11/cast.h#L1307-L1309
This PR unifies both usages to use the latter (more informative) one.
If there was a reason to have different error messages for both collector variants, please feel free to close this PR.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -971,11 +971,11 @@ template <return_value_policy policy = return_value_policy::automatic_reference,
     for (size_t i = 0; i < args.size(); i++) {
         if (!args[i]) {
 #if defined(NDEBUG)
-            throw cast_error("make_tuple(): unable to convert arguments to Python object (compile in debug mode for details)");
+            throw cast_error("Unable to convert call argument to Python object (compile in debug mode for details)");
 #else
             std::array<std::string, size> argtypes { {type_id<Args>()...} };
-            throw cast_error("make_tuple(): unable to convert argument of type '" +
-                argtypes[i] + "' to Python object");
+            throw cast_error("Unable to convert call argument '" + std::to_string(i) +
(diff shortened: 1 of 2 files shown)


rhaschke:
Closing and reopening to trigger CI.

rwgk (commented):
I don't have historical background to know why the messages aren't uniform, but it seems likely to me that this simply didn't get enough attention and just slipped through the cracks. I think this is a good change to make. My only suggestion is to add simple unit tests for both, if possible actually one unit test that exercises both implementations, to ensure that the implementations are being kept consistent in the future.

Skylion007 (approved):

rwgk:
I rebased this branch on current master and added a commit to consolidate the error reporting code. Waiting for the CI to finish.

rwgk:
@Skylion007 Could you please take another look?

With respect to my comment from June 1: https://github.com/pybind/pybind11/pull/3013#pullrequestreview-673435643

When looking at this PR again today, I decided it's better to ensure uniformity at the root (and reduce code duplication) rather than asserting something about the results (commit https://github.com/pybind/pybind11/pull/3013/commits/53fc821fc6cbf0465d2d45fe6afcfa7e6371d106).

I verified that the existing unit tests cover both call paths affected by the commit.

Skylion007 (approved):

rwgk:
Thanks @rhaschke for initiating this PR, and thanks @Skylion007 for the review!

PR #3038, opened 2021-06-15
Docs: Update py::kwargs example in function.rst to pass by reference

Skylion007:
## Description

A very common clang-tidy performance issue I have noticed is that py::dict is not trivially copyable, but I can't think of any case where where you want to actually create a shallow copy of the dict intentionally. Furthermore, not creating a shallow-copy of the dict would better match pythonic syntax in the C++ side. As such, I propose editing the documentation to have this be the default practice. There are probably a lot more areas where this could be improved, but here are the two most common pieces of code I see being copy and pasted around.

As an aside, I'd recommend enabling the clang-tidy performance checks at some point to catch these issues in the tests and other places. Not that they are performance sensitive per say, but that they should show best coding practices.

## Suggested changelog entry:

```rst
* Improved documentation for best practices regarding ``py::dict`` and ``py::kwargs``
```

### DIFF
--- a/docs/advanced/functions.rst
+++ b/docs/advanced/functions.rst
@@ -254,7 +254,7 @@ For instance, the following statement iterates over a Python ``dict``:
 
 .. code-block:: cpp
 
-    void print_dict(py::dict dict) {
+    void print_dict(const py::dict& dict) {
         /* Easily interact with Python types */
         for (auto item : dict)
             std::cout << "key=" << std::string(py::str(item.first)) << ", "
@@ -292,7 +292,7 @@ Such functions can also be created using pybind11:
 
 .. code-block:: cpp
 
-   void generic(py::args args, py::kwargs kwargs) {
+   void generic(py::args args, const py::kwargs& kwargs) {
        /// .. do something with args
        if (kwargs)
            /// .. do something with kwargs


rwgk:
The Centos 8 failure is unrelated (it appeared ~June 4; we're working on it).

rwgk:
Thanks for this change!

I really like your clang-tidy suggestion, but unless we find a volunteer to add this to the GitHub Actions it's very unlikely to happen.

Skylion007:
@rwgk Happy to do it, but the only way I know how to do to it properly is to have a hierarchical config (since we don't want it making changes to the obvious copies and movies that we do in the tests folder). These would require a new version of Clang-Tidy (>12) and the github action only uses clang-tidy 10. Happy to set this up if you don't mind bumping the clang-tidy version: https://reviews.llvm.org/D75184

rwgk:
> @rwgk Happy to do it, but the only way I know how to do to it properly is to have a hierarchical config (since we don't want it making changes to the obvious copies and movies that we do in the tests folder). These would require a new version of Clang-Tidy (>12) and the github action only uses clang-tidy 10. Happy to set this up if you don't mind bumping the clang-tidy version: https://reviews.llvm.org/D75184

I'm up for trying, assuming we can roll back to 10 if trying 12 gets us into trouble.
@henryiii to you have an opinion?

I'm not very familiar with the GitHub Action tooling. Is bumping the clang-tidy version just another PR? In that case I'd say let's just try and make decisions based on the outcome.

Skylion007:
@rwgk In the meantime, I applied the performance optimizations from clang-tidy in another PR: https://github.com/pybind/pybind11/pull/3046 . There were not too many (as expected of a well optimized codebase), but there were several on relatively hot code paths including arg_v, static_properties and such.

henryiii:
Sure, feel free to bump https://github.com/pybind/pybind11/blob/b4b67f026bc2aa1a1ae17e9858a9e61210c77c31/.github/workflows/format.yml#L30 to any supported number. I can do that if you prefer.

PR #3023, opened 2021-05-27
shared_from_this experiments

rwgk:
Experiments.

PR #3039, opened 2021-06-16
[smart_holder] return value policies allowed for shared_ptr

jakobandersen:
## Description

Allow ``copy`` and ``move`` policy for smart holder casting of ``std::shared_ptr``.

## Suggested changelog entry:

TODO?

```rst

```


rwgk:
Unfortunately I have to manually trigger the workflows each time you push something new, and I'm not getting automatic notifications. Please tag me here each time you want to run the CI (some dummy comment with @rwgk).

To make the Format workflow happy, I always blindly run `pre-commit run --all-files` before a git commit. See .github/CONTRIBUTING.md.

Maybe for later/general background: test_smart_ptrs.cpp,py is one of the few tests that does NOT exercise any smart_holder functionality. It would be better if you added to some test_class_sh_*.cpp,py or ideally simply make a new test_class_sh_xxx.cpp,py pair. The extra overhead is super tiny (one more line in tests/CMakeLists.txt). Then we can decide later if we want to keep the new tests separate, or fold them into something else (also super easy).

About the `copy` policy, when enabling it, I'd want to be certain we emulate the existing behavior exactly. We could either do this via a test that uses type_caster_base in CI-SH-AVL, and smart_holder in CI-SH-DEF, i.e. via the `PYBIND11_SH_AVL` macro, although I think the alternative of having one test that exercises both simultaneously would be ideal. It's actually pretty simple, combining these two tricks:

* https://github.com/pybind/pybind11/blob/8f6ca71390f4cc1faf8e060146f347fa387ecd6a/tests/test_class_sh_basic.cpp#L12
* https://github.com/pybind/pybind11/blob/8f6ca71390f4cc1faf8e060146f347fa387ecd6a/tests/test_class_sh_disowning.cpp#L10

The first tracks copy and move, the second adds a serial number, so that you can wrap with both `py::class_<Atype<0>, std::shared_ptr<Atype<0>>>` and `py::classh<Atype<1>>` in the same test\_\*.cpp, then in test\_\*.py you can loop over both wrapped types, for the exact same tests.

tests/test_smart_ptr.cpp
```diff
@@ -353,6 +363,15 @@ TEST_SUBMODULE(smart_ptr, m) {
...
     m.def("print_myobject3_4", [](const std::shared_ptr<MyObject3> *obj) { py::print((*obj)->toString()); });
 
+    py::class_<MyObject3a>(m, "MyObject3a");
+    m.def("make_myobject3_1", []() { return new MyObject3a(8); });
+    m.def("make_myobject3_2", []() { return std::make_shared<MyObject3a>(9); });
+    m.def("print_myobject3a_1", [](const MyObject3a *obj) { py::print(obj->toString()); });
+    m.def("print_myobject3a_2", [](std::shared_ptr<MyObject3a> obj) { py::print(obj->toString()); });
+    m.def("print_myobject3a_3", [](const std::shared_ptr<MyObject3a> &obj) { py::print(obj->toString()); });
+    // this doesn't compile, should it?
+    //m.def("print_myobject3a_4", [](const std::shared_ptr<MyObject3a> *obj) { py::print((*obj)->toString()); });
```
rhaschke, line 373: No. A `shared_ptr` is expected to be passed by some kind of reference (or value).

tests/test_smart_ptr.cpp
```diff
@@ -353,6 +363,15 @@ TEST_SUBMODULE(smart_ptr, m) {
     m.def("print_myobject3_3", [](const std::shared_ptr<MyObject3> &obj) { py::print(obj->toString()); });
     m.def("print_myobject3_4", [](const std::shared_ptr<MyObject3> *obj) { py::print((*obj)->toString()); });
 
+    py::class_<MyObject3a>(m, "MyObject3a");
+    m.def("make_myobject3_1", []() { return new MyObject3a(8); });
+    m.def("make_myobject3_2", []() { return std::make_shared<MyObject3a>(9); });
+    m.def("print_myobject3a_1", [](const MyObject3a *obj) { py::print(obj->toString()); });
+    m.def("print_myobject3a_2", [](std::shared_ptr<MyObject3a> obj) { py::print(obj->toString()); });
+    m.def("print_myobject3a_3", [](const std::shared_ptr<MyObject3a> &obj) { py::print(obj->toString()); });
```
rhaschke, line 371: All these functions are not yet used, are they?

rhaschke (commented):
At first sight, I don't see a particular reason for the double-free error - unfortunately.

tests/test_smart_ptr.cpp
rwgk: > No. A `shared_ptr` is expected to be passed by some kind of reference (or value).

I'm thinking of allowing const pointers, but definitely not non-const pointers.
Rationale: It could be useful to pass `nullptr`, that seems like a valid use case.

rhaschke: A `nullptr` can be represented by an empty `shared_ptr`. IMO we always want to pass a pointer to a held object.
The `shared_ptr` is just a convenient replacement for a raw pointer. Passing a pointer to a (shared) pointer doesn't make sense to me - no matter being it const or not: We cannot modify the pointer anyway.

rwgk: > A `nullptr` can be represented by an empty `shared_ptr`. IMO we always want to pass a pointer to a held object.
> The `shared_ptr` is just a convenient replacement for a raw pointer. Passing a pointer to a (shared) pointer doesn't make sense to me - no matter being it const or not: We cannot modify the pointer anyway.

OK, let's keep this as-is for now. Let's just wait and see if anyone later has a strong need for it.

tests/test_smart_ptr.cpp
rwgk: FWIW, I copied this also under my WIP #3023, but at the moment I'm thinking this block doesn't add anything in the smart_holder context. Focusing this PR narrowly on the `copy` return value policy question seems best to me.

jakobandersen: Right, these were just copy-paste artefacts, I'll clean it up when moving the test to a separate file.

jakobandersen:
@rwgk, thanks for the explanations. I think I made essentially the changes you suggested, and it is ready for another CI run.
The candidate fix is still the last commit (I'll squash the first two later), and before it the ``test_avl()`` fails with the casting error.

rwgk:
Thanks, that looks great! Please don't worry about squashing. In my experience it's easiest and best to not mess with the history. The PR will be squashed automatically when merging.

rwgk:
Please keep tagging me here for triggering the CI. 10 times a day is fine, no worries.

jakobandersen:
Ping @rwgk for CI :-).

tests/test_class_sh_shared_ptr_copy.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    m.def("test_avl", []() {
+        auto o = std::make_shared<FooAVL>("AVL");
+        auto l = py::list();
+        l.append(o);
+    });
+    m.def("test_def", []() {
+      auto o = std::make_shared<FooDEF>("DEF");
+      auto l = py::list();
+      l.append(o);
```
rhaschke, line 46: Previously, you also had a test for `l.append(std::move(o));`
This will require the `move` policy, which is not yet supported. Is there a reason, you dropped this?

include/pybind11/detail/smart_holder_type_casters.h
```diff
@@ -625,6 +625,7 @@ struct smart_holder_type_caster<std::shared_ptr<T>> : smart_holder_type_caster_l
     static handle cast(const std::shared_ptr<T> &src, return_value_policy policy, handle parent) {
         if (policy != return_value_policy::automatic
             && policy != return_value_policy::reference_internal
+            && policy != return_value_policy::copy
```
rhaschke, line 628: Why don't you need `move` as well??

jakobandersen: Indeed that is needed, should be added now.

tests/test_class_sh_shared_ptr_copy.cpp
jakobandersen: Indeed, I simply forgot while refactoring it. Added again.

rhaschke (approved):
LGTM.

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,62 @@
...
+PYBIND11_SMART_HOLDER_TYPE_CASTERS(pybind11_tests::class_sh_shared_ptr_copy_move::FooAVL)
+
+namespace pybind11_tests {
+namespace class_sh_shared_ptr_copy_move {
+
+TEST_SUBMODULE(class_sh_shared_ptr_copy_move, m) {
+    namespace py = pybind11;
+
+    py::class_<FooAVL, PYBIND11_SH_AVL(FooAVL)>(m, "FooAVL");
+    py::class_<FooDEF, PYBIND11_SH_DEF(FooDEF)>(m, "FooDEF");
```
rwgk, line 36: Suggested change:
```
    py::class_<FooShPtr, std::shared_ptr<FooAVL>>(m, "FooShPtr");
    py::classh<FooSmHld>(m, "FooSmHld");
```

What I had in mind is unconditionally comparing the `std::shared_ptr` vs `py::smart_holder` behavior.
Feel free to pick different variable names.

You will also need the `PYBIND11_TYPE_CASTER_BASE_HOLDER` macro for `FooShPtr`.

To make that easier to read, in more recent code I'm using the anonymous `namespace` instead of `pybind11_tests::class_sh_shared_ptr_copy_move`. I think it looks much better and is just as safe.

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    m.def("test_avl_copy", []() {
+        auto o = std::make_shared<FooAVL>("AVL");
+        auto l = py::list();
+        l.append(o);
```
rwgk, line 41:
```
return l;
```
(in all 4 functions)

tests/test_class_sh_shared_ptr_copy_move.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+
+from pybind11_tests import class_sh_shared_ptr_copy_move as m
+
+
+def test_avl_copy():
+    m.test_avl_copy()
```
rwgk, line 7: To be sure the copy and move behavior is as expected, something like:
```
    l = m.test_shp_copy()
    assert m.get_mtxt(l[0]) == "what we expect"
```
You probably have to allow for differences between platforms, using regular expressions, ideally this function from @rhaschke's still-open PR:

https://github.com/pybind/pybind11/pull/2916/files#diff-e614dc3a68da8c26363419d3c06816825b1b4344e35103c5dd004f7dcf3a6f84

```
def check_regex(expected, actual):
    result = re.match(expected + "$", actual)
    if result is None:
        pytest.fail("expected: '{}' != actual: '{}'".format(expected, actual))
```

For this you will also need the `get_mtxt` helper from test_class_sh_basic.cpp.

tests/test_class_sh_shared_ptr_copy_move.cpp
jakobandersen: Ah, ok. Should be updated accordingly now.

tests/test_class_sh_shared_ptr_copy_move.py
jakobandersen: Good idea,. Though, implemented as exact string compare, as in this case there should be no modification of the Foo objects after construction. The copy/move constructors are now deleted, just to ensure that.

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+template <int SerNo>
+struct Foo {
+    std::string mtxt;
+    Foo(const std::string &mtxt_) : mtxt(mtxt_) {}
+    Foo(const Foo &other) = delete;
+    Foo(Foo &&other) = delete;
+    std::string get_text() const {
+        std::string res = "Foo";
+        if (SerNo == 0)
```
rwgk, line 20: The appveyor build is (and probably all our 30+ MSVC builds would be) failing:
```
C:\projects\pybind11\tests\test_class_sh_shared_ptr_copy_move.cpp(20): warning C4127: conditional expression is constant [C:\projects\pybind11\tests\pybind11_tests.vcxproj]
C:\projects\pybind11\tests\test_class_sh_shared_ptr_copy_move.cpp(22): warning C4127: conditional expression is constant [C:\projects\pybind11\tests\pybind11_tests.vcxproj]
```

My recommendation is to keep the code here as simple as possible (so it's easier to reason about) and to use this approach here:

https://github.com/pybind/pybind11/blob/29fafcc1b28f27512715f28908fd00a946ba5d67/tests/test_class_sh_basic.cpp#L96

Then set the initial `mtxt` from Python.
BTW in some more recent code I called it `history`, if you like that better.

tests/test_class_sh_shared_ptr_copy_move.py
rwgk: I have to think about this one. We don't want to leave any holes and want to know exactly what happens in all situations incl. different platforms. Maybe we need multiple `Foo` types with different combinations of (copy, move) x (ctor, assignment-op)? Is some or all of that obviously non-sensical? But let's get the CI green first with what you have right now.

include/pybind11/detail/smart_holder_type_casters.h
```diff
@@ -623,11 +623,16 @@ struct smart_holder_type_caster<std::shared_ptr<T>> : smart_holder_type_caster_l
     static constexpr auto name = _<std::shared_ptr<T>>();
 
     static handle cast(const std::shared_ptr<T> &src, return_value_policy policy, handle parent) {
-        if (policy != return_value_policy::automatic
-            && policy != return_value_policy::reference_internal
-            && policy != return_value_policy::automatic_reference) {
-            // SMART_HOLDER_WIP: IMPROVABLE: Error message.
-            throw cast_error("Invalid return_value_policy for shared_ptr.");
+        switch (policy) {
```
rwgk, line 626: I really like this solution!

tests/test_class_sh_shared_ptr_copy_move.py
jakobandersen: Is there anywhere where the pybind code is conditional on the class being moveable/copyable?
It may be good to add a move-only Foo and copyable Foo to test that indeed the object is not modified in any way, and it's only pointer copying.

tests/test_class_sh_shared_ptr_copy_move.cpp
jakobandersen: Warning fixed and member renamed to ``history``.

jakobandersen:
Ping @rwgk for CI, now that Appveyor is green.

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+namespace pybind11_tests {
+namespace {
+
+const std::string fooNames[] = {"ShPtr_", "SmHld_"};
+
+template <int SerNo>
+struct Foo {
+    std::string hisotry;
```
rwgk, line 16: Consistently misspelled, accident?

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,71 @@
...
+    Foo(const std::string &hisotry_) : hisotry(hisotry_) {}
+    Foo(const Foo &other) = delete;
+    Foo(Foo &&other)      = delete;
```
rwgk, line 19: The purpose of the `history` member is completely lost: the move and copy assignment operators are not `delete`d, but also not tracked. How those behave is exactly what we want to test here. (At the moment idk what our conclusion will be when we see the results on all platforms, but it's important that we know what they are.)

rwgk (commented):
> Is there anywhere where the pybind code is conditional on the class being moveable/copyable?

Yes, there are certainly conditionals, I think both compile-time (using type traits) and runtime (looking at refcounts). In and around cast.h. I don't know what all the details are. The behavior is generally well exercised in test_copy_move.cpp/py.

I'll trigger the CI now although the missing history tracking means we won't get much information.

rwgk:
The many MSVC failures are due to some unrelated issue with downloading catch.hpp. I see it in other PRs as well. I'm trying to find out what's wrong.

tests/test_class_sh_shared_ptr_copy_move.cpp
```diff
@@ -0,0 +1,71 @@
...
+};
+
+using FooShPtr = Foo<0>;
+using FooSmHld = Foo<1>;
+
+} // namespace
+} // namespace pybind11_tests
+
+PYBIND11_TYPE_CASTER_BASE_HOLDER(pybind11_tests::FooShPtr)
+PYBIND11_SMART_HOLDER_TYPE_CASTERS(pybind11_tests::FooSmHld)
```
rwgk, line 30: I overlooked this before, this needs to be (untested but I hope it's correct; there are many examples in e.g. test_smart_ptr.cpp):
```
PYBIND11_TYPE_CASTER_BASE_HOLDER(pybind11_tests::FooShPtr, std::shared_ptr<pybind11_tests::FooShPtr>)
```

rwgk:
Another class of multiple CI failures are sad old compilers that complain about implicit copy constructors. Annoyingly we have to make them explicit to keep those compilers happy. I usually simply add them in without #ifdef clutter, but a comment, e.g. in test_class_sh_inheritance.cpp (`// Some compilers complain about implicitly defined versions of some of the following:`). That's a little overkill but I didn't want to torture our CI too much or spend a lot of time messing with docker containers.

tests/test_class_sh_shared_ptr_copy_move.cpp
jakobandersen: Oops, fixed.

tests/test_class_sh_shared_ptr_copy_move.cpp
jakobandersen: Indeed, there should now be tracking of the special member functions.

tests/test_class_sh_shared_ptr_copy_move.cpp
jakobandersen: Fixed.

rwgk:
Unfortunately we're currently blind to the MSVC behavior (2 x 18 builds), but everything else is working. Except for clang-tidy wanting two `noexcept`. See below.

Did you already clang-format the new code? (`clang-format -style=file -i test_class_sh_shared_ptr_copy_move.cpp`)
I'm trying to keep all smart_holder code clang-formatted, but currently that not enforced in the CI runs.

Do you feel you could add in something like this (snippet from test_shared_ptrs.cpp):
```
        .def_property_readonly(                                                 
            "holder_copy",                                                      
            [](const SharedPtrRef &s) { return s.shared; },                     
            py::return_value_policy::copy)                                      
```
And similar for `py::return_value_policy::move`.
I'm not sure what exactly makes sense. I'd probably start by emulating the existing test. What I think we need to know is the `history` of the returned value (copied or moved), and also that of `s.shared` in the snippet above. If you feel that's too involved for this PR we could merge first, although I'd feel a bit uneasy about enabling the return value policies without being sure what exactly the behavior is. If it's out there for a while it will become difficult to change, even if inconsistent with the non-smart_holder behavior, and then we may have to explain any inconsistencies forever.

```
/__w/pybind11/pybind11/tests/test_class_sh_shared_ptr_copy_move.cpp:19:5: error: move constructors should be marked noexcept [performance-noexcept-move-constructor,-warnings-as-errors]
    Foo(Foo &&other) : history(other.history + "_MvCtor") {}
    ^
                      noexcept 
/__w/pybind11/pybind11/tests/test_class_sh_shared_ptr_copy_move.cpp:24:10: error: move assignment operators should be marked noexcept [performance-noexcept-move-constructor,-warnings-as-errors]
    Foo &operator=(Foo &&other) {
         ^
                                 noexcept 
```

rwgk:
I just pushed a temporary patch to the smart_holder branch with a workaround for the Windows breakages. Could you please rebase your branch and force push? Then we should be able to get a full set of CI results.

jakobandersen:
> I just pushed a temporary patch to the smart_holder branch with a workaround for the Windows breakages. Could you please rebase your branch and force push? Then we should be able to get a full set of CI results.

Sure, done.

rwgk:
I pushed trivial automatic clang-tidy and clang-format fixes directly to your fork.
(Looks like it worked. This was the first time I tried that.)

rwgk:
The CI is looking great now. But the behavior test I'm most worried about (something like the `.def_property_readonly(..., py::return_value_policy::copy)` and `move`) is still missing.

jakobandersen:
> I pushed trivial automatic clang-tidy and clang-format fixes directly to your fork.
> (Looks like it worked. This was the first time I tried that.)
Great, thanks.

> The CI is looking great now. But the behavior test I'm most worried about (something like the `.def_property_readonly(..., py::return_value_policy::copy)` and `move`) is still missing.

I have pushed a few commits. Is it something along those lines you mean?
It tries retrieving an attribute of type ``FooShPtr`` and ``FooSmHld``, defined via ``def_readonly``, ``def_readwrite``, and ``def_property_readonly``, in policy combinations ``move``, ``copy``, and whatever the default is.

rwgk:
Wow, thanks! Those macros are interesting! (apparently clang-format has a rough time with them)

It looks like the Foo's copy/move/operator= are never invoked, because they never appear in the `history`. See below.

In other words: When returning `std::shared_ptr`, the copy and move return value policies have no effect at all, compared to the default policy (I think it's automatic).

Probably not what I would have done, but if that's established behavior, and we have a test for it now to ensure the behavior is identical for `std::shared_ptr` as holder and `smart_holder`, I agree it's good to get this in.

No need for more tests IMO (e.g. move-only Foo, copy-only Foo). It would be nice to parameterize `test_properties`, which could probably be done with `itertools.product`, but maybe not worth the trouble (it only adds value if some but not all tests are failing).

I see the CI is green! I'll merge this now. If we want to do anything else we can just create new PRs.

```
$ grep -e _CpCtor -e _MvCtor -e _OpEqLv -e _OpEqRv tests/test_class_sh_shared_ptr_copy_move.*
tests/test_class_sh_shared_ptr_copy_move.cpp:    Foo(const Foo &other) : history(other.history + "_CpCtor") {}
tests/test_class_sh_shared_ptr_copy_move.cpp:    Foo(Foo &&other) noexcept : history(other.history + "_MvCtor") {}
tests/test_class_sh_shared_ptr_copy_move.cpp:        history = other.history + "_OpEqLv";
tests/test_class_sh_shared_ptr_copy_move.cpp:        history = other.history + "_OpEqRv";
```

rwgk:
FYI: The Rebase merge was an accident (my intention was to Squash). I didn't realize the GUI keeps the previous choice. But it's OK, I'll pay more attention next time.

jakobandersen:
Thanks!

I tried a bit to make clang format not mangle it too much, but perhaps disabling it around that code may be better.

The semantics of the attributes/properties with the default policy also matches what happens if the code is directly transcribed to Boost.Python.
I guess a similar property test with unique_ptr could be interesting as well.

rwgk:
> Thanks!
> 
> I tried a bit to make clang format not mangle it too much, but perhaps disabling it around that code may be better.

Yeah, I was thinking the same.

> The semantics of the attributes/properties with the default policy also matches what happens if the code is directly transcribed to Boost.Python.

Interesting to know. One more reason to go with the flow.

> I guess a similar property test with unique_ptr could be interesting as well.

Again same thought, but I didn't want to bother you too much with requests :-)

jakobandersen:
> > The semantics of the attributes/properties with the default policy also matches what happens if the code is directly transcribed to Boost.Python.
> 
> Interesting to know. One more reason to go with the flow.

Indeed, there shouldn't be nasty surprises. Though, they do have a type of return value policy as well, but I'm not sure if they semantically map cleanly. But it may be worth checking out if there are strange edge cases. See e.g., https://wiki.python.org/moin/boost.python/CallPolicy

> > I guess a similar property test with unique_ptr could be interesting as well.
> 
> Again same thought, but I didn't want to bother you too much with requests :-)

No problem. I might make a PR for it at some point, now that I know the technical setup.

Btw. I couldn't immediately find it, but what is the status of the smart holder branch with respect to a merge into master?

rwgk:
> Btw. I couldn't immediately find it, but what is the status of the smart holder branch with respect to a merge into master?

Earlier this year I proposed a "road map" that @henryiii used as a starting point for https://github.com/pybind/pybind11/wiki/Roadmap.

We're still figuring out how to get this work done.

In the meantime I'm striving to merge everything from master asap, keeping the smart_holder behavior in line with master as much as possible.

PR #3041, opened 2021-06-17
`py::smart_holder` `std::shared_ptr` deleter simplification & optimization.

rwgk:
Replacing `guarded_builtin_delete` and `guarded_custom_deleter` with a simpler `guarded_delete`.

The primary motivation for this change was to enable straightforward use of `std::get_deleter`, but this happened to lead to a general simplification & optimization.

`guarded_delete` only manages a simple function pointer and a `bool`. `shared_ptr<bool>` is no longer needed and removed from `py::smart_holder`.

This change was developed and copied from PR #3023. It sets the stage for changing `py::smart_holder` to play nicely with `enable_shared_from_this`.

For completeness, calling out two collateral tweaks that are not strictly needed for the deleter change:

```
-    smart_holder &operator=(smart_holder &&) = default;
+    smart_holder &operator=(smart_holder &&) = delete;
```

```
     static smart_holder from_raw_ptr_unowned(T *raw_ptr) {                     
-        hld.vptr.reset(raw_ptr, guarded_builtin_delete<T>(hld.vptr_deleter_armed_flag_ptr));
+        hld.vptr.reset(raw_ptr, [](void *) {});                                
```

### DIFF
--- a/include/pybind11/detail/smart_holder_poc.h
+++ b/include/pybind11/detail/smart_holder_poc.h
@@ -56,36 +56,43 @@ Details:
 namespace pybindit {
 namespace memory {
 
-template <typename T>
-struct guarded_builtin_delete {
-    std::shared_ptr<bool> flag_ptr;
-    explicit guarded_builtin_delete(std::shared_ptr<bool> armed_flag_ptr)
-        : flag_ptr{armed_flag_ptr} {}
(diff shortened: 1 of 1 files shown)


rwgk:
This change will get reviewed Google-internally.

PR #3042, opened 2021-06-17
Avoid string copy if possible when passing a Python object to std::ostream

jonathan-conder-sm:
## Description

Changes `operator<<(std::ostream &, const handle &)` to delegate to `operator<<(std::ostream &, std::string_view)` if possible. If I understand the type caster for `string_view` correctly, I think this avoids an unnecessary copy.

The `PYBIND11_HAS_STRING_VIEW` macro is defined in `cast.h`, which is indirectly included by `stl.h`. If that's likely to change in future, it might be desirable to move the macro to `internals.h` or something.

## Suggested changelog entry:

```rst
Use ``std::string_view`` if available to avoid a copy when passing an object to a ``std::ostream``.
```


rwgk:
Thanks!

PR #3044, opened 2021-06-17
Docs: Fix typos

Skylion007:
## Description
* Fix various typos in the documentation

## Suggested changelog entry:
* Fix typos in the docs


rwgk:
The two failing checks are 1. known Centos 8 failure and 2. a flake I've seen many times before.
The only thing really that counts is that Format succeeds. I'll merge this now. Thanks for the fixes!

PR #3046, opened 2021-06-17
fix(clang-tidy): Apply performance fixes from clang-tidy

Skylion007:
## Description
* Spoke with @rwgk about applying performance fixes from clang-tidy which he was supportive of. While there are some blockers preventing this from being enabled automatically, (mainly the lack of a hierarchical config to prevent the test files from being edited), I went ahead and applied various fixes from the performance checks. I also formatted all diffs with clang-format-diff.py .
* Most of changes can be summarized as (unnecessary std::move)
* Copies that should be std::move instead
* Copying in values that should be const references.
* Forgetting to add noexcept to move constructors.
* Let me know if any of the changes should be spun off or such because they are ABI breaking or can be applied immediately. 

## Suggested changelog entry:
```rst
* Various performance micro-optimizations applied to the codebase using clang-tidy.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -164,7 +164,7 @@ struct npy_api {
             NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
     };
 
-    using PyArray_Dims = struct {
+    struct PyArray_Dims {
         Py_intptr_t *ptr;
         int len;
     };
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -164,10 +164,10 @@ struct npy_api {
             NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
     };
 
-    typedef struct {
+    using PyArray_Dims = struct {
         Py_intptr_t *ptr;
         int len;
-    } PyArray_Dims;
+    };
 
     static npy_api& get() {
         static npy_api api = lookup();
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1412,15 +1412,15 @@ public:
 
     template <typename D, typename... Extra>
     class_ &def_readwrite_static(const char *name, D *pm, const Extra& ...extra) {
-        cpp_function fget([pm](object) -> const D &{ return *pm; }, scope(*this)),
-                     fset([pm](object, const D &value) { *pm = value; }, scope(*this));
+        cpp_function fget([pm](const object &) -> const D & { return *pm; }, scope(*this)),
+            fset([pm](const object &, const D &value) { *pm = value; }, scope(*this));
         def_property_static(name, fget, fset, return_value_policy::reference, extra...);
         return *this;
     }
 
     template <typename D, typename... Extra>
     class_ &def_readonly_static(const char *name, const D *pm, const Extra& ...extra) {
-        cpp_function fget([pm](object) -> const D &{ return *pm; }, scope(*this));
+        cpp_function fget([pm](const object &) -> const D & { return *pm; }, scope(*this));
         def_property_readonly_static(name, fget, return_value_policy::reference, extra...);
         return *this;
     }
@@ -2054,7 +2054,7 @@ exception<CppException> &register_exception(handle scope,
 }
 
 PYBIND11_NAMESPACE_BEGIN(detail)
-PYBIND11_NOINLINE inline void print(tuple args, dict kwargs) {
+PYBIND11_NOINLINE inline void print(const tuple &args, const dict &kwargs) {
     auto strings = tuple(args.size());
     for (size_t i = 0; i < args.size(); ++i) {
         strings[i] = str(args[i]);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1671,30 +1671,36 @@ struct enum_base {
             }, name("__members__")), none(), none(), ""
         );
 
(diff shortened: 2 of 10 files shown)


rwgk:
This is **awesome**!

How much trouble is it for you to apply the exact same process to the smart_holder branch?

Background:
* The smart_holder branch is what we use Google-internally.
* I'm working to keep it in sync with master as much as possible, but there are some non-trivial diffs.

Hope:
* If you apply the exact same process to the smart_holder branch we won't have merge conflicts.

But:
* I'm not sure that's actually better or easier than dealing with the merge conflicts.
* In any event, I'd really want the same cleanup for the code that's only on the smart_holder branch.

Also:
* I can run the Google-internal global testing (includes exercising a ton of open source code) only against the smart_holder branch.

Which means if we do master-merge followed by smart_holder-merge-from-master followed by testing, that testing may uncover some things we need to fix on master, so we have to go back and forth through the branches. If it's reasonably automatic to apply your process to both master and smart_holder straightaway, we'd have a more holistic picture.

Skylion007:
> This is **awesome**!
> 
> How much trouble is it for you to apply the exact same process to the smart_holder branch?
> 
> Background:
> 
>     * The smart_holder branch is what we use Google-internally.
> 
>     * I'm working to keep it in sync with master as much as possible, but there are some non-trivial diffs.
> 
> 
> Hope:
> 
>     * If you apply the exact same process to the smart_holder branch we won't have merge conflicts.
> 
> 
> But:
> 
>     * I'm not sure that's actually better or easier than dealing with the merge conflicts.
> 
>     * In any event, I'd really want the same cleanup for the code that's only on the smart_holder branch.
> 
> 
> Also:
> 
>     * I can run the Google-internal global testing (includes exercising a ton of open source code) only against the smart_holder branch.
> 
> 
> Which means if we do master-merge followed by smart_holder-merge-from-master followed by testing, that testing may uncover some things we need to fix on master, so we have to go back and forth through the branches. If it's reasonably automatic to apply your process to both master and smart_holder straightaway, we'd have a more holistic picture.

Not that much trouble, I'll go ahead and submit a PR there as well.

Skylion007:
@rwgk Results on SH branch: https://github.com/Skylion007/pybind11/tree/clang-tidy-sh-perf

rwgk:
Looks like your branch is a day or two behind. Is that easy to fix?
I see clang-tidy fixes in code that I changed yesterday and today.
Could you please turn it into a PR against the smart_holder branch? Then I
could kick off the global testing using my standard workflow.

Skylion007:
Yeah, I am making a new branch with the latest fixes.

Skylion007:
@rwgk Here you go: https://github.com/Skylion007/pybind11/tree/clang-tidy-sh-perf-latest

rwgk:
Awesome, that's totally up-to-date. I'm still learning the github ropes. I
don't see a button to create a PR from the branch (like I do in my own
fork). What's the next step for me?

Skylion007:
Opened a PR assuming you want to make it in the main smart-holder branch. If you want to pull it somewhere else here is the dialogue for doing so: https://github.com/pybind/pybind11/compare/master...Skylion007:clang-tidy-sh-perf-latest?expand=1

rwgk:
I see it now: https://github.com/pybind/pybind11/pull/3048
That's what I was hoping for, for some reason it wasn't in the "pulls" list
a few minutes ago. Thanks!!

rwgk (approved):
Via #3048 I ran these changes with ASAN, MSAN, and the Google-internal global testing. I'd love to see this merged.

include/pybind11/numpy.h
```diff
@@ -164,10 +164,10 @@ struct npy_api {
             NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
     };
 
-    typedef struct {
+    using PyArray_Dims = struct {
```
wjakob, line 167: Why not just `struct PyArray_Dims`?

wjakob (approved):
That's actually pretty neat. They all look fine to me, except for one minor change I would have made.

include/pybind11/numpy.h
Skylion007: Was here when I got there. Changed.

include/pybind11/pybind11.h
```diff
@@ -1671,30 +1671,36 @@ struct enum_base {
...
-                name(op), is_method(m_base), arg("other"))
-
-        #define PYBIND11_ENUM_OP_CONV_LHS(op, expr)                                    \
-            m_base.attr(op) = cpp_function(                                            \
-                [](object a_, object b) {                                              \
-                    int_ a(a_);                                                        \
-                    return expr;                                                       \
-                },                                                                     \
-                name(op), is_method(m_base), arg("other"))
+#define PYBIND11_ENUM_OP_STRICT(op, expr, strict_behavior)                                        \
```
Skylion007, line 1674: clang-format required this change.

include/pybind11/numpy.h
```diff
@@ -164,7 +164,7 @@ struct npy_api {
             NPY_ULONG_, NPY_ULONGLONG_, NPY_UINT_),
     };
 
-    using PyArray_Dims = struct {
```
rwgk: I'm just curious, when I saw this initially I thought it was something automatically applied, thinking "probably for a good reason". Was that not the case?

Skylion007: It should have been automatically caught by the github actions tbh, but it wasn't. From modernize-use-using check

rwgk:
Thanks Aaron!

PR #3049, opened 2021-06-19
fix(clang-tidy) Add more modernize clang-tidy checks

Skylion007:
## Description

* This adds a bunch more reasonable modernization checks. Basically prevents contributors from using deprecated pre-C++11 features like std::bind. I doubt most of these changes would make it past code review, but it's good to have automated checks to detect them. These checks should find no issues with the current codebase as expected.

I also alphabetized the list and treated all warning as errors by default.

## Suggested changelog entry:
```rst
* Enable more clang-tidy modernization checks
```


Skylion007:
Ping @rwgk

tests/test_factory_constructors.cpp
```diff
@@ -20,12 +20,12 @@ class TestFactory1 {
     TestFactory1() : value("(empty)") { print_default_created(this); }
     TestFactory1(int v) : value(std::to_string(v)) { print_created(this, value); }
     TestFactory1(std::string v) : value(std::move(v)) { print_created(this, value); }
+public:
+    std::string value;
     TestFactory1(TestFactory1 &&) = delete;
```
Skylion007, line 25: It complained that deleted functions need to be public per the spec so I fixed it.

Skylion007: Specifically:
```
pybind11/tests/test_factory_constructors.cpp:23:5: error: deleted member function should be public [modernize-use-equals-delete,-warnings-as-errors]
    TestFactory1(TestFactory1 &&) = delete;
    ^
pybind11/tests/test_factory_constructors.cpp:24:5: error: deleted member function should be public [modernize-use-equals-delete,-warnings-as-errors]
    TestFactory1(const TestFactory1 &) = delete;
    ^
pybind11/tests/test_factory_constructors.cpp:25:19: error: deleted member function should be public [modernize-use-equals-delete,-warnings-as-errors]
    TestFactory1 &operator=(TestFactory1 &&) = delete;
pybind11/tests/test_factory_constructors.cpp:26:19: error: deleted member function should be public [modernize-use-equals-delete,-warnings-as-errors]
    TestFactory1 &operator=(const TestFactory1 &) = delete;
```

rwgk (approved):
Looks great, thanks! Just giving @henryiii  a chance to take a look, too.

henryiii (approved):
Looks great to me! More checks are always better if we can add them.

PR #3051, opened 2021-06-20
fix(clang-tidy): Apply all performance fixes to tests and enable performance checks on CI

Skylion007:
## Description

@rwgk  @henryiii I went ahead and tried to fix or ignore as many of the performance issues in tests as possible. It's a big PR, but by fixing all the issues in tests or ignoring them explicitly, this should allow us to automatically run the performance checks on CI. Feel free to directly edit this PR if I broke any tests.
I also caught a couple of issues in the core library that weren't caught by the automated fixes earlier.

Types of changes
I was very conservative with what changes I made manually since I didn't want to break any testing behavior.
The changes made are mostly as follows:
1. Add or removing missing std::move
1. Adding noexcept
1. Change certain variables to const ref. It's a very noisy error in tests, but it was also by far the most common performance error throughout the codebase, so I left this check enabled. I tried to fix it in several places, but others I left it as NOLINTNEXTLINE to try to avoid changing behavior of the test. Occasionally, changing it to a const ref broke things anyway due to typing errors. Many of these NOLINTNEXTLINES should probably be migrated over to proper const ref fixes if safe, but I'll let @rwgk, @henryiii or someone else more familiar with the tests down the line fix them later. It's not very high pri since testing isn't that performance sensitive. I tried to change it in examples that people might copy from test to try to demonstrate best coding practices. In other words, NOLINTNEXTLINE is the default for any test that caused issues when changing it to a const ref or I was uncertain whether it would change testing behavior. I also added it in args that are likely to be elided like static methods in tests to avoid having to change all those by hand. Feel free to get rid of as many of these as possible @rwgk .
Some of the unparameterized nolintnextlines was because changing the class to be a multiline class or adding std::move in the constructor initializer list would break the pygrep pre-commit hook for proper upper/lowercasing. 

@rwgk Plenty of tricky decisions, the most common one is not fixing the Cat  and Dog classes in tests because it would cause a formatting change that would break the pygrep pre-commit hook. The other tricky decisions was mostly whether to actually fix the pass by value or leave it as a NOLINT. I erred on the side of caution with NOLINT when possible to avoid breaking tests.
## Suggested changelog entry:

```rst
* Enable clang-tidy performance checks throughout the codebase to enforce performant coding practices.
```

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Configure
       run: >
         cmake -S . -B build
-        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);"
+        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy)"
         -DDOWNLOAD_EIGEN=ON
         -DDOWNLOAD_CATCH=ON
         -DCMAKE_CXX_STANDARD=17
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -27,7 +27,7 @@ jobs:
   clang-tidy:
     name: Clang-Tidy
     runs-on: ubuntu-latest
-    container: silkeh/clang:10
+    container: silkeh/clang:12
     steps:
     - uses: actions/checkout@v2
 
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -43,4 +43,4 @@ jobs:
         -DCMAKE_CXX_STANDARD=17
 
     - name: Build
-      run: cmake --build build -j 2
+      run: cmake --build build -j 2 -- --keep-going
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Configure
       run: >
         cmake -S . -B build
-        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--warnings-as-errors=*"
+        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);"
         -DDOWNLOAD_EIGEN=ON
         -DDOWNLOAD_CATCH=ON
         -DCMAKE_CXX_STANDARD=17
--- a/tests/pybind11_cross_module_tests.cpp
+++ b/tests/pybind11_cross_module_tests.cpp
@@ -104,7 +104,7 @@ PYBIND11_MODULE(pybind11_cross_module_tests, m) {
     m.def("return_self", [](LocalVec *v) { return v; });
     m.def("return_copy", [](const LocalVec &v) { return LocalVec(v); });
 
-    // Changint his broke things with pygrep. TODO fix
+    // Changing this broke things with pygrep. TODO fix
     // NOLINTNEXTLINE
     class Dog : public pets::Pet { public: Dog(std::string name) : Pet(name) {}; };
     py::class_<pets::Pet>(m, "Pet", py::module_local())
--- a/tests/pybind11_cross_module_tests.cpp
+++ b/tests/pybind11_cross_module_tests.cpp
@@ -104,6 +104,7 @@ PYBIND11_MODULE(pybind11_cross_module_tests, m) {
     m.def("return_self", [](LocalVec *v) { return v; });
     m.def("return_copy", [](const LocalVec &v) { return LocalVec(v); });
 
+    // Changint his broke things with pygrep. TODO fix
     // NOLINTNEXTLINE
     class Dog : public pets::Pet { public: Dog(std::string name) : Pet(name) {}; };
     py::class_<pets::Pet>(m, "Pet", py::module_local())
--- a/tests/pybind11_cross_module_tests.cpp
+++ b/tests/pybind11_cross_module_tests.cpp
@@ -104,6 +104,7 @@ PYBIND11_MODULE(pybind11_cross_module_tests, m) {
     m.def("return_self", [](LocalVec *v) { return v; });
     m.def("return_copy", [](const LocalVec &v) { return LocalVec(v); });
 
+    // NOLINTNEXTLINE
     class Dog : public pets::Pet { public: Dog(std::string name) : Pet(name) {}; };
     py::class_<pets::Pet>(m, "Pet", py::module_local())
         .def("name", &pets::Pet::name);
@@ -126,6 +127,7 @@ PYBIND11_MODULE(pybind11_cross_module_tests, m) {
     // test_missing_header_message
     // The main module already includes stl.h, but we need to test the error message
     // which appears when this header is missing.
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("missing_header_arg", [](std::vector<float>) { });
     m.def("missing_header_return", []() { return std::vector<float>(); });
 }
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -30,7 +30,11 @@ class type_caster<ConstRefCasted> {
   // cast operator.
   bool load(handle, bool) { return true; }
 
-  operator ConstRefCasted&&() { value = {1}; return std::move(value); }
+  operator ConstRefCasted &&() {
+      value = {1};
+      // NOLINTNEXTLINE(performance-move-const-arg)
+      return std::move(value);
+  }
   operator ConstRefCasted&() { value = {2}; return value; }
   operator ConstRefCasted*() { value = {3}; return &value; }
 
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -105,6 +105,7 @@ TEST_SUBMODULE(builtin_casters, m) {
 
     // test_bytes_to_string
     m.def("strlen", [](char *s) { return strlen(s); });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("string_length", [](std::string s) { return s.length(); });
 
 #ifdef PYBIND11_HAS_U8STRING
@@ -150,6 +151,7 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("int_passthrough_noconvert", [](int arg) { return arg; }, py::arg{}.noconvert());
 
     // test_tuple
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("pair_passthrough", [](std::pair<bool, std::string> input) {
         return std::make_pair(input.second, input.first);
     }, "Return a pair in reversed order");
@@ -181,10 +183,13 @@ TEST_SUBMODULE(builtin_casters, m) {
 
     // test_none_deferred
     m.def("defer_none_cstring", [](char *) { return false; });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("defer_none_cstring", [](py::none) { return true; });
     m.def("defer_none_custom", [](UserType *) { return false; });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("defer_none_custom", [](py::none) { return true; });
     m.def("nodefer_none_void", [](void *) { return true; });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("nodefer_none_void", [](py::none) { return false; });
 
     // test_void_caster
@@ -235,6 +240,7 @@ TEST_SUBMODULE(builtin_casters, m) {
     }, "copy"_a);
 
     m.def("refwrap_iiw", [](const IncType &w) { return w.value(); });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("refwrap_call_iiw", [](IncType &w, py::function f) {
         py::list l;
         l.append(f(std::ref(w)));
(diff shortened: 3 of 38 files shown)


rwgk (approved):
That's fantastic! I glanced through the changes but didn't go through with a comb, thinking that's not needed, given that this PR allows us to add automatic tooling for code health, and that we already have extensive testing. As far as I saw there are mostly or only (?) three types of changes:

1. `foo` -> `const foo&`
2. adding `noexcept`
3. `NOLINTNEXTLINE`

Did I miss something important? What do you think about adding that list + anything I missed to the PR description? If you could explain (outline is fine) the rationale for deciding between adding `const &` vs `NOLINTNEXTLINE` that would be a plus. If there are any special or tricky decisions you made, pointing those out would be good, too. I will take a closer look at those.

tests/pybind11_cross_module_tests.cpp
```diff
@@ -104,6 +104,7 @@ PYBIND11_MODULE(pybind11_cross_module_tests, m) {
     m.def("return_self", [](LocalVec *v) { return v; });
     m.def("return_copy", [](const LocalVec &v) { return LocalVec(v); });
 
+    // NOLINTNEXTLINE
     class Dog : public pets::Pet { public: Dog(std::string name) : Pet(name) {}; };
```
Skylion007, line 108: @rwgk One of the problematic lines of code. FIxing this with `: Pet(std::move(name))` would break things.

tests/test_builtin_casters.cpp
```diff
@@ -177,10 +183,13 @@ TEST_SUBMODULE(builtin_casters, m) {
 
     // test_none_deferred
     m.def("defer_none_cstring", [](char *) { return false; });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("defer_none_cstring", [](py::none) { return true; });
     m.def("defer_none_custom", [](UserType *) { return false; });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("defer_none_custom", [](py::none) { return true; });
```
Skylion007, line 190: It really surprises me py::none is not trivially copyable. Isn't it a singleton? Should we add it to the list of allowed values to pass by value in the .clang-tidy config?

tests/test_builtin_casters.cpp
```diff
@@ -231,6 +240,7 @@ TEST_SUBMODULE(builtin_casters, m) {
     }, "copy"_a);
 
     m.def("refwrap_iiw", [](const IncType &w) { return w.value(); });
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("refwrap_call_iiw", [](IncType &w, py::function f) {
```
Skylion007, line 244: Function here could have been a reference, but I wasn't sure it would break things.

tests/test_copy_move.cpp
```diff
@@ -49,9 +56,16 @@ class MoveOnlyInt {
 class MoveOrCopyInt {
 public:
     MoveOrCopyInt() { print_default_created(this); }
-    MoveOrCopyInt(int v) : value{std::move(v)} { print_created(this, value); }
-    MoveOrCopyInt(MoveOrCopyInt &&m) { print_move_created(this, m.value); std::swap(value, m.value); }
-    MoveOrCopyInt &operator=(MoveOrCopyInt &&m) { print_move_assigned(this, m.value); std::swap(value, m.value); return *this; }
+    MoveOrCopyInt(int v) : value{v} { print_created(this, value); }
```
Skylion007, line 59: Move is unnecessary. Feel free to add back if intentional.

tests/test_eigen.cpp
```diff
@@ -94,14 +93,16 @@ TEST_SUBMODULE(eigen, m) {
...
+    m.def("double_mat_cm", [](const Eigen::MatrixXf &x) -> Eigen::MatrixXf { return 2.0f * x; });
+    m.def("double_mat_rm", [](const DenseMatrixR &x) -> DenseMatrixR { return 2.0f * x; });
 
     // test_eigen_ref_to_python
     // Different ways of passing via Eigen::Ref; the first and second are the Eigen-recommended
+    // NOLINTNEXTLINE (performance-unnecessary-value-param)
     m.def("cholesky1", [](Eigen::Ref<MatrixXdR> x) -> Eigen::MatrixXd { return x.llt().matrixL(); });
     m.def("cholesky2", [](const Eigen::Ref<const MatrixXdR> &x) -> Eigen::MatrixXd { return x.llt().matrixL(); });
     m.def("cholesky3", [](const Eigen::Ref<MatrixXdR> &x) -> Eigen::MatrixXd { return x.llt().matrixL(); });
+    // NOLINTNEXTLINE (performance-unnecessary-value-param)
```
Skylion007, line 105: Wasn;t sure if Eigen::Ref should just be allowed to pass by values explicitly too.

tests/test_exceptions.cpp
```diff
@@ -199,6 +201,7 @@ TEST_SUBMODULE(exceptions, m) {
         throw py::error_already_set();
     });
 
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
     m.def("python_call_in_destructor", [](py::dict d) {
```
Skylion007, line 205: Changes this one broke things.

tests/test_kwargs_and_defaults.cpp
```diff
@@ -37,17 +39,21 @@ TEST_SUBMODULE(kwargs_and_defaults, m) {
     m.def("args_function", [](py::args args) -> py::tuple {
```
Skylion007, line 39: This can probably be safely changed, I just didn't know if you were testing fo rpass by value explicitly.

tests/test_local_bindings.cpp
```diff
@@ -86,6 +86,7 @@ TEST_SUBMODULE(local_bindings, m) {
     m.def("return_self", [](LocalVec *v) { return v; });
     m.def("return_copy", [](const LocalVec &v) { return LocalVec(v); });
 
+    // NOLINTNEXTLINE
     class Cat : public pets::Pet { public: Cat(std::string name) : Pet(name) {}; };
```
Skylion007, line 90: Same issue as before with std::move breaking the pygrep

tests/test_numpy_array.cpp
```diff
@@ -245,39 +248,59 @@ TEST_SUBMODULE(numpy_array, sm) {
     });
 
     // test_overload_resolution
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
```
Skylion007, line 251: These all are nameless but could all be const ref. I didn't know if that would break the overloads though.

tests/test_numpy_vectorize.cpp
```diff
@@ -53,11 +57,15 @@ TEST_SUBMODULE(numpy_vectorize, m) {
...
-    m.def("vec_passthrough", py::vectorize(
-        [](double *a, double b, py::array_t<double> c, const int &d, int &e, NonPODClass f, const double g) {
-            return *a + b + c.at(0) + d + e + f.value + g;
-        }
-    ));
+    m.def("vec_passthrough",
+          py::vectorize([](double *a,
+                           double b,
+                           // NOLINTNEXTLINE(performance-unnecessary-value-param)
+                           py::array_t<double> c,
```
Skylion007, line 64: Changing this to a const ref broke things

Skylion007:
@rwgk Feel free to edit the PR directly. I pointed out some of the tricky decisions I made. Mostly erred on the side of caution and just NOLINTed things instead of changing behavior.

.github/workflows/format.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     - name: Configure
       run: >
         cmake -S . -B build
-        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--warnings-as-errors=*"
+        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy)"
```
Skylion007, line 40: We already treat warning as errors in the config so this arg is unnecessary here.

Skylion007:
@rwgk I also wanted to enable readability-const-return-type , but it generated a bunch of errors I am not sure how to fix such as: 
```
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<1UL / 10, 1UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<2UL / 10, 2UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<32UL / 10, 32UL % 10>::digits)' (aka 'const pybind11::detail::descr<2>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<3UL / 10, 3UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<4UL / 10, 4UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<5UL / 10, 5UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<64UL / 10, 64UL % 10>::digits)' (aka 'const pybind11::detail::descr<2>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/descr.h:77:24: error: return type 'decltype(int_to_str<6UL / 10, 6UL % 10>::digits)' (aka 'const pybind11::detail::descr<1>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
/Users/agokaslan/git/my_forks/pybind11/tests/test_eigen.cpp:180:9: error: return type 'const Eigen::MatrixXd' (aka 'const Matrix<double, Dynamic, Dynamic>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
        static const Eigen::MatrixXd createConst() { return Eigen::MatrixXd::Ones(10, 10); }
        
../include/pybind11/detail/../pytypes.h:642:5: error: return type 'pybind11::detail::generic_iterator<pybind11::detail::iterator_policies::dict_readonly>::reference' (aka 'const pair<pybind11::handle, pybind11::handle>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
    reference operator*() const { return Policy::dereference(); }
    ^
../include/pybind11/detail/../pytypes.h:642:5: error: return type 'pybind11::detail::generic_iterator<pybind11::detail::iterator_policies::sequence_fast_readonly>::reference' (aka 'const pybind11::handle') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
../include/pybind11/detail/../pytypes.h:686:5: error: return type 'pybind11::detail::iterator_policies::sequence_fast_readonly::reference' (aka 'const pybind11::handle') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
    reference dereference() const { return *ptr; }
    ^
../include/pybind11/detail/../pytypes.h:730:5: error: return type 'pybind11::detail::iterator_policies::dict_readonly::reference' (aka 'const pair<pybind11::handle, pybind11::handle>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
    reference dereference() const { return {key, value}; }
    ^
../include/pybind11/detail/../pytypes.h:882:5: error: return type 'pybind11::iterator::reference' (aka 'const pybind11::handle') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
    reference operator*() const {
```

rwgk:
Hi Aaron, thanks for all the comments! I don't have the free bandwidth to get deeply into the weeds. My suggestion is to turn those github comments into terse code comments to go with the `NOLINTNEXTLINE`. Then over time we can weed them out, e.g. when someone is working on a related change and has built up the required context in their head already. And the cases that hang around probably aren't that important! What seems much more important to me is to get the automatic runs up and running.

rwgk:
> @rwgk I also wanted to enable readability-const-return-type , but it generated a bunch of errors I am not sure how to fix

No immediate idea. Sorry I don't know much about this code, too. My suggestion is to get this PR merged, then open a separate one for this issue while we have the rest in place already.

henryiii (approved):

PR #3052, opened 2021-06-20
Enable and apply clang-tidy readability and misc fixes.

Skylion007:
## Description
* Broke off this PR from the beefer performance once. This just applies some automated fixes to the code base. The main one is a redundant function ptr dereference and forgetting to mark a bunch const methods as such. Enables a bunch of other checks that ensure modern code.

There is one readability improvement with const return types I would like to enable but it causes error sadly that need to be fixed.  
## Suggested changelog entry:

```rst
* Enable and apply checks for more misc, readability, and code modernization improvements to the codebase. Remove rendundant function ptr dereference, makes methods const, and uses member-initializers when possible to allow for trivial construction.
```

### DIFF
--- a/tests/test_constants_and_functions.cpp
+++ b/tests/test_constants_and_functions.cpp
@@ -61,7 +61,8 @@ int f3(int x) noexcept(false) { return x+3; }
 #  pragma GCC diagnostic push
 #  pragma GCC diagnostic ignored "-Wdeprecated"
 #endif
-int f4(int x) noexcept { return x+4; } // Deprecated equivalent to noexcept(true)
+// NOLINTNEXTLINE(modernize-use-noexcept)
+int f4(int x) throw() { return x+4; } // Deprecated equivalent to noexcept(true)
 #if defined(__GNUG__) && !defined(__INTEL_COMPILER)
 #  pragma GCC diagnostic pop
 #endif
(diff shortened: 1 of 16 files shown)


tests/test_constants_and_functions.cpp
```diff
@@ -75,8 +77,10 @@ struct C {
 #  pragma GCC diagnostic push
 #  pragma GCC diagnostic ignored "-Wdeprecated"
 #endif
-    int m7(int x) throw() { return x-7; }
-    int m8(int x) const throw() { return x-8; }
+    // NOLINTNEXTLINE(modernize-use-noexcept)
```
Skylion007, line 80: This check discourages the user of throw, but since these are explicilty deprecated tests, I add nolints to prevent them from being changed.

Skylion007:
Ping @rwgk @henryiii

tests/test_constants_and_functions.cpp
```diff
@@ -1,5 +1,6 @@
 /*
-    tests/test_constants_and_functions.cpp -- global constants and functions, enumerations, raw byte strings
+    tests/test_constants_and_functions.cpp -- global constants and functions, enumerations, raw
+   byte strings
```
rwgk, line 3: another space for indentation?

rwgk (approved):
Thanks Aaron!

tests/test_constants_and_functions.cpp
Skylion007:
``` suggestion
    byte strings
```

henryiii (approved):
This all looks very good to me! Thanks! Love seeing that someone is finally filling out the clang-tidy checks, see my comment at  https://github.com/pybind/pybind11/pull/2478#issue-484607248 :)

henryiii:
I'm calling this a fix due to the fact that adding const technically is better, as you can only call const methods on a const object.

PR #3055, opened 2021-06-22
Updating and slightly enhancing instructions for running clang-tidy.

rwgk:
Follow-on to PR #3051 which updated the docker container version in .github/workflows/format.yml but not in the docs.

Minor clarification from where to run the docker command (for docker muggles like myself).

Adding `--yes` for convenience and `--keep-going` for engineering productivity.

Tested interactively.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -209,9 +209,9 @@ so you can use git to monitor the changes.
 
 ```bash
 docker run --rm -v $PWD:/mounted_pybind11 -it silkeh/clang:12
-apt-get update --yes && apt-get install --yes python3-dev python3-pytest
-cmake -S /mounted_pybind11/ -B build -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);-fix"
-cmake --build build -- --keep-going -j 2
+apt-get update && apt-get install -y python3-dev python3-pytest
+cmake -S /mounted_pybind11/ -B build -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);-fix" -DDOWNLOAD_EIGEN=ON -DDOWNLOAD_CATCH=ON -DCMAKE_CXX_STANDARD=17
+cmake --build build -j 2 -- --keep-going
 ```
 
 ### Include what you use
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -203,14 +203,15 @@ of the pybind11 repo.
 [`clang-tidy`][clang-tidy] performs deeper static code analyses and is
 more complex to run, compared to `clang-format`, but support for `clang-tidy`
 is built into the pybind11 CMake configuration. To run `clang-tidy`, the
-following recipe should work. Files will be modified in place, so you can
-use git to monitor the changes.
+following recipe should work. Run the `docker` command from the top-level
+directory inside your pybind11 git clone. Files will be modified in place,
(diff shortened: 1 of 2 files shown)


rwgk:
@Skylion007

.github/CONTRIBUTING.md
```diff
@@ -203,14 +203,15 @@ of the pybind11 repo.
...
+following recipe should work. Run the `docker` command from the top-level
+directory inside your pybind11 git clone. Files will be modified in place,
+so you can use git to monitor the changes.
 
 ```bash
-docker run --rm -v $PWD:/pybind11 -it silkeh/clang:10
-apt-get update && apt-get install python3-dev python3-pytest
-cmake -S pybind11/ -B build -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);-fix"
-cmake --build build
+docker run --rm -v $PWD:/mounted_pybind11 -it silkeh/clang:12
```
Skylion007, line 211: This works although docker is a little overkill for clang-tidy IMO. 👍

Skylion007 (approved):

.github/CONTRIBUTING.md
henryiii: Getting a consistent LLVM install on all platforms is not exactly easy. :)

.github/CONTRIBUTING.md
```diff
@@ -203,14 +203,15 @@ of the pybind11 repo.
...
+apt-get update --yes && apt-get install --yes python3-dev python3-pytest
+cmake -S /mounted_pybind11/ -B build -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);-fix"
+cmake --build build -- --keep-going -j 2
```
henryiii, line 214: I'm not sure why 2 is used here, we don't what the ideal number is. We could use `--parallel=2` instead if you want to make it a bit more obvious that it's the number of build threads (though maybe `-j` is so common that it's best). Also, this needs to be before the `--`, it should go through CMake, not directly to the build backend.

```suggestion
cmake --build build -j 2 -- --keep-going
```

rwgk: > I'm not sure why 2 is used here, we don't what the ideal number is. We could use `--parallel=2` instead if you want to make it a bit more obvious that it's the number of build threads (though maybe `-j` is so common that it's best). Also, this needs to be before the `--`, it should go through CMake, not directly to the build backend.

Applied (by copying the exact command from the workflow file).

In the meantime I stumbled over missing `-D...` options. Added here. I also added a comment in the workflow file, to make it less likely that the documentation is out-of-sync.

rwgk:
I forgot to mention, the `-j 2` is just to tell people "hey you can run this in parallel". That was a question I was uncertain about initially. I think people can take it from there, with the number of the CPUs.

.github/CONTRIBUTING.md
rwgk: I have a centrally managed "workstation" (in the cloud). Stuff I install manually disappears seemingly randomly. The behavior of a docker container is a bit more predictable.

henryiii (approved):

rwgk:
Thanks Aaron and Henry!
(I spent my afternoon clang-tidy fixing the smart_holder branch. 😅)

Skylion007:
@rwgk FYI: You can autofix many clang-tidy issues with the following script: https://clang.llvm.org/extra/doxygen/run-clang-tidy_8py_source.html

PR #3058, opened 2021-06-24
fix(perf): avoid unnecessary strlen

cyyever:


Skylion007 (approved):
Neat! These are on some pretty hot code paths so these changes should be helpful.

include/pybind11/attr.h
```diff
@@ -377,7 +377,7 @@ template <> struct process_attribute<is_new_style_constructor> : process_attribu
 };
 
 inline void process_kw_only_arg(const arg &a, function_record *r) {
-    if (!a.name || strlen(a.name) == 0)
+    if (!a.name || a.name[0] == '\0')
```
henryiii, line 380: Would this be cleaner this way?
```suggestion
    if (!a.name || !a.name[0])
```

Maybe the first way reads better, though. Just a thought.

Skylion007: I am personally against implicit boolean conversions for readability reasons.

henryiii: Yeah, I don't think I like it that way either. The current way clearly indicates what it's doing.

henryiii (approved):

include/pybind11/attr.h
cyyever: Both ways are equivalent. However, I also want to make the intention more explicit.

rwgk (approved):
We fixed the Windows issue a few days ago. Could you please rebase on current master?  We'll merge this when we see that the CI is green. (I'm tempted to just merge but better to not be too wild west.)

cyyever:
> We fixed the Windows issue a few days ago. Could you please rebase on current master? We'll merge this when we see that the CI is green. (I'm tempted to just merge but better to not be too wild west.)
OK

rwgk:
I think the one macos straggler will just time out again. Happens all the time recently. Safe to ignore.

PR #3059, opened 2021-06-25
fix: enable py::implicitly_convertible<py::none, ...> for py::class_-wrapped types

### LINKED ISSUE #2778: [FEAT] Cannot implicitly convert from None.
I am going to address this issue one more time. It was first brought up in #642, then again in #1382.

In short, we cannot make a custom type that implicitly converts from `None`. @bstaletic suggested [a solution](https://github.com/pybind/pybind11/issues/1382#issuecomment-664207065) that doesn't work.

Consider the following example code (paraphrased, might have typos):

```cpp
#include "pydip.h"

class A {
public:
   int value = 0;
   A() = default;
   A(int v) : value(v) {};
}

int X(A const&) {
   return A.value;
}

   auto a = py::class_<A>(m, "A");
   a(py::init<>());
   a("__init__", [](py::object const& o) { // @bstaletic's solution
      if (o.is_none()) {
         return A{};
      }
      throw py::type_error("boom");
   });
   //a(py::init([](py::none const&) { return A{}; }));  // I had tried this one too
   py::implicitly_convertible<py::none, A>();
   a(py::init<int>());
   py::implicitly_convertible<int, A>();

   m.def("X", X);
}
```

I can now do in Python:
```python
a = A(None)
a = A(1)
X(a)
X(1)
```
But I cannot do
```python
X(None)
```

However, if I edit `pybind11/cast.h`, at the top of `bool load_impl(handle src, bool convert)`, where it says:
```cpp
        if (src.is_none()) {
            // Defer accepting None to other overloads (if we aren't in convert mode):
            if (!convert) return false;
            value = nullptr;
            return true;
        }
```
If I remove the `value = nullptr; return true;` lines, then the above all works correctly.

I have not found any downsides to doing so. This means that the custom converter function gets a `py::none` object as input, it can choose to reject that or accept that. Any conversion function already tests what the type of the input object is, so this should not affect any existing code.

I have been using this "fix" for some months now, but would prefer a solution that doesn't require patching pybind11.

---

In real life, the above example is for an image class where the default-constructed image has no pixels, and is used to signal "I don't need this input" to specific functions. In C++ one can use `{}` for a default-constructed object, `None` is the most Pythonic substitution. See [this other comment](https://github.com/pybind/pybind11/issues/642#issuecomment-634961927) for details.

Sure, I could write a function for `X(None)`, but this is for a large project with a lot of functions, and creating multiple versions of each for various inputs being `None` is not doable.

crisluengo:
## Description

Enables implicit casts from `None` to a custom type, by allowing custom converters to run first, and doing the check for `py::none` at the end.
This fixes #2778.

The new test case is a nice illustration:
```cpp
py::class_<NoneCastTester>(m, "NoneCastTester")
    .def(py::init<>())
    .def(py::init<int>())
    .def(py::init([](py::none const&) { return NoneCastTester{}; }));
py::implicitly_convertible<py::none, NoneCastTester>();
m.def("ok_obj_or_none", [](NoneCastTester const& foo) { return foo.answer; });
```
This allows in Python:
```python
ok_obj_or_none(NoneCastTester())  # as was allowed previously
ok_obj_or_none(None)              # this is new
```

## Suggested changelog entry:

```rst
* Enable ``py::implicitly_convertible<py::none, ...>`` for ``py::class_``-wrapped types.
```

### DIFF
--- a/tests/test_methods_and_attributes.py
+++ b/tests/test_methods_and_attributes.py
@@ -431,6 +431,17 @@ def test_accepts_none(msg):
     assert "incompatible function arguments" in str(excinfo.value)
 
 
+def test_casts_none(msg):
+    """#2778: implicit casting from None to object (not pointer)"""
+    a = m.NoneCastTester()
+    assert m.ok_obj_or_none(a) == -1
+    a = m.NoneCastTester(4)
+    assert m.ok_obj_or_none(a) == 4
+    a = m.NoneCastTester(None)
+    assert m.ok_obj_or_none(a) == -1
+    assert m.ok_obj_or_none(None) == -1
+
+
 def test_str_issue(msg):
     """#283: __str__ called on uninitialized instance when constructor arguments invalid"""
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
       run: >
         cmake -S . -B .
         -DPYBIND11_WERROR=ON
-        -DDOWNLOAD_CATCH=ON
+        -DDOWNLOAD_CATCH=OFF
         -DDOWNLOAD_EIGEN=ON
         -DCMAKE_CXX_STANDARD=11
         ${{ matrix.args }}
@@ -111,10 +111,10 @@ jobs:
     - name: Python tests C++11
       run: cmake --build . --target pytest -j 2
 
-    - name: C++11 tests
-      # TODO: Figure out how to load the DLL on Python 3.8+
-      if: "!(runner.os == 'Windows' && (matrix.python == 3.8 || matrix.python == 3.9 || matrix.python == '3.10-dev'))"
-      run: cmake --build .  --target cpptest -j 2
+    #- name: C++11 tests
+    #  # TODO: Figure out how to load the DLL on Python 3.8+
+    #  if: "!(runner.os == 'Windows' && (matrix.python == 3.8 || matrix.python == 3.9 || matrix.python == '3.10-dev'))"
+    #  run: cmake --build .  --target cpptest -j 2
 
     - name: Interface test C++11
       run: cmake --build . --target test_cmake_build
@@ -127,7 +127,7 @@ jobs:
       run: >
         cmake -S . -B build2
         -DPYBIND11_WERROR=ON
-        -DDOWNLOAD_CATCH=ON
+        -DDOWNLOAD_CATCH=OFF
         -DDOWNLOAD_EIGEN=ON
         -DCMAKE_CXX_STANDARD=17
         ${{ matrix.args }}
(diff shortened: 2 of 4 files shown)


crisluengo:
There seems to be some issue downloading Catch on Windows. None of the failed tests actually ran.

henryiii:
Known issue for the last day or two. Guessing it's a networking issue on GHA's side?

henryiii (approved):
Seems fine to me, @EricCousineau-TRI or @rwgk would you like to take a look too? Also, @rwgk should we port your CI patch for now? Maybe we can get the file some other way, instead? (It's a GitHub to GitHub download that's failing on Windows only...)

henryiii:
@YannickJadoul You looked at this before, as well, along with @bstaletic, happy to take a once-over!

EricCousineau-TRI (changes requested):
LGTM for code!

That being said, there is an existing way to admit casting from `py::none` to a different representation (custom object) by using `py::detail::type_caster<>`.
Docs: https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html
Example (in pybind11) for `std::optional<T>` caster:
https://github.com/pybind/pybind11/blob/484b0f04332fff15543d4c2a49c355d6942455fa/include/pybind11/stl.h#L281-L283

Therefore, could you better qualify your commit / PR title to reflect that?
e.g.
"fix: allow casting from None to custom object when using type_caster_generic" (impl details)
or
"fix: allow casting from None to custom object for generic types"

rwgk:
> "fix: allow casting from None to custom object for generic types"

LGTM for code, too.

Using the word "custom" puts me on the wrong track. IIUC the feature change is for `py::class_`-wrapped types and ties in with `py::implicitly_convertible`, is that correct? Would this make sense as PR title?

Enable `py::implicitly_convertible<py::none, ...>` for `py::class_`-wrapped types.

crisluengo:
@rwgk
Indeed, it is about the implicit cast to a wrapped C++ class. I have updated the PR title and the suggested change log entry.

@EricCousineau-TRI
It is maybe possible to write a `py::detail::type_caster<>`, I honestly have not tried this avenue. I have written type casters for C++ types that I wanted to correspond to native Python types (i.e. I didn't want to wrap the C++ classes). I don't know how `py::detail::type_caster<>` interacts with a wrapped class. On the other hand, why write so much code for stuff that should be generated automatically? :)

rwgk:
To get full CI coverage, could you try to adopt my ci.yml patch from the smart_holder branch here in this PR?

One way to get the diff (git pull on master & smart_holder first to be up-to-date):
```
git diff master smart_holder -- .github/workflows/ci.yml
```
Ignore the first to `+` in the diff, apply everything else.

rwgk:
The one failing CI job is safe to ignore, GHA simply ran out of macos hosts (I saw that several times before):
```
2021-06-25T20:11:48.6474262Z Found online and idle hosted runner(s) in the current repository's enterprise account that matches the required labels: 'macos-latest'
2021-06-25T20:11:48.6474317Z Waiting for a hosted runner in 'enterprise' to pick this job...
```

rwgk (approved):
@henryiii I think merging this quickly including the ci.yml patch would be great.

EricCousineau-TRI:
> `@EricCousineau-TRI`
> It is maybe possible to write a py::detail::type_caster<>, I honestly have not tried this avenue.  [...]

Haha, my b. I was more or less trying to say the same thing as Ralf - your fix is great and stand on its own, just needed to clarify the title. Sorry about that, and thanks!

EricCousineau-TRI (approved):

rwgk:
Thanks everybody! I'll merge this now with the added benefit of restoring our CI to be fully functioning.

tests/test_methods_and_attributes.py
```diff
@@ -431,6 +431,17 @@ def test_accepts_none(msg):
     assert "incompatible function arguments" in str(excinfo.value)
 
 
+def test_casts_none(msg):
```
rwgk, line 434: Post-merge comment: `msg` is unused and can be removed. Looks like a copy-paste accident.
(I'll create a PR to fix.)

PR #3060, opened 2021-06-25
CONTRIBUTING: Ensure `pip` gets upgraded; install `wheel`

EricCousineau-TRI:
Homogenize how to activate venv (use `source` instead of `.`)

## Description

While briefly investigating a possible new issue (https://github.com/pybind/pybind11/pull/2972#discussion_r658848867), was doing my local build env. Re-read `CONTRIBUTING`, and noted the followigng:
- When following current version (484b0f043), noted that nominal pip version `pip-9.0.1`, as installed by Ubuntu 18.04's `python3-venv` package, was emitting warnings like `Cache entry deserialization failed, entry ignored` (due to version mismatch).
- I know in other workflows, it can be useful to have `wheel`.
- Ideally, we should choose one way to activate.

This addresess those points.

## Suggested changelog entry:

N/A

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -59,6 +59,7 @@ system with CMake 3.14+:
 ```bash
 python3 -m venv venv
 source venv/bin/activate
+pip install -U pip wheel
 pip install -r tests/requirements.txt
 cmake -S . -B build -DDOWNLOAD_CATCH=ON -DDOWNLOAD_EIGEN=ON
 cmake --build build -j4
@@ -233,7 +234,8 @@ recent CMake and Python 3):
 
 ```bash
 python3 -m venv venv
-. venv/bin/activate
+source venv/bin/activate
+pip install -U pip wheel
 pip install pytest
 cmake -S . -B build-intel -DCMAKE_CXX_COMPILER=$(which icpc) -DDOWNLOAD_CATCH=ON -DDOWNLOAD_EIGEN=ON -DPYBIND11_WERROR=ON
 ```


henryiii:
Why install wheel? Users should use `pyproject.toml`, wheel is an implementation detail that shouldn't be installed by hand unless you are dealing with legacy packages that don't have `pyproject.toml`'s. If one of our dependencies needs it, perhaps? Hmm, if it is installing a wheel, but an old wheel, and the minimums are not specified in pyproject.toml, it will reuse the cached wheel - is this installing wheel, though? I would assume not.

The embedded pip that's installed by the `ensurepip` (and therefore `venv`) module can be out of date (especially on package managers that hack around with the embedded pip, like Debian), so updating that is a good idea. But why not use `--upgrade-deps`?

.github/CONTRIBUTING.md
```diff
@@ -59,6 +59,7 @@ system with CMake 3.14+:
 ```bash
 python3 -m venv venv
 source venv/bin/activate
+pip install -U pip wheel
```
henryiii, line 62:
```suggestion
python3 -m venv --upgrade-deps venv
source venv/bin/activate
```

Not sure when this was added, but guessing it's been around for a while? Using `virutalenv` ensures a recent versions of everything and is faster, but is not builtin.

henryiii: Ouch, this was added in Python 3.9. One of the problems of being on a Mac... ;)

```console
$ docker run --rm -it quay.io/pypa/manylinux_2_24_x86_64:latest bash
# python3.6 -m venv --help
# python3.7 -m venv --help
# python3.8 -m venv --help
# python3.9 -m venv --help
```

.github/CONTRIBUTING.md
```diff
@@ -59,6 +59,7 @@ system with CMake 3.14+:
 ```bash
```
henryiii, line 59:
```suggestion
```console
```

Not something for now, but these should likely be changed.

henryiii (approved):
Okay, fine with me, we can't support `--upgrade-deps` unless we make this 3.9+ only, and installing wheel is likely not needed, but if you really want it, I'll leave that up to you.

We might want to add a `noxfile.py`, that might make "fly-by" contributing easier.

PR #3077, opened 2021-07-06
Splitting out pybind11/stl/filesystem.h.

rwgk:
Unfortunately PR #2730 forces users of pybind11/stl.h to link against `-lstdc++fs` (or similar, depending on the platform) even if they don't actually use the new type caster. This issue is resolved here by moving the new type caster to pybind11/stl/filesystem.h. The `std::filessytem`-related link options will be needed only if the new header is included.

Further refinements of the cmake files (see PR comments) is beyond the scope of this PR and left for follow-on PRs.

Example breakage due to PR #2730: https://github.com/deepmind/open_spiel/runs/2999582108

This PR is mostly following the suggestion here: https://github.com/pybind/pybind11/pull/2730#issuecomment-750507575

Except using pybind11/stl/filesystem.h instead of pybind11/stlfs.h, as decided via chat.

stl.h restored to the exact state before merging PR #2730 via:
```
git checkout 733f8de24feed964f96b639a0a44247f46bed868 stl.h
```

NOTE: I will edit the #2730 change log entry after this PR is merged.

### DIFF
--- a/include/pybind11/stl/filesystem.h
+++ b/include/pybind11/stl/filesystem.h
@@ -1,0 +1,92 @@
+// Copyright (c) 2021 The Pybind Development Team.
+// All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#pragma once
+
+#include "../detail/common.h"
+
+#ifdef __has_include
+#  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
+      PY_VERSION_HEX >= 0x03060000
+#    include <filesystem>
+#    define PYBIND11_HAS_FILESYSTEM 1
+#  endif
+#endif
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+#if defined(PYBIND11_HAS_FILESYSTEM)
+template<typename T> struct path_caster {
+
+private:
+    static PyObject* unicode_from_fs_native(const std::string& w) {
+#if !defined(PYPY_VERSION)
+        return PyUnicode_DecodeFSDefaultAndSize(w.c_str(), ssize_t(w.size()));
+#else
+        // PyPy mistakenly declares the first parameter as non-const.
+        return PyUnicode_DecodeFSDefaultAndSize(
+            const_cast<char*>(w.c_str()), ssize_t(w.size()));
+#endif
+    }
+
+    static PyObject* unicode_from_fs_native(const std::wstring& w) {
+        return PyUnicode_FromWideChar(w.c_str(), ssize_t(w.size()));
+    }
+
+public:
+    static handle cast(const T& path, return_value_policy, handle) {
+        if (auto py_str = unicode_from_fs_native(path.native())) {
+            return module::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
+                   .release();
+        }
+        return nullptr;
+    }
+
+    bool load(handle handle, bool) {
+        // PyUnicode_FSConverter and PyUnicode_FSDecoder normally take care of
+        // calling PyOS_FSPath themselves, but that's broken on PyPy (PyPy
+        // issue #3168) so we do it ourselves instead.
+        PyObject* buf = PyOS_FSPath(handle.ptr());
+        if (!buf) {
+            PyErr_Clear();
+            return false;
+        }
+        PyObject* native = nullptr;
+        if constexpr (std::is_same_v<typename T::value_type, char>) {
+            if (PyUnicode_FSConverter(buf, &native)) {
+                if (auto c_str = PyBytes_AsString(native)) {
+                    // AsString returns a pointer to the internal buffer, which
+                    // must not be free'd.
+                    value = c_str;
+                }
+            }
+        } else if constexpr (std::is_same_v<typename T::value_type, wchar_t>) {
+            if (PyUnicode_FSDecoder(buf, &native)) {
+                if (auto c_str = PyUnicode_AsWideCharString(native, nullptr)) {
+                    // AsWideCharString returns a new string that must be free'd.
+                    value = c_str;  // Copies the string.
+                    PyMem_Free(c_str);
+                }
+            }
+        }
+        Py_XDECREF(native);
+        Py_DECREF(buf);
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+            return false;
+        } else {
+            return true;
+        }
+    }
+
+    PYBIND11_TYPE_CASTER(T, _("os.PathLike"));
+};
+
+template<> struct type_caster<std::filesystem::path>
+    : public path_caster<std::filesystem::path> {};
+#endif // PYBIND11_HAS_FILESYSTEM
+
+PYBIND11_NAMESPACE_END(detail)
+PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/stl/filesystem.h
+++ b/include/pybind11/stl/filesystem.h
@@ -4,7 +4,14 @@
 
 #pragma once
 
+#include "../cast.h"
+#include "../pybind11.h"
+#include "../pytypes.h"
+
 #include "../detail/common.h"
+#include "../detail/descr.h"
+
+#include <string>
 
 #ifdef __has_include
 #  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
@@ -38,7 +45,7 @@ private:
 public:
     static handle cast(const T& path, return_value_policy, handle) {
         if (auto py_str = unicode_from_fs_native(path.native())) {
-            return module::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
+            return module_::import("pathlib").attr("Path")(reinterpret_steal<object>(py_str))
                    .release();
         }
         return nullptr;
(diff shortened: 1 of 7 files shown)


include/pybind11/stl/filesystem.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include "../detail/common.h"
+
+#ifdef __has_include
+#  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
+      PY_VERSION_HEX >= 0x03060000
+#    include <filesystem>
+#    define PYBIND11_HAS_FILESYSTEM 1
+#  endif
+#endif
```
henryiii, line 15: Just a thought, do we need this `PYBIND11_HAS_FILESYSTEM` now? If I include `<stl/filesystem.h>` do I want it to silently not cast things instead of getting a compile-time error?

henryiii: (If not, it should be moved to the include in the test ;) )

include/pybind11/stl/filesystem.h
```diff
@@ -0,0 +1,92 @@
...
+                    PyMem_Free(c_str);
+                }
+            }
+        }
+        Py_XDECREF(native);
+        Py_DECREF(buf);
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+            return false;
+        } else {
```
Skylion007, line 79: else after return is not ideal style.

Skylion007 (approved):

include/pybind11/stl/filesystem.h
henryiii: Yes, it's called "guard" style to put the short returning branch at the top without an else, and it's preferable.

Skylion007:
Need to fix some of the pip tests.

henryiii:
There's a place or two where this needs to be added to the Python build system too. I'd do a quick `git grep stl.h` or maybe a rarer file and see where there are lists. Possibly only in tests, maybe in setup.py somewhere.

include/pybind11/stl/filesystem.h
Skylion007: @henryiii Plan on enabling a clang-tidy check for this soon.

include/pybind11/stl/filesystem.h
rwgk: > (If not, it should be moved to the include in the test ;) )

Copying my question from the slack channel here:

> In this first pass, I simply kept the #ifdef block defining PYBIND11_HAS_FILESYSTEM in the header file. Does that still make sense? Would it be better to move that to the test? — It seem useful to keep it in the header, so others can easily use it if available without having to do anything much otherwise, but I'm not sure how this shakes out best.

Ideally we'd give people a choice? Bail out with a compile-time error or have them handle the runtime error.

It's not really great if we lead users down the path of copy-pasting the `__has_include` logic from the test.

rwgk:
> There's a place or two where this needs to be added to the Python build system too. I'd do a quick `git grep stl.h` or maybe a rarer file and see where there are lists. Possibly only in tests, maybe in setup.py somewhere.

I tried that already but apparently I need to do something more involved because `stl` is a new subdirectory. Working on it.

henryiii:
One Python 2 failure is our random known flake, the other is just the packaging test (which happens to be Python 2 as well), which tests the file list.

We already have some directories (like detail) so whatever we do for that should work?

rwgk:
> One Python 2 failure is our random known flake, the other is just the packaging test (which happens to be Python 2 as well), which tests the file list.
> 
> We already have some directories (like detail) so whatever we do for that should work?

No good deed gets unpunished ... I think I finally figured it out. (I overlooked the tools directory longer than I'd like to admit.)

include/pybind11/stl/filesystem.h
henryiii: But do users want "works if available" behavior for this? (The answer may be yes) It seems a little weird if your package only accepts `pathlib.Path` arguments if it was compiled in C++17 mode on some compilers. And only allows you to return std::filesystem on Python 3 (though that I'm less worried about).  I'm guessing there will be some extra work most users will have to do to make sure this works (including requiring macOS 10.15+), so forcing the user to explicit might be better.

That is, I expect most users will either make an effort to support std::filesystem/pathlib everywhere, or they will not support it, but very few will support it if available. But could easily be wrong. I don't find the "auto" selection of C++ level all that useful either, but some users want it.

henryiii: I'm okay to leave it as is, a user (who looks at the source) can check `PYBIND11_HAS_FILESYSTEM` if they want. But I'm also okay to (re)move it to the tests.

include/pybind11/stl/filesystem.h
rwgk: > else after return is not ideal style.

Done.

include/pybind11/stl/filesystem.h
rwgk: I added `PYBIND11_HAS_FILESYSTEM_IS_OPTIONAL`.
If not defined, users will see a compile-time error.
If that's not desired, they can opt in to handling the runtime error, without having to copy-paste the `__has_include` logic.

I don't think we need to document this. The safe behavior (compile-time error) is the default, the error message tells people what the problem is and what they can do to ignore it.

rwgk:
Happy day, the mac os straggler made it :-)

@henryiii I hope this PR is final now, could you please review and approve if this looks good to you?

jagerman:
I don't think any of the CI's are going to catch this, but with clang before 9 and gcc before 9 (or any clang with stdlibc++ before before 9) you also need to link with `-lstdc++fs` (stdlibc++) or `-lc++fs` (clang+llvm).  Might be worth documenting that, or perhaps providing some opt-in cmake machinery to help with setting up the linking?

https://github.com/oxen-io/lokinet/blob/dev/cmake/check_for_std_filesystem.cmake

is some cmake code we wrote to deal with figuring out the linking mess.

Skylion007:
There is a cmake community module for FileSystem: https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake

henryiii:
> Might be worth documenting that, or perhaps providing some opt-in cmake machinery to help with setting up the linking?

This is why I liked just making it an error if `pybind11/std/filesystem.h` is included but the filesystem library is not available, either due to linking requirements, macOS target settings, or for other reasons. Having a "try to use it if available" makes us responsible for deciding that correctly (which is hard), and for providing tools to make it optionally available elsewhere. The current, undocumented, tiny bit of code that enables this is not too bad, since we aren't advertising it.

It's not our responsibility to provide support for filesystem for the user, just for allowing conversion if a user already supports it. IMO. But docs on how to use it, including links to CMake modules that do the linking, etc, are fine and useful.

include/pybind11/stl/filesystem.h
henryiii: `__has_include` is not actually a C++17 feature, it's a [C++20 one]()https://en.cppreference.com/w/cpp/feature_test. Though I think all major compilers added it very early, but still, maybe it's best to remove this and just `#include` filesystem and let the failure happen if it's not available, rather than calling `#error`. This also makes `PYBIND11_HAS_FILESYSTEM` rather weird, though.

Skylion007: @henryiii This page suggest it is a C++17 feature though? https://en.cppreference.com/w/cpp/preprocessor/include

Skylion007: official spec: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0061r1.html

rwgk: `__has_include` currently appears in 10 source code locations on master (see below).

If think for this PR it's important to stay focused on the one problem we want to solve as soon as possible: users shouldn't be forced to work on the link options if they don't actually want to use the new caster. I'll update the PR description accordingly.

```
include/pybind11/cast.h:#  if defined(__has_include)
include/pybind11/cast.h:#    if __has_include(<string_view>)
include/pybind11/stl.h:#ifdef __has_include
include/pybind11/stl.h:#  if defined(PYBIND11_CPP17) && __has_include(<optional>)
include/pybind11/stl.h:#  if defined(PYBIND11_CPP14) && (__has_include(<experimental/optional>) && \
include/pybind11/stl.h:                                 !__has_include(<optional>))
include/pybind11/stl.h:#  if defined(PYBIND11_CPP17) && __has_include(<variant>)
include/pybind11/stl.h:#  if defined(PYBIND11_CPP17) && __has_include(<filesystem>) && \
include/pybind11/detail/common.h:#if defined(__has_include)
include/pybind11/detail/common.h:#  if __has_include(<version>)
```

rwgk:
> Having a "try to use it if available" makes us responsible for deciding that correctly (which is hard),

Yes exactly. This is the issue meant to be solved by this PR.

> It's not our responsibility to provide support for filesystem for the user, just for allowing conversion if a user already supports it. IMO. But docs on how to use it, including links to CMake modules that do the linking, etc, are fine and useful. (and the above caveats).

These could be nice follow-on PRs.

include/pybind11/stl/filesystem.h
henryiii: Yes, @Skylion007 is correct, I was muddling has_include with feature test macros - probably shouldn't comment on things while in a conference. :)

This plans seems fine, we need to fix this, then we can polish and or document a bit more later if needed.

henryiii (approved):

PR #3078, opened 2021-07-07
Prep-work for clang-format.

rwgk:
Prep-work for global clang-format (see **experimental/demo** PRs #3074, #3082).

NO bulk changes in this PR, just minimal changes such that a global clang-format is stable after a single pass.

This was several hours of tedious work running clang-format & clang-tidy until both are stable. Breaking out as a separate PR so that these manual changes do not have to be maintained separately while we decide how to integrate clang-format or clang-format-diff into the pre-commit hooks.

Notable steps:

* Systematically adding `// clang-format off`, `on` around all `\rst` `\endrst` blocks. — Possibly some could be omitted, but for blocks that are deliberately formatted as `\rst` it is safer to systematically exclude them from clang-format. Note that the sphinx error messages are often completely useless if something goes wrong, therefore letting clang-format loose on `\rst` blocks is likely to turn into time-consuming surprises.

* Removing stray semicolons (discovered by running clang-format v12 followed by tools/check-style.sh).

* Manually moving clang-tidy `// NOLINT` and `// NOLINTNEXTLINE` comments so that clang-format does not move them to the wrong places. (This was the most tedious part.)

* Manually reformatting comments related to `static_assert`s so that clang-format does not need two passes.

* `git diff -U0 --no-color HEAD^ | python3 $HOME/clone/llvm-project/clang/tools/clang-format/clang-format-diff.py -p1 -style=file -i`

## Suggested changelog entry:

```rst
Prep-work for adding clang-format to the pre-commit hooks.
```

PR #3086, opened 2021-07-09
Fixing select clang-tidy issues to reduce NOLINTNEXTLINE use.

rwgk:
This was really meant to be part of PR #3051 but was held back either out of an abundance of caution, or because of confusion caused by stray semicolons.

The particular `NOLINT`s removed here were stumbling stones for the ongoing work under PR #3078. Breaking out the fixes into a separate PR to disconnect them from other questions.

clang-format-diff was applied.


Skylion007 (approved):

rwgk:
Thanks @Skylion007 for the review, and for your help while I was working on this!

PR #3087, opened 2021-07-09
Prep-work for clang-format excluding `\rst`-related issues.

rwgk:
Prep-work to establish a clean base for clang-format or clang-format-diff (see **experimental/demo** PR #3074).

NO bulk changes in this PR, just minimal changes such that a global clang-format is stable after a single pass.

The demo PR #3074 has two automatic changes applied on top of this PR:
* Automatic insertion of `// clang-format off`, `on` pairs around `\rst`, `\endrst` comment blocks.
* Bulk global clang-format via pre-commit.
This is to prove that this prep work is the clean base we need.

This prep work was several hours of tedious work running clang-format & clang-tidy until both are stable. Breaking out as a separate PR so that these manual changes do not have to be maintained separately while we decide how to integrate clang-format or clang-format-diff into our GitHub Actions.

Notable steps:

* Removing stray semicolons (discovered by running clang-format v12 followed by tools/check-style.sh).

* Manually moving clang-tidy `// NOLINT` comments so that clang-format does not move them to the wrong places.

* Manually reformatting comments related to `static_assert`s so that clang-format does not need two passes.

* `git diff -U0 --no-color HEAD^ | python3 $HOME/clone/llvm-project/clang/tools/clang-format/clang-format-diff.py -p1 -style=file -i`

## Suggested changelog entry:

```rst
Prep-work to establish a clean base for clang-format or clang-format-diff.
```

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -62,7 +62,8 @@ struct metaclass {
     handle value;
 
     PYBIND11_DEPRECATED("py::metaclass() is no longer required. It's turned on by default now.")
-    metaclass() { } // NOLINT(modernize-use-equals-default): breaks MSVC 2015 when adding an attribute
+    // NOLINTNEXTLINE(modernize-use-equals-default): breaks MSVC 2015 when adding an attribute
+    metaclass() {}
 
     /// Override pybind11's default metaclass
     explicit metaclass(handle value) : value(value) { }
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -801,7 +801,8 @@ struct nodelete { template <typename T> void operator()(T*) { } };
 PYBIND11_NAMESPACE_BEGIN(detail)
 template <typename... Args>
 struct overload_cast_impl {
-    constexpr overload_cast_impl() {}; // NOLINT(modernize-use-equals-default):  MSVC 2015 needs this
+    // NOLINTNEXTLINE(modernize-use-equals-default):  MSVC 2015 needs this
+    constexpr overload_cast_impl() {}
 
     template <typename Return>
     constexpr auto operator()(Return (*pf)(Args...)) const noexcept
--- a/tests/pybind11_tests.h
+++ b/tests/pybind11_tests.h
@@ -1,10 +1,15 @@
 #pragma once
+
+// This must be kept first for MSVC 2015.
+// Do not remove the empty line between the #includes.
 #include <pybind11/pybind11.h>
+
 #include <pybind11/eval.h>
 
 #if defined(_MSC_VER) && _MSC_VER < 1910
 // We get some really long type names here which causes MSVC 2015 to emit warnings
-#  pragma warning(disable: 4503) // warning C4503: decorated name length exceeded, name was truncated
+#    pragma warning(                                                                              \
+        disable : 4503) // warning C4503: decorated name length exceeded, name was truncated
 #endif
 
 namespace py = pybind11;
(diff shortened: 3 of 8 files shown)


tests/test_numpy_array.cpp
```diff
@@ -275,10 +275,16 @@ TEST_SUBMODULE(numpy_array, sm) {
     // [workaround(intel)] ICC 20/21 breaks with py::arg().stuff, using py::arg{}.stuff works.
 
     // Only accept the exact types:
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    sm.def("overloaded3", [](py::array_t<int>) { return "int"; }, py::arg{}.noconvert());
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    sm.def("overloaded3", [](py::array_t<double>) { return "double"; }, py::arg{}.noconvert());
+    sm.def(
+        "overloaded3",
+        // NOLINTNEXTLINE(performance-unnecessary-value-param)
```
Skylion007, line 280: These probably could const refs to make it happy.

tests/test_stl.cpp
```diff
@@ -299,12 +299,15 @@ TEST_SUBMODULE(stl, m) {
     m.def("stl_pass_by_pointer", [](std::vector<int>* v) { return *v; }, "v"_a=nullptr);
 
     // #1258: pybind11/stl.h converts string to vector<string>
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    m.def("func_with_string_or_vector_string_arg_overload", [](std::vector<std::string>) { return 1; });
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    m.def("func_with_string_or_vector_string_arg_overload", [](std::list<std::string>) { return 2; });
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    m.def("func_with_string_or_vector_string_arg_overload", [](std::string) { return 3; });
+    m.def("func_with_string_or_vector_string_arg_overload",
```
Skylion007, line 302: Could be const refs to make it happy.

tests/test_numpy_array.cpp
rwgk: Done here and in tests/test_numpy_array.cpp, but systematically in both files.

But I want to propose to do this even more systematically: I volunteer to do this everywhere, in one PR.

Reason: picking out a place here and there, just because the associated original source code line happens to be longer than 99 characters, is really random. If people look later they will wonder why some are `const T &` while others are `T` with the `NOLINT` comment, and possibly read something into it. That's really misleading.

It was nice to use the `NOLINT` to enable activating the clang-tidy option without having to get into the cleanup, but I think doing the cleanup sooner rather than later is best.

If you agree I'll do the systematic `NOLINT` reduction*** first, then come back to this PR after merging and rebasing.

*** I'm assuming there could be a couple exceptions.

Skylion007: Okay sounds good. It's not in too many places I suppose.

rwgk:
This PR has a lot of merge conflicts now. I need to take care of some other things first but will come back here asap.

rwgk:
For future reference, here are a couple interesting links I found when searching for information about clang-format & `\rst`:
* https://stackoverflow.com/questions/48016641/force-clang-format-to-respect-doxygen-block-comments
* https://bugs.llvm.org/show_bug.cgi?id=44486

Skylion007 (approved):

henryiii (approved):

rwgk:
Thanks Aaron and Henry!

PR #3070, opened 2021-07-01
Some fixes

cyyever:
Fix some code problems found by PVS-Studio  and clang-tidy analyzers.

PR #3094, opened 2021-07-11
fix(clang-tidy): Add cppcoreguidelines-init-vars,slicing, and throw-by-value-catch-by-reference checks

Skylion007:
## Description

* This turns on a clang-tidy check for some of the issues found #3070. Basically, it just ensures that all simple local variables (ints, floats, etc) are initialized with some same default value so that if a codepath forgets to initialize them later. It will be easier to debug and won't produce potential UB behavior which is difficult to debug latter. Since it only checks primitive vars, they are very cheap to initialize and the long term benefits for maintainability are worth it.
* Adds a check to warn if do not you throw by value and catch by reference which is bad C++ style. We don't do that anywhere here, but this check.
* ~Also adds a check that warns against confusing slicing of the vtable. We don't do that here, but this check will guard against it.~ Our enum macros break this rule and would be annoying to fix.
* Also tweaks the clang-tidy config to warn on performance issues when using auto-incorrectly. Not currently an issue, but will guard against future issues.
## Suggested changelog entry:
* Enable cppcoreguidelines-init-variables, ~cppcoreguidelines-slicing~, and misc-throw-by-value-catch-by-reference checks (Update: removed coreguidelines-slicing since it doesn't like our enum op hacking).

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,7 +2,6 @@ FormatStyle: file
 
 Checks: '
 cppcoreguidelines-init-variables,
-cppcoreguidelines-slicing,
 clang-analyzer-optin.cplusplus.VirtualCall,
 llvm-namespace-comment,
 misc-misplaced-const,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,10 +1,13 @@
 FormatStyle: file
 
 Checks: '
+cppcoreguidelines-init-variables,
+cppcoreguidelines-slicing,
 clang-analyzer-optin.cplusplus.VirtualCall,
 llvm-namespace-comment,
 misc-misplaced-const,
 misc-static-assert,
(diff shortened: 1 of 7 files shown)


henryiii:
Needs a rebase, and we can add the PR number to the list in the changelog.

Skylion007:
@henryiii rebased

henryiii (approved):
I assume this might have a _tiny_ runtime cost? Or is the compiler smart enough to tell they are never read before writing? Either way, don't think it's measurable and this is safer.

henryiii:
PS: Not using it for this one (just adding the PR number to the list), but normally, do not delete the `rst` block, but place the entry inside it. The `nox -s make_changelog` script will read this from the GitHub API and combine it into a draft changelog.

PR #3097, opened 2021-07-12
[smart_holder] WIP: Fix for Python 3.9 MSAN/valgrind errors.

rwgk:
WIP — Currently this is NOT A FIX, just hoping to become one.

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -422,7 +422,7 @@ extern "C" inline void pybind11_object_dealloc(PyObject *self) {
     auto type = Py_TYPE(self);
     type->tp_free(self);
 
-#if PY_VERSION_HEX < 0x03080000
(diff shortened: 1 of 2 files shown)


henryiii:
Please prefix all PRs to non-master with `[branch-name]`. :)

rwgk:
> Please prefix all PRs to non-master with `[branch-name]`. :)

Done and will do.
FWIW: this one has a chance to turn into a PR against master, but not sure until we have an actual fix + reproducer.

henryiii:
That's fine, but that will be a new PR (or at least a changed PR), so the name can be different. :)

rwgk:
Drive-by cleanup: closing.

I was told the root issue was fixed in Python 3.9.something. I assume the fix trickled through the system in the meantime. Any more work here isn't going to be of great help to world at large.

PR #3099, opened 2021-07-12
docs: update changelog

henryiii:
## Description

Updates the changelog with the marked PRs.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,11 +9,153 @@ Starting with version 1.8.0, pybind11 releases use a `semantic versioning
 v2.7.0 (TBA, not yet released)
 ------------------------------
 
+New features:
+
+* Enable ``py::implicitly_convertible<py::none, ...>`` for
+  ``py::class_``-wrapped types.
+  `#3059 <https://github.com/pybind/pybind11/pull/3059>`_
+
+* Allow function pointer extraction from overloaded functions.
+  `#2944 <https://github.com/pybind/pybind11/pull/2944>`_
+
+* NumPy: added ``.char_()`` to type which gives the NumPy public ``char``
+  result, which also distinguishes types by bit length (unlike ``.kind()``).
+  `#2864 <https://github.com/pybind/pybind11/pull/2864>`_
+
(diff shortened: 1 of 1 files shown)


rwgk:
The flake is #2995 again.
I think it's very important to merge #2995 before the 2.7 release.

henryiii:
But on PyPy macOS this time?

docs/changelog.rst
```diff
@@ -9,12 +9,154 @@ Starting with version 1.8.0, pybind11 releases use a `semantic versioning
...
+* Checks for common misspellings were added to the pre-commit hooks.
+  `#3076 <https://github.com/pybind/pybind11/pull/3076>`_
+
+* Enable and apply checks for more misc, readability, and code modernization
+  improvements to the codebase. Remove redundant function ptr dereference,
+  makes methods const, and uses member-initializers when possible to allow for
+  trivial construction.
+  `#3052 <https://github.com/pybind/pybind11/pull/3052>`_
+
+* Enable clang-tidy performance checks throughout the codebase to enforce
```
rwgk, line 139: Might be nice to combine this with the next two, with one intro, then listing the 3 PRs.

For example:

Enable clang-tidy performance and modernization checks throughout the codebase to enforce best coding practices.

rwgk (commented):
This looks great, my main concern is that #2995 isn't included.

I think we need to be completely open about the undefined behavior, to not mislead users. I'm also thinking that we are more likely to attract a volunteer to develop a fix if we're open about the issue.

To add my own experience, it is no exaggeration that I scrolled through failed jobs 100+ times to be sure it's that flake, and it's an error each time I run TSAN.

I'd be happy quickly address suggestions on #2995.

henryiii:
This is not the final changelog before 2.7 (though I expect we are getting close), just an update to the current changelog that has been _long_ delayed.

docs/changelog.rst
henryiii: I mostly went with the automatic script, which is one per PR, then with minor touchups, but I agree, will modify this soon(ish). :)

PR #3102, opened 2021-07-12
chore: update pre-commit, add a check

henryiii:
- chore: update pre-commit hooks
- chore: move some config to pyproject.toml
- style: add pyupgrade check, 2.7+

### DIFF
--- a/tests/extra_python_package/test_files.py
+++ b/tests/extra_python_package/test_files.py
@@ -126,7 +126,7 @@ def test_build_sdist(monkeypatch, tmpdir):
     with tarfile.open(str(sdist)) as tar:
         start = tar.getnames()[0] + "/"
         version = start[9:-1]
-        simpler = set(n.split("/", 1)[-1] for n in tar.getnames()[1:])
+        simpler = {n.split("/", 1)[-1] for n in tar.getnames()[1:]}
 
         with contextlib.closing(
             tar.extractfile(tar.getmember(start + "setup.py"))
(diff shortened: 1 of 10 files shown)


tests/extra_python_package/test_files.py
```diff
@@ -138,9 +138,9 @@ def test_build_sdist(monkeypatch, tmpdir):
         ) as f:
             pyproject_toml = f.read()
 
-    files = set("pybind11/{}".format(n) for n in all_files)
```
rwgk, line 141: I'm fine with this, but JIC others feel the same: I think this obfuscates the code (reduces readability). With this change I have to squint to figure out: is this a dict? or a set? As-is it's totally obvious.

Skylion007 (approved):

rwgk (approved):

tests/extra_python_package/test_files.py
henryiii: There's no colon in it - in Python, you always have to see if there's a colon to distinguish between sets and dicts. Same would be true of `{a}` vs `{a:b}`. The fact that this is immediately used like a set helps too.

If it's unclear, I'd either add "set" to the name, or use a static type set (Python 3), rather than avoid writing it with the built-in language syntax.

henryiii: (if you aren't used to seeing it, perhaps, it could be due to the fact it was a 2.7+ feature, and most code used to support 2.6/2.5, etc and is not always updated)

PR #3106, opened 2021-07-13
Simple provision to help dealing with flakey CI jobs (repeat_command.py)

rwgk:
Experiments trying to understand the MSVC test_iostream.py flakiness.


Skylion007:
There is also pytest-flake which could be useful here.

rwgk:
If anyone wants to use my trick to deflake tests in the future, PR #3180 should make it straightforward to pin-point where the tests are run, and repeat_command.py now lives here: https://github.com/rwgk/rwgk_config/tree/master/bin

Permalink: https://github.com/rwgk/rwgk_config/blob/e8396e6126741b8916dae91b3024fa0d713e2cea/bin/repeat_command.py

The only slight wrinkle is that one ci.yml job runs the Python tests outside the pybind11 source directory. This is what I did for that one special case:
```
    - name: Copy tools/repeat_command.py to /tmp
      run: cp -a tools/repeat_command.py /tmp
...
      run: python3 /tmp/repeat_command.py 1 make pytest -j 2
```

PR #3111, opened 2021-07-14
style: pre-commit cleanup

henryiii:
- style: disallow PyTest (should be pytest)
- style: cleanup spell checking a bit
- style: add a few items to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ pybind11Targets.cmake
 /.vscode
 /pybind11/include/*
 /pybind11/share/*
+/docs/_build/*
+.ipynb_checkpoints/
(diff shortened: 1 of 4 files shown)


Skylion007:
Should the changelog update be here or in a seperate PR, we might have a few more PRs merge in before the release window.

henryiii:
If a PR is going to be merge quickly, it's okay to add a changelog update there, but normally, we should collect and update the changelog in batches (there's a script that does that). I don't think we need a changelog update for this specific PR, though, as it's just docs?

Skylion007:
@henryiii Blocked by #3114 which should improve the quality of this script as well.

Skylion007 (approved):

PR #3113, opened 2021-07-14
Use cpp17 constexpr for casting

Skylion007:
## Description
This is an attempt to modernize and revive: https://github.com/pybind/pybind11/pull/1589 for discussion. Particularly, the part about using constexpr for C++17 > . Closes #1589. This does increase the size of the so by about 224 bytes for C++17, but hopefully that will be worth it. I am using a lot of the STL I am unfamiliar with here. Open to closing this PR if it's determined to have no benefit or causes bizzarre compiler issues. Just wanted to start some machinery to use constexpr more aggressively in more modern C++ dialects that support

## Suggested changelog entry:

```rst
* Allows constexpr to be used during casting in C++17 or newer.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -887,7 +887,6 @@ T cast(const handle &handle) { return T(reinterpret_borrow<object>(handle)); }
 template <typename T, detail::enable_if_t<!detail::is_pyobject<T>::value, int> = 0>
 object cast(T &&value, return_value_policy policy = return_value_policy::automatic_reference,
             handle parent = handle()) {
-    // using no_ref_T = PYBIND11_CPP17_CONSTEXPR std::remove_reference<T>::type;
     if (policy == return_value_policy::automatic) {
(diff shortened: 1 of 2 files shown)


rwgk:
Your implementation avoiding the copied source code sections is definitely much better, but does it actually do anything?
Did you check if the +224 bytes size difference is reproducible going back and forth compiling from scratch a few times? Maybe it's just noise?

The compiler already knows that the `::value` expressions are compile-time constant, the added `constexpr` seem redundant, the optimizer will probably (certainly?) figure out that it can discard the branch, but it could be that I'm missing something.

IIUC the main benefit of `if constexpr` is that - inside templates - template expansion is NOT applied to the code inside the `if` if false, which enables a much nicer and more intuitive code structure compared to the usually mind-bending alternative of SFINAE acrobatics. But that's not the situation in this PR, the `if constexpr` bodies are just simple enum assignments.

PR #3114, opened 2021-07-14
style:(pre-commit) Add shellcheck style checking

Skylion007:
## Description
* Adds a linter for shell scripts which helps improve our shellscripts

## Suggested changelog entry:

```rst
* Adds a shellcheck-py pre-commit hook and improve style of shell scripts
```


henryiii (approved):

henryiii:
We don't have many shell scripts ( :smile: ), but good.

PR #3116, opened 2021-07-15
Move object in pop method of List.

blacktea:
I suggest to move object because it should be deleted from vector, and move constructor likely faster than copy constructor.
BTW, If the class hasn't got a move constructor it would be copied.

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -191,7 +191,7 @@ void vector_modifiers(enable_if_t<is_copy_constructible<typename Vector::value_t
         [](Vector &v) {
             if (v.empty())
                 throw index_error();
-            T t{ std::move(v.back()) };
+            T t = std::move(v.back());
             v.pop_back();
             return t;
         },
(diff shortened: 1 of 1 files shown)


Skylion007:
@blacktea Seems like all the compilers aren't happy about this.

blacktea:
> @blacktea Seems some of the compilers aren't happy about this.

Yep. It seems a compiler bug.
Please, look at it: https://godbolt.org/z/8ocjEzs8d
Clang 3.6 warns, Clang 3.7 is happy.
I will change using copy initialization.

Skylion007:
Actually according to the official spec, I am pretty sure these moves are unnecessary as returned local variables are automatically moved per the specification as long as they aren't marked as const or something: https://clang.llvm.org/extra/clang-tidy/checks/performance-no-automatic-move.html , so I am not sure this PR is even necessary.

Skylion007 (approved):
Looks good to me, @rwgk @henryiii any objections?

rwgk (approved):
LGTM. Feeling good enough about it that it doesn't need the Google testing. (And it's small so easy to roll back in the unlikely event that something goes wrong.)

PR #3120, opened 2021-07-15
Introducing PYBIND11_VERSION_HEX

rwgk:
## Description

Better late than never!

## Suggested changelog entry:

```rst
A ``PYBIND11_VERSION_HEX`` define was added, similar to ``PY_VERSION_HEX``.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,36 @@ VERSION_REGEX = re.compile(
     r"^\s*#\s*define\s+PYBIND11_VERSION_([A-Z]+)\s+(.*)$", re.MULTILINE
 )
 
+
+def build_expected_version_hex(matches):
+    patch_level_serial = matches["PATCH"]
+    serial = None
+    try:
+        major = int(matches["MAJOR"])
+        minor = int(matches["MINOR"])
+        flds = patch_level_serial.split(".")
+        if flds:
+            patch = int(flds[0])
+            level = None
+            if len(flds) == 1:
+                level = "0"
+                serial = 0
+            elif len(flds) == 2:
+                level_serial = flds[1]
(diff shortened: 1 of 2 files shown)


henryiii:
This should be checked for consistency, just like the C++ and Python versions are verified to match by setup.py.

rwgk:
> This should be checked for consistency, just like the C++ and Python versions are verified to match by setup.py.

I agree!
But UGH ... so much code for something so trivial.
Please take another look.

Skylion007 (approved):

henryiii (approved):

setup.py
```diff
@@ -49,6 +79,15 @@
     )
     raise RuntimeError(msg)
 
+version_hex = matches.get("HEX", "MISSING")
+expected_version_hex = build_expected_version_hex(matches)
```
henryiii, line 83: At this point, already have the `version`, can't you just compute the expected version hex from that?

rwgk: IIUC that's what I'm doing. The subproblem that makes the code long is handling `patch_level_serial`, e.g. `"3.dev1"`. And then we also need the conversion from base-10 to base-16 (hex), which doesn't happen anywhere else.

The only alternative I see is to do it the other way around, building expected_major, expected_minor, expected_patch from the hex value, but I think it'll be a very similar amount of code.

In any case, it's done "good enough for our purposes" I feel, and not that much code :-)

henryiii: I might take a stab at refactoring it eventually, but good for now. :)

rwgk: > I might take a stab at refactoring it eventually, but good for now. :)

Sounds good, as long as it doesn't look like write-only Perl at the end ;-)

PR #3109, opened 2021-07-13
docs: try using Furo

henryiii:
## Description

The current theme is broken; trying out a new theme. Using Furo at the moment, which used by pip and several other PyPA projects, along with others. Has really nice dark mode support.

Preview the site by clicking the readthedocs check below.

## Suggested changelog entry:

```rst
* New theme for the documentation.
```

PR #3121, opened 2021-07-15
chore: support PDF from nox

henryiii:
## Description

Include PDF creation from nox.

Pulled from #3109.


henryiii:
Already reviewed in #3109.

PR #3124, opened 2021-07-16
chore: increase CMake upper limit

henryiii:
## Description

We were relying on https://cmake.org/cmake/help/latest/policy/CMP0112.html OLD. Only affects tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,13 +7,13 @@
 
 cmake_minimum_required(VERSION 3.4)
 
-# The `cmake_minimum_required(VERSION 3.4...3.18)` syntax does not work with
+# The `cmake_minimum_required(VERSION 3.4...3.21)` syntax does not work with
 # some versions of VS that have a patched CMake 3.11. This forces us to emulate
 # the behavior using the following workaround:
-if(${CMAKE_VERSION} VERSION_LESS 3.18)
+if(${CMAKE_VERSION} VERSION_LESS 3.21)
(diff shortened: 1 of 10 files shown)


henryiii:
Well that didn't work. Will have to investigate later to see what we are depending on that changes when the limit is increased.

CMakeLists.txt
```diff
@@ -10,10 +10,10 @@ cmake_minimum_required(VERSION 3.4)
 # The `cmake_minimum_required(VERSION 3.4...3.18)` syntax does not work with
 # some versions of VS that have a patched CMake 3.11. This forces us to emulate
 # the behavior using the following workaround:
-if(${CMAKE_VERSION} VERSION_LESS 3.18)
+if(${CMAKE_VERSION} VERSION_LESS 3.21)
```
henryiii, line 13:
```suggestion
if(${CMAKE_VERSION} VERSION_LESS 3.19)
```

CMakeLists.txt
```diff
@@ -10,10 +10,10 @@ cmake_minimum_required(VERSION 3.4)
...
   cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
 else()
-  cmake_policy(VERSION 3.18)
+  cmake_policy(VERSION 3.21)
```
henryiii, line 16:
```suggestion
  cmake_policy(VERSION 3.19)
```

Skylion007 (approved):

rwgk (approved):

PR #3125, opened 2021-07-16
WIP: pragma block reduction

rwgk:
Spreadsheet: **[pybind11.h v2.7.0 pragma block analysis](https://docs.google.com/spreadsheets/d/1XDKkq1w7d9f2dPcVsn-vmZHleMWQkVCl7xb4cilpGiI/edit?usp=sharing)**

Approach:
```
for pragma in pragmas:
    remove_from_pybind11_h(pragma)
    run_ci_with_tests_disabled()
    collect_logs()
manually_sort_through_logs()
manually_add_local_pragma_push_pop()
```

Keeping track of some simple results here:
* GCC `-Wunused-but-set-parameter` needs work
* GCC `-Wunused-but-set-variable` rm_ok
* GCC `-Wmissing-field-initializers` rm_ok
* GCC `-Wattributes` needs work
* GCC `-Wnoexcept-type` needs work (gcc 7 only)
* MSVC `C4100` needs work
* MSVC `C4127` needs work
* MSVC `C4512` rm_ok
* MSVC `C4800` needs work
* MSVC `C4996` needs work
* MSVC `C4702` rm_ok
* MSVC `C4522` needs work
* MSVC `C4505` needs work
* ICC `186` rm_ok
* ICC `878` needs work
* ICC `1334` rm_ok
* ICC `1682` rm_ok
* ICC `1786` rm_ok
* ICC `1875` rm_ok
* ICC `2196` needs work

PR #3127, opened 2021-07-17
Removing all warning pragmas that have not effect.

rwgk:
The 10 (out of 21) obsolete pragmas were identified via experiments under draft/scratch PR #3125.


rwgk:
Deflaking.

henryiii:
I was guessing we'd have removed the need for some of these with all the cleanups we have been doing!

henryiii (approved):

PR #3129, opened 2021-07-17
Removing MSVC C4996 from pragma block at the top of pybind11.h

rwgk:

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -78,14 +78,9 @@ inline bool apply_exception_translators(std::forward_list<ExceptionTranslator>& 
 }
 
 #if defined(_MSC_VER)
-#    pragma warning(push)
-#    pragma warning(disable : 4996) // warning C4996: The POSIX name for this item is deprecated.
-#endif
-
-inline char *strdup_helper(const char *s) { return strdup(s); }
-
-#if defined(_MSC_VER)
-#    pragma warning(pop)
+#    define PYBIND11_COMPAT_STRDUP _strdup
+#else
+#    define PYBIND11_COMPAT_STRDUP strdup
 #endif
 
 PYBIND11_NAMESPACE_END(detail)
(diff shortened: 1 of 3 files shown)


rwgk:
I'm having second thoughts after review of this spreadsheet: **[pybind11.h v2.7.0 pragma block analysis](https://docs.google.com/spreadsheets/d/1XDKkq1w7d9f2dPcVsn-vmZHleMWQkVCl7xb4cilpGiI/edit?usp=sharing)**

Marking this PR as draft for now.

include/pybind11/pybind11.h
```diff
@@ -56,6 +57,21 @@
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+#if defined(_MSC_VER)
```
Skylion007, line 62: Instead of poping the warning, I think the recommended way to handle this error is actually to call the  ```_strdup``` on MSC: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=msvc-160

rwgk: The 2nd thoughts I had when marking this PR as draft yesterday:

* Cluttering pybind11 with MSVC-specific things is ugly.
* Better to simply disable the warning on the command line?

To me it looks like just another MSVC oddity, dirt that I want to keep out of my house, to use stronger words.

But we have similar pros-and-cons for most of the 11 items that I enumerated here yesterday:
**[pybind11.h v2.7.0 pragma block analysis](https://docs.google.com/spreadsheets/d/1XDKkq1w7d9f2dPcVsn-vmZHleMWQkVCl7xb4cilpGiI/edit?usp=sharing)**

I was thinking of suggesting a focused action meeting to go through the list and decide what to do about each one. I'll ping you and @henryiii on slack about it.

henryiii: I think we should do what MSVC wants; like it or not, they are a major target for us. We should not requires any special flags to build without warnings; this is a header-only library and is not _always_ called through build methods we control.

rwgk:
@henryiii @Skylion007 Please take another look. The latest version is completely different compared to the initial one.

Skylion007 (approved):

henryiii (approved):

rwgk:
Thanks! I'll merge this now, then rebase the other PRs.

PR #3132, opened 2021-07-18
Fix Mingw64 build and add a CI job to test it

jeromerobert:
This is a tentative to fix mingw64 build and add a CI job to test it. So far it does not work. I would need some help or explanation to fix it.
* A `__declspec(dllexport)` symbol cannot be hidden, it must have the default visibility (this is required by gcc). https://github.com/pybind/pybind11/commit/3ac690b88 make `builtin_exception` `__declspec(dllexport)` while the `pybind11` namespace is `visibility(hidden)`.
* I get a huge amount of deprecation like `warning: 'pybind11::object::object(pybind11::handle, bool)' is deprecated: Use reinterpret_borrow<object>() or reinterpret_steal<object>()`

This is the follow-up.
* https://github.com/pybind/pybind11/pull/2921
* https://github.com/pybind/python_example/pull/75

## Suggested changelog entry:

```rst
* Fix Mingw64 and add to the CI testing matrix.
```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -89,6 +89,16 @@
 #  endif
 #endif
 
+#if !defined(PYBIND11_EXPORT_EXCEPTION)
+#  ifdef __MINGW32__
+// workaround for:
+// error: 'dllexport' implies default visibility, but xxx has already been declared with a different visibility
+#    define PYBIND11_EXPORT_EXCEPTION
+#  else
+#    define PYBIND11_EXPORT_EXCEPTION PYBIND11_EXPORT
+#  endif
+#endif
+
 #if defined(_MSC_VER)
 #  define PYBIND11_NOINLINE __declspec(noinline)
 #else
@@ -740,7 +750,7 @@ PYBIND11_NAMESPACE_END(detail)
 #  pragma warning(disable: 4275) // warning C4275: An exported class was derived from a class that wasn't exported. Can be ignored when derived from a STL class.
 #endif
 /// C++ bindings of builtin Python exceptions
-class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
+class PYBIND11_EXPORT_EXCEPTION builtin_exception : public std::runtime_error {
 public:
     using std::runtime_error::runtime_error;
     /// Set the error using the Python C API
@@ -751,7 +761,7 @@ public:
 #endif
 
 #define PYBIND11_RUNTIME_EXCEPTION(name, type) \
-    class PYBIND11_EXPORT name : public builtin_exception { public: \
+    class PYBIND11_EXPORT_EXCEPTION name : public builtin_exception { public: \
         using builtin_exception::builtin_exception; \
         name() : name("") { } \
         void set_error() const override { PyErr_SetString(type, what()); } \
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -105,7 +105,11 @@
 #  define PYBIND11_NOINLINE __attribute__ ((noinline))
 #endif
 
-#if defined(PYBIND11_CPP14)
+#if defined(__MINGW32__)
+// For unknown reasons all PYBIND11_DEPRECATED member trigger a warning when declared
+// whether it is used or not
+#  define PYBIND11_DEPRECATED(reason)
+#elif defined(PYBIND11_CPP14)
 #  define PYBIND11_DEPRECATED(reason) [[deprecated(reason)]]
 #else
 #  define PYBIND11_DEPRECATED(reason) __attribute__((deprecated(reason)))
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -327,7 +327,7 @@ PYBIND11_NAMESPACE_END(detail)
 /// thrown to propagate python-side errors back through C++ which can either be caught manually or
 /// else falls back to the function dispatcher (which then raises the captured error back to
 /// python).
-class PYBIND11_EXPORT error_already_set : public std::runtime_error {
+class PYBIND11_EXPORT_EXCEPTION error_already_set : public std::runtime_error {
 public:
     /// Constructs a new exception from the current Python error indicator, if any.  The current
     /// Python error indicator will be cleared.
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -59,8 +59,7 @@ except ImportError:
 import distutils.errors
 import distutils.ccompiler
 
-
-WIN = sys.platform.startswith("win32") and sysconfig.get_platform() != "mingw"
+WIN = sys.platform.startswith("win32") and "mingw" not in sysconfig.get_platform()
 PY2 = sys.version_info[0] < 3
 MACOS = sys.platform.startswith("darwin")
 STD_TMPL = "/std:c++{}" if WIN else "-std=c++{}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -859,3 +859,34 @@ jobs:
 
     - name: Run all checks
       run: cmake --build build -t check
+
+  mingw:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: msys2/setup-msys2@v2
+      with:
+        install: >-
+          mingw-w64-x86_64-gcc
+          mingw-w64-x86_64-python-pip
+          mingw-w64-x86_64-cmake
+          mingw-w64-x86_64-make
+          mingw-w64-x86_64-python-pytest
+          mingw-w64-x86_64-eigen3
+          mingw-w64-x86_64-boost
+          mingw-w64-x86_64-catch
+
+    - uses: actions/checkout@v1
+
+    - name: Configure
+      # LTO leads to many undefined reference like
+      # `pybind11::detail::function_call::function_call(pybind11::detail::function_call&&)
+      run: cmake -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF -G "MinGW Makefiles" -S . -B build
+
+    - name: Build
+      run: cmake --build build -j 2
+
+    - name: Python tests
+      run: cmake --build build --target pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -883,7 +883,7 @@ jobs:
     - name: Configure
       # LTO leads to many undefined reference like
       # `pybind11::detail::function_call::function_call(pybind11::detail::function_call&&)
-      run: cmake -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF -G "MinGW Makefiles" -S . -B build
+      run: cmake -G "MinGW Makefiles" -S . -B build
 
     - name: Build
       run: cmake --build build -j 2
--- a/tests/test_exceptions.h
+++ b/tests/test_exceptions.h
@@ -4,7 +4,7 @@
 
 // shared exceptions for cross_module_tests
 
-class PYBIND11_EXPORT shared_exception : public pybind11::builtin_exception {
+class PYBIND11_EXPORT_EXCEPTION shared_exception : public pybind11::builtin_exception {
 public:
     using builtin_exception::builtin_exception;
     explicit shared_exception() : shared_exception("") {}
--- a/tools/pybind11Common.cmake
+++ b/tools/pybind11Common.cmake
@@ -308,7 +308,7 @@ function(_pybind11_generate_lto target prefer_thin_lto)
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT APPLE)
       # Clang Gold plugin does not support -Os; append -O3 to MinSizeRel builds to override it
       set(linker_append ";$<$<CONFIG:MinSizeRel>:-O3>")
-    elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND NOT MINGW)
       set(cxx_append ";-fno-fat-lto-objects")
     endif()
 
--- a/tools/pybind11Common.cmake
+++ b/tools/pybind11Common.cmake
@@ -302,6 +302,11 @@ function(_pybind11_return_if_cxx_and_linker_flags_work result cxxflags linkerfla
 endfunction()
 
 function(_pybind11_generate_lto target prefer_thin_lto)
+  if(MINGW)
+    message(STATUS "${target} disabled (problems with undefined symbols for MinGW for now)")
+    return()
+  endif()
+
   if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     set(cxx_append "")
     set(linker_append "")


jeromerobert:
> A `__declspec(dllexport)` symbol cannot be hidden, it must have the default visibility (this is required by gcc). https://github.com/pybind/pybind11/commit/3ac690b88 make `builtin_exception` `__declspec(dllexport)` while the `pybind11` namespace is `visibility(hidden)`.

An example from the GHA log:

```
D:/a/pybind11/pybind11/include/pybind11/detail/common.h:735:23: error: 'dllexport' implies default visibility, but 'class pybind11::builtin_exception' has already been declared with a different visibility
  735 | class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
      |                       ^~~~~~~~~~~~~~~~~
```

Should I disable exception `dllexport` on Mingw64 or make the pybind11 namespace default visible ? Would @oraluben know ?

oraluben:
Sorry I have few idea about the issue here, I only tested against gcc on Linux (with `__attribute__`), I guess the root cause could be GCC handling `__declspec(dllexport)` and `__attribute__(visibility(hidden))` differently?

If that's the case then I think disable exception dllexport on Mingw64 (or any other gcc build on Windows) is a workaround good enough.

pybind11/setup_helpers.py
```diff
@@ -59,8 +59,7 @@
 import distutils.errors
 import distutils.ccompiler
 
-
-WIN = sys.platform.startswith("win32") and sysconfig.get_platform() != "mingw"
+WIN = sys.platform.startswith("win32") and "mingw" not in sysconfig.get_platform()
```
henryiii, line 62: Mind pulling out _just_ this change in a PR? I'd like to make a patch release soon and it looks like the other parts might not be solved in time, but this is a clear fix for something we list in the release nodes for 2.7.0.

henryiii: Nevermind. Looks like this needs to be fixed first.

henryiii:
I think we should handle it in pybind11Common, when we add the LTO flags. I"m trying just scaling it back a little first.

henryiii:
Eventually, we might be able to come up with the correct LTO flags for MinGW, but for now, I think this is a clear improvement.

include/pybind11/detail/common.h
```diff
@@ -735,6 +735,12 @@ using expand_side_effects = bool[];
 
 PYBIND11_NAMESPACE_END(detail)
 
+#ifdef __MINGW32__
+#  pragma push_macro("PYBIND11_EXPORT")
```
rwgk, line 739: This really worries me: it's the kind of thing that's easily overlooked, highly surprising, can be extremely confusing when this hits in the middle of complex bigger changes (refactoring), and likely to cost hours to finally realize and work around. But I want to be pragmatic:

1. Is there a way this could be avoided, for example by being smarter in the primary definition of PYBIND11_EXPORT?

2. What about something like PYBIND11_EXPORT_ALL and PYBIND11_EXPORT_SOME (with a more meaningful name)?

3. If there is really no other way, could you please add a comment to explain in detail why this is needed, what is the scope of code where this needs to be blank for MINGW, and what is the code that needs the "original" define?

include/pybind11/pytypes.h
```diff
@@ -319,6 +319,12 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 inline std::string error_string();
 PYBIND11_NAMESPACE_END(detail)
 
+#ifdef __MINGW32__
```
rwgk, line 322: Having to repeat this worries me even more: when we refactor, will we have to create even more copies? How do we ensure the copies stay in sync?

include/pybind11/detail/common.h
henryiii: PYBIND11_EXPORT_EXCEPTION ?

henryiii: PYBIND11_EXPORT_STDLIB ?

rwgk: Yeah, that sounds like a good direction!
Looking systematically:
```
include/pybind11/pytypes.h:class PYBIND11_EXPORT error_already_set : public std::runtime_error {
include/pybind11/detail/common.h:#if !defined(PYBIND11_EXPORT)
include/pybind11/detail/common.h:#    define PYBIND11_EXPORT __declspec(dllexport)
include/pybind11/detail/common.h:#    define PYBIND11_EXPORT __attribute__ ((visibility("default")))
include/pybind11/detail/common.h:    extern "C" PYBIND11_MAYBE_UNUSED PYBIND11_EXPORT PyObject *PyInit_##name(); \
include/pybind11/detail/common.h:    extern "C" PYBIND11_EXPORT PyObject *PyInit_##name()
include/pybind11/detail/common.h:    extern "C" PYBIND11_MAYBE_UNUSED PYBIND11_EXPORT void init##name(); \
include/pybind11/detail/common.h:    extern "C" PYBIND11_EXPORT void init##name() {                      \
include/pybind11/detail/common.h:class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
include/pybind11/detail/common.h:    class PYBIND11_EXPORT name : public builtin_exception { public: \
docs/advanced/misc.rst:    class PYBIND11_EXPORT Dog : public Animal {
CMakeLists.txt:  if(NOT PYBIND11_EXPORT_NAME)
CMakeLists.txt:    set(PYBIND11_EXPORT_NAME "${PROJECT_NAME}Targets")
CMakeLists.txt:  install(TARGETS pybind11_headers EXPORT "${PYBIND11_EXPORT_NAME}")
CMakeLists.txt:    EXPORT "${PYBIND11_EXPORT_NAME}"
tests/test_cmake_build/subdirectory_embed/CMakeLists.txt:set(PYBIND11_EXPORT_NAME test_export)
tests/test_cmake_build/subdirectory_embed/CMakeLists.txt:# Test custom export group -- PYBIND11_EXPORT_NAME
tests/test_exceptions.h:class PYBIND11_EXPORT shared_exception : public pybind11::builtin_exception {
tests/cross_module_gil_utils.cpp:extern "C" PYBIND11_EXPORT
```

Trying to understand:

* Are the hits in CMakeLists.txt unrelated to the C++ macro?
* `PYBIND11_EXPORT` is documented, probably it will need some care?
* There aren't that many hits overall, cleanly & systematically separating them by stdlib, exceptions, (something else?) sounds like a great move.

But I also wouldn't want to block the MINGW support on making this too big a project.

henryiii: From what is mentioned above, it's rather bad to export something that's in the pybind11:: namespace. Actually, what I would have though from looking around, it's that we are exporting a class that inherits from something that is not exported, which is usually disallowed, but is "safe" if that something is a standard library object.

Regardless, exporting this is "okay" in a very special situations, but MinGW doesn't understand this. So for the exception only, we need to avoid exporting this for now (to restore 2.6 behavior on MinGW). So PYBIND11_EXPORT would remain the same, but PYBIND11_EXPORT_EXCEPTION would only export on non-MinGW.

rwgk: I don't understand all the nuances, but IIUC the upshot is that we'd have `PYBIND11_EXPORT` as before, `PYBIND11_EXPORT_EXCEPTION` in addition, each only defined once in one location, then we replace a few `PYBIND11_EXPORT` with `PYBIND11_EXPORT_EXCEPTION`? If that's true: looks good to me.

include/pybind11/detail/common.h
```diff
@@ -89,6 +89,16 @@
 #  endif
 #endif
 
+// Exceptions can't be marked with __declspec(dllexport) in MinGW
+#if !defined(PYBIND11_EXPORT)
```
rwgk, line 93: #if !defined(PYBIND11_EXPORT_EXCEPTION)
?

include/pybind11/detail/common.h
```diff
@@ -751,7 +761,7 @@ class PYBIND11_EXPORT builtin_exception : public std::runtime_error {
 #endif
 
 #define PYBIND11_RUNTIME_EXCEPTION(name, type) \
-    class PYBIND11_EXPORT name : public builtin_exception { public: \
+    class PYBIND11_EXPORT_EXCEPTION name : public builtin_exception { public: \
```
rwgk, line 764: Does line 753 (above) also need to be changed?

Similarly, line 7 in tests/test_exceptions.h?

rwgk (commented):
I looked at the logs ... a ton of errors, wow.

jeromerobert:
Oops it looks like a pushed `PYBIND11_EXPORT_EXCEPTION` without having read the last hour discussions.

jeromerobert:
Fortunately this seems to be what you expect.

> I looked at the logs ... a ton of errors, wow.

You mean tons of warnings ? Yes, I don't understand why (and where) `object(handle h, bool is_borrowed)` is called.

include/pybind11/detail/common.h
jeromerobert: Yes they need.

rwgk:
> Fortunately this seems to be what you expect.
> 
> > I looked at the logs ... a ton of errors, wow.
> 
> You mean tons of warnings ?

On all (or at least most) other platforms we're using `-Werror` or equivalent.

> Yes, I don't understand why (and where) `object(handle h, bool is_borrowed)` is called.

Do you have an interactive setup? A trick I often use is to add some extra arg(s) to the function, then build with `--keep-going`, it'll show all callers. Doesn't always work, but this looks like it could.

```diff
-object(handle h, bool is_borrowed)
+object(handle h, bool is_borrowed, const std::string &)
```

jeromerobert:
I tried to apply the following patch which remove all deprecated member. All deprecation warnings disappeared and I get no error.

```patch
diff --git a/include/pybind11/attr.h b/include/pybind11/attr.h
index 20d119f..b65ecd8 100644
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -41,8 +41,6 @@ struct sibling { handle value; sibling(const handle &value) : value(value.ptr())
 /// Annotation indicating that a class derives from another given type
 template <typename T> struct base {

-    PYBIND11_DEPRECATED("base<T>() was deprecated in favor of specifying 'T' as a template argument to class_")
-    base() { } // NOLINT(modernize-use-equals-default): breaks MSVC 2015 when adding an attribute
 };

 /// Keep patient alive while nurse lives
@@ -61,10 +59,6 @@ struct buffer_protocol { };
 struct metaclass {
     handle value;

-    PYBIND11_DEPRECATED("py::metaclass() is no longer required. It's turned on by default now.")
-    // NOLINTNEXTLINE(modernize-use-equals-default): breaks MSVC 2015 when adding an attribute
-    metaclass() {}
-
     /// Override pybind11's default metaclass
     explicit metaclass(handle value) : value(value) { }
 };
index 139a411..a04f9aa 100644
@@ -984,16 +984,6 @@ class module_ : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(module_, object, PyModule_Check)

-    /// Create a new top-level Python module with the given name and docstring
-    PYBIND11_DEPRECATED("Use PYBIND11_MODULE or module_::create_extension_module instead")
-    explicit module_(const char *name, const char *doc = nullptr) {
-#if PY_MAJOR_VERSION >= 3
-        *this = create_extension_module(name, doc, new PyModuleDef());
-#else
-        *this = create_extension_module(name, doc, nullptr);
-#endif
-    }
-
     /** \rst
         Create Python binding for a new function within the module scope. ``Func``
         can be a plain C++ function, a function pointer, or a lambda function. For
index 4cf606e..3fa36e4 100644
@@ -232,8 +232,6 @@ protected:
 class object : public handle {
 public:
     object() = default;
-    PYBIND11_DEPRECATED("Use reinterpret_borrow<object>() or reinterpret_steal<object>()")
-    object(handle h, bool is_borrowed) : handle(h) { if (is_borrowed) inc_ref(); }
     /// Copy constructor; always increases the reference count
     object(const object &o) : handle(o) { inc_ref(); }
     /// Move constructor; steals the object from ``other`` and preserves its reference count
@@ -854,8 +852,6 @@ PYBIND11_NAMESPACE_END(detail)
 //       isn't even used.
 #define PYBIND11_OBJECT_COMMON(Name, Parent, CheckFun) \
     public: \
-        PYBIND11_DEPRECATED("Use reinterpret_borrow<"#Name">() or reinterpret_steal<"#Name">()") \
-        Name(handle h, bool is_borrowed) : Parent(is_borrowed ? Parent(h, borrowed_t{}) : Parent(h, stolen_t{})) { } \
         Name(handle h, borrowed_t) : Parent(h, borrowed_t{}) { } \
         Name(handle h, stolen_t) : Parent(h, stolen_t{}) { } \
         PYBIND11_DEPRECATED("Use py::isinstance<py::python_type>(obj) instead") \
@@ -1292,8 +1288,6 @@ public:
 class capsule : public object {
 public:
     PYBIND11_OBJECT_DEFAULT(capsule, object, PyCapsule_CheckExact)
-    PYBIND11_DEPRECATED("Use reinterpret_borrow<capsule>() or reinterpret_steal<capsule>()")
-    capsule(PyObject *ptr, bool is_borrowed) : object(is_borrowed ? object(ptr, borrowed_t{}) : object(ptr, stolen_t{})) { }

     explicit capsule(const void *value, const char *name = nullptr, void (*destructor)(PyObject *) = nullptr)
         : object(PyCapsule_New(const_cast<void *>(value), name, destructor), stolen_t{}) {
@@ -1301,13 +1295,6 @@ public:
             pybind11_fail("Could not allocate capsule object!");
     }

-    PYBIND11_DEPRECATED("Please pass a destructor that takes a void pointer as input")
-    capsule(const void *value, void (*destruct)(PyObject *))
-        : object(PyCapsule_New(const_cast<void*>(value), nullptr, destruct), stolen_t{}) {
-        if (!m_ptr)
-            pybind11_fail("Could not allocate capsule object!");
-    }
-
     capsule(const void *value, void (*destructor)(void *)) {
         m_ptr = PyCapsule_New(const_cast<void *>(value), nullptr, [](PyObject *o) {
             auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
```

jeromerobert:
Here is the very first warning:

```
C:/msys2/home/jerome/pybind11/include/pybind11/pytypes.h: In constructor 'pybind11::object::object(pybind11::handle, bool)':
C:/msys2/home/jerome//pybind11/include/pybind11/pytypes.h:236:82: warning: 'pybind11::object::object(pybind11::handle, bool)' is deprecated: Use reinterpret_borrow<object>() or reinterpret_steal<object>() [-Wdeprecated
-declarations]
  236 |     object(handle h, bool is_borrowed) : handle(h) { if (is_borrowed) inc_ref(); }
      |                                                                                  ^
C:/msys2/home/jerome/pybind11/include/pybind11/pytypes.h:236:5: note: declared here
  236 |     object(handle h, bool is_borrowed) : handle(h) { if (is_borrowed) inc_ref(); }
      |     ^~~~~~
```

It looks that a warning is triggered each time a deprecated member is declared whether it is used or not. That does not make sens. Also, I cannot reproduce that on simple example.

include/pybind11/detail/common.h
henryiii: That's what I get for doing this right before going to sleep...

rwgk:
I'd be fine with `#if !defined(__MINGW32__)` around the deprecated code.

henryiii:
> It looks that a warning is triggered each time a deprecated member is declared whether it is used or not. That does not make sense.

I notices this, and don't know why it's behaving this way.

> I'd be fine with #if !defined(__MINGW32__) around the deprecated code.

As long as you get the tests to pass, I'd be okay with that too. Though this seems like a weird bug, I wonder if there's a flag we are passing that's affecting this. @jeromerobert have you compared the compile commands we use with your simple example?

jeromerobert:
> I'd be fine with #if !defined(__MINGW32__) around the deprecated code.

I disabled PYBIND11_DEPRECATED on mingw32

> have you compared the compile commands we use with your simple example

Yes. I copied them from the verbose make log.

henryiii:
> I disabled PYBIND11_DEPRECATED on mingw32

Actually that's the simplest and obvious solution. :)

rwgk (approved):

henryiii (approved):
Poke me if I don't notice when this finishes, and I'll merge & backport to the v2.7 branch.

jeromerobert:
All green !

PR #3135, opened 2021-07-20
Removing __INTEL_COMPILER section from pragma block at the top of pybind11.h

rwgk:

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -15,7 +15,6 @@
 //
 #if defined(__INTEL_COMPILER)
 #  pragma warning push
-#  pragma warning disable 878   // incompatible exception specifications
 #  pragma warning disable 2196  // warning #2196: routine is both "inline" and "noinline"
 #elif defined(_MSC_VER)
 #  pragma warning(push)
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -10,6 +10,9 @@
 
 #pragma once
 
+//
+// THE `push` HERE NEED TO BE KEPT IN SYNC WITH THE CORRESPONDING `pop` AT THE BOTTOM OF THIS FILE.
+//
 #if defined(__INTEL_COMPILER)
 #  pragma warning push
 #  pragma warning disable 878   // incompatible exception specifications
@@ -2376,8 +2379,13 @@ inline function get_overload(const T *this_ptr, const char *name) {
 
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
 
-#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
-#  pragma warning(pop)
+//
+// THE `pop` HERE NEED TO BE KEPT IN SYNC WITH THE CORRESPONDING `push` AT THE TOP OF THIS FILE.
+//
+#if defined(__INTEL_COMPILER)
+#    pragma warning pop
+#elif defined(_MSC_VER)
+#    pragma warning(pop)
 #elif defined(__GNUG__) && !defined(__clang__) && !defined(__INTEL_COMPILER)
-#  pragma GCC diagnostic pop
+#    pragma GCC diagnostic pop
 #endif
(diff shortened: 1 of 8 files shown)


tests/CMakeLists.txt
```diff
@@ -310,8 +310,13 @@ function(pybind11_enable_warnings target_name)
...
           -Werror-all
-          # "Inlining inhibited by limit max-size", "Inlining inhibited by limit max-total-size"
-          -diag-disable 11074,11076)
+          # 2196: routine is both "inline" and "noinline" (no way to `pop` and not get the warning:
+          #       https://community.intel.com/t5/Intel-C-Compiler/Inline-and-no-inline-warning/td-p/1216764)
+          # 11074: Inlining inhibited by limit max-size
+          # 11074: Inlining inhibited by limit max-total-size
+          # 11076: To get full report use -qopt-report=4 -qopt-report-phase ipo
+          -diag-disable
+          2196,11074,11076)
```
henryiii, line 319: This should be added by the main CMakeLists to the correct target, not in tests. Someone using CMake (at least) should get the warnings added. ("This" being the inline/noinline one, not the max-size, as those can be handled by users)

rwgk: Could you please guide me how?

* "the main CMakeLists"  —  is that https://github.com/pybind/pybind11/blob/master/CMakeLists.txt?
* "correct target"  — I looked through top to bottom but didn't see how to find the "correct target". Somewhere around line 184? (I'm completely lost TBH.)

I have this: https://gist.github.com/rwgk/9c4305ccc492ad78654f842f58f36b09

It shows exactly which header files trigger `#2196`. Is that relevant here?

henryiii: Here, actually: https://github.com/pybind/pybind11/blob/master/tools/pybind11Common.cmake

I wish the names of the targets were a bit more general - windows helper, etc. are too specific to add this too. Maybe we need an intel_helper?

rwgk: > Here, actually: https://github.com/pybind/pybind11/blob/master/tools/pybind11Common.cmake
> 
> I wish the names of the targets were a bit more general - windows helper, etc. are too specific to add this too. Maybe we need an intel_helper?

Gulp ... will it be simpler like e.g.
```
set(clang_4plus
    "$<AND:$<CXX_COMPILER_ID:Clang>,$<NOT:$<VERSION_LESS:$<CXX_COMPILER_VERSION>,3.9>>>")
set(no_register "$<OR:${clang_4plus},$<CXX_COMPILER_ID:AppleClang>>")
```
?
If you could help out with this part that would be really awesome.

rwgk: Wait ... thinking about it ...

> This should be added by the main CMakeLists to the correct target, not in tests.

If that's what we want, we could just add this in common.h and be done with it:

```
#if defined(__INTEL_COMPILER)
// This pragma cannot be pop'ed:
// https://community.intel.com/t5/Intel-C-Compiler/Inline-and-no-inline-warning/td-p/1216764
#  pragma warning disable 2196  // warning #2196: routine is both "inline" and "noinline"
#endif
```

I'll make that revision asap.

rwgk:
Deflaking (test_iostream.py).

rwgk:
Hi @henryiii and @Skylion007, please take another look. It's really straightforward now.

Skylion007 (approved):

include/pybind11/pybind11.h
```diff
@@ -2376,6 +2375,9 @@ inline function get_overload(const T *this_ptr, const char *name) {
 
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
 
+//
+// THE `pop` HERE NEED TO BE KEPT IN SYNC WITH THE CORRESPONDING `push` AT THE TOP OF THIS FILE.
```
henryiii, line 2379:
```suggestion
// THE `pop` HERE NEEDS TO BE KEPT IN SYNC WITH THE CORRESPONDING `push` AT THE TOP OF THIS FILE.
```

IMO "push" implies "pop", otherwise you wouldn't bother to "push" in the first place, but not horribly against this note if you want it.

henryiii (approved):

include/pybind11/pybind11.h
rwgk: I removed the notes completely. Hopefully a few PRs down the road this block will be gone entirely!

Note that I started out wanting to fix the missing pop for Intel, but in the meantime the direction of this PR changed a lot and the notes don't make that much sense anymore.

rwgk:
Thanks for the reviews :-)

PR #3137, opened 2021-07-20
Specified encoding in setup.py calls of open()

ka-bo:
## Description

The files of the project are encoded in UTF-8, yet the opening of files in the `setup.py` does not specify this encoding. When trying to install on a system with a non-UTF-8 locale using `pip install pybind11/.` the process fails with the the following error:

```
Processing /pybind11
    Complete output from command python setup.py egg_info:
      File "/tmp/pip-mi6rymcm-build/setup.py", line 74, in <module>
        matches = dict(VERSION_REGEX.findall(f.read()))
      File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 16164: ordinal not in range(128)
```

Specifying that the files should be opened using UTF-8 encoding fixes the problem.

## Suggested changelog entry:

```rst
-  Specified UTF8-encoding in setup.py calls of open()
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ import string
 import subprocess
 import sys
 import tempfile
+import io
 
 import setuptools.command.sdist
 
@@ -63,14 +64,14 @@ to_src = (
 )
 
 # Read the listed version
-with open("pybind11/_version.py", encoding="utf8") as f:
+with open("pybind11/_version.py") as f:
     code = compile(f.read(), "pybind11/_version.py", "exec")
 loc = {}
 exec(code, loc)
 version = loc["__version__"]
 
 # Verify that the version matches the one in C++
-with open("include/pybind11/detail/common.h", encoding="utf8") as f:
+with io.open("include/pybind11/detail/common.h", encoding="utf8") as f:
     matches = dict(VERSION_REGEX.findall(f.read()))
 cpp_version = "{MAJOR}.{MINOR}.{PATCH}".format(**matches)
 if version != cpp_version:
--- a/setup.py
+++ b/setup.py
@@ -63,14 +63,14 @@ to_src = (
 )
 
 # Read the listed version
-with open("pybind11/_version.py") as f:
+with open("pybind11/_version.py", encoding="utf8") as f:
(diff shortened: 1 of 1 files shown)


Skylion007 (changes requested):
This isn't the right syntax for Python2 @ka-bo

setup.py
```diff
@@ -63,14 +63,14 @@ def build_expected_version_hex(matches):
 )
 
 # Read the listed version
-with open("pybind11/_version.py") as f:
+with open("pybind11/_version.py", encoding="utf8") as f:
```
Skylion007, line 66: This needs to be io.open for Python2

ka-bo: Unfortunately, that gives `SyntaxError: encoding declaration in Unicode string` in python2 package building.

ka-bo: Something like this seems to work, but is a somewhat ugly solution, i think:

```python
encoding_for_py3 = {"encoding" : "utf8"} if sys.version_info.major >= 3 else {}
with open("pybind11/_version.py", **encoding_for_py3) as f:
```

henryiii: Are you sure you need encoding here? I think it's correct below, but this one is being exec'd, so it can pick up the encoding from the file. Maybe it could even be opened explicitly in binary?

setup.py
```diff
@@ -63,14 +63,14 @@ def build_expected_version_hex(matches):
...
     code = compile(f.read(), "pybind11/_version.py", "exec")
 loc = {}
 exec(code, loc)
 version = loc["__version__"]
 
 # Verify that the version matches the one in C++
-with open("include/pybind11/detail/common.h") as f:
+with open("include/pybind11/detail/common.h", encoding="utf8") as f:
```
henryiii, line 73: This one should be safe to io.open.

ka-bo:
Thanks for the hint @henryiii, exactly that exec seemed to be the problem.

The updated version should work now.

Skylion007 (approved):

PR #3141, opened 2021-07-25
Removing MSVC C4800 from pragma block at the top of pybind11.h

rwgk:

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -322,7 +322,7 @@ public:
             }
             #endif
             if (res == 0 || res == 1) {
-                value = PYBIND11_COMPAT_BOOL_CAST(res);
+                value = (res != 0);
                 return true;
             }
             PyErr_Clear();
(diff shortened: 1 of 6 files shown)


include/pybind11/cast.h
```diff
@@ -322,7 +322,7 @@ template <> class type_caster<bool> {
             }
             #endif
             if (res == 0 || res == 1) {
-                value = (bool) res;
```
henryiii, line 325: Why not just make these `value = res != 0`; ? The problem is that (bool) int_value is poor for performance (according to the warning), so we shouldn't be pretending this is okay to cast.

Skylion007:
Actually never mind, I can add that in a separate PR as that check is much more strict than this one even is and will touch a lot more code.

Skylion007 (approved):

rwgk:
> Actually never mind, I can add that in a separate PR as that check is much more strict than this one even is and will touch a lot more code.

Awesome, I'm glad. Combining (convoluting) things backfires too often. Best to keep things compartmentalized.

include/pybind11/cast.h
rwgk: Done. Applied globally. — I was aiming for a robust general solution before, but you're right, we only need specific solutions, and `!= 0` is simpler therefore better.

PR #3142, opened 2021-07-26
Removing MSVC C4522 from pragma block at the top of pybind11.h

rwgk:


henryiii (approved):
This is a weird warning...

rwgk:
> This is a weird warning...

Yeah, the fact that it went away with MSVC 2019 makes me think the warning was more of a bug than a feature.

Thanks Henry!

PR #3144, opened 2021-07-26
fix: the Python package had a hard coded path

### LINKED ISSUE #3136: [BUG] Incorrect cmake install path in 2.7.0 release?
I have a local requirements.txt file that just uses the latest pybind11 and when I rebuilt my environment, I got the 2.7.0 release, which looks new. However, CMake is detecting include and lib paths that look suspiciously like they are from the VM that built it, not the actual install location on my machine. Example CMake output:

```
-- found (/home/stella/src/iree-forge/build/iree-forge.venv/lib/python3.8/site-packages/pybind11/share/cmake/pybind11)
-- Found pybind11: /home/runner/work/pybind11/pybind11/pybind11/include (found version "2.7.0" )
-- Found pybind11 v2.7.0: /home/runner/work/pybind11/pybind11/pybind11/include
```

This was found while building the LLVM/MLIR python bindings from head with recommended instructions: https://mlir.llvm.org/docs/Bindings/Python/ (but I imagine there are easier ways to repro). I'm about to push a commit that excludes that version from our dev setup.

henryiii:
## Description

Fix #3136 - regression in #3005.

## Suggested changelog entry:

```rst
* Fix regression in CMake Python package config: improper use of absolute path.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,6 +203,12 @@ if(PYBIND11_INSTALL)
       "${CMAKE_INSTALL_DATAROOTDIR}/cmake/${PROJECT_NAME}"
       CACHE STRING "install path for pybind11Config.cmake")
 
+  if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(pybind11_INCLUDEDIR "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
+  else()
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

rwgk (approved):
Pro-forma LGTM (my cmake knowledge is extremely superficial).

The suggested changelog entry needs more context (and has a typo). Maybe:
< * Fix abolute path in Python package.
> * Fix regression in cmake Python package config: improper use of absolute path.

"cmake" needs to appear somewhere.

PR #3150, opened 2021-07-28
Removing MSVC C4100 from pragma block at the top of pybind11.h

rwgk:

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -516,18 +516,22 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
 /// Recursively iterate over variadic template arguments
 template <typename... Args> struct process_attributes {
     static void init(const Args&... args, function_record *r) {
+        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
         ignore_unused(unused);
     }
     static void init(const Args&... args, type_record *r) {
+        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
         ignore_unused(unused);
     }
     static void precall(function_call &call) {
+        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call);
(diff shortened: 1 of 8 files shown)


include/pybind11/detail/common.h
```diff
@@ -913,5 +913,15 @@ inline static std::shared_ptr<T> try_get_shared_from_this(std::enable_shared_fro
 #endif
 }
 
+#if defined(_MSC_VER) && _MSC_VER <= 1916
+// warning C4100: Unreferenced formal parameter
+template <typename... Args>
+inline constexpr void workaround_incorrect_msvc_c4100(Args &&...) {}
+#    define PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(...)                                         \
+        detail::workaround_incorrect_msvc_c4100(__VA_ARGS__)
```
henryiii, line 921: The structure here is very visually confusing to me. If this was on a single line it would be easier to read, or if it had a space above it, and/or if the define was not indented, or if the template was indented.

I think its largely due to the double-indent of the body but no indent on the define above.

include/pybind11/attr.h
```diff
@@ -545,6 +549,7 @@ template <typename... Extra,
           size_t named = constexpr_sum(std::is_base_of<arg, Extra>::value...),
           size_t self  = constexpr_sum(std::is_same<is_method, Extra>::value...)>
 constexpr bool expected_num_args(size_t nargs, bool has_args, bool has_kwargs) {
+    PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(nargs, has_args, has_kwargs);
```
henryiii, line 552: This is a really buggy warning, I guess? Because the parameters are literally used on the next line.

rwgk: > This is a really buggy warning, I guess?

Absolutely. We're working around a bug in old MSVC versions.

include/pybind11/detail/common.h
rwgk: It's clang-format'ed.
I could make it different now, but later clang-format would change it straight back to this.
Do you want me to change it anyway?

henryiii: Adding a blank line or two won't be removed, and that would help.

rwgk: Please take another look. I added three, thinking that's most clear.
I also ran it through clang-format-diff again, which left it like what you see.

henryiii (approved):

rwgk:
Thanks Henry! (Only two more MSVC pragmas to go.)

PR #3151, opened 2021-07-28
Always check __STDC_LIB_EXT1__ before __STDC_WANT_LIB_EXT1__

rwgk:
Accommodating environments that define `__STDC_WANT_LIB_EXT1__` even if `__STDC_LIB_EXT1__` is not defined by the implementation.

Follow-on to PR #3129.

Fixes Google-internal macos builds, but this is generally more correct.

Changelog not needed: (PR #3129 was not released yet.


henryiii (approved):

rwgk:
Thanks Henry!

PR #3152, opened 2021-07-29
Removing MSVC C4127 from pragma block at the top of pybind11.h

rwgk:

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -941,8 +941,16 @@ inline constexpr void workaround_incorrect_msvc_c4100(Args &&...) {}
 #    define PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(...)
 #endif
 
-// Suppresses MSVC warning C4127: Conditional expression is constant
-constexpr inline bool constexpr_bool(bool cond) { return cond; }
+#if defined(_MSC_VER) // All versions (as of July 2021).
+
+// warning C4127: Conditional expression is constant
+constexpr inline bool silence_msvc_c4127(bool cond) { return cond; }
+
+#    define PYBIND11_SILENCE_MSVC_C4127(...) detail::silence_msvc_c4127(__VA_ARGS__)
+
+#else
+#    define PYBIND11_SILENCE_MSVC_C4127(...) __VA_ARGS__
+#endif
 
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -941,5 +941,8 @@ inline constexpr void workaround_incorrect_msvc_c4100(Args &&...) {}
 #    define PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(...)
 #endif
 
+// Suppresses MSVC warning C4127: Conditional expression is constant
+constexpr inline bool constexpr_bool(bool cond) { return cond; }
+
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -778,7 +778,7 @@ struct pyobject_caster {
         // For Python 2, without this implicit conversion, Python code would
         // need to be cluttered with six.ensure_text() or similar, only to be
         // un-cluttered later after Python 2 support is dropped.
-        if (std::is_same<T, str>::value && isinstance<bytes>(src)) {
+        if (constexpr_bool(std::is_same<T, str>::value) && isinstance<bytes>(src)) {
             PyObject *str_from_bytes = PyUnicode_FromEncodedObject(src.ptr(), "utf-8", nullptr);
             if (!str_from_bytes) throw error_already_set();
             value = reinterpret_steal<type>(str_from_bytes);
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -384,7 +384,11 @@ template <typename StringType, bool IsView = false> struct string_caster {
 
         const auto *buffer = reinterpret_cast<const CharT *>(PYBIND11_BYTES_AS_STRING(utfNbytes.ptr()));
         size_t length = (size_t) PYBIND11_BYTES_SIZE(utfNbytes.ptr()) / sizeof(CharT);
-        if (constexpr_bool(UTF_N > 8)) { buffer++; length--; } // Skip BOM for UTF-16/32
+        // Skip BOM for UTF-16/32
+        if (PYBIND11_SILENCE_MSVC_C4127(UTF_N > 8)) {
+            buffer++;
+            length--;
+        }
         value = StringType(buffer, length);
 
         // If we're loading a string_view we need to keep the encoded Python object alive:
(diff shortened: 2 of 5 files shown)


include/pybind11/detail/common.h
```diff
@@ -926,5 +926,8 @@ inline constexpr void workaround_incorrect_msvc_c4100(Args &&...) {}
 #    define PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(...)
 #endif
 
+// Suppresses MSVC warning C4127: Conditional expression is constant
+constexpr inline bool constexpr_bool(bool cond) { return cond; }
```
henryiii, line 930: What about something like:

```cpp
// if C++17+
#define PYBIND11_CONSTEXPR_IF(arg) if constexpr(arg)
// else
constexpr inline bool constexpr_bool(bool cond) { return cond; }
#define PYBIND11_CONSTEXPR_IF(arg) pybind11::detail::constexpr_bool(arg)
```

Skylion007: Just copy this machinery from this experimental PR: https://github.com/pybind/pybind11/pull/3113

henryiii: Does this warning show up for C++ < 17? It really shouldn't, as there's not an easy "fix" there. But I expect it does?

henryiii: That wouldn't fix the warning if this is produced in C++11 or C++14 mode.

henryiii: `if PYBIND11_CPP17_CONSTEXPR (constexpr_bool(...))` is not horrible, though, that is an option.

Skylion007: ``` constexpr inline bool constexpr_bool(bool cond) { return cond; } ``` can be the else. I don't think there will be any performance benefit in < CPP17 though? OR will the constexpr actually propogate? I don't think so right?

henryiii: The only goal of the constexpr_bool is to remove this warning. Most compilers don't warn here, since it's not easy to avoid an if with a static condition until C++17. That's why I'm not wildly excited about adding constexpr_bool; using a macro that can also utilize `if constexpr` would be an actual possible benefit.

`if constexpr(constexpr_bool(...))` will be constexpr, but there's no difference with `if constexpr(...)`, just for hiding this warning when the "constexpr" is not present. That's why I'd be much more tempted if MSVC doesn't produce this warning in C++11/14 mode, but I _think_ it does?

rwgk:
I'm outside and cannot look carefully right now, but please click on the
Worksheet link in the PR description for the full ground truth. From
memory: every MSVC ever in existence, and in any mode we use, produces that
warning.

include/pybind11/detail/common.h
henryiii: You can pick `PYBIND11_CONSTEXPR_IF` vs. `PYBIND11_IF_CONSTEXPR`, but this should help us move to using if constexpr` when available - while the warning is a little irritating, it does nicely show all the places we can use `if constexpr`. :)

rwgk: I have two concerns:

1. The `PYBIND11_EXPORT_EXCEPTION` alternative would mean that we need more involved rearrangements of the existing code. For example this one would get pretty complicated:
```
         if (constexpr_bool(StringCaster::UTF_N == 8) && str_len > 1 && str_len <= 4) {
         ...
         else if (constexpr_bool(StringCaster::UTF_N == 16) && str_len == 2) {
```

Note that the `constexpr_bool` is only for a subexpression.

2. I feel making this look too much like the C++17 `if constexpr` is actually misleading. In my understanding, the primary distinguishing feature is that the body of the `if constexpr` does not participate in template instantiation if the condition is false:

https://en.cppreference.com/w/cpp/language/if
> If a constexpr if statement appears inside a templated entity, and if condition is not value-dependent after instantiation, the discarded statement is not instantiated when the enclosing template is instantiated .
 
It's actually a very big deal, because it enables much more intuitive and easier to write code than the alternative SFINAE acrobatics we have to resort to before C++17. But clearly, for the existing code that triggers C4127, that distinguishing feature doesn't apply. It just so happens that the condition is a compile-time constant.

> while the warning is a little irritating

I think what is a hack is the implementation behind the warning. I was pretty happy about the minimally obtrusive solution I found. But I'll try something else as soon as I get a chance, more similar to the C4100 approach:

`PYBIND11_SILENCE_MSVC_C4127`

Then it is totally clear why exactly it's there, without getting into the fairly complex territory of `constexpr` semantics.

henryiii: `if constexpr` also avoids checking the condition at runtime, and ensures that a runtime "if" is not placed in the code. So if it reduces to `if constexpr(True)` then the else block is not included in the compilation. While `if(True)` still compiles the else. So you get a small benefit. We likely will not be dropping C++14 (or even 11, possibly) support for a long time, so being able to simplify this would be a nice side benefit. MSVC is correctly indicating this should be `if constexpr` with the warning; we just can't require it. It's actually really helpful, because it tells us exactly what if statements are constant without having to read the code.

The compiler _may_ turn a runtime if into a compile time `if` if the expression is simple enough. In fact, MSVC doesn't warn if there's a simple "true" or similar in the statement. For example, this is a compile time if on all GCC optimization levels:

```cpp
int check(int j) {
    if (true)
      return 2*j;
    else
      return 3*j;
}
```

However, this is _not_ a compile time if on GCC if optimization is off (`-O0`):

```cpp
constexpr bool f(bool x) {return x;}

int check(int j) {
    if (f(true))
      return 2*j;
    else
      return 3*j;
}
```

This is, even at `-O0`:

```cpp
constexpr bool f(bool x) {return x;}

int check(int j) {
    if constexpr (f(true))
      return 2*j;
    else
      return 3*j;
}
```

See https://gcc.godbolt.org/z/WhzndzTTx

> It's actually a very big deal

IIRC, it actually doesn't work in quite as many places as I would have wanted it too, but yes, it's very, very useful for this. Dramatically simplifies SFINAE (most of the time).

Given that this _usually_ optimizes to the same thing, I'm fine with not adding the constexprs.

henryiii: (Though adding PYBIND11_CPP17_CONSTEXPR on all these in a followup might be nice)

henryiii: > Note that the constexpr_bool is only for a subexpression.

You could just as easily apply it to the whole expression. The other part is constant too (it has to be to trigger the warning).

henryiii: No, I take that back - why would that trigger the warning then? A constant expression combined with a non-constant one triggers the warning?

rwgk: > No, I take that back - why would that trigger the warning then? A constant expression combined with a non-constant one triggers the warning?

I think what happened: sometime in the 1990s or even 80s someone implemented C4127 for C. It made total sense: the warning would tell you that you could simply remove the (sub)expression (and maybe an entire if or else branch with it). Then templates were added to the language. But nearly a quarter century later, Microsoft still hasn't updated the diagnostic to take that into account, therefore in modern template heavy code most of these warnings are false positives, irritating noise.

I don't think it's useful to convolute constexpr modernization into dealing with the antiquated warning, beyond the coincidence-to-some-degree that a constexpr function can be used as a trick to silence the warning.

henryiii (approved):

rwgk:
Awesome, thanks Henry!
I'm just in the middle of rebasing: this PR will now remove the entire MSVC section of the pragma block!
One two GCC warnings left to do.

PR #3161, opened 2021-07-30
Limiting pragma for ignoring GCC 7 -Wnoexcept-type to the scope of pybind11.h.

rwgk:

Note that the pragma for `-Wnoexcept-type` is moved from above to below all `#include`s. This unblocks refactoring and IWYU cleanup (and eventually merging the smart_holder branch).

It is somewhat of a coincidence that the moved pragma is still in the same header file. Potentially, the pragma for `-Wnoexcept-type` may need to be copied to refactored header files, but that's the price we have to pay for improved code organization with the added requirement to not suppress warnings from the command line (for portability).

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -50,9 +50,7 @@
    It is only actually needed in a couple places, but apparently GCC 7 "generates this warning if
    and only if the first template instantiation ... involves noexcept" [stackoverflow], therefore
    it could get triggered from seemingly random places, depending on user code.
-   It seems very unlikely that this warning will be useful to anyone (no other GCC version
-   generates it), but to be maximally accommodating we turn it off here instead of using a
-   command line option.
+   No other GCC version generates this warning.
  */
 #if defined(__GNUC__) && __GNUC__ == 7
 #    pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


include/pybind11/pybind11.h
```diff
@@ -48,6 +45,20 @@
...
 #endif
 
+/* https://stackoverflow.com/questions/46798456/handling-gccs-noexcept-type-warning
+   This warning is about ABI compatibility, not code health.
+   It is only actually needed in a couple places, but apparently GCC 7 "generates this warning if
+   and only if the first template instantiation ... involves noexcept" [stackoverflow], therefore
+   it could get triggered from seemingly random places, depending on user code.
+   It seems very unlikely that this warning will be useful to anyone (no other GCC version
+   generates it), but to be maximally accommodating we turn it off here instead of using a
+   command line option.
```
henryiii, line 55: >  other GCC version generates it
🤣 

```suggestion
   No other GCC version generates this warning.
```

Too much information, we want pybind11 to compile out of the box with no warnings if possible, and that's why we carefully remove a few warnings, including this one.

henryiii (approved):

include/pybind11/pybind11.h
rwgk: Change applied. Thanks!

rwgk:
Thanks Henry!

The one failing CI job is a test_iostream flake. Safe to ignore, especially because it worked in the previous CI run and the only change was to a comment.

PR #3164, opened 2021-07-31
Removing GCC -Wunused-but-set-parameter from pragma block at the top of pybind11.h

rwgk:

Consolidating the new warning suppressions for MSVC & GCC and the existing `ignore_unused()` function.

Introducing a new macro `PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER`, similar to the existing one for `MSVC`.

This PR is mostly trivial, but includes an important commit by @henryiii that is more-or-less independent (but significantly reduces the required workarounds):

https://github.com/pybind/pybind11/pull/3164/commits/218732d402c80c9ce6308e52a63e1378a067da24

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -517,13 +517,13 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
 template <typename... Args> struct process_attributes {
     static void init(const Args&... args, function_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
-        PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER(r);
+        PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
         silence_unused_warnings(unused);
     }
     static void init(const Args&... args, type_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
-        PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER(r);
+        PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
         silence_unused_warnings(unused);
     }
@@ -534,7 +534,7 @@ template <typename... Args> struct process_attributes {
     }
     static void postcall(function_call &call, handle fn_ret) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call, fn_ret);
-        PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER(fn_ret);
+        PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(fn_ret);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::postcall(call, fn_ret), 0) ... };
         silence_unused_warnings(unused);
     }
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -518,26 +518,26 @@ template <typename... Args> struct process_attributes {
     static void init(const Args&... args, function_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
         PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
-        int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        silence_unused_warnings(unused);
+        PYBIND11_INT_ARRAY_WORKING_AROUND_MSVC_CLANG_ISSUES(
+            {0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0)...});
     }
     static void init(const Args&... args, type_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
         PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
-        int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        silence_unused_warnings(unused);
+        PYBIND11_INT_ARRAY_WORKING_AROUND_MSVC_CLANG_ISSUES(
+            {0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0)...});
     }
     static void precall(function_call &call) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call);
-        // int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
-        // silence_unused_warnings(unused);
-        (int []) { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
+        PYBIND11_INT_ARRAY_WORKING_AROUND_MSVC_CLANG_ISSUES(
+            {0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0)...});
     }
     static void postcall(function_call &call, handle fn_ret) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call, fn_ret);
         PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(fn_ret);
-        int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::postcall(call, fn_ret), 0) ... };
-        silence_unused_warnings(unused);
+        PYBIND11_INT_ARRAY_WORKING_AROUND_MSVC_CLANG_ISSUES(
+            {0,
+             (process_attribute<typename std::decay<Args>::type>::postcall(call, fn_ret), 0)...});
     }
 };
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -517,23 +517,26 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
 template <typename... Args> struct process_attributes {
     static void init(const Args&... args, function_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
+        PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        ignore_unused(unused);
+        silence_unused_warnings(unused);
     }
     static void init(const Args&... args, type_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
+        PYBIND11_WORKAROUND_INCORRECT_GCC_UNUSED_BUT_SET_PARAMETER(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        ignore_unused(unused);
+        silence_unused_warnings(unused);
     }
     static void precall(function_call &call) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
-        ignore_unused(unused);
+        silence_unused_warnings(unused);
     }
     static void postcall(function_call &call, handle fn_ret) {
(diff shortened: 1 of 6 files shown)


rwgk:
For completeness, below is the list of cmake `The CXX compiler identification`s currently used in the CI workflow.

From the CI results we know for certain that GCC 4 needs the `PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER` macro, and GCC 6 does not. GCC 5 is unknown.

I tried to capture this super concisely in the comment, with total line length < 100, but assuming a person caring enough and thinking about it for a minute will be able to figure it out:

```
#    if defined(__GNUC__) && __GNUC__ <= 5 // 4 is certain, 5 unknown, 6 is certain
```

```
AppleClang 12.0.0.12000032
Clang 10.0.0
Clang 13.0.0
Clang 3.6.2
Clang 3.7.1
Clang 3.9.1
Clang 5.0.2
Clang 7.0.1
Clang 9.0.0
GNU 10.3.0
GNU 11.2.0
GNU 4.8.5
GNU 6.3.0
GNU 7.5.0
GNU 8.4.1
GNU 9.3.0
Intel 2021.3.0.20210609
MSVC 19.0.24245.0
MSVC 19.16.27045.0
MSVC 19.29.30040.0
PGI 20.9.0
```

include/pybind11/attr.h
```diff
@@ -517,23 +517,26 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
...
     static void init(const Args&... args, type_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
+        PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        ignore_unused(unused);
+        silence_unused_warnings(unused);
     }
     static void precall(function_call &call) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(call);
         int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
```
henryiii, line 532:
```suggestion
        (int []) { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
```

Why can't we just do this? I'm guessing the point of this is to "trick" the compile to get it to call all these in an expansion, but we could at least save several lines and forcing the reader to rely on what "unused" and "silence unused warnings" is.

I may not fully understand the trick here though, since I don't understand the `0,` at the start is doing. That's not part of the usual trick I've seen elsewhere.

rwgk: Just quick: that's existing code, the only reason I touched it is to use the more general new function (I'm not wedded to the function name). But I'll play with your idea to see how it pans out.

henryiii: Yes, I know, and I don't mind the name, I just think we should fix the issue rather than hack around with the hack for it. I don't think we want to create a variable here, since we don't use it; making a variable should be telling the reader you are going to do something with it.

henryiii: I naively would put `(int []) {process_attribute<typename std::decay<Args>::type>::precall(call) ... };` - we should at least comment on why this structure is needed if it is.

rwgk: I did a quick experiment:

https://github.com/pybind/pybind11/actions/runs/1086709363

I didn't look in detail, but as far as I can see all MSVC and clang builds break. Only gcc and icc don't issue warnings.

An example clang warning (my local machine) is below. Note that I don't use `-Werror` on my local machine, therefore the build succeeds and the tests run successfully. Your suggested change is clearly correct, but not compatible with MSVC and clang without some sort of warning suppression we'd need to figure out.

@henryiii It's a fair question to ask, but tangential to this PR. I'm already paying a very high price unblocking refactoring and IWYU cleanup by painstakingly nibbling away on the pragma block in pybind11.h. I need to stay focused on that. I could open a bug or another PR with your suggested change, for someone to work on separately. Please let me know if you think that's worthwhile. Note that there are only 5 lines total (4 in attr.h, 1 in cast.h) that this effort might improve, if successful.

`Debian clang version 11.0.1-21`
```
/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/detail/../attr.h:534:9: warning: expression result unused [-Wunused-value]
        (int []) { 0, (process_attribute<typename std::decay<Args>::type>::precall(call), 0) ... };
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

rwgk: I'll try to make use of what I just learned found out by introducing another macro, that does the warning suppression only for MSVC and clang.

rwgk: Done. Holding my breath that all compilers in our CI are happy with this.

It's a bit on the overkill side of the spectrum, but the comments that go with the new macro clearly document why some sort of workaround is needed.

Note that the code in the last commit is `clang-format`ed.

rwgk: Not even a flake, hooray!
@henryiii Please take another look.

include/pybind11/detail/type_caster_base.h
```diff
@@ -939,6 +940,7 @@ template <typename type> class type_caster_base : public type_caster_generic {
     template <typename T, typename = enable_if_t<std::is_move_constructible<T>::value>>
     static auto make_move_constructor(const T *x) -> decltype(new T(std::move(*const_cast<T *>(x))), Constructor{}) {
```
henryiii, line 941: "INCORRECT"

I don't see any usage of x, actually - GCC is correct, this is ugly and wrong. It is used in the type expression, but it doesn't need to be - you could just as well use `std::declval` instead.

```suggestion
    static auto make_move_constructor(const T *x) -> decltype(&std::declval<T>())), Constructor{}) {
```

include/pybind11/detail/type_caster_base.h
```diff
@@ -939,6 +940,7 @@ template <typename type> class type_caster_base : public type_caster_generic {
...
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(x);
+        PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER(x);
```
henryiii, line 943:
```suggestion
```

include/pybind11/detail/type_caster_base.h
```diff
@@ -931,6 +931,7 @@ template <typename type> class type_caster_base : public type_caster_generic {
     template <typename T, typename = enable_if_t<is_copy_constructible<T>::value>>
     static auto make_copy_constructor(const T *x) -> decltype(new T(*x), Constructor{}) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(x);
+        PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER(x);
```
henryiii, line 934:
```suggestion
    static auto make_copy_constructor(const T *x) -> decltype(&std::declval<T>()), Constructor{}) {
```

include/pybind11/attr.h
```diff
@@ -517,23 +517,27 @@ template <size_t Nurse, size_t Patient> struct process_attribute<keep_alive<Nurs
 template <typename... Args> struct process_attributes {
     static void init(const Args&... args, function_record *r) {
         PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(r);
-        int unused[] = { 0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ... };
-        ignore_unused(unused);
+        PYBIND11_WORKAROUND_INCORRECT_ALL_GCC_UNUSED_BUT_SET_PARAMETER(r);
+        PYBIND11_INT_ARRAY_WORKING_AROUND_MSVC_CLANG_ISSUES(
+            {0, (process_attribute<typename std::decay<Args>::type>::init(args, r), 0)...});
```
henryiii, line 522: Could you try this?

```suggestion
        using expander = int[];
        (void) expander{0, ((void) process_attribute<typename std::decay<Args>::type>::init(args, r), 0) ...};
```

That's how I'd expect this to look in C++11 (much nicer in C++17). The extra "void" cast is not needed (the one in front of process_attribute), since that's already void, but it's an extra protection that is commonly in this pattern (if it's not void and has a comma operator overload, it will not behave correctly). This works locally for me in Clang. If it doesn't work, then fine, we can go with what you have. :)

include/pybind11/detail/type_caster_base.h
henryiii: (And, obviously, remove the "x"  and leave the parameter unnamed and remove the lines hiding warnings)

rwgk: It builds (after removing extra parenthesis) but fails to run. See below.

I only tried both suggestions together. I'd highly appreciate if we could reserve these side battles for separate PRs. It's fine with me if someone wants to spend the time going above and beyond, but in my position it's difficult to justify spending more time on a very localized, very small enhancement of existing code, only tangentially related to the bigger problem I need to solve.

```
ImportError while loading conftest '/usr/local/google/home/rwgk/forked/pybind11/tests/conftest.py'.
conftest.py:19: in <module>
    import pybind11_tests  # noqa: F401
E   ImportError: return_value_policy = copy, but type MyEnum is non-copyable!
```

Here is the diff that led to the error above, in case I made a small mistake and someone wants to try it more:
```
diff --git a/include/pybind11/detail/type_caster_base.h b/include/pybind11/detail/type_caster_base.h
index a3ce4204..e84e8d99 100644
--- a/include/pybind11/detail/type_caster_base.h
+++ b/include/pybind11/detail/type_caster_base.h
@@ -929,18 +929,14 @@ protected:
     /* Only enabled when the types are {copy,move}-constructible *and* when the type
        does not have a private operator new implementation. */
     template <typename T, typename = enable_if_t<is_copy_constructible<T>::value>>
-    static auto make_copy_constructor(const T *x) -> decltype(new T(*x), Constructor{}) {
-        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(x);
-        PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER(x);
+    static auto make_copy_constructor(const T *) -> decltype(&std::declval<T>(), Constructor{}) {
         return [](const void *arg) -> void * {
             return new T(*reinterpret_cast<const T *>(arg));
         };
     }
 
     template <typename T, typename = enable_if_t<std::is_move_constructible<T>::value>>
-    static auto make_move_constructor(const T *x) -> decltype(new T(std::move(*const_cast<T *>(x))), Constructor{}) {
-        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(x);
-        PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER(x);
+    static auto make_move_constructor(const T *) -> decltype(&std::declval<T>(), Constructor{}) {
         return [](const void *arg) -> void * {
             return new T(std::move(*const_cast<T *>(reinterpret_cast<const T *>(arg))));
         };
```

include/pybind11/detail/type_caster_base.h
rwgk: See other comment (make_move_constructor).

include/pybind11/detail/type_caster_base.h
henryiii: The thing you are replacing is different in the two cases, why would you expect it to become the same expression? `new T(*x)` is not the same as `new T(std::move(*const_cast<T *>(x))`. If it makes you feel better, I did exactly the same mistake about 6 times in a row trying to understand why this didn't work. :) 

The good news is that by doing it so many times, I have a better idea of what's going on here and was able to simplify the expression a bit - it's easier and more natural when you can make any type you want via declval, rather than having to cast around a preset value.

henryiii: Ahh, `gh pr checkout 3164` failed, so I modified and pushed to master without noticing it. Reverted, will reset locally and apply to this branch.

henryiii: I wondered why git apply didn't work...

henryiii (approved):

include/pybind11/detail/type_caster_base.h
```diff
@@ -927,18 +927,17 @@ template <typename type> class type_caster_base : public type_caster_generic {
     using Constructor = void *(*)(const void *);
 
     /* Only enabled when the types are {copy,move}-constructible *and* when the type
-       does not have a private operator new implementation. */
+       does not have a private operator new implementation. A comma operator is used in the decltype
+       argument to apply SFINAE to the public copy/move constructors.*/
     template <typename T, typename = enable_if_t<is_copy_constructible<T>::value>>
-    static auto make_copy_constructor(const T *x) -> decltype(new T(*x), Constructor{}) {
-        PYBIND11_WORKAROUND_INCORRECT_MSVC_C4100(x);
+    static auto make_copy_constructor(const T *) -> decltype(new T(std::declval<const T>()), Constructor{}) {
```
Skylion007, line 933: This should be changed to not required a new in the return type right?

Skylion007 (approved):

henryiii:
My commit was not really independent - you added a different workaround for this warning, including a new macro `PYBIND11_WORKAROUND_INCORRECT_OLD_GCC_UNUSED_BUT_SET_PARAMETER`, I fixed the warning instead. Fixing "unused-but-set-parameter" instead of finding a new way to hide it is a perfectly valid way to "remove" a warning and it not unrelated, it fits perfectly under the title.

PR #3166, opened 2021-08-02
maint(clang-tidy): Bugprone enable checks

Skylion007:
## Description
* Enables additional clang-tidy checks for bugprone-behavior. 

## Suggested changelog entry:

```rst
* Enable most bugprone checks in clang-tidy and fix the found potential bugs and poor coding styles.
```


.clang-tidy
```diff
@@ -43,6 +44,10 @@ readability-static-accessed-through-instance,
 readability-static-definition-in-anonymous-namespace,
 readability-string-compare,
 readability-uniqueptr-delete-release,
+-bugprone-exception-escape,
```
Skylion007, line 47: We call py::print a lot in the destructors in tests which is quite noisy. Probably should nolint them all.

include/pybind11/numpy.h
```diff
@@ -1551,6 +1551,7 @@ struct vectorize_helper {
             "pybind11::vectorize(...) requires a function with at least one vectorizable argument");
 
 public:
+    // NOLINTNEXTLINE(bugprone-forwarding-reference-overload) //TODO FIX BEFORE MERGE
```
Skylion007, line 1554: @rwgk This is the prefect forwarding code that I was struggling with fixing the error for.

Skylion007: Okay, so we do need to the copy and move constructors after all... Hmm... so we need to do some SFINAE acrobatics as described in https://moodle.ufsc.br/pluginfile.php/2377667/mod_resource/content/0/Effective_Modern_C__.pdf Item 25-26 Still struggling with it though.

rwgk:
I just took a super quick look after seeing the emails about your last two commits. Does that mean `enable_if` isn't needed in this case, just deleting the copy and move constructors does the trick?

Skylion007:
> 
> 
> I just took a super quick look after seeing the emails about your last two commits. Does that mean `enable_if` isn't needed in this case, just deleting the copy and move constructors does the trick?

Unfortunately, the default copy and move destructors are actually used so I do need to do the enableif trick.

.clang-tidy
rwgk: How many nolints are needed?
Are they only under tests/?

Skylion007: Quite a few, and yes only under tests. Under virtually ever destructor or operator that calls py::print.

rwgk: I'm thinking:
* We really want that check under include/
* But minimize clutter under tests/
Could we have a tests/.clang-tidy? How much trouble (copy-paste)?

Another completely different idea, possibly very naive/silly: could we make the prints noexcept?
```
void print() noexcept {
  try {
    whatever_we_have_already();
  }
  catch (...) {
    std::cout << "something useful" << std::endl;
    std::terminate();
  }
}
```

Skylion007: The latter could definitely work and would fundamentally solve the problem. I think the compiler's main complaint is that you may not want to call std::terminate implicitly by throwing an exception in a destructor, but calling it explicitly should be fine. 

The former doesn't because we can only specify where clang-tidy is run by build taargets (ie. cpp) and since it's a header only library we can't ask it to run on headers unless they are actually used somewhere (and therefore have their templates expanded etc.)

Skylion007: Alternatively, someone can just go an add all the NOLINT comments manually.

rwgk: > Alternatively, someone can just go an add all the NOLINT comments manually.

That sounds like a quick mostly mechanical effort, based on clang-tidy diagnostics, is that true? I'm thinking it would be best to do that straightaway in this PR, but it's fine with me to do it separately. Please let me know.

BTW: It looks like the PR description is out of date. Could you please update it to reflect the final state, if necessary?

Skylion007: @rwgk Let's do it separetely. Having a no-except print version with std::terminate might be a better approach longterm.

include/pybind11/embed.h
```diff
@@ -189,6 +189,7 @@ class scoped_interpreter {
     scoped_interpreter &operator=(const scoped_interpreter &) = delete;
     scoped_interpreter &operator=(scoped_interpreter &&) = delete;
 
+    // NOLINTNEXTLINE(bugprone-exception-escape)
```
Skylion007, line 192: If we caught the exception and explicitly call std::terminate that would also work well.

include/pybind11/detail/common.h
```diff
@@ -220,16 +220,16 @@
...
-#define PYBIND11_LONG_FROM_UNSIGNED(o) PyLong_FromSize_t((size_t) o)
-#define PYBIND11_BYTES_NAME "bytes"
-#define PYBIND11_STRING_NAME "str"
-#define PYBIND11_SLICE_OBJECT PyObject
-#define PYBIND11_FROM_STRING PyUnicode_FromString
-#define PYBIND11_STR_TYPE ::pybind11::str
-#define PYBIND11_BOOL_ATTR "__bool__"
-#define PYBIND11_NB_BOOL(ptr) ((ptr)->nb_bool)
-#define PYBIND11_BUILTINS_MODULE "builtins"
+#    define PYBIND11_LONG_FROM_SIGNED(o) PyLong_FromSsize_t((ssize_t) (o))
```
Skylion007, line 223: I can revert the formatting here if necessary @rwgk . Applied by clang-format.

Skylion007:
@henryiii Ping. Would appreciated your feedback on this PR. :)

rwgk:
ICC flaked, I don't remember seeing that exact flake before, but it's download related, it doesn't even get to the configure stage. Maybe just try again?
(I have one suggestion though (I think). Hang on.)

.clang-tidy
rwgk: Sounds good.

include/pybind11/detail/common.h
rwgk: Up to you.
I often do/did, because I'm not sure I/we want to keep the current clang-format choices for preprocessor directives.

include/pybind11/detail/type_caster_base.h
```diff
@@ -38,6 +38,7 @@ class loader_life_support {
     }
 
     /// ... and destroyed after it returns
+    // NOLINTNEXTLINE(bugprone-exception-escape) See issue #2215
```
rwgk, line 41: This is my only strong suggestion:
IIUC we're NOT turning on bugprone-exception-escape in this PR. Is that true?
In that case it will be confusing to have the NOLINT in this PR, could you please remove them?
I realize it's a little bit of extra work adding them back in later, but IMM making this PR as straightforward/clean as possible is an important enough reason to invest those 5 minutes.

include/pybind11/gil.h
```diff
@@ -149,6 +149,7 @@ class gil_scoped_release {
         active = false;
     }
 
+    // NOLINTNEXTLINE(bugprone-exception-escape) See issue #2215
```
rwgk, line 152: Same here (other comment).

include/pybind11/numpy.h
```diff
@@ -1551,8 +1551,10 @@ struct vectorize_helper {
             "pybind11::vectorize(...) requires a function with at least one vectorizable argument");
 
 public:
-    template <typename T>
-    explicit vectorize_helper(T &&f) : f(std::forward<T>(f)) { }
+    template <typename T,
```
rwgk, line 1554: A 1-line hint (comment) that this is to not shadow the default copy constructor would be ideal. Maybe with a link to the clang-tidy documentation, for background.

tests/test_class.cpp
```diff
@@ -510,8 +523,9 @@ CHECK_HOLDER(6, shared); CHECK_HOLDER(7, shared); CHECK_HOLDER(8, shared);
 // failures occurs).
 
 // We have to actually look into the type: the typedef alone isn't enough to instantiate the type:
-#define CHECK_BROKEN(N) static_assert(std::is_same<typename Breaks##N::type, BreaksBase<-N>>::value, \
-        "Breaks1 has wrong type!");
+#define CHECK_BROKEN(N)                                                                           \
```
rwgk, line 526: For my own benefit, merging this into the smart_holder branch, if you could minimize the whitespace changes here that would definitely be helpful. But optional.

rwgk (approved):

henryiii:
Looks good, would second @rwgk's suggestions; a couple of places need comments as to why they are doing what they are doing and `bugprone-exception-escape` is probably less confusing if handled in a different PR (though personally not highly against adding a nolint for it if you know it won't pass, even if it's not enabled, but fine to wait and I do see that seeing a no lint for something that is not being linted is likely quite confusing).

Skylion007:
@henryiii Need your approval to merge. :)

rwgk (approved):
Thanks Aaron!

rwgk:
Looks like you lost a comma in numpy.h.

rwgk:
> @henryiii Need your approval to merge. :)

We are already two maintainers.
This is pure code health.
I don't want to by-pass anyone, but to me this clearly is fine with just one approval.
Note that even in a shop like Google, where you could in theory bring down a service for billions, this is enough approvals coverage. Here we're just working on master codehealth.
Caveat: at Google we cannot submit with a broken CI. But I trust nobody would do that here.

PR #3169, opened 2021-08-03
Fix func_handle for rule of two

Skylion007:
## Description
If the copy constructor is defined, the copy assignment constructor should also be defined. Found in #3070 but that was never merged. Found in LGTM static analysis. https://lgtm.com/projects/g/pybind/pybind11/ I looked into turning the clang-tidy check for this, but the only clang-tidy check is way too noisy and will give way too many false positive (only rule of 5 is implemented, ie all constructors, dtor, and assignment operators have to be explicitly defined or defaulted which seem excessive). 
The effect is that if we do anything more complex with this struct later, the code will be more maintainable.

### DIFF
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -70,9 +70,11 @@ public:
         struct func_handle {
             function f;
             func_handle(function &&f_) noexcept : f(std::move(f_)) {}
-            func_handle(const func_handle& f_) {
+            func_handle(const func_handle &f_) { operator=(f_); }
+            func_handle &operator=(const func_handle &f_) {
                 gil_scoped_acquire acq;
                 f = f_.f;
+                return &this;
             }
             ~func_handle() {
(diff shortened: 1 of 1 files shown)


include/pybind11/functional.h
```diff
@@ -70,9 +70,11 @@ struct type_caster<std::function<Return(Args...)>> {
         struct func_handle {
             function f;
             func_handle(function &&f_) noexcept : f(std::move(f_)) {}
-            func_handle(const func_handle& f_) {
+            func_handle(const func_handle &f_) { operator=(f_); }
+            func_handle &operator=(const func_handle &f_) {
                 gil_scoped_acquire acq;
                 f = f_.f;
+                return &this;
```
henryiii, line 77:
```suggestion
                return *this;
```

rwgk (approved):
I think this is a good change to get in asap (incl 2.7.1).

PR #3180, opened 2021-08-07
Consistent CI step name "Python tests".

rwgk:
Consistent naming makes it easy to pin-point where Python tests are run, to apply temporary changes.

See PR #3106 for how that might be useful (running tests repeatedly for deflaking).


henryiii (approved):

rwgk:
Thanks Henry!

PR #3183, opened 2021-08-08
maint(perf): Optimize Numpy constructor to remove copies by value.

Skylion007:
## Description
* Was looking through the numpy code and found a bunch of unneccessary copies. I added some std::move, but this constructor may want to use const-refs instead. I am not sure we have build coverage of this constructor hence why clang-tidy didn't detect this perf error.

## Suggested changelog entry:

```rst
* Optimize numpy array construction with additional moves.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -478,11 +478,11 @@ public:
 
     dtype(list names, list formats, list offsets, ssize_t itemsize) {
         dict args;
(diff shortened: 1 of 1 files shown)


henryiii (approved):

rwgk (commented):
I'm late here, but that's great! :-)
Logging my approval only to track that I looked.

PR #3186, opened 2021-08-09
Removing last remnants of pragma block at the top of pybind11.h

rwgk:
Defaulting CUDA, GCC7, GCC8 to `PYBIND11_NOINLINE_DISABLED`, with the option to define `PYBIND11_NOINLINE_FORCED`.  Relevant `inline`/`noinline` [shared-library size differences are tabulated here](https://docs.google.com/spreadsheets/d/1bkg1HtmU1AjotPTXCCLSbTBGARRet8W91noh1DnKOHU/edit#gid=0https://docs.google.com/spreadsheets/d/1bkg1HtmU1AjotPTXCCLSbTBGARRet8W91noh1DnKOHU/edit#gid=0). The measured size saving when using `noinline` are only 1.7% for CUDA, -0.2% for GCC7, and 0.0% for GCC8 (using `-DCMAKE_BUILD_TYPE=MinSizeRel`, the default under pybind11/tests).

_______________________________________________________________________________

This PR changes 2 files, removes 9 lines, adds 11 lines (6 of which are for a comment).
_______________________________________________________________________________
## Suggested changelog entry:

```rst
The warnings-suppression "pragma clamp" at the top/bottom of pybind11 was removed, clearing the path to refactoring and IWYU cleanup.
```

CUDA, GCC7, GCC8 builds that prioritize `noinline` over `-Wattributes` diagnostics need to add `-DPYBIND11_NOINLINE_FORCED -Wno-attributes` to the `g++` command lines. However, the measured shared-library size saving when using `noinline` are only 1.7% for CUDA, -0.2% for GCC7, and 0.0% for GCC8 (using `-DCMAKE_BUILD_TYPE=MinSizeRel`, the default under pybind11/tests).

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -99,7 +99,7 @@
 #  endif
 #endif
 
-#define PYBIND11_NOINLINE_FORCED
+#define PYBIND11_NOINLINE_DISABLED
 // For CUDA, GCC7, GCC8:
 // PYBIND11_NOINLINE_FORCED is incompatible with `-Wattributes -Werror`.
 // When defining PYBIND11_NOINLINE_FORCED, it is best to also use `-Wno-attributes`.
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -99,7 +99,6 @@
 #  endif
 #endif
 
-#define PYBIND11_NOINLINE_FORCED
 // For CUDA, GCC7, GCC8:
 // PYBIND11_NOINLINE_FORCED is incompatible with `-Wattributes -Werror`.
 // When defining PYBIND11_NOINLINE_FORCED, it is best to also use `-Wno-attributes`.
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -99,6 +99,7 @@
 #  endif
 #endif
 
+#define PYBIND11_NOINLINE_FORCED
 // For CUDA, GCC7, GCC8:
 // PYBIND11_NOINLINE_FORCED is incompatible with `-Wattributes -Werror`.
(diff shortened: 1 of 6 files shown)


Skylion007 (approved):

include/pybind11/detail/common.h
```diff
@@ -99,6 +99,17 @@
...
 
+// For CUDA, GCC7, GCC8:
+// PYBIND11_NOINLINE_FORCED is incompatible with `-Wattributes -Werror`.
+// When defining PYBIND11_NOINLINE_FORCED, it is best to also use `-Wno-attributes`.
+// However, the measured shared-library size saving when using noinline are only
+// 1.7% for CUDA, -0.2% for GCC7, and 0.0% for GCC8 (using -DCMAKE_BUILD_TYPE=MinSizeRel,
+// the default under pybind11/tests).
+#if !defined(PYBIND11_NOINLINE_FORCED) && \
+    (defined(__CUDACC__) || (defined(__GNUC__) && (__GNUC__ == 7 || __GNUC__ == 8)))
+#  define PYBIND11_NOINLINE_DISABLED
```
Skylion007, line 110: I am guessing there is no clean way to detect if the warning is enabled is there? If there is seems like it might make sense to add it to the conditional.

rwgk: > I am guessing there is no clean way to detect if the warning is enabled is there?

I don't think so, not even `noinline` is portable, even less so introspection of compiler-specific warnings related to it.

> If there is seems like it might make sense to add it to the conditional.

If it was easy, yes.

But how likely is it that anyone cares enough about a 1.7% binary size reduction when building with CUDA? If someone does, they can add the suggested command line options. If they don't like that, they could send us a PR. — The group of affected people and the binary size saving both seem so marginal to me, I cannot imagine it's worth anyone's time, cost vs benefit.

PR #3195, opened 2021-08-12
maint(precommit): Apply isort

Skylion007:
## Description

* Applies the isort formatter to the repo and enables the pre-commit hook for it. It is a simple formatter which is compatible with black and ensure that all imports in the repo occur in a standard (alphabetical) order unless otherwise indicated. It also otherwise can format the imports and ensure standard, third party, and first party libraries all go in their standard order. The ordering is automatically applied and enforced with a pre-commit hook.

## Suggested changelog entry:

```rst
* Format pybind11 with isort consistent ordering of imports
* Enable isort pre-commit hook
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,9 @@ ignore = [
 
 [tool.isort]
 known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
+# For black compatibility
 profile = "black"
+multi_line_output = 3
 
 [tool.mypy]
 files = "pybind11"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ ignore = [
 ]
 
 [tool.isort]
-known_first_party = "env,pybind11_tests,"
+known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
 profile = "black"
 
 [tool.mypy]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ ignore = [
 ]
 
 [tool.isort]
+# Needs the compiled .so modules and env.py from tests
 known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
 # For black compatibility
 profile = "black"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,9 @@ ignore = [
     "noxfile.py",
 ]
 
+[tool.isort]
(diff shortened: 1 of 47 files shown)


rwgk (approved):
Cool!

pyproject.toml
```diff
@@ -15,6 +15,10 @@ ignore = [
     "noxfile.py",
 ]
 
+[tool.isort]
+known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
+profile = "black"
```
henryiii, line 20:
```suggestion
profile = "black"
multi_line_output = 3
```

I think this is slightly more black-like, but was added after the black profile was created.

henryiii:
I'm [obviously a fan.](https://scikit-hep.org/developer/style#isort-extra) :)

pyproject.toml
```diff
@@ -15,6 +15,10 @@ ignore = [
     "noxfile.py",
 ]
 
+[tool.isort]
+known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
```
Skylion007, line 19: Normally, these are autodetected, but it's a bit trickier in this project since the compiled modules don't live in the root, so we have to include any top level modules it might have missed (in tests)

pyproject.toml
henryiii: Ahh, you are correct in your commit message, this is was only true on older black versions - it's now part of the standard black Profile. Please feel free to drop. We only support the linter versions as specified in the pre-commit file.

pyproject.toml
```diff
@@ -15,6 +15,13 @@ ignore = [
     "noxfile.py",
 ]
 
+[tool.isort]
+# Needs the compiled .so modules and env.py from tests
+known_first_party = "env,pybind11_cross_module_tests,pybind11_tests,"
+# For black compatibility
+profile = "black"
+multi_line_output = 3
```
henryiii, line 23:
```suggestion
profile = "black"
```

henryiii (approved):
Sorry about the confusion, multi_line_output is part of "black".

Skylion007:
> 
> 
> Sorry about the confusion, multi_line_output is part of "black".

Should I revert it? Or just leave it as is?

pyproject.toml
henryiii: Yes, please revert - it's completely redundant (can just accept this).

PR #3198, opened 2021-08-15
Removing obsolete eigen.h warning suppression pragmas.

rwgk:
Long story short: the `pragma`s in eigen.h are simply not needed when using `-isystem`, as all non-Windows CI jobs do.

A `HINT ... use -isystem` was added to eigen.h.

For MSVC the existing `PYBIND11_SILENCE_MSVC_C4127` macro was added in 4 locations.

For completeness, the existing `pragma`s have several issues:

* The suppressions are applied to all code in eigen.h, not just the Eigen headers.
* The `__INTEL_COMPILER` `pragma` is not pushed (it applies globally from the point of including eigen.h).
* The `#ifdef` logic for `push`/`pop` is inconsistent, which may or may not matter (see e.g. https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers).

Luckily, we can just purge the `pragma`s completely.

## Suggested changelog entry:
```rst
Legacy warning suppression pragma were removed from eigen.h. On Unix platforms, please use -isystem for Eigen include directories, to suppress compiler warnings originating from Eigen headers. Note that cmake does this by default. No adjustments are needed for Windows.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,18 +106,18 @@ jobs:
         ${{ matrix.args }}
 
     - name: Build C++11
-      run: cmake --build . --verbose -j 2
+      run: cmake --build . -j 2
 
     - name: Python tests C++11
-      run: cmake --build . --target pytest
+      run: cmake --build . --target pytest -j 2
 
     - name: C++11 tests
       # TODO: Figure out how to load the DLL on Python 3.8+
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

rwgk:
Thanks @Skylion007!

PR #3207, opened 2021-08-19
extend msys2 mingw CI

lkeegan:
## Description

extension of #3132 with

- additional 32-bit mingw CI job
- c++11/17 c++/interface tests copied from standard ci
- numpy/scipy installed to also run these tests

note: 32-bit numpy tests are currently failing due to a difference in the padding of structs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -911,17 +911,35 @@ jobs:
     - name: Clean directory
       run: git clean -fdx
 
-    - name: Configure C++17
-      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=17 -S . -B build2
+    - name: Configure C++14
+      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=14 -S . -B build2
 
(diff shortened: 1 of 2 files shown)


Skylion007 (approved):

.github/workflows/ci.yml
```diff
@@ -861,32 +861,67 @@ jobs:
...
+    - name: C++11 tests
+      run: cmake --build build --target cpptest -j 2
+
+    - name: Interface test C++11
+      run: cmake --build build --target test_cmake_build
+
+    - name: Clean directory
+      run: git clean -fdx
+
+    - name: Configure C++17
```
Skylion007, line 914: Any reason we aren't testing C++14 as well?

henryiii: Isn't that just above?

henryiii: Oh, it's been added. The reason we don't generally check it is if 11 and 17 both work, 14's pretty likely to work too, and it saves time.

henryiii (approved):

PR #3219, opened 2021-08-25
Adding `ssize_t_cast` to support passing `size_t` or `ssize_t` values where `ssize_t` is needed.

rwgk:
## Description
This PR solves the problem described under PR #2692, but with a more general and more thorough implementation, including full unit test coverage for all functions with modified signatures.

The entire include/pybind11 directory was inspected for `static_cast<ssize_t>` and `(ssize_t)` C-style casts. A couple tangential minor enhancements are included in this PR.

clang-format-diff was applied.

## Suggested changelog entry:

```rst
The pybind11 proxy types `str`, `bytes`, `bytearray`, `tuple`, `list` now consistently support passing `ssize_t` values for sizes and indexes. Previously, only `size_t` was accepted in several interfaces.
```


rwgk:
The one macos failure is just a flake (the job never actually ran).

Skylion007 (approved):

rwgk:
Thanks @Skylion007 for the approval!

PR #3223, opened 2021-08-27
bytes_ssize_t -Wnarrowing reproducer (see PR #2692).

rwgk:
Just to see what breaks.


rwgk:
Closing: #3219 was just merged, with many more similar tests that would not have built without the changes.

Issue #3228, opened 2021-08-29
[BUG]: Constructing pybind11::cpp_function with a closure causes memory leak


yuantailing:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

When constructing a `pybind11::cpp_function` object with lvalue reference of a closure, the captures are copied but not to be deconstructed. The expected behavior is that all copies of captures are deconstructed.

Versions:
- python 3.8
- pybind11-2.7.1
- x86_64-linux-gnu-gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

The issue may be related to #1510 and https://github.com/pytorch/pytorch/issues/5161.

### Reproducible example code

Add the following code to https://github.com/pybind/python_example/blob/master/src/main.cpp

```cpp
    class A {
    public:
        A() { printf("A()\n"); }
        A(A const &) { printf("A(A const &)\n"); }
        A(A &&) { printf("A(A &&)\n"); }
        ~A() { printf("~A()\n"); }
    };

    m.def("foo1", [] {
        A a;
        auto bar = [a] { };
        return py::cpp_function(bar);
    });

    m.def("foo2", [] {
        A a;
        auto bar = [a] { };
        return py::cpp_function(std::move(bar));
    });
```

Then call `m.foo1()`. It shows that one object of `A` is not deconstructed:

```
A()
A(A const &)
A(A const &)
~A()
~A()
```

Note that `m.foo2()` does not cause memory leak:
```
A()
A(A const &)
A(A &&)
~A()
~A()
~A()
```

Issue #1510, opened 2018-08-31
Leaks from Address Sanitizer


mbrucher:
The issue is related to https://github.com/pytorch/pytorch/issues/5161

It seems that pybind11::cpp_function::initialize_generic (and others) are allocating data that never gets freed before the end of the run, when ASAN is unloaded.

Is it a known issue in pybond11? Boost.Python also has similar issues, so wondering if there shouldn't be a documentation on how to avoid these or valgrind/ASAN blacklist?

bstaletic:
Running Valgrind on a python executable is always troublesome. CPython, while written in C, doesn't use `malloc`, instead it has its own implementation called `PyMalloc`. When you compile your interpreter you can pass `--with-valgrind` and `--without-pymalloc`, yet this is still not enough.

According to Valgrind, python2 leaks a lot more than python3, but has orders of magnitude fewer memory access errors.
It is possible to suppress Valgrind errors and to even generate suppression rules with `--gen-suppressions`, but once I got a clean Valgrind run on my local machine it didn't work on CI builds.
There's [an outdated valgrind suppression file](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Misc/valgrind-python.supp) in the CPython repository too.

All this said, I think there might be a few issues coming from pybind11, not just CPython. For that part take a look at https://github.com/pybind/pybind11/pull/1454, which touches upon the discussion of ridding pybind11 of (known) leaks.

mbrucher:
These are the issues that ASAN finds on pybind11 code. As Python is not generated with asan, there is no issue with Python code. Only the memory allocated by the pyblind11-wrapped modules and not freed by them.
I don't care currently about valgrind, only ASAN builds (because they are faster and better).

micbou:
Look at the memory leaks. They are all coming from CPython functions (mostly `_PyCFunction_FastCallDict`). I doubt pybind11 can do much about that.

bstaletic:
@mbrucher ASAN begin "better" is debatable. Try running valgrind with `--track-origins=full` or `--track-origins=yes`. It will make valgrind even slower, but it will tell you where an error occured.

mbrucher:
I know both of them quite well ;)

mbrucher:
@micbou perhaps, but I didn't see anything in the documentation about how to solve this issue...

bstaletic:
[...]

PR #3229, opened 2021-08-29
fix: memory leak in cpp_function (#3228)

yuantailing:
## Description

The PR is related to #3228, and may be related to #1510 and pytorch/pytorch#5161.

Let's focus on `pybind11::cpp_function::initialize`. If the first argument is lvalue reference, the type `Func` is deduced to lvalue reference (according to [Template argument deduction](https://en.cppreference.com/w/cpp/language/template_argument_deduction)). In this case, `std::is_trivially_destructible<Func>` is always true, therefore `rec->data` will never be deconstructed.

To solve this issue, we should determine whether `rec->data` should be deconstructed according whether the type of `rec->data` (i.e., `capsule`) is trivially destructible.

### DIFF
--- a/tests/test_embed/test_interpreter.cpp
+++ b/tests/test_embed/test_interpreter.cpp
@@ -1,5 +1,4 @@
 #include <pybind11/embed.h>
-#include "../constructor_stats.h"
 
 #ifdef _MSC_VER
 // Silence MSVC C++17 deprecation warning from Catch regarding std::uncaught_exceptions (up to catch
@@ -323,32 +322,4 @@ TEST_CASE("sys.argv gets initialized properly") {
         REQUIRE(cpp_widget.argv0() == "a.out");
     }
     py::initialize_interpreter();
-}
-
-class MyFunctionObject {
-public:
-    MyFunctionObject() { track_default_created(this); }
-    ~MyFunctionObject() { track_destroyed(this); }
-    MyFunctionObject(const MyFunctionObject &) { track_copy_created(this); }
-    MyFunctionObject(MyFunctionObject &&) noexcept { track_move_created(this); }
-    void operator()() { }
-};
-
-// Related issue: https://github.com/pybind/pybind11/issues/3228
-// Related PR: https://github.com/pybind/pybind11/pull/3229
-TEST_CASE("Check objects are deconstructed in cpp_function") {
-    MyFunctionObject func_obj;
-    ConstructorStats &stat = ConstructorStats::get<MyFunctionObject>();
-    {
-        py::cpp_function func(func_obj);
-        func();
-        CHECK(stat.alive() == 2);
-    }
-    CHECK(stat.alive() == 1);
-    {
-        py::cpp_function func(std::move(func_obj));
-        func();
-        CHECK(stat.alive() == 2);
-    }
-    CHECK(stat.alive() == 1);
 }
--- a/tests/test_embed/test_interpreter.cpp
+++ b/tests/test_embed/test_interpreter.cpp
@@ -341,12 +341,14 @@ TEST_CASE("Check objects are deconstructed in cpp_function") {
     ConstructorStats &stat = ConstructorStats::get<MyFunctionObject>();
     {
         py::cpp_function func(func_obj);
-        (void)func;
+        func();
+        CHECK(stat.alive() == 2);
     }
     CHECK(stat.alive() == 1);
     {
         py::cpp_function func(std::move(func_obj));
-        (void)func;
+        func();
+        CHECK(stat.alive() == 2);
     }
     CHECK(stat.alive() == 1);
 }
--- a/tests/test_embed/test_interpreter.cpp
+++ b/tests/test_embed/test_interpreter.cpp
@@ -323,3 +323,32 @@ TEST_CASE("sys.argv gets initialized properly") {
     }
     py::initialize_interpreter();
 }
+
+class Counter {
+public:
+    Counter() { cnt_++; }
+    Counter(const Counter &) { cnt_++; }
+    Counter(Counter &&) { cnt_++; }
+    ~Counter() { cnt_--; }
+    static int get_cnt() { return cnt_; }
+    void operator()() { }
+
+private:
+    static int cnt_;
+};
+
+int Counter::cnt_ = 0;
+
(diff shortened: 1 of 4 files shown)


Skylion007:
Very interesting and surprised Valgrind and our test suite isn't catching this.

Do you think it's possible that this is related to this PyBind failing test case as well: https://github.com/pybind/pybind11/pull/1301? Does that test case pass when added to this PR?

Can the MRE be reduced to a test case to repro the error in a way our test suite can detect?

rwgk:
Amazing indeed that this bug survived until now.

This seems to be the commit that introduced the line you're changing:

https://github.com/pybind/pybind11/commit/19208fe9a4e27e38dfc17456e411e38f53de5f2c

I don't know if it was a bug from the start, or if the bug is the result of subsequent changes.

@yuantailing could you help by converting your #3228 reproducer to a unit test? Maybe an addition to test_callbacks.cpp,py. constructor_stats.h may be useful. But I'm very uncertain about the details, I haven't touched this code myself, and I don't know how the original unit test (see commit above) relates to constructor_stats.h.

yuantailing:
> Do you think it's possible that this is related to this PyBind failing test case as well: #1301? Does that test case pass when added to this PR?

I think they are not related. The issue happens when a lvalue reference is passed to `pybind11::cpp_function`. Test case #1301 tests that a python function can be converted to `std::function`. They are not the same.

> Can the MRE be reduced to a test case to repro the error in a way our test suite can detect?

Added.

Skylion007:
> Do you think it's possible that this is related to this PyBind failing test case as well: #1301? Does that test case pass when added to this PR?
Yeah, I just realized it's a subtle issue with finalize_interperter and therefore is likely not an actual bug.

tests/test_embed/test_interpreter.cpp
```diff
@@ -323,3 +323,32 @@ TEST_CASE("sys.argv gets initialized properly") {
     }
     py::initialize_interpreter();
 }
+
+class Counter {
+public:
+    Counter() { cnt_++; }
+    Counter(const Counter &) { cnt_++; }
+    Counter(Counter &&) { cnt_++; }
```
Skylion007, line 331:
```suggestion
    Counter(Counter &&) noexcept { cnt_++; }

File: tests/test_embed/test_interpreter.cpp [hunk 2]
```diff
@@ -323,3 +323,32 @@ TEST_CASE("sys.argv gets initialized properly") {
...
+    ~Counter() { cnt_--; }
+    static int get_cnt() { return cnt_; }
```
Skylion007, line 333:
```suggestion
    static int get_cnt() const { return cnt_; }
```

yuantailing:
> @yuantailing could you help by converting your #3228 reproducer to a unit test? Maybe an addition to test_callbacks.cpp,py. constructor_stats.h may be useful. But I'm very uncertain about the details, I haven't touched this code myself, and I don't know how the original unit test (see commit above) relates to constructor_stats.h.

I have added a test case to *tests/test_embed/test_interpreter.cpp*. Maybe the code can be moved to a more proper place.

tests/test_embed/test_interpreter.cpp
yuantailing: error: static member function `static int Counter::get_cnt()` cannot have cv-qualifier

Skylion007:
> 
> 
> > @yuantailing could you help by converting your #3228 reproducer to a unit test? Maybe an addition to test_callbacks.cpp,py. constructor_stats.h may be useful. But I'm very uncertain about the details, I haven't touched this code myself, and I don't know how the original unit test (see commit above) relates to constructor_stats.h.
> 
> I have added a test case to _tests/test_embed/test_interpreter.cpp_. Maybe the code can be moved to a more proper place.

Constructor_Stats.h just defines a helper class with keeps track of all its' constructions / destructions. You may want to use that class instead of your counter class if it makes sense: https://github.com/pybind/pybind11/blob/e08a58111dbea38d667b209f7543864d51a3b185/tests/constructor_stats.h#L73

tests/test_embed/test_interpreter.cpp
Skylion007: Ah, sorry forgot it was static.

yuantailing:
> Constructor_Stats.h just defines a helper class with keeps track of all its' constructions / destructions. You may want to use that class instead of your counter class if it makes sense:
> 
> https://github.com/pybind/pybind11/blob/e08a58111dbea38d667b209f7543864d51a3b185/tests/constructor_stats.h#L73

Thanks for your suggestions, fixed now.

tests/test_embed/test_interpreter.cpp
yuantailing: Thanks for your suggestions. Fixed in newer commits.

Skylion007 (approved):

rwgk:
The `(void)func;` are no-ops I think. Was that intentional? I just pushed a commit that changes them to `func();`. Could you please let me know if you think that's not correct?

I also added two `CHECK(stat.alive() == 2);` to make it more obvious and certain that the count actually goes down.

I will now look into moving the test elsewhere, because it's not specific to embedding, but specific to callbacks.

rwgk:
> I will now look into moving the test elsewhere, because it's not specific to embedding, but specific to callbacks.

Done.

@Skylion007, if my changes look good to you, please go ahead and merge (pending completion of CI jobs).

rwgk (approved):
(Assuming CI will be green.)

yuantailing:
> The `(void)func;` are no-ops I think. Was that intentional? I just pushed a commit that changes them to `func();`. Could you please let me know if you think that's not correct?

The purpose is to avoid the potential **unused variable** warning. Change it to `func()` is also correct.

> I also added two `CHECK(stat.alive() == 2);` to make it more obvious and certain that the count actually goes down.
> 
> I will now look into moving the test elsewhere, because it's not specific to embedding, but specific to callbacks.

Looks good.

rwgk:
> The purpose is to avoid the potential **unused variable** warning. Change it to `func()` is also correct.

Ah, got it. Thanks for the explanation!
We have a handy function to make that obvious: `py::detail::silence_unused_warnings(py_func);`

PR #3233, opened 2021-09-01
fix: Set __file__ constant when using eval_file (#1300)

### LINKED ISSUE #1300: Setting __file__ in eval_file
Hi,
I wonder if it wouldn't make sense to set __file__ global variable by default inside pybind11::eval_file executed scripts might want to use it.
A kind of recurring pattern in python is:
```
import os
script_path = os.path.realpath(__file__)
```

Skylion007:
## Description
Closes #1300 by setting the `__file__` constant when running eval_file.

## Suggested changelog entry:

```rst
* Sets __file__ constant when running eval_file in an embedded interpreter.
```

### DIFF
--- a/include/pybind11/eval.h
+++ b/include/pybind11/eval.h
@@ -136,16 +136,14 @@ object eval_file(str fname, object global = globals(), object local = object()) 
         pybind11_fail("File \"" + fname_str + "\" could not be opened!");
     }
 
-    if (!global.contains("__file__")) {
+    // In Python2, this should be encoded by getfilesystemencoding.
+    // We don't boher setting it since Python2 is past EOL anyway.
+    // See PR#3233
 #if PY_VERSION_HEX >= 0x03000000
+    if (!global.contains("__file__")) {
         global["__file__"] = std::move(fname);
-#else
-        // In Python2, this should be encoded by getfilesystemencoding.
-        // We are just assuming it's UTF-8 since Python2 is past EOL anyway.
-        // See PR#3233
-        global["__file__"] = bytes(fname);
-#endif
     }
+#endif
 
 #if PY_VERSION_HEX < 0x03000000 && defined(PYPY_VERSION)
     PyObject *result = PyRun_File(f, fname_str.c_str(), start, global.ptr(),
--- a/include/pybind11/eval.h
+++ b/include/pybind11/eval.h
@@ -88,20 +88,20 @@ void exec(const char (&s)[N], object global = globals(), object local = object()
 
 #if defined(PYPY_VERSION) && PY_VERSION_HEX >= 0x03000000
 template <eval_mode mode = eval_statements>
-object eval_file(str, object, object) {
+object eval_file(const str&, object, object) {
     pybind11_fail("eval_file not supported in PyPy3. Use eval");
 }
 template <eval_mode mode = eval_statements>
-object eval_file(str, object) {
+object eval_file(const str&, object) {
     pybind11_fail("eval_file not supported in PyPy3. Use eval");
 }
 template <eval_mode mode = eval_statements>
-object eval_file(str) {
+object eval_file(const str&) {
     pybind11_fail("eval_file not supported in PyPy3. Use eval");
 }
 #else
 template <eval_mode mode = eval_statements>
-object eval_file(str fname, object global = globals(), object local = object()) {
+object eval_file(const str &fname, object global = globals(), object local = object()) {
     if (!local)
         local = global;
 
--- a/include/pybind11/eval.h
+++ b/include/pybind11/eval.h
@@ -115,6 +115,10 @@ object eval_file(str fname, object global = globals(), object local = object()) 
         default: pybind11_fail("invalid evaluation mode");
     }
 
+    if (!global.contains("__file__")) {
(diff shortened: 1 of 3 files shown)


tests/test_cmake_build/test.py
```diff
@@ -3,5 +3,7 @@
 
 import test_cmake_build
 
+__file__  # Test this is properly set
```
rwgk, line 6: What do you think about making this slightly stronger (and look more like a test)? E.g.
```
assert isinstance(__file__, str)  # Test this is properly set.
```

rwgk (approved):

tests/test_cmake_build/test.py
Skylion007: It's actually confusing if it's suppose to a byte or str or if that itself is implementation dependant. Were you able to find the spec that says it's suppose to a str? And a unicode str or an ascii str at that?

Skylion007: I just left it as is since if it's a bug, it's a bug with our entire function signature for eval.

Skylion007: I knew it! 
```
isinstance(__file__, str) == False on Python2
```

rwgk: Oh, because it is unicode even on Python 2? — I think that counts as a bug, too!
Although not a surprising one given how pybind11 works internally. I think in that case we want to keep the test as is (isinstance str), but encode in the C++ code.
Give me a moment to look around a bit what the expections are for __file__ with Python 2.

rwgk: A bit random, but this looks very similar:

https://github.com/pypa/pip/pull/1538/files

rwgk: Maybe more pragmatically: guard the new C++ code with `#ifdef` for Python 3, and the test with `not env.PY2`.
I.e. we're simply leaving Python 2 behind (instead of going through hoops for something that is practically dead already).

Skylion007: I don't like that because it should be defined in Python 2, the previous code was checking for NameError.

Skylion007: > 
> 
> A bit random, but this looks very similar:
> 
> https://github.com/pypa/pip/pull/1538/files

Nah, this is same code snippit is suppose to be done in the Python2 side to convert __file__ to the appropriate format.

rwgk: I'm good with either:
* status quo for Python 2 (carry forward existing shortcoming),
* or fixing properly.
For the latter, is there a Python C API for `sys.getfilesystemencoding()`? If not we have to import, getattr, call.

tests/test_cmake_build/test.py
```diff
@@ -3,5 +3,10 @@
 
 import test_cmake_build
 
+from .. import env
+
+if env.PY3:
```
rwgk, line 8: Without the matching guard in C++ we are as much creating a problem as we are solving one.

Skylion007:
@rwgk Can you investigate fixing the encoding then for Python2, not familiar with searching the CPython documentation.

include/pybind11/eval.h
```diff
@@ -136,6 +136,14 @@ object eval_file(str fname, object global = globals(), object local = object())
         pybind11_fail("File \"" + fname_str + "\" could not be opened!");
     }
 
+// Python2 API requires this to be encoded to filesystemencoding.
```
rwgk, line 139: This looks great.
I see the `env.PY2` is giving us trouble. There is another easy trick, JIC there isn't a better one:
```
if str is not bytes:  # Python 2 is not supported.
```
(We'll find that easily when we clear out Python 2 with the firehose in a few months.)

Skylion007:
Looks like a flake.

rwgk (approved):
Cool! Thanks for picking this up.

Skylion007:
Actually, after thinking it over, I am tempted to just leave the global __file__ in Python2 (as UTF-8 bytes) because there is chance it could be right (if the filesystem encoding is UTF-8), and it shouldn't break any previous code because any code that required __file__ had to set it byitself with the global dict and this code snippit will not interfere with that path. @henryiii Is there an easy way to encode to the getfilesystemencoding ? I know there is a C-API for codecs which makes this easier.

@rwgk Going to revert to the bytes method for now as it will usually be correct, if someone wants to fix it they can.

henryiii (approved):
I’m not very worried about Python 2 support, as long as it doesn’t regress for now, I’m fine with it. AFAICT, this just means that `__file__` wasn't previously available, but now it is - but on Python 2, it might be misencoded if not unicode.

It is possible that an invalid unicode sequence could show up here, but if something really bad happens, Python users can just set `__file__` manually.

henryiii:
(I'd also be okay to leave it unset on Python 2 at this stage)

rwgk:
> (I'd also be okay to leave it unset on Python 2 at this stage)

I think that really is the best, because ...

> but if something really bad happens, Python users can just set __file__ manually.
(I read this as "Python 2 users")

nobody wants to mess with Python 2 changes anymore. Just leaving everything as-is as much as possible is sure to be the least disruptive.

PR #3236, opened 2021-09-02
(perf): Add a missing noexcept to a pytype constructor

Skylion007:
Adds a missing noexcept to a PyType constructors for perf reasons.

## Description
* I noticed a move constructor that didn't have noexcept which prevents it from being used in certain contexts (like being moved into a STL vector), this rectifies that issue.

## Suggested changelog entry:

```rst
* Perf: Mark a pytype move constructor as noexcept
```


rwgk (approved):
Thanks Aaron!

Issue #2765, opened 2020-12-31
[BUG] pybind11:cast is not thread safe


zachariahreed:
in particular use of `loader_life_support` seems problematic. This function keeps state in the global `loader_patient_stack`: https://github.com/pybind/pybind11/blob/b7dfe5cc84ca1891b50e728e83c9b5b393cf5272/include/pybind11/detail/internals.h#L105 In the case where a cast can release the GIL, two thread's push/pop of `loader_patient_stack` can become interleaved. 

While releasing the GIL in a conversion seems like a bad idea, it can be surprisingly hard to avoid. Any cast that results in the bytecode interpreter being run (either directly or through something like a `__del__` method) can cause python itself to release GIL (`sys.setswitchinterval`).

YannickJadoul:
@zachariahreed, that's interesting. Do you have concrete code that manages to show this, or how did you run into this?

I was about to shout "But of course you should hold the GIL", but then you already thought of that.
However, by the time the interpreter stops running and hands back control to pybind11, I'd expect the GIL to be held again? Do you mind explaining slightly more in depth how the GIL can be released without pybind11 knowing about it?

zachariahreed:
The GIL is held when you get control back. The problem comes when another thread with pybind11-wrapped code gets scheduled during the time when you don't have control. Then you end up in a situation like:

```
Thread1          Thread2
--------------   --------------
Push             ...
ByteCode         ...
     Yield       ...
        ...      Push
        ...      ByteCode
        ...          Yield
Pop -> Boom      ...

```


The specific code that triggered this was use of `pybind11::module::import` inside of a custom caster. Much of the import machinery is written in pure python these days and its complex enough that it can certainly run longer than the multiplexing period.

YannickJadoul:
Argh, right, yes. So rather than `pop`ing, we should remember which pointer was added?

zachariahreed:
I don't think that's sufficient. During the time between when thread 1 gets control back and before the pop, if additional temporaries are created they'll end up in thread2's "arena".  The flow control is not linear, so I don't think there's any way a `vector`-based strategy can work. I looked very briefly at moving this into thread-local storage, but I was a bit concerned about performance as this seems to be a hot path.

YannickJadoul:
[...]

PR #3237, opened 2021-09-02
Fix thread safety for pybind11 loader_life_support

laramiel:
## Description

Fixes issue: https://github.com/pybind/pybind11/issues/2765

This converts the vector of PyObjects to either a single `void*` or a per-thread `void*` depending on the WITH_THREAD define.

The new `void*` field is used to construct a stack of loader_life_support frames that can extend the life of python objects.

The pointer is updated when the loader_life_support object is allocated (which happens before a `call`) as well as on release.

Each `loader_life_support` frame maintains a set of PyObject references that need to be lifetime extended; this is done by storing them in a c++ `std::unordered_set` and clearing the references when the object goes out of scope when the call has completed.

## Suggested changelog entry:

```rst
Fixes thread safety for some pybind11::type_caster which require lifetime extension, such as for std::string_view.
```


include/pybind11/detail/internals.h
```diff
@@ -106,14 +106,16 @@ struct internals {
     std::unordered_map<const PyObject *, std::vector<PyObject *>> patients;
     std::forward_list<ExceptionTranslator> registered_exception_translators;
     std::unordered_map<std::string, void *> shared_data; // Custom data to be shared across extensions
-    std::vector<PyObject *> loader_patient_stack; // Used by `loader_life_support`
     std::forward_list<std::string> static_strings; // Stores the std::strings backing detail::c_str()
     PyTypeObject *static_property_type;
     PyTypeObject *default_metaclass;
     PyObject *instance_base;
```
daniel-smith, line 112: Do we need to update `PYBIND11_INTERNALS_VERSION`?

laramiel:
Sure, a PyList or a vector or some other container would work as well.

The idea of using an unordered_set is intended to also address 
https://github.com/pybind/pybind11/issues/1251

daniel-smith:
Is it worth adding a python test similar to https://github.com/pybind/pybind11/issues/2765#issuecomment-907572642 somewhere?

laramiel:
I added a similar version; perhaps I should use a convertible reference rather than a string view. Let me know what you think.

daniel-smith:
> I added a similar version; perhaps I should use a convertible reference rather than a string view. Let me know what you think.

I believe that going via the `string_view` caster would exhibit the same issues - though I've not tested this out myself. I think as long as the test exercises the relevant code path, and verifies that things are thread-safe, then we're good.

daniel-smith:
Actually @laramiel, one consideration of using string views for the test is that this is a C++17 feature - if we want to support compilation of tests on older C++ standards (?)

include/pybind11/detail/type_caster_base.h
```diff
@@ -32,46 +32,37 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 /// Adding a patient will keep it alive up until the enclosing function returns.
 class loader_life_support {
 public:
```
daniel-smith, line 34: Can these members be private?

laramiel: Done.

laramiel:
@daniel-smith Converted test to use lifetime-extended references.

include/pybind11/detail/common.h
```diff
@@ -11,11 +11,11 @@
 
 #define PYBIND11_VERSION_MAJOR 2
 #define PYBIND11_VERSION_MINOR 8
-#define PYBIND11_VERSION_PATCH 0.dev1
+#define PYBIND11_VERSION_PATCH 0.dev2
```
Skylion007, line 14: You have to update __version.py as well.

laramiel: Done.

include/pybind11/detail/type_caster_base.h
laramiel: Done.

tests/test_thread.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+import concurrent.futures
+
+import pytest
+
+import env  # noqa: F401
+from pybind11_tests import thread as m
+
+
+def test_implicit_conversion():
```
Skylion007, line 10: Concurrent Futures are not in Python2 so you will need add skips here to make sure they aren't run.

laramiel: Changed to use threading.Thread

Skylion007:
Also please remove unused imports from the added test python file and ensure clang-tidy is happy.

tests/test_thread.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+
+import threading
+
+import pytest
+
+import env  # noqa: F401
```
Skylion007, line 7: This import isn't needed if you aren't using it for skipifs.

include/pybind11/detail/internals.h
```diff
@@ -154,7 +157,7 @@ struct type_info {
 };
 
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version
-#define PYBIND11_INTERNALS_VERSION 4
+#define PYBIND11_INTERNALS_VERSION 5
```
henryiii, line 160: If we are bumping the ABI, we should probably move through several of the PRs sitting round that bump the ABI and get them in too. Also, this is going to create a massive mess, as we'll have to start a conda-forge migration, and quite a few packages don't use the conda-forge's pybind11, so...  It will be a mess. But I guess it's going to have to happen.

laramiel: There may be a non-abi bumping strategy available; if you like I can take that approach.

tests/test_thread.py
```diff
@@ -4,7 +4,6 @@
 
 import pytest
```
Skylion007, line 5: pytest is also unnecessary.

Skylion007:
Clang-tidy is still unhappy about a missing comment: (see the check failure) and the tests are failing on 2.7 and PyPy

tests/test_thread.cpp
```diff
@@ -0,0 +1,64 @@
...
+   m.def("test_no_gil", [](const IntStruct& in) {
+        IntStruct copy = in;
+        std::this_thread::sleep_for(std::chrono::milliseconds(5));
+        if (in.value != copy.value) {
+            throw std::runtime_error("Reference changed!!");
+        }
+    },
+    py::call_guard<py::gil_scoped_release>());
+
+    // NOTE: std::string_view also uses loader_life_support to ensure that
```
Skylion007, line 62: You can test string_view too if you want using the appropiate if_defs (see our current stringview tests)

include/pybind11/detail/type_caster_base.h
```diff
@@ -28,34 +28,44 @@
...
 /// Adding a patient will keep it alive up until the enclosing function returns.
 class loader_life_support {
 private:
-    void* parent = nullptr;
+    loader_life_support* parent = nullptr;
     std::unordered_set<PyObject *> keep_alive;
 
+    static loader_life_support** get_stack_pp() {
+#if defined(WITH_THREAD)
+        thread_local static loader_life_support* per_thread_stack = nullptr;
```
Skylion007, line 41: Does this really maintain compatibility though? @wjakob ? Sounds like the better thing to do would be to break the ABI.

laramiel: It does. All the older code is not thread safe, but the newer code is.

I personally think that this is probably better; it's certainly simpler and more conforming to C++ style. If thread_local doesn't work here there are larger problems in the system.

Skylion007: @laramiel But is no way to accomplish this with Python's TLS?

laramiel: It can be done, however the python TLS version is much more complicated than this version, and likely performs worse. I don't see why we'd eschew a simple elegant fix in favor of a python api call.

Skylion007: Mainly referring to: https://github.com/pybind/pybind11/issues/2765#issuecomment-911846758

laramiel: I think that thread_local is likely to generate code that's at least equivalent or better than by-hand calls to a python thread api with a lot less code, and which the compiler platform can optimize, makes it preferable. Particularly since it doesn't have to be sharable across .so boundaries.

As for the generated code, it's pretty concise, really: https://godbolt.org/z/K57oj9M4P

tests/test_thread.py
```diff
@@ -0,0 +1,33 @@
...
+    a = threading.Thread(target=loop, args=(10,))
+    b = threading.Thread(target=loop, args=(10,))
+    c = threading.Thread(target=loop, args=(10,))
+    for x in [a, b, c]:
+        x.start()
+    for x in [c, b, a]:
+        x.join()
+
+
+def test_implicit_conversion_no_gil():
```
daniel-smith, line 22: Have you been able to see this test fail without the fix applied? I've not been able to on my machine locally. I think the move away from thread pool executor to just "raw" python threads means we'll have to do something extra to capture the exceptions. See https://stackoverflow.com/questions/2829329/catch-a-threads-exception-in-the-caller-thread for a few ideas.

laramiel: Fair point; fixed.

tests/test_thread.cpp
laramiel: It shouldn't be necessary here.

laramiel:
@henryiii @Skylion007 @daniel-smith What's the next step here?

daniel-smith:
> @henryiii @Skylion007 @daniel-smith @rwgk
> What's the next step here?

I guess this is a question more for official contributors, but I'm personally happy with these changes if everyone else is happy.

include/pybind11/detail/type_caster_base.h
```diff
@@ -28,53 +28,60 @@
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
 
+
```
rwgk, line 31: Was this line added by accident?
Also below, line 84.

tests/test_thread.cpp
```diff
@@ -0,0 +1,62 @@
...
+    Copyright (c) 2017 Laramie Leavitt (Google LLC) <lar@google.com>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#include <chrono>
+#include <thread>
+
+#include <pybind11/cast.h>
```
rwgk, line 13: The pybind11 includes must be before the system includes. (I.e. simply move lines 9,10,11 after current line 17.)

rwgk (commented):
I couple very minor suggestions. I'll look more tomorrow.

The PR description still claims this is an ABI change. Is that an oversight?

tests/test_thread.cpp
laramiel: Done.

rwgk:
All: I don't think we have an active maintainer with deep background related to this PR (including myself). If you feel you have a good background, a full review (approval, optionally with short rationale) will be helpful for moving this PR forward.

include/pybind11/detail/type_caster_base.h
```diff
@@ -31,47 +31,52 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 /// A life support system for temporary objects created by `type_caster::load()`.
 /// Adding a patient will keep it alive up until the enclosing function returns.
 class loader_life_support {
+private:
+    loader_life_support* parent = nullptr;
+    std::unordered_set<PyObject *> keep_alive;
+
+    static loader_life_support** get_stack_pp() {
+#if defined(WITH_THREAD)
+        thread_local static loader_life_support* per_thread_stack = nullptr;
```
jbms, line 40: It occurred to me that an unfortunate effect of using C++ thread_local is that every single pybind11 extension module will add an additional TLS variable --- that could make the thread-local storage rather large, as there can easily be a large number of extension modules.

To avoid that, the Python TLS API could be used instead --- the key that is allocated would have to be stored in the internals struct (which would make the ABI incompatible), or if we really want to maintain ABI compatibility, could be accessed via a separate PyCapsule that is handled in the same way as the existing internals struct.  Probably it would make more sense to just break the ABI, and take the opportunity to merge in the other ABI-breaking changes, though I don't have sufficient context to really offer much judgement on that decision.

include/pybind11/detail/type_caster_base.h
```diff
@@ -31,47 +31,52 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 /// A life support system for temporary objects created by `type_caster::load()`.
 /// Adding a patient will keep it alive up until the enclosing function returns.
 class loader_life_support {
+private:
+    loader_life_support* parent = nullptr;
+    std::unordered_set<PyObject *> keep_alive;
```
jbms, line 36: Probably better to use `pybind11::object` here, as then the reference counting doesn't have to be done manually.

jbms: Note: Previously the possibility of using PySet was discussed, but that cannot be used since if the type defines a custom hash and equality function then it won't work correctly.

tests/test_thread.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <thread>
+
+#include "pybind11_tests.h"
+
+namespace py = pybind11;
+
+namespace {
+
+struct IntStruct {
+    int value;
```
jbms, line 23: Could add a destructor that modified `value` or modifies some global variable, so that we aren't relying on debugging features of the memory allocator as much.

Separately, perhaps add a note that this test should be run with asan for greater effectiveness.

include/pybind11/detail/type_caster_base.h
laramiel: And why does that concern you? I suspect that many extensions may already use thread_local variables without this concern.

The problem with using the python TLS API is that without versioning the internal data structure we're left with the same issue.

jbms: Some extensions may already use TLS, but I expect that is only a tiny fraction of them, and presumably it is providing useful functionality.

This will add an extra separate TLS variable for every pybind11 extension module (due to the use of -fvisibility=hidden, they will not be merged).  As you can easily have hundreds of extension modules loaded in a program, this is effectively adding potentially a very large number of TLS variables, and the memory usage scales with `number_of_threads_in_program * number_of_pybind11_extensions`.  Furthermore, with https://github.com/pybind/pybind11/pull/3257 taking care of the `string_view` use case, the main remaining use case is  `pybind11::implicitly_convertible`, which I think is a relatively niche feature.

Yes, the Python TLS API only helps if we ensure there is a single key shared by all pybind11 extensions.  As far as versioning the `internals` data structure, note that pybind11 already has such a mechanism, it is just a matter of bumping the abi version number --- the cost is that the type registry is not shared between different ABI versions, which may create an incompatibility between extensions compiled.  The internals data is currently accessed via a PyCapsule set as an attribute of the builtins module in the Python interpreter.  To avoid breaking the ABI, it would be possible to create a separate PyCapsule using the same mechanism that holds any additional data, in this case just the single tls key.

include/pybind11/detail/type_caster_base.h
laramiel: Unfortunately py::object is not hashable, so that would require intrusive adaptors or similar. Since the control flow here is so limited, I think that the simplest answer is to just use PyObject* here and refcount it.

tests/test_thread.cpp
laramiel: Done.

include/pybind11/detail/type_caster_base.h
jbms: Ah, good point.

include/pybind11/detail/type_caster_base.h
laramiel: You're talking about *potentially* hundreds of instances, which is admittedly an outlier.

As it stands now, I'd rather that we get this in to fix the threading issue, which I see as significant, and then we can rework it with an API version revision to use a shared python TLS variable or a pycapsule.

I'd prefer broader consensus w.r.t. the pycapsule encapsulation, though.

jbms: I don't think hundreds is an outlier.  For example, tensorflow alone seems to have ~76.

rwgk: After we finish the PyCLIF-pybind11 integration, "hundreds" is probably very common.
But I really like this plan:
> As it stands now, I'd rather that we get this in to fix the threading issue, which I see as significant, and then we can rework it with an API version revision to use a shared python TLS variable or a pycapsule.

jbms: Another issue that occurred to me is that if you call `py::cast` from translation unit `a.cc`, and that call is nested within an invocation of a pybind11-bound function defined in translation unit `b.cc`, and these two translation units have separate copies of pybind11 due to `-fvisibility=hidden`, then previously this would have worked (albeit not thread safe) but now it will fail.

For example, within Google's code base, this could occur if we define some utility function in a pybind_cc_library target that calls `py::cast`, then use that utility function from within a pybind11-bound function defined in a pybind_extension.  All of the pybind_cc_library targets will share a single copy of the thread_local, while each pybind_extension will have its own copy of the thread_local.

laramiel: If in the implementation of such a call you call `pybind11::cast` in such a way that a loader_life_support temporary is necessary then you would need to add a scoped instance of `loader_life_support` in the `-fvisibility=hidden` method implementation. I think that this is a really rare edge case, and as I said, I think that the state of the world here is vastly better than before, and we can change the implementation with an internal API bump.

jbms: I noticed that there is already a `shared_data` member of the internals struct that could be used to store the TLS key without breaking ABI --- a macro could control whether to use that or a regular data member of the internals struct.  For efficiency the TLS key could also be copied to the local_internals struct.  Then when ABI is bumped the mechanism using shared_data could be removed.

daniel-smith: Just my two cents - I would say that this is a critical bug, of which the impact is far worse than any of the aforementioned side effects/edge cases. The team I work with have taken a fork of pybind11 and applied this fix so that we can move forward - as we have a strong requirement to run IO-bound and computationally expensive functions in parallel. Without this fix, we cannot reliably do so. I would second @laramiel's suggestion to get this fix in, and then rework/optimise later if required.

jbms: I don't have any objection to merging this as is, especially given that there appears to be a relatively easy path to revising it to avoid duplicating the TLS variables without introducing an ABI break.

rwgk (approved):
My approval is based on Jeremy's and Daniel's reviews.

@Skylion007 or @henryiii, could you please also weigh in?

(One `explicit` needs to be added before we merge.)

Skylion007 (approved):

Skylion007:
Merging this in as a stop-gap. Definitely should be refactored later as per the above discussion.

tests/test_thread.py
```diff
@@ -0,0 +1,44 @@
...
+class Thread(threading.Thread):
+    def __init__(self, fn):
+        super(Thread, self).__init__()
+        self.fn = fn
+        self.e = None
+
+    def run(self):
+        try:
+            for i in range(10):
+                self.fn(i, i)
```
rainwoodman, line 17: Can I avoid implicit cast (and triggering life support) by changing this line to `self.fn(i, m.IntStruct(i))`?

daniel-smith:
Thanks everyone. Looking forward to seeing any future changes w.r.t. the issues @jbms raised.

PR #3254, opened 2021-09-09
maint(Clang-Tidy): readability-const-return

Skylion007:
## Description
* Ensure we don't use const return types since they are useless and can prevent useful optimizations (const references are allowed to be returned though).
* Fix some bad templating code that allowed

* The annoying edge case to account for here is that our Eigen cast actually has a different cast depending on whether the Eigen array is const (it controls the writability flag). For that reason, we cannot ban const return types from the caster in case the user does similarly off spec things. I have however added the NOLINTS and ensured that all other errant const return types (such as const handles) have been removed.
* Note: remove_cv_t will not strip the constness of references :)

## Suggested changelog entry:

```rst
* Enable readability clang-tidy-const-return and remove useless consts
```


include/pybind11/pybind11.h
```diff
@@ -2002,13 +2002,13 @@ template <return_value_policy Policy = return_value_policy::reference_internal,
           typename KeyType = decltype((*std::declval<Iterator>()).first),
 #endif
           typename... Extra>
-iterator make_key_iterator(Iterator first, Sentinel last, Extra &&... extra) {
+iterator make_key_iterator(Iterator first, Sentinel last, Extra &&...extra) {
```
henryiii, line 2005: Intentional?

tests/test_eigen.cpp
```diff
@@ -244,6 +245,7 @@ TEST_SUBMODULE(eigen, m) {
 
     // test_fixed, and various other tests
     m.def("fixed_r", [mat]() -> FixedMatrixR { return FixedMatrixR(mat); });
+    // NOLINTNEXTLINE(readability-const-return-type)
     m.def("fixed_r_const", [mat]() -> const FixedMatrixR { return FixedMatrixR(mat); });
```
henryiii, line 249: This used to return a non-writable numpy array, now it returns a writable one.

henryiii: *Unchanged files with check annotations (Beta)* - very nice to see!

include/pybind11/pybind11.h
Skylion007: Clang-Format. I can revert.

tests/test_eigen.cpp
Skylion007: How annoyingly offspec, Fixing this.

henryiii: It's a kind-of-handy and "cute" trick, but not at all "normal" C++!

rwgk: Is that trick explained somewhere? (I still don't know what it is.) A one-line comment here with a hint would be nice.

rwgk (approved):

rwgk (approved):
Thanks for the Eigen comment! That makes it totally clear.

henryiii (approved):

rwgk:
I'll merge this now, to then merge into to smart_holder branch, cleanup there, and then run the Google global testing overnight.

Skylion007:
Oh darn, I didn't merge this yes because there was a useless a NOLINT I hadn't got around to deleting.

rwgk:
Oops, sorry. Is that a one-line delete? Where? I could test it via the smart_holder testing, we could simply push directly to master.

Skylion007:
cast.h 1187

rwgk:
Thanks! I still have everything open (local clang-tidy testing). I'll take care of it.

rwgk:
I'm running into a snag with my grand overnight testing plan. See below.
I'm too tired right now to debug. I'll work on this tomorrow. I'll hold off removing the NOLINT until I understand the error.

```
ERROR: /google/src/cloud/rwgk/pybind11_sh_update/google3/third_party/pybind11_protobuf/BUILD:9:15: Compiling third_party/pybind11_protobuf/proto_utils.cc failed: (Exit 1) wrapped_clang failed: error executing command third_party/crosstool/v18/stable/toolchain/bin/wrapped_clang '-frandom-seed=blaze-out/k8-asan-fastbuild/bin/third_party/pybind11_protobuf/_objs/proto_utils/proto_utils.pic.o' -DABSL_ALLOCATOR_NOTHROW ... (remaining 260 arguments skipped).  [forge_remote_host=ixdf7]
third_party/pybind11_protobuf/proto_utils.cc:520:16: error: non-const lvalue reference to type 'pair<...>' cannot bind to a temporary of type 'pair<...>'
    for (auto& item : values) SetItem(item.first, item.second);
               ^    ~
third_party/pybind11/include/pybind11/pytypes.h:1470:27: note: selected 'begin' function with iterator type 'detail::dict_iterator' (aka 'generic_iterator<iterator_policies::dict_readonly>')
    detail::dict_iterator begin() const { return {*this, 0}; }
                          ^
third_party/pybind11_protobuf/proto_utils.cc:1072:14: error: non-const lvalue reference to type 'pair<...>' cannot bind to a temporary of type 'pair<...>'
  for (auto& item : kwargs_in) {
             ^    ~
third_party/pybind11/include/pybind11/pytypes.h:1470:27: note: selected 'begin' function with iterator type 'detail::dict_iterator' (aka 'generic_iterator<iterator_policies::dict_readonly>')
    detail::dict_iterator begin() const { return {*this, 0}; }
                          ^
2 errors generated.
```

rwgk:
Never mind, a quick guess worked: the proto_utils.cc code needs to be changed to `const auto&` in the two places pointed to in the error message.
Clearly your clang-tidy change is doing something good already for code health beyond pybind11 core :-)

rwgk:
I actually managed to run the Google global testing overnight. There is a pretty high number of failures (hundreds), with estimated 9 root causes. Unfortunately I cannot share the full list publicly, but one root cause points to pytorch. I'll work on this asap and will log my updates here.

PR #3257, opened 2021-09-09
Avoid use of temporary `bytes` object in string_caster for UTF-8

### LINKED ISSUE #3252: [BUG]: string_caster should not use temporary on Python >= 3.3
### Required prerequisites

### Problem description

Currently, `string_caster` always creates a temporary PyBytes object by calling `PyUnicode_AsEncodedString`.  However, in the common case of `UTF_N == 8`, on Python >= 3.3 we can instead use `PyUnicode_AsUTF8AndSize`, which manages and caches the UTF-8 encoding internally within the PyUnicode object.  If the UTF-8 representation is cached, then the encoding does not have to be done at all, avoiding an extra copy of the string.

This is particularly advantageous in the `IsView == true` case: users likely expect casting from PyUnicode to `std::string_view` to be low cost, but currently it always involves a copy of the string.  Additionally, it the `IsView == true` case has the additional cost of relying on `loader_life_support::add_patient`, which introduces additional cost and additional memory allocations (e.g. with the change in #3237, an allocation of the unordered_set bucket array on first use of loader_life_support, and an additional allocation of the node).

### Reproducible example code

_No response_

jbms:
## Description

Fixes #3252

## Suggested changelog entry:

```rst
Conversion to ``std::string`` and ``std::string_view`` now avoids making an extra copy of the data on Python >= 3.3.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -377,6 +377,22 @@ template <typename StringType, bool IsView = false> struct string_caster {
 #endif
         }
 
+#if PY_VERSION_HEX >= 0x03030000
+        // On Python >= 3.3, for UTF-8 we avoid the need for a temporary `bytes`
+        // object by using `PyUnicode_AsUTF8AndSize`.
+        if (PYBIND11_SILENCE_MSVC_C4127(UTF_N == 8)) {
+            Py_ssize_t size = -1;
+            const auto *buffer
+                = reinterpret_cast<const CharT *>(PyUnicode_AsUTF8AndSize(load_src.ptr(), &size));
+            if (!buffer) {
+                PyErr_Clear();
+                return false;
+            }
+            value = StringType(buffer, static_cast<size_t>(size));
+            return true;
+        }
+#endif
+
         auto utfNbytes = reinterpret_steal<object>(PyUnicode_AsEncodedString(
             load_src.ptr(), UTF_N == 8 ? "utf-8" : UTF_N == 16 ? "utf-16" : "utf-32", nullptr));
         if (!utfNbytes) { PyErr_Clear(); return false; }


include/pybind11/cast.h
```diff
@@ -377,6 +377,21 @@ template <typename StringType, bool IsView = false> struct string_caster {
 #endif
         }
 
+#if PY_VERSION_HEX >= 0x03030000
+        // On Python >= 3.3, for UTF-8 we avoid the need for a temporary `bytes`
+        // object by using `PyUnicode_AsUTF8AndSize`.
+        if (UTF_N == 8) {
+            Py_ssize_t size;
+            const CharT *buffer
```
Skylion007, line 385:
```suggestion
            Py_ssize_t size = 0;
            const auto *buffer
```

include/pybind11/cast.h
```diff
@@ -377,6 +377,21 @@ template <typename StringType, bool IsView = false> struct string_caster {
 #endif
         }
 
+#if PY_VERSION_HEX >= 0x03030000
+        // On Python >= 3.3, for UTF-8 we avoid the need for a temporary `bytes`
+        // object by using `PyUnicode_AsUTF8AndSize`.
+        if (UTF_N == 8) {
```
Skylion007, line 383:
```suggestion
        if (PYBIND11_SILENCE_MSVC_C4127(UTF_N == 8)) {
```
I think this is the macro that silences that annoying MSVC warning.

include/pybind11/cast.h
```diff
@@ -377,6 +377,21 @@ template <typename StringType, bool IsView = false> struct string_caster {
...
+            Py_ssize_t size;
+            const CharT *buffer
+                = reinterpret_cast<const CharT *>(PyUnicode_AsUTF8AndSize(load_src.ptr(), &size));
+            if (!buffer) {
+                PyErr_Clear();
+                return false;
+            }
+            value = StringType(buffer, static_cast<size_t>(size));
+        }
+#endif
```
Skylion007, line 393: Rest of the code after this is still run after this. You need to further modify the control flow else the value will still get overwritten below.

Skylion007 (changes requested):

include/pybind11/cast.h
jbms: Fixed.

include/pybind11/cast.h
jbms: Fixed.

include/pybind11/cast.h
jbms: Changed to use `auto`.  Since it is not actually necessary to initialize `size`, I added `NOLINT` instead to silence clang-tidy.

include/pybind11/cast.h
```diff
@@ -377,6 +377,22 @@ template <typename StringType, bool IsView = false> struct string_caster {
 #endif
         }
 
+#if PY_VERSION_HEX >= 0x03030000
+        // On Python >= 3.3, for UTF-8 we avoid the need for a temporary `bytes`
+        // object by using `PyUnicode_AsUTF8AndSize`.
+        if (PYBIND11_SILENCE_MSVC_C4127(UTF_N == 8)) {
+            Py_ssize_t size; // NOLINT(cppcoreguidelines-init-variables)
```
Skylion007, line 384: Initializing this is cheap and prevents potential UB. I would recommend setting it to obvious error vals such as size=0 or size=-1. That way it makes it more obvious when an error occurs.

jbms: In this case `size` is an output-only parameter of `PyUnicode_AsUT8AndSize`.  The idea of initializing it would be to protect against a bug in `PyUnicode_AsUT8AndSize`?

In general I don't agree with adding extra initializations as it makes it less obvious what is happening, and prevents sanitizers like memorysanitizer from detecting bugs.

However if the extra initialization is the preferred style in this code base I am happy to add it

rwgk: I agree with this. Looks much nicer, surely correct, runtime overhead certainly in the noise, if any.

rwgk (approved):

include/pybind11/cast.h
Skylion007: Yes, the extra initialization is the preferred style in this code base and the runtime overhead is minimal (if the compiler doesn't eliminate the unused_value entirely).

Skylion007: @rwgk To clarify, who are you agreeing with?

rwgk: Sorry: I agree with setting a value, `Py_ssize_t size = -1;` preferred.

jbms: Done.

Skylion007 (approved):

PR #3260, opened 2021-09-10
Fixing NOLINT mishap

rwgk:
Removing NOLINT accidentally added with PR #3254.

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1200,7 +1200,6 @@ public:
     }
 
     template <typename Return, typename Guard, typename Func>
-    // NOLINTNEXTLINE(readability-const-return-type)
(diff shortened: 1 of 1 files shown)


rwgk:
I mindlessly tried to remove all NOLINT one-by-one, running clang-tidy locally for each. Two can be removed, the three remaining are definitely needed.

Question @Skylion007: could there be valid use cases that need the extra NOLINT I removed (beyond the one you pointed out)? If yes, would it make sense to add a test case?

Skylion007:
Not really, if the Eigen tests pass then everything should be fine as the only meaningful const should be the one that goes into the caster and only as an edge case like how we used it for Eigen.

rwgk:
Awesome, thanks Aaron!

PR #3263, opened 2021-09-10
Follow-on to PR #3254, to address user code breakages.

rwgk:
Google-internal testing after merging PR #3254 came back with ~560 build breakages, going back to ~9 root causes. The tests added in this PR are reflective of the broken user code.

One way to fix the breakages is to remove `&` in user code, e.g.
```diff
-        for (auto &it : tup) {
+        for (auto it : tup) {
```

Strictly speaking that's an API break. Managing such sprawling fixes has a high cost.

To avoid the API break, this PR restores the `const` removed from pytypes.h in PR #3254 and adds 6 `// NOLINTNEXTLINE(readability-const-return-type)`. This decouples activating the clang-tidy feature from the API break.

### DIFF
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -462,10 +462,17 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("weakref_from_object_and_function",
           [](py::object o, py::function f) { return py::weakref(std::move(o), std::move(f)); });
 
+// See https://github.com/pybind/pybind11/pull/3263 for background.
+#if (defined(__clang__) && defined(__apple_build_version__)) || defined(PYPY_VERSION)
+#    define PYBIND11_CONST_FOR_STRICT_PLATFORMS const
+#else
+#    define PYBIND11_CONST_FOR_STRICT_PLATFORMS
+#endif
+
     m.def("tuple_iterator", []() {
         auto tup = py::make_tuple(5, 7);
         int tup_sum = 0;
-        for (auto &it : tup) {
+        for (PYBIND11_CONST_FOR_STRICT_PLATFORMS auto &it : tup) {
             tup_sum += it.cast<int>();
         }
         return tup_sum;
@@ -476,7 +483,7 @@ TEST_SUBMODULE(pytypes, m) {
         dct[py::int_(3)] = 5;
         dct[py::int_(7)] = 11;
         int kv_sum = 0;
-        for (auto &it : dct) {
+        for (PYBIND11_CONST_FOR_STRICT_PLATFORMS auto &it : dct) {
             kv_sum += it.first.cast<int>() * 100 + it.second.cast<int>();
         }
         return kv_sum;
@@ -484,11 +491,13 @@ TEST_SUBMODULE(pytypes, m) {
 
     m.def("passed_iterator", [](const py::iterator &py_it) {
         int elem_sum = 0;
-        for (auto &it : py_it) {
+        for (PYBIND11_CONST_FOR_STRICT_PLATFORMS auto &it : py_it) {
             elem_sum += it.cast<int>();
         }
         return elem_sum;
     });
+
+#undef PYBIND11_CONST_FOR_STRICT_PLATFORMS
 
     // Tests below this line are for pybind11 IMPLEMENTATION DETAILS:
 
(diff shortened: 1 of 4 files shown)


rwgk:
It turns out the "tests reflective of user code" do not work on all platforms, even when added to the v2.7 branch. See below.

Failing platforms are: all macos-latest, all PyPy (9 total, master or v2.7).

This suggests such user code is not portable and needs to be fixed. Unfortunately that is (in my situation) or can be (in general) very disruptive.

```
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:441:20: error: loop variable 'it' is always a copy because the range of type 'pybind11::tuple' does not return a reference [-Werror,-Wrange-loop-analysis]
        for (auto &it : tup) {
                   ^
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:441:14: note: use non-reference type 'const pybind11::handle'
        for (auto &it : tup) {
             ^~~~~~~~~~
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:452:20: error: loop variable 'it' is always a copy because the range of type 'py::dict' does not return a reference [-Werror,-Wrange-loop-analysis]
        for (auto &it : dct) {
                   ^
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:452:14: note: use non-reference type 'const std::__1::pair<pybind11::handle, pybind11::handle>'
        for (auto &it : dct) {
             ^~~~~~~~~~
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:460:20: error: loop variable 'it' is always a copy because the range of type 'const py::iterator' does not return a reference [-Werror,-Wrange-loop-analysis]
        for (auto &it : py_it) {
                   ^
/Users/runner/work/pybind11/pybind11/tests/test_pytypes.cpp:460:14: note: use non-reference type 'const pybind11::handle'
        for (auto &it : py_it) {
             ^~~~~~~~~~
```

rwgk:
Hi @Skylion007, if you get a chance to look, that would be great. I'm not entirely sure what to do. I'm leaning towards `#ifdef`ing out the new tests on platforms where they don't work.

rwgk:
I created a Google-internal bug for the breakages. Current thinking:

* I'll apply a local patch to pytypes.h simply adding back the `const`. That will allow me to keep our internal pybind11 copy current with github master/smart_holder.
* In the background I will work on patching the user code that breaks when the `const` are removed. This could take a week or so to trickle through the system.
* When all user code is cleaned up, I'll purge my local patch.

Beyond Google: we need to explain the situation clearly in the upgrade guide for PR #3254.

But let me try my plan above first, to see if it works out, and what we learn from it.

tests/test_pytypes.cpp
```diff
@@ -462,6 +462,34 @@ TEST_SUBMODULE(pytypes, m) {
...
+        int tup_sum = 0;
+        for (auto &it : tup) {
+            tup_sum += it.cast<int>();
+        }
+        return tup_sum;
+    });
+
+    m.def("dict_iterator", []() {
+        py::dict dct;
+        dct[py::int_(3)] = 5;
```
Skylion007, line 476: Okay, so this is a bug with storing const values in keys?

Skylion007 (approved):

tests/test_pytypes.cpp
rwgk: I'm not sure I understand the question, but I'm thinking the answer is no.
This code here is just to have a couple random dict entries. The actual lines that matter are only the three `for (auto &it : xxx)`.
IIUC, a new `handle` is constructed in each iteration, a temporary. `auto &it` creates a reference to a temporary. The stricter compilers that reject this code even if we keep the `const` are doing the "right thing" IMO.
For us the critical question is: do we want to make pybind11 a forcing function (forcing clean user code by removing the `const`)? On that one I'm ambivalent, leaning to no, to not have to deal with the fallout (is that lazy or smart?).
The next question: what do we want to do about these tests here?
If we remove them from this PR, it will not be obvious at all why we need the NOLINTs.
If we keep them, we need ifdefs do make them work with the stricter compilers.
Hm ... writing this I'm beginning to think we just need to insert `const` for those strict compilers. It'll look a little convoluted, but then we have all situations covered in code, as concise as it can be made.

rwgk:
Hi @Skylion007, could you please take another look?

This PR is still very similar to the version you approved already, but in tests/test_pytypes.cpp I added a fairly chunky `#ifdef` + comments block, to make the new tests build on all platforms, and to explain the situation. All changes are also labeled with `PR #3263`, for easy reference, and to make it easy to undo them if we decide to do so in the future.

Skylion007 (approved):

Issue #1177, opened 2017-11-09
There can be only one (enum instance)


pschella:
Currently enums compare equal with `==` but not with `is`.
I think it would be good if enums would be singletons like `None` instead.

televator:
This is still an issue with 2.2.4 and very error-prune as the python doc [explicitly encourages to check enum values for identity](https://docs.python.org/3/library/enum.html#comparisons).

Minimal example to reproduce
```
#include <pybind11/pybind11.h>

namespace py = pybind11;

PYBIND11_MODULE(example_py, m) {

    enum class ScopedEnum {
        Two = 2,
        Three
    };
    m.def("return_two", [](){return ScopedEnum::Two;});
    py::enum_<ScopedEnum>(m, "ScopedEnum")
        .value("Two", ScopedEnum::Two)
        .value("Three", ScopedEnum::Three);
}
```
produces this situation
```
In [1]: import example_py

In [2]: x = example_py.ScopedEnum.Two

In [3]: y = example_py.ScopedEnum.Two

In [4]: z = example_py.return_two()

In [5]: x is y # BUT...
Out[5]: True

In [6]: x is z
Out[6]: False

In [7]: x == z
Out[7]: True
```

lucmans:
@rwgk Were still running into this issue on pybind11 2.9.1 with the minimal repro provided by @televator. Any update on this?

rwgk:
> @rwgk We're still running into this issue on pybind11 2.9.1 with the minimal repro provided by @televator. Any update on this?

I could be wrong, but I'm operating under the assumption that the `py::enum_` behavior is practically unchangeable/unfixable, merely because it's in wide use.

Therefore I added `py::native_enum` as an **alternative** here:

* https://github.com/google/pywrapcc/blob/1a9ac015439b7cecafffae039d138223a10dd5a1/include/pybind11/native_enum.h

* https://github.com/google/pywrapcc/blob/1a9ac015439b7cecafffae039d138223a10dd5a1/tests/test_native_enum.py

Note that that's in another repo, although it's actually more-or-less just another branch.

Could this be merged into pybind11 master? — In theory yes, easily (`git merge` will just work), in practice it seems extremely unlikely to me. pybind11 master has been in extremely conservative mode for a very long time.

You could use the pywrapcc branch and it'll fully interoperate with pybind11 master. There is just no pip-installable release.

lucmans:
[...]

Issue #3253, opened 2021-09-09
[BUG]: Overriding __new__ (or type-level __call__) for a pybind11 class should be supported


jbms:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

It would be nice if, for a given pybind11-defined class `Foo`, there were a mechanism for overriding the object returned by a call to `Foo(...)`.

In Python, this can be done by overriding `__new__` or by using a custom metaclass with an overridden `__call__` method:

https://stackoverflow.com/questions/6760685/creating-a-singleton-in-python

One case where this would be useful is to be able to return a reference to an existing object.  Currently, even if we use `std::shared_ptr` as the holder type and return a reference to an existing object from a `pybind11::init` function, we will end up with a new Python object, which is unfortunate.

As far as I am aware, that problem is impossible to remedy from `__init__`, since by the time `__init__` is called, the Python object has already been created.  Instead, we must either use `__new__` or the metaclass `__call__`.

Currently, pybind11 does not support defining `__new__`.  Attempting to define a method named `"__new__"` results in an error due to this line:

https://github.com/pybind/pybind11/blob/6abf2baa62063a39ccc4a9535351de571004ddd4/include/pybind11/pybind11.h#L420

The sibling (the existing `__new__` function) is not a PyCapsule, which causes this to fail.  Incidentally, `pybind11::capsule::{get,set}_pointer` have a bug, in that they call `pybind11_fail` without calling `PyErr_Clear()`, which leads to an assertion failure elsewhere due to an unexpected error indicator state.

I suspect that it may take a decent amount of work beyond addressing this immediate error to make `__new__` work, though.

An alternative workaround is to define a metaclass.  That works, but is non-trivial since pybind11 itself provides no support for defining metaclasses, and you must take care to inherit from pybind11's own metaclass.

Skylion007:
PR welcome on this front. 
[...]

Issue #1693, opened 2019-02-13
Support classmethod


asadchev:
I have a workaround to create classmethods in PyBind.
```
  template<class Func, typename ... Args>
  pybind11::object classmethod(Func f, Args ... args) {
    pybind11::object cf = pybind11::cpp_function(f, args...);
    return pybind11::object(PyClassMethod_New(cf.ptr()), true);
  }
```
used like this
```
void foo(py::object cls, ...);
cls.attr("foo") = classmethod(&foo);
```
Be useful to have such mechanism in pybind proper

henryiii:
@YannickJadoul were you going to look into this? This is a possible workaround without calling CPython API:

```cpp
class Example {
    int x;
    Example(int x) : x(x) {}
public:
    static Example make_x(int y) { return Example(y); }
    int get_x() const { return x; }
};

PYBIND11_MODULE(tmp, m) {
    py::class_<Example>(m, "Example")
        .def("get_x", &Example::get_x)
        .def_property_readonly_static("make_x", [](py::object){
              return py::cpp_function([](int x){return Example::make_x(3);
        });});
}
```

YannickJadoul:
@henryiii, I was. I thought it was easy as the `staticmethod` PR I contributed a few years ago. But the question is a bit what to do with the first argument to a `classmethod`. Should it be a `py::type`?

Skylion007:
@YannickJadoul Pretty sure it should be a py::type.

Skylion007:
> @YannickJadoul were you going to look into this? This is a possible workaround without calling CPython API:
> 
> ```c++
> class Example {
>     int x;
>     Example(int x) : x(x) {}
> public:
>     static Example make_x(int y) { return Example(y); }
>     int get_x() const { return x; }
> };
> 
> PYBIND11_MODULE(tmp, m) {
>     py::class_<Example>(m, "Example")
>         .def("get_x", &Example::get_x)
>         .def_property_readonly_static("make_x", [](py::object){
>               return py::cpp_function([](int x){return Example::make_x(3);
>         });});
> }
> ```

If follow the implementation of property_readonly_static though it does eventually call the C++ API. We should just officially support it IMO as this is a lot of indirection and all our static methods call attr anyway.

Skylion007:
[...]

PR #3265, opened 2021-09-11
Enable defining custom __new__

Skylion007:
## Description

* This solves the bugs preventing us from defining our own __new__ methods on classes that was brought up in #3253 .

## Suggested changelog entry:

```rst
* Enable defining custom __new__ methods on classes.
* Fixes bug preventing overriding methods if they have non-pybind11 siblings.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -416,7 +416,7 @@ protected:
         detail::function_record *chain = nullptr, *chain_start = rec;
         if (rec->sibling) {
             if (PyCFunction_Check(rec->sibling.ptr())) {
-                auto self = PyCFunction_GET_SELF(rec->sibling.ptr());
+                auto *self = PyCFunction_GET_SELF(rec->sibling.ptr());
                 capsule rec_capsule;
                 if (isinstance<capsule>(self)) {
                     rec_capsule = reinterpret_borrow<capsule>(self);
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -322,7 +322,6 @@ protected:
         rec->is_constructor
             = (strcmp(rec->name, "__init__") == 0) || (strcmp(rec->name, "__setstate__") == 0);
 
-#define PYBIND11_DISABLE_NEW_STYLE_INIT_WARNING
 #if !defined(NDEBUG) && !defined(PYBIND11_DISABLE_NEW_STYLE_INIT_WARNING)
         if (rec->is_constructor && !rec->is_new_style_constructor) {
             const auto class_name = detail::get_fully_qualified_tp_name((PyTypeObject *) rec->scope.ptr());
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -322,6 +322,7 @@ protected:
         rec->is_constructor
             = (strcmp(rec->name, "__init__") == 0) || (strcmp(rec->name, "__setstate__") == 0);
 
+#define PYBIND11_DISABLE_NEW_STYLE_INIT_WARNING
 #if !defined(NDEBUG) && !defined(PYBIND11_DISABLE_NEW_STYLE_INIT_WARNING)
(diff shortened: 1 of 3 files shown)


rwgk:
How did I arrive at https://github.com/pybind/pybind11/pull/3265/commits/194033e40433fcdfd381837bb228a73629f6584e?

When testing in the Google environment, I ran into this `assert`:

https://github.com/python/cpython/blob/d5650a1738fe34f6e1db4af5f4c4edb7cae90a36/Objects/typeobject.c#L3117

The error wasn't obvious at all, therefore I inserted this right before that `assert`:
```
    if (PyErr_Occurred()) {                                                     
      PyErr_Print();                                                            
      assert(0);                                                                
    } 
```
That way I got to see the `FutureWarning`. After disabling it with the unconditional `#define`, all tests ran ASAN clean.

While looking around, I saw that all new-style `"__init__"` return a pointer (from a placement `new`). Therefore I changed the new code accordingly, before I discovered the `FutureWarning`. I didn't go back to test if it also works when returning a reference (I believe returning a pointer is preferable regardless, although I don't have deep insights).

That's all I know. 🙂

Skylion007:
Defining a custom __new__ function for enums could help close #1177. Also we are putting a class method in the tests now so we might want to add a class method shim as mentioned in #1693 now to py_class as the current way of defining class methods is bulky and not very intuitive. 

We don't seem to have a way to check the py::type of a the C++ class for the first argument though, which is a bit annoying.

Skylion007:
I think even if we reduce the scope to calling attr("get") it would be a useful shim.

Skylion007:
This syntax is at the very least not terrible until we have proper classmethod support.

include/pybind11/pybind11.h
```diff
@@ -416,7 +416,13 @@ class cpp_function : public function {
...
         if (rec->sibling) {
             if (PyCFunction_Check(rec->sibling.ptr())) {
-                auto rec_capsule = reinterpret_borrow<capsule>(PyCFunction_GET_SELF(rec->sibling.ptr()));
+                auto *self = PyCFunction_GET_SELF(rec->sibling.ptr());
+                capsule rec_capsule;
+                if (isinstance<capsule>(self)) {
+                    rec_capsule = reinterpret_borrow<capsule>(self);
+                } else {
+                    rec_capsule = capsule(self);
+                }
```
henryiii, line 425:
```suggestion
                rec_capsule = isinstance<capsule>(self) ? reinterpret_borrow<capsule>(self) : capsule(self);
```

henryiii, line 425:
```suggestion
                capsule rec_capsule = isinstance<capsule>(self) ? reinterpret_borrow<capsule>(self) : capsule(self);
```

henryiii (approved):

Skylion007:
I actually just realized those PTRs in test should probably be ref since they can't be null. Probably can go back and fix them when we write the actual documentation for this feature at some point @henryiii along with proper classmethod support perhaps

tests/test_class.cpp
```diff
@@ -47,10 +47,25 @@ TEST_SUBMODULE(class_, m) {
...
+            return ptr;
+        }
+        ~NoConstructorNew() { print_destroyed(this); }
+    };
 
     py::class_<NoConstructor>(m, "NoConstructor")
         .def_static("new_instance", &NoConstructor::new_instance, "Return an instance");
 
+    py::class_<NoConstructorNew>(m, "NoConstructorNew")
+        .def(py::init([](NoConstructorNew *self) { return self; })) // Need a NOOP __init__
```
Skylion007, line 66: I also wish there was a cleaner way to not have to implement `__init__` if __new__ is defined, but I can't figure out how to hack to the metaclass not to complain.

rwgk: Maybe something like `.def_new(...)` or `.def(py::new(...))` that wraps the two methods you have right now?

Issue #3269, opened 2021-09-15
[BUG]: pybind11 broken with the default gcc on ubuntu-trusty, linuxmint-17


mkoeppe:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

In the SageMath 9.5 development series we have just upgraded to `pybind11` 2.7.1 and have noticed that building `scipy` 1.6.3 now fails with the default compilers on the platforms `linuxmint-17` (`gcc 4.8.4-2ubuntu1~14.04.3`) and `ubuntu-trusty` (`gcc 4.8.4-2ubuntu1~14.04.4`). pybind11's README still claims "GCC 4.8 or newer" is supported, so this seems to be a regression.

Likely caused by https://github.com/pybind/pybind11/pull/2956 . The complicated condition for `__GLIBCXX__`, 
```
#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
```
not fully explained in that PR, misses the value on the two failing platforms, which is `20150426`.

Looks like the pybind11 CI does not catch this -- you test `centos-7`, which uses gcc 4.8.5 (`__GLIBCXX__ = 20150623`), which works.

Tracked in https://trac.sagemath.org/ticket/32515

mkoeppe:
Proposed fix in #3270

Skylion007:
Closed by #3270

PR #3270, opened 2021-09-15
include/pybind11/numpy.h: gcc 4.8.4 does not have is_trivially_copyable

mkoeppe:
## Description

Proposed fix for (#3269), tested on `ubuntu-trusty`.

## Suggested changelog entry:

```rst
Restore compatibility with gcc 4.8.4 as distributed by ubuntu-trusty, linuxmint-17.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -319,7 +319,7 @@ template <typename T> using remove_all_extents_t = typename array_info<T>::type;
 
 template <typename T> using is_pod_struct = all_of<
     std::is_standard_layout<T>,     // since we're accessing directly in memory we need a standard layout type
-#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
+#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150426 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
     // libstdc++ < 5 (including versions 4.8.5, 4.9.3 and 4.9.4 which were released after 5)
     // don't implement is_trivially_copyable, so approximate it
     std::is_trivially_destructible<T>,


include/pybind11/numpy.h
```diff
@@ -319,7 +319,7 @@ template <typename T> using remove_all_extents_t = typename array_info<T>::type;
 
 template <typename T> using is_pod_struct = all_of<
     std::is_standard_layout<T>,     // since we're accessing directly in memory we need a standard layout type
-#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
+#if defined(__GLIBCXX__) && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150426 || __GLIBCXX__ == 20150623 || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
```
Skylion007, line 322: Yikes. Disappointed there isn't a better way to query this information.

Skylion007 (approved):

include/pybind11/numpy.h
rwgk: +1 but I'll merge this now to get us back to normal.

Question, to maybe do something different later in another PR: would there anything be lost if we simply made this
```
#if defined(__GLIBCXX__) && __GLIBCXX__ < 20160803
```

rwgk (approved):

rwgk:
TBH: I missed that this was against v2.7 (I assumed master).
@henryiii what's the best way to handle this? Also cherry-pick to master?

henryiii:
> Also cherry-pick to master?

Yep.

include/pybind11/numpy.h
henryiii: I think the dates are the release dates, even if they are not sequential?

henryiii:
Also, @mkoeppe  it's best to target master, and then ask for it to be backported. If you do target a branch, I'd start the PR title with the branch name (usually in `[]`, but I've seen other styles).

mkoeppe:
Not exactly practical to ask downstream projects to test with your master branch -- we are using a release and our concern is to fix that. 

By the way, I noticed that the directory layout of an sdist of the project differs from the directory layout of your repo, forcing me to hand-edit the patch. That's a bit unusual and something that you may want to fix in the medium term.

henryiii:
This will not fix the release unless we make a new release. Generally, either make a PR against master or clearly note you are doing something non-standard. You are welcome to ask for a backport, we will do it if we can. If you make a patch to an upstream project, you should try to target the active development branch first. And clearly note if you don't.

If someone hadn't caught this, 2.8 would be missing this fix, even if there is a 2.7.3 (or whatever is next).

Not a huge criticism, and very happy for the patch, just a recommendation.

The SDist has the CMake install instead of the source, so that the SDist is installable without requiring CMake. However the directory structure should be somewhat close, the main (only?) difference is the includes are already placed where they need to go for Python.

mkoeppe:
Thanks for the explanations.

PR #3287, opened 2021-09-20
Add `custom_type_setup` attribute

jbms:
This allows for custom modifications to the PyHeapTypeObject prior to
calling `PyType_Ready`.  This may be used, for example, to define
`tp_traverse` and `tp_clear` functions.

## Suggested changelog entry:

```rst
`pybind11::custom_type_setup` was added, for customizing the `PyHeapTypeObject` corresponding to a class, which may be useful for enabling garbage collection support, among other things.
```

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1261,3 +1261,37 @@ object, just like ``type(ob)`` in Python.
     Other types, like ``py::type::of<int>()``, do not work, see :ref:`type-conversions`.
 
 .. versionadded:: 2.6
+
+Custom type setup
+=================
+
+For advanced use cases, such as enabling garbage collection support, you may
+wish to directly manipulate the `PyHeapTypeObject` corresponding to a
+``py::class_`` definition.
+
+You can do that using ``py::custom_type_setup``:
+
+.. code-block:: cpp
+
+   struct OwnsPythonObjects {
+       py::object value = py::none();
+   };
+   py::class_<OwnsPythonObjects> cls(
+       m, "OwnsPythonObjects", py::custom_type_setup([](PyHeapTypeObject *heap_type) {
+           auto *type = &heap_type->ht_type;
+           type->tp_flags |= Py_TPFLAGS_HAVE_GC;
+           type->tp_traverse = [](PyObject *self_base, visitproc visit, void *arg) {
+               auto &self = py::cast<OwnsPythonObjects&>(py::handle(self_base));
+               Py_VISIT(self.value.ptr());
+               return 0;
+           };
+           type->tp_clear = [](PyObject *self_base) {
+               auto &self = py::cast<OwnsPythonObjects&>(py::handle(self_base));
+               self.value = py::none();
+               return 0;
+           };
+       }));
+   cls.def(py::init<>());
+   cls.def_readwrite("value", &OwnsPythonObjects::value);
+
+.. versionadded:: 2.8
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -12,6 +12,8 @@
 
 #include "cast.h"
 
+#include <functional>
+
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 /// \addtogroup annotations
@@ -77,6 +79,23 @@ struct metaclass {
 
     /// Override pybind11's default metaclass
     explicit metaclass(handle value) : value(value) { }
+};
+
+/// Specifies a custom callback with signature `void (PyHeapTypeObject*)` that
+/// may be used to customize the Python type.
+///
+/// The callback is invoked immediately before `PyType_Ready`.
+///
+/// Note: This is an advanced interface, and uses of it may require changes to
+/// work with later versions of pybind11.  You may wish to consult the
+/// implementation of `make_new_python_type` in `detail/classes.h` to understand
+/// the context in which the callback will be run.
+struct custom_type_setup {
+    using callback = std::function<void(PyHeapTypeObject *heap_type)>;
+
+    explicit custom_type_setup(callback value) : value(std::move(value)) {}
+
+    callback value;
 };
 
 /// Annotation that marks a class as local to the module:
@@ -271,6 +290,9 @@ struct type_record {
 
     /// Custom metaclass (optional)
     handle metaclass;
+
+    /// Custom type setup.
+    custom_type_setup::callback custom_type_setup_callback;
 
     /// Multiple inheritance marker
     bool multiple_inheritance : 1;
@@ -474,6 +496,13 @@ struct process_attribute<multiple_inheritance> : process_attribute_default<multi
 template <>
 struct process_attribute<dynamic_attr> : process_attribute_default<dynamic_attr> {
     static void init(const dynamic_attr &, type_record *r) { r->dynamic_attr = true; }
+};
+
+template <>
+struct process_attribute<custom_type_setup> {
+    static void init(const custom_type_setup &value, type_record *r) {
+        r->custom_type_setup_callback = value.value;
+    }
 };
 
 template <>
--- a/tests/test_custom_type_setup.cpp
+++ b/tests/test_custom_type_setup.cpp
@@ -1,0 +1,41 @@
+/*
+    tests/test_custom_type_setup.cpp -- Tests `pybind11::custom_type_setup`
+
+    Copyright (c) Google LLC
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#include <pybind11/pybind11.h>
+
+#include "pybind11_tests.h"
+
+namespace py = pybind11;
+
+namespace {
+
+struct OwnsPythonObjects {
+    py::object value = py::none();
+};
+} // namespace
+
+TEST_SUBMODULE(custom_type_setup, m) {
+    py::class_<OwnsPythonObjects> cls(
+        m, "OwnsPythonObjects", py::custom_type_setup([](PyHeapTypeObject *heap_type) {
+            auto *type = &heap_type->ht_type;
+            type->tp_flags |= Py_TPFLAGS_HAVE_GC;
+            type->tp_traverse = [](PyObject *self_base, visitproc visit, void *arg) {
+                auto &self = py::cast<OwnsPythonObjects &>(py::handle(self_base));
+                Py_VISIT(self.value.ptr());
+                return 0;
+            };
+            type->tp_clear = [](PyObject *self_base) {
+                auto &self = py::cast<OwnsPythonObjects &>(py::handle(self_base));
+                self.value = py::none();
+                return 0;
+            };
+        }));
+    cls.def(py::init<>());
+    cls.def_readwrite("value", &OwnsPythonObjects::value);
+}
--- a/tests/test_custom_type_setup.py
+++ b/tests/test_custom_type_setup.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+import gc
+import weakref
+
+import pytest
+
+import env  # noqa: F401
+from pybind11_tests import custom_type_setup as m
+
+
+@pytest.fixture
+def gc_tester():
+    """Tests that an object is garbage collected.
+
+    Assumes that any unreferenced objects are fully collected after calling
+    `gc.collect()`.  That is true on CPython, but does not appear to reliably
+    hold on PyPy.
+    """
+
+    weak_refs = []
+
+    def add_ref(obj):
+        # PyPy does not support `gc.is_tracked`.
+        if hasattr(gc, "is_tracked"):
+            assert gc.is_tracked(obj)
+        weak_refs.append(weakref.ref(obj))
+
+    yield add_ref
+
+    gc.collect()
+    for ref in weak_refs:
+        assert ref() is None
+
+
+# PyPy does not seem to reliably garbage collect.
+@pytest.mark.skipif("env.PYPY")
+def test_self_cycle(gc_tester):
+    obj = m.OwnsPythonObjects()
+    obj.value = obj
+    gc_tester(obj)
+
+
+# PyPy does not seem to reliably garbage collect.
+@pytest.mark.skipif("env.PYPY")
+def test_indirect_cycle(gc_tester):
+    obj = m.OwnsPythonObjects()
+    obj_list = [obj]
+    obj.value = obj_list
+    gc_tester(obj)
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -683,11 +683,13 @@ inline PyObject* make_new_python_type(const type_record &rec) {
     if (rec.buffer_protocol)
         enable_buffer_protocol(heap_type);
 
+    if (rec.custom_type_setup_callback)
+        rec.custom_type_setup_callback(heap_type);
+
     if (PyType_Ready(type) < 0)
         pybind11_fail(std::string(rec.name) + ": PyType_Ready failed (" + error_string() + ")!");
 
-    assert(rec.dynamic_attr ? PyType_HasFeature(type, Py_TPFLAGS_HAVE_GC)
-                            : !PyType_HasFeature(type, Py_TPFLAGS_HAVE_GC));
+    assert(!rec.dynamic_attr || PyType_HasFeature(type, Py_TPFLAGS_HAVE_GC));
 
     /* Register type with the parent scope */
     if (rec.scope)
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -259,8 +259,11 @@ public:
 
     object& operator=(const object &other) {
         other.inc_ref();
-        dec_ref();
+        // Use temporary variable to ensure `*this` remains valid while
+        // `Py_XDECREF` executes, in case `*this` is accessible from Python.
+        handle temp(m_ptr);
         m_ptr = other.m_ptr;
+        temp.dec_ref();
         return *this;
     }
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -104,6 +104,7 @@ set(PYBIND11_TEST_FILES
     test_constants_and_functions.cpp
     test_copy_move.cpp
     test_custom_type_casters.cpp
+    test_custom_type_setup.cpp
     test_docstring_options.cpp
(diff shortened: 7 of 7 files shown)


rwgk (commented):
Initially the suggested changelog entry made me wonder: does `custom_type_setup` exist already? Is this PR enhancing the functionality?
But no, it's new.
How about:

`pybind11::custom_type_setup` was added, for customizing the `PyHeapTypeObject` corresponding to a `class_`, which may be useful for enabling garbage collection support, among other things.

jbms:
Thanks.

An alternative is to add direct support for garbage collection, similar to the other protocols.  But still I think `custom_type_setup` is useful as an escape hatch, so that users can implement any protocol they like without waiting for pybind11 to specifically support it.

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Creates a view that references an existing function.
+    ///
+    /// The constructed view is only valid for the lifetime of `f`.
+    template <typename F,
+              typename = enable_if_t<
+                  (!std::is_same<remove_reference_t<remove_cv_t<F>>, function_view>::value
+                   && is_expected_return_type<R>::template compatible_with<
+                       call_result_t<F, Arg...>>::value)>>
+    function_view(F &&f) noexcept // NOLINT
```
Skylion007, line 28: Please fully qualify the NOLINT

jbms: Done.  Thanks.

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,53 @@
...
+class function_view<R(Arg...)> {
+public:
+    function_view() = default;
+
+    /// Creates a view that references an existing function.
+    ///
+    /// The constructed view is only valid for the lifetime of `f`.
+    template <typename F,
+              typename = enable_if_t<
+                  (!std::is_same<remove_reference_t<remove_cv_t<F>>, function_view>::value
```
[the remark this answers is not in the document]
jbms: Done.  Good catch about the order of remove_cv and remove_reference.

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,59 @@
...
+    ///
+    /// The constructed view is only valid for the lifetime of `f`.
+    ///
+    /// Requires that `F` is a function object type that is callable with
+    /// arguments `(Args...)` and has a return type convertible to `R`.
+    template <typename F,
+              typename = enable_if_t<
+                  // Prevent construction from the same `function_view` type, to avoid inefficient
+                  // double wrapping that also commonly leads to dangling references.
+                  (!std::is_same<remove_cv_t<remove_reference_t<F>>, function_view>::value
```
Skylion007, line 30: Any reason not to use std::decay here?

jbms: That is fine, it just does extra conversions that aren't applicable, like array decaying and function reference decaying.  I think remove_cvref more precisely reflects what is needed here.

tests/test_custom_type_setup.py
```diff
@@ -0,0 +1,45 @@
...
+            assert gc.is_tracked(obj)
+        weak_refs.append(weakref.ref(obj))
+
+    yield add_ref
+
+    gc.collect()
+    if hasattr(gc, "collect_step"):
+        # PyPy may require additional encouragement to actually collect the
+        # garbage.
+        for i in range(100):
```
Skylion007, line 29:
```suggestion
        for _ in range(100):
```
will stop the linter from complaining

docs/advanced/classes.rst
```diff
@@ -1259,3 +1259,40 @@ object, just like ``type(ob)`` in Python.
...
+Custom type setup
+=================
+
+For advanced use cases, such as setting certain type object slots, you may wish
+to directly manipulate the `PyHeapTypeObject` corresponding to a ``py::class_``
+definition.
+
+You can do that using ``py::custom_type_setup``:
+
+.. code-block:: cpp
```
rwgk, line 1272: I'm really worried about this very large & complex block of code in the documentation. Bit rot seems almost guaranteed. Could this be moved to the test, with a simple pointer here? (I wish we had an examples directory in this repo, but that'd be a project on its own.)

jbms:
All tests pass now.

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "common.h"
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+template <typename Signature>
+class function_view;
+
+template <typename R, typename... Arg>
```
rwgk, line 15: Would it make sense to explain how this compares to `std::function`?

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,59 @@
...
+
+private:
+    template <typename F>
+    static R wrapper(void *obj, Arg... arg) {
+        return (*static_cast<typename std::add_pointer<F>::type>(obj))(
+            static_cast<Arg &&>(arg)...);
+    }
+
+    R (*erased_fn_)(void *, Arg...) = nullptr;
+    const void *erased_obj_;
```
rwgk, line 55: `type_erased_obj_`?
If that's correct, I'd definitely spend the extra few characters for clarity.

tests/test_custom_type_setup.py
```diff
@@ -0,0 +1,50 @@
...
+    def add_ref(obj):
+        # PyPy does not support `gc.is_tracked`.
+        if hasattr(gc, "is_tracked"):
+            assert gc.is_tracked(obj)
+        weak_refs.append(weakref.ref(obj))
+
+    yield add_ref
+
+    gc.collect()
+    if hasattr(gc, "collect_step"):
```
rwgk, line 27: Is this still needed?

tests/test_custom_type_setup.py
```diff
@@ -0,0 +1,50 @@
...
+
+import pytest
+
+import env  # noqa: F401
+from pybind11_tests import custom_type_setup as m
+
+
+@pytest.fixture
+def gc_tester():
+    """Tests that an object is garbage collected."""
```
rwgk, line 14: Might be helpful to add something like:

Assumes that the Python VM including garbage collector are designed to satisfy this test. This is generally true for CPython, but is known to not be true for PyPy.

tests/test_custom_type_setup.py
```diff
@@ -0,0 +1,50 @@
...
+
+    weak_refs = []
+
+    def add_ref(obj):
+        # PyPy does not support `gc.is_tracked`.
```
rwgk, line 19: Is this still needed? (PyPy seems to be skipped entirely.)

rwgk (commented):
I believe this is a useful addition, although I also wonder if this will lead to `N` almost correct implementations for adding GC support (in other repos), instead of `1` firmly correct implementation (in this repo). — On balance I'm coming down on the side of "let it grow and maybe clean up later" vs slipping into over-engineering.

tests/test_custom_type_setup.py
jbms: Okay, I removed it.  It seemed to increase reliability on pypy, but ultimately I could not make it reliable so I disabled the tests there.

include/pybind11/detail/function_view.h
jbms: Done.

tests/test_custom_type_setup.py
jbms: Done.

tests/test_custom_type_setup.py
jbms: I removed PyPy because it is unreliable but it still seemed somewhat useful to retain those workarounds in case someone wants to make the test more reliable in the future.

docs/advanced/classes.rst
jbms: This is also in a test.  But it seemed useful to provide a working example.  I was able to simplify it slightly since I also fixed `object::operator=`.

jbms:
It looks like all tests now pass --- the one failure was just a spurious one.

include/pybind11/detail/function_view.h
```diff
@@ -0,0 +1,63 @@
...
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+template <typename Signature>
+class function_view;
+
+// Type-erased unowned reference to a function/function object.
+//
+// This is similar to `std::function`, but is much lighter weight (size of two
```
rwgk, line 17: With 64-bit ubuntu & clang++:
* `sizeof(py::detail::type_record)` is 15 pointers.
* `sizeof(std::function<void()>)` is 4 pointers.

That makes me think the optimization is 17 pointers vs 19 pointers.

I wonder about the size of the generated machine code: if `std::function` is reused, could that be a net saving, even though sizeof(std::function) is 4 vs sizeof(function_view) is 2?

I noticed this existing comment in attr.h:
```
/// Special data structure which (temporarily) holds metadata about a bound class
struct type_record {                                                            
```

IIUC, if we reuse `std::function` we'd have the same functionality with less code ("less is more").

Do we have measurements that could justify adding this extra code?

docs/advanced/classes.rst
rwgk: I still feel strongly that
* the code duplication by itself is worrisome.
* the potential for bit rot (untested code) is worrisome.
* the large code block in the documentation gives too much weight to a feature that I think will be fairly esoteric to most pybind11 users.

rwgk (commented):
I'd feel much more comfortable with this PR if the documentation had a pointer to tested code elsewhere, and if we either reuse `std::function`, or have measurements to show that the `function_view` optimization is worth the extra code.

But I think it'll be best to first let @henryiii and @Skylion007 weigh in. I'm OK with this PR as-is if it gets support from a majority.

henryiii:
> I'd feel much more comfortable with this PR if the documentation had a pointer to tested code elsewhere

The current documentation does not do this. We probably should use include markers and test the documentation examples that way, but we don't, and I don't think we should change styles in the middle of an unrelated PR. We in general have to be careful changing any public interface, and this makes that a public interface, so the docs shouldn't bit-rot too fast.

include/pybind11/detail/function_view.h
jbms: The space usage isn't really an issue since `type_record` is temporary and only lives on the stack.

We could definitely use `std::function` --- that would be just fine.  I avoided it in part because pybind11 currently doesn't use `std::function` except in the separate, not default-included header, `pybind11/functional.h`.

I don't think there is really any situation in which `std::function` is more efficient or faster to compile or produces smaller generated code than `function_view`:

https://godbolt.org/z/57rMnnjTT

But in any case, I am happy to change to `std::function` if that is preferred.  Certainly if we don't anticipate any further uses of `function_view` (basically equivalent to `absl::FunctionRef` by the way) within pybind11 then it is probably not worth adding it for this purpose.

include/pybind11/attr.h
```diff
@@ -272,6 +290,9 @@ struct type_record {
     /// Custom metaclass (optional)
     handle metaclass;
 
+    /// Custom type setup.
+    ::pybind11::custom_type_setup::callback custom_type_setup;
```
rwgk, line 294: Using the same name for two different things is easily confusing. How about:
```
custom_type_setup::callback custom_type_setup_cb;
```
Or `custom_type_setup_callback` fully spelled out (I actually like that best; no need to go to a shortcut for something that's reference only a couple times).

include/pybind11/detail/function_view.h
rwgk: > We could definitely use std::function --- that would be just fine.

Could you please make that change? I think then this PR is good to go in.

include/pybind11/attr.h
jbms: Done.

include/pybind11/detail/function_view.h
jbms: Done.

jbms:
Note: Tests are now failing on Windows, due to an unrelated warning, presumably due to a compiler upgrade on github actions.

henryiii:
The last build on master compiled after this, with no errors. Also Appveyor also failed.

include/pybind11/attr.h
```diff
@@ -10,6 +10,8 @@
 
 #pragma once
 
+#include <functional>
+
 #include "cast.h"
```
henryiii, line 15:
```suggestion
#include "cast.h"

#include <functional>
```

Pybind11 headers must always be included before stdlib.

henryiii: @Skylion007 is there a way to force this in `clang-format/tidy`?

jbms:
This is the error I'm seeing on both AppVeyor and github actions:

```
C:\projects\pybind11\include\pybind11/embed.h(108): warning C4996: 'mbstowcs': This function or variable may be unsafe. Consider using mbstowcs_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\projects\pybind11\tests\test_embed\test_embed.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdlib.h(920): note: see declaration of 'mbstowcs'
C:\projects\pybind11\include\pybind11/embed.h(112): warning C4996: 'mbstowcs': This function or variable may be unsafe. Consider using mbstowcs_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\projects\pybind11\tests\test_embed\test_embed.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdlib.h(920): note: see declaration of 'mbstowcs'
  test_interpreter.cpp
C:\projects\pybind11\include\pybind11/embed.h(108): error C2220: warning treated as error - no 'object' file generated [C:\projects\pybind11\tests\test_embed\test_embed.vcxproj]
C:\projects\pybind11\include\pybind11/embed.h(108): warning C4996: 'mbstowcs': This function or variable may be unsafe. Consider using mbstowcs_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\projects\pybind11\tests\test_embed\test_embed.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdlib.h(920): note: see declaration of 'mbstowcs'
C:\projects\pybind11\include\pybind11/embed.h(112): warning C4996: 'mbstowcs': This function or variable may be unsafe. Consider using mbstowcs_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\projects\pybind11\tests\test_embed\test_embed.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdlib.h(920): note: see declaration of 'mbstowcs'
```

include/pybind11/attr.h
jbms: Thanks, done.

include/pybind11/attr.h
```diff
@@ -12,6 +12,9 @@
 
 #include "cast.h"
 
+// Include after pybind11 headers.
```
henryiii, line 15: This is always true, and we don't already have a note like this for each header. Let's see if we can enforce it with CI, and not write a comment in a single random header.

jbms: Done.

jbms:
Okay, all tests are passing now.

docs/advanced/classes.rst
```diff
@@ -1261,3 +1261,37 @@ object, just like ``type(ob)`` in Python.
...
+
+   struct OwnsPythonObjects {
+       py::object value = py::none();
+   };
+   py::class_<OwnsPythonObjects> cls(
+       m, "OwnsPythonObjects", py::custom_type_setup([](PyHeapTypeObject *heap_type) {
+           auto *type = &heap_type->ht_type;
+           type->tp_flags |= Py_TPFLAGS_HAVE_GC;
+           type->tp_traverse = +[](PyObject *self_base, visitproc visit, void *arg) {
```
rwgk, line 1283: tbh
```
type->tp_traverse = +[](
----------_---------^
```
I don't know what the `+` does here. IIRC previously you had those also in test_custom_type_setup.cpp, but then you took them out there. I tried, you could also take them out here. Are they somehow important here? But not in the unit test?

jbms: The `+` is applying unary `operator+` to the lambda, which causes it to be converted to a function pointer.  It is a somewhat common idiom.  However, it is true that it is not needed here, since that implicit conversion happens anyway, and in fact on Windows it causes problems which is why I took it out.

jbms: This is now fixed.

tests/test_custom_type_setup.cpp
```diff
@@ -0,0 +1,41 @@
...
+                Py_VISIT(self.value.ptr());
+                return 0;
+            };
+            type->tp_clear = [](PyObject *self_base) {
+                auto &self = py::cast<OwnsPythonObjects &>(py::handle(self_base));
+                self.value = py::none();
+                return 0;
+            };
+        }));
+    cls.def(py::init([] { return OwnsPythonObjects{}; }));
```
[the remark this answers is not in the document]
jbms: Done.

include/pybind11/detail/function_view.h
jbms: I'm not sure if there are currently other places where it would be useful at the moment.  I don't see any uses of `std::function` outside of pybind11/functional.h.

Just let me know if it should be brought back or not.

rwgk (approved):
Looks good to me.

Potentially adding `function_view` in another PR sounds fine, although only if we can show that the extra code has some measurable benefit.

docs/advanced/classes.rst
henryiii: FYI, I'm fine if someone wants to setup testing for our docs, I think it would be extremely useful. Probably by moving the examples to files, then using inclusion markers to pull it into the docs, but maybe the other direction (making files out of the docs then testing those) is possible too. But that should be a separate project.

jbms:
Is there anything left to do here?

rwgk:
> Is there anything left to do here?

@henryiii, @Skylion007, are you OK with this PR as it is now?

henryiii (approved):

rwgk:
Thanks @henryiii !

PR #3293, opened 2021-09-22
feat: reapply fixed version of #3271

henryiii:
## Description

SFINAE was broken, C++ does not perform SFINAE inside classes/structs. If someone wants to implement a better test (one that actually supports keep_alive), happy to take a contribution!

Activating SFINAE shows the new code is broken on some compilers.

## Suggested changelog entry:

```rst
- Add ``make_value_iterator()``.
- Fix ``make_key_iterator()`` to return references instead of copies.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1966,27 +1966,29 @@ struct iterator_state {
 };
 
 // Note: these helpers take the iterator by non-const reference because some
-// iterators in the wild can't be dereferenced when const.
-template <typename Iterator>
+// iterators in the wild can't be dereferenced when const. C++ needs the extra parens in decltype
+// to enforce an lvalue. The & after Iterator is required for MSVC < 16.9. SFINAE cannot be
+// reused for result_type due to bugs in ICC, NVCC, and PGI compilers. See PR #3293.
+template <typename Iterator, typename SFINAE = decltype((*std::declval<Iterator &>()))>
 struct iterator_access {
-    using result_type = decltype((*std::declval<Iterator>()));
+    using result_type = decltype((*std::declval<Iterator &>()));
     // NOLINTNEXTLINE(readability-const-return-type) // PR #3263
     result_type operator()(Iterator &it) const {
         return *it;
     }
 };
 
-template <typename Iterator>
+template <typename Iterator, typename SFINAE = decltype(((*std::declval<Iterator &>()).first)) >
 struct iterator_key_access {
-    using result_type = decltype(((*std::declval<Iterator>()).first));
+    using result_type = decltype(((*std::declval<Iterator &>()).first));
     result_type operator()(Iterator &it) const {
         return (*it).first;
     }
 };
 
-template <typename Iterator>
+template <typename Iterator, typename SFINAE = decltype(((*std::declval<Iterator &>()).second))>
 struct iterator_value_access {
-    using result_type = decltype(((*std::declval<Iterator>()).second));
+    using result_type = decltype(((*std::declval<Iterator &>()).second));
     result_type operator()(Iterator &it) const {
         return (*it).second;
     }
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1955,25 +1955,52 @@ inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_t
     return res;
 }
 
-template <typename Iterator, typename Sentinel, bool KeyIterator, return_value_policy Policy>
+/* There are a large number of apparently unused template arguments because
+ * each combination requires a separate py::class_ registration.
+ */
+template <typename Access, return_value_policy Policy, typename Iterator, typename Sentinel, typename ValueType, typename... Extra>
 struct iterator_state {
     Iterator it;
     Sentinel end;
     bool first_or_done;
 };
 
-PYBIND11_NAMESPACE_END(detail)
+// Note: these helpers take the iterator by non-const reference because some
+// iterators in the wild can't be dereferenced when const.
+template <typename Iterator>
+struct iterator_access {
+    using result_type = decltype((*std::declval<Iterator>()));
+    // NOLINTNEXTLINE(readability-const-return-type) // PR #3263
+    result_type operator()(Iterator &it) const {
+        return *it;
+    }
+};
 
-/// Makes a python iterator from a first and past-the-end C++ InputIterator.
-template <return_value_policy Policy = return_value_policy::reference_internal,
+template <typename Iterator>
+struct iterator_key_access {
+    using result_type = decltype(((*std::declval<Iterator>()).first));
+    result_type operator()(Iterator &it) const {
+        return (*it).first;
+    }
+};
+
+template <typename Iterator>
+struct iterator_value_access {
+    using result_type = decltype(((*std::declval<Iterator>()).second));
+    result_type operator()(Iterator &it) const {
+        return (*it).second;
+    }
+};
+
+template <typename Access,
+          return_value_policy Policy,
           typename Iterator,
           typename Sentinel,
-#ifndef DOXYGEN_SHOULD_SKIP_THIS  // Issue in breathe 4.26.1
-          typename ValueType = decltype(*std::declval<Iterator>()),
-#endif
+          typename ValueType,
           typename... Extra>
-iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
-    using state = detail::iterator_state<Iterator, Sentinel, false, Policy>;
+iterator make_iterator_impl(Iterator first, Sentinel last, Extra &&... extra) {
+    using state = detail::iterator_state<Access, Policy, Iterator, Sentinel, ValueType, Extra...>;
+    // TODO: state captures only the types of Extra, not the values
 
     if (!detail::get_type_info(typeid(state), false)) {
         class_<state>(handle(), "iterator", pybind11::module_local())
@@ -1987,7 +2014,7 @@ iterator make_iterator(Iterator first, Sentinel last, Extra &&... extra) {
                     s.first_or_done = true;
                     throw stop_iteration();
                 }
-                return *s.it;
+                return Access()(s.it);
             // NOLINTNEXTLINE(readability-const-return-type) // PR #3263
             }, std::forward<Extra>(extra)..., Policy);
     }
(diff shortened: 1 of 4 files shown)


henryiii:
@bmerry this should be a good starting point. Fixing this seems to be tripping up the new code, it's trying to make a copy of NonCopyableInt. Could you take a look? I can give you access to the branch if needed.

henryiii:
Fixed most of them by adding the double parentheses back in - should probably make a note that double parentheses are required there. Don't know why, would like to be able to say why - it was true across all compilers. Looks like at least ICC is unhappy now, but most are fine.

Skylion007:
Ping @rainwoodman

henryiii:
These are the three compilers that can't handle it:

```
X  🐍 3 • CentOS7 / PGI 20.9 • x64                            15m19s  https://github.com/pybind/pybind11/runs/3680311051
X  🐍 3 • ICC latest • x64                                    4m56s   https://github.com/pybind/pybind11/runs/3680313285
X  🐍 3.8 • CUDA 11 • Ubuntu 20.04                            5m23s   https://github.com/pybind/pybind11/runs/3680312770
X  🐍 3.9-dbg (deadsnakes) • Valgrind • x64                   14m39s  https://github.com/pybind/pybind11/runs/3680309032
```

PGI, NVCC, Intel ICC, and I expect my test is triggering valgrind. Also okay to revert #3271 then try again with this on top, if @bmerry is not available to fix this till next week and no one can fix these compilers. This is a pretty serous regression in master that was discovered <24 hours after merging by a user.

rwgk:
Reverting sounds good to me.

include/pybind11/pybind11.h
```diff
@@ -1966,27 +1966,27 @@ struct iterator_state {
 };
 
 // Note: these helpers take the iterator by non-const reference because some
-// iterators in the wild can't be dereferenced when const.
-template <typename Iterator>
+// iterators in the wild can't be dereferenced when const. C++ needs the extra parens in decltype.
+template <typename Iterator, typename ResultType = decltype((*std::declval<Iterator>()))>
```
henryiii, line 1970:
```suggestion
template <typename Iterator, typename ResultType = decltype(*std::declval<Iterator>())>
```

henryiii:
Going to revert, then move the rest of the changes into here.

henryiii:
Ideally, tests should be pulled out and applied to master first.

bmerry:
Hi @henryiii. Thanks for working on this - seems like you have more time for it this week than I do. From a quick look you're taking a similar approach to what I was thinking of. However, I'd suggest you try *not* using the `ResultType` template parameter to define the `result_type` nested typedef (I'll make a code suggestion to show what I mean). You're hitting the same problem I hit originally, where ICC (and I'm guessing the other failing compilers) gets the wrong type when decltype is used as the template parameter default (it loses the referenceness). That shouldn't matter if it's only used for SFINAE.

include/pybind11/pybind11.h
```diff
@@ -1955,25 +1955,52 @@ inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_t
...
     Sentinel end;
     bool first_or_done;
 };
 
-PYBIND11_NAMESPACE_END(detail)
+// Note: these helpers take the iterator by non-const reference because some
+// iterators in the wild can't be dereferenced when const. C++ needs the extra parens in decltype.
+template <typename Iterator, typename ResultType = decltype((*std::declval<Iterator>()))>
+struct iterator_access {
+    using result_type = ResultType;
```
bmerry, line 1972:
```suggestion
template <typename Iterator, typename SFINAE = decltype((*std::declval<Iterator>()))>
struct iterator_access {
    // Note: can't use SFINAE template parameter to define result_type due to a bug in ICC
    using result_type = decltype((*std::declval<Iterator>()));
```
See if that works (will need to be applied similarly to iterator_key_access and iterator_value_access.

henryiii: I believe this fixes all three "special" compilers, but breaks every version of MSVC.

```
D:\a\pybind11\pybind11\include\pybind11/pybind11.h(1991,1): error C2440: 'return': cannot convert from 'const _Ty2' to 'const int &&' [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
          with
          [
              _Ty2=int
          ]
D:\a\pybind11\pybind11\include\pybind11/pybind11.h(1991,21): message : You cannot bind an lvalue to an rvalue reference [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\include\pybind11/pybind11.h(1990): message : while compiling class template member function 'const int &&pybind11::detail::iterator_value_access<Iterator,const _Ty2 &&>::operator ()(Iterator &) const' [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
          with
          [
              Iterator=const std::pair<int,int> *,
              _Ty2=int
          ]
```

bmerry: That is extremely weird. Is this just by adding the SFINAE template parameter (relative to what was in #3271), or is it some interaction with the additional testing you added?

I see you've tried to make a workaround with `add_lvalue_reference`. I suspect that is going to cause other problems if the iterator dereferencing doesn't produce a reference (such as an input iterator). The next thing I would try is changing `std::declval<Iterator>()` to `std::declval<Iterator &>()` (all over) since the former will have type `Iterator &&` while the latter will have type `Iterator &`. I don't see why applying **dereferencing** to an rvalue reference should give another rvalue reference though.

henryiii: I think it's the additional testing, though I could be wrong. I think after adding the tests I've not been able to get MSVC to work. SFINAE seems to be working, it's just failing inside the correct overload. This would be easier if I set this up with MSVC locally on a Windows box. With nox it might be easier to setup than in the past. `add_lvalue_reference` and adding a `&` should be very similar except for void, so I'm expecting similar results, but let's see.

If we have an MSVC solution and an everyone-else solution, that would be doable.

bmerry: > add_lvalue_reference and adding a & should be very similar except for void, so I'm expecting similar results, but let's see.

Not really - you were using add_lvalue_reference to modify the result type, whereas I'm proposing modifying the iterator type. But since I don't really understand what MSVC was doing I'm also going to wait and see. I'm trying to use compiler explorer to reproduce the issue (which was pretty handy for the ICC bug), but haven't managed it yet.

henryiii: Ah, yes, I see that now. Looks like this is working a lot better! 90% complete with no failures yet.

Skylion007 (approved):
Looks good. May want to add a few comments explaining the black magic with the decltypes and how the & and the parens are needed.

henryiii:
Just to be sure, I'm going to verify that using ResultType still breaks those three compilers, now that this is fixed. I'll also add a comment as to why this is needed if it still is. I'll keep it in draft until then.

henryiii:
@Skylion007 do we need a couple more `NOLINTNEXTLINE(readability-const-return-type)` too? Not getting any linting errors, though.

bmerry:
I've managed to reproduce the MSVC bug in Compiler Explorer: https://godbolt.org/z/343qzxe6z.

bmerry:
It looks like this is a known bug in MSVC: https://developercommunity2.visualstudio.com/t/decltypepointerToArray-produces-rva/291161?entry=problem&space=62

henryiii:
> It looks like this is a known bug in MSVC: 

Doesn't that say fixed in 2019, specifically VS16.9, but it still seems to be present in VS16.11?

henryiii:
@rwgk and @bmerry okay to merge?

bmerry (changes requested):

tests/test_sequences_and_iterators.cpp
```diff
@@ -308,19 +336,63 @@ TEST_SUBMODULE(sequences_and_iterators, m) {
...
+            return py::make_value_iterator(NonZeroIterator<std::pair<int, int>>(s.begin()), NonZeroSentinel());
+        }, py::keep_alive<0, 1>())
+        .def("simple_iterator", [](IntPairs& self) {
+            return py::make_iterator(self);
+        }, py::keep_alive<0, 1>())
+        .def("simple_keys", [](IntPairs& self) {
+            return py::make_key_iterator(self);
+        }, py::keep_alive<0, 1>())
+
+        // test iterator with keep_alive (doesn't work so not used at runtime, but tests compile)
```
bmerry, line 349: Wasn't this whole PR triggered by someone whose previously working code used py::keep_alive here?

tests/test_sequences_and_iterators.cpp
```diff
@@ -308,19 +336,63 @@ TEST_SUBMODULE(sequences_and_iterators, m) {
...
         }, py::keep_alive<0, 1>())
+        .def("nonzero_values", [](const IntPairs& s) {
+            return py::make_value_iterator(NonZeroIterator<std::pair<int, int>>(s.begin()), NonZeroSentinel());
+        }, py::keep_alive<0, 1>())
+        .def("simple_iterator", [](IntPairs& self) {
+            return py::make_iterator(self);
+        }, py::keep_alive<0, 1>())
+        .def("simple_keys", [](IntPairs& self) {
+            return py::make_key_iterator(self);
+        }, py::keep_alive<0, 1>())
```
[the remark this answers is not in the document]
henryiii: Oops, bad merge, thanks!

tests/test_sequences_and_iterators.cpp
henryiii: Yes, in https://github.com/pybind/pybind11/commit/ee0c5ee405e7a532410797687da28a20b89cd62b#r56899065. But they had an object that supported keep_alive, and here we don't, we just have a pair. I think you have to have a python wrapped iterator to use a second keep_live like this. This at least makes sure the keep_alive don't trigger the wrong overload - it will fail to compile if it does. If there's another handy Extra we could add, that would work too.

I'd be happy to add a keep_alive test, but don't want to write the proper class & iterator. Maybe @wojdyr could?

bmerry: Since we're just checking that SFINAE works rather than the behaviour of the policy, you could probably replace it with something like `py::call_guard<int>()` (which won't do anything useful).

bmerry:
> > It looks like this is a known bug in MSVC:
> 
> Doesn't that say fixed in 2019, specifically VS16.9, but it still seems to be present in VS16.11?

You're right. So presumably it's not exactly the same bug, but some variant of it.

include/pybind11/pybind11.h
```diff
@@ -1955,25 +1955,54 @@ inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_t
...
     Sentinel end;
     bool first_or_done;
 };
 
-PYBIND11_NAMESPACE_END(detail)
+// Note: these helpers take the iterator by non-const reference because some
+// iterators in the wild can't be dereferenced when const. C++ needs the extra parens in decltype
+// to enforce an lvalue. The & after Iterator is required for MSVC < 16.9. ResultType cannot be
+// reused for result_type due to bugs in ICC, NVCC, and PGI. See #3293.
+template <typename Iterator, typename ResultType = decltype((*std::declval<Iterator &>()))>
```
bmerry, line 1972: How about 
```suggestion
template <typename Iterator, typename SFINAE = decltype((*std::declval<Iterator &>()))>
```
so that it's self-documenting that this argument is here for SFINAE purposes? Otherwise it looks like it could be removed.

bmerry (changes requested):

include/pybind11/pybind11.h
henryiii: Done, thanks!

henryiii:
> You're right. So presumably it's not exactly the same bug, but some variant of it.

Are you going to report it? Especially with a nice reproducer, should be easy to report?

tests/test_sequences_and_iterators.cpp
henryiii: I've used that, and also added one for each of the other two.

rwgk (approved):
I was not able to follow the details, but with so much attention given to this work, by so many people, I'm sure this will be fine!
After it's merged I'll run the google testing again (tonight). I think it's safe to not block on waiting for the results.

bmerry:
> > You're right. So presumably it's not exactly the same bug, but some variant of it.
> 
> Are you going to report it? Especially with a nice reproducer, should be easy to report?

Yep: https://developercommunity2.visualstudio.com/t/decltype-infers-rvalue-reference-when-de/1537933?entry=myfeedback&space=62

Do any of you have any experience reporting ICC bugs? I tried searching for a bug reporting system, and the most useful thing I found was a forum rant from a user complaining that the only way to report bugs was on the forum and that the bug reports were usually ignored, so I didn't look any further.

bmerry (approved):
Thanks again for fixing this up for me @henryiii. Assuming the CI passes I reckon it's good to go. I've got some work in progress on wiring this up to `bind_map` that I'll just need to rebase and write tests for next week.

henryiii:
They aren't all that great at responding to them, as they are leaving ICC behind for the new clang-based Intel compiler. I think @ax3l has reported before?

henryiii:
Thanks all!

rainwoodman:
Some reflection after the merge. I am astonished by the amount of C++ quirks we ran into for a relatively straightforward code reuse.

Could it have been easier if we instead pursued three separate function bodies, one for each of make_iterator, make_key_iterator, and make_value_iterator?

henryiii:
SFINAE would have still been broken (the reuse part actually worked seamlessly). Trying to support non-copyable on all compilers was where much of the effort came in. Now if there thinks simply had different names for the one that takes an iterator and a sentinel and the one that takes a object supporting iteration, it would have been really easy, as then SFINAE would not have been needed.

rainwoodman:
Thanks for the explanation!

PR #3296, opened 2021-09-23
tests: check simple iteration of pairs

henryiii:
Signed-off-by: Henry Schreiner <henryschreineriii@gmail.com>

## Suggested changelog entry:

```rst

```

PR #3297, opened 2021-09-23
fix: iteration of non-copyable

henryiii:
## Description

Making just the fix in #3271. Seeing if this passes more compilers by avoiding the bigger changes. Builds on #3296.

## Suggested changelog entry:

```rst
* Fix ``make_key_iterator()`` to return references instead of copies.
```

WIP: I've actually dropped the key part at the moment, seeing what compiles where.

### DIFF
--- a/tests/test_embed/CMakeLists.txt
+++ b/tests/test_embed/CMakeLists.txt
@@ -7,7 +7,7 @@ if("${PYTHON_MODULE_EXTENSION}" MATCHES "pypy" OR "${Python_INTERPRETER_ID}" STR
   return()
 endif()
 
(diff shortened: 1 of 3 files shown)


henryiii:
Unrelated, but not sure why updating Catch2 isn't working, seems here https://packages.msys2.org/package/mingw-w64-x86_64-catch?repo=mingw64 that it's available.

Skylion007:
@henryiii Any reason the PR is still closed? We don't need to update Catch2 in it, do we?

henryiii:
(Implemented in #3293, except for the Catch2 update)

PR #3311, opened 2021-09-27
[Minor] Fixing typo in docs

philbucher:
## Description
Fixing a typo in the docs

## Suggested changelog entry:
None

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -347,7 +347,7 @@ standard explicitly with
 
     set(CMAKE_CXX_STANDARD 14 CACHE STRING "C++ version selection")  # or 11, 14, 17, 20
     set(CMAKE_CXX_STANDARD_REQUIRED ON)  # optional, ensure standard is supported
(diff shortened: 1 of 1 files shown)


henryiii (approved):

henryiii:
I've cherry-picked this to master, will likely make a mess when we merge, but hopefully it will at least not be wiped out.

philbucher:
my apologies :/

I am not sure why it targeted `stable` as target branch, as you seem to be using `master` as default branch

in any case, having it in the description of the PR template will help! => #3313

henryiii:
You probably clicked "edit this page" on the docs, which technically is from stable (the stable docs are the default, though both are rendered). So not sure if, following this path, users will be aware that they are targeting `stable` in order to follow #3313, but it's at least worth adding.

philbucher:
yes exactly thats what I did

I think with the hint you added it should be possible to notice in time and switch the target branch before opening the PR

PR #3317, opened 2021-10-01
docs: update CHANGELOG for 2.8

henryiii:
Will release after this.


Skylion007 (approved):

rwgk (approved):

docs/changelog.rst
```diff
@@ -109,6 +109,9 @@ Build system improvements:
 
 * Fix a harmless warning from CMake 3.21 with the classic Python discovery.
   `#3220 <https://github.com/pybind/pybind11/pull/3220>`_
+  
+* Eigen repo and version can now be selected, and reconfigures faster.
```
rwgk, line 113: "reconfigures" or "reconfigured"? (I cannot tell)

rwgk (approved):

docs/changelog.rst
Skylion007: Does the std::free deserve a mention?

henryiii: I didn't think so (just took off the needs-changlog), but if you think it does, we can?

Skylion007: On the fence, it does fix a rare bug, but it's also a really minor change.

rwgk: For free: I think not mentioning is fine.

For eigen, how about simply:

Eigen repo and version can now be specified as cmake options.

Rationale: Mentioning "cmake" is important to establish relevant context. It's obvious that it's then easier to reconfigure (it doesn't have to be stated).

henryiii: Before, it had to reach out to the Internet every time you reconfigured to see if the version (3.3.8) was still the thing you had locally. Now, we use the hash instead, so it knows that the hash is unchanged, and doesn't reach out to the internet to look at the repository for possible tag updates.

I don't think it affects anyone that much, though, as it is tests only, so we can leave off mentioning the improvement.

henryiii: I'm fine to leave off rare/small bugfixes, to highlight the bigger ones. The really tiny ones can be seen in the git logs.

rwgk: > Before, it had to reach out to the Internet every time you reconfigured to see if the version (3.3.8) was still the thing you had locally. Now, we use the hash instead, so it knows that the hash is unchanged, and doesn't reach out to the internet to look at the repository for possible tag updates.

I was wondering but that totally wasn't clear to me. I think it's worth spelling that out in the changelog (~copy-paste from here).

henryiii: I don't think we have that many users building our tests with DOWNLOAD_EIGEN=ON, so I think it's fine as is (you have to have a pretty recent CMake to even do that). Too much text would highlight this change as more important than it is.

henryiii:
Assuming this passes (turned off Eigen on the CentOS tests, as there's not a newer image than we are using yet), then I think we are ready for release.

PR #3324, opened 2021-10-03
tests: allow Eigen repo and version to be overridden

rwgk:
To work around gitlab eigen outage.

The eigen repo is still unavailable (seems to be 2+ days now):

Screen Shot 2021-10-04 at 06 11 00

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -171,6 +171,14 @@ set(PYBIND11_CROSS_MODULE_TESTS test_exceptions.py test_local_bindings.py test_s
 
 set(PYBIND11_CROSS_MODULE_GIL_TESTS test_gil_scoped.py)
 
+set(PYBIND11_EIGEN_REPO
+    "https://gitlab.com/libeigen/eigen.git"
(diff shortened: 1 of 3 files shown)


rwgk:
It's very easy to roll back. Better than driving blind while the eigen problem is being worked on.

henryiii:
Will need a rebase and merge, since we want to roll back the second commit after they have the original repo restored.

tests/CMakeLists.txt
```diff
@@ -171,6 +171,14 @@ set(PYBIND11_CROSS_MODULE_TESTS test_exceptions.py test_local_bindings.py test_s
 
 set(PYBIND11_CROSS_MODULE_GIL_TESTS test_gil_scoped.py)
 
+set(PYBIND11_EIGEN_REPO
+    "https://gitlab.com/libeigen/eigen.git"
+    CACHE STRING "Eigen repository to use for tests")
+# This hash is for 3.3.8, using a hash for security reasons
```
henryiii, line 177: Technically, this is also faster on re-configure, too, since it doesn't have to reach out again.

PR #3343, opened 2021-10-08
Correct options on Eigen::MappedSparseMatrix.

cantonios:
## Description

The only accepted `Options` for `SparseMatrix` are `RowMajor` and `ColMajor`.
Note that `Options` != `Flags`, so passing `Type::Flags` leads to invalid
options.  This is checked in the latest version of Eigen via a static
assertion, which causes a build break.


Skylion007:
We should bump the Eigen version here as well: https://github.com/pybind/pybind11/blob/f4c81e0877a7d302eb580886f8552bc81d3b3d65/tests/CMakeLists.txt#L179 to run the tests properly on the latest version.

henryiii:
Should we just add `-wd4127` to the tests when compiling with MSVC? Ideally Eigen should suppress or handle this.

henryiii:
If it could be localized enough. Ideally on the Eigen include, if MSVC was happy with that (sometimes it has to be at usage, not at include).

Issue #3357, opened 2021-10-12
[BUG]: Recursive dispatch fails to find python function override. 


dyershov:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

Consider a simple C++ visitor pattern:
```C++
class Data{};
using DataVisitor = std::function<void (const Data&)>;

class Adder {
public:
    // interface to add two data objects, the result is passes to data visitor
    virtual void operator()(const Data& first, const Data& second, const DataVisitor& visitor) const = 0;
};

void add(const Data& first, const Data& second,  const Adder& adder, const DataVisitor& visitor) {
    adder(first, second, visitor);
}

void add3(const Data& first, const Data& second, const Data& third,  const Adder& adder, const DataVisitor& visitor) {
    adder(first, second, [&] (const Data& first_plus_second) {
        adder(first_plus_second, third, visitor);
    });
}
```
In this pattern, Data and Adder are abstract classes which can be used in two algorithms `add` and `add3` which, as the name suggest, add two or three objects of the type `Data` and send the result to a visitor. This pattern is intended to eliminate allocating memory for Data pointer in case we need to store temporary results, e.g., see `add3` algorithm.

Next, we define the corresponding python bindings:
```C++
class PyAdder : public Adder {
public:
    void operator()(const Data& first, const Data& second, const DataVisitor& visitor) const override {
        PYBIND11_OVERRIDE_PURE_NAME(void, Adder, "__call__", operator(), first, second, visitor);
    }
};

class PyData : public Data {
public:
    PyData() = default;
};

PYBIND11_MODULE(module, m) {
    pybind11::class_<Adder, PyAdder>(m, "Adder")
        .def(pybind11::init<>())
        .def("__call__", &Adder::operator());

    m.def("add", &add);
    m.def("add3", &add3);

    pybind11::class_<Data, PyData>(m, "Data")
        .def(pybind11::init<>());
}
```

Finally, let's test it in Python. Firs,t some boilerplate:
```Python
class Adder(module.Adder):
[...]

PR #3365, opened 2021-10-15
Test recursive dispatch using visitor pattern.

dyershov:
## Description

Implemented unit test aims at replicating BUG #3357. The implementation defines abstract C++ class that performs an operation on abstract member Data class and sends the result to a visitor. A trampoline class is used to bind Adder python class. Additionally two algorithms that use previously defined abstractions to perform operation on two or three objects. These algorithms are tested using a concrete python implementation. The recursive dispatch of overloaded operator fails in the case of algorithm that uses three objects. When calling a visitor indirectly from lambda function inside overloaded  operation, the expected functionality is restored.

## Suggested changelog entry:

```rst
- implement tests for recursive dispatch using visitor pattern
```

### DIFF
--- a/tests/test_virtual_functions.cpp
+++ b/tests/test_virtual_functions.cpp
@@ -174,6 +174,25 @@ struct DispatchIssue : Base {
     }
 };
 
+// An abstract adder class that uses visitor pattern to add two data
+// objects and send the result to the visitor functor
+struct AdderBase {
+    struct Data {};
+    using DataVisitor = std::function<void (const Data&)>;
+
+    virtual void operator()(const Data& first, const Data& second, const DataVisitor& visitor) const = 0;
+    virtual ~AdderBase() = default;
+    AdderBase() = default;
+    AdderBase(const AdderBase&) = delete;
+};
+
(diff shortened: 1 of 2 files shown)


dyershov:
@rwgk Tagging you in this PR as per https://github.com/pybind/pybind11/issues/3357#issuecomment-942328778.

Apologies for CI failing. I have to learn how to run checks locally so that I can fix things.

henryiii:
Try `nox`, maybe with `pipx run nox` if you have `pipx` installed, or just with `nox` if you have that installed. That should run all the possible tests, and `pre-commit` too. You can select as well, see `nox -l`.

dyershov:
Thanks @henryiii! Running `nox` didn't show failures in windows build. I'm suspecting, based on the comment for DispatchTest, pure virtual functions [here](https://github.com/pybind/pybind11/pull/3365/files#diff-7228bcb06c17310cb9b3156c0f11e7f3d3b834c6932ed0a037b156b7df2eb238R183) somehow not allowed in windows. I can fix it if it is causing problems.

PR #3368, opened 2021-10-15
fix: improve support for Python 3.11-dev

### LINKED ISSUE #3367: [BUG]: error: no member named 'f_code' in '_frame'
### Required prerequisites

### Problem description

Compilation error:  latest python 3.11 from github, latest pybind11:

`python-embedded-example-project/libs/pybind11/include/pybind11/detail/type_caster_base.h:473:38: error: no member named 'f_code' in '_frame'
                "  " + handle(frame->f_code->co_filename).cast<std::string>() +`

### Reproducible example code

```text
compile python-embedded-example-project with latest code

https://github.com/matusnovak/python-embedded-example-project
```

henryiii:
Also update 3.10 to final, better PyPy usage

Fixes #3367. However, 3.11 segfaults, so this doesn't work yet, and we shouldn't be running 3.11 in the normal workflow anyway, but in some separate workflow that's not tied to general PRs.

## Suggested changelog entry:

```rst
* Modernize usage of PyCodeObject on Python 3.9 (moving toward support for Python 3.11a1)
* Fix harmless warning on upcoming CMake 3.22
```

Also slims down our PyPy numpy/scipy testing to official wheels only.

Also correctly displays the Eigen version when downloaded automatically.

### DIFF
--- a/include/pybind11/detail/type_caster_base.h
+++ b/include/pybind11/detail/type_caster_base.h
@@ -469,19 +469,18 @@ PYBIND11_NOINLINE std::string error_string() {
         errorString += "\n\nAt:\n";
         while (frame) {
 #if PY_VERSION_HEX >= 0x03090000
-            PyCodeObject *code = PyFrame_GetCode(frame);
+            PyCodeObject *f_code = PyFrame_GetCode(frame);
 #else
-            PyCodeObject *code = frame->f_code;
+            PyCodeObject *f_code = frame->f_code;
+            Py_INCREF(f_code);
 #endif
             int lineno = PyFrame_GetLineNumber(frame);
             errorString +=
-                "  " + handle(code->co_filename).cast<std::string>() +
+                "  " + handle(f_code->co_filename).cast<std::string>() +
                 "(" + std::to_string(lineno) + "): " +
-                handle(code->co_name).cast<std::string>() + "\n";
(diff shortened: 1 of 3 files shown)


henryiii:
PyTest doesn't support 3.11 yet (see https://github.com/pypa/build/pull/376), so 3.11 is broken. I've fixed the API issues, though. Feel free to suggest cleanups if there are nicer ways to do things now, I focused on replicating the current design using 3.9+ API.

henryiii:
For the latest python 3.11, we are broken due to #3374. For 3.11a1, we get a segfault halfway through the tests. :/

I think we should move the 3.11 tests to a job that runs on a schedule or if a label is set, it will likely randomly break on Python updates and that will interfere with normal PRs.

henryiii:
> Do we need to jump in that early?

The earlier we start, the smaller the changelog between each failure, and we can find what changed more easily, and it's much easier to fix one issue at a time, instead of being hit with a dozen changes. Basically the best parts of Google's live-at-HEAD philosophy. ;)

But I think we agree here; I don't want to add this to our test matrix yet (ignore what you see in the current state of the PR), I want to make a separate workflow that just checks 3.11 and runs on a schedule or by request, and not on normal PRs. I don't want to do this locally, because I have no interest in manually building 3.11 or running this by hand every so often, that's what CI is for.

PR #3370, opened 2021-10-15
ci: cancel in-progress on repeated pushes

henryiii:
## Description

Can help if development is moving faster than the CI builds.


henryiii:
@lkeegan or @jeromerobert, any idea why minGW jobs are broken?

jeromerobert:
@henryii see https://github.com/msys2/setup-msys2/issues/167

PR #3374, opened 2021-10-17
fix: deprecate make_simple_namespace, fix Python 3.11

henryiii:
## Description

This was using an internal API detail, and just traded a few characters for extra header complexity and a special name users have to look up and remember. See https://github.com/pybind/pybind11/pull/2840, specifically https://github.com/pybind/pybind11/pull/2840/files#r730350459. This also broke Python 3.11, since that internal API moved.

We have some other usage of internal API; not a lot, but some (`git -C include grep '\b_Py'`) - internal API can change at any time, and CPython is explicitly cleaning up API for the next couple of versions.

## Suggested changelog entry:

```rst
* The simple namespace creation shortcut added in 2.8.0 was deprecated due to usage of CPython internal API, and will be removed soon. Use ``py::module_::import("types").attr("SimpleNamespace")`` instead.
```

Any usage of the recently added `py::make_simple_namespace` should be converted to using `py::module_::import("types").attr("SimpleNamespace")` instead.

### DIFF
--- a/docs/advanced/pycpp/object.rst
+++ b/docs/advanced/pycpp/object.rst
@@ -44,8 +44,9 @@ A `simple namespace`_ can be instantiated using
 
 .. code-block:: cpp
 
-    using namespace pybind11::literals; // to bring in the `_a` literal
-    py::object ns = py::module_::import("types").attr("SimpleNamespace")("spam"_a=py::none(), "eggs"_a=42);
+    using namespace pybind11::literals;  // to bring in the `_a` literal
+    py::object SimpleNamespace = py::module_::import("types").attr("SimpleNamespace");
+    py::object ns = SimpleNamespace("spam"_a=py::none(), "eggs"_a=42);
 
 Attributes on a namespace can be modified with the :func:`py::delattr`,
 :func:`py::getattr`, and :func:`py::setattr` functions. Simple namespaces can
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,10 @@ Starting with version 1.8.0, pybind11 releases use a `semantic versioning
 IN DEVELOPMENT
 --------------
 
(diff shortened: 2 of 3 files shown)


joukewitteveen:
> [...] just traded a few characters for extra header complexity and a special name users have to look up and remember.

Allow me to defend `make_simple_namespace` a bit. If we go with "simple things should be easy (and hard things should be possible", _and_ we consider a `make_simple_namespace` primitive easier than chained module loading and attribute lookup, then the question is whether making a simple namespace should be a simple action with `pybind`.

Suppose we have some Python code that expects a class instance with some methods defined on it. If we have some C++ implementation of those methods that we want to use, it would be nice if we could create an object with those methods. I don't think this is in anyway an exotic usage pattern, yet without `make_simple_namespace` I see no way of doing this that reflects it's simplicity.

The reason I think using an internal API call is warranted here is because the above use case is specific for the purpose of `pybind`, namely combining C++ and Python code. I don't think this use case is given much consideration by upstream Python when they decide whether an API call should be public or private (e.g. https://bugs.python.org/issue15003, which in some ways is the predecessor of https://github.com/python/cpython/pull/28970). Additionally, it saves a module import so it should be more performant.

In short, I think `pybind` should offer the functionality that was provided by `make_simple_namespace` as directly as possible (i.e. without extra module loading and attribute lookup).

rwgk:
Henry wrote:

> If you can talk CPython and @vstinner into adding a public API for this, I'd be fine to use that (and provide the old private usage for 3.3-3.10), but it should happen there first.

I agree with that. If 3.11 is released with a public API for this, I'd say it's a yes to keep for pybind11, mainly because it makes user code more readable and uses more discoverable. Discoverability is a big one for maintenance of a large-scale code base; I see runtime and compile time considerations as practically insignificant for something as tiny as this.

PR #3419, opened 2021-10-27
ci: support development releases of Python

henryiii:
## Suggested changelog entry:

```rst
* Test python-3.11 on PRs with the ``python dev`` label.
* Support PyPy 7.3.7 and the PyPy3.8 beta.
```


Skylion007:
@henryiii I think the crash might be related to calling the CPython APIs in get_type_override when being dispatched from a PURE_OVERRIDE call, specifically here: https://github.com/pybind/pybind11/blob/6abf2baa62063a39ccc4a9535351de571004ddd4/tests/test_factory_constructors.cpp#L123 There is something about that edge case that is causing the segfault and why the other tests in pass.

Skylion007:
@henryiii @rwgk It's not the best fix, but all tests pass for Python 3.11. We really need to open an issue about the fact that the new frame struct API should be public or at least have some getter for covarnames.

henryiii:
Could you do that? You've looked into it the most.

tests/test_exceptions.py
```diff
@@ -109,6 +109,7 @@ def test_python_alreadyset_in_destructor(monkeypatch, capsys):
 
         def hook(unraisable_hook_args):
             exc_type, exc_value, exc_tb, err_msg, obj = unraisable_hook_args
+            print(locals())
```
rwgk, line 112: Oversight?

tests/test_builtin_casters.py
```diff
@@ -299,7 +299,7 @@ def cant_convert(v):
     assert noconvert(7) == 7
     cant_convert(3.14159)
     # TODO: Avoid DeprecationWarning in `PyLong_AsLong` (and similar)
-    if (3, 8) <= env.PY < (3, 10):
+    if (3, 8) <= env.PY < (3, 10) and env.CPYTHON:
```
rwgk, line 302: TODO here and below, to remove the `and env.CPYTHON` after the 3.11 `co_varnames` issue is resolved?

tests/test_buffers.py
```diff
@@ -62,6 +62,8 @@ def test_to_python():
     assert cstats.alive() == 1
     del mat2  # holds a mat reference
     pytest.gc_collect()
+    pytest.gc_collect()  # Needed for PyPy
```
rwgk, line 65: Can these be removed again? (Looks like a last-ditch brute force attempt.)

include/pybind11/pybind11.h
```diff
@@ -2365,16 +2365,17 @@ inline function get_type_override(const void *this_ptr, const type_info *this_ty
 
     /* Don't call dispatch code if invoked from overridden function.
        Unfortunately this doesn't work on PyPy. */
-#if !defined(PYPY_VERSION)
-
+#if !defined(PYPY_VERSION) && PY_VERSION_HEX < 0x030B0000
+    // TODO: Remove PyPy workaround for Python 3.11.
+    // Current API fails on 3.11 since covarnames can be null.
```
rwgk, line 2370: co_varnames — consistent spelling helps pin-pointing later.

tests/test_builtin_casters.py
henryiii: This is a PyPy todo, pypy38 didn't adapt to the changes in CPython.

tests/test_buffers.py
henryiii: I think it didn't work, PyPy still didn't like this. So yes.

henryiii:
If you want, we could just disable the two failing tests on PyPy and move forward with PyPy testing. The garbage collector one doesn't clean up, even on 3.7, and the hook one doesn't work for mysterious reasons (I can't make a MWE, it seems to work in MWE form).

rwgk (approved):
Looks great, thanks!

.github/workflows/upstream.yml
```diff
@@ -0,0 +1,112 @@
...
+  cancel-in-progress: true
+
+env:
+  PIP_ONLY_BINARY: numpy
+
+jobs:
+  standard:
+    name: "🐍 3.11 dev • ubuntu-latest • x64"
+    runs-on: ubuntu-latest
+    if: "contains(github.event.pull_request.labels.*.name, 'python dev')"
```
henryiii, line 19: I don't think this will trigger for workflow dispatch. And I wanted it to run weekly. Can do in a followup PR.

PR #3423, opened 2021-10-28
fix: avoiding usage of _ if already defined

### LINKED ISSUE #3401: [BUG]: Underscore function conflicts with gettext common usage
### Required prerequisites

### Problem description

The use of `_` as a function in `detail/descr.h` conflicts with the common macro usage of `_()` to denote `gettext()`.

The use of `_` for translation is baked into multiple programs, e.g. wxFormBuilder to generate code for frameworks (wxWidgets) that expect the macro.

While this macro is only a suggestion in the gettext manual [1], it is built into expectations[2][3][4][5].

Since the pybind11 usage is internal, would we be able to rename the pybind11 functions from `_` to something similarly short like `x`?

[1] http://www.gnu.org/software/gettext/manual/gettext.html#Overview
[2] https://forums.wxwidgets.org/viewtopic.php?t=19062
[3] https://stackoverflow.com/q/15244397/995714
[4] https://stackoverflow.com/q/15244397/995714
[5] https://stackoverflow.com/questions/3336056/underscore-before-the-format-string

### Reproducible example code

```text
#define _(s) gettext(s)

namespace py = pybind11;

PYBIND11_MODULE(gettext_test, m) {
    m.attr("test") = "test";
}
```

henryiii:
## Description

Closes #3401 by avoiding using a raw `_`.

## Suggested changelog entry:

```rst
* Replace ``_`` with ``const_name`` in internals, avoid defining ``pybind::_`` if ``_`` defined as macro (common gettext usage).
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -208,7 +208,7 @@ public:
         return PyLong_FromUnsignedLongLong((unsigned long long) src);
     }
 
-    PYBIND11_TYPE_CASTER(T, _<std::is_integral<T>::value>("int", "float"));
+    PYBIND11_TYPE_CASTER(T, const_str<std::is_integral<T>::value>("int", "float"));
 };
 
 template<typename T> struct void_caster {
@@ -221,7 +221,7 @@ public:
     static handle cast(T, return_value_policy /* policy */, handle /* parent */) {
         return none().inc_ref();
     }
-    PYBIND11_TYPE_CASTER(T, _("None"));
+    PYBIND11_TYPE_CASTER(T, const_str("None"));
 };
 
 template <> class type_caster<void_type> : public void_caster<void_type> {};
@@ -264,7 +264,7 @@ public:
 
     template <typename T> using cast_op_type = void*&;
     explicit operator void *&() { return value; }
-    static constexpr auto name = _("capsule");
+    static constexpr auto name = const_str("capsule");
 private:
     void *value = nullptr;
 };
@@ -315,7 +315,7 @@ public:
     static handle cast(bool src, return_value_policy /* policy */, handle /* parent */) {
         return handle(src ? Py_True : Py_False).inc_ref();
     }
-    PYBIND11_TYPE_CASTER(bool, _("bool"));
+    PYBIND11_TYPE_CASTER(bool, const_str("bool"));
 };
 
 // Helper class for UTF-{8,16,32} C++ stl strings:
@@ -405,7 +405,7 @@ template <typename StringType, bool IsView = false> struct string_caster {
         return s;
     }
 
-    PYBIND11_TYPE_CASTER(StringType, _(PYBIND11_STRING_NAME));
+    PYBIND11_TYPE_CASTER(StringType, const_str(PYBIND11_STRING_NAME));
 
 private:
     static handle decode_utfN(const char *buffer, ssize_t nbytes) {
@@ -542,7 +542,7 @@ public:
         return one_char;
     }
 
-    static constexpr auto name = _(PYBIND11_STRING_NAME);
+    static constexpr auto name = const_str(PYBIND11_STRING_NAME);
     template <typename _T> using cast_op_type = pybind11::detail::cast_op_type<_T>;
 };
 
@@ -579,7 +579,7 @@ public:
         return cast(*src, policy, parent);
     }
 
-    static constexpr auto name = _("Tuple[") + concat(make_caster<Ts>::name...) + _("]");
+    static constexpr auto name = const_str("Tuple[") + concat(make_caster<Ts>::name...) + const_str("]");
 
     template <typename T> using cast_op_type = type;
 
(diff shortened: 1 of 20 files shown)


include/pybind11/detail/descr.h
```diff
@@ -53,9 +53,16 @@ constexpr descr<N1 + N2, Ts1..., Ts2...> operator+(const descr<N1, Ts1...> &a, c
     return plus_impl(a, b, make_index_sequence<N1>(), make_index_sequence<N2>());
 }
 
+template <size_t N>
+constexpr descr<N - 1> const_str(char const(&text)[N]) { return descr<N - 1>(text); }
+constexpr descr<0> const_str(char const(&)[1]) { return {}; }
+
+// The "_" might be defined as a macro - don't define it if so.
+#ifndef _
```
Skylion007, line 61: Do we should add a check to ensure people don't accidentally add the old syntax back into PyBind11 code.

henryiii: We could probably just add a define to a CMake job? `-D_=1` for example?

jagerman:
This needs a doc update, too: the `_` is still used in the examples in https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html

include/pybind11/detail/descr.h
rwgk: @Skylion007 What kind of check did you have in mind?

The only thing I can think of is something like:
```
#ifndef PYBIND11_DO_NOT_DEFINE_UNDERSCORE
#ifndef _
...
#endif
#endif
```

Only loosely related, would this work?
```
#define _ const_str
```

Just for background / FYI:

I love this PR, although it will be a one or two day cleanup job for me to replace `_` with `const_str` throughout Google's code base (including imported 3rd party code). After I'm done I'd add `PYBIND11_DO_NOT_DEFINE_UNDERSCORE` to our central build rule, to ensure the underscore cannot be used anymore.

henryiii:
Do we want to change the docs over? I think it's more readable, so I'd be in favor. We can leave `_` available indefinitely due to its previous prevalence.

rwgk:
> Do we want to change the docs over? I think it's more readable, so I'd be in favor. We can leave `_` available indefinitely due to its previous prevalence.

I'm strongly in favor if changing the docs to `const_str`. I'm really glad to see this change, the `_` is so easy to overlook, so difficult to pin-point, and gives no clue to what it's for.

include/pybind11/detail/descr.h
rwgk: Got it, thanks; sorry the history wasn't apparent to me.

A define similar to `PYBIND11_DO_NOT_DEFINE_UNDERSCORE` will be very important for my purposes, to guard against backsliding. I could patch it in when importing pybind11 from github, but I believe it will be useful in other environments, too, and it's just a couple lines extra in the code, and maybe one or two sentences in the changelog.

rwgk: Hi @henryiii, did you see my comment above? Does `PYBIND11_DO_NOT_DEFINE_UNDERSCORE` work for everybody? I'd be happy to add that here.

henryiii: I don't see the point of adding yet another configuration option when you can just define `-D_=1` (which is the original problem with this) and that works fine? You could even do:

```c++
#define _
#undef _
```

rwgk: Actually at 2nd thought, what I really want is an opt-in rather than an opt-out.
I'll just apply an automatic patch when importing from github.

About my other question, do we need to repeat the complicated `const_str` code? Is there a catch with this?
```
#ifndef _
#define _ const_str
#endif
```

I just tried it quickly locally and it seems to work.

henryiii: Because you should always avoid defines if at all possible. The problem here is that _other_ people are defining `_`; we are not, this was and is `pybind11::_`. We are trying to play nice with bad actors, not become one ourselves! :)

rwgk: In that case a comment would be useful, e.g.
```
// Repeating the const_str code to avoid introducing a #define.
```

rwgk (approved):

rwgk (approved):
Randomly crossed my mind:

`const_char_array` would be a more truthful / expressive name.

`py::str` is something completely different from `py::detail::const_str`.

jagerman:
> Randomly crossed my mind:
> 
> `const_char_array` would be a more truthful / expressive name.

Kind of, but the point of this wrapper isn't to be a const char array, it's meant only to be a compile-time static name for a class, and being a const char array just happens to be the current implementation detail.  So perhaps `"py::type_name"` or something along that line would be a better choice?

include/pybind11/cast.h
```diff
@@ -208,7 +208,7 @@ struct type_caster<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_t
         return PyLong_FromUnsignedLongLong((unsigned long long) src);
     }
 
-    PYBIND11_TYPE_CASTER(T, _<std::is_integral<T>::value>("int", "float"));
+    PYBIND11_TYPE_CASTER(T, const_str<std::is_integral<T>::value>("int", "float"));
```
jagerman, line 211: I think we should add one test that keeps the `_` that only gets built/tested when `_` is not externally defined (so that we still test backwards compatibility for existing code still using `_`).

henryiii:
I think the overlap between `py::detail::const_str` and `py::str` is fine - there can't be a const version of `py::str`, since it's a Python object, and you really don't see the `py::detail` part, it's always just `const_str` in the type caster. It's important to keep it fairly clear and short. `_` is a bit too much, but `const_char_array` would be too long. (And, eventually, it might be able to use a constexpr string, perhaps?)

`const_name` or `static_name` or `static_str` are ideas, though.

I can add a test. I'm wondering if adding a `_` define is tripping up Boost.

rwgk:
> const_name or static_name or static_str are ideas, though.

`const_name` sounds good to me, in line with Jason's idea to emphasize high level vs impl detail.

`static` has so many meanings already ... thinking, better not.

henryiii:
So `const_str`, `const_name`, and `type_name` (which I missed from before).

rwgk:
Const name is my strong favorite

(Typing on my phone)

rwgk (approved):
I just took another look. Perfect, thanks a lot for doing this!

henryiii:
Can I get one more approval? @jagerman, @Skylion007, or someone else?

jagerman (approved):
Looks good to me.

PR #3424, opened 2021-10-29
docs: rework CI a bit, more modern skipping

henryiii:
## Description

Minor docs modernization.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -57,10 +57,10 @@ def docs(session: nox.Session) -> None:
     session.chdir("docs")
 
     if "pdf" in session.posargs:
-        session.run("sphinx-build", "-M", "latexpdf", ".", "_build")
+        session.run("sphinx-build", "-b", "latexpdf", ".", "_build")
         return
 
-    session.run("sphinx-build", "-M", "html", ".", "_build")
+    session.run("sphinx-build", "-b", "html", ".", "_build")
 
     if "serve" in session.posargs:
         session.log("Launching docs at http://localhost:8000/ - use Ctrl-C to quit")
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -88,4 +88,6 @@ def build(session: nox.Session) -> None:
     session.log("Building normal files")
     session.run("python", "-m", "build", *session.posargs)
     session.log("Building pybind11-global files (PYBIND11_GLOBAL_SDIST=1)")
-    session.run("python", "-m", "build", *session.posargs, env={"PYBIND11_GLOBAL_SDIST": "1"})
+    session.run(
+        "python", "-m", "build", *session.posargs, env={"PYBIND11_GLOBAL_SDIST": "1"}
```
Skylion007, line 92:
```suggestion
        sys.executable, "-m", "build", *session.posargs, env={"PYBIND11_GLOBAL_SDIST": "1"}
```

noxfile.py
```diff
@@ -88,4 +88,6 @@ def build(session: nox.Session) -> None:
     session.log("Building normal files")
     session.run("python", "-m", "build", *session.posargs)
```
Skylion007, line 89:
```suggestion
    session.run(sys.executable, "-m", "build", *session.posargs)
```

Skylion007 (changes requested):

Skylion007:
All the Python calls in the nox.py file should be replaced with sys.executable to ensure ti's calling the same Python module as Nox.

henryiii:
You don't _want_ the same python calling nox inside the session; you want the one set up by the session inside the virtual environment nox creates. It explicitly always prepares the path, so `python` (or any other executable) is safe, and it will show a warning if anything external to the virtual environment is used (unless you pass `external=True`).

Skylion007:
Ah, nvm then.

Skylion007 (approved):

PR #3429, opened 2021-10-31
Update basics.rst

hovey:
Correct "which" versus "that" and add macOS equivalent to Linux.

## Suggested changelog entry:

```rst

```


hovey:
Mac instructions appear in later pages, so revert.

PR #3436, opened 2021-11-03
Don't add -g0 to Pybind11Extension if $CFLAGS sets -g.

anntzer:
On Unix, setuptools prepends $CFLAGS and $CPPFLAGS to the compiler flags
(they always come before extra_compile_args and anything else; see
distutils.sysconfig.customize_compiler).  In practice, the environment
variables are useful e.g. to quickly generate a debug build (e.g. by
setting CFLAGS=-g), but Pybind11Extension currently unconditionally
overwrites this with -g0.

Instead, check the environment variables and only insert -g0 if not
overridden by them.

## Suggested changelog entry:

```rst
Flags starting with ``-g`` in ``$CFLAGS`` and ``$CPPFLAGS`` are no longer overridden by ``.Pybind11Extension``.
```

### DIFF
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -42,6 +42,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import contextlib
 import os
 import platform
+import shlex
 import shutil
 import sys
 import sysconfig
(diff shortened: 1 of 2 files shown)


pybind11/setup_helpers.py
```diff
@@ -143,7 +144,12 @@ def __init__(self, *args, **kwargs):
         if WIN:
             cflags += ["/EHsc", "/bigobj"]
         else:
-            cflags += ["-fvisibility=hidden", "-g0"]
+            cflags += ["-fvisibility=hidden"]
+            env_cflags = os.environ.get("CFLAGS", "")
+            env_cppflags = os.environ.get("CFLAGS", "")
```
henryiii, line 149:
```suggestion
            env_cppflags = os.environ.get("CXXFLAGS", "")
```

anntzer: Oops, fixed.  It's CPPFLAGS though (again, per the distutils source).

Skylion007 (approved):

PR #3443, opened 2021-11-05
fix: add missing std::forward calls

Boris-Rasin:
Two of the four cpp_function overloads are missing std::forward calls, which seems like a simple oversight.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -115,7 +115,7 @@ public:
     template <typename Return, typename Class, typename... Arg, typename... Extra>
     // NOLINTNEXTLINE(google-explicit-constructor)
     cpp_function(Return (Class::*f)(Arg...)&, const Extra&... extra) {
-        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(args...); },
+        initialize([f](Class *c, Arg... args) -> Return { return (c->*f)(std::forward<Arg>(args)...); },
                    (Return (*) (Class *, Arg...)) nullptr, extra...);
     }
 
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):
This looks like an oversight from #2213 . Looks good to me. Ping @rwgk

rwgk:
Hi @Boris-Rasin, thanks for fixing the oversight, looks good to me.
I'm guessing you noticed because something broke. Ideally we'd add corresponding new tests. Distilling tests from a known breakage is much easier than having to imagine them out of thin air. Is there a chance you could try distilling out unit tests? Alternatively, could you please post the error message(s) here?

Boris-Rasin:
Hi @rwgk, ref qualified methods with rvalue reference parameters were broken:

```c++
struct RValueParam {
    void func1(std::string&&) {}
    void func2(std::string&&) const {}
    void func3(std::string&&) & {}
    void func4(std::string&&) const & {}
};

void init(pybind11::module& m)
{
  pybind11::class_<RValueParam>(m, "RValueParam")
    .def("func1", &RValueParam::func1) //ok
    .def("func2", &RValueParam::func2) //ok
    .def("func3", &RValueParam::func3) //error
    .def("func4", &RValueParam::func4) //error
  ;
}
```
// error message:
// pybind11/pybind11.h(118,74): error C2664: 'Return (std::string &&) &': cannot convert argument 1 from 'std::string' to 'std::string &&'

rwgk:
> Hi @rwgk, ref qualified methods with rvalue reference parameters were broken:

Thanks! I'd be great if you could add that to this PR, similar to what you see here (really straightforward):

https://github.com/pybind/pybind11/pull/2213/files

rwgk:
I triggered the CI, although I'm not sure if you're still working on tests/test_methods_and_attributes.py, to prove that the methods are actually runnable. Anything trivial is good enough for this PR, e.g (untested).
```
std::size_t func1(std::string&& s) { return s.size(); }
...
```
```
assert obj.func1('boo') == 3
...
```

rwgk (approved):
Thanks a lot for adding the tests!
Holding off merging only to give @Skylion007 a chance to look at the tests, too.

Skylion007 (approved):

Issue #3369, opened 2021-10-15
[BUG]: Wrong caching of the overrides


Trigve:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

pybind11 2.6.2 (or smart holder branch), Python 3.7.10, VS 2019 C++20 (compiled as x86), Windows 10 x64

There is problem of caching the python functions overrides. The problem could be simulated using the code below.

In the example, there are 2 function in python.
`func()` does returns the python-derived class in which the method is overriden from the base class.
`func2()` does returns python-derived class BUT NO method is overriden, therefor C++ function is called in the end.

The problem is, that the cache (`get_internals().inactive_override_cache`) does store the pair<handle of the type, func name address>. But if the type is deallocated, it isn't removed from the cache and in future there could be match of the given type handle and the function name address. The problem seems to be with `pybind11_meta_dealloc()`. Because the type isn't the one registered with pybind11, it doesn't erase it from the override cache. (**edit: Looking to it more carefully, the actual problem could be in `all_type_info_get_cache()`. Because, there type is removed from the `registered_types_py`, but the override cache isn't checked**).

On the line marked "// (2)" the `p_obj2->func();` call should be store/retrieved from the override cache as it always resolve to the C++ function.
The code on line marked "// (1)"  should ALWAYS returns `42` but on some random iteration it would return 0 (and assertion is hit "// (3)"), because the cache would hold the same handle and function name address and that means that function is in C++ and not in python instance (which is wrong).

This does occurs using https://github.com/pybind/pybind11/issues/1546#issuecomment-526712704 and also using smart holder branch https://github.com/pybind/pybind11/tree/smart_holder (@rwgk).

[...]

PR #3462, opened 2021-11-13
[smart_holder] Wrong caching of overrides 

Trigve:
## Description
There was a problem when the python type, which was stored in override
cache for C++ functions, was destroyed and the record wasn't removed from the
override cache. Therefor, dangling pointer was stored there. Then when the
memory was reused and new type was allocated at the given address and the
method with the same name (as previously stored in the cache) was actually
overridden in python, it would wrongly find it in the override cache for C++
functions and therefor override from python wouldn't be called.
The fix is to erase the type from the override cache when the type is destroyed.

Closes: #3369

## Suggested changelog entry:

```rst
Fix caching of the C++ overrides
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2093,6 +2093,16 @@ inline std::pair<decltype(internals::registered_types_py)::iterator, bool> all_t
         // gets destroyed:
         weakref((PyObject *) type, cpp_function([type](handle wr) {
             get_internals().registered_types_py.erase(type);
+
+            // Actually just `std::erase_if`, but that's only available in C++20
+            auto &cache = get_internals().inactive_override_cache;
+            for (auto it = cache.begin(), last = cache.end(); it != last; ) {
+                if (it->first == reinterpret_cast<PyObject *>(type))
+                    it = cache.erase(it);
+                else
+                    ++it;
+            }
+
             wr.dec_ref();
         })).release();
     }
--- a/tests/test_class_sh_inheritance.cpp
+++ b/tests/test_class_sh_inheritance.cpp
@@ -49,6 +49,24 @@ struct drvd2 : base1, base2 {
     int id() const override { return 3 * base1::base_id + 4 * base2::base_id; }
 };
 
+class test_derived {
+
+public:
+    virtual int func() { return 0; }
+
+    test_derived() = default;
+    ~test_derived() = default;
+    // Non-copyable
+    test_derived &operator=(test_derived const &Right) = delete;
+    test_derived(test_derived const &Copy) = delete;
+};
+
+class py_test_derived : public test_derived {
+    virtual int func() override { PYBIND11_OVERRIDE(int, test_derived, func); }
+};
+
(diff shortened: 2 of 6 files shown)


rwgk:
Awesome, thanks! This looks great at first sight. I triggered the CI.

Important, **so I can follow your work**: please do not force push changes other than rebase-on-master. Ideally force-push immediately after a rebase-on-master, then use only non-force pushes for your work. When this PR is merged, the commits will be squashed, therefore it doesn't matter for that purpose how many commits we have here.

I still have to look in more detail, but I believe we need to turn this into a PR against `master`, moving the added code in tests/test_class_sh_inheritance.* to tests/test_class.* (first guess).  —  All fixes that are not in the added code on the smart_holder branch need to go to master first, otherwise it will become very difficult to merge the smart_holder branch into master, which I'm hoping to do very soon (needs some work fixing docs etc.).

rwgk:
The ClangTidy failure looks very easy to fix, could you please do that?

https://github.com/pybind/pybind11/runs/4199226407?check_suite_focus=true

```
/__w/pybind11/pybind11/tests/test_embed/test_interpreter.cpp:54:17: error: 'virtual' is redundant since the function is already declared 'override' [modernize-use-override,-warnings-as-errors]
    virtual int func() override { PYBIND11_OVERRIDE(int, test_derived, func); }
    ~~~~~~~~    ^
5405 warnings generated.
```

rwgk:
I believe if you fix the ClangTidy error, we will have a lot less failures. All compilers except MSVC seem to be upset about the missing virtual destructor. I'll stop looking at the pile of errors. Let's just run the CI again after the two ClangTidy fixes.

Trigve:
> 
> Awesome, thanks! This looks great at first sight. I triggered the CI.
> 
> Important, **so I can follow your work**: please do not force push changes other than rebase-on-master. Ideally force-push immediately after a rebase-on-master, then use only non-force pushes for your work. When this PR is merged, the commits will be squashed, therefore it doesn't matter for that purpose how many commits we have here.
> 
> I still have to look in more detail, but I believe we need to turn this into a PR against `master`, moving the added code in tests/test_class_sh_inheritance.* to tests/test_class.* (first guess). — All fixes that are not in the added code on the smart_holder branch need to go to master first, otherwise it will become very difficult to merge the smart_holder branch into master, which I'm hoping to do very soon (needs some work fixing docs etc.).

Ok. I've already have changes against `master` branch, but I the problem is, that in the `smart_holder` branch the tests would need to be modified (which is not a big deal).

So should I close this PR and open one against the `master`?

rwgk:
Typing on my phone.
Could you please just leave this PR as is for now and transfer this work
with adjustments to a second new PR against master? (When we have the
master PR stable we can revisit this one.)

Trigve:
> Typing on my phone. Could you please just leave this PR as is for now and transfer this work with adjustments to a second new PR against master? (When we have the master PR stable we can revisit this one.)

Done as https://github.com/pybind/pybind11/pull/3465

rwgk:
Hi @Trigve, I just merged the master changes into the smart_holder branch. I believe this PR is obsolete now, but please let me know if I overlooked something.

PR #3466, opened 2021-11-13
Revert "style: drop pycln"

henryiii:
Reverts pybind/pybind11#3397


henryiii:
The artificial limit has been removed so this won't bite us again in 3.11 (unless there's a real issue, but that's fine, that can happen to anyone).

Skylion007 (approved):

PR #3478, opened 2021-11-18
Minimal reproducer for breakage introduced by PR #3402.

rwgk:
## Description
WIP — Starting with a minimal reproducer.

Breakage introduced by PR https://github.com/pybind/pybind11/pull/3402#issuecomment-963341987

## Suggested changelog entry:

```rst

```

### DIFF
--- a/tests/test_kwargs_and_defaults.cpp
+++ b/tests/test_kwargs_and_defaults.cpp
@@ -167,4 +167,12 @@ TEST_SUBMODULE(kwargs_and_defaults, m) {
         "class_default_argument",
         [](py::object a) { return py::repr(std::move(a)); },
         "a"_a = py::module_::import("decimal").attr("Decimal"));
+
+    // https://github.com/pybind/pybind11/pull/3402#issuecomment-963341987
+    // Reduced from tensorstore.Dim
+    struct tensorstore_dim {};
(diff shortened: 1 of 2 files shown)


henryiii:
@jagerman, it looks like the new code doesn't handle self correctly. The old code counted from the end, so self was ignored. But the new code counts from the start, so it has to account for self.

rwgk:
Hi @jagerman, could you please take a look? This is a really tiny reproducer. Looks like an edge case that's not handled correctly, maybe something simple like an off-by-one. I'm guessing a minor tweak will fix it, the trick is to know where. I'm hoping that you as the original author may have a clue straightaway.

rwgk:
I expanded the new test in a trivially systematic way, placing the `py::kw_only` at different positions:
```
0: *, int i
1: int, *, int i
2: int, int, *, int i
```
Then `def`ing (A) a simple function and (B) `init`.
Calling with `i=1` works consistently for (A), fails for (B) 0, but works for (B) 1 & 2.

@jagerman does that give you any clues?

jagerman:
I will look into this later today.  I think you're right that there's some edge case being triggered here but will likely take some debugging through it to figure it out.

rwgk:
Issue fixed by @jagerman's PR #3488 :-)

Issue #3477, opened 2021-11-17
VS2022 _invalid_parameter


gearslogy:
### Problem description

Not a pybind bug, it should be vs2022 bug 

![image](https://user-images.githubusercontent.com/14259542/142109750-e42400d8-0387-42b5-8510-69cee784faa0.png)

solution:

https://www.reddit.com/r/cpp_questions/comments/qpo93t/error_c2039_invalid_parameter_is_not_a_member_of/

in the pybind11.h , Insert this code after line 22 ,  it can fix the problem
```c
#elif defined(_MSC_VER)
#include <corecrt.h>
```

henryiii:
We can turn on 2022 testing, GHA now has 2022 images.

henryiii:
I think this is a bug in MSVC 2022. You shouldn't have to include a file (like corecrt.h) before accessing stdlib files like `atomic` and `memory`.

Boris-Rasin:
I think this is pybind11 bug, not MSVC 2022 bug.

Problem arises because pybind11 redefines _DEBUG variable in the middle of the build.

henryiii:
If we don't do that, then you have to link to the (often non-existent) debug lib of Python, which is usually not what you want.

henryiii:
We undefine it just for python.h, then redefine it. It shouldn't affect anything in C++, just C via Python.

Boris-Rasin:
These headers (<python.h> and others) include standard C/C++ headers, which leads to this bug. I believe if we include relevant standard C/C++ headers before redefining _DEBUG, it will fix the problem. These standard C/C++ headers will first be processed with correct _DEBUG definition, and will not be processed second time from <python.h> etc. because they all have include guards.

henryiii:
Sorry, I thought I responded here. CPython headers _must_ be included before any system headers get included. This is a requirement from CPython, and it's why we require pybind11 to be the first include too. However, this _might_ not be a requirement for MSVC; I know it breaks things on linux if not first, but not sure that's an issue if you are using GCC.

henryiii:
(See https://docs.python.org/3/c-api/intro.html#include-files - this just says:

> you _must_ include Python.h before any standard headers are included

(no mention of type of system)

rwgk:
> Sorry, I thought I responded here.
I got a bit confused, too.
@gearslogy, I think it would be better to close this with comment: this is being worked on under PR #3477, please continue the discussion there.

Issue #3459, opened 2021-11-12
[BUG]: undefing _DEBUG breaks build on VS 2022


libbooze:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

After updating to VS2022 my build was broken, I originally reported it to [Microsoft](https://github.com/microsoft/STL/issues/2335), but turns out it is probably pybind bug. 

You are not allowed to undef _DEBUG macro.

https://github.com/pybind/pybind11/search?q=_DEBUG

### Reproducible example code

```text
My code was broken in Debug build in VS 2022, I used 2.6, but master seems to have same undef.
```

snnn:
I think the primary reason is we can’t build numpy in debug mode(with the debug version msvcrt). ONNX and ONNX Runtime follow the same pattern. Either they all define the macro, or they all should undef it.

fsb4000:
related: https://github.com/pybind/pybind11/issues/3403

libbooze:
> I think the primary reason is we can’t build numpy in debug mode(with the debug version msvcrt). ONNX and ONNX Runtime follow the same pattern. Either they all define the macro, or they all should undef it.

Is this some limitation in MSVC or something in numpy is broken under debug build? From what I understand in other issue MSFT has 0 intention of supporting people undefing _DEBUG (and I agree with them, although I know very little about build hacks like this)

henryiii:
You can always set `Py_DEBUG` to disable the hiding of `_DEBUG` as a workaround, by the way.

Skylion007:
Relevant PR: https://github.com/pybind/pybind11/issues/1713#issue-416071000

WillemvJ:
> You can always set `Py_DEBUG` to disable the hiding of `_DEBUG` as a workaround, by the way.

Not sure if I understand this. Tried to add (in cmakelists.txt):

    target_compile_definitions(pybind_example PUBLIC Py_DEBUG=1) 

but that didn't work. Can build pybind module in release, but unfortunately not in debug at present on VS2022.

[...]

PR #3480, opened 2021-11-18
ci: try newer images

henryiii:
## Description

It seems MSVC 2022 is causing some issues (#3459, #3477). Let's try it explicitly.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ubuntu-latest, windows-latest, macos-latest]
+        runs-on: [ubuntu-latest, windows-2022, macos-11]
         python:
         - '2.7'
         - '3.5'
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -153,10 +153,6 @@
 #  pragma warning(push)
 // C4505: 'PySlice_GetIndicesEx': unreferenced local function has been removed (PyPy only)
 #  pragma warning(disable: 4505)
-#  if defined(_DEBUG) && !defined(Py_DEBUG)
-#    define PYBIND11_DEBUG_MARKER
-#    undef _DEBUG
-#  endif
 #endif
 
 // https://en.cppreference.com/w/c/chrono/localtime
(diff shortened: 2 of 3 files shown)


henryiii:
So it seems removing _DEBUG fixes the issue with needing the extra include (that is from an optional component), but reintroduces the original `LINK : fatal error LNK1104: cannot open file 'python35_d.lib'` type errors.

snnn:
From my understanding, in that case you will need to used python_d.exe instead of python.exe to run the package. Then I will lost numpy because I don't know how to build numpy in such a way.

henryiii:
Yes, this is why we disable _DEBUG when including Python.h unless this is a Py_DEBUG build. Python.h can't include C++ sources, though, so I don't know why MSVC 2022 is unhappy. I'm not sure adding a PYBIND11_KEEP_YOUR_HANDS_OFF_DEBUG variable would actually provide a way around the problem due to that.

libbooze:
> Yes, this is why we disable _DEBUG when including Python.h unless this is a Py_DEBUG build. Python.h can't include C++ sources, though, so I don't know why MSVC 2022 is unhappy. 

Not a STL person but I read their comments as:
 your undefing of _DEBUG breaks their code because they think they can use a function that is not available in release version of their libs, so linker errors happen.
https://github.com/microsoft/STL/issues/2335#issuecomment-967311942

also they say _DEBUG is NOT a customization point so you are not allowed to do what you do (I learned from that issue that clang even has a [warning ](https://godbolt.org/z/cnE3jeaKf)when you mess with reserved identifiers).
https://github.com/microsoft/STL/issues/2335#issuecomment-967489428

tl;dr I don't think MS will ever agree to supporting this, if you explain underlying issue why originally you undef _DEBUG then maybe somebody reading this will be able to figure out a different hack/solution.

If something like that is not feasible then maybe just add to README that debug build is not supported.

henryiii:
99% of users making debug builds want to debug _their_ code, not CPython itself; they usually don't have a CPython debug library and wouldn't know how to link to it even if they did. That's why we do this. If there's another way to get CPython to not require the debug library but to still build user code in debug mode, I'm all ears. :)

libbooze:
> If there's another way to get CPython to not require the debug library but to still build user code in debug mode, I'm all ears. :)
Thank you for reply. I think best would be to ask CPython people if they would support this, maybe they can make their dependency in debug configurable. 

Beside that I think this hack might work:

https://github.com/pybind/pybind11/issues/3477#issuecomment-974899737

Issue #747, opened 2017-03-20
Update custom type_caster doc with a more complete example


tjcorona:
I have recently come across strange errors (```XXX undetected error```) when calling overloaded methods that take as input objects that have custom type_caster instantiations. After some debugging, I found that a custom pybind11 ```type_caster<>::load(handle,bool)``` method must not only return a ```bool``` to denote its successful conversion, but must also throw a ```reference_cast_error()``` when the ```handle``` is not the expected type (to facilitate its use in pybind11's try/catch dispatch for overloaded methods). Also, the failed attempts to convert the ```handle``` likely raise a python error which, if left uncleared, will manifest in strange behavior elsewhere in the program. The [readthedocs](https://pybind11.readthedocs.io/en/master/advanced/cast/custom.html) for custom type casters doesn't really cover these crucial requirements on the ```load``` method. It would have been useful for me to read about this in the documentation, so I figure it would probably be useful for someone else to read it too!

tjcorona:
If you'd like, I can take a stab at updating your doc and put up a merge request. Just let me know!

wjakob:
Please do -- that would be great!

PR #3486, opened 2021-11-19
remove redundant copy operation. fix warning which can be interpreted as an error in some projects

Lishen1:
## Suggested changelog entry:

```rst
* Fix a rare warning about extra copy in an Eigen constructor. 
```


Skylion007 (changes requested):
You are missing a conditional here. the actual stride of the EigenConformable must be (0,0) if the negative stride boolean is not set (read the comment about bug #747). As this is implemented, I am worried it will break compatibility with old Eigen versions.

Skylion007 (changes requested):
Yeah, older versions of Eigen will hit the static assert, so we need to make the Stride object is strictly positive for maximum backwards compatibility.

include/pybind11/eigen.h
```diff
@@ -80,13 +80,12 @@ template <bool EigenRowMajor> struct EigenConformable {
     // Matrix type:
     EigenConformable(EigenIndex r, EigenIndex c,
             EigenIndex rstride, EigenIndex cstride) :
-        conformable{true}, rows{r}, cols{c} {
+        conformable{true}, rows{r}, cols{c},
+        stride{EigenRowMajor ? rstride > 0 ? rstride : 0 : cstride > 0 ? cstride : 0 /* outer stride */,
+               EigenRowMajor ? cstride > 0 ? cstride : 0 : rstride > 0 ? rstride : 0 /* inner stride */ } {
```
henryiii, line 85: Add parenthesis, please, I can't easily tell what's happening here.

include/pybind11/eigen.h
```diff
@@ -80,13 +80,12 @@ template <bool EigenRowMajor> struct EigenConformable {
...
         // TODO: when Eigen bug #747 is fixed, remove the tests for non-negativity. http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747
         if (rstride < 0 || cstride < 0) {
             negativestrides = true;
```
henryiii, line 88:
```suggestion
        negativestrides = (rstride < 0 || cstride < 0);
```

include/pybind11/eigen.h
```diff
@@ -80,13 +80,12 @@ template <bool EigenRowMajor> struct EigenConformable {
...
-        } else {
-            stride = {EigenRowMajor ? rstride : cstride /* outer stride */,
-                      EigenRowMajor ? cstride : rstride /* inner stride */ };
         }
```
henryiii, line 89:
```suggestion
```

include/pybind11/eigen.h
henryiii: This could also be moved to the member initializers above.

include/pybind11/eigen.h
```diff
@@ -80,14 +80,12 @@ template <bool EigenRowMajor> struct EigenConformable {
...
-            negativestrides = true;
-        } else {
-            stride = {EigenRowMajor ? rstride : cstride /* outer stride */,
-                      EigenRowMajor ? cstride : rstride /* inner stride */ };
-        }
+        conformable{true}, rows{r}, cols{c},
+        //TODO: when Eigen bug #747 is fixed, remove the tests for non-negativity. http://eigen.tuxfamily.org/bz/show_bug.cgi?id=747
+        negativestrides{rstride < 0 || cstride < 0},
+        stride{EigenRowMajor ? (rstride > 0 ? rstride : 0) : (cstride > 0 ? cstride : 0) /* outer stride */,
+               EigenRowMajor ? (cstride > 0 ? cstride : 0) : (rstride > 0 ? rstride : 0) /* inner stride */ } {
```
henryiii, line 87: Last tiny request from my side, could you swap the order, so that this matches the declaration order above?

```suggestion
        stride{EigenRowMajor ? (rstride > 0 ? rstride : 0) : (cstride > 0 ? cstride : 0) /* outer stride */,
               EigenRowMajor ? (cstride > 0 ? cstride : 0) : (rstride > 0 ? rstride : 0) /* inner stride */ },
        negativestrides{rstride < 0 || cstride < 0} {
```

henryiii: Doesn't actually pass the build with the order backwards due to warnings, so went ahead and commented it.

Skylion007 (approved):

PR #3488, opened 2021-11-20
Fix py::kw_only when used before the first arg of a method

jagerman:
The implicit space for the `self` argument isn't added until we hit the first argument, but this wasn't being done for `kw_only` or `pos_only` that happen *before* the first argument, and so a `kw_only` before the first argument on methods/constructors would break.

This fixes it by properly checking whether we need to add the self arg in those places as well.

(The `pos_only` issue here was extremely mild -- you didn't get the `/` in the docstring, but AFAICT it has no other effect since there are no meaningful arguments before it anyway).

Closes #3789 (thanks to @rwgk for the minimal test case in that PR).

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -414,11 +414,15 @@ inline void check_kw_only_arg(const arg &a, function_record *r) {
         pybind11_fail("arg(): cannot specify an unnamed argument after a kw_only() annotation or args() argument");
 }
 
+inline void check_have_self_arg(function_record *r) {
+    if (r->is_method && r->args.empty())
+        r->args.emplace_back("self", nullptr, handle(), true /*convert*/, false /*none not allowed*/);
+}
+
 /// Process a keyword argument attribute (*without* a default value)
 template <> struct process_attribute<arg> : process_attribute_default<arg> {
     static void init(const arg &a, function_record *r) {
-        if (r->is_method && r->args.empty())
-            r->args.emplace_back("self", nullptr, handle(), true /*convert*/, false /*none not allowed*/);
+        check_have_self_arg(r);
         r->args.emplace_back(a.name, nullptr, handle(), !a.flag_noconvert, a.flag_none);
 
         check_kw_only_arg(a, r);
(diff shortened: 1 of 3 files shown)


include/pybind11/attr.h
```diff
@@ -414,11 +414,15 @@ inline void check_kw_only_arg(const arg &a, function_record *r) {
         pybind11_fail("arg(): cannot specify an unnamed argument after a kw_only() annotation or args() argument");
 }
 
+inline void check_have_self_arg(function_record *r) {
```
rwgk, line 417: Maybe `append_self_arg_if_needed`?
"check" sounds like a const function that returns a bool.

include/pybind11/attr.h
```diff
@@ -414,11 +414,15 @@ inline void check_kw_only_arg(const arg &a, function_record *r) {
...
+    if (r->is_method && r->args.empty())
+        r->args.emplace_back("self", nullptr, handle(), true /*convert*/, false /*none not allowed*/);
```
rwgk, line 419: `/*convert=*/ true`
?
Super minor of course, but I don't remember ever seeing `value /*keyword*/`, it flips compared to expectations (Python).

rwgk (approved):
Awesome, thanks a lot Jason!

rwgk:
For the record: I ran the clang sanitizers (asan, msan, tsan, ubsan) with this PR included. All fine. (There are only a handful of known long-standing failures.)

I also initiated Google global testing (includes testing many OSS packages). Results in ~7 hours. But I think it's safe to merge this PR asap.

include/pybind11/attr.h
jagerman: Works for me; I didn't actually add this, just moved it as-is.

jagerman: Okay I *did* actually put that comment there, nearly 5 years ago in abc29cad020bb035f698fdc3e5d33c5c551ed772  :flushed:

include/pybind11/attr.h
jagerman: Updated.

include/pybind11/attr.h
jagerman: Updated.

rwgk:
Thanks Jason! — In the meantime the Google-global testing finished without issues. I'll go ahead with merging this PR.

PR #3492, opened 2021-11-21
[smart_holder] Fixing `stict` vs `strict` typo.

rwgk:
## Description
Very minor fix.

## Suggested changelog entry:

```rst

```

PR #3493, opened 2021-11-21
Fixing `stict` vs `strict` typo.

rwgk:
## Description
Interestingly, this caused a failure only when testing the smart_holder branch, and only with `py::smart_holder` as the default holder.

## Suggested changelog entry:

```rst

```


rwgk:
Corresponding smart_holder CI run: https://github.com/pybind/pybind11/actions/runs/1487747056

(I will discard draft PR #3492 and merge from master instead, as usual.)

PR #3497, opened 2021-11-22
fix: vs2022 compilation, issue #3477

Boris-Rasin:


include/pybind11/detail/common.h
```diff
@@ -154,6 +154,8 @@
 // C4505: 'PySlice_GetIndicesEx': unreferenced local function has been removed (PyPy only)
 #  pragma warning(disable: 4505)
 #  if defined(_DEBUG) && !defined(Py_DEBUG)
+// fix for vs2022: https://github.com/pybind/pybind11/issues/3477
+#    include <crtdefs.h>
```
Skylion007, line 158: This seems to be causing issues for older MSVCs so I think the ifdef need to be more specific for new MSVCs if you want this work.

henryiii: I believe this introduces a new dependency on an optional component of MSVC / Windows.

henryiii: Since it's a specific case, it's probably okay (and might not be true, anyway). How about gating it for MSVC 2022+? We could probably also add `_CRT_SECURE_NO_WARNINGS` on Python 2.7, I hope people are not mixing 2.7 and MSVC 2022. Actually I hope people are not using 2.7 on Windows at all. Or at all.

Boris-Rasin: Yes, defining _CRT_SECURE_NO_WARNINGS will fix this problem.

I think it's a good idea to always define _CRT_SECURE_NO_WARNINGS, for all MSVC versions and all python versions. These warnings (treated by this build as errors) complain about use of standard C library functions and direct users toward non-standard Microsoft functions, which are not present on other platforms.

henryiii: > I think it's a good idea

Yes, but we shouldn't mess with other people's code. Using pybind11 should be as light weight as possible. If the compiler wants to tell the user something, we shouldn't get in the way, unless it's about pybind11 itself. The fewer surprises, the better. Every time someone tries to be "helpful" with something like this in a library, that causes issues down the road. Applications can do this, we could even suggest it, but we should not "just do it" for everyone.

Boris-Rasin: Then maybe we could stop treating these warnings as errors?

henryiii: I'm perfectly fine to hide these warnings inside pybind11, highly recommended actually, that's why we make the warnings errors, so we fix them. And we don't treat warnings as errors for anyone else.

(I honestly don't care enough about Python 2 to worry about leaking this if it's hard do to, by the way)

rwgk: Would it work to `#define _CRT_SECURE_NO_WARNINGS` only in this context, and only for VS2022? (Maybe that was suggested already?)
I think that would be an OK tradeoff.

In any case, I'd find it useful to generate a message like "INFO: Mixing non-debug Python build with debug pybind11 build. Please see *section in pybind11 docs* for background."
Ideally with added "_CRT_SECURE_NO_WARNINGS disabled" in this particular situation.
Although I have no idea how easy this is to implement.

Re "OK tradeoff":
* Clearly, playing tricks with a symbol that has a leading underscore would make any purist frown. We're obviously in "hack" territory right there.
* I'm strongly in favor of it through, as a matter of practicality.
* In that hack territory, it is nice if we can keep up as much code health tooling as possible, but if we run into a conflict between mixing debug modes and enabling warnings, I come down on the side of disabling the warnings that trip us up. Ultimately code that matters in real life will be built in release mode, the warnings will show up then.

include/pybind11/detail/common.h
```diff
@@ -154,6 +154,11 @@
...
+// python 2.7 header uses 'mbstowcs' function
```
rwgk, line 159: Oh, if this is actually only needed for Python 2.7 (?), an `#ifdef` here would be best. That way we will know for sure that we can remove this code again when we purge Python 2 support in a couple months.
With the `#ifdef`, it definitely looks good to me. (Even without, but only if Python 3 actually needs it.)

include/pybind11/detail/common.h
henryiii: The warning is just due to some weird old CPython 2 stuff. The problem here is unrelated; we are intentionally going against the CPython requirement to try to hack around something that changed in MSVC 2022, this hack requires Python.h to be included before any system headers, directly in opposition to the docs.

Ideally:
* CPython should workaround this (if we can make a MWE, which is not too hard, I'd think, it would be a good report to make)
* MSVC could possibly do something to ensure this worked like before?
* This workaround (scoped as closely as possible) is fine for now, but could be broken by a future MSVC or CPython release, as neither supports this at all right now.

include/pybind11/detail/common.h
henryiii: No, see the comment I forget to send above. :)

include/pybind11/detail/common.h
rwgk: I guess I'm still a bit lost, and I forgot to mention before, I'd `#ifdef` guard any workarounds also with `_MSC_VER`, to not change or break anything existing.

Henry wrote:
> this hack requires Python.h to be included before any system headers, directly in opposition to the docs.

Did you mean "this hack requires a system header to be included before Python.h, directly in opposition to the docs."

Meta... I just realized we have not two but three open threads:

#3477 an issue by @gearslogy
#3480 a PR by @henryiii 
#3497 a PR by @Boris-Rasin 

all about VS 2022. No wonder I'm confused :-) Please, let's close all but #3480 (maybe change the title) and work together / discuss there.

include/pybind11/embed.h
```diff
@@ -97,6 +97,12 @@ struct wide_char_arg_deleter {
     }
 };
 
+// warning C4996: 'mbstowcs': This function or variable may be unsafe.
+#if defined(_MSC_VER)
```
henryiii, line 101: This is also only needed for Py2, correct?

include/pybind11/detail/common.h
henryiii, line 158: This needs to be enabled only for msvc 22+. It might add a new dependency on an optional component. Plus it will help reduce the impact if something breaks.

include/pybind11/embed.h
Boris-Rasin: This disable warning is needed whenever mbstowcs is called. Which, according to the code, is when PY_VERSION_HEX < 0x030500f0 and HAVE_BROKEN_MBSTOWCS is not defined.

include/pybind11/detail/common.h
Boris-Rasin: This doesn't add any dependencies which are not already there. Pybind11 is already dependent on standard C library.

include/pybind11/embed.h
henryiii: Great, then could that be added here? Though I'd just put `PY_VERSION_HEX < 0x030500f0`. That way we will remember to remove it at the end of the year when we drop Python 2 & 3.5 support.

henryiii:
Would you mind turning on a 2022 build? See #3480.

include/pybind11/embed.h
henryiii: Nice solution to this one. :)

henryiii:
We can disable the setuptools_helper test on 3.5 - unless it's our fault (I don't think so), then this is just an incompatibility between setuptools & Python 3.5 & MSVC 2022 that I don't really care about.

henryiii:
I still like gating the `include <crtdefs.h>` for MSVC 2022+. Other than those two things, then I think this is good to go.

Boris-Rasin:
How do I "disable setuptools_helper test on 3.5"?

henryiii (approved):

Boris-Rasin:
This problem is not with vs2022 C++ compiler, it is with vs2022 C++ standard library, which could be used by different compilers, for example clang bundled with vs2022 (or clang for windows installed separately).

Therefore, if we want to limit workaround for specific version, it should be version of msvc stdlib (_MSVC_STL_VERSION >= 143), not compiler version (_MSC_VER), which is in fact old on clang for windows.

The problem in this case is, we need to include system header (yvals.h) anyway to get _MSVC_STL_VERSION value.

henryiii:
When did they start bundling clang? Nice. Maybe lfortran will be more readily available?

Anyway, at least this is much more controlled, and lets the reader know more about the issue. Hopefully `yvals.h` doesn't include much else.

.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ubuntu-latest, windows-latest, macos-latest]
+        runs-on: [ubuntu-latest, windows-2022, macos-latest]
```
rwgk, line 26: IIUC we're implicitly dropping `windows-latest`. Mostly a question for @henryiii: could we simply add `windows-2022` but keep `windows-latest` (until it becomes the same)?

include/pybind11/embed.h
rwgk: From what I see it's not gated on `PY_VERSION_HEX` or the value of `_MSC_VER`. I missed why not and why Henry wrote "Nice solution". A terse comment to explain would be helpful.

include/pybind11/detail/common.h
```diff
@@ -154,6 +154,11 @@
 // C4505: 'PySlice_GetIndicesEx': unreferenced local function has been removed (PyPy only)
 #  pragma warning(disable: 4505)
 #  if defined(_DEBUG) && !defined(Py_DEBUG)
+// fix for vs2022: https://github.com/pybind/pybind11/issues/3477
+#    include <yvals.h>
+#    if _MSVC_STL_VERSION >= 143
+#      include <crtdefs.h>
```
rwgk, line 160: A comment to explain that this is knowingly violating the Python.h include order would be good. For example:

```
// Workaround for a VS 2022 issue.
// NOTE: This workaround knowingly violates the Python.h include order requirement:
// https://docs.python.org/3/c-api/intro.html#include-files
// See https://github.com/pybind/pybind11/pull/3497 for full context.
```

Boris-Rasin:
@ henryiii the single failed test seems to be completely unrelated. Can you take a look? Otherwise, I think this is ready.

rwgk:
Yes, that's a known flake, safe to ignore.

Do you see my review from 3 hours ago?

include/pybind11/embed.h
Boris-Rasin: No additional gates are needed here. Whenever we call mbstowcs function on MSVC stdlib, we need to (temporarily) disable "warning C4996: 'mbstowcs': This function or variable may be unsafe".

include/pybind11/detail/common.h
rwgk: Before I wrote:  From what I see it's not gated on PY_VERSION_HEX or the value of _MSC_VER.

I see it now, it's already gated on Python <= 3.5. Sorry I missed it because the github view didn't show it.

rwgk (approved):
Looks good to me, thanks!

I already tested with the smart_holder branch (good; just one flake).

Leaving this for @henryiii to merge, because of my windows-latest vs windows-2022 question.

.github/workflows/ci.yml
rwgk: @henryiii I simply tried it out: #3517 is the same as this PR but with `windows-latest` added back. Here we have 73 checks, on the other PR 79. I verified that we actually have distinct builds (below). — Resource wise it's definitely not a problem to have both. I'm thinking: Why not?

```
$ grep ' -- The CXX compiler identification is' *.txt | grep windows-latest.*MSVC | sed 's/2021-11-30[^ ]*//' | sed 's/^[^_]*//' | uniq | sort
______2.7_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______3.10_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______3.5_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______3.6_____windows-latest_____x64_-DPYBIND11_FINDPYTHON=ON.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______3.9_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______pypy-3.7-v7.3.7_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
______pypy-3.8-v7.3.7_____windows-latest_____x64.txt: -- The CXX compiler identification is MSVC 19.29.30137.0
```

```
$ grep ' -- The CXX compiler identification is' *.txt | grep windows-2022.*MSVC | sed 's/2021-11-30[^ ]*//' | sed 's/^[^_]*//' | uniq | sort
______2.7_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______3.10_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______3.5_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______3.6_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______3.9_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______pypy-3.7-v7.3.7_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
______pypy-3.8-v7.3.7_____windows-2022_____x64.txt: -- The CXX compiler identification is MSVC 19.30.30705.0
```

henryiii: Let's just add a few windows-2019 builds, but don't think we need the full set. Also, I think the difference is that 2022 has the MSVC 2022 compiler _also_, but we could select the 2019 or 2017 or even 2015 compilers as well on the newer images (not sure how far back they go, used to be 2015, 2017, and 2019). If we force the older compiler, we don't need older Windows images (the windows-2016 image is being removed in March).

rwgk: What should we do on this PR?

My feeling: add windows-2022 here, then trim down in a separate PR.

henryiii: We can inject a 2022 build and leave the bulk of them 2019, or we can inject a 2019 build and leave the bulk of them 2022. No need to explode the build matrix.

rwgk: Sounds good to me either way. The simpler the better (without losing coverage). When trading of human time (for more complicated config) vs. 3 or 4 extra jobs, I'd come down on the side of minimizing human effort.

.github/workflows/ci.yml
```diff
@@ -50,6 +50,9 @@ jobs:
               -DPYBIND11_FINDPYTHON=ON
           - runs-on: macos-latest
             python: pypy-2.7
+          # Inject a Windows 2019 run
+          - runs-on: windows-2019
+            python: 3.9
```
rwgk, line 55: Nice! I didn't know it's that easy.
While we're still advertising Python 2.7 support, could we add that in here, too?

rwgk (approved):
Looks perfect now, thanks Henry!

rwgk:
Re-testing after git rebase master.

rwgk:
pypy-2.7 • macos-latest • x64 seems to be stuck on Post Cache wheels
Ignoring.
I'll merge this now, to get the lengthy process going -> smart_holder -> Google.

PR #3499, opened 2021-11-22
Replace usage of deprecated Eigen class MappedSparseMatrix.

rmlarsen:
Replace usage of deprecated Eigen class MappedSparseMatrix.

## Description
`Eigen::MappedSparseMatrix` has been deprecated since Eigen 3.3 from 2016. Use the equivalent modern syntax `Eigen::Map<Eigen::SparseMatrix<...>>`.

The deprecated class was removed on the Eigen master branch in:

https://gitlab.com/libeigen/eigen/-/commit/7e586635ba85ba926ba3148f8d14beca3535f970

## Suggested changelog entry:

```rst
Replace usage of deprecated ``Eigen::MappedSparseMatrix`` with ``Eigen::Map<Eigen::SparseMatrix<...>>`` for Eigen 3.3+.
```

### DIFF
--- a/include/pybind11/eigen.h
+++ b/include/pybind11/eigen.h
@@ -50,8 +50,12 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 
 #if EIGEN_VERSION_AT_LEAST(3,3,0)
 using EigenIndex = Eigen::Index;
+template<typename Scalar, int Flags, typename StorageIndex>
+using eigen_mapped_sparse_matrix = Eigen::Map<Eigen::SparseMatrix<Scalar, Flags, StorageIndex>>;
 #else
 using EigenIndex = EIGEN_DEFAULT_DENSE_INDEX_TYPE;
+template<typename Scalar, int Flags, typename StorageIndex>
+using eigen_mapped_sparse_matrix = Eigen::MappedSparseMatrix<Scalar, Flags, StorageIndex>;
 #endif
 
 // Matches Eigen::Map, Eigen::Ref, blocks, etc:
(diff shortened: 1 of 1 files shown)


Skylion007:
Currently, the minimum supported Eigen version in PyBind11 is [3.2.7](https://github.com/pybind/pybind11/blob/7c580586f853b5f48603594ff0c11a5211de9007/include/pybind11/eigen.h#L40). Did this API exist in 3.2? Otherwise, we need to either raise the minimum Eigen version or add an ifdef backport.

rmlarsen:
@rmlarsen I don't think this existed in Eigen 3.2. What would you prefer, raising the minimum version or an ifdef? Eigen 3.2 is pretty ancient.

rmlarsen:
@Skylion007 what do you prefer?

Skylion007:
If it's not too much trouble to enable a backport with a "using"/typedef, it would be nice to keep support for Eigen 3.2.

rmlarsen:
@Skylion007 PTAL

henryiii (approved):

include/pybind11/eigen.h
```diff
@@ -50,8 +50,12 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 
 #if EIGEN_VERSION_AT_LEAST(3,3,0)
 using EigenIndex = Eigen::Index;
+template<typename Scalar, int Flags, typename StorageIndex>
+using eigen_mapped_sparse_matrix = Eigen::Map<Eigen::SparseMatrix<Scalar, Flags, StorageIndex>>;
```
henryiii, line 54: For consistency with the line above, how about EigenMapSparseMatrix?

rmlarsen: I was aiming for consistency with the numerous other template alias below line 61.

henryiii: I don't think this is public, so no strong opinion. But those are combining snake_case stdlib names and Eigen, while this is purely an Eigen CamelCase name.

Skylion007 (approved):

rwgk (approved):
Thanks all!
I'll go ahead merge this now.

@rmlarsen I'll import this into the Google env asap.

rmlarsen:
@rwgk Thanks!

Issue #3504, opened 2021-11-23
[BUG]: std::filesystem::path does not automatically convert to pathlib.Path on Windows with pybind11 2.8.1


srydell:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

Hi,

First of all, thank you for all the work on this amazing project. I've been trying out the automatic filesystem::path conversion, but there seems to be an issue with it at least on Windows. While trying to build the example below I get the following:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: toString(): incompatible function arguments. The following argument types are supported:
    1. (p: std::filesystem::path) -> str

Invoked with: WindowsPath('Example')

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
```

Here is a minimal repo reproducing the same thing:

[https://github.com/srydell/pybind11_filesystembug](https://github.com/srydell/pybind11_filesystembug)

I used Visual Studio 2019 Professional Version 16.11.7

If there is anything else I can provide, I'll gladly do so. Thanks!

### Reproducible example code

```text
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
#include <filesystem>
#include <string>

namespace py = pybind11;

std::string toString(std::filesystem::path const& p) {
	return p.string();
}

PYBIND11_MODULE(m, m) {
	m.def("toString", &toString, "", py::arg("p"));
}


from pathlib import Path
import m

print(m.toString(Path("Example")))
```

Skylion007:
You need to include `<pybind11/stl/filesystem.h>`.

PR #3505, opened 2021-11-23
Update required include file for std::filesystem::path

srydell:
From `pybind11/stl.h`  to `pybind11/stl/filesystem.h` after opening https://github.com/pybind/pybind11/issues/3504

## Suggested changelog entry:

```rst

```


Skylion007:
This change is already in master.

PR #3517, opened 2021-11-30
Clone of #3497 with windows-latest added back.

rwgk:
Helper/scratch PR for testing. #3497 is the primary PR, if we decide we want to make modifications, they will be applied back there.


rwgk:
This PR is obsolete. PR #3497 adopted a more selective approach.

PR #3521, opened 2021-12-01
Expand std::string_view support to str, bytes, memoryview

jagerman:
## Description

1. Allows constructing a str or bytes implicitly from a string_view; this is essentially a small shortcut allowing a caller to write `py::bytes{sv}` rather than `py::bytes{sv.data(), sv.size()}`.

    For `py::str` this also allows `std::u8string_view`, but not for `py::bytes` because that didn't seem entirely appropriate to me.

2. Allows implicit conversion *to* `std::string_view` from `py::bytes`—this plugs a current hole where there's no simple way to get such a view of the bytes without copying it (or resorting to Python API calls).

   (This is not done for `str` because when the str contains unicode we have to allocate to a temporary and so there might not be some string data we can properly view without owning.)

3. Allows `memoryview::from_memory` to accept a string_view.  As with the other from_memory calls, it's entirely your responsibility to keep it alive.

This also required moving the string_view availability detection into detail/common.h because this PR needs it in pytypes.h, which is higher up the include chain than cast.h where it was being detected currently.

## Suggested changelog entry:

```rst
* Make str/bytes/memoryview more interoperable with ``std::string_view``.
```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -244,9 +244,6 @@
 #    include <version>
 #  endif
 #endif
-#ifdef PYBIND11_HAS_STRING_VIEW
-#  include <string_view>
-#endif
 
 // #define PYBIND11_STR_LEGACY_PERMISSIVE
 // If DEFINED, pybind11::str can hold PyUnicodeObject or PyBytesObject
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -183,6 +183,21 @@
 #  define PYBIND11_HAS_VARIANT 1
 #endif
 
+#if defined(PYBIND11_CPP17)
+#  if defined(__has_include)
+#    if __has_include(<string_view>)
+#      define PYBIND11_HAS_STRING_VIEW
+#    endif
+#  elif defined(_MSC_VER)
+#    define PYBIND11_HAS_STRING_VIEW
+#  endif
+#endif
+
+#if defined(__cpp_lib_char8_t) && __cpp_lib_char8_t >= 201811L
+#  define PYBIND11_HAS_U8STRING
+#endif
+
+
 #include <Python.h>
 #include <frameobject.h>
 #include <pythread.h>
@@ -228,6 +243,9 @@
 #  if __has_include(<version>)
 #    include <version>
 #  endif
+#endif
+#ifdef PYBIND11_HAS_STRING_VIEW
+#  include <string_view>
 #endif
 
 // #define PYBIND11_STR_LEGACY_PERMISSIVE
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1090,8 +1090,10 @@ public:
     str(const std::string &s) : str(s.data(), s.size()) { }
 
 #ifdef PYBIND11_HAS_STRING_VIEW
+    // enable_if is needed to avoid "ambiguous conversion" errors (see PR #3521).
+    template <typename T, detail::enable_if_t<std::is_same<T, std::string_view>::value, int> = 0>
     // NOLINTNEXTLINE(google-explicit-constructor)
-    str(std::string_view s) : str(s.data(), s.size()) { }
+    str(T s) : str(s.data(), s.size()) { }
 
 # ifdef PYBIND11_HAS_U8STRING
     // reinterpret_cast here is safe (C++20 guarantees char8_t has the same size/alignment as char)
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -18,6 +18,10 @@
 #  include <optional>
 #endif
 
+#ifdef PYBIND11_HAS_STRING_VIEW
+#  include <string_view>
+#endif
+
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 /* A few forward declarations */
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1186,8 +1186,8 @@ public:
 
 #ifdef PYBIND11_HAS_STRING_VIEW
     // enable_if is needed to avoid "ambiguous conversion" errors (see PR #3521).
-    // NOLINTNEXTLINE(google-explicit-constructor)
     template <typename T, detail::enable_if_t<std::is_same<T, std::string_view>::value, int> = 0>
+    // NOLINTNEXTLINE(google-explicit-constructor)
     bytes(T s) : bytes(s.data(), s.size()) { }
 
     // Obtain a string view that views the current `bytes` buffer value.  Note that this is only
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1185,8 +1185,10 @@ public:
     }
 
 #ifdef PYBIND11_HAS_STRING_VIEW
+    // enable_if is needed to avoid "ambiguous conversion" errors (see PR #3521).
     // NOLINTNEXTLINE(google-explicit-constructor)
-    bytes(std::string_view s) : bytes(s.data(), s.size()) { }
+    template <typename T, detail::enable_if_t<std::is_same<T, std::string_view>::value, int> = 0>
+    bytes(T s) : bytes(s.data(), s.size()) { }
 
     // Obtain a string view that views the current `bytes` buffer value.  Note that this is only
     // valid so long as the `bytes` instance remains alive and so generally should not outlive the
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1085,6 +1085,18 @@ public:
     // NOLINTNEXTLINE(google-explicit-constructor)
     str(const std::string &s) : str(s.data(), s.size()) { }
 
+#ifdef PYBIND11_HAS_STRING_VIEW
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    str(std::string_view s) : str(s.data(), s.size()) { }
+
+# ifdef PYBIND11_HAS_U8STRING
+    // reinterpret_cast here is safe (C++20 guarantees char8_t has the same size/alignment as char)
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    str(std::u8string_view s) : str(reinterpret_cast<const char*>(s.data()), s.size()) { }
+# endif
+
+#endif
+
     explicit str(const bytes &b);
 
     /** \rst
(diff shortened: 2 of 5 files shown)


include/pybind11/pytypes.h
```diff
@@ -1085,6 +1085,17 @@ class str : public object {
     // NOLINTNEXTLINE(google-explicit-constructor)
     str(const std::string &s) : str(s.data(), s.size()) { }
 
+#ifdef PYBIND11_HAS_STRING_VIEW
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    str(std::string_view s) : str(s.data(), s.size()) { }
+
+# ifdef PYBIND11_HAS_U8STRING
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    str(std::u8string_view s) : str(s.data(), s.zie()) { }
```
Skylion007, line 1094: Is this a typo?
```suggestion
    str(std::u8string_view s) : str(s.data(), s.size()) { }
```
Can't find any reference to `.zie` in the std::library. If it is a typo and our test suite isn't picking it up that is a bit concerning and indicates that the constructor isn't ever compiled / is shadowed by another constructor and is never therefore called or compiled, right?

Skylion007: @henryiii Another reason why we should look into adding codecoverage to our CI even if only for gcc / clang.

include/pybind11/pytypes.h
```diff
@@ -1167,6 +1178,20 @@ class bytes : public object {
...
+
+#ifdef PYBIND11_HAS_STRING_VIEW
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    bytes(std::string_view s) : bytes(s.data(), s.size()) { }
+
+    operator std::string_view() const {
+        char *buffer = nullptr;
+        ssize_t length = 0;
+        if (PYBIND11_BYTES_AS_STRING_AND_SIZE(m_ptr, &buffer, &length))
+            pybind11_fail("Unable to extract bytes contents!");
```
rwgk, line 1190: `Unable to obtain a string_view for this object!`

?

rwgk (approved):
Nice!

rwgk:
The 2.7 ubuntu failure probably just needs a leading `u` (`u"i'm a string"`, a trick we use in many places).

I've never seen the ITERATOR_DEBUG_LEVEL failure before.

include/pybind11/pytypes.h
```diff
@@ -1167,6 +1178,20 @@ class bytes : public object {
             pybind11_fail("Unable to extract bytes contents!");
         return std::string(buffer, (size_t) length);
     }
+
+#ifdef PYBIND11_HAS_STRING_VIEW
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    bytes(std::string_view s) : bytes(s.data(), s.size()) { }
+
+    operator std::string_view() const {
```
rwgk, line 1186: This is a comment from an unfinished Google-internal PR Ken Oslund (sorry I cannot find his github handle) sent me a long time ago:
```
    // Python retains ownership of the string, so returned view is only valid
    // as long as the python object lives.
```

I think some version of this would be very useful to add to the added `operator std::string_view` in this PR.

tests/test_builtin_casters.py
```diff
@@ -206,6 +206,14 @@ def test_string_view(capture):
         """
         )
 
+    assert m.string_view_bytes() == b"abc \x80\x80 def"
+    assert m.string_view_str() == "abc ‽ def"
+    assert m.string_view_from_bytes("abc ‽ def".encode()) == "abc ‽ def"
+    if hasattr(m, "has_u8string"):
+        assert m.string_view8_str() == "abc ‽ def"
```
Skylion007, line 213:
```suggestion
    assert m.string_view_str() == u"abc ‽ def"
    assert m.string_view_from_bytes(u"abc ‽ def".encode()) == u"abc ‽ def"
    if hasattr(m, "has_u8string"):
        assert m.string_view8_str() == u"abc ‽ def"
```

include/pybind11/detail/common.h
```diff
@@ -183,6 +183,24 @@
...
+#  if defined(__has_include)
+#    if __has_include(<string_view>)
+#      define PYBIND11_HAS_STRING_VIEW
+#    endif
+#  elif defined(_MSC_VER)
+#    define PYBIND11_HAS_STRING_VIEW
+#  endif
+#endif
+#ifdef PYBIND11_HAS_STRING_VIEW
+#include <string_view>
```
Skylion007, line 196: There are not suppose to be any includes before "Python.h" please move this block down.

include/pybind11/pytypes.h
jagerman: No, I don't think so: if this fails it's because we couldn't get the pointer+length out of the underlying Python `bytes` object.  (I don't know under what situation this could actually fail, though).

include/pybind11/pytypes.h
jagerman: Added.

include/pybind11/pytypes.h
```diff
@@ -1091,7 +1091,7 @@ class str : public object {
 
 # ifdef PYBIND11_HAS_U8STRING
     // NOLINTNEXTLINE(google-explicit-constructor)
-    str(std::u8string_view s) : str(s.data(), s.size()) { }
+    str(std::u8string_view s) : str(reinterpret_cast<const char*>(s.data()), s.size()) { }
```
rwgk, line 1094: Is there a cleaner way to obtain a `char*` from a `char8_t*`?
I look at `reinterpret_cast` as a method of last resort.
If there is no alternative, adding a comment would be good, to explain why this is sufficiently safe and portable.
E.g.: // The C++ standard guarantees that sizeof(char) and sizeof(char8_t) are identical.

include/pybind11/pytypes.h
rwgk: Got it, thanks!

rwgk (approved):

include/pybind11/pytypes.h
rwgk: Looks like there is no alternative (sigh ... omg ... sigh).

https://stackoverflow.com/questions/57603013/how-to-safely-convert-const-char-to-const-char8-t-in-c20

jagerman: I've added a comment.

> Looks like there is no alternative (sigh ... omg ... sigh).

Yet another reason to dislike char8_t/u8string_view/etc.  It all feels a bit like a solution in search of an artificial problem that no one actually has *and* breaks code in the wild that actually thought it might be safe to use u"" to properly indicate utf-8 literals before C++20 (e.g. `std::string s = u8"asdf";` no longer compiles under -std=c++20).  Hurray.
\</rant\>

include/pybind11/detail/common.h
jagerman: Done.

jagerman:
Squashed it.

include/pybind11/detail/common.h
```diff
@@ -229,6 +244,9 @@
 #    include <version>
 #  endif
 #endif
+#ifdef PYBIND11_HAS_STRING_VIEW
+#  include <string_view>
```
Skylion007, line 248: Minor nit: To better follow an IWYU model, we should really only add the includes where they are needed in `pytypes.h and cast.h`

Skylion007 (approved):

Skylion007 (approved):

Skylion007:
@rwgk Does this pass the Google Test suite?

jagerman:
I don't understand why the 3.6 - windows-latest build here started failing.  Is that a known flakey build or something?

rwgk:
> I don't understand why the 3.6 - windows-latest build here started failing. Is that a known flakey build or something?

Yes, that's our most common known flake, **safe to ignore**. It's a lot better than it used to be before PR #2995, but it's still not stable.

There are other common flakes: 1. print from destructor; 2. various transient issues installing or downloading dependencies.

rwgk:
> @rwgk Does this pass the Google Test suite?

I'll initiate this now (will take several hours).

henryiii (approved):
Happy when Google's happy. :)

rwgk:
> Happy when Google's happy. :)

I missed the 12:00 global testing opportunity, but got this PR into the 16:00 batch. Results are expected about 4-5 hours later.

I did comprehensive manual testing before sending this PR for global testing, including testing with ASAN. No issues.

rwgk:
This PR didn't make it through the basic "smoke check" :-(
Below is one breakage. I still have to try to understand myself.
IIRC we recently had some trouble with `tensorflow::tstring` already, I forget what exactly ...

Error message:

```
third_party/tensorflow/python/lib/io/file_io_wrapper.cc:316:21: error: ambiguous conversion for functional-style cast from 'tensorflow::tstring' to 'py::bytes'
             return py::bytes(result);
                    ^~~~~~~~~~~~~~~~
./third_party/pybind11/include/pybind11/detail/../pytypes.h:1174:5: note: candidate constructor
    bytes(const std::string &s) : bytes(s.data(), s.size()) { }
    ^
./third_party/pybind11/include/pybind11/detail/../pytypes.h:1189:5: note: candidate constructor
    bytes(std::string_view s) : bytes(s.data(), s.size()) { }
    ^
1 error generated.
```

While compiling:

```
      .def("read",
           [](BufferedInputStream* self, int64_t bytes_to_read) {
             py::gil_scoped_release release;
             tensorflow::tstring result;
             const auto status = self->ReadNBytes(bytes_to_read, &result);
             if (!status.ok() && !tensorflow::errors::IsOutOfRange(status)) {
               result.clear();
               tensorflow::MaybeRaiseRegisteredFromStatusWithGIL(status);
             }
             return py::bytes(result);
           })
```

rwgk:
Trying a fix. I don't know if this is the best approach, but at least it fixes the failures in the wild (at least some; I still have to try the global testing again).

The force push was needed because I also rebased on master.

rwgk:
I think this will work now, pending full global testing; just initiated. Results when I wake up (hopefully).

I already verified that all failing targets from the previous sample run pass with the latest version of this PR.

jagerman:
> I think this will work now, pending full global testing; just initiated. Results when I wake up (hopefully).

`str()` will have the same issue; I've pushed an update to apply the workaround there, too.

jagerman:
I'd kind of prefer that, in the case of ambiguous conversions, we were preferring string_view to string (because it likely avoids an extra allocation), but I don't really see a nice way to do that without introducing potential breakage for existing code.

rwgk:
> I'd kind of prefer that, in the case of ambiguous conversions, we were preferring string_view to string (because it likely avoids an extra allocation), but I don't really see a nice way to do that without introducing potential breakage for existing code.

Initially I tried `#ifdef`ing out the `std::string` constructor, which also fixed the `tensorflow::tstring` issue in the wild, but broke existing unit tests. With a pushing-and-shoving mentality I tried to rescue it, but quickly got into a mess and gave up.

rwgk:
The one CI failure is a flake (Run jwlawson/actions-setup-cmake@v1.11 Error: connect ETIMEDOUT 140.82.114.6:443).

Google-global testing was successful (it did not run into the `str` issue fixed here by https://github.com/pybind/pybind11/pull/3521/commits/5709ccfae019a4b32e7b910f8f52f56352145560 in the meantime).

Definitely good to merge Jason!

PR #3548, opened 2021-12-12
Ensure ThreadPool is closed in setup_helpers

bobbyi:
The ParallelCompile setup helper using a ThreadPool to enable its parallelism. It does not properly close the pool when it is done with it.

This can lead to a `Exception ignored in: <function Pool.__del__...` message with traceback being printed at shutdown.

### DIFF
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -466,8 +466,12 @@ class ParallelCompile(object):
                     threads = 1
 
             if threads > 1:
-                for _ in ThreadPool(threads).imap_unordered(_single_compile, objects):
-                    pass
+                pool = ThreadPool(threads)
+                try:
+                    for _ in pool.imap_unordered(_single_compile, objects):
+                        pass
+                finally:
+                    pool.terminate()
             else:
                 for ob in objects:
                     _single_compile(ob)
--- a/pybind11/setup_helpers.py
+++ b/pybind11/setup_helpers.py
@@ -467,6 +467,8 @@ class ParallelCompile(object):
 
             if threads > 1:
                 pool = ThreadPool(threads)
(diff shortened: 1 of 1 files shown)


henryiii (approved):

bobbyi:
@henryiii I just updated this PR because CI failed against Python 2.7. Turns out that the pool does not support being used as a context manager in that version. I've changed it to use a finally block that calls `pool.terminate`, which is what the pool's `__exit__` method does

rwgk:
I've not seen the "ICC latest" error before. Seems unrelated to this PR. Maybe coincidentally the compiler was just updated?

henryiii:
I'd either add a note or maybe even use an if statement - we are very close to dropping Python 2 support (Jan 1) and long term I'd like the nicer syntax. There should be something making it easy to remember to clean up.

bobbyi:
@henryiii If it's only two weeks or so until Python 2 support is dropped, I'd lean towards reverting this back to the first version I had and having you not merge it until then.

henryiii:
I'd like to get as many fixes as possible in before dropping Py 2. So how about we merge this one, then you could make a PR to the nicer version that we merge in 2 weeks? :)

bobbyi:
@henryiii Sounds good. I've added a comment that mentions "Python 2". There are already couple others in this same file (related to Distutils using old-style class in Python 2) that can maybe be cleaned up at the same time as this one. This should be ready to merge.

henryiii:
The note is fine. :) I'll be expecting the old-style class stuff, but won't be likely to remember this context manager simplification without the note.

henryiii:
Please do not delete the changelog entry portion of the template. It fills our changelog. (fixed manually).

henryiii:
Thanks!

PR #3571, opened 2021-12-23
fix: restore full range of _ functions

henryiii:
## Description

CC @rwgk - this should restore the missing forms of `_`. I wish C++ supported function aliases! :)


Skylion007 (approved):

rwgk (approved):
Approving to get us back on track.
Just noting two obvious non-idealisms:
* Significant duplication of non-trivial code.
* Missing test coverage

But let's take care of that separately.

henryiii:
This is a strict improvement, so putting it in to make it easier to test. Let's see what it looks like with this.

Help writing a better dedicated test appreciated. :)

PR #3577, opened 2021-12-28
fix(setup): support overriding CMake args

henryiii:
## Description

Supports setting CMAKE_ARGS to control the CMake file generation. Needed for conda-forge, which does not find the compiler otherwise.

See https://github.com/conda-forge/pybind11-feedstock/pull/74.

## Suggested changelog entry:

```rst
* Allow CMAKE_ARGS to override CMake args in pybind11's own setup.py.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -146,6 +146,8 @@ with remove_output("pybind11/include", "pybind11/share"):
             "-DBUILD_TESTING=OFF",
             "-DPYBIND11_NOPYTHON=ON",
         ]
+        if "CMAKE_ARGS" in os.environ:
+            cmd += os.environ["CMAKE_ARGS"].split()
         cmake_opts = dict(cwd=DIR, stdout=sys.stdout, stderr=sys.stderr)
         subprocess.check_call(cmd, **cmake_opts)
         subprocess.check_call(["cmake", "--install", tmpdir], **cmake_opts)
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):
We do this for Habitat-Sim as well. Surprised this issue wasn't already brought up and fixed.

henryiii:
(On hold until I work out the actual problem with conda-forge - I think they are going 2017 -> 2019 and the transition is broken). If the original code works after this is sorted, I'd be up in the air about putting this in - we might want the ability to override the CMake command, so I think I'd still be in favor of it.

henryiii:
> Surprised this issue wasn't already brought up and fixed.

Building from true source is very rare - honestly, I probably could move conda-forge over to using the SDist, and not bother with the build from git source. However, I think the reason I don't do that is then I'd have two source files and it would be harder to automate the maintenance.

henryiii:
This sort of patch is not needed to support conda-forge (see https://github.com/conda-forge/pybind11-feedstock/pull/75). It might still be a good idea, though?

setup.py
```diff
@@ -146,6 +146,8 @@ def remove_output(*sources):
             "-DBUILD_TESTING=OFF",
             "-DPYBIND11_NOPYTHON=ON",
         ]
+        if "CMAKE_ARGS" in os.environ:
+            cmd += os.environ["CMAKE_ARGS"].split()
```
henryiii, line 150:
```suggestion
            fcommand = [c for c in os.environ["CMAKE_ARGS"].split() if "DCMAKE_INSTALL_PREFIX" not in c]
            cmd += fcommand
```

PR #3578, opened 2021-12-28
Adding dedicated test_const_name.

rwgk:
## Description
Also exercises `pybind11::detail::_` backward compatibility.

See also: PR #3423

Notes:

* clang-format was applied to test_const_name.cpp
* Testing with MSVC 2015 is skipped due to bizarre errors that are not worth debugging. Completeness of the duplicated code for `pybind11::detail::_` is sufficiently exercised on the many other platforms.
* `const_name<0>()` and similar only works with C++17 or higher ([source code commented added](https://github.com/pybind/pybind11/pull/3578/commits/0f763169458dbc71a5d461f753bb9126bab1ffc9)). This functionality is not actually used, therefore the corresponding tests were backed out. 

An easy way to disable backward compatibility is to patch include/pybind11/detail/descr.h like this:
```diff
-#ifndef _
-#define PYBIND11_DETAIL_UNDERSCORE_BACKWARD_COMPATIBILITY
+#if defined(_) && defined(PYBIND11_DETAIL_UNDERSCORE_BACKWARD_COMPATIBILITY)
```
This can be useful after a cleanup (i.e. replacing `_` with `const_name`), to prevent backsliding.

## Suggested changelog entry:

```rst

```


tests/test_const_name.py
```diff
@@ -0,0 +1,35 @@
...
+import itertools
+
+import pytest
+
+from pybind11_tests import const_name as m
+
+
+@pytest.mark.parametrize(
+    "func, selector_expected",
+    itertools.product(
```
Skylion007, line 11: If you parameterize `func`, and `selector_expected` separately, it will implicitly do an itertools.product between them.

rwgk: Done. Thanks!

Skylion007 (approved):

henryiii (approved):

rwgk:
Thank you for the reviews!

pastasonica:
Gg

PR #3581, opened 2021-12-29
[smart_holder] .def_readonly, .def_readwrite adaptors

rwgk:
## Description
The original motivation for this PR was to fix the issue observed here: [disowning creates a dangling pointer](https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/testing/python/nested_fields_test.py#L56).

To ensure internal consistency down the road, the scope of this PR was expanded to cover `.def_readonly` and `.def_readwrite` behavior for

* value
* raw pointer
* `std::unique_ptr`
* `std::shared_ptr`

members (mutable & const for the pointer types).

For `shared_ptr` members, the existing behavior is as desired (no work needed).

For value and raw-pointer members, the [PyCLIF approach here](https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/python/instance.h#L233) was adopted: aliasing `shared_ptr` tying the lifetime of the "outer" object to the lifetime of the Python object pointing to the member. In this way the existing mechanisms prevent the "outer" object from being disowned while pointers to members are still alive.

For unique_ptr members, the situation is complex. This PR only establishes a proof of concept. See the code and comments in tests/test_class_sh_property.py for details. Productizing the proof of concept is left for a future separate PR, as needed.


rwgk:
To track progress (FYI @wangxf123456):

I've verified that the [PyCLIF TODO here](https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/testing/python/nested_fields_test.py#L56) can be removed when using this PR (ASAN clean).

```diff
     a = nested_fields.AA()
     a.b.c.i = 123
     c = a.b.c
-    if 'pybind11' not in nested_fields.__doc__:
-      with self.assertRaises(ValueError):
-        # Cannot give up |a| as |c| still is alive
-        nested_fields.ConsumeAA(a)
-      del c
-      # Since |c| is now gone, we can give up |a|
+    with self.assertRaises(ValueError):
+      # Cannot give up |a| as |c| still is alive
       nested_fields.ConsumeAA(a)
-    else:
-      # TODO(rwgk): This disowns |a| even though |c| is still alive. Accessing
-      # |c| beyond this point generates an ASAN heap-use-after-free error.
-      nested_fields.ConsumeAA(a)
+    del c
+    # Since |c| is now gone, we can give up |a|
+    nested_fields.ConsumeAA(a)

     a = nested_fields.AA()
     a.b.c.i = 123
```

include/pybind11/pybind11.h
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+
+// smart_holder specializations for members held by-value.
+// The read functions return a shared_ptr to the member, emulating PyCLIF approach:
+// https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/python/instance.h#L233
+// This prevents disowning of the Python object owning the member.
+template <typename T, typename D>
+struct xetter_cpp_function<
+    T,
+    D,
+    detail::enable_if_t<detail::type_uses_smart_holder_type_caster<T>::value
```
Skylion007, line 1446: This would be cleaner using detail::any_of

include/pybind11/pybind11.h
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+// read disowns the member unique_ptr.
+// write disowns the passed Python object.
+// readonly is disabled (static_assert) because there is no safe & intuitive way to make the member
+// accessible as a Python object without disowning the member unique_ptr. A .def_readonly disowning
+// the unique_ptr member is deemed highly prone to misunderstandings.
+template <typename T, typename D>
+struct xetter_cpp_function<
+    T,
+    D,
+    detail::enable_if_t<
```
Skylion007, line 1487: Also could be cleaner using detail::any_of

include/pybind11/pybind11.h
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+// This prevents disowning of the Python object owning the member.
+template <typename T, typename D>
+struct xetter_cpp_function<
+    T,
+    D,
+    detail::enable_if_t<detail::type_uses_smart_holder_type_caster<T>::value
+                        && detail::type_uses_smart_holder_type_caster<D>::value
+                        && !std::is_pointer<D>::value           //
+                        && !detail::is_std_unique_ptr<D>::value //
+                        && !detail::is_std_shared_ptr<D>::value>> {
```
Skylion007, line 1450:
```suggestion
    detail::enable_if_t<detail::all_of<
                        detail::type_uses_smart_holder_type_caster<T>,
                        detail::type_uses_smart_holder_type_caster<D>,
                        !std::is_pointer<D>,           //
                        !detail::is_std_unique_ptr<D>, //
                        !detail::is_std_shared_ptr<D>>>::value> {

File: include/pybind11/pybind11.h [hunk 4]
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+        return cpp_function(
+            [pm](const std::shared_ptr<T> &c_sp) -> std::shared_ptr<drp> {
+                D ptr = (*c_sp).*pm;
+                return std::shared_ptr<drp>(c_sp, ptr);
+            },
+            is_method(hdl));
+    }
+
+    template <typename PM>
+    static cpp_function read(PM pm, const handle &hdl) {
```
Skylion007, line 1428: If you don't change the call signature, you should perfect forward PM.

include/pybind11/pybind11.h
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+    T,
+    D,
+    detail::enable_if_t<detail::type_uses_smart_holder_type_caster<T>::value
+                        && detail::type_uses_smart_holder_type_caster<D>::value
+                        && std::is_pointer<D>::value>> {
+
+    using drp = typename std::remove_pointer<D>::type;
+
+    template <typename PM>
+    static cpp_function readonly(PM pm, const handle &hdl) {
```
Skylion007, line 1418:
```suggestion
    static cpp_function readonly(PM &pm, const handle &hdl) {
```
and capture by reference in the lambda. At the very least, the PM should be forwarded otherwise.

I think original pybind11 it's a pointer. being passed around.

This could invoke copy or moves depending on the type PMs.

include/pybind11/pybind11.h
```diff
@@ -1374,6 +1374,141 @@ using default_holder_type = smart_holder;
...
+// WARNING: Like the classic implementation, this implementation can lead to dangling pointers.
+// See test_ptr() in tests/test_class_sh_property.py
+// However, the read functions return a shared_ptr to the member, emulating PyCLIF approach:
+// https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/python/instance.h#L233
+// This prevents disowning of the Python object owning the raw pointer member.
+template <typename T, typename D>
+struct xetter_cpp_function<
+    T,
+    D,
+    detail::enable_if_t<detail::type_uses_smart_holder_type_caster<T>::value
```
Skylion007, line 1411: This can also be done cleaner with `detail::all_of`

Skylion007 (changes requested):
@rwgk Made some nits about potentially bugs that could arise from the typing of the templates and that all the helpers could benefit from using the detail::all_of helper defined in details/commons.h

include/pybind11/pybind11.h
rwgk: Done.

include/pybind11/pybind11.h
rwgk: Thanks for pointing this out! It looks a lot nicer.

include/pybind11/pybind11.h
rwgk: See other reponse.

include/pybind11/pybind11.h
rwgk: I couldn't get this to work. I tried a few variations of sprinkling in `&`, `&&`, `std::forward<PM>`, but either it didn't build or I got segfaults. Then I decided it's better to just make it obvious that we're only dealing with pointers anyway, and the perfect forwarding doesn't buy us anything. See the long added comment for the new `must_be_member_function_pointer` SFINAE helper.

include/pybind11/pybind11.h
rwgk: Done.

rwgk:
The CI failures are sure to be flakes or (hopefully transient) infrastructure failures. The previous CI run had no failures. I only added a couple comments (and ran `pre-commit run --all-files` before pushing).

tests/test_class_sh_property.py
```diff
@@ -0,0 +1,163 @@
...
+@pytest.mark.parametrize("m_attr", ("m_valu_readonly", "m_valu_readwrite"))
+def test_valu_getter(msg, m_attr):
+    # Reduced from PyCLIF test:
+    # https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/testing/python/nested_fields_test.py#L56
+    outer = m.Outer()
+    field = getattr(outer, m_attr)
+    assert field.num == -99
+    with pytest.raises(ValueError) as excinfo:
+        m.DisownOuter(outer)
+    assert msg(excinfo.value) == "Cannot disown use_count != 1 (loaded_as_unique_ptr)."
```
rainwoodman, line 18: The test case is clear and well defined behavior.

My concern is on the practical side -- it is hard to know where the reference to the object is. field could have been passed and kept in many remote part of the application.

A very canonical pattern I know of, for this type of situation is to mark the object as stale, and all uses of it errors. A very typical example is the 'NFS stale handle' error. Then the responsibility is distributed to the consumers of the inner objects, rather than the owner needing to reason a large code base 'who is reference my inner parts?'.

tests/test_class_sh_property.py
```diff
@@ -0,0 +1,163 @@
...
+    del field
+    m.DisownOuter(outer)
+    with pytest.raises(ValueError) as excinfo:
```
rainwoodman, line 21: ` pytest.raises(ValueError, match=r".* 123 .*")`?

tests/test_class_sh_property.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const ClassicField *m_cptr = nullptr;
+};
+
+struct Field {
+    int num = -99;
+};
+
+struct Outer {
+    // The compact 4-character naming matches that in test_class_sh_basic.cpp
+    // (c = const, m = mutable).
```
charlesbeattie, line 24: Maybe comment sh = shared_ptr and uq = unique_ptr as well.

tests/test_class_sh_property.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include <memory>
+
+namespace test_class_sh_property {
+
+struct ClassicField {
+    int num = -88;
+};
+
+struct ClassicOuter {
```
charlesbeattie, line 13: Is the `Classic` prefix imply that this lass in not wrapped in `PYBIND11_SMART_HOLDER_TYPE_CASTERS`. A comment above each of these would help with readability.

tests/test_class_sh_property.cpp
```diff
@@ -0,0 +1,90 @@
+#include "pybind11_tests.h"
+
+#include "pybind11/smart_holder.h"
```
charlesbeattie, line 3: I've seen
`#include <pybind11/smart_holder.h>`
Elsewhere is there a consistent choice we should make with these includes?

charlesbeattie:
Made some small style comments othwise it looks good to me.

rwgk:
I'm continuing this work under PR #3844. See the **Note** at top of the PR description there.

PR #3588, opened 2022-01-03
[smart_holder] fix: names of types held by smart holder should be the actual type

virtuald:
Required for pybind11-stubgen to work properly


rwgk:
Thanks Dustin!

Issue #3600, opened 2022-01-07
[BUG]: Regression from 2.5.0 to 2.6.0: Existance of operator breaks binding of some collections


marlamb:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

In the following example the `Struct` has a Python binding and is convertible on its own. Also `std::unordered_set` works in general for build in types. However for the combination `std::unordered_set<Struct>` the conversion fails at runtime with the error by calling `test_func()`
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: Unable to convert function return value to a Python type! The signature was
	() -> Set[test_module.Struct]
```

 * The removal of `py::is_operator()` in the binding fixes the problem.
 * The example does work for version 2.5.0, but fails for 2.6.0, 2.7.1, 2.8.1 and 2.9.0.
 * It does also not work for an `std::set`, but for an `std::vector`.
 * The code was compiled with `clang13.0.0` on `x86_64` (Ubuntu20.04)

### Reproducible example code

```cpp
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>

namespace py = ::pybind11;

struct Struct
{
  std::string member;
};

bool operator==(const Struct& lhs, const Struct& rhs)
{
  return lhs.member == rhs.member;
}

namespace std
{
template <>
struct hash<Struct>
{
  std::size_t operator()(const Struct& selector) const;
};
}  // namespace std

std::size_t std::hash<Struct>::operator()(const Struct& selector) const
{
  return std::hash<std::string>()(selector.member);
}

PYBIND11_MODULE(test_module, module)
{
  py::class_<Struct>(module, "Struct")
      // Removing the binding makes the example work.
      .def(
          "__eq__",
          [](const Struct& lhs, const Struct& rhs) {
            return lhs == rhs;
          },
          py::is_operator());

  module.def("test_func", []() {
    return std::unordered_set<Struct>{{"one"}, {"two"}};
  });
}
```

Skylion007:
[...]

PR #3605, opened 2022-01-09
Add additional info to TypeError when C++->Python casting fails

Skylion007:
## Description
Adds additional information when raising a type error during function return. Inspired by the unexpected behavior from #3600 .

## Suggested changelog entry:

```rst
* If possible, attach Python exception with ``py::raise_from`` to TypeError when casting from C++ to Python. This will give additional info if Python exceptions occur in the caster. Adds a test case of trying to convert a set from C++ to Python when the hash function is not defined in Python. 
```

As a meta note, this is an issue because we are a bit inconsistent in how we handle Python exceptions with ``py::set``, ``py::dict``, etc...). We currently not raising an exception from Python to C++ when interacting with them, maybe we should consider doing that.

### DIFF
--- a/tests/test_operator_overloading.cpp
+++ b/tests/test_operator_overloading.cpp
@@ -7,10 +7,11 @@
     BSD-style license that can be found in the LICENSE file.
 */
 
-#include "pybind11_tests.h"
 #include "constructor_stats.h"
-#include <pybind11/operators.h>
+#include "pybind11_tests.h"
 #include <functional>
+#include <pybind11/operators.h>
+#include <pybind11/stl.h>
 
 class Vector2 {
 public:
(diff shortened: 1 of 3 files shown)


tests/test_operator_overloading.cpp
```diff
@@ -82,6 +89,13 @@ namespace std {
         // Not a good hash function, but easy to test
         size_t operator()(const Vector2 &) { return 4; }
     };
+
+    template <>
```
rwgk, line 93: Might be nice to add a comment (if you feel it's useful):
```
// HashMe has a hash function in C++ but no `__hash__` for Python.
```

tests/test_operator_overloading.py
```diff
@@ -144,3 +145,10 @@ def test_overriding_eq_reset_hash():
 
         assert hash(hashable(15)) == 15
         assert hash(hashable(15)) == hash(hashable(15))
+
+
+def test_return_set_of_unhashable():
+    with pytest.raises(TypeError) as excinfo:
```
rwgk, line 151: Also optional: while we're at it, updating lines 139-140 accordingly would be nice. (I'd delete the then-redundant comment there.)

rwgk (approved):
Thanks Aaron!

tests/test_operator_overloading.py
Skylion007: @rwgk I did even better and added it as an explicit test.

henryiii (approved):

PR #3179, opened 2021-08-07
PYBIND11_NOINLINE-related cleanup.

rwgk:
Pure cleanup for correctness and consistency.

Motivated by ongoing work on removing GCC `-Wattributes` from the pragma block at the top of pybind11.h (see PR #3125 for full context), but technically completely independent and a worthwhile cleanup in its own right.

Correctness aspect:
* Removal of incorrectly placed `inline`.
* Relevant background: https://stackoverflow.com/questions/9317473/forward-declaration-of-inline-functions
* **This resolves GCC 9, 10, 11 warnings that surface after removal of GCC `-Wattributes` from the pragma block at the top of pybind11.h**. That's not obvious with just this PR, but is based on the data collected under PR #3125.

Consistency aspect:
* Current master has 4 x `inline PYBIND11_NOINLINE` and 17 x `PYBIND11_NOINLINE inline`. See lists below.
* To remove the potential for such inconsistencies, and to make the code nicer to read, `inline` is added to the `PYBIND11_NOINLINE` macro.
_______________________________________________________________________________

Pointing out a special case in this PR (the only special case):

Converted from `inline` to `PYBIND11_NOINLINE`, for consistency with an overload:
```
pybind11.h:inline void keep_alive_impl(handle nurse, handle patient) {
```
_______________________________________________________________________________

Current master:

`inline` first:
```
numpy.h:inline PYBIND11_NOINLINE void load_numpy_internals(numpy_internals* &ptr) {
numpy.h:inline PYBIND11_NOINLINE void register_structured_dtype(
detail/internals.h:inline PYBIND11_NOINLINE void *get_shared_data(const std::string &name) {
detail/internals.h:inline PYBIND11_NOINLINE void *set_shared_data(const std::string &name, void *data) {
```

`inline` second:
```
pybind11.h:PYBIND11_NOINLINE inline void keep_alive_impl(size_t Nurse, size_t Patient, function_call &call, handle ret) {
pybind11.h:PYBIND11_NOINLINE inline void print(const tuple &args, const dict &kwargs) {
detail/type_caster_base.h:PYBIND11_NOINLINE inline void all_type_info_populate(PyTypeObject *t, std::vector<type_info *> &bases) {
detail/type_caster_base.h:PYBIND11_NOINLINE inline detail::type_info* get_type_info(PyTypeObject *type) {
detail/type_caster_base.h:PYBIND11_NOINLINE inline detail::type_info *get_type_info(const std::type_index &tp,
[...]

PR #3608, opened 2022-01-10
Add support for nested C++11 exceptions

### LINKED ISSUE #1913: Map std::nested_exceptions to Python equivalent
I want to collect ideas whether and how one could map [`std::nested_exception`s](https://en.cppreference.com/w/cpp/error/nested_exception) to Python's chained exceptions from a [`raise ... from`](https://www.python.org/dev/peps/pep-3134/) clause and vice versa. These seem to be the corresponding language features in the two languages, and hence could be an interesting addition to pybind. My expectation is that this would provide more information in case of exceptions that cross language barriers.

I have done only very initial research, but if there is interest from the side of this project, I might be able to invest some more time, potentially with ideas and help from others.

Skylion007:
## Description
* This PR closes #1913 and maps C++11 nested exception to their Python equivalent in Python 3.3 and greater using the newly introduced raise_from utility.

## Suggested changelog entry:

```rst
* Adds a mapping of C++11 nested exceptions to their Python exception equivalent using ``py::raise_from``. This attaches the nested exceptions in Python using the ``__cause__`` field.
```

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -327,7 +327,7 @@ void handle_nested_exception(T, std::exception_ptr) {
 
 inline bool raise_err(PyObject *exc_type, const char *msg) {
     PyErr_SetString(exc_type, msg);
-    return false
+    return false;
 }
 #endif
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -10,7 +10,6 @@
 #pragma once
 
 #include "../pytypes.h"
-#include <__threading_support>
 #include <exception>
 #include <type_traits>
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -382,7 +382,7 @@ inline void translate_exception(std::exception_ptr p) {
         raise_err(PyExc_RuntimeError, e.what());
         return;
     } catch (const std::nested_exception &e) {
-        handle_nested_exception(e, p);.
+        handle_nested_exception(e, p);
         raise_err(PyExc_RuntimeError, "Caught an unknown nested exception!");
     } catch (...) {
         raise_err(PyExc_RuntimeError, "Caught an unknown exception!");
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -283,7 +283,7 @@ inline internals **&get_internals_pp() {
 
 #if PY_VERSION_HEX >= 0x03030000
 inline bool raise_err(PyObject *exc_type, const char *msg) {
-    bool err_occurred = PyErr_Occurred();
+    bool err_occurred = PyErr_Occurred() != nullptr;
     if (err_occurred) {
         raise_from(exc_type, msg);
     } else {
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -321,7 +321,7 @@ bool handle_nested_exception(const T &exc, const std::exception_ptr &p) {
 }
 #else
 template <class T>
-void handle_nested_exception(T, std::exception_ptr p) {
+void handle_nested_exception(T, std::exception_ptr) {
     return;
 }
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -384,6 +384,7 @@ inline void translate_exception(std::exception_ptr p) {
     } catch (const std::nested_exception &e) {
         handle_nested_exception(e, p);
         raise_err(PyExc_RuntimeError, "Caught an unknown nested exception!");
+        return;
     } catch (...) {
         raise_err(PyExc_RuntimeError, "Caught an unknown exception!");
         return;
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -282,15 +282,6 @@ inline internals **&get_internals_pp() {
 }
 
 #if PY_VERSION_HEX >= 0x03030000
-inline bool raise_err(PyObject *exc_type, const char *msg) {
-    if (PyErr_Occurred()) {
-        raise_from(exc_type, msg);
-        return true;
-    }
-    PyErr_SetString(exc_type, msg);
-    return false;
-};
-
 // forward decl
 inline void translate_exception(std::exception_ptr);
 
@@ -315,16 +306,23 @@ bool handle_nested_exception(const T &exc, const std::exception_ptr &p) {
 }
 
 #else
-inline bool raise_err(PyObject *exc_type, const char *msg) {
-    PyErr_SetString(exc_type, msg);
-    return false;
-}
 
 template <class T>
-bool handle_nested_exception(const T &, std::exception_ptr) {
+bool handle_nested_exception(const T &, std::exception_ptr &) {
     return false;
 }
 #endif
+
+inline bool raise_err(PyObject *exc_type, const char *msg) {
+#if PY_VERSION_HEX >= 0x03030000
+    if (PyErr_Occurred()) {
+        raise_from(exc_type, msg);
+        return true;
+    }
+#endif
+    PyErr_SetString(exc_type, msg);
+    return false;
+};
 
 inline void translate_exception(std::exception_ptr p) {
     if (!p) {
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -11,7 +11,6 @@
 
 #include "../pytypes.h"
 #include <exception>
-#include <type_traits>
 
 /// Tracks the `internals` and `type_info` ABI version independent of the main library version.
 ///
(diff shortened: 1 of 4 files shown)


rwgk:
Seems to be mostly working, but MSVC is unhappy, and possibly there is a separate issue for Python 2.

Which reminds me: let's get rid of Python 2 asap!

Are you looking at the MSVC errors already?

Skylion007:
@rwgk The python2 failures are missing skipif, I am less sure about MSVC, but I would like to remove ```rethrow_if_nested``` since I only need to do a convert std::exception to a nested_exception if I can do and access the nested_ptr member. However, I keep getting some weird bugs when I enable the templates for that. Based on the implementation from this link: https://en.cppreference.com/w/cpp/error/rethrow_if_nested

Skylion007:
@rwgk We are all green and I further optimized and simplified the code.

include/pybind11/detail/internals.h
```diff
@@ -281,20 +282,59 @@ inline internals **&get_internals_pp() {
...
+            raise_err = raise_from;
+        }
+    } catch (...) {
+    }
+#endif
+    try {
+        std::rethrow_exception(p);
+    } catch (error_already_set &e) {
+        e.restore();
+        return;
```
rwgk, line 308: Are the many `return;` needed (to make some compiler(s) happy)?
Would be nice to get rid of them. Or add a comment (maybe) right before the `try`, something like:

// The returns inside the catch blocks are needed to silence compiler warnings.
(If that's true.)

include/pybind11/detail/internals.h
```diff
@@ -281,20 +282,59 @@ inline internals **&get_internals_pp() {
...
+            translate_exception(nested);
+            raise_err = raise_from;
+        }
+    } catch (...) {
+    }
+#endif
+    try {
+        std::rethrow_exception(p);
+    } catch (error_already_set &e) {
+        e.restore();
```
rwgk, line 307: Here and for `builtin_exception` below `raise_err` isn't referenced.
Could these two `catch` be moved up?
Hm, it's getting a bit tricky with the #ifdef unfortunately.
We're almost removing 2.7 and 3.5 support, which would make that idea straightforward.
If you think my suggestion has merit, could you please add a comment, e.g.

// TODO: Move `error_already_set` and `builtin_exception` `catch` blocks up after removal of Python 2 & 3.5 support (because `raise_err` is not needed).

rwgk (approved):
Cool!

include/pybind11/detail/internals.h
Skylion007: I don't know why the return's are here. There were here before the diff.

include/pybind11/detail/internals.h
Skylion007: I suppose they could be moved up. The only purpose of the first try catch is to figure out if an exception_ptr refers to a nested_exception and recur down on the stack for it. The two try blocks could be merged if I duplicated the catch logic for the handling of the nested_exception.

(Alternatively, I could write a function that handles a nested exception on every single catch block. That would remove the try catch, but complicate other parts like raise-err selection etc.

Skylion007: I could just duplicate these two catch statements, it's only 4 lines of code and the return code should ensure that it's not run twice.

rwgk: > I could just duplicate these two catch statements, it's only 4 lines of code and the return code should ensure that it's not run twice.

Hm... I'd be OK with that, but when we remove 2.7 & 3.5 support we have to keep that code and remove it further down. I'd probably settle on a TODO comment, just enough so we don't forget when we're done with the more sweeping changes.

For the returns, good to clean out later (maybe).

henryiii (approved):
After the missing comment (or however that is resolved), this is good by me.

include/pybind11/detail/internals.h
Skylion007: @rwgk Happy to do that but it complicates the logic a bit.
I also realized I can't actually move it up or it will cause bugs if either of those errors are thrown in a nested exception (only 1 catch statement executes in C++ per throw to my knowledge). The other thing I wanted feedback is that I could disable this nested exception throwing with an optional potentially if there is interest.

Skylion007:
@rwgk @henryiii I refactored it a bit to remove the need to rethrow the exception. 
It does add a bit of code duplication, but the code is quite sensible now.

include/pybind11/detail/internals.h
rwgk: > I also realized I can't actually move it up or it will cause bugs if either of those errors are thrown in a nested exception (only 1 catch statement executes in C++ per throw to my knowledge).

oh ... now I'm thinking I need to read up on `std::nested_exception`.

big picture thinking: the simpler the better => I really hope we don't need an option to disable the feature.

henryiii: This is already opt-in, that is, if you don't use nested_exceptions in C++, you won't be using the feature here. (FWIU)

Skylion007: Yeah refactored it now so the performance penalty is minimal (no extra std::rethrow)

include/pybind11/detail/internals.h
```diff
@@ -280,21 +281,112 @@ inline internals **&get_internals_pp() {
...
+    if (err_occurred) {
+        raise_from(exc_type, msg);
+    } else {
+        PyErr_SetString(exc_type, msg);
+    }
+    return err_occurred;
+};
+
+// forward decl
+inline void translate_exception(std::exception_ptr);
```
rwgk, line 296: Forward declarations should never have `inline`:

https://stackoverflow.com/questions/9317473/forward-declaration-of-inline-functions

(I find this highly counter-intuitive myself.)

include/pybind11/detail/internals.h
```diff
@@ -280,21 +281,112 @@ inline internals **&get_internals_pp() {
...
+    }
+    return false;
+}
+
+template <
+    class T,
+    enable_if_t<!std::is_same<std::nested_exception, remove_cv_t<remove_reference_t<T>>>::value,
+                int> = 0>
+bool handle_nested_exception(const T &exc, const std::exception_ptr &p) {
+    if (auto *nep = dynamic_cast<const std::nested_exception *>(std::addressof(exc))) {
```
rwgk, line 317: Cooool!
I'd probably add
```
assert(nep != nullptr);
```
as a sanity check. (overhead only in debug mode, but even then it's negligible)

include/pybind11/detail/internals.h
```diff
@@ -280,21 +281,112 @@ inline internals **&get_internals_pp() {
...
+        return handle_nested_exception(*nep, p);
+    }
+    return false;
+}
+#else
+template <class T>
+void handle_nested_exception(T, std::exception_ptr) {
```
rwgk, line 324: This is `void` while the >=3.3 version returns a `bool`. Is that intentional? A terse comment to explain would be good.

rwgk (approved):
I want to look more, but don't let that stop you. Unfortunately I have only super thin time slices at the moment.

include/pybind11/detail/internals.h
henryiii: `if` literally checks for this being nullptr, I don't think we need to add an assert too. If a reader doesn't understand the `if`, they might not understand the `assert` either. A comment could be added if you _really_ want it, but I'm not a fan of explaining C++ syntax in comments unless it's obscure or easy to miss - you can't miss a block. (If that's what you are referring to here).

include/pybind11/detail/internals.h
henryiii: C99 requires the inline on the definition; are you sure C++ cares? Looking at the spec, it seems clearly allowed unless it's nested inside another function https://en.cppreference.com/w/cpp/language/inline & https://en.cppreference.com/w/cpp/language/declarations#Specifiers clearly seem to indicate it's allowed, and I didn't have any problems on any complier I quickly checked on godbolt.

include/pybind11/detail/internals.h
rwgk: Oh, sorry, you're right. I was going too fast (basically I was overlooking the `if`).
I'd write this as
```
auto *nep = dynamic_cast<const std::nested_exception *>(std::addressof(exc));
if (nep != nullptr) {
```
because I believe that's less likely to get misread, but I'm totally fine with this code as is.

include/pybind11/detail/internals.h
rwgk: > C99 requires the inline on the definition;

Ouch.

My previous comment was based on what I remembered from working on
https://github.com/pybind/pybind11/pull/3179.

From memory, in a Google-internal chat I was advised to not use `inline` in forward declarations.

Am I sure? No. I was asking for advice and then going with that. The work on PR #3179 left me believing that even compilers are unusually confused about what is correct.

To be pragmatic, we can reduce it to: do we want to be consistent within pybind11? PR #3179 removed some `inline` based on the advice I got, and to be consistent.

rwgk:
Over night I ran this through the Google global testing. There were no problems.

For completeness: I tested this PR in combination with the smart_holder git-merge-master helper PR #3613.

include/pybind11/detail/internals.h
rwgk: I intentionally kept #3179 limited to functions involving `PYBIND11_NOINLINE`. I didn't look around too much, as I very often do, to not let a good project die the death of a thousand cuts.

For this PR, idk. If we're inconsistent already and the advice I got in Aug last year isn't universally accepted (as I was thinking until now) ... shrug :-)

include/pybind11/detail/internals.h
```diff
@@ -309,29 +307,27 @@ bool handle_nested_exception(const T &exc, const std::exception_ptr &p) {
...
-bool handle_nested_exception(T, std::exception_ptr) {
+inline bool raise_err(PyObject *exc_type, const char *msg) {
+    PyErr_SetString(exc_type, msg);
     return false;
 }
 
-inline bool raise_err(PyObject *exc_type, const char *msg) {
-    PyErr_SetString(exc_type, msg);
+template <class T>
+bool handle_nested_exception(const T&, std::exception_ptr) {
```
rwgk: `const T &`
?

rwgk (approved):

rwgk:
Yay!

PR #3643, opened 2022-01-25
Function pybind11_select_caster uses ADL to select the caster.

amauryfa:
## Description

Instead of the `type_caster` template specialization, it is possible to register the conversions with a function *declaration*, which will be found with [ADL](https://en.cppreference.com/w/cpp/language/adl).

```c++
namespace std {
string_caster pybind11_select_caster(std::string*);
}
```

The parameter is a pointer to the C++ (intrinsic) type. The returned type is the caster class. This function should have no implementation!

- There is no ODR violation when different translation units define different conversions: distinct modules might choose different behavior. This is the main motivation of this change.
- The declarations need to be defined in the original namespace of the converted C++ type.
- The default behavior did not change: `type_caster<T>` will be used as a fallback.
- Because `type_caster<T>` not always used, I had to update some places with `make_caster<T>`.

So far I do not propose to convert the existing casters. I updated the one for `std::string` just to prove the concept, and check that it works on all compilers.

### DIFF
--- a/docs/advanced/cast/custom.rst
+++ b/docs/advanced/cast/custom.rst
@@ -101,12 +101,12 @@ There are two ways to do this:
       in every compilation unit of the Python extension module. Otherwise,
       undefined behavior can ensue.
 
-* The preferred method is to *declare* a function named `pybind11_declare_caster`,
+* The preferred method is to *declare* a function named `pybind11_select_caster`,
   its only purpose is to associate the C++ type with its caster class:
 
   .. code-block:: cpp
 
-      inty_type_caster pybind11_declare_caster(inty*);
+      inty_type_caster pybind11_select_caster(inty*);
 
   The argument is a *pointer* to the C++ type, the return type is the
   `caster_type`. This function has no implementation!
--- a/docs/advanced/cast/custom.rst
+++ b/docs/advanced/cast/custom.rst
@@ -31,63 +31,82 @@ The following Python snippet demonstrates the intended usage from the Python sid
 
     print(A())
 
-To register the necessary conversion routines, it is necessary to add an
-instantiation of the ``pybind11::detail::type_caster<T>`` template.
-Although this is an implementation detail, adding an instantiation of this
-type is explicitly allowed.
+To register the necessary conversion routines, it is necessary to define a
+`type_caster` class, and register it with the other pybind11 casters:
 
 .. code-block:: cpp
 
-    namespace pybind11 { namespace detail {
-        template <> struct type_caster<inty> {
-        public:
-            /**
-             * This macro establishes the name 'inty' in
-             * function signatures and declares a local variable
-             * 'value' of type inty
-             */
-            PYBIND11_TYPE_CASTER(inty, const_name("inty"));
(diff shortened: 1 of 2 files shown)


Skylion007:
@henryiii Do we finally need to update MSYS2?

rwgk:
The 2 mingw failures is some infrastructure issue that we started seeing yesterday, unrelated to this PR.

laramiel:
There's an ICE when building with MSVC 2015. Now that it's 2022, I suspect that it's time to retire that compiler.

 ```
 d:\a\pybind11\pybind11\include\pybind11\cast.h(35): fatal error C1001: An internal error has occurred in the compiler. [D:\a\pybind11\pybind11\build\tests\pybind11_tests.vcxproj]
  (compiler file 'msc1.cpp', line 1468)
   To work around this problem, try simplifying or changing the program near the locations listed above.
```

This ICE was reported to MS--it appears to be a MSVC 2015 update 3 specific bug--but it was long enough ago that the link has become dead:
https://connect.microsoft.com/VisualStudio/feedback/details/2869042

It appears similar to a decltype() adjacent bug reported to boost in 2018:
https://svn.boost.org/trac10/ticket/13393#ticket

henryiii:
We are going to drop Python 2.7-3.5 soon, that was when I was planning to retire the compiler. With `windows-2016` being removed in March, March is also a good time to retire it. But I'm honestly okay to retire it sooner.

What about the failures on 2017?

rwgk:
> But I'm honestly okay to retire it sooner.

Fantastic!

* That gives me a strong argument to prioritize the Python 2.7 & 3.5 removal over other things.

* It will also get this oddity off my back: https://github.com/pybind/pybind11/blob/aa4fe8e4ffb8ef772cdd952d111480f721b629df/include/pybind11/detail/smart_holder_type_casters.h#L634

* And I can be more relaxed when our toddler decides to bang on the keyboard of the only laptop I have for MSVC 2015 testing.

(It's late here, sorry for rambling.)

laramiel:
@henryiii I haven't yet looked too closely at those, so hard to say yet. I may look tomorrow, but the environment setup is somewhat challenging.

henryiii:
MSVC 2015 only really supports C++14. They basically implemented 11 and 14 in one shot, because they were behind.

henryiii:
To actually merge this, we probably should either revert the `std::string` change, or add similar changes for the other casters, they should be consistent. If there's a chance it will work on MSVC 2015, maybe we could see if reverting the `std::string` change (hopefully dropping something similar in as a test) causes this to pass? If the only problem is related to using this, I'd not be that worried about 2015. (I think this is not why it's breaking, though).

rwgk:
> If there's a chance it will work on MSVC 2015, maybe we could see if reverting the std::string

Yeah, I was thinking the same, and maybe @amauryfa, too?
IIUC those changes were mainly to have a quick way to exercise the new feature.
I'm thinking undoing those changes and adding dedicated new tests is best.

rwgk:
Work continued under #3862.

PR #3648, opened 2022-01-26
Fix a typo in class.rst

cauliyang:
Fix a typo in class.rst

## Description

A Typo in class.rst.

`we’d like to to`

## Suggested changelog entry:

`we’d like to`


rwgk:
Thanks for the fix!

PR #3649, opened 2022-01-26
Add tests demonstrating smart_holder issues with multiple inheritance

virtuald:
As requested by @rwgk these tests show that everything works fine on master.

From #3635


auscompgeek:
Looks like this got merged in with #3650.

PR #3650, opened 2022-01-26
Propagate py::multiple_inheritance to all children

virtuald:
## Description

This builds on the tests split from #3665 into #3649, which should probably be merged first.

This started with a comment that @Skylion007 made on my tests for #3635, and I realized the fix was straightforward. Previously, you had to mark any child of a class that participated in multiple inheritance with `py::multiple_inheritance`, even if the child derived from a single base. Now, pybind11 can autodetect this in most cases.

I'm not confident this is quite the right fix. It's not clear to me what the difference between `simple_type` and `simple_ancestors` are, and I wonder if the purpose of `simple_type` has evolved over time as it's only used in one place.

A different fix might be to change type_caster_base's simple_type check to simple_ancestors, but then simple type wouldn't be used anywhere? 

... actually, I just realized this change makes simple_type and simple_ancestors identical? Simple ancestors is only used for registering/deregistering instances.

 It's late here, I'll sleep on it. There must be a subtle side effect of simple_type that I'm just missing and that the tests don't currently cover.

## Suggested changelog entry:

```rst
* ``py::multiple_inheritance`` is now only needed when C++ bases are hidden from pybind11
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1206,10 +1206,13 @@ protected:
         if (rec.bases.size() > 1 || rec.multiple_inheritance) {
             mark_parents_nonsimple(tinfo->type);
             tinfo->simple_ancestors = false;
+            tinfo->simple_type = false;
         }
         else if (rec.bases.size() == 1) {
             auto parent_tinfo = get_type_info((PyTypeObject *) rec.bases[0].ptr());
             tinfo->simple_ancestors = parent_tinfo->simple_ancestors;
+            // a child of a non-simple type can never be a simple type
+            tinfo->simple_type = parent_tinfo->simple_type;
         }
 
         if (rec.module_local) {
--- a/tests/test_multiple_inheritance.cpp
+++ b/tests/test_multiple_inheritance.cpp
@@ -304,9 +304,7 @@ TEST_SUBMODULE(multiple_inheritance, m) {
         .def("get_e_d1", &MVE::get_e_d1)
         .def("get_e_e", &MVE::get_e_e)
         .def_readwrite("e", &MVE::e);
-    // TODO: py::multiple_inheritance is required here, but pybind11 should
-    // be able to detect this by looking at MVE which is already bound...
-    py::class_<MVF, MVE>(m, "MVF", py::multiple_inheritance())
+    py::class_<MVF, MVE>(m, "MVF")
         .def(py::init<>())
         .def("get_f_b", &MVF::get_f_b)
         .def("get_f_c", &MVF::get_f_c)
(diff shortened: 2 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -1206,10 +1206,13 @@ class generic_type : public object {
...
             mark_parents_nonsimple(tinfo->type);
             tinfo->simple_ancestors = false;
+            tinfo->simple_type = false;
         }
         else if (rec.bases.size() == 1) {
             auto parent_tinfo = get_type_info((PyTypeObject *) rec.bases[0].ptr());
             tinfo->simple_ancestors = parent_tinfo->simple_ancestors;
+            // a child of a non-simple type can never be a simple type
+            tinfo->simple_type = parent_tinfo->simple_type;
         }
```
Skylion007, line 1217: According to the documentation for the TypeInfo type, a type with non simple_ancestors can never be a simple type. We should add an assert somewhere that verifies this:
```suggestion
        // a child of a non-simple type can never be a simple type
        assert (!tinfo->simple_type || tinfo->simple_ancestors));
```

include/pybind11/pybind11.h
```diff
@@ -1206,10 +1206,13 @@ class generic_type : public object {
         if (rec.bases.size() > 1 || rec.multiple_inheritance) {
             mark_parents_nonsimple(tinfo->type);
             tinfo->simple_ancestors = false;
+            tinfo->simple_type = false;
         }
         else if (rec.bases.size() == 1) {
             auto parent_tinfo = get_type_info((PyTypeObject *) rec.bases[0].ptr());
             tinfo->simple_ancestors = parent_tinfo->simple_ancestors;
+            // a child of a non-simple type can never be a simple type
+            tinfo->simple_type = parent_tinfo->simple_type;
```
Skylion007, line 1215: Nit: Technically should be "&&" because non-simple parent implies non-simple type, but simple parent may not imply simple type by it's current definition.

Skylion007 (approved):

henryiii (approved):

rwgk:
@jagerman for comment, who I believe was the main author of the multiple inheritance code.

rwgk:
I'll try to get this through the global testing asap (hopefully the 12:00 batch, in ~2 hours).

rwgk:
This PR is now in the global testing pipeline (12:00 batch), after first testing manually with ASAN.

rwgk (approved):
The global testing went fine. Results are a bit noisy again, but after looking through I'm convinced we're good to go ahead.

rwgk:
This is also on the smart_holder branch now. Thanks again!

PR #3659, opened 2022-01-28
Fix caster optimization regression introduced in #3650

Skylion007:
## Description
So I thought about this some more, and I realized the difference between simple type and simple ancestors in the caster. A non-simple type is any ancestor who has any descendants who have multi-inheritance from C++ classes. The descendant itself is a simple type though, unless it also has descendants. This PR fixes a mistake made in #3650 where the descendants themselves can remain simple types until they are extended. Marking a type as non-simple adds some additional pointer checks in the casters and while this is an edge case, we should be careful not to undo the optimization here.

Ping @virtuald 
TLDR: Figured out what the difference is between simple_type and simple_ancestors. 
1. A type can be simple if it has MI, but has no descendants. 
2. As soon as it gets PyBind11 descendants, it ceases to be simple. 
In otherwise, `!simple_ancestors => !simple_type` only if it has PyBind11 descendants. 
I figured this out by thinking about why the rec.multiple_inheritance pathway didn't cause issues before as long multiple_inheritance was specified. The difference being that the single base pathway failed to update the type of the parent to non-simple.

## Suggested changelog entry:

```rst
* Fix previous PR.
```
(just a reminder to include both PR numbers)

### DIFF
--- a/tests/test_multiple_inheritance.py
+++ b/tests/test_multiple_inheritance.py
@@ -472,3 +472,23 @@ def test_pr3635_diamond_f():
     assert o.get_f_e() == 5
 
     assert o.get_f_f() == 6
+
+
+def test_python_inherit_from_mi():
+    class PYMVF(m.MVF):
+        g = 7
+
+        def get_g_g(self):
+            return self.g
+
+    o = PYMVF()
+
+    assert o.b == 1
+    assert o.c == 2
+    assert o.d0 == 3
+    assert o.d1 == 4
+    assert o.e == 5
+    assert o.f == 6
+    assert o.g == 7
+
+    assert o.get_g_g() == 7
--- a/tests/test_multiple_inheritance.py
+++ b/tests/test_multiple_inheritance.py
@@ -475,13 +475,15 @@ def test_pr3635_diamond_f():
 
 
 def test_python_inherit_from_mi():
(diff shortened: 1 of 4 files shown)


virtuald (changes requested):
Your explanation sounds like it captures the subtlety that I wasn't sure I had completely understood.

Can you update the documentation for the `simple_type` and `simple_ancestors`? Having it in this PR is ok, but they're really sparse on detail so making them a bit more clear would be good IMO.

tests/test_multiple_inheritance.py
```diff
@@ -472,3 +472,23 @@ def test_pr3635_diamond_f():
     assert o.get_f_e() == 5
 
     assert o.get_f_f() == 6
+
+
+def test_python_inherit_from_mi():
+    class PYMVF(m.MVF):
```
rwgk, line 478: On chat Aaron told me this is redundant, pulled out for focused debugging.
Maybe add a comment to that effect? E.g.

`# This test is redundant but was useful for focused debugging while working on PR #3659`

rwgk (approved):
Makes sense to me, although I cannot claim that I have a full grasp of all subtleties.

Tagging @jagerman in hopes that he can confirm or correct.

henryiii (approved):
LGTM

PR #3663, opened 2022-01-29
Fix module extension detection for python 3.10

mbasaglia:
## Description
I was having issues with python 3.10, looks like `distutils.sysconfig` has been merged into `sysconfig` (https://docs.python.org/3/distutils/apiref.html#module-distutils.sysconfig) so this adds a condition to import sysconfig on python >= 3.10.

Probably not the most "pythonic" code but I kept it as a 1 liner.

## Suggested changelog entry:

```rst
* Fix CMake extension suffix computation on Python 3.10+
```

### DIFF
--- a/tools/pybind11NewTools.cmake
+++ b/tools/pybind11NewTools.cmake
@@ -110,7 +110,7 @@ if(NOT DEFINED PYTHON_MODULE_EXTENSION)
   execute_process(
     COMMAND
       "${${_Python}_EXECUTABLE}" "-c"
-      "from distutils import sysconfig as s;print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
+      "import sys; s = __import__('distutils.sysconfig' if sys.version_info < (3, 10) else 'sysconfig'); print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
     OUTPUT_VARIABLE _PYTHON_MODULE_EXTENSION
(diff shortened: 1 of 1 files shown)


tools/pybind11NewTools.cmake
```diff
@@ -110,7 +110,7 @@ if(NOT DEFINED PYTHON_MODULE_EXTENSION)
   execute_process(
     COMMAND
       "${${_Python}_EXECUTABLE}" "-c"
-      "from distutils import sysconfig as s;print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
+      "import sys; s = __import__('distutils.sysconfig').sysconfig if sys.version_info < (3, 10) else __import__('sysconfig'); print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
```
henryiii, line 113:
```suggestion
      "import sys, importlib; s = importlib.import_module('distutils.sysconfig' if sys.version_info < (3, 10) else 'sysconfig'); print(s.get_config_var('EXT_SUFFIX') or s.get_config_var('SO'))"
```

https://docs.python.org/3/library/importlib.html#importlib.import_module

henryiii (approved):
Please don't delete the changelog entry from the PR template, we build the changelog from that.

Looks great to me! Nice to see distutils disappearing. ;)

rwgk:
LGTM (but it's out of my wheelhouse).

Skylion007 (approved):

PR #3665, opened 2022-01-29
Document how to bind templates

virtuald:
I was very surprised that the documentation doesn't explicitly call out how to bind templates. While it's fairly obvious, it may not be to those minimally familiar with C++.

### DIFF
--- a/docs/advanced/classes.rst
+++ b/docs/advanced/classes.rst
@@ -1161,6 +1161,58 @@ error:
 
 .. versionadded:: 2.6
 
+Binding classes with template parameters
+========================================
+
+pybind11 can also wrap classes that have template parameters. Consider these classes:
+
+.. code-block:: cpp
+
+    struct Cat {};
+    struct Dog {};
+
+    template <typename PetType>
+    struct Cage {
+        Cage(PetType& pet);
+        PetType& get();
+    };
+
+C++ templates may only be instantiated at compile time, so pybind11 can only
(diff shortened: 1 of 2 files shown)


docs/advanced/classes.rst
```diff
@@ -1161,6 +1161,58 @@ error:
...
+        Cage(PetType& pet);
+        PetType& get();
+    };
+
+C++ templates may only be instantiated at compile time, so pybind11 can only
+wrap instantiated templated classes. You cannot wrap a non-instantiated template:
+
+.. code-block:: cpp
+
+    // this will not compile
```
rwgk, line 1185: // BROKEN (this will not compile)
?
To make oversights & resulting misunderstandings less likely.

(Similarly in the other file, for functions.)

docs/advanced/functions.rst
```diff
@@ -578,3 +578,37 @@ prefers earlier-defined overloads to later-defined ones.
...
+You must bind each instantiated function template separately. You may bind
+each instantiation with the same name, which will be treated the same as
+an overloaded function:
+
+.. code-block:: cpp
+
+    m.def("set", &set<int>);
+    m.def("set", &set<std::string>);
+
+Sometimes it's more clear to bind them with separate names, which is also ok:
```
rwgk, line 609: which is also an option
?

"ok" could be understood as a judgement (ok but not as good).

rwgk (approved):
Nice!

rwgk:
Merging, thanks Dustin!

PR #3666, opened 2022-01-29
Remove extra semicolon

mbasaglia:
Just a stray semicolon triggering a warning, it was causing issues with `-Werror` for me.


rwgk:
Oh, wow, amazing that it made it through the many checks we have. Thanks!
The one CI failure is a well-known flake, safe to ignore.
Merging.

PR #3671, opened 2022-01-31
Ensure python errors are propogated for Python->C++

Skylion007:
## Description
* Ensures that any other exceptions that occur during function overload resolution propogates to the user if possible

## Suggested changelog entry:

```rst
* Propagate Python exception traceback using ``raise_from`` if a pybind11 function runs out of overloads.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -988,6 +988,13 @@ protected:
             }
 
             append_note_if_missing_header_is_suspected(msg);
+#if PY_VERSION_HEX >= 0x03030000
+            // Attach additional error info to the exception if supported
+            if (PyErr_Occurred()) {
+                raise_from(PyExc_TypeError, msg.c_str());
+                return nullptr;
+            }
+#endif
             PyErr_SetString(PyExc_TypeError, msg.c_str());
             return nullptr;
         }


henryiii (approved):

rwgk:
Looks great, thanks Aaron!
How much trouble is it to update a test to exercise this change? (The dispatcher code is very complex. Would be great to have an assurance that refactoring doesn't break this feature.)

Skylion007:
@rwgk We would need an example to trigger it easily. (Maybe a dummy caster just sets an error and then fails to load)?  One of the tests could already be triggering it, but I would need to check the `__cause__` value on all the exceptions to verify.

rwgk:
> @rwgk We would need an example to trigger it easily. (Maybe a dummy caster just sets an error and then fails to load)? One of the tests could already be triggering it, but I would need to check the `__cause__` value on all the exceptions to verify.

I played around a few minutes, adding `assert()`s. It turns out no existing test reaches the new branch. (To be sure I'm not fooling myself, I also put the `assert()` in the older similar code a few lines down, and that fired.)

Based on @Skylion007's comment, it seems to be quite a bit of trouble adding a test. I think it's fine to not invest the time, but it would be best to leave some low-key warning for people working on the code in the future. Here is a super simple idea:

https://github.com/pybind/pybind11/pull/3673

What do you think?

PR #3675, opened 2022-02-02
ci: move centos 8 to stream

henryiii:
## Description

CentOS 8 is at EOL.

https://forums.centos.org/viewtopic.php?f=54&t=78708


rwgk (approved):
Nice to see my `VAR_BUILD_TYPE` workaround purged!
Could we remove the `-DCMAKE_BUILD_TYPE=MinSizeRel` line entirely?

Skylion007 (approved):

PR #3678, opened 2022-02-02
fix: support sysconfig direclty in 3.10+

### LINKED ISSUE #3677: [BUG]: Distutils is going to be deprecated with Python3.12
### Required prerequisites

### Problem description

In Python 3.12 distutils is going to be removed and with 3.10 it's marked as deprecated. Currently `tools/FindPythonLibsNew.cmake` is still using

["${PYTHON_EXECUTABLE}" "-c" "from distutils import sysconfig as s;import sys;import struct;](https://github.com/pybind/pybind11/blob/ce18721d830a9deb8cf1883b513bcb8f98090c35/tools/FindPythonLibsNew.cmake#L115)

It might be replaceable by a simple `import sysconfig as s;`

See [[0]](https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated) and [[1]](https://docs.python.org/3/library/distutils.html)

Kind regards,
Arno

### Reproducible example code

_No response_

henryiii:
## Description

Closes #3677.

## Suggested changelog entry:

```rst
* Use sysconfig over distutils for Python 3.10 using classic CMake discovery.
```

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -112,7 +112,8 @@ endif()
 # VERSION. VERSION will typically be like "2.7" on unix, and "27" on windows.
 execute_process(
   COMMAND
-    "${PYTHON_EXECUTABLE}" "-c" "from distutils import sysconfig as s;import sys;import struct;
+    "${PYTHON_EXECUTABLE}" "-c" "import sys, importlib, struct;
+s = importlib.import_module('distutils.sysconfig' if sys.version_info < (3, 10) else 'sysconfig');
 print('.'.join(str(v) for v in sys.version_info));
 print(sys.prefix);
 print(s.get_python_inc(plat_specific=True));


rwgk (approved):
This seems to be very similar to a recent change in tools/pybind11NewTools.cmake. Evidently, someone works on one copy, unaware of the other, and it takes two iterations. I usually flag such situations with a comment, e.g.
```
# Note that there is similar code in X.
```

henryiii:
This is a lot more complicated because we'll have to work out how to get each value, and have some sort of nicer toggle to keep the code from exploding. (notice it doesn't work on Python 3.10 with the current change) The pybind11NewTools.cmake change is only partially related.

henryiii:
(pytorch/pytorch@4b96fc0 will likely be very helpful)

Given that nothing is broken, I think we are fine to continue with a release without doing this.

stefanor:
Filed PR #3764, which adds a sysconfig path for >= 3.10, and uses the distutils mechanism for older pythons.

PR #3679, opened 2022-02-03
fix: update catch to 2.13.5 to fix glibc 2.34 failures

mgorny:
## Description

Update the downloaded Catch version to 2.13.8, in order to fix build
failure on glibc 2.34:

```
In file included from /usr/include/signal.h:328,
                 from /tmp/pybind11/.nox/tests-3-9/tmp/tests/catch/catch.hpp:8030,
                 from /tmp/pybind11/tests/test_embed/catch.cpp:13:
/tmp/pybind11/.nox/tests-3-9/tmp/tests/catch/catch.hpp:10818:58: error: call to non-‘constexpr’ function ‘long int sysconf(int)’
10818 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
      |                                                          ^~~~~~~~~~~
In file included from /usr/include/python3.9/Python.h:36,
                 from /tmp/pybind11/include/pybind11/detail/common.h:215,
                 from /tmp/pybind11/include/pybind11/pytypes.h:12,
                 from /tmp/pybind11/include/pybind11/cast.h:13,
                 from /tmp/pybind11/include/pybind11/attr.h:13,
                 from /tmp/pybind11/include/pybind11/pybind11.h:13,
                 from /tmp/pybind11/include/pybind11/embed.h:12,
                 from /tmp/pybind11/tests/test_embed/catch.cpp:4:
/usr/include/unistd.h:640:17: note: ‘long int sysconf(int)’ declared here
  640 | extern long int sysconf (int __name) __THROW;
      |                 ^~~~~~~
In file included from /tmp/pybind11/tests/test_embed/catch.cpp:13:
/tmp/pybind11/.nox/tests-3-9/tmp/tests/catch/catch.hpp:10877:45: error: size of array ‘altStackMem’ is not an integral constant-expression
10877 |     char FatalConditionHandler::altStackMem[sigStackSize] = {};
      |                                             ^~~~~~~~~~~~
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -932,7 +932,7 @@ jobs:
       run: git clean -fdx
 
     - name: Configure C++14
-      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=14 -S . -B build2
+      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=14 -DDOWNLOAD_CATCH=ON -S . -B build2
 
     - name: Build C++14
       run: cmake --build build2 -j 2
@@ -950,7 +950,7 @@ jobs:
       run: git clean -fdx
 
     - name: Configure C++17
-      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=17 -S . -B build3
+      run: cmake -G "MinGW Makefiles" -DCMAKE_CXX_STANDARD=17 -DDOWNLOAD_CATCH=ON -S . -B build3
 
     - name: Build C++17
       run: cmake --build build3 -j 2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
             python: '3.6'
             args: >
               -DPYBIND11_FINDPYTHON=ON
-              -DPYBIND11_DEFINE_UNDERSCORE=ON
+              -DCMAKE_CXX_FLAGS="-D_=1"
           - runs-on: windows-latest
             python: '3.6'
             args: >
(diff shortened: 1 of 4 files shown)


henryiii (approved):

Skylion007:
Do we not need to update anything with the DOWNLOAD_CATCH cmake code?

henryiii:
Hmmm,

```
/home/runner/work/pybind11/pybind11/tests/catch/catch.hpp: In member function ‘void Catch::RunContext::invokeActiveTestCase()’:
[23](https://github.com/pybind/pybind11/runs/5053666845?check_suite_focus=true#step:11:23)
<command-line>: error: expected unqualified-id before numeric constant
```

Not very descriptive.

henryiii:
https://github.com/catchorg/Catch2/commit/8f277a54c0b9c1d1024dedcb2dec1d206971e745 broke this, I've submitted a fix in https://github.com/catchorg/Catch2/pull/2364.

henryiii:
I think we can go to 2.13.5, around 2.13.6 introduced the bug. What version fixes this? Was it the constexpr change in 2.13.8?

mgorny:
Thank you for your help. I've confirmed that 2.13.5 is good enough (2.13.4 is still broken). Let's hope it doesn't have that bug yet.

henryiii:
It turns out, by using `git grep '\<_\>' src/* include/*`, that 2.13.5 introduced the bug. So 2.13.4 would be valid to update to, but anything past that is broken.

Skylion007:
Hmm, the test that is broken seems to be because our CMake system doesn't force the remaining CI to downgrade catch. Thoughts @henryiii ?

henryiii:
There is no working version. 2.13.4 doesn't support GLIBC 2.34 and 2.13.5 introduced the lock that has a `_` name.

mgorny:
For the record, I'm not in a hurry to have this fixed (as normally I use system catch), so personally I'm fine with waiting for a fixed Catch version.

henryiii:
Okay, I think I have an option that is working. It's not testing `_` for the embed test anymore, which avoids our only use of Catch2 - not ideal to loose some coverage, but maybe necessary. I'd kind of like to intentionally break it to make sure it's actually still working, but I _think_ it is.

PR #3683, opened 2022-02-03
style: clang-format in pre-commit

henryiii:
## Description

Full clang-format, in pre-commit.

## Suggested changelog entry:

```rst
* Format the codebase with clang-format
```


Skylion007:
@henryiii I have a readability check that's been waiting on this clang-format. Let me add in it as well.

include/pybind11/detail/common.h
```diff
@@ -873,39 +987,59 @@ PYBIND11_RUNTIME_EXCEPTION(type_error, PyExc_TypeError)
...
+    static constexpr bool value = false;
+};
+template <typename T>
+struct is_fmt_numeric<T, enable_if_t<std::is_arithmetic<T>::value>> {
     static constexpr bool value = true;
-    static constexpr int index = std::is_same<T, bool>::value ? 0 : 1 + (
-        std::is_integral<T>::value ? detail::log2(sizeof(T))*2 + std::is_unsigned<T>::value : 8 + (
-        std::is_same<T, double>::value ? 1 : std::is_same<T, long double>::value ? 2 : 0));
+    static constexpr int index
+        = std::is_same<T, bool>::value
```
Skylion007, line 1019: yikes to some of this formatting...

include/pybind11/cast.h
```diff
@@ -10,11 +10,11 @@
 
 #pragma once
 
-#include "pytypes.h"
 #include "detail/common.h"
 #include "detail/descr.h"
 #include "detail/type_caster_base.h"
 #include "detail/typeid.h"
+#include "pytypes.h"
```
Skylion007, line 17: Wow, this isn't right. @henryiii

Skylion007:
This also broke a lot of clang-tidy comments.

include/pybind11/detail/common.h
henryiii: I'm fine to change the format rules, especially if they make it more LLVM like. Autoformatting is not perfectly "automatic", a little fiddling can be done to get it to format nicely (like moving comments).

henryiii:
Looks like `decltype(PyThread_create_key())` is an int for Python < 3.7, which is muddling with `readability-qualified-auto`. I'm impressed I have that showing up in my VIM. :)

henryiii:
I've set the include order to be pybind11 first, detail/* second, other local includes third. The order shouldn't matter, we should IWYU, but for now, this hopefully will work.

henryiii:
Ahh, NOLINTNEXTLINE should count logical lines. :'(

henryiii:
```console
$ brew install act
$ act -j clang-tidy
```

Very nice. :)

include/pybind11/detail/common.h
```diff
@@ -635,177 +651,271 @@ using remove_cvref_t = typename remove_cvref<T>::type;
 using std::index_sequence;
 using std::make_index_sequence;
 #else
-template<size_t ...> struct index_sequence  { };
-template<size_t N, size_t ...S> struct make_index_sequence_impl : make_index_sequence_impl <N - 1, N - 1, S...> { };
-template<size_t ...S> struct make_index_sequence_impl <0, S...> { using type = index_sequence<S...>; };
-template<size_t N> using make_index_sequence = typename make_index_sequence_impl<N>::type;
+template <size_t...>
+struct index_sequence {};
```
henryiii, line 655: These sorts of things would be a lot more readable in the new form with an extra blank line in-between.

rwgk:
Quick question: when I tried this previously I had trouble with
```
/** \rst
\endrst */
```
comments getting garbled. I even wrote a [small script](https://github.com/rwgk/pybind11_scons/blob/master/rst_clang_format_off_on.py) to massage those, to not get messed up by clang-format.

But now it seems to be fine without doing anything special. (?)

Mostly curious: Does anyone know what changed?

include/pybind11/pytypes.h
```diff
@@ -1738,35 +1894,36 @@ class memoryview : public object {
 
         See also: Python C API documentation for `PyMemoryView_FromBuffer`_.
 
-        .. _PyMemoryView_FromMemory: https://docs.python.org/c-api/memoryview.html#c.PyMemoryView_FromMemory
-     \endrst */
```
rwgk, line 1742: clang-format still seems to be unaware of reStructuredText.

Previously the doc generation would fail after I ran clang-format (clang 12).

Are we just getting lucky now?

rwgk:
I wrote an even simpler [script to extract the `\rst`...`\endrst` blocks](https://github.com/rwgk/pybind11_scons/blob/master/rst_block_extractor.py) to a file,
* ran it on current master (36813cfa1228f9414aac669d239a0d3255fc25c3),
* then again on this PR (e379d5d2d52f487d57d9e0026ec0cffff932a68c).

Here is the diff: https://github.com/rwgk/pybind11_scons/commit/fa4ffb49c325430d2f36084a2e98f495a3ef7c2c

I think we need to curate the changed blocks, at least some, and protect them from clang-format.

I believe it's best to systematically protect all such `\rst` blocks, which is what the [other script](https://github.com/rwgk/pybind11_scons/blob/master/rst_clang_format_off_on.py) does. It's slightly annoying, but really only a few lines extra and easy to maintain.

rwgk:
An alternative line of thought:
* I believe clang-format getting into reflowing comments will be generally more annoying than helpful.
* All I want it to do is warn about lines being longer than 99 chars, with the usual exceptions for (apparent) URLs and filenames.
* Then leave it to users to fix manually.

henryiii:
I'm quite okay to turn off comment formatting, I have it off for other areas (including CMake comments, maybe even in this repo).

Can we keep the "nice" changes (like reflowing text to the right length) and use your script to recover the rst formatting, then just turn it off?

henryiii:
(I'm also okay for comments to go over length, but I don't know how to turn that off, including in black)

rwgk:
> Can we keep the "nice" changes (like reflowing text to the right length) and use your script to recover the rst formatting, then just turn it off?

Sounds great, I'll try to work on it tomorrow morning.

> (I'm also okay for comments to go over length, but I don't know how to turn that off, including in black)

Yeah, that's a minor loss. After we get this PR merged, we could maybe hook in a small script that flags long comments. I bet a page worth of Python will do most of the job, with some simple approach to allowing exceptions.

henryiii:
I'm hoping to avoid too much customization, we are getting rid of our custom formatting script. :) I think we may still get line length warnings - and I'm really not too worried about comments. The thing I don't like is this:

```cpp
AVeryLongLongLongClass a_very_long_long_long_varaible_or_other_expression(); // NOLINT
```
Turning into this:

```cpp
AVeryLongLongLongClass a_very_long_long_long_varaible_or_other_expression();
// NOLINT
```

or this:

```cpp
AVeryLongLongLongClass
    a_very_long_long_long_varaible_or_other_expression(); // NOLINT
```

But I think we are stuck with that. For new code, you usually can rework it to make it shorter, us a `NOLINTNEXTLINE` or similar construct, etc.

rwgk:
> Can we keep the "nice" changes (like reflowing text to the right length) and use your script to recover the rst formatting, then just turn it off?

Done. My tweaks are in commit 60b7eb410fefe2b23aeb6906f2a9184e91b11a15. Setting `ReflowComments: false` is in the follow-on commit 59572e65598b4b9f2c9ae0b8a0e5fcb6d65c7f92. I was surprised to see that turning off comment reflow changed how pragma comments are handled (clang-format wanted to add backslashes, which looks weird to me). I went through a couple iterations until I settled on what you see.

rwgk:
I spent a few minutes glancing through the net diffs. It sure does odd things with comments:

Screen Shot 2022-02-04 at 17 02 25

I manually changed that block to this:
```
    void *ptr = nullptr; // Pointer to the underlying storage
    ssize_t itemsize = 0; // Size of individual items in bytes
    ssize_t size = 0; // Total number of entries
    std::string format; // For homogeneous buffers, this should be set to format_descriptor<T>::format()
    ssize_t ndim = 0; // Number of dimensions
    std::vector<ssize_t> shape; // Shape of the tensor (1 entry per dimension)
    std::vector<ssize_t> strides; // Number of bytes between adjacent entries (for each per dimension)
    bool readonly = false; // flag to indicate if the underlying storage may be written to
```
But then rerunning the pre-commit changed it straight back.

What I'd really want: enforce one space (or two?) exactly before `//`, but then pretend everything after that space doesn't exist, to not trigger wrapping of the code just because of the trailing comment.

The rest looks acceptable, although I cannot say that I like how it deals with `m.def()` and member `.def()`. But I'm OK sacrificing that for the bigger gain of not having to manually move whitespaces around.

henryiii:
Okay, I'll see if we can handle those things better.

henryiii:
Looks like it’s hard to improve WRT comments. What should be done is long trailing comments need to be placed above the line, not on the same one. We need some new blank lines separating sequences of definitions, too. Or we could turn up the line length a lot.

include/pybind11/chrono.h
```diff
@@ -115,71 +127,80 @@ inline std::tm *localtime_thread_safe(const std::time_t *time, std::tm *buf) {
...
+            cal.tm_hour = PyDateTime_TIME_GET_HOUR(src.ptr());
+            cal.tm_mday = 1;  // This date (day, month, year) = (1, 0, 70)
+            cal.tm_mon = 0;   // represents 1-Jan-1970, which is the first
+            cal.tm_year = 70; // earliest available date for Python's datetime
             cal.tm_isdst = -1;
-            msecs        = microseconds(PyDateTime_TIME_GET_MICROSECOND(src.ptr()));
-        }
-        else return false;
+            msecs = microseconds(PyDateTime_TIME_GET_MICROSECOND(src.ptr()));
+        } else
```
Skylion007, line 175: Stuff like this is why I woould vote always endorcing braces.

rwgk:
> What should be done is long trailing comments need to be placed above the line, not on the same one.

I think so, too.

With this in my mind: when developing pybind11 in the future, clang-format will always do one thing or another that doesn't look great. Then we will look and tweak to make it look ok before sending PRs for review.
Moving these comments around is basically retro-active tweaking. I.e. perfectly fine.

Related but separate note: when I worked on this previously I was organizing the work to have all manual changes *before* running clang-format. It's extra work, but it looks much better in the git history: it will be easy to see what was manually changed. If we notice problems later, we will more easily understand how they came about, and therefore feel more certain how to properly fix. We will also have complete certainty which issues were introduced by clang-format, and again be more certain about how to fix; and we have something to point to for filing bugs.

Once we agree here, I could work on splitting into two PRs (part 1: manual, part 2: exclusively automatic).

include/pybind11/chrono.h
henryiii: If you have code automatically formatted, then whitespace is always in sync with braces, so it doesn't bother me at all. I just completely ignore braces and pretend it's Python. :) Though I'm totally fine with requiring the braces, as I said, I ignore them anyway. ;)

henryiii:
Clang-format can't do this, actually, but `readability-braces-around-statements` for clang-tidy can.

Issue #597, opened 2017-01-06
Broken mro when using multiple inheritance with py::metaclass


pschella:
When a derived class `MroDerived` needs to expose a static attribute (which necessitates using `py::metaclass`), but also inherits from multiple bases Python raises the following error:

`ImportError: MroDerived: PyType_Ready failed (TypeError: mro() returned base with unsuitable layout ('pybind11_tests.MroBaseB'))!`

For a complete example see https://github.com/lsst-dm/pybind11/tree/broken_mro

pschella:
This is on Python 3.5.2.

r-owen:
I will also say that the need to add `py::metaclass` to all our code is a bit of a nuisance. I wish there was some way to offer backwards compatibility while still giving whatever benefit this adds (such as making that the default and having a way to disable it when needed). Why might one omit it?

pschella:
While I agree with it being a nuisance, this was already ruled upon by the lead developer.
However, adding `py::metaclass` to all classes does not resolve this problem.

PR #3694, opened 2022-02-07
Improve Python 3.11 support

### LINKED ISSUE #3530: [BUG]: Remove Python 3.11/PyPy dispatch workaround
### Required prerequisites

### Problem description

When implementing Python 3.11 support, due to API changes on the frame object we added a hack to get it temporarily working by falling back on an on old PyPy API to prevent the infinite recursion dispatch hack workaround as seen in https://github.com/pybind/pybind11/pull/3419. As of now, a python code snippit has to be interpreted and evaluated for each dispatch which is far from ideal. Another solution was brought up in https://github.com/pybind/pybind11/issues/2549, but that would require an API change. 

Either we need to find a way to workaround this once the 3.11 private APIs stabilize or to find a better workaround to prevent this recursion issue. I would like to point out that this is the cause of numerous other weird edge cases in function dispatch so cleaning it up would be preferred.

The workaround can be found here: https://github.com/pybind/pybind11/blob/a224d0cca5f1752acfcdad8e37369e4cda42259e/include/pybind11/pybind11.h#L2368

### Reproducible example code

_No response_

Skylion007:
## Description
* Unfortunately, 3.11a uncovered a bug  in our multi-inheritance code that may cause memory leaks with dynamic_attrs, will deal with a follow up PR.
* Fixes a compilation error caused by 3.11 requiring the use of the helper to get the f_back. This PR also backports the reference counting behavior of newer pythons (f_back being a strong reference that needs to be decremented). 
* Closes #3530 by removing the workaround

## Suggested changelog entry:

```rst

```


Skylion007:
Okay, I have narrowed down the problem. It causes an MRO issue under these conditions:
1. A type has multiple inheritance
2. One of the bases has dynamic_attrs enabled
3. The first base does NOT have dynamic_attrs enabled.

This causes the multiple inheritance code complain similar to issue https://github.com/pybind/pybind11/issues/597 . We fixed that one by ensuring all classes have the same metaclass. Something breaks this fix in 3.11.

include/pybind11/detail/type_caster_base.h
```diff
@@ -514,7 +515,14 @@ PYBIND11_NOINLINE std::string error_string() {
             errorString += "  " + handle(f_code->co_filename).cast<std::string>() + "("
                            + std::to_string(lineno)
                            + "): " + handle(f_code->co_name).cast<std::string>() + "\n";
-            frame = frame->f_back;
+#    if PY_VERSION_HEX >= 0x030900B1
```
Skylion007, line 518: Nit: Maybe we should add a backport section details/common.h . I just wasn't sure if the recommend static inline stuff would cause any nasty ODR / LTO issues so I left it as is.

tacaswell:
This fixes compiling scipy with py311.

rwgk:
I looked a bit but have questions for my understanding:

Is it true that

* the change in pybind11.h is purely a 3.11 compatibility fix?

* the change in type_caster_base.h is for 3.11 compatibility, but also an actual fix for a bug that exists already pre 3.11? (Does the fix involve the `Py_XINCREF(frame);` and matching `Py_DECREF(frame);`?)

It would be useful to explain in a couple sentences in the PR description.

Skylion007:
@rwgk This fixes 3.11 compatability. The issue is the new API returns a strong reference (instead of a weak reference) which means we need to decref it. The easiest way to handle this was to have the old code path make a strong reference too. I'll edit the PR description.

henryiii (approved):

rwgk (approved):
The changes LGTM, too, but I'm wondering about
* @Skylion007 I didn't see edits to the PR description, would be nice to have them.
* the valgrind error looks like it needs attention
* is it OK to ignore the upstream failure?

Skylion007:
@rwgk Good catch on the Valgrind error, seems I am not doing the incref / decrefs properly. Thoughts @henryiii ?

include/pybind11/detail/type_caster_base.h
```diff
@@ -514,7 +515,14 @@ PYBIND11_NOINLINE std::string error_string() {
...
+            auto *b_frame = PyFrame_GetBack(frame);
+#    else
+            auto *b_frame = frame->f_back;
+            Py_XINCREF(b_frame);
```
henryiii, line 522: This the leak?

Skylion007: It's on 3.9 so it seems the issue is `PyFrame_GetBack(frame)`? It will be decremented on the next iteration. If it's null, it doesn't need to be decremented. I am not seeing the issue with the control flow.

henryiii: I think it's the fact that the b_frame doesn't get DECREF'd at the end of the loop. The frame does, but not the b_frame.

include/pybind11/pybind11.h
```diff
@@ -2636,19 +2636,18 @@ get_type_override(const void *this_ptr, const type_info *this_type, const char *
...
         // f_code is guaranteed to not be NULL
         if ((std::string) str(f_code->co_name) == name && f_code->co_argcount > 0) {
             PyObject *locals = PyEval_GetLocals();
-            if (locals != nullptr && f_code->co_varnames != nullptr) {
-                PyObject *self_caller
-                    = dict_getitem(locals, PyTuple_GET_ITEM(f_code->co_varnames, 0));
+            if (locals != nullptr) {
+                PyObject *self_arg = PyTuple_GET_ITEM(
+                    PyObject_GetAttrString((PyObject *) f_code, "co_varnames"), 0);
+                PyObject *self_caller = dict_getitem(locals, self_arg);
```
henryiii, line 2650:
```suggestion
                PyObject *co_varnames = 
                    PyObject_GetAttrString((PyObject *) f_code, "co_varnames");
                PyObject *self_arg = PyTuple_GET_ITEM(co_varnames, 0);
                PyObject *self_caller = dict_getitem(locals, self_arg);
                Py_DECREF(co_varnames);
```

henryiii: I'll just try this to see if it changes anything, easy to revert.

henryiii: This made valgrind happy.

Skylion007:
Thanks for pointing out that issue @henryiii. Looks good to me now.

rwgk:
I'll run some quick interactive tests in with the Google toolchain.

include/pybind11/pybind11.h
```diff
@@ -2636,19 +2636,18 @@ get_type_override(const void *this_ptr, const type_info *this_type, const char *
...
         PyCodeObject *f_code = PyFrame_GetCode(frame);
         // f_code is guaranteed to not be NULL
         if ((std::string) str(f_code->co_name) == name && f_code->co_argcount > 0) {
             PyObject *locals = PyEval_GetLocals();
-            if (locals != nullptr && f_code->co_varnames != nullptr) {
-                PyObject *self_caller
-                    = dict_getitem(locals, PyTuple_GET_ITEM(f_code->co_varnames, 0));
+            if (locals != nullptr) {
+                PyObject *co_varnames = PyObject_GetAttrString((PyObject *) f_code, "co_varnames");
+                PyObject *self_caller = dict_getitem(locals, PyTuple_GET_ITEM(co_varnames, 0));
```
rwgk, line 2649: I just looked, `dict_getitem` can throw, and then `co_varnames` will leak.
There may be more issues like this here, but to at least not make it worse, I'd something like this:
```
PyObject *co_varnames = PyObject_GetAttrString((PyObject *) f_code, "co_varnames");
PyObject *vn0 = PyTuple_GET_ITEM(co_varnames, 0);
Py_DECREF(co_varnames);
PyObject *self_caller = dict_getitem(locals, vn0);
```

rwgk (approved):

rwgk:
I did a manual & embarrassingly simple leak check (diff below + `top` command), with Python 3.7 and 3.9. I'm happy to report I didn't see a leak. (I identified that test by inserting an assert in the middle of the changed code in type_caster_base.h)

ASAN testing was happy, too.

```diff
+++ b/tests/test_exceptions.py
@@ -246,6 +246,7 @@ def test_throw_nested_exception():

 # This can often happen if you wrap a pybind11 class in a Python wrapper
 def test_invalid_repr():
+  while True:
     class MyRepr:
         def __repr__(self):
             raise AttributeError("Example error")
```

PR #3700, opened 2022-02-08
fix: __index__ on Enum should always be present.

henryiii:
## Description

Tiny bit of minor cleanup.

## Suggested changelog entry:

```rst
* Enum now has an ``__index__`` method on Python <3.8 too
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1913,8 +1913,7 @@ public:
         def("__int__", [](Type value) { return (Scalar) value; });
         #if PY_MAJOR_VERSION < 3
             def("__long__", [](Type value) { return (Scalar) value; });
-        #endif
-        #if PY_MAJOR_VERSION > 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 8)
+        #else
             def("__index__", [](Type value) { return (Scalar) value; });
         #endif
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1913,7 +1913,7 @@ public:
         def("__int__", [](Type value) { return (Scalar) value; });
         #if PY_MAJOR_VERSION < 3
             def("__long__", [](Type value) { return (Scalar) value; });
-        #else
+        #elif PY_VERSION_HEX >= 0x03080000
             def("__index__", [](Type value) { return (Scalar) value; });
         #endif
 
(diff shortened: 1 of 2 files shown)


include/pybind11/pybind11.h
```diff
@@ -1913,8 +1913,7 @@ template <typename Type> class enum_ : public class_<Type> {
         def("__int__", [](Type value) { return (Scalar) value; });
         #if PY_MAJOR_VERSION < 3
             def("__long__", [](Type value) { return (Scalar) value; });
-        #endif
-        #if PY_MAJOR_VERSION > 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 8)
+        #else
             def("__index__", [](Type value) { return (Scalar) value; });
```
Skylion007, line 1917: Isn't this function valid in >= 3.8? Or are we purposefully backporting index to early to python versions. We still will officially support 3.7

henryiii: Ahh, I misread the statement, this is a very confusing way to write it. Though I don't think this actually hurts older Pythons, it was just required for conversion in 3.8+.

henryiii: Actually, I prefer making this always present. `__index__` can be used other places, like in an index `x[y]` will call `y.__index__()`, even if it took till Python 3.8 to standardize that `__int__` meant lossy conversions, and `__index__` meant lossless conversions.

Skylion007 (approved):

rwgk (approved):
I had the same question last night (but no chance to ask here). I agree with the conclusion!

If you think it's useful to mention in the PR description or changelog: https://github.com/pybind/pybind11/pull/2801 also "backported" 3.8 behavior. The PR description there mentions another situation where that was the case.

rwgk:
I believe (!) this PR broke a couple Google-internal tests. I don't have time to fully dig into the details, but what I see is that generated .pyi files now have extra `def __index__(self) -> int: ...` lines.

@henryiii it might be good to add a warning to the upgrade guide.

henryiii:
Let me know if Python behavior actually changed, I still think that's really unlikely.

rwgk:
> Let me know if Python behavior actually changed, I still think that's really unlikely.

In the meantime the affected team accepted the pybind11 update (and the side-effect changes in their .pyi files) without hesitation. I'll move on with this belief for now: they are harvesting pybind11 docstrings to build their .pyi files. It's actually the same set of .pyi files that was affected when I tested PR #2621 a couple weeks ago, which adds to my belief.

Yes, I agree there will likely be issues with the .pyi files again when we upgrade our Python version.

PR #3707, opened 2022-02-09
Cast bytearray to string 

kururu002:
## Description

- Support bytearray to string casting
- Add test case 
- Rename load_bytes to load_raw

A follow-up change for #2799 to complete bytearray functionalities.

## Suggested changelog entry:

```rst
Support bytearray casting to string
```


tests/test_builtin_casters.py
```diff
@@ -137,6 +137,14 @@ def to_bytes(s):
     assert m.string_length(u"💩".encode("utf8")) == 4
 
 
+def test_bytearray_to_string():
+    """Tests the ability to pass bytearray to C++ string-accepting functions.  Note that this is
+    one-way: the only way to return bytes to Python is via the pybind11::bytearray class."""
+    # Issue #2799
+    assert m.string_length(bytearray(b"Hi")) == 2
+    assert m.strlen(bytearray(b"bytearray")) == 9
+
```
rwgk, line 146: Could you please add tests with empty `bytearray`s?

And with malformed utf-8, to prove that there really isn't any encoding (see e.g. `malformed_utf8` in test_pytypes.py).

I looked at the Python C code underneath `PyByteArray_AsString`, it special-cases empty arrays. But in any case, tests for corner cases are best practice.

rwgk (commented):
I think this is great, with a couple extra tests. I'm not too sure about my `if (bytes)` worry. Let's see what other maintainers think.

tests/test_builtin_casters.py
```diff
@@ -137,6 +137,14 @@ def to_bytes(s):
     assert m.string_length(u"💩".encode("utf8")) == 4
 
 
+def test_bytearray_to_string():
+    """Tests the ability to pass bytearray to C++ string-accepting functions.  Note that this is
+    one-way: the only way to return bytes to Python is via the pybind11::bytearray class."""
```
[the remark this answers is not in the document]
kururu002: The correct term should be `return bytearray to Python` here, but after all, I think one-way is true that bytearray has to form explicitly with py::bytearray similarly as py::bytes. I can correct that or do you think both bytes/bytearray tests are self-explainable?

tests/test_builtin_casters.py
kururu002: Sure, an empty bytearray test case will be added.
For malformed_utf8, I'm not sure it's the same case as casting to python str. In test_pytypes, it's casting a single byte to str, and str is not fulfilled with a utf-8 resulting "b'\\x80'". While passing b"\x80" to C++ string it would be treated as a byte and has size 1. Is this the case you're looking for?

include/pybind11/cast.h
```diff
@@ -435,12 +435,21 @@ template <typename StringType, bool IsView = false> struct string_caster {
                 return true;
             }
         }
+        if (PyByteArray_Check(src.ptr())) {
+            // We were passed a bytearray; accept it into a std::string or char*
+            // without any encoding attempt.
+            const char *bytes = PyByteArray_AsString(src.ptr());
+            if (bytes) {
```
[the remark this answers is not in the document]
kururu002: I'm uncertain if there will be a case bytes is nullptr/NULL. I agree with you we better shout out for such an error. 
I'll wait for more comments for this one.

tests/test_builtin_casters.py
rwgk: The motivation/idea is very simple and high-level:

* We want to be sure this works for malformed utf-8.

We know the current implementation does. We want to be sure that's not accidentally broken somehow (e.g. refactoring).

My suggestion was based on past experience, where that actually happened.

include/pybind11/cast.h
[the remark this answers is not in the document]
rwgk: I see you added the checks, thank you!

This PR looks great to me. The only reason I'm holding off clicking the approve button is that I want to run this through our (Google's) global testing. It will probably take me a couple days to restore my ability to do that; I need to merge master into the smart_holder branch first, which will take some effort after the sweeping clang-format changes.

I'll get back here asap.

tests/test_builtin_casters.py
kururu002: I see your point, the test is added. I'm trying to make sure `assert m.string_length(bytearray(b"\x80")) == 1` is the behavior we're looking for :blush:

rwgk (approved):
This passed global testing!

rwgk:
@kururu002 could you please add a suggestion for the changelog in the PR description, and also give some context for the reference to 2799 (just one terse sentence is fine)?

tests/test_builtin_casters.py
Skylion007: Wouldn't it just be better to parameterize the bytes_to_str tests and have them be rerun with all the bytes args wrapped as bytearrays(via pytest fixture for instance? or by having a function that either returns the bytes or the bytes wrapped in a bytearray depending on test parameters)

kururu002: Hi @Skylion007, I also came up wrapping up the params for better coverage. Considering there're at least 3 dimensions (bytearray vs bytes, string_lenth vs strlen, bytearray with/without params), and a few corner cases(empty bytearray), I would suggest the current intuitive version should give us similar coverage with . What do you think?

kururu002:
> @kururu002 could you please add a suggestion for the changelog in the PR description, and also give some context for the reference to 2799 (just one terse sentence is fine)?

Thank you for testing! I updated the description, not sure what the suggestion for the changelog is for, so I follow characteristics from other PRs.

tests/test_builtin_casters.py
Skylion007: Sounds good.

Skylion007 (approved):

henryiii (approved):

PR #3709, opened 2022-02-09
(clang-tidy): Enable static downcast check

Skylion007:
## Description
* Enable a clang-tidy check to ensure we don't do static_cast for downcasting and always use dynamic_cast. This is safer, more likely to be correct, and most compilers will optimize them to be the same speed anyway if valid. There are no other changes to the codebase currently since we do it properly, but this check ensures the regression doesn't get introduced in a  PR.

## Suggested changelog entry:

```rst
* Enable cppcoreguidelines-pro-type-static-cast-downcast in clang-tidy.
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,6 +3,7 @@ FormatStyle: file
 Checks: '
 *bugprone*,
 cppcoreguidelines-init-variables,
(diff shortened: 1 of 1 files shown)


rwgk (approved):
Nice!

PR #3711, opened 2022-02-09
DO NOT MERGE: trying out setuptools PR

henryiii:
## Description

This is based on #3719.

See https://github.com/henryiii/pybind11/pull/9 too.


henryiii:
So far:

```
E  'pybind11_global.egg-info/not-zip-safe',
```
is missing with the TOML config.

abravalheri:
Thanks @henryiii, I will check!

henryiii:
Okay, that's weird, a single job has this:

```
AttributeError: module 'pybind11' has no attribute 'get_include'
```

I think it's related to the not-zip-safe issue, at least that's my guess. That's from `pybind11-global`, which is the weird legacy version that has "global" data and headers, rather than being inside a Python package.

abravalheri:
Ok, I tried the following:

```bash
% PYBIND11_GLOBAL_SDIST=1 .python -m buildls dist
% ls dist
pybind11-2.10.0.dev1.tar.gz  pybind11_global-2.10.0.dev1-py3-none-any.whl
```

and I noticed the names of the files produced... The `sdist` name does not contain `_global`...

I think this is caused by the fact that PEP 621 requires `name` to be statically defined in the `pyproject.toml` (so in theory trying to redefine it dynamically should not be an option).

The `tar.gz` file however does contain a `not-zip-safe` member:

```bash
% tar tf dist/pybind11-2.10.0.dev1.tar.gz | grep zip
pybind11-2.10.0.dev1/pybind11.egg-info/not-zip-safe
```

abravalheri:
@henryiii, I think that for `pybind11_global` you will have to edit `pyproject.toml` before running the build 😅

abravalheri:
> ```
> E         -  'pybind11_global.egg-info/not-zip-safe',
> ```
> 
> is missing with the TOML config.

I added a test case for that in https://github.com/pypa/setuptools/pull/3068/commits/ed73bc52fde8d3106295f24b9a45dea8292374bd. Let's see what happens with the test suite, but running the tests locally the `not-zip-safe` file seems to be generated.

henryiii:
Ah, yes, I wasn't really expecting this to fully work with PEP 621 metadata, I was originally seeing if it would still work with the changes and the old system. The static name is correct. I'd be perfectly fine adding some sort of pre-processing system to make the global SDist.

abravalheri:
I updated the PR to reflect the changes discussed in https://discuss.python.org/t/help-testing-experimental-features-in-setuptools/13821/59, regarding
the handling of `license` to better complain with PEP 621.

pyproject.toml
```diff
@@ -47,10 +48,6 @@ Chat = "https://gitter.im/pybind/Lobby"
 [tool.setuptools]
 zip-safe = false
 
-[tool.setuptools.dynamic]
-license = "BSD"
-license-files = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]
```
abravalheri: These are the default values already considered by `setuptools` and `wheel`.

Since it was clarified in the Python discourse that specifying `project.license` statically does not prevent the backend to dynamically populate `License-files` in the core metadata, having a dynamic `license` is no longer necessary.

pyproject.toml
```diff
@@ -1,7 +1,53 @@
...
+[project.urls]
+Homepage = "https://github.com/pybind/pybind11"
+Documentation = "https://pybind11.readthedocs.io/"
+"Bug Tracker" = "https://github.com/pybind/pybind11/issues"
+Discussions = "https://github.com/pybind/pybind11/discussions"
+Changelog = "https://pybind11.readthedocs.io/en/latest/changelog.html"
+Chat = "https://gitter.im/pybind/Lobby"
+
+[tool.setuptools]
+zip-safe = false
```
abravalheri, line 49: I wonder if this option is even necessary nowadays.

According to the docs: `it will consider the project unsafe if it contains any C extensions or datafiles whatsoever.`

But more importantly... This flag works for eggs right? I believe `pip` will always extract the package anyway...

PR #3708, opened 2022-02-09
clang-format comment prep wip

rwgk:
## Description
Helper/scratch PR. See #3703 for game plan.

## Suggested changelog entry:

```rst

```

PR #3712, opened 2022-02-10
Final manual curation in preparation for global `clang-format`ing

rwgk:
## Description
After this PR is merged we are ready for the last fully-automatic step: PR #3713.

Informed by experimental work under https://github.com/pybind/pybind11/pull/3708, which is this PR + clang-format in pre-commit.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -44,10 +44,12 @@ struct buffer_info {
     void *ptr = nullptr;          // Pointer to the underlying storage
     ssize_t itemsize = 0;         // Size of individual items in bytes
     ssize_t size = 0;             // Total number of entries
-    std::string format;           // For homogeneous buffers, this should be set to format_descriptor<T>::format()
+    std::string format; // For homogeneous buffers, this should be set to
+                        // format_descriptor<T>::format()
     ssize_t ndim = 0;             // Number of dimensions
     std::vector<ssize_t> shape;   // Shape of the tensor (1 entry per dimension)
-    std::vector<ssize_t> strides; // Number of bytes between adjacent entries (for each per dimension)
+    std::vector<ssize_t> strides; // Number of bytes between adjacent entries
+                                  // (for each per dimension)
     bool readonly = false;        // flag to indicate if the underlying storage may be written to
 
     buffer_info() = default;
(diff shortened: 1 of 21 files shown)


include/pybind11/buffer_info.h
```diff
@@ -44,10 +44,12 @@ struct buffer_info {
     void *ptr = nullptr;          // Pointer to the underlying storage
     ssize_t itemsize = 0;         // Size of individual items in bytes
     ssize_t size = 0;             // Total number of entries
-    std::string format;           // For homogeneous buffers, this should be set to format_descriptor<T>::format()
+    std::string format; // For homogeneous buffers, this should be set to
```
henryiii, line 47: FWIW, for these sorts of things, I'd do something like:

```
<empty line>
// Comment
std::string thing;
```

That will format much better, remain readable, and use less horizontal space for some extra vertical space.

No need to do that now. But that would be my general solution.

include/pybind11/cast.h
```diff
@@ -994,7 +994,9 @@ PYBIND11_NAMESPACE_BEGIN(detail)
 template <typename T, enable_if_t<!is_pyobject<T>::value, int>>
 object object_or_cast(T &&o) { return pybind11::cast(std::forward<T>(o)); }
 
-struct override_unused {}; // Placeholder type for the unneeded (and dead code) static variable in the PYBIND11_OVERRIDE_OVERRIDE macro
+// Placeholder type for the unneeded (and dead code) static variable in the
+// PYBIND11_OVERRIDE_OVERRIDE macro
+struct override_unused {};
```
henryiii, line 999: Like this. :)

henryiii (approved):

include/pybind11/buffer_info.h
Skylion007: Totally agree.

rwgk: I was going back and forth myself exactly here, then decided clang-format is actually doing a very nice job, preserving the original intent (no semi-pun intended). It's not obvious here, but if you look under #3713 it's great there. — But I'm happy to change it around. What do you prefer? Put only some comments on top, as needed? Or all, systematically?

henryiii: Let's go with that then. Merge at your discretion.

rwgk:
The testing for #3713 succeeded. I will merge this now. I will then rebase #3713 on master and run the testing once again, although it should be exactly the same, out of an abundance of caution.

PR #3713, opened 2022-02-10
Fully-automatic clang-format with include reordering

rwgk:
## Description
To be merged after PR #3712.

## Suggested changelog entry:

```rst
clang-format was added to the pre-commit actions, and the entire code base automatically reformatted (after several iterations preparing for this leap).
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -145,12 +145,3 @@ repos:
   rev: "v13.0.0"
   hooks:
   - id: clang-format
-
-- repo: local
-  hooks:
-  - id: check-style
-    name: Classic check-style
-    language: system
-    types:
-    - c++
-    entry: ./tools/check-style.sh
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -141,6 +141,11 @@ repos:
     entry: PyBind|Numpy|Cmake|CCache|PyTest
     exclude: .pre-commit-config.yaml
 
+- repo: https://github.com/pre-commit/mirrors-clang-format
+  rev: "v13.0.0"
+  hooks:
+  - id: clang-format
+
 - repo: local
   hooks:
   - id: check-style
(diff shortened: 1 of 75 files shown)


.clang-format
```diff
@@ -9,11 +9,28 @@ BinPackParameters: false
...
+IncludeBlocks: Regroup
 IndentCaseLabels: true
 IndentPPDirectives: AfterHash
 IndentWidth: 4
 Language: Cpp
 SpaceAfterCStyleCast: true
 Standard: Cpp11
+StatementMacros: ['PyObject_HEAD']
 TabWidth: 4
+IncludeCategories:
```
henryiii, line 21: This could go in before / be rebased in, doesn't need to be in the automatic commit, since we aren't running clang-format till this point.

henryiii (approved):

.clang-format
```diff
@@ -9,11 +9,28 @@ BinPackParameters: false
 BreakBeforeBinaryOperators: All
 BreakConstructorInitializers: BeforeColon
 ColumnLimit: 99
```
Skylion007, line 11: i would add some comment pragmas for clang-tidy nolints

rwgk: Do you have a pointer handy to some time-tested config? Or a concrete suggestion? Happy to fold that in. (I just don't want to invent an incompatible wheel, not currently needing anything in particular.)

henryiii: [CommentPragmas](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) is a regular expression, can you just match it with a nolint comment? That's quite nice, actually.

rwgk:
(The [🐍 3.5 • MSVC 2017 • x64](https://github.com/pybind/pybind11/runs/5145935901?check_suite_focus=true#logs) failure is an infrastructure flake, connection timed out. Safe to ignore.)

rwgk:
Screen Shot 2022-02-10 at 12 16 42

PR #3715, opened 2022-02-10
chore(clang-tidy): Add missing clang-tidy fixes

Skylion007:
## Description
* The recent clang-tidy fixes did not have full coverage because of various ifdefs. I ran clang-tidy on my local machine and found lots of errors. Adding the fixes in this PR. We may want to use cmake to run clang-tidy on every CI to catch this or at least have the clang-tidy config use the debug config.


rwgk:
Awesome, thanks for catching this! I'll merge this now to get this into #3712 & #3713.

For later: clang-tidy is very slow unfortunately, running in every CI could be a drag. Maybe 4 total: (clang, gcc) x (NDEBUG yes/no)?

PR #3719, opened 2022-02-11
chore: drop Python 3.5

henryiii:
Dropping Python 3.5 support.

- chore: drop Python 3.5 support
- chore: more fstrings with flynt's help
- ci: drop Python 3.5
- chore: bump dependency versions

## Suggested changelog entry:

```rst
* Removed Python 3.5 support.
```


henryiii:
> Note that you currently cannot combine this with a ``py::args`` argument.

I think this was fixed…

henryiii:
Forgot we have a pytest.ini for some situations. Will have to check.

setup.py
```diff
@@ -41,12 +41,10 @@ def build_expected_version_hex(matches: Dict[str, str]) -> str:
                     serial = int(level_serial[len(level) :])
                     break
     if serial is None:
-        msg = 'Invalid PYBIND11_VERSION_PATCH: "{}"'.format(patch_level_serial)
+        msg = f'Invalid PYBIND11_VERSION_PATCH: "{patch_level_serial}"'
```
Skylion007, line 44: Won't this just cause a parse error before the old python version even gets to the version check. I don't like using f-strings in setup.py for this reason.

henryiii: Our `setup.py` shouldn't run in Python <3.6, since we use python_requires. And if it does, the string error is an immediate indication that this is clearly Python 3.6+ only, as it's the most common error 3.6+ only error - I'd say it's only a hair worse than a dedicated version check & message, which we don't have (since I think they are rather pointless inside setup.py, due to python_requires). Finally, this is very much developer only, as this setup.py doesn't end up in SDists.

henryiii:
I'd rather disallow non-strings via pylint, vs. have flynt always running to try to convert them. It's actually only a hair stronger at converting them than pyupgrade unless you pass `--aggressive`, anyway.

henryiii:
I left off the Flynt addition, but kept the smart quotes one + fix.

include/pybind11/detail/common.h
```diff
@@ -211,8 +211,9 @@
 #endif
 
 #include <Python.h>
-#if PY_VERSION_HEX < 0x030500f0
-#    error "PYTHON 2 IS NO LONGER SUPPORTED. pybind11 v2.9 was the last to support Python 2."
+#if PY_VERSION_HEX < 0x03060000
+#    error                                                                                        \
+        "PYTHON < 3.6 IS NO LONGER SUPPORTED. pybind11 v2.9 was the last to support Python 2 and 3.5."
```
henryiii, line 216: Ouch, that's weird. Probably should shorten the line a little. UNSUPPORTED for example.

henryiii:
Any thoughts to the failing valgrind run? Will not be able to work on this more for now.

henryiii:
We have 3 new errors in valgrind when going from numpy 1.21 to numpy 1.22, so I left this at numpy 21 for Python < 3.10. Ready for review.

PR #3720, opened 2022-02-11
style: pylint

henryiii:
## Description

Builds on #3719, adds pylint, which discovered a mistake in the recent reworking of intree helpers.

## Suggested changelog entry:

None needed.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -22,14 +22,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
+    - name: Add matchers
+      run: echo "::add-matcher::$GITHUB_WORKSPACE/.github/matchers/pylint.json"
     - uses: pre-commit/action@v2.0.3
       with:
         # Slow hooks are marked with manual - slow is okay here, run them too
         extra_args: --hook-stage manual --all-files
-    - name: PyLint
-      run: |
-        echo "::add-matcher::$GITHUB_WORKSPACE/.github/matchers/pylint.json"
-        pipx run nox -s pylint
 
   clang-tidy:
     # When making changes here, please also review the "Clang-Tidy" section
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -12,6 +12,9 @@ on:
     - stable
     - "v*"
 
+env:
+  FORCE_COLOR: 3
+
 jobs:
   pre-commit:
     name: Format
@@ -23,6 +26,10 @@ jobs:
       with:
         # Slow hooks are marked with manual - slow is okay here, run them too
         extra_args: --hook-stage manual --all-files
+    - name: PyLint
+      run: |
+        echo "::add-matcher::$GITHUB_WORKSPACE/.github/matchers/pylint.json"
+        pipx run nox -s pylint
 
   clang-tidy:
     # When making changes here, please also review the "Clang-Tidy" section
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,22 +87,29 @@ repos:
   - id: rst-directive-colons
   - id: rst-inline-touching-normal
 
-# Flake8 also supports pre-commit natively (same author)
-- repo: https://github.com/PyCQA/flake8
-  rev: "4.0.1"
-  hooks:
-  - id: flake8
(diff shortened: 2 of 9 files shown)


.github/workflows/format.yml
```diff
@@ -23,6 +26,10 @@ jobs:
       with:
         # Slow hooks are marked with manual - slow is okay here, run them too
         extra_args: --hook-stage manual --all-files
+    - name: PyLint
```
Skylion007, line 29: Just add pylint as another pre-commit hook, it has native support.

henryiii: Pylint usually requires it be installed into the same environment that your package is installed into. Except in this case, where local runs are actually just fine. 🤦 It's also often slower, but again, not really since this is small.

Skylion007: Ah nvm, it has the same problem as flake8, mypy, and other checks. :/

setup.py
```diff
@@ -41,12 +41,10 @@ def build_expected_version_hex(matches: Dict[str, str]) -> str:
                     serial = int(level_serial[len(level) :])
                     break
     if serial is None:
-        msg = 'Invalid PYBIND11_VERSION_PATCH: "{}"'.format(patch_level_serial)
+        msg = f'Invalid PYBIND11_VERSION_PATCH: "{patch_level_serial}"'
```
Skylion007, line 44: I would drop fstrings from here, they will just cause syntax errors if someone tries to pip install from an old setup.py before they hit the version checks, right?

.github/CONTRIBUTING.md
```diff
@@ -93,11 +93,10 @@ cmake --build build -j4
 
 Tips:
 
-* You can use `virtualenv` (from PyPI) instead of `venv` (which is Python 3
-  only).
+* You can use `virtualenv` (faster, from PyPI) instead of `venv`
 * You can select any name for your environment folder; if it contains "env" it
   will be ignored by git.
-* If you don’t have CMake 3.14+, just add “cmake” to the pip install command.
+* If you don’t have CMake 3.14+, just add "cmake" to the pip install command.
```
Skylion007, line 99: @henryiii I know some commit hooks we can add to ban smart quotes, if you want.

henryiii: :heart: (though a simple pygrep would do it, as well, I'd think)

.pre-commit-config.yaml
```diff
@@ -76,8 +82,7 @@ repos:
   - id: python-check-blanket-noqa
   - id: python-check-blanket-type-ignore
   - id: python-no-log-warn
-  # Python 3.6
-  # - id: python-use-type-annotations
+  - id: python-use-type-annotations
```
Skylion007, line 85: You can change a mypy setting to have it complain about these too.

.github/CONTRIBUTING.md
Skylion007: @henryiii It also banned some smart apostophe, we can tweak the commit hook to allow them if you want.

henryiii:
Ready whenever any/every one else is. Does include a fix for a recent regression in intree helpers.

pybind11/__main__.py
```diff
@@ -1,3 +1,5 @@
+# pylint: disable=missing-function-docstring
```
rwgk, line 1: I think it would be better to move this to the `messages_control.disable` list in pyproject.toml.

I often see people writing completely redundant (useless) docstrings just to appease the linter. In my experience it is better left for the review stage, requesting docstrings as needed, if something isn't clear. (Even then, I usually encourage clearer variable and function naming over adding docstrings.)

rwgk (approved):

pybind11/__main__.py
henryiii: We already have docstrings on everything except `__main__` (where they are very much not needed), and I don't expect this to change much. I did disable the module docstring one, as that would be overkill. How about we only disable it (further) if it gets in the way?

Skylion007 (approved):

pybind11/__main__.py
rwgk: It's fine.
Internally we have our linter configured to warn about function docstrings only if the function is more than N lines (10 maybe, not sure). I find that useful.
But we can revisit as needed.

PR #3722, opened 2022-02-12
Dropping MSVC 2015

rwgk:
## Description
This PR drops support for MSVC 2015, and systematically removes workarounds for MSVC 2015.

Some workarounds marked as needed for MSVC 2015 are actually also needed for some other compilers, even including an old version of clang, and the newest version of MSVC. The `#ifdef`s for these workarounds were made as specific as possible, based on CI failures.

The only remaining AppVeyor build was moved from MSVC 2015 to 2017.

## Suggested changelog entry:

```rst
Support for MSVC 2015 was dropped. Workarounds for MSVC 2015 were systematically removed.
```

Upgrade guide:
```rst
The minimum version for MSVC is now 2017.
```


rwgk:
Reference point:
* 59 successful (we don't want that number to be lower for the rest of this PR)
* 3 failing (2 GHA, 1 appveyor)

Screen Shot 2022-02-11 at 17 23 46

Skylion007 (approved):

include/pybind11/detail/common.h
```diff
@@ -824,10 +822,12 @@ struct is_template_base_of_impl {
 /// Check if a template is the base of a type. For example:
 /// `is_template_base_of<Base, T>` is true if `struct T : Base<U> {}` where U can be anything
 template <template <typename...> class Base, typename T>
-#if !defined(_MSC_VER)
+#if defined(PYBIND11_CPP20) || !defined(_MSC_VER)                                                 \
+    || _MSC_VER > 1930 /* Increment as needed, let MS know please. */
+//     https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros
```
henryiii, line 827: A little confused, what's the current MSVC number, 1930 or 1931? This isn't something we have to increment every MSVC release, is it? If it doesn't work now, we should assume we don't know when it will start working. Once it does, then we can set this number. Only put real, known limits, not "I hope they fix this" numbers. I"m guessing it is fine in C++20 mode, though - we test on MSVC C++20?

include/pybind11/detail/common.h
```diff
@@ -1031,8 +1028,12 @@ PYBIND11_NAMESPACE_END(detail)
 ///  - regular: static_cast<Return (Class::*)(Arg0, Arg1, Arg2)>(&Class::func)
 ///  - sweet:   overload_cast<Arg0, Arg1, Arg2>(&Class::func)
 template <typename... Args>
+#    if (defined(_MSC_VER) && _MSC_VER < 1920) /* MSVC 2017 */                                    \
+        || (defined(__clang__) && __clang_major__ == 5)
 static constexpr detail::overload_cast_impl<Args...> overload_cast = {};
-// MSVC 2015 only accepts this particular initialization syntax for this variable template.
+#    else
+static constexpr detail::overload_cast_impl<Args...> overload_cast;
+#    endif
```
henryiii, line 1036: Why is the workaround longer now?

henryiii (commented):
More workarounds removed than I thought, nice! Couple of minor things.

docs/upgrade.rst
```diff
@@ -411,7 +411,7 @@ error is raised if the compiler does not meet the requirements:
 
 * GCC >= 4.8
 * clang >= 3.3 (appleclang >= 5.0)
-* MSVC >= 2015u3
+* MSVC >= 2017
```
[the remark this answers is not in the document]
Skylion007: I have the same nit here.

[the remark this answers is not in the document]
rwgk: Oops — thanks for catching this! I was too much in lawn-mower mode when going through here.

Reverted.

I added mention for the upgrade guide to the PR description.

include/pybind11/detail/common.h
rwgk: Changed as suggested.

Current is 1930.

We have 3 jobs total with C++20:
* Clang 10 x64
* GCC 10 x64
* MSVC 2019 x86

I think I overreacted. I was a pretty unhappy to discover that even MSVC 2019 and 2022 ICE without this workaround. Thinking: heck, if we didn't mask it here for any MSVC, it may have gotten fixed in all those years. But at second thought, my idea to force us to increment the version with every MSVC release — until there is finally a fix — will backfire badly, because it'll force people to update pybind11, or patch locally, to be able to use a new MSVC release.

Another idea is to gate on `_MSC_VER >= 1940`, with the idea to catch ICEs for the next major update of MSVC, but who knows, really, what the correct number will be? — Staying away from it.

include/pybind11/detail/common.h
rwgk: Before there was **only** the workaround, with a comment that marked it as such.

Now there is what (I assume) the original authors wanted to use, but couldn't, plus the workaround.

I could change it back to only-workaround-with-updated-comment, but at this point in time we're just a couple years away from being able to purge the workaround. I think it's nice to be sure that there aren't any modern compilers having trouble with the default construction. It's kind-of the point of beta testing compilers, that we want to catch such things.

rwgk:
@henryiii Could you please take another look?

include/pybind11/detail/common.h
henryiii: Yes. Do we know it works on C++20? The old version simply always used the workaround, the new one seems to think MSVC will be happy with this on C++20 mode. Is it?

rwgk: We don't know before or after my change. The old version was guilty (needs workaround) until proven innocent, which has the drawback of covering up flaws even for new compilers. The new is innocent until proven guilty. I'm trying it out now. My first attempt seems to only kind-of work:
```
Run cmake -S . -B . -DPYBIND11_WERROR=ON -DDOWNLOAD_CATCH=ON -DDOWNLOAD_EIGEN=ON -DCMAKE_CXX_STANDARD=11 -DCMAKE_CXX_STANDARD=20
-- Building for: Visual Studio 17 2022
```
If it doesn't get too messy we can fold it in here.

rwgk: eigen is acting up:
```
test_eigen.cpp
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/Matrix.h([46](https://github.com/pybind/pybind11/runs/5179409281?check_suite_focus=true#step:11:46),40): error C2220: the following warning is treated as an error [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/PlainObjectBase.h(98): message : see reference to class template instantiation 'Eigen::internal::traits<Derived>' being compiled [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
          with
          [
              Derived=Eigen::Matrix<double,-1,-1,0,-1,-1>
          ]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/Matrix.h(180): message : see reference to class template instantiation 'Eigen::PlainObjectBase<Eigen::Matrix<double,-1,-1,0,-1,-1>>' being compiled [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\tests\test_eigen.cpp(39): message : see reference to class template instantiation 'Eigen::Matrix<double,-1,-1,0,-1,-1>' being compiled [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/Matrix.h(46,40): warning C[50](https://github.com/pybind/pybind11/runs/5179409281?check_suite_focus=true#step:11:50)54: operator '&': deprecated between enumerations of different types [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/Matrix.h(46,40): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(211,29): warning C5054: operator '&': deprecated between enumerations of different types [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(211,29): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\tests\test_eigen.cpp(110): message : see reference to class template instantiation 'Eigen::TriangularView<const MatrixType,1>' being compiled [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
          with
          [
              MatrixType=Eigen::Matrix<double,-1,-1,1,-1,-1>
          ]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(212,29): warning C5054: operator '&': deprecated between enumerations of different types [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(212,29): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(213,29): warning C5054: operator '&': deprecated between enumerations of different types [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(213,29): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(214,29): warning C5054: operator '&': deprecated between enumerations of different types [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\_deps\eigen-src\Eigen\src/Core/TriangularMatrix.h(214,29): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
D:\a\pybind11\pybind11\include\pybind11\cast.h(1014,1): error C27[52](https://github.com/pybind/pybind11/runs/5179409281?check_suite_focus=true#step:11:52): 'pybind11::detail::type_caster<Eigen::Ref<Eigen::Matrix<double,-1,-1,0,-1,-1>,0,Eigen::OuterStride<-1>>,void>': more than one partial specialization matches the template argument list [D:\a\pybind11\pybind11\tests\pybind11_tests.vcxproj]
```

henryiii: The old version had a harmless workaround that might not be needed. The new version has a probability of breaking things that would have worked. We try to guess the safest path unless we test it to see it work. If we can test it and it works in C++20 mode, great! Otherwise err on the side of caution.

henryiii: A nice developer way to disable workarounds might be nice, but all at once would be too much. Maybe a standard searchable comment?

rwgk: > Otherwise err on the side of caution.

Idk, I always cared about helping new compilers getting better. It diminishes the benefit of beta testing-compilers if we cover up anyway.

rwgk:
Merging this now, to start work on the smart_holder branch, catching up with the clang-format changes (PR #3713), the Python drops (PRs #3688 & #3719), and this PR.

PR #3732, opened 2022-02-14
Adding MSVC 2022 C++20 GitHub Action

rwgk:
## Description
test_eigen is currently disabled (similar to the existing MSVC 2019 C++20 GHA), awaiting Eigen compatibility fixes.

See also:
* Immediate motivation for adding this GHA: https://github.com/pybind/pybind11/pull/3722#discussion_r805530113
* Eigen issues: https://github.com/pybind/pybind11/pull/3722#discussion_r805532743

## Suggested changelog entry:

```rst
MSVC 2022 C++20 coverage was added to GitHub Actions (test_eigen is currently disabled).
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -709,15 +709,22 @@ jobs:
         - 3.7
         - 3.8
         - 3.9
+        year:
+        - 2019
 
         include:
           - python: 3.9
+            year: 2019
             args: -DCMAKE_CXX_STANDARD=20 -DDOWNLOAD_EIGEN=OFF
           - python: 3.8
+            year: 2019
             args: -DCMAKE_CXX_STANDARD=17
+          - python: 3.9
+            year: 2022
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

rwgk:
> Can't we just add this to the 2019 jobs?

Thanks a lot for jumping in!

One difference I noticed is that the 2019 job doesn't include `--target cpptest` and `--target test_cmake_build`, therefore I decided to start by adding a completely new block.

henryiii:
Okay, reverted. We can clean up eventually. I also notice this is 32-bit only, which is odd.

I'm sad to see we actually had MSVC 2015 working on windows-2019, that does prove it's available there (which I had noticed in Scikit-build), but we can't run 2017 on GHA at all post windows-2016.

rwgk:
Thanks Henry! Merging, then retesting #3722.

PR #3734, opened 2022-02-14
tests: use 'build' in tests instead of running setup.py

henryiii:
## Description

Use build instead of direct setup.py commands in our tests. Mostly by @abravalheri, pulled from #3711.

## Suggested changelog entry:

```rst
* Avoid ``setup.py <command>`` usage in internal tests.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "cmake>=3.18", "ninja"]
+requires = ["setuptools>=42", "cmake>=3.18", "ninja"]
 build-backend = "setuptools.build_meta"
 
 [tool.check-manifest]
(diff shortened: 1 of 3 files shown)


rwgk (approved):
Out of my wheelhouse but LGTM.
Would be nice to add a sentence or two to the PR description, for build muggles like myself:
* Was it the Python 2 or 3.5 drop that made this possible?
* Why are we keeping setup.py? (Needed externally?)

henryiii:
`setup.py <command>` is [very deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). We shouldn't be using or promoting it. Using `build` is the correct alternative. This still runs setuptools, which still uses `setup.py` internally, but the use of commands is an implementation detail. We still heavily (for now) use setup.py. That might change in the future, though. :)

No, don't think it was the Python drop, it was just... Not sure why I used raw setup.py commands before, probably didn't realize the proper way to do it. Though the drop does make it easier to just add a specific build version.

PR #3740, opened 2022-02-15
Disable ipo by property

mathstuf:
## Description

IPO is expensive and it can cause out-of-memory in CI. It would be nice to be able to disable IPO in these cases.

## Suggested changelog entry:

```rst
* Targets may now set the `PYBIND11_NO_IPO` target property to `1` to opt-out of IPO for PyBind11 targets.
```

### DIFF
--- a/tools/pybind11Common.cmake
+++ b/tools/pybind11Common.cmake
@@ -328,11 +328,12 @@ function(_pybind11_generate_lto target prefer_thin_lto)
     # CONFIG takes multiple values in CMake 3.19+, until then we have to use OR
     set(is_debug "$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>")
     set(not_debug "$<NOT:${is_debug}>")
+    set(not_optout "$<NOT:$<BOOL:$<TARGET_PROPERTY:PYBIND11_NO_IPO>>>")
     set(cxx_lang "$<COMPILE_LANGUAGE:CXX>")
     if(MSVC AND CMAKE_VERSION VERSION_LESS 3.11)
-      set(genex "${not_debug}")
+      set(genex "$<AND:${not_debug},${not_optout}>")
     else()
-      set(genex "$<AND:${not_debug},${cxx_lang}>")
+      set(genex "$<AND:${not_debug},${cxx_lang},${not_optout}>")
     endif()
     set_property(
       TARGET ${target}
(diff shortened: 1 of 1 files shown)


henryiii:
Please use https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html and CMake 3.9+.

mathstuf:
The problem is that Pybind is adding the IPO flags for us and a way to turn it off would be nice.

henryiii:
Set CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF and it will not add flags, you can add them if you want them per-target by setting the `INTERPROCEDURAL_OPTIMIZATION` property.

henryiii:
`-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF` will even do this from the command line.

mathstuf:
Looking into it more, it is coming from Pybind11, but not here; it comes from the `_pybind11_add_lto_flags` function in the CMake interface. I see `NO_EXTRAS` seems to skip this. But the `INTERPROCEDURAL_OPTIMIZATION` property doesn't do anything about this source of the IPO flags.

henryiii:
You have to the variable `CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF` explicitly; there's no way to set a property before the target is made inside the function. But if you do, then INTERPROCEDURAL_OPTIMIZATION controls the target. There's no harm in setting `CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF`, as that's just like the default.

mathstuf:
I looked at this again. The LTO flags are coming from pybind11, *not* any `INTERPROCEDURAL_OPTIMIZATION` setting. Turning it off explicitly does nothing (because that is just the default).

henryiii:
No, we are supposed to check to see if `INTERPROCEDURAL_OPTIMIZATION` is defined when we are making the targets; if it is, we do nothing. If that doesn't work, it's a bug. From the CMake side, explicit OFF and not defined are the same, but for us, explicit OFF is how we know to not try to manually set IPO flags.

mathstuf:
Hmm. Maybe we're talking about different versions here…it seems we are pinned to a hash of some kind. Let me check about syncing that up with a more recent release.

henryiii:
https://github.com/pybind/pybind11/blob/da15bb206c21be7180bd428e4e69907756eb70dc/tools/pybind11Tools.cmake#L188

and

https://github.com/pybind/pybind11/blob/da15bb206c21be7180bd428e4e69907756eb70dc/tools/pybind11NewTools.cmake#L228

PR #3743, opened 2022-02-16
ci: test pypy upstream

henryiii:
## Description

Adding a nox test for PyPy upstream before their next release. I think this works on 3.7 and 3.8, though we should see if we still need our workarounds, and fails on PyPy 3.9, but haven't really looked into it more than just adding the tests. One test was regularly xpassing for me. CC @mattip.

## Suggested changelog entry:

```rst
* Support testing development versions of PyPy (UNIX only) via nox
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,11 +1,26 @@
 import os
+import sys
+from pathlib import Path
 
 import nox
 
 nox.needs_version = ">=2022.1.7"
 nox.options.sessions = ["lint", "tests", "tests_packaging"]
 
-PYTHON_VERISONS = ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.7", "pypy3.8"]
+DIR = Path(__file__).parent.resolve()
+
+PYTHON_VERISONS = [
+    "3.6",
+    "3.7",
+    "3.8",
+    "3.9",
+    "3.10",
+    "3.11",
+    "pypy3.7",
+    "pypy3.8",
+    "pypy3.9",
+]
+PYPY_VERSIONS = ["3.7", "3.8", "3.9"]
 
 if os.environ.get("CI", None):
     nox.options.error_on_missing_interpreters = True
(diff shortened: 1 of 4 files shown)


henryiii:
Hmmm, CMake failing to find PyPy 3.9 but reporting 3.9 is a valid version was not what I was expecting. That worked locally.

henryiii:
This is the local error:

```
nox > cmake --build .nox/pypy_upstream-3-9/tmp --target pytest
[0/2] Re-checking globbed directories...
[0/1] cd /Users/henryschreiner/git/software/pybind11/.nox/pypy_upstream-3-9/tmp/tests && /Users...tests/test_union.py /Users/henryschreiner/git/software/pybind11/tests/test_virtual_functions.py
libc++abi: terminating with uncaught exception of type pybind11::error_already_set: IndentationError: ('unexpected indent', ('<string>', 2, 1, '        class pybind11_static_property(property):\n', 2))
```

Skylion007:
Okay, so we should definitely make a PR to fix that leading indent then.

Skylion007:
Nice, PyPy, 3.7 and 3.8 are working now.

henryiii:
3.7 and 3.8 were always working. It's just 3.9 that's broken - via CMake on CI and by `_PyPyCMethod_New` missing (have no idea what that is) locally - only in one of the three CMake compile tests, I think.

noxfile.py
```diff
@@ -96,3 +110,58 @@ def build(session: nox.Session) -> None:
     session.run(
         "python", "-m", "build", *session.posargs, env={"PYBIND11_GLOBAL_SDIST": "1"}
     )
+
+
+@nox.session
+@nox.parametrize("pypy", PYPY_VERSIONS, ids=PYPY_VERSIONS)
+def pypy_upstream(session: nox.Session, pypy: str) -> None:
+    """
+    Test against upstream PyPI (64-bit UNIX only)
```
mattip, line 119:
```suggestion
    Test against upstream PyPy (64-bit UNIX only)
```

tests/test_exceptions.py
```diff
@@ -95,7 +95,11 @@ def ignore_pytest_unraisable_warning(f):
 
 
 # TODO: find out why this fails on PyPy, https://foss.heptapod.net/pypy/pypy/-/issues/3583
-@pytest.mark.xfail(env.PYPY, reason="Failure on PyPy 3.8 (7.3.7)", strict=False)
+@pytest.mark.xfail(
+    env.PYPY and sys.version_info >= (3, 8),
```
mattip, line 99:
```suggestion
    env.PYPY and sys.implementation.version < (7, 3, 8),
```

mattip:
As far as I can tell, PyPy3.9 has the include file `pypy_decl.h` with `#define PyCMethod_New PyPyCMethod_New`, and CPython has the include file `methodobject.h` with `PyAPI_FUNC(PyObject *) PyCMethod_New(PyMethodDef *, PyObject *, ...`. I wonder where `_PyPyCMethod_New` with the underline is coming from?

tests/test_exceptions.py
henryiii: I _think_ I saw this not xpassing. But I could try it.

henryiii: Nope.

mattip: Sorry for the noise. I wonder if we can get this to a more minimal reproducer

mattip:
I think CMAKE is not finding the include files nor the library for PyPy3.9.

henryiii:
Yes, but it is finding it for 3.8 & 3.7, AND it is finding it locally on my Mac for 3.9, even though everything should be identical - same pypy download, same cmake, etc...

mattip:
>  it is finding it locally on my Mac for 3.9,

Do the tests pass?

henryiii:
The normal tests pass, then the build tests fail with the error listed above.

mattip:
What would be the easiest way to run those specific tests locally?

henryiii:
Let me revert the last commit, that should get past that test.

henryiii:
Ahaha, I bet this is what's breaking CMake: https://github.com/scikit-build/scikit-build/pull/673 (at least it's a start)

DarkWingMcQuack:
> This is the local error:
> 
> ```
> nox > cmake --build .nox/pypy_upstream-3-9/tmp --target pytest
> [0/2] Re-checking globbed directories...
> [0/1] cd /Users/henryschreiner/git/software/pybind11/.nox/pypy_upstream-3-9/tmp/tests && /Users...tests/test_union.py /Users/henryschreiner/git/software/pybind11/tests/test_virtual_functions.py
> libc++abi: terminating with uncaught exception of type pybind11::error_already_set: IndentationError: ('unexpected indent', ('<string>', 2, 1, '        class pybind11_static_property(property):\n', 2))
> ```
> 
> This actually seems correct, the code has a leading indent. Not sure why PyPy < 3.9 likes it, actually.

I have this bug in a project currently which i try porting to pypy. Is there any fix for this currently?
Using python 3.8 is not an option unfortunatly for me

mattip:
@DarkWingMcQuack the indent is fixed on HEAD. You can install pybind11 from git until the fix is released: `pip install git+https://github.com/pybind/pybind11.git`.

henryiii:
It should also be available in the `v2.9` branch. Hopefully we can cut a new patch release soon.

henryiii:
The library name change in 3.9 is also likely breaking things, though, possibly including all older versions of CMake's FindPython support. I want to look into that before making a release. Might not be fixable from our end, though, for CMake.

mattip:
> The library name change in 3.9 is also likely breaking things

bummer.

nachovizzo:
I'm also seeing this error on a CI build, for Python 3.9

```sh
[root@bbf72c22780d ~]# "${PYBIN}/pytest" --capture=sys /io/
================================================ test session starts ================================================
platform linux -- Python 3.9.10[pypy-7.3.8-final], pytest-7.1.1, pluggy-1.0.0
rootdir: /io
collecting ... terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  IndentationError: ('unexpected indent', ('<string>', 2, 1, '        class pybind11_static_property(property):\n', 2))
Fatal Python error: Aborted
```

mattip:
@nachovizzo The indent is fixed on HEAD. You can install pybind11 from git until the fix is released: pip install git+https://github.com/pybind/pybind11.git.

nachovizzo:
> @nachovizzo The indent is fixed on HEAD. You can install pybind11 from git until the fix is released: pip install git+https://github.com/pybind/pybind11.git.

Great, already try this and fix the problem. Looking forward to a new release ;)

Issue #2101, opened 2020-01-29
make_iterator throws exception after re-initializing interpreter


bkloster-sma:
## Issue description

1. Initialize an interpreter
2. Call `make_iterator` on some begin/end pair of a container
3. Finalize interpreter, initialize it again
4. Call `make_iterator` again, same as above

Result: An exception is thrown "instance allocation failed: new instance has no pybind11-registered base types".

## Reproducible example code

```
    std::vector<int> container;

    // Initialize an interpreter
    pybind11::initialize_interpreter();

    // Call make_iterator
    {
        pybind11::gil_scoped_acquire gil;
        auto iter = pybind11::make_iterator(container.begin(), container.end());
    }

    // Finalizer interpreter, initialize it again
    pybind11::finalize_interpreter();
    pybind11::initialize_interpreter();

    // Call make iterator again
    {
        pybind11::gil_scoped_acquire gil;
        auto iter = pybind11::make_iterator(container.begin(), container.end());
    }
    pybind11::finalize_interpreter();
```

## Proposed Fix

The culprit seems to be static state in `pybind11::detail::registered_local_types_cpp()`. Clearing that container (deleting the heap-allocated values in it beforehand) in `finalize_interpreter()` appears to fix the issue.

bstaletic:
This does look like an actual bug. Here's a cleaner example that reproduces the error:
```
#include <vector>
#include <pybind11/embed.h>
#include <pybind11/stl.h>
#include <pybind11/pybind11.h>
int main() {
	std::vector<int> container;
	{
		pybind11::scoped_interpreter g;
		auto iter = pybind11::make_iterator(container.begin(), container.end());
	}
	{
		pybind11::scoped_interpreter g;
		auto iter = pybind11::make_iterator(container.begin(), container.end());
	}
}
```

bobbay:
I would really like to see a fix to this one. I am depending on this feature and my users can only run once and then close my app and start over.
I tried to find the location of the Proposed Fix in the headers files but could not find "registered_local_types_cpp".

Skylion007:
@bobbay It probably was refactored, but if it's static registered state than it's likely in get_internals()
https://github.com/pybind/pybind11/blob/4c7e509fa475d0a0ede726267bc451e4f0583446/include/pybind11/detail/internals.h#L101 Seems like registered_types_cpp might be what was referred to.

Skylion007:
Confirmed that the above link is correct as #2650 renamed it.

PR #3744, opened 2022-02-17
fix: clear local internals after finalizing interpreter #2101

StarQTius:
## Description

The author of #2101 proposed to clear the local internals after finalizing an interpreter. It seems that when registering a type as local, some of the data is managed by the current interpreter through pointers, which are then dangling when the interpreter is finalized. I implemented the solution suggested by the author, as it does not seem that keeping the local internals between two consecutive interpreters is the expected behavior.

### DIFF
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -198,7 +198,7 @@ inline void finalize_interpreter() {
     if (builtins.contains(id) && isinstance<capsule>(builtins[id])) {
         internals_ptr_ptr = capsule(builtins[id]);
     }
-
+    // Local internals contains data managed by the current interpreter, so we must clear them to avoid undefined behaviors when initializing another interpreter
     detail::get_local_internals().registered_types_cpp.clear();
     detail::get_local_internals().registered_exception_translators.clear();
 
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -199,6 +199,9 @@ inline void finalize_interpreter() {
         internals_ptr_ptr = capsule(builtins[id]);
     }
 
+    detail::get_local_internals().registered_types_cpp.clear();
+    detail::get_local_internals().registered_exception_translators.clear();
+
     Py_Finalize();
 
     if (internals_ptr_ptr) {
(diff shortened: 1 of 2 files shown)


rwgk:
> Seems fine to me, though I don't use this functionality much. @Skylion007 or @rwgk?

+1

I could run this through Google's global testing, then merge if there are no issues.

There is a chance that even the global testing doesn't hit this change, embedding is on our "don't use" list.

But this PR looks good to me, nice test.

Skylion007 (approved):
This looks good to me. Although I would appreciate in a comment in the test which links back to the issue, and maybe another comment in embed.h about why clearing the interpreters is necessary.

StarQTius:
> This looks good to me. Although I would appreciate in a comment in the test which links back to the issue, and maybe another comment in embed.h about why clearing the interpreters is necessary.

It's done. I'm sorry for the extra commit.

henryiii (approved):

rwgk:
The more the merrier. They get squashed anyway when we merge. I'll try to
get this globally tested asap.

include/pybind11/embed.h
```diff
@@ -198,6 +198,10 @@ inline void finalize_interpreter() {
     if (builtins.contains(id) && isinstance<capsule>(builtins[id])) {
         internals_ptr_ptr = capsule(builtins[id]);
     }
+    // Local internals contains data managed by the current interpreter, so we must clear them to
+    // avoid undefined behaviors when initializing another interpreter
+    detail::get_local_internals().registered_types_cpp.clear();
+    detail::get_local_internals().registered_exception_translators.clear();
```
rwgk, line 204: Just to note: I verified that new `TEST_CASE` covers  `.registered_types_cpp.clear()`, but it does not cover `.registered_exception_translators.clear()`.
I think that's fine though, although if someone could add that test coverage in a follow-on PR that would be ideal.

** "covers" in the sense that the test breaks if the call is removed.

rwgk (approved):
Global testing went fine.

I actually ran it twice:
1. Exactly this PR: everything worked (10s of millions of tests).
2. With an added `pybind11_fail` in pybind11/embed.h: exactly 1 failure ([this one fwiw](https://github.com/google/atheris/blob/d75570b8635a452cc459f56ac53b5fdeb1c0b984/src/native/util_test.cc)).

Learned:
* This PR is good.
* I'm beginning to think adding test_embed to the build config for internal testing isn't difficult. (But later, maybe.)

PR #3758, opened 2022-02-24
cast: Qualify symbol usage in PYBIND11_TYPE_CASTER

EricCousineau-TRI:
## Description

Permits using macro outside of pybind11::detail

Motivated by https://github.com/RobotLocomotion/drake/pull/16625#pullrequestreview-892816740

## Suggested changelog entry:

```rst
``PYBIND11_TYPE_CASTER`` now uses fully qualified symbols, so it can be used outside of ``pybind11::detail``.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -85,11 +85,15 @@ protected:                                                                      
                                                                                                   \
 public:                                                                                           \
     static constexpr auto name = py_name;                                                         \
-    template <typename T_, enable_if_t<std::is_same<type, remove_cv_t<T_>>::value, int> = 0>      \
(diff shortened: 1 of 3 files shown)


rwgk:
Quick drive-by comment/question:
* Looks great!
* Any chance you could add a test? (separate test_type_caster_outside_detail (or similar) would be my first choice)
Without the test we will always be in doubt (in the best case) or break you (more likely).

EricCousineau-TRI:
Thanks! And great point on testing - fixed an error :sweat_smile: See 1d1dbb4

rwgk (approved):
Awesome. I agree it fits well into tests/test_custom_type_casters.cpp

Skylion007 (approved):
Yet another reason to avoid using macros.

PR #3761, opened 2022-02-26
Adding "make_alias" to use a different `__name__` as shared-object

tdegeus:
## Description

I have a set of functions written in C++ that I compile in a module `_foo`.
My python module simply imports all functions from it. I.e. `foo/__init__.py` reads
```python
from _foo import *
```
This was giving issues with my documentation, and did not give nice error messages. This PR fixes this.
See https://github.com/pybind/pybind11/discussions/3748

## Suggested changelog entry:

```rst
Adding "make_alias" to use a different `__name__` as shared-object
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1245,6 +1245,28 @@ public:
         //       For Python 2, reinterpret_borrow was correct.
         return reinterpret_borrow<module_>(m);
     }
+
+    struct strip_leading_underscore_from_name;
+
+    static module_ make_alias(const char *name) {
+        module_ ret = *this;
+        ret.attr("__name__") = this.attr(name);
+        return ret;
+    }
+
+    static module_ make_alias(strip_leading_underscore_from_name) {
+        const char *name = PyModule_GetName(*this);
+        size_t n = strlen(name);
+        if (n == 0) {
+            return this->make_alias(name);
+        }
+        if (name[0] != "_") {
+            return this->make_alias(name);
+        }
+        char alias[n - 1];
+        std::copy(&name[1], &name[n - 1], &alias[0]);
+        return this->make_alias(alias);
+    }
 };
 
 // When inside a namespace (or anywhere as long as it's not the first item on a line),
--- a/tests/test_make_alias.cpp
+++ b/tests/test_make_alias.cpp
@@ -1,0 +1,22 @@
+/*
+    tests/make_alias -- make alias
+
+    Copyright (c) 2016 Wenzel Jakob <wenzel.jakob@epfl.ch>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#include <pybind11/pybind11.h>
+
+#include "constructor_stats.h"
+#include "pybind11_tests.h"
+
+#if defined(_MSC_VER)
+#    pragma warning(disable : 4996) // C4996: std::unary_negation is deprecated
+#endif
+
+TEST_SUBMODULE(_make_alias, m) {
+    ma = m.make_alias(pybind11::module::strip_leading_underscore_from_name);
+    ma.def("foo", []() -> { });
+}
--- a/tests/test_make_alias.py
+++ b/tests/test_make_alias.py
@@ -1,0 +1,8 @@
+import pytest
+
+from pybind11_tests import ConstructorStats
+
+m = pytest.importorskip("pybind11_tests.make_alias")
+
+def assert_name(mat):
+    assert m.__name__ == "make_alias"


tdegeus:
I think I need some help...

rwgk:
Awesome, thanks! I can only glance through atm, I believe this is definitely on the right path. I look more asap.

include/pybind11/pybind11.h
```diff
@@ -1245,6 +1245,28 @@ class module_ : public object {
...
+        size_t n = strlen(name);
+        if (n == 0) {
+            return make_alias(name);
+        }
+        if (strcmp(&name[0], "_") != 0) {
+            return make_alias(name);
+        }
+        char alias[n - 1];
+        std::copy(&name[1], &name[n - 1], &alias[0]);
+        return make_alias(alias);
```
skoslowski, line 1268:
```suggestion
        if (name && name[0] == '_')
            name += 1;
        return make_alias(name);
```

tests/test_make_alias.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "constructor_stats.h"
+#include "pybind11_tests.h"
+
+#if defined(_MSC_VER)
+#    pragma warning(disable : 4996) // C4996: std::unary_negation is deprecated
+#endif
+
+TEST_SUBMODULE(_make_alias, m) {
+    ma = m.make_alias(pybind11::module::strip_leading_underscore_from_name);
```
skoslowski, line 20:
```suggestion
    ma = m.make_alias(pybind11::module::strip_leading_underscore_from_name{});
```

tests/test_make_alias.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+from pybind11_tests import ConstructorStats
+
+m = pytest.importorskip("pybind11_tests.make_alias")
```
skoslowski, line 5:
```suggestion
m = pytest.importorskip("pybind11_tests._make_alias")
```

tests/test_make_alias.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def assert_name(mat):
+    assert m.__name__ == "make_alias"
```
skoslowski, line 9: Not sure this is an effective test. You would probably need to check the qualname of an object (class) in the module, maybe have another class in there that has not been aliased

skoslowski (commented):
I like this feature! Thanks for working on this. 

I hope these comments help you working towards a solution.
Note, that you'll have to register your test in tests/CMakeLists.txt:117 to have it executed.

rwgk:
I looked at this for >1 hour and got something that at least compiles and more-or-less runs, but I'm afraid & sorry my plan is making a crash landing: `__name__` lives in the same `__dict__` that's populated by `.def` and `class_`:

https://github.com/python/cpython/blob/de2c91a89aa0e89d5d01d71697d0f27a9f60c11d/Objects/moduleobject.c#L17

`PyModuleObject` actually has this:
```
PyObject *md_name;  /* for logging purposes after md_dict is cleared */
```
which is almost what we need (separate name) but it's unfortunately only used for "logging".

Therefore there is no way to achieve what I wrote before:

> to create a `module_alias`, a new object, which has all the attributes we need, exactly the same is `m`, except `__name__`

We could backtrack to the original `ScopedModuleOverride`, but it really is a hack. Good enough? Idk.

Another idea is to introduce a special attribute used by by `.def` and `class_`. Currently they look at `__module__` first then `__name__`. They could also look at `__name_alias__` or similar. Could be really simple. Could be a can of worms. I don't have time at the moment to look into this more.

If someone wants to experiment, I believe (but I'm NOT sure), here is where `.def` looks for __module__` first then `__name__`:

https://github.com/pybind/pybind11/blob/f495dfc4339df1454620ac2e7c0f5561959e86af/include/pybind11/pybind11.h#L499

And here `class_`:

https://github.com/pybind/pybind11/blob/f495dfc4339df1454620ac2e7c0f5561959e86af/include/pybind11/detail/class.h#L630

tdegeus:
Your assessment makes sense. Too bad though, because I do think that the problem that this PR was supposed to address is relevant, as:
1. Having a hybrid C++/Python module, with functions in the same namespace, is something quite some users might want.
2. `scikit-build`, which does simplify builds quite a bit, enforces the structure above.

rwgk:
After sleeping over this for a couple nights, I believe going into pybind11, the places I pointed out, is "the right thing". What we **really** want is to give the wrapped functions and classes a custom __module__ or __name__ (I'm currently uncertain about the exact details of those). Messing with the module object — via transient in-place manipulations — to achieve that is obviously a hack. Doing what we want more directly looks easy*** and avoids the pitfalls.

*** of course I could be wrong again. someone needs to try. someone who believes they need it. (I don't, for my own purposes, atm)

PR #3779, opened 2022-03-04
Docs: No Strip in Debug

ax3l:
## Description

The docs were not 100% the same as we advertise and do with our high-level tooling function: most users do not want to strip symbols in Debug builds.

## Suggested changelog entry:

```rst
* compilation documentation: guard ``pybind11_strip`` to keep symbols in debug builds
```

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -509,7 +509,10 @@ You can use these targets to build complex applications. For example, the
     target_link_libraries(example PRIVATE pybind11::module pybind11::lto pybind11::windows_extras)
 
     pybind11_extension(example)
-    pybind11_strip(example)
+    if(NOT MSVC AND NOT ${CMAKE_BUILD_TYPE} MATCHES Debug|RelWithDebInfo)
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

henryiii:
Did you mean https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html ?

ax3l:
No, I mean I just copied what we do in *Tools.cmake

ax3l:
https://github.com/pybind/pybind11/blob/v2.9.1/tools/pybind11Tools.cmake#L203

ax3l:
@henryiii does this clarify things? :)

henryiii:
We have to support CMake 3.4+ in our code. But 3.9+ can use `GENERATOR_IS_MULTI_CONFIG`, which is what we actually mean - it's not MSVC specific, it's just not valid on multi config generators, including Xcode & Ninja in multiconfig mode.

PR #3789, opened 2022-03-09
ci: test pypy 3.9

henryiii:
## Description

Trying out pypy 3.9

## Suggested changelog entry:

```rst

```


Skylion007 (approved):

henryiii:
I think we need to see if the FindPython mode works...

Skylion007:
@henryiii Does our CI Tests not cover that?

henryiii:
Only a few jobs do, with it injected via `include:`.

henryiii:
(Even though it's not showing up here, CI is running)

Issue #3788, opened 2022-03-09
[BUG]: __getattr__ and smart_holders


petersteneteg:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

This is a problem in the `smart_holder` branch

When defining `__getattr__` in the bindings for a class (see example code below) using the `master` pybind11 branch I can do the following:
```
In [1]: import foo
In [2]: f = foo.Foo()
In [3]: f.bar()
Out[3]: 42
In [5]: f.something
Out[5]: 'GetAttr: something'
```
and everything works as expected. 

But when using the `smart_holder` branch and defining `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT`
either calling `f.bar()` or `f.something` results in infinite recursion, hanging the runtime. Using "conservative mode" gives the same result. 

Im not sure if I'm doing anything bad, and I can not find anything about it not being supported it in the documentation, and it works perfectly fine in the `master` branch. 

### Reproducible example code

```text
#include <pybind11/pybind11.h>

struct Foo {
    Foo() = default;
    int bar() const { return 42; }
};

namespace py = pybind11;

PYBIND11_MODULE(foo, m) {
    py::class_<Foo>(m, "Foo")
        .def(py::init<>())
        .def("bar", &Foo::bar)
        .def("__getattr__", [](Foo&, std::string key) { return "GetAttr: " + key; });
}
```

petersteneteg:
Btw, just using the `smart_holder` without defining `PYBIND11_USE_SMART_HOLDER_AS_DEFAULT` or using `py::classy` works just fine.

petersteneteg:
The recursive call stack I get looks like this:
```
    .
    .
    .
    frame #16945: 0x0000000100e8fb44 foo.cpython-310-darwin.so`pybind11::cpp_function::dispatcher(self=0x0000000100c36760, args_in=0x0000000100c18300, kwargs_in=0x0000000000000000) at pybind11.h:925:30
[...]

PR #3795, opened 2022-03-10
Fix for #3788 __getattr__ issue

rwgk:
## Description

Moving `try_as_void_ptr_capsule()` to be last in `load_impl()`. I.e. it will only be called if there is no other way to load `src`. This avoids the infinite recursion reported under #3788.

## Suggested changelog entry:

```rst

```


rwgk:
@wangxf123456 is continuing the work under #3796.
I'll close this PR as soon as the CI is finished (almost done). It's on track to be completely successful.

PR #3797, opened 2022-03-11
PYBIND11_OBJECT_CVT should use namespace for error_already_set()

oleksandr-pavlyk:
This change makes the macro usable outside of pybind11 namespace.

## Description

Change to `PYBIND11_OBJECT_CVT` macro to fully qualify `error_already_set` used in constructors. This change allows macro to be used outside of `pybind11` namespace.

## Suggested changelog entry:

```rst
PYBIND11_OBJECT_CVT and PYBIND11_OBJECT_CVT_DEFAULT macro can be used to define classes in namespaces other than pybind11.
```

### DIFF
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -35,12 +35,12 @@ namespace external {
 	}
     }
     class float_ : public py::object {
-	PYBIND11_OBJECT_CVT(float_, py::object, ::detail::check, ::detail::conv)
+	PYBIND11_OBJECT_CVT(float_, py::object, external::detail::check, external::detail::conv)
 
-	float_() : py::object(::detail::default_constructed(), borrowed_t{}) {}
+	float_() : py::object(external::detail::default_constructed(), borrowed_t{}) {}
 
 	double get_value() const { return PyFloat_AsDouble(this->ptr()); }
-    }
+    };
 }
 
 TEST_SUBMODULE(pytypes, m) {
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -33,7 +33,7 @@ PyObject *default_constructed() { return PyFloat_FromDouble(0.0); }
 class float_ : public py::object {
     PYBIND11_OBJECT_CVT(float_, py::object, external::detail::check, external::detail::conv)
 
-    float_() : py::object(external::detail::default_constructed(), borrowed_t{}) {}
+    float_() : py::object(external::detail::default_constructed(), stolen_t{}) {}
 
     double get_value() const { return PyFloat_AsDouble(this->ptr()); }
 };
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -16,16 +16,16 @@ namespace detail {
 bool check(PyObject *o) { return PyFloat_Check(o) != 0; }
 
 PyObject *conv(PyObject *o) {
+    PyObject *ret = nullptr;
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

rwgk:
How did you discover this?
Is there a chance you could add a test based on the situation you had? Otherwise it's likely that this will get broken again in refactoring work in the future.
Recently we had a similar PR, with added test: #3758

oleksandr-pavlyk:
I discovered it by building a toy pybind11 extensions `a.cpp` in https://github.com/oleksandr-pavlyk/cython_using_number_magic_methods.git 

To test the change I can try building a simple class outside of pybind11 namespace which wraps `PyFloatObject` inside. 
Let me try to whip something up and add it to this PR.

oleksandr-pavlyk:
Test has been added in `test/test_pytypes.cpp` and companion `.py` file. Please review again.

rwgk (approved):
Awesome, thanks!
Just waiting for the CI to finish.

rwgk:
valgrind is detecting a leak and clang-tidy has two complaints:

```
/__w/pybind11/pybind11/tests/test_pytypes.cpp:25:7: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
    } else {
      ^~~~~~
/__w/pybind11/pybind11/tests/test_pytypes.cpp:577:48: error: the parameter 'x' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    m.def("square_float_", [](external::float_ x) -> double {
                                               ^
```

I missed the leak when I looked at the test code before. Looking again.

oleksandr-pavlyk:
> valgrind is detecting a leak and clang-tidy has two complaints:
> 
> ```
> /__w/pybind11/pybind11/tests/test_pytypes.cpp:25:7: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
>     } else {
>       ^~~~~~
> /__w/pybind11/pybind11/tests/test_pytypes.cpp:577:48: error: the parameter 'x' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
>     m.def("square_float_", [](external::float_ x) -> double {
>                                                ^
> ```
> 
> I missed the leak when I looked at the test code before. Looking again.

I am about to push the change to fix this.

tests/test_pytypes.cpp
```diff
@@ -11,6 +11,34 @@
...
+        return nullptr;
+    }
+}
+
+PyObject *default_constructed() { return PyFloat_FromDouble(0.0); }
+} // namespace detail
+class float_ : public py::object {
+    PYBIND11_OBJECT_CVT(float_, py::object, external::detail::check, external::detail::conv)
+
+    float_() : py::object(external::detail::default_constructed(), borrowed_t{}) {}
```
rwgk, line 36: Is this the leak? I'm thinking it needs to be `stolen_t` (to take ownership).

("stolen" always sounds wrong to me btw)

oleksandr-pavlyk: Yes, right. Yes in numpy.h `ensure` is using `reinterpret_steal`. Another fix is on its way.

oleksandr-pavlyk:
@rwgk Thanks for your help. The CI is now green.

rwgk:
Thanks @oleksandr-pavlyk!

Issue #3776, opened 2022-03-04
[BUG]: make_iterator causes runtime error in second scoped_interpreter


jasjuang:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

This issue is similar to #2101, the reported minimal example problem in #2101 is fixed in the latest master because of #3744. However, the below minimal example will produce the error
```
1
2
3
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  RuntimeError: instance allocation failed: new instance has no pybind11-registered base types

At:
  <string>(4): <module>

Aborted (core dumped)
```
I can reproduce this problem in Ubuntu and macOS. 

The real use case is that each gtest `TEST` block will have its own `py::scoped_interpreter`, and this bug is preventing me from writing `for i in testclass:` more than once, but I would like to write `for i in testclass:` in multiple different `TEST` blocks. 

### Reproducible example code

```text
CMakeLists.txt

cmake_minimum_required(VERSION 3.20)
project(pybindsample)

set(pybind11_DIR "/home/jasjuang/install/share/cmake/pybind11")
find_package(pybind11 REQUIRED)

pybind11_add_module(${PROJECT_NAME} pybindsample.cpp)
target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_17)

add_executable(pybindsample_test pybindsample_test.cpp)
target_link_libraries(pybindsample_test pybind11::module pybind11::embed)
```
pybindsample.cpp
```
#include "pybind11/pybind11.h"
#include "pybind11/stl.h"

namespace py = pybind11;

class TestClass {
public:
  TestClass() = default;
  std::vector<int>::iterator begin() noexcept { return vec.begin(); }
  std::vector<int>::iterator end() noexcept { return vec.end(); }

private:
  std::vector<int> vec = {1, 2, 3};
};

PYBIND11_MODULE(pybindsample, m) {
  py::class_<TestClass>(m, "TestClass")
      .def(py::init<>())
      .def(
          "__iter__",
[...]

PR #3798, opened 2022-03-12
fix: Make non-embedded module clear their locals before deallocation

StarQTius:
## Description
This fix aims to solve the issue #3776. The problem was that every non-embedded module had their own locals, which could only be accessed through the `get_local_internals` function of the shared library they were in. So when these modules were deallocated, their locals were not cleared, and if the modules were initialized again, C++ types could not be registered in the locals.
I used the [PyModuleDef::m_free](https://docs.python.org/3/c-api/module.html#c.PyModuleDef.m_free) field to register a cleanup callback which clears the locals on non-embedded module deallocation. I added another function for creating embedded module, so that embedded modules do not clear the locals on deallocation.
I also registered a local class in the [external_module](https://github.com/pybind/pybind11/blob/master/tests/test_embed/external_module.cpp) used for testing. This new line was making the ["Restart the interpreter"](https://github.com/pybind/pybind11/blob/91a6e129d9274a0b10943044e5aeae08471e00fe/tests/test_embed/test_interpreter.cpp#L174) test case fail.

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -510,9 +510,17 @@ struct local_internals {
 };
 
 /// Works like `get_internals`, but for things which are locally registered.
+/// There is one variable storing the local internals for each module, but every embedded module
+/// shares the same locals.
 inline local_internals &get_local_internals() {
     static local_internals locals;
     return locals;
+}
+
+/// Clears the locally registered types and exception translators.
+inline void clear_local_internals() {
+    detail::get_local_internals().registered_types_cpp.clear();
+    detail::get_local_internals().registered_exception_translators.clear();
 }
 
 /// Constructs a std::string with the given arguments, stores it in `internals`, and returns its
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1232,13 +1232,45 @@ public:
                         /* m_slots */ nullptr,
                         /* m_traverse */ nullptr,
                         /* m_clear */ nullptr,
-                        /* m_free */ nullptr};
+                        /* m_free */ [](void *) { detail::clear_local_internals(); }};
         auto *m = PyModule_Create(def);
         if (m == nullptr) {
             if (PyErr_Occurred()) {
                 throw error_already_set();
             }
             pybind11_fail("Internal error in module_::create_extension_module()");
+        }
+        // TODO: Should be reinterpret_steal for Python 3, but Python also steals it again when
+        //       returned from PyInit_...
+        //       For Python 2, reinterpret_borrow was correct.
+        return reinterpret_borrow<module_>(m);
+    }
+
+    /** \rst
+        Create a new top-level embedded module.
+
+        ``def`` should point to a statically allocated module_def.
+    \endrst */
+    static module_
+    create_embedded_extension_module(const char *name, const char *doc, module_def *def) {
+        // module_def is PyModuleDef
+        // Placement new (not an allocation).
+        def = new (def)
+            PyModuleDef{/* m_base */ PyModuleDef_HEAD_INIT,
+                        /* m_name */ name,
+                        /* m_doc */ options::show_user_defined_docstrings() ? doc : nullptr,
+                        /* m_size */ -1,
+                        /* m_methods */ nullptr,
+                        /* m_slots */ nullptr,
+                        /* m_traverse */ nullptr,
+                        /* m_clear */ nullptr,
+                        /* m_free */ nullptr};
+        auto *m = PyModule_Create(def);
+        if (m == nullptr) {
+            if (PyErr_Occurred()) {
+                throw error_already_set();
+            }
+            pybind11_fail("Internal error in module_::create_embedded_extension_module()");
         }
         // TODO: Should be reinterpret_steal for Python 3, but Python also steals it again when
         //       returned from PyInit_...
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -42,7 +42,7 @@
     static ::pybind11::module_::module_def PYBIND11_CONCAT(pybind11_module_def_, name);           \
     static void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_ &);                     \
     static PyObject PYBIND11_CONCAT(*pybind11_init_wrapper_, name)() {                            \
-        auto m = ::pybind11::module_::create_extension_module(                                    \
+        auto m = ::pybind11::module_::create_embedded_extension_module(                           \
             PYBIND11_TOSTRING(name), nullptr, &PYBIND11_CONCAT(pybind11_module_def_, name));      \
         try {                                                                                     \
             PYBIND11_CONCAT(pybind11_init_, name)(m);                                             \
@@ -200,8 +200,7 @@ inline void finalize_interpreter() {
     }
     // Local internals contains data managed by the current interpreter, so we must clear them to
     // avoid undefined behaviors when initializing another interpreter
-    detail::get_local_internals().registered_types_cpp.clear();
-    detail::get_local_internals().registered_exception_translators.clear();
+    detail::clear_local_internals();
 
     Py_Finalize();
 
--- a/tests/test_embed/external_module.cpp
+++ b/tests/test_embed/external_module.cpp
@@ -13,7 +13,10 @@ PYBIND11_MODULE(external_module, m) {
         int v;
     };
 
+    class B {};
+
     py::class_<A>(m, "A").def(py::init<int>()).def_readwrite("value", &A::v);
+    py::class_<B>(m, "B", py::module_local());
 
     m.def("internals_at",
           []() { return reinterpret_cast<uintptr_t>(&py::detail::get_internals()); });


include/pybind11/pybind11.h
```diff
@@ -1245,6 +1245,38 @@ class module_ : public object {
...
         return reinterpret_borrow<module_>(m);
     }
+
+    /** \rst
+        Create a new top-level embedded module.
+
+        ``def`` should point to a statically allocated module_def.
+    \endrst */
+    static module_
+    create_embedded_extension_module(const char *name, const char *doc, module_def *def) {
```
Skylion007, line 1255: Hmm, this a lot of code duplication for changing the default value of one argument and the generic pybind11 fail, isn't it? If you think from API/ABI design they should remain seperate, maybe they can both call a common private method. Thoughts @henryiii @rwgk ?

Skylion007: Maybe this would actually be better to subclass?

rwgk: I agree with @Skylion007, thinking it will be easy to avoid the code duplication, maybe as simple as `bool is_embedded = false`? (same idea as Aaron's?)

My second concern is testing: what would it take to add a test that fails without the change here?

rwgk (changes requested):

include/pybind11/pybind11.h
Skylion007: This just feels like bad code smell, the MACRO we are using is already a bit hacky. Is there a way we can improve create_embedded_extension_module to automatically detect when it's being called from that macro? It seems like a rather specific and limited use case that the end user should never call directly so this method should really be hidden somewhere as an implementation detail.

StarQTius: Unless there is a reason to allow embedded and non-embedded modules in the same executable / library, maybe detecting whether `embed.h` has been included, or having the user indicate if they want the modules to be embedded with a `#define PYBIND11_EMBED` are options to consider ?

StarQTius: @rwgk About the testing: I added a line in `external_module` which was making the "Restart an interpreter" test case fail in a way that was consistent with the original issue (ie. having the library complain about an already registered type). I could have added another test but I didn't see the point since it would have done exactly the same thing and I needed to add that line in the external module in any case.

Skylion007: Okay, here is my proposal @StarQTius:
* Factor out the implementation of the static module function into to a function in a pybind11 `detail` namespace. You can have it use the is_embedded to denote the different behavior. Have the static function call the new detail one.
* Have the macro call the new detail function with the proper bool from the macro.

That way we don't create a new static method API that users can accidentally use or that we have to mantain and the API is the same as before with minimal code duplication. (The detail namespace will indicate this is not a new part of the API).

rwgk: > bout the testing: I added a line in `external_module.cpp` which was making the "Restart an interpreter" test case fail in a way that was consistent with the original issue (ie. having the library complain about an already registered type).

I see, thanks! I saw that line but didn't make the connection. Could you please add a comment, maybe one or two lines to explain in a nutshell, plus reference to this PR? So that it will be quick for anyone looking at the test code in the future, which no specific context, to make the connection.

StarQTius: As you asked, I moved the extension module creation implementation detail into a non-API function. I had to add forward declarations to make it works though. I also added comments to clarify how local internals are tested, and a reference to the PR.

rwgk (approved):
Looks awesome now, thanks!

include/pybind11/detail/internals.h
```diff
@@ -510,11 +510,19 @@ struct local_internals {
 };
 
 /// Works like `get_internals`, but for things which are locally registered.
+/// There is one variable storing the local internals for each module, but every embedded module
```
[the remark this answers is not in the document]
StarQTius: I'm sorry if it does not sound clear, English is not my native language. What I meant is that since the static local variable of `get_local_internals` does not get dynamically linked when the extension module is loaded by the interpreter, you end up with one definition of the locals for the interpreter, and one definition for each non-embedded modules.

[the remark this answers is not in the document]
rwgk: I think in that case my suggested wording is clearer. But I'm not a native speaker, too!
Let's see what @Skylion007 thinks.

Skylion007 (approved):

include/pybind11/detail/internals.h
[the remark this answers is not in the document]
rwgk: Sorry I'm having serious 2nd thoughts.

E.g. at Google, Python and all extension modules for a given binary (or test) are linked together statically. Each binary or test is hermetic.

Others may use `RTLD_GLOBAL`. Probably rare, but maybe in certain environment that's what's needed?

What does this PR mean in those situations? I'm not sure. **Does someone else have a clear understanding of how the `m_clear` functionality plays out across all environments?**

Guessing a lot: I think to be completely safe, we'd need some kind of lookup table, to keep track of which `module_def` pointers are associated with a given `local_internals` pointer, and if more than one, we cannot safely `clear_local_internals` until the last one disappears.

Or we could have `static std::unordered_map<module_def *, local_internals>` with all the required logic to register & clear types given a `module_def` pointer, but search for types over all keys of the map.

It seems pretty tricky to get the `m_clear` functionality correct in all environments.

Given those uncertainties, and that pybind11 existed to this point without using `m_clear`, maybe this should be an opt-in, e.g. via some `#define`, or runtime option? That would be a safer and less sudden way to find out what's possible and what not.

rwgk (changes requested):
Sorry after more thought I became too uncertain how this plays out. I someone else feels certain, please speak up.

rwgk:
Tagging @jbms: is there a chance that you could look at this PR (and my uncertainties under https://github.com/pybind/pybind11/pull/3798#discussion_r829737887)?

jbms:
I wasn't too clear on exactly when the `PyModuleDef.m_free` function is called, so I did a bit of experimentation.  My conclusion is that this change is NOT safe.

First let's consider just the single interpreter case:

In the most common case, you import some pybind11 modules, and don't do anything funny.  In that case, the `PyModuleDef.m_free` functions of all the modules will be called during interpreter finalization.  In that case it is probably fairly safe --- even if the locals are shared by multiple modules, clearing the locals multiple times doesn't really hurt anything, as long as none of the module free functions/destructors invoked by pybind11 attempt to use any pybind11 apis.

However, it appears that the `PyModuleDef.m_free` function can be called before finalization, if you import the extension, remove it from `sys.modules`, then reload it.  However, note that even though `m_free` is called, the `PyInit_modulename` function is not called again.  Here is a test case I put together:

mymodule.cc:

```cpp
#include <stdio.h>

namespace py = ::pybind11;

struct MyClass {
  std::string get_name() { return "myclass"; }
};

  m.doc() = "pybind11 example plugin";  // optional module docstring

  py::class_<MyClass>(m, "MyClass", py::module_local())
      .def(py::init([] { return MyClass{}; }))
      .def("get_name", &MyClass::get_name);

  fprintf(stderr, "Loading mymodule\n");
  fflush(stderr);
}
```

test.py:

```
import sys

import mymodule
cls = mymodule.MyClass()
assert cls.get_name() == "myclass"

del sys.modules['mymodule']
del mymodule
import mymodule
print('Reloaded module')
cls = mymodule.MyClass()
assert cls.get_name() == "myclass"
```

I also modified the `m_free` function to print a message.

The result with Python 3.9.9 is:

```
Loading mymodule
m_free called
Reloaded module
zsh: segmentation fault  python3 test.py
```

The crash happens in:
```
pybind11::class_<MyClass>::init_instance(pybind11::detail::instance*, void const*)
```

Presumably due to attempting to reference a type record that has been already deallocated.

This may be a bug in Python.

I don't really have any experience with starting and finalizing the Python interpreter multiple times in a single process, but from what I have seen when looking closely at the Python finalization process it is not really properly supported by Python --- some memory is sure to be leaked, especially if threads are used.  Therefore I wouldn't recommend doing that in production code, and even doing it in tests seems kind of questionable since why test something that doesn't match what happens in production?

It appears there is support for multi-phase initialization intended to better support module reloading, and also sub-interpreters:
https://docs.python.org/3/c-api/module.html#multi-phase-initialization

We might consider changing over to using that (which would probably require API changes in pybind11) ,
though my understanding is that there are still some Python APIs where the module state cannot easily be obtained, which would make it challenging to use this module state mechanism.  Additionally, I think it would require major API changes in pybind11 to support that.

I think the objective of this PR, fixing module_local registration when restarting the interpreter, could be solved by instead arranging to clear the local internals only when the interpreter is finalized, or after the interpreter is restarted, rather than using m_free.  I'm not sure what the best way to do that would be --- maybe adding an `atexit` handler would work, although that would add some overhead.  Alternatively we could attempt to detect if the interpreter has been restarted, and clear the locals only in that case.  I'm not sure how to do that exactly.

More generally, I think it would be best if module_local went away, and instead casters and exception translators could be overridden via explicitly passed-around context objects or something, so that there is no hidden semi-global state dependent on symbol visibility.

Skylion007:
> I think the objective of this PR, fixing module_local registration when restarting the interpreter, could be solved by instead arranging to clear the local internals only when the interpreter is finalized, or after the interpreter is restarted, rather than using m_free. I'm not sure what the best way to do that would be --- maybe adding an atexit handler would work, although that would add some overhead. Alternatively we could attempt to detect if the interpreter has been restarted, and clear the locals only in that case. I'm not sure how to do that exactly.

Good thing Python has an auditing event just for this: https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx

Skylion007:
Upon looking at it, atexit is a viable alternative, but I do dislike having to manage yet another global static state.

jbms:
You can run the Python executable itself in a debugger --- obviously works better if you have the debugging symbols, though.

What I mean, though, is that running Py_Initialize more than once is not something well-tested or well-supported by CPython, and there will be a lot of behavior differences compared to running Python normally.  For example, some state may stick around from a previous interpreter run, and therefore a test may pass even if the functionality does not actually work in normal Python usage, or a test may fail even though the functionality does work in normal Python usage.

Of course some tradeoffs have to be made to make testing more efficient, though.

jasjuang:
It seems problematic if there are a lot of behavior differences between Python and CPython. I think most users (at least me) would implicitly assume that CPython is a C version of Python, therefore their behavior should ideally be identical. If the consensus is they shouldn't be identical, is there a README somewhere that documents the behavior differences between Python and CPython that I can read?

jbms:
CPython is the main implementation of Python --- PyPy is another less widely used implementation.  I mentioned CPython because I have some familiarity with its implementation, while I have essentially no familiarity with PyPy.

rwgk:
General comment: we (mostly CPython) is sending people into traps by giving the impression that restarting the interpreter is safe, well-defined behavior. It simply isn't. It probably never will be. It would be best if people gave up trying. It's just a trap, ultimately, that's burning human time, like here, for no good reason.

If it was just me, I'd put a `static bool` into pybind11, that made it impossible to initialize the interpreter a second time, terminating the process with a message pointing here:

> **Bugs and caveats:** The destruction of modules and objects in modules is done in random order; this may cause destructors (`__del__()` methods) to fail when they depend on other objects (even functions) or modules. Dynamically loaded extension modules loaded by Python are not unloaded. Small amounts of memory allocated by the Python interpreter may not be freed (if you find a leak, please report it). Memory tied up in circular references between objects is not freed. Some memory allocated by extension modules may not be freed. Some extensions may not work properly if their initialization routine is called more than once; this can happen if an application calls [Py_Initialize()](https://docs.python.org/3/c-api/init.html#c.Py_Initialize) and [Py_FinalizeEx()](https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx) more than once.

A couple years ago I did some really simple experiments: Leaks!

https://github.com/rwgk/stuff/blob/71f5300061aaecb0f499aab77343abd4ca360c6c/noddy/embedded_noddy_main_run.c

PR #3805, opened 2022-03-16
changed localtime to gmtime to fix daylight savings issues

JamesUnicomb:
I was having issues with pybind using UTC times properly. As well, daylight savings datetimes were failing "roundtrip" tests. 

Linked to this discussion:
https://github.com/pybind/pybind11/discussions/3451#discussioncomment-2291585

### DIFF
--- a/include/pybind11/chrono.h
+++ b/include/pybind11/chrono.h
@@ -99,7 +99,7 @@ public:
     PYBIND11_TYPE_CASTER(type, const_name("datetime.timedelta"));
 };
 
-inline std::tm *localtime_thread_safe(const std::time_t *time, std::tm *buf) {
+inline std::tm *gmtime_thread_safe(const std::time_t *time, std::tm *buf) {
 #if (defined(__STDC_LIB_EXT1__) && defined(__STDC_WANT_LIB_EXT1__)) || defined(_MSC_VER)
     if (localtime_s(buf, time))
         return nullptr;
@@ -107,7 +107,7 @@ inline std::tm *localtime_thread_safe(const std::time_t *time, std::tm *buf) {
 #else
     static std::mutex mtx;
     std::lock_guard<std::mutex> lock(mtx);
-    std::tm *tm_ptr = std::localtime(time);
+    std::tm *tm_ptr = std::gmtime(time);
     if (tm_ptr != nullptr) {
         *buf = *tm_ptr;
     }
@@ -166,7 +166,7 @@ public:
             return false;
         }
 
-        value = time_point_cast<Duration>(system_clock::from_time_t(std::mktime(&cal)) + msecs);
+        value = time_point_cast<Duration>(system_clock::from_time_t(timegm(&cal)) + msecs);
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


include/pybind11/chrono.h
```diff
@@ -99,15 +99,15 @@ class duration_caster {
     PYBIND11_TYPE_CASTER(type, const_name("datetime.timedelta"));
 };
 
-inline std::tm *localtime_thread_safe(const std::time_t *time, std::tm *buf) {
+inline std::tm *gmtime_thread_safe(const std::time_t *time, std::tm *buf) {
 #if (defined(__STDC_LIB_EXT1__) && defined(__STDC_WANT_LIB_EXT1__)) || defined(_MSC_VER)
     if (localtime_s(buf, time))
```
rwgk, line 104: All the Windows tests are failing. Do you already have a plan for that? Could we simply remove the `#if` and use the same code, with `mutex`, everywhere?

include/pybind11/chrono.h
```diff
@@ -194,17 +194,17 @@ class type_caster<std::chrono::time_point<std::chrono::system_clock, Duration>>
...
+        if (!gmtime_ptr) {
+            throw cast_error("Unable to represent system_clock in GMT time");
         }
-        return PyDateTime_FromDateAndTime(localtime.tm_year + 1900,
-                                          localtime.tm_mon + 1,
-                                          localtime.tm_mday,
-                                          localtime.tm_hour,
-                                          localtime.tm_min,
-                                          localtime.tm_sec,
+        return PyDateTime_FromDateAndTime(gmtime.tm_year + 1900,
```
rwgk, line 202: Doesn't this change the behavior?
Before: local timezone
After: actually UTC but (incorrectly?) appearing as local timezone?

Same question from a different angle: as-is this PR fixes the roundtrip issue, i.e. fixes an internal consistency issue, but is it still externally inconsistent?

Do we have to tell `PyDateTime_FromDateAndTime()` somehow that the passed time is UTC?

I just looked a bit, seeing this warning 3 times under https://docs.python.org/3/library/datetime.html:

> Warning Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing a specific timestamp in UTC is by calling datetime.fromtimestamp(timestamp, tz=timezone.utc).

How can handle this correctly from the C API?

include/pybind11/chrono.h
Skylion007, line 104: This function needs to be swapped out too to stop the tests from failing

include/pybind11/chrono.h
Skylion007, line 202: @JamesUnicomb I am not sure about the UTC issue, but the datetime roundtrip issue can be easily resolved. We just need to use `PyDateTime_FromDateAndTimeAndFold` instead of `PyDateTime_FromDateAndTime` and make sure to properly set the `is_fold` argument.

Skylion007, line 202: @JamesUnicomb I am not sure about the UTC issue, but the datetime roundtrip issue can be easily resolved. We just need to use `PyDateTime_FromDateAndTimeAndFold` instead of `PyDateTime_FromDateAndTime` and make sure to properly set the `is_fold` argument.

include/pybind11/chrono.h
```diff
@@ -166,7 +166,7 @@ class type_caster<std::chrono::time_point<std::chrono::system_clock, Duration>>
             return false;
         }
 
-        value = time_point_cast<Duration>(system_clock::from_time_t(std::mktime(&cal)) + msecs);
+        value = time_point_cast<Duration>(system_clock::from_time_t(timegm(&cal)) + msecs);
```
CrustyAuklet, line 169: `timegm` isn't available on windows, so this line needs to change to something like:

```c++
#if defined(_MSC_VER)
    time_t tt = _mkgmtime(&cal);
#else
    time_t tt = timegm(&cal);
#endif

    value = time_point_cast<Duration>(system_clock::from_time_t(tt) + msecs);
```

rwgk: As @Skylion007 pointed out before, we don't have to go to UTC to solve the roundtrip problem: https://github.com/pybind/pybind11/pull/3805#discussion_r830116635
I think it's better to keep using the local time zone, but make proper use of the `fold` feature.

CrustyAuklet: with that solution would `std::system_clock` time points go into python as time aware (UTC) datetime objects, and python datetime objects go into C++ as UTC?

C++ system_clock is Unix Time (so UTC modulo leap seconds and such), so the time zone conversions are a little surprising.

Skylion007: I believe python datetime objects should be in local time in both? I am not sure we are actually attaching the time zone to the Python objects (we should but it doesn't seem straight forward using the raw C - API). 

If it's too tricky, we might want to call the Python datetime function directly if we need to.

voznesenskym:
I'm a noob here, so please, disregard readily and with prejudice if what I write does not jive with the spirit of the project.

It feels like localtime_thread_safe is a critical part of the contract for this cast, and that theres probably upstream customer assumptions around it, esp if its not a 1:1 conversion (as you said, daylight savings time can cause off-by-some-amount-of-time discrepancies). 

It feels, then, like the right thing to do here would be make this a perfectly backward compatible change, via pre-processor macro, something like:

PYBIND11_USE_LEGACY_LOCAL_TIME (where the default is now the gmtime) (my preference) 

or

PYBIND11_USE_GMTIME (where the default is the old behavior)

PR #3806, opened 2022-03-16
Expand string_caster support to uint8_t and std::byte

Guillaume227:
## Description

I would like to get easier handling of `std::basic_string` and `std::basic_string_view` for `uint8_t` and `std::byte` underlying types.
Those types are meant to represent raw data and thus should skip decoding / encoding steps, which the existing code already has support for using `string_caster::load_raw` and py::bytes.

Let me know if that looks like it belongs there, I still need to add test coverage before that goes in (have tested on my external use case).

## Suggested changelog entry:

Add support for conversion of std::string_view<uint8_t> and std::string_view<std::byte>.


Skylion007:
Hmm, okay I am not sure why the CI isn't running for this PR... Thoughts @henryiii @rwgk?

include/pybind11/cast.h
```diff
@@ -434,11 +438,19 @@ struct string_caster {
...
-        handle s = decode_utfN(buffer, nbytes);
-        if (!s) {
-            throw error_already_set();
+        if (std::is_same_v<CharT, uint8_t>
+#if defined(PYBIND11_CPP17)
+            or std::is_same_v<CharT, std::byte>
+#endif
+        ) {
+            return PyBytes_FromStringAndSize(buffer, nbytes);
+        } else {
```
Skylion007, line 447: else-after-return is bad style and clang-tidy will complain when the CI is back up. Just remove the else block since the return terminates the function anyway.

include/pybind11/cast.h
```diff
@@ -434,11 +438,19 @@ struct string_caster {
     cast(const StringType &src, return_value_policy /* policy */, handle /* parent */) {
         const char *buffer = reinterpret_cast<const char *>(src.data());
         auto nbytes = ssize_t(src.size() * sizeof(CharT));
-        handle s = decode_utfN(buffer, nbytes);
-        if (!s) {
-            throw error_already_set();
+        if (std::is_same_v<CharT, uint8_t>
```
Skylion007, line 441: `is_same_v` is not introduced until C++17 and we don't have this function available in the backports in the common.h yet.

include/pybind11/cast.h
```diff
@@ -434,11 +438,19 @@ struct string_caster {
...
+#if defined(PYBIND11_CPP17)
+            or std::is_same_v<CharT, std::byte>
```
Skylion007, line 443: You also may need to wrap this in special MACRO to prevent it MSVC from complaining.

include/pybind11/cast.h
```diff
@@ -461,20 +473,26 @@ struct string_caster {
...
+    bool load_raw(enable_if_t<accept_bytes_as_is<C>::value, handle> src) {
         if (PYBIND11_BYTES_CHECK(src.ptr())) {
             // We were passed raw bytes; accept it into a std::string or char*
             // without any encoding attempt.
             const char *bytes = PYBIND11_BYTES_AS_STRING(src.ptr());
             if (!bytes) {
                 pybind11_fail("Unexpected PYBIND11_BYTES_AS_STRING() failure.");
             }
-            value = StringType(bytes, (size_t) PYBIND11_BYTES_SIZE(src.ptr()));
+            value = StringType((const C *) bytes, (size_t) PYBIND11_BYTES_SIZE(src.ptr()));
```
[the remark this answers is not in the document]
Guillaume227: I want with a C-style cast in keeping with the surrounding usage style (see `(size_t)` cast next to it). I can definitely use a `static_cast`, that's what I would have favored personally.

include/pybind11/cast.h
Guillaume227: regarding that flag, what's the threshold for introducing a new PYBIND11_xxx flag (such as `PYBIND11_HAS_U8STRING`, `PYBIND11_HAS_STRING_VIEW`) to hide a check for presence of `std::byte`?
I went for that simple check since std::byte is introduced with c++17 but you never know what c++ partial-support compiler versions their could be out there.

include/pybind11/cast.h
```diff
@@ -434,11 +438,19 @@ struct string_caster {
     cast(const StringType &src, return_value_policy /* policy */, handle /* parent */) {
         const char *buffer = reinterpret_cast<const char *>(src.data());
         auto nbytes = ssize_t(src.size() * sizeof(CharT));
-        handle s = decode_utfN(buffer, nbytes);
-        if (!s) {
-            throw error_already_set();
+        if (std::is_same<CharT, uint8_t>::value
+#if defined(PYBIND11_CPP17)
+            || std::is_same<CharT, std::byte>::value
+#endif
```
jagerman, line 444: I'd prefer if the logic of what is a binary type was defined in one place, perhaps via splitting up the `is_std_char_type` into this:
```C++
template <typename CharT>
using is_std_binary_char_type = any_of<
#if defined(PYBIND11_CPP17)
                                       std::is_same<CharT, std::byte>,
#endif
                                       std::is_same<CharT, uint8_t>
                                       >;

template <typename CharT>
using is_std_char_type = any_of<std::is_same<CharT, char>, /* std::string */
#if defined(PYBIND11_HAS_U8STRING)
                                std::is_same<CharT, char8_t>, /* std::u8string */
#endif
                                std::is_same<CharT, char16_t>, /* std::u16string */
                                std::is_same<CharT, char32_t>, /* std::u32string */
                                std::is_same<CharT, wchar_t>,  /* std::wstring */
                                is_std_binary_char_type<CharT>
                                >;
```
and then here you can simplify to just:
```C++
        if (is_std_binary_char_type<CharT>::value)
            return PyBytes_FromStringAndSize(buffer, nbytes);
```

jagerman: Another random suggestion (entirely optional, some people might hate it, it just occurred to me as something that might clean up a bit):
```C++
// in detail/common.h:
template <typename T, typename... Ts>
using is_any_same = any_of<std::is_same<T, Ts>...>;
```
and then the `any_of<std::is_same<CharT, A>, std::is_same<CharT, B>, ...>` can become `is_any_same<CharT, A, B, ...>`.

I'd be surprised if there aren't a few other places in the code base that could be simplified using it.

Issue #3801, opened 2022-03-15
[BUG]: __getattr__ and multiple inheritance with smart holders


petersteneteg:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

Hi, I've found an other issue related to #3788 
this is in regard to the smart holder branch

The example code below fails with infinite recursive getattr calls. 

removing either `Base2`, `Derived2` or the binding of `__getattr__` removes the problem. 

removing the following lines 
https://github.com/pybind/pybind11/blob/9d4b4dffce6677a7c9a9af37a45f4436d353f5e1/include/pybind11/detail/smart_holder_type_casters.h#L295-L297
also removes the issue.  

### Reproducible example code

```text
// Binding code

#include <pybind11/pybind11.h>

namespace py = pybind11;

struct Base1 {int a1{};};
struct Base2 {int a2{};};

struct Base12 : Base1, Base2 {
    virtual ~Base12() = default;
    int foo() const { return 0; }
};

struct Derived1 : Base12 {
    int bar() const { return 1; }
};

struct Derived2 : Base12 {
    int bar() const { return 2; }
};


PYBIND11_MODULE(foo, m) {
    py::class_<Base1>(m, "Base1");
    py::class_<Base2>(m, "Base2");

    py::class_<Base12, Base1, Base2>(m, "Base12")
        .def(py::init<>())
        .def("foo", &Base12::foo)
       .def("__getattr__", [](Base12&, std::string key) {
            return "Base GetAttr: " + key;
        });

    py::class_<Derived1, Base12>(m, "Derived1")
        .def(py::init<>())
        .def("bar", &Derived1::bar);

    py::class_<Derived2, Base12>(m, "Derived2")
        .def(py::init<>())
        .def("bar", &Derived2::bar);
}


// Test code
import foo

d1 = foo.Derived1()
print(f"d1.foo(): {d1.foo()}")
print(f"d1.bar(): {d1.bar()}")
print(f"d1.prop1: {d1.prop1}")

d2 = foo.Derived2()
[...]

PR #3807, opened 2022-03-16
Fixes issue #3801

wangxf123456:
## Description

Fixes https://github.com/pybind/pybind11/issues/3801 by only calling `try_as_void_ptr_capsule` when the Python object is not wrapped with pybind11.

Also modified the test cases to make them work with the new implementation.

### DIFF
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -292,9 +292,11 @@ public:
             loaded_v_h = value_and_holder();
             return true;
         }
-        const auto &bases = all_type_info(srctype);
-        if (convert && cpptype && bases.empty() && try_as_void_ptr_capsule(src)) {
-            return true;
+        if (convert && cpptype) {
+            const auto &bases = all_type_info(srctype);
+            if (bases.empty() && try_as_void_ptr_capsule(src)) {
+                return true;
+            }
         }
         return false;
     }
--- a/include/pybind11/detail/smart_holder_type_casters.h
+++ b/include/pybind11/detail/smart_holder_type_casters.h
@@ -292,7 +292,8 @@ public:
             loaded_v_h = value_and_holder();
             return true;
         }
-        if (convert && cpptype && try_as_void_ptr_capsule(src)) {
+        const auto &bases = all_type_info(srctype);
+        if (convert && cpptype && bases.empty() && try_as_void_ptr_capsule(src)) {
             return true;
         }
         return false;
--- a/tests/test_class_sh_void_ptr_capsule.cpp
+++ b/tests/test_class_sh_void_ptr_capsule.cpp
@@ -94,7 +94,8 @@ TEST_SUBMODULE(class_sh_void_ptr_capsule, m) {
     py::classh<Base12, Base1, Base2>(m, "Base12")
         .def(py::init<>())
         .def("foo", &Base12::foo)
-        .def("__getattr__", [](Base12 &, const std::string &key) { return "Base GetAttr: " + key; });
+        .def("__getattr__",
+             [](Base12 &, const std::string &key) { return "Base GetAttr: " + key; });
 
     py::classh<Derived1, Base12>(m, "Derived1").def(py::init<>()).def("bar", &Derived1::bar);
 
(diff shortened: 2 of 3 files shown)


include/pybind11/detail/smart_holder_type_casters.h
```diff
@@ -292,7 +292,8 @@ class modified_type_caster_generic_load_impl {
             loaded_v_h = value_and_holder();
             return true;
         }
-        if (convert && cpptype && try_as_void_ptr_capsule(src)) {
+        const auto &bases = all_type_info(srctype);
+        if (convert && cpptype && bases.empty() && try_as_void_ptr_capsule(src)) {
```
rwgk, line 296: The (possibly expensive) `all_type_info` call is needed only if `convert && cpptype` is true:
```
if (convert && cpptype) {
  const auto &bases = all_type_info(srctype);
  if (bases.empty() && try_as_void_ptr_capsule(src)) {
    return true;
  }
}
```
But even that is still doing too much work.
After everything is working (CI is green), but before marking this PR as ready for review, I'd look into something like `has_type_info(src_type)` which returns immediately when it find the first one.
Not sure if that actually makes sense, but I'd definitely look to find out.
Another idea: have a const pointer for bases right under the `Case 2` comment, if set already inside `Case 2`, reuse here.
But that one may be a completely inconsequential optimization, although also an easy one, not sure if it's worth it or not.

tests/test_class_sh_void_ptr_capsule.py
```diff
@@ -3,26 +3,79 @@
 from pybind11_tests import class_sh_void_ptr_capsule as m
 
 
+class Valid:
+
+    capsule_generated = False
```
rwgk, line 8: This looks dangerous.
```
def __init__(self):
  self.capsule_generated = False
```
maybe?
Even if the test only calls it once at the moment, I'd add that `__init__`. Otherwise someone could trip over this when working or experimenting with this test later.

include/pybind11/detail/smart_holder_type_casters.h
wangxf123456: There is `get_type_info`: https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/type_caster_base.h#L184. But `all_type_info` is called inside. I did not find anything else that is related to this right now. But I will spend more time on this direction. 

For `all_type_info`, looks like it will cache the result: https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/type_caster_base.h#L167. So I believe if it reached `Case 2`, the cached result will be returned.

tests/test_class_sh_void_ptr_capsule.cpp
```diff
@@ -7,122 +7,96 @@
...
-struct NoConversion : public HelperBase {
-    int get() const override { return 102; }
-};
+struct AsAnotherObject {};
 
-struct NoCapsuleReturned : public HelperBase {
-    int get() const override { return 103; }
+// Create a void pointer capsule for test. The encapsulated object does not
+// matter for this test case.
+PyObject *create_test_void_ptr_capsule() { return PyCapsule_New((void *) 12345, "int", nullptr); }
```
rwgk, line 20: While this is evidently fine on all platforms today, I'm afraid casting an arbitrary integer to a pointer could trigger some sanitizer error in the future. Ideas:
```
// Deemed safe not to incref. None surely outlives this capsule.
static_cast<void *>(Py_None)
```
or
```
static int just_to_have_something_to_point_to = 0;
static_cast<void *>(&just_to_have_something_to_point_to)
```

tests/test_class_sh_void_ptr_capsule.py
```diff
@@ -3,26 +3,73 @@
...
+
+
+class NoCapsuleReturned:
+    def __init__(self):
+        self.capsule_generated = False
+
+    def as_pybind11_tests_class_sh_void_ptr_capsule_NoCapsuleReturned(  # noqa: N802
+        self,
+    ):
+        return
```
rwgk, line 27:
```
pass
```
or
```
return None
```
would seem either most idiomatic or most explicit.

rwgk (approved):

rwgk:
Google global testing with this PR passed (2022-03-18 05:00 batch, test token OCL:435561515:BASE:435643811:1647617193745:53ea324).

PR #3811, opened 2022-03-18
Use rvalue reference for std::variant cast_op<T>

laramiel:
Nearly every call site of `cast_op<T>` uses an r-value reference.

Except in stl.h `variant_caster::load_alternative` which handles `std::variant`.

Fix that.

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -372,7 +372,7 @@ struct variant_caster<V<Ts...>> {
     bool load_alternative(handle src, bool convert, type_list<U, Us...>) {
         auto caster = make_caster<U>();
         if (caster.load(src, convert)) {
-            value = cast_op<U>(caster);
+            value = cast_op<U>(std::move(caster));
             return true;
         }
         return load_alternative(src, convert, type_list<Us...>{});


laramiel:
NOTE: I have not yet run extensive tests on this.

rwgk (approved):
The CI is happy, what could possibly go wrong! :-)
I'll try to run our global testing tonight with this included.

laramiel:
It should be safe, the`caster` is essentially discarded on the next line.  It may change a few call sites from a copy to a move, which is more consistent with basically all other cast operations. So it might improve casting of things like  `std::variant<std::unique_ptr<X>>`.

laramiel:
Also, it's possible that this overload of `cast_op` isn't a good idea: (cast.h#58)

```
typename make_caster<T>::template cast_op_type<T> cast_op(make_caster<T> &caster) {
    return caster.operator typename make_caster<T>::template cast_op_type<T>();
}
```

rwgk:
Google global testing with this PR passed (2022-03-18 05:00 batch, test token OCL:435561515:BASE:435643811:1647617193745:53ea324).

rwgk:
I'll go ahead merging this 1-line change, based on the extensive testing, to get it into the smart_holder update I'm working on right now.

PR #3823, opened 2022-03-23
(perf): Add missing move in sp matrix caster and microopt char concats

Skylion007:
## Description
* Noticed some missing moves in the Eigen.h casters
* Fixed some micro-opts that were bugging me here and there (single char length strings that should have been chars).


rwgk (approved):
Looks good. The MSVC 2022 C++20 failure was in the setup phase, safe to ignore I think.

PR #3825, opened 2022-03-24
fix: better exception and error handling for capsules

Skylion007:
## Description
* This solves several issues when dealing with capsules. Firstly, we were leaving the pybind11 in a bad exception state when we throw pybind11_fail without clearing the error first. We really should be throwing error_already_set to ensure that exception messages are as consistent and informative as vanilla cpython.
* There really isn't a good reason to be catching these errors so the changes should be pretty backwards compatible. In fact, a lot of the invalid error states occurred because our test suite never hit them in production. 
* This guards against some nasty segfaults that could occur when mishandling capsules.
* Many of the python capsule calls only return null on error states which makes exception handling trivial. There was only one case where I had to add an additional pybind11_fail as a catch.
* Finally while browsing the repo, I found one more missing std::move and added that in. 

## Suggested changelog entry:

```rst
* The bindings for capsules now have more consistent exception handling.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1288,7 +1288,8 @@ public:
     static pybind11::dtype dtype() {
         list shape;
         array_info<T>::append_extents(shape);
-        return pybind11::dtype::from_args(pybind11::make_tuple(base_descr::dtype(), shape));
+        return pybind11::dtype::from_args(
+            pybind11::make_tuple(base_descr::dtype(), std::move(shape)));
     }
 };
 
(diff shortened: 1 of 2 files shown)


rwgk (approved):
Thanks a lot for this cleanup, the revised code makes a lot more sense!
I'll try to run this through our global testing tonight, but if you need this before I have the results, please go ahead.

rwgk:
> Suggested changelogentry:
> PyCapsules standing bindings now have more consistent exception handling.

standing?

standard?

Maybe: `py::capsule` now has more consistent exception handling.

?

rwgk:
The test ID for the global testing is `OCL:437135289:BASE:437101826:1648173046939:234ba7aa`, 21:00 batch.

rwgk:
The global testing finished without finding any issues.

Skylion007:
@rwgk Fixed spelling errors in the PR description.

henryiii (approved):
Looks fantastic to me. :)

PR #2903, opened 2021-03-15
Raise codec errors when casting to std::string

sloretz:
## Description

This allows [the codec's exception](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8String) to be raised instead of `RuntimeError` when casting from `py::str` to `std::string`. I ran into this when porting a CPython extention to pybind11 with an existing test that expected a `UnicodeEncodeError` when given a string with [UCS surrogate characters](https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates).

## Suggested changelog entry:

```rst
Allow the codec's exception to be raised instead of :code:`RuntimeError` when casting from :code:`py::str` to :code:`std::string`.
```

PR #3826, opened 2022-03-25
fix: improve str exceptions and consistency with python

Skylion007:
## Description
* Use error_already_set() to provide more informative exceptions and better alignment with CPython exception strings
* Replaces the related compatibility macro. 
* Raise CodecError instead of TypeError in the byte constructor to be consistent with the fix in #2903 

## Suggested changelog entry:

```rst
* Improve exception handling in python str bindings
```


jagerman:
> [Revert macro change just in case](https://github.com/pybind/pybind11/pull/3826/commits/3e38dd8c615d8f49f8c8b78217b262dd1dfbc87e)

Why keep the macro?

henryiii:
I see at least a few users:

https://cs.github.com/?scopeName=All+repos&scope=&q=PYBIND11_BYTES_AS_STRING_AND_SIZE

Tencent/MMKV: POSIX/Python/libmmkv_python.cpp‎ for example.

henryiii (approved):

PR #3827, opened 2022-03-25
[v2.9] chore: backport a few more commits to 2.9

henryiii:
Fixing the removal of MSVC will be fun. Probably will just ignore it and hope for the best.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -687,7 +687,7 @@ jobs:
 
 
   # This verifies that the documentation is not horribly broken, and does a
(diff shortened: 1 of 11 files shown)


henryiii:
Wow, even fully pinned dependencies break with Jinja. Weird. ~~We can bump to sphinx4, which will fix the problem~~ No, Python too old., or add a pin for Jinja (I think).

henryiii:
I don't think there's anything critical here, so we could also just release what's there now as 2.9.x, and close this (it will be released in 2.10 anyway).

PR #3838, opened 2022-03-30
[smart_holder] Add a new return value policy `return_as_bytes`

wangxf123456:
## Description
Add a new return value policy `return_as_bytes` to make C++ functions return `bytes` to Python instead of `str`.

We can convert return values to `bytes` by applying `py::bytes`, but this might be hard when dealing with nested types.

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -461,8 +461,7 @@ enum class return_value_policy : uint8_t {
         return_value_policy::reference and the keep_alive call policy */
     reference_internal,
 
-    /** This policy is experimental and likely to change in the future. With
-        this policy, C++ string types are converted to Python bytes,
+    /** With this policy, C++ string types are converted to Python bytes,
         instead of str. This is most useful when a C++ function returns a
         container-like type with nested C++ string types, and py::bytes cannot
         be applied easily. Dictionary like types might not work, for example,
@@ -470,7 +469,8 @@ enum class return_value_policy : uint8_t {
         to be converted to bytes. Note that this return_value_policy is not
         concerned with lifetime/ownership semantics, like the other policies,
         but the purpose of return_as_bytes is certain to be orthogonal, because
-        C++ strings are always copied to Python bytes or str. */
+        C++ strings are always copied to Python `bytes` or `str`.
+        NOTE: This policy is NOT available on master. */
     return_as_bytes
 };
 
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -463,15 +463,15 @@ enum class return_value_policy : uint8_t {
 
     /** With this policy, C++ string types are converted to Python bytes,
         instead of str. This is most useful when a C++ function returns a
-        container-like type with nested C++ string types, and py::bytes cannot
+        container-like type with nested C++ string types, and `py::bytes` cannot
         be applied easily. Dictionary like types might not work, for example,
         `Dict[str, bytes]`, because this policy forces all string return values
         to be converted to bytes. Note that this return_value_policy is not
         concerned with lifetime/ownership semantics, like the other policies,
-        but the purpose of return_as_bytes is certain to be orthogonal, because
+        but the purpose of _return_as_bytes is certain to be orthogonal, because
         C++ strings are always copied to Python `bytes` or `str`.
         NOTE: This policy is NOT available on master. */
-    return_as_bytes
+    _return_as_bytes
 };
 
 PYBIND11_NAMESPACE_BEGIN(detail)
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -461,13 +461,16 @@ enum class return_value_policy : uint8_t {
         return_value_policy::reference and the keep_alive call policy */
     reference_internal,
 
-    /** With this policy, C++ string types are converted to Python bytes,
+    /** This policy is experimental and likely to change in the future. With
+        this policy, C++ string types are converted to Python bytes,
         instead of str. This is most useful when a C++ function returns a
         container-like type with nested C++ string types, and py::bytes cannot
-        be applied easily. Note that this return_value_policy is not concerned
-        with lifetime/ownership semantics, like the other policies, but the
-        purpose of return_as_bytes is certain to be orthogonal, because C++
-        strings are always copied to Python bytes or str. */
+        be applied easily. Dictionary like types might not work, for example,
+        `Dict[str, bytes]`, because this policy forces all string return values
+        to be converted to bytes. Note that this return_value_policy is not
+        concerned with lifetime/ownership semantics, like the other policies,
+        but the purpose of return_as_bytes is certain to be orthogonal, because
+        C++ strings are always copied to Python bytes or str. */
     return_as_bytes
 };
 
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -461,11 +461,13 @@ enum class return_value_policy : uint8_t {
         return_value_policy::reference and the keep_alive call policy */
     reference_internal,
 
-    /** Use this policy to make C++ functions return bytes to Python instead of
-        str. This is useful when C++ function returns nested type of string,
-        where we cannot apply `py::bytes` easily. Note that when C++ function
-        returns string, we don't have to deal ownership for sure, because we
-        are always copying the C++ string to a new Python str/bytes object.*/
+    /** With this policy, C++ string types are converted to Python bytes,
+        instead of str. This is most useful when a C++ function returns a
+        container-like type with nested C++ string types, and py::bytes cannot
+        be applied easily. Note that this return_value_policy is not concerned
+        with lifetime/ownership semantics, like the other policies, but the
+        purpose of return_as_bytes is certain to be orthogonal, because C++
+        strings are always copied to Python bytes or str. */
     return_as_bytes
 };
 
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -459,7 +459,14 @@ enum class return_value_policy : uint8_t {
         collected while Python is still using the child. More advanced
         variations of this scheme are also possible using combinations of
         return_value_policy::reference and the keep_alive call policy */
-    reference_internal
+    reference_internal,
+
+    /** Use this policy to make C++ functions return bytes to Python instead of
+        str. This is useful when C++ function returns nested type of string,
+        where we cannot apply `py::bytes` easily. Note that when C++ function
+        returns string, we don't have to deal ownership for sure, because we
+        are always copying the C++ string to a new Python str/bytes object.*/
+    return_as_bytes
 };
 
 PYBIND11_NAMESPACE_BEGIN(detail)
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -386,27 +386,27 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("takes_const_ref_wrap",
           [](std::reference_wrapper<const ConstRefCasted> x) { return x.get().tag; });
 
-    // test return_value_policy::return_as_bytes
+    // test return_value_policy::_return_as_bytes
     m.def(
         "invalid_utf8_string_as_bytes",
         []() { return std::string("\xba\xd0\xba\xd0"); },
-        py::return_value_policy::return_as_bytes);
+        py::return_value_policy::_return_as_bytes);
     m.def("invalid_utf8_string_as_str", []() { return std::string("\xba\xd0\xba\xd0"); });
     m.def(
         "invalid_utf8_char_array_as_bytes",
         []() { return "\xba\xd0\xba\xd0"; },
-        py::return_value_policy::return_as_bytes);
+        py::return_value_policy::_return_as_bytes);
     py::class_<StringAttr>(m, "StringAttr")
         .def(py::init<std::string>())
         .def_property(
             "value",
             py::cpp_function([](StringAttr &self) { return self.value; },
-                             py::return_value_policy::return_as_bytes),
+                             py::return_value_policy::_return_as_bytes),
             py::cpp_function([](StringAttr &self, std::string v) { self.value = std::move(v); }));
 #ifdef PYBIND11_HAS_STRING_VIEW
     m.def(
         "invalid_utf8_string_view_as_bytes",
         []() { return std::string_view("\xba\xd0\xba\xd0"); },
-        py::return_value_policy::return_as_bytes);
+        py::return_value_policy::_return_as_bytes);
 #endif
 }
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -380,18 +380,19 @@ TEST_SUBMODULE(builtin_casters, m) {
           [](std::reference_wrapper<const ConstRefCasted> x) { return x.get().tag; });
 
     // test return_value_policy::return_as_bytes
-    m.def("invalid_utf8_string_as_bytes", []() {
-        return std::string("\xba\xd0\xba\xd0");
-    }, py::return_value_policy::return_as_bytes);
-    m.def("invalid_utf8_string_as_str", []() {
-        return std::string("\xba\xd0\xba\xd0");
-    });
-    m.def("invalid_utf8_char_array_as_bytes", []() {
-        return "\xba\xd0\xba\xd0";
-    }, py::return_value_policy::return_as_bytes);
+    m.def(
+        "invalid_utf8_string_as_bytes",
+        []() { return std::string("\xba\xd0\xba\xd0"); },
+        py::return_value_policy::return_as_bytes);
+    m.def("invalid_utf8_string_as_str", []() { return std::string("\xba\xd0\xba\xd0"); });
+    m.def(
+        "invalid_utf8_char_array_as_bytes",
+        []() { return "\xba\xd0\xba\xd0"; },
+        py::return_value_policy::return_as_bytes);
 #ifdef PYBIND11_HAS_STRING_VIEW
-    m.def("invalid_utf8_string_view_as_bytes", []() {
-        return std::string_view("\xba\xd0\xba\xd0");
-    }, py::return_value_policy::return_as_bytes);
+    m.def(
+        "invalid_utf8_string_view_as_bytes",
+        []() { return std::string_view("\xba\xd0\xba\xd0"); },
+        py::return_value_policy::return_as_bytes);
 #endif
 }
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -378,4 +378,20 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("takes_const_ref", [](const ConstRefCasted &x) { return x.tag; });
     m.def("takes_const_ref_wrap",
           [](std::reference_wrapper<const ConstRefCasted> x) { return x.get().tag; });
+
+    // test return_value_policy::return_as_bytes
+    m.def("invalid_utf8_string_as_bytes", []() {
+        return std::string("\xba\xd0\xba\xd0");
+    }, py::return_value_policy::return_as_bytes);
+    m.def("invalid_utf8_string_as_str", []() {
+        return std::string("\xba\xd0\xba\xd0");
+    });
+    m.def("invalid_utf8_char_array_as_bytes", []() {
+        return "\xba\xd0\xba\xd0";
+    }, py::return_value_policy::return_as_bytes);
+#ifdef PYBIND11_HAS_STRING_VIEW
+    m.def("invalid_utf8_string_view_as_bytes", []() {
+        return std::string_view("\xba\xd0\xba\xd0");
+    }, py::return_value_policy::return_as_bytes);
+#endif
 }
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -11,12 +11,14 @@
 
 #include "pybind11_tests.h"
 
+#include <utility>
+
 struct ConstRefCasted {
     int tag;
 };
 
 struct StringAttr {
-    StringAttr(std::string v) : value(v) {}
+    explicit StringAttr(std::string v) : value(std::move(v)) {}
     std::string value;
 };
 
@@ -396,10 +398,11 @@ TEST_SUBMODULE(builtin_casters, m) {
         py::return_value_policy::return_as_bytes);
     py::class_<StringAttr>(m, "StringAttr")
         .def(py::init<std::string>())
-        .def_property("value",
-                      py::cpp_function([](StringAttr &self) { return self.value; },
-                                       py::return_value_policy::return_as_bytes),
-                      py::cpp_function([](StringAttr &self, std::string v) { self.value = v; }));
+        .def_property(
+            "value",
+            py::cpp_function([](StringAttr &self) { return self.value; },
+                             py::return_value_policy::return_as_bytes),
+            py::cpp_function([](StringAttr &self, std::string v) { self.value = std::move(v); }));
 #ifdef PYBIND11_HAS_STRING_VIEW
     m.def(
         "invalid_utf8_string_view_as_bytes",
--- a/tests/test_builtin_casters.cpp
+++ b/tests/test_builtin_casters.cpp
@@ -15,6 +15,11 @@ struct ConstRefCasted {
     int tag;
 };
 
+struct StringAttr {
+    StringAttr(std::string v) : value(v) {}
+    std::string value;
+};
+
 PYBIND11_NAMESPACE_BEGIN(pybind11)
 PYBIND11_NAMESPACE_BEGIN(detail)
 template <>
@@ -389,6 +394,12 @@ TEST_SUBMODULE(builtin_casters, m) {
         "invalid_utf8_char_array_as_bytes",
         []() { return "\xba\xd0\xba\xd0"; },
         py::return_value_policy::return_as_bytes);
+    py::class_<StringAttr>(m, "StringAttr")
+        .def(py::init<std::string>())
+        .def_property("value",
+                      py::cpp_function([](StringAttr &self) { return self.value; },
+                                       py::return_value_policy::return_as_bytes),
+                      py::cpp_function([](StringAttr &self, std::string v) { self.value = v; }));
 #ifdef PYBIND11_HAS_STRING_VIEW
     m.def(
         "invalid_utf8_string_view_as_bytes",
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -559,7 +559,7 @@ TEST_SUBMODULE(stl, m) {
 #ifdef PYBIND11_TEST_VARIANT
     m.def(
         "invalid_utf8_variant_string_as_bytes",
-        []() { return std::variant<std::string, std::monostate>{"\xba\xd0\xba\xd0"}; },
+        []() { return variant<std::string, int>{"\xba\xd0\xba\xd0"}; },
         py::return_value_policy::return_as_bytes);
 #endif
 }
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -545,10 +545,10 @@ TEST_SUBMODULE(stl, m) {
     // test return_value_policy::return_as_bytes
     m.def(
         "invalid_utf8_string_array_as_bytes",
-        []() { return std::array<std::string, 1>{"\xba\xd0\xba\xd0"}; },
+        []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}} },
         py::return_value_policy::return_as_bytes);
     m.def("invalid_utf8_string_array_as_str",
-          []() { return std::array<std::string, 1>{std::string("\xba\xd0\xba\xd0")}; });
+          []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; });
 #ifdef PYBIND11_HAS_OPTIONAL
     m.def(
         "invalid_utf8_optional_string_as_bytes",
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -545,7 +545,11 @@ TEST_SUBMODULE(stl, m) {
     // test return_value_policy::return_as_bytes
     m.def(
         "invalid_utf8_string_array_as_bytes",
-        []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}} },
+        []() {
+            return std::array<std::string, 1> {
+                { "\xba\xd0\xba\xd0" }
+            }
+        },
         py::return_value_policy::return_as_bytes);
     m.def("invalid_utf8_string_array_as_str",
           []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; });
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -541,4 +541,23 @@ TEST_SUBMODULE(stl, m) {
         []() { return new std::vector<bool>(4513); },
         // Without explicitly specifying `take_ownership`, this function leaks.
         py::return_value_policy::take_ownership);
+
+    // test return_value_policy::return_as_bytes
+    m.def("invalid_utf8_string_array_as_bytes", []() {
+        return std::array<std::string, 1>{"\xba\xd0\xba\xd0"};
+    }, py::return_value_policy::return_as_bytes);
+    m.def("invalid_utf8_string_array_as_str", []() {
+        return std::array<std::string, 1>{std::string("\xba\xd0\xba\xd0")};
+    });
+#ifdef PYBIND11_HAS_OPTIONAL
+    m.def("invalid_utf8_optional_string_as_bytes", []() {
+        return std::optional<std::string>{"\xba\xd0\xba\xd0"};
+    }, py::return_value_policy::return_as_bytes);
+#endif
+
+#ifdef PYBIND11_TEST_VARIANT
+    m.def("invalid_utf8_variant_string_as_bytes", []() {
+        return std::variant<std::string, std::monostate>{"\xba\xd0\xba\xd0"};
+    }, py::return_value_policy::return_as_bytes);
+#endif
 }
--- a/tests/test_stl.cpp
+++ b/tests/test_stl.cpp
@@ -542,24 +542,24 @@ TEST_SUBMODULE(stl, m) {
         // Without explicitly specifying `take_ownership`, this function leaks.
         py::return_value_policy::take_ownership);
 
-    // test return_value_policy::return_as_bytes
+    // test return_value_policy::_return_as_bytes
     m.def(
         "invalid_utf8_string_array_as_bytes",
         []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; },
-        py::return_value_policy::return_as_bytes);
+        py::return_value_policy::_return_as_bytes);
     m.def("invalid_utf8_string_array_as_str",
           []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; });
 #ifdef PYBIND11_HAS_OPTIONAL
     m.def(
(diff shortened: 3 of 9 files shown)


Skylion007:
This isn't necessary though as we already have bindings for 'py::bytes` as well as `py::bytearray` and they both have constructors for std::string. Adding an additional return_value_policy just for this specific use case seems ill-advised and will make maintenance much more difficult. All one needs to do is wrap the std::string  or std::stringview with the py::bytes constructor to achieve the desired result.

tests/test_builtin_casters.cpp
```diff
@@ -378,4 +378,21 @@ TEST_SUBMODULE(builtin_casters, m) {
     m.def("takes_const_ref", [](const ConstRefCasted &x) { return x.tag; });
     m.def("takes_const_ref_wrap",
           [](std::reference_wrapper<const ConstRefCasted> x) { return x.get().tag; });
+
+    // test return_value_policy::return_as_bytes
+    m.def(
+        "invalid_utf8_string_as_bytes",
+        []() { return std::string("\xba\xd0\xba\xd0"); },
```
Skylion007, line 385:
```suggestion
        []() { return py::bytes(std::string("\xba\xd0\xba\xd0")); },
```
is all that was ever needed.

Skylion007:
If you are every unhappy with the behavior of casters, you can always just return the python objects themselves to the same effect.

tests/test_stl.cpp
```diff
@@ -541,4 +541,25 @@ TEST_SUBMODULE(stl, m) {
         []() { return new std::vector<bool>(4513); },
         // Without explicitly specifying `take_ownership`, this function leaks.
         py::return_value_policy::take_ownership);
+
+    // test return_value_policy::return_as_bytes
+    m.def(
+        "invalid_utf8_string_array_as_bytes",
+        []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; },
```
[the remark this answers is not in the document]
Skylion007: Better yet, you could return the py::array directly or use py::memoryview::from_buffer

rwgk:
> If you are every unhappy with the behavior of casters, you can always just return the python objects themselves to the same effect.

This is for the PyCLIF pybind11 code generator. You're right, for a simple `std::string` return we could generate a lambda with `py::bytes`, but for nested types that's not possible, as far as we can see, without very complex logic somehow unraveling the nested types, to then generated nested new code. Concretely:

```
 -        []() { return std::array<std::string, 1>{{"\xba\xd0\xba\xd0"}}; },
 +       []() { return std::array<py::bytes, 1>{{"\xba\xd0\xba\xd0"}}; },
```

That is a transformation we'd have to make **automatically** and recursively.

In contrast, the core of this PR is super simple:
* adds one `enum` (cost essentially zero)
* adds one `if` in the caster

It got us from ~97.5% success to ~98.5%, i.e. it was the top-most-important fix by far that we needed. (We're in the "long tail" phase of the project.)

rwgk:
We really need this — essentially 6-line change.

Skylion007:
My biggest concern is that this is  to other return_value_policy. IE, we may want to combine this flag with reference_internal etc at some point in the future. Also, this is only valid for a single underlying type (std::string caster), which seems wrong. I agree we probably need a better way to handle this, but hacking the caster like this doesn't seem right. Maybe we need a proxy wrapper that automatically triggers this modified caster behavior through a templated function?

Skylion007:
What we really need way is the ability to override the value_conv and key_conv behavior of these templated types.

rwgk:
> What we really need way is the ability to override the value_conv and key_conv behavior of these templated types. Abusing return_value_policy for this has terrible code smell.

Hm ... could you explain more?

This PR is taking a very simple path to achieve the desired behavior. Could it be even simpler?

henryiii:
Taking a simple path is not always the correct path. If you have a clear meaning for return value policy (which is _not_ to control the conversion of types), then abusing it to do what you want here can spell disaster down the road. It might make it impossible to refactor, for example, is hard to document and confusing to newcomers. I think it's worth investigating to see if there's a way around it without making it a return_value_policy.

include/pybind11/detail/common.h
```diff
@@ -459,7 +459,14 @@ enum class return_value_policy : uint8_t {
         collected while Python is still using the child. More advanced
         variations of this scheme are also possible using combinations of
         return_value_policy::reference and the keep_alive call policy */
-    reference_internal
+    reference_internal,
+
+    /** Use this policy to make C++ functions return bytes to Python instead of
```
rwgk, line 464: A bit of word-smithing:

With this policy, C++ string types are converted to Python `bytes`, instead of `str`. This is most useful when a C++ function returns a container-like type with nested C++ string types, and `py::bytes` cannot be applied easily. Note that this `return_value_policy` is not concerned with lifetime/ownership semantics, like the other policies, but the purpose of `return_as_bytes` is certain to be orthogonal, because C++ strings are always copied to Python `bytes` or `str`.

rwgk:
> º> This PR is taking a very simple path to achieve the desired behavior. Could it be even simpler? Simpler is not necessarily the issue here. Here is a trivial use case the current API design would not cover:
> 
> ```python
>        []() { return std::array<py::bytes, 1>{{"\xba\xd0\xba\xd0"}}; },
> ```
> 
> and specify to have it return by reference, by value, or by copy. Since we return value policy is an Enum and not a flag, we cannot combine this return value policies easily with the other ones which are mutually exclusive to one another.
> 
> The real issue here is that we have no way disambiguate casters except by doing the cast ourselves in the lambda. This is normally trivial, but becomes non-trivial for container or other variant types. We could of course fix that with another caster which changes the behavior of current caster. The main issue is that we cannot disambiguate return types that multiple valid C++ casts. Strings actually have three:
> 
> 1. An immutable unicode or other codec variant string
> 2. An immutable byte object
> 3. A mutable bytearray (Do you suggest we add yet ANOTHER return value policy for bytearrays which are nested?
> 
> This exposes three issues with our current :
> 
> 1. We cannot pass any extra arguments to our C++ to Python casters.
> 2. We cannot pass any arguments to the value and key casters in our STL casters (std::vector, std::array, and std::unordered_map, etc...) (EXCEPT FOR return_value_policy, and the parent handle).
> 3. We cannot specify a custom type caster to be used only for a single function, we have to reimplement that type caster as part of the lambda.
> 
> We could just add a special wrapper that triggers an alternative version of the stl_casters, but I don't think that would solve your problem since presumably your issue is actually ABSL or other container types?
> 
> The easiest solution is probably to call the caster directly in the lambda with some modified optional args. The messiest but robust solution would be to do some STL magic to rip apart these nested casters and convert them to byte strings and automatically has a templated getter that does the casting as it is accessed: https://stackoverflow.com/a/14280396/2444240
> 
> We could abstract the list_caster, array_caster, map_caster, and set_caster further to allow for templates which modify the key_conv and value_conv as well though.
> 
> Another solution though would probably be to make this another extra arg that specified in the def() block that specializes the casters or allows the user to specify their own.
> 
> @henryiii @wjakob I would love to hear your thoughts on how to best solve this use case / ambiguity.

This is a very good analysis, thanks Aaron! For strategic reasons, we cannot afford to make this a huge general project at the moment, we have to approach this with a long-term view: we still have to prove that pybind11 actually works for Google, by successfully integrating it into PyCLIF, i.e. we have to get from 98.5% to 100%. Once we've made that hurdle, we can devote more time on bigger projects for pybind11 itself, like generalizing the return_value_policy concept.

`return_value_policy` as-is is certainly not capable of covering complex cases in general. The bytes/str issue is just a more (the most?) practically important example. In that sense we're not making anything worse, we're just brining a fundamental limitation to light. — Assume the enum type wasn't called `return_value_policy`, but return_value_ownership_policy`, clearly bytes/str would be a mismatch. But the name is `return_value_policy`, and bytes/str totally makes sense, it's just an orthogonal dimension that wasn't used before.

Will anyone ever need something more general? Will that just be over-engineering? I don't know.

With one enum and if we can cover what we need. We can generalize from there, as and if needed. I don't think that one extra enum targeted at (currently) one `if` in one caster can lead to terrible problems down the road.

Skylion007:
I would also be okay if it's hidden by some IFDEF flag. I just really don't want this becoming a part of the public API that we have to support later like the PYTHON2 compatability macros.

rwgk:
> > This is a very good analysis, thanks Aaron! For strategic reasons, we cannot afford to make this a huge general project at the moment, we have to approach this with a long-term view: we still have to prove that pybind11 actually works for Google, by successfully integrating it into PyCLIF, i.e. we have to get from 98.5% to 100%. Once we've made that hurdle, we can devote more time on bigger projects for pybind11 itself, like generalizing the return_value_policy concept.
> 
> Maybe it's best for this to live in another branch and merge in

I could easily maintain this in the smart_holder branch if you prefer. pybind11 will not work for PyCLIF without this, just like pybind11 won't work for PyCLIF without `py::smart_holder`. It's a pretty good fit there.

> once it's cleaned up and up to 100%?

My totally personal bet: nobody will ever have enough motivation (time/money) to generalize `return_value_policy` further, to address the general limitations you pointed out. Note that Dave Abrahams invented it 20+ years ago. It's limited, but does the job for the most part. After literally two decades it seems unlikely that anyone would seriously want to tear this up.

Skylion007:
Actually @wangxf123456 for internal usages, does this return_value_policy works with dicts? Or does it only work with sequences and sets?

include/pybind11/detail/common.h
```diff
@@ -459,7 +459,16 @@ enum class return_value_policy : uint8_t {
...
-    reference_internal
+    reference_internal,
+
+    /** With this policy, C++ string types are converted to Python bytes,
+        instead of str. This is most useful when a C++ function returns a
+        container-like type with nested C++ string types, and py::bytes cannot
+        be applied easily. Note that this return_value_policy is not concerned
+        with lifetime/ownership semantics, like the other policies, but the
+        purpose of return_as_bytes is certain to be orthogonal, because C++
+        strings are always copied to Python bytes or str. */
```
Skylion007, line 470: Can we mark this as experimental and likely to change in the future?

wangxf123456:
> Actually @wangxf123456 for internal usages, does this return_value_policy works with dicts? Or does it only work with sequences and sets?

This should only work with sequences and sets. For example, return types like `Dict[str, bytes]` does not work. But there are only a few cases like this internally and can be easily fixed. For more complex cases, I have not done enough tests. But there are only less than 10 cases like `Dict[bytes, object]` internally.

Skylion007:
Can you please update the relevant comment and also say the usage of this experimental and likely to change.

include/pybind11/detail/common.h
Skylion007: @wangxf123456 Please add this: https://github.com/pybind/pybind11/pull/3838#issuecomment-1085141637 here too

Skylion007:
@wangxf123456 @rwgk I also found some C++ template magic to rebind the type of (STL) container types: https://stackoverflow.com/a/32214640/2444240 We might be able to use a trick like this to recursively change the Return type parameter to disambiguate how we should return std::string.

The TLDR of this is we need to have a variant of std::string caster that prefers to output bytes and we need a way to signal that should be called. 

One idea is to change the behavior of out_cast with a special tag like we for is_operator. This is actually pretty easy, but would involve setting a static variable that I would like to avoid. We need some way to pass that info into the type caster or query that extra arg from inside the std::string caster.

There also may be a way to abuse the polymorphic_type_hook to do this.

I am by no means an expert on C++ templating idioms, but I feel like there has to be a better way to do this than abusing the return type.

henryiii:
I'd much rather @Skylion007's suggestion be attempted. This is adding a misusage of the return policies that doesn't even cover all cases like dict's, or complex types. It does not combine with other return types (since they are disjoint conceptual features). If we add it, it will be impossible to pull out (just see our "private" compatibility macros!). Just because it's easy don't make it right. If it is really, really needed and attempts to do it properly have failed, then the name should start with an underscore, and probably have a warning in the code that it is not guarantied to be kept in the future.

rwgk:
FYI: I'm systematically combing through include/pybind11 to see how the one new enum + if could disturb things, or set up traps. I'll report here when I'm done. Could take a few more days. (Our focus is on larger scale issues that we need to sort out to get to 100% success rate for PyCLIF + pybind11.)

henryiii:
> If we add it, it will be impossible to pull out (just see our "private" compatibility macros!). If it is really, really needed, **then the name should start with an underscore**, and probably have a warning in the code that it is not guarantied to be kept in the future.

I’m not worried about the current code. I’m worried about breaking the mental and programmatic model of a return value policies and casters.

henryiii:
> I'm systematically combing through include/pybind11 … Could take a few more days.

That time might be better spent trying @Skylion007’s ideas above.

henryiii:
What c++ standard are you targeting in PyCLIF?

rwgk:
> > I'm systematically combing through include/pybind11 … Could take a few more days.
> 
> That time might be better spent trying @Skylion007’s ideas above.

Data first. Also priorities. I don't want the big project die the death of a thousand cuts, getting sidetracked with too many side issues.

> What c++ standard are you targeting in PyCLIF?

C++17 required (already).

But I want to keep the smart_holder branch compatible with master. (I spent many hours keeping it compatible with all the old compilers.)

henryiii:
> Data first.

What data? What are you looking for? We already know it “works” so we know you’ll find “no issues”. The first time you break a model to introduce a shortcut it always works, because the model is clean and separated. The second time, it doesn’t, because you only get to break the rules once before the breakages interfere. For example, if you tomorrow decide that to get to 100%, you also need to change one more caster, so you add one more return policy enum, then find you needed them together, so now you need a combined enum… etc. Or maybe you need to actually use a different return policy on a bytes function and then you need a full set of 2x more return policies.

Return policies and casters are orthogonal. One describes how to handle copy/references to the return value, the other handles casting between C++ and Python types. That’s why this PR had to add a name to an unused parameter.

I’ve already stated my requirements for putting this in. It can be added with an underscore and seen as temporary, or behind an ifdef would work too if you really need it in “immediately”. I can try to put together an alternative attempt and see if it works. @Skylion007’s suggestions do not look that hard to attempt.

henryiii:
> C++17 required (already).

some of the suggestions do not require changing pybind11.

Are the tests in the PR representative of what you care about? If we find a way to make them pass you’d be happy?

rwgk:
> What data? What are you looking for?

Going through the sources to analyze how return value policy and casters play together (learning). To see if we can think of concrete problems created (anticipating). — I agree with the general sentiment, we need to be careful. But there is also what I believe is called FUD, just being afraid without knowing exactly why, and passing up good opportunities just out of fear. Where do we stand? idk yet.

> Are the tests in the PR representative of what you care about?

I think so, although what we care about primarily is: amenable to automatically generating pybind11-based bindings code. Recursive unravelling to insert something at a certain level is very difficult.

rwgk:
Review of `return_value_policy` uses
====================================


Big picture
-----------

* `return_value_policy` is an enum type defined in detail/common.h

* The enums target `static` `type_caster` functions for C++ to Python conversions, with this general signature:

```
    template <typename T>
    static handle cast(T, return_value_policy /* policy */, handle /* parent */);
```


Method used to pin-point source code locations to review
--------------------------------------------------------

* Searching for `return_value_policy::` under pybind11/include.

* Excluding default template parameters.

* Excluding code that passes `return_value_policy` as constants.

* This basically leaves code involving `return_value_policy` in `if` and `switch` statements.

* Systematically enumerating all `static handle cast` functions.


Conclusions
-----------

* 17/38 `static handle cast` functions ignore `return_value_policy policy`.
* All but two `static handle cast` do not guard against a `policy` that does not apply.
* For those two `static handle cast`, a custom `case` for `return_by_bytes` will be useful, to generate a tailored error message for clarity.
* In general, `return_value_policy` does not support differentiated handling of sub-components of composite types (e.g. `std::pair`).
* In general, using `return_as_bytes` is more concise than having to write a lambda function, especially for functions with many arguments.


Judgements
----------

* In retrospect, it almost seems like an oversight or obvious omission that the `string_caster` completely ignores the `return_value_policy` feature, although it has to support it in the `static handle cast` signature, but then users are directed anyway to write potentially complicated lambda functions to handle `str` vs `bytes` return values, even for the simple and most common case.

* Using a general limitation of `return_value_policy` (does not support differentiated handling of sub-components) against adding `return_by_bytes` has no merit as an arguments, at least not as an argument in isolation.


Code reviewed
=============

detail/type_caster_base.h
-------------------------

```
class type_caster_generic {
```
* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/detail/type_caster_base.h#L598
```
...
       switch (policy) {
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/detail/type_caster_base.h#L654
```
            default:
                throw cast_error("unhandled return_value_policy: should not happen!");
```

```
class type_caster_base : public type_caster_generic {
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/detail/type_caster_base.h#L977
```
    static handle cast(const itype &src, return_value_policy policy, handle parent) {
        if (policy == return_value_policy::automatic
            || policy == return_value_policy::automatic_reference) {
            policy = return_value_policy::copy;
        }
        return cast(&src, policy, parent);
    }
```


cast.h
------
```
template <typename type>
class type_caster<std::reference_wrapper<type>> {
```
* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/cast.h#L68
```
    static handle
    cast(const std::reference_wrapper<type> &src, return_value_policy policy, handle parent) {
        // It is definitely wrong to take ownership of this pointer, so mask that rvp
        if (policy == return_value_policy::take_ownership
            || policy == return_value_policy::automatic) {
            policy = return_value_policy::automatic_reference;
        }
        return caster_t::cast(&src.get(), policy, parent);
    }
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/cast.h#L92
```
    static ::pybind11::handle cast(                                                               \
        T_ *src, ::pybind11::return_value_policy policy, ::pybind11::handle parent) {             \
        if (!src)                                                                                 \
            return ::pybind11::none().release();                                                  \
        if (policy == ::pybind11::return_value_policy::take_ownership) {                          \
            auto h = cast(std::move(*src), policy, parent);                                       \
            delete src;                                                                           \
            return h;                                                                             \
        }                                                                                         \
        return cast(*src, policy, parent);                                                        \
    }                                                                                             \
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/cast.h#L645
```
    // copied from the PYBIND11_TYPE_CASTER macro
    template <typename T>
    static handle cast(T *src, return_value_policy policy, handle parent) {
        if (!src) {
            return none().release();
        }
        if (policy == return_value_policy::take_ownership) {
            auto h = cast(std::move(*src), policy, parent);
            delete src;
            return h;
        }
    }
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/cast.h#L1040
```
// C++ type -> py::object
template <typename T, detail::enable_if_t<!detail::is_pyobject<T>::value, int> = 0>
object cast(T &&value,
            return_value_policy policy = return_value_policy::automatic_reference,
            handle parent = handle()) {
    using no_ref_T = typename std::remove_reference<T>::type;
    if (policy == return_value_policy::automatic) {
        policy = std::is_pointer<no_ref_T>::value     ? return_value_policy::take_ownership
                 : std::is_lvalue_reference<T>::value ? return_value_policy::copy
                                                      : return_value_policy::move;
    } else if (policy == return_value_policy::automatic_reference) {
        policy = std::is_pointer<no_ref_T>::value     ? return_value_policy::reference
                 : std::is_lvalue_reference<T>::value ? return_value_policy::copy
                                                      : return_value_policy::move;
    }
    return reinterpret_steal<object>(
        detail::make_caster<T>::cast(std::forward<T>(value), policy, parent));
}
```

* https://github.com/pybind/pybind11/blob/b3ebd11d9851007f525e2e41de53b7913b24787c/include/pybind11/cast.h#L1247


pybind11.h
----------

The family of `def_property` member functions has these comments:
```
    /// Uses return_value_policy::reference_internal by default
    /// Uses cpp_function's return_value_policy by default
    /// Uses return_value_policy::reference by default
    /// Uses cpp_function's return_value_policy by default
    /// Uses return_value_policy::reference_internal by default
    /// Uses cpp_function's return_value_policy by default
    /// Uses return_value_policy::reference by default
    /// Uses cpp_function's return_value_policy by default
```


attr.h
------
```
struct function_record {
...
    return_value_policy policy = return_value_policy::automatic;
```


pytypes.h
---------

No `return_value_policy::` except as default template parameters.


stl_bind.h
----------

`return_value_policy::` only passed as constants.


eigen.h
-------

Similar to type_caster_base.h:

* 2 `switch` with `default` raising an exception.

* 2 `if` `return_value_policy::automatic` mapping to `return_value_policy::copy`.


All `static handle cast` functions
----------------------------------

* 38 total
* 21 use `return_value_policy policy`
* 17 ignore `return_value_policy`

```
cast.h:     cast(const std::reference_wrapper<type> &src, return_value_policy policy, handle parent) {
cast.h:     static handle cast(T, return_value_policy /* policy */, handle /* parent */) {
cast.h:     static handle cast(const void *ptr, return_value_policy /* policy */, handle /* parent */) {
cast.h:     static handle cast(bool src, return_value_policy /* policy */, handle /* parent */) {
cast.h:     cast(const StringType &src, return_value_policy /* policy */, handle /* parent */) {
cast.h:     static handle cast(const CharT *src, return_value_policy policy, handle parent) {
cast.h:     static handle cast(CharT src, return_value_policy policy, handle parent) {
cast.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
cast.h:     static handle cast(T *src, return_value_policy policy, handle parent) {
cast.h:     static handle cast(const holder_type &src, return_value_policy, handle) {
cast.h:     static handle cast(holder_type &&src, return_value_policy, handle) {
cast.h:     static handle cast(const handle &src, return_value_policy /* policy */, handle /* parent */) {
numpy.h:     static handle cast(const handle &src, return_value_policy /* policy */, handle /* parent */) {
functional.h:     static handle cast(Func &&f_, return_value_policy policy, handle /* parent */) {
stl.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
stl.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
stl.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
stl.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
stl.h:     static handle cast(T &&src, return_value_policy policy, handle parent) {
stl.h:     static handle cast(Variant &&src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(Type &&src, return_value_policy /* policy */, handle parent) {
eigen.h:     static handle cast(const Type &&src, return_value_policy /* policy */, handle parent) {
eigen.h:     static handle cast(Type &src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(const Type &src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(Type *src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(const Type *src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(const MapType &src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(const Type &src, return_value_policy /* policy */, handle /* parent */) {
eigen.h:     static handle cast(const Type *src, return_value_policy policy, handle parent) {
eigen.h:     static handle cast(const Type &src, return_value_policy /* policy */, handle /* parent */) {
chrono.h:     static handle cast(const type &src, return_value_policy /* policy */, handle /* parent */) {
chrono.h:     static handle cast(const std::chrono::time_point<std::chrono::system_clock, Duration> &src,
complex.h:     cast(const std::complex<T> &src, return_value_policy /* policy */, handle /* parent */) {
filesystem.h:     static handle cast(const T &path, return_value_policy, handle) {
type_caster_base.h:     PYBIND11_NOINLINE static handle cast(const void *_src,
type_caster_base.h:     static handle cast(const itype &src, return_value_policy policy, handle parent) {
type_caster_base.h:     static handle cast(itype &&src, return_value_policy, handle parent) {
type_caster_base.h:     static handle cast(const itype *src, return_value_policy policy, handle parent) {
```

rwgk:
> I think so, although what we care about primarily is: amenable to automatically generating pybind11-based bindings code. Recursive unravelling to insert something at a certain level is very difficult.

Xiaofei tried for an hour or two, but we ran into issues and decided to give up. There are far higher priority items for us to tackle first.

tests/test_builtin_casters.cpp
```diff
@@ -15,6 +15,11 @@ struct ConstRefCasted {
     int tag;
 };
 
+struct StringAttr {
+    StringAttr(std::string v) : value(v) {}
```
Skylion007, line 19:
```suggestion
    StringAttr(std::string v) : value(std::move(v)) {}
```

laramiel:
The summary of the problem is pretty succinct:  **C++ std::string types are not Unicode**.  Full Stop.

Given this, all std::string / std::string_view types should be returned as py::bytes.  So really this should go the other way, and there should be a policy that converts to string as an extra step.

henryiii:
The point is not that it shouldn’t be possible to change this, it’s just that return value policies are the wrong way to change it. I think it would be great to be able to configure this.

And yes, string -> Unicode is an irritating historical and convenience mismatch.

laramiel:
Maybe we should break that "historical mismatch".

henryiii:
I've already explained it. Return value policies control the lifetime of the return value. They have nothing to do with type casters. We have two distinct, unrelated concepts and mixing them is wrong and will cause problems. The problem isn't really anything to do with return values anyway, it's that you want to "fix" the type caster to be `std::string` -> `py::bytes`. And in this case, you are trying to do it via return policies.

> Maybe we should break that "historical mismatch".

You skipped the word "convenience" above.

Only if you explain to all 100,000+ pybind11 users why you broke their code and why they now need to add decoding everywhere and why `std::string` is not `str`, but rather `bytes`. You'd also probably need to explain why there's no native type to use to convert to `str` while you are at it, along with why both `std::string` and `std::bytearray` convert to `bytes`. These things _can_ be explained - but they are very, very breaking changes and make pybind11 a bit harder to use for simple cases (and easier for hard cases).

It's a very handy convenience that a huge number of users are happy with, and the most obvious behavior for newcomers. The problem with this "simplify this for users even if it's wrong" behavior is you currently can't turn it off if you know what you are doing. (For the record, I'd be very tempted not to make this mistake again if replmementing pybind11 - I don't like this sort of incorrectness!)

If there was a single, global level compile-time toggle controlling `std::string` -> `str`/`bytes`, would that be useable?

rwgk:
> I've already explained it. Return value policies control the lifetime of the return value.

I addressed that already: https://github.com/pybind/pybind11/pull/3838#issuecomment-1085004387. Just because it was that way doesn't suffice as a reason to keep it that way.

> They have nothing to do with type casters.

That is obviously false. See the top of https://github.com/pybind/pybind11/pull/3838#issuecomment-1094390333.

> We have two distinct, unrelated concepts

"Distinct" is true. "Unrelated" obviously false: strings are always copied. That is a lifetime aspect.

> and mixing them is wrong

This is a judgement, not a reason. You have already two flaws in your premises, therefore it is difficult to accept your judgement.

> and will cause problems. The problem isn't really anything to do with return values anyway,

You don't explain what the problems are. "anyway" is tell-tale sign of evasiveness.

>  it's that you want to "fix" the type caster to be `std::string` -> `py::bytes`. And in this case, you are trying to do it via return policies.

It's exactly about the return value. That a type caster is involved is strictly speaking an implementation detail, although it leaks in that the `return_value_policy` exactly targets `type_caster` `static handle cast`.

laramiel:
A couple points:

* `return_value_policy` absolutely does affect `type_caster::cast()` calls. Quite a few type casters behave a bit differently depending on the policy.

* There's no other mechanism beyond wrapping in a lambda and manually calling `cast` to affect `type_caster<>` selection for method return types.

laramiel:
> > Maybe we should break that "historical mismatch".
> 
> You skipped the word "convenience" above.

Yes, I did. It's not convenient if your std::string does not hold unicode. So sometimes it's convenient, and sometimes it isn't.

> Only if you explain to all 100,000+ pybind11 users why you broke their code and why they now need to add decoding everywhere and why `std::string` is not `str`, but rather `bytes`. You'd also probably need to explain why there's no native type to use to convert to `str` while you are at it, along with why both `std::string` and `std::bytearray` convert to `bytes`. These things _can_ be explained - but they are very, very breaking changes and make pybind11 a bit harder to use for simple cases (and easier for hard cases).
> 

Perhaps we break this with a major version release? We would need a mechanism that allows selection of the type_caster<> used for return types. While we're at it, perhaps we could specify non-default type casters for method parameters via the py::args() call?

Anyway, I expect that the 100,000+ pybind11 users is an over-exaggeration; for open-source projects it's more likely in the 100s of projects, so more along the lines of 1000s of pybind11 users total, I expect.  That's still a pretty large problem, but over-exaggeration doesn't help.

> If there was a single, global level compile-time toggle controlling `std::string` -> `str`/`bytes`, would that be useable?

A global compile-time toggle will not work very well for the same reason that you claim that we can't break the "100,000s of users".  Namely, if we could change the default to return bytes for all of the google projects, then we could probably universally change the default anyway, and the users wouldn't be a problem. Also, changing how the `type_caster<std::string>` behaves in one CU vs. another CU is a textbook case of an ODR violation, and it will likely blow up at some point in ways that are impossible to debug.

To attempt to distill this:

* converting std::string to str is the wrong thing to do in many cases
* there is no mechanism to select the `type_caster<>` for return types
* the only existing mechanism to affect the `type_caster<>` for return types is `return_value_policy`
* a single global toggle (e.g. changing the default for `type_caster<std::string>` using a #define) is unlikely to work
* we don't want to introduce ODR violations in whatever solution we have.

Perhaps an alternate proposal would help the discussion?

henryiii:
There are 29.8k results for PYBIND11_MODULE in sourcegraph, and that's not every repo, just the top 5 million open source ones. In fact, we have over 10K stars alone. Our python module on PyPI (which many users do not use) gets over 1M downloads a week. So about 50k-100k is not an exaggeration. This includes PyTorch, SciPy, etc. It's not just a Google project.

Many, many users simply have simple ascii text in strings, and they want `str` in Python, because `bytes` are annoying. Putting anything other than ascii in std::string is already problematic for some other libraries too (including unicode!). I understand the "correct" thing to do is produce bytes, and I'd love to see a way to do that, but changing this as a default is not happing, even in 3.0. It would basically be the Python 2->3 transition, but for pybind11. (Ironically, same 2->3 number!) To the best of my knowledge, most C++ libraries avoid anything past ASCII unless they really have to, since C++ has terrible encoding support.

I don't like the "smart" return either - it would then mean that you would not know whether to encode/decode the return (even between calls!) - that would be terrible and would introduce a huge amount of bugs.

I would like to see an alternative proposal - this is very much a workaround and a complete change for the meaning of return_value_policy. If any alternative looks terrible, that would make swallowing this a bit easier.

> Namely, if we could change the default to return bytes for all of the google projects, then we could probably universally change the default anyway, and the users wouldn't be a problem. 

Google is _not_ all open source. Specifically, you can work out and set something across all projects, run tests across all projects, etc. It could very much be that Google would always have "PYBIND11_STRING_IS_BYTES" defined (as part of PyCLIF) and other projects could opt in. Getting all Google on one pages is thousands of times easier than getting all users of pybind11 converted. There's already a smart pointer toggle?

Also, this is needed for a very specific thing (supporting the PyCLIF pybind11 code generator), and hasn't been that big of an issue before. The docs specifically say `std::string` should hold unicode, and if it doesn't, to wrap in `py::bytes`.

Specific points: 
* What type caster returns a different _type_ based on the return value policy? The docs clearly say it controls the _lifetime_, not the type.
* Is there any place where you'd need to select a return value policy and still control bytes/str, say for a nested structure?
* If we allow `return_value_policy` to control type conversions, what's to stop other ones, like selecting `std::filesystem` to `Path`/`str`, `char` to `int`/`str`/`bytes`, `std::vector` to `list`/`tuple`, etc? If the answer to the question above was no, is it still no for all of these? Note that this is hardcoded in, too, so users can't add or change these themselves (which is why you want to hard-code in this workaround).

wjakob:
Just for the record, I am also not in favor for repurposing return value policies for this unrelated task. It's not the problem that they were intended to solve, and it adds a distraction to a concept that is already difficult to grasp (object ownership is complicated when interfacing C++ and  Python).

laramiel:
I think that users vs. uses is the difference. Your data is good; I was looking more at the projects-level uses of pybind11, and am well aware of uses beyond google, and the hundred-ish project level uses that are quite prominent.

I'm not actually arguing for or against this proposal. I'm arguing that the problem needs to be solved and just saying that we don't like repurposing `return_value_policy` isn't, in and of itself, a great answer.  It's completely possible to rework the meaning of return_value_policy to handle this. It's also possible to introduce something else.

Having thought about it a bit more, `Union[str, bytes]` is not very good, I agree.

Since I'm still not seeing alternate proposals, I'll throw out another strawman proposal, but it's more invasive:

**Introduce a `pbind11::returns()` that we pass similar to `pybind11::args("name")`.**

It might look something like this:

```
template<typename T>
struct returns {
  using type = T;
  returns() = default;
  /// We could put return_value_policy in here, kind of like args()
};
```

Then we'd have to update `type_caster<>::cast()` to a signature like:
(Imagine that this is `type_caster<std::string>`)

```
template<typename T, typename R>
static pybind11::handle(T&& src, R returns, pybind11::handle parent) {
  if constexpr (std::is_same<R::type, py::bytes>) {
    /// convert to bytes.
  }
  /// Convert to str.
}
```

It could be called like:

```
  m.def("returns_str", []() { return std::string("bytes"); });
  m.def("returns_bytes", []() { return std::string("bytes"); }, pybind11::returns<pybind11::bytes>());
```

Now that's clearly pretty intrusive to all the existing pybind type_caster implementations, but perhaps that can be worked around since most callers won't pass in a `pybind11::returns()`.

For it to be useful, it might require that returns accept an arg-pack for the recursive dict case, and a method to query the type of a particular index, a la std::tuple.

rwgk:
> It's not the problem that they were intended to solve

When Dave introduced return value policies there was no `str` & `bytes`. When those were introduced, handling that with return value policies would have been a logical choice. `py::bytes` lambdas really don't look very elegant and are clearly ...

> and it adds a distraction to a concept that is already difficult to grasp (object ownership is complicated when interfacing C++ and Python).

... a much bigger mental leap to make (distracting) than understanding that `return_as_bytes` returns `bytes`.

Even if someone just looks at `return_as_bytes` and doesn't make any connection to lifetime handling, or vice versa , there isn't really a potential for distraction. Maybe at some point "oh, it does both", but "oh, of course, strings are always copied". That's fine, learning on the go without getting pulled into traps.

Re traps, coincidentally we just discovered that the `py::bytes` lambda trick got us into one: https://github.com/google/clif/commit/c0e1e91a05923b506d6795bce8e0cd9ee1e13820

rwgk:
> m.def("returns_bytes", []() { return std::string("bytes"); }, pybind11::returns<pybind11::bytes>());

I really like this, too, it has the potential to become a comprehensive solution for complex cases. But I need to be pragmatic to stay on track with the bigger project. We are already critically dependent on this PR for the simple, common case, and this existing solution could very easily be converted to the comprehensive one when it becomes available. To stay organized, I will merge this on the smart_holder branch, with a note in the enum comment that the feature isn't available on master.

henryiii:
If it helps: I've already said I'd be okay with it if it has a private name (`_return_as_bytes`) and a warning that it might be removed. This does need to be fixed properly soon, but a quick solution is okay as long as no one users it beyond users who can move to the new solution quickly once it's ready.

rwgk:
> If it helps: I've already said I'd be okay with it if it has a private name (`_return_as_bytes`) and a warning that it might be removed. This does need to be fixed properly soon, but a quick solution is okay as long as no one users it beyond users who can move to the new solution quickly once it's ready.

Thanks, I added the underscore. I'll merge this now on the smart_holder branch, to keep the PyCLIF-pybind11 integration work on track. If @wjakob supports having this on master, I'll back-port.

PR #3840, opened 2022-03-31
feature: support compilers that use std::experimental::filesystem 

Llcoolsouder:
feature: support compilers that use std::experimental::filesystem such as gcc7

## Description
Minor changes to support older C++ compilers where filesystem is not yet part of the standard library and is instead included in std::experimental::filesystem. 

## Suggested changelog entry:

```rst
Minor changes to support older C++ compilers where filesystem is not yet part of the standard library and is instead included in std::experimental::filesystem. 
```

### DIFF
--- a/include/pybind11/stl/filesystem.h
+++ b/include/pybind11/stl/filesystem.h
@@ -104,7 +104,8 @@ public:
 };
 
 template <>
-struct type_caster<pybind11_filesystem_alias::path> : public path_caster<pybind11_filesystem_alias::path> {};
+struct type_caster<pybind11_filesystem_alias::path>
+    : public path_caster<pybind11_filesystem_alias::path> {};
 #endif // PYBIND11_HAS_FILESYSTEM
 
 PYBIND11_NAMESPACE_END(detail)
--- a/include/pybind11/stl/filesystem.h
+++ b/include/pybind11/stl/filesystem.h
@@ -18,11 +18,9 @@
           PY_VERSION_HEX >= 0x03060000
 #            include <filesystem>
 #            define PYBIND11_HAS_FILESYSTEM 1
-namespace pybind11::filesystem_alias = std::filesystem;
 #        elif __has_include(<experimental/filesystem>)
 #            include <experimental/filesystem>
 #            define PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM 1
-namespace pybind11::filesystem_alias = std::experimental::filesystem;
 #        endif
 #    endif
 #endif
@@ -103,10 +101,17 @@ public:
     PYBIND11_TYPE_CASTER(T, const_name("os.PathLike"));
 };
 
+#endif // PYBIND11_HAS_FILESYSTEM || defined(PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM)
+
+#if defined(PYBIND11_HAS_FILESYSTEM)
 template <>
-struct type_caster<pybind11::filesystem_alias::path>
-    : public path_caster<pybind11::filesystem_alias::path> {};
-#endif // PYBIND11_HAS_FILESYSTEM
+struct type_caster<std::filesystem::path>
+    : public path_caster<std::filesystem::path> {};
+#elif defined(PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM)
+template <>
+struct type_caster<std::experimental::filesystem::path>
+    : public path_caster<std::experimental::filesystem::path> {};
+#endif
 
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/stl/filesystem.h
+++ b/include/pybind11/stl/filesystem.h
@@ -105,8 +105,7 @@ public:
 
 #if defined(PYBIND11_HAS_FILESYSTEM)
 template <>
-struct type_caster<std::filesystem::path>
-    : public path_caster<std::filesystem::path> {};
+struct type_caster<std::filesystem::path> : public path_caster<std::filesystem::path> {};
 #elif defined(PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM)
 template <>
 struct type_caster<std::experimental::filesystem::path>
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

include/pybind11/stl/filesystem.h
```diff
@@ -13,21 +13,30 @@
...
 #ifdef __has_include
-#    if defined(PYBIND11_CPP17) && __has_include(<filesystem>)
-#        include <filesystem>
-#        define PYBIND11_HAS_FILESYSTEM 1
+#    if defined(PYBIND11_CPP17)
+#        if __has_include(<filesystem>) && \
+          PY_VERSION_HEX >= 0x03060000
+#            include <filesystem>
+#            define PYBIND11_HAS_FILESYSTEM 1
+namespace fs = std::filesystem;
```
rwgk, line 21: This is in the global namespace, and a very short name => I'm worried about name clashes. Could you please make this longer, e.g. `pybind11_filesystem_alias` or similar?

include/pybind11/stl/filesystem.h
```diff
@@ -13,21 +13,30 @@
...
+#        elif __has_include(<experimental/filesystem>)
+#            include <experimental/filesystem>
+#            define PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM
```
rwgk, line 24: Slightly silly, but for local consistency, could you also make this
```
define PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM 1
```
like the existing define?

include/pybind11/stl/filesystem.h
Llcoolsouder: Hey, great catch. I'll go ahead and update it.

include/pybind11/stl/filesystem.h
```diff
@@ -13,21 +13,30 @@
...
+#    if defined(PYBIND11_CPP17)
+#        if __has_include(<filesystem>) && \
+          PY_VERSION_HEX >= 0x03060000
+#            include <filesystem>
+#            define PYBIND11_HAS_FILESYSTEM 1
+namespace pybind11_filesystem_alias = std::filesystem;
+#        elif __has_include(<experimental/filesystem>)
+#            include <experimental/filesystem>
+#            define PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM 1
+namespace pybind11_filesystem_alias = std::experimental::filesystem;
```
Skylion007, line 25: Actually, any reason it needs to / should be in the global namespace? Do we want people downstream to reuse it? Even if we do wouldn't be better to just make it part of the pybind11 namespace?

Skylion007: Thoughts @rwgk ?

Llcoolsouder: I agree, totally doesn't need to be global

include/pybind11/stl/filesystem.h
```diff
@@ -13,21 +13,30 @@
 #include <string>
 
 #ifdef __has_include
-#    if defined(PYBIND11_CPP17) && __has_include(<filesystem>)
-#        include <filesystem>
-#        define PYBIND11_HAS_FILESYSTEM 1
+#    if defined(PYBIND11_CPP17)
```
Skylion007, line 16: This doesn't match experimental optional. Experimental optional can work even if only CPP14 is defined see: https://github.com/pybind/pybind11/blob/9aa676d38db4b18d9952c70486412d20c0213f13/include/pybind11/detail/common.h#L184

Skylion007: Actually, nvm as this file has C++17 syntax that would break and complicate the implementation anyhow.

include/pybind11/stl/filesystem.h
Skylion007: Yeah, I would try sticking it in detail and see if anything blows up.

Llcoolsouder: On second glance, I can't actually think of any way to completely remove this from the public API here. I can put it under the pybind11 namespace if that makes us feel better. Between `pybind11_filesystem_alias` or `pybind11::filesystem_alias`, I think the intent is clear enough. They're both unlikely to conflict with any user namespaces, they both have clear intent, and even if you were to use it, the can't cause any real damage; it's just a namespace alias.

Skylion007: @Llcoolsouder the `detail` namespace is how we denote private APIs / functions in pybind11. Putting under the pybind11 namespace would still be better though.

include/pybind11/stl/filesystem.h
```diff
@@ -13,21 +13,30 @@
...
-#if !defined(PYBIND11_HAS_FILESYSTEM) && !defined(PYBIND11_HAS_FILESYSTEM_IS_OPTIONAL)
+#if !defined(PYBIND11_HAS_FILESYSTEM) && !defined(PYBIND11_HAS_EXPERIMENTAL_FILESYSTEM)           \
+    && !defined(PYBIND11_HAS_FILESYSTEM_IS_OPTIONAL)
 #    error                                                                                        \
-        "#include <filesystem> is not available. (Use -DPYBIND11_HAS_FILESYSTEM_IS_OPTIONAL to ignore.)"
+        "Neither #include <filesystem> nor #include <experimental/filesystem is available. (Use -DPYBIND11_HAS_FILESYSTEM_IS_OPTIONAL to ignore.)"
 #endif
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
```
Skylion007, line 37: detail namespace begins here. Can you not just move the include logic after line 28?

Skylion007: Why do we even need the namespace btw? It's only used on the type caster right? Just mimic what we do for experimental optional and spell it out twice with an IFDEF?

Llcoolsouder: Addressed in the latest commit.

Since we're here though, what is the point of the `PYBIND11_NAMESPACE_BEGIN` and `PYBIND11_NAMESPACE_END` macros? It looks like we're just replacing the language syntax with a lesser known syntax.

rwgk: From memory: in some environments the macros insert "visibility hidden" attributes.

Skylion007 (approved):

rwgk (approved):
Looks very clean now.

PR #3844, opened 2022-04-02
[smart_holder] .def_readonly, .def_readwrite adaptors (continuation of PR #3581).

rwgk:
## Description

**Note**: This PR is a continuation of PR #3581, to preserve the original development history. The large-scale changes from Feb 2022 (Python 2 drop, global clang-format) make it difficult to meaningfully preserve the original development history in the same PR.

The original motivation for this PR was to fix the issue observed here: [disowning creates a dangling pointer](https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/testing/python/nested_fields_test.py#L56).

To ensure internal consistency down the road, the scope of this PR was expanded to cover `.def_readonly` and `.def_readwrite` behavior for

* value
* raw pointer
* `std::unique_ptr`
* `std::shared_ptr`

members (mutable & const for the pointer types).

For `shared_ptr` members, the existing behavior is as desired (no work needed).

For value and raw-pointer members, the [PyCLIF approach here](https://github.com/google/clif/blob/c371a6d4b28d25d53a16e6d2a6d97305fb1be25a/clif/python/instance.h#L233) was adopted: aliasing `shared_ptr` tying the lifetime of the "outer" object to the lifetime of the Python object pointing to the member. In this way the existing mechanisms prevent the "outer" object from being disowned while pointers to members are still alive.

For unique_ptr members, the situation is complex. This PR only establishes a proof of concept. See the code and comments in tests/test_class_sh_property.py for details. Productizing the proof of concept is left for a future separate PR, as needed.

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1875,12 +1875,11 @@ public:
     class_ &def_readwrite(const char *name, D C::*pm, const Extra &...extra) {
         static_assert(std::is_same<C, type>::value || std::is_base_of<C, type>::value,
                       "def_readwrite() requires a class member (or base class member)");
-        def_property(
-            name,
-            xetter_cpp_function<type, D>::read(pm, *this),
-            xetter_cpp_function<type, D>::write(pm, *this),
-            return_value_policy::reference_internal,
-            extra...);
+        def_property(name,
+                     xetter_cpp_function<type, D>::read(pm, *this),
+                     xetter_cpp_function<type, D>::write(pm, *this),
+                     return_value_policy::reference_internal,
+                     extra...);
         return *this;
     }
 
@@ -1888,11 +1887,10 @@ public:
     class_ &def_readonly(const char *name, const D C::*pm, const Extra &...extra) {
         static_assert(std::is_same<C, type>::value || std::is_base_of<C, type>::value,
                       "def_readonly() requires a class member (or base class member)");
-        def_property_readonly(
-            name,
-            xetter_cpp_function<type, D>::readonly(pm, *this),
-            return_value_policy::reference_internal,
-            extra...);
+        def_property_readonly(name,
+                              xetter_cpp_function<type, D>::readonly(pm, *this),
+                              return_value_policy::reference_internal,
+                              extra...);
         return *this;
     }
 
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1514,6 +1514,154 @@ using default_holder_type = smart_holder;
 
 #endif
 
+// Helper for the xetter_cpp_function static member functions below. xetter_cpp_function is
+// a shortcut for 'getter & setter adapters for .def_readonly & .def_readwrite'.
+// The only purpose of these adapters is to support .def_readonly & .def_readwrite.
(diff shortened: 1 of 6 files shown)


rwgk:
@rainwoodman
@charlesbeattie
@Skylion007 

Thanks for the comments on PR #3581, and sorry for taking extremely long to get back to it. I closed the other PR and continued with this one because we did the global clang-formatting in the meantime.

I think I addressed all comments, in separate small commits here, hoping that makes it easy for you to re-review. If you get a chance, could you please take another look?

Re CI: we have two existing failures on master, therefore 4 failures here, 2 for CI (smart_holder is just an add-on), 2 for CI-SH-DEF (smart_holder is the default holder).

include/pybind11/pybind11.h
```diff
@@ -1514,6 +1514,153 @@ using default_holder_type = smart_holder;
...
+    template <typename PM, must_be_member_function_pointer<PM> = 0>
+    static cpp_function read(PM pm, const handle &hdl) {
+        return cpp_function(
+            [pm](const std::shared_ptr<T> &c_sp) -> D { return D{std::move(c_sp.get()->*pm)}; },
+            is_method(hdl));
+    }
+
+    template <typename PM, must_be_member_function_pointer<PM> = 0>
+    static cpp_function write(PM pm, const handle &hdl) {
+        return cpp_function([pm](T &c, D &&value) { c.*pm = std::move(value); }, is_method(hdl));
```
Skylion007, line 1660: Any reason why std::move instead of std::forward?

include/pybind11/pybind11.h
```diff
@@ -1514,6 +1514,153 @@ using default_holder_type = smart_holder;
...
+    }
+
+    template <typename PM, must_be_member_function_pointer<PM> = 0>
+    static cpp_function read(PM pm, const handle &hdl) {
+        return readonly(pm, hdl);
+    }
+
+    template <typename PM, must_be_member_function_pointer<PM> = 0>
+    static cpp_function write(PM pm, const handle &hdl) {
+        return cpp_function([pm](T &c, D value) { c.*pm = value; }, is_method(hdl));
```
Skylion007, line 1586:
```suggestion
        return cpp_function([pm](T &c, D value) { c.*pm = std::forward<D>(value); }, is_method(hdl));
```

include/pybind11/pybind11.h
Skylion007: In other words, is this really an rvalue or should it be a universal forwarding reference?

Skylion007: Ah, nvm I think I get it because D is always a unique ptr.

include/pybind11/pybind11.h
rwgk: Thanks Aaron, applied (with clang-format).

rwgk:
The 4 CI failures are still the 2 x 2 failures that were fixed only recently on master.

I'll go ahead and merge this PR now, as a starting-point for updating the smart_holder branch (git merge master), global testing, and importing into the Google environment. Thanks for all the feedback! (If you have any additional suggestions, I'll work on them in a new PR.)

PR #3854, opened 2022-04-09
ci: drop win2016

henryiii:
- ci: drop dead windows CI jobs
- chore: touch up pre-commit

## Description

Long removed from GHA. Currently we have MSVC 2017 coverage only through Appveyor.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -814,55 +814,6 @@ jobs:
     - name: Python tests
       run: cmake --build build --config Debug -t pytest
 
-  win32-msvc2017:
-    name: "🐍 ${{ matrix.python }} • MSVC 2017 • x64"
-    runs-on: windows-2016
-    strategy:
-      fail-fast: false
-      matrix:
-        python:
-          - 3.6
-          - 3.7
-        std:
-          - 14
-
-        include:
-          - python: 3.7
-            std: 17
-            args: >
-              -DCMAKE_CXX_FLAGS="/permissive- /EHsc /GR"
-
-    steps:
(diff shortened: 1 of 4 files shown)


rwgk (approved):
Thanks for the cleanup!
Might be helpful to add a note to the PR description, something like "Currently we have MSVC 2017 coverage only through appveyor." (IIUC)

henryiii:
You can edit PR descriptions too. ;)

rwgk:
> You can edit PR descriptions too. ;)

I wasn't entirely sure, you're the authority for build-related stuff.

henryiii:
Yes, I'm just trying to understand why one check is failing (repeatedly). Probably happening on master too, so this is still an improvement.

henryiii:
Ahh, I think I know why. You can't have an older CMake than MSVC version, and that's why the config is failing now.

rwgk:
Yes, you're right, I overlooked that in the other errors. E.g. https://github.com/pybind/pybind11/actions/runs/2150571824

My vote is to merge this to reduce the distracting noise, then tackle that one separately.

rwgk:
No mercy :-)
```
CMake Error at tools/FindPythonLibsNew.cmake:[17](https://github.com/pybind/pybind11/runs/5978924057?check_suite_focus=true#step:7:17)6 (message):
  Python config failure: Python is 32-bit, chosen compiler is 64-bit
```

henryiii:
I'm switching my "I care about 32-bit" hat (https://github.com/scipy/oldest-supported-numpy/issues/45) for my "I don't care about 32-bit" hat for now. :)

PR #3674, opened 2022-02-01
Experiments on top of PR #3643 (pybind11_select_caster)

rwgk:
## Description
WIP on top of #3643.

## Suggested changelog entry:

```rst

```

PR #3862, opened 2022-04-11
Allow type_caster<> selection via ADL

laramiel:
## Description

Instead of the type_caster template specialization, it is possible to register the conversions with a function declaration, which will be found with [ADL](https://en.cppreference.com/w/cpp/language/adl).

```
namespace std {
string_caster pybind11_select_caster(std::string*);
}
```

The parameter is a pointer to the C++ (intrinsic) type. The returned type is the caster class. This function should have no implementation!

There is no ODR violation when different translation units define different conversions: distinct modules might choose different behavior. This is the main motivation of this change.

* The declarations need to be defined in the original namespace of the converted C++ type.
* The default behavior did not change: type_caster<T> will be used as a fallback.

Because type_caster<T> not always used, load_caster has to accept a generic Caster argument.

This is manually adapted from #3643 and #3674

## Suggested changelog entry:

```rst
Allow type_caster<> selection via ADL.
```

### DIFF
--- a/docs/advanced/cast/custom.rst
+++ b/docs/advanced/cast/custom.rst
@@ -31,63 +31,81 @@ The following Python snippet demonstrates the intended usage from the Python sid
 
     print(A())
 
-To register the necessary conversion routines, it is necessary to add an
-instantiation of the ``pybind11::detail::type_caster<T>`` template.
-Although this is an implementation detail, adding an instantiation of this
-type is explicitly allowed.
+To register the necessary conversion routines, it is necessary to define a
+caster class, and register it with the other pybind11 casters:
 
 .. code-block:: cpp
 
-    namespace pybind11 { namespace detail {
-        template <> struct type_caster<inty> {
-        public:
-            /**
-             * This macro establishes the name 'inty' in
-             * function signatures and declares a local variable
-             * 'value' of type inty
-             */
-            PYBIND11_TYPE_CASTER(inty, const_name("inty"));
-
-            /**
-             * Conversion part 1 (Python->C++): convert a PyObject into a inty
-             * instance or return false upon failure. The second argument
-             * indicates whether implicit conversions should be applied.
-             */
-            bool load(handle src, bool) {
-                /* Extract PyObject from handle */
-                PyObject *source = src.ptr();
-                /* Try converting into a Python integer value */
-                PyObject *tmp = PyNumber_Long(source);
-                if (!tmp)
-                    return false;
-                /* Now try to convert into a C++ int */
-                value.long_value = PyLong_AsLong(tmp);
-                Py_DECREF(tmp);
-                /* Ensure return code was OK (to avoid out-of-range errors etc) */
-                return !(value.long_value == -1 && !PyErr_Occurred());
-            }
-
-            /**
-             * Conversion part 2 (C++ -> Python): convert an inty instance into
-             * a Python object. The second and third arguments are used to
-             * indicate the return value policy and parent object (for
-             * ``return_value_policy::reference_internal``) and are generally
-             * ignored by implicit casters.
-             */
-            static handle cast(inty src, return_value_policy /* policy */, handle /* parent */) {
-                return PyLong_FromLong(src.long_value);
-            }
-        };
-    }} // namespace pybind11::detail
+    struct inty_type_caster {
+    public:
+        /**
+         * This macro establishes the name 'inty' in
+         * function signatures and declares a local variable
+         * 'value' of type inty
+         */
+        PYBIND11_TYPE_CASTER(inty, const_name("inty"));
+        /**
+         * Conversion part 1 (Python->C++): convert a PyObject into a inty
+         * instance or return false upon failure. The second argument
+         * indicates whether implicit conversions should be applied.
+         */
+        bool load(handle src, bool) {
+            /* Extract PyObject from handle */
+            PyObject *source = src.ptr();
+            /* Try converting into a Python integer value */
+            PyObject *tmp = PyNumber_Long(source);
+            if (!tmp)
+                return false;
+            /* Now try to convert into a C++ int */
+            value.long_value = PyLong_AsLong(tmp);
+            Py_DECREF(tmp);
+            /* Ensure return code was OK (to avoid out-of-range errors etc) */
+            return !(value.long_value == -1 && !PyErr_Occurred());
+        }
+        /**
+         * Conversion part 2 (C++ -> Python): convert an inty instance into
+         * a Python object. The second and third arguments are used to
+         * indicate the return value policy and parent object (for
+         * ``return_value_policy::reference_internal``) and are generally
+         * ignored by implicit casters.
+         */
+        static handle cast(inty src, return_value_policy /* policy */, handle /* parent */) {
+            return PyLong_FromLong(src.long_value);
+        }
+    };
 
 .. note::
 
-    A ``type_caster<T>`` defined with ``PYBIND11_TYPE_CASTER(T, ...)`` requires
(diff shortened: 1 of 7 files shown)


docs/advanced/cast/custom.rst
```diff
@@ -31,63 +31,86 @@ The following Python snippet demonstrates the intended usage from the Python sid
...
+  is to *declare* a function named ``pybind11_select_caster``, its only purpose
+  is to associate the C++ type with its caster class:
+
+  .. code-block:: cpp
+
+      struct inty_type_caster;
+
+      struct inty {
+        ...
+        friend inty_type_caster pybind11_select_caster(inty*);
```
rwgk, line 112: This looks like the feature is intrusive on the wrapped C++ code, it is not immediately obvious that that's a choice. I think the previous version of the doc was actually better, because it makes it clear that `pybind11_select_caster` is usually only in the pybind11 bindings code (not user code). Maybe adding a note could nice:

Note that `pybind11_select_caster` can also be implemented as a `friend` function if that is preferred.

rwgk (approved):

docs/advanced/cast/custom.rst
laramiel: I think that it's preferred to use it as a friend function. But it is permitted to be used as a non-friend function.

rwgk: > I think that it's preferred to use it as a friend function.

My main worry is that people walk away with the wrong impression that this feature is intrusive on their C++ code, and therefore don't use `pybind11_select_caster` at all. With the doc as is, I'd bet that's fairly likely to happen.

henryiii: I agree. Maybe a small comment stating it is not required if you can use only public variables?

docs/advanced/cast/custom.rst
```diff
@@ -31,63 +31,86 @@ The following Python snippet demonstrates the intended usage from the Python sid
...
+            PyObject *source = src.ptr();
+            /* Try converting into a Python integer value */
+            PyObject *tmp = PyNumber_Long(source);
+            if (!tmp)
+                return false;
+            /* Now try to convert into a C++ int */
+            value.long_value = PyLong_AsLong(tmp);
+            Py_DECREF(tmp);
+            /* Ensure return code was OK (to avoid out-of-range errors etc) */
+            return !(value.long_value == -1 && !PyErr_Occurred());
```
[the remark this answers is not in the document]
henryiii: Would be a good followup, unless you want to suggest the update inline. It might be elsewhere?

henryiii (approved):
A couple of very minor nits would be nice to address, but looks pretty safe and useful.

tests/test_make_caster_adl.cpp
```diff
@@ -0,0 +1,92 @@
...
+struct unnamed_ns_type_mock {};
+struct unnamed_ns_mock_caster {
+    static int num() { return 303; }
+};
+unnamed_ns_mock_caster pybind11_select_caster(unnamed_ns_type_mock *);
+#if !defined(_MSC_VER)
+// Dummy implementation, purely to avoid compiler warnings (unused function).
+// Easier than managing compiler-specific pragmas for warning suppression.
+// MSVC happens to not generate any warnings. Leveraging that to prove that
+// this test actually works without an implementation.
```
henryiii, line 33: Maybe mention `[[maybe_unused]]` for C++17? If we ever go C++17+ only, we can then just use that.

rwgk:
@laramiel 
@amauryfa
@henryiii 
@Skylion007 

I revised custom.rst top to bottom. Could you please review?

I only changed custom.rst, the rest is as before.

I kept example code changes to a minimum, but did address @Skylion007's concern (`PyErr_Clear()`).

docs/advanced/cast/custom.rst
```diff
@@ -1,93 +1,135 @@
...
+  to discover the association.
+
+  Note that ``pybind11_select_caster`` can alternatively be declared as a ``friend``
+  function of the C++ type, if that is practical and preferred:
+
+  .. code-block:: cpp
+
+      struct inty {
+          ...
+          friend inty_type_caster pybind11_select_caster(inty*);
```
amauryfa, line 113: Does it work? the type `inty_type_caster` probably cannot be defined before this declaration.

docs/advanced/cast/custom.rst
```diff
@@ -1,93 +1,135 @@
...
+  .. note::
+      ``type_caster` specializations may be full (as in this simple example) or partial.
 
-.. warning::
+  .. warning::
 
-    When using custom type casters, it's important to declare them consistently
-    in every compilation unit of the Python extension module. Otherwise,
-    undefined behavior can ensue.
+      When using this method, it is important to declare the specializations
```
amauryfa, line 131: Unfortunately, the ADL method does not solve the ODR issue.
For example, for a given `T` there can be only one definition of `load_caster<T>(const handle&)` (defined below), so `make_caster<T>` must be consistent across all modules.

docs/advanced/cast/custom.rst
laramiel: It would have to be forward declared:

```
class inty_type_caster;

struct inty {
  ...
  friend inty_type_caster pybind11_select_caster(inty*);
};

```

tests/test_make_caster_adl.cpp
rwgk: Thanks for the suggestion! We already have `PYBIND11_MAYBE_UNUSED`, trying that now in the CI.

rwgk:
@laramiel @amauryfa @henryiii @Skylion007

Thanks Amaury and Laramie for the feedback! And Henry for the `[[maybe_unused]]` hint from 3 weeks ago. I think I addressed everything now, and beyond, by porting the rst code blocks to the new tests/test_docs_advanced_cast_custom.cpp, making a pass revising the source code comments, clang-formatting, then copying back from the test to the rst file.

I also renamed test_make_caster_adl to test_select_caster, to make it more clear what is tested. (I tried to use test_pybind11_select_caster but that leads to a name clash, because test_ is stripped off for the submodule name.) 

Regrading test_docs_advanced_cast_custom.cpp, the three alternatives shown in the documentation are handled with a `#define`:

```
// 0: Use the older pybind11::detail::type_caster<T> specialization mechanism.
// 1: Use pybind11_select_caster declaration in user_space.
// 2: Use pybind11_select_caster friend function declaration.
#ifndef PYBIND11_TEST_DOCS_ADVANCED_CAST_CUSTOM_ALTERNATIVE
#    define PYBIND11_TEST_DOCS_ADVANCED_CAST_CUSTOM_ALTERNATIVE 1
#endif
```

It would be ideal to exercise all 3 alternatives, but I'm not sure how to best do that. I'm not even sure if it's worth the trouble, the extra coverage is just a few lines of C++. I'm thinking of leaving that for "later, maybe". What's in this PR is already clearly beyond the status quo (rst code blocks not tested at all; outdated formatting).

rwgk:
Putting this PR into Draft mode. The rationale:

* The primary motivation behind the original PR #3643 was to add support for translation-unit-specific casters.
* But the idea that `pybind11_select_caster` could be a stepping stone towards TU-specific casters was conclusively laid to rest by what we learned from the experiments under #3931.
* We are still looking into another idea, which is not very concrete at the moment (it involves leveraging the `typename... Extra` mechanism).
* To not make the starting point for that work more complicated, we think it is best to hold off merging `pybind11_select_caster` for now.

Let's wait and see where the new idea leads, and reevaluate adding `pybind11_select_caster` when there is more certainty.

rwgk:
2+ years have passed: Closing.

(It would be great to rescue the custom.rst edits.)

PR #3866, opened 2022-04-12
Adds set_name method of pybind11::capsule class

oleksandr-pavlyk:
This calls `PyCapsule_SetName` on the underlying Python object.

## Description

Definition of `pybind11::capsule` class was missing `set_name` method.

This is useful if using pybind11 to implement DLPack protocol which [requires](https://dmlc.github.io/dlpack/latest/python_spec.html#implementation) to rename consumed capsule.

## Suggested changelog entry:

```rst
pybind11::capsule::set_name added to mutate the name of the capsule instance.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1639,9 +1639,7 @@ public:
         }
     }
 
-    const char *name() const {
-        return get_name_or_throw(m_ptr);
-    }
+    const char *name() const { return get_name_or_throw(m_ptr); }
 
     /// Replaces a capsule's name *without* calling the destructor on the existing one.
     void set_name(const char *new_name) {
@@ -1649,13 +1647,14 @@ public:
             throw error_already_set();
         }
     }
+
 private:
-    static const char * get_name_or_throw(PyObject *o) {
+    static const char *get_name_or_throw(PyObject *o) {
         const char *name = PyCapsule_GetName(o);
         if (name == nullptr && PyErr_Occurred()) {
-	    throw error_already_set();
+            throw error_already_set();
         }
-	return name;
+        return name;
     }
 };
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1588,7 +1588,7 @@ public:
                 }
                 pybind11_fail("Unable to get capsule context");
             }
-            const char *name = get_name_no_throw(o);
+            const char *name = get_name_in_error_scope(o);
             void *ptr = PyCapsule_GetPointer(o, name);
             if (ptr == nullptr) {
                 throw error_already_set();
@@ -1603,7 +1603,7 @@ public:
 
     explicit capsule(void (*destructor)()) {
         m_ptr = PyCapsule_New(reinterpret_cast<void *>(destructor), nullptr, [](PyObject *o) {
-            const char *name = get_name_no_throw(o);
+            const char *name = get_name_in_error_scope(o);
             auto destructor = reinterpret_cast<void (*)()>(PyCapsule_GetPointer(o, name));
             if (destructor == nullptr) {
                 throw error_already_set();
@@ -1655,7 +1655,7 @@ public:
     }
 
 private:
-    static const char *get_name_no_throw(PyObject *o) {
+    static const char *get_name_in_error_scope(PyObject *o) {
         error_scope error_guard;
 
         const char *name = PyCapsule_GetName(o);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1588,12 +1588,7 @@ public:
                 }
                 pybind11_fail("Unable to get capsule context");
             }
-            const char *name = PyCapsule_GetName(o);
-            if (name == nullptr) {
-                if (PyErr_Occurred()) {
-                    throw error_already_set();
-                }
-            }
+            const char *name = get_name_or_throw(o);
             void *ptr = PyCapsule_GetPointer(o, name);
             if (ptr == nullptr) {
                 throw error_already_set();
@@ -1608,12 +1603,7 @@ public:
 
     explicit capsule(void (*destructor)()) {
         m_ptr = PyCapsule_New(reinterpret_cast<void *>(destructor), nullptr, [](PyObject *o) {
-            const char *name = PyCapsule_GetName(o);
-            if (name == nullptr) {
-                if (PyErr_Occurred()) {
-                    throw error_already_set();
-                }
-            }
+            const char *name = get_name_or_throw(o);
             auto destructor = reinterpret_cast<void (*)()>(PyCapsule_GetPointer(o, name));
             if (destructor == nullptr) {
                 throw error_already_set();
@@ -1650,12 +1640,7 @@ public:
     }
 
     const char *name() const {
-        const char *name = PyCapsule_GetName(m_ptr);
-        if (name == nullptr) {
-            if (PyErr_Occurred()) {
-                throw error_already_set();
-            }
-        }
+        return get_name_or_throw(m_ptr);
     }
 
     /// Replaces a capsule's name *without* calling the destructor on the existing one.
@@ -1663,6 +1648,14 @@ public:
         if (PyCapsule_SetName(m_ptr, new_name) != 0) {
             throw error_already_set();
         }
+    }
+private:
+    static const char * get_name_or_throw(PyObject *o) {
+        const char *name = PyCapsule_GetName(o);
+        if (name == nullptr && PyErr_Occurred()) {
+	    throw error_already_set();
+        }
+	return name;
     }
 };
 
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1649,7 +1649,7 @@ public:
     }
 
 private:
-    static const char * get_name_or_throw(PyObject *o) {
+    static const char *get_name_or_throw(PyObject *o) {
         /* an exception may be in-flight, we must save it in case we create another one */
         PyObject *type, *value, *traceback;
         PyErr_Fetch(&type, &value, &traceback);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1588,7 +1588,7 @@ public:
                 }
                 pybind11_fail("Unable to get capsule context");
             }
-            const char *name = get_name_or_throw(o);
+            const char *name = get_name_no_throw(o);
             void *ptr = PyCapsule_GetPointer(o, name);
             if (ptr == nullptr) {
                 throw error_already_set();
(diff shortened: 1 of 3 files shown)


include/pybind11/pytypes.h
```diff
@@ -1638,6 +1638,13 @@ class capsule : public object {
     }
 
     const char *name() const { return PyCapsule_GetName(m_ptr); }
+
+    /// Replaces a capsule's name *without* calling the destructor on the existing one.
+    void set_name(const char *&new_name) {
```
Skylion007, line 1643: Why `const char *&new_name` instead of just `const char *`?
```suggestion
    void set_name(const char *new_name) {
```

Also: 
"If non-NULL, the name must outlive the capsule. If the previous name stored in the capsule was not NULL, no attempt is made to free it."

Shouldn't we try to free the old name? In our `set_name` function?

Skylion007, line 1643: Why `const char *&new_name` instead of just `const char *`?
```suggestion
    void set_name(const char *new_name) {
```

Also: 
"If non-NULL, the name must outlive the capsule. If the previous name stored in the capsule was not NULL, no attempt is made to free it."

Shouldn't we try to free the old name? In our `set_name` function?

oleksandr-pavlyk: No, usually the names are `static const char *`, so we should not attempt to free it.

Re: why `const char *&new_name` instead of `const char *` was to avoid copying the value of the pointer. If you think I am overthinking this, I am OK with using `const char *` instead.

Skylion007: Any decent compiler should optimize this out.

The arg would still have to be copied for PyCapsule_SetName as that takes a `const char *` so this wouldn't be a useful optimization, it just makes the method overly complex.

Skylion007: @oleksandr-pavlyk Actually, I just realized that the default destructors need to be changed for the explicit constructor since some of the make the assumption that the name of the capsule ("null") doesn't change, where not it could in those scenarios. They need to call name() for the calls to PyCapsule_GetPointer

Skylion007: Or better yet, just call `PyCapsule_GetName`

rwgk: My thinking: Unless the function is inlined, the machine code will have to put *something* on the process stack. AFAIK it cannot get any better than a pointer. A reference is just a syntax trick, but ultimately it's still a pointer underneath.

But maybe more importantly: what if the capsule deleter manages the name, without storing the original char pointer elsewhere? — Obviously that cannot be solved here, the Python C API is fundamentally unsafe. But it might be good to hint on it in the comment. Trying:

/// Thin wrapper for PyCapsule_SetName. As with the Python C API call, it is the responsibility of the caller to ensure that the lifetime of the old and new pointees is managed properly.

Skylion007: @rwgk The Capsule Deleter can just retrieve the name use PyCapsule_GetName(). Currently, we have it hard coded for two destructors we defined.

Skylion007:
Okay, we just need to fix our custom destructors with PyCapsule_GetName() instead of assuming the name is null.

rwgk:
> Okay, we just need to fix our custom destructors with PyCapsule_GetName() instead of assuming the name is null.

I think it's too complicated, trying to establish a safe context for managing the lifetime of name. Maybe even impossible, because we don't name who owns the passed name (is it static, or system malloc, from a pool, etc.). To not turn this into something big, I'd just add the thin wrapper (as-is) with the suggested comment.

include/pybind11/pytypes.h
```diff
@@ -1588,7 +1588,7 @@ class capsule : public object {
                 }
                 pybind11_fail("Unable to get capsule context");
             }
-            void *ptr = PyCapsule_GetPointer(o, nullptr);
+            void *ptr = PyCapsule_GetPointer(o, PyCapsule_GetName(o));
```
Skylion007, line 1591: PyCapsule_GetName can throw exceptins so the error handling is a little more subtle here. I'll go and fix it when I have time.

Skylion007 (changes requested):

include/pybind11/pytypes.h
rwgk: Oh, before I wasn't aware of the fact that we need to pass the name here. I think I misunderstood.

Now looking around, I see there is already similar code in `get_pointer`.
That uses `this->name();` without checking error conditions as documented here:

https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_GetName

This looks like an existing bug.

Current thinking:
* add error handling to the `name()` implementation.
* re-use the thus indirectly fixed-up `get_pointer` implementation here and below.

oleksandr-pavlyk: I just pushed a change to add error handling in `name()` method.

include/pybind11/pytypes.h
```diff
@@ -1637,7 +1649,21 @@ class capsule : public object {
         }
     }
 
-    const char *name() const { return PyCapsule_GetName(m_ptr); }
+    const char *name() const {
+        const char *name = PyCapsule_GetName(m_ptr);
+        if (name == nullptr) {
+            if (PyError_Occurred()) {
```
rwgk, line 1655:
```
if (name == nullptr && PyErr_Occurred()) {
```

I'm strongly opposed to copy-pasting this much code. If it wasn't for the error handling, fine, but reimplementing `name` three times ... totally no.

If you change to re-use, this looks good to me.

Best practice: we need tests covering the changed code. I think it's pretty straightforward. Do you know where to start?

oleksandr-pavlyk: Sorry will fix this up. And will add tests too. Perhaps this should be converted to a draft until it is ready to minimize distraction.

rwgk: Draft is for before you want feedback.
It's fine with me either way.

rwgk:
Looks good to me but (3.9-dbg (deadsnakes) • Valgrind • x64) ...
```
  what():  RuntimeError: Unable to cast Python instance of type <class 'numpy.ndarray'> to C++ type 'Eigen::Ref<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 0, Eigen::Stride<-1, -1> >'
Fatal Python error: Aborted
```
huh?
(just to say I don't see what's wrong; needs debugging)

include/pybind11/pytypes.h
```diff
@@ -1637,7 +1639,29 @@ class capsule : public object {
...
+    /// Replaces a capsule's name *without* calling the destructor on the existing one.
+    void set_name(const char *new_name) {
+        if (PyCapsule_SetName(m_ptr, new_name) != 0) {
+            throw error_already_set();
+        }
+    }
+
+private:
+    static const char *get_name_or_throw(PyObject *o) {
+        /* an exception may be in-flight, we must save it in case we create another one */
```
rwgk, line 1653: Oh ... your code is great, but that we need it here isn't.

After thinking about it for a moment: it kind-of makes sense that it can happen in a capsule destructor (objects are deleted while exceptions unwind). But then we need to be super careful how we handle cascading errors, we don't want to mask the original error with the `PyCapsule_GetName()` error, that can be very confusing & waste a lot of time debugging (I lost a few days b/o masked errors only a couple months ago). I think to avoid that we need to chain the errors with `raise_from`.

Ideally the complex error handling should only be done from the destructors, not from the plain `name()` member function. It's a bug if `name()` is called while `PyErr_Occurred()` is true, unless it is from a destructor.

This is turning out far more complex than I would have thought. Maybe wait until @Skylion007 can comment on my suggestions before making more changes.

oleksandr-pavlyk: Yes, I can wait, and your suggestions make perfect sense. 

I have pushed a change that initializes variables in `get_name_or_throw` but will wait to give @Skylion007 time to comment.

oleksandr-pavlyk: I think you are absolutely correct, and `name()` should not call `get_name_or_throw`. 

Also the `get_name_or_throw` should use `raise_from` in the clause where call to `PyCapsule_GetName` raised an exception.

I will prepare the change, but will defer pushing it.

rwgk: Please go ahead with the push.
I just didn't want to rush you.
But certainly don't want to stop you from doing what you feel is a good direction.

oleksandr-pavlyk: The [recommended practice](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WriteUnraisable) for handling exceptions raised during execution of destructors is to use `PyErr_WriteUnraisable` passing the object being destroyed as it's argument.

This is simpler and more straightforward that trying to chain exceptions. The issue with chaining exceptions is that since I can only pass string error message to `raise_from`, we loose the traceback of the exception caused by `PyCapsule_Name`.
To turn `value` associated with this exception we need to call `PyObject_Repr` on it, which can raise additional exceptions. 

I am thus giving up on the chaining idea and will use `PyErr_WriteUnraisable` instead.

rwgk: Sounds awesome. I wasn't aware of `PyErr_WriteUnraisable` before. Great to know we're in a situation that's formally covered by the Python C API.

include/pybind11/pytypes.h
```diff
@@ -1637,7 +1639,39 @@ class capsule : public object {
...
+        PyObject *type = nullptr, *value = nullptr, *traceback = nullptr;
+        PyErr_Fetch(&type, &value, &traceback);
```
rwgk, line 1661: We have `error_scope` in detail/common.h. Do you think that could work here? My guess is it does.
(Although now that I look at it, we're currently not following the rule-of-3 in the `error_scope` implementation, but that's best fixed in a separate PR IMO.)

include/pybind11/pytypes.h
oleksandr-pavlyk: I think the right thing to do if `PyCapsule_GetName` raises is to write unraisable, and then simply restore the error in flight (if any) and return the name. 

With this, `get_name_or_throw` should probably be `get_name_no_throw`.

include/pybind11/pytypes.h
oleksandr-pavlyk: Very nice! Thanks for the suggestion @rwkg. It totally belongs here and make code much cleaner too.

include/pybind11/pytypes.h
```diff
@@ -1637,7 +1639,33 @@ class capsule : public object {
...
+
+    /// Replaces a capsule's name *without* calling the destructor on the existing one.
+    void set_name(const char *new_name) {
+        if (PyCapsule_SetName(m_ptr, new_name) != 0) {
+            throw error_already_set();
+        }
+    }
+
+private:
+    static const char *get_name_no_throw(PyObject *o) {
```
rwgk, line 1658: get_name_preserve_error
get_name_in_error_scope
get_name_with_error_scope
...
?

The "no throw" makes sense in the context of the development history of this PR, but not so much without that context.

rwgk (approved):
Thanks Oleksandr!

include/pybind11/pytypes.h
oleksandr-pavlyk: Yes, I like `get_name_in_error_scope`.

Skylion007 (approved):
This is actually a better solution then what I was going to suggest (as we still throw errors for if we can't get the context or the destructor is nullptr). Regardless, we can consider fixing those in a separate PR.

Issue #3858, opened 2022-04-11
[ENH]: Add DLPACK castersto PyBind11 


Skylion007:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

* Recently, nanobind added a [new type of array caster](https://github.com/wjakob/nanobind/commit/5d92f75fabb439b1172a33627e7d33557f4ed9d1) that uses DLPack to convert to and from array. We would like to backport this caster type to pybind. DLPack is a protocol for quickly and efficiently sharing arrays between programs that are backed by either CPU or GPU memory. For example, with a properly written caster one could transfer between a CUDA GPU Array to a PyTorch Tensor. While this is currently possible with custom caster in PyBind11, it would be nice to have official caster support to make this type of generic array sharing easier. Recent version of Numpy also have native APIs to create Numpy arrays from DLPack.

1. This is not a replacement for the Eigen / Numpy bindings, but a supplement to them. Our current bindings support versions of Numpy that are fairly old. Additionally, we have a lot of convenience functions for dealing with numpy arrays that DLPack likely will not support (reshaping, resizing etc.
2. This would be implemented as custom type casters. To use them, the user would need to specify an optional include (similar to how they already need to for Eigen/Numpy Bindings.
3. The testing suite and much of the code can be ripped from nanobind. Note that there are signficant differences between the type casters though: https://github.com/wjakob/nanobind#api-differences . We should ensure that the ported casters handle errors the pybind11 way.

Skylion007:
@oleksandr-pavlyk Are you interested in pursuing this issue? Given the text in PR https://github.com/pybind/pybind11/pull/3866 seems like you already are working on an implementation, it would be nice to have official DLPack casters supported in pybind11.

oleksandr-pavlyk:
I would love to.

PeterDykas:
Sorry if I am reading this wrong but it seems like there may have been support added for DLPack but I was struggling to find information on it in the docs.  Was the support added for DLPack tensors and is there anywhere you could point me to for finding out more?

[...]

PR #3868, opened 2022-04-13
Expand dtype accessors

oleksandr-pavlyk:
## Description

Added explicit constructor for `py::dtype` from type number. Also added accessor to retrieve the `type_num` field of the struct.

While at it, added other accessors. 

Added tests.

## Suggested changelog entry:

```rst
dtype constructor from type number added, accessors corresponding to Python API `dtype.num`, `dtype.byteorder`, `dtype.flags` and `dtype.alignment` added.
```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -562,6 +562,14 @@ public:
         m_ptr = from_args(std::move(args)).release().ptr();
     }
 
+    explicit dtype(int typenum) {
+        if (auto *ptr = detail::npy_api::get().PyArray_DescrFromType_(typenum)) {
+            m_ptr = ptr;
+        } else {
+            throw error_already_set();
+        }
+    }
+
     /// This is essentially the same as calling numpy.dtype(args) in Python.
     static dtype from_args(object args) {
         PyObject *ptr = nullptr;
@@ -595,6 +603,23 @@ public:
         // C API (``PyArray_Descr::type``).
         return detail::array_descriptor_proxy(m_ptr)->type;
     }
+
+    /// type number of dtype.
+    ssize_t num() const {
+        // Note: The signature, `dtype::num` follows the naming of NumPy's public
+        // Python API (i.e., ``dtype.num``), rather than its internal
+        // C API (``PyArray_Descr::type_num``).
+        return detail::array_descriptor_proxy(m_ptr)->type_num;
+    }
+
+    /// Single character for byteorder
+    char byteorder() const { return detail::array_descriptor_proxy(m_ptr)->byteorder; }
+
+    /// Single character for byteorder
+    int alignment() const { return detail::array_descriptor_proxy(m_ptr)->alignment; }
+
+    /// Flags for the array descriptor
+    char flags() const { return detail::array_descriptor_proxy(m_ptr)->flags; }
 
 private:
     static object _dtype_from_pep3118() {
--- a/tests/test_numpy_dtypes.cpp
+++ b/tests/test_numpy_dtypes.cpp
@@ -291,6 +291,7 @@ py::list test_dtype_ctors() {
     list.append(py::dtype(names, formats, offsets, 20));
     list.append(py::dtype(py::buffer_info((void *) 0, sizeof(unsigned int), "I", 1)));
     list.append(py::dtype(py::buffer_info((void *) 0, 0, "T{i:a:f:b:}", 1)));
+    list.append(py::dtype(py::detail::npy_api::NPY_DOUBLE_));
     return list;
 }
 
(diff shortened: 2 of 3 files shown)


include/pybind11/numpy.h
```diff
@@ -596,6 +604,23 @@ class dtype : public object {
...
+        // Note: The signature, `dtype::num` follows the naming of NumPy's public
+        // Python API (i.e., ``dtype.num``), rather than its internal
+        // C API (``PyArray_Descr::type_num``).
+        return detail::array_descriptor_proxy(m_ptr)->type_num;
+    }
+
+    /// Single character for byteorder
+    char byteorder() const { return detail::array_descriptor_proxy(m_ptr)->byteorder; }
+
+    /// Single character for byteorder
```
rwgk, line 618: Oversight? (copy-paste forgot to modify?)

rwgk (approved):
Nice, thanks!

include/pybind11/numpy.h
oleksandr-pavlyk: Fixed, thanks for spotting this.

Skylion007 (approved):

include/pybind11/numpy.h
```diff
@@ -562,6 +562,14 @@ class dtype : public object {
         m_ptr = from_args(std::move(args)).release().ptr();
     }
 
+    explicit dtype(int typenum) {
+        if (auto *ptr = detail::npy_api::get().PyArray_DescrFromType_(typenum)) {
+            m_ptr = ptr;
+        } else {
+            throw error_already_set();
+        }
+    }
```
Skylion007, line 571:
```suggestion
    explicit dtype(int typenum) : object(detail::npy_api::get().PyArray_DescrFromType_(typenum), stolen_t{}){
        if (m_ptr == nullptr) {
            throw error_already_set();
        }
    }

Comment (Skylion007) [reply include/pybind11/numpy.h:571]: See pytypes.h for how we usually handle errors in constructors like this.

**Comment by oleksandr-pavlyk** [comment]
Failures are due to error while cloning the Eigen repo.

**Comment by Skylion007** [comment]
Actually, now that it's merged, not sure the new ctor is really necessary. Wouldn't
```
dtype::from_args(py::int_(typenum))
```
have worked?

**Comment by oleksandr-pavlyk** [comment]
Tried that in `test_numpy_dtype.cpp`:

```
~/pybind11/tests/test_numpy_dtypes.cpp: In function ‘pybind11::list test_dtype_ctors()’:
~/pybind11/tests/test_numpy_dtypes.cpp:294:68: error: no matching function for call to ‘pybind11::int_::int_(pybind11::detail::npy_api::constants
)’
  294 |     list.append(py::dtype(py::int_(py::detail::npy_api::NPY_DOUBLE_)));
      |                                                                    ^                                                                                                 I
``` 

but even if it worked, the need to cast native `int` into `py::int_` would have made it less efficient, so the new ctor is still beneficial.

**Comment by oleksandr-pavlyk** [comment]
Actually, since `py::from_args` is equivalent of calling `np.dtype`, it dos not work:

```
In [2]: np.dtype(12)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-dea73ed57aec> in <module>
----> 1 np.dtype(12)

TypeError: Cannot interpret '12' as a data type
```

In fact modifying the test file:

```diff
diff --git a/tests/test_numpy_dtypes.cpp b/tests/test_numpy_dtypes.cpp
index 7de36f2f..8ed6ff8d 100644
--- a/tests/test_numpy_dtypes.cpp
+++ b/tests/test_numpy_dtypes.cpp
@@ -291,7 +291,7 @@ py::list test_dtype_ctors() {
     list.append(py::dtype(names, formats, offsets, 20));
     list.append(py::dtype(py::buffer_info((void *) 0, sizeof(unsigned int), "I", 1)));
     list.append(py::dtype(py::buffer_info((void *) 0, 0, "T{i:a:f:b:}", 1)));
-    list.append(py::dtype(py::detail::npy_api::NPY_DOUBLE_));
+    list.append(py::dtype::from_args(py::int_(static_cast<int>(py::detail::npy_api::NPY_DOUBLE_))));
     return list;
 }

```

and running raises the same error.

Even if this worked, it would still be less efficient that the added ctor, since the argument Python object needs to be constructed.

Skylion007:
I am just wondering if these is a less ambiguous way to build that ctor. Perhaps we could have it take an enum instead of an int or something along those lines? Or maybe it would make more sense to have it be a static factory method. It's a bit weird for the dtype to have an int constructor as it stands currently. 

Right now it's a public constructor that could only be used via an implementation detail enum, which is actually really weird. Thoughts @henryiii @rwgk ?

Skylion007:
I am just having second thoughts about having this ctor be a part of the public API for dtypes considering the values that are used for it are themselves not public.

oleksandr-pavlyk:
Yes, agreed, the non-public nature of type numbers has bugged me too, but I dismissed it since it is easy to `#include "numpy/ndarraytypes.h"`, or use `py::detail`.

Also the exposure of `num()` accessor allows for use of private enums, `dtype(dt.num())`.

Skylion007:
@oleksandr-pavlyk We can revisit this PR in connection to casters we design for https://github.com/pybind/pybind11/issues/3858

PR #3869, opened 2022-04-14
chore: simplify numpy dtype ctor

Skylion007:
## Description
When reviewing another PR #3868, I noticed one of the ctors could be simplified using ctor delegation so I went ahead an added a tiny PR to do that.
Also removed a few python 2 compat macros in numpy.h


oleksandr-pavlyk (approved):

rwgk (approved):

rwgk:
post-merge CI failure is a known flake

PR #3874, opened 2022-04-16
Use py::module instead of py::module_ in example code.

gsmecher:
The changelog for v2.6 explains the reasons for py::module_, but also states "qualified usage is unaffected." User code should probably make use of py::module, since it is (1) uncluttered by the underscore, and (2) backwards compatible with pre-v2.6 builds.

## Description

Example code uses py::module_, which is neither backwards compatible, nor necessary. The underscore is an internal collision that needn't leak out.

## Suggested changelog entry:

```rst
Use py::module instead of py::module_ in example code.
```

### DIFF
--- a/docs/advanced/embedding.rst
+++ b/docs/advanced/embedding.rst
@@ -108,11 +108,11 @@ The two approaches can also be combined:
 Importing modules
 =================
 
-Python modules can be imported using ``module_::import()``:
+Python modules can be imported using ``module::import()``:
 
 .. code-block:: cpp
 
-    py::module_ sys = py::module_::import("sys");
+    py::module sys = py::module::import("sys");
     py::print(sys.attr("path"));
 
 For convenience, the current working directory is included in ``sys.path`` when
@@ -129,12 +129,12 @@ embedding the interpreter. This makes it easy to import local Python files:
 
 .. code-block:: cpp
 
-    py::module_ calc = py::module_::import("calc");
+    py::module calc = py::module::import("calc");
     py::object result = calc.attr("add")(1, 2);
     int n = result.cast<int>();
     assert(n == 3);
 
-Modules can be reloaded using ``module_::reload()`` if the source is modified e.g.
+Modules can be reloaded using ``module::reload()`` if the source is modified e.g.
 by an external process. This can be useful in scenarios where the application
 imports a user defined data processing script which needs to be updated after
 changes by the user. Note that this function does not reload modules recursively.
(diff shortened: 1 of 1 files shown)


henryiii:
We are several releases past 2.6 (2.7, 2.8, 2.9), all of which have the same requirements and ABI compatibility, so I don't see much reason to be backward compatible with 2.5 in 2.9+ docs. Also, anything before 2.9.2 will segfault or crash on PyPy 3.9. so users should be upgrading.

The internal type _is_ `py::module_`, and `py::module` is just a convenience for backward compatibility. `module` is now a soft keyword in C++, and could cause issues - for example, users might ignore our advice and put `using pybind11`, which unqualifies this and can collide with the C++20 soft keyword. Also the type name will be `module_` if you print it out in C++; that would be surprising if you thought it really was `module`. We use a trailing underscore if we collide with keywords other places (like `class_`), so it's not too bad. C++ (partially) took away module and gave it a new meaning.

I'm mildly in favor of keeping it truthful vs. pretty. But not strongly against it if someone else disagrees. We had already decided on making this change a year or two ago, so I don't think other maintainers will be likely to disagree.

I might run the CI anyway just to see if Python 3.11a7 is fine, that just came to GHA.

henryiii:
If there isn't one, a little side note that `py::module` is allowed instead of `py::module_` as long as you use the qualified name, or something like that, might be acceptable, though.

henryiii:
Ahh, this is targeting stable. We do not target stable with PRs, even to docs.

gsmecher:
Oops - I did not intend to target stable.

In either case, thanks for the thoughtful answer. I bumped into an older version of pybind11 using Github's CI runner but am well ahead of any backwards compatibility problems locally. I have no strong opinions here, especially if it's already been settled once.

henryiii:
Don't know what's up with the CI, don't think it likes targeting stable first. I also think we might have to add a label to get the 3.11 dev runs to turn on, so maybe this won't work as a quick test of the 3.11a7 release. Might try once more, since this PR does't change any code, it's kind of a nice test bed for it.

I think if you click "edit this page" or something like that, RtD drops you into a PR against stable. I mistakenly merged one once. 🤦 

Anyway, if there's no `py::module` note as mentioned above, I think adding one would be acceptable if you were interested.

gsmecher:
I don't think a 2.6-specific note in the documentation makes sense - 2.6 is pretty far in the rear-view mirror, as you pointed out, and a note would just add cruft for most of us. I will close the issue.

Thank you for your efforts! I rely heavily on pybind11, and it lets my grasp vastly exceed my reach.

PR #3881, opened 2022-04-18
chore: enable clang-tidy check modernize-use-nullptr

Skylion007:
## Description
* I noticed we didn't have the check modernize-use-nullptr enabled. This replaces usages of the NULL macro or 0 with nullptr when appropriate. This makes the code a bit more readable, modern, and standard compliant.
* I also sorted some checks that were out of order.

## Suggested changelog entry:

```rst
* Enable modernize-use-nullptr clang-tidy checks
```


henryiii (approved):
Wow, could have sworn we had this one, it's usually one of the first ones I add.

PR #3886, opened 2022-04-19
Support frozenset, tuple as dict keys

ecatmur:
## Description

https://github.com/pybind/pybind11/discussions/3836

Add frozenset as a pybind11 type.
Allow type_caster<const T> to be (explicitly) selected and specialized, defaulting to (inherit from) type_caster<T>.
Use type_caster<const K> for std::set, std::map etc. key converter.
Specialize type_caster<const C> for std::set, std::vector etc.
Add tests.

## Suggested changelog entry:

```rst
`std::map<std::set<int>, int>` now converts to/from `Dict[FrozenSet[int], int]`; likewise `std::set<std::vector<int>>` to/from `Set[Tuple[int, ...]]`.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1784,14 +1784,11 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class set : public object {
+class set_base : public object {
+protected:
+    PYBIND11_OBJECT(set_base, object, PyAnySet_Check)
+
 public:
-    PYBIND11_OBJECT_CVT(set, object, PySet_Check, PySet_New)
-    set() : object(PySet_New(nullptr), stolen_t{}) {
-        if (!m_ptr) {
-            pybind11_fail("Could not allocate set object!");
-        }
-    }
     size_t size() const { return (size_t) PySet_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     template <typename T>
@@ -1802,6 +1799,26 @@ public:
     template <typename T>
     bool contains(T &&val) const {
         return PySet_Contains(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr()) == 1;
+    }
+};
+
+class set : public set_base {
+public:
+    PYBIND11_OBJECT_CVT(set, set_base, PySet_Check, PySet_New)
+    set() : set_base(PySet_New(nullptr), stolen_t{}) {
+        if (!m_ptr) {
+            pybind11_fail("Could not allocate set object!");
+        }
+    }
+};
+
+class frozenset : public set_base {
+public:
+    PYBIND11_OBJECT_CVT(frozenset, set_base, PyFrozenSet_Check, PyFrozenSet_New)
+    frozenset() : set_base(PyFrozenSet_New(nullptr), stolen_t{}) {
+        if (!m_ptr) {
+            pybind11_fail("Could not allocate frozenset object!");
+        }
     }
 };
 
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -94,6 +94,34 @@ def test_recursive_casting():
     assert z[0].value == 7 and z[1].value == 42
 
 
+def test_frozen_key(doc):
+    """Test that we special-case C++ key types to Python immutable containers, e.g.:
+    std::map<std::set<K>, V> <-> dict[frozenset[K], V]
+    std::set<std::set<T>> <-> set[frozenset[T]]
+    std::set<std::vector<T>> <-> set[tuple[T, ...]]
+    """
+    s = m.cast_set_map()
+    assert s == {frozenset({"key1", "key2"}): "value"}
+    s[frozenset({"key3"})] = "value2"
+    assert m.load_set_map(s)
+    assert doc(m.cast_set_map) == "cast_set_map() -> Dict[FrozenSet[str], str]"
+    assert doc(m.load_set_map) == "load_set_map(arg0: Dict[FrozenSet[str], str]) -> bool"
+
+    s = m.cast_set_set()
+    assert s == {frozenset({"key1", "key2"})}
+    s.add(frozenset({"key3"}))
+    assert m.load_set_set(s)
+    assert doc(m.cast_set_set) == "cast_set_set() -> Set[FrozenSet[str]]"
+    assert doc(m.load_set_set) == "load_set_set(arg0: Set[FrozenSet[str]]) -> bool"
+
+    s = m.cast_vector_set()
+    assert s == {(1, 2)}
+    s.add((3,))
+    assert m.load_vector_set(s)
+    assert doc(m.cast_vector_set) == "cast_vector_set() -> Set[Tuple[int, ...]]"
+    assert doc(m.load_vector_set) == "load_vector_set(arg0: Set[Tuple[int, ...]]) -> bool"
+
+
 def test_move_out_container():
     """Properties use the `reference_internal` policy by default. If the underlying function
     returns an rvalue, the policy is automatically changed to `move` to avoid referencing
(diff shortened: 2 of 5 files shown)


include/pybind11/pytypes.h
```diff
@@ -1784,14 +1784,11 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class set : public object {
+class set_base : public object {
+protected:
```
Skylion007, line 1788: I don't like having add as a public function on frozen-sets. We can still call the C-API directly on the pointer for casters, but other code shouldn't be able to add objects to frozen sets.

Skylion007:
The frozen set casters are a decent idea, but I don't know how I feel about the tuple ones considering we have std::tuple to fulfill that role already. Thoughts @rwgk @henryiii ?

Skylion007:
The frozen set casters are a decent idea, but I don't know how I feel about the tuple ones considering we have std::tuple to fulfill that role already. Thoughts @rwgk @henryiii ?

include/pybind11/pytypes.h
ecatmur: moved non-const methods to protected

Skylion007: They would still need to be public for normal set?

Skylion007: @ecatmur Why not just move them to set subclass and use the raw C-API to interact with the frozen set.

ecatmur: Sorry, I get it now. Fixed.

ecatmur:
> The frozen set casters are a decent idea, but I don't know how I feel about the tuple ones considering we have std::tuple to fulfill that role already. 

I think the problem is that historically (and currently) Python's tuple has a bit of a mixed role: it's the container you go for when you need an immutable homogeneous sequence, as well as being the heterogeneous sequence type. For example, in typing you can write `Tuple[int, ...]` as well as `Tuple[int, str]`. 

FWIW, when PEP 351 was rejected (over 15 years ago!) Raymond Hettinger made the argument that list->tuple and set->frozenset are the only two transformations you actually ever need: https://mail.python.org/pipermail/python-dev/2005-October/057586.html That suggests a simpler approach would be to adjust list->tuple and set->frozenset after casting before insertion into dict (or set), avoiding the hairiness of `type_caster<type const>` and transparently supports all types casting to list and set (including third party stuff) at a small runtime cost. I'll propose that as an alternative.

jiwaszki (commented):
@Skylion007 @ecatmur I will join as the interested party. Thanks for porting my draft into actual codebase!

I tried the branch and get multiple errors during compilation on my Mac:
> error: ambiguous partial specializations of 'type_caster<const int>' const_name<Const>("FrozenSet[", "Set[") + key_conv::name

Thus I was unable to test it further, however I have some observations.

1. Personally I do not support an idea of the implicit cast from `non-const std::set` to `frozenset`. As I mentioned in https://github.com/pybind/pybind11/discussions/3836 -- there should be a clear line between `set` and it's immutable version as it might confuse Python/pybind users. Ideally functions should only accept dicts coming (or getting out) in form of `std::map<const std::set<...>, ...>`. Is there a way to restrict that?
2. I have also noticed that tests lack some of essential scenarios. What happens in case when code mix const and non-cost types? What happens when arguments are passed by copy?
3. Following @Skylion007 concern about tuples as I have similar thoughts. `std::tuple` is already filling that niche, there is no need to expand in that area. As Python tuple will be converted to `std::vector` it will not only allow appending, but actually allow to modify it's contents -- I do not think this is an expected behaviour. If a user wants to do it, just use `py::tuple` and manually transfer the data to the mutable container.

ecatmur:
> I tried the branch and get multiple errors during compilation on my Mac:
> 
> > error: ambiguous partial specializations of 'type_caster' const_name("FrozenSet[", "Set[") + key_conv::name

Oh, sorry. Maybe #3887 would work? It's a simpler approach but the observable behavior should be the same.

> Thus I was unable to test it further, however I have some observations.
> 
> 1. Personally I do not support an idea of the implicit cast from `non-const std::set` to `frozenset`. As I mentioned in #3836 -- there should be a clear line between `set` and it's immutable version as it might confuse Python/pybind users. Ideally functions should only accept dicts coming (or getting out) in form of `std::map<const std::set<...>, ...>`. Is there a way to restrict that?

There isn't really such a thing as `std::map<const std::set<...>, ...>` - it isn't possible to get a non-const reference to the key in a std::map (not without an illegal const_cast, anyway) so no-one would bother putting `const` on the key. The `value_type` is `std::pair<const Key, Value>` so the keys are const as long as they're in the map.

So there isn't a conversion from non-const std::set to frozenset, it's from `const std::set` to frozenset.

> 2. I have also noticed that tests lack some of essential scenarios. What happens in case when code mix const and non-cost types? What happens when arguments are passed by copy?

I'm only implementing this for the caster interface, not the binder interface, so everything is being passed by copy already and there's no difference between const and non-const at the top level.

> 3. Following @Skylion007 concern about tuples as I have similar thoughts. `std::tuple` is already filling that niche, there is no need to expand in that area. As Python tuple will be converted to `std::vector` it will not only allow appending, but actually allow to modify it's contents -- I do not think this is an expected behaviour. If a user wants to do it, just use `py::tuple` and manually transfer the data to the mutable container.

We already allow passing `tuple` to C++ functions that take a `std::vector` (etc.) - there's one or two tests that show that already, and I've just added some more.

If Python code inserts a tuple key into a `dict[tuple[str, ...], int]` and passes it to C++ code as a `std::map<std::vector<std::string>, int>` then the C++ code can't modify the `std::vector`s (since they're const within the map) and if it removes them, inserts modified keys and passes the map back to Python then they'll be completely distinct objects. There's no way for a std::vector to alias a tuple and allows modifying it.

tests/test_stl.py
```diff
@@ -94,6 +94,38 @@ def test_recursive_casting():
     assert z[0].value == 7 and z[1].value == 42
 
 
+def test_frozen_key(doc):
```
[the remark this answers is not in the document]
ecatmur: Added https://github.com/pybind/pybind11/pull/3886/commits/fcaff44f09b936fb933d55c839863413ce55cc8c that tests you can pass `frozenset` to a C++ function that takes `std::set` (just as you can pass `tuple` to a C++ function taking `std::vector`). Is there's anything else you can do with `frozenset` on its own?

rwgk:
I looked up and down both PRs quite a bit, but didn't get a chance yet to fully appreciate all details. It looks very interesting: not a lot of code, but makes something work that currently doesn't.

Which of the PRs do you prefer? I'd like to focus on the one that you think is best.

Another idea: could you please transfer the additional tests for tuple and frozenset that already work on current HEAD to a separate PR, even if it's a tiny PR? That will make it more clear here what already works and what doesn't. The more immediately improved test coverage is a nice plus.

ecatmur:
> I looked up and down both PRs quite a bit, but didn't get a chance yet to fully appreciate all details. It looks very interesting: not a lot of code, but makes something work that currently doesn't.
> 
> Which of the PRs do you prefer? I'd like to focus on the one that you think is best.

#3887. While this one (#3886) is more idiomatic C++, #3887 is more Pythonic and that's appropriate for code that affects the Python side of things - it's automatically creating a Python API, so should stick to types that would appear naturally in Python programs. The clever recursive stuff would work in the STL binder API, perhaps.

> Another idea: could you please transfer the additional tests for tuple and frozenset that already work on current HEAD to a separate PR, even if it's a tiny PR? That will make it more clear here what already works and what doesn't. The more immediately improved test coverage is a nice plus.

Sure: #3900 is for tuple tests. #3901 splits out the core frozenset functionality (any_set as a common base type, and std::set cast from any_set).

Skylion007:
Let's focus on merging #3901 for now.

PR #3887, opened 2022-04-20
Support frozenset, tuple as dict keys (alternative approach)

ecatmur:
## Description

Alternative to #3886
Inspired by https://github.com/pybind/pybind11/discussions/3836

Add frozenset as a pybind11 type.
Add freeze() function converting set to frozenset and list to tuple; use
it in std::set and std::map casters.
Add tests.

## Suggested changelog entry:

```rst
When used as keys (e.g. in std::map), sequence and set types now convert to tuple and frozenset (instead of unhashable list and set).
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1617,6 +1617,25 @@ object object_api<Derived>::call(Args &&...args) const {
     return operator()<policy>(std::forward<Args>(args)...);
 }
 
+// Convert list -> tuple and set -> frozenset for use as keys in dict, set etc.
+// https://mail.python.org/pipermail/python-dev/2005-October/057586.html
+inline object freeze(object &&obj) {
+    if (isinstance<list>(obj)) {
+        return tuple(std::move(obj));
+    } else if (isinstance<set>(obj)) {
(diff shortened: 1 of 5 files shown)


Skylion007:
I am wondering though if there is anyway we can allow users to extend the 'freeze' function for custom Python types, perhaps using template specialization tricks? Thoughts @rwgk @henryiii?

Skylion007:
Hmm, also I noticed the freeze PEP was rejected as bugprone. Really this is a symptom of us needing to have better specialization and user configuration of key / value converters. Related to the issue we have found in https://github.com/pybind/pybind11/pull/3838

rwgk:
Hi @ecatmur is this ready for re-review? (Or are you still working on it?)

ecatmur:
> Hi @ecatmur is this ready for re-review? (Or are you still working on it?)

Sorry, yeah. I was exploring some alternate approaches (I don't particularly like `frozen_type_name`) but nothing worked out any better.

leslinice:
Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #3900, opened 2022-04-24
Add tests for cast from tuple to sequence containers

ecatmur:
## Description

We already test that tuple can cast to `std::vector` and `std::deque`; add tests for `std::vector<bool>`, `std::list` and
`std::valarray`.

Extracted from #3886.


rwgk:
Thanks Ed!

The one test failure is an infrastructure issue, unrelated to this PR. I'll go ahead with merging.

PR #3901, opened 2022-04-24
Add anyset & frozenset, enable copying (cast) to std::set

ecatmur:
## Description

Add `anyset` & `frozenset`, enable copying (cast) to `std::set`.

For the reverse direction, `std::set` still casts to `set`. This is in concordance with the behavior for sequence containers, where e.g. `tuple` casts to `std::vector`, but `std::vector` casts to `list`.

Extracted from https://github.com/pybind/pybind11/pull/3886.

## Suggested changelog entry:

`anyset` & `frozenset` were added, with copying (cast) to `std::set` (similar to `set`).

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -910,6 +910,8 @@ struct pyobject_caster {
     template <typename T = type, enable_if_t<std::is_same<T, handle>::value, int> = 0>
     pyobject_caster() : value() {}
 
+    // `type` may not be default constructible (e.g. frozenset, anyset).  Initializing `value`
+    // to a nil handle is safe since it will only be accessed if `load` succeeds.
     template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
     pyobject_caster() : value(reinterpret_steal<type>(handle())) {}
 
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -908,6 +908,12 @@ struct handle_type_name<kwargs> {
 template <typename type>
 struct pyobject_caster {
     template <typename T = type, enable_if_t<std::is_same<T, handle>::value, int> = 0>
+    pyobject_caster() : value() {}
+
+    template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
+    pyobject_caster() : value(reinterpret_steal<type>(handle())) {}
+
+    template <typename T = type, enable_if_t<std::is_same<T, handle>::value, int> = 0>
     bool load(handle src, bool /* convert */) {
         value = src;
         return static_cast<bool>(value);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1784,12 +1784,12 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class any_set : public object {
+class anyset : public object {
 protected:
-    PYBIND11_OBJECT(any_set, object, PyAnySet_Check)
+    PYBIND11_OBJECT(anyset, object, PyAnySet_Check)
 
 public:
-    size_t size() const { return (size_t) PySet_Size(m_ptr); }
+    size_t size() const { return static_cast<size_t>(PySet_Size(m_ptr)); }
     bool empty() const { return size() == 0; }
     template <typename T>
     bool contains(T &&val) const {
@@ -1797,10 +1797,10 @@ public:
     }
 };
 
-class set : public any_set {
+class set : public anyset {
 public:
-    PYBIND11_OBJECT_CVT(set, any_set, PySet_Check, PySet_New)
-    set() : any_set(PySet_New(nullptr), stolen_t{}) {
+    PYBIND11_OBJECT_CVT(set, anyset, PySet_Check, PySet_New)
+    set() : anyset(PySet_New(nullptr), stolen_t{}) {
         if (!m_ptr) {
             pybind11_fail("Could not allocate set object!");
         }
@@ -1812,10 +1812,10 @@ public:
     void clear() /* py-non-const */ { PySet_Clear(m_ptr); }
 };
 
-class frozenset : public any_set {
+class frozenset : public anyset {
 public:
-    PYBIND11_OBJECT_CVT(frozenset, any_set, PyFrozenSet_Check, PyFrozenSet_New)
-    frozenset() : any_set(PyFrozenSet_New(nullptr), stolen_t{}) {
+    PYBIND11_OBJECT_CVT(frozenset, anyset, PyFrozenSet_Check, PyFrozenSet_New)
+    frozenset() : anyset(PyFrozenSet_New(nullptr), stolen_t{}) {
         if (!m_ptr) {
             pybind11_fail("Could not allocate frozenset object!");
         }
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1784,24 +1784,41 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class set : public object {
+class set_base : public object {
+protected:
+    PYBIND11_OBJECT(set_base, object, PyAnySet_Check)
+
 public:
-    PYBIND11_OBJECT_CVT(set, object, PySet_Check, PySet_New)
-    set() : object(PySet_New(nullptr), stolen_t{}) {
+    size_t size() const { return (size_t) PySet_Size(m_ptr); }
+    bool empty() const { return size() == 0; }
+    template <typename T>
+    bool contains(T &&val) const {
+        return PySet_Contains(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr()) == 1;
+    }
+};
+
+class set : public set_base {
+public:
+    PYBIND11_OBJECT_CVT(set, set_base, PySet_Check, PySet_New)
+    set() : set_base(PySet_New(nullptr), stolen_t{}) {
         if (!m_ptr) {
             pybind11_fail("Could not allocate set object!");
         }
     }
-    size_t size() const { return (size_t) PySet_Size(m_ptr); }
-    bool empty() const { return size() == 0; }
     template <typename T>
     bool add(T &&val) /* py-non-const */ {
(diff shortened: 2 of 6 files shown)


include/pybind11/pytypes.h
```diff
@@ -1784,24 +1784,41 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class set : public object {
+class any_set : public object {
+protected:
+    PYBIND11_OBJECT(any_set, object, PyAnySet_Check)
+
+public:
+    size_t size() const { return (size_t) PySet_Size(m_ptr); }
```
rwgk, line 1792: While you're at it, could you please change this to `static_cast<size_t>`? (mainly for readability)

include/pybind11/pytypes.h
```diff
@@ -1784,24 +1784,41 @@ class kwargs : public dict {
...
+};
+
+class set : public any_set {
 public:
-    PYBIND11_OBJECT_CVT(set, object, PySet_Check, PySet_New)
-    set() : object(PySet_New(nullptr), stolen_t{}) {
+    PYBIND11_OBJECT_CVT(set, any_set, PySet_Check, PySet_New)
+    set() : any_set(PySet_New(nullptr), stolen_t{}) {
         if (!m_ptr) {
             pybind11_fail("Could not allocate set object!");
```
rwgk, line 1805: Just a remark, not actionable in this PR: `throw error_already_set()` would seem better, similar to the changes @Skylion007 made elsewhere recently. But strictly speaking that's a behavior change, therefore clearly better left for a separate PR IMO.

rwgk (approved):
Looks great, thanks!

rwgk (approved):
This just crossed my mind: could you please also add minimal coverage for the `frozenset` C++ API (constructors, `size`, `empty`, `contains`)?

include/pybind11/pytypes.h
Skylion007: @rwgk No, this is correct to my knowledge. `PySet_New` only throws one type of an exception: a TypeError if the sequence fed into it is not a valid type. The sequence fed into it is a nullptr, so it won't raise that exception. It will return NULL if another unspecified error occurs, but it won't raise an exception to my knowledge: https://docs.python.org/3/c-api/set.html . 
If we add a C++ constructor that takes in an iterable though and passes it into the arg for `PySet_New`, then we should add the `throw error_already_set` in that ctor though.

Skylion007 (changes requested):
The Macro here still is checking against PySet. Needs to be change to PyAnySet. There is also no convert function for AnySet so we should the COMMON MACRO instead.

include/pybind11/pytypes.h
```diff
@@ -1784,24 +1784,41 @@ class kwargs : public dict {
...
     bool add(T &&val) /* py-non-const */ {
         return PySet_Add(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr()) == 0;
     }
     void clear() /* py-non-const */ { PySet_Clear(m_ptr); }
-    template <typename T>
-    bool contains(T &&val) const {
-        return PySet_Contains(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr()) == 1;
+};
+
+class frozenset : public any_set {
```
Skylion007, line 1815: NIt: seems odd to me that it's any_set and frozenset. Shouldn't it be anyset for consistency (I know that's annoyingly not readable, but it is consistent) or frozen_set. Thoughts @rwgk ?

include/pybind11/pytypes.h
```diff
@@ -1784,24 +1784,41 @@ class kwargs : public dict {
...
+public:
+    PYBIND11_OBJECT_CVT(frozenset, any_set, PyFrozenSet_Check, PyFrozenSet_New)
+    frozenset() : any_set(PyFrozenSet_New(nullptr), stolen_t{}) {
```
Skylion007, line 1818: A default ctor doesn't really make sense for frozen set if there is not a good way to add to it. We should instead have a ctor that takes iterable and does appropriate error checking with `throw error_already_set`

Skylion007 (changes requested):
Actually, looking over it I do have some nits before merging.

include/pybind11/pytypes.h
Skylion007: Or actually, if the cvt functions works for converting it from set, we don't need a default ctor at all, do we?

rwgk:
> C-API docs unfortunately seem to indicate it should be "anyset" instead of "any_set" to be consistent: https://docs.python.org/3/c-api/set.html#c.PySet_Contains

I don't have a strong opinion about this one, but usually I try to emulate what I see elsewhere, `anyset` in this case would be my choice.

Regarding the ctor questions, not sure, usually I figure this out by writing unit tests fully covering the public API, then it's easy to see/decide what is useful and practical.

include/pybind11/pytypes.h
ecatmur: Yes, the converting constructors work fine; an iterable ctor would be slightly more efficient but it doesn't seem worth worrying about; anyone who really cares can use the C API anyway.

ecatmur:
Hope I've addressed all concerns.

Note the change to pyobject_caster (to default-initialize with `value` a nil handle) - this should be very slightly more efficient and still safe, since `load` always assigns to `value` (so previously it was being allocated to no purpose). This is necessitated by anyset and frozenset not being default constructible.

include/pybind11/cast.h
```diff
@@ -907,6 +907,12 @@ struct handle_type_name<kwargs> {
 
 template <typename type>
 struct pyobject_caster {
+    template <typename T = type, enable_if_t<std::is_same<T, handle>::value, int> = 0>
+    pyobject_caster() : value() {}
+
+    template <typename T = type, enable_if_t<std::is_base_of<object, T>::value, int> = 0>
```
Skylion007, line 913: I am not sure about the changes on the caster: So without a default ctor, this ctor is needed on pyobject_caster? At the very least, we should have a comment about why there are necessary. I am also worried that there are edge cases where this might break when it is used in downstream applications.

Thoughts @rwgk @henryiii

rwgk: IIUC we have a choice:

* Either the change here.
* Or we need to make `anyset` and `frozenset` default constructible.

> https://github.com/pybind/pybind11/pull/3901#issuecomment-1114261772

I agree but think it is not just "still safe" but actually safer: if the previously default-constructed object is accidentally used the symptoms could be highly confusing (silent failures). A noisy failure (segfault b/o nullptr) seems better.

I'll run our (Google's) global testing with this PR to see if there are breakages and report back here. I might have the results only tomorrow.

rwgk: No breakages! This PR is great. If anyone was previously relying on the default-constructed `value`, they will probably be glad to see that exposed as an obvious bug now.

rwgk (approved):
I tried to enhance the PR description. Please review & correct as needed.

include/pybind11/pytypes.h
```diff
@@ -1784,25 +1784,37 @@ class kwargs : public dict {
     PYBIND11_OBJECT_DEFAULT(kwargs, dict, PyDict_Check)
 };
 
-class set : public object {
+class anyset : public object {
+protected:
```
Skylion007, line 1788: Okay, last concern. Why is the scope here protected? I don't see any other Pytypes doing that.

Skylion007: Specifically, I am worried it might interfere with the static "check_" method that isinstance is suppose to be using. https://github.com/pybind/pybind11/blob/e8e229fa0b486118bf91321b923f8158055c053c/include/pybind11/pytypes.h#L483

rwgk: Good question, it turns out the `protected:` here doesn't do anything, because the `PYBIND11_OBJECT` macro expands to starting with `public:`. See `clang -E` output below.

(Note that there is an `isinstance<anyset>` in stl.h, i.e. this had to be the case somehow.)

I think it's best to not get fancy here and to simply remove the `protected:` line.

```
class anyset : public object {
protected:
    public: [[deprecated("Use reinterpret_borrow<" "anyset" ">() or reinterpret_steal<" "anyset" ">()")]] anyset(handle h, bool is_borrowed) : object(is_borrowed ? object(h, borrowed_t{}) : object(h, stolen_t{})) {} anyset(handle h, borrowed_t) : object(h, borrowed_t{}) {} anyset(handle h, stolen_t) : object(h, stolen_t{}) {} [[deprecated("Use py::isinstance<py::python_type>(obj) instead")]] bool check() const { return m_ptr != nullptr && ((_Py_IS_TYPE(((const PyObject*)(m_ptr)), &PySet_Type) || _Py_IS_TYPE(((const PyObject*)(m_ptr)), &PyFrozenSet_Type) || PyType_IsSubtype((((PyObject*)(m_ptr))->ob_type), &PySet_Type) || PyType_IsSubtype((((PyObject*)(m_ptr))->ob_type), &PyFrozenSet_Type)) != 0); } static bool check_(handle h) { return h.ptr() != nullptr && (_Py_IS_TYPE(((const PyObject*)(h.ptr())), &PySet_Type) || _Py_IS_TYPE(((const PyObject*)(h.ptr())), &PyFrozenSet_Type) || PyType_IsSubtype((((PyObject*)(h.ptr()))->ob_type), &PySet_Type) || PyType_IsSubtype((((PyObject*)(h.ptr()))->ob_type), &PyFrozenSet_Type)); } template <typename Policy_> anyset(const ::pybind11::detail::accessor<Policy_> &a) : anyset(object(a)) {} anyset(const object &o) : object(o) { if (m_ptr && !check_(m_ptr)) throw ::pybind11::type_error("Object of type '" + ::pybind11::detail::get_fully_qualified_tp_name((((PyObject*)(m_ptr))->ob_type)) + "' is not an instance of '" "anyset" "'"); } anyset(object &&o) : object(std::move(o)) { if (m_ptr && !check_(m_ptr)) throw ::pybind11::type_error("Object of type '" + ::pybind11::detail::get_fully_qualified_tp_name((((PyObject*)(m_ptr))->ob_type)) + "' is not an instance of '" "anyset" "'"); }

public:
    size_t size() const { return static_cast<size_t>(PySet_Size(m_ptr)); }
    template <typename T>
    bool contains(T &&val) const {
        return PySet_Contains(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr()) == 1;
    }
};
```

ecatmur: Oops, yes. Removed.

Skylion007 (approved):

PR #3905, opened 2022-04-26
chore: rule of 3 for strdup guard

Skylion007:
## Description
I noticed our strdup_guard wasn't properly following the rule of 5 and some of the implicit copy constructors / assignment operators could in theory cause a double free. This PR explicitly deletes the copy ctors to guard against accidents in future refactoring work.


rwgk (approved):
Thanks!

PR #3906, opened 2022-04-26
Test pickling a simple callable (does not work).

rwgk:
## Description
Currently only documents that it does not work for C Python (it only works for PyPy). Starting point for future fix.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/tests/test_pickling.py
+++ b/tests/test_pickling.py
@@ -1,4 +1,5 @@
 import pickle
+import re
 
 import pytest
 
@@ -10,7 +11,7 @@ def test_pickle_simple_callable():
     assert m.simple_callable() == 20220426
     with pytest.raises(TypeError) as excinfo:
         pickle.dumps(m.simple_callable)
-    assert str(excinfo.value) == "cannot pickle 'PyCapsule' object"
+    assert re.search("can.*t pickle .*PyCapsule.* object", str(excinfo.value))
 
 
 @pytest.mark.parametrize("cls_name", ["Pickleable", "PickleableNew"])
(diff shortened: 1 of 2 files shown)


tests/test_pickling.py
```diff
@@ -1,11 +1,24 @@
...
 from pybind11_tests import pickling as m
 
 
+def test_pickle_simple_callable():
+    assert m.simple_callable() == 20220426
+    if env.PYPY:
+        serialized = pickle.dumps(m.simple_callable)
+        deserialized = pickle.loads(serialized)
+        assert deserialized() == 20220426
+    else:
```
Skylion007, line 16: Do we really want to test the broken behavior of CPython or just wrap this test with a non-strict xfail? At the very least, we should add a comment that this is a CPython bug not a PyPy bug and that PyPY is doing the preferred behavior for once.

rwgk: I added a comment to be explicit.

The test as-is gives unambiguous information about the where & what, in 3 lines.

strict xfail would lose where & what exactly the failure is.

non-strict xfail would lose almost all information about the where & what.

unless it's tracked in a comment, which is about as many characters, only not executable code, therefore leaving room for doubts and misunderstandings.

Skylion007 (approved):

tests/test_pickling.py
Skylion007: @rwgk You can add a raises arg to xfail, but I see what you mean.

rwgk:
Thanks Aaron! Merging this now, to get this into the smart_holder update.

PR #3913, opened 2022-04-28
Replace error printing code gated by NDEBUG with a new flag: PYBIND11_DETAILED_ERROR_MESSAGES 

voznesenskym:
## Description

Does what it says in the title. We have some internal use cases of pybind where we would love to get richer logging from within pybind, but are not running a debug build.

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include "cast.h"
+#include "detail/common.h"
 
 #include <functional>
 
@@ -477,7 +478,7 @@ struct process_attribute<arg_v> : process_attribute_default<arg_v> {
         }
 
         if (!a.value) {
-#if !defined(NDEBUG)
+#if defined(PYBIND11_DETAILED_ERROR_MESSAGES)
             std::string descr("'");
             if (a.name) {
                 descr += std::string(a.name) + ": ";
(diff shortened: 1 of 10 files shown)


voznesenskym:
Apologies upfront if this is not desired. I am not 100% sure how to test it, I did run all unit tests and nothing exploded.

Skylion007 (approved):

voznesenskym:
> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

@Skylion007 Thanks for the stamp, I do not have merge access :)

Skylion007:
Any objections @henryiii or @rwgk ? If not, let's merge it.

voznesenskym:
@rwgk thanks for the great feedback. Applied.

w/r/t

> Optional, or follow-on PR: define the macro for at least one non-debug CI job (very easy) and verify that some error is complete with details (needs some thought, unless someone already has a good idea how to implement that).


There are tests in `tests/test_methods_and_attributes.py` and `tests/test_pytypes.py` that test that the correct mode prints the correct value. It's not perfect, as it relies on `debug_enabled` being set, which is set from reading the right #define, but can of course be set by anyone, anywhere.

rwgk (approved):
Looks great to me!

@Skylion007, do you want to take another look?

Skylion007 (approved):

Issue #3902, opened 2022-04-25
[BUG]: PYBIND11_OVERRIDE_IMPL issue with multiple inheritance with QObject


Holt59:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

When using `PYBIND11_OVERRIDE_PURE` inside of class with multiple inheritance with QObject, e.g.

```cpp
class PyClass : public MyClass, public QObject {
    Q_OBJECT
    Q_INTERFACES(MyClass)
public:
    QString function() const override {
        PYBIND11_OVERRIDE_PURE(QString, MyClass, function, );
    }
};

py::class_<MyClass, PyClass>(m, "MyClass")
    .def(py::init<>()) // should this be init_alias?
    .def("function", &MyClass::function);
```

This functions throws a "Tried to call pure virtual function ..." exception when calling `function()` on a class returned from Python.

Removing this `static_cast`:

    https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L2729

...solves the issue.

Is there any reason to have this `static_cast` here? Is there something I'm missing?

Skylion007:
@Holt59 Could you add a PR adding this test to our test suite? That would make debugging a solution easier. The static_cast is probably to ensure consistent behavior and fix some other edge case. Does replacing it with a dynamic_cast also solve the problem?

Holt59:
@Skylion007 I will see if I can add a PR. Replacing with `dynamic_cast` does not solve the issue, I think the problem is that the `py::object` is not retrieved properly from the `this` when `this` is a `B*` (and not a `PyB*`).

PR #3915, opened 2022-04-29
Add tests for virtual function and multiple inheritance, #3902.

Holt59:
## Description

Test cases for #3902 - The tests are currently crashing, I did not find a way to have them only failing.

### DIFF
--- a/tests/test_virtual_functions.cpp
+++ b/tests/test_virtual_functions.cpp
@@ -187,6 +187,22 @@ struct Adder : AdderBase {
     }
 };
 
+// Multiple inheritance, issue #3902
+struct BaseMultipleInheritance1 {
+    virtual ~BaseMultipleInheritance1() {}
+};
+
+struct BaseMultipleInheritance2 {
+    virtual std::string run() const = 0;
+    virtual ~BaseMultipleInheritance2() {}
+};
+
+struct PyMultipleInheritance : BaseMultipleInheritance1, BaseMultipleInheritance2 {
+    virtual std::string run() const {
+        PYBIND11_OVERRIDE_PURE(std::string, BaseMultipleInheritance2, run, );
+    }
+};
+
 static void test_gil() {
     {
         py::gil_scoped_acquire lock;
@@ -411,6 +427,15 @@ TEST_SUBMODULE(virtual_functions, m) {
         .def("func", &test_override_cache_helper::func);
 
     m.def("test_override_cache", test_override_cache);
+
+    // test multiple inheritance
+    // #3902
+    py::class_<BaseMultipleInheritance2, PyMultipleInheritance>(
+        m, "MultipleInheritance", py::multiple_inheritance())
+        .def(py::init<>())
+        .def("run", &BaseMultipleInheritance2::run);
+    m.def("test_multiple_inheritance",
+          [](BaseMultipleInheritance2 *p) -> std::string { return p->run(); });
 }
 
 // Inheriting virtual methods.  We do two versions here: the repeat-everything version and the
--- a/tests/test_virtual_functions.py
+++ b/tests/test_virtual_functions.py
@@ -457,3 +457,14 @@ def test_python_override():
     for _ in range(1500):
         assert m.test_override_cache(func()) == 42
         assert m.test_override_cache(func2()) == 0
+
+
+def test_multiple_inheritance():
+    """Issue #3902."""
+
+    class A(m.MultipleInheritance):
+        def run(self):
+            return "success"
+
+    a = A()
+    assert a.run() == "success"


tests/test_virtual_functions.cpp
```diff
@@ -187,6 +187,22 @@ struct Adder : AdderBase {
...
+struct BaseMultipleInheritance1 {
+    virtual ~BaseMultipleInheritance1() {}
+};
+
+struct BaseMultipleInheritance2 {
+    virtual std::string run() const = 0;
+    virtual ~BaseMultipleInheritance2() {}
+};
+
+struct PyMultipleInheritance : BaseMultipleInheritance1, BaseMultipleInheritance2 {
```
rwgk, line 200: This looks invalid.

The trampoline class (aka "alias" in the sources) is meant to inherit only from `BaseMultipleInheritance2` in this case, the `T` in the corresponding `class_<T>` more generally speaking.

> https://github.com/pybind/pybind11/issues/3902#issuecomment-1109932084

Yes, that's the supported situation.

Why do you want the trampoline class to inherit from `BaseMultipleInheritance1`?

rwgk: I glanced through the ModOrganizer2 code but there are so many similar-looking names I got lost.

Let's try to distill out a minimal reproducer that we can agree on is valid.

My understanding, starting very basic:

* You have a `class ClsA` with virtual functions that you want override in Python, therefore you need a `class ClsATrampoline`.
* Now the same for `ClsB` and so on.
* Where does the multiple inheritance come into play? In this PR I only see that for the trampoline, which is where I get lost. I'm thinking the multiple inheritance has to be between the `ClsX`es, only, and that multiple inheritance has to be reflected accurately in the `class_` bindings. The trampolines remain simple: 1-to-1 between `ClsX` and `ClsXTrampoline`.

I could be wrong, but that's been my understanding until now. Could you please point to existing documentation or tests that use multiple inheritance only for the trampoline class?

Holt59: In MO2, we have "plugins", so we have some interfaces: `IPlugin`, `IPluginGame`, `IPluginDiagnose` for instance.

Unfortunately, these have weird inheritance relationship: 
- `IPlugin` does not inherit anything,
- `IPluginGame` inherits `IPlugin` and `QObject` (see below),
- `IPluginDiagnose` does not inherit anything.

In C++, if you implement a plugin, you can:
- Extend a "large" plugin, e.g., `IPluginGame`, in which case you get the `IPlugin` and `QObject` inheritence,
- Extend a "light" plugin, e.g. `IPlugin` or `IPluginDiagnose`, in which case you also need to inherit `QObject` manually,
- You can also extend one large and one light, e.g., extends both `IPluginDiagnose` and `IPluginGame`.

In Python, it's impossible to have such flexibility, so everything you extend from Python must inherit `IPlugin` and `QObject`, which is why we have weird trampoline.

Ideally:
- The `IPlugin` trampoline (`PyPlugin`) would inherit `IPlugin` and `QObject`,
- The `IPluginDiagnose` trampoline (`PyPluginDiagnose`) would inherit `IPlugin`, `IPluginDiagnose` and `QObject`,
- The `IPluginGame` trampoline (`PyPluginGame`) would inherit only `IPluginGame`.

Then when I declare the `class_`, I can do so with the raw plugins:

```cpp
py::class_<IPlugin, PyPlugin>(...);
py::class_<IPluginDiagnose, PyPluginDiagnose, IPlugin>(...);
py::class_<IPluginGame, PyPluginGame, IPlugin>(...);
```

This currently does not work, as per the above issues - I'm not sure there would not be other issues even if the cast was working.

This is an extreme edge-case, if this is not going to be supported, maybe it would just be nice to have a paragraph in the documentation about multiple inheritance and virtual function bindings.

rwgk: > In C++, if you implement a plugin, you can:
> 
> Extend a "large" plugin, e.g., IPluginGame, in which case you get the IPlugin and QObject inheritence,
> Extend a "light" plugin, e.g. IPlugin or IPluginDiagnose, in which case you also need to inherit QObject manually,
> You can also extend one large and one light, e.g., extends both IPluginDiagnose and IPluginGame.

> In Python, it's impossible to have such flexibility,

Bug "huh?" here. I cannot get into how MO2 is designed, but just superficially, it looks straightforward:

```
class_<IPlugin, IPluginTrampoline>
class_<QObject, QObjectTrampoline>
class_<IPluginGame, IPlugin, QObject, IPluginGameTrampoline>
class_<IPluginDiagnose, IPluginDiagnoseTrampoline>
```

Then in Python inherit as you would in C++. My only worry would be MRO differences. If that bites work on the MO2 design, so that it works in both languages. I expect that you'll arrive at an improved design in all respects.

> so everything you extend from Python must inherit IPlugin and QObject, which is why we have weird trampoline.

Sorry I cannot follow that logic. Let's close this PR.

Holt59: What you propose is not possible:
- We cannot bind `QObject` because it would conflict with the `PyQt6` bindings and we need type casters for `QObject` to be able to retrieve objects constructed from `PyQt6`.
- We need all trampoline classes to inherit `QObject` and `IPlugin` - With what you propose, I can indeed inherit multiple classes in Python, but then I don't have a single object in C++ that has the 3 interfaces.

> > so everything you extend from Python must inherit IPlugin and QObject, which is why we have weird trampoline.
>
> Sorry I cannot follow that logic. Let's close this PR.

We need a common interface when passing plugins between components, and since the C++ plugins are Qt plugins, we use `QObject*` to be consistent. Thus, the C++ manager only receives `QObject*` and then checks the various sub-interface to see which ones are implemented. 

That's the design choice of MO2, I'm not the one that made it, but that's how it is right now and I cannot change it.

Holt59: @rwgk I do have a working solution for the MO2 implementation, which is convoluted but works. This PR was for the much simpler use-case.

If this does not work "per-design" (which I can totally understand), then maybe it would be nice to simply add a remark about it on the documentation, maybe perform a check in the `PYBIND11_OVERRIDE(_PURE)` with a clearer message?

rwgk: > @rwgk I do have a working solution for the MO2 implementation, which is convoluted but works.

I recommend leaving it at that. You seem to have a complicate binding-system interop challange (`QObject`), for which we don't have a general solution.

> If this does not work "per-design" (which I can totally understand), then maybe it would be nice to simply add a remark about it on the documentation,

What would that remark be?

> maybe perform a check in the `PYBIND11_OVERRIDE(_PURE)` with a clearer message?

The macro family is very complex already: seems tricky, and getting people to agree on anything that adds runtime overhead is usually an uphill battle.

Holt59: > > If this does not work "per-design" (which I can totally understand), then maybe it would be nice to simply add a remark about it on the documentation,
> 
> What would that remark be?

Simple that trampoline classes should only inherit a single class, and for use-case similar to the one in the PR, a simple solution is to make an intermediate class.

That might be a trivial remark, but this remark would have same me a lot of time (and I think it can help others since Boost.Python wrappers allow this kind of things, and I guess many people are moving from Boost.Python to pybind11).

rwgk: > Simply that trampoline classes should only inherit a single class

That wouldn't be accurate, it totally depends on what the multiple inheritance is for, e.g.:

https://github.com/pybind/pybind11/blob/smart_holder/README_smart_holder.rst#trampolines-and-stdunique_ptr

In your test here, `BaseMultipleInheritance1` can simply be removed and the test runs fine (I actually tried it out). Therefore I'm thinking the test is not conforming to the "as simple as possible, but not simpler" principle. Can you please show a self-contained test that makes it clear why `BaseMultipleInheritance1` is needed, and that breaks if it is removed as a base of the trampoline class?

Note that your test here also passes if you reverse the order of the bases. Is there a reason why `BaseMultipleInheritance1` needs to be first? If so, that needs to be captured by the self-contained test as well.

As an aside, I'm thinking the title of this PR is very misleading. Apparently you want to change some detail of how trampolines are implemented (remove a `static_cast`), but not anything to do with how pybind11 `class_` handles multiple inheritance.

Holt59: @rwgk Thanks for the feedback but since that PR is now closed, and looking at how this is going and the many hours I already lost on this problem, I'm not going to invest more time on it.

PR #3916, opened 2022-04-29
Fix issue #3912 : invalid docstring generation of std::array <-> list caster

### LINKED ISSUE #3912: [BUG]: Invalid typing for `std::array`?
### Required prerequisites

### Problem description

When returning a `std::array`, the typing information is invalid, e.g. `List[str[2]]` for the code below. This causes tool such as mypy's stubgen or pybind11-stubgen to produce invalid stubs.

I guess it would be valid to generate `List[str]` in this case since Python does not have information about the size anyway?

### Reproducible example code

```text
#include <pybind11/stl.h>

namespace py = pybind11;

PYBIND11_MODULE(cpp_module, m)
{
    m.def("arr", []() -> std::array<const char*, 2> {
        return {"foo", "bar"};
    });
}
```

Skylion007:
## Description
* Fixes #3912, by removing size information from list, while this information is helpful, it generates invalid type signatures for mypy and other type checkers.

## Suggested changelog entry:

```rst
* Fixes doc generation for std::array-list caster. Previously, signatures included the size of the list in a non-standard, non-spec compliant way.
```

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -251,12 +251,7 @@ public:
         return l.release();
     }
 
-    PYBIND11_TYPE_CASTER(ArrayType,
-                         const_name("List[") + value_conv::name
-                             + const_name<Resizable>(const_name(""),
-                                                     const_name("[") + const_name<Size>()
-                                                         + const_name("]"))
-                             + const_name("]"));
+    PYBIND11_TYPE_CASTER(ArrayType, const_name("List[") + value_conv::name + const_name("]"));
 };
 
 template <typename Type, size_t Size>
--- a/tests/test_stl.py
+++ b/tests/test_stl.py
@@ -39,8 +39,8 @@ def test_array(doc):
     assert m.load_array(lst)
     assert m.load_array(tuple(lst))
 
-    assert doc(m.cast_array) == "cast_array() -> List[int[2]]"
-    assert doc(m.load_array) == "load_array(arg0: List[int[2]]) -> bool"
+    assert doc(m.cast_array) == "cast_array() -> List[int]"
+    assert doc(m.load_array) == "load_array(arg0: List[int]) -> bool"
 
 
 def test_valarray(doc):


wjakob:
PR trades MyPy compliance for potentially very confusing situation where a pybind11-bound function does not accept an input array, and the error message provides zero context why that is (input has the wrong size)

wjakob:
These signatures in pybind11 precede the time where MyPy was even a thing, so "non-standard non-spec compliant" seems a bit strong of a qualifier.

Skylion007:
@Wjacob, in >=3.9 we can use the Annotated typing extension to specify the length of the list. Hmm, not sure if there is a standard compliant way to specify the type for anything besides Tuples though.

include/pybind11/stl.h
```diff
@@ -251,12 +251,7 @@ struct array_caster {
         return l.release();
     }
 
-    PYBIND11_TYPE_CASTER(ArrayType,
-                         const_name("List[") + value_conv::name
-                             + const_name<Resizable>(const_name(""),
-                                                     const_name("[") + const_name<Size>()
-                                                         + const_name("]"))
-                             + const_name("]"));
+    PYBIND11_TYPE_CASTER(ArrayType, const_name("List[") + value_conv::name + const_name("]"));
```
virtuald, line 254: I think if you make the string `typing.Annotated[List[type], size]` then pybind11-stubgen will generate the correct thing from this (and if it doesn't then you can just modify it to do so).

virtuald: Also note, it's just a string. You don't need typing extensions in pybind11, that's the responsibility of whoever interprets the string.

felixvd: `Annotated[List[type], size]` appears to be the intended standard way to hint the length of a list. [PEP 593](https://peps.python.org/pep-0593/) shows this use case as an example. 

I would just remove `typing.` from your suggestion since `List` is not prefixed either.

> This PR achieves MyPy compliance in exchange for potentially very confusing situations where a pybind11-bound function does not accept an input array, and the error message provides zero context why that is (input has the wrong size)

@wjakob My understanding is that this PR only changes the type hint and does not affect the error message otherwise. Is there a situation where `Annotated[List[str], 2]` would be inferior to `List[str[2]]`?

cielavenir:
@Skylion007 @rwgk can we close this one?

rwgk:
Closing after #4679 was merged.

PR #3919, opened 2022-05-01
Fix file extension on windows with cp36 and cp37

HaarigerHarald:
Hi,

I noticed that with #3764 pybind11 modules now have `.pyd` as file extension on Windows for CPython 3.6 and CPython 3.7 while it previously was `.cp36-win_amd64.pyd` and `.cp37-win_amd64.pyd`, respectively. That was probably just an oversight, which should be fixed with this PR.

PS: Tried to be consistent with the usage of `;` but I don't know why it's used here in the first place.


henryiii:
Thanks! Looks like an oversight. Worth back porting if we do any more back porting and/or don't release 2.10 soon.

PR #3922, opened 2022-05-03
Chore: update NVIDIA-PGI CI workflow

Skylion007:
## Description
* Update NVIDIA-PGI's compiler to the latest and greatest version


henryiii (approved):
We only officially support the latest versions for secondary compilers.

PR #3923, opened 2022-05-04
(dev): 3.11 testing

Skylion007:
## Description
* Testing 3.11 

## Suggested changelog entry:

```rst
* Fix issues and test against CPython 3.11 betas.
```

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -14,7 +14,7 @@ env:
 
 jobs:
   standard:
-    name: "🐍 3.11 dev • ubuntu-latest • x64"
+    name: "🐍 3.11 latest internals • ubuntu-latest • x64"
     runs-on: ubuntu-latest
     if: "contains(github.event.pull_request.labels.*.name, 'python dev')"
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#include "detail/common.h"
 #include "cast.h"
 
 #include <functional>
@@ -477,7 +478,7 @@ struct process_attribute<arg_v> : process_attribute_default<arg_v> {
         }
 
         if (!a.value) {
-#if !defined(NDEBUG)
+#if defined(PYBIND11_DETAILED_ERROR_MESSAGES)
             std::string descr("'");
             if (a.name) {
                 descr += std::string(a.name) + ": ";
@@ -498,7 +499,8 @@ struct process_attribute<arg_v> : process_attribute_default<arg_v> {
 #else
             pybind11_fail("arg(): could not convert default argument "
                           "into a Python object (type not registered yet?). "
-                          "Compile in debug mode for more information.");
+                          "#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for "
+                          "more information.");
 #endif
         }
         r->args.emplace_back(a.name, a.descr, a.value.inc_ref(), !a.flag_noconvert, a.flag_none);
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -345,8 +345,9 @@ struct type_record {
 
         bases.append((PyObject *) base_info->type);
 
+#if PY_VERSION_HEX < 0x030B0000
         dynamic_attr |= base_info->type->tp_dictoffset != 0;
-#if PY_VERSION_HEX >= 0x030B0000
+#else
         dynamic_attr |= (base_info->type->tp_flags & Py_TPFLAGS_MANAGED_DICT) != 0;
 #endif
 
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -345,9 +345,10 @@ struct type_record {
 
         bases.append((PyObject *) base_info->type);
 
-        if (base_info->type->tp_dictoffset != 0) {
-            dynamic_attr = true;
-        }
+        dynamic_attr |= base_info->type->tp_dictoffset != 0;
+#if PY_VERSION_HEX >= 0x030B0000
+        dynamic_attr |= (base_info->type->tp_flags & Py_TPFLAGS_MANAGED_DICT) != 0;
+#endif
 
         if (caster) {
             base_info->implicit_casts.emplace_back(type, caster);
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -157,12 +157,14 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
         // A failure here indicates a character-encoding failure or the python
         // interpreter out of memory. Give up.
         PyConfig_Clear(&config);
-        throw std::runtime_error("Failed to prepare CPython");
+        throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
+                                                          : "Failed to prepare CPython");
     }
     status = Py_InitializeFromConfig(&config);
     PyConfig_Clear(&config);
     if (PyStatus_Exception(status)) {
-        throw std::runtime_error("Failed to init CPython");
+        throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
+                                                          : "Failed to init CPython");
     }
     if (add_program_dir_to_path) {
         PyRun_SimpleString("import sys, os.path; "
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -86,37 +86,6 @@ inline wchar_t *widen_chars(const char *safe_arg) {
     return widened_arg;
 }
 
-/// Python 2.x/3.x-compatible version of `PySys_SetArgv`
-inline void set_interpreter_argv(int argc, const char *const *argv, bool add_program_dir_to_path) {
-    // Before it was special-cased in python 3.8, passing an empty or null argv
-    // caused a segfault, so we have to reimplement the special case ourselves.
-    bool special_case = (argv == nullptr || argc <= 0);
-
-    const char *const empty_argv[]{"\0"};
-    const char *const *safe_argv = special_case ? empty_argv : argv;
-    if (special_case) {
-        argc = 1;
-    }
-
-    auto argv_size = static_cast<size_t>(argc);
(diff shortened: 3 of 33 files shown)


Skylion007:
@rwgk Any ideas how to address this failure? I would like to get PyBind11 fully supported before 3.11 leaves alpha.

rwgk:
> @rwgk Any ideas how to address this failure? I would like to get PyBind11 fully supported before 3.11 leaves alpha.

No, unfortunately not. First thought: this could be really tricky to debug and fix. MRO is complex, especially between native Python & pybind11:

https://pybind11.readthedocs.io/en/stable/advanced/classes.html

> ... but once you begin mixing Python and C++ multiple inheritance, things will fall apart due to differences between Python’s MRO and C++’s mechanisms.

That's all I know though, I don't have first-hand experience working on the MI code.

What I'd try first: comment out tests until everything that's left passes. Then look at the overall picture to decide what to do next.

Skylion007:
@rwgk I know what causes the issue. It's a corner case caused when a PyBind11 Class has multiple inheritance and the first base does not have dynamic-attrs, but another base class does not.

rwgk:
I only remembered that very vaguely.

Variation of my "comment out" strategy: manually add ` py::dynamic_attr` to see if that makes everything pass. If that's do-able: could that even be the final solution, maybe with a better error message?

Skylion007:
@rwgk Adding it to the children classes did not help. Adding it to the base classes does work, but then  that makes this unit test stops doing anything meaningful and the base class and any of its parents inherit the performance penalty of the dynamic-attrs.

rwgk:
> @rwgk Adding it to the children classes did not help. Adding it to the base classes does work, but then that makes this unit test stops doing anything meaningful and the base class and any of its parents inherit the performance penalty of the dynamic-attrs.

Sound fine to me:
* Most importantly, everything works. (I guess)
* In all likelihood, the performance penalty is inconsequential. (If not, they have a bigger problem IMO.)
* If someone disagrees regarding the performance penalty, invite them to work out a fix.

My net conclusion: Add `py::dynamic_attrs` to the base class with a terse comment and link pointing here.

henryiii:
FYI, beta1 is out, no more non-bug fix changes.

include/pybind11/embed.h
```diff
@@ -146,9 +115,54 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
...
+        widened_argv[ii] = widened_argv_entries.back().get();
+    }
+
+    auto *pysys_argv = widened_argv.get();
+
+    PySys_SetArgvEx(argc, pysys_argv, static_cast<int>(add_program_dir_to_path));
+#else
+    PyConfig config;
+    PyConfig_InitPythonConfig(&config);
+    config.safe_path = add_program_dir_to_path ? 0 : 1;
```
henryiii, line 153: @vstinner Do you know why this might not be working in 3.11? I've swapped the third argument of `PySys_SetArgvEx` with this setting (reversed, since it's true to remove the local directory from the search path), but 3.11 is unable to find the local modules in the tests (and then segfaults for good measure). I think we are working around https://github.com/python/cpython/issues/92678 correctly but the tests break on this.

include/pybind11/embed.h
```diff
@@ -146,9 +115,54 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
...
+    config.install_signal_handlers = init_signal_handlers ? 1 : 0;
+
+    PyStatus status = PyConfig_SetBytesArgv(&config, argc, const_cast<char *const *>(argv));
+    if (PyStatus_Exception(status)) {
+        // A failure here indicates a character-encoding failure or the python
+        // interpreter out of memory. Give up.
+        PyConfig_Clear(&config);
+        return;
+    }
+    Py_InitializeFromConfig(&config);
```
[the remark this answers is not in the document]
henryiii: Thanks!

include/pybind11/embed.h
henryiii: Okay, I think what's happening is the newer API is not adding this to the path (regardless of `safe_path`, which only affects `Py_RunMain` - the new API pre-dates safe_path). So I have to manually add the [old algorithm](https://docs.python.org/3/c-api/init.html#c.PySys_SetArgvEx) of pulling out the script directory if an existing path is the first argument, and adding an empty string otherwise.

henryiii: `_PyPathConfig_ComputeSysPath0` is private. :( It seems looking at PEP 587 that all the `sys.path` handling is now tied to the high level API (`Py_Main`, `Py_BytesMain`, and `Py_RunMain`), and line-by-line runners like us are stuck reimplementing the logic ourselves now.

vstinner:
In the Python 3.10 code path, PySys_SetArgvEx() is called *after* Python initialization, when sys.path was already computed. In this case, the function inserts a new path at ``sys.path[0]``:

```
If argv[0] is not '-c' nor '-m', prepend argv[0] to sys.path.
```

If you don't call PySys_SetArgvEx(), Py_RunMain() inserts a new path at ``sys.path[0]`` if safe_path=0: the path depends on ``PyConfig.argv[0]``.

I tried to document everything about the Python Path Configuration at: https://docs.python.org/dev/c-api/init_config.html#python-path-configuration See the "Py_RunMain() and Py_Main() modify sys.path: (...)" section.

If you control argv and you know what should be in ``sys.argv[0]``, you can just modify sys.path manually. Dummy example without error handling:

```
PyRun_SimpleString("import sys; sys.path.insert(0, '/my/path')");
```

Note: the code to compute sys.path has been rewritten in pure Python in Python 3.11: https://github.com/python/cpython/blob/main/Modules/getpath.py But this code doesn't include changes done afterwards by Py_RunMain(), as I wrote.

include/pybind11/embed.h
```diff
@@ -146,9 +115,62 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
...
+    PyConfig config;
+    PyConfig_InitIsolatedConfig(&config);
+    config.install_signal_handlers = init_signal_handlers ? 1 : 0;
+
+    PyStatus status = PyConfig_SetBytesArgv(&config, argc, const_cast<char *const *>(argv));
+    if (PyStatus_Exception(status)) {
+        // A failure here indicates a character-encoding failure or the python
+        // interpreter out of memory. Give up.
+        PyConfig_Clear(&config);
+        throw std::runtime_error("Failed to prepare CPython");
```
Skylion007, line 160: @henryiii Is there a reason it's throwing an runtime error instead of calling pybind11_fail? Is it to avoid assert for ensuring a clear error indicator?

henryiii: Nothing major, was just trying the simplest change to get it working, feel free to switch to that if it works. I wasn't sure how far along it was in the process it needed to be to call `PyErr_Occcured` (this is the embedded interpreter, not an extension module). This also should be rare, I hope (also hard to write a test for).

henryiii: Same thing just below.

Skylion007: Makes sense, fine as is then.

include/pybind11/attr.h
```diff
@@ -345,9 +345,10 @@ struct type_record {
 
         bases.append((PyObject *) base_info->type);
 
-        if (base_info->type->tp_dictoffset != 0) {
-            dynamic_attr = true;
-        }
+        dynamic_attr |= base_info->type->tp_dictoffset != 0;
```
Skylion007, line 348: @henryiii I am not sure if this should be an exclusive if/else . This seems to be the simplest delta though.

Skylion007: other option do an #IF and #ELSE

henryiii: Seems like it's better as an `#else` for the if below?

include/pybind11/embed.h
vstinner, line 160: You might want to include the error message: ``status.err_msg``, it's only filled if ``PyStatus_IsError()`` (otherwise, it's NULL).

include/pybind11/attr.h
henryiii: If `base_info->type->tp_dictoffset` is still important in 3.11 I think this is best. If not, it could be placed in an `#else`.

.github/workflows/upstream.yml
```diff
@@ -14,7 +14,7 @@ env:
 
 jobs:
   standard:
-    name: "🐍 3.11 dev • ubuntu-latest • x64"
+    name: "🐍 3.11 latest internals • ubuntu-latest • x64"
```
rwgk, line 17: Consistency would be nice, e.g.

3.11-latest above (line 33) and here.

I often look at CI logs in bulk, having to manually map `latest internals` here to `-dev` elsewhere is confusing/distracting.

include/pybind11/embed.h
```diff
@@ -146,9 +115,64 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
...
+            return;
+        }
+        widened_argv[ii] = widened_argv_entries.back().get();
+    }
+
+    auto *pysys_argv = widened_argv.get();
+
+    PySys_SetArgvEx(argc, pysys_argv, static_cast<int>(add_program_dir_to_path));
+#else
+    PyConfig config;
```
rwgk, line 151: A comment here would be useful (similar to Henry's comment from a few days ago):

* Link to the corresponding Python sources.
* Something like "unfortunately the Python code is private, therefore we are forced to duplicate here".

rwgk (commented):
Looks like Henry was faster by a few minutes.

include/pybind11/embed.h
vstinner: What has to be duplicated?

vstinner: PyConfig API initializes the default value of sys.path. But Py_RunMain() then inserts a path to sys.path. Python 3.11 safe_path=1 disables this behavior:
* https://docs.python.org/dev/c-api/init_config.html#c.PyConfig.safe_path
* https://docs.python.org/dev/c-api/init_config.html#python-path-configuration

It might be interesting to move more code changing sys.path into the Python initialization (PyConfig), but it's complicated.

PR #3928, opened 2022-05-06
chore: remove unused include from stl.h

Skylion007:
## Description
* Removes the iostream include from stl.h. Only the header ostream is used which is included transitively. 
* Should help speed up compilation when pybind11/stl.h is included but istream objects are not needed.

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -15,6 +15,7 @@
 #include <deque>
 #include <list>
 #include <map>
+#include <ostream>
 #include <set>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


rwgk (approved):

rwgk:
For the record / entertainment: This turns out to be a breaking change: Google-global testing hit 4 projects (including https://github.com/google/atheris) that are broken by this PR (13 broken tests in total).

We should mention this in the changelog. The global testing suggests a number of people will stumble over this.

Skylion007:
@rwgk I updated the changelog to be more specific (we still include iostream.h in pybind11/iostream.h).

PR #3929, opened 2022-05-07
[smart_holder] Fix "extra ';' outside of a function" warning

tomba:
Fix the following warning seen with clang:

include/pybind11/detail/smart_holder_poc.h:109:2: error: extra ';' outside of a function is incompatible with C++98 [-Werror,-Wc++98-compat-extra-semi]

Signed-off-by: Tomi Valkeinen <tomi.valkeinen@ideasonboard.com>


rwgk:
Thanks for the fix! (I'm a little amazed that our heavy-duty code health tooling didn't catch this.)

The 2 CI failures are our currently known tool failures (2 x the test that is already failing on master).

PR #3938, opened 2022-05-12
Add install info for CMake/Hunter

caseymcc:
[Master] Add install info for CMake/Hunter

## Description

Adds information on how to install pybind11 when using CMake/Hunter

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -82,6 +82,44 @@ community contributors. If the version is out of date, please `create an issue
 or pull request <https://github.com/Microsoft/vcpkg/>`_ on the vcpkg
 repository.
 
+Include with CMake/Hunter
+==================
+If you are already using CMake you can include pybind11 using Hunter. 
+If you are not already using Hunter, download 
+`HunterGate <https://github.com/cpp-pm/gate/tree/master/cmake>`_ to a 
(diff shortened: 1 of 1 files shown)


rwgk (commented):
This is @henryiii's domain, but I looked because I was curious and left some tiny suggestions. 

I don't know much about cmake and never heard of hunter, but from a high-level perspective, is this the best place for this documentation? I think it would be better to have this in the hunter docs, and a link to that doc here. Something like:

Other build tools and package managers with pybind11 support:
* hunter
* ... I guess there are or will be more. Is that a reasonable guess?

Such links seem helpful, and having the landing pages maintained by the owners will scale nicely. (Vs pybind11 maintainers having to keep an eye on them.)

henryiii:
Looks like the latest version of pybind11 on hunter is 2.8.1. I remember the update process being slightly annoying with CLI11, others keep it up to date nowadays and I'm happy I haven't had to push the update in a while. It mostly hurts that it's not automated or easy to automate. I'm not against docs here, though I'd push an update to hunter first (after the next release, which probably will be pretty soon because there are a couple of CMake issues that should be fixed soon (PRs open)). I am worried about this growing forever - maybe we could make this a "details" expander? Maybe after updating the docs formatting (which is on my long-term todo list)?

PR #3943, opened 2022-05-16
Do we have a unit test for the traceback code in error_string()?

rwgk:
## Description
**Just a quick experiment.** Local testing passes without the removed code. Does testing pass on all platforms?

## Suggested changelog entry:

```rst

```


rwgk:
Everything passes, except two existing failures already on master.
Evidently the traceback code in `error_string()` does not have unit test coverage for the results.
**But** there is one test that fails if a `pybind11_fail()` is inserted: test_exceptions.py test_nested_throws
I.e. the C++ code is covered, only the result is never checked.

PR #3949, opened 2022-05-16
Fix py::cast from pytype rvalue to pytype

MaartenBaert:
Previously, py::cast blindly assumed that the destination type was a C++ type rather than a python type when the source type was an rvalue.

## Description

The following piece of code:
```
py::int_ value = py::float_(1.0).cast<py::int_>();
```
... results in the following bizarre runtime error:
```
RuntimeError: Unable to cast Python instance of type <class 'float'> to C++ type 'int_'
```
Whereas this works fine:
```
py::float_ temp = 1.0;
py::int_ value = temp.cast<py::int_>();
```
The problem seems to be that in the first case, the `py::float_` is an rvalue, which causes pybind to use a different implementation of `py::cast` which blindly assumes that the destination type is always a C++ type and attempts to use the `type_caster` mechanism - unlike the normal `py::cast` implementation which checks the destination type to determine whether to use `type_caster` or just call the constructor of the destination type.

This commit does three things:
- Add a static assertion to `load_type` such that if it gets invoked with a python type rather than a C++ type, it will produce a meaningful compile-time error rather than the bizarre runtime error shown above.
- Modify `py::cast(object &&)` such that it actually checks whether the destination type is a python object, and if so, invokes the move constructor instead of using `type_caster`.
- Add a test to make sure that the problem is fixed.

## Suggested changelog entry:

```rst
Fix cast from pytype rvalue to another pytype.
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1009,7 +1009,8 @@ struct return_value_policy_override<
 // Basic python -> C++ casting; throws if casting fails
 template <typename T, typename SFINAE>
 type_caster<T, SFINAE> &load_type(type_caster<T, SFINAE> &conv, const handle &handle) {
-    static_assert(!detail::is_pyobject<T>::value, "Internal error: type_caster should only be used for C++ types");
+    static_assert(!detail::is_pyobject<T>::value,
+                  "Internal error: type_caster should only be used for C++ types");
     if (!conv.load(handle, true)) {
 #if !defined(PYBIND11_DETAILED_ERROR_MESSAGES)
         throw cast_error("Unable to cast Python instance to C++ type (#define "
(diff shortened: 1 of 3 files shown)


include/pybind11/cast.h
```diff
@@ -1099,21 +1101,30 @@ detail::enable_if_t<!detail::move_never<T>::value, T> move(object &&obj) {
 // - If both movable and copyable, check ref count: if 1, move; otherwise copy
 // - Otherwise (not movable), copy.
 template <typename T>
-detail::enable_if_t<detail::move_always<T>::value, T> cast(object &&object) {
+detail::enable_if_t<!detail::is_pyobject<T>::value && detail::move_always<T>::value, T>
```
Skylion007, line 1104: This could be cleaner using detail::all_of or detail::satisifes_all_of, none_of satisfies_none_of

Skylion007 (approved):

Skylion007:
@rwgk Yeah, I looked more closely at the logic and realized all_of any_of would not have helped.

PR #3958, opened 2022-05-18
chore: add err guard to capsule destructor and add a move to iostream

Skylion007:
## Description
* One fixes a potential bug in pycapsule destructor by adding an error_guard to one of the dtors.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/iostream.h
+++ b/include/pybind11/iostream.h
@@ -100,7 +100,7 @@ private:
 
             if (size > remainder) {
                 str line(pbase(), size - remainder);
-                pywrite(line);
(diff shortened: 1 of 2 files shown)


henryiii (approved):

rwgk (approved):
I didn't get a chance to look closely today.

I don't fully understand how this plays with the pybind11_fail, but I feel sufficiently confident that this change will only make things better.

It would be good to note somewhere (here or PR description) how you came to add the error_guard (something failed? or did you just happen to notice?)

PR #3959, opened 2022-05-19
fix: support cross-compiles with classic Python

henryiii:
Signed-off-by: Henry Schreiner <henryschreineriii@gmail.com>

## Description

Drop the bit check for cross compiles (like for Pyodide). Also allow found values to be overridden if needed for future proofing cross compiles. See https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/pybind11/patches/0001-no-bitsize-check.patch. I believe FindPython already supports cross-compiling.

## Suggested changelog entry:

```rst
* Drop the bitness check on cross compiles (like targeting WebAssembly via Emscripten)
```

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -151,26 +151,36 @@ if(NOT _PYTHON_SUCCESS MATCHES 0)
   return()
 endif()
 
+# Can manually set values when cross-compiling
+macro(_PYBIND11_GET_IF_UNDEF lst index name)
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

henryiii:
Looks like FindPython has the same check: https://github.com/Kitware/CMake/blob/07d351b535dde56a89aa6786a3f0701e3b159d44/Modules/FindPython/Support.cmake#L815-L817 so I'm pretty happy with it, though I'm having a different error later on for what I'm doing.

PR #3962, opened 2022-05-19
Experiment: error_already_set_terminate_if_gil_not_held

rwgk:
## Description
Experimental/scratch PR, to answer this question:

```
catch (const py::error_already_set &e) {
    throw e; // Uses copy ctor. Are we sure we have the GIL in all situations?
}
```

The copy ctor uses INCREF, which could race if the GIL is not held.

## Suggested changelog entry:

```rst

```


rwgk:
Cleaning up.

PR #1895, opened 2019-08-28
error_already_set::what() is now constructed lazily

superbobry:
Prior to this commit throwing error_already_set was expensive due to the
eager construction of the error string (which required traversing the
Python stack). See #1853 for more context and an alternative take on the
issue.

Note that error_already_set no longer inherits from std::runtime_error
because the latter has no default constructor.

PR #3964, opened 2022-05-20
Unladen `error_already_set`

rwgk:
## Description
Adopts the Boost.Python approach for `error_already_set`.

Motivations (see PR #1895):

* Fixes undefined behavior of existing `error_already_set` copy constructor: it is very unlikely to trigger Python reference count corruption, but that is a possibility in the general case.
* Maximizes performance, in particular avoids building `what()`, which is very rarely actually used.

**This is a breaking change.**

Based on the changes to the unit tests, I believe most required adjustments in user code are backward compatible:

* Additionally needed `PyErr_Clear();` are no-ops with released pybind11 versions.

It seems to be very rare that `what()` is called. `#ifdef`s (in user code) will be needed to replace those calls in a backward-compatible way.

More work is needed to be sure that there are no bigger obstacles.

See also:
* https://github.com/pybind/pybind11/pull/1895#issuecomment-1124576917
* https://github.com/pybind/pybind11/pull/1895#issuecomment-1131356883

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -326,7 +326,7 @@
     catch (pybind11::error_already_set &) {                                                       \
         std::string err_str = py::detail::error_string();                                         \
         PyErr_Clear();                                                                            \
-        PyErr_SetString(PyExc_ImportError, err_str.c_str());                                      \
+        PyErr_SetString(PyExc_ImportError, ("initialization failed: " + err_str).c_str());        \
         return nullptr;                                                                           \
     }                                                                                             \
     catch (const std::exception &e) {                                                             \
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -324,9 +324,9 @@
 
 #define PYBIND11_CATCH_INIT_EXCEPTIONS                                                            \
     catch (pybind11::error_already_set &) {                                                       \
-        std::string err_str = pybind11::detail::error_string();                                   \
-        PyErr_Clear();                                                                            \
-        PyErr_SetString(PyExc_ImportError, ("initialization failed: " + err_str).c_str());        \
+        PyErr_SetString(                                                                          \
+            PyExc_ImportError,                                                                    \
+            ("initialization failed: " + pybind11::get_error_string_and_clear_error()).c_str());  \
         return nullptr;                                                                           \
     }                                                                                             \
     catch (const std::exception &e) {                                                             \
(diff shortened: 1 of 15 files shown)


Skylion007:
@rwgk what() is called quite a bit in big projects like PyTorch. I would not recommend removing the what() method.

rwgk:
> @rwgk what() is called quite a bit in big projects like PyTorch. I strongly oppose removing the what() method.

Currently the price we pay for `what()` is UB. I'd have to explain a lot, making that part of a core technology at Google.

pytorch is included in our global testing. I'll learn the hard way how bad it is :-)

But first thing, I have to get the CI green here.

rwgk:
> Currently the price we are paying for `what()` is UB.

I need to correct myself: computing the `what()` non-lazily is fine. The UB is because we're holding `py::objects` that may be INCREFed without the GIL held. (And if we're trying to acquire the GIL, we're in trouble when the interpreter is finalizing.)

Still trying to get everything green. It's too early to make a judgement.

jbms:
This is the behavior that one might expect based on the name `error_already_set`.  However I think there are some drawbacks:

- Most Python API functions cannot be called with an error set.  Therefore, the user needs to be aware that one of these exceptions is in flight and make sure not to call any Python API functions without first clearing/saving the python error indicator.  That constraint is manageable with Python C code that manually propagates the NULL return up the stack, but I think would not be very manageable with C++ exceptions that propagate implicitly.

- Related to above point, can't support nested errors.

- The user must not catch the exception without knowing what it is.

- Major breaking change to how error handling works in pybind11, likely to break a lot of users, and may break in ways not caught by tests since it affects only error paths.

rwgk:
The last commit (0a4d0c0b6a0ae87bbc18d288762a2c4a1aa60cf7) is what I used for exploratory global testing.  Based on that, a quick estimate of what it would take to adapt to this change Google-internally:

* "Everything": About 2000 files that match `'#include.*/pybind11\.h'` (not counting multiple copies of the pybind11 sources).
* "Portion likely affected": About 100 files that match `pybind11 error_already_set`.
* <20 files that use `.restore()`, `.what()`, `.type()`, `.value()`, `.trace()`. Only a handful of those are non-trivial to change. (This estimation is based on global testing, only fixing some build failures. Internal test ID OCL:450289283:BASE:450349644:1653281009482:4931dc01)

Very rough estimation: some of the 100 will probably not need to be touched, but a few changes may be needed elsewhere.

Net files that need changes will be around 100, i.e. about 5% of pybind11 user code.

rwgk:
I'll close this PR for now. For future reference, the CI was all green:

All checks have passed
2 skipped and 62 successful checks

PR #3968, opened 2022-05-24
ci: try CUDA 11.7

henryiii:
## Suggested changelog entry:

```rst

```


Skylion007:
@henryiii 
```/__w/pybind11/pybind11/include/pybind11/detail/../cast.h:653:1: error: 'void operator delete(void*, std::size_t)' called on unallocated object 'int_string_pair' [-Werror=free-nonheap-object]
``` 
weird, I thought we fixed this issue in a recent PR. Maybe we do need the more complicated fix they initially proposed instead of just modifying the return value policy.

henryiii:
The main issue is:

```
/__w/pybind11/pybind11/include/pybind11/stl_bind.h:91:7: error: no matching function for call to 'pybind11::class_<std::vector<int>, std::unique_ptr<std::vector<int, std::allocator<int> >, std::default_delete<std::vector<int, std::allocator<int> > > > >::def(pybind11::detail::op_<pybind11::detail::op_eq, pybind11::detail::op_l, pybind11::detail::self_t, pybind11::detail::self_t>)'
       cl.def(self == self);
```

I think the other issue might be new to 11.7.

ax3l:
Looks like the issue is present in 11.4 and newer: https://github.com/pybind/pybind11/issues/4193#issuecomment-1255646048

ax3l:
I reported the issue with the NVCC compiler team last week, but have not yet received a ticket number of tracking.

Shall we try to find a work-around for this construct in the meantime?

Skylion007:
@henryiii @ax3l Should we also try the newer PGI compilers? https://github.com/pybind/pybind11/blob/da8c730a62a7f9654433f5e2206bd1135a8b57d8/.github/workflows/ci.yml#L394

ax3l:
added in  #4220

ax3l:
Let's do NVHPC (former PGI) compilers separately. They are a different stack.

PR #3970, opened 2022-05-24
perf: Add object rvalue overload for accessors. Enables reference stealing

Skylion007:
## Description
The attr accessors had a fairly inefficient code path. Specifically, every access of that use a py::object to access it (like a py::str, py::int_, etc...), was converted to a handle and then copied back into an object, causing unnecessary reference count operations (INCREF/DECREF). We can simplify this significantly for the common case of accessing an attr using an rvalue. We just need to add an additional `object&&` specialization and using the corresponding move ctor. This doesn't change directly observable behavior, it's just a simple performance optimization eliminating unnecessary reference count operations by using the object's move ctor. 

Testing has observed up to 30% speed up in code with heavy attr accesses. Any code that accesses key or values of accessor with rvalues should benefit. 

## Suggested changelog entry:

```rst
* Added an accessor overload of `(object &&key)`to reference steal the object when using python types as keys. This prevents unnecessary reference count overhead for attr, dictionary, tuple, and sequence look ups. Added additional regression tests.
* Fixed a performance bug the caused accessor assignments to potentially perform unnecessary copies. 
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -87,7 +87,7 @@ public:
     item_accessor operator[](handle key) const;
     /// See above (the only difference is that they key is provided as a string literal)
     item_accessor operator[](const char *key) const;
-    /// See above (the only difference is that they key's reference is stolen)
+    /// See above (the only difference is that the key's reference is stolen)
     item_accessor operator[](object &&key) const;
 
     /** \rst
@@ -99,7 +99,7 @@ public:
     obj_attr_accessor attr(handle key) const;
     /// See above (the only difference is that they key is provided as a string literal)
     str_attr_accessor attr(const char *key) const;
-    /// See below  (only difference is that the key's reference is stolen)
+    /// See above (only difference is that the key's reference is stolen)
     obj_attr_accessor attr(object &&key) const;
 
     /** \rst
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -85,7 +85,7 @@ public:
         or `object` subclass causes a call to ``__setitem__``.
     \endrst */
     item_accessor operator[](handle key) const;
-    /// See above (the only difference is that they key is provided as a string literal)
+    /// See above (the only difference is that the key is provided as a string literal)
     item_accessor operator[](const char *key) const;
     /// See above (the only difference is that the key's reference is stolen)
     item_accessor operator[](object &&key) const;
@@ -97,9 +97,9 @@ public:
         or `object` subclass causes a call to ``setattr``.
     \endrst */
     obj_attr_accessor attr(handle key) const;
-    /// See above (the only difference is that they key is provided as a string literal)
+    /// See above (the only difference is that the key is provided as a string literal)
     str_attr_accessor attr(const char *key) const;
-    /// See above (only difference is that the key's reference is stolen)
+    /// See above (the only difference is that the key's reference is stolen)
     obj_attr_accessor attr(object &&key) const;
 
     /** \rst
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -85,10 +85,10 @@ public:
         or `object` subclass causes a call to ``__setitem__``.
     \endrst */
     item_accessor operator[](handle key) const;
-    /// See above (the only difference is that the key is provided as a string literal)
-    item_accessor operator[](const char *key) const;
     /// See above (the only difference is that the key's reference is stolen)
     item_accessor operator[](object &&key) const;
+    /// See above (the only difference is that the key is provided as a string literal)
+    item_accessor operator[](const char *key) const;
 
     /** \rst
         Return an internal functor to access the object's attributes. Casting the
@@ -97,10 +97,10 @@ public:
         or `object` subclass causes a call to ``setattr``.
     \endrst */
     obj_attr_accessor attr(handle key) const;
-    /// See above (the only difference is that the key is provided as a string literal)
-    str_attr_accessor attr(const char *key) const;
     /// See above (the only difference is that the key's reference is stolen)
     obj_attr_accessor attr(object &&key) const;
+    /// See above (the only difference is that the key is provided as a string literal)
+    str_attr_accessor attr(const char *key) const;
 
     /** \rst
         Matches * unpacking in Python, e.g. to unpack arguments out of a ``tuple``
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1687,7 +1687,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
+    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
     detail::tuple_iterator begin() const { return {*this, 0}; }
     detail::tuple_iterator end() const { return {*this, PyTuple_GET_SIZE(m_ptr)}; }
 };
@@ -1748,7 +1748,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::sequence_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
+    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
     detail::sequence_iterator begin() const { return {*this, 0}; }
     detail::sequence_iterator end() const { return {*this, PySequence_Size(m_ptr)}; }
 };
@@ -1768,7 +1768,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
+    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
     detail::list_iterator begin() const { return {*this, 0}; }
     detail::list_iterator end() const { return {*this, PyList_GET_SIZE(m_ptr)}; }
     template <typename T>
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -689,8 +689,8 @@ public:
     }
 
 private:
-    object ensure_object(object &&o) { return std::move(o); }
-    object ensure_object(handle h) { return reinterpret_borrow<object>(h); }
+    static object ensure_object(object &&o) { return std::move(o); }
+    static object ensure_object(handle h) { return reinterpret_borrow<object>(h); }
 
     object &get_cache() const {
         if (!cache) {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -87,6 +87,8 @@ public:
     item_accessor operator[](handle key) const;
     /// See above (the only difference is that they key is provided as a string literal)
     item_accessor operator[](const char *key) const;
+    /// See above (the only difference is that they key's reference is stolen)
+    item_accessor operator[](object &&key) const;
 
     /** \rst
         Return an internal functor to access the object's attributes. Casting the
@@ -97,6 +99,8 @@ public:
     obj_attr_accessor attr(handle key) const;
     /// See above (the only difference is that they key is provided as a string literal)
     str_attr_accessor attr(const char *key) const;
+    /// See below  (only difference is that the key's reference is stolen)
+    obj_attr_accessor attr(object &&key) const;
 
     /** \rst
         Matches * unpacking in Python, e.g. to unpack arguments out of a ``tuple``
@@ -1683,6 +1687,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
+    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
     detail::tuple_iterator begin() const { return {*this, 0}; }
     detail::tuple_iterator end() const { return {*this, PyTuple_GET_SIZE(m_ptr)}; }
 };
@@ -1743,6 +1748,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::sequence_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
+    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
     detail::sequence_iterator begin() const { return {*this, 0}; }
     detail::sequence_iterator end() const { return {*this, PySequence_Size(m_ptr)}; }
 };
@@ -1762,6 +1768,7 @@ public:
     bool empty() const { return size() == 0; }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }
     detail::item_accessor operator[](handle h) const { return object::operator[](h); }
+    detail::item_accessor operator[](object &&k) const { return object::operator[](k); }
     detail::list_iterator begin() const { return {*this, 0}; }
     detail::list_iterator end() const { return {*this, PyList_GET_SIZE(m_ptr)}; }
     template <typename T>
@@ -2057,12 +2064,20 @@ iterator object_api<D>::end() const {
     return iterator::sentinel();
 }
 template <typename D>
+item_accessor object_api<D>::operator[](object &&key) const {
+    return {derived(), std::move(key)};
+}
+template <typename D>
 item_accessor object_api<D>::operator[](handle key) const {
     return {derived(), reinterpret_borrow<object>(key)};
 }
 template <typename D>
 item_accessor object_api<D>::operator[](const char *key) const {
     return {derived(), pybind11::str(key)};
+}
+template <typename D>
+obj_attr_accessor object_api<D>::attr(object &&key) const {
+    return {derived(), std::move(key)};
 }
 template <typename D>
 obj_attr_accessor object_api<D>::attr(handle key) const {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -661,7 +661,7 @@ public:
     }
     template <typename T>
     void operator=(T &&value) & {
-        get_cache() = reinterpret_borrow<object>(object_or_cast(std::forward<T>(value)));
+        get_cache() = ensure_object(object_or_cast(std::forward<T>(value)));
     }
 
     template <typename T = Policy>
@@ -689,6 +689,9 @@ public:
     }
 
 private:
+    object ensure_object(object &&o) { return std::move(o); }
+    object ensure_object(handle h) { return reinterpret_borrow<object>(h); }
+
     object &get_cache() const {
         if (!cache) {
             cache = Policy::get(obj, key);
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1689,8 +1689,10 @@ public:
     size_t size() const { return (size_t) PyTuple_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<detail::is_pyobject<T>::value, int> = 0>
+    detail::item_accessor operator[](T o) const {
+        return object::operator[](std::forward<T>(o));
+    }
     detail::tuple_iterator begin() const { return {*this, 0}; }
     detail::tuple_iterator end() const { return {*this, PyTuple_GET_SIZE(m_ptr)}; }
 };
@@ -1750,8 +1752,10 @@ public:
     }
     bool empty() const { return size() == 0; }
     detail::sequence_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<detail::is_pyobject<T>::value, int> = 0>
+    detail::item_accessor operator[](T o) const {
+        return object::operator[](std::forward<T>(o));
+    }
     detail::sequence_iterator begin() const { return {*this, 0}; }
     detail::sequence_iterator end() const { return {*this, PySequence_Size(m_ptr)}; }
 };
@@ -1770,8 +1774,10 @@ public:
     size_t size() const { return (size_t) PyList_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<detail::is_pyobject<T>::value, int> = 0>
+    detail::item_accessor operator[](T o) const {
+        return object::operator[](std::forward<T>(o));
+    }
     detail::list_iterator begin() const { return {*this, 0}; }
     detail::list_iterator end() const { return {*this, PyList_GET_SIZE(m_ptr)}; }
     template <typename T>
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -2073,24 +2073,24 @@ iterator object_api<D>::end() const {
     return iterator::sentinel();
 }
 template <typename D>
-item_accessor object_api<D>::operator[](object &&key) const {
-    return {derived(), std::move(key)};
-}
-template <typename D>
 item_accessor object_api<D>::operator[](handle key) const {
     return {derived(), reinterpret_borrow<object>(key)};
+}
+template <typename D>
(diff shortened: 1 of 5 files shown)


henryiii (approved):

rwgk:
I tend to be deliberately a bit ignorant towards low-level mechanics***, but trying to understand this PR, is the following correct?

* Both `handle` and `object` are just `PyObject *` underneath.
* Copying a pointer or moving a pointer does not make a difference in terms of runtime performance.
* Deleting an `object` means DECREF, and creating a new one with `reinterpret_borrow<object>` means `INCREF`.

Are those DECREF/INCREF what this PR is eliminating?

If that's the point, what kind of test could ensure that this optimization is not accidentally undone?

Also, did you already run a micro benchmark to quantify the best-case performance gain? — Something temporary and quick & dirty to get a rough idea would seem sufficient.

________________________________________________________________

*** I'm generally more focused on high-level aspects & safety than low-level efficiency considerations, trusting that modern compilers will optimize out obvious inefficiencies. I believe cluttering or complicating code to squeeze out a tiny bit of extra efficiency negatively impacts readability & long-term maintainability, and it makes future development work & refactoring more bug-prone. That could result in a significant waste of human time, the most expensive resource we have, just to save a tiny amount of very cheap machine time.

Skylion007:
>> Are those DECREF/INCREF what this PR is eliminating?
Yes

rwgk:
> > > Are those DECREF/INCREF what this PR is eliminating?
> > > Yes

What test could ensure that this optimization is not accidentally undone?

Skylion007:
>>> What test could ensure that this optimization is not accidentally undone?
Writing the tests for that would be pretty difficult. I would normally use constructor_stats.h, but we would need to override the ctors and dtor of the object class to measure that. Any thoughts on how to do it @rwgk? I think it's more of a style issue TBH. reinterpret_borrow should not be used on objects. Maybe we could add an overload to reinterpret_borrow that throws an exception if it's an pyobject (to prevent the implicit conversion of a handle). There is no reason to use it with objects, but it can be done due to the implicit handle conversion. Thoughts @rwgk?

Skylion007:
@rwgk TLDR: the real evil here is the antipattern of `objects -> handle -> objects` it does incref / decrefs for no reason. I am not sure of a way to guard against it though since `handle -> object` are a valid use case. I normally agree with you about trusting the compiler to do this stuff for us, but this is an obvious case where it is failing, the overhead of maintainig these overloads is negligible.

Skylion007:
@rwgk Also, I wouldn't normally add these extra overloads just for that, but the fact that is such a hot code path (any attr accesses) made me want to optimize it.

rwgk:
To me it feels like we're driving blindly (no performance measurements) and without guard rails (test to ensure the optimization is not accidentally undone). Re performance measurements: I believe that is only a few minutes worth of effort, to get a rough idea at least; i.e. we're driving blindly unnecessarily. Re tests, letting basic best practices slip needs to be clearly stated in the PR description, with a rationale, to prove (incl. to ourselves) that we carefully thought about it.

Skylion007:
@rwgk Care to do the benchmarking then? We currently don't have any benchmarking framework on master at all currently.

Skylion007:
@rwgk Googling, this is effectively the same as benchmarking incref / decref and I don't expect any performance difference there from the CPython benchmarks. Do not expect any speed significant difference from this PR anyhow. I'll update the PR description with this explanation.

rwgk:
> @rwgk Care to do the benchmarking then? We currently don't have any benchmarking framework on master at all currently.

We don't need one for this (although it would be nice to have), a quick 10 minute thing like this will do:

https://github.com/rwgk/rwgk_tbx/blob/main/test_perf_error_already_set.cpp
https://github.com/rwgk/rwgk_tbx/blob/main/test_perf_error_already_set.py

Just post a link to your code and the results here.

rwgk:
I've been staring at this code for a while:
```
    template <typename T>
    void operator=(T &&value) & {
        get_cache() = ensure_object(object_or_cast(std::forward<T>(value)));
    }
```
With
```
private:
    static object ensure_object(object &&o) { return std::move(o); }
    static object ensure_object(handle h) { return reinterpret_borrow<object>(h); }

    object &get_cache() const {
        if (!cache) {
            cache = Policy::get(obj, key);
        }
        return cache;
    }
private:
    handle obj;
    key_type key;
    mutable object cache;```
```

I.e. in `operator=(T &&value)`, on the lhs we have a `object &` (r-value) and on the right side `object` (r-value) IIUC.

Is the compiler forced or only permitted to elide the copy here?

Mainly I was looking for a way to test that the INCREF/DECREF cycle is in fact eliminated, but I failed to see one.

But what stood out to me, does `cache` have to be `object`? Could we make it `handle`? Then we have obvious & full control over when INCREF and DECREF are invoked. AFAICT, if we update the (copy, move) x (ctor, operator=) and dtor accordingly we should be fine. Could that unlock other manual optimizations?

Generally thinking:

* If we had a test that proves the INCREF/DECREF cycle is in fact eliminated, ok. We don't know how much that helps, but at least we're sure it's gone and won't come back.
* If we could at least measure a gain, at least one platform, ok.

But without either I feel very uneasy about making the code more complex.

If none of us can see a way to implement a test, then a rough benchmark would be useful. If and only if that gives us measurements that prove it's worth the trouble, the `handle` approach might be interesting, although with such measurements I'd also be OK with this PR as-is.

Skylion007:
>> Is the compiler forced or only permitted to elide the copy here? 
It's explicitly allowed to on C++14, but not requited to elide it. It's explicitly required to elide the copy on C++17.

rwgk:
> It's explicitly allowed to on C++14, but not requited to elide it. It's explicitly required to elide the copy on C++17. I think it should elide it on C++11.

I looked around some more and now I'm thinking that my "Is the compiler forced or only permitted to elide the copy here?" was actually besides the point. The correct question: "Is the compiler forced or only permitted to use the move ctor here?" I don't know all the rules that determine that, but my guess is it is actually forced to use the move ctor here. So that's great.

Could you suggest a code snippet for a best-case micro benchmark? Just the part to loop over, I could plug that into my ad-hoc benchmark code I pointed out earlier. That's really quick.

Skylion007:
> The correct question: "Is the compiler forced or only permitted to use the move ctor here?" I don't know all the rules that determine that, but my guess is it is actually forced to use the move ctor here. So that's great.

Yes, it's forced to use the move ctor.

Skylion007:
@rwgk I could also just do:
```
@@ -1689,8 +1689,8 @@ public:
     size_t size() const { return (size_t) PyTuple_Size(m_ptr); }
     detail::tuple_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<!std::is_integral<T>::value, int> = 0>
+    detail::tuple_accessor operator[](T o) const { return object::operator[](std::forward<T>(o)); }
     detail::tuple_iterator begin() const { return {*this, 0}; }
     detail::tuple_iterator end() const { return {*this, PyTuple_GET_SIZE(m_ptr)}; }
 };
@@ -1750,8 +1750,8 @@ public:
     }
     detail::sequence_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<!std::is_integral<T>::value, int> = 0>
+    detail::sequence_accessor operator[](T o) const { return object::operator[](std::forward<T>(o)); }
     detail::sequence_iterator begin() const { return {*this, 0}; }
     detail::sequence_iterator end() const { return {*this, PySequence_Size(m_ptr)}; }
 };
@@ -1770,8 +1770,8 @@ public:
     size_t size() const { return (size_t) PyList_Size(m_ptr); }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
-    detail::item_accessor operator[](object &&o) const { return object::operator[](std::move(o)); }
+    template <typename T, detail::enable_if_t<!std::is_integral<T>::value, int> = 0>
+    detail::list_accessor operator[](T o) const { return object::operator[](std::forward<T>(o)); }
     detail::list_iterator begin() const { return {*this, 0}; }
     detail::list_iterator end() const { return {*this, PyList_GET_SIZE(m_ptr)}; }
     template <typename T>
```
which I guess is a little less readable, but more maintainable?

Skylion007:
@rwgk I used templates to make it more verbose, but also more future proof. Hopefully this addresses your maintainability concerns.

include/pybind11/pytypes.h
```diff
@@ -1761,7 +1774,10 @@ class list : public object {
     size_t size() const { return (size_t) PyList_Size(m_ptr); }
     bool empty() const { return size() == 0; }
     detail::list_accessor operator[](size_t index) const { return {*this, index}; }
-    detail::item_accessor operator[](handle h) const { return object::operator[](h); }
+    template <typename T, detail::enable_if_t<detail::is_pyobject<T>::value, int> = 0>
+    detail::item_accessor operator[](T o) const {
```
Skylion007, line 1778: @rwgk Okay, this is a bit more verbose now and uses templates, but at least it's more general.

rwgk:
Is this on the right track?

https://github.com/pybind/pybind11/pull/3977/commits/5cd9f3ff58cf3040fde4aedb06b290df4f72a136

```C++
obj.attr(attr_name) = attr;
```

Embedded like this:

```C++
void perf_object(std::size_t num_iterations,
                 py::object obj,
                 const py::object &attr_name,
                 const py::object &attr) {
    while (num_iterations) {
        obj.attr(attr_name) = attr;
        num_iterations--;
    }
}
```

I got these initial measurements (num_iterations, test_time_secs, per_call_μs):
```
master PERF 126533164,4.193s,0.331413μs
pr3970 PERF 125816446,4.189s,0.332976μs
```

I.e. identical times, but I don't really know what I'm doing with the `obj.attr(attr_name) = attr;` line, it's only based on how I understood your suggestion under https://github.com/pybind/pybind11/pull/3970#issuecomment-1138696796. And there is always the possibility that I screwed up somewhere in this first attempt; i.e. it definitely needs more scrutiny before deriving conclusions.

I'll try to drill down, to be sure it is actually going through the code path changed by this PR, but if you see something that is obviously wrong please let me know. — My goal is to have a best-case (but "real") piece of code where you currently see `obj.attr(attr_name) = attr;`.

For completeness, some info about my platform & build:

```
Debian clang version 13.0.1-3+build2
Target: x86_64-pc-linux-gnu
Thread model: posix

clang++ -o pybind11/tests/test_perf_accessors.os -c -std=c++17 -fPIC -fvisibility=hidden -Os -flto -Wall -Wextra -Wconversion -Wcast-qual -Wdeprecated -Wnon-virtual-dtor -Wunused-result -isystem /usr/include/python3.9 -isystem /usr/include/eigen3 -DPYBIND11_STRICT_ASSERTS_CLASS_HOLDER_VS_TYPE_CASTER_MIX -DPYBIND11_TEST_BOOST -Ipybind11/include -I/usr/local/google/home/rwgk/forked/pybind11/include -I/usr/local/google/home/rwgk/clone/pybind11/include /usr/local/google/home/rwgk/forked/pybind11/tests/test_perf_accessors.cpp
clang++ -o lib/pybind11_tests.so -shared -fPIC -Os -flto -shared pybind11/tests/pybind11_tests.os pybind11/tests/test_perf_accessors.os

platform linux -- Python 3.9.12, pytest-6.2.3, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python3
```

rwgk:
Just to let you know, **in a rush** (I'll be out for a few hours now)

the previous was not hitting the code path we discussed yesterday

by running all unit tests I found exactly one case that hits it

this is what I'm seeing right now

```
master PERF 204265501,4.193s,0.205279μs
pr3970 PERF 216456859,4.168s,0.192548μs
```

```
                 py::object obj,
                 const py::object &attr_name,
                 const py::object &attr) {
    auto l = py::list(1);
    l[0] = 0;
    auto var = l[0];
    while (num_iterations) {
        var = num_iterations;
        num_iterations--;
    }
}
```

rwgk:
PR #3977 now has a unit test for the list accessor that is sensitive to this PR.

https://github.com/rwgk/pybind11/pull/18 is a clone of this PR rebased on PR #3977, with the expected `inc_refs` adjusted.

First I'm running with `-DNDEBUG` `-DPYBIND11_HANDLE_REF_DEBUG` (here the `inc_refs` are important but NOT the per-iteration times in μs, because I'm also incrementing a global counter in `handle`):

With PR #3977:
```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=[1, 1]
PERF 0,246538867,2.106s,0.085436μs
PERF 1,246538867,3.293s,0.133584μs
```

With https://github.com/rwgk/pybind11/pull/18:
```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=[1, 0]
PERF 0,240448964,2.130s,0.088568μs
PERF 1,240448964,3.080s,0.128111μs
```

The diff in `inc_refs` proves that this PR successfully eliminates an INCREF/DECREF pair.

Then I'm running just with `-DNDEBUG` (here the per-iteration times in μs ARE important, the global counter is ifdef-ed out):

With PR #3977:
```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,275052739,2.101s,0.076378μs
PERF 1,275052739,3.182s,0.115703μs
```

With https://github.com/rwgk/pybind11/pull/18:
```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,233334018,2.095s,0.089781μs
PERF 1,233334018,2.958s,0.126780μs
```

When running the test repeatedly the variations of the per-iteration μs times are in the 3rd decimal. Additionally, I hand-picked the worst of 3 for "With PR #3977" and the best of 3 for "With https://github.com/rwgk/pybind11/pull/18".

I don't understand why the per-iteration times with this PR are slightly longer, even though there is a proof that the INCREF/DECREF pair is eliminated. Can you see a problem with my test?

Also, I'm still not certain that I have a best-case situation, but it is down to a plain `var = 42;`

I don't know the trick for getting those outputs when using cmake. I'm using these commands:

```
scons -j 16 extra_defines=NDEBUG,PYBIND11_HANDLE_REF_DEBUG meta_opts=lto selected_test_cpp=test_perf_accessors.cpp
/usr/bin/python3 $HOME/clone/pybind11_scons/run_tests.py ../pybind11 test_perf_accessors.py -v -s
scons -j 16 extra_defines=NDEBUG meta_opts=lto selected_test_cpp=test_perf_accessors.cpp
/usr/bin/python3 $HOME/clone/pybind11_scons/run_tests.py ../pybind11 test_perf_accessors.py -v -s
```

Probably if there is a way to pass `-DPYBIND11_HANDLE_REF_DEBUG` and `pytest -v -s` you can get the outputs also with cmake.

For completeness, here are the exact commands as issued by scons:
```
clang++ -o pybind11/tests/test_perf_accessors.os -c -std=c++17 -fPIC -fvisibility=hidden -Os -flto -Wall -Wextra -Wconversion -Wcast-qual -Wdeprecated -Wnon-virtual-dtor -Wunused-result -isystem /usr/include/python3.9 -isystem /usr/include/eigen3 -DNDEBUG -DPYBIND11_HANDLE_REF_DEBUG -DPYBIND11_STRICT_ASSERTS_CLASS_HOLDER_VS_TYPE_CASTER_MIX -DPYBIND11_TEST_BOOST -Ipybind11/include -I/usr/local/google/home/rwgk/forked/pybind11/include -I/usr/local/google/home/rwgk/clone/pybind11/include /usr/local/google/home/rwgk/forked/pybind11/tests/test_perf_accessors.cpp

clang++ -o lib/pybind11_tests.so -shared -fPIC -Os -flto -shared pybind11/tests/pybind11_tests.os pybind11/tests/test_perf_accessors.os
```

rwgk:
One more quick measurement, after local `git revert` of the "future proof operator overloads" commit, picking the best of 3 runs:

```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,232804504,2.096s,0.090036μs
PERF 1,232804504,2.966s,0.127395μs
```

It looks pretty much identical to the numbers in the previous comment, a tiny amount slower, but I'd bet that's just noise.

rwgk:
Now all the same except using `-O3` instead of `-Os`, only reporting 1. the worst of 3 for master, 2. the best of 3 for this PR:

```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,438143155,2.103s,0.048007μs
PERF 1,438143155,3.687s,0.084145μs
```

```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,439159638,2.097s,0.047740μs
PERF 1,439159638,2.613s,0.059508μs
```

Now this PR is helping!
```
>>> 0.059508 / 0.084145
0.7072077960663141
```

Yay!

rwgk:
For completeness, here with the "future proof operator overloads" commit re-applied:

```
test_perf_accessors.py::test_perf_list_accessor
inc_refs=None
PERF 0,440602540,2.101s,0.047685μs
PERF 1,440602540,2.615s,0.059353μs
```

Same as above (without the commit).

Skylion007:
@rwgk Yeah, that https://github.com/pybind/pybind11/pull/3970#issuecomment-1138956401 unit test was wrong and not triggering. Both attr and attr_name are const reference, aka pointers, which make them lvalues. They should therefore hit the py handle codepath. To trigger this code path, you need the attr to be an rvalue (ie. be a nameless temporary or be an std::move from value) and also be an object.


Now that we know this helps on a pretty hot code path (at least in normal release mode) and I forward the arguments to the appropriate overload with templates, are you satisfied with the current PR?

rwgk:
If the optimization is considered important, there has to be a test for it, in this PR. IMO it would be an abuse of maintainer status to let that slip. A test is the only assurance we have that the optimization doesn't accidentally get undone. The mechanics involved here are brittle and need a lot of attention to detail. It cannot be expected that everybody working on this library is aware of all nuances everywhere, especially in the middle of some complex refactoring.

There is still a glitch with my approach: the global counter could race if pytype runs multiple threads and `var = 42;` somehow gets Python to release & reacquire the GIL. If there is a race, the `inc_refs=[1, 0]` could become `[1+N, M]`. Does anyone know, does pytype run multiple threads (e.g. I see `make pytest -j 2` in the CI)? Is there a reasonably easy trick to avoid a race? — A cheap & dirty but maybe OK way around this could be to produce a message explaining the race possibility if we see something different from `[1, 0]`. Quite possibly, it will never fire, because `var = 42;` never touches Python code that releases & reacquires the GIL.

Also, the test needs to be extended to tuple, sequence, and attr.

I have to say, just doing the clever fun part, but then asking others "care to" do the due diligence, not offering any clues or answering questions until there is a breakthrough, is very poor team spirit IMO. If this optimization is important to you, please complete the work adhering to basic best practices. If you feel that's a burden, that casts serious doubts on the importance of this work. Next time, better don't start if you feel seeing it all the way through is too much trouble. I teased out the right combination of things to see a best-case 30% gain, but I also know that in real applications it's likely to do next to nothing. You need months worth of effort working on many micro optimizations like this before you see any impact for average applications. If you have observations like "in my application the runtime went from X to Y", please report it, so that we know it's worth an effort.

Skylion007:
@rwgk Sorry about missing your initial post from 21 hours ago, and thank you so much for helping with the due dilligence on this PR. I am happy to answer any other questions you have about it. Thanking you so much for providing these benchmark numbers. I am happy to enumerate test cases that should hit this PR.

The numbers you report in the other comment are correct (from extracting the data from our unit tests), I just wanted to explain why the code path wasn't being hit. 30% speed up looks correct to me. The most recent commit just fixes a template subtly that might have caused issues in the future proof version.

>  the global counter could race if pytype runs multiple threads and var = 42; somehow gets Python to release & reacquire the GIL

That's exactly why I want to get rid of the incref / decrefs, more than the optimization use case. Reduce the number of incs and decs to the global counter. In regards to the optimization being undone, this is also pretty stable, rather untouched part of the code base (the methods I am adding to overloads have been untouched since 2016).

I do wish we had optimization tests more globally. I would like to point out this is a very hotpath. I am happy to write tests that test out the fast path (although it should already be tested by pybind11 itself, pybind11.h triggers those code path several places in the `py::enum_base` and `py::class` classes.

rwgk:
> That's exactly why I want to get rid of the incref / decrefs, more than the optimization use case. Reduce the number of incs and decs to the global counter.

That is a very slippery slope:

* I'd need to check what the documentation says at the minute, but I hope it's either nothing or "For any operations involving pytype.h types you must hold the GIL."
* If someone asks, we could say "There are a few operations that do not actually require the GIL, be we do not recommend exploiting that."

Because a year later someone else looks at their code, thinking "sure they are holding the GIL here", and adds an innocently looking line that does require the GIL. Two weeks later we have an emergency-level production bug. — Better don't play with fire.

> In regards to the optimization being undone, this is also pretty stable, rather untouched part of the code base (the methods I am adding to overloads have been untouched since 2016).

It's like running a red light. You do it once, twice, ten times, eventually you need a new car. And a new arm. — I'd want to make exceptions only for really tiny things that are OK if we lose them, or if it is unreasonably onerous. I don't think it is here. On the contrary, establishing a way to test such optimization could pay off for other things as well.

> pybind11.h triggers those code path several places in the py::enum_base and py::class classes.

Yesterday or the day before, I inserted `pybind11_fail()` in `operator=`, ran all unit tests, and there was only exactly one subtest that failed: test_pytypes.py `test_accessors()` `d = m.accessor_assignment()`. But that's the only code path for which I did that.

We have global performance monitoring of some sorts looking at production jobs. That might be one way to pin-point high-value optimization targets (assuming our usage is representative for the world at large; probably). I never use those, though, would need to learn.

Skylion007:
@rwgk I've also gone ahead and added some additional unit tests to address your concerns. This ensures that regardless of other refactoring, we should have coverage of the new overload path.

Skylion007:
> Yesterday or the day before, I inserted pybind11_fail() in operator=, ran all unit tests, and there was only exactly one subtest that failed: test_pytypes.py test_accessors() d = m.accessor_assignment(). But that's the only code path for which I did that.

When testing, I inserted pybind11_fail for `object_api<D>::operator[](object &&)` and `object_api<D>::attr(object &&)` and I wasn't even able to import any pybind11 modules due to that codepath being used so often in pybind11. The operator= bit is definitely more of a niche improvement.

rwgk:
(I'll be out for a few hours. I'll catch up when I'm back.)

rwgk:
The new tests look great!

I'm trying `thread_local` for the `handle::inc_ref_counter()` (https://github.com/pybind/pybind11/pull/3977/commits/d6dda2df61175b895aa39f59ceb0f2cd7c6f92d5). If that works reliably on just some platforms, I believe that's all we need for a simple and rock-solid test that ensures the moves are working and will not get lost accidentally.

I don't think we need to drag the benchmark code with us, even though it's pretty simple, too. But it's easy enough to pull in as needed to answer specific questions.

rwgk:
`thread_local` works with all 16 debug builds that we currently have (below; determined by intentionally breaking #3977).

15 Windows debug builds.
Only 1 non-Windows: deadsnakes Valgrind

Not a great sampling, but plenty of coverage for tests using handle::inc_ref_counter(). Only just one would give us the assurance that the new moves will not regress.

I'll fix up #3977 again.

```
1______3.6_____MSVC_2019_____x86.txt
1______3.8_____MSVC_2019__Debug______x86_-DCMAKE_CXX_STANDARD=17.txt
1______3.9-dbg__deadsnakes______Valgrind_____x64.txt
1______3.9_____MSVC_2022_C++20_____x64.txt
2______3.7_____MSVC_2019_____x86.txt
2______3.9_____MSVC_2019__Debug______x86_-DCMAKE_CXX_STANDARD=20.txt
3______3.8_____MSVC_2019_____x86_-DCMAKE_CXX_STANDARD=17.txt
4______3.9_____MSVC_2019_____x86_-DCMAKE_CXX_STANDARD=20.txt
7______3.6_____windows-2022_____x64.txt
8______3.9_____windows-2022_____x64.txt
9______3.10_____windows-2022_____x64.txt
10______pypy-3.7_____windows-2022_____x64.txt
11______pypy-3.8_____windows-2022_____x64.txt
12______pypy-3.9_____windows-2022_____x64.txt
19______3.6_____windows-2019_____x64_-DPYBIND11_FINDPYTHON=ON.txt
20______3.9_____windows-2019_____x64.txt
```

rwgk:
https://github.com/pybind/pybind11/pull/3977 is now almost what I believe we need here. The perf test is gone. The inc_refs code is in test_pytypes. It covers list and tuple, but sequence and attr are still missing. It's a pretty small test, except for the really clunky repetitive added code in test_pytypes.cpp. But I'm too sleepy to complete it right now.

rwgk:
I'm done with https://github.com/pybind/pybind11/pull/3977, see the PR description.

Could you please review the new unit tests? I know they are fully covering the behavior changes, but if you can think of improvements, could you please apply them directly to 3977? After you're done, I will go through again top-to-bottom to verify that the behavior change coverage is still complete.

When we're both happy with 3977, could you please transfer the unit tests here, without the `PYBIND11_PR3970` `#ifdef`s?

include/pybind11/pytypes.h
```diff
@@ -2057,6 +2073,10 @@ iterator object_api<D>::end() const {
     return iterator::sentinel();
 }
 template <typename D>
+item_accessor object_api<D>::operator[](object &&key) const {
```
rwgk, line 2076: The declaration appears after the overload for `handle`. Could you please put them in the same order here?

include/pybind11/pytypes.h
```diff
@@ -2065,6 +2085,10 @@ item_accessor object_api<D>::operator[](const char *key) const {
     return {derived(), pybind11::str(key)};
 }
 template <typename D>
+obj_attr_accessor object_api<D>::attr(object &&key) const {
```
rwgk, line 2088: Please keep the order of declarations here, too.

rwgk:
I went ahead and modified the PR description, to be explicit out the reference counting overhead. Please correct in case I didn't get it right.

Skylion007:
@rwgk Do I need to add a special define to the CI? It doesn't seem to be running the incref tests.

PR #3973, opened 2022-05-26
Add missing error handling to `module_::def_submodule`

rwgk:
## Description
Extracted from PR #3964, where the missing error handling happened to cause some confusion.

## Suggested changelog entry:

```rst
`module_::def_submodule` was missing proper error handling. This is fixe now.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1173,9 +1173,16 @@ public:
             py::module_ m3 = m2.def_submodule("subsub", "A submodule of 'example.sub'");
     \endrst */
     module_ def_submodule(const char *name, const char *doc = nullptr) {
-        std::string full_name
-            = std::string(PyModule_GetName(m_ptr)) + std::string(".") + std::string(name);
-        auto result = reinterpret_borrow<module_>(PyImport_AddModule(full_name.c_str()));
+        const char *this_name = PyModule_GetName(m_ptr);
+        if (this_name == nullptr) {
+            throw error_already_set();
+        }
+        std::string full_name = std::string(this_name) + "." + name;
+        handle submodule = PyImport_AddModule(full_name.c_str());
+        if (!submodule) {
+            throw error_already_set();
+        }
+        auto result = reinterpret_borrow<module_>(submodule);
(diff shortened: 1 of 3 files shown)


include/pybind11/pybind11.h
```diff
@@ -1173,9 +1173,16 @@ class module_ : public object {
...
     \endrst */
     module_ def_submodule(const char *name, const char *doc = nullptr) {
-        std::string full_name
-            = std::string(PyModule_GetName(m_ptr)) + std::string(".") + std::string(name);
-        auto result = reinterpret_borrow<module_>(PyImport_AddModule(full_name.c_str()));
+        const char *this_name = PyModule_GetName(m_ptr);
+        if (this_name == nullptr) {
+            throw error_already_set();
+        }
+        std::string full_name = std::string(this_name) + "." + name;
```
Skylion007, line 1180:
```suggestion
        std::string full_name = std::string(this_name) + '.' + name;
```

include/pybind11/pybind11.h
```diff
@@ -1173,9 +1173,16 @@ class module_ : public object {
...
+        handle submodule = PyImport_AddModule(full_name.c_str());
+        if (!submodule) {
+            throw error_already_set();
+        }
+        auto result = reinterpret_borrow<module_>(submodule);
```
Skylion007, line 1185: For reference, wouldn't you also been able to do? 
```
if(!result){
            throw error_already_set();
}
```
This looks good either way though.

Skylion007 (approved):
Approve with minor nit: (Append a char instead of a 1char string literal, it's a faster overload.)

Skylion007 (approved):

include/pybind11/pybind11.h
rwgk: IIU your question correctly, that would assume `reinterpret_borrow<module_>` is OK with `nullptr`, and does not care about the active error. While that is usually true, what if someone temporarily adds debugging code to `reinterpret_borrow`, or the `module_` constructor in this case? Their fault, of course, but why build a trap like that?
Generally, I think it is bad style to separate error handling from the source of the error more than necessary.

PR #3974, opened 2022-05-26
Avoid `catch (...)` for expected `import numpy` failures

rwgk:
## Description
Extracted from PR #3964, where the `catch (...)` happened to cause significant confusion because they are silently dropping any kind of error. This is likely to cause more confusion in future development work. Narrowing down to `catch (const py::error_already_set &)` is a simple way to significantly reduce the potential for silently dropping important errors (although that is still not perfect).

## Suggested changelog entry:

```rst

```

### DIFF
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -162,7 +162,7 @@ static int data_i = 42;
 TEST_SUBMODULE(numpy_array, sm) {
     try {
         py::module_::import("numpy");
-    } catch (...) {
+    } catch (const py::error_already_set &) {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


henryiii (approved):

henryiii:
Can we restrict any usage of `catch (...)` in the codebase?

Skylion007 (approved):
LGTM, let's merge it.

Skylion007:
> Can we restrict any usage of `catch (...)` in the codebase?

@henryiii Unfortunately, no. We need to use it for exception translators / exception handling so at least need to be present there. We also need it for noexcept handling as well. Finally, I am not aware of clang-tidy rules or such to restrict this anyway.

PR #3980, opened 2022-05-28
chore: perfectly forward all make_iterator args

Skylion007:
## Description
This forwards all args in make_iterator. This should be more efficient and allow downstream functions to use the rvalue / lvalue nature of the args).
* I also added a missing move argument and emplace_back for an lambda.

## Suggested changelog entry:

```rst
* Perfect forward all args of make_iterator.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2326,7 +2326,7 @@ template <typename Access,
           typename Sentinel,
           typename ValueType,
           typename... Extra>
-iterator make_iterator_impl(Iterator first, Sentinel last, Extra &&...extra) {
+iterator make_iterator_impl(Iterator &&first, Sentinel &&last, Extra &&...extra) {
     using state = detail::iterator_state<Access, Policy, Iterator, Sentinel, ValueType, Extra...>;
     // TODO: state captures only the types of Extra, not the values
 
@@ -2352,7 +2352,7 @@ iterator make_iterator_impl(Iterator first, Sentinel last, Extra &&...extra) {
                 Policy);
     }
 
-    return cast(state{first, last, true});
+    return cast(state{std::forward<Iterator>(first), std::forward<Sentinel>(last), true});
 }
 
 PYBIND11_NAMESPACE_END(detail)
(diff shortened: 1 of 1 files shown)


rwgk (approved):
To explain my rationale, in contrast to PR #3970: here you are just inserting a few `&&` and `std::move()`, and use `emplace_back()` instead of `push_back()`. If we lose any of those, optimizations again, so be it.

In PR #3970, you are adding new functions and member functions. That's clearly a level up in terms of code complexity, and there is also more danger that overloads get lost unnoticed. I want to maintain a quality standard that includes: if an entire (member) function is removed accidentally, we need to see test breakages.

henryiii:
FYI, I think this change is causing ambiguous resolution:

```
In file included from /Users/henryschreiner/git/scikit-hep/boost-histogram/src/register_axis.cpp:10:
/Users/henryschreiner/git/scikit-hep/boost-histogram/include/bh_python/register_axis.hpp:252:24: error: call to 'make_iterator' is ambiguous
                return py::make_iterator(begin, end);
                       ^~~~~~~~~~~~~~~~~
/Users/henryschreiner/git/scikit-hep/boost-histogram/src/register_axis.cpp:22:19: note: in instantiation of function template specialization 'register_axis<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0>>>' requested here
        auto ax = register_axis<T>(mod);
                  ^
/Users/henryschreiner/git/scikit-hep/boost-histogram/src/register_axis.cpp:28:5: note: in instantiation of function template specialization 'register_axis_each<boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<0>>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<1>>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<2>>, boost::histogram::axis::regular<double, boost::use_default, metadata_t>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<11>>, boost::histogram::axis::regular<double, boost::use_default, metadata_t, boost::histogram::axis::option::bitset<6>>, axis::regular_numpy, (lambda at /Users/henryschreiner/git/scikit-hep/boost-histogram/src/register_axis.cpp:34:50)>' requested here
    register_axis_each<axis::regular_none,
    ^
/Users/henryschreiner/git/scikit-hep/boost-histogram/extern/pybind11/include/pybind11/pybind11.h:2373:10: note: candidate function [with Policy = pybind11::return_value_policy::reference_internal, Iterator = iterator &, Sentinel = iterator &, ValueType = pybind11::tuple, Extra = <>]
iterator make_iterator(Iterator &&first, Sentinel &&last, Extra &&...extra) {
         ^
/Users/henryschreiner/git/scikit-hep/boost-histogram/extern/pybind11/include/pybind11/pybind11.h:2425:10: note: candidate function [with Policy = pybind11::return_value_policy::reference_internal, Type = iterator, Extra = <iterator &>]
iterator make_iterator(Type &value, Extra &&...extra) {
         ^
```

I can fix it by adding a `std::move`, but I'd think that shouldn't be required to avoid ambiguity.

PR #3985, opened 2022-06-02
STL map bind naming fix for Sphinx and Pylance

themarpe:
## Description

Addresses a change in: https://github.com/pybind/pybind11/pull/3310
Discussion/rationale can be found over at: https://github.com/pybind/pybind11/pull/3310#discussion_r887191310

## Suggested changelog entry:

```rst
Fixes the class names generated by ``bind_map``:
- Change the naming from ``*View[name]`` to ``*View_name``
```

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -674,11 +674,11 @@ class_<Map, holder_type> bind_map(handle scope, const std::string &name, Args &&
 
     Class_ cl(scope, name.c_str(), pybind11::module_local(local), std::forward<Args>(args)...);
     class_<KeysView> keys_view(
-        scope, ("KeysView[" + name + "]").c_str(), pybind11::module_local(local));
+        scope, ("KeysView_" + name).c_str(), pybind11::module_local(local));
     class_<ValuesView> values_view(
-        scope, ("ValuesView[" + name + "]").c_str(), pybind11::module_local(local));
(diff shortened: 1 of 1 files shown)


Skylion007:
Okay, so this probably isn't the right the fix. This is caused by a bug where the typing name should be `KeyView[class_name]` , not `class_name.KeyView[class_name].` If we fix this, MyPy etc... should be happy with it. This just makes the KeyView, ViewName etc function as it's own class, when really it should just be a specializing of the typing.KeyView, ItemView etc...

themarpe:
Thanks @Skylion007
Let me know how you'd go about doing this change - feel free to request changes here or if separate PR is needed

Skylion007:
Okay, so I tried something and it didn't quite work unfortunately. We need some way to get mypy to pickup on that fact that class_name.KeyView[class_name] is an alias / subclass of typing.KeyView, but I am not sure what the best way to do that is. We would need a reproducer of the mypy complaints and see if there isn't away we can fix it with an alias, subclass, typing mixin, etc... 

@rwgk @henryiii I am stumped any idea how to proceed? I am going to open an issue to track this.

themarpe:
Thanks for the information layed out in the issue - this was beyond my understanding of the issue at hand. 
Looking forward on the debate evolving there

themarpe:
Seems to be superseded by: https://github.com/pybind/pybind11/pull/4353

PR #3994, opened 2022-06-06
chore: optimize dictionary access in strip_padding numpy

Skylion007:
## Description
Further optimize strip_padding by making construction of field_descriptors faster by adding a ctor, using emplace, and trying to use the native C++ APIs for iterating through the dictionary.

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -648,10 +648,9 @@ private:
         std::vector<field_descr> field_descriptors;
         field_descriptors.reserve(field_dict.size());
 
-        for (auto field : field_dict.attr("items")()) {
-            auto spec = field.cast<tuple>();
-            auto name = spec[0].cast<pybind11::str>();
-            auto spec_fo = spec[1].cast<tuple>();
+        for (auto field : field_dict) {
+            auto name = field.first.cast<pybind11::str>();
+            auto spec_fo = field.second.cast<tuple>();
             auto format = spec_fo[0].cast<dtype>();
             auto offset = spec_fo[1].cast<pybind11::int_>();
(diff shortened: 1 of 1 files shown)


include/pybind11/numpy.h
```diff
@@ -641,20 +641,23 @@ class dtype : public object {
...
+        auto field_dict = attr("fields").cast<dict>();
         std::vector<field_descr> field_descriptors;
+        field_descriptors.reserve(field_dict.size());
 
-        for (auto field : attr("fields").attr("items")()) {
-            auto spec = field.cast<tuple>();
-            auto name = spec[0].cast<pybind11::str>();
-            auto spec_fo = spec[1].cast<tuple>();
+        for (auto kv_item : field_dict) {
+            auto name = reinterpret_borrow<pybind11::str>(kv_item.first);
```
rwgk, line 652: IIUC this replaces `obj.cast<pybind11::str>()` — which does type checking — with `reinterpret_borrow<pybind11::str>` — which blindly assumes `kv_item.first` is a `str`. If that's true (?), and there is some upstream oversight/bug/API change, this could lead to a crash (relatively good) or a silent failure and time-consuming debugging.

Skylion007: @rwgk Okay, I'll revert that part of the change. the dict_iterator returns an iterator of a std::pair<handles> which doesn't have a clean way of handling the steal / borrow distinction. I think you can call cast<> to a pyobject directly on it, but I don't know how it handles that code path to be honest.

Skylion007: @rwgk I've reverted to the older dict api.

rwgk (approved):

PR #3997, opened 2022-06-08
chore: bump clang-tidy to 13

henryiii:
Signed-off-by: Henry Schreiner <henryschreineriii@gmail.com>

## Suggested changelog entry:

```rst

```


Skylion007 (approved):

.clang-tidy
```diff
@@ -73,5 +73,3 @@ CheckOptions:
   value:           true
 
 HeaderFilterRegex: 'pybind11/.*h'
-
-WarningsAsErrors: '*'
```
Skylion007, line 77: @henryiii Any reason this is removed? If there are explicit checks which are giving warnings, they can be negated as errors in this part of the list.

.github/workflows/format.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     - name: Configure
       run: >
         cmake -S . -B build
-        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy)"
+        -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--use-color;--warnings-as-errors=*"
```
henryiii, line 47: @Skylion007 Warnings as errors is set here. It's more inconvenient running this locally with warnings as errors, since it stops the compilation (requiring `--keep-going`, for example). Notice the "local run" text changes too. It's better to have this as a user-control via `CMAKE_CXX_CLANG_TIDY` vs. always forcing it. (IMO)

henryiii: Do you like it then, or would you rather it the old way?

Issue #1309, opened 2018-03-06
Implicit conversion of ints (and char const*) to handles?


anntzer:
## Issue description

Is there a reason why handles are not implicitly constructible from ints (I may well have missed something obvious...)?  e.g. something like
```
py::object func() {
    py::object tpl = py::make_tuple(1, 2);
    return tpl[1];
}
```
seems to have a clear meaning but fails to compile (`error: invalid conversion from ‘int’ to ‘const char*’` (there are two overloads for operator[], one taking a handle and the other a const char*)).

If char const* was also implicitly convertible to handle then the second overload would not be necessary either.

For additional confusion, 0 *is* implicitly convertible to a handle (an invalid one: the null pointer).

Asked on gitter (a more specific version of this) but got no reply.

## Reproducible example code

Include the above-mentioned code in python_example (replacing the `add` function there, and the `m.def(...` line).

jagerman:
I don't think there's any particularly good reason why it's not there--other than that no one has proposed/implemented it yet.  (As a side note, for tuples in particular, you're better off keeping the object as a `py::tuple`, which *does* work as you expect; it also bypasses the heavier `PyObject_GetItem` in favour of the lighter `PyTuple_GetItem`.  Likewise for `py::list`).

For the more generic `py::object`, I suppose we could accept any arbitrary argument type with a cast to Python.  Then strings, floats, custom types, etc. would all work as arguments to `obj[]`.  ISTM that this could replace *both* the `handle` and `char *key` versions, since casting either of those does the right thing (i.e. nothing for the former, and making a python string for the latter).

anntzer:
OK, I really thought that was such an obvious thing to do that I must have had missed a good reason :-)  For example, is there a way to ensure that the literal 0 becomes the integer 0 rather than a null pointer (as it does now)?
Tuple indexing is only an example, of course; there are plenty of other places where this could be used (e.g., `py::getattr(foo, "bar", 42 <- this argument)`).

jagerman:
Yup.  Meanwhile there are other places that already *do* accept arbitrary arguments--the arguments to `make_tuple`, for example.  So you're right that's it's not a stretch (and indeed expected) to get automatic Python casting.

cielavenir:
@jagerman @anntzer do you think https://github.com/woodychow/pybind11/commit/98c9f77e5481af4cbc7eb092e1866151461e3508 works?

/cc @woodychow

anntzer:
[...]

PR #4004, opened 2022-06-14
Support arbitrary type in item accessor's subscript operator

woodychow:
Resolves https://github.com/pybind/pybind11/issues/1309

PR made at [cielavenir](https://github.com/cielavenir)'s request.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -92,11 +92,8 @@ public:
         subclass causes a corresponding call to ``__getitem__``. Assigning a `handle`
         or `object` subclass causes a call to ``__setitem__``.
     \endrst */
-    item_accessor operator[](handle key) const;
-    /// See above (the only difference is that the key's reference is stolen)
-    item_accessor operator[](object &&key) const;
-    /// See above (the only difference is that the key is provided as a string literal)
-    item_accessor operator[](const char *key) const;
+    template <typename T>
+    item_accessor operator[](T &&key) const;
 
     /** \rst
         Return an internal functor to access the object's attributes. Casting the
@@ -2270,16 +2267,9 @@ iterator object_api<D>::end() const {
     return iterator::sentinel();
 }
 template <typename D>
-item_accessor object_api<D>::operator[](handle key) const {
-    return {derived(), reinterpret_borrow<object>(key)};
-}
-template <typename D>
-item_accessor object_api<D>::operator[](object &&key) const {
-    return {derived(), std::move(key)};
-}
-template <typename D>
-item_accessor object_api<D>::operator[](const char *key) const {
-    return {derived(), pybind11::str(key)};
+template <typename T>
+item_accessor object_api<D>::operator[](T &&key) const {
+    return {derived(), reinterpret_borrow<object>(detail::object_or_cast(std::forward<T>(key)))};
 }
 template <typename D>
 obj_attr_accessor object_api<D>::attr(handle key) const {
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -126,6 +126,10 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("tuple_ssize_t", []() { return py::tuple{(py::ssize_t) 0}; });
     m.def("tuple_size_t", []() { return py::tuple{(py::size_t) 0}; });
     m.def("get_tuple", []() { return py::make_tuple(42, py::none(), "spam"); });
+    m.def("access_tuple_with_int_index", []() {
+        py::object tpl = py::make_tuple(1, 2);
+        return tpl[1];
+    });
 
     // test_simple_namespace
     m.def("get_simple_namespace", []() {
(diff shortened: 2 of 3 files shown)


include/pybind11/pytypes.h
```diff
@@ -2270,16 +2267,9 @@ iterator object_api<D>::end() const {
...
-}
-template <typename D>
-item_accessor object_api<D>::operator[](object &&key) const {
-    return {derived(), std::move(key)};
-}
-template <typename D>
-item_accessor object_api<D>::operator[](const char *key) const {
-    return {derived(), pybind11::str(key)};
+template <typename T>
+item_accessor object_api<D>::operator[](T &&key) const {
```
Skylion007, line 2271: This actually will break quite a few things and cause some perf issues with a needless temporary. I don't quite get the motivation here either. What is wrong with the current overloads?

Skylion007: Additionally, this PR should include tests to demonstrate the new behavior.

include/pybind11/pytypes.h
```diff
@@ -2270,16 +2267,9 @@ iterator object_api<D>::end() const {
...
+    return {derived(), reinterpret_borrow<object>(detail::object_or_cast(std::forward<T>(key)))};
```
Skylion007, line 2272:
```suggestion
    return {derived(), detail::object_or_cast(std::forward<T>(key)).cast<object>()};
```
albeit not the most readable at least it won't give as much perf issues

tests/test_pytypes.cpp
```diff
@@ -126,6 +126,10 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("tuple_ssize_t", []() { return py::tuple{(py::ssize_t) 0}; });
     m.def("tuple_size_t", []() { return py::tuple{(py::size_t) 0}; });
     m.def("get_tuple", []() { return py::make_tuple(42, py::none(), "spam"); });
+    m.def("access_tuple_with_int_index", []() {
+        py::object tpl = py::make_tuple(1, 2);
+        return tpl[1];
```
Skylion007, line 131: int is implicitly converted to size_t first, I believe so it uses the tuple_accessor, not the item_accessor.

Skylion007:
@woodychow Seems this change breaks pickling in pypy for some reason.

Skylion007:
@woodychow If the remaining tests pass, I'll consider merging.

include/pybind11/pytypes.h
```diff
@@ -2270,16 +2267,9 @@ iterator object_api<D>::end() const {
...
-template <typename D>
-item_accessor object_api<D>::operator[](object &&key) const {
-    return {derived(), std::move(key)};
-}
-template <typename D>
-item_accessor object_api<D>::operator[](const char *key) const {
-    return {derived(), pybind11::str(key)};
+template <typename T>
+item_accessor object_api<D>::operator[](T &&key) const {
+    return {derived(), detail::object_or_cast(std::forward<T>(key)).template cast<object>()};
```
Skylion007, line 2272: I dug around and found it that going from handle to object this way automatically does reinterpret_borrow so it's just some syntactic sugar in that regard.

tests/test_pytypes.cpp
Skylion007: Okay so it looks this test case itself is bugged and makes pypy complain.

Skylion007: @woodychow Would you please add a dict case that actually tests the broken behavior? Like the dict example given in the issue?

cielavenir: @Skylion007 for this particular case py::tuple could be used but what about multi-dimensional tuple/list?

cielavenir: this test would be more proper here

```
    m.def("access_tuple_with_int_index", []() {
        py::object tpl = py::make_tuple(py::make_tuple(1, 2), py::make_tuple(3, 4));
        return tpl[0][1];
    });
```

this `tpl[0][1]` cannot be compiled on stock pybind11 even if tpl is declared as py::tuple.

include/pybind11/pytypes.h
cielavenir: interestingly, https://github.com/cielavenir/pybind11/commit/c59a10d1db42ffffc7cd44ead7b5734722ae62ca works locally on both pypy2 and pypy3.

tests/test_pytypes.cpp
cielavenir: actually we need both single dimension case and multidimension cases...

cielavenir: dict test https://github.com/cielavenir/pybind11/commit/fd82f0ab118f9e42d53abf9d6ba3cca6bb089670

somehow this one works fine, why tuple is broken like this...

cielavenir:
- trying to add tests as https://github.com/pybind/pybind11/pull/4007
- seems we cannot access py::tuple as object (i.e. PyObject_GetItem) for new ones, unlike PyTuple_GetItem. So I fixed the test accordingly.

include/pybind11/pytypes.h
cielavenir: @Skylion007 sorry to bother you but is this "syntactic sugar" of @woodychow 's original code or does your code have improvement?

PR #4006, opened 2022-06-14
maint: disable implicit conversion of 0 to handle

Skylion007:
## Description
* Disable implicit conversion of int to pyobject* to handle.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -214,6 +214,7 @@ public:
     /// Creates a ``handle`` from the given raw Python object pointer
     // NOLINTNEXTLINE(google-explicit-constructor)
     handle(PyObject *ptr) : m_ptr(ptr) {} // Allow implicit conversion from PyObject*
+    handle(int) = delete;
 
     /// Return the underlying ``PyObject *`` pointer
     PyObject *ptr() const { return m_ptr; }


include/pybind11/pytypes.h
```diff
@@ -214,6 +214,7 @@ class handle : public detail::object_api<handle> {
     /// Creates a ``handle`` from the given raw Python object pointer
     // NOLINTNEXTLINE(google-explicit-constructor)
     handle(PyObject *ptr) : m_ptr(ptr) {} // Allow implicit conversion from PyObject*
```
rwgk, line 216: Maybe this is a job for `enable_if`?
I'd bet though `enable_if` `PyObject*` won't work, `is_pointer` or similar might. Or maybe to be fancy, look for `ob_refcnt` in the pointee?

Skylion007: @rwgk Hmm... can't figure out the way to construct the enable_if_t.

Skylion007: This strategy doesn't work because it causes ambiguous overloads, which is ridiculous since the func is deleted.

rwgk:
It's a can of worms indeed, but I got pretty far:

```
rwgk.c.googlers.com:~/forked/pybind11 $ git diff
index f9625e77..370b3f9a 100644
@@ -213,7 +213,13 @@ public:
     handle() = default;
     /// Creates a ``handle`` from the given raw Python object pointer
-    handle(PyObject *ptr) : m_ptr(ptr) {} // Allow implicit conversion from PyObject*
+    template <typename T, detail::enable_if_t<std::is_pointer<T>::value || std::is_same<T, std::nullptr_t>::value, int> = 0>
+    handle(T ptr) : m_ptr(ptr) {} // Allow implicit conversion from PyObject*
+
+    handle(const handle &) = default;
+    handle(handle &&) = default;
+    handle& operator=(const handle &) = default;
+    handle& operator=(handle &&) = default;

     /// Return the underlying ``PyObject *`` pointer
     PyObject *ptr() const { return m_ptr; }
diff --git a/tests/test_numpy_array.cpp b/tests/test_numpy_array.cpp
index 69ddbe1e..f8a777e4 100644
--- a/tests/test_numpy_array.cpp
+++ b/tests/test_numpy_array.cpp
@@ -218,7 +218,7 @@ TEST_SUBMODULE(numpy_array, sm) {
     // test_empty_shaped_array
     sm.def("make_empty_shaped_array", [] { return py::array(py::dtype("f"), {}, {}); });
     // test numpy scalars (empty shape, ndim==0)
-    sm.def("scalar_int", []() { return py::array(py::dtype("i"), {}, {}, &data_i); });
+    sm.def("scalar_int", []() { return py::array(py::dtype("i"), {}, {}/*, &data_i*/); });

     // test_wrap
     sm.def("wrap", [](const py::array &a) {
```

There is only one subtest that does not pass (test_make_empty_shaped_array) with this diff. Is there a chance you could help debugging that?

Skylion007:
Closing in favor of #4008

PR #4007, opened 2022-06-15
Draft: Support arbitrary type in item accessor's subscript operator [with more tests]

cielavenir:
(to run ci)

https://github.com/pybind/pybind11/pull/4004


include/pybind11/pytypes.h
```diff
@@ -2270,16 +2267,9 @@ iterator object_api<D>::end() const {
...
-template <typename D>
-item_accessor object_api<D>::operator[](object &&key) const {
-    return {derived(), std::move(key)};
-}
-template <typename D>
-item_accessor object_api<D>::operator[](const char *key) const {
-    return {derived(), pybind11::str(key)};
+template <typename T>
+item_accessor object_api<D>::operator[](T &&key) const {
+    return {derived(), reinterpret_borrow<object>(detail::object_or_cast(std::forward<T>(key)))};
```
Skylion007, line 2272: You also need to use the cast trick used in the other PR or this will fail when run in debug mode.

cielavenir: rebased based on https://github.com/pybind/pybind11/pull/4004#discussion_r899256872 comment, thank you

tests/test_pytypes.cpp
```diff
@@ -71,6 +71,20 @@ TEST_SUBMODULE(pytypes, m) {
...
+        py::list l1 = py::list();
+        l1.append(1);
+        l1.append(2);
+        return l1[1];
+    });
+    m.def("access_list_as_object", []() {
+        py::list l1 = py::list();
+        l1.append(1);
+        l1.append(2);
+        py::object l2 = l1;
```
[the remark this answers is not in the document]
Skylion007: Actually super odd that only Python built in debug mode causes the segfault. Thoughts @rwgk @henryiii ?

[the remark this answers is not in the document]
rwgk: No, but I looked a little bit only.
Sorry, I don't have the free bandwidth to help debugging this.

tests/test_pytypes.cpp
```diff
@@ -120,13 +134,45 @@ TEST_SUBMODULE(pytypes, m) {
...
+    m.def("access_dict_as_object_with_str", []() {
+        py::dict d1 = py::dict();
+        d1["x"] = 1;
+        py::object d2 = d1;
+        return d2["x"];
+    });
+    m.def("access_dict_as_object_with_int", []() {
+        py::dict d1 = py::dict();
+        d1[1] = 1;
+        py::object d2 = d1;
```
Skylion007, line 156:
```suggestion
        py::object d2 = std::move(d1);
```

Skylion007 (commented):
Nits: should make clang-tidy happy and are better style more general (allows reference stealing).

tests/test_pytypes.cpp
```diff
@@ -120,13 +134,45 @@ TEST_SUBMODULE(pytypes, m) {
...
+    });
+    m.def("access_dict_with_int", []() {
+        py::dict d1 = py::dict();
+        d1[1] = 1;
+        return d1[1];
+    });
+    m.def("access_dict_as_object_with_str", []() {
+        py::dict d1 = py::dict();
+        d1["x"] = 1;
+        py::object d2 = d1;
```
Skylion007, line 150:
```suggestion
        py::object d2 = std::move(d1);
```

cielavenir: thank you

cielavenir:
by the way, L62 is `access_list`, which is this:

```
    m.def("access_list", []() {
        py::list l1 = py::list();
        l1.append(1);
        l1.append(2);
        return l1[1];
    });
```

why this one segfaults?

cielavenir:
@rwgk @Skylion007 do you have ideas about segfault in debug build?

Skylion007:
> @rwgk @Skylion007 do you have ideas about segfault in debug build?

Yeah, unfortunately and I am not sure it's an easy fix (at least if it's really the method after (access_list_as_object) that is segfaulting. 

It's because depending on the type of the object, the accessor either borrows or steals the reference to the underlying object. What's happening here is that the list is deleted which is the only remaining reference to the object since it's not properly borrowed by the item accessor. The GC then frees the item since the ref_count is 0. However, we still have a pointer to that object (returned from the function), but it's now pointing to free memory. Hence, the segfault.

Not sure these is an easy solution here as the type of the accessor that needs to be called depends on the object. You may be able to get away with this through virtual inheritance / overloading, but that itself seems bugprone and currently our class inheritance method is non-virtual and pybind11 even supports compilation without RTTI (which this solution may break). 

We could also try to return strong reference for the accessors by seeing if we can use different CPython APIs, but this may both have performance implications and could introduce ref_count based memory leaks if we are not careful.

Not sure what a good solution is here other than ensure that object is properly downcasted or that the argument is a proper Python object. We already pushed a PR that disable implicit conversions of int to handles so that should fix some of the issues that this PR brought up.

PR #4017, opened 2022-06-18
chore: use explicit defaulting in pyobject macros

Skylion007:
## Description
* This change makes the pyobject macro uses the equals-default instead of defining an empty ctor.  The behavior should be identical to before, but the ctor is now trivial, enabling more compiler optimizations.

## Suggested changelog entry:

```rst
* Explicitly default all relevant ctors for pytypes in the PYBIND11_OBJECT macros and enforce the clang-tidy checks modernize-use-equals-default in macros as well.
```


rwgk (approved):

Skylion007:
Sigh, this is now only a bug with the NVIDIA compilers.

rwgk:
> Sigh, this is now only a bug with the NVIDIA compilers.

I'm a bit confused TBH, IIRC there were more failures before, but I cannot figure out what changed.

Before and now, I was/am only seeing this trivial issue:
```
"/__w/pybind11/pybind11/include/pybind11/numpy.h", line 540: error: extra ";" ignored
      PYBIND11_OBJECT_DEFAULT(dtype, object, detail::npy_api::get().PyArrayDescr_Check_);
```

In theory this PR could break user projects with extra ";" & -Werror, but Google-internally I just did a (very quick) global code search for `PYBIND11_OBJECT_DEFAULT` and `PYBIND11_OBJECT_CVT_DEFAULT`: the only matches are in the pybind11 sources.

Therefore I believe this PR is safe, we just need to remove the extra ";" and add a small warning in the changelog.

PR #4020, opened 2022-06-20
ci: use almalinux instead of centos, add 9

henryiii:
## Description

AlmaLinux is used by manylinux, so let's use that here.

## Suggested changelog entry:

None needed, CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -584,18 +584,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        centos:
-          - centos7  # GCC 4.8
-          - stream8
+        container:
+          - "centos:7"  # GCC 4.8
+          - "almalinux:8"
+          - "almalinux:9"
 
-    name: "🐍 3 • CentOS ${{ matrix.centos }} • x64"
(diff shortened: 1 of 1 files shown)


rwgk (approved):
Do you think it would be useful to add this to the changelog?

> Testing with Centos 8 was replaced with AlmaLinux 8. Testing with AlmaLinux 9 was added.

I think that would be useful to highlight in the changelog because from a practical viewpoint that means we are dropping support for Centos 8, but are adding support for AlmaLinux 8 & 9.

henryiii:
Not really, we support RHEL 7, 8, and 9. We just use CentOS's repackaging for 7 and AlmaLinux repackaging for 8 & 9. We could use RockyLinux and it would be the same. Or, since I'm at Princeton, maybe Springdale. ;) For an end user, RHEL 7-9 is the important thing, not how it's been debranded and rebuilt.

henryiii:
I also believe we don't advertise the list of OS's we build on, just compiler versions. Could be wrong though.

We are avoiding stream distros - I think we were supporting 8 stream, which we aren't now. But we are now testing on 9 anyway, and we've never supported/tested Fedora or even non LTS Ubuntu either.

PR #4031, opened 2022-06-28
Support loading `unique_ptr<Derived>` as `unique_ptr<Base>`.

wangxf123456:
## Description
This is to support the following use case. C++ code:
```
class Base {
 public:
  Base() = default;
  virtual int get() { return 10; }
  virtual ~Base() { }
};

class Derived: public Base {
 public:
  static std::unique_ptr<Derived> create() {
    return std::make_unique<Derived>();
  }
  int get() override { return 100; }
  ~Derived() override { }
};

int consume_base(std::unique_ptr<Base> base) {
  return base->get();
}
```

The following should be allowed, but without the change, it raises exception `Incompatible unique_ptr deleter (as_unique_ptr).`:
```
derived = Derived.create();
unique_ptr_deleter.consume_base(derived)
```

### DIFF
--- a/tests/pure_cpp/smart_holder_poc_test.cpp
+++ b/tests/pure_cpp/smart_holder_poc_test.cpp
@@ -39,12 +39,12 @@ private:
 
 struct base {
     virtual int get() { return 10; }
-    virtual ~base() { }
+    virtual ~base() {}
 };
 
-struct derived: public base {
+struct derived : public base {
     int get() override { return 100; }
-    ~derived() override { }
+    ~derived() override {}
 };
 
 } // namespace helpers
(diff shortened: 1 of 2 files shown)


tests/pure_cpp/smart_holder_poc_test.cpp
```diff
@@ -37,6 +37,16 @@ struct indestructible_int {
...
 };
 
+struct base {
+    virtual int get() { return 10; }
+    virtual ~base() {}
+};
+
+struct derived : public base {
+    int get() override { return 100; }
+    ~derived() override {}
```
rwgk: Could we delete this override?
(It seems unusual to override the virtual dtor in this situation.)

tests/pure_cpp/smart_holder_poc_test.cpp
```diff
@@ -227,6 +237,25 @@ TEST_CASE("from_unique_ptr+as_shared_ptr", "[S]") {
     REQUIRE(*new_owner == 19);
 }
 
+TEST_CASE("from_unique_ptr_derived+as_unique_ptr_base", "[E]") {
```
rwgk, line 240: Could you please make this `"[S]"` here?

`[E]` is for from/to combinations that generate an error, `[S]` for valid combinations.

rwgk (approved):

PR #4033, opened 2022-06-28
Expose exception cause as std::nested_exception

ecatmur:
## Description

Expose Python exception cause (`raise ... from ...`) to C++ via `std::nested_exception`. Inverse of #3608.

Useful after #3605 etc.

Does not set std::nested_exception base if a C++ exception is already present, for backwards compatibility if anyone is already using `std::throw_with_nested(py::error_already_set())`.

## Suggested changelog entry:

```rst
`error_already_set` now inherits from `std::nested_exception` to expose the Python exception cause (`raise ... from ...`).
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -546,6 +546,8 @@ struct error_fetch_and_normalize {
         return (PyErr_GivenExceptionMatches(m_type.ptr(), exc.ptr()) != 0);
     }
 
+    std::nested_exception get_cause_as_nested() const;
+
     // Not protecting these for simplicity.
     object m_type, m_value, m_trace;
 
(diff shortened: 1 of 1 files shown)


include/pybind11/pytypes.h
```diff
@@ -2327,6 +2333,20 @@ bool object_api<D>::rich_compare(object_api const &other, int value) const {
     return rv == 1;
 }
 
+inline std::nested_exception error_fetch_and_normalize::get_cause_as_nested() const {
+    auto cause = reinterpret_steal<object>(PyException_GetCause(m_value.ptr()));
+    if (not cause or cause.is_none())
+        return std::nested_exception();
+    auto typ = type::of(cause);
+    auto tb = reinterpret_steal<object>(PyException_GetTraceback(cause.ptr()));
```
Skylion007, line 2341: PyException_GetTraceback can return null. Will PyErr_Restore work if this is the case?

include/pybind11/pytypes.h
```diff
@@ -2327,6 +2333,20 @@ bool object_api<D>::rich_compare(object_api const &other, int value) const {
...
+    PyErr_Restore(typ.release().ptr(), cause.release().ptr(), tb.release().ptr());
+    try {
+        throw error_already_set();
```
Skylion007, line 2344: Throwing is very slow and something that is best avoided here. is there a way we avoid having to throw the exception just to catch and return an empty std::nested_exception?

PR #4034, opened 2022-06-28
[docs] Fix "Enumerations and internal types" example

crcrpar:
## Description
This is a patch for the example mentioned in the title to be built because when building the snippet, I got the build error which includes:

```console
    src/main.cpp:71:60: error: expected ';' at end of declaration
      py::class_<Pet::Attributes> attributes(pet, "Attributes")
                                                               ^
                                                               ;
    1 error generated.

```

Could you kindly let me know if there's a better change?

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -482,7 +482,8 @@ The binding code for this example looks as follows:
         .value("Cat", Pet::Kind::Cat)
         .export_values();
 
-    py::class_<Pet::Attributes> attributes(pet, "Attributes")
+    py::class_<Pet::Attributes> attributes(pet, "Attributes");
+    attributes
         .def(py::init<>())
         .def_readwrite("age", &Pet::Attributes::age);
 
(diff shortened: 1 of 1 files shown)


docs/classes.rst
```diff
@@ -482,7 +482,8 @@ The binding code for this example looks as follows:
         .value("Cat", Pet::Kind::Cat)
         .export_values();
 
-    py::class_<Pet::Attributes> attributes(pet, "Attributes")
+    py::class_<Pet::Attributes> attributes(pet, "Attributes");
```
rwgk, line 485: Thanks for reporting! This is clearly a bug in the docs, but I think it would be better to delete ` attributes` because it's not used:
```
py::class_<Pet::Attributes>(pet, "Attributes")
```

Could you please update this PR to make that change?

rwgk:
Thanks! Looks good. I'm just waiting for the CI to finish. (I cannot imagine what could to wrong, but it's pretty quick.)

PR #4043, opened 2022-07-05
`PYBIND11_NAMESPACE` consistency fixes.

rwgk:
## Description
This PR fixes a couple long-standing `PYBIND11_NAMESPACE` `__attribute__((visibility("hidden")))` inconsistencies. The fixes only affect builds with `__GNUG__` defined.

Based on experiments under PR #4044, this (currently) only makes a difference in an unusual situation: macOS, NOT using `-fvisibility=hidden`. However, if someone ventures there, surprises due to inconsistencies can be extremely time-consuming to troubleshoot.

(Discovered while experimenting related to PR #4022.)

## Suggested changelog entry:

```rst
A couple long-standing `PYBIND11_NAMESPACE` `__attribute__((visibility("hidden")))` inconsistencies are now fixed (affects only unusual environments).
```

PR #4044, opened 2022-07-06
Add `test_namespace_visibility` (experiment, not for merging)

rwgk:
## Description
To answer the question: Is there a difference in behavior if some `__attribute__((visibility("hidden")))` are missing (PR #4043)?

A second motivation for this experiment was to gain clarity about Google-internal vs external behavior.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -846,7 +846,7 @@ struct always_construct_holder {
 
 /// Create a specialization for custom holder types (silently ignores std::shared_ptr)
 #define PYBIND11_DECLARE_HOLDER_TYPE(type, holder_type, ...)                                      \
-    namespace pybind11 {                                                                          \
+    PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)                                                  \
     namespace detail {                                                                            \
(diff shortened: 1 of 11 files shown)


rwgk:
To explain:
```
      6 UNEXPECTED code_line: AAA:AAc:AaA:Aac:aAC:aAc:aaC:aac
     30 UNEXPECTED code_line: AAA:AAc:AaC:Aac:aAC:aAc:aaC:aac
----------------------------------------^
```
The dynamic loaders behave differently on macOS (1st line above) vs Linux (2nd line).

If a namespace has `hidden` visibility in one Translation Unit (TU), but `default` in another TU linked into the same `PYBIND11_MODULE`, the net result is that it is

* macOS: visible  from another `PYBIND11_MODULE`.
* Linux: `hidden` from another `PYBIND11_MODULE`.

Note that this is withOUT `-fvisibility=hidden` and `RTLD_GLOBAL` on Linux, i.e. for an unusual situation.

rwgk:
The code added here was moved to separate files (tests/*namespace_visibility*) under PR #4050.

PR #4051, opened 2022-07-08
Properly visit self in >=3.9 traverse

Skylion007:
## Description
* This PR fixes a small bug that was already fixed in nanobind. Specifically, the our traverse behavior does not follow the [clarified behavior of the tp-traverse function Python >=3.9](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse). 
* I originally discovered this bug when researching the 3.11 fixes, but this did not fix the root cause there. Still, it's better to have proper behavior for >= 3.9 so this PR cherry picks that fix and ensure we properly traverse the classes when calling gc on dynamically allocated classes.

## Suggested changelog entry:

```rst
* Ensure proper behavior when garbage collecting classes with dynamic attributes in Python >=3.9.
```

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -531,6 +531,7 @@ extern "C" inline int pybind11_set_dict(PyObject *self, PyObject *new_dict, void
 extern "C" inline int pybind11_traverse(PyObject *self, visitproc visit, void *arg) {
     PyObject *&dict = *_PyObject_GetDictPtr(self);
     Py_VISIT(dict);
+// https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse
 #if PY_VERSION_HEX >= 0x03090000
     Py_VISIT(Py_TYPE(self));
 #endif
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/class.h
```diff
@@ -531,6 +531,9 @@ extern "C" inline int pybind11_set_dict(PyObject *self, PyObject *new_dict, void
 extern "C" inline int pybind11_traverse(PyObject *self, visitproc visit, void *arg) {
     PyObject *&dict = *_PyObject_GetDictPtr(self);
     Py_VISIT(dict);
+#if PY_VERSION_HEX >= 0x03090000
+    Py_VISIT(Py_TYPE(self));
```
rwgk, line 535: Optional: I'd add this comment (your link from the PR description) so the next person to look here has an easier time understanding this code:
```
// https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse
```

rwgk (approved):

henryiii (approved):

rwgk:
Thanks @Skylion007, I'll go ahead merging this PR, to get it into the next smart_holder update.

PR #4064, opened 2022-07-13
chore: optimize sparse matrix casting with python tuple

Skylion007:
## Description
* Optimizes the Eigen sparse matrix construction by using a py::tuple instead of requiring a cast from std::tuple.

## Suggested changelog entry:

```rst
* optimizes Eigen sparse matrix casting by removing unnecessary temporary.
```


rwgk (approved):
It would be nice to have some kind of estimate how much this helps (to inform where it is worth it in general to optimize), but this is obviously a good change.

Skylion007:
Probably not worth much perf to be honest, but I was staring at the code and noticed it was obviously wrong / bad style. Will just remove some unnecessary instructions.

PR #4065, opened 2022-07-14
Fix #3812 and fix const of inplace assignments

### LINKED ISSUE #3812: [BUG]: py::object augmented assignment doesn't work with immutable types
### Required prerequisites

### Problem description

The augmented assignment operators (`operator+=`, etc) on `py::object` are implemented analogously to the regular binary operators (`operator+`, etc): they call one of the `PyNumber_` API functions and return the result. For immutable types such as `py::str`, `PyNumber_InPlaceAdd` is implemented the same as `PyNumber_Add`, so you need to look at the result in order to see any change; in both C++ and Python, though, `foo += bar` is supposed to be sufficient to update `foo` on its own without observing the result of the expression.

### Reproducible example code

```text

namespace py = pybind11;

PYBIND11_MODULE(example, m) {
    m.def("test", []() -> py::str {
        py::str text = "foo"_s;
        text += " bar"_s;
        return text;
    });
}

// import example
// assert example.test() == "foo bar"
//    --> fails: example.test() actually returns "foo"
```

Skylion007:
## Description
So I looked through the inplace assignments and found that many of them did not work for immutable types (int_, str, etc...). I actually couldn't find any situation where they appeared to work properly at all. Additionally, it didn't seem possible to actually properly +=, -= etc in the object_api without some virtual functionsince the = operator is overridden in the subclasses. The methods I suppose could remain const for handles since they don't modify the handle, but we need to make those const methods can't be accidentally used on the object class (perhaps through explicit deletion?).  

Fix #3812.

## Suggested changelog entry:

```rst
* Make augmented assignment operators non-const for the object-api
* Behavior was previously broken for augmented assignment operators.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -190,6 +190,11 @@ private:
     bool rich_compare(object_api const &other, int value) const;
 };
 
+template <typename T>
+using is_pyobj_ptr_or_nullptr_t = detail::any_of<std::is_same<T, PyObject *>,
+                                                 std::is_same<T, PyObject *const>,
+                                                 std::is_same<T, std::nullptr_t>>;
+
 PYBIND11_NAMESPACE_END(detail)
 
 #if !defined(PYBIND11_HANDLE_REF_DEBUG) && !defined(NDEBUG)
@@ -211,9 +216,23 @@ class handle : public detail::object_api<handle> {
 public:
     /// The default constructor creates a handle with a ``nullptr``-valued pointer
     handle() = default;
-    /// Creates a ``handle`` from the given raw Python object pointer
+
+    /// Enable implicit conversion from ``PyObject *`` and ``nullptr``.
+    /// Not using ``handle(PyObject *ptr)`` to avoid implicit conversion from ``0``.
+    template <typename T,
+              detail::enable_if_t<detail::is_pyobj_ptr_or_nullptr_t<T>::value, int> = 0>
     // NOLINTNEXTLINE(google-explicit-constructor)
-    handle(PyObject *ptr) : m_ptr(ptr) {} // Allow implicit conversion from PyObject*
+    handle(T ptr) : m_ptr(ptr) {}
+
+    /// Enable implicit conversion through ``T::operator PyObject *()``.
+    template <
+        typename T,
+        detail::enable_if_t<detail::all_of<detail::none_of<std::is_base_of<handle, T>,
+                                                           detail::is_pyobj_ptr_or_nullptr_t<T>>,
+                                           std::is_convertible<T, PyObject *>>::value,
+                            int> = 0>
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    handle(T &obj) : m_ptr(obj) {}
 
     /// Return the underlying ``PyObject *`` pointer
     PyObject *ptr() const { return m_ptr; }
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -334,12 +334,14 @@ public:
     }
 
     object &operator=(const object &other) {
-        other.inc_ref();
-        // Use temporary variable to ensure `*this` remains valid while
-        // `Py_XDECREF` executes, in case `*this` is accessible from Python.
-        handle temp(m_ptr);
-        m_ptr = other.m_ptr;
-        temp.dec_ref();
+        if (this != &other) {
+            other.inc_ref();
+            // Use temporary variable to ensure `*this` remains valid while
+            // `Py_XDECREF` executes, in case `*this` is accessible from Python.
+            handle temp(m_ptr);
+            m_ptr = other.m_ptr;
+            temp.dec_ref();
+        }
         return *this;
     }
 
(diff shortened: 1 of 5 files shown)


henryiii:
Ahh, this is the operator on pyobject, not `py::self`, which is what I immediately thought of (because I use it at https://github.com/scikit-hep/boost-histogram/blob/d438def268b8bfb9a51a5fed42ad88683de0e6f9/include/bh_python/register_histogram.hpp#L61 ). Seems like a reasonable fix, we could add const handle versions later if someone cares (guessing these weren't heavily used before?)

Skylion007:
@henryiii Not sure this would actually work because we need to do the same thing on accessors as well. I was thinking about whether it would be better to do a general mixin class like we do the object_api. Thoughts?

include/pybind11/pytypes.h
```diff
@@ -353,6 +353,20 @@ class object : public handle {
         return *this;
     }
 
+#define PYBIND11_INPLACE_OP(inp)                                                                  \
```
rwgk, line 356: Very minor technically, but I'd definitely make this `iop` (a lot more intuitive).

include/pybind11/pytypes.h
```diff
@@ -2364,26 +2378,35 @@ bool object_api<D>::rich_compare(object_api const &other, int value) const {
         return result;                                                                            \
     }
 
+#define PYBIND11_MATH_OPERATOR_BINARY_INPLACE(op, fn)                                             \
```
rwgk, line 2381: I'd change this to `iop`, too (to be consistent & intuitive).

tests/test_pytypes.cpp
```diff
@@ -756,4 +756,30 @@ TEST_SUBMODULE(pytypes, m) {
...
+        return a;
+    });
+    m.def("inplace_or", [](py::object &a, const py::object &b) {
+        a |= b;
+        return a;
+    });
+    m.def("inplace_and", [](py::object &a, const py::object &b) {
+        a &= b;
+        return a;
+    });
```
rwgk, line 784: Are `<<=` and `>>=` omitted intentionally?

rwgk (commented):
Very minor comments, the meat of this looks good to me, but can you please give me a chance to run this through global testing? (Results late tonight or tomorrow morning.)
At the moment I'm thinking the `const` versions are an accident that just happened to survive amazingly long, and that we don't need to bring them back.

include/pybind11/pytypes.h
Skylion007: I actually think this is insufficient for attributes. Basically, we need to define the augmented operators for every object_api have custom operator=.

rwgk: Did you count already how many we have?
I'm not sure about the big picture TBH (I may still be misunderstanding), but this PR seemed like a necessary good step in the right direction, just by itself. Does that make sense?

tests/test_pytypes.cpp
rwgk: I just pushed a commit adding test_inplace_lshift, test_inplace_rshift, assuming it was an oversight that they were missing.

rwgk:
I still don't have global testing results unfortunately, due to some internal outage.

rwgk:
Global testing passes for this PR (internal test ID OCL:461063257:BASE:461249872:1657923213067:e8a3caed).

Therefore I'm still thinking: no need to worry about `const` versions, at least not for the current scope of changes.

@Skylion007 Are there reasons for holding back this PR? (see my previous question)

Skylion007:
@rwgk Main reason is I am trying to figure if there is a better way to solve this. Using the augmented assigned operators on handles is still broken with this PR and likely broken for accessors as well.

Skylion007:
Okay, so I tested it and the augmented assignment operators for accessors are currently broken anyway due to another bug which results in a compile error (unless you are doing augmented assignments from another accessor). As such, it's not as big of a deal as I thought. I would like to actually remove the augmented assignment operators from handles if possible, but that requires further re-architecting.

Skylion007:
@henryiii @rwgk Is this build error a flake or a compiler bug? Seems rather specific. Might have to revert the self-assignment optimizations if its the former.

rwgk:
Looks like flakes I've seen before. I clicked the button to rerun only the failed job. Let's see what happens.

include/pybind11/pytypes.h
```diff
@@ -334,17 +334,19 @@ class object : public handle {
     }
 
     object &operator=(const object &other) {
-        other.inc_ref();
-        // Use temporary variable to ensure `*this` remains valid while
-        // `Py_XDECREF` executes, in case `*this` is accessible from Python.
-        handle temp(m_ptr);
-        m_ptr = other.m_ptr;
-        temp.dec_ref();
+        if (!this->is(other)) {
```
rwgk, line 337: Is this purely an optimization that came to your attention because of the current focus on in-place operators, or is this `if` critical in some way?
A terse comment might be useful for readers, e.g.
```
if (!this->is(other)) { // Optimization for in-place operators.
```
if that's true, or something else if not.

rwgk (approved):
It was a flake, the rerun worked.

> I would like to actually remove the augmented assignment operators from handles if possible, but that requires further re-architecting.

Sounds good (I assume you mean "in a separate PR").

include/pybind11/pytypes.h
Skylion007: It came to my attention due to inplace-assigment, but it's a more general optimization.

PR #4072, opened 2022-07-16
Remove cmake `PROPERTIES CXX_VISIBILITY_PRESET "hidden"`

rwgk:
## Description
* Remove cmake `PROPERTIES CXX_VISIBILITY_PRESET "hidden"`
* Add `PYBIND11_NS_VISIBILITY` in pybind11/detail/common.h
* Use to resolve all "... declared with greater visibility than ..." warnings across all platforms.

TODO:
* Decide if it is actually desirable to remove `PROPERTIES CXX_VISIBILITY_PRESET "hidden"` from the cmake files.
* If not, revert the cmake changes but keep the test changes (so that the tests are compatible with default visibility).
* Clean up the documentation including FAQ.

Background: Issue #2479

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -25,8 +25,10 @@
 // on the main `pybind11` namespace.
 #if !defined(PYBIND11_NAMESPACE)
 #    ifdef __GNUG__
-#        define PYBIND11_NAMESPACE pybind11 __attribute__((visibility("hidden")))
+#        define PYBIND11_NS_VISIBILITY(...) __VA_ARGS__ __attribute__((visibility("hidden")))
+#        define PYBIND11_NAMESPACE PYBIND11_NS_VISIBILITY(pybind11)
 #    else
+#        define PYBIND11_NS_VISIBILITY(...) __VA_ARGS__
 #        define PYBIND11_NAMESPACE pybind11
 #    endif
 #endif
--- a/tests/test_custom_type_setup.cpp
+++ b/tests/test_custom_type_setup.cpp
@@ -13,14 +13,19 @@
 
 namespace py = pybind11;
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
 namespace {
 
 struct OwnsPythonObjects {
     py::object value = py::none();
 };
+
 } // namespace
+PYBIND11_NAMESPACE_END(pybind11_tests)
 
 TEST_SUBMODULE(custom_type_setup, m) {
+    using namespace pybind11_tests;
+
     py::class_<OwnsPythonObjects> cls(
         m, "OwnsPythonObjects", py::custom_type_setup([](PyHeapTypeObject *heap_type) {
             auto *type = &heap_type->ht_type;
--- a/tests/test_exceptions.cpp
+++ b/tests/test_exceptions.cpp
@@ -15,6 +15,8 @@
 #include <stdexcept>
 #include <utility>
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
+
 // A type that should be raised as an exception in Python
 class MyException : public std::exception {
 public:
@@ -110,7 +112,11 @@ std::string error_already_set_what(const py::object &exc_type, const py::object 
     return py::error_already_set().what();
 }
 
+PYBIND11_NAMESPACE_END(pybind11_tests)
+
 TEST_SUBMODULE(exceptions, m) {
+    using namespace pybind11_tests;
+
     m.def("throw_std_exception",
           []() { throw std::runtime_error("This exception was intentionally thrown."); });
 
--- a/tests/test_exceptions.h
+++ b/tests/test_exceptions.h
@@ -5,9 +5,13 @@
 
 // shared exceptions for cross_module_tests
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
+
 class PYBIND11_EXPORT_EXCEPTION shared_exception : public pybind11::builtin_exception {
 public:
     using builtin_exception::builtin_exception;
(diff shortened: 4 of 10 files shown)


rwgk:
@henryiii
@eacousineau

This PR proves that we do not need `PROPERTIES CXX_VISIBILITY_PRESET "hidden"` if we do not want to. Note that nothing is changed in the core library, except a tiny factoring to break out `PYBIND11_NS_VISIBILITY(...)` from `PYBIND11_NAMESPACE`. All other non-cmake/setup_helpers.py changes are to the tests only.

What is your opinion regarding this question (currently a TODO in the PR description):

* Decide if it is actually desirable to remove `PROPERTIES CXX_VISIBILITY_PRESET "hidden"` from the cmake files.

From my perspective, pushing "hidden" as strongly as it is done currently in the documentation and cmake files is more on the harmful than helpful side. Google-internally, after significant back & forth and confusion, we removed `-fvisibility=hidden` from our BUILD files, for reasons very similar to what @eacousineau reported under #2479.

EricCousineau-TRI:
I'm down for this!
Any chance you were able to check binary size produced before/after this PR for one platform+Python version+impl?
I would expect it might increase, but ideally it only increases due to other non-pybind11 symbols being (correctly) exposed.

include/pybind11/detail/common.h
```diff
@@ -25,8 +25,10 @@
 // on the main `pybind11` namespace.
 #if !defined(PYBIND11_NAMESPACE)
 #    ifdef __GNUG__
```
EricCousineau-TRI, line 27: nit `VISIBILITY` sounds like what is being changed, but this doesn't provide an option to choose.

Consider using `HIDDEN` to make intent more clear.

include/pybind11/detail/common.h
```diff
@@ -25,8 +25,10 @@
...
-#        define PYBIND11_NAMESPACE pybind11 __attribute__((visibility("hidden")))
+#        define PYBIND11_NS_VISIBILITY(...) __VA_ARGS__ __attribute__((visibility("hidden")))
```
EricCousineau-TRI, line 28: nit While `NS` is a good abbreviation, perhaps it is still good to full expand `NAMESPACE`?

Or if this is agnostic to the type of symbol, consider just removing the `NS` qualifier?

tests/test_custom_type_setup.cpp
```diff
@@ -13,14 +13,19 @@
 
 namespace py = pybind11;
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
```
EricCousineau-TRI, line 16: Not sure if I understand need for applying `hidden` when it only contains an anonymous namespace.

Is it possible to revert this (possibly redundant) change?

tests/test_exceptions.cpp
```diff
@@ -15,6 +15,8 @@
 #include <stdexcept>
 #include <utility>
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
```
EricCousineau-TRI, line 18: Same as above - why not just use anonymous `namespace {}`?

tests/test_exceptions.h
```diff
@@ -5,9 +5,13 @@
 
 // shared exceptions for cross_module_tests
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
```
EricCousineau-TRI, line 8: Same as above - why not just use anonymous `namespace {}`?

tests/test_numpy_array.cpp
```diff
@@ -15,6 +15,8 @@
 #include <cstdint>
 #include <utility>
 
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(pybind11_tests))
```
EricCousineau-TRI, line 18: Same as above - why not just use anonymous `namespace {}`?

tests/test_pytypes.cpp
```diff
@@ -11,7 +11,7 @@
 
 #include <utility>
 
-namespace external {
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NS_VISIBILITY(external))
```
EricCousineau-TRI, line 14: nit Er, hiding a namespace `external` smells odd.

Can we use a better name here? e.g. `feaux_external`, with a comment of what is going on?

tools/pybind11NewTools.cmake
```diff
@@ -203,19 +203,6 @@ function(pybind11_add_module target_name)
     target_link_libraries(${target_name} PRIVATE pybind11::windows_extras)
   endif()
 
-  # -fvisibility=hidden is required to allow multiple modules compiled against
```
EricCousineau-TRI, line 206: BTW yay! deleting code makes me 10000% happier!

EricCousineau-TRI (changes requested):
First pass done, PTAL!

tests/test_custom_type_setup.cpp
rwgk: Thanks Eric, I'll wait for Henry's opinion before investing more time, but to explain quickly, this is purely to appease some sad compilers that don't "realize" that the unnamed namespace is already hidden. I could add comments, or remove the unnamed namespace, I'll figure out something after we decided the big question.

henryiii:
Let's wait till after 3.10.1 to decide on this. I know the original design was to do the right thing for most users, and most users are not using the ABI compat, and are not that familiar with symbol visibility. So suddenly asking everyone to start hiding their symbols or the binaries expand is not a very great thing to do. There's also a very easy workaround - a user can simply unset it, which is making the more advanced use case take more configuration, but that's nicer.

I think the corrections on the visibility are good, and we can see how much this tends to affect user code. We could also update the docs recommending users set visibility for a release before we actually pull the plug on the default.

PR #4077, opened 2022-07-17
add installation support for pkg-config dependency detection

### LINKED ISSUE #230: Add pkg-config support
Adding pybind11.pc (https://www.freedesktop.org/wiki/Software/pkg-config/) to setup.py would greatly simplify the task of finding pybind11 via CMake.

eli-schwartz:
pkg-config is a build-system agnostic alternative to `pybind11Config.cmake` that can be used from build systems other than cmake.

Fixes #230

## Suggested changelog entry:

```rst
* Include a pkg-config file in the Python package.
```

### DIFF
--- a/tools/JoinPaths.cmake
+++ b/tools/JoinPaths.cmake
@@ -1,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
+function(join_paths joined_path first_path_segment)
+    set(temp_path "${first_path_segment}")
+    foreach(current_segment IN LISTS ARGN)
+        if(NOT ("${current_segment}" STREQUAL ""))
+            if(IS_ABSOLUTE "${current_segment}")
+                set(temp_path "${current_segment}")
+            else()
+                set(temp_path "${temp_path}/${current_segment}")
+            endif()
+        endif()
+    endforeach()
+    set(${joined_path} "${temp_path}" PARENT_SCOPE)
+endfunction()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,9 @@
 #
 # See https://github.com/pre-commit/pre-commit
 
+# third-party content
+exclude: ^tools/JoinPaths.cmake$
+
 repos:
 # Standard hooks
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,6 +198,9 @@ else()
 endif()
 
 include("${CMAKE_CURRENT_SOURCE_DIR}/tools/pybind11Common.cmake")
+# https://github.com/jtojnar/cmake-snips/#concatenating-paths-when-building-pkg-config-files
+# TODO: cmake 3.20 adds the cmake_path() function, which obsoletes this snippet
+include("${CMAKE_CURRENT_SOURCE_DIR}/tools/JoinPaths.cmake")
 
 # Relative directory setting
 if(USE_PYTHON_INCLUDE_DIR AND DEFINED Python_INCLUDE_DIRS)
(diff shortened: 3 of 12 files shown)


tools/JoinPaths.cmake
```diff
@@ -5,19 +5,21 @@
...
-            if(IS_ABSOLUTE "${current_segment}")
-                set(temp_path "${current_segment}")
-            else()
-                set(temp_path "${temp_path}/${current_segment}")
-            endif()
-        endif()
-    endforeach()
-    set(${joined_path} "${temp_path}" PARENT_SCOPE)
+  set(temp_path "${first_path_segment}")
+  foreach(current_segment IN LISTS ARGN)
```
eli-schwartz, line 13: Is there some way to tell pre-commit.ci that this file belongs to another repo and shouldn't be autoformatted as that will result in a confusing file delta between here and the upstream version? :eyes:

Skylion007: Yes, you can add an exclude key regex to the pre-commit-config.yaml for the relevant hook.

eli-schwartz: Filename seems to have a dot in front of it...

This file is very big and I have basically zero knowledge of how it works other than "frustrating thing that messes up my commit messages", can I have a small hint at what I'm looking at, here? :)

henryiii: https://github.com/pyodide/pyodide/blob/be1b3c2a6f3ae869300b5630f523dbfb15e3d2e3/.pre-commit-config.yaml#L1

is an example.

henryiii: Also could be worth mentioning this file can be removed once we depend on 3.20+.

eli-schwartz: Ah, thanks. Global exclude sounds helpful.

eli-schwartz: I added comments in CMakeLists.txt, marked with `# TODO: ` for easy searchability.

henryiii:
I'm traveling & at conferences, I'd like to take a look at the generated file. I might want to tweak the CMake code just a bit too. I'll have a bit more time starting Wednesday, so ping me ~Thursday if I don't get back to this by then. Thanks!

eli-schwartz:
Ping. :)



> I'd like to take a look at the generated file.

BTW, here is the generated code for the regular wheel:

```
$ bsdtar -xOf dist/pybind11-2.11.0.dev1-py3-none-any.whl pybind11/share/pkgconfig/pybind11.pc
prefix=${pcfiledir}/../../
includedir=${prefix}/include

Name: pybind11
Description: Seamless operability between C++11 and Python
Version: 2.11.0
Cflags: -I${includedir}
```

It's the same content for the pybind11_global wheel except, of course, the location.
File lists, more or less:

pybind11:
```
pybind11/share/cmake/pybind11/FindPythonLibsNew.cmake
pybind11/share/cmake/pybind11/pybind11Common.cmake
pybind11/share/cmake/pybind11/pybind11Config.cmake
pybind11/share/cmake/pybind11/pybind11ConfigVersion.cmake
pybind11/share/cmake/pybind11/pybind11NewTools.cmake
pybind11/share/cmake/pybind11/pybind11Targets.cmake
pybind11/share/cmake/pybind11/pybind11Tools.cmake
pybind11/share/pkgconfig/pybind11.pc
```


pybind11_global
```
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/FindPythonLibsNew.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11Common.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11Config.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11ConfigVersion.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11NewTools.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11Targets.cmake
pybind11_global-2.11.0.dev1.data/data/share/cmake/pybind11/pybind11Tools.cmake
pybind11_global-2.11.0.dev1.data/data/share/pkgconfig/pybind11.pc
```

eli-schwartz:
@henryiii ping

henryiii:
Thanks! I'm back in Princeton now after nearly a month of travel.

Is there any reason it's joining the path rather than hardcoding the separator, as in `${pcfiledir}/../../`? The generated file gets including in the wheel, and must be usable from all OS's (and ideally creatable on all OSs identically), so `${pcfiledir}\..\..\` would be invalid.

eli-schwartz:
The JoinPaths function uses forward slashes anyway, exactly for portability reasons. And the paths need to be joined because cmake's includedir may or may not be an absolute path. (It will be a relative path when invoking cmake from setup.py, though.)

eli-schwartz:
@henryiii the test that you added is failing due to platform differences between `\r\n` and `\n`. Not sure what to do about that, I do notice that the `pybind11Config.cmake` check only tries to assert that a `set(...)` exists in the file, it's very newline style agnostic.

eli-schwartz:
By the way, merging this PR would help Void Linux successfully cross-compile SciPy. See https://github.com/scipy/scipy/issues/16783

henryiii:
Sorry, in the third of four conferences in a row. And preparation for CPython 3.10rc1 is taking free time too. This just needs to address the failing test. Not too worried about the line endings, happy to normalize them in the test. I hopefully can do it by Monday, ping me if I don’t! (Or you can do it if you want)

eli-schwartz:
No problem, take your time. I won't have time before then either, anyway. :)

henryiii (approved):

rwgk:
This is totally over my head but I looked out of curiosity. Is this impression correct?

* pkg-config is build-system agnostic, but the implementation of the pybind11 pkg-config support relies on cmake?

If that's true, it might be nice to add that as a note to the PR description.

Initially the phrase "alternative to pybind11Config.cmake" in the PR description led me to believe the pkg-config support will not involve cmake at all.

eli-schwartz:
CMake is the build system that pybind11 uses, so CMake is used to generate the pkg-config file. Once generated (for example by producing a python wheel, or by running `cmake && make install` in a Linux distro) the resultant installable files are agnostic.

For the purposes of e.g. SciPy (and more generally meson) we don't involve CMake.

rwgk:
Got it, thanks for the explanation!

eli-schwartz:
@henryiii thanks for figuring out the CI. Really happy to see this landing. :)

PR #4074, opened 2022-07-16
CI job name "🐍 3 • GCC 10 • C++2a • x64" (changed from C++20, which is misleading)

rwgk:
## Description
Distinguish `cmake_cxx_standard` & `std_passed_to_compiler`, to make it more obvious what is actually being tested, and why `PYBIND11_CPP20` is not defined.

## Suggested changelog entry:

```rst

```

PR #4083, opened 2022-07-18
More systematic gcc & clang coverage

rwgk:
## Description
Based on https://github.com/pybind/pybind11/pull/4074#issuecomment-1188385580

Drops GCC 10 `--std=c++2a` (because it is an odd choice), adds 6 GCC / `-std` combinations not covered already but assumed to be what users would pick.

Also adds clang 11, 12, 13 (not covered at all before), with `--std=c++20`.

Resolves GCC 11 & 12 "redundant move in return statement" warnings.

Suppresses misleading clang warnings in the same locations.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -1865,6 +1865,10 @@ private:
         }
 
         auto result = returned_array::create(trivial, shape);
+#ifdef PYBIND11_DETECTED_CLANG_WITH_MISLEADING_CALL_STD_MOVE_EXPLICITLY_WARNING
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wreturn-std-move"
+#endif
 
         if (size == 0) {
             return result;
@@ -1879,6 +1883,9 @@ private:
         }
 
         return result;
+#ifdef PYBIND11_DETECTED_CLANG_WITH_MISLEADING_CALL_STD_MOVE_EXPLICITLY_WARNING
+#    pragma clang diagnostic pop
+#endif
     }
 
     template <size_t... Index, size_t... VIndex, size_t... BIndex>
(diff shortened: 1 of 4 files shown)


Skylion007 (approved):
This looks good to me.

include/pybind11/numpy.h
```diff
@@ -1865,9 +1865,13 @@ struct vectorize_helper {
...
         auto result = returned_array::create(trivial, shape);
+#ifdef PYBIND11_DETECTED_CLANG_WITH_MISLEADING_CALL_STD_MOVE_EXPLICITLY_WARNING
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wreturn-std-move"
+#endif
 
         if (size == 0) {
-            return std::move(result);
+            return result;
         }
```
Skylion007, line 1875: If you remove the std::move are the pragmas necessary here, anymore?

rwgk: Here a GCC 11 & 12 warning is at conflict with a clang 7 through 12 warning.

When GCC 11 & 12 complained, I simply removed the `std::move`. (Thinking: "Let's see what happens.")
Next CI: many clang failures.
Next CI: complete the clang coverage to better understand what is going on.

Looking at the results I concluded (**guessed** TBH): clang was wrong for a long time, but that was fixed in clang 13. The fix appears to stick: clang 14 & 15 agree.

Next CI: with pragmas for clang added.

I considered introducing something like
```
PYBIND11_SUPPRESS_MISLEADING_CLANG_WARNING_CALL_STD_MOVE_EXPLICITLY_BEGIN
...
PYBIND11_SUPPRESS_MISLEADING_CLANG_WARNING_CALL_STD_MOVE_EXPLICITLY_END
```
or
```
    return PYBIND11_KNOWN_REDUNDANT_STD_MOVE(result);
```
but decided against it because there are only two affected locations, and the implementation you see now is easier to understand just looking at the local code, without having to look elsewhere.

PR #4087, opened 2022-07-21
Fix missing undef in pytypes

Skylion007:
## Description
* Fixes a missing undef from #4065 PR.


rwgk (approved):
Definitely a good change. Just curious: did this cause an actual problem? (Thinking: no, but maybe I'm overlooking something?)

Skylion007:
> Definitely a good change. Just curious: did this cause an actual problem? (Thinking: no, but maybe I'm overlooking something?)

Nope, didn't cause a problem. Just noticed it when looking over the code base.

PR #4098, opened 2022-07-30
Open pybind11 namespace with consistent visility.

thomaseding:
## Description

Pybind opens the `pybind11` namespace in its library code via `PYBIND11_NAMESPACE` to capture visibility settings. The documentation does not open the `pybind11` namespace the same way, This implies that forward declaring some pybind11 classes can be done without `PYBIND11_NAMESPACE`, but this is incorrect in the general case. Ultimately this leads to visibility mismatches depending on definition and include order.

## Suggested changelog entry:

```rst
Users should use `PYBIND11_NAMESPACE` instead of using `pybind11` when opening namespaces. Using namespace declarations and namespace qualification remain the same as `pybind11`. This is done to ensure consistent symbol visibility.
```

### DIFF
--- a/docs/advanced/cast/custom.rst
+++ b/docs/advanced/cast/custom.rst
@@ -38,7 +38,7 @@ type is explicitly allowed.
 
 .. code-block:: cpp
 
-    namespace pybind11 { namespace detail {
+    namespace PYBIND11_NAMESPACE { namespace detail {
         template <> struct type_caster<inty> {
         public:
             /**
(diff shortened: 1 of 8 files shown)


rwgk (approved):
Thanks Thomas, this looks good to me.

From a small experiment under PR #4050 (merged into the smart_holder branch) I can confirm that this matters on some platforms (e.g. macOS while not using `-fvisibility=hidden`).

On a related note, I have a PR that shows that `-fvisibility=hidden` can be removed without losing any pybind11 features (#4072). When working on that PR I overlooked the changes this PR makes to the tests. I'm thinking that will never really matter, because nobody will run them with a mix of pybind11 versions, does that sound right? But regardless, being consistent is better, because test code is copied quite often as a starting point for something new.

Skylion007 (approved):

Issue #3815, opened 2022-03-21
[BUG]: Documentation on static member variables does not carry over


srydell:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

Hi,

First of all, thank you for this amazing library.

I have a small problem. Documentation written on `def_readonly_static` and `def_readwrite_static` does not carry over to python. Example:

```cpp
struct MyStruct {
  static int const i;
  static int j;
}
```

Initialized in another `C++` source file:

```cpp
int const MyStruct::i = 0;
int Mystruct::j = 0;
```

And with the `pybind11`:

```cpp
PYBIND11_MODULE(m, m) {
  pybind::class_<MyStruct>(m, "MyStruct", "")
    .def_readonly_static("i", &MyStruct::i, "This is not readable from python")
    .def_readwrite_static("j", &MyStruct::j, "This is not readable from python")
}
```

Reading the documentation string from python results in the help text for the type (in this case `int`):

```python
>>> import m
>>> print(m.MyStruct.i.__doc__)
int([x]) -> integer
int(x, base=10) -> integer

Convert a number or string to an integer, ...
```

The same is shown for `j`. When using a class property that is not static the documentation string is shown instead.

Is it possible to have the documentation for static member variables show in python aswell?

I am using

pybind11 v2.9.0
Linux

Thanks for taking the time.

2bndy5:
+1 to this. This is annoying because I can't properly document the static member using sphinx autodoc.

At first, I thought it was my binding code, but I'm beginning to doubt that. @pybind Given the steps outlined to reproduce (which align with my problematic experience), is there really a need for a reproducible example code/project?

Issue #2900, opened 2021-03-13
[BUG] Descriptions of static members are not printed via help command


burak-yildizoz:
## Issue description

The documentation gives an example about [providing description for a definition](https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments) which can be seen by `help`.

Static members of a class can be accessed via [`def_readonly_static` and `def_readwrite_static`](https://pybind11.readthedocs.io/en/stable/classes.html#instance-and-static-fields).

The problem is that all descriptions are visible with `help` command except that of static members.

## Reproducible example code

`include/ClassTest.hpp`
``` c++
#ifndef CLASSTEST_HPP
#define CLASSTEST_HPP

#include <string>

struct ClassTest
{
    static std::string static_data;
    static std::string print_static_data();

    std::string _data;
    ClassTest(const std::string& data);
    std::string print_data() const;
    std::string& change_data(std::string& data);
};

#endif  // CLASSTEST_HPP
```

---

`cpp2py.cpp`
``` c++
#include "ClassTest.hpp"

#include <pybind11/pybind11.h>
namespace py = pybind11;

PYBIND11_MODULE(class_test, m)
{
    m.doc() = "pybind11 example class";

    py::class_<ClassTest>(m, "ClassTest", "the example class")
        .def_readonly ("data_readonly", &ClassTest::_data, "read-only data" )
        .def_readwrite("data"         , &ClassTest::_data, "read-write data")
        .def_readonly_static ("static_data_readonly", &ClassTest::static_data, "static read-only data" )
        .def_readwrite_static("static_data"         , &ClassTest::static_data, "static read-write data")
        .def_static("print_static_data", &ClassTest::print_static_data,
            "static method that prints and returns a copy of static_data")
        .def(py::init<const std::string &>(), "initialize data")
        .def("print_data" , &ClassTest::print_data , "const method that prints data")
        .def("change_data", &ClassTest::change_data, "modifier method that returns reference to data");
}
```

---

`test_class.py`
``` python
import class_test
help(class_test)
```

Output:
```
Help on module class_test:

NAME
    class_test - pybind11 example class

CLASSES
    pybind11_builtins.pybind11_object(builtins.object)
        ClassTest
    
    class ClassTest(pybind11_builtins.pybind11_object)
     |  the example class
     |  
     |  Method resolution order:
     |      ClassTest
     |      pybind11_builtins.pybind11_object
     |      builtins.object
     |  
     |  Methods defined here:
     |  
[...]

PR #4101, opened 2022-07-31
feat: add type signatures to docstrings of properties.

frosch123:
## Description

Currently signatures with type information are generated into docstrings for functions.

This PR adds type information in docstrings of properties, like
```
value_property: int

Property documentation.
```

Generation of this type information is enabled by default, but can be disabled/enabled with the same setting `py::options::disable_function_signatures` as for functions.
I did not consider it necessary to have separate settings for function and property signatures.

## Suggested changelog entry:

```rst
* Add type signatures in docstrings of properties.
```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1378,14 +1378,32 @@ protected:
                                   handle fset,
                                   detail::function_record *rec_func) {
         const auto is_static = (rec_func != nullptr) && !(rec_func->is_method && rec_func->scope);
-        const auto has_doc = (rec_func != nullptr) && (rec_func->doc != nullptr)
-                             && pybind11::options::show_user_defined_docstrings();
+
+        std::string doc;
+        if (rec_func != nullptr) {
+            if (pybind11::options::show_function_signatures()) {
+                doc += name;
+                if (rec_func->signature != nullptr) {
+                    std::string sig = rec_func->signature;
+                    size_t ret = sig.rfind(" -> ");
+                    if (ret != std::string::npos) {
+                        doc += ": ";
+                        doc += sig.substr(ret + 4);
+                    }
+                }
+                doc += "\n\n";
+            }
+            if (rec_func->doc != nullptr && pybind11::options::show_user_defined_docstrings()) {
+                doc += rec_func->doc;
+            }
+        }
+
         auto property = handle(
             (PyObject *) (is_static ? get_internals().static_property_type : &PyProperty_Type));
         attr(name) = property(fget.ptr() ? fget : none(),
                               fset.ptr() ? fset : none(),
                               /*deleter*/ none(),
-                              pybind11::str(has_doc ? rec_func->doc : ""));
+                              pybind11::str(doc));
     }
 };
 
--- a/tests/test_docstring_options.cpp
+++ b/tests/test_docstring_options.cpp
@@ -79,10 +79,31 @@ TEST_SUBMODULE(docstring_options, m) {
             void setValue(int v) { value = v; }
             int getValue() const { return value; }
         };
-        py::class_<DocstringTestFoo>(m, "DocstringTestFoo", "This is a class docstring")
-            .def_property("value_prop",
+        py::class_<DocstringTestFoo> c(m, "DocstringTestFoo", "This is a class docstring");
+
+        c.def_property("value_prop1",
                           &DocstringTestFoo::getValue,
                           &DocstringTestFoo::setValue,
                           "This is a property docstring");
+
+        {
+            py::options nested_options;
+            nested_options.enable_user_defined_docstrings();
+
+            c.def_property("value_prop2",
+                          &DocstringTestFoo::getValue,
+                          &DocstringTestFoo::setValue,
+                          "This is a property docstring");
+
+            c.def_property("value_prop3",
+                          &DocstringTestFoo::getValue,
+                          &DocstringTestFoo::setValue);
+
+            options.disable_function_signatures();
+
+            c.def_property("value_prop4",
+                          &DocstringTestFoo::getValue,
+                          &DocstringTestFoo::setValue);
+        }
     }
 }
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -88,7 +88,14 @@ def test_docstrings(doc):
         Get value using a method
     """
     )
-    assert doc(UserType.value) == "Get/set value using a property"
+    assert (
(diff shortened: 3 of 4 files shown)


rwgk (commented):
I initiated our (Google's) global testing for this PR, to see if there is anything that trips over this change.

See also: PR #2621, which unfortunately seems stalled over questions how other tools harvest the docstrings.

Skylion007:
@frosch123 See also https://github.com/pybind/pybind11/issues/3815 which may also cause issues here with static properties.

tests/test_docstring_options.cpp
```diff
@@ -82,28 +82,26 @@ TEST_SUBMODULE(docstring_options, m) {
...
-                          &DocstringTestFoo::setValue);
+            c.def_property(
+                "value_prop3", &DocstringTestFoo::getValue, &DocstringTestFoo::setValue);
 
             options.disable_function_signatures();
 
-            c.def_property("value_prop4",
-                          &DocstringTestFoo::getValue,
-                          &DocstringTestFoo::setValue);
+            c.def_property(
```
Skylion007, line 103: What about static properties? Any ideas on how to support those as well?

rwgk:
> I initiated our (Google's) global testing for this PR, to see if there is anything that trips over this change.

Quick update: Unfortunately I'm seeing breakages with at least two root causes. The total number of breakages is a few thousands (out of millions). I will try to drill down as soon as I get a chance.

include/pybind11/pybind11.h
```diff
@@ -1378,14 +1378,32 @@ class generic_type : public object {
...
-        const auto has_doc = (rec_func != nullptr) && (rec_func->doc != nullptr)
-                             && pybind11::options::show_user_defined_docstrings();
+
+        std::string doc;
+        if (rec_func != nullptr) {
+            if (pybind11::options::show_function_signatures()) {
+                doc += name;
+                if (rec_func->signature != nullptr) {
+                    std::string sig = rec_func->signature;
+                    size_t ret = sig.rfind(" -> ");
```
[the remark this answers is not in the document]
frosch123: One could extend function_record with an additional member "return_type" or similar, but it would only be used for properties.
This appeared to invasive to me, so I did it the same as for constructors:
https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1046

tests/test_docstring_options.cpp
frosch123: There are multiple issues for this (#3815, #2900). I looked into it.

Expectation: The docstring of static property `FooClass.bar_prop` is shown in `help(FooClass)` and `help(FooClass.bar_prop)`.

Problem:
Both `help` calls read the documentation from `FooClass.bar_prop.__doc__`. However, `FooClass.bar_prop` is already the value of the property, so `__doc__` refers to the class of the value, not to the property.
So fundamentally this is as impossible as adding docstrings to attributes. (See rejected PEP 224: https://legacy.python.org/dev/peps/pep-0224/ )

Possible work-around:
It could be done like for enums. List the static properties in the documentation of the parent class:
https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L1961
Not sure whether this would look any good.

rwgk:
After a couple hours drilling into a deep pile of tooling, and a guess ... the diff below fixes all breakages reported in my previous comment.

I don't have a full explanation. I know that MyPy's stubgen is in the mix, but not much more. There is also a script (currently not open source AFAIK) that post-processes the stubgen output to generate a .pyi file for pytype. I don't know if it is stubgen or the post-processing that stumbles over the name of the property.

What I'm seeing is something like this:

This part is manually edited and untested but hopefully close enough:
```
    .def(py::init([](int left, int bottom, int width, int height) {
              return Rect{left, bottom, width, height};
            }),
            py::arg("left"), py::arg("bottom"), py::arg("width"),
            py::arg("height"));
    .def_readwrite("left", &Rect::left)
    .def_readwrite("bottom", &Rect::bottom)
    .def_readwrite("width", &Rect::width)
    .def_readwrite("height", &Rect::height);
```

.pyi file before this PR:
```
class Rect:
    bottom: int
    height: int
    left: int
    width: int
```

.pyi file with this PR:
```
class Rect:
    bottom: bottom
    height: height
    left: left
    width: width
```

I won't have time to drill into this much more. Does someone familiar with stubgen have a clue?

```diff
diff --git a/include/pybind11/pybind11.h b/include/pybind11/pybind11.h
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1384,18 +1384,18 @@ protected:
         std::string doc;
         if (rec_func != nullptr) {
             if (pybind11::options::show_function_signatures()) {
-                doc += name;
                 if (rec_func->signature != nullptr) {
                     std::string sig = rec_func->signature;
                     size_t ret = sig.rfind(" -> ");
                     if (ret != std::string::npos) {
-                        doc += ": ";
                         doc += sig.substr(ret + 4);
                     }
                 }
-                doc += "\n\n";
             }
             if (rec_func->doc != nullptr && pybind11::options::show_user_defined_docstrings()) {
+                if (!doc.empty()) {
+                    doc += "\n\n";
+                }
                 doc += rec_func->doc;
             }
         }
diff --git a/tests/test_class.py b/tests/test_class.py
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -91,7 +91,7 @@ def test_docstrings(doc):
     assert (
         doc(UserType.value)
         == """
-        value: int
+        int

         Get/set value using a property
     """
diff --git a/tests/test_docstring_options.py b/tests/test_docstring_options.py
--- a/tests/test_docstring_options.py
+++ b/tests/test_docstring_options.py
@@ -39,13 +39,13 @@ def test_docstring_options():
     # Suppression of user-defined docstrings for non-function objects
     assert not m.DocstringTestFoo.__doc__

-    assert m.DocstringTestFoo.value_prop1.__doc__ == "value_prop1: int\n\n"
+    assert m.DocstringTestFoo.value_prop1.__doc__ == "int"

     assert (
         m.DocstringTestFoo.value_prop2.__doc__
-        == "value_prop2: int\n\nThis is a property docstring"
+        == "int\n\nThis is a property docstring"
     )

-    assert m.DocstringTestFoo.value_prop3.__doc__ == "value_prop3: int\n\n"
+    assert m.DocstringTestFoo.value_prop3.__doc__ == "int"

     assert not m.DocstringTestFoo.value_prop4.__doc__
```

PR #4106, opened 2022-08-02
chore: Use PyObject_GenericGetDict and PyObject_GenericSetDict functions

Skylion007:
## Description
Remove custom implementation of PyObject_GenericGetDict and PyObject_GenericSetDict. These functions have been exposed since Python 3.3 and have both been part of the stable API since 3.10. Since we don't support Python 2 anymore, there is no reason to keep them around. 

## Suggested changelog entry:

```rst
* Use the modern PyObject_GenericGetDict and PyObject_GenericSetDict for handling dynamic attribute dictionaries.
```

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -533,11 +533,16 @@ inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type) {
     type->tp_traverse = pybind11_traverse;
     type->tp_clear = pybind11_clear;
 
-    static PyGetSetDef getset[] = {{const_cast<char *>("__dict__"),
-                                    PyObject_GenericGetDict,
-                                    PyObject_GenericSetDict,
-                                    nullptr,
-                                    nullptr},
+    static PyGetSetDef getset[] = {{
+#if PY_VERSION_HEX <= 0x03060000
+                                       const_cast<char *>("__dict__"),
+#else
+                                       "__dict__",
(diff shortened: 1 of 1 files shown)


include/pybind11/detail/class.h
```diff
@@ -558,9 +533,12 @@ inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type) {
     type->tp_traverse = pybind11_traverse;
     type->tp_clear = pybind11_clear;
 
-    static PyGetSetDef getset[] = {
-        {const_cast<char *>("__dict__"), pybind11_get_dict, pybind11_set_dict, nullptr, nullptr},
-        {nullptr, nullptr, nullptr, nullptr, nullptr}};
+    static PyGetSetDef getset[] = {{const_cast<char *>("__dict__"),
```
rwgk, line 536: When we drop 3.6 the `const_cast` can be removed (according to https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef 3.7 is the first version where `name` is `const char *`).

While at it, could you please make the `const_cast` conditional on `PY_VERSION_HEX`?

Maybe (untested):
```
#if PY_VERSION_HEX <= 0x03060000
const_cast<char *>("__dict__"),
#else
"__dict__",
#endif
```

Then we will not overlook this when we drop 3.6.

rwgk (approved):
Nice!

PR #4138, opened 2022-08-11
chore: Add pytests for constructing pytypes from iterable

Skylion007:
## Description
* I was going through the code and wanted to add some tests for the iterable to pytype ctors since I noticed are unit tests for the iterable class are working. Since I wrote the tests to verify this use case worked properly, I thought I might as well commit them.

### DIFF
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -109,9 +109,9 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("get_iterator", [] { return py::iterator(); });
     // test_iterable
     m.def("get_iterable", [] { return py::iterable(); });
-    m.def("get_set_from_iterable", [](const py::iterable &iter) { return py::set(iter); });
     m.def("get_frozenset_from_iterable",
           [](const py::iterable &iter) { return py::frozenset(iter); });
+    m.def("get_set_from_iterable", [](const py::iterable &iter) { return py::set(iter); });
     m.def("get_tuple_from_iterable", [](const py::iterable &iter) { return py::tuple(iter); });
     m.def("get_list_from_iterable", [](const py::iterable &iter) { return py::list(iter); });
     // test_float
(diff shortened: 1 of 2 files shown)


tests/test_pytypes.cpp
```diff
@@ -109,6 +109,13 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("get_iterator", [] { return py::iterator(); });
     // test_iterable
     m.def("get_iterable", [] { return py::iterable(); });
+    m.def("get_frozenset_from_iterable",
+          [](const py::iterable &iter) { return py::frozenset(iter); });
+    m.def("get_list_from_iterable", [](const py::iterable &iter) { return py::list(iter); });
+    m.def("get_set_from_iterable", [](const py::iterable &iter) { return py::set(iter); });
+    m.def("get_tuple_from_iterable", [](const py::iterable &iter) { return py::tuple(iter); });
+    // test_float
+    // test_float
```
rwgk, line 118: Oversight (two stray line)?

tests/test_pytypes.py
```diff
@@ -26,6 +26,22 @@ def test_iterator(doc):
...
+@pytest.mark.parametrize(
+    "pytype, from_iter_func",
+    [
+        (frozenset, m.get_frozenset_from_iterable),
+        (list, m.get_list_from_iterable),
+        (set, m.get_set_from_iterable),
+        (tuple, m.get_tuple_from_iterable),
+    ],
+)
+def test_from_iterable(capture, doc, pytype, from_iter_func):
```
rwgk, line 38: `capture` and `doc` don't appear to be used. Does it still work if you remove these two args?

rwgk (approved):
Nice!

tests/test_pytypes.py
Skylion007: @rwgk yea, I just wanted to be consistent with the rest of the functions in this file. Should I still remove them?

rwgk: Yes I think it's better to remove them. At least that's what I've been doing when adding new tests, only keep what's actually needed.

If you see unused args elsewhere, I wouldn't touch those in this PR, to not distract from the main purpose. (And not enter the slippery slope of feature creep.)

PR #4155, opened 2022-08-23
Words matter updates

messerb5467:
## Description

An industry trend towards inclusive language ([IBM's stance](https://www.ibm.com/blogs/think/2020/08/words-matter-driving-thoughtful-change-toward-inclusive-language-in-technology/), [Inclusive Naming](https://inclusivenaming.org/), [INCITS](https://www.incits.org/news-events/news-coverage/incits-leads-the-way-in-adopting-inclusive-terminology-in-standards)), has us scrubbing code and documentation of outdated terms and connotations as much as possible in favor of friendlier alternatives.

A scan of pybind caught the below changes that were easy to get. There were others regarding `PYBIND11_MASTER_PROJECT` and the like I deemed too aggressive because it would be a fundamental alteration to the build system. Any thoughts on that would be appreciated though.

Thanks!

### DIFF
--- a/tests/test_eigen.py
+++ b/tests/test_eigen.py
@@ -251,14 +251,14 @@ def array_copy_but_one(a, r, c, v):
 def test_eigen_return_references():
     """Tests various ways of returning references and non-referencing copies"""
 
-    master = np.ones((10, 10))
+    main = np.ones((10, 10))
     a = m.ReturnTester()
     a_get1 = a.get()
     assert not a_get1.flags.owndata and a_get1.flags.writeable
-    assign_both(a_get1, master, 3, 3, 5)
+    assign_both(a_get1, main, 3, 3, 5)
     a_get2 = a.get_ptr()
     assert not a_get2.flags.owndata and a_get2.flags.writeable
-    assign_both(a_get1, master, 2, 3, 6)
+    assign_both(a_get1, main, 2, 3, 6)
 
     a_view1 = a.view()
     assert not a_view1.flags.owndata and not a_view1.flags.writeable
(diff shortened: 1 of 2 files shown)


tests/test_eigen.py
```diff
@@ -251,14 +251,14 @@ def array_copy_but_one(a, r, c, v):
 def test_eigen_return_references():
     """Tests various ways of returning references and non-referencing copies"""
 
-    master = np.ones((10, 10))
```
rwgk, line 254: What do you think of `primary` as the new name?

`main` is a heavily used name for other things (C++ `main()`, Python `if __name__ == "__main__"`).

rwgk (commented):
Is there a chance that you could also scan the smart_holder branch?
(That's what we deploy from regularly for Google-internal use.)

tests/test_eigen.py
messerb5467: Good point. I'll change that around!

messerb5467:
> Is there a chance that you could also scan the smart_holder branch? (That's what we deploy from regularly for Google-internal use.)

That won't take me long at all. Just for your awareness, the tool uses a much stricter set of words than what I've shown so far. I'll have the updates soon.

messerb5467:
Here's the results for the smart holder branch:
[pybind11_smart_holder_results.tar.gz](https://github.com/pybind/pybind11/files/9410181/pybind11_smart_holder_results.tar.gz)

rwgk:
> Here's the results for the smart holder branch: [pybind11_smart_holder_results.tar.gz](https://github.com/pybind/pybind11/files/9410181/pybind11_smart_holder_results.tar.gz)

Thanks!

I looked through and it seems the only additional finding with respect to the master branch is this one:
```
include_scan_results.csv:84,include,white hat hacker,/include/pybind11/detail/descr.h:32:// here exploits the gray area (similar to a white hat hacker).
```

I happen to be the author of that line. I could just delete the `(similar to a white hat hacker)` part, it doesn't add a lot, but maybe there is already an established replacement? Do you happen to know?

rwgk:
BTW: I'd love to change the 'master' branch to 'main', although it will be a lot of work (I'd be OK to put up with it), but more importantly, there will be numerous broken external references. That's the part that always makes me hesitate: effectively making a decision about other people's time; I'd bet some people get stuck badly.

messerb5467:
> > Here's the results for the smart holder branch: [pybind11_smart_holder_results.tar.gz](https://github.com/pybind/pybind11/files/9410181/pybind11_smart_holder_results.tar.gz)
> 
> Thanks!
> 
> I looked through and it seems the only additional finding with respect to the master branch is this one:
> 
> ```
> include_scan_results.csv:84,include,white hat hacker,/include/pybind11/detail/descr.h:32:// here exploits the gray area (similar to a white hat hacker).
> ```
> 
> I happen to be the author of that line. I could just delete the `(similar to a white hat hacker)` part, it doesn't add a lot, but maybe there is already an established replacement? Do you happen to know?

Offensive security researcher would be the appropriate term from what I can see.

rwgk (approved):

messerb5467:
> BTW: I'd love to change the 'master' branch to 'main', although it will be a lot of work (I'd be OK to put up with it), but more importantly, there will be numerous broken external references. That's the part that always makes me hesitate: effectively making a decision about other people's time; I'd bet some people get stuck badly.

Gotcha. Usually you do the best you can and put out a change notice. Changing `master` to `main` is less worrisome. With how pervasive pybind is though I'd rather play it safe than sorry. Enough people already had their say about pypi's upgrade to 2FA and that was only 1% of the projects being deemed critical. Then you had maintainers trying to circumvent the critical label by deleting their release from pypi, breaking tons of people, and it just got reapplied via automation again anyways. I don't know if you have a user's email list, discord, slack etc, but you could put out a poll and see what the community thinks about the change and see if everyone thinks it appropriate. Should you decide to, maintaining master and main for some time period could be helpful for people to migrate.

henryiii:
Oh, I'd very much recommend _not_ having both `main` and `master`, as that breaks all GitHub's smart redirection.

henryiii (approved):

messerb5467:
> GitHub automatically redirects when you rename a branch, so most links still work. I haven't had much problem moving to main branches in other repos, like CLI11.

That's good to know. Thanks for letting us know :)

Skylion007:
There is a pre-commit hook that does this kind of semantic sanitization. I can look into enabling it in a bit.

henryiii:
> There is a pre-commit hook that does this kind of semantic sanitization. I can look into enabling it in a bit.

Which one? Would like it for another repo. :)

PR #4158, opened 2022-08-24
Implement classmethod pytype

### LINKED ISSUE #1693: Support classmethod

Skylion007:
## Description
I've been mulling over how to implement this for a while now and finally got it working without segfaults.

Closes #1693 and implements classmethod support in pybind11. This allows defining classmethods which have the class of the caller automatically prepended to the list of args. This is nice because it allows adding/modifying class variables through the python bindings directly without a helper C++ class or struct.

NB: One thing I have noticed during my testing is that while `__new__` has a classmethod signature, it is actually implemented as a static method. Therefore trying to add a classmethod onto `__new__` will cause errors.

@rwgk I would like some thoughts on how to think about designing some proper tests for this functionality as there are probably a lot of edge cases having to do with potential mem leaks and other odd behavior. We also do play fast and loose with the terminology of classmethod with our add_classmethod function in pybind11 so I might need to refactor this. 

Tests that may need to be added:
1. memleaks? (I am not if the added attributes are properly cleaned up).
2. docstring
3. overloading precedence / siblings

Additional things needed: documentation.

## Suggested changelog entry:

```rst
* Implement classmethod support
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1224,6 +1224,7 @@ inline bool PyUnicode_Check_Permissive(PyObject *o) {
 #endif
 
 inline bool PyStaticMethod_Check(PyObject *o) { return o->ob_type == &PyStaticMethod_Type; }
+inline bool PyClassMethod_Check(PyObject *o) { return o->ob_type == &PyClassMethod_Type; }
 
 class kwargs_proxy : public handle {
 public:
@@ -2087,6 +2088,11 @@ public:
 class staticmethod : public object {
 public:
     PYBIND11_OBJECT_CVT(staticmethod, object, detail::PyStaticMethod_Check, PyStaticMethod_New)
+};
+
+class classmethod : public object {
+public:
+    PYBIND11_OBJECT_CVT(classmethod, object, detail::PyClassMethod_Check, PyClassMethod_New)
 };
 
 class buffer : public object {
--- a/tests/test_class.cpp
+++ b/tests/test_class.cpp
@@ -62,7 +62,15 @@ TEST_SUBMODULE(class_, m) {
     };
 
     py::class_<NoConstructor>(m, "NoConstructor")
-        .def_static("new_instance", &NoConstructor::new_instance, "Return an instance");
+        .def_static("new_instance", &NoConstructor::new_instance, "Return an instance")
+        .def_classmethod(
+            "new_instance_uuid",
+            [](py::object &cls) {
+                py::int_ uuid = getattr(cls, "uuid", py::int_(0));
+                cls.attr("uuid") = uuid + py::int_(1);
+                return NoConstructorNew::new_instance();
+            },
+            "Returns a new instance and then increment the uuid");
 
     py::class_<NoConstructorNew>(m, "NoConstructorNew")
         .def(py::init([](const NoConstructorNew &self) { return self; })) // Need a NOOP __init__
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -31,6 +31,12 @@ def test_instance_new(msg):
     assert cstats.alive() == 0
 
 
+def test_classmethod(num_instances=10):
+    for i in range(num_instances):
+        assert getattr(m.NoConstructor, "uuid", 0) == i
+        m.NoConstructor.new_instance_uuid()
+
+
 def test_type():
     assert m.check_type(1) == m.DerivedClass1
     with pytest.raises(RuntimeError) as execinfo:
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -1578,6 +1578,18 @@ public:
         return *this;
     }
 
+    template <typename Func, typename... Extra>
+    class_ &def_classmethod(const char *name_, Func &&f, const Extra &...extra) {
+        cpp_function cf(std::forward<Func>(f),
+                        name(name_),
+                        is_method(*this),
+                        sibling(getattr(*this, name_, none())),
+                        extra...);
+        auto cf_name = cf.name();
+        attr(std::move(cf_name)) = classmethod(std::move(cf));
+        return *this;
+    }
+
     template <detail::op_id id, detail::op_type ot, typename L, typename R, typename... Extra>
     class_ &def(const detail::op_<id, ot, L, R> &op, const Extra &...extra) {
         op.execute(*this, extra...);


Skylion007:
Seems pypy doesn't expose the PyClassMethod_Type, is there a stable API way to implement the check_ method? Or even a good way to test it to verify it works. @rwgk @henryiii thoughts ?

rwgk:
> Seems pypy doesn't expose the PyClassMethod_Type, is there a stable API way to implement the check_ method? Or even a good way to test it to verify it works. @rwgk @henryiii thoughts ?

@mattip for the PyPy PyClassMethod_Type question

(I can look at this PR only later.)

Skylion007:
Interestingly noticed it's not part of the CPython Stable-API, which is really odd for a pytype. It appears that CPython wants projects to move to the PyMethodDef API, but that only supports 3.9 and higher: https://docs.python.org/3/c-api/structures.html#c.PyMethodDef

tests/test_class.cpp
```diff
@@ -62,7 +62,15 @@ TEST_SUBMODULE(class_, m) {
     };
 
     py::class_<NoConstructor>(m, "NoConstructor")
-        .def_static("new_instance", &NoConstructor::new_instance, "Return an instance");
+        .def_static("new_instance", &NoConstructor::new_instance, "Return an instance")
+        .def_classmethod(
+            "new_instance_uuid",
```
rwgk, line 67: Maybe `serial` or `seq` or similar? (It doesn't look like a universally unique identifier.)

tests/test_class.py
```diff
@@ -31,6 +31,12 @@ def test_instance_new(msg):
     assert cstats.alive() == 0
 
 
+def test_classmethod(num_instances=10):
+    for i in range(num_instances):
+        assert getattr(m.NoConstructor, "uuid", 0) == i
+        m.NoConstructor.new_instance_uuid()
+
```
rwgk, line 38: Maybe (untested):
```
    assert no hasattr(m.NoConstructor, "uuid")
    for i in range(num_instances):
        m.NoConstructor.new_instance_uuid()
        assert m.NoConstructor.uuid == i + 1
```

rwgk (commented):
> Tests that may need to be added:
> memleaks? (I am not if the added attributes are properly cleaned up).

GHA valgrind is happy, ASAN & MSAN in the Google environment, too. That's a great start.

The only other thing I'd do, because it's super easy and conclusive beyond a doubt: locally modify the new test(s) by inserting `while True:`, run the test, in another shell (on the same machine) run the top command for a couple minutes and watch `RES` to see if it is stable for the test.

> overloading precedence / siblings

I'd add a test similar to what you have, but with one argument, as a way to ensure that `extra...` is plumbed through.

TBH I need to learn what siblings are.

I wouldn't worry about overload precedence, that logic is somewhere completely different and I don't see how the new `.def_classmethod` could break anything there.

include/pybind11/pytypes.h
```diff
@@ -2089,6 +2090,11 @@ class staticmethod : public object {
     PYBIND11_OBJECT_CVT(staticmethod, object, detail::PyStaticMethod_Check, PyStaticMethod_New)
 };
 
+class classmethod : public object {
```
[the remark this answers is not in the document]
Skylion007: Yes, it is used in def_classmethod ?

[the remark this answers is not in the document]
rwgk: Oh, sorry, now I see it. (I was staring at the cpp_function there too much and overlooked the much simpler line below.)

virtuald:
Seems useful. Just add some docs and skip the old versions of pypy and we're done here?

tests/test_class.py
rwgk: Oh ...
Commit 2e3d29d from two months ago escaped my attention.
If you reply with Done or similar here it's more likely that that doesn't happen.

rwgk:
> Seems useful. Just add some docs and skip the old versions of pypy and we're done here?

Agreed.

virtuald:
FWIW, using this locally, docstring has 'self' as first parameter instead of 'cls'. The type is correct however. pybind11-stubgen will also need a fix to generate the classmethod correctly.

virtuald:
Here's a format-patch for the docstring:

```patch
From 7ebb255157d077440956fe9721bdd33ca364c1d1 Mon Sep 17 00:00:00 2001
From: Dustin Spicuzza <dustin@virtualroadside.com>
Date: Sat, 24 Dec 2022 14:39:47 -0500
Subject: [PATCH] Fix classmethod docstring

---
 include/pybind11/attr.h     | 16 ++++++++++++++++
 include/pybind11/pybind11.h |  2 +-
 tests/test_class.py         |  2 ++
 3 files changed, 19 insertions(+), 1 deletion(-)

diff --git a/include/pybind11/attr.h b/include/pybind11/attr.h
index db7cd8ef..53231e65 100644
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -26,6 +26,12 @@ struct is_method {
     explicit is_method(const handle &c) : class_(c) {}
 };
 
+/// Annotation for classmethods
+struct is_classmethod {
+    handle class_;
+    explicit is_classmethod(const handle &c) : class_(c) {}
+};
+
 /// Annotation for operators
 struct is_operator {};
 
@@ -426,6 +432,16 @@ struct process_attribute<is_method> : process_attribute_default<is_method> {
     }
 };
 
+/// Process an attribute which indicates that this function is a classmethod
+template <>
+struct process_attribute<is_classmethod> : process_attribute_default<is_classmethod> {
+    static void init(const is_classmethod &s, function_record *r) {
+        r->is_method = true;
+        r->scope = s.class_;
+        r->args.emplace_back("cls", nullptr, handle(), /*convert=*/true, /*none=*/false);
+    }
+};
+
 /// Process an attribute which indicates the parent scope of a method
 template <>
 struct process_attribute<scope> : process_attribute_default<scope> {
index 9fd4e214..bd2fd548 100644
@@ -1593,7 +1593,7 @@ public:
     class_ &def_classmethod(const char *name_, Func &&f, const Extra &...extra) {
         cpp_function cf(std::forward<Func>(f),
                         name(name_),
-                        is_method(*this),
+                        is_classmethod(*this),
                         sibling(getattr(*this, name_, none())),
                         extra...);
         auto cf_name = cf.name();
index dca01b9d..31fb6b33 100644
--- a/tests/test_class.py
+++ b/tests/test_class.py
@@ -37,6 +37,8 @@ def test_classmethod(num_instances=10):
         m.NoConstructor.new_instance_seq_id()
         assert m.NoConstructor.seq_id == i + 1
 
+    assert m.NoConstructor.new_instance_seq_id.__doc__.startswith("new_instance_seq_id(cls: type) ->")
+
 
 def test_type():
     assert m.check_type(1) == m.DerivedClass1
-- 
2.36.1
```

virtuald:
Just kidding, that doesn't work if you specify arguments. Updated comment. It seems like the two options are to modify the function record (triggers an internals API bump?) or go this route. This seems less invasive, but it's also less consistent with how args are done elsewhere.

PR #4162, opened 2022-08-29
Modify Include as a submodule doc

jenhaoyang:
## Description

The [cmake example](https://github.com/pybind/cmake_example) did not include extern dir, remove extern in case confusing reader.


henryiii:
Just below:

> This assumes you are placing your dependencies in ``extern/``, and that you are

jenhaoyang:
Got it.

PR #4167, opened 2022-08-29
Update Test: ConstructorStats with Extra Newline

ax3l:
This looks like it lacks a newline.

## Suggested changelog entry:

```rst
* update test: ``ConstructorStats`` with extra newline
```


henryiii:
While this is very likely what was intended, due to the indentation, putting it on the same line is also valid.

ax3l:
Sounds good. Changed label & changelog entry.

rwgk (approved):

PR #4168, opened 2022-08-31
fix: Python-3.12 compatibility

### LINKED ISSUE #4115: [BUG]: Python 3.12 support
### Required prerequisites

### Problem description

@tacaswell reported that https://github.com/python/cpython/pull/23205 from Python 3.12 breaks pybind11, which in turn breaks matplotlib. We should probably both fix this and see if we can reenable the `dev` tag for 3.12 testing (Might also be good to start playing with using `PyType_FromMetaclass` too).

### Reproducible example code

_No response_

sizmailov:
Enable dynamic attributes for `pybind11_static_property` for compatility with  cpython-3.12

## Description

The problem is that `property`-derived classes now use `__doc__` attribute, which in turn requires to enable "dynamic attributes". All that seems like a pessimisation which is not welcome here. I'm submitting the PR to demonstrate a possible solution if the compatibility issue will be solved on pybind side.

Sorry for all the trouble. I thought I'd been fixing `cpython`, not breaking `pybind`. :disappointed:  

## Suggested changelog entry:

```rst
Enable dynamic attributes for `pybind11_static_property` for Python-3.12
```

Closes #4115

### DIFF
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -55,6 +55,9 @@ extern "C" inline int pybind11_static_set(PyObject *self, PyObject *obj, PyObjec
     return PyProperty_Type.tp_descr_set(self, cls, value);
 }
 
+// Forward declaration to use in `make_static_property_type()`
+inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type);
+
 /** A `static_property` is the same as a `property` but the `__get__()` and `__set__()`
     methods are modified to always use the object type instead of a concrete instance.
     Return value: New reference. */
@@ -86,7 +89,11 @@ inline PyTypeObject *make_static_property_type() {
     if (PyType_Ready(type) < 0) {
         pybind11_fail("make_static_property_type(): failure in PyType_Ready()!");
     }
-
+#    if PY_VERSION_HEX >= 0x030C0000
+    // Since Python-3.12 property-derived types are required to
+    // have dynamic attributes (to set `__doc__`)
+    enable_dynamic_attributes(heap_type);
+#    endif
     setattr((PyObject *) type, "__module__", str("pybind11_builtins"));
     PYBIND11_SET_OLDPY_QUALNAME(type, name_obj);
 
--- a/include/pybind11/detail/class.h
+++ b/include/pybind11/detail/class.h
@@ -89,11 +89,14 @@ inline PyTypeObject *make_static_property_type() {
     if (PyType_Ready(type) < 0) {
         pybind11_fail("make_static_property_type(): failure in PyType_Ready()!");
     }
+
 #    if PY_VERSION_HEX >= 0x030C0000
+    // PRE 3.12 FEATURE FREEZE. PLEASE REVIEW AFTER FREEZE.
     // Since Python-3.12 property-derived types are required to
     // have dynamic attributes (to set `__doc__`)
     enable_dynamic_attributes(heap_type);
 #    endif
+
     setattr((PyObject *) type, "__module__", str("pybind11_builtins"));
     PYBIND11_SET_OLDPY_QUALNAME(type, name_obj);
 


rwgk:
Hi level questions, these may be naive or infeasible:

> Sorry for all the trouble. I thought I'd been fixing cpython, not breaking pybind. 

* Could the change in cpython still be undone or done differently? -- Background thoughts: In the most general case, `py::dynamic_attr()` / `enable_dynamic_attributes()` opens up pitfalls, therefore I look at that feature as a last resort or escape hatch. But then again, this PR only changes what's more-or-less a pybind11-internal type (IIUC; I haven't actually worked on that part of pybind11).

* Is this PR in any way related to `Py_TPFLAGS_MANAGED_DICT`? How does it play with PR #4142? (I'm still unclear what's better, dropping or merging the PR. What's your advice?)

* Is there a reasonable and meaningful way to set up a pybind11 CI job to test against cpython 3.12? At the moment we're only testing with cpython 3.11, using [something pre-packed](https://github.com/pybind/pybind11/blob/8756f16ed842e40406018df901f3219b231e2105/.github/workflows/upstream.yml#L70) (I don't know how it works under the hood, but I know with had/have to wait for b1, b2, rc1, rc2 releases). -- I.e. what I have in mind is a CI job that first builds Python from sources, then uses that to test pybind11.

sizmailov:
I believe python method resolution rules leave no alternative to dynamics attributes to correctly implement the `property.__init__` _and_ preserve `__doc__` in subclasses. At least in a non-intrusive way. 

There are more troubles for `__doc__` specifically because each derived class (implicitly) defines its own `__doc__` which takes precedence over the instance's `__doc__` defined in its parent class.

The feature freeze for Python 3.12 should be around May 2023, so I believe it can be undone or done differently.

Another possible solution would be to avoid subclassing built-in `property` and implement descriptor protocol from scratch, getting all the performance we want and correctness at the same time. But this approach would add 100+ lines to pybind source and potentially breaks compatibility. Do you think this would be worth a try? 

I read through the lined issues/PR, and I believe this issue is not related to `Py_TPFLAGS_MANAGED_DICT`. After all, we have just base class here, so this should not interfere.

WRT CI, I think someone just adds more images to the list once a new python version gets released. I think having an opportunity to test against an upstream python version is a nice feature. Although such CI builds will not be reproducible and prone to spurious errors due to rapid changes in CPython.

rwgk:
> Another possible solution would be to avoid subclassing built-in `property` and implement descriptor protocol from scratch, getting all the performance we want and correctness at the same time. But this approach would add 100+ lines to pybind source and potentially breaks compatibility. Do you think this would be worth a try?

First thought: it doesn't seem attractive, because it sounds like complex functionality is duplicated, with small tweaks. Is that a fair way to look at it? — If yes, we'll have the potential for divergence over time, with all the confusion that usually arises from such a situation.

> I read through the lined issues/PR, and I believe this issue is not related to `Py_TPFLAGS_MANAGED_DICT`. After all, we have just base class here, so this should not interfere.

Thanks for the explanation!

> WRT CI, I think someone just adds more images to the list once a new python version gets released. I think having an opportunity to test against an upstream python version is a nice feature. Although such CI builds will not be reproducible and prone to spurious errors due to rapid changes in CPython.

Yeah, we'd need some way to easily pin on a CPython commit, but that's probably easy to achieve. (I'm itching to try ... but someone must have done something similar already ...)

Coming back to this PR, it's tiny, and if it helps testing with pybind11 while 3.12 is still fluid, I think we should merge, but maybe with a special comment? E.g. `// PRE 3.12 FEATURE FREEZE. PLEASE REVIEW AFTER FREEZE.`

sizmailov:
I totally agree, re-implementing the `property` is not a picnic. Although, the maintenance should be rather low. I don't think Python would break C API or descriptor protocol too often. Overall, yeah, this option is not too appealing.

I [tried](https://github.com/sizmailov/pybind11/commit/30519d0d4ba7b97da72a264830a3aeb6209074e0) a no-brainer python build from source on CI, but it [failed](https://github.com/sizmailov/pybind11/runs/8181665753?check_suite_focus=true#step:17:33) at pybind linking stage. You are right, pinning a version is not a problem. I'll submit a PR if I squeeze something meaningful out of it.

rwgk:
> I [tried](https://github.com/sizmailov/pybind11/commit/30519d0d4ba7b97da72a264830a3aeb6209074e0) a no-brainer python build from source on CI, but it [failed](https://github.com/sizmailov/pybind11/runs/8181665753?check_suite_focus=true#step:17:33) at pybind linking stage. You are right, pinning a version is not a problem. I'll submit a PR if I squeeze something meaningful out of it.

JIC you find this useful: https://github.com/rwgk/pybind11_scons/blob/8c15a4c54a010beb650084c801a5c48464e930b7/SConscript#L155

(That small SConscript has pretty much all tricks required in one easy view, but Linux only.)

rwgk:
I forgot to mention, I find `-DCMAKE_VERBOSE_MAKEFILE=ON` (config option) super useful to be sure about what is actually happening.

rwgk (approved):
@henryiii @Skylion007 Do you want to take a look?

rwgk:
No response from @henryiii @Skylion007 for 9 days, I assume this PR is fine with them.

In the meantime 311rc2 was released.

I will try to:

* rebase this PR on current master
* add the python dev label (which will activate testing with 311rc2, to get two birds with one stone ... assuming it goes well)
* then merge

Hang on.

include/pybind11/detail/class.h
```diff
@@ -87,6 +90,13 @@ inline PyTypeObject *make_static_property_type() {
         pybind11_fail("make_static_property_type(): failure in PyType_Ready()!");
     }
 
+#    if PY_VERSION_HEX >= 0x030C0000
```
henryiii, line 93:
```suggestion
#if PY_VERSION_HEX >= 0x030C0000
```

include/pybind11/detail/class.h
```diff
@@ -87,6 +90,13 @@ inline PyTypeObject *make_static_property_type() {
...
+    // PRE 3.12 FEATURE FREEZE. PLEASE REVIEW AFTER FREEZE.
+    // Since Python-3.12 property-derived types are required to
+    // have dynamic attributes (to set `__doc__`)
+    enable_dynamic_attributes(heap_type);
+#    endif
```
henryiii, line 98:
```suggestion
#endif
```

Not that it's great, but we should follow the current styling. Would be okay if it was updated (older versions of clang-format didn't handle it, and still not sure if newer versions indent based on the surrounding code).

henryiii (commented):
Looks okay if tests pass.

include/pybind11/detail/class.h
rwgk: I just tried this change locally, but pre-commit clang-format changes it straight back the way it was (with the spaces that you want to delete here).

I'm confused ... did you assume clang-format accepts it either way? Or is there something else I'm missing?

henryiii: Okay, though it didn't format this since I saw code below without it, okay if it's being autoformatted!

henryiii (approved):
Fine by me if it passes 3.12 tests.

rwgk:
> Fine by me if it passes 3.12 tests.

Thanks!

tacaswell:
I can confirm that this fixes the problem for me.

henryiii:
We'll try to get a release out in the "near" future. "Near" is a fluid term, though.

Skylion007:
@henryiii @rwgk Do we need what commit in CPython necessitated this change? Want to make sure this expected behavior as opposed to a 3.12 bug.

Skylion007:
Ah, for future reference it's https://github.com/python/cpython/pull/23205

PR #4183, opened 2022-09-16
chore: Delete copy ctor/assign for GIL RAIIs

Skylion007:
## Description
I noticed some of the GIL RAIIs still had their potential copy assignment operators so I made sure they were explicitly deleted. This should prevent bugprone behavior of trying to copy them / move them around.

## Suggested changelog entry:

```rst
* GIL RAII scopes are non-copyable to avoid potential bugs.
```


henryiii (approved):

rwgk (approved):
Nice!
It might be good to add the two lines to `gil_scoped_acquire_local` in detail/internals.h as well. Not that it matters for the code as is, but these kind of fragments tend to get copied around. Having it complete there as well, the oversight/omission won't spread.

rwgk (approved):
Thanks!

PR #4184, opened 2022-09-16
chore: optimize buffer_info ctors with some extra std::move

Skylion007:
## Description
chore: remove some extra potential copies of strings in the buffer info ctors at the expense of a few extra std::move operations.

## Suggested changelog entry:

```rst
* chore: reduce some extra copies / allocations in pybind11 buffer_info the ctors
```

### DIFF
--- a/include/pybind11/buffer_info.h
+++ b/include/pybind11/buffer_info.h
@@ -11,6 +11,8 @@
 
 #include "detail/common.h"
 
+#include <utility>
+
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 PYBIND11_NAMESPACE_BEGIN(detail)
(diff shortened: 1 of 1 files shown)


rwgk (commented):
I'm fine with this, it's a small change, but are we sure this is an actual gain?
My understanding:
* The code as-is only copies references (i.e. pointers in disguise), the only `string` copy is at the end when `format` is copied to the `buffer_info` member.
* With the new version, the copy is in the two ctors, then there a a couple moves for the string to land in the member.

Now guessing:
It's much easier for an optimizer to get rid of the reference/pointer copies, than it is to optimize out the moves (the string move ctor code probably isn't completely trivial, but I didn't go to look).
Does that make sense?

Skylion007:
> The code as-is only copies references (i.e. pointers in disguise), the only string copy is at the end when format is copied to the buffer_info member.

Yeah, I was thinking of all the format() calls in the delegating ctors, but I forgot about NRVO which comes into play here.

Issue #4125, opened 2022-08-08
[PLAN] Tracking issue for 2.10.1


henryiii:
Known regressions https://github.com/pybind/pybind11/labels/regression :

- #4100
- #4105
- #4117
- #4195 - https://github.com/pybind/scikit_build_example/pull/45
- #4234

Failed fixes:

- #4123

Planned features:

- #4077
- #4258
- #4216

Fixes:

- #4119
- #4115
- #4168
- #4294

The next release of PyPy will help too. :)

henryiii:
I'm still strongly in favor of a 2.10.1. It will be be no worse than 2.10.0, and it needs to be done before 3.11 is released, which is combine up soon. Both of those "bigger" changes above really should sit in master for a bit before release.

Also, I don't think an ABI bump with a "go back" option is all that useful. Distributions like conda-forge can't handle "selectable" ABI bumps. A version of pybind11 has a set ABI, and they can run a migration if it changes. Let's release a 2.10.1 later this week, then put in those larger changes, maybe go ahead and merge some of the other ABI bump ideas we've had in the past, then do a 2.11 properly.

henryiii:
I'd prefer not to leave a release series "broken" - some users will undoubtedly be stuck on 2.10 for a while. A 2.10.1 with better Python 3.11 support would be seen as a "safer" upgrade than 2.11.0. I've moved to 2.10 everywhere and have been very happy with it. The ABI problem doesn't affect all users.

Even if we release 2.11 a couple of weeks later, I think we should leave the 2.10 series in the best possible state.

> If we explain well (a couple sentences will do here I think) we give people a choice.

Only if people read it, and only if they can. Conda-forge, for example, doesn't have a choice and this will trigger a migration. Given 2.10 is the 'last' of a very long ABI series, I wouldn't be surprised if it's a mini Python 2.

> I'm happy to cut the 2.11 release, as soon as https://github.com/pybind/pybind11/pull/4201 is merged.

"bigger" changes really should sit in master for a bit before release.

> How much longer do you think you need to fix up the scikit_build_example?

Probably this week. By next Monday, at the latest.

henryiii:
By "broken" I'm referring to the Python 3.11 fixes, not the ABI incompatibility, which I'm not too worried about, at the most I think it's likely quite small.

rwgk:
How about:

* You cut the 2.10.1 release when you get to it.
* We branch 2.10 before merging #4201 or #4228 (which ever comes first).
[...]

PR #4188, opened 2022-09-19
Fix for Windows compiler, missing object initializer

jiwaszki:
## Description

Added missing object initializer due to error while compiling with Visual Studio 2019 MSC v. 1924:
```
error: C2737: 'pybind11::overload_cast': 'constexpr' object must be initialized
```

## Suggested changelog entry:

```rst
* Fix MSVC 2019 v.1924 & C++14 mode error for `overload_cast`.
```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -1037,7 +1037,7 @@ template <typename... Args>
         || (defined(__clang__) && __clang_major__ == 5)
 static constexpr detail::overload_cast_impl<Args...> overload_cast = {};
 #    else
-static constexpr detail::overload_cast_impl<Args...> overload_cast;
+static constexpr detail::overload_cast_impl<Args...> overload_cast{};
 #    endif
 #endif
 
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -1033,12 +1033,7 @@ PYBIND11_NAMESPACE_END(detail)
 ///  - regular: static_cast<Return (Class::*)(Arg0, Arg1, Arg2)>(&Class::func)
 ///  - sweet:   overload_cast<Arg0, Arg1, Arg2>(&Class::func)
 template <typename... Args>
-#    if (defined(_MSC_VER) && _MSC_VER < 1920) /* MSVC 2017 */                                    \
-        || (defined(__clang__) && __clang_major__ == 5)
-static constexpr detail::overload_cast_impl<Args...> overload_cast = {};
-#    else
 static constexpr detail::overload_cast_impl<Args...> overload_cast{};
-#    endif
 #endif
 
 /// Const member function selector for overload_cast


jiwaszki:
One question - is it possible to port this change to 2.10.0 or add to future fix version?

include/pybind11/detail/common.h
```diff
@@ -1037,7 +1037,7 @@ template <typename... Args>
         || (defined(__clang__) && __clang_major__ == 5)
 static constexpr detail::overload_cast_impl<Args...> overload_cast = {};
 #    else
-static constexpr detail::overload_cast_impl<Args...> overload_cast;
+static constexpr detail::overload_cast_impl<Args...> overload_cast{};
```
rwgk, line 1040: Uhm ... wait, do we actually need this ugly `#ifdef`?

henryiii: You mean maybe the new version works on all compilers?

henryiii: Also why weren't we seeing the errors with newer MSVC? We should be testing with 2019 & 2022 (maybe not 2017 anywhere anymore).

rwgk: > You mean maybe the new version works on all compilers?

Yes, I tried it out, CI is green: https://github.com/pybind/pybind11/pull/4190

Whatever made the `#ifdef` necessary in the past, it doesn't seem to be an issue anymore with the compilers we support today.

jiwaszki: > Uhm ... wait, do we actually need this ugly #ifdef?

Actually if the `{}` is present, this `#ifdef` can be removed. I will cover that in next commit and run checks.

> Also why weren't we seeing the errors with newer MSVC? 

My bad, I have got lost in the logs of the project and actual MSVC build version is `1924`. Maybe this version is not covered? Also I see that `-DCMAKE_CXX_STANDARD=17/20` is set for Windows builds, is `C++14` covered as well?

Skylion007 (approved):

include/pybind11/detail/common.h
henryiii: I bet no one ever tried it with `{}`, but only with `= {}`.

I'm hoping to get a patch release out in ~1 week or so. Still have a few items on #4125 left.

henryiii (approved):

Skylion007:
@rwgk Did we actually test if this works C++14 on MSVC? Seems like our CI might not be running that edge case?

Skylion007:
Opened https://github.com/pybind/pybind11/pull/4191 to make sure

include/pybind11/detail/common.h
jiwaszki: @henryiii Thanks for the answer, can't wait to finally bump version in OpenVINO!

jiwaszki: @henryiii As fix version 2.10.1 is going to be released, will it include this PR and everything from the master branch up to this point? I am not accustom to pybind's release practices, I am really sorry for such questions.

(I do not want to spam in #4125 comments, hope you don't mind:))

henryiii: It should. I think the two breaking changes might go in before it releases, which means we'll branch, but so far master is 2.10.1. I'm going to finish investigation into the CMake caching issue in the next couple of days, then make a release.

henryiii: Missed my target, but 2.10.1 is out now.

PR #4194, opened 2022-09-21
perf: call reserve method in set and map casters

Skylion007:
## Description
* Our STL caster was smart enough to reserve space in the list_caster. However, std::unordered_map and std::unordered_set also have the ability to reserve space ahead of time. This PR updates our other casters to make use of this. Reserving the size of the container is even more important for these types than list_caster since resizes will also call all elements to be rehashed. I am rather surprised it took so long for someone to notice this issue.
* I wanted to separate out the messy enable_if_t condition into a has_reserve_method alias, but kept getting syntax errors.
* Tested by adding static_assert false to make sure the proper reserve_method overload is called when appropriate.

## Suggested changelog entry:

```rst
* Reserve space in set and STL map casters if possible. This will prevent unnecessary rehashing / resizing by knowing the number of keys ahead of time for Python to C++ casting. This improvement will greatly speed up the casting of large unordered maps and sets.
```

### DIFF
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -50,7 +50,7 @@ constexpr forwarded_type<T, U> forward_like(U &&u) {
 }
 
 // Checks if a container has a STL style reserve method.
-// The reserve method should also have a void return type.
+// This will only return true for a `reserve()` with a `void` return.
 template <typename C>
 using has_reserve_method = std::is_same<decltype(std::declval<C>().reserve(0)), void>;
 
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -59,6 +59,14 @@ struct set_caster {
     using type = Type;
     using key_conv = make_caster<Key>;
 
+private:
+    template <typename T = Type, enable_if_t<has_reserve_method<T>::value, int> = 0>
+    void reserve_maybe(const anyset &s, Type *) {
+        value.reserve(s.size());
+    }
+    void reserve_maybe(const anyset &, void *) {}
+
+public:
     bool load(handle src, bool convert) {
         if (!isinstance<anyset>(src)) {
             return false;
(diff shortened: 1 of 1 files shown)


henryiii (approved):

include/pybind11/stl.h
```diff
@@ -136,6 +147,15 @@ struct map_caster {
     PYBIND11_TYPE_CASTER(Type,
                          const_name("Dict[") + key_conv::name + const_name(", ") + value_conv::name
                              + const_name("]"));
+
+private:
+    template <
+        typename T = Type,
+        enable_if_t<std::is_same<decltype(std::declval<T>().reserve(0)), void>::value, int> = 0>
```
rwgk, line 154: Is this meant to be a stand-in for "value has a .reserve() method?" ?

Might be good to add a terse comment:
```
// This will work only for `reserve()` with `void` return (all std containers).
```
(Is "all std containers" true?)

rwgk (approved):

include/pybind11/stl.h
henryiii: "All" sounds like this works on non-set or non-map containers too. This is a fairly standard idiom for checking to see if a method exists, I don't think it needs a comment. Not against one, though.

Skylion007: @rwgk Not all std::containers have reserve (std::map, std::set, std::forward list) etc.. don't.

rwgk: I realize, what I had in mind is that all existing `reserve()` have a `void` return. If we're not sure, simply:

```
// This will work only for `reserve()` with `void` return.
```

So people don't have to parse the complicated `enable_if` to figure that out.

rwgk (approved):
Very nice, I really like this version!

PR #4195, opened 2022-09-22
fix: add flag for overriding classic Python search values

henryiii:
## Description

Should fix https://github.com/pybind/scikit_build_example/pull/45.

## Suggested changelog entry:

```rst
* CMake: revert overwrite behavior, now opt-in with ``PYBIND11_PYTHONLIBS_OVERRWRITE OFF``
```

### DIFF
--- a/tools/FindPythonLibsNew.cmake
+++ b/tools/FindPythonLibsNew.cmake
@@ -95,8 +95,23 @@ if(NOT PythonLibsNew_FIND_VERSION)
   set(PythonLibsNew_FIND_VERSION "3.6")
 endif()
 
+# Save variables that get set by PythonInterp
+macro(_PYBIND11_PUSH_IF_DEFINED name)
+  if(DEFINED "${name}")
+    set("_PYBIND11_ORIG_${name}" "${${name}}")
+  endif()
+endmacro()
+
+_pybind11_push_if_defined(PYTHON_INCLUDE_DIR)
+_pybind11_push_if_defined(PYTHON_MODULE_EXTENSION)
+_pybind11_push_if_defined(PYTHON_IS_DEBUG)
+
 find_package(PythonInterp ${PythonLibsNew_FIND_VERSION} ${_pythonlibs_required}
              ${_pythonlibs_quiet})
+
+unset(PYTHON_INCLUDE_DIR)
+unset(PYTHON_MODULE_EXTENSION)
+unset(PYTHON_IS_DEBUG)
 
 if(NOT PYTHONINTERP_FOUND)
   set(PYTHONLIBS_FOUND FALSE)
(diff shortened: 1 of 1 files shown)


henryiii:
Actually I think it's our stuff around FindPython that's breaking, not this one. Will make a new PR just in case I'm wrong and this history is useful.

henryiii:
No, I take that back, this is running the old one - where are these values being set? Not by FindPythonInterp as far as I can tell...

Skylion007 (approved):

tools/FindPythonLibsNew.cmake
```diff
@@ -95,9 +95,24 @@ if(NOT PythonLibsNew_FIND_VERSION)
   set(PythonLibsNew_FIND_VERSION "3.6")
 endif()
 
+# Save variables that get set by PythonInterp
```
rwgk, line 98: From my viewpoint as someone not knowing much about cmake, copying the title of this PR here, ideally even with a link to this PR, would seem super helpful. Otherwise I'd give too much credit to this code looking fresh. Maybe:

# For PyPy, broken FindPythonInterp values need to be overridden (see PR #4195).
# This macro saves variables that get set by PythonInterp.

henryiii: This isn't working, so I'm going to need to address this differently anyway. I'll probably do one of two things. If I can stop these from being set at all, then the current form is fine. If not, I can introduce a toggle and then you could override all or none. Something like that. :/

henryiii: (So far sill have no idea who is (incorrectly) setting these values)

tools/FindPythonLibsNew.cmake
```diff
@@ -151,9 +151,13 @@ if(NOT _PYTHON_SUCCESS MATCHES 0)
...
 endif()
 
+option(
+  PYBIND11_PYTHONLIBS_OVERRWRITE
+  "Overwrite cached values read from Python library (classic search). Turn off if cross-compiling and manually setting these values."
+  ON)
 # Can manually set values when cross-compiling
 macro(_PYBIND11_GET_IF_UNDEF lst index name)
-  if(NOT DEFINED "${name}")
+  if(PYBIND11_PYTHONLIBS_OVERRWRITE OR NOT DEFINED "${name}")
```
rwgk, line 160: Looks like one `R` to many: `OVERRWRITE`

I don't understand the high level enough to say something meaningful. What I see looks fine with my very superficial understanding.

henryiii:
```suggestion
  if(PYBIND11_PYTHONLIBS_OVERWRITE OR NOT DEFINED "${name}")
```

tools/FindPythonLibsNew.cmake
```diff
@@ -151,9 +151,13 @@ if(NOT _PYTHON_SUCCESS MATCHES 0)
   return()
 endif()
 
+option(
+  PYBIND11_PYTHONLIBS_OVERRWRITE
```
henryiii, line 155:
```suggestion
  PYBIND11_PYTHONLIBS_OVERWRITE
```

PR #4208, opened 2022-10-03
bugfix: Add error checking to list append and insert

Skylion007:
## Description
I was looking through our python bindings and noticed we didn't do any error checking for insert or append in the list object in pytypes.h. While the only way append or insert can fail if is the list is unable to grow (allocation failure)This change will properly propagate exceptions for both these methods in this rare case.

## Suggested changelog entry:

```rst
* (bugfix): Add proper error checking to C++ bindings for Python list append and insert.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -2022,14 +2022,20 @@ public:
     detail::list_iterator end() const { return {*this, PyList_GET_SIZE(m_ptr)}; }
     template <typename T>
     void append(T &&val) /* py-non-const */ {
-        PyList_Append(m_ptr, detail::object_or_cast(std::forward<T>(val)).ptr());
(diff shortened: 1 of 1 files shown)


rwgk (approved):
For `.insert()`: wow, how did this go unnoticed for so long? Great catch!

It would be ideal to add a test for the `.insert()` failure. I guess it's just a line or two.

Skylion007:
> For `.insert()`: wow, how did this go unnoticed for so long? Great catch!
> 
> It would be ideal to add a test for the `.insert()` failure. I guess it's just a line or two.

Whelp, I was wrong.
```python
a = []
a.insert(999, 1)
a == [1,]
```
That is perfectly valid and does not raise any exceptions. Yeah Python! :/ So yeah, this exception will only be raised on out of memory errors then. I've updated the PR description.

rwgk (approved):
> That is perfectly valid and does not raise any exceptions.

Oh! I would not have guessed.

PR #4212, opened 2022-10-04
ci: update pre-commit schedule

henryiii:
CC @rwgk. Monthly, also matching the commit messages to the same style as Dependabot.


rwgk:
> The final alternative options is `quarterly`, but I think `monthly` is fine. ;)

Yeah, I'm sure I can keep up with that, thanks!

PR #4216, opened 2022-10-05
Add `PYBIND11_SIMPLE_GIL_MANAGEMENT` option (cmake, C++ define)

Chekov2k:
## Description

This PR adds a `PYBIND11_SIMPLE_GIL_MANAGEMENT` option (cmake, C++ define). The new option may be useful to try when debugging GIL-related issues, to determine if the more complex default implementation is or is not to blame. See comments here for background.

This PR was triggered by https://github.com/pytorch/pytorch/issues/83101. While we could not reproduce the original problem in a unit test, many tests were added to test_gil_scoped.py trying to do so. That work exposed that test_gil_scoped.py has been finding deadlocks for a long time (years), but those were thus far ignored as "flakes". test_gil_scoped.py was updated to make it crystal clear when a `DEADLOCK` is detected. It turns out deadlocks are detected at a fairly high rate (ballpark 1 in 1000). This needs to be worked on separately. To not continue to pollute the CI results in the meantime, `SKIP_IF_DEADLOCK = True` was added in test_gil_scoped.py. The GitHub Actions CI logs can be harvested in the future to determine what platforms exactly are affected, and at what rate. Hopefully the information accumulating over time will lead to a better understanding and fixes, so that `SKIP_IF_DEADLOCK` can be removed eventually.

This PR also adds ThreadSanitizer compatibility to test_gil_scoped.py (closes #2754).

**WARNING: Please be careful to not create ODR violations when using the new option: everything that is linked together with mutual symbol visibility needs to be rebuilt.**

## Suggested changelog entry:

```rst
* A `PYBIND11_SIMPLE_GIL_MANAGEMENT` option was added (cmake, C++ define), along with many additional tests in test_gil_scoped.py. The option may be useful to try when debugging GIL-related issues, to determine if the more complex default implementation is or is not to blame. See #4216 for background. WARNING: Please be careful to not create ODR violations when using the option: everything that is linked together with mutual symbol visibility needs to be rebuilt.
```

PR #4218, opened 2022-10-06
Strip down gil.h to the simple version only.

rwgk:
## Description
Prompted by discussion under PR #4216. Please see there for background.

This PR removes ~170 lines of code, and two `internals` variables (guarded by internals version).

This passes Google-global testing (internal testing ID OCL:479233883:BASE:479302840:1665064423602:50af055b).

## Suggested changelog entry:

```rst
An outdated version of `gil_scoped_acquire` was removed from pybind11/gil.h, resolving #1276 and pytorch/pytorch#83101.
```

### DIFF
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -170,7 +170,9 @@ struct internals {
     PyTypeObject *default_metaclass;
     PyObject *instance_base;
 #if defined(WITH_THREAD)
+#    if PYBIND11_INTERNALS_VERSION > 5
     PYBIND11_TLS_KEY_INIT(tstate)
+#    endif // PYBIND11_INTERNALS_VERSION > 5
 #    if PYBIND11_INTERNALS_VERSION > 4
     PYBIND11_TLS_KEY_INIT(loader_life_support_tls_key)
 #    endif // PYBIND11_INTERNALS_VERSION > 4
@@ -180,6 +182,7 @@ struct internals {
         PYBIND11_TLS_FREE(loader_life_support_tls_key);
 #    endif // PYBIND11_INTERNALS_VERSION > 4
 
+#    if PYBIND11_INTERNALS_VERSION > 5
         // This destructor is called *after* Py_Finalize() in finalize_interpreter().
         // That *SHOULD BE* fine. The following details what happens when PyThread_tss_free is
         // called. PYBIND11_TLS_FREE is PyThread_tss_free on python 3.7+. On older python, it does
@@ -188,6 +191,7 @@ struct internals {
         // Neither of those have anything to do with CPython internals. PyMem_RawFree *requires*
         // that the `tstate` be allocated with the CPython allocator.
         PYBIND11_TLS_FREE(tstate);
+#    endif // PYBIND11_INTERNALS_VERSION > 5
     }
 #endif
 };
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "../gil.h"
 #include "../pytypes.h"
 
 #include <exception>
@@ -408,15 +409,7 @@ PYBIND11_NOINLINE internals &get_internals() {
         return **internals_pp;
     }
 
-    // Ensure that the GIL is held since we will need to make Python calls.
-    // Cannot use py::gil_scoped_acquire here since that constructor calls get_internals.
-    struct gil_scoped_acquire_local {
-        gil_scoped_acquire_local() : state(PyGILState_Ensure()) {}
-        gil_scoped_acquire_local(const gil_scoped_acquire_local &) = delete;
-        gil_scoped_acquire_local &operator=(const gil_scoped_acquire_local &) = delete;
-        ~gil_scoped_acquire_local() { PyGILState_Release(state); }
-        const PyGILState_STATE state;
-    } gil;
+    gil_scoped_acquire gil;
     error_scope err_scope;
 
     PYBIND11_STR_TYPE id(PYBIND11_INTERNALS_ID);
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -170,19 +170,21 @@ struct internals {
     PyTypeObject *default_metaclass;
     PyObject *instance_base;
 #if defined(WITH_THREAD)
-#    if PYBIND11_INTERNALS_VERSION > 5
+#    if PYBIND11_INTERNALS_VERSION < 6
     PYBIND11_TLS_KEY_INIT(tstate)
-#    endif // PYBIND11_INTERNALS_VERSION > 5
+#    endif // PYBIND11_INTERNALS_VERSION < 6
 #    if PYBIND11_INTERNALS_VERSION > 4
     PYBIND11_TLS_KEY_INIT(loader_life_support_tls_key)
 #    endif // PYBIND11_INTERNALS_VERSION > 4
+#    if PYBIND11_INTERNALS_VERSION < 6
     PyInterpreterState *istate = nullptr;
+#    endif // PYBIND11_INTERNALS_VERSION < 6
     ~internals() {
 #    if PYBIND11_INTERNALS_VERSION > 4
         PYBIND11_TLS_FREE(loader_life_support_tls_key);
 #    endif // PYBIND11_INTERNALS_VERSION > 4
 
-#    if PYBIND11_INTERNALS_VERSION > 5
+#    if PYBIND11_INTERNALS_VERSION < 6
         // This destructor is called *after* Py_Finalize() in finalize_interpreter().
         // That *SHOULD BE* fine. The following details what happens when PyThread_tss_free is
         // called. PYBIND11_TLS_FREE is PyThread_tss_free on python 3.7+. On older python, it does
@@ -191,7 +193,7 @@ struct internals {
         // Neither of those have anything to do with CPython internals. PyMem_RawFree *requires*
         // that the `tstate` be allocated with the CPython allocator.
         PYBIND11_TLS_FREE(tstate);
-#    endif // PYBIND11_INTERNALS_VERSION > 5
+#    endif // PYBIND11_INTERNALS_VERSION < 6
     }
 #endif
 };
@@ -442,11 +444,13 @@ PYBIND11_NOINLINE internals &get_internals() {
 #    if PY_VERSION_HEX < 0x03090000
         PyEval_InitThreads();
 #    endif
+#    if PYBIND11_INTERNALS_VERSION < 6
         PyThreadState *tstate = PyThreadState_Get();
         if (!PYBIND11_TLS_KEY_CREATE(internals_ptr->tstate)) {
             pybind11_fail("get_internals: could not successfully initialize the tstate TSS key!");
         }
         PYBIND11_TLS_REPLACE_VALUE(internals_ptr->tstate, tstate);
+#    endif
 
 #    if PYBIND11_INTERNALS_VERSION > 4
         if (!PYBIND11_TLS_KEY_CREATE(internals_ptr->loader_life_support_tls_key)) {
@@ -454,7 +458,9 @@ PYBIND11_NOINLINE internals &get_internals() {
                           "loader_life_support TSS key!");
         }
 #    endif
+#    if PYBIND11_INTERNALS_VERSION < 6
         internals_ptr->istate = tstate->interp;
+#    endif
 #endif
         builtins[id] = capsule(internals_pp);
         internals_ptr->registered_exception_translators.push_front(&translate_exception);
--- a/include/pybind11/gil.h
+++ b/include/pybind11/gil.h
@@ -10,175 +10,11 @@
 #pragma once
 
 #include "detail/common.h"
-#include "detail/internals.h"
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
-PYBIND11_NAMESPACE_BEGIN(detail)
+#if defined(WITH_THREAD)
 
-// forward declarations
-PyThreadState *get_thread_state_unchecked();
-
-PYBIND11_NAMESPACE_END(detail)
-
-#if defined(WITH_THREAD) && !defined(PYPY_VERSION)
-
-/* The functions below essentially reproduce the PyGILState_* API using a RAII
- * pattern, but there are a few important differences:
- *
- * 1. When acquiring the GIL from an non-main thread during the finalization
- *    phase, the GILState API blindly terminates the calling thread, which
- *    is often not what is wanted. This API does not do this.
- *
- * 2. The gil_scoped_release function can optionally cut the relationship
- *    of a PyThreadState and its associated thread, which allows moving it to
- *    another thread (this is a fairly rare/advanced use case).
- *
- * 3. The reference count of an acquired thread state can be controlled. This
- *    can be handy to prevent cases where callbacks issued from an external
- *    thread would otherwise constantly construct and destroy thread state data
- *    structures.
- *
- * See the Python bindings of NanoGUI (http://github.com/wjakob/nanogui) for an
- * example which uses features 2 and 3 to migrate the Python thread of
- * execution to another thread (to run the event loop on the original thread,
- * in this case).
- */
-
-class gil_scoped_acquire {
-public:
-    PYBIND11_NOINLINE gil_scoped_acquire() {
-        auto &internals = detail::get_internals();
-        tstate = (PyThreadState *) PYBIND11_TLS_GET_VALUE(internals.tstate);
-
-        if (!tstate) {
-            /* Check if the GIL was acquired using the PyGILState_* API instead (e.g. if
-               calling from a Python thread). Since we use a different key, this ensures
-               we don't create a new thread state and deadlock in PyEval_AcquireThread
-               below. Note we don't save this state with internals.tstate, since we don't
-               create it we would fail to clear it (its reference count should be > 0). */
-            tstate = PyGILState_GetThisThreadState();
-        }
-
-        if (!tstate) {
-            tstate = PyThreadState_New(internals.istate);
-#    if defined(PYBIND11_DETAILED_ERROR_MESSAGES)
-            if (!tstate) {
-                pybind11_fail("scoped_acquire: could not create thread state!");
-            }
-#    endif
-            tstate->gilstate_counter = 0;
-            PYBIND11_TLS_REPLACE_VALUE(internals.tstate, tstate);
-        } else {
-            release = detail::get_thread_state_unchecked() != tstate;
-        }
-
-        if (release) {
-            PyEval_AcquireThread(tstate);
-        }
-
-        inc_ref();
-    }
-
-    gil_scoped_acquire(const gil_scoped_acquire &) = delete;
-    gil_scoped_acquire &operator=(const gil_scoped_acquire &) = delete;
-
-    void inc_ref() { ++tstate->gilstate_counter; }
-
-    PYBIND11_NOINLINE void dec_ref() {
-        --tstate->gilstate_counter;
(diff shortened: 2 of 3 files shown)


rwgk:
All CI failures appear to be of the kind below.

Speculation: The failing test relies on `gil_scoped_acquire` to call `detail::get_internals()` as a side effect.

```
2022-10-06T14:35:24.7574365Z ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2022-10-06T14:35:24.7574761Z test_embed is a Catch v2.13.9 host application.
2022-10-06T14:35:24.7575357Z Run with -? for options
2022-10-06T14:35:24.7575524Z
2022-10-06T14:35:24.7575908Z -------------------------------------------------------------------------------
2022-10-06T14:35:24.7576241Z Threads
2022-10-06T14:35:24.7576689Z -------------------------------------------------------------------------------
2022-10-06T14:35:24.7577136Z /__w/pybind11/pybind11/tests/test_embed/test_interpreter.cpp:285
2022-10-06T14:35:24.7577818Z ...............................................................................
2022-10-06T14:35:24.7577989Z
2022-10-06T14:35:24.7578196Z /__w/pybind11/pybind11/tests/test_embed/test_interpreter.cpp:296: FAILED:
2022-10-06T14:35:24.7578582Z   REQUIRE( has_pybind11_internals_static() )
2022-10-06T14:35:24.7578878Z with expansion:
2022-10-06T14:35:24.7579097Z   false
2022-10-06T14:35:24.7579523Z
2022-10-06T14:35:24.8253489Z ===============================================================================
2022-10-06T14:35:24.8254019Z test cases:   12 |   11 passed | 1 failed
2022-10-06T14:35:24.8254379Z assertions: 1553 | 1552 passed | 1 failed
```

rwgk:
@jbms Question regarding e70b38bd0ebe0b849ed59174298a59261a8f0be8:

```diff
+        // This used to be triggered by py::gil_scoped_release().
+       // TODO: Explain why/how this is relevant to this unit test.
+       py::detail::get_internals();
        REQUIRE(has_pybind11_internals_static());
```

Locally I tried to remove those 4 lines and the test passes, but does that make sense?

I also tried to move the `REQUIRE(has_pybind11_internals_static());`

1. After the loop creating the threads,
2. To the very end of that unit test,

**without** adding the explicit `get_internals()`  call, but then the test fails in both cases. I.e. `get_internals()` is apparently never called as a side-effect of the rest of the code.

jbms:
It looks like that test isn't testing anything that uses pybind11 internals anymore, so it seems like we could just remove `REQUIRE(has_pybind11_internals_static());`.  Maybe that was just testing a side effect of the old `gil_scoped_release` implementation.

rwgk:
> It looks like that test isn't testing anything that uses pybind11 internals anymore, so it seems like we could just remove `REQUIRE(has_pybind11_internals_static());`. Maybe that was just testing a side effect of the old `gil_scoped_release` implementation.

Cool, thanks! Done: ad126f58ee5cbfb910bf3253162fba44870a93a1

While I was at it, I also took care of this one: be83638a22ad79e9cb3de4e0b1554abe757b3671

But I see the CI is unhappy, need to look & fix ...

jbms:
Given that the default internals version is still 4, I think the `tstate` change can also be gated by `PYBIND11_INTERNALS_VERSION > 4`.

rwgk:
@wjakob Could you please let us know if we still need the code introduced with https://github.com/pybind/pybind11/commit/39e97e6a490474af7a4fcc3da58b132f0f00e091#diff-4249f25293fdbc32ba7cb0a011c27ff420af64a7f95931553f953dc41f38915e?

rwgk:
I looked in nanobind: https://github.com/wjakob/nanobind/blob/526e2c798547185e37f66f77e3876311157eca95/include/nanobind/nb_misc.h

It is almost exactly the simple version here (although it is missing #4183). That strongly suggests @wjakob has decided that the complicated `gil_scoped_acquire` version in pybind11 is no longer useful. I believe we need to follow that lead and purge the code here as well, to stop sending pybind11 users into situations like #1276 and pytorch/pytorch#83101.

rwgk:
@henryiii @Skylion007 could you please comment on this PR?

My preference is to have this as a bug fix in 2.10.1.

Note that this was globally tested, which has been highly conclusive in the past. I'm not aware of any arguments for keeping the code that is known to cause issues in practice, and was found to have problems by inspection (see #4216).

Skylion007 (approved):

henryiii:
Have you checked performance? My first thought when seeing large amounts of code with a very simple fallback is that it was done that way for performance. Some codes might acquire and release the gil quite a bit, and they'll care a lot about performance. I'm also a little worried about dropping in a big change right before releasing a patch release, I'd rather have it sit in master a bit. That why I suggested making it opt-in for the patch release.

What about the three points listed in the code that you've deleted? Specifically, have you compiled NanoGUI with this change? It is supposed to demonstrate the need for the more complex custom code.

henryiii:
If you can prove https://github.com/mitsuba-renderer/nanogui works with this change, I'm in.

henryiii:
I tested this, and nanogui can't even compile with this change.

jbms:
Nanogui could copy the old code to its own repo if it wants to use it, or build using an older version of pybind11.  But since the old code is buggy and doesn't appear to be fixable without adding new apis to cpython, it would probably be better for nanogui to use a different approach entirely.

I don't think we should keep buggy behavior by default just to support a single user.

henryiii:
That's fine, but we don't knowingly break users in patch releases. This code has been linked to from the pybind11 source for years, so I'd be quite surprised if this was truly the only user. **This can't be stripped out till 2.11**. I'm okay to have the new behavior opt-in in 2.10.1, or not, but the default can't change till 2.11. If there are several more users, then I'd say 3.0 is needed for _breaking_ change, I'm only okay with it in a patch release because I'm still assuming the number of users is small.

I'll probably at least make a PR to nanogui to keep from breaking them in 2.11, either by limiting pybind11 or by coping the removed code there.

henryiii:
Actually, I don't think NanoGUI can copy it, because they can't add the required internals pointers. This will kill NanoGUI eventually if they can't adapt to the new code, either by limiting pybind11 version to something that will eventually no longer work, or by requiring an old internals version that we eventually kill off (which I'd like to do, setting the internals version is ugly, it's better if it's linked to pybind11 version). NanoGUI is just the obvious user, there could easily be more. (It's also a very nice looking project, there aren't many cross platform OSS graphic libraries that can build from scratch. Though it's also lacking working examples and tests...)

rwgk:
Thanks @henryiii for trying out the nanogui build. I left a message there:

https://github.com/mitsuba-renderer/nanogui/discussions/133

henryiii:
You should also open an issue in `Kitware/SMTK`, which would also be broken by this change.

https://github.com/Kitware/SMTK/blob/ac60e3e83857b450d61fb20d52e64d4f8fbc057b/smtk/operation/pybind11/PyOperation.h#L127

https://github.com/Kitware/SMTK/blob/ac60e3e83857b450d61fb20d52e64d4f8fbc057b/smtk/operation/pybind11/PyOperation.h#L140

rwgk:
> You should also open an issue in `Kitware/SMTK`, which would also be broken by this change.
> 
> https://github.com/Kitware/SMTK/blob/ac60e3e83857b450d61fb20d52e64d4f8fbc057b/smtk/operation/pybind11/PyOperation.h#L127
> 
> https://github.com/Kitware/SMTK/blob/ac60e3e83857b450d61fb20d52e64d4f8fbc057b/smtk/operation/pybind11/PyOperation.h#L140

Alright, two is a crowd. Thanks for digging this up, too.

The first thing we need is a test in pybind11, if we keep that functionality.

henryiii:
I expect it's really hard to to test threaded code like this, which is why it's missing tests. Also, I'm okay to have the simple version opt-in, then switch to opt-out eventually (Using the "advanced" (buggy) version breaks pypy compatibility anyway, so making the complex version require setup seems quite reasonable even if we can't remove it). Maybe the projects can work out how to avoid this, and then we can drop it.

rwgk:
https://github.com/Kitware/SMTK/issues/354

rwgk:
Let's first hear from them, what they actually need. Maybe we can change-and-reduce the non-simple version to something that works for 100% of pybind11 users?

> I expect it's really hard to to test threaded code like this,

I agree, but the lack of discipline has a cost, too. My gut estimate: the spread-out cost for people dealing with the buggy behavior, in accumulation, exceeds the cost for getting this tested and right. The problem of course is to collectively get organized enough to fix that.

rwgk:
Pragmatically, for 2.10.1, my current thinking:

* Keep the non-simple version but make the simple version the default.
* All the presumably few affected projects need to do is pass an extra define to keep working. We could add a comment near our define, asking them to link their PRs here. That will give us a nice overview what we're up against.
* While the vast majority of pybind11 users is already protected from repeats of the reported bugs.

henryiii:
Even if the two users we've identified any all the others we haven't found yet are able to quickly adapt, _I'm not comfortable shipping a known breaking change for an old feature in a patch release._ So we either don't put it in 2.10.1 at all, or we have an option that allows the simpler version to be activated. Since PyTorch would like the simpler version, I'm inclined for the latter.

---

Let's make the simple version optional and non-default in 2.10.1, and default in 2.11.0 but with an opt-out. We can make PRs fixing those two projects to be ready for 2.11. I think 2.11's not terribly far way.

I don't think "repeats of the reported bugs" is that big of an issue - many pybind11 users have happily been using pybind11, and there are lots of other bugs in issues that we can work on if we really want to work out all possible bugs. Let's not cause regressions and grief for ourselves and users, and take this slow.

rwgk:
I don't know, I've never knowingly released something untested & broken like this, I'm very uncomfortable having my name associated with such a trap. But let's see if we can get more data, maybe it's not as severe as I think?

* #1286 — @KyleFromKitware
* pytorch/pytorch#83101 — @Chekov2k

How much time did you and your teams spend dealing with issues related to the non-simple `gil_scoped_acquire` version? What's your opinion for resolving the conflict of interests (trap for many, needed for some)?

henryiii:
> I've never knowingly released something untested & broken

Pybind11 has over 400 issues open. Our test coverage is rather poor, too. We have lots of edge cases and ways to do broken things. Our release and maintenance are on a "best effort" basis. In this case, we do support dissociation of the gil (and have for many years) and we don't support reentry (except on PyPy, apparently). We can change what we support in a minor or major release, but not a patch release. We can add an opt-in flag in a patch release safely.

Patch releases are releases that try to only provide fixes. If we make it an opt-in fix, it's still a fix. If we make it opt-out, it's a breaking change. We follow Python's deprecation period for breaking changes - we will release at least one or two releases with warnings before we make a breaking change.

I'd like to get a patch release out before Monday. That doesn't provide much time. We have more time to decide for 2.11 if we want to change it from opt-in to opt-out.

jbms:
> Actually, I don't think NanoGUI can copy it, because they can't add the required internals pointers. This will kill NanoGUI eventually if they can't adapt to the new code, either by limiting pybind11 version to something that will eventually no longer work, or by requiring an old internals version that we eventually kill off (which I'd like to do, setting the internals version is ugly, it's better if it's linked to pybind11 version). NanoGUI is just the obvious user, there could easily be more (and it has a combined 5k stars and ~800 forks). (It's also a very nice looking project, there aren't many cross platform OSS graphic libraries that can build from scratch. Though it's also lacking working examples and tests...)

I suppose nanogui can't literally copy the code, but there is no reason it needs to use pybind11's internals to hold the thread local storage identifier --- it can create it its own thread local storage identifier that it stores wherever it wishes.

jbms:
> Even if the two users we've identified and all the others we haven't found yet are able to quickly adapt, _I'm not comfortable shipping a known breaking change for an old feature in a patch release._ So we either don't put it in 2.10.1 at all, or we have an option that allows the simpler version to be activated. Since PyTorch would like the simpler version, I'm inclined for the latter.

Is there reason that the next release needs to be called 2.10.1?  It could be called 2.11 instead.

henryiii:
Because we need to make a release before Python 3.11 release on Monday, and we aren't ready to ship breaking changes. We are not making a breaking change then releasing a couple of days later. Master has already moved on past the 2.10 branch, so we can make it opt-out immediately and then make it opt-in in the backport, I don't care. But "fixing" a perceived bug in pybind11 that's existed for multiple years and forcing your fix on everyone else does not need an emergency minor release tomorrow.

If it makes everyone feel better, I can just not backport #4216 at all and you can pretend "2.10.0" and "2.10.1" are the same thing.

henryiii:
To be clear, the pybind11 gil handling is not _untested & broken_. It is tested, and works for the 100K+ pybind11 users. It simply does not support nesting, and this change would allow it to support nesting, which at least a few users (one identified), at the expense of being able to dissociate the handling from the current thread, which at least a couple of (identified) users need.

rwgk:
> untested

It is obviously untested in pybind11 (which is what I meant), otherwise the CI for this PR wouldn't pass.

henryiii:
If we take this more slowly, we can copy the code needed for nanogui, provide that as an example in the upgrade guide for 3.11, and gently move to the new version in a few _weeks_ - 2.11 is not years away or anything terrible. The release in the next couple of days could include this as an opt-in preview, which PyTorch (which I think already worked around this anyway?) could then opt-in to, along with anyone else that encounters this issue. There is no need to rush 2.11 just because we need to ship a few fixes for 2.10 before Python 3.11.

henryiii:
> It is obviously untested in pybind11 (which is what I meant), otherwise the CI for this PR wouldn't pass.

I meant the gil handling is tested. The only thing that's not tested is the dissociation feature (which is hard to test in a test running a python process), or nesting (which is unsupported before this). But the gil handling code is very much tested, and used in many projects.

rwgk:
A bit random, but maybe useful for later, this PR does NOT fix the TSAN error I've seen ever since I first ran the pybind11 tests with the clang sanitizers (~2 years ago). See also PR #2754. (Note that the assertion failures also appear regularly in flakes, (almost?) always Windows.)

```
============================= test session starts ==============================
platform linux -- Python 3.9.15, pytest-6.2.5, py-1.10.0, pluggy-0.9.0
cachedir: /build/work/.../google3/tmp
C++ Info: Clang google3-trunk (11897708c0229c92802e747564e7c34b722f045f) C++17 __pybind11_internals_v4_clang_libcpp_cxxabi1002__
rootdir: /build/work/.../google3/runfiles/google3/third_party/pybind11/tests, configfile: pytest.ini
collected 5 items

test_gil_scoped.py ==10081==ThreadSanitizer: starting new threads after multi-threaded fork is not supported. Dying (set die_after_fork=0 to override)
F==10082==ThreadSanitizer: starting new threads after multi-threaded fork is not supported. Dying (set die_after_fork=0 to override)
F==10083==ThreadSanitizer: starting new threads after multi-threaded fork is not supported. Dying (set die_after_fork=0 to override)
F..

=================================== FAILURES ===================================
___________________ test_python_to_cpp_to_python_from_thread ___________________

    def test_python_to_cpp_to_python_from_thread():
        """Makes sure there is no GIL deadlock when running in a thread.

        It runs in a separate process to be able to stop and assert if it deadlocks.
        """
>       assert _run_in_process(_python_to_cpp_to_python_from_threads, 1) == 0
E       assert 66 == 0
E        +  where 66 = _run_in_process(_python_to_cpp_to_python_from_threads, 1)


test_gil_scoped.py:59: AssertionError
__________ test_python_to_cpp_to_python_from_thread_multiple_parallel __________

    def test_python_to_cpp_to_python_from_thread_multiple_parallel():
        """Makes sure there is no GIL deadlock when running in a thread multiple times in parallel.

        It runs in a separate process to be able to stop and assert if it deadlocks.
        """
>       assert _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=True) == 0
E       assert 66 == 0
E        +  where 66 = _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=True)


test_gil_scoped.py:68: AssertionError
_________ test_python_to_cpp_to_python_from_thread_multiple_sequential _________

    def test_python_to_cpp_to_python_from_thread_multiple_sequential():
        """Makes sure there is no GIL deadlock when running in a thread multiple times sequentially.

        It runs in a separate process to be able to stop and assert if it deadlocks.
        """
>       assert (
            _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=False) == 0
        )
E       assert 66 == 0
E        +  where 66 = _run_in_process(_python_to_cpp_to_python_from_threads, 8, parallel=False)


test_gil_scoped.py:77: AssertionError
- generated xml file: /build/work/5db35a65bbd644542f198253cf4c0968ea07/google3/blaze-out/k8-tsan-fastbuild/testlogs/third_party/pybind11/tests/test_gil_scoped/test.xml -
=========================== short test summary info ============================
FAILED test_gil_scoped.py::test_python_to_cpp_to_python_from_thread - assert ...
FAILED test_gil_scoped.py::test_python_to_cpp_to_python_from_thread_multiple_parallel
FAILED test_gil_scoped.py::test_python_to_cpp_to_python_from_thread_multiple_sequential
========================= 3 failed, 2 passed in 0.55s ==========================
```

henryiii:
Did you test this? I copied the contents of #1276 on ubuntu 18.04 (since I was there for something else) and still get:

```console
$ git fetch origin pull/4218/head:MASTER
$ g++ -I/usr/include/python3.6m -I/usr/include/python3.6m -I/pybind11/include -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g   -fstack-protector-strong -Wformat -Werror=format-security  -g -flto -fuse-linker-plugin -ffat-lto-objects -std=c++11 -Wl,-Bsymbolic-functions  -Wl,-z,relro -fPIC master.cpp -o master -std=c++14 -L/usr/lib/python3.6/config-3.6m-x86_64-linux-gnu -L/usr/lib -lpython3.6m -lpthread -ldl  -lutil -lm  -Xlinker -export-dynamic
./master
Calling C++ code
Segmentation fault
```

rwgk:
> Did you test this on anything that was supposed to be fixed by it?

To answer this question (sorry for delay): I didn't test with the #1276 code and only glanced at that bug and #1322 a little bit before.

My starting point was #4216, the idea here was to try what happens if we only have the simple code: the CI passed, and Google global testing.

#1276 is from 4 yr 8 mo ago. I don't see references to Python 3 in the comments, only Python 2.7. I didn't spot details about platforms and Python versions. With so much uncertainty after such a long time, is this still something useful to look at?

I'm thinking it's best to leave that old issue behind (not reference it anymore) and only focus on #4216.

Considering the uses of the non-simple version you found (comments here), this PR is too radical. I agree we need to keep the non-simple version around for a while. Our only difference seems to be when to flip the default, with me on the "as soon as possible" side, and you on the "let's wait" IIUC.

For Google-internal use, I still want to flip the default as soon as possible. I don't want other Google engineers to find out the hard way (like the PyTorch people) that they need the simple version, when they probably never even became aware of the features of the non-simple version, but just wanted a convenient way to ensure the GIL is held in a section of code.

I could flip the default only on the smart_holder branch, that would achieve my goal for Google's purposes.

I'll convert this PR to draft mode. Rationale: Q: Is it better to just add a define as in #4216, or also fold in the `internals` changes here? A: I believe purging the two members (`tstate`, `istate`) doesn't do anything measurable to compile times and binary sizes, and if we cannot purge the non-simple code, the complication with the `#ifdef`s here is just distracting. I.e. this PR is mainly useful as a starting point for purging the non-simple code later.

rwgk:
Closing this PR now, I think it will be completely stale by the time we get to removing the non-simple code. The change to test_embed/test_interpreter.cpp needed to be transferred to #4216, so that the tests pass when `PYBIND11_SIMPLE_GIL_MANAGEMENT` is defined. I also marked the `tstate` and `istate` members of `internals` with comments under #4216, so that we don't forget to remove them eventually. Everything else is really easy to figure out from scratch, without looking at this PR again.

PR #4220, opened 2022-10-06
Work-Around: NVCC 11.4.0 - 11.8.0

ax3l:
Adds a targeted NVCC work around for limited number of CUDA releases. Fixed in NVCC development.

Fix #4193

## Suggested changelog entry:

```rst
* work-around for Nvidia's CUDA nvcc compiler in versions 11.4.0 - 11.8.0
```

PR #4227, opened 2022-10-07
[smart_holder] git merge master

rwgk:
## Description
Helper/scratch PR for testing.

## Suggested changelog entry:

```rst

```

PR #4228, opened 2022-10-07
Reproducer and fix for issue encountered in smart_holder update.

rwgk:
## Description
Reproducer and fix for compilation failure related to PR #4220.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -1178,5 +1178,12 @@ constexpr inline bool silence_msvc_c4127(bool cond) { return cond; }
 #    define PYBIND11_DETAILED_ERROR_MESSAGES
 #endif
 
+#if defined(__CUDACC__) && (__CUDACC_VER_MAJOR__ == 11) && (__CUDACC_VER_MINOR__ >= 4)            \
+    && (__CUDACC_VER_MINOR__ <= 8) && !defined(PYBIND11_USING_WORKAROUND_FOR_CUDA_11_4_THROUGH_8)
+// Nvidia's NVCC is broken between 11.4.0 and 11.8.0
+//   https://github.com/pybind/pybind11/issues/4193
+#    define PYBIND11_USING_WORKAROUND_FOR_CUDA_11_4_THROUGH_8
+#endif
+
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -1178,12 +1178,5 @@ constexpr inline bool silence_msvc_c4127(bool cond) { return cond; }
 #    define PYBIND11_DETAILED_ERROR_MESSAGES
 #endif
 
-#if defined(__CUDACC__) && (__CUDACC_VER_MAJOR__ == 11) && (__CUDACC_VER_MINOR__ >= 4)            \
-    && (__CUDACC_VER_MINOR__ <= 8) && !defined(PYBIND11_USING_WORKAROUND_FOR_CUDA_11_4_THROUGH_8)
-// Nvidia's NVCC is broken between 11.4.0 and 11.8.0
-//   https://github.com/pybind/pybind11/issues/4193
-#    define PYBIND11_USING_WORKAROUND_FOR_CUDA_11_4_THROUGH_8
-#endif
-
 PYBIND11_NAMESPACE_END(detail)
 PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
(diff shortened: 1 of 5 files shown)


ax3l (approved):
Thanks, that better to narrow the work-around.

Can we make the smart holder branch part of CI?

ax3l (changes requested):
One thing to consider: sfinae is slow. Potentially keep the ifdefs for the work-around?

Also the comments are gone to describe why we do the construct.

rwgk:
> One thing to consider: sfinae is slow.

Do you have measurements or references?

I believe this change will have an unmeasurable impact on the compilation time, but it would be interesting to see data, if someone wants to spend the time collecting them. However, I believe (very strongly) that any such micro-level optimizations have a negative value (timer better spent somewhere else), at least from my perspective (I regularly get asked what I did with my time and what the impact was).

> Potentially keep the ifdefs for the work-around?

It's not a workaround anymore IMO, but a clearly better alternative: it keeps implementation details of operators.h (`<op_id id, op_type ot, typename L, typename R>`) out of pybind11.h. It just so happened that we saw that potential only as a side-effect of looking for a workaround. Of course, it's not a generally applicable approach, because it's intrusive on the type we're targeting with the `enable_if_t`, but the target type here is in _our_ detail namespace, therefore it would be silly IMO to not take advantage of the opportunity, now that we know we have it (works with all compilers on GitHub and passed global testing with the Google toolchain).

> Also the comments are gone to describe why we do the construct.

Because it's a universally better alternative. I believe it's wise to not touch that code again unless we have some kind of breakage again.

I'm hoping to deploy the smart_holder branch update (#4227) Google-internally tomorrow morning. I jumped in here only to not set you back with a rollback because of my timeline. However, if you feel this needs more work, could you please merge #4230 and work on it in your time? It's pretty quick for me to redo #4227 after the rollback.

> Can we make the smart holder branch part of CI?

The smart_holder branch doubles the CI load, because it needs to be tested with 1. smart_holder as add-on/option, 2. smart_holder as the default holder. I believe it would be too burdensome at this time to double the CI load that on master. (There are a few other reasons to keep it separate that are complicated to explain.) That's why I reduced the test that fails to build on the smart_holder branch and added it to test_class on master.

ax3l:
Ref: SFINAE is generally the worst for compile time

https://thephd.dev/sol3-compile-times-binary-sizes

ax3l (approved):

rwgk:
Thanks Axel! I can't read the article tonight but will look tomorrow.
Merging here, to then redo the smart_holder update (for simpler merge history; I don't squash), so it's ready for tomorrow morning.

rwgk:
@ax3l I haven't looked at the paper yet, but I made a sheet with [Compile times before & after merging PR #4228](https://docs.google.com/spreadsheets/d/18Nir1-pfsr1OJVqQhBnff6IKU2EO6wgVs9qNk8s1Z7o/edit#gid=439133634) ("cpp only sorted" subsheet).

Everything needed to reproduce it (minus my hardware & exact OS) is archived here:

https://github.com/rwgk/pybind11_scons/tree/b168c1e10b2fbf525a8a3856c0c372670dc5f960/compile_times_before_and_after_merging_pr4228

My conclusion:

* If anything, this PR made the compile times a little bit shorter.
* But I don't really believe much happened: note that column `before #3` in the sheet has systematically longer times. See also the plot in the sheet, the yellow spikes. Why? I have no clue.
* From those data I think it's definitely fair to say that the impact this PR had on the compile times is in the noise.

Please let me know if you see a flaw in how I obtained the data.

rwgk:
I just took a super quick look at the paper. Nice that they put this right at the top:

```
Version	sol2	sol3 (patron-only alpha)	sol3 (now)
Build Time	2 hours, 3 minutes	2 hours, 11 minutes	1 hour, 53 minutes
Binary Size	119.0 MB	138.0 MB	103.0 MB
```

After seeing those numbers I zoned out straightaway. Savings, yes, but that's barely enough to get a cup of coffee, in a 2 hour wait.

My take home: Just buy another core and code on.

**Human time is so much more expensive.**

ax3l:
Thank you for doing the detailed comparison, great that is stays the same / is potentially a bit faster now.
For curiosity, which compiler did you use?

> My take home: Just buy another core and code on.

Generally yes, but for header-only libs that go into a single TU, you cannot parallelize your critical path (your longest TU).

rwgk:
> similar as in the big rule of Chiel figure in that post?

I learned the hard way to focus on the "What do actually I want to do?" more than worrying about the details, although they are interesting. Then profile, and optimize the bits that provably matter.

> For curiosity, which compiler did you use?

```
$ clang++ --version
Debian clang version 14.0.6-2
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

(It's in the README.md but buried in the middle.)

> Thinking about it, `.def()` in pybind11 is not such a location, since we can indeed parallelize that by splitting in user code 👍

Yes, it's slightly on the cumbersome side, but you can basically break it down to any level of granularity.

PR #4230, opened 2022-10-08
Revert "Work-Around: NVCC 11.4.0 - 11.8.0"

henryiii:
Reverts pybind/pybind11#4220

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -324,8 +324,8 @@ jobs:
   # Testing NVCC; forces sources to behave like .cu files
   cuda:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


rwgk:
Trying a forward fix: e9070b27e5ff81c365f5673db8bb323b52b44483

At this second tested only interactively.

Needs to pass #4228 (master)

Needs to pass #4227 (smart_holder)

rwgk:
The commit mentioned in the previous comment had a flaw due to a misunderstanding, fixed in 7557684964e52c05b9d1fc633d4c383462116f11. Retesting.

henryiii:
Okay, will go with that one, then.

PR #4232, opened 2022-10-10
chore: Improve PyCapsule exception handling

Skylion007:
## Description
This further improves another error in exception handling pointed out by #4221 . 
It makes the following changes
* Only check if PyErrOccurred() only if the destructor is nullptr (the exceptions are only set in if the destructor is nullptr according the spec). This should be a bit faster and be more robust as it better follows the spec.
* ~There is a pretty rare edge case that could occur when wrapping a capsule around a function ptr if the destructor was somehow changed to nullptr. I couldn't actually get this one to fire at all, however, it's better to have proper handling in case this corner case is every encountered since it's a valid return state according the CPython API.~ I checked the code and it is actually impossible to get to have GetPointer return a nullptr and NOT set up an exception
* Change a cstyle cast to an explicit C++ reinterpret_cast for readability.

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1834,7 +1834,7 @@ public:
             }
             const char *name = get_name_in_error_scope(o);
             void *ptr = PyCapsule_GetPointer(o, name);
-            if (ptr == nullptr && PyErr_Occurred()) {
+            if (ptr == nullptr) {
                 throw error_already_set();
             }
 
@@ -1852,13 +1852,10 @@ public:
         m_ptr = PyCapsule_New(reinterpret_cast<void *>(destructor), nullptr, [](PyObject *o) {
             const char *name = get_name_in_error_scope(o);
             auto destructor = reinterpret_cast<void (*)()>(PyCapsule_GetPointer(o, name));
-            if (destructor != nullptr) {
-                destructor();
-            } else {
-                if (PyErr_Occurred()) {
-                    throw error_already_set();
-                }
+            if (destructor == nullptr) {
+                throw error_already_set();
             }
+            destructor();
         });
 
         if (!m_ptr) {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1829,7 +1829,7 @@ public:
             // guard if destructor called while err indicator is set
             error_scope error_guard;
             auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
-            if (PyErr_Occurred()) {
+            if (destructor == nullptr && PyErr_Occurred()) {
                 throw error_already_set();
             }
             const char *name = get_name_in_error_scope(o);
@@ -1843,7 +1843,7 @@ public:
             }
         });
 
-        if (!m_ptr || PyCapsule_SetContext(m_ptr, (void *) destructor) != 0) {
+        if (!m_ptr || PyCapsule_SetContext(m_ptr, reinterpret_cast<void *>(destructor)) != 0) {
             throw error_already_set();
         }
     }
@@ -1852,10 +1852,13 @@ public:
         m_ptr = PyCapsule_New(reinterpret_cast<void *>(destructor), nullptr, [](PyObject *o) {
             const char *name = get_name_in_error_scope(o);
             auto destructor = reinterpret_cast<void (*)()>(PyCapsule_GetPointer(o, name));
-            if (destructor == nullptr) {
-                throw error_already_set();
+            if (destructor != nullptr) {
+                destructor();
+            } else {
+                if (PyErr_Occurred()) {
+                    throw error_already_set();
+                }
             }
-            destructor();
         });
 
         if (!m_ptr) {
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1834,7 +1834,7 @@ public:
             }
             const char *name = get_name_in_error_scope(o);
             void *ptr = PyCapsule_GetPointer(o, name);
-            if (ptr == nullptr) {
+            if (ptr == nullptr && PyErr_Occurred()) {
                 throw error_already_set();
             }
 


include/pybind11/pytypes.h
```diff
@@ -1829,12 +1829,12 @@ class capsule : public object {
...
             error_scope error_guard;
             auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
-            if (PyErr_Occurred()) {
+            if (destructor == nullptr && PyErr_Occurred()) {
                 throw error_already_set();
             }
             const char *name = get_name_in_error_scope(o);
             void *ptr = PyCapsule_GetPointer(o, name);
-            if (ptr == nullptr) {
+            if (ptr == nullptr && PyErr_Occurred()) {
```
rwgk: I looked at this recently in connection with something completely different:

https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_New

> Create a [PyCapsule](https://docs.python.org/3/c-api/capsule.html#c.PyCapsule) encapsulating the pointer. The pointer argument may not be NULL.

I found it impossible to create a capsule with a NULL pointer.

Therefore I think it's better to not change this line. If no error is set, something is very wrong, and it'll be better to let `error_already_set()` tell us:

https://github.com/pybind/pybind11/blob/da104a9efd0357b5c144c67eb641ae0b9e23012d/include/pybind11/pytypes.h#L451

Skylion007: @rwgk Ah, you are right. I just updated this PR.

Skylion007:
Okay, now this PR mainly just avoids a call to `PyErrOccurred()` if possible.

include/pybind11/pytypes.h
```diff
@@ -1829,7 +1829,7 @@ class capsule : public object {
             // guard if destructor called while err indicator is set
             error_scope error_guard;
             auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
-            if (PyErr_Occurred()) {
+            if (destructor == nullptr && PyErr_Occurred()) {
```
Skylion007, line 1832: Main thing here is PyErr_Occurred() is a much more expensive call, especially on the common path so might as well avoid doing so if possible

rwgk: Hm ... but what if the error indicator is set (for whatever reason)? We're leaving the process in an invalid state, follow-on Python C API calls may behave in confusing ways (painful debugging memories come back).

If the Python documentation guarantees that the error indicator is not set unless `destructor == nullptr`, this change is fine.

Looking: https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_GetContext
________ 
Return the current context stored in the capsule. On failure, set an exception and return NULL.

It is legal for a capsule to have a NULL context. This makes a NULL return code somewhat ambiguous; use [PyCapsule_IsValid()](https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_IsValid) or [PyErr_Occurred()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Occurred) to disambiguate.
________

To me it sounds like:

* If the error indicator is set the return value is certainly nullptr.
* But knowing that the return value is nullptr does not tell us anything about the state of the error indicator.

So checking the error indicator is mandatory, unless we want to risk leaving the process in an invalid state, which I don't think is a responsible thing to do, even it this was more than a minor optimization.

include/pybind11/pytypes.h
```diff
@@ -1829,7 +1829,7 @@ class capsule : public object {
             // guard if destructor called while err indicator is set
             error_scope error_guard;
             auto destructor = reinterpret_cast<void (*)(void *)>(PyCapsule_GetContext(o));
```
rwgk, line 1831: This line was introduced with PR #752:

https://github.com/pybind/pybind11/pull/752/files

@wjakob Why is this `PyCapsule_GetContext()` and not `PyCapsule_GetDestructor()`? 

I see it's self-consistent within pybind11, `PyCapsule_SetContext()` is used here:

https://github.com/pybind/pybind11/blob/da104a9efd0357b5c144c67eb641ae0b9e23012d/include/pybind11/pytypes.h#L1846

But I'd expect that the rest of the world handles destructors via `PyCapsule_SetDestructor()` & `PyCapsule_GetDestructor()`. Could this lead to leaks, because we're never calling PyCapsule_GetDestructor() anywhere in pybind11?

I just looked, even Python 2.7 had `PyCapsule_SetDestructor()` & `PyCapsule_GetDestructor()` already.

include/pybind11/pytypes.h
Skylion007: @rwgk The documentation and the [reference implementation](https://github.com/python/cpython/blob/f4095e53ab708d95e019c909d5928502775ba68f/Objects/capsule.c#L122) seem to guarantee that the error indicator is not set unless it returns a nullptr. 

A better way to state this is the following. Is it ever valid to set the ErrorIndicator and return a non-nullptr from this method? That is there any case where non-null and the error indicator is set (for the return to this function. Ignoring it could be set to an err due leakage from some other part of the code.

The answer is no, because if it returned any non-null ptr, then an error did not occur. The error only occurs if the capsule is [PyCapsule_IsValid()](https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_IsValid) returns false internally, it therefore unable to access the members of the pycapsule, and then return null. Therefore, this function only sets the error indicator IFF it returns a nullptr. The error guard guarantees that we the error indicator is not set. 

Therefore, destructor being non-null strictly implies that no error has occurred.

rwgk: I agree, sorry for getting this wrong. I didn't look at the implementation yesterday, from that it's totally clear.

rwgk (approved):

include/pybind11/pytypes.h
rwgk: It turns out my worries here were unfounded. Based on feedback from @wjakob, I came to realize that my core misunderstanding was that `py::capsule` is controlling the lifetime (in a C++ dtor) but there is no C++ dtor here and the lifetime it is (of course) tied to the Python refcount of the capsule. This is connected to having misunderstood the `nullptr` for the 2nd argument of `PyCapsule_New`, that is the `name`, not the `destructor`.

I.e. everything is fine here. But having looked at this pretty closely now, I decided this minor cleanup will be useful: #4238

PR #4234, opened 2022-10-11
fix: Revert pfect args make iterator

### LINKED ISSUE #4100: [BUG]: error: lvalue required as increment operand
### Required prerequisites

### Problem description

previously compiling code fails with gcc 12.1.0:

```c++
/usr/include/pybind11/pybind11.h:2347:29: error: lvalue required as increment operand
```

relevant code:
https://github.com/pybind/pybind11/blob/1e3400b6742288429f2069aaf5febf92d0662dae/include/pybind11/pybind11.h#L2336-L2360

### Reproducible example code

```text
#include <vector>
#include <pybind11/stl.h>

namespace py = pybind11;
using namespace py::literals;

class Vec {
    public:
        Vec(double x, double y, double z) {
            values[0] = x;
            values[1] = y;
            values[2] = z;
        };
        double values[3];
};

    py::class_<Vec>(m, "Vec")
        .def(py::init<double, double, double>())
        .def("__iter__", [](const Vec& v){
            return py::make_iterator(v.values, v.values+3);
        },  py::keep_alive<0, 1>());
            
}
```

Skylion007:
## Description
Closes #4100 . Apparently, the perfect forwarding here can cause some issues with the wrong template being selected here. One option is to revert the relevant commit that this PR demonstrates, but I am not sure that is the best long term fix as the ambiguous / wrong overload can be easily be selected since the two templates for make_iterator are under-constrained.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2467,7 +2467,7 @@ void implicitly_convertible() {
     };
 
     if (auto *tinfo = detail::get_type_info(typeid(OutputType))) {
-        tinfo->implicit_conversions.push_back(implicit_caster);
+        tinfo->implicit_conversions.emplace_back(std::move(implicit_caster));
     } else {
         pybind11_fail("implicitly_convertible: Unable to find type " + type_id<OutputType>());
     }
--- a/include/pybind11/pybind11.h
+++ b/include/pybind11/pybind11.h
@@ -2333,7 +2333,7 @@ template <typename Access,
           typename Sentinel,
           typename ValueType,
           typename... Extra>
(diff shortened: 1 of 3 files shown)


rwgk (approved):
Thanks for adding the test!

Skylion007:
@henryiii @rwgk  What do you think the best longterm fix for the caster is here? I think there is a pretty semantically ambiguous conversion between the two function. Maybe we should have an enable_if_t if std::begin/std::end is even supported on the value type?

henryiii (approved):

henryiii:
@Skylion007 could you undraft? I think this would be a good interim solution, maybe we can revisit before 2.11/2.12.

PR #4237, opened 2022-10-12
perf: Optimize iterator advance() call

Skylion007:
## Description
Inspired by #4232, I found one other place where we ignore the return value of C-API function and blindly call PyErr_Occurred(). Worse, it's in a very hot piece of code, (the advance() function for iterators in C++ bindings). The official [C-API documentation example code](https://docs.python.org/3/c-api/iter.html) clearly shows that PyErr_Occurred() check is only necessary when the iteration terminates (ie. the next pointer returned is not null). `PyErrOccurred()` involves a good deal of pointer chasing so best to avoid having to so every step: https://github.com/vstinner/pythondev/blob/d7a8d069e5a5d7ad26503f98a2d29c1508a03e73/assembly.rst

## Suggested changelog entry:

```rst
* perf: optimize iterator advancement in C++ bindings.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1382,7 +1382,7 @@ public:
 private:
     void advance() {
         value = reinterpret_steal<object>(PyIter_Next(m_ptr));
-        if (PyErr_Occurred()) {
+        if (value.ptr() == nullptr && PyErr_Occurred()) {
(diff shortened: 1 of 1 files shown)


rwgk (approved):
The Python documentation for this is indeed super clear.

PR #4238, opened 2022-10-12
Minor `py::capsule` cleanup. No functional change.

rwgk:
Use `PyCapsule_Destructor` (part of the stable Python ABI) instead of spelling out the C `typedef`.

The deprecation message is misleading. Replace with a message pointing to another existing ctor.

Background: According to @wjakob the original motivation for deprecating the ctor (in PR #752) was to hide Python C API details, but PR #902 brought those back with a new ctor, it cannot be avoided. Having a `PyCapsule_Destructor` or a `void (*destructor)(void *)` are two separate and valid use cases.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -1809,16 +1809,16 @@ public:
 
     explicit capsule(const void *value,
                      const char *name = nullptr,
-                     void (*destructor)(PyObject *) = nullptr)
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

rwgk:
Ignoring the CentOS7 / PGI 22.3 failure (also seen in another PR; it seems to be generally broken at the moment).

Issue #4137, opened 2022-08-11
[BUG]: free(): invalid pointer


whybeyoung:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

Like issue https://github.com/pybind/pybind11/issues/1472,
we still have problem in 2.10.0 

free(): invalid pointer

c++ code: 
```cpp
#include "pybind11/embed.h"
#include <iostream>
#include <thread>
#include <chrono>
#include <sstream>

namespace py = pybind11;
using namespace std::chrono_literals;

class Wrapper
{
public:
  Wrapper()
  {
    py::gil_scoped_acquire acquire;
    _obj = py::module::import("wrapper").attr("Wrapper")();
    _wrapperInit = _obj.attr("wrapperInit");
    _wrapperFini = _obj.attr("wrapperFini");

  }

  ~Wrapper()
  {
    _wrapperInit.release();
    _wrapperFini.release();
  }

  int wrapperInit()
  {
    py::gil_scoped_acquire acquire;
    return _wrapperInit(nullptr).cast<int>();
  }

  void wrapperFini(int x)
  {
    py::gil_scoped_acquire acquire;
    _wrapperFini(x);
  }

  private:
  py::object _obj;
  py::object _wrapperInit;
  py::object _wrapperFini;
};
void thread_func(int iteration)
{
  Wrapper w;

  for (int i = 0; i < 1; i++)
  {
    w.wrapperInit();
    std::stringstream msg;
    msg << "iteration: " << iteration << " thread: " << std::this_thread::get_id() << std::endl;
    std::cout << msg.str();
        std::this_thread::sleep_for(100ms);
  }
}

int main() {
  py::scoped_interpreter guard{};
  py::gil_scoped_release release; // add this to release the GIL

  std::vector<std::thread> threads;

  for (int i = 0; i < 1; ++i)
    threads.push_back(std::thread(thread_func, 1));

  for (auto& t : threads)
    t.join();

  return 0;
}
```
wrapper.py code is 

```python

class Wrapper():
[...]

PR #4246, opened 2022-10-17
`pybind11::handle` `inc_ref()` & `dec_ref()` `PyGILState_Check()` **excluding** `nullptr`

rwgk:
## Description
Background: https://docs.python.org/3/glossary.html#term-global-interpreter-lock

The GIL must be held when calling any Python C API functions. In multithreaded applications that use callbacks this requirement can easily be violated by accident. A general tool to ensure GIL health is not available, but the this PR integrates a very simple basic health check into `pybind11::handle`.

A comparison of test runtimes with and without this PR showed that **the runtime overhead for the added checks is practically undetectable for non-optimized builds**, but Google global testing with this PR uncovered a number (10+) of often very subtle GIL-not-held issues, including bugs in a few 3rd party projects (e.g. [pikepdf](https://github.com/pikepdf/pikepdf/commit/9fdb4740e0d94fc80fb91393e0d810736ec86258)).

The added GIL checks are guarded by `PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF`, which is the default only if `NDEBUG` is *not* defined:

```diff
+#if !defined(NDEBUG) && !defined(PY_ASSERT_GIL_HELD_INCREF_DECREF)                                \
+    && !defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
+#    define PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF
+#endif
```

The test for the `PY_ASSERT_GIL_HELD_INCREF_DECREF` macro is for cooperation with a more general patch in the Python sources, inserting the GIL checks directly in `Py_INCREF` & `Py_DECREF`, which makes the checks in pybind11 redundant.

Note that the GIL checks are skipped if `py::handle` `ptr() == nullptr`. This could mask bugs, in particular for classes with `py::object` members for which `ptr() == nullptr` in all unit tests, but not in production. Unfortunately Google global testing revealed that such members are much more common than one might expect. It is currently unclear if the cleanup cost for enforcing that the GIL is held even for `nullptr` is commensurate with the benefits of preventing a potentially only very small number of production bugs.

## Suggested changelog entry:

```rst
`PyGILState_Check()`s were integrated to `pybind11::handle` `inc_ref()` & `dec_ref()`. The added GIL checks are guarded by `PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF`, which is the default only if `NDEBUG` is not defined
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -247,6 +247,11 @@ public:
 #ifdef PYBIND11_HANDLE_REF_DEBUG
         inc_ref_counter(1);
 #endif
+#if defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
+        if (m_ptr != nullptr && !PyGILState_Check()) {
+            throw std::runtime_error("pybind11::handle::inc_ref() PyGILState_Check() failure.");
+        }
+#endif
         Py_XINCREF(m_ptr);
         return *this;
     }
@@ -257,6 +262,11 @@ public:
         this function automatically. Returns a reference to itself.
     \endrst */
     const handle &dec_ref() const & {
+#if defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
+        if (m_ptr != nullptr && !PyGILState_Check()) {
+            throw std::runtime_error("pybind11::handle::dec_ref() PyGILState_Check() failure.");
+        }
+#endif
         Py_XDECREF(m_ptr);
         return *this;
     }
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -262,7 +262,9 @@
 #    define PYBIND11_HAS_U8STRING
 #endif
 
+// See description of PR #4246:
 #if !defined(NDEBUG) && !defined(PY_ASSERT_GIL_HELD_INCREF_DECREF)                                \
+    && !(defined(PYPY_VERSION) && defined(_MSC_VER)) /* Tests hang indefinitely at startup. */    \
     && !defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
 #    define PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF
 #endif
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -262,6 +262,11 @@
 #    define PYBIND11_HAS_U8STRING
 #endif
 
+#if !defined(NDEBUG) && !defined(PY_ASSERT_GIL_HELD_INCREF_DECREF)                                \
+    && !defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
+#    define PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF
+#endif
+
 // #define PYBIND11_STR_LEGACY_PERMISSIVE
 // If DEFINED, pybind11::str can hold PyUnicodeObject or PyBytesObject
 //             (probably surprising and never documented, but this was the
--- a/include/pybind11/detail/common.h
+++ b/include/pybind11/detail/common.h
@@ -264,7 +264,9 @@
 
 // See description of PR #4246:
 #if !defined(NDEBUG) && !defined(PY_ASSERT_GIL_HELD_INCREF_DECREF)                                \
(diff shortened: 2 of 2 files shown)


rwgk:
Hi @mattip, in this PR I had to disable the new feature for Windows PyPy. This isn't a problem for my purposes, but I'm curious why Windows PyPy is hanging with the new code? Note that macOS and Linux PyPy are fine.

mattip:
> why Windows PyPy is hanging with the new code

Is there a way to boil the test down to a minimal reproducer and open a PyPy issue?

rwgk:
> > why Windows PyPy is hanging with the new code
> 
> Is there a way to boil the test down to a minimal reproducer and open a PyPy issue?

Ooo, that's hard, I'd have to reduce via the GitHub CI. I don't have access to a suitable Windows machine. And not really any clue where the problem could be, beyond observing that pytest didn't even start up.

rwgk:
FYI — This PR was deployed to Google's production systems today (after a relatively small internal cleanup). I'll keep this in draft mode for a week or so, JIC there is some pushback.

include/pybind11/pytypes.h
```diff
@@ -246,6 +246,11 @@ class handle : public detail::object_api<handle> {
     const handle &inc_ref() const & {
 #ifdef PYBIND11_HANDLE_REF_DEBUG
         inc_ref_counter(1);
+#endif
+#if defined(PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF)
+        if (m_ptr != nullptr && !PyGILState_Check()) {
+            throw std::runtime_error("pybind11::handle::inc_ref() PyGILState_Check() failure.");
```
Skylion007, line 252: Is there a reason no pybind11_fail? Because of the corresponding `!PyErrOccurred() ` assertion?

rwgk: Yes.
But even if that was taken out in `pybind11_fail()`, I think a "raw" C++ exception is best here, we want the process to terminate as directly as possible. I was thinking of using `std::terminate()` directly, but then it is troublesome to also produce the message, so I decided simpler is better. In our environment at least, `std::runtime_error()` works out very nicely (useful traceback with message).

rwgk:
> It seems the CI failed on [this test](https://github.com/pybind/pybind11/actions/runs/3267538328/jobs/5372851326#step:12:78)
> 
> ```
> assert _run_in_process(_python_to_cpp_to_python_from_threads, 1) == 0
> ```
> 
> The code for that test is [here](https://github.com/pybind/pybind11/blob/6493f496e30c80f004772c906370c8f4db94b6ec/tests/test_gil_scoped.py#L7).

Oh, thanks for looking and pointing this out! I was misled by what I saw in the web view of the logs yesterday before I cancelled the jobs. There was just a spinning wheel for a long time without any output.

I'll try to get back here as soon as I get a chance.

rwgk:
> FYI — This PR was deployed to Google's production systems today (after a relatively small internal cleanup). I'll keep this in draft mode for a week or so, JIC there is some pushback.

15 days later: I only got two questions re problems that turned out to be completely unrelated, and one report of a bug uncovered by this PR and then fixed.

Also note that this PR would have immediately flagged the root cause of #4137, quite likely that issue would not even have been filed if we had made this change earlier.

Skylion007 (approved):

henryiii (approved):
Given you've checked runtime, and this helps with safety, I'm for it.

include/pybind11/pytypes.h
wjakob: FYI, functions that can raise exceptions do come with some subtle costs in C++. The compiler needs to keep track of the many ways in which might have to destroy objects if an exception arises at various different points of the program and generate code for each one. I don't think it's a performance issue, but this will likely make the debug binaries bigger (compared to `pybind11_fail`) given that such a central operation is affected.

rwgk: Narrow focus:

Currently `pybind11_fail()` includes this:
```
assert(!PyErr_Occurred());
```
That needs the GIL, but here (`handle`) we just determined that we don't have it.

Bigger picture:

There is a general tradeoff: a faster or leaner binary today vs a future with more features.

Boundary condition: human time is ~constant.

So ultimately we always trade one for the other.

If you hire person A who likes to optimize, you may get something that runs in 1.0 time units with 1.0 resources, with 10 features.

If you hire person B, who's more like me, you get something that runs in 1.1 time units with 1.1 resources, with 20 features.

What's "better"?

It depends.

PR #4247, opened 2022-10-17
Allow specifying custom base classes

virtuald:
## Description

In the GObject* (and others, such as QEMU's qdev) type system, C-compatible 'inheritance' is implemented by embedding the parent struct in the child struct:

```
struct Child {
    Parent object;
    ...
};
```

One can already implement a [polymorphic_type_hook](https://pybind11.readthedocs.io/en/stable/advanced/classes.html?highlight=llvm#custom-automatic-downcasters) to go from `Parent` to `Child` objects, but currently you can't do:

```
py::class_<Child, Parent> c(m, "Child");
```

This allows specification of arbitrary base classes for a bound object, which allows this use case (and others as well).

## Suggested changelog entry:

```rst
Base classes that don't follow normal C++ inheritance can now be specified using py::custom_base<T>
```

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -82,13 +82,13 @@ struct base {
     .. note:: This is an advanced feature. If you use this, you likely need
               to implement polymorphic_type_hook for your type hierarchy.
  \endrst */
-template <typename T> struct custom_base {
-    using caster = void *(*)(void *);
+template <typename T>
+struct custom_base {
+    using caster = void *(*) (void *);
 
     explicit custom_base(const caster &f) : fn(f) {}
     caster fn;
 };
-
 
 /// Keep patient alive while nurse lives
 template <size_t Nurse, size_t Patient>
@@ -579,8 +579,7 @@ struct process_attribute<base<T>> : process_attribute_default<base<T>> {
 /// Process a custom base attribute
 template <typename T>
 struct process_attribute<custom_base<T>> : process_attribute_default<custom_base<T>> {
-    static void init(const custom_base<T> &b, type_record *r)
-    {
+    static void init(const custom_base<T> &b, type_record *r) {
         r->add_base(typeid(T), b.fn);
         // TODO: rename this to 'nonsimple'?
         r->multiple_inheritance = true;
--- a/tests/test_custom_base.cpp
+++ b/tests/test_custom_base.cpp
@@ -28,7 +28,8 @@ TEST_SUBMODULE(custom_base, m) {
 
     py::class_<Derived>(m, "Derived", py::custom_base<Base>([](void *o) -> void * {
                             return &reinterpret_cast<Derived *>(o)->base;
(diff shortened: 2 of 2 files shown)


include/pybind11/attr.h
```diff
@@ -551,6 +576,16 @@ struct process_attribute<base<T>> : process_attribute_default<base<T>> {
     static void init(const base<T> &, type_record *r) { r->add_base(typeid(T), nullptr); }
 };
 
+/// Process a custom base attribute
+template <typename T>
+struct process_attribute<custom_base<T>> : process_attribute_default<custom_base<T>> {
+    static void init(const custom_base<T> &b, type_record *r) {
+        r->add_base(typeid(T), b.fn);
+        // TODO: rename this to 'nonsimple'?
+        r->multiple_inheritance = true;
```
Skylion007, line 585: You really mean simple_ancestors=False, right?

virtuald: Can't set that here, but that's what setting multiple_inheritance causes. See the TODO comment. :)

rwgk:
First impression: amazing 1. how small the core changes are and 2. CI full success on first try.

I'm wondering about limitations and pitfalls, without having anything specific in mind, due to lack of familiarity with the internals you're touching. Can you think of any?

Some slightly random questions:

* Would this work with more than one custom base?

* Copying my question from #4241 here, do you already know if this could work?

```
py::class_<MyException, py::custom_base<PyExc_Exception>(...)>
```

* Do you have an example that needs the `polymorphic_type_hook`?

virtuald:
I expect that as I play with it I'll run into some pitfalls, but I have some other priorities to attend to before doing that. I suspect trampolines might not work, and if someone really wanted to bind GObject you'd need a custom metaclass.

> Would this work with more than one custom base?

Definitely expect it to, since that's what the multiple-inheritance stuff is for.

> py::class_<MyException, py::custom_base<PyExc_Exception>(...)>

Maybe. The limitation is that in `py::custom_base<T>`, I believe that `T` needs to be something that pybind11 has previously bound using `py::class_`, or at least py::cast<T> needs to work? Though, honestly I'm not sure that's true either.

> Do you have an example that needs the polymorphic_type_hook?

If you had a function that was of the form `Base * get_base(Derived *)`, if I bind `get_base` and call it from python, I believe without a polymorphic type hook I believe that `isinstance(get_base(d), Derived) == False`. Haven't tried it yet either.

virtuald:
This weekend I played a bit with this concept. There's definitely a bit of work one needs to do to fully flesh this out. In particular, to support inheritance I needed to add an `__init_subclass__` implementation and an override for `__init__` that did construction in a special way. It's still not fully working, but once that's done I should be able to put together a full example and some API extensions we can add here. Might be awhile though.

Skylion007:
@virtuald Glad to find someone found a usecase for #4158, happy to merge that if it unblocks this.

Issue #2810, opened 2021-01-19
[FEAT] [RFC] Class builder API for high performance bindings


lqf96:
## Motivation
Pybind11 provides easy-to-use APIs for writing Python bindings. However, these bindings are often implemented with multiple level of indirections, which can have negative impact on the performance. For example, `class_::def_property` and `class_::def_property_readonly` are currently implemented as:

* A `PyCFunction` wrapper capturing a capsule holding a `detail::function_record`
* ... wrapped in an `instancemethod`
* ... wrapped in a `property` stored in the class dictionary.

Compared to a `getset_descriptor` directly implemented with `PyGetSetDef`, this can be [5x-20x slower](https://github.com/pytorch/pytorch/pull/49988#issuecomment-757678543) both in terms of execution time and instruction counts. Similarly, constructors implemented in pybind11 can be slower than a low-level constructor function directly bound to type with the `tp_init` slot. Both problems show the necessity of having a set of high performance binding APIs for pybind11.

## Design
The new API is called "class builder" because it employs the builder pattern:

```c++
pybind11::class_builder<Example>(m, "Example")
    .def("method", [](Example& e) {})
    .def_class("class_method", [](Example& e) {})
    .def_static("static_method", []() { return 1; })
    .def_attr_readonly("readonly_attr", []() { return 1; })
    .build();
```

To create high-performance bindings for a class, we would call `pybind11::class_builder` instead of `pybind11::class_` with almost exactly the same parameters. This gives us a class builder, which we can use to define a series of methods and attributes (note these are `getset_descriptor`s not `property`s). After we define all the members of the class, we call the `build` method, which will create the Python class and return a `pybind11::class_<...>` instance.

Internally, `def_attr` and `def_attr_readonly` is implemented as adding a new entry to the `tp_getset` slot, and `def`, `def_class` and `def_static` is implemented as adding a new entry to the `tp_methods` slot. The `build` method will call `PyType_Ready` method on the type being built, and wraps the ready `PyTypeObject` in a `pybind11::class_<...>` instance.

## Misc
* With the class builder API, we are going to directly set the slots of the type object. Hence, we need to split `cpp_function` into two parts: object wrapper and dispatching logic. For the class builder API, the previous part isn't needed. We'll however keep it for backward compatibility.
[...]

PR #4250, opened 2022-10-18
[Early stage prototype] Optimize calls using vectorcall, 35% reduction in call overhead

EthanSteinberg:
This PR is an early stage prototype of using Python 3.7's vectorcall support to reduce function call overhead in pybind11.

This is very early stage, it doesn't support all of pybind11's features (kwargs don't quite work) and is in really rough shape code / documentation wise.

I am using a simple benchmark, included in this PR under the folder pybench, to measure performance.

Time to call a C++ function 100000000 times:

Raw CPython API: 5.2974734380841255 seconds
Pybind11 master Branch:  17.06675560772419 seconds
Pybind11 this PR:  11.004007823765278 seconds

I believe I can get this down even further (to around 8 seconds or so) by optimizing the loader_life_support.

The main downsides of this approach:

- We are storing temporary data in function_record objects, so we can't support multiple threads calling the same function_record at a time. This shouldn't matter though as that's always the case due to the GIL

- For older versions of Python (<= 3.6), I will have to implement a fallback that will have a tiny drop in performance due to an extra python tuple and python dict allocation and construction.

What are people's thoughts? Should I fully flush out this PR and add full support for the remaining pybind11 functionality? Do people think this is a useful optimization to add to pybind11?

CC @rwgk


Lalaland:
@lqf96 I think you might find this interesting given your prior work on https://github.com/pybind/pybind11/issues/2810

rwgk:
> For older versions of Python (<= 3.6)

3.6 is already ~10 months past EOL (https://endoflife.date/python). This PR is a good motivation to drop it. You could simply remove the corresponding CI jobs in .github/workflows/ci.yml

I only glanced through the code. At first sight it looks like a relatively small change, given what it does. If you want to drive this to completion, I'll look more, run sanitizers & global testing when you get the CI (minus 3.6) healthy here.

Under the smart_holder branch I already introduced a [ubench (micro benchmark) directory](https://github.com/pybind/pybind11/tree/smart_holder/ubench). Could you please use the same name here, too? Actually, I think it would make sense to copy over the code and strip out the very few lines for smart_holder, to be added back in under the smart_holder branch (I could do that part). You could also remove the python/number_bucket.clif file.

FWIW [ubench results from Feb 2021 are here](https://docs.google.com/spreadsheets/d/1InapCYws2Gt-stmFf_Bwl33eOMo3aLE_gc9adveY7RU/edit#gid=1197468011)

include/pybind11/pybind11.h
```diff
@@ -458,7 +459,16 @@ class cpp_function : public function {
 
         rec->signature = guarded_strdup(signature.c_str());
         rec->args.shrink_to_fit();
+        for (size_t i = 0; i < rec->args.size(); i++) {
```
Skylion007, line 462: There is a lowbound of argument_index.size() at least. Argument_index reserve method should be called to prevent many reallocations.

Skylion007:
Lalaland, to try shorten the commit log a bit for this PR, please do the following:
`pip install pre-commit
# cd to git project root
pre-commit install
`
That will run pre-commit locally before every commit.

Lalaland:
Thanks for the feedback. I'll push this PR to completion. I highly recommend both of you turn off notifications for this PR as I'll probably have to spam the CI quite a bit to get this in a working shape.

@Skylion007 I purposely don't run pre-commit to trigger a CI run. I don't have CI permissions, but the precommit bot does so if I can trigger the precommit bot, I get a nice CI run.

Skylion007:
@Lalaland You should now have CI permissions.

henryiii:
3.8 added vectorcall. That’s why nanobind requires 3.8+.

include/pybind11/detail/function_record.h
```diff
@@ -477,7 +477,7 @@ struct args<Return, std::tuple<Args...>, std::tuple<Extra...>> {
                 }
                 // Separator for keyword-only arguments, placed before the kw
                 // arguments start (unless we are already putting an *args)
-                if (!has_args && arg_index == nargs_pos) {
+                if (PYBIND11_SILENCE_MSVC_C4127(!has_args && arg_index == nargs_pos)) {
```
Skylion007, line 480: Only `has_args` should be wrapped.

Lalaland: Will do. As an aside, can we just disable this warning period? It makes metaprogramming a pain ...

include/pybind11/detail/function_record.h
```diff
@@ -0,0 +1,674 @@
...
+
+/// Internal data structure which holds metadata about a keyword argument
+struct argument_record {
+    std::string name;
+    std::string desc;
+    object value;     ///< Associated Python object
+    bool convert : 1; ///< True if the argument is allowed to convert when loading
+    bool none : 1;    ///< True if None is allowed when loading
+
+    argument_record() : name(""), desc(""), value(), convert(true), none(true) {}
```
Skylion007, line 45:
```suggestion
    std::string name = "";
    std::string desc = "";
    object value();     ///< Associated Python object
    bool convert : 1 = true; ///< True if the argument is allowed to convert when loading
    bool none : 1 = true;    ///< True if None is allowed when loading

    argument_record() = default;
```

include/pybind11/attr.h
```diff
@@ -505,97 +442,128 @@ struct process_attribute<arg_v> : process_attribute_default<arg_v> {
...
+    template <typename RecordType>
+    static void init(const is_final &, RecordType &r) {
+        r.is_final = true;
+    }
 };
 
 template <>
 struct process_attribute<buffer_protocol> : process_attribute_default<buffer_protocol> {
-    static void init(const buffer_protocol &, type_record *r) { r->buffer_protocol = true; }
+    template <typename RecordType>
```
Skylion007, line 542: Wow, these are a lot of template types that are just ignored. Is there a more efficient way to do this that doesn't cause a lot of template initialization? Like making the 2nd arg a void pointer or something?

include/pybind11/detail/common.h
```diff
@@ -1178,5 +1237,24 @@ constexpr inline bool silence_msvc_c4127(bool cond) { return cond; }
...
+
+// From https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57350#c11
+inline void *align(std::size_t alignment, std::size_t size, void *&ptr, std::size_t &space) {
+    std::uintptr_t pn = reinterpret_cast<std::uintptr_t>(ptr);
+    std::uintptr_t aligned = (pn + alignment - 1) & -alignment;
+    std::size_t padding = aligned - pn;
+    if (space < size + padding)
+        return nullptr;
+    space -= padding;
+    return ptr = reinterpret_cast<void *>(aligned);
```
Skylion007, line 1250:
```suggestion
    return ptr = static_cast<void *>(aligned);
```

Skylion007 (changes requested):
A lot of nits.

As I mentioned before, a lot of the slowdown in master is also from unnecessary memory allocations (std::vector, std::string, etc...) be wary of that as you build out this PR and take a look at the one I linked.

Also be wary of unused template args / such as we are a header only lib so we do want to be careful about slowdown/bloat from unnecessary template instationation. Of course it's fine if those template args are actually used (like for converting an std::vector into an std::array for n_args.

Also you only need the MSVC warning override around stuff that should be constexpr. Unfortunately, there is no easy way to override it in MSVC as the recommended solution is to use constexpr if etc, but that only works if you are > C++14 of course.

include/pybind11/detail/python_compat.h
```diff
@@ -0,0 +1,99 @@
...
+// Actual vectorcall implementation
+
+constexpr unsigned long pybind_vectorcall_flag = Py_TPFLAGS_HAVE_VECTORCALL;
+
+inline void set_vectorcall_set_offset(PyTypeObject &type, size_t offset) {
+    type.tp_vectorcall_offset = static_cast<Py_ssize_t>(offset);
+}
+using pybind_vectorcallfunc = vectorcallfunc;
+
+constexpr auto pybind_vectorcall_call = PyVectorcall_Call;
```
Skylion007, line 24:
```suggestion
using constexpr auto pybind_vectorcall_call = PyVectorcall_Call;
```

include/pybind11/detail/function_record.h
Lalaland: It's probably also good to disable it to help the optimizer. We want the compiler to know these conditions are constant so they can removed from the resulting executable.

jblespiau:
(FYI, if it's useful, I also extracted and polished https://github.com/jblespiau/pybind11_benchmarks from the nanobind repository, in hope to bring within pybind11 a more extensive benchmarking suite.)

include/pybind11/detail/common.h
```diff
@@ -1178,5 +1237,24 @@ constexpr inline bool silence_msvc_c4127(bool cond) { return cond; }
...
+}
+
+#else
+
+constexpr auto align = std::align;
```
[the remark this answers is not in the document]
Lalaland: This doesn't seem to work? Are you sure this is the correct syntax?

include/pybind11/detail/function_record.h
Lalaland: The bitfields mean that this doesn't work.

include/pybind11/detail/function_record.h
```diff
@@ -0,0 +1,674 @@
...
+            /* Override policy for rvalues -- usually to enforce rvp::move on an rvalue */
+            return_value_policy final_policy
+                = return_value_policy_override<Return>::policy(metadata.policy);
+
+            /* Function current_scope guard -- defaults to the compile-to-nothing `void_type` */
+            using Guard = extract_guard_t<Extra...>;
+
+            std::array<handle, sizeof...(Args)> call_args;
+
+            typename metadata_type::cast_in loader;
```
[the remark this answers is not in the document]
Lalaland: I am creating an instance, here, not creating an alias.

include/pybind11/detail/common.h
Lalaland: This isn't support on centos7 for some reason.

Lalaland:
Looks like this is very close to v1 being ready. Passes all the tests on regular Python and is quite fast to boot.

The main issue is that I need to do a lot more space optimization

Using @jblespiau's test harness

Time:
python: 0.18168139830231667
nanobind: 0.1628
old pybind11: 0.3840818479657173
this PR: 0.22408869490027428

The problem is space
old pybind11: 1.5625
this PR: 6.1168060302734375

rwgk:
> old pybind11: 1.5625
> this PR: 6.1168060302734375

What exactly is 4 times bigger?

Sorry I don't have a lot of experience with micro benchmarks and binary size comparison.

Lalaland:
> What exactly is 4 times bigger?

The .so files generated when the extensions are complied.

Fixing this should be quite easy. I just need to create non inlined helper functions in several places.

rwgk:
> > What exactly is 4 times bigger?
> 
> The .so files generated when the extensions are complied.

Wow ... I don't think I've ever seen such an explosion of binary sizes.

> I just need to create non inlined helper functions in several places.

Hm ... I did measurements back in Aug 2021:

https://docs.google.com/spreadsheets/d/1bkg1HtmU1AjotPTXCCLSbTBGARRet8W91noh1DnKOHU/edit#gid=0

noinline hardly made a difference, but maybe your situation is somehow special?

Lalaland:
> noinline hardly made a difference, but maybe your situation is somehow special?

The situation here is special because of the templates involved. So pulling an even small amount of code out of the template can have huge savings because of the hundreds of copies.

> Wow ... I don't think I've ever seen such an explosion of binary sizes.

Keep in mind that these microbenchmarks are sorta contrived. In order to make the sizes / compile times more significant the same code is copied 719 times.

Skylion007:
> > noinline hardly made a difference, but maybe your situation is somehow special?
> 
> The situation here is special because of the templates involved. So pulling an even small amount of code out of the template (and forcing it out of the template with non-inline) can have huge savings because of the hundreds of copies.
> 
> > Wow ... I don't think I've ever seen such an explosion of binary sizes.
> 
> Keep in mind that these microbenchmarks are sorta contrived. In order to make the sizes / compile times more significant the same code is copied 719 times.

This was also the case in the PR I linked here: https://github.com/pybind/pybind11/pull/2013/ they were eventually able to resolve it.

include/pybind11/detail/common.h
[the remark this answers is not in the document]
Skylion007: Oh sorry, yeah that's the wrong syntax.
```suggestion
using std::align;
```
See https://github.com/pybind/pybind11/blob/09db6445d8da6e918c2d2be3aa4e7b0ddd8077c7/include/pybind11/detail/common.h#L637 for example

henryiii:
Python 3.6 now dropped! Planning on dropping 3.7 immediately after the next release, too. That should make this much easier.

PR #4254, opened 2022-10-19
Fix functional.h bug + introduce test to verify that it is fixed

Lalaland:
## Description

Functional.h is currently buggy when working with manually defined CPython functions. You get segfaults whenever you try to convert a manually defined CPython function to a std::function.

This PR adds a test that illustrates the bug as well as some #ifdef guards around the broken code in functional.h

https://github.com/pybind/pybind11/pull/4254/files#diff-89e29c03817f899ff72bdb90772cb781df9c82044ae712fb68487be02dd7e3cdL49-R72 in particular is the cause of the bug.

As that code path is just an optimization, it can be removed without breaking other pieces of code. However, it will be a significant performance regression.

## Suggested changelog entry:

```rst
* Fix segfault bug when passing foreign native functions to functional.h
```

### DIFF
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -189,7 +189,7 @@ struct argument_record {
 struct function_record {
 
     // Update the following whenever this struct is changed
-    static constexpr const char* function_capsule_name = "pybind11_function_capsule_v1";
+    static constexpr const char *function_capsule_name = "pybind11_function_capsule_v1";
 
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -185,7 +185,12 @@ struct argument_record {
 
 /// Internal data structure which holds metadata about a bound function (signature, overloads,
 /// etc.)
+// Note that function_capsule_name must be updated whenever this struct is changed
 struct function_record {
+
+    // Update the following whenever this struct is changed
+    static constexpr const char* function_capsule_name = "pybind11_function_capsule_v1";
+
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
           is_operator(false), is_method(false), has_args(false), has_kwargs(false),
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -185,11 +185,7 @@ struct argument_record {
 
 /// Internal data structure which holds metadata about a bound function (signature, overloads,
 /// etc.)
-// Note that function_capsule_name must be updated whenever this struct is changed
 struct function_record {
-
-    // Update the following whenever this struct is changed
-    static constexpr const char *function_capsule_name = "pybind11_function_capsule_v1";
 
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
--- a/include/pybind11/attr.h
+++ b/include/pybind11/attr.h
@@ -186,7 +186,6 @@ struct argument_record {
 /// Internal data structure which holds metadata about a bound function (signature, overloads,
 /// etc.)
 struct function_record {
-
     function_record()
         : is_constructor(false), is_new_style_constructor(false), is_stateless(false),
           is_operator(false), is_method(false), has_args(false), has_kwargs(false),
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -187,7 +187,7 @@ struct internals {
     PyInterpreterState *istate = nullptr;
 
 #    if PYBIND11_INTERNALS_VERSION > 4
-    // Note that we have to invoke strdup to allocate memory to ensure a unique address
+    // Note that we have to use a std::string to allocate memory to ensure a unique address
     // We want unique addresses since we use pointer equality to compare function records
     std::string function_record_capsule_name = internals_function_record_capsule_name;
 #    endif
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -34,7 +34,7 @@
 /// further ABI-incompatible changes may be made before the ABI is officially
 /// changed to the new version.
 #ifndef PYBIND11_INTERNALS_VERSION
-#    define PYBIND11_INTERNALS_VERSION 4
+#    define PYBIND11_INTERNALS_VERSION 5
 #endif
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
@@ -491,7 +491,7 @@ PYBIND11_NOINLINE internals &get_internals() {
 struct local_internals {
     type_map<type_info *> registered_types_cpp;
     std::forward_list<ExceptionTranslator> registered_exception_translators;
-#if defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
+#if defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION >= 4
 
     // For ABI compatibility, we can't store the loader_life_support TLS key in
     // the `internals` struct directly.  Instead, we store it in `shared_data` and
@@ -523,7 +523,10 @@ struct local_internals {
         loader_life_support_tls_key
             = static_cast<shared_loader_life_support_data *>(ptr)->loader_life_support_tls_key;
     }
-#endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
+#endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION >= 4
+#if PYBIND11_INTERNALS_VERSION == 5
+    const char* function_capsule_name = strdup("pybind11_function_capsule");
+#endif
 };
 
 /// Works like `get_internals`, but for things which are locally registered.
@@ -535,6 +538,14 @@ inline local_internals &get_local_internals() {
     // https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables
     static auto *locals = new local_internals();
     return *locals;
+}
+
+inline const char* get_function_capsule_name() {
+#if PYBIND11_INTERNALS_VERSION == 5
+    return get_local_internals().function_capsule_name;
+#else
+    return nullptr;
+#endif
 }
 
 /// Constructs a std::string with the given arguments, stores it in `internals`, and returns its
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -34,7 +34,11 @@
 /// further ABI-incompatible changes may be made before the ABI is officially
 /// changed to the new version.
 #ifndef PYBIND11_INTERNALS_VERSION
-#    define PYBIND11_INTERNALS_VERSION 5
+#    define PYBIND11_INTERNALS_VERSION 4
+#endif
+
+#ifndef PYBIND11_LOCAL_INTERNALS_VERSION
+#    define PYBIND11_LOCAL_INTERNALS_VERSION 1
 #endif
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
@@ -491,7 +495,7 @@ PYBIND11_NOINLINE internals &get_internals() {
 struct local_internals {
     type_map<type_info *> registered_types_cpp;
     std::forward_list<ExceptionTranslator> registered_exception_translators;
-#if defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION >= 4
+#if defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
 
     // For ABI compatibility, we can't store the loader_life_support TLS key in
     // the `internals` struct directly.  Instead, we store it in `shared_data` and
@@ -523,9 +527,10 @@ struct local_internals {
         loader_life_support_tls_key
             = static_cast<shared_loader_life_support_data *>(ptr)->loader_life_support_tls_key;
     }
-#endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION >= 4
-#if PYBIND11_INTERNALS_VERSION == 5
-    const char* function_capsule_name = strdup("pybind11_function_capsule");
+#endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
+
+#if PYBIND11_LOCAL_INTERNALS_VERSION >= 1
+    const char *function_capsule_name = strdup("pybind11_function_capsule");
 #endif
 };
 
@@ -540,8 +545,8 @@ inline local_internals &get_local_internals() {
     return *locals;
 }
 
-inline const char* get_function_capsule_name() {
-#if PYBIND11_INTERNALS_VERSION == 5
+inline const char *get_function_capsule_name() {
+#if PYBIND11_LOCAL_INTERNALS_VERSION >= 1
     return get_local_internals().function_capsule_name;
 #else
     return nullptr;
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -43,8 +43,7 @@ using ExceptionTranslator = void (*)(std::exception_ptr);
 
 PYBIND11_NAMESPACE_BEGIN(detail)
 
-static constexpr const char *internals_function_record_capsule_name
-    = "pybind11_function_record_capsule";
+constexpr const char *internals_function_record_capsule_name = "pybind11_function_record_capsule";
 
 // Forward declarations
 inline PyTypeObject *make_static_property_type();
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -197,6 +197,11 @@ struct internals {
         PYBIND11_TLS_FREE(tstate);
     }
 #endif
+#if PYBIND11_INTERNALS_VERSION >= 5
+    // Note that we have to invoke strdup to allocate memory to ensure a unique address
+    // We want unique addresses since we use pointer equality to compare function records
+    const char *function_record_capsule_name = strdup("pybind11_function_record_capsule");
+#endif
 };
 
 /// Additional type information which does not fit into the PyTypeObject.
@@ -546,6 +551,25 @@ const char *c_str(Args &&...args) {
     auto &strings = get_internals().static_strings;
     strings.emplace_front(std::forward<Args>(args)...);
     return strings.front().c_str();
+}
+
+inline const char *get_function_record_capsule_name() {
+#if PYBIND11_INTERNALS_VERSION >= 5
+    return get_internals().function_record_capsule_name;
+#else
+    return nullptr;
+#endif
+}
+
+// Determine whether or not the following capsule contains a pybind11 function record.
+// Note that we use internals to make sure that only ABI compatible records are touched.
+//
+// This check is currently used in two places:
+// - An important optimization in functional.h to avoid overhead in C++ -> Python -> C++
+// - The sibling feature of cpp_function to allow overloads
+inline bool is_function_record_capsule(const capsule &cap) {
+    // Pointer equality as we rely
+    return cap.name() == get_function_record_capsule_name();
 }
 
 PYBIND11_NAMESPACE_END(detail)
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -568,7 +568,7 @@ inline const char *get_function_record_capsule_name() {
 // - An important optimization in functional.h to avoid overhead in C++ -> Python -> C++
 // - The sibling feature of cpp_function to allow overloads
 inline bool is_function_record_capsule(const capsule &cap) {
-    // Pointer equality as we rely
+    // Pointer equality as we rely on internals() to ensure unique pointers
     return cap.name() == get_function_record_capsule_name();
 }
 
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -37,10 +37,6 @@
 #    define PYBIND11_INTERNALS_VERSION 4
 #endif
 
-#ifndef PYBIND11_LOCAL_INTERNALS_VERSION
-#    define PYBIND11_LOCAL_INTERNALS_VERSION 1
-#endif
-
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 
 using ExceptionTranslator = void (*)(std::exception_ptr);
@@ -528,10 +524,6 @@ struct local_internals {
             = static_cast<shared_loader_life_support_data *>(ptr)->loader_life_support_tls_key;
     }
 #endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
-
-#if PYBIND11_LOCAL_INTERNALS_VERSION >= 1
-    const char *function_capsule_name = strdup("pybind11_function_capsule");
-#endif
 };
 
 /// Works like `get_internals`, but for things which are locally registered.
@@ -543,14 +535,6 @@ inline local_internals &get_local_internals() {
     // https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables
     static auto *locals = new local_internals();
     return *locals;
-}
-
-inline const char *get_function_capsule_name() {
-#if PYBIND11_LOCAL_INTERNALS_VERSION >= 1
-    return get_local_internals().function_capsule_name;
-#else
-    return nullptr;
-#endif
 }
 
 /// Constructs a std::string with the given arguments, stores it in `internals`, and returns its
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -43,6 +43,9 @@ using ExceptionTranslator = void (*)(std::exception_ptr);
 
 PYBIND11_NAMESPACE_BEGIN(detail)
 
+static constexpr const char *internals_function_record_capsule_name
+    = "pybind11_function_record_capsule";
+
 // Forward declarations
 inline PyTypeObject *make_static_property_type();
 inline PyTypeObject *make_default_metaclass();
@@ -186,7 +189,7 @@ struct internals {
 #    if PYBIND11_INTERNALS_VERSION > 4
     // Note that we have to invoke strdup to allocate memory to ensure a unique address
     // We want unique addresses since we use pointer equality to compare function records
-    std::string function_record_capsule_name = "pybind11_function_record_capsule";
+    std::string function_record_capsule_name = internals_function_record_capsule_name;
 #    endif
 
     internals() = default;
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -182,6 +182,16 @@ struct internals {
 #    endif // PYBIND11_INTERNALS_VERSION > 4
     // Unused if PYBIND11_SIMPLE_GIL_MANAGEMENT is defined:
     PyInterpreterState *istate = nullptr;
+
+#    if PYBIND11_INTERNALS_VERSION > 4
+    // Note that we have to invoke strdup to allocate memory to ensure a unique address
+    // We want unique addresses since we use pointer equality to compare function records
+    std::string function_record_capsule_name = "pybind11_function_record_capsule";
+#    endif
+
+    internals() = default;
+    internals(const internals &other) = delete;
+    internals &operator=(const internals &other) = delete;
     ~internals() {
 #    if PYBIND11_INTERNALS_VERSION > 4
         PYBIND11_TLS_FREE(loader_life_support_tls_key);
@@ -196,11 +206,6 @@ struct internals {
         // that the `tstate` be allocated with the CPython allocator.
         PYBIND11_TLS_FREE(tstate);
     }
-#endif
-#if PYBIND11_INTERNALS_VERSION >= 5
-    // Note that we have to invoke strdup to allocate memory to ensure a unique address
-    // We want unique addresses since we use pointer equality to compare function records
-    const char *function_record_capsule_name = strdup("pybind11_function_record_capsule");
 #endif
 };
 
@@ -554,15 +559,15 @@ const char *c_str(Args &&...args) {
 }
 
 inline const char *get_function_record_capsule_name() {
-#if PYBIND11_INTERNALS_VERSION >= 5
-    return get_internals().function_record_capsule_name;
+#if PYBIND11_INTERNALS_VERSION > 4
+    return get_internals().function_record_capsule_name.c_str();
 #else
     return nullptr;
 #endif
 }
 
 // Determine whether or not the following capsule contains a pybind11 function record.
-// Note that we use internals to make sure that only ABI compatible records are touched.
+// Note that we use `internals` to make sure that only ABI compatible records are touched.
 //
 // This check is currently used in two places:
 // - An important optimization in functional.h to avoid overhead in C++ -> Python -> C++
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -49,9 +49,19 @@ public:
 #ifndef BAD
         if (auto cfunc = func.cpp_function()) {
             auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
-            if (isinstance<capsule>(cfunc_self)) {
+            if (cfunc_self == nullptr) {
+                PyErr_Clear();
+            } else if (isinstance<capsule>(cfunc_self)) {
                 auto c = reinterpret_borrow<capsule>(cfunc_self);
-                auto *rec = (function_record *) c;
+
+                function_record *rec = nullptr;
+                // THIS IS THE PROBLEM: UNSAFE NO WAY TO FIGURE OUT IF CAPSULE IS FOREIGN
+                // We would need to refactor to store a special string such as
+                // pybind11_function_record Doing so is almost certainly an ABI break though Best
+                // we can do without an ABI break is ignore named capsules
+                if (c.name() == nullptr) {
+                    rec = static_cast<function_record *>(c);
+                }
 
                 while (rec != nullptr) {
                     if (rec->is_stateless
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -59,7 +59,7 @@ public:
                 // We would need to refactor to store a special string such as
                 // pybind11_function_record Doing so is almost certainly an ABI break though Best
                 // we can do without an ABI break is ignore named capsules
-                if (c.name() == nullptr) {
+                if (c.name() == function_capsule_name()) {
                     rec = static_cast<function_record *>(c);
                 }
 
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -54,11 +54,7 @@ public:
                 auto c = reinterpret_borrow<capsule>(cfunc_self);
 
                 function_record *rec = nullptr;
-                // THIS IS THE PROBLEM: UNSAFE NO WAY TO FIGURE OUT IF CAPSULE IS FOREIGN
-                // We would need to refactor to store a special string such as
-                // pybind11_function_record Doing so is almost certainly an ABI break though Best
-                // we can do without an ABI break is ignore capsules with the wrong name
-                //
+                // Check that we can safely reinterpret the capsule into a function_record
                 // Note that we compare pointers, not values to ensure each extension is unique
                 if (c.name() == function_capsule_name()) {
                     rec = static_cast<function_record *>(c);
--- a/include/pybind11/functional.h
+++ b/include/pybind11/functional.h
@@ -55,8 +55,7 @@ public:
 
                 function_record *rec = nullptr;
                 // Check that we can safely reinterpret the capsule into a function_record
-                // Note that we compare pointers, not values to ensure each extension is unique
-                if (c.name() == function_capsule_name()) {
+                if (detail::is_function_record_capsule(c)) {
                     rec = static_cast<function_record *>(c);
                 }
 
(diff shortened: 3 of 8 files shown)


Skylion007:
Oof, @rwgk got any Python experts to ping about this?

rwgk:
```
            auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
            if (isinstance<capsule>(cfunc_self)) {
````

Did you already try to do something like this?

```
            auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
            if (cfunc_self == nullptr) {
                PyErr_Clear();
            } else if (isinstance<capsule>(cfunc_self)) {
````

File: include/pybind11/functional.h [hunk 1]
```diff
@@ -38,14 +38,15 @@ struct type_caster<std::function<Return(Args...)>> {
...
+   function via Python, every function call would normally involve
+   a full C++ -> Python -> C++ roundtrip, which can be prohibitive.
+   Here, we try to at least detect the case where the function is
+   stateless (i.e. function pointer or lambda function without
+   captured variables), in which case the roundtrip can be avoided.
+ */
+#ifndef BAD
         if (auto cfunc = func.cpp_function()) {
             auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
             if (isinstance<capsule>(cfunc_self)) {
```
Skylion007, line 52: @rwgk It looks like the issue is here where we have a capsule, but the contents of the capsule is not what we expect (a function record pointer). Therefore our reinterepret_cast is unsafe. We need a way to disambiguate function_record pointers from other contents in the capsule, thoughts @rwgk?

rwgk: General / without having looked very closely: AFAIK the capsule `name` is the only way to achieve some level of type safety.

rwgk: > we store it as nullptr.

The `name` is nullptr? (just to make sure I'm not misunderstanding)

Skylion007:
@rwgk @Lalaland @henryiii I added workaround, but unclear if I should keep the new pytype ctor with it's current argument ordering. It might be the best way to disambiguate whether the dtor func takes the capsule itself or the underlying void value. Alternatively, we may want to create a custom tag long term.

include/pybind11/functional.h
Skylion007: Yes.

include/pybind11/functional.h
```diff
@@ -38,19 +38,30 @@ struct type_caster<std::function<Return(Args...)>> {
...
                 auto c = reinterpret_borrow<capsule>(cfunc_self);
-                auto *rec = (function_record *) c;
+
+                function_record *rec = nullptr;
+                // THIS IS THE PROBLEM: UNSAFE NO WAY TO FIGURE OUT IF CAPSULE IS FOREIGN
+                // We would need to refactor to store a special string such as
+                // pybind11_function_record Doing so is almost certainly an ABI break though Best
+                // we can do without an ABI break is ignore named capsules
+                if (c.name() == nullptr) {
+                    rec = static_cast<function_record *>(c);
```
Skylion007, line 63: @rwgk Why does a static_cast even work here and not give a compiler error. Do we define a custom conversion somewhere?

include/pybind11/pytypes.h
```diff
@@ -1824,8 +1824,8 @@ class capsule : public object {
         }
     }
 
-    capsule(const void *value, void (*destructor)(void *)) {
-        m_ptr = PyCapsule_New(const_cast<void *>(value), nullptr, [](PyObject *o) {
+    capsule(const void *value, void (*destructor)(void *), const char *name = nullptr) {
```
Skylion007, line 1827: note: this argument order is purposefully different from other ctors. Also currently the dtor does not free it, we may want to copy it and ensure we free it.

include/pybind11/functional.h
```diff
@@ -38,19 +38,30 @@ struct type_caster<std::function<Return(Args...)>> {
...
             auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
-            if (isinstance<capsule>(cfunc_self)) {
+            if (cfunc_self == nullptr) {
+                PyErr_Clear();
+            } else if (isinstance<capsule>(cfunc_self)) {
                 auto c = reinterpret_borrow<capsule>(cfunc_self);
-                auto *rec = (function_record *) c;
+
+                function_record *rec = nullptr;
+                // THIS IS THE PROBLEM: UNSAFE NO WAY TO FIGURE OUT IF CAPSULE IS FOREIGN
```
Skylion007, line 58: This is actually the problem in several places in pybind11.h, but hitting a custom def is much rarer there.

include/pybind11/functional.h
rwgk: Looking (Python 3.9) to see if we can find an explanation:

```
#define PyCFunction_GET_SELF(func) \
        (((PyCFunctionObject *)func) -> m_ml -> ml_flags & METH_STATIC ? \
         NULL : ((PyCFunctionObject *)func) -> m_self)
```

Looks like the type of `cfunc_self` is `PyCFunctionObject *`.

Looking that up:
```
typedef struct {
    PyObject_HEAD
    PyMethodDef *m_ml; /* Description of the C function to call */
    PyObject    *m_self; /* Passed as 'self' arg to the C func, can be NULL */
    PyObject    *m_module; /* The __module__ attribute, can be anything */
    PyObject    *m_weakreflist; /* List of weak references */
    vectorcallfunc vectorcall;
} PyCFunctionObject;
```

We're casting that to to a `capsule`:

```
auto c = reinterpret_borrow<capsule>(cfunc_self);
```
With

https://github.com/pybind/pybind11/blob/fab1eebe2c4c52e0abac249f3d058787bc83b5ec/include/pybind11/pytypes.h#L1803

Then here the `capsule` to a `function_record *`:
```
static_cast<function_record *>(c);
```

With

https://github.com/pybind/pybind11/blob/fab1eebe2c4c52e0abac249f3d058787bc83b5ec/include/pybind11/attr.h#L188

And that works? — I don't know how. My way of finding out would be to reduce the code.

rwgk: > @rwgk Nah, we are just extracting m_self from the function. We set that to be a py capsule in other parts of the code. I don't know why the static_cast works from a py::capsule works though.

Oh, yes, sorry I skipped that. So it makes sense that we cast from `PyObject *` via `handle` to `capsule`, but how from `capsule` to `function_record *` I also still don't know.

jagerman: > I don't know why the static_cast works from a py::capsule works though.

`py::capsule` has an implicit `operator T*() const` that looks like it will happily convert to any pointer type you like (and even does so implicitly).  That's... a bit ick.

Lalaland:
@Skylion007 I don't think it would be an abi break to set a name and check it in functional.h

This is an optimization path, so if we incorrectly ignore a capsule, that's ok as it will just run slower.

Skylion007:
Okay @Lalaland/@rwgk, whenever I try to set_name on the rec_capsules I am getting segfaults (likely from the lifetime of the name string???). 

We could try to strdup the cstr in the relevant bindings and ensure to free it on our dtor, but that seems error prone. This is super weird as I never call free it on and I set it a global constexpr like `constexpr const char *FUNCTION_RECORD_CAPSULE_NAME = "pybind11_function_record" just set_name with either the method or the ctor so we shouldn't have to mange the lifetime at all since it's constexpr.

The current iteration at least ensures we won't try to convert pointers for named_capsules, but it would be great if we could give func_record_capsules their own name without segfaulting. Unfortunately, I do not have time to look into this further.

include/pybind11/pybind11.h
```diff
@@ -69,6 +69,12 @@ inline bool apply_exception_translators(std::forward_list<ExceptionTranslator> &
     return false;
 }
 
+// Need to use a wrapper function to ensure 1 address
+inline const char *function_capsule_name() {
+    static const char *name = "pybind11_function_capsule";
```
Skylion007, line 74: Should be constexpr

include/pybind11/pybind11.h
```diff
@@ -468,13 +474,20 @@ class cpp_function : public function {
...
-                                                                : capsule(self);
-                chain = (detail::function_record *) rec_capsule;
-                /* Never append a method to an overload chain of a parent class;
-                   instead, hide the parent's overloads in this case */
-                if (!chain->scope.is(rec->scope)) {
+                if (!isinstance<capsule>(self)) {
                     chain = nullptr;
+                } else {
+                    auto rec_capsule = reinterpret_borrow<capsule>(self);
+                    if (rec_capsule.name() == detail::function_capsule_name()) {
```
Skylion007, line 481: Is it really safe to do pointer comparison here? Addresses from different submodules could be different? Or maybe if a capsule from a different module is somehow added? Should be std::strcmp, right?

Lalaland: The pointer comparison is on purpose to avoid binary compatibility issues. Otherwise we would have multiple detail::function_record's with different ABIs trying to interact with each other (which could be a disaster if they are from different compilers / STL versions).

I can switch it to std::strcmp, but that is much less safe.

include/pybind11/pybind11.h
Lalaland: That requires C++23 to support static variables within constexpr functions.

Skylion007: Actually, can we use a constexpr function here? I still don't get why this function is necessary. We don't need the address to be the same (just the value)

Skylion007: Even if the function is not constexpr, I think that variable can be a `static constexpr const char *`

Lalaland: We do actually need the address to be the same to enforce ABI safety. We don't want function_records from different ABI versions interacting.

Lalaland: Let's resolve this discussion and keep the ABI talk in the thread below, no need to split it.

include/pybind11/pybind11.h
Skylion007: I am a bit confused as to why it's less safe. Are we sure a foreign capsule could never be a self object here? I think we may be able to construct a case where it is? (Ie. a pybind11 class the subclasses an imported pybind11 class, right?)

Lalaland: > Even if the function is not constexpr, I think that variable can be a static constexpr const char *

static constexpr const char* and my current function have different semantics.

static constexpr const char* creates a single variable per translation unit.

My current function creates a single actually global variable that is shared across an *.so.

We want to ensure that things are shared across an *.so so we can't use a static variable here.

Lalaland: > I am a bit confused as to why it's less safe. Are we sure a foreign capsule could never be a self object here? I think we may be able to construct a case where it is? (Ie. a pybind11 class the subclasses an imported pybind11 class, right?)

The case where it would be less safe is right here in functional.h 

If you pass a capsule function from a foreign pybind11 class, it will break.

Lalaland: > Ie. a pybind11 class the subclasses an imported pybind11 class, right?

This is undefined behavior unless both classes are on the exact same ABI (same compiler, same STL, same pybind11 version, etc).

Skylion007: We should add a comment for these comparisons that these are pointer comparisons and not string comparison bugs.

Skylion007:
@Lalaland One scenario I am worried about. Suppose we have multiple different translation units, is the address guaranteed to be the same across them? I know particularly it will be with gcc/llvm, but is this guaranteed or just their default behavior? Particularly with link time optimized turned off? I am worried where splitting up a large translation unit could accidentally cause a performance regression in this case which is unexpected behavior. Part of the issue is that we do not have a regression test for this optimization. 

Thoughts on this behavior @rwgk? Should we really avoid an std::strcmp here and do this pointer black magic?

Lalaland:
Hmm. Just realized that the pointer black magic here is technically incorrect. 

Should be `static char[] name = "pybind11_function_capsule";`

Lalaland:
Actually, my pointer black magic logic is/was incorrect. I just pushed a fixed version.

` inline const char *function_capsule_name() {
-    static const char *name = "pybind11_function_capsule";
+    static char name[] = "pybind11_function_capsule";
     return name;
 }
`

Skylion007:
> Actually, my pointer black magic logic is/was incorrect. I just pushed a fixed version.
> 
> ```
> inline const char *function_capsule_name() {
> -    static const char *name = "pybind11_function_capsule";
> +    static char name[] = "pybind11_function_capsule";
>      return name;
> }
> ```

How did you test it? Do you have a URL link to why this works and why a const char pointer doesn't?

> Actually, my pointer black magic logic is/was incorrect. I just pushed a fixed version.
> 
> ```
> inline const char *function_capsule_name() {
> -    static const char *name = "pybind11_function_capsule";
> +    static char name[] = "pybind11_function_capsule";
>      return name;
> }
> ```

How did you test this? Do you have a link to why this works and the const char pointer did not? I also extracted out a helper function to make it a bit more readable.

Lalaland:
>  I know particularly it will be with gcc/llvm, but is this guaranteed or just their default behavior?

Per the documentation of local_internals (which uses the same trick), this should work. Each .so should get their own copy.

https://github.com/pybind/pybind11/blob/f743bdf8e61b2dc4a8995a0485111bedd9c2cbb1/include/pybind11/detail/internals.h#L471

Lalaland:
Actually, let's just put this key in local_internals. So we only have this hack implemented in one location.

Skylion007:
@Lalaland It should definitely put in global internals, not local one.

Lalaland:
Ok, I moved this to local_internals, and bumped the ABI version flag. 

> How did you test this? Do you have a URL link to why this works and the const char pointer did not?

Per the C++ standard, in order to ensure that you have a unique pointer you need to actually allocate memory. See the text in https://stackoverflow.com/questions/50610274/what-is-the-rationale-behind-returning-unique-addresses-for-allocations-of-zero

Skylion007 (changes requested):
@Lalaland It really should be in internals, not local internals. Callbacks are frequently and often shared between modules so this will make performance worse.

Skylion007:
Also putting this in internals does not solve the issue I have with this, mainly that the check is too strict and overly brittle. Any minor change will cause issues. Instead, we could even have a function_record_specific ABI and store that in the function_name string itself. No reason to rely on all this pointer stuff and it would be a lot less brittle. We can manually increment the rec_capsule name when we change the function record too.

include/pybind11/detail/internals.h
```diff
@@ -510,6 +514,10 @@ struct local_internals {
             = static_cast<shared_loader_life_support_data *>(ptr)->loader_life_support_tls_key;
     }
 #endif //  defined(WITH_THREAD) && PYBIND11_INTERNALS_VERSION == 4
+
+#if PYBIND11_LOCAL_INTERNALS_VERSION >= 1
+    const char *function_capsule_name = strdup("pybind11_function_capsule");
```
Skylion007, line 519: Why does it need to be strdup? Doesn't making it a pointer instead of a  const char[] array break the pointer comparison again?

Lalaland:
> Also putting this in internals does not solve the issue I have with this, mainly that the check is too strict and overly brittle. Any minor change will cause issues. Instead, we could even have a function_record_specific ABI and store that in the function_name string itself. No reason to rely on all this pointer stuff and it would be a lot less brittle. We can manually increment the rec_capsule name when we change the function record too.

Yeah, this is too smart by half. Simply having a regular name that we strcmp against with a manual version check would be a lot easier to understand.

It's a tiny bit more error-prone, but that should be OK. I'll swap it out for that.

include/pybind11/detail/internals.h
Lalaland: Idea was to allocate memory to ensure a unique address. But this code is now gone since you are right that this should be simpler.

include/pybind11/pybind11.h
```diff
@@ -69,9 +69,14 @@ inline bool apply_exception_translators(std::forward_list<ExceptionTranslator> &
     return false;
 }
 
+constexpr const char* function_capsule_name = "pybind11_funciton_capsule_v1";
+
 inline bool is_function_record_capsule(const capsule &cap) {
     // Compare the pointers, not the values to ensure that each extension is unique
```
Skylion007, line 75: This comment no longer is valid.

include/pybind11/pybind11.h
```diff
@@ -69,9 +69,14 @@ inline bool apply_exception_translators(std::forward_list<ExceptionTranslator> &
     return false;
 }
 
+constexpr const char* function_capsule_name = "pybind11_funciton_capsule_v1";
```
Skylion007: There is a typo. And maybe this should be colocated with the function_record struct with a comment to update when relevant.

tests/test_callbacks.cpp
```diff
@@ -16,6 +16,8 @@
 
 int dummy_function(int i) { return i + 1; }
 
+static PyMethodDef def;
```
rwgk, line 19: Minor, but ... it's unfortunate that this very short and generic name is used only 220+ lines later.

But down there a longer name is annoying.

General: lumping chunks of unrelated code together is error prone (it's common practice in this code base, but that doesn't make it good).

Two birds with one stone: move the code below into a helper function up here? I'm thinking this line can be inside the helper function.

rwgk (commented):
Some comments, first glance through.

I'll try some basic health check (sanitizers, global testing) although I don't expect surprises.

include/pybind11/pybind11.h
```diff
@@ -69,6 +69,13 @@ inline bool apply_exception_translators(std::forward_list<ExceptionTranslator> &
     return false;
 }
 
+inline bool is_function_record_capsule(const capsule &cap) {
+    if (cap.name() == nullptr) {
+        return false;
+    }
+    return (std::strcmp(cap.name(), function_record::function_capsule_name) == 0);
```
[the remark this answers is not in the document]
Lalaland: The primary thing we are worried about isn't speed, it's ABI safety.

The question is what function_record's should pybind11 try to manipulate?

Options are:
1. function_records only from the same .so (pointer comparison using local_internals)
2. function_records only from the same PYBIND11_INTERNALS_ID (pointer comparison using internals)
3. all function_records (std::strcmp using manual string)

See my response to the above discussion for what "manipulation" pybind11 does (but note that the exact type of manipulation probably doesn't matter that much for ABI; any use is UB and a bad idea).

include/pybind11/attr.h
```diff
@@ -185,7 +185,12 @@ struct argument_record {
 
 /// Internal data structure which holds metadata about a bound function (signature, overloads,
 /// etc.)
+// Note that function_capsule_name must be updated whenever this struct is changed
```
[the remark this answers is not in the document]
Lalaland: Good point. I'll add a comment here.

There are two places where this is used:

1. Functional.h for a performance improvement if a pybind11 function is detected. On failure, runs slower.

2. cpp_function's sibling feature for function overloading.

Think of the following

module 1:
module_<..> a;
a.attr("foo") = cpp_function([](int a));

module 2:
object a = import(a);
a.attr("bar") = cpp_function([](string a){}, sibling(a.attr("foo"));

The sibling feature means that the "bar" function will call "foo" if the input cannot be cast to string.

include/pybind11/pybind11.h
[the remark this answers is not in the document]
Lalaland: I would argue that 1 is safer than 2, and 2 is safer than 3.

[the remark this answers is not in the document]
rwgk: Thanks a lot for the explanation! I'll need a little bit of time to look more carefully. But one quick high-level question: do we have timings, even just crude, for what the performance hit is for a "typical situation in the wild" (vs. micro-benchmark only)?

A super crude idea (not sure it's feasible & meaningful, but just to start somewhere): run certain existing unit tests always returning `false` here.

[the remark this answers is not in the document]
Lalaland: The performance benefit of functional.h's optimization will be extremely dependent on the setup. It's really just the cost of the type_converters for each input argument and return value. For something like std::vector, it will be very high, for something like integers it's basically zero.

The bigger question is to what degree do applications rely on this optimization across pybind11 modules ...

Our unit tests only try this optimization for functions that take integers so there is practically performance difference (well, 170 nanoseconds).

include/pybind11/functional.h
Skylion007: @rwgk Well, we could implicitly convert without a static cast, but that would be even more surprising. I suppose we could also add a special explicit cast method to capsules and deprecate the implicit one somehow? That sounds like another PR though.

rwgk: Based on glancing through this code

* https://github.com/pybind/pybind11/blob/17c1e27b3d202aaab6d5443b1800124bbf24d9ca/include/pybind11/pytypes.h#L1866-L1880

I'm thinking `rec = c.get_pointer<function_record>();` may work.

rwgk:
> I'll try some basic health check (sanitizers, global testing) although I don't expect surprises.

Everything passed (asan, msan, ubsan, tsan), including the global testing (snapshot at 51f62b224d548420f8804aea12bb7e87053fa6b0 ~22 hours ago, still HEAD).

include/pybind11/pybind11.h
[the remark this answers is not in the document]
rwgk: Thanks, that's very clear!

Hm ... I'd think people generally expect no significant performance regressions, that would rule out 1.

2. The internals version trick has known limitations and pitfalls (potential UB) that are generally accepted.

3. means modules are not ABI compatible, presumably not expected to cooperate, presumably have very little expectations re performance, apart maybe "it used to be faster" for a handful of users (who may be blissfully unaware of the functional.h mechanics).

I'm making this up, I don't think this can be decided in a pure scientific way.

I'd settle on 2.

include/pybind11/attr.h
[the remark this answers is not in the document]
rwgk: Wow! I didn't even know about the `sibling` feature TBH. I looked, there is exactly 1 use in our extended codebase (1st and a lot of 3rd party code). This is the gem, JIC someone is curious:

https://github.com/llvm/llvm-project/blob/3f8d2c917c3da4ad33ad4aa6f988ae3e8dfacdf0/mlir/include/mlir/Bindings/Python/PybindAdaptors.h#L298

What does "On failure" mean for the `sibling` case? That "bar" will not be able to call "foo"?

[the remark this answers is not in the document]
Lalaland: > What does "On failure" mean for the sibling case? That "bar" will not be able to call "foo"?

Yep. You will get a type error instead.

include/pybind11/functional.h
```diff
@@ -48,9 +48,16 @@ struct type_caster<std::function<Return(Args...)>> {
          */
         if (auto cfunc = func.cpp_function()) {
             auto *cfunc_self = PyCFunction_GET_SELF(cfunc.ptr());
```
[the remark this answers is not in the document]
Lalaland: Not really. I tried it and it failed. Perhaps a future PR could try cleaning this up further, but right now I don't want to heavily touch this very fiddly bit of logic.

include/pybind11/pybind11.h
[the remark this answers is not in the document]
Lalaland: 2 is now implemented.

include/pybind11/detail/internals.h
```diff
@@ -189,6 +189,11 @@ struct internals {
         PYBIND11_TLS_FREE(tstate);
     }
 #endif
+#if PYBIND11_INTERNALS_VERSION >= 5
+    // Note that we have to invoke strdup to allocate memory to ensure a unique address
+    // We want unique addresses since we use pointer equality to compare function records
+    const char *function_record_capsule_name = strdup("pybind11_function_record_capsule");
```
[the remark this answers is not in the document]
Lalaland: > We're leaking this, that's fine.

Yep, all of internals leaks by design. We can never deallocate anything in it.

> We're assuming only one instance is constructed per process.

Incorrect, it's one instance per PYBIND11_INTERNALS_ID. If there are multiple pybind11 modules with different PYBIND11_INTERNALS_IDs, you will get different copies by design.

> Tangential, but since we are making a change here already, I think this would be a good addition, to make the intent clear, and to guard against accidents:

Yep, that's a good idea. Although note that we should also add

> // Always leak, never delete
>  ~internals() = delete;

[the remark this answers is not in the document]
rwgk: > Incorrect, it's one instance per PYBIND11_INTERNALS_ID. If there are multiple pybind11 modules with different PYBIND11_INTERNALS_IDs, you will get different copies by design.

Oh, of course, thanks for the correction!

> // Always leak, never delete
> ~internals() = delete;

I was curious and tried it out, it fails with a compiler error because there is a destructor already.

Then I added a segfault trigger in the existing destructor (`long *BAD = nullptr; *BAD = 101;`). With that test_embed fails but the pytest tests all pass.

I'm not sure what to make of those observations, but I'm thinking let's not do anything about it in this PR (or anytime soon). I'm not worried at all about the `function_record_capsule_name` leak.

**Tangent**: I was about to write: I know for sure cpython leaks a lot more: https://github.com/rwgk/stuff/blob/f6549ecadb46169feb5f3c2456dcd13852382f7a/noddy/embedded_noddy_main_run.c

But I just found out: it still leaks with Python 3.9, but Python 3.10 is clean!

Anyway, I'm still not worried about the leak.

Lalaland:
Oh, oops, didn't think about the embedding case. Where we really don't want
to leak. I'll swap this to a std::string.

Lalaland:
I think I have resolved everyone's comments. This PR should now be mostly good to go.

The only caveat is that we probably want to only merge this after 2.10.1 gets released. It's not an ABI break, but it's too risky of a change to include in 2.10.1 (#4125).

I'll rebase this PR once 2.10.1 is released.

tests/test_callbacks.cpp
```diff
@@ -240,4 +240,38 @@ TEST_SUBMODULE(callbacks, m) {
...
+                throw std::runtime_error("Invalid argument to example_name");
+            }
+            auto result = py::cast(PyLong_AsLong(first) * 9);
+            return result.release().ptr();
+        };
+        def.ml_flags = METH_VARARGS;
+        return &def;
+    }();
+
+    // rec_capsule with custom name
```
rwgk, line 263: Question before maybe making a suggestion, is the point here:
* the capsule name matches that defined in `internals`,
* but it's a different pointer?

Lalaland: Yep. This is how we ensure ABI safety. Only exact pointer matches are allowed.

Lalaland: If we were using unsafe string matching (and not the safe pointer matching), this test would fail and cause a default.

rwgk: Cool, thanks!

I'm worried about the names here and in internals.h getting out of sync in the long run.

Suggestion:

In internals.h, `namespace pybind11::detail`:
```
static constexpr const char *internals_function_record_capsule_name = "pybind11_function_record_capsule";
```

Then use that in internals.h and here.

Add a comment here, something like:

```
// Ensure that the capsule is not accidentally used as a pybind11 function_record, even if the name matches.
```

Lalaland: Done!

include/pybind11/detail/internals.h
```diff
@@ -182,6 +185,16 @@ struct internals {
 #    endif // PYBIND11_INTERNALS_VERSION > 4
     // Unused if PYBIND11_SIMPLE_GIL_MANAGEMENT is defined:
     PyInterpreterState *istate = nullptr;
+
+#    if PYBIND11_INTERNALS_VERSION > 4
+    // Note that we have to invoke strdup to allocate memory to ensure a unique address
```
rwgk, line 190: Sorry I missed this before: this still mentions strdup

rwgk (approved):
Awesome work, thanks a lot for finding & fixing this bug!

include/pybind11/detail/internals.h
Lalaland: Oops, good catch!

include/pybind11/pybind11.h
```diff
@@ -663,8 +671,10 @@ class cpp_function : public function {
         using namespace detail;
```
Skylion007, line 672:
```suggestion
        using namespace detail;
        assert(isinstance<capsule>(self));
        
```

include/pybind11/detail/internals.h
```diff
@@ -182,6 +185,16 @@ struct internals {
 #    endif // PYBIND11_INTERNALS_VERSION > 4
     // Unused if PYBIND11_SIMPLE_GIL_MANAGEMENT is defined:
     PyInterpreterState *istate = nullptr;
+
+#    if PYBIND11_INTERNALS_VERSION > 4
+    // Note that we have to use a std::string to allocate memory to ensure a unique address
+    // We want unique addresses since we use pointer equality to compare function records
+    std::string function_record_capsule_name = internals_function_record_capsule_name;
```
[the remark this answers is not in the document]
rwgk: > I am still not entirely sold on this since it could cause unexpected slowdowns when using passing callback builtins from other compilation units.

The `internals` object is reused between compilation units, even if using different pybind11 versions, via a capsule stored in builtins.

If the `internals` keys for retrieving that capsule don't match, using the `function_record` pointer is UB, for sure (unless people manipulate the macros for building `PYBIND11_INTERNALS_ID` in weird ways, but that's their own responsibility if they incur slowdowns then).

> We did test that this currently works to trigger the callback optimization right? It's just not skipping it everytime, right?

Yes, I inserted a print to check:

```
test_callbacks.py::test_cpp_function_roundtrip
LOOOK /usr/local/google/home/rwgk/forked/pybind11/include/pybind11/functional.h:70

LOOOK /usr/local/google/home/rwgk/forked/pybind11/include/pybind11/functional.h:70

LOOOK /usr/local/google/home/rwgk/forked/pybind11/include/pybind11/functional.h:70

LOOOK /usr/local/google/home/rwgk/forked/pybind11/include/pybind11/functional.h:70
PASSED
```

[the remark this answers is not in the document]
Lalaland: > We did test that this currently works to trigger the callback optimization right? It's just not skipping it everytime, right?

That test is actually part of our existing test suite. 

https://github.com/pybind/pybind11/blob/14b7e335306907f2ebd2b84d635f035076689622/tests/test_callbacks.py#L93

rwgk:
The one CI failure is a notorious flake again. Good to ignore.

I locally retested with all sanitizers (ASAN, MSAN, UBSAN, TSAN), everything passes (this PR rebased on the smart_holder branch).

I'll also redo the global testing (easy, but will take a few hours).

include/pybind11/detail/internals.h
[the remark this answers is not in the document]
Skylion007: That makes me feel a bit better at least.

include/pybind11/detail/internals.h
```diff
@@ -43,6 +43,9 @@ using ExceptionTranslator = void (*)(std::exception_ptr);
 
 PYBIND11_NAMESPACE_BEGIN(detail)
 
+static constexpr const char *internals_function_record_capsule_name
```
Skylion007, line 46: What does static do in this context? Add a comment?

rwgk: That was my suggestion: I'm never entirely sure where and why `static` is needed.
But clang-tidy was happy.
I don't think it needs a comment, the intent is clear. If a tool complains in the future, we'll fix it.

Lalaland: @rwgk requested static there, not sure why? I also don't quite understand the reason for the static when it is already constexpr?

rwgk: In class context it's definitely needed, e.g:

```
rwgk.c.googlers.com:~/forked/pybind11/include $ grep -r 'static constexpr' | wc -l
117
```

`namespace`, I'm not sure.

But generally I try not to waste too many of the few neurons that I have on clearly inconsequential details of an extremely quirky language. If a tool tells me to do it differently I comply. Our tooling is almost insanely thorough. I stop thinking there.

Lalaland: > I'm never entirely sure where and why static is needed

You need static whenever you ODR use a constexpr variable that is in a header file to avoid linking errors. 

One of the annoying traps in the C++ language. This case probably dosen't need static because we don't ODR use it.

rwgk:
> I'll also redo the global testing (easy, but will take a few hours).

Global testing passed again (internal ID OCL:482903768:BASE:485514840:1667371815628:a4b009b9).

Note that the global testing now includes running all pybind11 unit tests with all sanitizers (recent change).

Skylion007 (approved):

rwgk:
Thanks again Ethan for the fix, and Aaron for the reviews!

rwgk:
Hi @Lalaland, based on what I learned & implemented under PR #4329, I'm wondering if we could improve on this PR, by **not** making any changes to the `internals` struct.

What I have in mind would be similar to this commit: 8737beaad503ce5136cb6c6119254e201132d60d

In case the hash changes in the future (b/o rebase), this is the commit message:

```
    Split out `get_native_enum_type_map()`, leaving `struct internals` untouched (compared to master).
```

What do you think?

The current implementation under PR #4329 has a lot of code duplication, but we could work on that pretty easily I think.

Lalaland:
@rwgk If I understand your suggested change correctly, it looks like you are proposing to split our global variable "internals" into multiple "internals-like" variables? (When I say "internals-like" I mean a global variable that is shared across modules and thus needs to be ABI stable).

The main advantage of that approach is that it would give us a bit more flexibility, but I think the maintenance costs would be really really bad. 

Right now ABI issues are theoretically straightforward. There is one shared struct "internals" that has all the stuff that has to be ABI stable with one ABI version. Code on different ABI versions just doesn't talk to each other, keeping things safe.

With your change there would be many "internals" structs that all have to have ABI stability with a very confusing mix of versions. It would be a lot harder to understand what code under what ABI version is talking to what other code.

rwgk:
> The main advantage of that approach is that it would give us a bit more flexibility, but I think the maintenance costs would be really really bad.

I totally agree, but ... the alternative is far worse, and it's connected to how widely pybind11 is in use: the moment we bump the internals version, we're essentially cutting off interoperability with all existing wheels, **silently**. There will suddenly be `TypeErrors` without any direct information about the root cause, and how to fix it.

Say, there are 10 extensions — in wheels — with cross-extension type conversion dependencies, in the general sense that's a graph with 10 nodes. Now one of them is rebuilt with a new `internals` version. It is disconnected from that graph, or to offer an alternative viewpoint, a new graph is started. The 10 interdependent extensions will be fully operational again only after all 10 wheels are rebuilt.

Now go from 10 to 20 to 100.

In the Google-internal world there is no problem: I can switch **everybody** with a single commit.

Externally, waiting for 10 wheels to get updated takes 10 weeks, 20 10 months, 100 takes years.

I think the only way to mitigate that is to
* look as the `internals`-like structures as carved into stone when they are released,
* copy-paste-edit new versions,
* provide backwards compatibility for N years, before the old `internals`-like structures are finally purged.

Lalaland:
@rwgk Why don't we tie our internals version to the Python version? And only release new internals when a new Python version gets released?

Python already has an unstable ABI. If people are recompiling anyways, might as well force them to use a newer version of pybind11.

Lalaland:
Tying our internals version to the Python version will make upgrading the internals version much much easier, as we simply wait for the next Python release and people are already being forced to recompile the world.

rwgk:
Interesting idea. I need to think about this more.

One issue that comes to mind: that basically freezes the pybind11 `internals` version for a given Python release. Unless we additionally have a migration mechanism***, we're stuck with it for 5 years (EOL for that Python release).

***If we have an additional migration mechanism, there isn't much gained by tying to Python releases.

High-level impression: It'll only help a little bit. A momentary free lunch at Python release time, only benefiting that Python release and future releases 1, 2, ... years later.

Lalaland:
We would of course add an option to force use the latest ABI for users like Google that do control everything.

rwgk:
> We would of course add an option to force use the latest ABI for users like Google that do control everything.

Two days ago (no kidding) I learned the hard way that that does not really work:

* Since 2021-09-20 Google is running with `internals` version 5.
* When the mujoco project is exported from the Google mono repo to https://github.com/deepmind/mujoco, @saran-t undoes my patch, to put it back to `internals` version 4, when building binaries for testing (I'm not sure Saran: are those uploaded to GitHub?).
* Saran also ships a cmake file with the hash of the smart_holder branch that matches what we are using internally (modulo the `internals` version patch).

Stepping back:

* Now imagine I put the `native_enum` feature I'm working on under PR #4329 behind an internals bump: people can happily use `native_enum`s internally, but it will not be available externally.
* Unless I bump the internals version on the smart_holder branch directly, instead of using a Google-internal patch.
* But then smart_holder-based extensions will no longer interoperate with "normal" pybind11 extensions.

I think it will be hell, every day a little bit more, as both the use of pybind11 in general, and the smart_holder branch (exported Google projects), grow.

Going to "rolling internals" ...

removed - deprecated - current

would allow us to manage change, rather than hitting walls all the time.

Concretely:

internals == v4 (forever / until removed)

internals_like_a_v1, 2, 3 ...
internals_like_b_v1, 2, 3 ...

requires a little bookkeeping (in our minds & documentation), and keeping compatibility code around for a while.

But it will work much better for the world at large.  —  IOW a moderate amount of extra work for a handful of people working on pybind11 code (but also more flexibility!), happiness for hundreds or thousands of other projects.

What is the minimum version for any of the internals-like structures could be tied to the Python version. Actually, this could work out very nicely: when we drop a Python version, we know we can drop all deprecated internals-like structures that are only used with that Python version.

Lalaland:
Yeah, that sounds good to me. The key part is letting us bump the internals version with every Python release.

rwgk:
> Yeah, that sounds good to me. The key part is letting us bump the internals version with every Python release.

FYI: I folded `cross_extension_shared_state` into PR #4329. With that it is really easy to handle multiple internals-like shared states.

(That PR is way too big now and will need to be split up, but I'm currently focused on testing the new `native_enum` in the context I need it in, which could take a while.)

Issue #3986, opened 2022-06-02
[BUG]: stl_bind generates invalid type signature for Sphinx/Pylance


Skylion007:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

STL_bind creates subclasses that try to mimic keyview. However, our subclasses have no explicit relation with typing.collections.KeyView or collection.abc.KeyView etc... https://github.com/pybind/pybind11/pull/3985 attempts to fix it by removing the [], but this seems like a cludge and we really should let the typing system represent these as the correct type. I am wondering if there is a way we can still have Sphinx / Pylance recognize these KeyView, ItemViews etc as what they are instead of just an opaque class. Maybe just having them inherit from / mixin collections.abc.KeyViews would be appropiate?

Would appreciate thoughts on this front @rwgk @henryiii?

Skylion007:
@bmerry Since you created the original PR for this feature, any thoughts on if there is a clever way to solve this issue?

bmerry:
> @bmerry Since you created the original PR for this feature, any thoughts on if there is a clever way to solve this issue?

No, no clever ideas. I don't think I've touched pybind11 since I wrote that code, and I think the square bracket convention predated the changes I made.

odel4y:
The current naming with e.g. `KeysView[BindedMap]` also prevents formatting the stubs with black, because it stumbles on the IMO incorrect Python syntax. Is there a reason against a name with valid syntax such as `KeysViewBindedMap`?

I guess the reason to inherit the class from e.g. `collections.abc.KeysView` would be to allow checking its capability with something like this:
```python
isinstance(keys_view_binded_map, collections.abc.KeysView)
```
If so, could this be done with creating a Python class like https://github.com/pybind/pybind11/issues/1193#issuecomment-429451094 or specifically indicating the base class like https://github.com/pybind/pybind11/issues/1193#issuecomment-660192531 ?

Skylion007:
[...]

PR #4259, opened 2022-10-19
Provide better type hints for a variety of generic types

virtuald:
## Description

* Makes for better documentation
* Added support for annotating tuple, dict, list, set, function

@sizmailov I'm pretty sure this will allow users to make better `.pyi` files too. :)

Note: I'm open to bikeshedding on the names, but adding an underscore seemed like a reasonable approach.

## Suggested changelog entry:

```rst
Special 'typed' wrappers now available in typing.h to annotate tuple, dict, list, set, and function
```

### DIFF
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -410,12 +410,12 @@ type.
 .. code-block:: cpp
 
     PYBIND11_MODULE(example, m) {
-        m.def("pass_list_of_str", [](py::List<py::str> arg) {
+        m.def("pass_list_of_str", [](py::typing::List<py::str> arg) {
             // arg can be used just like py::list
         ));
     }
 
-The resulting docstring will be ``pass_list_of_str(arg0: List[str]) -> None``.
+The resulting docstring will be ``pass_list_of_str(arg0: list[str]) -> None``.
 
 The following special types are available in ``pybind11/typing.h``:
 
--- a/docs/advanced/misc.rst
+++ b/docs/advanced/misc.rst
@@ -398,3 +398,32 @@ before they are used as a parameter or return type of a function:
         pyFoo.def(py::init<const ns::Bar&>());
         pyBar.def(py::init<const ns::Foo&>());
     }
+
+Setting inner type hints in docstrings
+======================================
+
+When you use pybind11 wrappers for ``list``, ``dict``, and other generic python
+types, the docstring will just display the generic type. You can convey the
+inner types in the docstring by using a special 'typed' version of the generic
+type.
+
+.. code-block:: cpp
+
+    PYBIND11_MODULE(example, m) {
+        m.def("pass_list_of_str", [](py::List<py::str> arg) {
+            // arg can be used just like py::list
+        ));
+    }
+
+The resulting docstring will be ``pass_list_of_str(arg0: List[str]) -> None``.
+
+The following special types are available in ``pybind11/typing.h``:
+
+* ``py::Tuple<Args...>``
+* ``py::Dict<K, V>``
+* ``py::List<V>``
+* ``py::Set<V>``
+* ``py::Callable<Signature>``
+
+.. warning:: Just like in python, these are merely hints. They don't actually
+             enforce the types of their contents at runtime or compile time.
--- a/include/pybind11/typing.h
+++ b/include/pybind11/typing.h
@@ -2,7 +2,7 @@
     pybind11/typing.h: Convenience wrapper classes for basic Python types
     with more explicit annotations.
 
-    Copyright (c) 2016 Wenzel Jakob <wenzel.jakob@epfl.ch>
+    Copyright (c) 2023 Dustin Spicuzza <dustin@virtualroadside.com>
 
     All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file.
--- a/include/pybind11/typing.h
+++ b/include/pybind11/typing.h
@@ -1,0 +1,94 @@
+/*
+    pybind11/typing.h: Convenience wrapper classes for basic Python types
+    with more explicit annotations.
+
+    Copyright (c) 2016 Wenzel Jakob <wenzel.jakob@epfl.ch>
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include "detail/common.h"
+#include "cast.h"
+#include "pytypes.h"
+
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+
+/*
+    The following types can be used to direct pybind11-generated docstrings
+    to have have more explicit types (e.g., `List[str]` instead of `list`).
+    Just use these in place of existing types.
+
+    There is no additional enforcement of types at runtime.
+*/
+
+template <typename... Types>
+class Tuple : public tuple {
+    using tuple::tuple;
+};
+
+template <typename K, typename V>
+class Dict : public dict {
+    using dict::dict;
+};
+
+template <typename T>
+class List : public list {
+    using list::list;
+};
+
+template <typename T>
+class Set : public set {
+    using set::set;
+};
+
+template <typename Signature>
+class Callable;
+
+template <typename Return, typename... Args>
+class Callable<Return(Args...)> : public function {
+    using function::function;
+};
+
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+template <typename... Types>
+struct handle_type_name<Tuple<Types...>> {
+    static constexpr auto name
+        = const_name("Tuple[") + concat(make_caster<Types>::name...) + const_name("]");
+};
+
+template <>
+struct handle_type_name<Tuple<>> {
+    // PEP 484 specifies this syntax for an empty tuple
+    static constexpr auto name = const_name("Tuple[()]");
+};
+
+template <typename K, typename V>
+struct handle_type_name<Dict<K, V>> {
+    static constexpr auto name = const_name("Dict[") + make_caster<K>::name + const_name(", ")
+                                 + make_caster<V>::name + const_name("]");
+};
+
+template <typename T>
+struct handle_type_name<List<T>> {
+    static constexpr auto name = const_name("List[") + make_caster<T>::name + const_name("]");
+};
+
+template <typename T>
+struct handle_type_name<Set<T>> {
+    static constexpr auto name = const_name("Set[") + make_caster<T>::name + const_name("]");
+};
+
+template <typename Return, typename... Args>
+struct handle_type_name<Callable<Return(Args...)>> {
+    using retval_type = conditional_t<std::is_same<Return, void>::value, void_type, Return>;
+    static constexpr auto name = const_name("Callable[[") + concat(make_caster<Args>::name...)
+                                 + const_name("], ") + make_caster<retval_type>::name
+                                 + const_name("]");
+};
+
+PYBIND11_NAMESPACE_END(detail)
+PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
--- a/include/pybind11/typing.h
+++ b/include/pybind11/typing.h
@@ -15,10 +15,11 @@
 #include "pytypes.h"
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(typing)
 
 /*
     The following types can be used to direct pybind11-generated docstrings
-    to have have more explicit types (e.g., `List[str]` instead of `list`).
+    to have have more explicit types (e.g., `list[str]` instead of `list`).
     Just use these in place of existing types.
 
     There is no additional enforcement of types at runtime.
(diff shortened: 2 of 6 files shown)


Skylion007 (changes requested):
Instead of wrappers, is there anyway we can abuse the existing C++ classes? Like having an optional template arg that has a default value of "Any" or "object". Then if that special template arg is specified we revert to the old caster?

virtuald:
I think that would work.... though, would it break users that are using multiple modules with different versions of pybind11?

Skylion007:
Or if we want to go with legacy Python names, just copy the capitalized nature of the collections `Tuple, Set...`

virtuald:
I like that better, and it matches the annotation as well. I also changed `function_` to `Callable` for consistency.

Lalaland:
If we are trying to support stub generation, we should really add some unit tests for that stub generation.

virtuald:
@Skylion007 

>  i worry that this may be confusing to a user who would be surprised when a particular cast breaks

Then let's put a note in there that it's just for documentation purposes? It's python, we're never guaranteed that anything gets validated at runtime.

> Also, some more information could specified with typing.Annotated, do we have know how we would support that in the future? 

I'm less interested in mypy validation and more interested in documentation and better IDE support. Historically IDEs haven't tried to do introspection on compiled packages, so things like autocomplete and documentation wouldn't work. In that spirit, if there are things we can add that make pybind11-stubgen's life easier, then that makes a lot of sense to me... but other things seem less useful. With a `.pyi` file next to my compiled library, when I click on the function in my code in my properly configured IDE, it brings me right to the pyi documentation, and things like autocomplete and types work as expected. I think mypy will use the pyi file if present as well.

@Lalaland I did add tests for stub generation? pybind11-stubgen generates stubs from the docstrings that pybind11 generates, and the test I added ensures they're in the right form. Unless you're going to integrate pybind11-stubgen into pybind11, there's nothing more to do here.

Skylion007:
@virtuald What about the edge case where a user wants to specify `Tuple` for consistency. IE. no template args.

virtuald:
Added that.

I note that the capitalized form of the builtin types (eg, `Tuple`) is deprecated as of python 3.9, so we could conditionally change the caseness based on the python version. Thinking about it, I think maybe since this is just a docstring, that we should leave the desired form up to the stub generator. It'll also be consistent with our other casters that have the uppercase form.

Skylion007:
What about untyped List, Set, etc...

virtuald:
You have to use the lowercase version for those.

virtuald:
I was contemplating a way to add an arbitrary annotation to types (like `py::annotate<int, "typing.Annotate[int, (2, 3)]">`, but it seems that you can't make a string a template parameter. These changes are probably fine for now though.

Lalaland:
@virtuald You can make a string a template parameter in C++ now

https://ctrpeach.io/posts/cpp20-string-literal-template-parameters/

Simply add some #include guards to only enable this for new enough compilers

virtuald:
Ha, neat:

```c++
template <typename T, detail::descr Name>
struct Annotate : T {
    using T::T;
};

template <typename T, descr Name>
struct handle_type_name<Annotate<T, Name>> {
    static constexpr auto name = Name;
};

...

m.def("annotate_int", [](const py::Annotate<py::int_, py::detail::const_name("Annotate[int, (1,2)]")> &) {});
```

That feels like a separate PR though, there's lots of room for bikeshedding and depending on how one feels about const_name could be more invasive.

Skylion007:
> You have to use the lowercase version for those.

Then we should have a static_assert to enforce that at least one template is specified.

virtuald:
@Skylion007 Only tuple has variable templates, everything else has a required template arg, no need for static assert:

```c++
template <typename T>
class List : public list {
    using list::list;
};
```

I don't think python type hints allow per-element hints for lists, dicts, sets.

sizmailov:
@virtuald I like this idea.
It feels like all "typed" types could be a third-party library or could be defined in a separate header within pybind (e.g. `typing.h`). 
I don't have a strong preference for one over the other; just sharing a thought.

Skylion007:
> @Skylion007 Only tuple has variable templates, everything else has a required template arg, no need for static assert:
> 
> ```c++
> template <typename T>
> class List : public list {
>     using list::list;
> };
> ```
> 
> I don't think python type hints allow per-element hints for lists, dicts, sets.

@virtuald , Ah typo, I meant handling the template args for Callable.

virtuald:
> It feels like all "typed" types could be a third-party library or could be defined in a separate header within pybind (e.g. typing.h).

I think having at least the basic python types in pybind11 makes sense, a third party library would be harder to discover. I'm.. +0.5 on putting all of this in `typing.h` if there are no objections... though, there are a lot of headers in pybind11 these days.

Lalaland:
@virtuald Why can't annotations should be part of pybind11::arg?

a.def("foo_function", [](std::string name){}, py::arg("name",  "List[str]") ;

py::annotate is going to be extremely confusing.

virtuald:
@Lalaland that's a neat approach, I like that we could specify arbitrarily complex type hints. It would have to also include a way to change the hint for return values and properties as well.

In theory that approach could replace this current PR. However, I do like having the ability to do `using MyCb = py::Callable<void(int)>` and reuse it in several places -- so I think they're complementary.

Lalaland:
>  so I think they're complementary

Yep. Totally agree. I think the goal should be that pybind11 should automatically generate the correct types from the C++ argument types, with manual annotations in py::arg only as a fallback solution for when the automatic approach fails.

Skylion007:
@virtuald Could this be extended to solve this typing issue too? https://github.com/pybind/pybind11/issues/3986#issue-1258469342

Skylion007:
@virtuald, Actually, this should just be automatic for newer versions of Python for return types, right? Something to consider if we decide to refactor casters.

virtuald:
I moved the typing related classes to `typing.h`.

I've been using this for awhile with no ill effects as far as I can tell. I don't believe there are any further changes to be made to this specific PR -- other suggestions mentioned above are better suited for their own PRs.

virtuald:
Any reason not to merge this?

rwgk:
> Any reason not to merge this?

@Skylion007 @Lalaland could you help reviewing this? When you guys feel it's ready I'll run the global testing.

henryiii:
Should we be using the "List" form instead of "list", even on modern Python? Also, would it add that much to just change the default to do this instead of making it duplicated and opt-in? (I know that's failed a few times, but IMO typing is very important for all Python these days).

virtuald:
> Should we be using the "List" form instead of "list", even on modern Python?

Meh, who cares?  Also, our STL type casters all use List[], so this uses List. 

> Also, would it add that much to just change the default to do this instead of making it duplicated and opt-in? (I know that's failed a few times, but IMO typing is very important for all Python these days).

pybind11 already does what you're asking for -- if you explicitly return a `py::list` then it will show the return type as a `list`, but there's no way to divine the inner type at compile time without the user explicitly specifying it.

I don't think it's possible to change the existing `py::list` to have the ability to show an inner type without potential severe backwards compatibility issues.

henryiii:
Okay. The capitalized versions are deprecated, but aren't going to disappear soon. Also `set` and `collection.abc.Set` are not the same thing (`MutableSet` is closer). But no need to work on that here.

docs/advanced/misc.rst
```diff
@@ -398,3 +398,32 @@ before they are used as a parameter or return type of a function:
...
+
+The following special types are available in ``pybind11/typing.h``:
+
+* ``py::Tuple<Args...>``
+* ``py::Dict<K, V>``
+* ``py::List<V>``
+* ``py::Set<V>``
+* ``py::Callable<Signature>``
+
+.. warning:: Just like in python, these are merely hints. They don't actually
```
Lalaland, line 428: Please move this warning up, right above the code example. I think it's too easy to miss here.

tests/test_pytypes.cpp
```diff
@@ -811,4 +813,11 @@ TEST_SUBMODULE(pytypes, m) {
         a >>= b;
         return a;
     });
+
+    m.def("annotate_tuple_float_str", [](const py::Tuple<py::float_, py::str> &) {});
+    m.def("annotate_tuple_empty", [](const py::Tuple<> &) {});
+    m.def("annotate_dict_str_int", [](const py::Dict<py::str, int> &) {});
+    m.def("annotate_list_int", [](const py::List<int> &) {});
+    m.def("annotate_set_str", [](const py::Set<std::string> &) {});
+    m.def("annotate_fn", [](const py::Callable<int(py::List<py::str>, py::str)> &) {});
```
Lalaland, line 822: Might be good to add a couple more tests here to be double sure.

I would add 

`m.def("annotate_empty_fn", [](const py::Callable<void()> &) {});` and `m.def("annotate_nested_tuple", [](const py::Tuple<py::Tuple<<>> &) {});`


I think both of those edge cases should work, but good to double check.

Lalaland (approved):
Looks good with just two minor suggestions!

Skylion007:
A colleague just pointed out that something like this could actually be really useful for specifying return types of methods / functions that are Union as well.

virtuald:
The PR says that there are changes requested, but I'm not sure what they are? If there are no objections, would be good to merge this.

rwgk:
I'll run this through global testing, to see if there are any surprises. (That will take 6+ hours.)

rwgk:
Global testing passed (the internal test id is OCL:549398140:BASE:549657321:1689871259083:7fa91c99)

I'll review asap.

include/pybind11/typing.h
```diff
@@ -0,0 +1,94 @@
...
+
+/*
+    The following types can be used to direct pybind11-generated docstrings
+    to have have more explicit types (e.g., `List[str]` instead of `list`).
+    Just use these in place of existing types.
+
+    There is no additional enforcement of types at runtime.
+*/
+
+template <typename... Types>
```
rwgk, line 27: I think this feature will be clearer as such by having the types below in a `typing` sub-namespace.

I'm not sure about (e.g.) `py::typing::Tuple` vs `py::typing::tuple`. Pure Python `typing.tuple` does not exist (I tried with Python 3.11), therefore I'm guessing `py::typing::Tuple` might be less surprising.

include/pybind11/typing.h
```diff
@@ -0,0 +1,94 @@
...
+    using tuple::tuple;
+};
+
+template <typename K, typename V>
+class Dict : public dict {
+    using dict::dict;
+};
+
+template <typename T>
+class List : public list {
```
rwgk, line 38: Could this be `template <typename... Types>` like or `Tuple`?

include/pybind11/typing.h
```diff
@@ -0,0 +1,94 @@
...
+class Callable<Return(Args...)> : public function {
+    using function::function;
+};
+
+PYBIND11_NAMESPACE_BEGIN(detail)
+
+template <typename... Types>
+struct handle_type_name<Tuple<Types...>> {
+    static constexpr auto name
+        = const_name("Tuple[") + concat(make_caster<Types>::name...) + const_name("]");
```
rwgk, line 60: Another thing I learned from the typing expert team mate:

>> Would it be better to have lower-case names (tuple, dict, list, set) in the output?

> It depends on what Python versions you need to support. In Python 3.10 and higher, it is indeed preferred to use the lowercase builtin names. 

That makes me think `#ifdef`s of some sort would be best (to produce `Tuple` for <= 3.9 and `tuple` otherwise), although that adds a bit of clutter until we can drop 3.9 support.

Tying that back to my other comment, `py::typing::tuple` vs `py::typing::Tuple`:

Pure Python `typing.Tuple` will likely be deleted at some point in the future, although IIUC there is no timeline yet.

We cannot reuse `py::tuple` like pure Python reuses `tuple` for typing purposes.

Should we have `py::typing::tuple` or `py::typing::Tuple`? I don't know. It's on the awkward side either way.

I think I'd settle for `py::typing::tuple` (for any Python version), to not look old fashioned in the long run.

include/pybind11/typing.h
henryiii: IMO, `list[float | str]` is fine / best, that's valid for older versions of Python as long as you only use it for typing and not at runtime.

henryiii: Also, the equivalent for `tuple` is `tuple[float | str, ...]`. `tuple[float, str]` is a size 2 tuple with a float as the first item, and a str as the second.

include/pybind11/typing.h
henryiii: For Python 3.7-3.9, you still can use lower case as long as you only use them for typing. In this case, it can only be used for typing, not runtime, so please use the lowercase, un-namespaced names, no ifdefs. I've mostly removed typing.* stuff from libraries that are 3.7+.

rwgk: +1

Note that we have already firmly decided to remove Python 3.6 support asap (as soon as I find a moment).

henryiii: (Yes, rather assuming that in this discussion :) - besides, typing in 3.6 is pretty hard since everything has dropped it a while back)

virtuald: I would add a `class Union<T...> : object` to implement union markers in a more general way.

If there was a `py::typing` namespace it would make more sense to implement such a thing, since `py::Union` is a bit weird.

rwgk: > I would add a `class Union<T...> : object` to implement union markers in a more general way.
> 
> If there was a `py::typing` namespace it would make more sense to implement such a thing, since `py::Union` is a bit weird.

Sounds great to me. I felt strongly already that `py::typing` is the way to go, now even more.

virtuald: I gave it a shot, but I think adding  would be out of scope for this current PR because of two messy issues that came up.

If you add a `Union` type then you get compilation errors when loading the argument because one would need to define a `PYBIND11_OBJECT_COMMON` for it... which, you could add a check function to check if the incoming PyObject is one of the arg types, but the common types also have implicit conversions which wouldn't be caught here. 

Additionally this would require defining a `concat_pipe` version of `concat` which is mostly copy/paste, but it's another thing to maintain.

rwgk (approved):
This looks great to me now.

The only small doubt I still have, do we want `py::typing::Tuple` or `py::typing::tuple`?

I slightly favor `py::typing::Tuple` as we have right now in this PR. It might look old fashioned from a pure Python viewpoint, but in the context of C++ it's a nice way to not have surprise name collisions (between `py::tuple` and `py::typing::tuple`).

virtuald:
I prefer `Tuple` over `tuple`.

rwgk:
> I prefer `Tuple` over `tuple`.

Cool, thanks.

@EthanSteinberg, @Skylion007, @henryiii, @sizmailov does anyone have a strong reason to prefer `py::tuple`?

I'll wait a couple days for responses. If there are no objections, I'll merge this PR.

rwgk:
(I just fixed my previous comment: should have been `py::typing::tuple`).

EthanSteinberg:
@rwgk 

> does anyone have a strong reason to prefer py::typing::tuple?

I would support upper case Tuple to avoid name collisions.

rwgk:
> @rwgk
> 
> > does anyone have a strong reason to prefer py::typing::tuple?
> 
> I would support upper case Tuple to avoid name collisions.

Perfect, thanks, that makes 3 votes for upper case `Tuple` (counting Dustin's), I'll go ahead merge this PR now.

rwgk:
Oh, wait, @Skylion007, merging is blocked because you requested changes. Could you please look again?

Skylion007 (approved):
I also prefer Tuple

include/pybind11/typing.h
```diff
@@ -0,0 +1,97 @@
+/*
+    pybind11/typing.h: Convenience wrapper classes for basic Python types
+    with more explicit annotations.
+
+    Copyright (c) 2016 Wenzel Jakob <wenzel.jakob@epfl.ch>
```
rwgk, line 5: Sorry I forgot to comment on this before, could you please either put your name & current year, or maybe:

```
// Copyright (c) 2023 The pybind Community.
```

(That's what I usually use.)

rwgk:
@virtuald I took the liberty to update the copyright line: 072049a0092e95a7dd32d002d164af35e72eaebd

I'll merge this as soon as I see that the CI is done, then I'll update the smart_holder branch.

PR #4264, opened 2022-10-20
ci(fix): don't label weekly dep updates & ci fixes

henryiii:
## Description

Trying this to reduce changelog effort a bit.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -10,7 +10,10 @@ jobs:
     steps:
 
     - uses: actions/labeler@main
-      if: github.event.pull_request.merged == true
+      if: >
+        github.event.pull_request.merged == true &&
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -10,7 +10,10 @@ jobs:
     steps:
 
     - uses: actions/labeler@main
-      if: github.event.pull_request.merged == true
+      if: >
+        github.event.pull_request.merged == true &&
+        !startsWith(github.event.pull_request.title, 'chore(deps):') &&
+        !startsWith(github.event.plul_request.title, 'ci(fix):'
```
rwgk, line 16: looks like a typo: `plul_request`

henryiii: Yes, it didn’t work on the next PR so I fixed it. It’s working now.

PR #4270, opened 2022-10-21
Revert "feat: add entrypoint for cmake modules dir"

henryiii:
Going to back out on #4258 before we release, we can add it more carefully later, I don't want to get the name and suggested usage wrong.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -103,9 +103,6 @@ Performance and style:
 
 Build system improvements:
 
-* Experimental support for ``cmake.modules`` entrypoint.
-  `#4258 <https://github.com/pybind/pybind11/pull/4258>`_
-
 * Include a pkg-config file when installing pybind11, such as in the Python
   package.
   `#4077 <https://github.com/pybind/pybind11/pull/4077>`_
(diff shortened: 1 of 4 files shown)


Skylion007 (approved):

henryiii:
For a little background: there are two ways to specify where this is; one is by building a list of CMAKE_PREFIX_PATH=dir1;dir2, and the other is by setting a specific path, pybind11_ROOT=dir1 in CMake 3.13+. The latter is better unless you have to support older CMake's. It's also easier to not override or intentionally override. So I'm thinking:

```python
# Better
module_dirs = importlib.metadata.entry_points(group="cmake.root")
packages = {ep.name:importlib.resources.files(ep.load()) for ep in module_dirs}
cmake_args = [f"-{k}_ROOT={v}" for k,v in packages]

# Old CMakes
module_dirs = importlib.metadata.entry_points(group="cmake.root")
dirs = (importlib.resources.files(ep.load()) for ep in module_dirs)
paths_str = ";".join(str(p) for p in dirs)
cmake_arg = [f"-DCMAKE_PREFIX_PATH={paths_str}"]
```

Same structure as before, but we need to pin down the name. `cmake.module` would be for modules, `cmake.prefix` could stay open just in case we want prefixes, and I'm currently experimenting with adding site-packages to cmake_prefix.

PR #4273, opened 2022-10-22
Fix casts to void*

### LINKED ISSUE #4117: [BUG]: void value not ignored as it ought to be
### Required prerequisites

### Problem description

the bind code like this :
```cpp
#include <iostream>
namespace py = pybind11;
using std::cout;
using std::endl;
class Animal {
public:
    virtual ~Animal() { }
    virtual void* go() = 0;
};

class PyAnimal : public Animal {
public:
    /* Inherit the constructors */
    using Animal::Animal;
    void* go() override {
        PYBIND11_OVERRIDE_PURE(
            void*, /* Return type */
            Animal,      /* Parent class */
            go,          /* Name of function in C++ (must match Python name) */
        );
    }
};

PYBIND11_MODULE(example, m) {
    py::class_<Animal, PyAnimal /* <--- trampoline*/>(m, "Animal")
        .def(py::init<>())
        .def("go", &Animal::go);
}
```

use pybind11 2.10.0,get error below:

```
example.cpp: In member function ‘virtual void* PyAnimal::go(int)’:
example.cpp:21:9: error: void value not ignored as it ought to be
   21 |         PYBIND11_OVERRIDE_PURE(
      |         ^~~~~~~~~~~~~~~~~~~~~~
error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

but use pybind 2.9.2 is ok ，how to deal with it ？？

### Reproducible example code

_No response_

Lalaland:
## Description

The current code is broken for safe casts to void* as those are incorrectly converted to casts to void. This fixes the problem.

Fixes #4117, needed for #4125

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1179,11 +1179,9 @@ enable_if_t<cast_is_temporary_value_reference<T>::value, T> cast_safe(object &&)
     pybind11_fail("Internal error: cast_safe fallback invoked");
 }
 template <typename T>
-enable_if_t<std::is_same<void, intrinsic_t<T>>::value, void> cast_safe(object &&) {}
+enable_if_t<std::is_same<void, T>::value, void> cast_safe(object &&) {}
 template <typename T>
-enable_if_t<detail::none_of<cast_is_temporary_value_reference<T>,
-                            std::is_same<void, intrinsic_t<T>>>::value,
-            T>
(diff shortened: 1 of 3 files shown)


rwgk:
Oh ... did you close this because you saw #4274?

henryiii:
I liked the way this just added to an existing tests (+ it works on clang since there's not a new class to complain about missing constructors about).

rwgk:
Outside.
Sounds good to me: take fix from 4274 and the test from here.
I'm not able to do that before tomorrow.

PR #3861, opened 2022-04-11
Cleanup cast_safe<void> specialization

laramiel:
Replace explicit specialization of cast_safe<void> with SFINAE.
It's better for SFINAE cases to cover all type-sets rather than mixing SFINAE and explicit specialization.

Extracted from #3674

PR #4274, opened 2022-10-22
Fix and test for issue #4117

rwgk:
## Description
See #3861 and #4117.

## Suggested changelog entry:

```rst
TODO
```

### DIFF
--- a/tests/test_virtual_functions.cpp
+++ b/tests/test_virtual_functions.cpp
@@ -589,3 +589,32 @@ void initialize_inherited_virtuals(py::module_ &m) {
     m.def("test_gil", &test_gil);
     m.def("test_gil_from_thread", &test_gil_from_thread);
 };
+
+namespace issue4117 { // Broken by PR #3861
+
+class Animal {
+public:
+    virtual ~Animal() {}
+    virtual void *go() = 0;
+};
+
+class PyAnimal : public Animal {
+public:
+    /* Inherit the constructors */
+    using Animal::Animal;
+    /* Trampoline (need one for each virtual function) */
+    void *go() override {
+        PYBIND11_OVERRIDE_PURE(void *, /* Return type */
+                               Animal, /* Parent class */
+                               go,     /* Name of function in C++ (must match Python name) */
+        );
+    }
+};
+
+void bindings(py::module_ m) {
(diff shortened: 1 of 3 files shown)


rwgk:
@laramiel thanks for suggesting the test!

(I wasn't getting very lucky looking under https://github.com/wassimj/Topologic)

rwgk:
@henryiii there is a test here.

rwgk:
@Lalaland @laramiel It's interesting that the CI succeeded for #4273, it simply replaces `intrinsic_t<T>` with `T`.

@Lalaland did you come to the idea that `intrinsic_t` is unnecessary by inspecting the calling code?

I almost think it will be best to drop this PR and reopen and merge #4273 (the only hiccup was the unused ctypes import). Although it might be useful to add in the Python part of the test from here, which has a Python override returning `self` (@laramiel's idea). I believe it is redundant for the purpose of exercising the exact core fix, but it is a different situation that we evidently don't cover otherwise, and is very little extra code.

@henryiii for awareness.

BTW: Amazing coincidence that 3 people worked on fixes in 2 PRs at the same time.

Lalaland:
> @Lalaland <https://github.com/Lalaland> did you come to the idea that
intrinsic_t is unnecessary by inspecting the calling code?

I switched it to T just to revert to the old behavior in 2.9

Your version with remove_cv is more correct (and an improvement over 2.9,
as opposed to just fixing the regression).

I can quickly reopen #4273 with the needed changes (remove_cv and returning
self).

Lalaland:
Don't actually have permission to reopen PRs, so created #4275

tests/test_virtual_functions.cpp
```diff
@@ -230,6 +230,35 @@ inline int test_override_cache(std::shared_ptr<test_override_cache_helper> const
 // are rather long).
 void initialize_inherited_virtuals(py::module_ &m);
 
+namespace issue_4117 { // Broken by PR #3861
+
+class Animal {
+public:
+    virtual ~Animal() {}
```
Skylion007, line 237:
```suggestion
    virtual ~Animal() = default;
```

rwgk:
Closing, superseded by #4275

rwgk:
FWIW / future reference: I rebased and squashed my branch after #4275 was merged, JIC we want to resurrect the test later. I also verified that the test succeeds as expected.

PR #4275, opened 2022-10-22
Fix casts to void*

### LINKED ISSUE #4117: [BUG]: void value not ignored as it ought to be

Lalaland:
## Description

The current code is broken for safe casts to void* as those are incorrectly converted to casts to void. This fixes the problem.

Fixes #4117, needed for #4125

This is a reopened version of #4273 with some minor test changes. Closes #4117.

## Suggested changelog entry:

```rst
* Fix support for safe casts to void* (regression in 2.10.0)
```

### DIFF
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1179,11 +1179,9 @@ enable_if_t<cast_is_temporary_value_reference<T>::value, T> cast_safe(object &&)
     pybind11_fail("Internal error: cast_safe fallback invoked");
 }
 template <typename T>
-enable_if_t<std::is_same<void, intrinsic_t<T>>::value, void> cast_safe(object &&) {}
+enable_if_t<std::is_same<void, T>::value, void> cast_safe(object &&) {}
 template <typename T>
-enable_if_t<detail::none_of<cast_is_temporary_value_reference<T>,
-                            std::is_same<void, intrinsic_t<T>>>::value,
-            T>
+enable_if_t<detail::none_of<cast_is_temporary_value_reference<T>, std::is_same<void, T>>::value, T>
 cast_safe(object &&o) {
     return pybind11::cast<T>(std::move(o));
 }
--- a/include/pybind11/cast.h
+++ b/include/pybind11/cast.h
@@ -1179,9 +1179,9 @@ enable_if_t<cast_is_temporary_value_reference<T>::value, T> cast_safe(object &&)
     pybind11_fail("Internal error: cast_safe fallback invoked");
 }
 template <typename T>
-enable_if_t<std::is_same<void, T>::value, void> cast_safe(object &&) {}
+enable_if_t<std::is_void<T>::value, void> cast_safe(object &&) {}
 template <typename T>
-enable_if_t<detail::none_of<cast_is_temporary_value_reference<T>, std::is_same<void, T>>::value, T>
+enable_if_t<detail::none_of<cast_is_temporary_value_reference<T>, std::is_void<T>>::value, T>
 cast_safe(object &&o) {
     return pybind11::cast<T>(std::move(o));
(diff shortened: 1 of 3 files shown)


henryiii (approved):

Skylion007 (approved):

Skylion007:
Actually, hold on confused whether this or #4274 is correct. This is slightly different: [std::is_void](https://en.cppreference.com/w/cpp/types/is_void) is equivalent to std::remove_cv_t not std::remove_cvref_t. Which one is actually appropiate here? Which cast should be used for void pointer and references @rwgk @Lalaland ?

rwgk:
> Which cast should be used for void pointer and references @rwgk @Lalaland ?

If we go by testing, all tests pass with 4 different variants now, from that viewpoint any of those is fine.

The best way to answer the question above is to create a full set of tests (c v ref), see what compilers actually accept, then make the tests work the way we want them to. (In this case that's probably only a very small effort.)

From an elegance viewpoint, `std::is_void` looks nicest.

Lalaland:
> Actually, hold on confused whether this or https://github.com/pybind/pybind11/pull/4274 is correct. This is slightly different: [std::is_void](https://en.cppreference.com/w/cpp/types/is_void) is equivalent to std::remove_cv_t not std::remove_cvref_t. Which one is actually appropiate here? Which cast should be used for void pointer and references @rwgk @Lalaland ?

References to void are illegal, so std::remove_cv_t and std::remove_cvref_t are equivalent. So std::is_void is preferred.

include/pybind11/cast.h
```diff
@@ -1179,11 +1179,9 @@ enable_if_t<cast_is_temporary_value_reference<T>::value, T> cast_safe(object &&)
     pybind11_fail("Internal error: cast_safe fallback invoked");
 }
 template <typename T>
-enable_if_t<std::is_same<void, intrinsic_t<T>>::value, void> cast_safe(object &&) {}
+enable_if_t<std::is_void<T>::value, void> cast_safe(object &&) {}
```
laramiel, line 1182: Yeah, is std::is_void<T> is more elegant.

laramiel:
Yes, as noted in https://github.com/pybind/pybind11/pull/3861, references to void are not allowed in C++, and `std::is_void<T>` is essentially the shorthand for what I wrote there.

https://en.cppreference.com/w/cpp/types/is_void

rwgk (approved):
Awesome, thanks @Lalaland and @laramiel! I'll merge this now.

PR #4282, opened 2022-10-24
tests: add ABI check

henryiii:
## Description

Adding tests for ABI. Just a very basic implementation at the moment. Didn't add it the tests because we don't run nox on most jobs. Maybe could move it to pytest-virtualenv.


Lalaland:
One suggestion: Have this test assert that the sizeof()s all the major structures are the same

So internals, type_info, instance, etc.

Even better would be to compare the raw bytes and verify those are equivalent. 

(Simply write the bytes of interals to a file or something and then compare the files).

henryiii:
Ahh, can't do an ABI check against and older version of pybind11 and Python 3.11, duh. :)

```
/tmp/tmpco8vd2na/.env/lib/python3.11/site-packages/pybind11/include/pybind11/detail/type_caster_base.h:473:36: error: invalid use of incomplete type ‘PyFrameObject’ {aka ‘struct _frame’}
[232](https://github.com/pybind/pybind11/actions/runs/3316298068/jobs/5477937033#step:25:233)
        473 |                 "  " + handle(frame->f_code->co_filename).cast<std::string>() +
[233](https://github.com/pybind/pybind11/actions/runs/3316298068/jobs/5477937033#step:25:234)
            |                                    ^~
```

> One suggestion

Hmm, interesting, okay, I'll try to add that soon. Won't have too much time for a while, though.

.github/workflows/ci.yml
```diff
@@ -176,6 +176,10 @@ jobs:
       run: pytest tests/extra_setuptools
       if: "!(matrix.runs-on == 'windows-2022')"
 
+    # This tests ABI compatibility
+    - name: ABI test
+      run: pytest tests/extra_abi
```
henryiii, line 181:
```suggestion
      if: matrix.python != '3.11-dev'
      run: pytest tests/extra_abi
```

Issue #4288, opened 2022-10-26
Bug: exception in noexcept `what()` when Python exception contains a surrogate character


Skylion007:
### Discussed in https://github.com/pybind/pybind11/discussions/4287

<div type='discussions-op-text'>

<sup>Originally posted by **TheShiftedBit** October 26, 2022</sup>
By default, Python produces errors when converting encoding `str`s with utf-8 if the `str` contains surrogate characters. This can be disabled by passing `surrogatepass` as a second argument to `.encode()`. Pybind11 has this same behavior with its `str` -> `std::string` conversion. However, the bug is this: if an exception message contains a surrogate character, calling `.what()` on an `error_already_set` with such an exception causes another exception to be thrown, but since `.what()` is `noexcept`, that exception cannot be caught and the program `std::terminate`s.

I'm not sure what the correct behavior regarding surrogate characters is. Perhaps pybind11 should always use `surrogatepass`, perhaps not. However, even if that's not the right choice, it should probably use it during exception handling, or Python exceptions like this are extremely difficult to diagnose.</div>

rwgk:
Without having looked very closely here, it seems similar to what we were running into with the `absl::Status` type, which also reports error messages. We settled on this:

https://github.com/pybind/pybind11_abseil/blob/ef7a738c6472f661aa3a77c228ff0b456fee3f2f/pybind11_abseil/utils_pybind11_absl.cc

That approach could be adopted in `error_already_set`.

Although that throws `error_already_set`, which we don't want to do when we're already inside `error_already_set` I think. I'm not sure how to best error out if there is really no way to recover the error message (some kind of system error or undefined behavior, presumably).

Skylion007:
On MacOS at least: I'm not getting an exception, I am just junk characters as the output. I suspect this may be an issue on Windows though.
```

    def test_surrogate_exception():
        ucs_surrogates_str = "\udcc3"
>       m.test_surrogate_str(ucs_surrogates_str)
\x1b[1m\x1b[31mE       AttributeError: 'str' object has no attribute '\udcc3'\x1b[0m

ucs_surrogates_str = '\udcc3'
```

TheShiftedBit:
You asked for a reproducer, I posted it on the other issue: https://github.com/pybind/pybind11/discussions/4287#discussioncomment-3995628

I temporarily fixed it by adapting your `what()` implementation. I roughly do this:

```cc
[...]

PR #4294, opened 2022-10-28
fix: improve bytes to str decoding error handling

Skylion007:
## Description
* I tried to replicate the bug reported where an error occurred in decoding surrogate pairs #4288. My initial try failed to reproduce the error, but discovered some bugs which I have patched and add tests for.

## Suggested changelog entry:

```rst
* Fixed a bug where ``UnicodeDecodeError`` was not propagated from various ``py::str`` ctors when decoding surrogate utf characters.
```

### DIFF
--- a/tests/test_pytypes.cpp
+++ b/tests/test_pytypes.cpp
@@ -219,6 +219,9 @@ TEST_SUBMODULE(pytypes, m) {
         return py::make_tuple(s1, s2);
     });
 
+    m.def("bytes_to_str_explicit",
+          [](const py::bytes &encoded_str) { return py::str(encoded_str); });
+
     // test_bytes
     m.def("bytes_from_char_ssize_t", []() { return py::bytes{"green", (py::ssize_t) 5}; });
     m.def("bytes_from_char_size_t", []() { return py::bytes{"purple", (py::size_t) 6}; });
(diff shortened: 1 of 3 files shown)


henryiii (approved):

tests/test_pytypes.cpp
```diff
@@ -206,7 +206,11 @@ TEST_SUBMODULE(pytypes, m) {
     m.def("str_from_char_ssize_t", []() { return py::str{"red", (py::ssize_t) 3}; });
     m.def("str_from_char_size_t", []() { return py::str{"blue", (py::size_t) 4}; });
     m.def("str_from_string", []() { return py::str(std::string("baz")); });
+    m.def("str_from_string_input", [](const std::string &stri) { return py::str(stri); });
```
rwgk, line 209: I'd probably make this `str_from_std_string_input` because `string` is so easily misunderstood (in the call site).

rwgk (approved):
Nice!

tests/test_pytypes.cpp
henryiii: I don't think it's critical, it's not too hard to infer. It is slightly better though.

I think this is fine for 2.10.1.

henryiii:
```suggestion
    m.def("str_from_std_string_input", [](const std::string &stri) { return py::str(stri); });
```

tests/test_pytypes.py
```diff
@@ -244,6 +244,20 @@ def __repr__(self):
         m.str_from_string_from_str(ucs_surrogates_str)
 
 
+@pytest.mark.parametrize(
+    "func",
+    [
+        m.str_from_bytes_input,
+        m.str_from_cstr_input,
+        m.str_from_string_input,
```
henryiii, line 252:
```suggestion
        m.str_from_std_string_input,
```

If you did want to change it.

tests/test_pytypes.cpp
Skylion007: I just wanted to be consistent with the out pytest unit tests, but yeah this change is fine.

PR #4297, opened 2022-10-31
Fix & test for issue #4288 (unicode surrogate character in Python exception message).

### LINKED ISSUE #4288: Bug: exception in noexcept `what()` when Python exception contains a surrogate character

rwgk:
## Description
Closes #4288. See there for background.

While at it:

* Also adding `test_error_already_set_message_with_malformed_utf8()` (passes without changes).
* Removing a stray block of code (oversight in #1895).

## Suggested changelog entry:

```rst
Bug fix: Unicode surrogate character in Python exception message leads to process termination in `error_already_set::what()`.
```

### DIFF
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -505,7 +505,23 @@ struct error_fetch_and_normalize {
                 message_error_string = detail::error_string();
                 result = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
             } else {
-                result = value_str.cast<std::string>();
+                // Not using `value_str.cast<std::string>()`, to not potentially throw a secondary
+                // error_already_set that will then result in process termination (#4288).
+                auto value_bytes = reinterpret_steal<object>(
+                    PyUnicode_AsEncodedString(value_str.ptr(), "utf-8", "backslashreplace"));
+                if (!value_bytes) {
+                    message_error_string = detail::error_string();
+                    result = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
+                } else {
+                    char *buffer = nullptr;
+                    Py_ssize_t length = 0;
+                    if (PyBytes_AsStringAndSize(value_bytes.ptr(), &buffer, &length) == -1) {
+                        message_error_string = detail::error_string();
+                        result = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
+                    } else {
+                        result = std::string(buffer, static_cast<std::size_t>(length));
+                    }
+                }
             }
         } else {
             result = "<MESSAGE UNAVAILABLE>";
--- a/include/pybind11/pytypes.h
+++ b/include/pybind11/pytypes.h
@@ -501,9 +501,10 @@ struct error_fetch_and_normalize {
         std::string message_error_string;
         if (m_value) {
             auto value_str = reinterpret_steal<object>(PyObject_Str(m_value.ptr()));
+            const char *message_unavailable_exc = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
             if (!value_str) {
                 message_error_string = detail::error_string();
-                result = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
+                result = message_unavailable_exc;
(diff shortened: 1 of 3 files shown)


henryiii:
PyPy 3.7 Windows gave:

```
MSBuild version 17.3.1+2badb37d1 for .NET Framework
  Checking File Globs
  cross_module_gil_utils.vcxproj -> D:\a\pybind11\pybind11\build2\tests\cross_module_gil_utils.pypy37-pp73-win_amd64.pyd
  cross_module_interleaved_error_already_set.vcxproj -> D:\a\pybind11\pybind11\build2\tests\cross_module_interleaved_error_already_set.pypy37-pp73-win_amd64.pyd
  pybind11_cross_module_tests.vcxproj -> D:\a\pybind11\pybind11\build2\tests\pybind11_cross_module_tests.pypy37-pp73-win_amd64.pyd
  pybind11_tests.vcxproj -> D:\a\pybind11\pybind11\build2\tests\pybind11_tests.pypy37-pp73-win_amd64.pyd
  ------ pybind11_tests.pypy37-pp73-win_amd64.pyd file size: 23571968 (no change)
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(247,5): error MSB8066: Custom build for 'D:\a\pybind11\pybind11\build2\CMakeFiles\c74e89d6f5f1ea108a7628c93f4b5083\pytest.rule;D:\a\pybind11\pybind11\tests\CMakeLists.txt' exited with code -1073741819. [D:\a\pybind11\pybind11\build2\tests\pytest.vcxproj]
```

That's the expected flake, or something else? either way, I can rerun. You can access old runs at the top, there's a selectable run number there.

tests/test_exceptions.cpp
```diff
@@ -105,11 +105,6 @@ struct PythonAlreadySetInDestructor {
     py::str s;
 };
 
-std::string error_already_set_what(const py::object &exc_type, const py::object &exc_value) {
-    PyErr_SetObject(exc_type.ptr(), exc_value.ptr());
-    return py::error_already_set().what();
-}
```
henryiii, line 111: Was this just not being used?

henryiii: Ahh, "Removing a stray block of code", got it.

rwgk: Yeah, I just forgot to remove that block after moving the code into a lambda function.

rwgk:
> PyPy 3.7 Windows gave:
> That's the expected flake, or something else?

Something else. I've seen it for the first time 2-3 months ago. I don't know what's behind it, at all. But clearly not related to this PR.

include/pybind11/pytypes.h
```diff
@@ -501,11 +501,28 @@ struct error_fetch_and_normalize {
         std::string message_error_string;
         if (m_value) {
             auto value_str = reinterpret_steal<object>(PyObject_Str(m_value.ptr()));
+            const char *message_unavailable_exc = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
```
Skylion007, line 504:
```suggestion
            constexpr const char *message_unavailable_exc = "<MESSAGE UNAVAILABLE DUE TO ANOTHER EXCEPTION>";
```

Skylion007 (approved):

Skylion007 (approved):

PR #4284, opened 2022-10-25
Effectively undoing #2999, to see what platforms are affected.

rwgk:
## Description
EXPERIMENTAL, related to #4105, #2999, #4283

## Suggested changelog entry:

```rst

```

PR #4283, opened 2022-10-24
WIP cross module exception ODR experiments

rwgk:
## Description
**Guessing** what could be behind #4105

Local testing passes. Let's see what happens in the CI.

## Suggested changelog entry:

```rst

```

Issue #4105, opened 2022-08-02
[BUG]: ABI incompatibility between 2.9.2 and 2.10


whophil:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### Problem description

I have two pybind11 extension libraries, A and B, where B uses classes from A.

My library A was compiled with pybind11 2.9 and B was compiled with pybind11 2.10. They both declare the same `PYBIND11_INTERNALS_VERSION` of 4, yet they do not seem to work with each other. They do work if I compile both A and B with pybind11 2.9.

Based on discussion with @henryiii on [Gitter](https://gitter.im/conda-forge/conda-forge.github.io?at=62e893c36837563d1cae8fbb), these _should_ be ABI compatible.

### Reproducible example code

```text
N/A
```

henryiii:
From Gitter:

> pybind11 2.10 issue
>
> ocp is a package on conda-forge consisting collection of pybind11 extensions for OpenCascade.
>
> I have another closed source project, Project B, which contains its own pybind11 code, using classes from ocp.
>
> In the past, this all worked fine so long as both projects were built using the same compiler version and pybind11-abiversion, using those from conda-forge-pinning.
>
> Recently, however, the build for Project B started breaking. This seems to be related to the release of pybind11 version 2.10. If I build Project B with pybind11=2.10 the two extension libraries can't talk to each other. If I pin the build of Project B to pybind11=2.9.*, which is the version that the ocp package was built with, then everything works fine.
>
> I thought that this is what thepybind11-abi global pin was supposed to solve. Is it possible that pybind11-abi is not sufficient for this use case? If so, should pybind11 itself be a global pin as well?

rwgk:
[...]

PR #4298, opened 2022-10-31
Define (non-empty) `PYBIND11_EXPORT_EXCEPTION` only under macOS.

rwgk:
## Description
Background: #2999, #4105, #4283, #4284

In a nutshell:

* Only macOS actually needs `PYBIND11_EXPORT_EXCEPTION` (#4284).

* Evidently (#4283), under macOS `PYBIND11_EXPORT_EXCEPTION` does not run the risk of introducing ODR violations,

* but evidently (#4283) under Linux it does, in the presumably rare/unusual situation that `RTLD_GLOBAL` is used.

* Windows does no have the equivalent of `RTLD_GLOBAL`, therefore `PYBIND11_EXPORT_EXCEPTION` has no practical benefit, on the contrary, noisy warning suppression pragmas are needed, therefore it is best left empty.

## Suggested changelog entry:

```rst
`PYBIND11_EXPORT_EXCEPTION` was made non-empty only under macOS. This makes Linux builds safer, and enables the removal of warning suppression pragmas for Windows.
```

### DIFF
--- a/docs/advanced/exceptions.rst
+++ b/docs/advanced/exceptions.rst
@@ -177,9 +177,12 @@ section.
     may be explicitly (re-)thrown to delegate it to the other,
     previously-declared existing exception translators.
 
-    Note that ``libc++`` and ``libstdc++`` `behave differently <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
-    with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI boundaries need to be explicitly exported, as exercised in ``tests/test_exceptions.h``.
-    See also: "Problems with C++ exceptions" under `GCC Wiki <https://gcc.gnu.org/wiki/Visibility>`_.
+    Note that ``libc++`` and ``libstdc++`` `behave differently under macOS
+    <https://stackoverflow.com/questions/19496643/using-clang-fvisibility-hidden-and-typeinfo-and-type-erasure/28827430>`_
+    with ``-fvisibility=hidden``. Therefore exceptions that are used across ABI
+    boundaries need to be explicitly exported, as exercised in
+    ``tests/test_exceptions.h``. See also:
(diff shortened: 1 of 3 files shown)


rwgk:
I think this would be good to have in 2.10.1, too. Nobody complained about ODR violations, or the need to deal with the Windows warnings, but this gets rid of both stumbling stones. We may just be saving a handful of people a few hours of unhappiness.

rwgk:
@oraluben JIC you want to look or comment.

oraluben:
> * ...therefore `PYBIND11_EXPORT_EXCEPTION` has no practical benefit, on the contrary, noisy warning suppression pragmas are needed, therefore it is best left empty.

LGTM

Skylion007:
>> under Linux it does, in the presumably rare/unusual situation that RTLD_GLOBAL is used.

We have used that quite frequently to work around issues with Magnum / Habitat-Sim bindings playing nicely together... To clarify, are you saying this PR would cause issues there? Or fix a potential regression and therefore issues?

rwgk:
> > > under Linux it does, in the presumably rare/unusual situation that RTLD_GLOBAL is used.
> 
> We have used that quite frequently to work around issues with Magnum / Habitat-Sim bindings playing nicely together...

Oh!

> To clarify, are you saying this PR would cause issues there? Or fix a potential regression and therefore issues? I don't think we worry that much about pybind11 exceptions propogating between the two though.

Did you also mix extensions built with different pybind11 versions? In that case I'd expect pretty terrible undefined behavior caused by the ODR violation for `error_already_set`, similar to what you wee in the failures under #4283.

Now thinking: We **really** need to have this in 2.10.1.

In case you used pybind11 from sources, but a mix of source versions between extension: you're really on thin ice with `RTLD_GLOBAL` ever since #2999 was merged. "thin ice" == maybe ODR violation but maybe not.

In case you only used releases: mixing any two release that have #2999 is still on thin ice (I don't have time to look up the release history right now).

But especially after #1895 was merged, it's not just thin ice, ODR violations are a certainty, on Linux, with`RTLD_GLOBAL`, mixing pre-1895 with post-1895.

Skylion007 (approved):

PR #4305, opened 2022-11-01
Use `multiprocessing` `start_method` `"spawn"`

rwgk:
## Description
Quoting from

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

* fork

  ... Note that safely forking a multithreaded process is problematic.

  Available on Unix only. The default on Unix.

The idea to try `"spawn"` is the result of deadlock debugging. After reviewing tracebacks @jbms wrote:

> This actually looks like a general fork thread-safety issue, not related to pybind11. Using fork in a multithreaded program is generally not safe.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,8 @@ Adds docstring and exceptions message sanitizers.
 import contextlib
 import difflib
 import gc
+import multiprocessing
+import os
 import re
 import textwrap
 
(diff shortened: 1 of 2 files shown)


rwgk:
Works! That's one way to get rid of the deadlocks (hoping this will do the trick). But I'll try "forkserver", too, in another PR.

rwgk:
This PR was made obsolete by #4306.

PR #4306, opened 2022-11-01
Use `multiprocessing` `start_method` `"forkserver"`

rwgk:
## Description
Please see https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592 for full background.

Alternative to PR #4305

Quoting from

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

* fork

  ... Note that safely forking a multithreaded process is problematic.

  Available on Unix only. The default on Unix.

The idea to try `"spawn"` or `"forkserver"`is the result of deadlock debugging. After reviewing tracebacks @jbms wrote:

> This actually looks like a general fork thread-safety issue, not related to pybind11. Using fork in a multithreaded program is generally not safe.

This stackoverflow posting provides some details:

* https://stackoverflow.com/questions/6078712/is-it-safe-to-fork-from-within-a-thread

Excerpts:

>  The problem is that fork() only copies the calling thread, and any mutexes held in child threads will be forever locked in the forked child.
...
> The problem is that the order that pthread_atfork handlers are called for unrelated libraries is undefined (it depends on the order that the libraries are loaded by the program). So this means that technically a deadlock can happen inside of a prefork handler because of a race condition.
________

Note that this PR only changes how test_gil_scoped.py is run, therefore a changelog entry is not needed.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,8 @@ Adds docstring and exceptions message sanitizers.
 import contextlib
 import difflib
 import gc
+import multiprocessing
+import os
 import re
 import textwrap
 
@@ -14,6 +16,9 @@ import pytest
 
 # Early diagnostic for failed imports
 import pybind11_tests
+
+if os.name != "nt":
+    multiprocessing.set_start_method("forkserver")
 
 _long_marker = re.compile(r"([0-9])L")
 _hexadecimal = re.compile(r"0x[0-9a-fA-F]+")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,7 +18,13 @@ import pytest
 import pybind11_tests
 
 if os.name != "nt":
-    # https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
+    # Full background: https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
+    # In a nutshell: fork() after starting threads == flakiness in the form of deadlocks.
+    # It is actually a well-known pitfall, unfortunately without guard rails.
+    # "forkserver" is more performant than "spawn" (~9s vs ~13s for tests/test_gil_scoped.py,
+    # visit the issuecomment link above for details).
+    # Windows does not have fork() and the associated pitfall, therefore it is best left
+    # running with defaults.
     multiprocessing.set_start_method("forkserver")
 
 _long_marker = re.compile(r"([0-9])L")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@ import pytest
 import pybind11_tests
 
 if os.name != "nt":
+    # https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
     multiprocessing.set_start_method("forkserver")
 
 _long_marker = re.compile(r"([0-9])L")
--- a/tests/test_gil_scoped.py
+++ b/tests/test_gil_scoped.py
@@ -144,7 +144,7 @@ def _intentional_deadlock():
 
 
 ALL_BASIC_TESTS_PLUS_INTENTIONAL_DEADLOCK = ALL_BASIC_TESTS + (_intentional_deadlock,)
-SKIP_IF_DEADLOCK = True  # See PR #4216
+SKIP_IF_DEADLOCK = False  # See PR #4216
 
 
 def _run_in_process(target, *args, **kwargs):
--- a/tests/test_gil_scoped.py
+++ b/tests/test_gil_scoped.py
@@ -145,7 +145,6 @@ def _intentional_deadlock():
 
 
 ALL_BASIC_TESTS_PLUS_INTENTIONAL_DEADLOCK = ALL_BASIC_TESTS + (_intentional_deadlock,)
-SKIP_IF_DEADLOCK = False  # See PR #4216
 
 
 def _run_in_process(target, *args, **kwargs):
@@ -182,7 +181,7 @@ def _run_in_process(target, *args, **kwargs):
         elif process.exitcode is None:
             assert t_delta > 0.9 * timeout
             msg = "DEADLOCK, most likely, exactly what this test is meant to detect."
-            if SKIP_IF_DEADLOCK or (env.PYPY and env.WIN):
+            if env.PYPY and env.WIN:
                 pytest.skip(msg)
             raise RuntimeError(msg)
         return process.exitcode
--- a/tests/test_gil_scoped.py
+++ b/tests/test_gil_scoped.py
@@ -5,6 +5,7 @@ import time
 
 import pytest
 
+import env
 from pybind11_tests import gil_scoped as m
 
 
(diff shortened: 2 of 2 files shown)


rwgk:
Both failures here are unrelated: test_iostream flake, eigen download problem. Trying again.

rwgk:
Windows PyPy is still giving us DEADLOCK grief:

https://github.com/pybind/pybind11/actions/runs/3379884767/jobs/5612023977

@mattip Do you have an interest and the tools for debugging these deadlocks? I think I've only ever seen this under Windows, but with with different PyPy versions (3.7, 3.8, 3.9).

test_gil_scoped.py on master (or in the 2.10.1 release) is already set up for easy deadlock debugging, all you'd need to do is change these lines:

```diff
-ALL_BASIC_TESTS_PLUS_INTENTIONAL_DEADLOCK = ALL_BASIC_TESTS + (_intentional_deadlock,)
+ALL_BASIC_TESTS_PLUS_INTENTIONAL_DEADLOCK = ALL_BASIC_TESTS # (_intentional_deadlock,)
-SKIP_IF_DEADLOCK = True  # See PR #4216
+SKIP_IF_DEADLOCK = False  # See PR #4216
...
-    timeout = 0.1 if test_fn is _intentional_deadlock else 10
+    timeout = 0.1 if test_fn is _intentional_deadlock else 1000
```

Then run test_gil_scoped.py and wait until it hits a deadlock, then attach a debugger to inspect the stack traces for the two threads waiting on the GIL.

I don't have a suitable Windows machine and don't know anything about debugging under Windows.

rwgk:
@Skylion007 Sorry I still need to put back the `skipif` for Windows PyPy, until we find someone who can help debugging the deadlocks for that platform.

rwgk:
The CI passed now without any issues.

I'm not expecting any deadlocks, except for PyPy Windows, which are now reported as such via `pytest.skip()`.

After this PR is merged, deadlocks for other platforms will show up as CI failures, rather than getting skipped. I hope not seeing such CI failures will confirm over time that we're good. I don't know of another way to get that confirmation quicker. See https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592.

rwgk:
@henryiii could you please help reviewing this tiny PR? I think it will take less than 5 minutes. **I'm really keen to restore more rigorous testing after the 2.10.1 release went out.**

I wasn't completely comfortable effectively silencing the deadlock detection, it was mainly to support the 2.10.1 release.

2 days later I learned that multiprocessing "fork" is generally not safe in combination with threads, IOW we were using it wrong the whole time. The fix is super easy. Getting this merged soon means that we will easily see again when deadlocks are detected. I have high hopes that there will be none. Being sure about it will be very useful.

rwgk:
@henryiii I'd appreciate if you could please comment on this PR.

rwgk:
@eacousineau It would be great if you could help out here as well. It's a really tiny PR. As I wrote above, I'm still very keen to restore more rigorous testing after the 2.10.1 release went out.

tests/conftest.py
```diff
@@ -15,6 +17,10 @@
 # Early diagnostic for failed imports
 import pybind11_tests
 
+if os.name != "nt":
```
EricCousineau-TRI, line 20: nit Can you add a comment as to why we're doing this here?
`git blame` is decent for more context, but I think two small sentences can help explain, e.g.:

```
# We want to shy away from Python's default choice for Linux of "fork", as it can incur
# post-fork resource contention issues (such as thread ids, etc.).
```

I'm vaguely familiar with this, and with our codebase we hit it with either sledgehammers of:
- "yes fork, but pin down threading for fast-but-reckless multiprocessing spawns"
- "no fork, please use spawn, and don't think about it" (I still need to read about `forkserver` and how it may interact with imports like `numpy` or `cv2` that may configure / spawn threads at import)
https://github.com/EricCousineau-TRI/repro/commit/2cff386a1c0dd6e19c0fe3de4d314a62f7dcbe76

tests/test_gil_scoped.py
```diff
@@ -181,7 +182,7 @@ def _run_in_process(target, *args, **kwargs):
         elif process.exitcode is None:
             assert t_delta > 0.9 * timeout
             msg = "DEADLOCK, most likely, exactly what this test is meant to detect."
-            if SKIP_IF_DEADLOCK:
+            if SKIP_IF_DEADLOCK or (env.PYPY and env.WIN):
```
EricCousineau-TRI, line 185: nit (readability) Why not hoist this logic up to the top? e.g.
```py
if env.PYPY or env.WIN:
   SKIP_IF_DEADLOCK = True  # See PR #4216
else:
   SKIP_IF_DEADLOCK = False
```

tests/conftest.py
EricCousineau-TRI, line 20: nit Unclear why you can't allow Windows and use `spawn`?
(On that matter, why not just use `spawn` carte blanch?)

For testing, and given the size / speed of imports (small and fast!) for here, I doubt `forkserver` will give that much of an improvement?

EricCousineau-TRI (changes requested):
Sounds good! First pass, PTAL!

tests/conftest.py
rwgk: ~9s vs ~13s

See added comment.

I think it's best / most future proof to not interfere under Windows: "If it's not broken, don't fix it."

tests/test_gil_scoped.py
rwgk: I removed it completely: 29cea64554c19d5d7bc412d46c0b39d1c9705bd7

tests/conftest.py
rwgk: > Can you add a comment as to why we're doing this here?

Done. I tried to keep it concise.

tests/test_gil_scoped.py
EricCousineau-TRI: Yay, thanks!

tests/conftest.py
EricCousineau-TRI: Gotcha, thanks for the info!

EricCousineau-TRI: Hm... While this and the issue describes the symptoms, is there any way you can link to something that provides readers more context about the root cause of why threads + fork == bad / flaky?
(feel free to dismiss if you wanna add to issue discussion itself, or punt)

EricCousineau-TRI (approved):
Thanks! Remaining comment, but feel free to dismiss / punt / what not.

rwgk:
> is there any way you can link to something that provides readers more context about the root cause of why threads + fork == bad / flaky?

TBH I was going only by the Python documentation (link in PR description: "Note that safely forking a multithreaded process is problematic.") and what @jbms told me. But I just found this:

https://stackoverflow.com/questions/6078712/is-it-safe-to-fork-from-within-a-thread

I'll add that to the PR description here.

Thanks for the review!

rwgk:
The one CI failure is a well-known PyPy flake: expected refcount mismatch.

PR #4316, opened 2022-11-06
[Experiment] `std::type_index` & un/named namespace

rwgk:
## Description
Related to PRs #4313, #4315, #4319

## Suggested changelog entry:

```rst

```

PR #4313, opened 2022-11-04
Minimal reproducer for clash when binding types defined in the unnamed namespace.

rwgk:
## Description
This PR

* works as expected only with **gcc**, but
* fails with any **clang** in otherwise identical environments, and also
* fails with any **MSVC**.

The root cause for the failures is:

* here for clang except macOS https://github.com/pybind/pybind11/blob/ee2b5226295d67b690faddd446a329bb2840a1a8/include/pybind11/detail/internals.h#L120
* here  for MSVC and clang macOS https://github.com/pybind/pybind11/blob/ee2b5226295d67b690faddd446a329bb2840a1a8/include/pybind11/detail/internals.h#L139)

Full details of which platform is or is not using the `__GLIBCXX__` branch: https://github.com/pybind/pybind11/pull/4313#issuecomment-1303436722

**Unknown**: How could this be fixed?

@wjakob Are you aware of this situation already? — Reason for bringing this up now: If the `internals` version is bumped, this would be a good fix to include (assuming the fix might be an ABI break, which isn't clear until we have a fix).

@charlesbeattie FYI It turns out the root cause is not [what we thought it is](https://github.com/pybind/pybind11/blob/ee2b5226295d67b690faddd446a329bb2840a1a8/include/pybind11/detail/typeid.h#L39) when we looked at this back in February.

## Suggested changelog entry:

```rst

```

PR #4319, opened 2022-11-06
Use `std::hash<std::type_index>`, `std::equal_to<std::type_index>` everywhere **except when libc++ is in use**

rwgk:
## Description
This PR was informed by experiments under PR #4316. See there for background.

The only production code change is this:

```diff
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -123,7 +123,8 @@ inline void tls_replace_value(PYBIND11_TLS_KEY_REF key, void *value) {
 // libstdc++, this doesn't happen: equality and the type_index hash are based on the type name,
 // which works.  If not under a known-good stl, provide our own name-based hash and equality
 // functions that use the type name.
-#if defined(__GLIBCXX__)
+#if (PYBIND11_INTERNALS_VERSION <= 4 && defined(__GLIBCXX__))                                     \
+    || (PYBIND11_INTERNALS_VERSION >= 5 && !defined(_LIBCPP_VERSION))
 inline bool same_type(const std::type_info &lhs, const std::type_info &rhs) { return lhs == rhs; }
 using type_hash = std::hash<std::type_index>;
 using type_equal_to = std::equal_to<std::type_index>;
```

Net change for **`PYBIND11_INTERNALS_VERSION 5` only** (default only for Python 3.12+):

* For MSVC this resolves clashes when binding types defined in the unnamed namespace (#4313).

Note for completeness: This PR does **not** help with the template alias problem seen under #4581.

## Suggested changelog entry:

```rst

```


rwgk:
The 2 CI failures are unrelated flakes:

* CI / 🐍 3.6 • macos-latest • x64  —  `Fatal Python error: This thread state must be current when releasing` test_embed/test_interpreter.cpp:285
* CI / 🐍 3 • GCC 11 • C++20• x64  —  `Unable to connect to deb.debian.org:http`

Not rerunning because I want to fold the tests under PR #4313 into this PR anyway.

rwgk:
The 1 CI failure is a test_iostream flake (most common flake we have): CI / 🐍 pypy-3.7 • windows-2022 • x64

include/pybind11/detail/internals.h
```diff
@@ -114,7 +114,7 @@ inline void tls_replace_value(PYBIND11_TLS_KEY_REF key, void *value) {
 // libstdc++, this doesn't happen: equality and the type_index hash are based on the type name,
 // which works.  If not under a known-good stl, provide our own name-based hash and equality
 // functions that use the type name.
-#if defined(__GLIBCXX__)
+#if !defined(__apple_build_version__)
```
Skylion007, line 117: What about non_apple_clang? I am actually not sure we test that in our CI surprisingly, but llvm supports Linux, Windows, and MacOS.

rwgk: Good question.
* Windows clang: I'm trying internal testing, which in the past at least triggered some Windows clang testing, although I never fully understood what exactly.
* Non-Apple macOS clang: Could someone please help with testing?

But:
* I'm not in a rush to merge this PR.
* I just want this to be in the pipeline if and when we decide to increment `PYBIND11_INTERNALS_VERSION`.

FWIW, initially I thought the problem is elsewhere and it may be an easy fix. I didn't anticipate that it would go in the direction you see now.

Skylion007: @rwgk Is there an issue opened up on the LLVM repo for this bug btw?

rwgk: > @rwgk Is there an issue opened up on the LLVM repo for this bug btw?

Not to my knowledge.

I think it'll be a couple hours work (at least) to create a minimal reproducer for them, assuming that they will not be able to work with a pybind11-based reproducer.

rwgk:
To explain 8f0c28e86c3776ed35050c5f4aeafa0789e5e3a1:

Internally we have tests under Ubuntu and Debian that use `-stdlib=libc++ -nostdlib++`. Those broke.

The commit is backtracking to a more conservative approach: instead of "all but" the conditions for using `std::hash<std::type_index>` & `std::equal_to<std::type_index>` are now back to "only these".

The current version passes testing here and limited internal testing (global testing takes more time), but I don't think "only these" is ideal, as the situation for MSVC shows: it went unnoticed that newer MSVC versions have a working stl.

rwgk:
Commit 883cadfd9b5ec752f39d5cafd422ba62b4dbcecc restores the "all but" approach, and passes limited internal testing and the CI here.

The idea is: if we encounter other non-cooperating stl implementations, our unit tests will fail and we can then "opt out" those as needed, by tweaking the `#ifdef`.

rwgk:
This PR passed Google global testing (internal testing ID OCL:486686530:BASE:486954198:1667923752339:e2b79b0f).

This includes https://github.com/deepmind/mujoco internal testing with Windows clang 15.0.3, Debian libc++, Ubuntu libc++, which are not covered in the GitHub CI here.

rwgk:
The CI failure is a rare flake (PyPy refcount mismatch). Ignoring.

include/pybind11/detail/internals.h
rwgk: > What about non_apple_clang Clang15, Clang16? I am actually not sure we test that in our CI surprisingly, but llvm supports Linux, Windows, and MacOS.

https://github.com/pybind/pybind11/pull/4316#issuecomment-1310444102

That platform is actually the worst: `std::type_index` does not work as desired, and cross-module exception translation is broken, even if the exceptions have default visibility (IOW "are exported").

rwgk:
The CI failure is a very common flake (test_iostream). Ignoring.

include/pybind11/detail/internals.h
ax3l: > Non-Apple macOS clang: Could someone please help with testing?

We could test this using the conda-forge compilers (they are vanilla clang on macOS).

rwgk: >> Non-Apple macOS clang: Could someone please help with testing?
>
> We could test this using the conda-forge compilers (they are vanilla clang on macOS). This also work in CI.

In the meantime I solved this (but forgot to note that here):

https://github.com/pybind/pybind11/pull/4326

ax3l:
Very cool!

Btw, do we know any libc++ developers that we could report this to?

rwgk:
> Very cool!
> 
> Btw, do we know any libc++ developers that we could report this to?

Back in November I asked our experts and after some back and forth there seemed to be a consensus that the libstdc++ behavior is buggy. I filed an internal bug but there was no related activity since.

rwgk:
> > there seemed to be a consensus that the libstdc++ behavior is buggy.
> 
> I think you mean `libc++` (from LLVM) here :)

Yes, sorry I didn't get back to correcting that here before. Corrected now.

> We could open a public bug then or do you want to ping people on it first? :)

I doubt a ping will do much. The other reason I'm hesitating: even if someone implements a fix today, it'll take many years before we/users can fully rely on it.

What I had (slightly edited):
______________________________

I found a behavior difference between gcc and clang for this situation:

2 Translation Units:

test_unnamed_namespace_a.cpp:
```
namespace {
struct any_struct {};
} // namespace
```

test_unnamed_namespace_b.cpp:
```
namespace {
struct any_struct {};
} // namespace
```

These 2 structs compare equal with clang (but NOT gcc) here:

* https://github.com/pybind/pybind11/blob/cf7d2e6f15be7622515a77466e84bd9414737680/include/pybind11/detail/internals.h#L119-L120
* https://github.com/pybind/pybind11/blob/cf7d2e6f15be7622515a77466e84bd9414737680/include/pybind11/detail/internals.h#L144-L145

```
using type_hash = std::hash<std::type_index>;
using type_equal_to = std::equal_to<std::type_index>;
...
template <typename value_type>
using type_map = std::unordered_map<std::type_index, value_type, type_hash, type_equal_to>;
```

Reproducer: https://github.com/pybind/pybind11/pull/4319 (**patch `#ifdef` in line 117 of internals.h!**)

* Is this intended behavior (for clang)?
* Is there a way to make clang behave like gcc?

Skylion007:
@ax3l @rwgk Have we opened a public bug on LLVM yet? They are usually responsive. Even if it does not solve our issue today, we will thank our future selves when it does eventually make into our oldest supported Clang.

rwgk:
@ax3l @henryiii @Lalaland @Skylion007 @wjakob

What's your vote on this PR?

The production code change is really minimal. It boils down to: for Python 3.12+ we're fixing an MSVC problem (binding types defined in the unnamed namespace).

The only downside I see: it's a little more involved to reason about pybind11 behavior under Windows, 1. simply because we're makaing a change and 2. conditional indirectly on the Python version.

Lalaland:
I think this looks good to me!

henryiii (approved):
I'm okay with it.

Skylion007:
Are we sure it's a LLVM or could it be a GCC bug btw?

Skylion007 (approved):

rwgk:
> Are we sure it's a LLVM or could it be a GCC bug btw?

I looked at the internal discussion from November last year again. It concludes with this sentence:

> So, I guess clang+libc++ could improve the implementation that uses strcmp.

This came from someone apparently very familiar with clang and libc++, based on other things they wrote (including pointers to libc++ sources).

That makes me think: yes, it's almost certainly a libc++ issue.

rwgk:
> The downside is that we don't have any `PYBIND11_TEST_OVERRIDE` anymore in GHA. Do we want to add one back, but decoupled from `PYBIND11_INTERNALS_VERSION`?

I didn't get any feedback to this question, so I decided to try d0276c01cd9e0a24aafe74c698d666908a31a397. The 3 CI logs look good (https://github.com/pybind/pybind11/actions/runs/4762077158: 3.9___MSVC_2022_C++20___x64.txt, macos-latest___brew_install_llvm.txt, 3___GCC_12___C++20__x64.txt). I'll break out the ci.yml change as a separate PR.

The Python 3.12 failure (Upstream workflow) [is expected](https://github.com/python/cpython/issues/102594) and needs special attention (separate PR).

rwgk:
Thanks for the reviews @henryiii, @Lalaland, @Skylion007!

With #4635 merged today, I think this PR is also good to get merged. I'll go ahead now.

The only non-reviewed tweak is a one-character change for Python 3.12 testing (`skipif` for alpha 7 instead of 6): 1b4a508d5e5d44d74664d8fa06c27cea568ff36a

With that Python 3.12 testing passes.

PR #4307, opened 2022-11-03
Scratch PR: clone of #4293 that could be merged immediately without disrupting anything

rwgk:
## Description
Scratch PR for testing.

Rebased on current master.

## Suggested changelog entry:

```rst

```

PR #4329, opened 2022-11-12
WIP: Conversions between Python's native (stdlib) enum and C++ enums.

rwgk:
## Description
Please ignore for now.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/native_enum.h
+++ b/include/pybind11/native_enum.h
@@ -6,6 +6,7 @@
 
 #include "detail/common.h"
 #include "detail/native_enum_data.h"
+#include "detail/type_caster_base.h"
 #include "cast.h"
 
 #include <limits>
@@ -32,7 +33,7 @@ public:
                 "pybind11::native_enum<...>(\"" + enum_name_encoded
                 + "\") is already registered as a `pybind11::enum_` or `pybind11::class_`!");
         }
-        if (detail::native_enum_type_map::get().count(enum_type_index)) {
+        if (cross_extension_shared_states::native_enum_type_map::get().count(enum_type_index)) {
             pybind11_fail("pybind11::native_enum<...>(\"" + enum_name_encoded
                           + "\") is already registered!");
         }
--- a/include/pybind11/native_enum.h
+++ b/include/pybind11/native_enum.h
@@ -19,14 +19,17 @@ public:
     native_enum(const object &scope, const char *name) : m_scope(scope), m_name(name) {}
 
     /// Export enumeration entries into the parent scope
-    native_enum &export_values() { return *this; }
+    native_enum &export_values() {
+        m_export_values = true;
+        return *this;
+    }
 
     /// Add an enumeration entry
     native_enum &value(char const *name, Type value, const char *doc = nullptr) {
-        if (doc) {
-            // IGNORED.
-        }
         m_members.append(make_tuple(name, static_cast<Underlying>(value)));
+        if (doc) {
+            m_docs.append(make_tuple(name, doc));
+        }
         return *this;
     }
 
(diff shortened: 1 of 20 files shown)


include/pybind11/native_enum.h
```diff
@@ -0,0 +1,67 @@
...
+    using Underlying = typename std::underlying_type<Type>::type;
+    // Scalar is the integer representation of underlying type
+    using Scalar = detail::conditional_t<detail::any_of<detail::is_std_char_type<Underlying>,
+                                                        std::is_same<Underlying, bool>>::value,
+                                         detail::equivalent_integer_t<Underlying>,
+                                         Underlying>;
+
+    template <typename... Extra>
+    native_enum(const handle &scope, const char *name, const Extra &.../*extra*/)
+        : m_scope(reinterpret_borrow<object>(scope)), m_name(name) {
```
Skylion007, line 26: If you are storing the scope anyway, why not allow it take an object in a delegating ctor?

rwgk: I'm only at the stage of trying to get the basic mechanisms in place. I'm thinking in the final version I may need to make `native_enum` a function, to avoid the dtor trick (will kill the process if something goes wrong).

At this point I'm emulating the existing `enum_` interface as much as possible.

Skylion007:
@rwgk There has been a lot of discussion about doing this and other limitation with the enum class for a while btw: see https://github.com/pybind/pybind11/issues/2332 especially and https://github.com/pybind/pybind11/issues/1177 (for a long standing enum issue). Also see this gist from #2332 https://gist.github.com/anntzer/96f27c0b88634dbc61862d08bff46d10

rwgk:
> @rwgk There has been a lot of discussion about doing this and other limitation with the enum class for a while btw: see especially #2332 and #1177 (for a long standing enum issue). Also see this gist from #2332 https://gist.github.com/anntzer/96f27c0b88634dbc61862d08bff46d10 . 2332 has a lot of ideas for how to do it properly and how people do it in their downstream project.

Thanks for the pointers. I was vaguely aware of #2332 but didn't actually look very much...

... because I basically don't have a choice: the `native_enum` is for PyCLIF compatibility. We have hundreds (hand-waving estimate) of extensions that use a `native_enum` equivalent. Global-test failures due to `pybind11:enum_` vs native enum mismatches are one of the biggest remaining blockers for us.

Note that I'm not aiming to replace `pybind11::enum_`, but to add an alternative option.

I'll focus on making the `native_enum` implementation complete for all corner cases, then with that in hand

1. the PyCLIF-pybind11 integration has one blocker less,
2. we can step back and review what we want to do long term.

Skylion007:
> > @rwgk There has been a lot of discussion about doing this and other limitation with the enum class for a while btw: see especially #2332 and #1177 (for a long standing enum issue). Also see this gist from #2332 https://gist.github.com/anntzer/96f27c0b88634dbc61862d08bff46d10 . 2332 has a lot of ideas for how to do it properly and how people do it in their downstream project.
> 
> Thanks for the pointers. I was vaguely aware of #2332 but didn't actually look very much...
> 
> ... because I basically don't have a choice: the `native_enum` is for PyCLIF compatibility. We have hundreds (hand-waving estimate) of extensions that use a `native_enum` equivalent. Global-test failures due to `pybind11:enum_` vs native enum mismatches are one of the biggest remaining blockers for us.
> 
> Note that I'm not aiming to replace `pybind11::enum_`, but to add an alternative option.
> 
> I'll focus on making the `native_enum` implementation complete for all corner cases, then with that in hand
> 
>     1. the PyCLIF-pybind11 integration has one blocker less,
> 
>     2. we can step back and review what we want to do long term.

Now that we only support Python >3.5, I don't see why we can't implement some of those changes to replace the old enum if needed.

Skylion007:
Also, some of the proposed methods don't require an ABI break. The Python Enum objects are singletons so we just need to generate and map between them in the caster somehow.

rwgk:
> Also, some of the proposed methods don't require an ABI break. The Python Enum objects are singletons so we just need to generate and map between them in the caster somehow.

I think it needs to go into "something like `internals`", but not necessarily `internals`.

@wjakob I think it would be great to have #4293 (with the tiny #4307 tweak) merged, for the `get_internals_state_dict()` function, which I could reuse for something like `get_native_enums_from_state_dict()`. (Maybe a better name for the helper function would be `get_state_dict()`.)

rwgk:
This PR currently includes this code:

```
        // Intentionally leak Python reference.
        detail::get_native_enum_type_map()[data.enum_type_index] = py_enum.release().ptr();
```

This will only matter for embedding.

This small experiment shows that it currently does not even matter for embedding:

https://github.com/rwgk/issues/tree/0e7ebbda7494a3d1165aaf45523fad4c506199f7/pybind11/initialize_finalize_loop

That experiment shows:

* A simple initialize-finalize loop does not leak with Python 3.10.
* But if `class_<>(...);` is called it leaks substantially.

`top` output files are included under the link above.

Conclusion: Any effort trying to avoid the "Intentional leak" above would have no tangible benefit. There is a bigger general problem that would need to be solved first, and probably that solution would make it straightforward to also avoid this documented leak.

PR #4330, opened 2022-11-13
scoped_interpreter. overloaded constructor: PyConfig param

arman-novikov:
## Description

In some cases it is convenient to have an opportunity to construct a scoped_interpreter with a custom PyConfig instance supplied.

```rst
PyConfig my_config{};
//...
pybind11::scoped_interpreter{my_config}
```
## Suggested changelog entry:

```rst
* feat: scoped_interpreter overloaded ctor: PyConfig param
* tests: Add program dir to path, Custom PyConfig, Custom PyConfig with argv
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,7 +9,7 @@ platform:
 - x86
 environment:
   matrix:
-  - PYTHON: 36
+  - PYTHON: 38
     CONFIG: Debug
 install:
 - ps: |
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,7 +9,7 @@ platform:
 - x86
 environment:
   matrix:
-  - PYTHON: 38
+  - PYTHON: 36
     CONFIG: Debug
 install:
 - ps: |
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -55,7 +55,7 @@
         PYBIND11_TOSTRING(name), PYBIND11_CONCAT(pybind11_init_impl_, name));                     \
     void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_                                \
                                                & variable) // NOLINT(bugprone-macro-parentheses)
-#define PYBIND11_PYCONFIG_SUPPORT_PY_VERSION (0x030B0000)
+#define PYBIND11_PYCONFIG_SUPPORT_PY_VERSION (0x03080000)
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -99,18 +99,18 @@ inline void initialize_interpreter(PyConfig *config,
                                    const char *const *argv = nullptr,
                                    bool add_program_dir_to_path = true) {
     PyStatus status = PyConfig_SetBytesArgv(config, argc, const_cast<char *const *>(argv));
-    if (PyStatus_Exception(status)) {
+    if (PyStatus_Exception(status) != 0) {
         // A failure here indicates a character-encoding failure or the python
         // interpreter out of memory. Give up.
         PyConfig_Clear(config);
-        throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
-                                                          : "Failed to prepare CPython");
+        throw std::runtime_error(PyStatus_IsError(status) != 0 ? status.err_msg
+                                                               : "Failed to prepare CPython");
     }
     status = Py_InitializeFromConfig(config);
-    if (PyStatus_Exception(status)) {
+    if (PyStatus_Exception(status) != 0) {
         PyConfig_Clear(config);
-        throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
-                                                          : "Failed to init CPython");
+        throw std::runtime_error(PyStatus_IsError(status) != 0 ? status.err_msg
+                                                               : "Failed to init CPython");
     }
     if (add_program_dir_to_path) {
         PyRun_SimpleString("import sys, os.path; "
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -95,13 +95,13 @@ inline void precheck_interpreter() {
 }
 
 #if PY_VERSION_HEX >= 0x030B0000
-class config_guard
-{
+class config_guard {
 public:
-    config_guard(PyConfig& config): m_config{config}{}
-    ~config_guard(){PyConfig_Clear(&m_config);}
+    config_guard(PyConfig &config) : m_config{config} {}
+    ~config_guard() { PyConfig_Clear(&m_config); }
+
 private:
-    PyConfig& m_config;
+    PyConfig &m_config;
 };
 
 /*!
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -264,6 +264,25 @@ public:
         initialize_interpreter(init_signal_handlers, argc, argv, add_program_dir_to_path);
     }
 
+#if PY_VERSION_HEX >= 0x030B0000
+    explicit scoped_interpreter(const PyConfig& config,
+                                bool add_program_dir_to_path = true) {
+        if (Py_IsInitialized() != 0) {
+            pybind11_fail("The interpreter is already running");
+        }
+        PyStatus status = Py_InitializeFromConfig(&config);
+        if (PyStatus_Exception(status)) {
+            throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg : "Failed to init CPython");
+        }
+        if (add_program_dir_to_path) {
+            PyRun_SimpleString("import sys, os.path; "
+                               "sys.path.insert(0, "
+                               "os.path.abspath(os.path.dirname(sys.argv[0])) "
+                               "if sys.argv and os.path.exists(sys.argv[0]) else '')");
+        }
+    }
+#endif
+
     scoped_interpreter(const scoped_interpreter &) = delete;
     scoped_interpreter(scoped_interpreter &&other) noexcept { other.is_valid = false; }
     scoped_interpreter &operator=(const scoped_interpreter &) = delete;
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -55,7 +55,6 @@
         PYBIND11_TOSTRING(name), PYBIND11_CONCAT(pybind11_init_impl_, name));                     \
     void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_                                \
                                                & variable) // NOLINT(bugprone-macro-parentheses)
-#define PYBIND11_PYCONFIG_SUPPORT_PY_VERSION_HEX (0x03080000)
 
 PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
 PYBIND11_NAMESPACE_BEGIN(detail)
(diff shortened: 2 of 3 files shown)


include/pybind11/embed.h
```diff
@@ -262,14 +262,14 @@ class scoped_interpreter {
...
+    explicit scoped_interpreter(const PyConfig &config, bool add_program_dir_to_path = true) {
         if (Py_IsInitialized() != 0) {
             pybind11_fail("The interpreter is already running");
         }
         PyStatus status = Py_InitializeFromConfig(&config);
         if (PyStatus_Exception(status)) {
-            throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg : "Failed to init CPython");
+            throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
+                                                              : "Failed to init CPython");
         }
```
Skylion007, line 273: This code appears to be entirely duplicated from the end on initialize. Furthermore, for consistency py::initialize should also be able to take a config on recent python versions. Therefore, I would suggest abstracting this into an overload for initialize(const PyConfig & config) that is only built when the python is recent enough. That would reduce code duplicate and make this a bit cleaner without changing the API that much.

Skylion007 (changes requested):
Also, it would be great to have a test added to our test suite as well.

include/pybind11/embed.h
arman-novikov: Thank you for your feedback. I find your suggestions reasonable.

However, I would rather get a sign of approval from the maintainers that they [want/do not mind] to have the feature introduced into the project before I rush into refactoring and extending tests. Is it at all possible to find out their opinion?

Skylion007: Seem reasonable to me, @rwgk @henryiii thoughts?

rwgk:
I think it'll be great to support passing `PyConfig`, and ideally to clean up:

The current implementation of `initialize_interpreter()` was merged in a rush despite me being "really concerned": https://github.com/pybind/pybind11/pull/3923#discussion_r915224856

I stopped looking when I saw the merge commit back then while working on my feedback, but looking again now, I see that the `argc` and `argv` arguments are completely and silently ignored in the Python 3.11 branch, which is obviously error prone, especially because the function documentation was not updated when the Python 3.11 branch was added.

I think a good start for this PR would be to cleanly split the two implementations, which will make the `argc` / `argv` situation immediately obvious, then to figure out how to best design passing `PyConfig`

* without duplicating code, and
* without silently (runtime or documentation) ignoring arguments, depending on the Python version.

It's not immediately obvious to me how to provide an API that works intuitively for Python < 3.11 and >= 3.11. Can it be done in a reasonable way without expecting

```
#if PY_VERSION_HEX < 0x030B0000
...
#else
...
#endif
```

constructs in user code? @arman-novikov what do you think?

rwgk:
Sorry our codesearch hover-over tooling failed me: it doesn't highlight the `argc`, `argv` in the 3.11 branch for some reason (I just double-checked). I only realized when searching in a normal editor.

That changes my thinking a lot:

Now I'm thinking we need all arguments for >= 3.11, rough start:

```
    explicit scoped_interpreter(bool init_signal_handlers = true,
                                int argc = 0,
                                const char *const *argv = nullptr,
                                bool add_program_dir_to_path = true
                                PyConfig *config = nullptr) {
```

Although `init_signal_handlers` is redundant if `config` is passed. Not sure how to best deal with that.

Also not sure what to do about `PyConfig` for < 3.11, maybe like this? (Note: that code was removed already when we dropped Python 2 support.)

https://github.com/pybind/pybind11/blob/ec24786eab4d170f21536c92422a59959e14dde4/include/pybind11/pybind11.h#L1234-L1238

rwgk (commented):
I really like the direction this is going.

include/pybind11/embed.h
```diff
@@ -88,6 +88,42 @@ inline wchar_t *widen_chars(const char *safe_arg) {
...
+}
+
+#if PY_VERSION_HEX >= 0x030B0000
+class config_guard
+{
+public:
+    config_guard(PyConfig& config): m_config{config}{}
+    ~config_guard(){PyConfig_Clear(&m_config);}
+private:
+    PyConfig& m_config;
```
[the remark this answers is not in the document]
arman-novikov: That is an excellent move. I will introduce it asap.

include/pybind11/embed.h
```diff
@@ -88,6 +88,42 @@ inline wchar_t *widen_chars(const char *safe_arg) {
...
+inline void precheck_interpreter() {
+    if (Py_IsInitialized() != 0) {
+        pybind11_fail("The interpreter is already running");
+    }
+}
+
+#if PY_VERSION_HEX >= 0x030B0000
+class config_guard {
+public:
+    config_guard(PyConfig &config) : m_config{config} {}
```
Skylion007, line 100: delete copy ctor, and copy assignment.

include/pybind11/embed.h
```diff
@@ -88,6 +88,42 @@ inline wchar_t *widen_chars(const char *safe_arg) {
 
 PYBIND11_NAMESPACE_END(detail)
 
+inline void precheck_interpreter() {
```
Skylion007, line 91: This should DEFINITELY be in detail namespace as well.

rwgk:
BTW

While looking at [the documentation for PyConfig](https://docs.python.org/3/c-api/init_config.html) I noticed "New in version 3.8."

I'm wondering, would it make sense / should we try to change the condition to `#if PY_VERSION_HEX >= 0x03080000`?

But what I'd probably do:
* get it to pass with 3.11 first,
* only then try how far back we can go.

arman-novikov:
@Skylion007, the test has been introduced in a [commit](https://github.com/pybind/pybind11/pull/4330/commits/a6fea8b3ca34ba5280e7f30b1523199bd7134b7a). I was forced to set up a newer version of python (3.8) in CI. The CI change has been introduced in [commit](https://github.com/pybind/pybind11/pull/4330/commits/1f6d7952ec35733c30ac07dc72faef8ab4f5bd92). Would you please review the change?

arman-novikov:
> BTW
> 
> While looking at [the documentation for PyConfig](https://docs.python.org/3/c-api/init_config.html) I noticed "New in version 3.8."
> 
> I'm wondering, would it make sense / should we try to change the condition to `#if PY_VERSION_HEX >= 0x03080000`?
> 
> But what I'd probably do:
> 
>     * get it to pass with 3.11 first,
> 
>     * only then try how far back we can go.

It seems python 3.8 was the first version with PyConfig support.

.appveyor.yml
```diff
@@ -9,7 +9,7 @@ platform:
 - x86
 environment:
   matrix:
-  - PYTHON: 36
+  - PYTHON: 38
```
rwgk, line 12: > I was forced to set up a newer version of python (3.8) in CI.

This sounds like a misunderstanding. Whatever is done in this PR should not break the existing test (because it would also break users). Could you please undo this change?

(I think dropping Python 3.6 support is past due, but that's a separate step.)

include/pybind11/embed.h
```diff
@@ -55,6 +55,7 @@
         PYBIND11_TOSTRING(name), PYBIND11_CONCAT(pybind11_init_impl_, name));                     \
     void PYBIND11_CONCAT(pybind11_init_, name)(::pybind11::module_                                \
                                                & variable) // NOLINT(bugprone-macro-parentheses)
+#define PYBIND11_PYCONFIG_SUPPORT_PY_VERSION (0x03080000)
```
rwgk, line 58: `PYBIND11_PYCONFIG_SUPPORT_PY_VERSION_HEX` (for clarity)

include/pybind11/embed.h
```diff
@@ -261,6 +283,13 @@ class scoped_interpreter {
         initialize_interpreter(init_signal_handlers, argc, argv, add_program_dir_to_path);
     }
 
+#if PY_VERSION_HEX >= PYBIND11_PYCONFIG_SUPPORT_PY_VERSION
+    explicit scoped_interpreter(const PyConfig &config, bool add_program_dir_to_path = true) {
```
rwgk, line 287: I think this needs to become:
```
    explicit scoped_interpreter(PyConfig *config,
                                int argc = 0,
                                const char *const *argv = nullptr,
                                bool add_program_dir_to_path = true) {
````

I'm not sure how this will shake out best.

I'm thinking the `scoped_config` will not actually be needed.

I'd definitely split out `initialize_interpreter_pre_pyconfig()` for clarity and safety (unused variables in one branch or another), in the `detail` namespace.

Then there would be two `initialize_interpreter()` overloads:

```
#if PY_VERSION_HEX >= PYBIND11_PYCONFIG_SUPPORT_PY_VERSION_HEX
inline void initialize_interpreter(PyConfig *config,
                                   int argc = 0,
                                   const char *const *argv = nullptr,
                                   bool add_program_dir_to_path = true) {
    ... implementation here ...
}
#endif

inline void initialize_interpreter(bool init_signal_handlers = true,
                                   int argc = 0,
                                   const char *const *argv = nullptr,
                                   bool add_program_dir_to_path = true) {
#if PY_VERSION_HEX < PYBIND11_PYCONFIG_SUPPORT_PY_VERSION_HEX
    detail::initialize_interpreter_pre_pyconfig(
        init_signal_handlers, argc, argv, add_program_dir_to_path);
#else
    PyConfig config;
    PyConfig_InitIsolatedConfig(&config);
    config.isolated = 0;
    config.use_environment = 1;
    config.install_signal_handlers = init_signal_handlers ? 1 : 0;
    initialize_interpreter(&config, argc, argv, add_program_dir_to_path);
#endif
}
```

Then the old `scoped_interpreter` constructor can stay as-is, and the new one will simply be:

```
    explicit scoped_interpreter(PyConfig *config,
                                int argc = 0,
                                const char *const *argv = nullptr,
                                bool add_program_dir_to_path = true) {
        initialize_interpreter(config, argc, argv, add_program_dir_to_path);
    }
```

Note that I'd choose to make this `PyConfig *config` instead of `PyConfig &config`, so that it is more obvious that the `config` is changed in place (`PyConfig_Clear()`). It seems very unlikely that anyone would want `config` to be `const`, although that's only a best guess. When you're done with this PR (CI passes), we should run it by Victor Stinner (not tagging him just now intentionally).

File: include/pybind11/embed.h [hunk 8]
```diff
@@ -261,6 +281,16 @@ class scoped_interpreter {
         initialize_interpreter(init_signal_handlers, argc, argv, add_program_dir_to_path);
     }
 
+#if PY_VERSION_HEX >= PYBIND11_PYCONFIG_SUPPORT_PY_VERSION_HEX
+    explicit scoped_interpreter(PyConfig *config,
+                                int argc = 0,
+                                const char *const *argv = nullptr,
+                                bool add_program_dir_to_path = true) {
+        detail::precheck_interpreter();
+        detail::initialize_interpreter(config, argc, argv, add_program_dir_to_path);
```
rwgk, line 290: I think it will be most useful

* to move this `initialize_interpreter()` overload out of the `detail` namespace,
* and the `detail::precheck_interpreter();` call to the very top of that overload, just like for the other overload.

The end result will be that the pair of constructors and the pair of `initialize_interpreter()` overloads each are almost identical, except for the first argument (`bool` vs `PyCconfig`).

It's very little extra work, it looks great already.

rwgk:
The force push was for `git rebase master`.

I added one simple commit, which was much easier than explaining here: b94d6ba965248b266a62bfb9017b0458d4467b29

I just looked, our existing test coverage isn't complete, unless I'm overlooking something, `add_program_dir_to_path` is not exercised at all. Covering this systematically for both pre-`PyConfig` and `PyConfig` would be great, but I see that as optional, covering that just for the new `PyConfig` overload would be sufficient for this PR I think.

This existing test case looks like a good starting point:

https://github.com/pybind/pybind11/blob/9c18a74e377dece2f2acd22c2c6e63ecb2a59c77/tests/test_embed/test_interpreter.cpp#L363-L383

Could you please try to morph that into an additional test case, with this at the core?

```
        py::scoped_interpreter argv_scope(&config, 1, argv, true);
```

I don't know how to tricky it is to assert that `add_program_dir_to_path` is handled properly. I hope it's within reason, I'd definitely try.

rwgk:
LGTM but

* 🐍 3 • ICC latest • x64 (using Python 3.8)
* 🐍 3 • almalinux:9 • x64 (using Python 3.9)

now produce similar errors. Do you have ideas already?

I think the segfaults could be side-effects of the failed `REQUIRE( get_sys_path_size() == sys_path_default_size + 1 )` and are possibly resolved when the `REQUIRE` is fixed.

rwgk:
Interesting, I see you changed the expecations from `== default + 1` to `> default`.

Did you look why it is not `default + 1` in the two failure cases?

arman-novikov:
> 

it seems a value remains in sys path left by the previous call of scoped_interpreter ctor

rwgk:
> it seems a value remains in sys path left by the previous call of scoped_interpreter ctor

Not good ... but that sounds like a problem in CPython.

From the previous run it seems to be a problem only for Python 3.8 and 3.9. Maybe only some platforms?

* But — I don't want to turn this into a science project.
* But also — I want to keep the test as strict as possible. 

What do you think about a simple "future-strict" approach? Something like:

```
#if PY_VERSION_HEX < 0x030A0000
    // It seems a value remains in sys.path left by the previous call of scoped_interpreter ctor.
    REQUIRE(get_sys_path_size() > sys_path_default_size);
#else
    REQUIRE(get_sys_path_size() == sys_path_default_size + 1);
#endif
```

IIRC the 1st `REQUIRE(get_sys_path_size() == sys_path_default_size + 1);` succeeded everywhere, only the 2nd failed and would need that `#if`.

rwgk (approved):
I really like the last commit!

@Skylion007 could you please take another look?

rwgk:
I forgot to mention: it would be great if you could update the PR description and changelog entry.

arman-novikov:
@Skylion007  would you please share your review? I've already covered new _(and some current)_ features with tests

Skylion007 (approved):
Looks good to me.

rwgk:
Thanks @arman-novikov for all the work, thanks @Skylion007 for the review!

PR #4336, opened 2022-11-14
Include GLIBCXX==20160726 in list of exceptional versions

aws-taylor:
std::is_trivially_copyable is not available in 20160726, so include it in the list of exceptional versions.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/include/pybind11/numpy.h
+++ b/include/pybind11/numpy.h
@@ -367,7 +367,7 @@ using is_pod_struct
                                          // we need a standard layout type
 #if defined(__GLIBCXX__)                                                                          \
     && (__GLIBCXX__ < 20150422 || __GLIBCXX__ == 20150426 || __GLIBCXX__ == 20150623              \
-        || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160803)
+        || __GLIBCXX__ == 20150626 || __GLIBCXX__ == 20160726 || __GLIBCXX__ == 20160803)
              // libstdc++ < 5 (including versions 4.8.5, 4.9.3 and 4.9.4 which were released after
              // 5) don't implement is_trivially_copyable, so approximate it
              std::is_trivially_destructible<T>,


rwgk:
Do you know if it would make sense to simply use a date more recent than `_GLIBCXX__ < 20150422`, instead of making the list of special dates longer and longer?

aws-taylor:
I was wondering the same thing. The original author has `__GLIBCXX__ < 20150422` as part of the conditional, so clearly they knew that such constraints were possible. I opted to err on the side of caution and consistency; if you'd prefer I can try to relax the date as you propose.

rwgk:
> I was wondering the same thing. The original author has `__GLIBCXX__ < 20150422` as part of the conditional, so clearly they knew that such constraints were possible. I opted to err on the side of caution and consistency; if you'd prefer I can try to relax the date as you propose.

@mvoelkle-cern (author of #PR 2956) what do you think?

What is the likelihood that there are `__GLIBCXX__` versions in the range from `20150422` to `20160803` that do not need the "approximate" workaround?

My gut feeling: "very low", and considering that even `20160803` is a very old date already, I'd opt for simply `__GLIBCXX__ < 20160803`, and hope that we never have to revisit that condition again.

rwgk:
The two CI failures are flakes: PyPy refcount mismatch, intermittent Debian download failure. Safe to ignore.

mvoelkle-cern:
Hello, thanks for pulling me into the discussion! What is the effect of the approximation on versions which do support `is_trivially_copyable`? If it still works, ideed the check could be simplified. BTW, which version was released on 20160726?

My PR was trying to make it compile with RHEL 7's system compiler (gcc 4.8.5 aka 20150623). Gcc's timeline is here: [[https://gcc.gnu.org/develop.html#timeline]], I should have added it to the code comment. Can you please do that?

aws-taylor:
I'm going to have to withdraw this PR. I did some more research and discovered that GLIBCXX isn't useful for detecting versions. @mvoelkle-cern already discovered this, but I didn't internalize the nuance of their comment:

> (including versions 4.8.5, 4.9.3 and 4.9.4 which were released after 5)

Here's the relevant bug report - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69699

And an associated comment on the documentation:

>  This macro can be used in code to detect whether the C++ Standard Library implementation in use is libstdc++, but is not useful for detecting the libstdc++ version, nor whether particular features are supported. The macro value might be a date after a feature was added to the development trunk, but the release could be from an older branch without the feature. For example, in the 5.4.0 release the macro has the value 20160603 which is greater than the 20160427 value of the macro in the 6.1.0 release, but there are features supported in the 6.1.0 release that are not supported in the 5.4.0 release. You also can't test for the exact values listed below to try and identify a release, because a snapshot taken from the gcc-5-branch on 2016-04-27 would have the same value for the macro as the 6.1.0 release despite being a different version. Many GNU/Linux distributions build their GCC packages from snapshots, so the macro can have dates that don't correspond to official releases.

It seems the right solution is probably something more robust, perhaps by using [try_compile().](https://cmake.org/cmake/help/v3.4/command/try_compile.html)

rwgk:
> Many GNU/Linux distributions build their GCC packages from snapshots, so the macro can have dates that don't correspond to official releases.

Interesting. In that case I'd be OK simply making the list longer as needed, if that's a simple way to deal with the problem. I guess eventually there will be no GNU/Linux distribution being made anymore that needs it. Up to you.

aws-taylor:
This is probably not the last time that Pybind is going to have need of modern-ish C++ features, and I don't think it's appropriate to subject the world to the nuances of when my particular distribution happened to build libstdc++, so I'd prefer to take a different tack. There's already some work that's been done using _try_compile()_ to detect _stdc++fs_; I'd propose that's the right direction. Now I just need to find time...

https://github.com/pybind/pybind11/blob/36ccb08b0d8f007f674ce373095b4b278689fc10/tests/CMakeLists.txt#L355

PR #4342, opened 2022-11-17
Experimental: Python 3.12 workflow

rwgk:
## Description
New python312.yml, based on upstream.yml, but with many tweaks.

## Suggested changelog entry:

```rst

```


rwgk:
Final result: https://github.com/google/pywrapcc/pull/30006

PR #4350, opened 2022-11-21
Use `PyEval_InitThreads()` as intended

rwgk:
## Description
This PR actually matters

* only for Python 3.6 (according to the Python documentation),
* only with `-DPYBIND11_SIMPLE_GIL_MANAGEMENT=ON` (evidently, based on CI flakes observed since PR #4216 was merged),
* and only if embed.h is used.

Hypothesis: `-DPYBIND11_SIMPLE_GIL_MANAGEMENT=OFF` covers up that `PyEval_InitThreads()` was misplaced.

This PR almost does not matter anymore at all:

* https://docs.python.org/3.11/c-api/init.html#c.PyEval_InitThreads

> Changed in version 3.7: This function is now called by `Py_Initialize()`, so you don’t have to call it yourself anymore.

Note that Python 3.6 EOL was 2021-12-23 (https://endoflife.date/python), but since PR #4216 was merged we had distracting CI flakes that this PR aims to get rid of, e.g.:

* 🐍 3.6 • windows-2022 • x64

```
Run cmake --build .  --target cpptest -j 2
MSBuild version 17.3.1+2badb37d1 for .NET Framework
  Checking File Globs
  external_module.vcxproj -> D:\a\pybind11\pybind11\tests\test_embed\external_module.cp36-win_amd64.pyd
  test_embed.vcxproj -> D:\a\pybind11\pybind11\tests\test_embed\Debug\test_embed.exe
CUSTOMBUILD : Fatal Python error : This thread state must be current when releasing [D:\a\pybind11\pybind11\tests\test_embed\cpptest.vcxproj]
  
  Current thread 0x0000[14](https://github.com/pybind/pybind11/actions/runs/3511087576/jobs/5881522679#step:13:15)84 (most recent call first):
  
  Thread 0x00000fc4 (most recent call first):
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(247,5): error MSB8066: Custom build for 'D:\a\pybind11\pybind11\CMakeFiles\bf7d1[16](https://github.com/pybind/pybind11/actions/runs/3511087576/jobs/5881522679#step:13:17)f1[18](https://github.com/pybind/pybind11/actions/runs/3511087576/jobs/5881522679#step:13:19)90487c30aedc390ae4de7\cpptest.rule;D:\a\pybind11\pybind11\tests\test_embed\CMakeLists.txt' exited with code -1073740791. [D:\a\pybind11\pybind11\tests\test_embed\cpptest.vcxproj]
Error: Process completed with exit code 1.
```

## Suggested changelog entry:

```rst
Bug fix affecting only Python 3.6 under very specific, uncommon conditions: correction of misplaced `PyEval_InitThreads()` call.
```

### DIFF
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -118,7 +118,7 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
 #if PY_VERSION_HEX < 0x030B0000
 
     Py_InitializeEx(init_signal_handlers ? 1 : 0);
-#    if PY_VERSION_HEX < 0x03070000
+#    if defined(WITH_THREAD) && PY_VERSION_HEX < 0x03070000
     PyEval_InitThreads();
 #    endif
 
@@ -171,7 +171,7 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
         throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
                                                           : "Failed to init CPython");
     }
-#    if PY_VERSION_HEX < 0x03070000
+#    if defined(WITH_THREAD) && PY_VERSION_HEX < 0x03070000
     PyEval_InitThreads();
 #    endif
     if (add_program_dir_to_path) {
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -118,6 +118,9 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
 #if PY_VERSION_HEX < 0x030B0000
 
     Py_InitializeEx(init_signal_handlers ? 1 : 0);
+#    if PY_VERSION_HEX < 0x03070000
+    PyEval_InitThreads();
+#    endif
 
     // Before it was special-cased in python 3.8, passing an empty or null argv
     // caused a segfault, so we have to reimplement the special case ourselves.
@@ -168,6 +171,9 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
         throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
                                                           : "Failed to init CPython");
     }
+#    if PY_VERSION_HEX < 0x03070000
+    PyEval_InitThreads();
+#    endif
     if (add_program_dir_to_path) {
         PyRun_SimpleString("import sys, os.path; "
                            "sys.path.insert(0, "
--- a/include/pybind11/embed.h
+++ b/include/pybind11/embed.h
@@ -171,9 +171,6 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
         throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
                                                           : "Failed to init CPython");
     }
-#    if defined(WITH_THREAD) && PY_VERSION_HEX < 0x03070000
-    PyEval_InitThreads();
-#    endif
     if (add_program_dir_to_path) {
         PyRun_SimpleString("import sys, os.path; "
                            "sys.path.insert(0, "
--- a/include/pybind11/detail/internals.h
+++ b/include/pybind11/detail/internals.h
@@ -468,9 +468,6 @@ PYBIND11_NOINLINE internals &get_internals() {
         internals_ptr = new internals();
 #if defined(WITH_THREAD)
 
-#    if PY_VERSION_HEX < 0x03090000
-        PyEval_InitThreads();
-#    endif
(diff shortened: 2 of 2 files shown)


rwgk:
@eacousineau there is no specific reason for requesting a review from you, but @henryiii has been unresponsive for 3 weeks. It would be great if you could help out with a second set of eyes.

include/pybind11/embed.h
```diff
@@ -168,6 +171,9 @@ inline void initialize_interpreter(bool init_signal_handlers = true,
         throw std::runtime_error(PyStatus_IsError(status) ? status.err_msg
                                                           : "Failed to init CPython");
     }
+#    if defined(WITH_THREAD) && PY_VERSION_HEX < 0x03070000
```
EricCousineau-TRI, line 174: This is in the `else` portion of `#if PY_VERSION_HEX < 0x030B0000`.
`PY_VERSION_HEX >= 0x030B0000 && PY_VERSION_HEX < 0x03070000` seems like it'd always be false.

If so, consider removing?

EricCousineau-TRI (changes requested):
Sounds good, and thanks for the writeup! Though I'm afraid I may be a bit slow to catch up on the issues.

It's hard for me to posit from above Windows-based failure that the above indicated misplacement. Can I ask if the following is the correct reasoning, reasoning within Python 3.6?

- Calling `PyEval_InitThreads()` in `get_internals()` via nominal usage did nothing per [3.6 docs](https://docs.python.org/3.6/c-api/init.html#c.PyEval_InitThreads) stating that "This is a no-op when called for a second time", so this is a harmless (but noisy) misplacement.
- The missing call in `embed.h` are why Windows (and others?) fail. This smells like it'd happen when the embedded interpreter exits before `get_internals()` was called. Is this the case? (is it spurious, or deterministic?)

include/pybind11/embed.h
rwgk: Yes, thanks for catching that! Sorry I was going through with lawnmower mentality and completely missed that.

Skylion007 (approved):

include/pybind11/embed.h
EricCousineau-TRI: Welcome!

EricCousineau-TRI (approved):
Still not sure on the whole of the changes (per questions in first post), but def. won't block progress - thanks!

rwgk:
> Still not sure on the whole of the changes (per questions in first post), but def. won't block progress - thanks!

Sorry I wasn't able to finish the below before a long meeting, continuing now ...

> Sounds good, and thanks for the writeup! Though I'm afraid I may be a bit slow to catch up on the issues.
> 
> It's hard for me to posit from above Windows-based failure that the above indicated misplacement. Can I ask if the following is the correct reasoning, reasoning within Python 3.6?
> 
> * Calling `PyEval_InitThreads()` in `get_internals()` via nominal usage did nothing per [3.6 docs](https://docs.python.org/3.6/c-api/init.html#c.PyEval_InitThreads) stating that "This is a no-op when called for a second time", so this is a harmless (but noisy) misplacement.

Yes, in the sense that it is harmless if there is actually a correct placement. That's not the case before this PR. I was thinking of "misplaced" as in "we're moving it from the wrong place to the right place".

> * The missing call in `embed.h` are why Windows (and others?) fail. This smells like it'd happen when the embedded interpreter exits before `get_internals()` was called. Is this the case? (is it spurious, or deterministic?)

Not deterministic. "flaky" is the term I see getting used around Google.

My understanding is not complete, but IIUC the Python documentation correctly, prior to Python 3.7, `PyEval_InitThreads()` needs to be called before the first attempt to acquire the GIL. `get_internals()` clearly violates that, because this:

* https://github.com/pybind/pybind11/blob/296615ad34f9d8f680efbab22553881ad9843063/include/pybind11/detail/internals.h#L433

is before this:

* https://github.com/pybind/pybind11/blob/296615ad34f9d8f680efbab22553881ad9843063/include/pybind11/detail/internals.h#L472

Now **pure speculation**: I think that the "non-simple" code in gil.h covered that up, somehow. I'm thinking that because we didn't see the flakiness I'm trying to fix with this PR before PR #4216.

Thanks for the approvals!

EricCousineau-TRI:
Sweet, that all makes sense - thanks for the thorough response!

PR #4276, opened 2022-10-22
chore: improve issue template

henryiii:
Small improvement to issue template, hopefully helping highlight regressions.

PR #4363, opened 2022-11-25
fix(.github): fix bug-report issue template

XuehaiPan:
## Description

Documented in [Building communities / Issue & PR templates / Syntax for GitHub's form schema #Keys](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#keys)

The correct type is `input` rather than `Input`. The bug-report template is not enabled at all and users cannot create a bug-report on the issue page [here](https://github.com/pybind/pybind11/issues/new/choose).

image

Related issue #4276.


Skylion007 (approved):

rwgk:
Thanks for the fix!

Issue #4365, opened 2022-11-28
[BUG]: invalid shared ptr conversion leads to crash


bluescarni:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://pybind11.readthedocs.io). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/pybind/pybind11/issues) and [Discussions](https:/pybind/pybind11/discussions) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in the [Gitter chat room](https://gitter.im/pybind/Lobby) or in a [Discussion](https:/pybind/pybind11/discussions/new).

### What version (or hash if on master) of pybind11 are you using?

2.10.1

### Problem description

We are experiencing a crash in the exposition of a C++ class hierarchy which seems to be related to conversions between ``std::shared_ptr``s within the class hierarchy. The attached minimal code is enough to reproduce the issue. Let me give a brief description.

On the C++ side, we have to base classes, ``Base0`` and ``Base1``, and a derived class ``Derived`` which inherits from the two bases. All classes are exposed via pybind11 using ``std::shared_ptr`` holder types, and the exposition of ``Derived`` signals that ``Base0`` and ``Base1`` are its bases.

Then, we expose a factory function ``make_object()`` which internally creates a shared ptr to ``Derived``, converts it to ``std::shared_ptr< Base0 >``, and then returns it. In the implementation of ``make_object()``, we print to screen the memory addresses of the base and derived objects (more on that later).

Finally, we expose a ``print_object()`` function that takes in input a pointer to derived and tries to print to screen the size of its vector data member. It also prints to screen the pointer contained  in the input shared pointer.

Now for the Python code:

```ipython
In [1]: import example

In [2]: o = example.make_object(42)
ret der ptr: 0x557ad38c7a30
ret base ptr: 0x557ad38c7a50

In [3]: example.print_object(o)
der ptr: 0x557ad38c7a50
23501544953261
```

So there's something obviously going wrong here, since:
- the vector data member definitely does not have ``23501544953261`` elements, and
- the pointers seem to be all wrong. In particular, the memory address ``0x557ad38c7a50`` is a pointer to ``Base0`` in ``make_object()``, but it has magically been converted into a pointer to ``Derived`` in ``print_object(o)``, even though casting between ``Base0`` and ``Derived`` **should** change the memory address in this hierarchy (as evidenced in the pointer values printed from ``make_object()``).

[...]

PR #4374, opened 2022-12-01
Add files for testing MI issue.

bluescarni:
## Description

Test case for the issue described in #4365 .

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -156,7 +156,8 @@ set(PYBIND11_TEST_FILES
     test_tagbased_polymorphic
     test_thread
     test_union
-    test_virtual_functions)
+    test_virtual_functions
+    test_mi_debug)
 
 # Invoking cmake with something like:
 #     cmake -DPYBIND11_TEST_OVERRIDE="test_callbacks.cpp;test_pickling.cpp" ..
--- a/tests/test_mi_debug.cpp
+++ b/tests/test_mi_debug.cpp
@@ -1,0 +1,59 @@
+#include <memory>
+#include <vector>
+#include <iostream>
+
+#include <pybind11/pybind11.h>
+
+// The first base class.
+struct Base0
+{
+    virtual ~Base0() { }
+};
+
+using Base0Ptr = std::shared_ptr< Base0 >;
+
+// The second base class.
+struct Base1
+{
+    virtual ~Base1() { }
+    std::vector< int > vec = { 1, 2, 3, 4, 5 };
+};
+
+using Base1Ptr = std::shared_ptr< Base1 >;
+
+// The derived class.
+struct Derived : Base1, Base0
+{
+    virtual ~Derived() { }
+};
+
+using DerivedPtr = std::shared_ptr< Derived >;
+
+PYBIND11_MODULE(example, m) {
+    // Expose the bases.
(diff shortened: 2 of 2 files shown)


rwgk:
Quick thoughts:
* The `reinterpret_cast` you found is definitely not good.
* But if it also fails with `smart_holder`, that's not the (only) problem.
* I don't fully understand the MI code, but I do know that it uses `dynamic_cast<void *>` which casts to the most derived type, and then handles the Python object as an object of that type, even if the original type was a base class.

**Pure speculation:** is there a general problem in how the MI code handles your types?

Which leads to the question: do you also see a problem if you use raw pointers instead of `shared_ptr`?

rwgk:
The CI seems to be mostly working (still running at this minute). That's a great start.
From here it's easy: could you please add some basic test(s) in test_mi_debug.py?
To see how many jobs are failing.
Sometimes that gives useful clues.

bluescarni:
> Quick thoughts:
> 
>     * The `reinterpret_cast` you found is definitely not good.
> 
>     * But if it also fails with `smart_holder`, that's not the (only) problem.
> 
>     * I don't fully understand the MI code, but I do know that it uses `dynamic_cast<void *>` which casts to the most derived type, and then handles the Python object as an object of that type, even if the original type was a base class.

I am not sure whether or not this is a problem specific to MI. I think this has more to do with a situation in which, for the same object, the pointer to base differs from the pointer to derived, and MI might be only an way to trigger this occurrence.

My understanding of what is going on largely agrees with yours: pybind11 infers the most derived type of the pointer to the base object, somehow encodes this type information somewhere and concludes that the two shared pointers to ``Base`` and ``Derived`` have the same most derived type.

Then, completely inexplicably to me, a direct, unchecked ``reinterpret_cast`` from ``std::shared_ptr<Base>`` to ``std::shared_ptr<Derived>`` is done. This is **ALWAYS** undefined behaviour (because the cast is done directly on the holder types, not on the internal pointers), it just happens to work in practice if the numerical values of the base and derived pointers are identical (as it is often the case for simple class hierarchies, though it is nowhere guaranteed in the C++ standard).

Thus, my impression is that automatic Base <-> Derived conversions of smart pointers in pybind11 might have always been broken from the C++ point of view in pybind11, they just happen to work in practice in many cases due to implementation-defined behaviour.

bluescarni:
> Which leads to the question: do you also see a problem if you use raw pointers instead of `shared_ptr`?

I don't know about raw pointers, but if I change the function to take as input a reference to ``Derived`` (rather than a shared pointer to ``Derived``), everything works fine and the pointer values are correct.

rwgk:
> conversions of smart pointers in pybind11 might have always been broken from the C++ point of view

Yes, true. That's why I had no choice but to do something else, which became the smart_holder branch.

I'm very worried that you're seeing a problem also with the smart_holder branch. I need to find a block of time to debug, that's the biggest problem right now.

> I don't know about raw pointers, but if I change the function to take as input a reference to Derived (rather than a shared pointer to Derived), everything works fine and the pointer values are correct.

That's an indication that the ["up-down-sideways"](https://en.cppreference.com/w/cpp/language/dynamic_cast) cast code (which is ~completely independent of the smart pointer handling code) is working correctly.

But then again, why does the smart_holder not work for you? We need to find out.

PR #4376, opened 2022-12-01
[smart_holder] git merge master (attempt 2 after #4372 was merged)

rwgk:
## Description
Helper/scratch PR for testing.

**REPLACED by #4379.**

## Suggested changelog entry:

```rst

```

PR #4377, opened 2022-12-01
chore: Convert direct multiprocessing.set_start_method("forkserver") call to a pytest fixture.

Skylion007:
## Description
* This PR updates the conftest logic to call multiprocessing.set_start_method in a fixture from #4306. This should be a bit safer since it ensure it can only be called once and if we import anything from conftest for whatever reason, it won't try to run the code snippit again. This is what we should have done at the start as putting any "raw" code in conftest.py is ill advised and bugprone.
* [This doc](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) has some really good explanation about the order the fixtures are called and when. a scope='session` is called once when pytest startups and its state is shared among all tests that include it. By adding autouse=True, it is automatically added to all tests affected by conftest.py meaning that it is run regardless of which test is run, and it is only once per session (hence scope="session"). You can also have fixtures run once per module, once per class, etc... to store global state, or to have a piece of code run once before the first instance of the class is created etc...
* Prevents this code from being run twice through use of `importlib.reload` or autoreload etc from Jupyter notebooks and better documents what functions are affecting the global state before a test is run

## Suggested changelog entry:

```rst
* multiprocessing_set_spawn in pytest fixture for added safety.
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,15 +17,19 @@ import pytest
 # Early diagnostic for failed imports
 import pybind11_tests
 
-if os.name != "nt":
-    # Full background: https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
-    # In a nutshell: fork() after starting threads == flakiness in the form of deadlocks.
-    # It is actually a well-known pitfall, unfortunately without guard rails.
-    # "forkserver" is more performant than "spawn" (~9s vs ~13s for tests/test_gil_scoped.py,
-    # visit the issuecomment link above for details).
-    # Windows does not have fork() and the associated pitfall, therefore it is best left
-    # running with defaults.
-    multiprocessing.set_start_method("forkserver")
+
+@pytest.fixture(scope="session", autouse=True)
+def always_forkserver_on_unix():
+    if os.name != "nt":
+        # Full background: https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
+        # In a nutshell: fork() after starting threads == flakiness in the form of deadlocks.
+        # It is actually a well-known pitfall, unfortunately without guard rails.
+        # "forkserver" is more performant than "spawn" (~9s vs ~13s for tests/test_gil_scoped.py,
+        # visit the issuecomment link above for details).
+        # Windows does not have fork() and the associated pitfall, therefore it is best left
+        # running with defaults.
+        multiprocessing.set_start_method("forkserver")
+
 
 _long_marker = re.compile(r"([0-9])L")
 _hexadecimal = re.compile(r"0x[0-9a-fA-F]+")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,15 +20,17 @@ import pybind11_tests
 
 @pytest.fixture(scope="session", autouse=True)
 def always_forkserver_on_unix():
-    if os.name != "nt":
-        # Full background: https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
-        # In a nutshell: fork() after starting threads == flakiness in the form of deadlocks.
-        # It is actually a well-known pitfall, unfortunately without guard rails.
-        # "forkserver" is more performant than "spawn" (~9s vs ~13s for tests/test_gil_scoped.py,
-        # visit the issuecomment link above for details).
-        # Windows does not have fork() and the associated pitfall, therefore it is best left
-        # running with defaults.
-        multiprocessing.set_start_method("forkserver")
+    if os.name == "nt":
+        return
+
+    # Full background: https://github.com/pybind/pybind11/issues/4105#issuecomment-1301004592
+    # In a nutshell: fork() after starting threads == flakiness in the form of deadlocks.
+    # It is actually a well-known pitfall, unfortunately without guard rails.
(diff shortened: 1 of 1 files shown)


rwgk:
Oh, this looks interesting! (I think it much more than a chore!)

Do you already know if that gets rid of the repeat
```
DISABLED std::system_error: ODR VIOLATION DETECTED: pybind11::detail::type_caster<mrc_ns::type_mrc>: SourceLocation1="/__w/pybind11/pybind11/tests/test_type_caster_odr_guard_1.cpp:17", SourceLocation2="/__w/pybind11/pybind11/tests/test_type_caster_odr_guard_2.cpp:17"
```
that popped up under #4376?

Skylion007:
> DISABLED std::system_error: ODR VIOLATION DETECTED: pybind11::detail::type_caster<mrc_ns::type_mrc>: SourceLocation1="/__w/pybind11/pybind11/tests/test_type_caster_odr_guard_1.cpp:17", SourceLocation2="/__w/pybind11/pybind11/tests/test_type_caster_odr_guard_2.cpp:17"

@rwgk I am not sure? This is just functionally the same as it was before, just prevents the code from being run twice by having it be a pytest.fixture .

Skylion007:
All the failures are flakes. Happy to merge pending approval.

rwgk:
> All the failures are flakes. Happy to merge pending approval.

Please give me a moment to try if it helps getting rid of the spurious `ODR VIOLATION DETECTED` messages.

I believe this PR is good even if it doesn't help, but I want to try first, in case a tweak here is needed for something.

TBH I don't really understand this sentence:

> This should be a bit safer since it ensure it can only be caused once and if we import anything from conftest for whatever reason, it won't try to run the code snippit again. 

Did you mean "called once"?

I think I'm missing something fundamental about how fixtures work. Could you please help me (and others) understand?

My (maybe incorrect) best-guess understanding is: conftest.py is imported only once, when pytest starts up. There is no direct reference to `confest` in all of pybind11.

Skylion007:
> Did you mean "called once"? 

Yep

[This doc](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) has some really good explanation about the order the fixtures are called and when. a scope='session`  is called once when pytest startups and its state is shared among all tests that include it. By adding autouse=True, it is automatically added to all tests affected by conftest.py meaning that it is run regardless of which test is run, and it is only once per session (hence scope="session"). You can also have fixtures run once per module, once per class, etc... to store global state, or to have a piece of code run once before the first instance of the class is created etc...

Skylion007:
> My (maybe incorrect) best-guess understanding is: conftest.py is imported only once, when pytest starts up. There is no direct reference to confest in all of pybind11.

Yeah, true, but let's say you just wanted to import code snippits from it outside of a pytest context for debugging, it would still set the multiprocessing state, which may be undesirable. Fixture are also nice because they document how the state of the program is setup for each test.

rwgk:
> > Did you mean "called once"?
> 
> Yep
> 
> [This doc](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) has some really good explanation about the order the fixtures are called and when. a scope='session` is called once when pytest startups and its state is shared among all tests that include it. By adding autouse=True, it is automatically added to all tests affected by conftest.py meaning that it is run regardless of which test is run, and it is only once per session (hence scope="session"). You can also have fixtures run once per module, once per class, etc... to store global state, or to have a piece of code run once before the first instance of the class is created etc...

Thanks, I need a moment to look.

Quick results:
* this PR does not get rid of the `ODR VIOLATION DETECTED` messages
* when running the pybind11 pytest unit tests **without** this PR, the `"forkserver"` line is executed only once. I verified that via a print to /tmp/somefile with open append.

Skylion007:
> 

Yeah, it should be executed once either way, this just make it more robust and it ensures its' not executed more than once by more edge case things like abusing `importlib.reload` etc. Regardless, this is the recommended way of doing it.

tests/conftest.py
```diff
@@ -17,15 +17,19 @@
...
-    # It is actually a well-known pitfall, unfortunately without guard rails.
-    # "forkserver" is more performant than "spawn" (~9s vs ~13s for tests/test_gil_scoped.py,
-    # visit the issuecomment link above for details).
-    # Windows does not have fork() and the associated pitfall, therefore it is best left
-    # running with defaults.
-    multiprocessing.set_start_method("forkserver")
+
+@pytest.fixture(scope="session", autouse=True)
+def always_forkserver_on_unix():
+    if os.name != "nt":
```
rwgk, line 23: With this being moved to a function, I have a slight preference for

```
if os.name == "nt":
  return
```
Then the rest of the code unindented one level.
That's slightly more future proof, in case we want to add more conditions.
But up to you.

rwgk (approved):
Could you please work your latest explanations into the PR description before merging?

Even more importantly: update the title? (because that goes into the commit hash)
I don't think "bugfix" is appropriate.
"best practices" seem more fitting.
But I'd just say what this does in the title and carefully explain in the description.

Suggested title:

Convert direct `multiprocessing.set_start_method("forkserver")` call to a pytest fixture.

PR #4381, opened 2022-12-02
Auto select return value policy for clif_automatic

wangxf123456:
## Description

PyCLIF automatically chooses lifetime management of return values based on the properties of them: https://github.com/google/clif/tree/main/clif/python#pointers-references-and-object-ownership. This PR should make the return value policy selection consistent with PyCLIF.

### DIFF
--- a/tests/test_return_value_policy_override.cpp
+++ b/tests/test_return_value_policy_override.cpp
@@ -1,11 +1,69 @@
+#include <pybind11/smart_holder.h>
+
 #include "pybind11_tests.h"
 
 namespace test_return_value_policy_override {
 
 struct some_type {};
 
+struct obj {
+    std::string mtxt;
+    obj(const std::string &mtxt_) : mtxt(mtxt_) {}
+    obj(const obj &other) { mtxt = other.mtxt + "_CpCtor"; }
+    obj(obj &&other) { mtxt = other.mtxt + "_MvCtor"; }
+    obj &operator=(const obj &other) {
+        mtxt = other.mtxt + "_CpCtor";
+        return *this;
+    }
+    obj &operator=(obj &&other) {
+        mtxt = other.mtxt + "_MvCtor";
+        return *this;
+    }
+};
+
+struct nocopy {
+    std::string mtxt;
+    nocopy(const std::string &mtxt_) : mtxt(mtxt_) {}
+    nocopy(nocopy &&other) { mtxt = other.mtxt + "_MvCtor"; }
+    nocopy &operator=(nocopy &&other) {
+        mtxt = other.mtxt + "_MvCtor";
+        return *this;
+    }
+};
+
+obj *return_pointer() {
+    static obj value("pointer");
+    return &value;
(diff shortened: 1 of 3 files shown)


tests/test_return_value_policy_override.cpp
```diff
@@ -1,11 +1,69 @@
...
+    }
+    obj &operator=(obj &&other) {
+        mtxt = other.mtxt + "_MvCtor";
+        return *this;
+    }
+};
+
+struct nocopy {
+    std::string mtxt;
+    nocopy(const std::string &mtxt_) : mtxt(mtxt_) {}
```
Skylion007, line 26: Don't you still need to delete the copy ctor?

wangxf123456: Thanks for pointing this out. I forgot this. I deleted the copy ctor.

tests/test_return_value_policy_override.cpp
```diff
@@ -51,9 +53,13 @@ const obj &return_const_reference() {
     return value;
 }
 
-std::shared_ptr<obj> return_shared_pointer() { return std::make_shared<obj>("shared_pointer"); }
+std::shared_ptr<obj> return_shared_pointer() {
+    return std::shared_ptr<obj>(new obj("shared_pointer"));
```
Skylion007, line 57: Should be `std::make_shared`, right?

tests/test_return_value_policy_override.cpp
```diff
@@ -1,11 +1,75 @@
+#include <pybind11/smart_holder.h>
+
 #include "pybind11_tests.h"
 
 namespace test_return_value_policy_override {
 
 struct some_type {};
 
+struct obj {
```
rwgk, line 9: `obj` is usually used for instances, it's surprising as a type name.

The combinations of copyable/movable keep coming up, it would be good have some systematic naming that we could use throughout. What do you think about

```
struct type_cp1_mv1
struct type_cp1_mv0
struct type_cp0_mv1
struct type_cp0_mv0
```

?

With a comment like:
```
// cp = copyable, mv = movable, 1 = yes, 0 = no
```
Question that's maybe beyond this PR: do we have (or would it be safest) to add tests for the other two combinations?

IIUC here you have only `type_cp1_mv1` and `type_cp0_mv1`, is that correct?

tests/test_return_value_policy_override.cpp
wangxf123456: Initially I used `std::make_shared` and `std::make_unqiue`, but then I see errors `note: ‘std::make_unique’ is only available from C++14 onwards`, so I modified both of them. I should be able to use `std::make_shared` with C++11. I will update the PR.

tests/test_return_value_policy_override.cpp
wangxf123456: Thanks for your suggestions. I updated the PR with more tests.

rwgk (approved):

PR #4382, opened 2022-12-04
Replace `ubuntu-latest` with `ubuntu-22.04`

rwgk:
## Description
Using previous configure.yml logs to pin-point when `ubuntu-latest` changed from 20.04 to 22.04:

* `2022-12-02T01:51:20.8819798Z` Image: ubuntu-20.04  —  last success
* `2022-12-02T13:46:45.3899378Z` Image: ubuntu-22.04  —  first failure

There are multiple problems:

* configure.yml: Python 3.6 no longer available (for cmake 3.4 testing). When changing to Python 3.9, there is boost compilation error (related to `variant` or boost/mpl).
* configure.yml: cmake 3.4 no longer working because of missing libidn11.
* ci.yml: Python 3.6 no longer available; valgrind testing broken.

General: it does not make much sense to test Python 3.6, cmake 3.4, valgrind with Python 3.9 (all very or somewhat outdated) on the "tatest" platform.

Opinion: it will be best to pin those indefinitely, modernize valgrind testing with the latest platform & Python (when someone gets to it), and drop cmake 3.4 and Python 3.6 entirely.

Conservative resolution: change `ubuntu-latest` to `ubuntu-20.04`, but **only as required to resolve breakages**.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ubuntu-latest, windows-2022, macos-latest]
+        runs-on: [ubuntu-20.04, windows-2022, macos-latest]
         python:
         - '3.6'
         - '3.9'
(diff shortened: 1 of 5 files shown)


rwgk:
Commit  cdfd995 confirms that we get back to all-green by systematically changing all `ubuntu-latest` to `ubuntu-20.04`.

I'll try to make the changes more selective, to only address breakages.

Skylion007 (approved):

rwgk:
Thanks @Skylion007! I wrote up the rationale in the PR description.

PR #4387, opened 2022-12-06
chore: update clang-tidy to 15

Skylion007:
## Description
* Update clang-tidy docker container to clang 15 and the new way of installing / generating Silekh docker containers. 

## Suggested changelog entry:

```rst
* Update clang-tidy to 15 in CI
```

### DIFF
--- a/include/pybind11/stl_bind.h
+++ b/include/pybind11/stl_bind.h
@@ -354,11 +354,15 @@ void vector_accessor(enable_if_t<vector_needs_copy<Vector>::value, Class_> &cl) 
     cl.def("__getitem__", [](const Vector &v, DiffType i) -> T {
         if (i < 0) {
             i += v.size();
+            if (i < 0) {
+                throw index_error();
+            }
         }
-        if (i < 0 || (SizeType) i >= v.size()) {
+        auto i_st = static_cast<SizeType>(i);
+        if (i_st >= v.size()) {
             throw index_error();
         }
-        return v[(SizeType) i];
+        return v[i_st];
     });
 
     cl.def(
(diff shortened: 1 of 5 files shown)


include/pybind11/stl_bind.h
```diff
@@ -352,10 +352,10 @@ void vector_accessor(enable_if_t<vector_needs_copy<Vector>::value, Class_> &cl)
     using DiffType = typename Vector::difference_type;
     using ItType = typename Vector::iterator;
     cl.def("__getitem__", [](const Vector &v, DiffType i) -> T {
-        if (i < 0 && (i += v.size()) < 0) {
-            throw index_error();
+        if (i < 0) {
+            i += v.size();
         }
-        if ((SizeType) i >= v.size()) {
+        if (i < 0 || (SizeType) i >= v.size()) {
```
Skylion007, line 358: This is microscopically slower than it was before since now it does the i<0 check twice even if ti doesn't need to, but it seems a bit more readable now. Also, if we do care about performance in this case, we should fix the wrap_i function as well.

rwgk: This is mostly to make it "nice" (subjectively), I'm really not worried about performance one way or another (untested):
```
if (i < 0) {
    i += v.size();
    if (i < 0) {
        throw index_errror();
    }
}
auto i_st = static_cast<SizeType>(i);
if (i_st >= v.size()) {
    throw index_error();
}
return v[i_st];
```

What's missing here IMO is nice error reporting (e.g. something like "index -8 is out of range for array size 5"), but that's beyond the scope of this PR.

rwgk (approved):
There is still one clang-tidy failure:
```
/__w/pybind11/pybind11/include/pybind11/stl.h:319:34: error: unchecked access to optional value [bugprone-unchecked-optional-access,-warnings-as-errors]
        return value_conv::cast(*std::forward<T>(src), policy, parent);
                                 ^
```

Please feel free to merge when that's resolved, with out without the suggested change.

rwgk (approved):

PR #4393, opened 2022-12-10
chore: future safe bugbear opinionated warnings

henryiii:
Future safe version of #4391. I've been applying this fix elsewhere.

The opinionated checks are not something you can just enable anymore, after B905 shows that non-EoL Pythons can be required for them. Instead, just turn on the ones that are helpful, and if more are added in the future, we can add them if we like them.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,5 +46,5 @@ zip_safe = False
 max-line-length = 120
 show_source = True
 exclude = .git, __pycache__, build, dist, docs, tools, venv
-extend-ignore = E203, E722, B903, B905, B950
-extend-select = B9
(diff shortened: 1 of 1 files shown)


rwgk (approved):
Thank you!

What made this particularly confusing: I first saw B905 only on my local workstation, but not in GHA, then suddenly in some GHA jobs but not others, only eventually everywhere. — I'm guessing it was some rollout window that overlapped with what I was doing.

Skylion007 (approved):

henryiii:
There’s quite a bit of caching with pre-commit, that’s why it’s so fast.

PR #4353, opened 2022-11-23
Correct class names for KeysView, ValuesView and ItemsView in bind_map

aimir:
## Description

Wrap KeysView, ValuesView and ItemsView with correct typing information in mappings wrapped via bind_map.

For example, for a mapping from strings to doubles, its `.keys()` function should have return type `KeysView[str]`.

Currently, things have been wrapped like `KeysView[MappingWrapperName]`, which is both misleading and inconsistent with the parallel python classes (especially for ItemsView, which usually has two type arguments, but here has only one).

This also means that we get differently-typed `KeysView` objects for different mappings with the same key type, even though they are functionally identical types.

This is solved by making `KeysView` etc. an abstract class that's a template on the relevant types only, rather than the entire map, so that maps with similar keys will have `KeysView` with the same type. The actual implementation is provided later, when wrapping the `.keys()` function of the mapping class, and similarly for values and items.

Closes #3986.

## Suggested changelog entry:

* Fixed typing of `KeysView`, `ValuesView` and `ItemsView` in `bind_map`

PR #4022, opened 2022-06-21
[smart_holder] `type_caster` ODR guard

rwgk:
## Description
In large-scale environments (e.g. Google) `type_caster` [ODR](https://en.cppreference.com/w/cpp/language/definition) violations are probably the biggest pybind11 pitfall. Without this PR, it is very difficult to ensure that a large application consistently uses the same `type_caster<T>` definition for a given `T` in all translation units linked together with mutual linker-level visibility. Accidents are to be expected and are very difficult to prevent & debug at scale, for example a mix of translation units that use `PYBIND11_MAKE_OPAQUE` + pybind11/stl_bind.h, and others that use pybind11/stl.h. Another type of accident is that custom `type_caster` includes are simply forgotten in some translation units, but by chance the linker picks all the right `type_caster` fragments (from other translation units), until one day it suddenly does not, because of some unrelated change in the environment (e.g. new compiler version, a system library update), resulting in seemingly inexplicable failures. Specific to the smart_holder branch, there is also the problem that the `PYBIND11_SMART_HOLDER_TYPE_CASTERS` macros may accidentally be missing in some translation units. In many cases runtime errors may give useful clues, but in general, the behavior is undefined, which is not acceptable in large-scale production environments.

Runtime detection of ODR violations is very elusive, especially when involving templates. Unfortunately, currently the [ASAN `detect_odr_violation` option](https://github.com/google/sanitizers/wiki/AddressSanitizerOneDefinitionRuleViolation) is known to not catch `type_caster` ODR violations. The `type_caster` ODR guard added with this PR is currently the only tool available, although it employs ODR itself to shield `type_caster::name.sloc` (source location) information from the linker mechanisms that discard all but one of multiple `type_caster` definitions. See the `WARNING`s in pybind11/detail/descr.h and pybind11/detail/type_caster_odr_guard.h for details. The current implementation is known to work on almost all platforms that meet the compiler requirements (C++17, C++20), except one (gcc 9.4.0 Debug build). See comments below for details.

Note that the new test_type_caster_odr_guard_1, test_type_caster_odr_guard_2 pair deliberately violates the ODR (see C++ source code comments), but in a way that is evidently compatible with reliable unit testing, very similar to how the ODR in the guard itself is evidently of a benign kind.

[...]

PR #4395, opened 2022-12-11
[KNOWN TRAP]: segfault with stl.h wrapper for std::map (and no optimization enabled)

rhaschke:
## Description

When using both, a `Debug` build and the `stl.h` wrapper for `std::map`, I ran into a surprising segfault. 
I spent a whole day tracking that error down to pybind11 being compiled with no optimization options enabled.
I tried to come up with a minimal example reproducing exactly my issue - but I failed.

However, I was able to reproduce the issue in a different context. For now, this PR provides this minimal example.
What is really weird is that adding a new wrapper in `test_stl.cpp` triggers a bug in `test_stl_binders`.

Note, that I had to add additional gcc and clang build configs to CI (with optimizations disabled) to trigger the bug in Linux. On Windows, it shows up with the existing jobs already.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,12 @@ jobs:
               -DPYBIND11_FINDPYTHON=ON
               -DCMAKE_CXX_FLAGS="-D_=1"
           - runs-on: ubuntu-20.04
+            python: '3.8'
+            # Add a build with no optimizations
+            args: >
+              -DPYBIND11_FINDPYTHON=ON
+              -DCMAKE_BUILD_TYPE=none -DCMAKE_CXX_FLAGS="-DNDEBUG"
+          - runs-on: ubuntu-20.04
             python: 'pypy-3.8'
             args: >
               -DPYBIND11_FINDPYTHON=ON
@@ -293,6 +299,9 @@ jobs:
             std: 20
           - clang: 14
             std: 20
+          - clang: 10
+            std: 14
+            args: -DCMAKE_BUILD_TYPE=none -DCMAKE_CXX_FLAGS="-DNDEBUG"
 
     name: "🐍 3 • Clang ${{ matrix.clang }} • C++${{ matrix.std }} • x64"
     container: "silkeh/clang:${{ matrix.clang }}"
@@ -310,6 +319,7 @@ jobs:
         -DPYBIND11_WERROR=ON
         -DDOWNLOAD_CATCH=ON
         -DCMAKE_CXX_STANDARD=${{ matrix.std }}
+        ${{ matrix.args }}
         -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)")
 
     - name: Build
--- a/include/pybind11/stl.h
+++ b/include/pybind11/stl.h
@@ -13,6 +13,7 @@
 #include "detail/common.h"
 
 #include <deque>
+#include <iostream>
 #include <list>
 #include <map>
 #include <ostream>
@@ -116,6 +117,10 @@ private:
     void reserve_maybe(const dict &, void *) {}
 
 public:
+#define DEBUG(value) std::cerr << __func__ << ": " << &value << " #: " << value.size() << "\n"
+
+    map_caster() { DEBUG(value); }
+    ~map_caster() { DEBUG(value); }
     bool load(handle src, bool convert) {
         if (!isinstance<dict>(src)) {
             return false;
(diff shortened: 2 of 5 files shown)


rhaschke:
The backtrace of the segfault reveals that the `std::map` passed to C++ is corrupted for some reason. As the behavior is very different across different compilers, I am suspecting a compiler bug. I'm not sure though. I wasn't able to come up with a minimal example just moving `std::map` around.

In the build with debug output (1f0677a), it is interesting to compare both clang `C++14` builds:
- 🐍 3 • Clang 5 • C++14 • x64 [succeeds, _not_ instantiating `map_caster`](https://github.com/pybind/pybind11/actions/runs/3667495765/jobs/6199999288#step:7:817).
- 🐍 3 • Clang 10 • C++14 • x64 [fails, instantiating `map_caster`](https://github.com/pybind/pybind11/actions/runs/3667495765/jobs/6199999452#step:7:818)

rhaschke:
I will work around this issue for now, by always ensuring some optimization options.
However, I just noticed that [all(?) gcc builds](https://github.com/pybind/pybind11/actions/runs/3667495765/jobs/6199997400#step:12:827) and [most Windows builds](https://github.com/pybind/pybind11/actions/runs/3667495765/jobs/6199997726#step:12:828) fail on that issue too when the map is actually used.

rwgk:
Hi Robert, you seem to have an ODR violation:

* PR #4095 (CI just started, but see PR description with error from local testing)
* PR #4022

rwgk:
I didn't realize this before: I'm afraid PR #4353 made it a lot more likely that ODR violations sneak in.

rwgk:
Actually, no, I take that back, after taking a fresh look at stl_bind.h ... and the error message in the #4396 PR description (showing the source locations).

rwgk:
The CI for #4396 is nearly complete. I'll close the PR soon, but the CI results will expire only later:

* https://github.com/pybind/pybind11/actions/runs/3667642607
* https://github.com/pybind/pybind11/actions/runs/3667642609

rwgk:
> Where does that come from? Isn't including `pybind11/stl.h` and defining a wrapper with a `std::map` argument a valid approach?

It depends on how you compile & link.

Generally, there are three lines of defense against ODR violations:
* [pybind11 namespace hidden](https://github.com/pybind/pybind11/blob/9727dcdae5b723a01ec04bec5eb1a5a681cbb0fd/include/pybind11/detail/common.h#L84-L93)
* `-fvisibility=hidden`
* `RTLD_LOCAL` (default under Linux, a given under Windows (no other option), I'm not sure about macOS)

**But**

if you link multiple **Translation Units** (critical keyword when it comes to ODR) into one extension, all those defenses do not exist, on all platforms.  —  Note that most pybind11 unit tests are linked together as pybind11_tests.so. (You can [see that more clearly here](https://github.com/rwgk/pybind11_scons/blob/9c537b4c4ead8435ca62b763887bc967e370a088/SConscript#L113-L130) (compared to the cmake files).)

Specific to `type_caster` ODR violations, the critical code to consider is:

https://github.com/pybind/pybind11/blob/9727dcdae5b723a01ec04bec5eb1a5a681cbb0fd/include/pybind11/cast.h#L39-L40

It needs to resolve to the exact same template instantiations in all TUs linked together. (There is zero wiggle room for this, as [we learned the hard way](https://github.com/pybind/pybind11/pull/3862#issuecomment-1126377796).)

With this PR, the ODR arises from

* test_stl.cpp  —  `SourceLocation1="/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/stl.h:161"`
* test_stl_binders.cpp 
 —  `SourceLocation2="/usr/local/google/home/rwgk/forked/pybind11/include/pybind11/detail/type_caster_base.h:913`

I.e. the `type_caster` specializations work out differently in the two TUs, and when those are linked together you get ODR violations.

I think of that has the "worst kind of trap set by pybind11". It really is not obvious.

One way to not have to worry about it so much: Set up your system with "one TU per extension". But that's not always feasible, then you have to think twice, and if things get really big (e.g. Google linking most things statically), test with the `type_caster_odr_guard` enabled (smart_holder branch only, but you can swap it in for testing without changing anything else).

rhaschke:
I'm fine to close this issue report. However, you might want to check the ODR warnings popping up in the current master branch: https://github.com/pybind/pybind11/actions/runs/3659994679/jobs/6186623695

rwgk:
> I'm fine to close this issue report. However, you might want to check the ODR warnings popping up in the current master branch: https://github.com/pybind/pybind11/actions/runs/3659994679/jobs/6186623695

Wow, thanks for pointing those out. Amazingly, I never noticed those before, even though I look at CI logs pretty regularly.

Continuing under PR #4398. I think it's a very different problem.

PR #4398, opened 2022-12-12
Systematically add `-DCMAKE_VERBOSE_MAKEFILE=ON` in ci.yml

rwgk:
## Description
@henryiii @Skylion007 I didn't really expect the added `-DCMAKE_VERBOSE_MAKEFILE=ON` to work everywhere, but it does!

The size difference in logs_*.zip is:

```
$ ls -l logs_25759.zip logs_25892.zip
3.9M Dec 12 02:47 logs_25759.zip
5.2M Dec 12 11:05 logs_25892.zip
```

I'm thinking, relative to all the binaries getting generated, that's probably a tiny increase, but it makes the difference between day a night for questions like we want to answer here. E.g. I see that `-Wodr` isn't actually specified explicitly on any command line, and I can now systematically mine the logs for clues what linker options (`-O3`, `-flto`?) trigger the ODR warnings.

I.e. I think it'll be great to merge this PR.

________

To obtain full command lines related to `-Wodr`.

https://github.com/pybind/pybind11/pull/4395#issuecomment-1346191424

https://github.com/pybind/pybind11/actions/runs/3659994679/jobs/6186623695

## Suggested changelog entry:

```rst

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
     - name: Configure C++11 ${{ matrix.args }}
       run: >
         cmake -S . -B .
+        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DPYBIND11_WERROR=ON
         -DPYBIND11_SIMPLE_GIL_MANAGEMENT=ON
         -DDOWNLOAD_CATCH=ON
@@ -135,6 +136,7 @@ jobs:
     - name: Configure C++17
       run: >
         cmake -S . -B build2
+        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DPYBIND11_WERROR=ON
         -DPYBIND11_SIMPLE_GIL_MANAGEMENT=OFF
         -DDOWNLOAD_CATCH=ON
@@ -157,6 +159,7 @@ jobs:
     - name: Configure (unstable ABI)
       run: >
         cmake -S . -B build3
+        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DPYBIND11_WERROR=ON
         -DDOWNLOAD_CATCH=ON
         -DDOWNLOAD_EIGEN=ON
@@ -243,6 +246,7 @@ jobs:
         SETUPTOOLS_USE_DISTUTILS: stdlib
       run: >
         cmake -S . -B build
+        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DCMAKE_BUILD_TYPE=Debug
         -DPYBIND11_WERROR=ON
         -DDOWNLOAD_CATCH=ON
@@ -307,6 +311,7 @@ jobs:
       shell: bash
       run: >
         cmake -S . -B build
+        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DPYBIND11_WERROR=ON
         -DDOWNLOAD_CATCH=ON
         -DCMAKE_CXX_STANDARD=${{ matrix.std }}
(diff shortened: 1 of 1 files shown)


rwgk:
The ODR warnings (link in PR description) seem to originate from

eigen3/unsupported/Eigen/CXX11/src/Tensor/TensorStorage.h, which was brought in via PR #4201.

@Lalaland have you seen those before?

A quick Google search for "Eigen TensorStorage.h Wodr" doesn't pull up anything:

https://www.google.com/search?sxsrf=ALiCzsbf1_qwrjfCNnNTRDHsSFW41iRVjQ:1670844029730&q=Eigen+TensorStorage.h+%22Wodr%22&sa=X&ved=2ahUKEwj1yPeG-_P7AhUFmeAKHTarBKsQ5t4CegQIKBAB&biw=2099&bih=1042&dpr=2

Another thing I don't understand: I see `-DPYBIND11_WERROR=ON` in log files that have the warning, but the `-Wodr` warnings are not treated as errors anyway.

I created this PR to find out what's going on.

Lalaland:
My initial guess is that it's due to our weird unit tests (where we have eigen with and without std::array).

Some of those symbols must be accidentally being made public, causing odr violations.

Easiest test would be to disable one of the two (disable the no std::array) test

rwgk:
> My initial guess is that it's due to our weird unit tests (where we have eigen with and without std::array).

Oh, of course, I believe it immediately!
Two definitions, in test_eigen_tensor.cpp & test_eigen_tensor_avoid_stl_array.cpp, but linked together into pybind11_tests.so, ouch. I completely missed that before, but it's very obvious in hindsight.

(I'm intrigued now btw that some compilers/linkers discover that. Before I implemented the `type_caster_odr_guard` I looked around pretty extensively, hoping to find an existing tool that does it.)

> Some of those symbols must be accidentally being made public, causing odr violations.

Everything linked together into pybind11_tests.so has mutual visibility. The only mechanism for hiding symbols from each other inside the same .so is to put them into different namespaces, incl. the unnamed namespace. (Or maybe also C-style `static` entities, not sure, but I don't think that's relevant here.)

> Easiest test would be to disable one of the two (disable the no std::array) test

Another option is to put test_eigen_tensor_avoid_stl_array into a separate extension, then the hiding will work. The only problem is that tests/CMakeLists.txt will become a little more complicated, but there are already similar cases:

* master: https://github.com/pybind/pybind11/blob/9727dcdae5b723a01ec04bec5eb1a5a681cbb0fd/tests/CMakeLists.txt#L219-L221

* smart_holder: https://github.com/pybind/pybind11/blob/53b80b4312d60375bbde55f8d587aec53ae7f02f/tests/CMakeLists.txt#L243-L246

Luckily PR #3590 by @iwanders has made this pretty easy, thanks again!

@Lalaland is there a chance you could help with this?

I think the end-result will be a double win:
* We're resolving the existing ODR.
* If we work on the visibility handling in the future (e.g. #4072) we know some compilers will tell us if we make a mistake, by generating the ODR warnings for TensorStorage.h again.

Lalaland:
> @Lalaland is there a chance you could help with this?

Yeah, I introduced the bug, so I should probably introduce a fix. I'll be able to work on it this weekend.

Skylion007 (approved):

rwgk:
Thanks Aaron!

henryiii:
Instead of changing every single invocation and making an already long file longer, why not just set `env: VERBOSE: 1` at the top? It's been supported since CMake 3.14: https://cmake.org/cmake/help/latest/envvar/VERBOSE.html

rwgk:
> Instead of changing every single invocation and making an already long file longer, why not just set `env: VERBOSE: 1` at the top? It's been supported since CMake 3.14: https://cmake.org/cmake/help/latest/envvar/VERBOSE.html

Oh, nice! I didn't know. I'll try that out.

PR #4401, opened 2022-12-12
Fix multi-context new Python linking mode

mablanchard:
## Description

Allow CMake find_package() from multiple directories.

Fixes https://github.com/pybind/pybind11/issues/4400

## Suggested changelog entry:

```rst
* Allow calling ``find_package(pybind11 CONFIG)`` multiple times from separate
  directories in the same CMake project and properly link Python (new mode).
```


henryiii (approved):
Seems fine. Have you verified it does work in this way?

mablanchard:
> Seems fine. Have you verified it does work in this way?

It seems to be working for me!

PR #4405, opened 2022-12-13
Set `env: VERBOSE: 1` as suggested by @henryiii

rwgk:
## Description
Improve on #4398 in two ways:

1. Set `env: VERBOSE: 1` in ci.yml as suggested by @henryiii.
2. Also apply this to all other .yml files using cmake.

To make it easy to see what the net changes are, compared to the state **before** #4398:

* commit 5136cb502a745428fa7b5fcfbbcf1853de7792ec rolls back `#4398`.
* commit 6abc64271ef66d5159115476a38a3476fbea95b4 brings back the functional equivalent of `#4398`, but instead of adding many `-DCMAKE_VERBOSE_MAKEFILE=ON` it removes a few.
* commit b939e13013c62d0f3087f49137fbd6c0d7914f07 applies the `env: VERBOSE: 1` to all other .yml files using cmake.

The CI logs were inspected manually to verify that this PR is working as intended.

## Suggested changelog entry:

```rst

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,6 @@ jobs:
     - name: Configure C++11 ${{ matrix.args }}
       run: >
         cmake -S . -B .
-        -DCMAKE_VERBOSE_MAKEFILE=ON
         -DPYBIND11_WERROR=ON
         -DPYBIND11_SIMPLE_GIL_MANAGEMENT=ON
         -DDOWNLOAD_CATCH=ON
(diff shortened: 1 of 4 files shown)


henryiii (approved):
I'm not sure I like increasing the verbosity - while it does help occasionally for certain things, most of the time it just adds a ton of irrelevant logging lines making checking the output for compiler errors and warnings harder to see. 95% of the time you aren't checking to see what's passed to the compiler. But this change over what's already been merged is a clear positive, IMO.

rwgk:
> 95% of the time you aren't checking to see what's passed to the compiler.

I would put that number even higher >99%

But for those <1%, I often found it impractical to get the insights I wanted and resorted to guessing. —  Impractical: e.g. looking back in the logs from last week, 10 or so commits ago: ouch, the information isn't there. It can be a real ordeal to backtrack to run the CI again for that state.

What guides my thinking here:
* Filtering is easy and conclusive.
* Guessing is hard and inconclusive.

PR #4406, opened 2022-12-14
fix(cmake): support Windows ARM cross-compilation

henryiii:
## Description

Currently, even if you manually set the libraries variable to an ARM library, this still gets the Intel library when cross-compiling on Windows.

## Suggested changelog entry:

```rst
* fix(cmake): support setting PYTHON_LIBRARIES manually for Windows ARM cross-compilation
```


henryiii:
Just shipped Windows ARM cross-compile support in scikit-build-core 0.1.4 and scikit-build 0.16.4. This patch would be great to get in for projects using pybind11, like  https://github.com/pybind/scikit_build_example/pull/65.

Skylion007 (approved):

